PR #8164, opened 2019-12-20
Update support for RHEL 8

hakman:
RHEL/CentOS 8 support is broken for various reasons:
1. check for Debian family doesn't recognise the distro
2. `iptables-setup` cannot be run from home dir (probably some extra security)
3. `python` package needs to be more explicit (`python2` or `python3`)
4. `ntpd` fails to install because it has been replaced with `chronny`

### DIFF
--- a/nodeup/pkg/model/firewall.go
+++ b/nodeup/pkg/model/firewall.go
@@ -46,7 +46,7 @@ func (b *FirewallBuilder) buildSystemdService() *nodetasks.Service {
 	manifest.Set("Unit", "Before", "network.target")
 	manifest.Set("Service", "Type", "oneshot")
 	manifest.Set("Service", "RemainAfterExit", "yes")
-	manifest.Set("Service", "ExecStart", "/home/kubernetes/bin/iptables-setup")
+	manifest.Set("Service", "ExecStart", "/opt/kops/helpers/iptables-setup")
 	manifest.Set("Install", "WantedBy", "basic.target")
 
 	manifestString := manifest.Render()
@@ -87,7 +87,7 @@ iptables -A FORWARD -w -p ICMP -j ACCEPT
 fi
 `
 	return &nodetasks.File{
-		Path:     "/home/kubernetes/bin/iptables-setup",
+		Path:     "/opt/kops/helpers/iptables-setup",
 		Contents: fi.NewStringResource(script),
 		Type:     nodetasks.FileType_File,
 		Mode:     s("0755"),
--- a/nodeup/pkg/model/miscutils.go
+++ b/nodeup/pkg/model/miscutils.go
@@ -57,7 +57,7 @@ func (b *MiscUtilsBuilder) Build(c *fi.ModelBuilderContext) error {
 		packages = append(packages, "apt-transport-https")
 	} else if b.Distribution.IsRHELFamily() {
 		packages = append(packages, "curl")
-		packages = append(packages, "python")
+		packages = append(packages, "python2")
 		packages = append(packages, "git")
 	} else {
 		klog.Warningf("unknown distribution, skipping misc utils install: %v", b.Distribution)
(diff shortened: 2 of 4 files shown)


rifelpet:
This is great! do you think we could add an e2e job for this? It should be as easy as copying the centos7 one and changing the ami id that it uses.
/ok-to-test

hakman:
@rifelpet was thinking the same thing about the e2e job.
The only catch here would be that somehow we need to enable `calico.iptablesBackend: NFT`. Wanted to ask you what would be the best place to add it.

rifelpet:
For e2e tests the best we can do is kops' `--set` functionality to override specific fields in manifests that arent otherwise accessible by cli flags in `kops create cluster`.

See these here:

https://github.com/kubernetes/kops/blob/4621df3c0a6e924458ca2dac18c6d3b15052b60f/cmd/kops/create_cluster.go#L367-L369

each field needs to explicitly get added to this string->field mapping function:

https://github.com/kubernetes/kops/blob/4621df3c0a6e924458ca2dac18c6d3b15052b60f/pkg/commands/set_cluster.go#L75

and we'd set it in a test with this `--kops-overrides` flag: https://github.com/kubernetes/test-infra/blob/4beb96650aaee54c274a42fb0ca011333808b6f5/kubetest/kops.go#L74

Though if that field is required for centos8 to work, should kops just automatically set it somehow? I know that can be tricky to make an addon manifest's contents depend on an AMI (of which there can be many in a given cluster). maybe its worth discussing at office hours today.

hakman:
@rifelpet good news. Not sure why I thought tests would run with Calico. Just did a quick test with Kubenet (default) and seems to work without any issue. So the test can be added without any extra changes.

nodeup/pkg/model/miscutils.go
```diff
@@ -57,7 +57,7 @@ func (b *MiscUtilsBuilder) Build(c *fi.ModelBuilderContext) error {
 		packages = append(packages, "apt-transport-https")
 	} else if b.Distribution.IsRHELFamily() {
 		packages = append(packages, "curl")
-		packages = append(packages, "python")
+		packages = append(packages, "python2")
```
justinsb: Is this not `python` for Centos7?  Or is python an alias for python2?

nodeup/pkg/model/firewall.go
```diff
@@ -87,7 +87,7 @@ iptables -A FORWARD -w -p ICMP -j ACCEPT
 fi
 `
 	return &nodetasks.File{
-		Path:     "/home/kubernetes/bin/iptables-setup",
+		Path:     "/opt/kops/helpers/iptables-setup",
```
justinsb: We might have to create this directory explicitly, but I think it's a good idea to pick a more reasonable directory.

For immutable OSes, we might have to bind-mount these also, if we're seeing failures there.

I think we could also do `/opt/kops/bin/iptables-setup`

justinsb:
I guess we should get this in, and then double check if we broke centos7 and other immutable OSes like COS.

/approve
/lgtm

nodeup/pkg/model/miscutils.go
hakman: On CenOS 7 python2 is an alias for python.

nodeup/pkg/model/firewall.go
hakman: I chose the "helpers" dir because that was the dir chosen for the Docker healthcheck.
Will give this a try with CoreOS also when this is merged and check again.
Changing to `/opt/kops/bin/iptables-setup` sounds good to me.

hakman:
Thanks @justinsb. This will make testing simpler for the tar.gz package for containerd.
Once the https://github.com/kubernetes/kops/pull/8172 gets checked we can enable the e2e test for containerd + rhel8.

PR #8168, opened 2019-12-20
added k8 audit logs to logrotate when enabled

simonmacklin:
small commit just to add the api audit logs to log rotate when its enabled

### DIFF
--- a/nodeup/pkg/model/logrotate.go
+++ b/nodeup/pkg/model/logrotate.go
@@ -71,6 +71,10 @@ func (b *LogrotateBuilder) Build(c *fi.ModelBuilderContext) error {
 	b.addLogRotate(c, "etcd", "/var/log/etcd.log", logRotateOptions{})
 	b.addLogRotate(c, "etcd-events", "/var/log/etcd-events.log", logRotateOptions{})
 
+	if b.Cluster.Spec.KubeAPIServer.AuditLogPath != nil {
+		b.addLogRotate(c, "kube-audit", *b.Cluster.Spec.KubeAPIServer.AuditLogPath, logRotateOptions{})
+	}
+
 	if err := b.addLogrotateService(c); err != nil {
 		return err
 	}


rifelpet:
Looks like the CLA check passed!

/ok-to-test

simonmacklin:
Thanks @rifelpet 

Looks like one of the test failed.  Looks like it may be a issue with the kops part and not the nodeup though.

justinsb:
I assume that without this the audit logs just build up forever (until we rotate the masters?) - is that right @simonmacklin ?

If so, I think this is a good idea, and we probably shouldn't even offer an option to turn off logrotation, though we probably should have a release note depending on how people are capturing these logs.  (I don't think we have yet started the kops 1.18 relnotes so this could be the first)

One other gotcha is that I don't think that all OSes support logrotate, and I'm wondering if we should do something for them (but definitely not in this PR, just that you might have insight into how important it is @simonmacklin ?)

simonmacklin:
Hey @justinsb 

I have noticed the logs building up on my masters even though I have the settings applied below to the api server through the kops cluster spec.  I have also validated the settings are present on the api server yaml config.

--audit-log-maxage **_defined the maximum number of days to retain old audit log files_**
--audit-log-maxbackup **_defines the maximum number of audit log files to retain_**
--audit-log-maxsize **_defines the maximum size in megabytes of the audit log file before it gets_** rotated

https://kubernetes.io/docs/tasks/debug-application-cluster/audit/

I am looking at the code for the audit logs which I think is hear 
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/options/audit.go 

and trying to figure why this is not working on the clusters I look after.  Because if I am reading the doc right we may not need to add audit logs to logrotate.  This would also solve the issue of OSes not supporting log rotate. 

Simon

simonmacklin:
Hi @justinsb 

I had a look at this yesterday and figured out why the api server was not rotating the logs right.  The API pod was not mounting the the directory where the logs was being written too.  

Kops already makes sure the right host path is mounted which is located https://github.com/kubernetes/kops/blob/master/nodeup/pkg/model/kube_apiserver.go#L492

Just for some reason it was missing on my clusters.  On this bases I don't think we need to add the audit logs to log rotate.  If you agree I can close this PR off.?

Many thanks 

Simon

rifelpet:
I agree, that is what should be happening so I think we can close this out. If you see an issue with Kops not mounting the expected host paths for the kube-apiserver pod please do open an issue.

/close

PR #8169, opened 2019-12-21
Run dns-controller and kops-controller as non-root user

johngmyers:
They do not need root privileges, so should not be given them.

### DIFF
--- a/images/dns-controller/Dockerfile
+++ b/images/dns-controller/Dockerfile
@@ -21,4 +21,6 @@ RUN apt-get update && apt-get install --yes ca-certificates \
 
 COPY /.build/artifacts/dns-controller /usr/bin/dns-controller
 
+USER 1000
+
 CMD /usr/bin/dns-controller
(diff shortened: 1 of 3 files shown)


images/dns-controller/Dockerfile
```diff
@@ -21,4 +21,6 @@ RUN apt-get update && apt-get install --yes ca-certificates \
 
 COPY /.build/artifacts/dns-controller /usr/bin/dns-controller
 
+USER 1000
```
joshbranham: For my own understanding, do we need to also set the user in the `securityContext` of the `dns-controller` here: https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/simple/dns-controller.addons.k8s.io-k8s-1.12.yaml

johngmyers: We will, but we need someone to release new images built from this `Dockerfile` first. So that's a different PR.

PR #8170, opened 2019-12-21
containerd: Use containerd 1.2.4 with Docker 18.09.3

hakman:
This PR is related to #7986:
1. adds the `containerd-1.2.4` package for a previously released Docker version
2. sets the default version only for Kubernetes versions >= 1.18
3. warns that Kubernetes versions >= 1.11 and < 1.18 are untested
4. exits with error when Kubernetes versions < 1.11

### DIFF
--- a/pkg/model/components/containerd.go
+++ b/pkg/model/components/containerd.go
@@ -19,6 +19,7 @@ package components
 import (
 	"fmt"
 
+	"k8s.io/klog"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/loader"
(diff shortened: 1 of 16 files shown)


pkg/model/components/containerd.go
```diff
@@ -41,64 +42,58 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
...
 
-			sv, err := KubernetesVersion(clusterSpec)
-			if err != nil {
-				return fmt.Errorf("unable to determine kubernetes version from %q", clusterSpec.KubernetesVersion)
+		// Set containerd based on Kubernetes version
+		if fi.StringValue(containerd.Version) == "" {
+			if b.IsKubernetesGTE("1.18") {
+				containerd.Version = fi.String("1.2.10")
+			} else if b.IsKubernetesGTE("1.11") {
+				return fmt.Errorf("contained version is required")
```
johngmyers: Typo:
```suggestion
				return fmt.Errorf("containerd version is required")
```

pkg/model/components/docker.go
```diff
@@ -50,40 +45,33 @@ func (b *DockerOptionsBuilder) BuildOptions(o interface{}) error {
...
+			docker.Version = fi.String("17.03.2")
+		} else if b.IsKubernetesGTE("1.8") {
+			docker.Version = fi.String("1.13.1")
+		} else if b.IsKubernetesGTE("1.6") {
+			docker.Version = fi.String("1.12.6")
+		} else if b.IsKubernetesGTE("1.5") {
+			docker.Version = fi.String("1.12.3")
+		} else if b.IsKubernetesGTE("1.4") {
+			docker.Version = fi.String("1.11.2")
+		} else {
```
johngmyers: This is a semantic change. Previously it would pick 1.11.2.

hakman: True. I think at that time I very tempted to remove the defaults these completely as they will be removed in 1.18.
I guess will do it in a future commit. Thanks,

pkg/model/components/containerd.go
hakman: Nice catch!

pkg/model/components/containerd.go
```diff
@@ -41,64 +42,58 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
...
-			if clusterSpec.KubernetesVersion == "" {
-				return fmt.Errorf("Kubernetes version is required")
-			}
+	if clusterSpec.ContainerRuntime == "containerd" {
+		if b.IsKubernetesLT("1.11") {
+			// Containerd 1.2 is validated against Kubernetes v1.11+
+			// https://github.com/containerd/containerd/blob/master/releases/v1.2.0.toml#L34
+			return fmt.Errorf("kubernetes %s is not compatible with containerd", clusterSpec.KubernetesVersion)
+		} else if b.IsKubernetesLT("1.18") {
+			klog.Warningf("kubernetes %s is untested with containerd", clusterSpec.KubernetesVersion)
```
justinsb: This might be a little strong, in that it's the combination kops + kubernetes that we haven't tested before 1.18, but we can let it stand :-)

justinsb:
Thanks @hakman - I appreciate your doing this additional work :-)

/approve
/lgtm

PR #8172, opened 2019-12-21
containerd: Add --container-runtime cli flag

hakman:
This PR is related to #7986.


justinsb:
Thanks @hakman - looks great!  I agree we should make the value explicit in the cluster, as we transition people from docker -> containerd.  (It's also best practice for union fields: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190325-unions.md#discriminator )

/approve
/lgtm
/lgtm

PR #8173, opened 2019-12-21
Update kubernetes versions used in integration tests

rifelpet:
In preparation for deprecating support for older Kubernetes versions, this updates our integration tests to use supported versions.

The only fields that change in the output are `kubernetesVersion`, `image`, and `kubelet.anonymousAuth`

### DIFF
--- a/tests/integration/create_cluster/complex/expected-v1alpha2.yaml
+++ b/tests/integration/create_cluster/complex/expected-v1alpha2.yaml
@@ -27,9 +27,11 @@ spec:
   iam:
     allowContainerRegistry: true
     legacy: false
+  kubelet:
+    anonymousAuth: false
   kubernetesApiAccess:
   - 0.0.0.0/0
-  kubernetesVersion: v1.7.1
+  kubernetesVersion: v1.15.6
   masterPublicName: api.complex.example.com
   networkCIDR: 172.20.0.0/16
   networking:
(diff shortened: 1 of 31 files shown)


joshbranham:
nice!

/lgtm

hakman:
@rifelpet @johngmyers Not sure if you already tried, but would be nice to remove the default versions of Docker for obsolete Kubernetes versions and fix the tests after that. It flags quite well tests that use very old Kubernetes versions.
https://github.com/kubernetes/kops/blob/1.17.0-alpha.1/pkg/model/components/docker.go#L64-L71

johngmyers:
My understanding is that we're waiting until deprecation notices are backported to 1.16 and 1.17 before removing code for obsolete Kubernetes versions in master.

hakman:
My idea was not to remove it yet, but to flag and change tests easier.
Though, removing the defaults would only make people more aware. They could still use the desired version of Docker, but would have to do it manually.

rifelpet:
I brought up the docker versions during the call on friday. Justin said theres not as much benefit to stripping it out compared to other areas of the code base that specifically handle older k8s versions, but removing the packages and/or the k8s version -> docker version logic could certainly be done. would it simplify your containerd work at all?

And yes, my plan was to get the deprecation announcement backported to 1.16 and then open a PR against master that removes the bulk of the old version support.

hakman:
In my case with containerd, there's not much to gain anymore. Was just thinking that removing the defaults would make people more inclined towards upgrading. Thanks.

PR #8175, opened 2019-12-21
[WIP] containerd: Check e2e tests

hakman:
This PR is related to #7986.

e2e tests need to be checked with the new `containerd` runtime. This is the simplest way to trigger the job and test any fixes.

Results:
* Passed - **696**
* Failed - 3
* Skipped - 4155

Conclusions:
* containerd doesn't work out of the box with `--networking=kubenet`- https://github.com/containerd/cri/blob/master/docs/config.md#cni-config-template
* containerd doesn't work with the default image because it is missing the `dbus` package, needed by Protokube to mount `/var/run/dbus`
* containerd doesn't work with the default image because Docker is already running there
* Docker health checks should be disabled (minor issue)
* some services still have references to Docker (`After=docker.service`) (minor issue)

@justinsb any thoughts about the conclusions and in particular about the default Kops images?


johngmyers:
Looks like a bunch more tests need to use more up-to-date Kubernetes versions.

hakman:
Thanks @johngmyers. I don't plan to make this a default in master. This was the easiest way to see how e2e tests are working with containerd.
I guess will have to wait for someone that can give the `ok-to-test`.

johngmyers:
This is failing Travis because `TestBootstrapChannelBuilder_BuildTasks` cannot calculate the containerd version for Kubernetes 1.4.6, among other reasons. This is a useful signal, that more tests should be updated to use more recent versions of Kubernetes.

Perhaps e2e will run with unit tests failing.

hakman:
Good point. Let me give it a try.

johngmyers:
I have a branch updating all the 1.4.6 in tests to 1.15.6. It passes Travis, but I need to look closer into the tests to see if that change invalidates them.

hakman:
Some tests are tricky. May require some extra attention. It is quite a big push to move 10 versions.

rifelpet:
@hakman is this safe to close now that we have cli support and the periodic job?

hakman:
e2e tests for containerd are not running periodically using the official RHEL 8 image and Calico network plugin:
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-containerd

Most issues mentioned here are fixed or have issues created for them:
* containerd doesn't work out of the box with - TBD (no plan to fix at the moment)
* containerd doesn't work with the default image because it is missing the `dbus` package - https://github.com/kubernetes-sigs/image-builder/issues/130
* Docker is already running in the default image - https://github.com/kubernetes-sigs/image-builder/issues/128
* Docker health checks should be disabled - https://github.com/kubernetes/kops/pull/8221
* some services still have references to Docker - https://github.com/kubernetes/kops/pull/8217

PR #8176, opened 2019-12-21
Add deprecation warning for older k8s versions

rifelpet:
As discussed during yesterday's office hours.

This should be cherry-picked back to 1.16.

We'll need to bump the minor version in the 1.17 and master branches. Once we officially drop support we could make this a sliding window that uses the `kops.Version` variable.

I think we could have deprecation warnings for the 2 oldest versions still supported by the kops version, announcing that the n+2 Kops version will no longer support the specified k8s version.

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -91,6 +91,8 @@ var (
 	AlphaAllowALI = featureflag.New("AlphaAllowALI", featureflag.Bool(false))
 	// CloudupModels a list of supported models
 	CloudupModels = []string{"proto", "cloudup"}
+	// OldestSupportedKubernetesVersion is the oldest kubernetes version that is supported in Kops
+	OldestSupportedKubernetesVersion = "1.9.0"
 )
 
 type ApplyClusterCmd struct {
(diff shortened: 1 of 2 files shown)


upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -91,6 +91,8 @@ var (
 	AlphaAllowALI = featureflag.New("AlphaAllowALI", featureflag.Bool(false))
 	// CloudupModels a list of supported models
 	CloudupModels = []string{"proto", "cloudup"}
+	// OldestSupportedKubernetesVersion is the oldest kubernetes version that is supported in Kops
+	OldestSupportedKubernetesVersion = "1.9.0"
```
johngmyers: When we remove code for handling old versions, we should add another cutoff below which kops should refuse to run. That should probably be what is named `OldestSupportedKubernetesVersion`. This should probably be `OldestRecommendedKubernetesVersion` or `NewestDeprecatedKubernetesVersion = 1.8.15`

I could also see a third cutoff `OldestUpgradeableKubernetesVersion` below which kops won't upgrade from.

justinsb: These are great suggestions ... we can probably do it separately?

We can recommend newer versions of k8s (and kops) in `channels/stable` ... with varying degrees of forcefulness :-)

justinsb:
I'd like to see us get this in to give people plenty of heads-up.  I do think that e.g. we need to be even more forceful about saying that 1.8 is a really bad idea, but ... as discussed we have to start somewhere!

/approve
/lgtm

PR #8177, opened 2019-12-22
Fix URL to Creative Commons license

LinshanYu:
The original address is unreachable, so I modify it to be truly accessible


johngmyers:
Thanks for the contribution! A better commit message would be "Fix URL to Creative Commons license".

LinshanYu:
> Thanks for the contribution! A better commit message would be "Fix URL to Creative Commons license".

ok,ths

joshbranham:
@LinshanYu I think you need a rebase against `kops:master`

LinshanYu:
> @LinshanYu I think you need a rebase against `kops:master`

I had rebased the pr

justinsb:
Looks like something went wrong on a rebase?

I think you want just the last commit, in which case the easy way out is to do this:

```
git reset --hard origin/master
git cherry-pick 1aa309688
# verify it's what you want
git push --force myfork
```

This is in the node_modules, so I'm not certain whether this will get overwritten.  (I've been meaning to move generation to https://github.com/ahmetb/gen-crd-api-reference-docs , but that would be a separate PR)

rifelpet:
It turns out these docs were very outdated and we now refer to the API docs at [godoc.org](https://godoc.org/k8s.io/kops/pkg/apis/kops). I ended up deleting this directory in https://github.com/kubernetes/kops/pull/7824 so I don't think we need this PR anymore.
/close

PR #8179, opened 2019-12-22
Add all flag to export cluster command

### LINKED ISSUE #140: kops export kubecfg should be able to export all clusters in one command
It would be great to have a single command that could export all the readable clusters, rather than having to pass them in one-by-one.

vvbogdanov87:
Closes #140

### DIFF
--- a/cmd/kops/export_kubecfg.go
+++ b/cmd/kops/export_kubecfg.go
@@ -20,8 +20,10 @@ import (
 	"io"
 
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/kops/cmd/kops/util"
+	api "k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/commands"
 	"k8s.io/kops/pkg/kubeconfig"
 	"k8s.io/kops/upup/pkg/fi"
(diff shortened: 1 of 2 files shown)


cmd/kops/export_kubecfg.go
```diff
@@ -67,42 +70,60 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
...
-	keyStore, err := clientset.KeyStore(cluster)
-	if err != nil {
-		return err
-	}
-
-	secretStore, err := clientset.SecretStore(cluster)
-	if err != nil {
-		return err
+	var clusterList []*api.Cluster
+	if options.all {
```
johngmyers: Perhaps fail if `len(args) != 0`?

cmd/kops/export_kubecfg.go
```diff
@@ -67,42 +71,63 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
...
+		if len(args) != 0 {
+			return fmt.Errorf("Cannot use both --args flag and positional arguments")
```
justinsb: Did you mean "..both --all flag and positional arguments" ?

justinsb:
I'm going to use this a lot - thanks @vvbogdanov87  :-)

I had one question about whether you wrote "--args" in the error message when you meant "--all", but we can easily fix that up separately if so.

/approve
/lgtm

PR #8180, opened 2019-12-22
Refactor: Add Region() method to fi.Cloud

justinsb:
This enables us to simply the apply logic.


rifelpet:
FWIW the vsphere cluster api uses tags to track regions, but since Kops' support does not do that yet I think leaving its value as the cluster name is fine.

https://github.com/kubernetes/cloud-provider-vsphere/blob/f11e58a0f15f435991d48addcf578568950dde0e/pkg/common/connectionmanager/zones.go#L45-L46

/lgtm

PR #8182, opened 2019-12-22
Fix verify-staticcheck prow job

rifelpet:
The new job was missing the generated files from go-bindata. I tested this locally on a clean checkout.

I noticed some of the prow jobs use make and others call hack/ scripts directly.

I think it'd be better to use make so we can more easily add dependencies like this, therefor I'd like to propose that we merge this, update the job to call `make verify-staticcheck` then remove the make command from hack/verify-staticcheck.sh

By fixing the existing job as well as the make target, we can unblock other PRs and not get another swarm of failure notifications when the job gets updated to use make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -553,7 +553,7 @@ verify-bazel:
 	hack/verify-bazel.sh
 
 .PHONY: verify-staticcheck
-verify-staticcheck:
+verify-staticcheck: ${BINDATA_TARGETS}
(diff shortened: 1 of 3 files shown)


hakman:
Thanks @rifelpet

joshbranham:
Well that explains all the emails I got :)

/lgtm

Issue #8091, opened 2019-12-11
single to multi-master docs suggest you can add etcdCluster memebers but Kops 1.16.0-alpha.2 says no


rayh:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Version 1.16.0-alpha.2 (git-dfaf0b633)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:42:50Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.6", GitCommit:"96fac5cd13a5dc064f7d9f4f23030a6aeface6cc", GitTreeState:"clean", BuildDate:"2019-08-19T11:05:16Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
changed cluster config from:
```
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-ap-southeast-2c
      name: c
    name: main
  - etcdMembers:
    - instanceGroup: master-ap-southeast-2c
      name: c
    name: events
```
to 
```
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-ap-southeast-2a
      name: a
    - instanceGroup: master-ap-southeast-2b
      name: b
    - instanceGroup: master-ap-southeast-2c
      name: c
    name: main
  - etcdMembers:
    - instanceGroup: master-ap-southeast-2a
      name: a
    - instanceGroup: master-ap-southeast-2b
      name: b
    - instanceGroup: master-ap-southeast-2c
      name: c
    name: events
```

and then ran:

`kops replace -f kops/all.yaml -v 9`

**5. What happened after the commands executed?**
```
[...]

PR #8183, opened 2019-12-23
Disable check for adding etcd members to the cluster

mccare:
Currently the cluster validation will disallow adding etcd members. (See #8091) IMHO Since etcd-manager supports adding members to the cluster, the check is not needed any more.

I've tested to go from a one master cluster to a three master cluster and it worked. It is probably still a risky operation. 

Added a section in the docs to test for a read/write into etcd to see if it works as expected. Are there other health checks that I could document to see if the etcd cluster started successfully? 

Or should this be gated with a feature flag?

### DIFF
--- a/docs/single-to-multi-master.md
+++ b/docs/single-to-multi-master.md
@@ -196,6 +196,8 @@ etcdClusters:
     name: events
 ```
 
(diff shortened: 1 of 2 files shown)


michaelajr:
@mccare Why was this branch deleted? Was hoping to move our dev clusters to multi-master without recreating.

PR #8185, opened 2019-12-23
fix url

LinshanYu:
fix url


joshbranham:
actually @LinshanYu these look to be vendored in regards to our netlify build. I don't think we should update the markdown one-off in PRs like this.

LinshanYu:
> actually @LinshanYu these look to be vendored in regards to our netlify build. I don't think we should update the markdown one-off in PRs like this.
ok,ths.I'll colse the pr.

Issue #7800, opened 2019-10-16
[Help] Clean up  static check failures 


tanjunchen:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
There are many packages that can't pass the static check and we need to clean them up even though hack/verify-staticcheck.sh exsits in the kops.
Step 1: Remove some packages(related packages) from the list in https://github.com/kubernetes/kops/blob/master/hack/.staticcheck_failures , 
Step 2: Run hack/verify-staticcheck.sh and look at outputs from console
Step 3: Fix them to ensure that  hack/verify-staticcheck.sh runs cleanly
Step 4: Run hack/verify-bazel.sh  and hack/verify-gofmt.sh  to make sure that you fix any problems with bazel or gofmt.
Step 5: How to make prs?
* For existing contributor,make sure there lists are up to date.
* advice title begins with staticchek: 
* Don't use Fixes: #7800 which will shut down the issue,just use Ref:#7800
* example for description
```
https://github.com/kubernetes/kops/pull/8203#issue-356845565
```
Attention: if you have any questions
read this https://github.com/kubernetes/kops/blob/master/docs/development/vendoring.md

rendhalver:
I am interested in getting started helping out with kops.
How much of this is left to do?

rifelpet:
Thanks for volunteering! Most of the staticcheck issues have been addressed but you can certainly check the remaining issues in hack/.staticcheck_failures for the remaining ones, besides the PR listed above that is currently open.

I'm adding an additional check in #8328 similar to staticcheck but for shell scripts: [shellcheck](https://github.com/koalaman/shellcheck). Feel free to check it out and start fixing some of the reported issues.

rendhalver:
Awesome. Thanks for the pointer @rifelpet!
I will see what I can work out.

rendhalver:
Hrm. There's only one entry in hack/.staticcheck_failures and it looks like #7839 fixes at least some of those.
I will see I can what I can do to help out with the shellcheck fixes. :)

rifelpet:
all the staticcheck failures have been fixed. good job everyone
/close

PR #2328, opened 2017-04-08
Move cluster delete files to pkg/resources

andrewsykim:
I'd like to generalize the cluster delete code so it can be pluggable with other cloud providers. Before doing that I wanted to move it to a package that made more sense. Related issue: https://github.com/kubernetes/kops/issues/2310. 

This PR only moves the contents of the file, you'll notice that `upup/pkg/kutil/delete_cluster.go` was not deleted since some code in `convert_kubeup_cluster.go` depends on it. I'm not sure when kubeup conversion to kops can be deprecated but seems like a possibility in the near future. Until then some of the code in `upup/pkg/kutil/delete_cluster.go` is duplicated in `pkg/resources` 😭 

There will be future PRs coming to refactor/clean up `pkg/resources`.

cc @kris-nova

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kops/2328)

PR #8188, opened 2019-12-23
pkg/resources/aws:simplify code and remove code

Aresforchina:
ref:#7800


rifelpet:
It looks like this function was never used even [when it was added](https://github.com/kubernetes/kops/pull/2328/files) so it should be safe to delete.

/lgtm
/approve

PR #8189, opened 2019-12-23
Remove make command from verify-staticcheck.sh

rifelpet:
Followup to https://github.com/kubernetes/kops/pull/8182 and https://github.com/kubernetes/test-infra/pull/15674

This script should now be called using `make verify-staticcheck` which will handle the go-bindata dependency.

Also using this PR to verify the changed prow jobs from the test-infra PR above.


tanjunchen:
good

PR #8190, opened 2019-12-24
Update links printed by Kops to use new docs site

rifelpet:
This updates all references to `https://github.com/kubernetes/kops/blob/master/docs` in the source code to instead use the new docs site.


joshbranham:
Nice! All about pushing people to the docs site!

/lgtm

PR #8191, opened 2019-12-24
fix staticcheck failures in pkg/resources

Aresforchina:
ref:#7800
Special notes:
pkg/resources/ali/ali.go:63:3: this value of region is never used (SA4006)
pkg/resources/ali/ali.go:427:27: this result of append is never used, except maybe in other appends (SA4010)
pkg/sshcredentials/fingerprint.go:57:6: func insertFingerprintColons is unused (U1000)
pkg/sshcredentials/fingerprint.go:25:2: imported and not used: "k8s.io/klog"
pkg/resources/ali/ali.go:427:13: append(roleToDelete, role.RoleId) evaluated but not used
pkg/resources/ali/ali.go:415:2: roleToDelete declared and not used

### DIFF
--- a/pkg/resources/ali/ali.go
+++ b/pkg/resources/ali/ali.go
@@ -59,9 +59,6 @@ type clusterDiscoveryALI struct {
 }
 
 func ListResourcesALI(aliCloud aliup.ALICloud, clusterName string, region string) (map[string]*resources.Resource, error) {
-	if region == "" {
-		region = aliCloud.Region()
-	}
 
 	resources := make(map[string]*resources.Resource)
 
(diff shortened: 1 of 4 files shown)


pkg/resources/ali/ali.go
```diff
@@ -60,7 +60,7 @@ type clusterDiscoveryALI struct {
 
 func ListResourcesALI(aliCloud aliup.ALICloud, clusterName string, region string) (map[string]*resources.Resource, error) {
 	if region == "" {
-		region = aliCloud.Region()
+		aliCloud.Region()
```
johngmyers: This seems nonsensical. If `region` is not used, then this whole if block should be removed.

Aresforchina: so...I will remove them together.

PR #8193, opened 2019-12-24
upup/pkg/fi-fix staticcheck

Aresforchina:
ref:#7800
special notes:
upup/pkg/fi/cloudup/apply_cluster.go:300:15: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:307:15: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:1003:14: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:1010:14: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:1014:14: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:1025:14: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:1069:14: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:1076:14: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:1080:14: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/apply_cluster.go:1091:14: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
upup/pkg/fi/cloudup/awstasks/sshkey.go:206:79: argument a is overwritten before first use (SA4009)
upup/pkg/fi/cloudup/loader.go:58:2: field templates is unused (U1000)
upup/pkg/fi/executor.go:162:6: type runnable is unused (U1000)

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -299,14 +299,14 @@ func (c *ApplyClusterCmd) Run() error {
 
 		if warn {
 			fmt.Println("")
-			fmt.Println(starline)
+			fmt.Printf("%s\n", starline)
 			fmt.Println("")
 			fmt.Println("Kubelet anonymousAuth is currently turned on. This allows RBAC escalation and remote code execution possibilities.")
 			fmt.Println("It is highly recommended you turn it off by setting 'spec.kubelet.anonymousAuth' to 'false' via 'kops edit cluster'")
 			fmt.Println("")
 			fmt.Println("See https://kops.sigs.k8s.io/security/#kubelet-api")
 			fmt.Println("")
-			fmt.Println(starline)
+			fmt.Printf("%s\n", starline)
 			fmt.Println("")
 		}
 	}
@@ -1001,20 +1001,20 @@ func (c *ApplyClusterCmd) validateKopsVersion() error {
 	}
 
 	if recommended != nil && !required {
-		fmt.Println("")
-		fmt.Println(starline)
-		fmt.Println("")
+		fmt.Printf("\n")
+		fmt.Printf("%s\n", starline)
+		fmt.Printf("\n")
 		fmt.Printf("A new kops version is available: %s", recommended)
-		fmt.Println("")
-		fmt.Println("upgrading is recommended")
+		fmt.Printf("\n")
+		fmt.Printf("Upgrading is recommended\n")
 		fmt.Printf("More information: %s\n", buildPermalink("upgrade_kops", recommended.String()))
-		fmt.Println("")
-		fmt.Println(starline)
-		fmt.Println("")
+		fmt.Printf("\n")
+		fmt.Printf("%s\n", starline)
+		fmt.Printf("\n")
 	} else if required {
-		fmt.Println("")
-		fmt.Println(starline)
-		fmt.Println("")
+		fmt.Printf("\n")
+		fmt.Printf("%s\n", starline)
+		fmt.Printf("\n")
 		if recommended != nil {
 			fmt.Printf("a new kops version is available: %s\n", recommended)
 		}
@@ -1023,9 +1023,9 @@ func (c *ApplyClusterCmd) validateKopsVersion() error {
 		fmt.Printf("(you can bypass this check by exporting KOPS_RUN_OBSOLETE_VERSION)\n")
 		fmt.Println("")
 		fmt.Printf("More information: %s\n", buildPermalink("upgrade_kops", recommended.String()))
-		fmt.Println("")
-		fmt.Printf(starline)
-		fmt.Println("")
+		fmt.Printf("\n")
+		fmt.Printf("%s\n", starline)
+		fmt.Printf("\n")
 	}
 
 	if required {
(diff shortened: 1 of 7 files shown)


upup/pkg/fi/cloudup/awstasks/sshkey.go
```diff
@@ -208,7 +208,7 @@ func (_ *SSHKey) RenderCloudformation(t *cloudformation.CloudformationTarget, a,
 
 	klog.Warningf("Cloudformation does not manage SSH keys; pre-creating SSH key")
 
-	a, err := e.find(cloud)
+	_, err := e.find(cloud)
```
rifelpet: this `a` variable is used by line 216, correct? I don't think replacing it with `_` is the correct fix here.

Aresforchina: this instance shouldn't get flagged – because it is only being set conditionally, in one branch.or can you give me some advice,please?

johngmyers: The mechanical fix would be to change the parameter `a` to `_`. But then, why is there a parameter? Is the `e.find(cloud)` call needed?

rifelpet: This function is called using reflection, so changing its signature may not be possible:

https://github.com/kubernetes/kops/blob/5cb2d3605c15e28fa077c1b9223e5ef357538086/upup/pkg/fi/context.go#L218-L223

Renaming the parameter to `_` might be the best approach. I'm sure there are other tasks that have unused `a` parameters in some of their Render functions, perhaps we look at how it is handled elsewhere?

johngmyers: My question was more along the lines of "Is the value of the `a` parameter semantically different than the `a` obtained from `e.find(cloud)`? If not, the call wouldn't be needed.

rifelpet: After reviewing the [commit](https://github.com/kubernetes/kops/commit/7d68f744f6#diff-cbd08d10644cc285349e5b764cd339eeR279) that added these lines, I understand the context a bit more. For reference, `a` is the "actual" state of the task, `e` is the "expected" state of the task.

For CloudFormation (and Terraform) targets we dont care about the actual state of the resource in AWS, we're generating the same terraform output regardless. Therefor most tasks' RenderTerraform functions don't use their `a` parameters. SSHKey is an exception because CloudFormation doesnt support creating them, so Kops needs to create the key AWS directly, which requires knowing the actual state of the ssh key. Now the question becomes "does the `a` parameter get populated with the true actual state of the resource even for cloudformation targets?" and based on some testing I believe the answer is yes. Therefor I think we can get rid of this line and the error checking directly below it.

rifelpet: Sorry I misspoke, `a` does not get populated with cloudformation target, therefor the best fix would be to do what [other tasks do](https://github.com/kubernetes/kops/blob/a5ea3809fc76d052f0541a90a0ac2d29294ad984/upup/pkg/fi/cloudup/awstasks/dnszone.go#L317) and use a different variable name, perhaps `keypair`.

Aresforchina: @rifelpet Hi!I have followed your suggestion and it occurs staticcheck fault.I have already type make and hack/verify-staticcheck.sh and they are passed.Can you tell me if you know about info about it,please?

tanjunchen:
@Aresforchina  
can you fix the conflicts , thanks

Aresforchina:
@tanjunchen yes,I will fix them,thanks!

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1082,18 +1082,18 @@ func (c *ApplyClusterCmd) validateKubernetesVersion() error {
...
-		fmt.Printf(starline)
+		fmt.Print(starline)
 		fmt.Printf("\n")
 		fmt.Printf("A new kubernetes version is available: %s\n", recommended)
 		fmt.Printf("Upgrading is recommended (try kops upgrade cluster)\n")
 		fmt.Printf("\n")
 		fmt.Printf("More information: %s\n", buildPermalink("upgrade_k8s", recommended.String()))
 		fmt.Printf("\n")
-		fmt.Printf(starline)
+		fmt.Print(starline)
```
hakman: To keep the same style, same in a few other places:
```suggestion
		fmt.Printf("%s\n"), starline)
```

Aresforchina: yes,sir.I have fixed them together.

Aresforchina:
@hakman you give me this version file that print style as `println() ` ,but last time suggest me update it into  fmt.Printf("%s\n"), starline). so could I  entirely change print style into `println() `  in  apply_cluster.go now?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -299,14 +299,14 @@ func (c *ApplyClusterCmd) Run() error {
 
 		if warn {
 			fmt.Println("")
-			fmt.Println(starline)
+			fmt.Printf(("%s\n"), starline)
```
johngmyers: Why all the extra parentheses?

hakman:
@Aresforchina sorry for the confusion. I wanted to suggest to change that in areas where there was mostly printf.

As @johngmyers pointed out, there were also some extra parenthesis and new lines in the current version. Seemed easier to just provide a full file as suggestion.

upup/pkg/fi/cloudup/apply_cluster.go
Aresforchina: This is my syntax fault .Thank you for pointing it out.

Aresforchina:
@hakman Thanks for your patience. I will fix it as soon as possible.

justinsb:
I think this looks good - thanks for sticking with it @Aresforchina 

/approve
/lgtm

PR #8196, opened 2019-12-24
For dev, don't preload docker images on nodes

justinsb:
If we're not going to use the docker images, we don't need to preload
them - saves a bit off the boot time.

### DIFF
--- a/nodeup/pkg/model/protokube.go
+++ b/nodeup/pkg/model/protokube.go
@@ -55,6 +55,15 @@ func (t *ProtokubeBuilder) Build(c *fi.ModelBuilderContext) error {
 		return nil
 	}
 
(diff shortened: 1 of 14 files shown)


hakman:
Nice! This saves me a PR.

In the same context, shouldn't the LoadImage task be in the ProtokubeBuilder?
https://github.com/kubernetes/kops/blob/0b52b994be03f143608c11f74a56532edd8cce7b/upup/pkg/fi/nodeup/command.go#L304-L310

rifelpet:
shoot, should have merged this before my mocks update

justinsb:
Also moved the protokube image loading into ProtokubeBuilder - thanks for the suggestion @hakman!

PR #8151, opened 2019-12-18
WIP: Support cluster-api for non-master instancegroups, on GCE

justinsb:

PR #8198, opened 2019-12-24
WIP: cluster-api - boot nodes without state store access

justinsb:
This is the other half of the initial cluster-api spike (the first half is #8151).

Instead of requiring nodes to read from the state store (s3 or GCS) in nodeup, we spin up a GRPC server in kops-controller that lets us serve the artifacts.  Currently they're still loaded from S3/GCS by kops-controller, but this means the node doesn't need permissions.  We're also getting closer to no S3/GCS dependency, if kops-controller can generate more of these artifacts on the fly (it's only the kubelet & kube-proxy keypairs).

What this does now allow is dynamic creation of InstanceGroups via the k8s api.


nodeup/pkg/model/kops_controller.go
```diff
@@ -0,0 +1,159 @@
...
+		Contents: fi.NewBytesResource(pkiutil.EncodeCertPEM(serverCert)),
+		Type:     nodetasks.FileType_File,
+		Mode:     s("0644"),
+	})
+	c.AddTask(&nodetasks.File{
+		Path:     filepath.Join(pkiDir, "server.key"),
+		Contents: fi.NewBytesResource(pkiutil.EncodePrivateKeyPEM(serverKey)),
+		Type:     nodetasks.FileType_File,
+
+		// TODO: Because we run as an unprivileged user (1000), we can't use hostPaths with mode 600
```
johngmyers: Perhaps specify the file `Owner`?

cmd/kops-controller/pkg/clusterapi/builder.go
```diff
@@ -0,0 +1,168 @@
...
+	Clientset    simple.Clientset
+}
+
+// ClusterAPIBuilder is implemented by the model task that knows how to map an InstanceGroup to k8s objects
+type ClusterAPIBuilder interface {
+	MapToClusterAPI(cluster *kops.Cluster, ig *kops.InstanceGroup) ([]*unstructured.Unstructured, error)
+}
+
+func (b *Builder) BuildMachineDeployment(clusterObj *kopsv1alpha2.Cluster, igObj *kopsv1alpha2.InstanceGroup) ([]*unstructured.Unstructured, error) {
+	cloudup.AlphaAllowGCE.SetEnabled(true)
```
mitch000001: I guess it is necessary to enable this builder to run GCE. But I wonder if we can move it somewhere else, where it does not bloat a unrelated context.

olemarkus:
This one is partially superseded by #10469, but this one also adds support for MachineDeployment resources

PR #8199, opened 2019-12-24
containerd: Add support for tar.gz package

hakman:
Each containerd release comes with a tar.gz that can be used to install it with minimal effort.
Ref: https://github.com/kubernetes/kops/pull/7986#discussion_r357925711

Summary of changes:
* Archive task will extract each required file to the desired location - easier to control what goes where, but may need versioning in the future
* the package contains static binaries, so no need to provide a distro like with each rpm/deb
* having a single package means dependencies need to go to Packages
* support for CentOS should be much simpler now that `container-selinux` is installed from repo and not vault (vault package is still needed for RHEL and Amazon Linux)

**PS1**: Docker could get some similar upgrade by moving the deps to Packages and would open the way to multi-arch.

**PS2**: Amazon Linux 2 doesn't have the packages required to run the latest containerd (and Docker). At least the deps for `container-selinux` from CentOS are missing. I suspect this distro will need special care if support for it is desired.

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -47,7 +47,6 @@ var containerdVersions = []packageVersion{
 		Version:        "1.2.4-1",
 		Source:         "https://download.docker.com/linux/debian/dists/stretch/pool/stable/amd64/containerd.io_1.2.4-1_amd64.deb",
 		Hash:           "48c6ab0c908316af9a183de5aad64703bc516bdf",
-		Dependencies:   []string{"libseccomp2", "pigz"},
 	},
 
 	// 1.2.10 - Debian Stretch
@@ -59,7 +58,6 @@ var containerdVersions = []packageVersion{
 		Version:        "1.2.10-3",
 		Source:         "https://download.docker.com/linux/debian/dists/stretch/pool/stable/amd64/containerd.io_1.2.10-3_amd64.deb",
 		Hash:           "186f2f2c570f37b363102e6b879073db6dec671d",
-		Dependencies:   []string{"libseccomp2", "pigz"},
 	},
 
 	// 1.2.10 - Debian Buster
@@ -71,7 +69,6 @@ var containerdVersions = []packageVersion{
 		Version:        "1.2.10-3",
 		Source:         "https://download.docker.com/linux/debian/dists/buster/pool/stable/amd64/containerd.io_1.2.10-3_amd64.deb",
 		Hash:           "365e4a7541ce2cf3c3036ea2a9bf6b40a50893a8",
-		Dependencies:   []string{"libseccomp2", "pigz"},
 	},
 
 	// 1.2.10 - Ubuntu Xenial
@@ -83,7 +80,6 @@ var containerdVersions = []packageVersion{
 		Version:        "1.2.10-3",
 		Source:         "https://download.docker.com/linux/ubuntu/dists/xenial/pool/stable/amd64/containerd.io_1.2.10-3_amd64.deb",
 		Hash:           "b64e7170d9176bc38967b2e12147c69b65bdd0fc",
-		Dependencies:   []string{"libseccomp2", "pigz"},
 	},
 
 	// 1.2.10 - Ubuntu Bionic
@@ -95,7 +91,6 @@ var containerdVersions = []packageVersion{
 		Version:        "1.2.10-3",
 		Source:         "https://download.docker.com/linux/ubuntu/dists/bionic/pool/stable/amd64/containerd.io_1.2.10-3_amd64.deb",
 		Hash:           "f4c941807310e3fa470dddfb068d599174a3daec",
-		Dependencies:   []string{"libseccomp2", "pigz"},
 	},
 
 	// 1.2.10 - CentOS / Rhel 7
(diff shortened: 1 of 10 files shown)


nodeup/pkg/model/containerd.go
```diff
@@ -220,11 +226,19 @@ func (b *ContainerdBuilder) Build(c *fi.ModelBuilderContext) error {
...
-					Source:          dv.Source,
-					Hash:            dv.Hash,
-					TargetDir:       "/usr/bin/",
-					StripComponents: 1,
+					Name:      "containerd.io",
+					Source:    dv.Source,
+					Hash:      dv.Hash,
+					TargetDir: "/",
+					MapFiles: map[string]string{
+						"./usr/local/bin/ctr":                     "/usr/bin",
```
justinsb: I'm wondering if we should map directories, as this likely makes it harder to support new tar.gz bundles we don't yet know about (they may have added a file)...

nodeup/pkg/model/containerd.go
```diff
@@ -293,21 +309,21 @@ func (b *ContainerdBuilder) buildSystemdService() *nodetasks.Service {
...
+	manifest.Set("Service", "OOMScoreAdjust", "-999")
 
 	manifest.Set("Service", "LimitNOFILE", "1048576")
 	manifest.Set("Service", "LimitNPROC", "infinity")
 	manifest.Set("Service", "LimitCORE", "infinity")
 	manifest.Set("Service", "TasksMax", "infinity")
 
-	manifest.Set("Service", "Restart", "always")
-	manifest.Set("Service", "RestartSec", "2s")
-	manifest.Set("Service", "StartLimitInterval", "0")
```
justinsb: We might actually need these to prevent systemd marking the unit as failed and not always restarting e.g. https://serverfault.com/questions/736624/systemd-service-automatic-restart-after-startlimitinterval

However, I think with RestartSec = 5 we're OK with the default values, but it is possible that systemd is not running with the default values.

Also annoying is that the field was renamed from StartLimitInterval to StartLimitIntervalSec at some stage; I don't know if any of our OSes uses StartLimitInterval.

We might need a helper function - so we can probably leave this as-is for now :-)

nodeup/pkg/model/docker.go
```diff
@@ -188,7 +188,7 @@ var dockerVersions = []packageVersion{
 				Hash:    "a6b0243af348140236ed96f2e902b259c590eefa",
 			},
 		},
-		Dependencies: []string{"libtool-ltdl", "libseccomp"},
+		Dependencies: []string{"libtool-ltdl"},
```
justinsb: These are the ones we're moving into packages.go?

upup/pkg/fi/nodeup/nodetasks/archive.go
```diff
@@ -155,20 +160,38 @@ func (_ *Archive) RenderLocal(t *local.LocalTarget, a, e, changes *Archive) erro
...
+			}
+		} else {
+			for src, dest := range e.MapFiles {
+				stripCount := strings.Count(src, "/")
+				targetDir := filepath.Join(e.TargetDir, dest)
+				if err := os.MkdirAll(targetDir, 0755); err != nil {
+					return fmt.Errorf("error creating directories %q: %v", targetDir, err)
+				}
+
+				args := []string{"tar", "xf", localFile, "-C", targetDir, "--strip-components=" + strconv.Itoa(stripCount), src}
```
justinsb: This is going to be slow I think, because we have to read the whole tar file every time (tar files aren't indexed, unlike zip files).

nodeup/pkg/model/containerd.go
```diff
@@ -107,14 +102,6 @@ var containerdVersions = []packageVersion{
 		Version:        "1.2.10",
 		Source:         "https://download.docker.com/linux/centos/7/x86_64/stable/Packages/containerd.io-1.2.10-3.2.el7.x86_64.rpm",
 		Hash:           "f6447e84479df3a58ce04a3da87ccc384663493b",
-		ExtraPackages: map[string]packageInfo{
-			"container-selinux": {
```
justinsb: I guess the problem here is that container-selinux might be tied to the behaviour of a particular version of containerd.  i.e. a new version of containerd introduces a new required permission (?)

justinsb:
Thanks @hakman ... generally looks good but I have some questions!

nodeup/pkg/model/containerd.go
hakman: My assumption was that newer container-selinux will only add permissions, which would make it backwards compatible with any containerd or docker version. In a regular install. where one has a valid subscription, any package upgrade would grab the latest version.

nodeup/pkg/model/containerd.go
hakman: I also saw that Kops doesn't use `StartLimitInterval` anywhere. Besides the name, it was also moved from `[Service]` to `[Unit]` at some point.
https://lists.freedesktop.org/archives/systemd-devel/2017-July/039255.html

Unless someone will complain, I also think we can skip it for now.

nodeup/pkg/model/docker.go
hakman: Yes. Moving to packages.go.

upup/pkg/fi/nodeup/nodetasks/archive.go
hakman: In theory, yes. In practice it's faster than `yum install containerd`:
```
I0112 04:43:20.381756     652 files.go:100] Hash matched for "/var/cache/nodeup/archives/containerd.io": sha1:f451d46280104588f236bee277bca1da8babc0e8
I0112 04:43:20.381805     652 archive.go:189] running command [tar xf /var/cache/nodeup/archives/containerd.io -C /usr --strip-components=3 ./usr/local/bin]
I0112 04:43:22.038539     652 archive.go:189] running command [tar xf /var/cache/nodeup/archives/containerd.io -C /usr --strip-components=3 ./usr/local/sbin]
I0112 04:43:23.601597     652 executor.go:103] Tasks: 2 done / 60 total; 45 can run
``

nodeup/pkg/model/containerd.go
hakman: **Very** good point. Updated and works perfectly.

hakman:
Thanks @justinsb. I think this will make it simpler to support multi-arch. I see that the guys working on containerd are improving their support for multi-arch and hopefully will release official builds soon.
https://github.com/containerd/containerd/pull/3758

Would it make sense to try and move Docker deps to Packages and add a tar.gz package there also?

justinsb:
Thanks for the tweaks - looks great!

/approve
/lgtm

justinsb:
For docker, my inclination would be to leave it as-is, and gradually let it age-out out of our code, as I think people will switch to pure containerd.  But if there's an advantage to refactoring it, we can do that!

PR #8200, opened 2019-12-25
Add missing priorityClassName for critical pods

### LINKED ISSUE #7605: Critical pod annotation - used by kube-proxy - is now deprecated
The special pod annotation `scheduler.alpha.kubernetes.io/critical-pod` - used by `kube-proxy` run by `kops` (via static manifest file) - has been [deprecated in Kubernetes `1.13`](https://github.com/kubernetes/kubernetes/pull/70298) and we should stop using it in `kops`.

As a side note, there's another issue open https://github.com/kubernetes/kops/issues/6527 discussing the option to run `kube-proxy` as a DaemonSet.

johngmyers:
/kind bug

Several pod manifests with the `scheduler.alpha.kubernetes.io/critical-pod` annotation (for which support was removed in Kubernetes 1.16) were missing a corresponding `priorityClassName`.

Fixes #7605


upup/models/cloudup/resources/addons/digitalocean-cloud-controller.addons.k8s.io/k8s-1.8.yaml.template
```diff
@@ -30,6 +30,7 @@ spec:
       serviceAccountName: cloud-controller-manager
       dnsPolicy: Default
       hostNetwork: true
+      priorityClassName: system-node-critical
```
rifelpet: According to [this table](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/), `priorityClassName` was added by default in Kubernetes 1.11. Since this manifest and the `networking.amazon-vpc-routed-eni` one below match older versions of kubernetes, we may need to fork the manifests for k8s 1.11.

johngmyers: The functions in [kubemanifest](https://github.com/kubernetes/kops/blob/master/pkg/kubemanifest/priority.go) unconditionally set these and are called without any version checks.

The field is also present in `coredns.addons.k8s.io/k8s-1.6.yaml.template`.

johngmyers: My attempt to bring up a 1.8.15 cluster with master kops failed. The API DNS entry never changed from the placeholder address.

johngmyers: According to the Kubernetes release notes, `priorityClassName` was added to the pod spec in 1.8.0. So this should not be a blocker.

johngmyers:
Fixes #7605

justinsb:
I believe we've now unblocked this, and we would like to merge this and cherry-pick back to >= 1.16 (?)

Thanks for driving this @johngmyers - though I think it needs a rebase (hopefully just one more!)

/approve
/lgtm

PR #8201, opened 2019-12-25
GCE: Naming of google_compute VPC route now includes cluster name

mccare:
The VPC was always named "default" when no NetworkID was given in the cluster spec.

* changed it to now be `vpc-{sanitized-cluster-name}`
* moved the length limit function into the util.go 
* SafeObjectName will now limit the name length to 63, only lower case chars. 

Didn't find a documentation on the specifics of the name but the website statet (often) that names should only contain lower alphanumerics and hyphens and should be max 63 chars long.

### DIFF
--- a/pkg/model/gcemodel/context.go
+++ b/pkg/model/gcemodel/context.go
@@ -37,7 +37,7 @@ func (c *GCEModelContext) LinkToNetwork() *gcetasks.Network {
 func (c *GCEModelContext) NameForNetwork() string {
 	networkName := c.Cluster.Spec.NetworkID
 	if networkName == "" {
-		networkName = "default"
+		networkName = c.SafeObjectName("vpc")
 	}
 	return networkName
 }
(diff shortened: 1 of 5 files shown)


rifelpet:
I definitely like the idea of naming the network after the cluster, however renaming an existing network doesn't appear to be supported by GCE based on the [terraform provider](https://github.com/terraform-providers/terraform-provider-google/blob/1594f147795d9389f63618184fa7cea7f24415c1/google/resource_compute_network.go#L48) which means Kops will just create a new network and I'm not sure what will happen to all of the other resources using the old network, and terraform will try to delete and recreate the network which will fail because of all the dependent resources already in the network. Therefor I don't know how we could feasibly update the network name on an existing cluster, it would need to be for new clusters only.

Thoughts @geojaz ?

mccare:
You are definitely right there. Existing installations need to define NetworkID in their clusterSpec to continue using "default" as VPC name.

PR #8203, opened 2019-12-26
Fix upup/tools/generators/pkg/codegen staticcheck failures

johngmyers:
Ref:#7800

```
upup/tools/generators/pkg/codegen/parse.go:40:2: field typeName is unused (U1000)
upup/tools/generators/pkg/codegen/parse.go:113:35: importer.For is deprecated: Use ForCompiler, which populates a FileSet with the positions of objects created by the importer.  (SA1019)
```

### DIFF
--- a/hack/.staticcheck_failures
+++ b/hack/.staticcheck_failures
@@ -9,4 +9,3 @@ upup/pkg/fi
 upup/pkg/fi/cloudup
 upup/pkg/fi/cloudup/awstasks
 upup/pkg/kutil
-upup/tools/generators/pkg/codegen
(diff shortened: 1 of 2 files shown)


tanjunchen:
/lgtm
Thanks

johngmyers:
Is Prow running those jobs with Go 1.11? I believe our minimum requirement is Go 1.12.

rifelpet:
interesting, you're right.

https://github.com/kubernetes/test-infra/blob/c628c3674eb0c9292812d760232afcdbb7c79c99/config/jobs/kubernetes/kops/kops-config.yaml#L378

That repo has a tool that automatically bumps image tags as they get periodically rebuilt but it seems the image we're using [hasn't been updated since January](https://console.cloud.google.com/gcr/images/k8s-testimages/GLOBAL/gcloud-in-go?gcrImageListsize=30) so there is probably a better image we should be using. I'll do some investigation.

rifelpet:
FWIW i ran the generator using both the master branch and this branch and the output was identical, so
/lgtm

hopefully https://github.com/kubernetes/test-infra/pull/15691 will get merged soon

rifelpet:
There we go!
/approve
/lgtm

PR #8204, opened 2019-12-27
Enable host logging for kops-controller

rifelpet:
This makes it easier to get the kops-controller logs from e2e tests since they [only dump logs](https://github.com/kubernetes/test-infra/blob/ec0fe6bd3603dd927cced9b398a8b3a951ab1157/kubetest/dump.go#L50-L74) from systemd services and /var/log files.

This is a bit different than the other components that we log to the host for... others are static pods that are ran on every master (or every node in the case of kube-proxy) but this is a deployment, so the log file will only be present on the master that happens to have the pod scheduled.

### DIFF
--- a/nodeup/pkg/model/logrotate.go
+++ b/nodeup/pkg/model/logrotate.go
@@ -70,6 +70,7 @@ func (b *LogrotateBuilder) Build(c *fi.ModelBuilderContext) error {
 	b.addLogRotate(c, "kubelet", "/var/log/kubelet.log", logRotateOptions{})
 	b.addLogRotate(c, "etcd", "/var/log/etcd.log", logRotateOptions{})
 	b.addLogRotate(c, "etcd-events", "/var/log/etcd-events.log", logRotateOptions{})
+	b.addLogRotate(c, "kops-controller", "/var/log/kops-controller.log", logRotateOptions{})
 
 	if err := b.addLogrotateService(c); err != nil {
 		return err
--- a/upup/models/cloudup/resources/addons/kops-controller.addons.k8s.io/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/kops-controller.addons.k8s.io/k8s-1.16.yaml.template
@@ -87,6 +87,8 @@ spec:
 {{ end }}
         - mountPath: /etc/kubernetes/kops-controller/
           name: kops-controller-config
+        - mountPath: /var/log/kops-controller.log
+          name: logfile
         command:
 {{ range $arg := KopsControllerArgv }}
         - "{{ $arg }}"
@@ -112,7 +114,10 @@ spec:
       - name: kops-controller-config
         configMap:
           name: kops-controller
-
+      - name: logfile
+        hostPath:
+          path: /var/log/kops-controller.log
+          type: FileOrCreate
 ---
 
 apiVersion: v1
--- a/upup/pkg/fi/cloudup/template_functions.go
+++ b/upup/pkg/fi/cloudup/template_functions.go
@@ -380,6 +380,8 @@ func (tf *TemplateFunctions) KopsControllerArgv() ([]string, error) {
 
 	argv = append(argv, "--conf=/etc/kubernetes/kops-controller/config.yaml")
 
+	argv = append(argv, "--logtostderr=false", "--alsologtostderr", "--log_file=/var/log/kops-controller.log")
+
 	return argv, nil
 }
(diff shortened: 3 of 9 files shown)


rifelpet:
looks like this e2e failure is actually related. will continue to troubleshoot. I think we need to add a FileTask like the other log files:

https://github.com/kubernetes/kops/blob/8436a3ea8dc75ba1a7e99957259ae28c37805b28/nodeup/pkg/model/kube_apiserver.go#L106-L112

which means we'd need to add a kops-controller model to nodeup/pkg which i'm not sure is the best design decision. will bring this up at the next office hours.
/hold

justinsb:
So actually I think in one of my open PRs I am creating a kops-controller module.  It's to create keypairs for serving GRPC, which means if we're using cluster-api we don't need to write to S3 (it also means that nodes might no longer need S3 access)

I'm not sure what the error is here - I hit some permissions problems with reading the key, so it might be the same thing.  You could also try setting `type: FileOrCreate` on the hostPath, but it's fine also to do this in nodeup IMO.

Also, kops-controller should be a daemonset, with leader-election.  It was a deployment briefly, but that was my mistake...

rifelpet:
You're right, i dont know why i still thought it was a deployment. thats what i get for opening these PRs from a plane :)
The kubelet logs are vague, just indicating the container dies immediately:

```
I1227 04:10:41.354129    6555 kube_docker_client.go:345] Stop pulling image "kope/kops-controller:1.15.0-alpha.1": "Status: Downloaded newer image for kope/kops-controller:1.15.0-alpha.1"
I1227 04:10:42.147482    6555 kubelet.go:1951] SyncLoop (PLEG): "kops-controller-wlrz7_kube-system(a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb)", event: &pleg.PodLifecycleEvent{ID:"a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb", Type:"ContainerDied", Data:"bee3cc9c16724ab57fe4aeaf10acd977efbcb2dcef74621138a0886645377be2"}
I1227 04:10:43.166432    6555 kubelet.go:1951] SyncLoop (PLEG): "kops-controller-wlrz7_kube-system(a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb)", event: &pleg.PodLifecycleEvent{ID:"a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb", Type:"ContainerDied", Data:"565aaa5a0ddaff32eefa2045cd00c87ea3372cc92403bf4c8f77bd6d760185e7"}
E1227 04:10:43.167967    6555 pod_workers.go:191] Error syncing pod a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb ("kops-controller-wlrz7_kube-system(a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb)"), skipping: failed to "StartContainer" for "kops-controller" with CrashLoopBackOff: "back-off 10s restarting failed container=kops-controller pod=kops-controller-wlrz7_kube-system(a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb)"
E1227 04:10:44.176691    6555 pod_workers.go:191] Error syncing pod a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb ("kops-controller-wlrz7_kube-system(a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb)"), skipping: failed to "StartContainer" for "kops-controller" with CrashLoopBackOff: "back-off 10s restarting failed container=kops-controller pod=kops-controller-wlrz7_kube-system(a52a88c9-877b-4f3a-a64e-a4ea0a43a3bb)"
```

I wonder if its a permission issue? will kubelet handle chowning the hostPath file if the kops-controller container runs with uid=1000?

https://github.com/kubernetes/kops/blob/master/cmd/kops-controller/BUILD.bazel#L40-L49

rifelpet:
@justinsb BTW is there a reason kops-controller is a DaemonSet rather than static pods?

rifelpet:
I fixed `--log-file` to `--log_file` and it now appears to be working. The E2E test now has the kops-controller logs as artifacts too:

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/8204/pull-kops-e2e-kubernetes-aws/1214592841498497030/artifacts/ip-172-20-53-129.eu-central-1.compute.internal/kops-controller.log

The E2E failure is unrelated so i'm removing the hold.
/cc @justinsb
/hold cancel

PR #8208, opened 2019-12-28
Remove support for Kubernetes 1.8 and earlier

johngmyers:


upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1047,13 +1049,22 @@ func (c *ApplyClusterCmd) validateKubernetesVersion() error {
 	}
 
 	if !util.IsKubernetesGTE(OldestSupportedKubernetesVersion, *parsed) {
+		fmt.Printf("This version of Kubernetes is no longer supported; upgrading Kubernetes is required\n")
+		fmt.Printf("\n")
+		fmt.Printf("More information: %s\n", buildPermalink("upgrade_k8s", OldestRecommendedKubernetesVersion))
```
rifelpet: Any reason to use `OldestRecommendedKubernetesVersion` in the url anchor? There isn't an anchor for `#1.10.0` on the page currently.

johngmyers: I'm following the example from later in the function where it deprecates or removes support based on what is in the channel. It gives us the flexibility to add version-specific sections to the file at a later date.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -93,6 +93,8 @@ var (
 	CloudupModels = []string{"proto", "cloudup"}
 	// OldestSupportedKubernetesVersion is the oldest kubernetes version that is supported in Kops
 	OldestSupportedKubernetesVersion = "1.9.0"
+	// OldestRecommendedKubernetesVersion is the oldest kubernetes version that is not deprecated in Kops
+	OldestRecommendedKubernetesVersion = "1.10.0"
```
hakman: Maybe it's just me, but I would not recommend any k8s version where the default etcd is not 3.x. Anyone having to got through the upgrade from 2.x to 3.x will be pretty annoyed.
https://github.com/kubernetes/kops/blob/298f79659a2bd9c1d6548c6c95b335b5e29e6466/pkg/model/components/etcdmanager/options.go#L60-L61

johngmyers: The current policy as I understand it is to remove support from 1.8 and earlier in kops 1.18, then do rolling removals in subsequent versions. That means removing support from 1.9 in kops 1.19. There was generalized talk about narrowing the gap in future releases, but no specific decisions there have been made.

hakman: I don't have anything against supporting these versions, but the name says "recommended" version. I think "non-deprecated" is not quite the same as "recommended".

johngmyers: I am not able to find a better word for "non-deprecated". The code uses "recommended" for "non-deprecated" later in the file, where it's dealing with channels.

hakman: Maybe we can use "legacy" here.

johngmyers: The word "recommended" is not made visible to the user. The constant only demarcates what is not described as "deprecated".

justinsb:
This LGTM - thanks @johngmyers 

Agree that we have a lot of version-thresholds now - deprecated, minimum-recommended, unsupported etc :-)   I imagine we'll figure this out over time - or maybe just align them!

/approve
/lgtm

upup/pkg/fi/cloudup/apply_cluster.go
justinsb: Yeah, if & when we come up with better words we can rename this constant without affecting anything user-visible!

justinsb:
Readding LGTM - thanks for fixing the staticcheck error @johngmyers 

/approve
/lgtm

PR #8209, opened 2019-12-28
Fix link printed in k8s version deprecation message

rifelpet:
Thanks for pointing this out @johngmyers 

It is meant to point [here](https://github.com/kubernetes/kops/blob/master/permalinks/upgrade_k8s.md)


justinsb:
Thanks for fixing @rifelpet and to @johngmyers for spotting

/approve
/lgtm

PR #8212, opened 2019-12-28
Move nodeup to /opt/kops/bin

hakman:
RHEL / CenOS 8 have some restrictions on binaries location for Systemd.
This is why initial configuration for `nodeup` cannot run and e2e tests are failing.
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-containerd

Moving `nodeup` to /opt/kops/bin solves this issue.

At same time, `iptables-setup` script should also move to /opt/kops/bin. https://github.com/kubernetes/kops/pull/8164#discussion_r361681377

```
/opt/kops/
├── bin
│   ├── iptables-setup
│   └── nodeup
└── conf
    ├── cluster_spec.yaml
    ├── ig_spec.yaml
    └── kube_env.yaml
```

### DIFF
--- a/pkg/model/resources/nodeup.go
+++ b/pkg/model/resources/nodeup.go
@@ -53,12 +53,9 @@ NODEUP_HASH={{ NodeUpSourceHash }}
 {{ ProxyEnv }}
 
 function ensure-install-dir() {
-  INSTALL_DIR="/var/cache/kubernetes-install"
-  # On ContainerOS, we install to /var/lib/toolbox install (because of noexec)
-  if [[ -d /var/lib/toolbox ]]; then
-    INSTALL_DIR="/var/lib/toolbox/kubernetes-install"
-  fi
-  mkdir -p ${INSTALL_DIR}
+  INSTALL_DIR="/opt/kops"
+  mkdir -p ${INSTALL_DIR}/bin
+  mkdir -p ${INSTALL_DIR}/conf
   cd ${INSTALL_DIR}
 }
 
@@ -140,6 +137,7 @@ function try-download-release() {
 
 function download-release() {
   # In case of failure checking integrity of release, retry.
+  cd ${INSTALL_DIR}/bin
   until try-download-release; do
     sleep 15
     echo "Couldn't download release. Retrying..."
@@ -147,7 +145,7 @@ function download-release() {
 
   echo "Running nodeup"
   # We can't run in the foreground because of https://github.com/docker/docker/issues/23793
-  ( cd ${INSTALL_DIR}; ./nodeup --install-systemd-unit --conf=${INSTALL_DIR}/kube_env.yaml --v=8  )
+  ( cd ${INSTALL_DIR}/bin; ./nodeup --install-systemd-unit --conf=${INSTALL_DIR}/conf/kube_env.yaml --v=8  )
 }
 
 ####################################################################################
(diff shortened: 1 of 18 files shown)


pkg/model/resources/nodeup.go
```diff
@@ -53,12 +53,9 @@ NODEUP_HASH={{ NodeUpSourceHash }}
 {{ ProxyEnv }}
 
 function ensure-install-dir() {
-  INSTALL_DIR="/var/cache/kubernetes-install"
-  # On ContainerOS, we install to /var/lib/toolbox install (because of noexec)
```
justinsb: I _think_ we're going to have to keep the COS workaround, because at this point /opt/kops isn't writeable.  I think we should make it writeable (and I started that in https://github.com/kubernetes/kops/pull/8269), but I don't particularly want to do so in bash.

That said, I'm also happy to break this briefly and then fix it in #8269...

pkg/model/resources/nodeup.go
```diff
@@ -53,12 +53,9 @@ NODEUP_HASH={{ NodeUpSourceHash }}
...
-  if [[ -d /var/lib/toolbox ]]; then
-    INSTALL_DIR="/var/lib/toolbox/kubernetes-install"
-  fi
-  mkdir -p ${INSTALL_DIR}
+  INSTALL_DIR="/opt/kops"
+  mkdir -p ${INSTALL_DIR}/bin
+  mkdir -p ${INSTALL_DIR}/conf
```
justinsb: Thought: If we're making our directories more standard, `/etc/kubernetes/kops` might be better than `/opt/kops/conf`

justinsb:
Generally looks good - thanks for doing this.

I think we need a workaround for COS noexec, and I'm wondering if we should start putting conf files under /etc

I think the test failure is unrelated as we don't seem to have logs...

/retest

pkg/model/resources/nodeup.go
hakman: Usually things that are in /opt have pretty much everything there, with their own structure, /etc seems more like a place for packages that respect the linux structure.
If you like /etc more, maybe `/etc/kops` would be quite nice. Do you think the `kubernetes` part adds any value?

hakman:
@justinsb it took me a while to realise what COS is. I think I can add the workaround back here if needed. But if we want /opt to be a place where we can put other files also, it probably it's better to do it in #8269.

The failed test message is a flake from some old test. Test infra was not working great during the last few weeks. The e2e is already passed in the prow list.

pkg/model/resources/nodeup.go
justinsb: I took a look and these are our options that I believe are mounted rw & exec at boot-time.  /var/lib/toolbox seems no worse than our other choices...

```
/dev/sda1 on /var/lib/google type ext4 (rw,nosuid,nodev,relatime,commit=30)
/dev/sda1 on /var/lib/docker type ext4 (rw,nosuid,nodev,relatime,commit=30)
/dev/sda1 on /var/lib/toolbox type ext4 (rw,nodev,relatime,commit=30)
```

pkg/model/resources/nodeup.go
justinsb: OK - the "self contained" argument sounds reasonable in this case :-)  We can consider /etc/kops or /etc/kubernetes/kops for the future (I don't have a strong preference either way on how important having kubernetes in the path is!)

justinsb:
OK let's get this in, then we can figure out the best of our choices for COS!

/approve
/lgtm

hakman:
Thanks @justinsb. Will add daily e2e next week for all supported distros to make these types changes easier to test.

PR #8217, opened 2019-12-29
containerd: Fix tiny nits

hakman:
This fixes various tiny nits observed after #7986 :
* run Protokube in the "default" container namespace instead of "k8s.io", same way is done with Docker
* delete Protokube container on restart without pausing first, not needed and can cause a deadlock
* update hardcoded service dependencies on "docker.service" to "containerd.service" when needed
* run each package hash check as a separate test, some feedback is nice when it takes 3 min


hakman:
Thanks @mikesplain. Looking forward for some feedback also. :)

PR #8218, opened 2019-12-29
Fix cloudmock/aws/mockelbv2 staticcheck failures

johngmyers:


tanjunchen:
ref:https://github.com/kubernetes/kops/issues/7800

PR #8220, opened 2019-12-29
Cilium nodeport

olemarkus:
Cilium is capable of fully replacing kube-proxy by running in NodePort mode.
This PR will prevent kops from installing kube-proxy if NodePort is enabled on the Cilium addon.

This is implemented similar to how kube-proxy is disabled by kube-router. I do wonder if forcing `kubeProxy.enabled=false` would be a bit cleaner than just disabling it implicitly by the `cilium.enableNodePort` setting.

Note: This feature requires kernel version 4.19.57 or later

### DIFF
--- a/pkg/apis/kops/validation/legacy.go
+++ b/pkg/apis/kops/validation/legacy.go
@@ -602,6 +602,10 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 
 	allErrs = append(allErrs, newValidateCluster(c)...)
 
+	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil && c.Spec.Networking.Cilium.EnableNodePort && c.Spec.KubeProxy != nil && *c.Spec.KubeProxy.Enabled {
+		allErrs = append(allErrs, field.Invalid(fieldSpec.Child("KubeProxy"), "enabled", "When Cilium NodePort is enabled, KubeProxy must be disabled"))
+	}
+
 	return allErrs
 }
 
--- a/pkg/apis/kops/validation/legacy.go
+++ b/pkg/apis/kops/validation/legacy.go
@@ -603,7 +603,7 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 	allErrs = append(allErrs, newValidateCluster(c)...)
 
 	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil && c.Spec.Networking.Cilium.EnableNodePort && c.Spec.KubeProxy != nil && *c.Spec.KubeProxy.Enabled {
-		allErrs = append(allErrs, field.Invalid(fieldSpec.Child("KubeProxy"), "enabled", "When Cilium NodePort is enabled, KubeProxy must be disabled"))
+		allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("KubeProxy").Child("enabled"), "When Cilium NodePort is enabled, KubeProxy must be disabled"))
 	}
 
 	return allErrs
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.7.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.7.yaml.template
@@ -363,6 +363,10 @@ spec:
               key: custom-cni-conf
               name: cilium-config
               optional: true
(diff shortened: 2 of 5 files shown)


upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/cilium/manifest.yaml
```diff
@@ -89,16 +89,16 @@ spec:
   - id: k8s-1.7
     kubernetesVersion: <1.12.0
     manifest: networking.cilium.io/k8s-1.7.yaml
-    manifestHash: 6928e95ec4b8359075e3dfb069f74e290e2e6eb2
+   manifestHash: 870d0a940ece8e98b38b3e8a20c062fc247e9b23
```
johngmyers: This removed space is probably the reason for the bazel failure.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.7.yaml.template
```diff
@@ -634,6 +638,11 @@ spec:
               key: AWS_DEFAULT_REGION
               name: cilium-aws
               optional: true
+        - name: KUBERNETES_SERVICE_HOST
+          value: "{{ .MasterInternalName }}"
+        - name: KUBERNETES_SERVICE_PORT
+          value: "443"
+{{ with .Networking.Cilium }}
```
johngmyers: Why this added `with` command?

pkg/apis/kops/validation/legacy.go
```diff
@@ -602,6 +602,10 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 
 	allErrs = append(allErrs, newValidateCluster(c)...)
 
+	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil && c.Spec.Networking.Cilium.EnableNodePort && c.Spec.KubeProxy != nil && *c.Spec.KubeProxy.Enabled {
+		allErrs = append(allErrs, field.Invalid(fieldSpec.Child("KubeProxy"), "enabled", "When Cilium NodePort is enabled, KubeProxy must be disabled"))
```
johngmyers: Should use `field.Forbidden()` instead.
Should use `fieldspec.Child("KubeProxy").Child("Enabled")` to refer to the correct field.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -909,7 +909,7 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 
 	if b.cluster.Spec.Networking.Cilium != nil {
 		key := "networking.cilium.io"
-		version := "1.6.4-kops.3"
+		version := "1.6.6-kops.0"
```
johngmyers: nit (not blocking): we usually start with `-kops.1`

rifelpet:
Just to help me understand, would specifying `Cilium.EnableNodePort=true` and `KubeProxy.Enabled=true` result in a functional cluster? If so, then no longer allowing that combination of fields would be a breaking change. You mention this being similar to KubeRouter and I'm wondering if we should update the logic here to exclude kubeproxy nodeup tasks when Cilium.EnableNodePort=true:

https://github.com/kubernetes/kops/blob/cbee4dccbaa1a6c0490a671500db921c34b521b7/upup/pkg/fi/nodeup/command.go#L252-L254

olemarkus:
Without this PR, enableNodePort doesn't work at all even though it was possible to set that flag. 

My first idea was actually to drop the kubeProxy task like kuberouter does, but I think it is better to make the user do the right thing instead of just disabling it explicitly. The first thing that happens in the kubeProxy task is to check the kubeProxy.enabled flag and return if it is false.

rifelpet:
I agree forcing the user to explicitly set kubeProxy.enabled=false is more clear. Maybe its worth mentioning both enableNodePort and its kubeProxy.enabled requirement in the docs? https://github.com/kubernetes/kops/blob/master/docs/networking.md#cilium-example-for-cni-and-network-policy

olemarkus:
I want to create documentation both for this feature and for #8316. Just need to see how these play out first in case I need to do more changes.

rifelpet:
ok great, thank you!

/lgtm
/approve

mazzy89:
asta la vista `kube-proxy` whoa

PR #8221, opened 2019-12-29
Move "docker-healthcheck" to DockerBuilder

hakman:
The Docker health check should be running only when the container runtime is Docker. At the moment it is also triggered for containerd.

This change moves the `docker-healthcheck` script, service and timer to DockerBuilder.

Based on Office Hours discussion, the health check will be enabled through a new config option:
```
spec:
  docker:
    healthCheck: true
```


hakman:
@justinsb did you get a chance ask around if this is still needed or if there is a better place for it in the Kops code?

justinsb:
The guidance I got was that much of the functionality has moved to Node Problem Detector and we likely don't _need_ the healthcheck itself any more.  So I propose we get this in so we stop using it at least for containerd, , and maybe also add NPD support.

Given release timings I'm not sure if / how far we want to backport this.  Probably not 1.16, maybe 1.17 (but maybe not).

We then have a choice: we can either actively stop using the healthcheck with docker, or we can nudge people towards containerd, effectively stopping the healthcheck in that way.

hakman:
@justinsb I added some release notes for this. Seems something people should read about before upgrading. I don't have any plans to backport this to any release branch, unless you think it would be good to have it in 1.17 also.

Support for Node Problem Detector seems interesting and something we should support. Will look into it.

Not sure if I understand the last part of your last comment. Do you think I should change the default for the health check to be enabled for Docker?

hakman:
@justinsb the last commit keeps the Docker "health-check" enabled for older versions of Kubernetes running on Debian distros (like the original behaviour). I think this should address the remaining concerns.
https://github.com/kubernetes/kops/pull/8221/files#diff-881677b9520c8452a2d4ce32dc41fdaaR1090

docs/releases/1.18-NOTES.md
```diff
@@ -6,6 +6,8 @@
 
 * Terraform users on AWS may need to rename some resources in their state file in order to prepare for Terraform 0.12 support. See Required Actions below.
 
+* The Docker `health-check` service is now disabled by default, it should not be needed anymore. See Required Actions below on how to enable.
```
johngmyers:
```suggestion
* The Docker `health-check` service is now disabled by default; it should not be needed anymore. See Required Actions below for how to enable.
```

johngmyers:
@k8s-ci-robot this needs rebase

hakman:
Will wait for approval before rebasing this. Any release notes change will mess it up again.

pkg/apis/kops/dockerconfig.go
```diff
@@ -34,6 +34,8 @@ type DockerConfig struct {
 	ExecRoot *string `json:"execRoot,omitempty" flag:"exec-root"`
 	// Experimental features permits enabling new features such as dockerd metrics
 	Experimental *bool `json:"experimental,omitempty" flag:"experimental"`
+	// HealthCheck enables the periodic health-check service
```
johngmyers: nit:
```suggestion
	// HealthCheck enables the periodic health-check service.
```

justinsb:
Thanks @hakman for this, and sorry it took so long to approve

/approve
/lgtm

Looks like we might need a test-output regeneration though, I'm afraid.

hakman:
Thank you @justinsb. With this, the support for containerd should be pretty much done.

PR #8225, opened 2019-12-30
staticcheck:remove duplicate import packages

yuxiaobo96:
Delete a package imported twice, for example:
```shell
metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

### DIFF
--- a/pkg/apis/kops/channel.go
+++ b/pkg/apis/kops/channel.go
@@ -22,7 +22,6 @@ import (
 
 	"github.com/blang/semver"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 	"k8s.io/kops/pkg/apis/kops/util"
 	"k8s.io/kops/util/pkg/vfs"
@@ -36,8 +35,8 @@ const (
 )
 
 type Channel struct {
-	v1.TypeMeta `json:",inline"`
-	ObjectMeta  metav1.ObjectMeta `json:"metadata,omitempty"`
+	metav1.TypeMeta `json:",inline"`
+	ObjectMeta      metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec ChannelSpec `json:"spec,omitempty"`
 }
--- a/pkg/client/simple/vfsclientset/instancegroup.go
+++ b/pkg/client/simple/vfsclientset/instancegroup.go
@@ -27,8 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/klog"
-	"k8s.io/kops/pkg/apis/kops"
-	api "k8s.io/kops/pkg/apis/kops"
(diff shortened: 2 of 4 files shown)


pkg/apis/kops/channel.go
```diff
@@ -22,7 +22,6 @@ import (
 
 	"github.com/blang/semver"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
tanjunchen: good catch

tanjunchen:
/retest
ref:https://github.com/kubernetes/kops/issues/7800

hakman:
@yuxiaobo96 please try and run `make gofmt`. It should fix the formatting.

yuxiaobo96:
@hakman 
Thanks for your advice, I have a question.
When I run `./hack/verify-gofmt.sh`:
```shell
xargs: bazel: without that file or directory
```
I don't understand why.

hakman:
Do you have `bazel` installed?

yuxiaobo96:
@hakman 
Thank you, I will install in a while.

pkg/client/simple/vfsclientset/instancegroup.go
```diff
@@ -27,7 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/klog"
-	"k8s.io/kops/pkg/apis/kops"
```
tanjunchen: kopsapi "k8s.io/kops/pkg/apis/kops" i see the alias in the other packages , so recommend kopsapi  i think , but it is ok.

rifelpet: I agree, `api` is too vague so if we switched the `k8s.io/kops/pkg/apis/kops` import in this file and `cloud_instance_group.go` to use `kopsapi` that could be more informative.

yuxiaobo96: I am going to update.

pkg/apis/kops/channel.go
```diff
@@ -36,7 +35,7 @@ const (
 )
 
 type Channel struct {
-	v1.TypeMeta `json:",inline"`
+	metav1.TypeMeta `json:",inline"`
 	ObjectMeta  metav1.ObjectMeta `json:"metadata,omitempty"`
```
hakman: This is the reason `verify-gofmt` fails, can you update?
```suggestion
	ObjectMeta      metav1.ObjectMeta `json:"metadata,omitempty"`
```

yuxiaobo96: You are great.

hakman: Np. 5 spaces to align with `metav1.TypeMeta`.

hakman: OK, tests are passing now. Please squash the commits and force push.

yuxiaobo96: OK.

rifelpet:
Thanks for the improvement!
/approve
/lgtm

PR #8232, opened 2019-12-30
Add code simplifications for staticheck

hakman:
Ref: #7800

This change adds back code simplification checks.
It also makes the style rules we are breaking explicit, to prevent other types of style checks to break and to make it easier to fix style these later if wanted.

```
Errors from staticcheck:
cmd/kops/toolbox_template.go:138:38: should use make(map[string]string) instead (S1019)
cmd/kops/toolbox_template.go:215:42: should use make(map[string]interface{}) instead (S1019)
cmd/kops/toolbox_template.go:228:40: should use make(map[string]interface{}) instead (S1019)
dns-controller/pkg/watchers/ingress.go:170:4: should merge variable declaration with assignment on next line (S1021)
dns-controller/pkg/watchers/service.go:206:5: should merge variable declaration with assignment on next line (S1021)
upup/pkg/fi/cloudup/alitasks/vpc.go:64:5: should omit nil check; len() for nil slices is defined as zero (S1009)
upup/pkg/fi/cloudup/openstacktasks/volume.go:73:2: unnecessary guard around call to delete (S1033)
upup/pkg/fi/cloudup/openstacktasks/volume.go:76:2: unnecessary guard around call to delete (S1033)
```

### DIFF
--- a/cmd/kops/toolbox_template.go
+++ b/cmd/kops/toolbox_template.go
@@ -135,7 +135,7 @@ func runToolBoxTemplate(f *util.Factory, out io.Writer, options *toolboxTemplate
 		templates = append(templates, list...)
 	}
 
-	snippets := make(map[string]string, 0)
+	snippets := make(map[string]string)
 	for _, x := range options.snippetsPath {
 		list, err := expandFiles(utils.ExpandPath(x))
 		if err != nil {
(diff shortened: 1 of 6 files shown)


rifelpet:
/ok-to-test

Nice work on enumerating the ignored checks! I also noticed that more recent versions of staticcheck have additional checks that kops fails, so it might be worth upgrading our vendored version at some point too.

hakman:
That would be nice. We should give it a try once the ignored failures list is gone.

cmd/kops/toolbox_template.go
```diff
@@ -135,7 +135,7 @@ func runToolBoxTemplate(f *util.Factory, out io.Writer, options *toolboxTemplate
 		templates = append(templates, list...)
 	}
 
-	snippets := make(map[string]string, 0)
```
tanjunchen: it seems that you repeat with me , but you first . i will rebase . hehe

tanjunchen:
good job!

PR #8237, opened 2019-12-31
[DigitalOcean] Add load balancer support for master HA

srikiz:
This PR adds load balancer support for digital ocean cloud provider.

- It would add a new load balancer based on a cluster master tag. 

- Incorporated the new load balancer address as part of the kuber config file. 

- Tested with 3 masters and a load balancer and was able to access the kubernetes cluster via the load balancer address from outside.

- Deleting the cluster also deletes the load balancer.

- Created a new cluster with the below command 

`./kops create cluster --cloud=digitalocean --name=dev5.k8s.local --networking=cilium --api-loadbalancer-type=public --master-count=3 --networking=cilium --zones=tor1 --ssh-public-key=~/.ssh/id_rsa.pub -v 10 --yes`

Will be creating a new PR to add documentation for DO's support to multi-master with gossip based clustering.


srikiz:
@timoreimann  - Please review when you get a chance, thanks !

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,201 @@
...
+	klog.V(10).Infof("Find IP address for load balancer ID=%s", fi.StringValue(lb.ID))
+	loadBalancer, _, err := loadBalancerService.Get(context.TODO(), fi.StringValue(lb.ID))
+	if err != nil {
+		klog.Errorf("Error fetching load balancer with Name=%s", fi.StringValue(lb.Name))
+		return nil, err
+	}
+
+	address := loadBalancer.IP
+
+	if len(address) > 0 && is_ipv4(address) {
```
timoreimann: I think you can drop the `len(address) > 0` check part, it's implicitly checked by `is_ipv4`.

pkg/model/domodel/api_loadbalancer.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (b *APILoadBalancerModelBuilder) Build(c *fi.ModelBuilderContext) error {
+	// Configuration where a load balancer fronts the API
+	if !b.UseLoadBalancerForAPI() {
+		return nil
+	}
+
+	lbSpec := b.Cluster.Spec.API.LoadBalancer
+	if lbSpec == nil {
+		// Skipping API ELB creation; not requested in Spec
```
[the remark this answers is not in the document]
srikiz: sure, will do.

pkg/model/domodel/api_loadbalancer.go
```diff
@@ -0,0 +1,93 @@
...
+	switch lbSpec.Type {
+	case kops.LoadBalancerTypeInternal:
+		// OK
+	case kops.LoadBalancerTypePublic:
+		// OK
+	default:
+		return fmt.Errorf("unhandled LoadBalancer type %q", lbSpec.Type)
+	}
+
+	loadbalancerName := "api-" + strings.Replace(b.ClusterName(), ".", "-", -1)
```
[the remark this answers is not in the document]
srikiz: Sure, done.

pkg/model/domodel/api_loadbalancer.go
```diff
@@ -0,0 +1,93 @@
...
+	clusterMasterTag := do.TagKubernetesClusterMasterPrefix + ":" + strings.Replace(b.ClusterName(), ".", "-", -1)
+
+	// Create LoadBalancer for API ELB
+	var loadbalancer *dotasks.LoadBalancer
+	{
```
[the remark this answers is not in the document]
srikiz: Sure, updated.

pkg/resources/digitalocean/cloud.go
```diff
@@ -128,7 +129,46 @@ func (c *Cloud) Droplets() godo.DropletsService {
...
 	return nil, errors.New("not implemented")
 }
+
+func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressStatus, error) {
+	var ingresses []kops.ApiIngressStatus
+	if cluster.Spec.MasterPublicName != "" {
+		// Note that this must match Digital Ocean's lb name
+		klog.V(2).Infof("Querying DO to find Loadbalancers for API (%q)", cluster.Name)
+
+		loadBalaners, _, err := c.LoadBalancers().List(context.TODO(), nil)
```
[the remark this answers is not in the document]
srikiz: yeah, that was the godo api to retrieve all the load balancers, so needed paging.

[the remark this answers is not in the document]
srikiz: fixed typo as well.

pkg/resources/digitalocean/cloud.go
```diff
@@ -128,7 +129,46 @@ func (c *Cloud) Droplets() godo.DropletsService {
...
+
+		if err != nil {
+			klog.Errorf("LoadBalancers.List returned error: %v", err)
```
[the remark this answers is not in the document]
srikiz: Yeah, caller is eventually logging the error. So, I have changed this to use fmt.Errorf and removed logging here. Thanks for pointing this out @timoreimann  👍

pkg/resources/digitalocean/cloud.go
```diff
@@ -128,7 +129,46 @@ func (c *Cloud) Droplets() godo.DropletsService {
...
+		}
+
+		lbName := "api-" + strings.Replace(cluster.Name, ".", "-", -1)
+
+		for _, lb := range loadBalaners {
+			if lb.Name == lbName {
+				klog.V(10).Infof("Matching LB name found for API (%q)", cluster.Name)
+
+				if lb.Status != "active" {
+					klog.Errorf("load-balancer is not yet active (current status: %s)", lb.Status)
```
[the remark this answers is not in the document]
srikiz: sure, updated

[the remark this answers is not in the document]
srikiz: I changed this to use fmt.Errorf instead.

pkg/resources/digitalocean/resources.go
```diff
@@ -315,6 +378,18 @@ func deleteRecord(cloud fi.Cloud, domain string, t *resources.Resource) error {
 	return nil
 }
 
+func deleteLoadBalancer(cloud fi.Cloud, t *resources.Resource) error {
+	c := cloud.(*Cloud)
+	lb := t.Obj.(godo.LoadBalancer)
+	_, err := c.Client.LoadBalancers.Delete(context.TODO(), lb.ID)
+
+	if err != nil {
+		return fmt.Errorf("failed to delete load balancer with name %s", lb.Name)
```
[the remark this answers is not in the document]
srikiz: updated.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
srikiz: Done.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,201 @@
...
+		klog.V(10).Infof("load balancer address=%s", address)
+		return &address, nil
+	}
+
+	klog.Errorf("Error fetching IP Address for lb Name=%s", fi.StringValue(lb.Name))
+	time.Sleep(10 * 3)
```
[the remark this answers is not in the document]
srikiz: changed that to 10 seconds with time.Sleep(10 * time.Second)

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,201 @@
...
+
+	klog.V(10).Infof("Sleeping for 30 seconds")
+
+	return nil, errors.New("IP Address is still empty.")
+}
+
+func is_ipv4(host string) bool {
```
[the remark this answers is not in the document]
srikiz: sure, updated.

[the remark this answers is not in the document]
srikiz: Done.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (lb *LoadBalancer) CompareWithID() *string {
+	return lb.ID
+}
+
+func (lb *LoadBalancer) Find(c *fi.Context) (*LoadBalancer, error) {
+	cloud := c.Cloud.(*digitalocean.Cloud)
+	lbService := cloud.LoadBalancers()
+
+	opt := &godo.ListOptions{}
```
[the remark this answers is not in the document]
srikiz: I realized we can use LB ID as you suggested below.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,201 @@
...
+
+	loadbalancers, _, err := lbService.List(context.TODO(), opt)
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, loadbalancer := range loadbalancers {
+		if loadbalancer.Name == fi.StringValue(lb.Name) {
```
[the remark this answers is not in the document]
srikiz: This really helps, thanks for letting me know. Changed the logic to use LB ID instead.

srikiz:
@timoreimann - I have incorporated all the comments. Please have a look when you get a chance. Thanks !

pkg/resources/digitalocean/cloud.go
```diff
@@ -128,7 +129,44 @@ func (c *Cloud) Droplets() godo.DropletsService {
...
 	return nil, errors.New("not implemented")
 }
+
+func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressStatus, error) {
+	var ingresses []kops.ApiIngressStatus
+	if cluster.Spec.MasterPublicName != "" {
+		// Note that this must match Digital Ocean's lb name
+		klog.V(2).Infof("Querying DO to find Loadbalancers for API (%q)", cluster.Name)
+
+		loadBalancers, _, err := c.LoadBalancers().List(context.TODO(), nil)
```
timoreimann: Does this need paging?

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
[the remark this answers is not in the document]
timoreimann: We should handle the case where the names don't match, e.g., logging.

I'm not sure I understand why we need the extra check though -- can we think of a case where the ID alone is insufficient?

pkg/resources/digitalocean/cloud.go
srikiz: yeah, I missed that. Updated it.

srikiz:
@timoreimann - I have incorporated the changes.
Please have a look when you get a chance.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,188 @@
...
+	lbService := cloud.LoadBalancers()
+
+	if fi.StringValue(lb.ID) != "" {
+		loadbalancer, _, err := lbService.Get(context.TODO(), fi.StringValue(lb.ID))
+
+		if err != nil {
+			return nil, fmt.Errorf("load balancer service get request returned error %v", err)
+		}
+
+		// do another check to double-sure we are talking to the right load balancer.
```
timoreimann: This comment looks like it's obsolete by now?

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,188 @@
...
+		return &LoadBalancer{
+			Name:      fi.String(loadbalancer.Name),
+			ID:        fi.String(loadbalancer.ID),
+			Lifecycle: lb.Lifecycle,
+			Region:    fi.String(loadbalancer.Region.Slug),
+		}, nil
+	}
+
+	// Loadbalancer = nil if not found
+	return nil, nil
```
timoreimann: nit: if we do

```go
if fi.StringValue(lb.ID) == "" {
  // Loadbalancer = nil if not found
  return nil, nil
}

cloud := c.Cloud.(*digitalocean.Cloud)
lbService := cloud.LoadBalancers()
loadbalancer, _, err := lbService.Get(context.TODO(), fi.StringValue(lb.ID))
[...]
```

then we do not need to have the intended if-branch for the case where the LB ID exists. The code also seems easier to read because the check for empty LB ID and the action we take are not so far apart anymore.

(I also moved the `cloud` and `lbService` variables below since they are not needed unless we have an ID.)

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return nil
+}
+
+func (lb *LoadBalancer) FindIPAddress(c *fi.Context) (*string, error) {
+	cloud := c.Cloud.(*digitalocean.Cloud)
+	loadBalancerService := cloud.LoadBalancers()
+
+	klog.V(10).Infof("Find IP address for load balancer ID=%s", fi.StringValue(lb.ID))
+	loadBalancer, _, err := loadBalancerService.Get(context.TODO(), fi.StringValue(lb.ID))
```
timoreimann: Are we certain at this point that we have an LB ID? Genuine question as I don't know the kops work flow in this context that well.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	address := loadBalancer.IP
+
+	if isIPv4(address) {
+		klog.V(10).Infof("load balancer address=%s", address)
+		return &address, nil
+	}
+
+	klog.Errorf("Error fetching IP Address for lb Name=%s", fi.StringValue(lb.Name))
```
timoreimann: Can we match the description with the error below and mention that we are sleeping so that users aren't surprised by the delay, e.g.:

```
const lbWaitTime = 10 * time.Second
klog.Errorf("IP address for LB %s not yet available -- sleeping %s", fi.StringValue(lb.Name), lbWaitTime)
```

I also extracted the wait time into a const so that we can reuse it in the actual `time.Sleep` call.

timoreimann: Should we do a `klog.Warnf` here? The delay should be expected.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,188 @@
...
+	time.Sleep(10 * time.Second)
+
+	klog.V(10).Infof("Sleeping for 10 seconds")
```
timoreimann: I'd drop this one given my improvement suggestion for the message above.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -0,0 +1,188 @@
...
+
+	ip := net.ParseIP(host)
+	if ip == nil {
+		return false
+	}
+
+	return ip.To4() != nil
+}
+
+// terraformVolume represents the digitalocean_volume resource in terraform
```
timoreimann: This one looks like a trailing comment that shouldn't be here?

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
srikiz: Done.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
srikiz: sure, that's a good suggestion. Updated accordingly.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
srikiz: Yes @timoreimann - we are certain we have an LB ID, but not certain about the LB IP Address. FindAddress is only called after RenderDO is invoked - we create a new LB Request in RenderDO that returns a LB ID. We use this LB ID to find if the IP Address is available.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
srikiz: sure, updated.

srikiz: done.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
srikiz: sure, done.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
srikiz: sure, updated.

pkg/model/domodel/api_loadbalancer.go
```diff
@@ -0,0 +1,89 @@
...
+	loadbalancer := &dotasks.LoadBalancer{
+		Name:       fi.String(loadbalancerName),
+		Region:     fi.String(b.Cluster.Spec.Subnets[0].Region),
+		DropletTag: fi.String(clusterMasterTag),
+		Lifecycle:  b.Lifecycle,
+	}
+	c.AddTask(loadbalancer)
+
+	// Temporarily do not know the role of the following function
+	if dns.IsGossipHostname(b.Cluster.Name) || b.UsePrivateDNS() {
```
rifelpet: I'm not a fan of this strategy either but I confirmed this same method is used for the other cloud providers as well: https://github.com/kubernetes/kops/blob/d70cc9e17faedd0190657f682de7987957e79741/pkg/model/awsmodel/api_loadbalancer.go#L272-L282

rifelpet:
The E2E failures are unrelated, so we can move forward with this regardless. Thanks!

/lgtm
/approve

PR #8241, opened 2020-01-01
Update copyrights for 2020

johngmyers:


tanjunchen:
/ok-to-test
why not consider update all copyright?

johngmyers:
I don't understand the question. This is the result of `make apimachinery`, needed for anything to pass CI.

tanjunchen:
> I don't understand the question. This is the result of `make apimachinery`, needed for anything to pass CI.

ok i get it

hakman:
@johngmyers I think this should go also to 1.15. I bet there would be at least one more release for it soon.

PR #8243, opened 2020-01-01
Fix aws-china.md to download SHA-256 checksums for kops assets

### LINKED ISSUE #8242: kops compares sha1 with sha256 while checking hash of nodeup
**1. What `kops` version are you running?**

Version 1.15.0

**2. What Kubernetes version are you running?**

v1.15.7

AWS

I created my cluster with --kubernetes-version according to the doc [How to use kops in AWS China Region](https://github.com/kubernetes/kops/blob/master/docs/aws-china.md). 
**Note that I only downloaded the "$url.sha1" file, not the sha256 file.**

kops generated user-data to initialize the master nodes, but compared  sha1 with sha256 while checking hash of the file "nodeup". So the initialization couldn't be done.

**6. The nodes `journalctl` logs**

    == nodeup corrupted, hash 9604ef18267ad7b5cf4cebbf7ab64423cf5bb0342d169c608ac6376e6af26d81 doesn't match expected f22d740fdbe2fc92e9b359bf23c5b6d7caa0aba9 ==
    == Hash validation of https://s3.cn-north-1.amazonaws.com.cn/my_bucket/kops/1.15.0/linux/amd64/nodeup failed. Retrying. ==
    All downloads failed; sleeping before retrying

**7. Anything else do we need to know?**

I notice that the method `findHash` in `pkg/assets/builder.go` will try ".sha256" first, then ".sha1".
It means that we use sha1 when couldn't get sha256.

    for _, ext := range []string{".sha256", ".sha1"} {
        hashURL := u.String() + ext
        b, err := vfs.Context.ReadFile(hashURL, vfs.WithBackoff(backoff))
        if err != nil {
            // Try to log without being too alarming - issue #7550
            if ext == ".sha256" {
                klog.V(2).Infof("unable to read new sha256 hash file (is this an older/unsupported kubernetes release?) %q: %v", hashURL, err)
            } else {
                klog.V(2).Infof("unable to read hash file %q: %v", hashURL, err)
            }
            continue
        }
        hashString := strings.TrimSpace(string(b))
        klog.V(2).Infof("Found hash %q for %q", hashString, u)

        // Accept a hash string that is `<hash> <filename>`
        fields := strings.Fields(hashString)
        if len(fields) == 0 {
            klog.Infof("hash file was empty %q", hashURL)
            continue
        }
        return hashing.FromString(fields[0])
    }

**But the `NodeUpTemplate` in `pkg/model/resources/nodeup.go` looks like below.**
**It just uses `sha256sum`.**

    validate-hash() {
      local -r file="$1"
      local -r expected="$2"
      local actual

      actual=$(sha256sum ${file} | awk '{ print $1 }') || true
      if [[ "${actual}" != "${expected}" ]]; then
        echo "== ${file} corrupted, hash ${actual} doesn't match expected ${expected} =="
        return 1
      fi
    }

**So, would it be better to do like this in `NodeUpTemplate`? (length of sha1 is 40.)**

    validate-hash() {
      local -r file="$1"
      local -r expected="$2"
      local actual
      local hash_cmd=sha256sum

      if [[ ${#expected} == 40 ]]; then
        hash_cmd=sha1sum
      fi

      actual=$(${hash_cmd} ${file} | awk '{ print $1 }') || true
      if [[ "${actual}" != "${expected}" ]]; then
        echo "== ${file} corrupted, hash ${actual} doesn't match expected ${expected} =="
        return 1
      fi
    }

johngmyers:
Since kops 1.15.0, nodeup verifies kops assets against the `.sha256` file.

Fixes #8242

### DIFF
--- a/docs/aws-china.md
+++ b/docs/aws-china.md
@@ -249,7 +249,7 @@ for asset in "${KOPS_ASSETS[@]}"; do
   mkdir -p "$dir"
   url="https://kubeupv2.s3.amazonaws.com/kops/$KOPS_VERSION/$asset"
   wget -P "$dir" "$url"
-  wget -P "$dir" "$url.sha1"
+  wget -P "$dir" "$url.sha256"
 done
 
 ## Upload assets


rifelpet:
Thanks!
/approve
/lgtm

rifelpet:
hm so much for the `retest-not-required-docs-only` label...

PR #6897, opened 2019-05-08
Set priority for static pods

vainu-arto:
Setting a high priority for kube-proxy is important in clusters that run with PodPriority enabled, otherwise it will get evicted from worker nodes whenever there are unschedulable pods. Also set priorities for the other static pods defined by kops, mainly for completeness.

This should be safe these days since the API always accepts PriorityClassName even if PodPriority isn't enabled.

Fixes #6615.

PR #8248, opened 2020-01-02
Fix issues with older versions of k8s for basic clusters

hakman:
Currently, Kops is not able to start a basic cluster with Kubernetes < 1.12 for a few reasons:
* PodPriority feature gate is needed for various manifests with Kubernetes < 1.11
* the kube-dns manifest cannot be applied because of an empty resource `{}`

### DIFF
--- a/docs/releases/1.18-NOTES.md
+++ b/docs/releases/1.18-NOTES.md
@@ -6,6 +6,8 @@
 
 * Terraform users on AWS may need to rename some resources in their state file in order to prepare for Terraform 0.12 support. See Required Actions below.
 
+* Kubernetes 1.9 users will need to enable the PodPriority feature gate. See Required Actions below.
+
 * A controller is now used to apply labels to nodes.  If you are not using AWS, GCE or OpenStack your (non-master) nodes may not have labels applied correctly.
 
 * Please see the notes in the 1.15 release about the apiGroup changing from kops to kops.k8s.io
@@ -34,6 +36,18 @@
   terraform apply
   ```
 
+* Kubernetes 1.9 users will need to enable the PodPriority feature gate. This is required for newer versions of Kops.
+
+  To enable the Pod priority feature, follow these steps:
+  ```
+  kops edit cluster
+  # Add the following section
+  spec:
+    kubelet:
+      featureGates:
+        PodPriority: "true"
+  ```
+ 
 * If a custom Kops build was used on a cluster, a kops-controller Deployment may have been created that should get deleted.
   Run `kubectl -n kube-system delete deployment kops-controller` after upgrading to Kops 1.16.0-beta.1 or later.
 
--- a/upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io/k8s-1.6.yaml.template
+++ b/upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io/k8s-1.6.yaml.template
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 {{- if or (.KubeDNS.UpstreamNameservers) (.KubeDNS.StubDomains) }}
----
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -28,9 +27,9 @@ data:
   stubDomains: |
     {{ ToJSON .KubeDNS.StubDomains }}
   {{- end }}
-{{- end }}
 
 ---
+{{- end }}
 
 apiVersion: extensions/v1beta1
 kind: Deployment


johngmyers:
I was able to create a validated cluster with Kubernetes 1.10.13. 1.9.11 did not come up.

If it's been broken with nobody complaining, I'm not sure the value of fixing it. Is it broken in kops 1.15?

hakman:
@johngmyers anything 1.12+ should be ok. The value is only for minimum supported version. If we want to say 1.9+, these should be fixed. Otherwise, minimum supported version should be 1.12.

johngmyers:
Appears to have been broken in kops 1.15.0 per [this commit](https://github.com/kubernetes/kops/commit/48974521e19e9b50513c530dfcece064ec36dad9)

johngmyers:
The change is minimal enough. No objections.

rifelpet:
I'd like to discuss this at office hours today. It seems that our two options are to either fix the broken support for these older k8s versions or update our deprecation policy to drop those versions as well.

hakman:
Agreed @rifelpet. In any case, best to know what the blockers are first.
The broken template affects also 1.11. Without the fix support should be 1.12+.

hakman:
With @rifelpet's help, e2e tests for older versions are up and running. Tweaks are still needed, but we can see what versions at least come up `e2e-kops-aws-k8s-1.x`:
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#Summary

In [some 1.11 jobs](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-1-11/1213731395319369731/artifacts/ip-172-20-49-96.ap-southeast-2.compute.internal/protokube.log), the clusters are up enough to run some tests, but `kube-dns` cannot be installed by Protokube:
```
NAME			CURRENT	UPDATE
kube-dns.addons.k8s.io	-	1.14.13-kops.1
I0105 08:53:37.834943   28975 addon.go:139] Applying update from "s3://k8s-kops-prow/e2e-kops-aws-k8s-1.11.test-cncf-aws.k8s.io/addons/kube-dns.addons.k8s.io/k8s-1.6.yaml"
I0105 08:53:37.834984   28975 s3fs.go:219] Reading file "s3://k8s-kops-prow/e2e-kops-aws-k8s-1.11.test-cncf-aws.k8s.io/addons/kube-dns.addons.k8s.io/k8s-1.6.yaml"
I0105 08:53:37.981011   28975 apply.go:67] Running command: kubectl apply -f /tmp/channel407381304/manifest.yaml
I0105 08:53:38.126740   28975 apply.go:70] error running kubectl apply -f /tmp/channel407381304/manifest.yaml
I0105 08:53:38.126766   28975 apply.go:71] error: error validating "/tmp/channel407381304/manifest.yaml": error validating data: [apiVersion not set, kind not set]; if you choose to ignore these errors, turn validation off with --validate=false
Error: error updating "kube-dns.addons.k8s.io": error applying update from "s3://k8s-kops-prow/e2e-kops-aws-k8s-1.11.test-cncf-aws.k8s.io/addons/kube-dns.addons.k8s.io/k8s-1.6.yaml": error running kubectl
```

For some reason the tests don't validate the cluster before preceding, maybe we should add this instead of the simple `Node Ready` check.

hakman:
Based on discussions we had at last office hours, assigning to Justin to check why PodPriority is needed.
/assign @justinsb

johngmyers:
We need `priorityClassName` because support for the `scheduler.alpha.kubernetes.io/critical-pod` annotation was removed in Kubernetes 1.16. It wouldn't be good having things like kubelet be evicted in order to make room for other pods.

johngmyers:
The PR for the aforementioned commit was #6897

justinsb:
So I dug into this.  As @hakman pointed out there are two things happening:

* The empty blocks mapping to `{}` is a bug, though one that only affects manifests that have empty objects.  So I like the idea of fixing it specifically for kube-dns here, and I'm trying to fix it in general in #8317 so that we'll just skip over them.  kubectl normally tolerates an empty section, but the `{}` is an artifact of our remapping process and kubectl doesn't like it.

* The pod priority class issue was indeed introduced through a series of commits, and at least in 1.9 we get the error `spec.priorityClassName: Forbidden: Pod priority is disabled by feature-gate` - normally kubernetes ignores unknown fields, so this is a little inconsistent, but I think @johngmyers has identified the sequence of PRs that led to this being broken.

I think we now have more testing for older versions (thanks @rifelpet and @hakman), and we understand how this happened.

The other option to the fix in this PR is to _not_ set the PodPriorityClass for the older versions of k8s (before 1.11) - i.e. either skip MarkPodAsClusterCritical or pass in the cluster and have it be a no-op for older clusters versions.  As we're likely dropping support for < 1.11 very soon, this would only matter for a cherry pick of this PR; arguably it is more consistent with our policy of not changing existing versions.  But OTOH it's already broken...

johngmyers:
I don't recall our deprecation policy as intending to take on a new requirement to fix preexisting brokenness in handling ancient versions. I don't object to taking a fix as minimal as this PR, but I would not want to spend the effort fixing, say, kops 1.16's handling of k8s 1.6. I think adding logic to make the setting of podPriorityClass conditional on version (and presumably whether the feature gate is enabled) would be a bit much, especially since we haven't yet gotten an actual user complaint.

hakman:
In theory, checking the k8s version before using MarkPodAsClusterCritical() should be pretty simple.

In practice, for etcd is not so trivial. I can't find a simple way to check the Kubernetes version here:
https://github.com/kubernetes/kops/blob/ff29cab20b868d5fe38d7cf10602fec2da07222a/protokube/pkg/protokube/etcd_manifest.go#L168

Unless there is something I'm missing, the feature gate is the simplest way to get things working again for 1.9 and 1.10.

hakman:
@justinsb removed the PodPriority part from this, will create a separate PR for explaining that 1.9 requires PodPriority feature gate.

hakman:
Release notes added for PodPriority feature gate requirement for Kubernetes 1.9. I think this is now in line with what we discussed in office hours.

PR #8254, opened 2020-01-03
Run Travis verifications in a separate parallel job

johngmyers:


johngmyers:
Also, can we remove the osx 1.12 job? I think one osx job would be sufficient for our needs.

geojaz:
/ok-to-test
(why aren't your PRs already being auto-tested... ? we need to get you added to that list)

PR #8021, opened 2019-11-27
Support tainting all nodes needing update during rolling update

johngmyers:
Adds a per-instancegroup option to taint all nodes needing update near the start of a rolling update. The expectation is that this would only be enabled on instancegroups which have autoscaling enabled and which have workloads which can tolerate waiting for scale-up if needed.

Extends the InstanceGroup API to support configuration of the rolling update strategy.  Extends the Cluster API to support per-cluster defaults for same. The configuration options are behind the new `ConfigurableRollingUpdate` feature flag, which defaults to off.

In order to limit the damage should the new instance specification produce instances which fail validation, in the case where there are no existing instances with the current spec the strategy will first cordon and update a single instance, waiting until its replacement validates before tainting the rest.

Fixes #7958

PR #8255, opened 2020-01-03
Adds johngmyers to reviewers because he's been awesome

geojaz:


geojaz:
@justinsb how can we get johngmyers into the k8s org? IMO, he's a shoe-in.

johngmyers:
Per [the requirements](https://github.com/kubernetes/community/blob/master/community-membership.md), reviewers are required to have been members for at least three months, so I am not eligible.

I have been waiting for #8021 to land before requesting membership.

geojaz:
IMO, you deserve k8s org membership, given your intensely active involvement over the past few months. I would be happy to see you join our OWNERS file and org. To that end, i'll take a pass at #8021 for ya

geojaz:
@rifelpet or @justinsb can we :+1: this yet? @johngmyers has joined the org and continues to be cool

johngmyers:
The issue here is the semantics of the reviewers list. As a member, I'm already able to lgtm PRs.

The reviewers list is the set of people in line to be automatically assigned to new PRs. Unlike what many people would assume, approvers are not automatically reviewers. Our current list contains two of our relatively less active members. If I were to be added, I would be assigned to and receive notifications for 2/3 of all incoming PRs, which I would not find helpful.

To be added, I would like some strict subset of approvers to also be added at the same time. Or we could go the other way and reduce the set to zero and work off of search queries. (I work off of `sort:updated-desc` and Gubernator.

The other thing I find frustrating is that the PRs that I spend the time to go through the multiple rounds and lgtm seem to languish just as much as the ones where I don't. I think it would be helpful if some approvers would keep a query of lgtm'ed unapproved PRs and give them priority. Those PRs should be approvable with less time spent.

geojaz:
Fair enough- my inbox and I definitely feel that pain and as a result, I also basically just use gubernator (when I'm able to contribute). 

I hadn't thought as thoroughly as you about the implications at the time. My goal was to empower you to put your insightful voice into action. At the time, I wanted you to be able to move code forward, and you couldn't without some approved person checking the box. 
I was happily surprised today when your `/lgtm` pushed my vendor PR through.
 
@johngmyers what do  _you_ want? Shall I close this? Do you want to be an approver? Or are you "happy" as is? (happy as in are you good with your ability to work on our project in github ;) This was for you, and for the project, but it looks like you may already have been unblocked.

johngmyers:
I would like us to review our code review procedures and change the configuration of the tools we have at hand. I think there are ways we could more efficiently use the contributed time we have.

I'll see if I can put something together. This PR we can probably
/close

PR #8258, opened 2020-01-03
Update mock kops version in integration tests

rifelpet:
To go along with our version deprecation we should update our mocked kops versions


mikesplain:
Totally agreed, good idea.

/lgtm
/approve

PR #8259, opened 2020-01-03
WIP add cloudbuild.yaml used by image pushing postsubmit job

rifelpet:
I dont know what to put for the docker registry, so leaving this as WIP until I get an answer

It will be used by the currently-failing [postsubmit job](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-postsubmit-push-to-staging).
Following [this example](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cloudbuild.yaml).

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,21 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+timeout: 1200s
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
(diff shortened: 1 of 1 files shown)


rifelpet:
/close
it turns out we already have a `ci/postsubmit/push-to-staging/cloudbuild.yaml` I'm working my way through the postsubmit job scripts to see if we can use it or if it needs to live in the repo root.

Issue #6796, opened 2019-04-19
Nodes of mixedinstancegroups are in "NeedsUpdate" state even after rolling-update.


vivekgarg20:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Version 1.12.0-beta.2 (git-d1453d22a)
**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
GitVersion:"v1.11.1", GitCommit:"b1b29978270dc22fecc592ac55d903350454310a"
**3. What cloud provider are you using?**
AWS
**4. What commands did you run?  What is the simplest way to reproduce this issue?**
Applied mixed instance policy to one of instance group as introduced
Sample spec
```
  mixedInstancesPolicy:
    instances:
    - m5.2xlarge
    - c5.2xlarge
    - r5.2xlarge
    onDemandAboveBase: 0
    spotInstancePools: 3
```
did rolling-update --yes for instance groupl
**5. What happened after the commands executed?**
Kops nodes remained in "NeedsUpdate" state.
**6. What did you expect to happen?**
Nodes to go into "Ready" state.

**7. Please provide your cluster manifest. Execute
specs provided above.

vivekgarg20:
@gambol99 you may want to look at it. Going to test with beta.1 as well. Will update status accordingly.

granular-ryanbonham:
Can you post the exact kops commands you ran to update your cluster and to do the rolling-update. I thought we had this fixed in beta 2.

vivekgarg20:
kops update cluster --yes
kops rolling-update cluster --instance-group node-m5-1b --yes

vasilgramov:
Could this be happening when explicity we are specifying which instance groups to be rolled out?

vivekgarg20:
@vasilgramov  Nopes, tried rolling-update for full cluster, same state. Though "validate cluster" command displays ready state. It says NeedsUpdate if I fire "rolling-update cluster" again

bzuelke:
Also running into this, tested out the mixed instance types and still shows that it needs an update. Is there a potential for a cherry pick fix for this, it's a pretty decent annoyance tbh.

clifford-sanders:
We also encountered this problem. Debugging revealed that the version of the launch templated associated with the ASG is not set.

[...]

PR #8038, opened 2019-12-03
Fix Handling of LaunchTemplate Versions for MixedInstancePolicy

granular-ryanbonham:
Correctly detect the version of the LaunchTemplate that should be used, handling Default, Latest, and a specified version.. 

Currently if you are using TF output and using mixed instances policies for your IGs, without this patch in 1.14+ the rolling update always thinks nodes need to be rolled. 

TF issue caused by this PR which was to fix handling of ASG set to use default version of the Launch Template.https://github.com/kubernetes/kops/pull/7445

TF is settings specific version, and not changing the default, so the logic in 7445 broke rolling updates for TF generated ASGs.

Recommend CP to 1.14, 1.15, 1.16 with new releases.

PR #7118, opened 2019-06-07
LaunchTemplate Version Name Fix .. 

gambol99:
The commit eca2b0a27aaa706dc760ce7d2708a54aa85a6f0b added the version to the name of the which created a loop on NeedUpdate, as the empty version will never be equal to the original name.

- Revert "User version number as well as name when checking LaunchTemplates

This reverts commit eca2b0a27aaa706dc760ce7d2708a54aa85a6f0b.

PR #8261, opened 2020-01-03
Fix RollingUpdate behaviour when using LaunchTemplates for both kops & terraform spec updates

KashifSaadat:
There have been a few regressions in the way InstanceGroups are marked for a RollingUpdate when `LaunchTemplates` are involved (e.g. when using Mixed Instance Policies). This bug resulted in one of two scenarios that I've seen (with previous attempted fixes regressing one of the scenarios):
- Updates via kops (`kops update cluster --yes`) did not set a LT Version, and so an Instance Group using a Mixed Instance Policy would always be marked as `NeedsUpdate` when performing a Rolling Update
- Updates via terraform (`kops update cluster --out=tf-files/ --target=terraform`) were not being marked for a RollingUpdate as the Version was not picked up and matched between changes

This PR aims to fix the above two scenarios. I've tested a cluster build with 2 different Instance Groups (one standard and another with mixed policy), via kops updates and terraform, and both appear to have expected behaviour with marking the correct nodes for an update between changes.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
+++ b/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 
 	"k8s.io/kops/upup/pkg/fi"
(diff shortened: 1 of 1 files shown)


KashifSaadat:
- Follows on from: #8038
- Fixes: #6796
- No longer required: #7118

paalkr:
Will this be added to the 1.16 branch, and included in the next 1.16 release?

PR #8262, opened 2020-01-03
Exclude one Travis osx job

johngmyers:
We don't need it and Travis OS/X resources appear to be scarce.


justinsb:
This could probably use a comment to remind us why we're excluding it, but git history will also reveal it, so looks good!

Thanks @johngmyers 

/approve
/lgtm

PR #8265, opened 2020-01-04
Add support for weave.npcExtraArgs

### LINKED ISSUE #7390: Add WeaveNet EXTRA_ARGS support
We have multiple kubernetes clusters that import all cluster logs to a logging service. Weave-net and weave-npc generate very large volumes of logs in our environment which causes us to use up our daily quote faster than desired. 

Weave net supports modifying the log level through the `--log-level=` command line argument and allows you to pass in this argument via the `EXTRA_ARGS` environmental variable. It would be nice if we could support this feature in Kops through addition to the Weave networking spec. 

Unfortunately weave-npc does not yet support this but I have added weaveworks/weave#3683 and once that or some variant of it is merged and a release of weave-net is cut we can add the environmental variable to weave-npc as well.

- I propose we add `NetExtraArgs` to the `WeaveNetworkingSpec` and when this is present add the `EXTRA_ARGS` to the weave-net container template
- `NpcExtraArgs` would be added at a later date to do the same with `weave-npc`

ReillyProcentive:
Closes #7390

References Weaveworks/Weave#3683

This change is excluded from k8s-1.6.yaml.template as this environmental variable is only effective on Weave 2.6.0 and above.

### DIFF
--- a/pkg/apis/kops/networking.go
+++ b/pkg/apis/kops/networking.go
@@ -67,6 +67,7 @@ type WeaveNetworkingSpec struct {
 	ConnLimit    *int32 `json:"connLimit,omitempty"`
 	NoMasqLocal  *int32 `json:"noMasqLocal,omitempty"`
 	NetExtraArgs string `json:"netExtraArgs,omitempty"`
+	NpcExtraArgs string `json:"npcExtraArgs,omitempty"`
 
 	// MemoryRequest memory request of weave container. Default 200Mi
 	MemoryRequest *resource.Quantity `json:"memoryRequest,omitempty"`
--- a/pkg/apis/kops/networking.go
+++ b/pkg/apis/kops/networking.go
@@ -63,11 +63,9 @@ type KopeioNetworkingSpec struct {
 
 // WeaveNetworkingSpec declares that we want Weave networking
 type WeaveNetworkingSpec struct {
-	MTU          *int32 `json:"mtu,omitempty"`
-	ConnLimit    *int32 `json:"connLimit,omitempty"`
-	NoMasqLocal  *int32 `json:"noMasqLocal,omitempty"`
-	NetExtraArgs string `json:"netExtraArgs,omitempty"`
-	NpcExtraArgs string `json:"npcExtraArgs,omitempty"`
+	MTU         *int32 `json:"mtu,omitempty"`
+	ConnLimit   *int32 `json:"connLimit,omitempty"`
+	NoMasqLocal *int32 `json:"noMasqLocal,omitempty"`
 
 	// MemoryRequest memory request of weave container. Default 200Mi
 	MemoryRequest *resource.Quantity `json:"memoryRequest,omitempty"`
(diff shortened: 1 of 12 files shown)


mmerrill3:
For the manifest's hash, you need to update them when you change the templates.  Run /hack/update-expected.sh to update the hashes.

rifelpet:
It'd be great if you could add this and netExtraArgs to [this test](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/weave/cluster.yaml). Then run `hack/update-expected.sh` as described above.

upup/models/cloudup/resources/addons/networking.weave/k8s-1.7.yaml.template
```diff
@@ -203,6 +203,10 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: spec.nodeName
+            {{- if .Networking.Weave.NpcExtraArgs }}
+            - name: EXTRA_ARGS
+              value: "{{ .Networking.Weave.NpcExtraArgs }}"
+            {{- end }}
```
johngmyers: There is actually no need to change this file as kops 1.18 doesn't support Kubernetes 1.7.

ReillyProcentive: I was hoping that we could get this in before kops 1.18. Is that no longer a possibility?

johngmyers: I don't think Kubernetes 1.7 works on any of the recent versions of kops.

pkg/apis/kops/networking.go
```diff
@@ -76,6 +75,9 @@ type WeaveNetworkingSpec struct {
 	MemoryLimit *resource.Quantity `json:"memoryLimit,omitempty"`
 	// CPULimit CPU limit of weave container.
 	CPULimit *resource.Quantity `json:"cpuLimit,omitempty"`
+	// Extra arguments that are passed to weave-kube
```
johngmyers:
```suggestion
	// NetExtraArgs are extra arguments that are passed to weave-kube.
```

pkg/apis/kops/networking.go
```diff
@@ -84,6 +86,8 @@ type WeaveNetworkingSpec struct {
 	NPCMemoryLimit *resource.Quantity `json:"npcMemoryLimit,omitempty"`
 	// NPCCPULimit CPU limit of weave npc container
 	NPCCPULimit *resource.Quantity `json:"npcCPULimit,omitempty"`
+	// Extra arguments that are passed to weave-npc
```
johngmyers:
```suggestion
	// NPCExtraArgs are extra arguments that are passed to weave-npc.
```

rifelpet:
looks good, and the consistency with `netExtraArgs` is good. Thanks!
/lgtm
/approve

PR #8268, opened 2020-01-04
Refactor rolling update tests

johngmyers:
In preparation for the next feature.

/area rolling-update


justinsb:
Thanks @johngmyers - much easier setup!  Not sure what happened with the tests (shows a failure and a pass), I'm guessing that two tests crossed, so I'm going to approve and we can fix if needed.

/approve
/lgtm

PR #7833, opened 2019-10-28
Use /opt/cni/bin on all distros (including COS)

justinsb:
Rather than remapping the cni-bin to /home/kubernetes/bin, which then
requires us to change the CNI manifests also to write to the new
location, we can instead bind mount /opt/cni/bin so that it is
writeable.

PR #8269, opened 2020-01-04
Make /opt/kops writeable on COS

justinsb:
This approach means that rather than trying to have different
directories on different OSes, we can instead make them look more
similar.


justinsb:
/retest

Storage flake

hakman:
https://github.com/kubernetes/kops/pull/7833 seems pretty similar. Maybe can be closed after this one.

Issue #7685, opened 2019-09-26
Prevent instance group from being managed by rolling-update


dene14:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
In some cases it's necessary to manage instance group manually (for e.g. when instance store implied and some time required for node to sync up after recreation).
I'm not sure if `instanceProtection` helps with that, please let me know... and if it's not:

**2. Feel free to provide a design supporting your feature request.**
Introduce flag `skipRollingUpdate` for IG configuration defaulted to `false`. It will be operator responsibility to manage instance replacement in ASG in a graceful way.

zetaab:
I do not see benefit this feature because it do exist already. When rolling updating you can specify instancegroups for it

dene14:
This is fine as long as you're the only cluster manager or you have a small amount of clusters under your control. Things are used to be forgotten, so kind of protection that's managed in code would be useful.

johngmyers:
@justinsb Now that surging has landed, I'd like to come to a conclusion on your deferred review comment about the method I chose to disable rolling updates for an instance group.

Currently rolling updates are disabled for an instance group by setting both `maxSurge` and `maxUnavailable` to `0`. Alternatively we could make that a validation error and introduce a separate boolean option under `rollingUpdate`.

/remove-lifecycle stale

johngmyers:
With node-local storage and post-replacement sync, the solution is to use a PodDisruptionBudget. One can include an additional Deployment in the scope of the PDB's selector, providing a pod that indicates ready only when the database has adequate replication.

PR #8271, opened 2020-01-04
Option to increase concurrency of rolling update within instancegroup

### LINKED ISSUE #1718: Parallelize and improve rolling updates even more
Repeated from https://github.com/kubernetes/kops/pull/1134#issuecomment-267540660
 It looks like rolling update only parallelizes by rolling one node from each instance group concurrently. In a medium or large cluster, that's going to take a while. Rather than having to wait for an hour or temporarily expand the ASG, it'd be nice if I could specify a roll concurrency parameter in absolute number or as a percentage, similar to how a rollingUpdateStrategy of a deployment has a maxUnavailable. It's true that in a heterogeneous cluster it could simultaneously roll all your big nodes, but that's an acceptable risk to save hours of waiting.

If you want to be fancy, you could attempt to auto-detect an acceptable roll rate by making sure there are no Pending pods, but that's probably too tricky for right now.

Another good suggestion was to pre-scale ASG before rolling.

johngmyers:
Adds a `RollingUpdate` field to instancegroup and cluster for configuring the rolling update strategy. Starts with `MaxUnavailable`, mostly cribbed from machine-api's `MachineDeployment`. Unlike `MachineDeployment`, permits disabling rolling updates for an instancegroup.

/kind feature
/area rolling-update
Fixes #1718, #7685


pkg/instancegroups/rollingupdate_test.go
```diff
@@ -602,6 +605,262 @@ func TestRollingUpdateTaintAllButOneNeedUpdate(t *testing.T) {
 	assertGroupInstanceCount(t, cloud, "node-1", 1)
 }
 
+func TestRollingUpdateSettingsIgnoredForMaster(t *testing.T) {
```
johngmyers: I've since been reconsidering whether we need this exception. Possibly we just need to add some PodDisruptionBudgets for workloads that run on the master.

pkg/instancegroups/instancegroups.go
```diff
@@ -148,15 +149,40 @@ func (r *RollingUpdateInstanceGroup) RollingUpdate(rollingUpdateData *RollingUpd
...
-	for _, u := range update {
-		err = r.drainTerminateAndWait(u, rollingUpdateData, isBastion, sleepAfterTerminate)
+	settings := resolveSettings(cluster, r.CloudGroup.InstanceGroup, numInstances)
+
+	concurrency := 0
+	maxConcurrency := 1
+
+	if r.CloudGroup.InstanceGroup.Spec.Role == api.InstanceGroupRoleNode && !rollingUpdateData.Interactive {
+		maxConcurrency = settings.MaxUnavailable.IntValue()
+		if maxConcurrency == 0 {
```
johngmyers: A subtlety of this placement is that masters, bastions, and interactive updates won't respect disabling of rolling updates. We might want to invert the sense of the enclosing if and have the body set `maxConcurrency = 1`

johngmyers:
Looks like my tests are a little flaky. Perhaps a GC is getting in and throwing off the timings. I will investigate.
/test pull-kops-bazel-test

pkg/instancegroups/instancegroups.go
```diff
@@ -290,6 +300,84 @@ func (r *RollingUpdateInstanceGroup) patchTaint(rollingUpdateData *RollingUpdate
...
+		klog.Errorf("error deleting instance %q, node %q: %v", instanceId, nodeName, err)
+		terminateChan <- err
+		return
+	}
+
+	// Wait for the minimum interval
+	klog.Infof("waiting for %v after terminating instance", sleepAfterTerminate)
+	time.Sleep(sleepAfterTerminate)
+
+	terminateChan <- nil
```
justinsb: It's more composable and less error-prone not to pass the channel in, IMO; but rather to return the error, and then call it with something like  `go result <- foo()` .  Then you can call it without needing channels; you know in the caller that the caller that the error is always sent to the channel (i.e. the parallelism logic is in one function) and e.g. you don't risk pushing a value onto the channel and forgetting to return.

pkg/instancegroups/settings.go
```diff
@@ -0,0 +1,56 @@
...
+	if rollingUpdate.MaxUnavailable == nil || rollingUpdate.MaxUnavailable.IntVal < 0 {
+		one := intstr.FromInt(1)
+		rollingUpdate.MaxUnavailable = &one
+	}
+
+	if rollingUpdate.MaxUnavailable.Type == intstr.String {
+		unavailable, err := intstr.GetValueFromIntOrPercent(rollingUpdate.MaxUnavailable, numInstances, false)
+		if err != nil {
+			// If unparseable use the default value
+			unavailable = 1
```
justinsb: Nit: we might want to log a warning here.  Ideally we would have already caught this in validation though.

pkg/instancegroups/settings_test.go
```diff
@@ -0,0 +1,167 @@
...
+	}{
+		{
+			name:            "MaxUnavailable",
+			defaultValue:    intstr.FromInt(1),
+			nonDefaultValue: intstr.FromInt(2),
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			defaultCluster := &kops.RollingUpdate{}
+			setFieldValue(defaultCluster, tc.name, tc.defaultValue)
```
justinsb: I think I know where this is going... it is _usually_ easier and clearer just to write it out, but ... let's see once the chain lands!

pkg/instancegroups/instancegroups.go
```diff
@@ -148,76 +149,49 @@ func (r *RollingUpdateInstanceGroup) RollingUpdate(rollingUpdateData *RollingUpd
...
 
-	for _, u := range update {
-		instanceId := u.ID
+	settings := resolveSettings(cluster, r.CloudGroup.InstanceGroup, numInstances)
 
-		nodeName := ""
-		if u.Node != nil {
-			nodeName = u.Node.Name
-		}
+	concurrency := 0
```
justinsb: Nit: `runningTasks` or similar might be a clearer name than `concurrency`

pkg/instancegroups/instancegroups.go
```diff
@@ -148,76 +149,49 @@ func (r *RollingUpdateInstanceGroup) RollingUpdate(rollingUpdateData *RollingUpd
...
+	maxConcurrency := settings.MaxUnavailable.IntValue()
 
-		if isBastion {
-			// We don't want to validate for bastions - they aren't part of the cluster
-		} else if rollingUpdateData.CloudOnly {
+	if maxConcurrency == 0 {
+		klog.Infof("Rolling updates for InstanceGroup %s are disabled", r.CloudGroup.InstanceGroup.Name)
+		return nil
```
justinsb: Should we return an error here (as we did not update the IG)?

pkg/instancegroups/instancegroups.go
```diff
@@ -227,11 +201,47 @@ func (r *RollingUpdateInstanceGroup) RollingUpdate(rollingUpdateData *RollingUpd
 				rollingUpdateData.Interactive = false
 			}
 		}
+
+	sweep:
+		for concurrency > 0 {
```
justinsb: I was pondering whether we needed this "sweep" block.  I think in most of these loops we wouldn't, except that we avoid some validate calls by having it here.  Is that right?  If so ... could we have a comment to save our future selves? :-)

justinsb:
LGTM in general, a few nits if you agree with them, but the only two I'd like to reach resolution on now is whether we should pass the channel in to `drainTerminateAndWait`, and having a comment on the `sweep` block about how it avoids validate calls.

pkg/instancegroups/instancegroups.go
johngmyers: We want to continue to update the subsequent instancegroups. The admin specifically wanted to have kops skip this instancegroup, so it isn't an error.

pkg/instancegroups/instancegroups.go
johngmyers: This allows us to keep the concurrency up. For now, validation will likely start returning failures once one node starts draining until its replacement is fully ready, so if we don't sweep we won't be able to sustain more than one or two concurrent drains.

pkg/instancegroups/instancegroups.go
johngmyers: One issue here is the placement. Where it is currently it will validate and taint before returning. Should this be moved up before the initial validation?

pkg/instancegroups/instancegroups.go
johngmyers: Done

pkg/instancegroups/settings.go
```diff
@@ -0,0 +1,56 @@
...
+		rollingUpdate = *group.Spec.RollingUpdate
+	}
+
+	if def := cluster.Spec.RollingUpdate; def != nil {
+		if rollingUpdate.MaxUnavailable == nil {
+			rollingUpdate.MaxUnavailable = def.MaxUnavailable
+		}
+	}
+
+	if rollingUpdate.MaxUnavailable == nil || rollingUpdate.MaxUnavailable.IntVal < 0 {
```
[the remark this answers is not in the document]
johngmyers: Added validation

pkg/instancegroups/instancegroups.go
johngmyers: Went with `runningDrains`

pkg/instancegroups/instancegroups.go
johngmyers: Added a comment

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -133,6 +133,12 @@ func ValidateInstanceGroup(g *kops.InstanceGroup) error {
 		return err
 	}
 
+	if g.Spec.RollingUpdate != nil {
+		if errs := validateRollingUpdate(g.Spec.RollingUpdate, field.NewPath("rollingUpdate")); len(errs) > 0 {
+			return errs.ToAggregate()
```
johngmyers: I'm curious as to why `ValidateInstanceGroup()` and `ValidateCluster()` return on the first found set of errors rather than gather and return all errors. Would this be something to fix?

justinsb: Probably yes - apimachinery introduced the idea of gathering all the errors, and we had something like validation that didn't use the machinery previously.  The apimachinery approach is better.  So yes, it would be great to gather and return all the errors (and print them all also, if we're not already doing so!)

johngmyers: I'll take that up in a separate PR.

pkg/instancegroups/instancegroups.go
justinsb: I may be misunderstanding, but I think you're saying that `maxUnavailable: 0` means "skip this IG for rolling updates".   I'm not sure I follow that implication, but I also don't know what `maxUnavailable: 0` would mean!

Is `maxUnavailable: 0` something you're actually using?  Because we could just treat it as a validation error until we have maxSurge support (I guess that maxUnavailable=0 and maxSurge=0 is a validation error on a Deployment?)

I'm not overly worry about the tainting, TBH, given that!

justinsb:
Thanks for the tweaks @johngmyers - this LGTM.  I do wonder if we should just make `maxUnavailable: 0` into a validation error for now - are you using it?  But anyway we can do that in a separate PR - I don't want to drag this one out any more - thanks for your patience!

/approve
/lgtm

johngmyers:
@justinsb `maxUnavailable: 0` is intended to address #7685. Do you have an issue with providing that feature, or just this PR's implementation of it? Perhaps we should reopen #7685 and discuss there?

Issue #7411, opened 2019-08-15
Allow "refreshing" instance groups


Moshem123:
I'd like to be able to refresh an instance group, masters as well.
In my eyes, this means a rolling-update with no changes, say I have nodes that are running for a long time, and over time the nodes degradate, I want to spin up x2 nodes, and terminate the old ones using oldestInstance policy.
Can this be achieved with the current capabilities?
Thanks.

Moshem123:
/remove-lifecycle stale
would allow me to keep track of the PR.

PR #8272, opened 2020-01-04
Support the kops.k8s.io/needs-update annotation on nodes

johngmyers:
Adds support for an annotation that nominates nodes for replacement by the kops rolling update mechanism. Possible uses would be external controllers to limit the age of nodes (ref #7411) or to remove unhealthy but still functioning nodes. By having the replacement performed by kops rolling update, it would necessarily coordinate with kops rolling updates and use kops's configured strategy, for example surging when we implement it.

/kind feature
/area rolling-update


pkg/instancegroups/rollingupdate.go
```diff
@@ -69,6 +69,31 @@ type RollingUpdateCluster struct {
 	ValidateSuccessDuration time.Duration
 }
 
+// AdjustNeedUpdate adjusts the set of instances that need updating, using factors outside those known by the cloud implementation
+func (c *RollingUpdateCluster) AdjustNeedUpdate(groups map[string]*cloudinstances.CloudInstanceGroup, cluster *api.Cluster, instanceGroups *api.InstanceGroupList) error {
+	for _, group := range groups {
+		if group.Ready != nil {
```
justinsb: FYI for reads golang treats a nil slice as empty, so I don't think you actually need this check.  Arguably it does make things clearer though!

pkg/instancegroups/rollingupdate.go
```diff
@@ -69,6 +69,31 @@ type RollingUpdateCluster struct {
...
+			var newReady []*cloudinstances.CloudInstanceGroupMember
+			for _, member := range group.Ready {
+				makeNotReady := false
+				if member.Node != nil && member.Node.Annotations != nil {
```
justinsb: (And FYI the same applies to nil maps, for reading - though not for writing, so again you don't need the Annotations != nil check)

justinsb:
A tip about nil slices & maps, but this LGTM.  Thanks @johngmyers 

Unfortunately I think we've conflicted on a doc file, so I suspect I'll need to re-apply lgtm.

/approve
/lgtm

justinsb:
Thanks for rebasing!

/approve
/lgtm

PR #8273, opened 2020-01-05
Remove kops-controller deployment

rifelpet:
kops-controller now runs as a DaemonSet. The migration was first made in 1.16.0-alpha.1, so that means 2 releases have been out that set the replicas to zero.
This removal negatively impacts anyone that created a cluster from kops HEAD between 1.15.0 and 1.16.0-alpha.1, and then upgraded kops directly to the 1.16.0 release that includes this commit, without having first upgraded to either of the alphas.

That seems like a reasonably small enough audience that this is safe to remove now.
Perhaps we mention in the release notes that anyone using HEAD or one of the alpha releases needs to `kubectl delete -n kube-system deployment kops-controller`

I'll add a release note in a separate PR since the release note wont get cherry-picked back to 1.16.

We'll also need to add a way to provide an iam role annotation to kops-controller or else any cluster running kube2iam with a restricted [--default-role](https://github.com/jtblin/kube2iam/blob/master/cmd/main.go#L23) wont be able to get AWS credentials with the necessary permissions to run, which will cause the rolling update to fail. Perhaps I'll open an issue to discuss this if we think there are multiple possible solutions for that.

### DIFF
--- a/upup/models/cloudup/resources/addons/kops-controller.addons.k8s.io/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/kops-controller.addons.k8s.io/k8s-1.16.yaml.template
@@ -11,38 +11,6 @@ data:
 
 ---
 
-# Deployment of size 0, to move from Deployment to DaemonSet
-# TODO: Remove in beta? (it's only been on master branch)
-
-kind: Deployment
-apiVersion: apps/v1
-metadata:
-  name: kops-controller
-  namespace: kube-system
-  labels:
-    k8s-addon: kops-controller.addons.k8s.io
-    k8s-app: kops-controller
-    version: v1.15.0-alpha.1
-spec:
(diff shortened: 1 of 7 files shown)


rifelpet:
I'll get these conflicts cleaned up in a bit.

@joshbranham I opened https://github.com/kubernetes/kops/issues/8306 to discuss :)

rifelpet:
@zetaab updated and passing tests.

joshbranham:
I do think we need to denote these types of cleanup steps in release notes. I ran into an issue with a `kube-dns` PodDisruptionBudget that was in the default namespace (that bug was fixed) and I had to clean it up from our clusters to get pod evictions to work.

rifelpet:
Yep, I'll open a PR that adds some release notes once this merges and is cherry-picked back.

PR #8276, opened 2020-01-06
Add support for custom env vars in amazon-vpc-cni

### LINKED ISSUE #7821: Kops should stop updating AWS VPC CNI config after bootstrapping
 this information.**

1.14.0

 a `kops` flag.**

1.12.10

AWS

I am doing a `replace` and `update` to synchronize YAML config to the kops config.

```
kops replace cluster --force -f kube-dev.yaml
kops update cluster --name=kube-dev
kops update cluster --name=kube-dev --yes
```

The AWS VPC CNI `aws-node` daemonset is reconfigured and it loses the important extra config I have added:

```
            - name: WARM_IP_TARGET
              value: "1"
            - name: WARM_ENI_TARGET
              value: "0"
```

these are important for my small VPC running a small subnet, carved out by my larger enterprise environment. I only have a `/23` to work with.

It'd be nice if running kops didn't reconfigure my networking stack. When the reconfiguration kicks in, it automatically restarts all the pods and my IP usage skyrockets.

```
apiVersion: kops/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: 2019-03-26T00:00:00Z
  name: kube-dev.meow.io
spec:
  docker:
    version: "18.06.2"
  kubeAPIServer:
    runtimeConfig:
      admissionregistration.k8s.io/v1alpha1: "true"
  api:
    loadBalancer:
      type: Internal
      idleTimeoutSeconds: 1800 # 30 minutes
  authorization:
    rbac: {}
  authentication:
    aws: {}
  additionalPolicies:
    master: |
      [
        {
          "Effect": "Allow",
          "Action": ["s3:ListObjects"],
          "Resource": ["arn:aws:s3:::ids-dev-kops/kube-dev.interactivedatastore.viasat.io/*"]
        },
        {
          "Effect": "Allow",
          "Action": ["elasticloadbalancing:DeregisterTargets"],
          "Resource": ["arn:aws:elasticloadbalancing:us-west-2:539703851677:targetgroup/*"]
        }
      ]
    node: |
      [
        {
          "Effect": "Allow",
          "Action": [
            "ec2:CreateNetworkInterface",
            "ec2:AttachNetworkInterface",
            "ec2:DeleteNetworkInterface",
            "ec2:DetachNetworkInterface",
            "ec2:DescribeNetworkInterfaces",
            "ec2:DescribeInstances",
            "ec2:ModifyNetworkInterfaceAttribute",
            "ec2:AssignPrivateIpAddresses",
            "ec2:UnassignPrivateIpAddresses"
          ],
          "Resource": ["*"]
        },
        {
          "Effect": "Allow",
          "Action": [
            "ec2:CreateTags"
          ],
          "Resource": [
            "arn:aws:ec2:*:*:network-interface/*"
          ]
        }
      ]
  channel: stable
  cloudProvider: aws
  configBase: s3://ids-dev-kops/kube-dev.meow.io
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-us-west-2c
      name: c
    name: main
    enableEtcdTLS: true
    version: 3.2.24
  - etcdMembers:
    - instanceGroup: master-us-west-2c
      name: c
    name: events
    enableEtcdTLS: true
    version: 3.2.24
  iam:
    allowContainerRegistry: true
    legacy: false
  kubernetesApiAccess:
  - 0.0.0.0/0
  kubernetesVersion: 1.13.12
  masterPublicname: api.kube-dev.meow.io
  networkCIDR: 10.43.202.0/23
  networkID: vpc-0b94766c8a8c871a4
  networking:
    amazonvpc:
      imageName: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.5.3
  nonMasqueradeCIDR: 10.43.202.0/23
  serviceClusterIPRange: 10.43.202.0/23
  sshAccess:
  - 0.0.0.0/0
  subnets:
  - cidr: 10.43.202.0/23
    id: subnet-03fa682131e608935
    name: us-west-2c
    type: Private
    zone: us-west-2c
    egress: External
  topology:
    dns:
      type: Private
    masters: private
    nodes: private
```

rifelpet:
There are a handful of env vars that users often want to set when using the AWS VPC CNI (`amazonvpc`):

https://github.com/aws/amazon-vpc-cni-k8s/tree/master#cni-configuration-variables

This opens them up through the cluster spec so they can be set more easily.

~~I was up in the air with the field name, `env` like in a pod spec's container, `envVar`, `variables` but settled with `environment` which is already used in [exec hooks](https://godoc.org/k8s.io/kops/pkg/apis/kops#ExecContainerAction). I'm open to using a different name though.~~ After discussion during office hours I went with `env` to match the pod spec's container field.

/hold since this relies on much of the test work of #8274, so waiting until that is merged

closes #7821


k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -2407,6 +2407,31 @@ spec:
...
+                    env:
+                      description: Env is a list of environment variables to set in
+                        the container.
+                      items:
+                        description: EnvVar represents an environment variable present
+                          in a Container.
+                        properties:
+                          name:
+                            description: Name of the environment variable. Must be
+                              a C_IDENTIFIER.
```
geojaz: Would it be helpful to either give an example of what a `C_IDENTIFIER` looks like/link? `Must be a C_IDENTIFIER (see https://reference_material for more information). I personally wasn't able to use the hint in the description because I don't remember any C, and was only somewhat successful in figuring out what this format should be with my googlemachine. 

I also would not be opposed to truncating this description to the first sentence. :)

k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -2407,6 +2407,31 @@ spec:
...
+                            type: string
+                          value:
+                            description: 'Variable references $(VAR_NAME) are expanded
```
geojaz: This is a bit awkward, imho. Perhaps something along these lines? (If i've understood correctly)
```suggestion
                            description: Variables are expanded by merging the default container variables and your custom variables. When a reference cannot be resolved, the input will be unchanged and will contain `$(VARIABLES)`. If a variable should never be expanded, use the format `$$(VARIABLE)` which will render as `$(VARIABLE)`.
```

upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/amazonvpc/cluster.yaml
```diff
@@ -27,7 +27,10 @@ spec:
   - proxy.api.minimal.example.com
   networkCIDR: 172.20.0.0/16
   networking:
-    amazonvpc: {}
+    amazonvpc:
+      env:
+      - name: WARM_IP_TARGET
+        value: "10"
```
geojaz: could you also show a test with an additional variable?

pkg/apis/kops/v1alpha2/networking.go
```diff
@@ -190,6 +190,8 @@ type AmazonVPCNetworkingSpec struct {
 	// The container image name to use, which by default is:
 	// 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.5.5
 	ImageName string `json:"imageName,omitempty"`
+	// Env is a list of environment variables to set in the container.
+	Env []EnvVar `json:"env,omitempty"`
```
geojaz: I think that `Environment` might be preferable to `EnvVar`, thoughts

k8s/crds/kops.k8s.io_clusters.yaml
rifelpet: The descriptions in these CRD fields actually come from the new API type we added a while back in #8455 specifically for this PR as well as #8402 

I copied those descriptions from upstream: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L1827-L1846

I'm fine with truncating the descriptions but it will need to be done on the api type itself so that `make crds` will update these new fields to match.

https://github.com/kubernetes/kops/blob/f400f74a3e498201efc77833b297184376710d61/pkg/apis/kops/cluster.go#L646-L661

pkg/apis/kops/v1alpha2/networking.go
rifelpet: When we discussed this during office hours we decided to follow upstream regarding environment variable terminology:

https://github.com/kubernetes/api/blob/d5ab0ddd8b7334701d758c13524fec4830b4c923/core/v1/types.go#L1826-L1827

upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/amazonvpc/cluster.yaml
rifelpet: done 👍

PR #8283, opened 2020-01-06
containerd: Set a default version even with Kubernetes 1.17

hakman:
For testing I need to be able to enable the containerd container runtime with Kubernetes 1.17 without setting the version in cluster spec.
It is a tiny change that should have minimal effect outside e2e tests.


rifelpet:
Since this still requires users explicitly opt in to containerd which isnt in a stable kops release yet, this is a pretty safe change. Thanks!

/lgtm
/approve

Issue #8040, opened 2019-12-04
Unexpected  "Subnet.CIDR: Ifield is immutable:" error - not trying to change CIDR


kmadathil:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Version 1.14.1 (git-b7c25f9a9)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"211047e9a1922595eaa3a1127ed365e9299a6c23", GitTreeState:"clean", BuildDate:"2019-10-15T12:02:12Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

I first did

```
kops create cluster ${CLUSTER_NAME} \
  --node-count ${NODE_COUNT} \
  --zones $ZONES \
  --node-size $NODE_SIZE \
  --master-size $MASTER_SIZE \
  --master-zones $ZONES \
  --vpc $VPC_ID \

kops replace cluster ${CLUSTER_NAME} \
  -f kops-cluster.yaml

kops update cluster ${CLUSTER_NAME} --yes

```
kops-cluster.yaml has a modified subnet CIDR, so that I can run on a pre-determined VPC without collisions. So far so good, everything runs well. 

Now, I need to reduce the number of nodes to zero temporarily. I ran `kops edit ig`, made the change, and then ran `kops update cluster --yes`. 

**5. What happened after the commands executed?**
I see the following error.

```
error running task "Subnet/us-west-2a.clustername.k8s.local" (8m58s remaining to succeed): Subnet.CIDR: Invalid value: "172.31.128.0/19": field is immutable: old="172.31.144.0/19" new="172.31.128.0/19"
```
I am not trying to modify the subnet CIDR, only maxSize/minSize so this error should not occur. 

**6. What did you expect to happen?**
[...]

PR #8284, opened 2020-01-07
Improve CIDR validation

johngmyers:
Would have caught the user error in #8040


fgbreel:
Thank you! :+1:

justinsb:
Thanks for fixing this @johngmyers - and our tests even made the same mistake!

/approve
/lgtm

PR #8286, opened 2020-01-08
Add wget and nfs-utils deps needed for e2e tests

hakman:
Periodic e2e tests are failing for both CentOS 7 and RHEL 8 because of missing `wget` and `nfs-utils` deps.
Debian family usually has `wget` already installed and Kops adds `nfs-common` alternative to utils already.

`nfs-utils` is required for NFS mounts to work in general, not just for e2e.

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kops-aws-imagecentos7/1214370995477942272
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kops-aws-containerd/1214741461090701315

### DIFF
--- a/nodeup/pkg/model/miscutils.go
+++ b/nodeup/pkg/model/miscutils.go
@@ -52,11 +52,14 @@ func (b *MiscUtilsBuilder) Build(c *fi.ModelBuilderContext) error {
 	if b.Distribution.IsDebianFamily() {
 		packages = append(packages, "socat")
 		packages = append(packages, "curl")
+		packages = append(packages, "wget")
 		packages = append(packages, "nfs-common")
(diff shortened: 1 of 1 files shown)


joshbranham:
Good catch!

/lgtm

hakman:
@rifelpet any concern about this?

mikesplain:
Good catch! Thanks so much!

Do you think we should cherry pick this back?

/lgtm
/approve

hakman:
@mikesplain we should cherrypick this and a few other PRs to 1.17 to enable RHEL 8 support maybe. What do you think?

Issue #7740, opened 2019-10-03
kube-scheduler fails to start when `usePolicyConfigMap` is enabled


rtluckie:
**1. What `kops` version are you running?**
1.14.0

**2. What Kubernetes version are you running?**
1.15.4

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?**

```bash
kops create -v 10 \
    --name cluster1.somedomain.io \
    --state "s3://cluster1.somedomain.io" \
    -f cluster_config.yaml

kops create -v 10 secret sshpublickey admin \
    --name cluster1.somedomain.io \
    --state "s3://cluster1.somedomain.io" \
    --config cluster_config.yaml \
    -i cluster1.somedomain.io.pub

kops -v 10 update cluster cluster1.somedomain.io \
    --state "s3://cluster1.somedomain.io" \
    --yes
```
[cluster_config.yaml](https://gist.github.com/rtluckie/b61ebea554e4b0f631c69bbd331da52a#file-cluster_config-yaml)

**5.  What is the simplest way to reproduce this issue?**

Create a cluster and set `spec.kubeScheduler.usePolicyConfigMap: true`

**6. What happened after the commands executed?**

* Master nodes come up but are in a `NotReady` state
* kube-scheduler pods continually crash

**7. What did you expect to happen?**
* All nodes would come online and enter a `Ready` state without manual intervention.

**8. Please provide your cluster manifest.**

[cluster manifest](https://gist.github.com/rtluckie/b61ebea554e4b0f631c69bbd331da52a#file-cluster_manifest-yaml)

**9. Please run the commands with most verbose logging by adding the `-v 10` flag.**

[kops log](https://gist.github.com/rtluckie/b61ebea554e4b0f631c69bbd331da52a#file-kops_commands-log)

**10. Anything else do we need to know?**

**This is not a new issue.  I've been working around it since K8s 1.13, maybe earlier.**

There are two different issues causing the kube-scheduler pods to crash.

1. The `system:kube-scheduler` clusterrole doesn't grant access to configmaps.

Related log messages:

```bash
I1003 05:31:39.339583       1 server.go:161] Starting Kubernetes Scheduler version v1.15.4
couldn't get policy config map kube-system/scheduler-policy: configmaps "scheduler-policy" is forbidden: User "system:kube-scheduler" cannot get resource "configmaps" in API group "" in the namespace "kube-system"
```

**Manual fix**: edit the `system:kube-scheduler` clusterrole and append the following:
[...]

PR #8057, opened 2019-12-08
Turn off dns-controller metrics that depend on kubernetes/kubernetes

justinsb:
This is the last dependency on kubernetes/kubernetes repo, which should now
make it much easier to do updates.

PR #8293, opened 2020-01-09
Add config to KubeSchedulerConfig

vvbogdanov87:
Related to #7740

PR adds `config` field to `kubeScheduler`. It enables the ability to use `--config` kube-scheduler flag instead of deprecated flags like `--kubeconfig`, `--policy-configmap`, `--policy-configmap-namespace`.


hakman:
@vvbogdanov87 I think there was some work here to reduce tight coupling. https://github.com/kubernetes/kops/pull/8057
Maybe you can ask some advice about this on Slack.

vvbogdanov87:
From [vendoring.md](https://github.com/kubernetes/kops/blob/master/docs/development/vendoring.md#adding-a-dependency-to-the-vendor-directory) 
```
Run make gomod
...
Open a pull request with these changes separately from other work so that it is easier to review.
```
Should I open a second PR with "vendor" changes to master? And merge master back to my branch after the second PR is merged to master?

vvbogdanov87:
verify-generated fails because few fields don't have Json tag
https://github.com/kubernetes/kubernetes/blob/ef69bc910f0e47bbe3cf396d4bebf4f678cf6f3a/staging/src/k8s.io/apimachinery/pkg/runtime/types.go#L119
Should I make a fix for this? Or it is a bad idea to embed the whole kube-scheduler config type into cluster spec? How should we handle the scheduler configuration in that case?

vvbogdanov87:
anyone?

Mhs-220:
is there any news for this pr?

mlushpenko:
It's still an issue with the latest 1.16 release, anyone knows when/if this one is planned to be implemented?

olemarkus:
We probably want something like this PR to go in. Since this was made, we have support for the config file now, and there are examples for how to set this in the API files. See `Burst` and `Qps` as examples.

PR #8056, opened 2019-12-07
Replace kubernetes mount code with utils

justinsb:
This will remove one of the main dependencies on the kubernetes/kubernetes repo.

PR #8055, opened 2019-12-07
Update dependencies to kubernetes 1.16

justinsb:
Helps pare down our dependency on kubernetes/kubernetes

PR #8294, opened 2020-01-09
Spotinst: Support for Headroom, Taints and Subnets in Ocean Launch Spec and User Data fixes

liranp:
This PR adds support for Headroom, Taints and Subnets in Ocean Launch Spec and minor User Data fixes.

### DIFF
--- a/pkg/model/spotinstmodel/BUILD.bazel
+++ b/pkg/model/spotinstmodel/BUILD.bazel
@@ -14,6 +14,7 @@ go_library(
         "//upup/pkg/fi:go_default_library",
         "//upup/pkg/fi/cloudup/awstasks:go_default_library",
         "//upup/pkg/fi/cloudup/spotinsttasks:go_default_library",
+        "//vendor/k8s.io/api/core/v1:go_default_library",
         "//vendor/k8s.io/klog:go_default_library",
     ],
 )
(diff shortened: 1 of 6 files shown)


joshbranham:
@liranp would you mind splitting the commits up a bit? This is a pretty large change (vendoring makes sense) but usually vendored code a standalone commit is easier for reviewers

liranp:
@joshbranham Done. Please let me know if you want me to split more. Thanks for reviewing this!

go.mod
```diff
@@ -139,6 +139,7 @@ require (
 	k8s.io/helm v2.9.0+incompatible
 	k8s.io/klog v1.0.0
 	k8s.io/kubectl v0.0.0
+	k8s.io/kubernetes v1.17.3
```
rifelpet: We've been working towards dropping k8s.io/kubernetes as a dependency (example: https://github.com/kubernetes/kops/pull/8055 and https://github.com/kubernetes/kops/pull/8056) since it can be a nightmare to update with each minor version and all of the transitive dependencies.

Seeing as you're only using the `ParseTaints` function and there doesn't appear to be any effort to move it to one of the other k8s.io repos, maybe we consider copying the function directly into the kops repo? It looks like ParseTaints and two unexported functions might be all that is needed. 

@justinsb what do you think? You can also provide more context on shifting away from depending on k/k.

liranp: I've removed the `k8s.io/kubernetes` dependency and [extracted the relevant parts only](https://github.com/kubernetes/kops/blob/2a17a50d007f3f0ef9f8fb1598838ab58fbbd99d/pkg/model/spotinstmodel/instance_group.go#L847-L883).

PR #8312, opened 2020-01-11
Add cloud ntp addresses

simonmacklin:
Hi 

This is a in working progress PR which will add the cloud specific NTP server into the NTP config.  This will prevent clock drift which could end up in resulting in downtime.  AWS as an example will prevent you speaking with there APIs once your clock is out by five minutes.

It would be great to get some feedback on whether you feel this is the best approach I am taking and also if you would prefer more parameters in the function I created rather pre determining the values inside the function with switch statements.

I used switch statements to prevent having to have the same logic again and again when determining the ntp install type.

I would like in another PR to maybe offer custom NTPs that a user could pass into the cluster spec.  This might be handy when you have custom NTP servers when hosting Kubernetes on prem.

@mikesplain hope you don't mind me tagging you into this and we already sort of discussed this on slack :)

Simon

### DIFF
--- a/nodeup/pkg/model/ntp.go
+++ b/nodeup/pkg/model/ntp.go
@@ -17,6 +17,10 @@ limitations under the License.
 package model
 
 import (
+	"fmt"
+	"io/ioutil"
+	"regexp"
+
 	"k8s.io/klog"
 	"k8s.io/kops/nodeup/pkg/distros"
 	"k8s.io/kops/upup/pkg/fi"
@@ -32,6 +36,13 @@ type NTPBuilder struct {
 
 var _ fi.ModelBuilder = &NTPBuilder{}
 
+type ntpDaemon string
+
+var (
+	chronyd ntpDaemon = "chronyd"
+	ntpd    ntpDaemon = "ntpd"
+)
+
 // Build is responsible for configuring NTP
 func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 	switch b.Distribution {
@@ -46,22 +57,107 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 		return nil
 	}
 
+	var ntpIP string
+	switch b.Cluster.Spec.CloudProvider {
+	case "aws":
+		ntpIP = "169.254.169.123"
+	case "gce":
+		ntpIP = "time.google.com"
+	default:
+		ntpIP = ""
+	}
+
 	if b.Distribution.IsDebianFamily() {
 		c.AddTask(&nodetasks.Package{Name: "ntp"})
+
+		if ntpIP != "" {
+			bytes, err := updateNtpIP(ntpIP, ntpd)
+			if err != nil {
+				return err
+			}
+			c.AddTask(&nodetasks.File{
+				Path:     "/etc/ntp.conf",
+				Contents: fi.NewBytesResource(bytes),
+				Type:     nodetasks.FileType_File,
+				Mode:     s("0644"),
+			})
+		}
+
 		c.AddTask((&nodetasks.Service{Name: "ntp"}).InitDefaults())
 	} else if b.Distribution.IsRHELFamily() {
 		switch b.Distribution {
 		case distros.DistributionCentos8, distros.DistributionRhel8:
 			c.AddTask(&nodetasks.Package{Name: "chrony"})
+
+			if ntpIP != "" {
+				bytes, err := updateNtpIP(ntpIP, chronyd)
+				if err != nil {
+					return err
+				}
+				c.AddTask(&nodetasks.File{
+					Path:     "/etc/chrony.conf",
+					Contents: fi.NewBytesResource(bytes),
+					Type:     nodetasks.FileType_File,
+					Mode:     s("0644"),
+				})
+			}
 			c.AddTask((&nodetasks.Service{Name: "chronyd"}).InitDefaults())
+
 		default:
 			c.AddTask(&nodetasks.Package{Name: "ntp"})
+
+			if ntpIP != "" {
+				bytes, err := updateNtpIP(ntpIP, ntpd)
+				if err != nil {
+					return err
+				}
+				c.AddTask(&nodetasks.File{
+					Path:     "/etc/ntp.conf",
+					Contents: fi.NewBytesResource(bytes),
+					Type:     nodetasks.FileType_File,
+					Mode:     s("0644"),
+				})
+			}
+
 			c.AddTask((&nodetasks.Service{Name: "ntpd"}).InitDefaults())
 		}
 	} else {
 		klog.Warningf("unknown distribution, skipping ntp install: %v", b.Distribution)
 		return nil
 	}
-
 	return nil
+}
+
+// updateNtpIP takes a ip and a ntpDaemon and will comment out
+// the default server or pool values and append the correct cloud
+// ip to the ntp config file.
+func updateNtpIP(ip string, daemon ntpDaemon) ([]byte, error) {
+	var address string
+	var r *regexp.Regexp
+	var path string
+	switch ntpd {
+	case ntpd:
+		address = fmt.Sprintf("server %s prefer iburst", ip)
+		// the regex strings might need a bit more work
+		r = regexp.MustCompile(`pool\s\d.*[a-z].[a-z].[a-z]\siburst`)
+		path = "/etc/ntp.conf"
+	case chronyd:
+		address = fmt.Sprintf("server %s prefer iburst minpoll 4 maxpoll 4", ip)
+		// the regex strings might need a bit more work
+		r = regexp.MustCompile(`server\s.*iburst.*`)
+		path = "/etc/chrony.conf"
+	default:
+		return nil, fmt.Errorf("%s is not a supported ntp application", ntpd)
+	}
+
+	f, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	new := r.ReplaceAllFunc(f, func(b []byte) []byte {
+		return []byte(fmt.Sprintf("#commented out by kops %s", string(b)))
+	})
+	new = append(new, []byte(address)...)
+	return new, nil
 }
--- a/nodeup/pkg/model/ntp.go
+++ b/nodeup/pkg/model/ntp.go
@@ -135,7 +135,7 @@ func updateNtpIP(ip string, daemon ntpDaemon) ([]byte, error) {
 	var address string
 	var r *regexp.Regexp
 	var path string
-	switch ntpd {
+	switch daemon {
 	case ntpd:
 		address = fmt.Sprintf("server %s prefer iburst", ip)
 		// the regex strings might need a bit more work
--- a/nodeup/pkg/model/ntp.go
+++ b/nodeup/pkg/model/ntp.go
@@ -133,21 +133,17 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 // ip to the ntp config file.
 func updateNtpIP(ip string, daemon ntpDaemon) ([]byte, error) {
 	var address string
-	var r *regexp.Regexp
 	var path string
+	r := regexp.MustCompile(`(?m)(^pool|^server)\s.*`)
 	switch daemon {
 	case ntpd:
-		address = fmt.Sprintf("server %s prefer iburst", ip)
-		// the regex strings might need a bit more work
-		r = regexp.MustCompile(`pool\s\d.*[a-z].[a-z].[a-z]\siburst`)
+		address = fmt.Sprintf("server %s prefer iburst\n", ip)
 		path = "/etc/ntp.conf"
 	case chronyd:
-		address = fmt.Sprintf("server %s prefer iburst minpoll 4 maxpoll 4", ip)
-		// the regex strings might need a bit more work
-		r = regexp.MustCompile(`server\s.*iburst.*`)
+		address = fmt.Sprintf("server %s prefer iburst minpoll 4 maxpoll 4\n", ip)
 		path = "/etc/chrony.conf"
 	default:
-		return nil, fmt.Errorf("%s is not a supported ntp application", ntpd)
+		return nil, fmt.Errorf("%s is not a supported ntp application", daemon)
 	}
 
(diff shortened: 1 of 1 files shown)


nodeup/pkg/model/ntp.go
```diff
@@ -46,22 +49,102 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 		return nil
 	}
 
+	var ntp string
+	switch b.Cluster.Spec.CloudProvider {
+	case "aws":
+		ntp = "169.254.169.123"
+	case "gce":
+		ntp = "time.google.com "
```
joshbranham: Extra whitespace

simonmacklin: Great spot!! I will fix that now

simonmacklin: I removed the white space thanks @joshbranham

nodeup/pkg/model/ntp.go
```diff
@@ -17,10 +17,13 @@ limitations under the License.
 package model
 
 import (
+	"fmt"
+	"io/ioutil"
 	"k8s.io/klog"
 	"k8s.io/kops/nodeup/pkg/distros"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
+	"regexp"
```
hakman: This needs to be moved before the k8s ones.
Try also `make goimports`.

nodeup/pkg/model/ntp.go
```diff
@@ -46,22 +49,102 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
...
 		}
 	} else {
 		klog.Warningf("unknown distribution, skipping ntp install: %v", b.Distribution)
 		return nil
 	}
-
 	return nil
 }
+
+func UpdateNtpIP(ip string, ntp string) ([]byte, error) {
```
johngmyers: Seems to me that `ntp` should be its own type, with constants.

nodeup/pkg/model/ntp.go
```diff
@@ -46,22 +49,102 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
...
 
+	var ntp string
+	switch b.Cluster.Spec.CloudProvider {
+	case "aws":
+		ntp = "169.254.169.123"
+	case "gce":
+		ntp = "time.google.com"
+	default:
+		ntp = ""
+	}
```
johngmyers: The code passes a variable named `ntp` to a parameter named `ip`, while there is a different parameter named `ntp`. This would be less confusing if the variables were named consistently.

nodeup/pkg/model/ntp.go
simonmacklin: Thanks @hakman I have moved It up before the k8 ones.  For some reason go imports is not working on my Mac.  Once I have fixed that I'll run it.

nodeup/pkg/model/ntp.go
simonmacklin: Thanks @johngmyers I was planning on cleaning that up a bit once I got it working.  Does the newest commit look better?

nodeup/pkg/model/ntp.go
```diff
@@ -120,12 +127,16 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
...
+// ip to the ntp config file
+func UpdateNtpIP(ip string, ntpd NtpDaemon) ([]byte, error) {
+	// the regex strings might need a bit more work
 	var address string
 	var r *regexp.Regexp
 	var path string
-	switch ntp {
-	case "ntp":
+	switch ntpd {
+	case "ntpd":
```
johngmyers: Should be using the constants here

nodeup/pkg/model/ntp.go
```diff
@@ -120,12 +127,16 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 	return nil
 }
 
-func UpdateNtpIP(ip string, ntp string) ([]byte, error) {
+// UpdateNtpIP takes a ip and a NtpDaemon and will comment out 
+// the default server or pool values and append the correct cloud 
+// ip to the ntp config file
+func UpdateNtpIP(ip string, ntpd NtpDaemon) ([]byte, error) {
```
johngmyers: The parameter is shadowing the constant of the same name. They should have different names.

nodeup/pkg/model/ntp.go
```diff
@@ -120,12 +127,16 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 	return nil
 }
 
-func UpdateNtpIP(ip string, ntp string) ([]byte, error) {
+// UpdateNtpIP takes a ip and a NtpDaemon and will comment out 
+// the default server or pool values and append the correct cloud 
+// ip to the ntp config file
```
johngmyers: nit: should end with period.

nodeup/pkg/model/ntp.go
```diff
@@ -35,6 +35,13 @@ type NTPBuilder struct {
 
 var _ fi.ModelBuilder = &NTPBuilder{}
 
+type NtpDaemon string
```
johngmyers: Perhaps it doesn't have to be public?

nodeup/pkg/model/ntp.go
simonmacklin: yep

nodeup/pkg/model/ntp.go
simonmacklin: Thanks @johngmyers would you be happy if I just used ```func UpdateNtpIP(ip string, n NtpDaemon)``` as its clear that it is now by the type?

nodeup/pkg/model/ntp.go
simonmacklin: I will add the . :)

nodeup/pkg/model/ntp.go
simonmacklin: sure I have unexport the const.  Shall I do the same with the func too?

nodeup/pkg/model/ntp.go
johngmyers: That is acceptable, though I would suggest `daemon`.

nodeup/pkg/model/ntp.go
johngmyers: Oh right, that would be good.

nodeup/pkg/model/ntp.go
simonmacklin: I have switched the param to daemon.  Thanks @johngmyers

simonmacklin:
I have made some changes requested by @johngmyers.  I have tested it with the default image and all seems to be working fine.  I just would like to run some more tests and double check the code before this is considered to be merged.  The regex strings are not perfect either so I need to work on those.

I am really short on time at the moment so I will probably not get a chance to sit and concentrate on this until Tuesday.  

Thanks

Simon

johngmyers:
/hold
per previous comment

rifelpet:
@simonmacklin you may need to squash your commits and rebase in order to fix the CLA issue. It looks like [this](https://github.com/kubernetes/kops/pull/8312/commits/1b995c4af179226ddd3c64baafa96cc77e4082b2) is the offending commit.

simonmacklin:
Thanks @rifelpet I think I have fixed it.  I formatted my mac and forgot to set my email in the git config.  Sorry about that

Simon

simonmacklin:
I have tested this on the default image and centos 8 and all seems to be working correctly.  

Kind Regards

Simon

nodeup/pkg/model/ntp.go
```diff
@@ -46,22 +57,103 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
...
 	return nil
 }
+
+// updateNtpIP takes a ip and a ntpDaemon and will comment out
+// the default server or pool values and append the correct cloud
+// ip to the ntp config file.
+func updateNtpIP(ip string, daemon ntpDaemon) ([]byte, error) {
+	var address string
+	var path string
+	r := regexp.MustCompile(`(?m)(^pool|^server)\s.*`)
```
johngmyers:
```suggestion
	r := regexp.MustCompile(`(?m)^(?:pool|server)\s.*`)
```

simonmacklin: Thanks @johngmyers that looks better :)

johngmyers:
The `AddTask` calls could possibly extracted out to single copy, using variables for the package name, config file path, and service name, but this looks good enough.

I've been trying to think of reasons someone might want this to be configurable and not coming up with anything likely.
/lgtm

simonmacklin:
Thanks @johngmyers 

I think there could be a use case to make this configurable when you are hosting kubernetes in house and want to use your own set of NTPs.  That being said you can always use DHCP to override the settings in the config anyhow.  

Also I really appreciate your help :)

johngmyers:
This doesn't have support for an "in house" cloud provider, so that case isn't relevant.

simonmacklin:
@johngmyers 

I stand corrected.  I thought KOPS supported self hosted OpenStack and VSphere.  So I assume they are only the cloud offerings.  

Thanks

johngmyers:
Kops does support those, but this added code only acts when the provider is AWS or GCE.

oded-dd:
I am not sure you support NTP for CoreOs, as it defaults to timesyncd NTP service

The way we support it is by Kops hooks:

‘’’
amazon_time_sync_service_ip_address =
                "169.254.169.123"
          
Type=oneshot
 RemainAfterExit=no
 ExecStart=/bin/sh -c "printf \"[Time]\nNTP=${amazon_time_sync_service_ip_address}\n\" > /etc/systemd/timesyncd.conf"
ExecStartPost=/bin/systemctl restart systemd-timesyncd.service
 ‘’’

Hope it can be supported as part of this PR

ReillyProcentive:
I've been of the opinion that it would be better to use timesyncd instead of chrony or ntpd on all of the supported platforms in fact.

johngmyers:
timesyncd would indeed seem to be better. Still, this PR reasonably accomplishes what it is intending to do. A competing or revised PR would also be welcome.

simonmacklin:
I agree but this PR was only intended to add the missing cloud NTPs.  I will be more then happy to create another PR to switch over to timesyncd another time.

justinsb:
Thanks @simonmacklin 

/approve
/lgtm

nodeup/pkg/model/ntp.go
```diff
@@ -46,22 +57,103 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 		return nil
 	}
 
+	var ntpIP string
+	switch b.Cluster.Spec.CloudProvider {
+	case "aws":
+		ntpIP = "169.254.169.123"
+	case "gce":
+		ntpIP = "time.google.com"
```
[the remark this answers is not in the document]
simonmacklin: I think you are right. I got the information from this doc https://developers.google.com/time/faq but it looks like time.google.com is public for all people to use outside of GCP.

I can't resolve metadata.google.internal from my laptop but I assume it will when running on GCP. 

Shall I update?

rifelpet:
It looks like this is causing issues on the [kuberouter e2e job](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-cni-kuberouter) which uses [Ubuntu 18.04](https://github.com/kubernetes/test-infra/blob/a0cadb92a2b48f5bbff706038ab709c9413002be/config/jobs/kubernetes/kops/kops-periodics-network-plugins.yaml#L229).
[nodeup is failing on all nodes](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-cni-kuberouter/1238966489453694977/artifacts/13.52.218.173/journal.log):

`Mar 14 23:15:07.576966 ip-172-20-54-66 nodeup[1634]: W0314 23:15:07.576951    1634 main.go:138] got error running nodeup (will retry in 30s): error building loader: open /etc/ntp.conf: no such file or directory`

rifelpet:
[Some research](https://ubuntu.com/blog/ubuntu-bionic-using-chrony-to-configure-ntp) shows that Ubuntu 18.04 uses chronyd rather than ntpd so we'll probably need to update the logic to be more specific than just debian family vs rhel family.

johngmyers:
Looks like Ubuntu uses timesyncd instead.

simonmacklin:
When I wrote the code back in Jan the e2e tests all passed.  Has Ubuntu 18.04 recently been added to the e2e tests?

hakman:
@simonmacklin the e2e tests in PR are run only on Debian 9. There are more extensive e2e tests that run periodically on master:
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#Summary

simonmacklin:
@hakman 

Right that makes sense.  

Do you think this could just be a race condition or maybe the wrong config path for ubuntu?  I understand like @johngmyers  said the default is timesyncd.  But nodeup installs ntpd anyway https://github.com/kubernetes/kops/blob/master/nodeup/pkg/model/ntp.go#L71 

So I am wondering if nodeup is trying to write the config before it installs the package.

I will have a closer look when I get home tonight 

Simon

hakman:
@simonmacklin I think the problem is the `updateNtpIP` function:
https://github.com/kubernetes/kops/blob/2b04d7d1114f5d49ea26ba67a5818daf99e4de92/nodeup/pkg/model/ntp.go#L110-L119

This runs before all the tasks are created and run. The config file will not exist at all at that time if the package is not installed.

IMHO, the correct solution would be to write the full config files as we do everywhere else in nodeup and as @justinsb suggested: https://github.com/kubernetes/kops/pull/8312#discussion_r392611069.

simonmacklin:
Wouldn't line https://github.com/kubernetes/kops/blob/2b04d7d1114f5d49ea26ba67a5818daf99e4de92/nodeup/pkg/model/ntp.go#L107 install ntp first before running the code under it?

hakman:
> Wouldn't line install ntp first before running the code under it?

`AddTask` just creates and queues a task to install. Ordering and the actual action is done later.

simonmacklin:
Right OK thanks @hakman

PR #8313, opened 2020-01-11
WIP Option to surge during rolling update

johngmyers:
Adds a `MaxSurge` field to the cluster/instancegroup `RollingUpdate` struct.

/kind feature
/area rolling-update


johngmyers:
WIP because it depends on #8271 landing first.

pkg/instancegroups/instancegroups.go
```diff
@@ -152,17 +149,61 @@ func (r *RollingUpdateInstanceGroup) RollingUpdate(rollingUpdateData *RollingUpd
...
+	}
+	maxConcurrency := maxSurge + settings.MaxUnavailable.IntValue()
 
 	if maxConcurrency == 0 {
 		klog.Infof("Rolling updates for InstanceGroup %s are disabled", r.CloudGroup.InstanceGroup.Name)
 		return nil
 	}
 
+	if r.CloudGroup.InstanceGroup.Spec.Role == api.InstanceGroupRoleMaster && maxSurge != 0 {
+		// Masters are incapable of surging because they rely on registering themselves through
```
justinsb: Not sure how chatty it would be if users don't set any values, but a warning might be great here if we're going to override user settings.

pkg/instancegroups/instancegroups.go
```diff
@@ -152,17 +149,61 @@ func (r *RollingUpdateInstanceGroup) RollingUpdate(rollingUpdateData *RollingUpd
...
+	if maxSurge > 0 && !rollingUpdateData.CloudOnly {
+		for numSurge := 1; numSurge <= maxSurge; numSurge++ {
+			u := update[len(update)-numSurge]
+			if !u.Detached {
+				if err := r.detachInstance(u); err != nil {
+					return err
+				}
+
+				// If noneReady, wait until after one node is detached and its replacement validates
+				// before the detaching more in case the current spec does not result in usable nodes.
```
justinsb: Nit typo in "before the detaching more"

justinsb:
I think this looks good - it's a clever idea to track the surge state using a tag on the infrastructure level - that's where we basically came unstuck previously.

I'm going to try to think a little bit more about this today (and we should probably talk about it during office hours), but I'm inclined to merge ... particularly if I can satisfy myself it works on non-AWS also :-)

johngmyers:
Other cloud providers have the option of either tagging/detaching like AWS does or temporarily increasing the desired size of the underlying ASG. When I designed out an interface to be agnostic to this implementation choice yet handle all the failure cases, it ended up looking the same as the detach interface in this PR.

pkg/instancegroups/instancegroups.go
johngmyers: I believe we should put on the `MaxSurge` field's documentation `Does not have any effect on instance groups with role "master".`

I could add an api validation that throws a `field.Forbidden` if `MaxSurge` is explicitly set to a non-zero value on an InstanceGroup with role "master", because that just doesn't make sense.

For the case where the user set a nonzero default `MaxSurge` on the Cluster but didn't override that at the InstanceGroup level, I believe the best user experience would be for kops to silently do the right thing. We shouldn't make the user have to explicitly override the value on each of their master InstanceGroups just to get rid of log noise.

justinsb: Ah - I missed the override point.  I agree with either/both of the suggestions, I also don't consider them a blocker to merging this.

justinsb:
Thanks @johngmyers - this is really great stuff

/approve
/lgtm

johngmyers:
One of my added tests has a flake. I was able to reproduce it once in 100 runs. Will investigate.

rifelpet:
It'd be great to have some documentation for this feature as well. perhaps in docs/instance_groups.md ?

johngmyers:
@rifelpet see #8673

PR #8314, opened 2020-01-11
WIP: Move kubectl dependency to k8s.io/kubectl

justinsb:
Helps pare down our dependency on kubernetes/kubernetes


justinsb:
Actually, this has to be incorporated into #8055 and can't be split out - cross-dependencies.

PR #8316, opened 2020-01-11
Cilium eni

olemarkus:
With this feature enabled, Cilium will create and allocate ENIs for pods similar to LyftVPC and AmazonVPC.

Marked this as work in progress since there are a couple of other Cilium PRs that should go in first.

### DIFF
--- a/nodeup/pkg/model/context.go
+++ b/nodeup/pkg/model/context.go
@@ -360,11 +360,7 @@ func (c *NodeupModelContext) UseNodeAuthorizer() bool {
 
 // UsesSecondaryIP checks if the CNI in use attaches secondary interfaces to the host.
 func (c *NodeupModelContext) UsesSecondaryIP() bool {
-	if (c.Cluster.Spec.Networking.CNI != nil && c.Cluster.Spec.Networking.CNI.UsesSecondaryIP) || c.Cluster.Spec.Networking.AmazonVPC != nil || c.Cluster.Spec.Networking.LyftVPC != nil {
-		return true
-	}
-
-	return false
+	return (c.Cluster.Spec.Networking.CNI != nil && c.Cluster.Spec.Networking.CNI.UsesSecondaryIP) || c.Cluster.Spec.Networking.AmazonVPC != nil || c.Cluster.Spec.Networking.LyftVPC != nil || (c.Cluster.Spec.Networking.Cilium != nil && c.Cluster.Spec.Networking.Cilium.Ipam == kops.CiliumIpamEni)
 }
 
 // UseBootstrapTokens checks if we are using bootstrap tokens
--- a/pkg/apis/kops/validation/legacy.go
+++ b/pkg/apis/kops/validation/legacy.go
@@ -204,7 +204,7 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 				allErrs = append(allErrs, field.Invalid(fieldSpec.Child("nonMasqueradeCIDR"), nonMasqueradeCIDRString, "Cluster had an invalid nonMasqueradeCIDR"))
 			}
 
-			if networkCIDR != nil && subnet.Overlap(nonMasqueradeCIDR, networkCIDR) && c.Spec.Networking != nil && c.Spec.Networking.AmazonVPC == nil && c.Spec.Networking.LyftVPC == nil {
+			if networkCIDR != nil && subnet.Overlap(nonMasqueradeCIDR, networkCIDR) && c.Spec.Networking != nil && c.Spec.Networking.AmazonVPC == nil && c.Spec.Networking.LyftVPC == nil && (c.Spec.Networking.Cilium == nil || c.Spec.Networking.Cilium.Ipam != kops.CiliumIpamEni) {
 				allErrs = append(allErrs, field.Invalid(fieldSpec.Child("nonMasqueradeCIDR"), nonMasqueradeCIDRString, fmt.Sprintf("nonMasqueradeCIDR %q cannot overlap with networkCIDR %q", nonMasqueradeCIDRString, c.Spec.NetworkCIDR)))
 			}
 
@@ -602,8 +602,22 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 
 	allErrs = append(allErrs, newValidateCluster(c)...)
 
-	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil && c.Spec.Networking.Cilium.EnableNodePort && c.Spec.KubeProxy != nil && *c.Spec.KubeProxy.Enabled {
-		allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubeProxy").Child("enabled"), "When kilium NodePort is enabled, kubeProxy must be disabled"))
+	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil {
+		ciliumSpec := c.Spec.Networking.Cilium
+
+		if ciliumSpec.EnableNodePort && c.Spec.KubeProxy != nil && *c.Spec.KubeProxy.Enabled {
+			allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubeProxy").Child("enabled"), "When Cilium NodePort is enabled, kubeProxy must be disabled"))
+		}
+
+		if ciliumSpec.Ipam == kops.CiliumIpamEni {
+			if c.Spec.CloudProvider != string(kops.CloudProviderAWS) {
+				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("cilium").Child("ipam"), "Cilum ENI IPAM is supported only in AWS"))
+			}
+			if !ciliumSpec.DisableMasquerade {
+				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("cilium").Child("disableMasquerade"), "Masquerade must be disabled when ENI IPAM is used"))
+			}
+
+		}
 	}
 
 	return allErrs
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -192,6 +192,10 @@ func (b *PolicyBuilder) BuildAWSPolicyMaster() (*Policy, error) {
 		addLyftVPCPermissions(p, resource, b.Cluster.Spec.IAM.Legacy, b.Cluster.GetName())
 	}
 
+	if b.Cluster.Spec.Networking != nil && b.Cluster.Spec.Networking.Cilium != nil && b.Cluster.Spec.Networking.Cilium.Ipam == kops.CiliumIpamEni {
+		addCiliumEniPermissions(p, resource, b.Cluster.Spec.IAM.Legacy)
+	}
+
 	return p, nil
 }
 
@@ -834,6 +838,34 @@ func addRomanaCNIPermissions(p *Policy, resource stringorslice.StringOrSlice, le
 }
 
 func addLyftVPCPermissions(p *Policy, resource stringorslice.StringOrSlice, legacyIAM bool, clusterName string) {
+	if legacyIAM {
+		// Legacy IAM provides ec2:*, so no additional permissions required
+		return
+	}
+
+	p.Statement = append(p.Statement,
+		&Statement{
+			Effect: StatementEffectAllow,
+			Action: stringorslice.Slice([]string{
+				"ec2:DescribeSubnets",
+				"ec2:AttachNetworkInterface",
+				"ec2:AssignPrivateIpAddresses",
+				"ec2:UnassignPrivateIpAddresses",
+				"ec2:CreateNetworkInterface",
+				"ec2:DescribeNetworkInterfaces",
+				"ec2:DescribeVpcPeeringConnections",
+				"ec2:DescribeSecurityGroups",
+				"ec2:DetachNetworkInterface",
+				"ec2:DeleteNetworkInterface",
+				"ec2:ModifyNetworkInterfaceAttribute",
+				"ec2:DescribeVpcs",
+			}),
+			Resource: resource,
+		},
+	)
+}
+
+func addCiliumEniPermissions(p *Policy, resource stringorslice.StringOrSlice, legacyIAM bool) {
 	if legacyIAM {
 		// Legacy IAM provides ec2:*, so no additional permissions required
 		return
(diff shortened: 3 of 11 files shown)


upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -625,6 +633,19 @@ spec:
           periodSeconds: 10
           timeoutSeconds: 3
       hostNetwork: true
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
johngmyers: This is more than "let" cilium-operator run on master nodes. This makes it always run on master nodes, does it not?

johngmyers:
Need to `make crds`

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
olemarkus: Indeed. The cilium-operator is responsible for provisioning ENIs and I only added the elevated privileges to master nodes.

johngmyers: But this also affects the Ipam != "eni" case.

olemarkus: I know. I am not sure that is a bad thing. The operator is getting increasingly more important and it may make sense that it runs as soon as the masters are up.

If there are any concerns, I can wrap that in an `if` but from my side, I think it makes sense that it unconditionally runs on masters.

rifelpet: I'd lean towards following the official upstream manifest when reasonable. Does it have the nodeSelector set?

johngmyers: Such a change should be called out more clearly. It should probably not be backported to 1.12, but only take effect on 1.18 on.

olemarkus: > I'd lean towards following the official upstream manifest when reasonable. Does it have the nodeSelector set?

Upstream, the focus has been on replacing AmazonVPC ENI provisioning with this, so there are no master nodes.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -127,6 +127,7 @@ data:
   enable-endpoint-routes: "true"
   auto-create-cilium-node-resource: "true"
   blacklist-conflicting-routes: "false"
+  masquerade: "false"
```
johngmyers: Doesn't this conflict with the `masquerade` on line 120?

This seems a reasonable thing to do, but I don't see it in the upstream charts.

olemarkus: It isn't directly in the charts, but it is in the documentation: https://docs.cilium.io/en/v1.6/gettingstarted/aws-eni/#prepare-deploy-cilium

I will change this to forcing the config on line 120 instead.

olemarkus:
On Cilium operator running on master nodes: there are two alternatives
a) the privileged IAM permissions is given to all nodes (and thus all pods by default)
b) cilium operator runs on master and only master nodes have privileged IAM permissions

Cilium operator is something I would call a system component as without it running, normal pods won't spawn.

The question then becomes what kind of security concerns do we have that outweights pods having the escalated privileges on normal nodes.

olemarkus:
What I can do is only force the operator on masters if ipam is set to ENI or gate it with a dedicated config option

johngmyers:
@k8s-ci-robot you should label this needs-rebase.

pkg/apis/kops/validation/legacy.go
```diff
@@ -602,6 +602,23 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 
 	allErrs = append(allErrs, newValidateCluster(c)...)
 
+	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil {
+		ciliumSpec := c.Spec.Networking.Cilium
+		if ciliumSpec.Ipam == "eni" {
+			if c.Spec.CloudProvider != "aws" {
+				allErrs = append(allErrs, field.Invalid(fieldSpec.Child("Cilium").Child("Ipam"), "eni", "Cilum ENI IPAM is supported only in AWS"))
```
johngmyers: Use `field.Forbidden()`

pkg/apis/kops/validation/legacy.go
```diff
@@ -602,6 +602,23 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+			}
+			if !ciliumSpec.DisableMasquerade {
+				allErrs = append(allErrs, field.Invalid(fieldSpec.Child("Cilium").Child("DisableMasquerade"), "false", "Masquerade must be disabled when ENI IPAM is used"))
```
johngmyers: Also `field.Forbidden()`

Is masquerading forbidden with ENI? the Note block at https://docs.cilium.io/en/v1.6/gettingstarted/aws-eni/#prepare-deploy-cilium seems to suggest either way is possible.

pkg/apis/kops/validation/legacy.go
```diff
@@ -602,6 +602,23 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+			}
+
+		}
+	}
+
+	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil && c.Spec.Networking.Cilium.Ipam == "eni" && c.Spec.CloudProvider != "aws" {
```
johngmyers: Appears to be a duplicate of the check/error at line 608.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
johngmyers: The kops approach is that when upgrading the kops version we don't make significant changes to the way existing clusters work. Significant changes are gated on K8s version so they only manifest when either upgrading Kubernetes or enabling a new feature in the cluster spec. I believe moving cilium-operator to the master is sufficiently significant to at least require either the k8s version gate or setting the ipam field.

My preference would be to only move to master when ENI IPAM is used. I would prefer our clusters not unnecessarily run cilium-operator on masters.

I could see a feature for adding a `nodeSelector` and `tolerations` to cilium-operator.

pkg/apis/kops/validation/legacy.go
olemarkus: If I masq, cluster networking doesn't work. I am guessing Cilium is masking behind the wrong interface or something. I couldn't quite figure out why. It may be that one have to explicitly set which interface to mask behind. But then I am unsure if we can really know that across all distros.

Masqing only makes sense if the SG on the ENI cilium creates have egress rules. But there isn't really any benefit of doing that since because of the masq, traffic through the ENI reaches internet anyway.

So I think the easiest is to force masqing off for now and if anyone has any good reason for having masq on, we loosen that restriction.

olemarkus:
I ran into an issue where etcd-manager is listening on one IP, but both primary IPs are added to the etcd-manager hosts file:
```
xxx.141   etcd-events-b.internal.xxx
xxx.229   etcd-events-b.internal.xxx
xxx.24    etcd-events-c.internal.xxx
xxx.157   etcd-events-c.internal.xxx
xxx.236    etcd-events-a.internal.xxx
xxx.181    etcd-events-a.internal.xxx
```
This will randomly break the etcd-cluster. Need to look into how to change this behaviour.
I would think this also breaks the other ENI CNIs
/hold

rifelpet:
If it helps your troubleshooting, I just confirmed that amazonvpc CNI does not have that problem

```
# Begin host entries managed by etcd-manager[etcd] - do not edit
xxx.255	etcd-a.internal.xxx
xxx.142	etcd-b.internal.xxx
xxx.110	etcd-c.internal.xxx
# End host entries managed by etcd-manager[etcd]

# Begin host entries managed by etcd-manager[etcd-events] - do not edit
xxx.255	etcd-events-a.internal.xxx
xxx.142	etcd-events-b.internal.xxx
xxx.110	etcd-events-c.internal.xxx
# End host entries managed by etcd-manager[etcd-events]
```

olemarkus:
Note that it took quite a while for this to happen. But thanks, that helps.

Would be useful with some clues on how etcd manager gets the info it uses for hosts. That has been quite hard to track

rifelpet:
I believe [this](https://github.com/kopeio/etcd-manager/blob/6dae4f146797e7156c4bd30dade19955fcf30655/pkg/etcd/etcdserver.go#L214-L232) is where etcd-manager writes the map of hosts to /etc/hosts and I believe it comes from the member list within the etcd cluster itself. Can you use etcdctl to see the list of members?

olemarkus:
From what I can tell, it looks like the issue only happens when using coreos. When using the default kops AMIs, I cannot reproduce this behaviour

/hold cancel

nodeup/pkg/model/context.go
```diff
@@ -360,11 +360,7 @@ func (c *NodeupModelContext) UseNodeAuthorizer() bool {
 
 // UsesSecondaryIP checks if the CNI in use attaches secondary interfaces to the host.
 func (c *NodeupModelContext) UsesSecondaryIP() bool {
-	if (c.Cluster.Spec.Networking.CNI != nil && c.Cluster.Spec.Networking.CNI.UsesSecondaryIP) || c.Cluster.Spec.Networking.AmazonVPC != nil || c.Cluster.Spec.Networking.LyftVPC != nil {
-		return true
-	}
-
-	return false
+	return (c.Cluster.Spec.Networking.CNI != nil && c.Cluster.Spec.Networking.CNI.UsesSecondaryIP) || c.Cluster.Spec.Networking.AmazonVPC != nil || c.Cluster.Spec.Networking.LyftVPC != nil || (c.Cluster.Spec.Networking.Cilium != nil && c.Cluster.Spec.Networking.Cilium.Ipam == "eni")
```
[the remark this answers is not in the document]
rifelpet: agreed, I think another constant along side [this one](https://github.com/kubernetes/kops/blob/cfe95a0efd68c99498c7d152ddc04a66cc833ba8/pkg/apis/kops/networking.go#L194) would be good.

pkg/apis/kops/validation/legacy.go
```diff
@@ -602,6 +602,19 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 
 	allErrs = append(allErrs, newValidateCluster(c)...)
 
+	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil {
+		ciliumSpec := c.Spec.Networking.Cilium
+		if ciliumSpec.Ipam == "eni" {
+			if c.Spec.CloudProvider != "aws" {
```
rifelpet: nit: could use [kops.CloudProviderAWS](https://github.com/kubernetes/kops/blob/1702b78ce33e3e4b5f1170531651980c8bbc2c31/pkg/apis/kops/channel.go#L250) here.

pkg/apis/kops/validation/legacy.go
tgraf: In ENI mode, there will be multiple possible egress paths for traffic leaving a node. Based on the destination, the right egress interface will be chosen. Masquerading is typically only ever needed for traffic heading out to the interface. This is typically routed via the default route pointing out `eth0` although there can be node configurations where this is different.

For this reason, the Cilium ENI guide sets the `egressMasqueradingInterfaces` helm option to limit masquerading to traffic leaving on eth0:
```
  --set global.egressMasqueradeInterfaces=eth0 \
```

Masquerading is *only* required if no NAT/internet gateway is set up in AWS which will perform the masquerading.

pkg/model/iam/iam_builder.go
```diff
@@ -861,6 +865,34 @@ func addLyftVPCPermissions(p *Policy, resource stringorslice.StringOrSlice, lega
 	)
 }
 
+func addCiliumEniPermissions(p *Policy, resource stringorslice.StringOrSlice, legacyIAM bool, clusterName string) {
```
rifelpet: this clusterName parameter is unused and can probably be removed.

pkg/apis/kops/validation/legacy.go
```diff
@@ -602,8 +602,22 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 
 	allErrs = append(allErrs, newValidateCluster(c)...)
 
-	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil && c.Spec.Networking.Cilium.EnableNodePort && c.Spec.KubeProxy != nil && *c.Spec.KubeProxy.Enabled {
-		allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("KubeProxy").Child("enabled"), "When Cilium NodePort is enabled, KubeProxy must be disabled"))
+	if c.Spec.Networking != nil && c.Spec.Networking.Cilium != nil {
+		ciliumSpec := c.Spec.Networking.Cilium
+
+		if ciliumSpec.EnableNodePort && c.Spec.KubeProxy != nil && *c.Spec.KubeProxy.Enabled {
+			allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("KubeProxy").Child("enabled"), "When Cilium NodePort is enabled, KubeProxy must be disabled"))
```
johngmyers:
```suggestion
			allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubeProxy").Child("enabled"), "When Cilium NodePort is enabled, kubeProxy must be disabled"))
```

pkg/apis/kops/validation/legacy.go
```diff
@@ -602,8 +602,22 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+		}
+
+		if ciliumSpec.Ipam == kops.CiliumIpamEni {
+			if c.Spec.CloudProvider != string(kops.CloudProviderAWS) {
+				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("Cilium").Child("Ipam"), "Cilum ENI IPAM is supported only in AWS"))
```
johngmyers:
```suggestion
				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("cilium").Child("ipam"), "Cilum ENI IPAM is supported only in AWS"))
```

pkg/apis/kops/validation/legacy.go
```diff
@@ -602,8 +602,22 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+			}
+			if !ciliumSpec.DisableMasquerade {
+				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("Cilium").Child("DisableMasquerade"), "Masquerade must be disabled when ENI IPAM is used"))
```
johngmyers:
```suggestion
				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("cilium").Child("disableMasquerade"), "Masquerade must be disabled when ENI IPAM is used"))
```

PR #8321, opened 2020-01-12
Add release notes for deleting the kops-controller deployment

rifelpet:
This is a followup to #8273. I guessed at the exact release names for the next 1.16 and 1.17 releases but we can adjust those if necessary.


joshbranham:
Thanks!
/lgtm

PR #8322, opened 2020-01-12
Don't share /etc/hosts when using etcd-manager

justinsb:
With etcd-manager the DNS names should only be used by the
etcd-manager pod itself, so we don't need to share /etc/hosts with the
host.

By not sharing we avoid:

(1) the temptation to address etcd directly
(2) problems of concurrent updates to /etc/hosts being hard from within a container (because locking is difficult across bind mounts)

Introducing with kubernetes 1.17 to avoid changing behavior of existing versions.


mikesplain:
Thanks, makes sense, good idea.  We probably need to update tests for this too though. LGTM but will need another after tests are fixed 😄 

/lgtm

PR #8328, opened 2020-01-13
Add shellcheck verification

rifelpet:
https://github.com/koalaman/shellcheck

[Copied from the k/k repo](https://github.com/kubernetes/kubernetes/blob/7b6369c803cd06f0084ac57b2b6266000b739aea/hack/verify-shellcheck.sh) with some of the utility functions inlined, just like verify-staticcheck.sh.
Once merged I'll add it to prow.

Side note: I noticed that k/k has a single "verify" job. This will be kops' 9th "verify" job and most of them take well under 1 minute. I think we should consider consolidating most or all of our verify jobs into one just so that the GH list of checks isn't so long.


tanjunchen:
good job

johngmyers:
The various verify jobs aren't particularly good at making the failures visible through the Prow UI. It would be good if the errors were prefixed with whatever Prow highlights in red instead of filtering out.

rifelpet:
I can make a new job that combines all the verify jobs and make it optional. That way it won't be ran automatically and isn't considered blocking for a PR to pass. It needs to be triggered with an explicit `/test`, so we can experiment with that for a while to make the output more human friendly.

rifelpet:
adding that new job here: https://github.com/kubernetes/test-infra/pull/15879
will be invoked with `/test pull-kops-verify`

rifelpet:
and yes I agree it can be difficult to find the actual error message. My work last week of migrating our prow jobs to use new functionality made available from the the test-infra team has greatly reduced the extraneous output so that should help as well.

johngmyers:
Consider creating a `make verify`. It could possibly have `make travis-ci` plus `verify-gofmt`.

rifelpet:
I opened [this PR](https://github.com/kubernetes/test-infra/pull/15892) to update the target to just `verify`. I also updated this PR to add the new target, so with both merged we should be ready to make the new prow job blocking which will trigger it on every open kops PR as well as remove the redundant verify jobs.

rifelpet:
using the new `make verify` target for this prow job now:

/test pull-kops-verify

justinsb:
Thanks @rifelpet - shellcheck is great!  Did you hit an issue with the bash - is there something in particular we need to fix?

/approve
/lgtm

PR #7883, opened 2019-11-05
Upgrading coredns to 1.4.0 for clusters >=1.14.0.

michalschott:
From kube 1.14 changelog:

> There is a known issue coredns/coredns#2629 in CoreDNS 1.3.1, wherein if the Kubernetes API shuts down while CoreDNS is connected, CoreDNS will crash. The issue is fixed in CoreDNS 1.4.0 in coredns/coredns#2529.

PR #8333, opened 2020-01-14
CoreDNS default image bump to 1.6.6 to resolve CVE

### LINKED ISSUE #8309: CoreDNS v1.3.1 emptyDir workaround blocks node draining
 this information.**
1.15.0

 a `kops` flag.**
1.15.6

The emptyDir volume used in fix #7492 prevents cluster-autoscaler from evicting some nodes. Default behavior is to skip draining of nodes using emptyDir volumes.

We've identified a few workarounds:
- Remove CoreDNS emptyDir volume, which requires a CoreDNS update (#7883)
- Allow custom annotations on the CoreDNS pod, allowing us to add annotation `"cluster-autoscaler.kubernetes.io/safe-to-evict": "true"` (not possible with the current template)
- Setting `"skip-nodes-with-local-storage=false"` on the Cluster Autoscaler.

### LINKED ISSUE #8332: CoreDNS CVE - CVE-2019-19794
 this information.**

1.14.1

 a `kops` flag.**

v1.14.9

AWS

Install kops with CoreDNS as the DNS provider.

CoreDNS v1.3.1 running by default.

CoreDNS to be running version >= 1.6.6 by default.

N/A

N/A

Per the maintainers of CoreDNS this CVE could result in forgery of DNS responses and thus poisoning of the DNS cache of pods running on a cluster.

Resolving this will have the side effect of resolving #8309 for those running the CA without `--skip-nodes-with-local-storage=false` as well as #7883 

This has been disclosed to the `security@kubernetes.io` mailing list and confirmed as OK to disclose/discuss publicly. 

Going to raise a PR with the required changes.

gjtempleton:
Also updates the default Corefile config to make use of the new lameduck functionality for healthcheck: https://coredns.io/plugins/health/ as per kubeADM: https://github.com/kubernetes/kubernetes/pull/86260/file

Builds on top of the work done by @michalschott in #7883 however updates the image for all clusters.

Resolves #8332 
Resolves #8309

### DIFF
--- a/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
@@ -64,7 +64,9 @@ data:
   {{- else }}
     .:53 {
         errors
-        health
+        health {
+          lameduck 5s
+        }
         kubernetes {{ KubeDNS.Domain }}. in-addr.arpa ip6.arpa {
           pods insecure
           upstream
@@ -165,7 +167,7 @@ spec:
           beta.kubernetes.io/os: linux
       containers:
       - name: coredns
-        image: {{ if KubeDNS.CoreDNSImage }}{{ KubeDNS.CoreDNSImage }}{{ else }}k8s.gcr.io/coredns:1.3.1{{ end }}
+        image: {{ if KubeDNS.CoreDNSImage }}{{ KubeDNS.CoreDNSImage }}{{ else }}k8s.gcr.io/coredns:1.6.6{{ end }}
         imagePullPolicy: IfNotPresent
         resources:
           limits:
@@ -175,9 +177,6 @@ spec:
             memory: {{ KubeDNS.MemoryRequest }}
         args: [ "-conf", "/etc/coredns/Corefile" ]
         volumeMounts:
-        # Workaround for 1.3.1 bug, can be removed after bumping to 1.4+. See: https://github.com/coredns/coredns/pull/2529
-        - name: tmp
-          mountPath: /tmp
         - name: config-volume
           mountPath: /etc/coredns
           readOnly: true
--- a/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
@@ -214,9 +214,6 @@ spec:
             scheme: HTTP
       dnsPolicy: Default
       volumes:
-        # Workaround for 1.3.1 bug, can be removed after bumping to 1.4+. See: https://github.com/coredns/coredns/pull/2529
-        - name: tmp
-          emptyDir: {}
         - name: config-volume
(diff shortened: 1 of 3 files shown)


mikesplain:
This mostly looks good. You probably also want to bump the versions here: https://github.com/kubernetes/kops/blob/de4a86faa2d4724b0687b08dc7fb764150db852f/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go#L288 and here: https://github.com/kubernetes/kops/blob/de4a86faa2d4724b0687b08dc7fb764150db852f/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go#L307 Those aren't really required anymore but I think it's good for cleanliness of existing deployments and code.

gjtempleton:
Good point, have updated as suggested.

joshbranham:
Do we care about the version of CoreDNS as it pertains to K8s version? I know we had a PR a while back we chose not to merge (I think) because the version was ahead of what k/k was using?

hakman:
> Do we care about the version of CoreDNS as it pertains to K8s version? I know we had a PR a while back we chose not to merge (I think) because the version was ahead of what k/k was using?

Good point @joshbranham. The bump to 1.6.x should happen for 1.16+.
Also should be cherrypicked to those branches.

mikesplain:
As discussed during office hours, lets check with @rajansandeep who has previously be in the know with CoreDNS.  

Should we consider backporting 1.6.* to k8s 1.15 regardless of what k/k is doing @rajansandeep?

michalschott:
My three cents, I'd keep CoreDNS to whatever version is suggested to certain version (aka keep 1.3.1 for <= 1.15 and bump to 1.6.X for > 1.16).

thomaspeitz:
Thanks for your PR @gjtempleton 

We are applying your change manually currently. During git diff @kforsthoevel  and I found this small issue:
https://github.com/kubernetes/kops/blob/b4bfdcbfacea1ed2e415fff37bd21ca2305a36c2/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template#L217 - There is still the volume definition in there. You only removed the mounting. From my perspective we should remove the definition as well.

Beside that i really would enjoy a merge of it. Having a CVE in the cluster sounds awful.

Backstory:
We are upgrading to 1.6.6 mainly because we have dns issues since upgrading from 1.14 to 1.15 and want to verify if this solves it. It happens during node exchange.

gjtempleton:
Hey @thomaspeitz !

Good catch, thanks. I've just pushed up a commit to remove that as well, hadn't noticed as we're running the CA ignoring local storage.

justinsb:
AFAICT k/k is going to move to 1.6.6 (I haven't seen any objections), and we probably shouldn't have a CVE in the default versions that kops installs, which trumps our "don't change behaviour of old versions" rule.  Ideally CoreDNS (and all addons) would backport CVE fixes so we could just bump the patch for older k8s versions, but nobody is doing that (yet!) to my knowledge.

/approve
/lgtm

justinsb:
(I suspect e2e will actually fail because of the skew issue that may be fixed in https://github.com/kubernetes/test-infra/pull/16029 )

PR #8338, opened 2020-01-15
Don't load nonexistent calico-client cert when CNI is Cilium

johngmyers:
/kind bug

6d0133698d82f578b1a0e78f1dca444ed77cb4d5 stopped generating the "calico-client" keypair and certificate when the CNI was set to Cilium, as the new version of Cilium no longer needed to use the apiserver's etcd cluster. Unfortunately, it failed to stop adding the EtcdTLSBuilder in that case. The result of this is that if the cluster is not using etcd-manager and has etcd TLS enabled, masters will fail to come up looking for "calico-client" certificates that don't exist.

### DIFF
--- a/upup/pkg/fi/nodeup/command.go
+++ b/upup/pkg/fi/nodeup/command.go
@@ -254,7 +254,7 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
 	} else {
 		loader.Builders = append(loader.Builders, &model.KubeRouterBuilder{NodeupModelContext: modelContext})
 	}
-	if c.cluster.Spec.Networking.Calico != nil || c.cluster.Spec.Networking.Cilium != nil {
+	if c.cluster.Spec.Networking.Calico != nil {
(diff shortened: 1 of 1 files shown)


olemarkus:
That diff is a part of https://github.com/kubernetes/kops/pull/8220 too

johngmyers:
So it is. It is, however, a separate concern that may need to be backported further.

olemarkus:
Newer Cilium support doesn't really go back that far though. But just merge in this one if this can make it into the releases that are supposed to be cut this week

olemarkus:
I removed this change from my PR. It was not really related anyway. So /lgtm

justinsb:
Thanks for figuring this out @johngmyers & @olemarkus !

/approve
/lgtm

justinsb:
This one is likely to fail aws e2e tests due to the issue being addressed in https://github.com/kubernetes/test-infra/pull/16029 (i.e. not due to this PR)

PR #8343, opened 2020-01-16
Fix crossbuild-nodeup-in-docker

johngmyers:
/kind bug

`make crossbuild-nodeup-in-docker` was missing necessary changes made to `crossbuild-in-docker`.


rifelpet:
I tested this out and it works well, thanks for the fix!
/lgtm
/approve

PR #8348, opened 2020-01-16
update gophercloud dependency

zetaab:
includes gophercloud/gophercloud#1829


zetaab:
@rifelpet any ideas what is going on with this PR? I do not understand why that test fails.

rifelpet:
@zetaab hm did you run `make gomod` ? that should create the BUILD files that the job complains are missing.

zetaab:
@rifelpet yes I updated go.mod and then executed make gomod.

vendor/golang.org/x/crypto/ed25519/BUILD.bazel
```diff
@@ -2,9 +2,15 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library")
...
+    srcs = [
+        "ed25519.go",
+        "ed25519_go113.go",
+    ],
     importmap = "k8s.io/kops/vendor/golang.org/x/crypto/ed25519",
     importpath = "golang.org/x/crypto/ed25519",
     visibility = ["//visibility:public"],
-    deps = ["//vendor/golang.org/x/crypto/ed25519/internal/edwards25519:go_default_library"],
+    deps = [
+        "//vendor/crypto/ed25519:go_default_library",
```
rifelpet: This is the line that is causing the error. Looking at the added comments in the file below this, I think it might be a go version issue. Are you able to rerun `make gomod` with go 1.12?

https://github.com/kubernetes/kops/blob/release-1.16/go.mod#L3

zetaab: yeah it is maybe the problem.. compiled now with 1.12.9

zetaab:
nah same problem

rifelpet:
hm now I'm wondering if its the go version used in the prow job. I dont think we have direct control over that but there might be a specific image we should use.

https://github.com/kubernetes/test-infra/blob/7ac332e23f3bd07972de08d1879296e3cecf7ff9/config/jobs/kubernetes/kops/kops-config.yaml#L160

perhaps switching to one of the images from here:

https://github.com/kubernetes/test-infra/blob/f219618f288f4e41a6dc62ddd1f33db48f6e93c9/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml

it should be easy to confirm the go version in each of the images since they're all publicly available. i can try to do that when i have a chance later today.

rifelpet:
Heres a list of the "image variants" https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/variants.yaml but it shows go 1.13 for the 1.16 variant. I also see that k/k 1.16 is built with go 1.13: https://github.com/kubernetes/kubernetes/blob/release-1.16/build/root/WORKSPACE#L56 so maybe we should upgrade to match.

PR #8353, opened 2020-01-16
Fix DNS loop on Ubuntu 18.04 (Bionic)

hakman:
Ubuntu 18.04 (Bionic) periodic e2e is failing. Seems to be related to DNS issues:
https://github.com/coredns/coredns/blob/master/plugin/loop/README.md#troubleshooting-loops-in-kubernetes-clusters

### DIFF
--- a/nodeup/pkg/model/kubelet.go
+++ b/nodeup/pkg/model/kubelet.go
@@ -595,6 +595,12 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 		}
 	}
 
+	// In certain configurations systemd-resolved will put the loopback address 127.0.0.53 as a nameserver into /etc/resolv.conf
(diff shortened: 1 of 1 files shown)


nodeup/pkg/model/kubelet.go
```diff
@@ -595,6 +595,12 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 		}
 	}
 
+	// In certain configurations systemd-resolved will put the loopback address 127.0.0.53 as a nameserver into /etc/resolv.conf
+	// https://github.com/coredns/coredns/blob/master/plugin/loop/README.md#troubleshooting-loops-in-kubernetes-clusters
+	if b.Distribution == distros.DistributionBionic {
```
rifelpet: Maybe we should only set it if the user hasn't already set it themselves?

hakman: Updated

PR #8360, opened 2020-01-17
Test package versions as well as hashes

justinsb:
Extend the existing "unit" test to check package versions, because
some of the docker packages now have a '5:' prefix.

### DIFF
--- a/nodeup/pkg/model/docker_test.go
+++ b/nodeup/pkg/model/docker_test.go
@@ -17,12 +17,11 @@ limitations under the License.
 package model
 
 import (
-	"crypto/sha1"
-	"encoding/hex"
-	"io"
-	"net/http"
+	"fmt"
 	"os"
+	"os/exec"
 	"path"
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -30,6 +29,7 @@ import (
 	"k8s.io/kops/pkg/flagbuilder"
 	"k8s.io/kops/pkg/testutils"
 	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/util/pkg/hashing"
 )
 
 func TestDockerPackageNames(t *testing.T) {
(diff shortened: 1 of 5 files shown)


nodeup/pkg/model/docker_test.go
```diff
@@ -86,34 +88,80 @@ func TestDockerPackageHashes(t *testing.T) {
...
+		out, err := cmd.CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("error running rpm %s: %v", strings.Join(cmd.Args, " "), err)
+		}
+
+		version := strings.TrimSpace(string(out))
+		if version != expectedVersion {
+			return fmt.Errorf("unexpected version, actual=%q, expected=%q", version, expectedVersion)
+		}
+	} else if strings.HasSuffix(u, ".tgz") {
```
hakman: Packages for conatinerd have a 'tar.gz' suffix:
https://github.com/kubernetes/kops/pull/8199/files#diff-ecf61990c0cdda1c13ad07df0440a537R116-R135
```suggestion
	} else if strings.HasSuffix(u, ".tgz") || strings.HasSuffix(u, ".tar.gz") {
```

nodeup/pkg/model/docker_test.go
```diff
@@ -86,34 +88,80 @@ func TestDockerPackageHashes(t *testing.T) {
...
+		if "" != expectedVersion {
+			return fmt.Errorf("did not expect version for tgz package")
+		}
+	} else {
+		return fmt.Errorf("unexpected suffix for file (not .rpm or .deb)")
```
hakman: Maybe also list `.tgz` and `.tar.gz` and the file name.

nodeup/pkg/model/docker_test.go
justinsb: Good call - added!

nodeup/pkg/model/docker_test.go
justinsb: Thanks - done!

PR #8361, opened 2020-01-17
Publish update AMIs into the alpha channel

justinsb:
* Add 1.16, 1.17 images with correct docker versions (cf
  https://github.com/kubernetes-sigs/image-builder/pull/127)

* Regular package updates on other stretch images

### DIFF
--- a/channels/alpha
+++ b/channels/alpha
@@ -23,21 +23,27 @@ spec:
       providerID: aws
       kubernetesVersion: ">=1.10.0 <1.11.0"
     # Stretch is the default for 1.11 (for nvme)
-    - name: kope.io/k8s-1.11-debian-stretch-amd64-hvm-ebs-2019-09-26
+    - name: kope.io/k8s-1.11-debian-stretch-amd64-hvm-ebs-2020-01-17
       providerID: aws
       kubernetesVersion: ">=1.11.0 <1.12.0"
-    - name: kope.io/k8s-1.12-debian-stretch-amd64-hvm-ebs-2019-09-26
+    - name: kope.io/k8s-1.12-debian-stretch-amd64-hvm-ebs-2020-01-17
       providerID: aws
       kubernetesVersion: ">=1.12.0 <1.13.0"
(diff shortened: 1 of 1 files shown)


rifelpet:
@justinsb do you know if these have the more recent kernel version that addresses #8224 ?

justinsb:
I do think we should try to adopt buster as default - I assume in 1.18, though we could still change the default for 1.17.  There was the iptables blocker but I think that is now fixed.  Assuming I can get it working for 1.18 (and it has the fixes) I can also build for 1.17 at the same time (I think).

--

These images are the default debian stretch kernel; the kernel package is:

```
> dpkg -l | grep linux-image
ii  linux-image-4.9.0-11-amd64     4.9.189-3+deb9u2                  amd64        Linux 4.9 for 64-bit PCs
ii  linux-image-amd64              4.9+80+deb9u9                     amd64        Linux for 64-bit PCs (meta-package)
```

```
> uname -rvm
4.9.0-11-amd64 #1 SMP Debian 4.9.189-3+deb9u2 (2019-11-11) x86_64
```


That kernel does seem to have http://patchwork.ozlabs.org/patch/1032812/, but does not have http://patchwork.ozlabs.org/patch/937963/ AFAICT.

---

I peeked at buster, I think it will have the latest LTS kernel which is 4.19: https://packages.debian.org/buster/linux-image-amd64

Right now it looks like it'll be 4.19.67-2+deb10u2

Downloading the [source package](https://packages.debian.org/buster/linux-source-4.19) it seems to have both http://patchwork.ozlabs.org/patch/937963/ and http://patchwork.ozlabs.org/patch/1032812/ ... so that's good.

I think the best strategy here is to keep using the stock kernels for both stretch and buster, and just get the buster images going asap.

---

There is a backports repository, so we could install even newer kernels from there in a more-supported way but without having to build them; we may pick up new bugs though so I'm wary to do this.

rifelpet:
great, thanks for the investigation. I agree with the proposed strategy
/lgtm

PR #8372, opened 2020-01-18
Remove some ancient release notes and migration docs

johngmyers:


mikesplain:
I personally think we shouldn't be removing those release notes.  K/K doesn't so why should we?  I'm fine with removing some of the other files but I think the history is important here.  Most open source projects keep their release notes in some form in perpetuity, I'm not sure why we should be any different.

johngmyers:
The Calico CIDR migration docs are linked to by the 1.7 release notes, so we either maintain them (including the script that fails shellcheck), we have a dead link, or we remove the old release notes.

People can research ancient history by going back in the git logs. I don't think we should spend effort maintaining these ancient, obsolete documents.

olemarkus:
I vote for removing old migration docs. K8s docs are versioned so it is not comparable to our docs. Should we start using versioned docs as well in the future, those old migration docs and scripts will resurface. 

Related: #9107 

/lgtm

rifelpet:
I'm a bit torn on this. On one side, [k/k keeps all changelogs in their master branch](https://github.com/kubernetes/kubernetes/tree/master/CHANGELOG). On the other side, many open source projects truncate their changelogs, either to the most recent major version or something like what this PR is doing, and the changelogs still exist in git history.

What exactly do we gain by deleting them? not spend time fixing broken links? Anything else?

johngmyers:
We wouldn't have to maintain obsolete docs, such as those for the Calico CIDR migration, that are referenced by obsolete release notes.

Also, these early release notes aren't named consistently.

hakman:
@johngmyers Now that https://github.com/kubernetes/kops/pull/9770 was merged, should this be closed?

johngmyers:
Closing

Issue #8371, opened 2020-01-18
Update API reference docs to use go.dev


rifelpet:
We have many references to [godoc.org](https://github.com/kubernetes/kops/search?utf8=✓&q=godoc.org&type=) in our documentation. It is most useful for viewing the Kops API, ClusterSpec, InstanceGroupSpec, etc.

Recently [golang announced](https://blog.golang.org/go.dev) a new go.dev website that provides similar functionality. Once benefit of go.dev is that it supports versioning, so users can view the Kops API for specific Kops versions:
 
https://pkg.go.dev/k8s.io/kops@v1.11.0/pkg/apis/kops?tab=doc

Unfortunately I don't see any more recent versions than v1.11.0 on the site. This sounds very similar to the bug where `GO111MODULES=on go get k8s.io/kops` downloads v1.11.0. We should get that fixed, and then update our docs to reference go.dev.

Once kops.sigs.k8s.io supports versioned docs we could have them link to the same version on go.dev as well.

rifelpet:
I've determined the source of the 1.11.0 problem. go modules requires semver tags prefixed with `v`.

```
go list -m -versions k8s.io/kops
k8s.io/kops v1.4.0-alpha.1 v1.4.0 v1.4.1 v1.4.2 v1.4.3 v1.4.4 v1.10.0 v1.11.0
```

While the [GitHub release for 1.11](https://github.com/kubernetes/kops/releases/tag/1.11.0) is tagged `1.11.0`, there is also a [v1.11.0](https://github.com/kubernetes/kops/releases/tag/v1.11.0) tag of the same commit. There is no such tag for any of the more recent stable releases.

I think we should prefix our GitHub releases with `v`. [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/releases/tag/v1.11.0) follows this pattern and is recommended by the [go modules documentation](https://github.com/golang/go/wiki/Modules#semantic-import-versioning). I may spin this off into its own GH issue.

PR #8373, opened 2020-01-19
Kops releases - prefix git tags with v

rifelpet:
I noticed this while investigating #8371 and there have been other reports of issues when trying to import kops with go modules (I can't find the GH issues at the moment)

[Go modules require the v prefix](https://github.com/golang/go/wiki/Modules#modules), and [k/k also tags with the v prefix](https://github.com/kubernetes/kubernetes/releases/tag/v1.17.0).

We have some inconsistent tags already, for 1.11.0 we have tags of both `1.11.0` and `v1.11.0` which is the most recent tag with the prefix.
This is also why 1.11.0 is the default version imported by `go get`:

```
go get -v k8s.io/kops
go: downloading k8s.io/kops v1.11.0
```

and the latest version in `go list`:

```
go list -m -versions k8s.io/kops
k8s.io/kops v1.4.0-alpha.1 v1.4.0 v1.4.1 v1.4.2 v1.4.3 v1.4.4 v1.10.0 v1.11.0
```

I'm proposing we switch to only tagging with the v prefix. I'm only updating the actual git tag and not the entire version string used throughout kops due to its larger impact:

* Output by `kops version`
* Public URLs for kops assets
* Protokube tag

I'm hoping this is the least invasive way we can make this change.
If we think advanced notice is required, we could tag with both formats for a number of releases before tagging only with the v prefix.

/hold for Justin's approval
/cc @justinsb

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -585,10 +585,11 @@ ${CHANNELS}:
 .PHONY: release-tag
 release-tag:
 	git tag ${KOPS_RELEASE_VERSION}
+	git tag v${KOPS_RELEASE_VERSION}
 
 .PHONY: release-github
 release-github:
-	shipbot -tag ${KOPS_RELEASE_VERSION} -config .shipbot.yaml
+	shipbot -tag v${KOPS_RELEASE_VERSION} -config .shipbot.yaml
 
 # --------------------------------------------------
 # API / embedding examples


justinsb:
Yes, we should adopt the "v" prefix.  I thought we would have to do both, so I did it for 1.11.0 IIRC, but then I guess I forgot to make it systematic so I stopped doing it :-(

Should we support both?  I don't think it costs us much to do so, and it shouldn't break anyone, but I also don't see how we ever discover if people are relying on the form without the "v" if we keep supporting it...

johngmyers:
We have a chance to find any of our own reliance on the non-prefixed tag through alpha/beta releases. We also have the option of tagging without the prefix at a later time. I would lead with prefix only for 1.18. Maybe 1.17. I wouldn't object to both for 1.16.

rifelpet:
Ok, I've updated it so that now both tags are created, and the GH Release is created off of the v prefixed one only. The `git push --tags` [documented here](https://github.com/kubernetes/kops/blob/master/docs/development/release.md#tag-new-version) will push both tags.

Similar to many of our other recent changes, we could tag both in 1.16 stable and announce that 1.18  will only be tagged with the v prefix.

We could also consider retroactively creating v prefixed tags because as you mention, they dont cost anything. I'm fairly certain we can fake the git timestamp so that GH still sorts them in order. Perhaps we do this after tagging 1.16 stable, just for posterity?

PR #8379, opened 2020-01-19
Add Cilium.EnablePolicy back into templates

olemarkus:
This may have dropped out when we bumped Cilium to 1.6
It is possible to set this value in the cluster spec, but it isn't used anywhere


olemarkus:
As far as I can tell, the build skips uploading any artifacts like protokube etc. Which perhaps makes sense since I don't change any of them.
But then the download job in the kops-configuration service complains about hash mismatch when downloading these artifacts.

olemarkus:
A rebase fixed it, it seems :man_shrugging:

PR #8386, opened 2020-01-21
Fix scheduler policy configmap args

vvbogdanov87:
`--policy-configmap` and `--policy-configmap-namespace` were added as one string. The result spec is:
```yaml
spec:
  containers:
  - args:
    - --kubeconfig=/var/lib/kube-scheduler/kubeconfig
    - --leader-elect=true
    - --policy-configmap=scheduler-policy --policy-configmap-namespace=kube-system
    - --v=2
    - --logtostderr=false
    - --alsologtostderr
    - --log-file=/var/log/kube-scheduler.log
```
It leads to an error: `couldn't get policy config map kube-system/scheduler-policy --policy-configmap-namespace=kube-system: configmaps "scheduler-policy --policy-configmap-namespace=kube-system" not found`

The spec should be
```
spec:
  containers:
  - args:
    - --kubeconfig=/var/lib/kube-scheduler/kubeconfig
    - --leader-elect=true
    - --policy-configmap=scheduler-policy
    - --policy-configmap-namespace=kube-system
    - --v=2
    - --logtostderr=false
    - --alsologtostderr
    - --log-file=/var/log/kube-scheduler.log
```

The PR fixes the bug

### DIFF
--- a/nodeup/pkg/model/kube_scheduler.go
+++ b/nodeup/pkg/model/kube_scheduler.go
@@ -104,7 +104,7 @@ func (b *KubeSchedulerBuilder) buildPod() (*v1.Pod, error) {
 	flags = append(flags, "--kubeconfig="+"/var/lib/kube-scheduler/kubeconfig")
 
 	if c.UsePolicyConfigMap != nil {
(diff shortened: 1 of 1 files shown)


rifelpet:
/approve
/lgtm

Thanks! We should consider cherry picking this back to 1.16

Issue #8306, opened 2020-01-10
Support custom AWS IAM credentials for kops-controller


rifelpet:
kops-controller is a new component introduced for k8s 1.16+ clusters. It runs on the master nodes and interacts with the cloud provider. For AWS it uses the EC2 instance's instance profile to get credentials using the instance metadata service.

Often users have access to the instance profile locked down so that pods cannot use its credentials and instead require a more explicit method.

Kube2iam for example has a `--default-role` which it will use when intercepting requests for instance profile credentials from a pod that does not have a role-arn annotation. We use this with a permission-less role, so that pods without an annotation behave similar to EC2 instances without an instance profile. This behavior cannot easily be restricted to certain namespaces (in order to whitelist kube-system) or certain nodes (in order to whitelist the masters). This setup will break kops-controller.

I'm proposing that Kops supports specifying annotations on the kops-controller pods so that we can provide it a role-arn to be recognized by kube2iam and other pod identity solutions. This would be exposed through an API field in the ClusterSpec - exact path TBD.

rifelpet:
It will also help to document the exact AWS permissions required, for users providing their own IAM role. I believe it is just [ec2:DescribeInstances](https://github.com/kubernetes/kops/blob/ebd7a29612cd47a7613870621c23f93e29a10965/pkg/nodeidentity/aws/identify.go#L120).

joshbranham:
So the `kops-controller` just runs on Masters, and needs to read tags or also put tags?

rifelpet:
Currently it just reads the EC2 instance tags. That may change as we add more functionality to the controller though.

rifelpet:
@justinsb i forgot to mention this yesterday, any thought on this? I was thinking of a podAnnotations field for kops-controller in the cluster spec but I wasn’t sure where it would go exactly

johngmyers:
For kiam, we could just bake in the `iam.amazonaws.com/role` annotation directly. That could also serve as the documentation of what is needed. Kube2iam would seem to require either custom annotations or IAM role creation.

mazzy89:
Would this bind kops too much to use external toolings like `kiam` or `kube2iam`? Couldn't we adopt an approach more like the cloud controller manager?

rifelpet:
[...]

PR #8404, opened 2020-01-23
Add ability to customize weave annotation

gianrubio:
kiam requires the pod to annotate the aws iam role Arn to indicate which role should be assumed.


rifelpet:
Are you sure this is necessary? We were actually just discussing this in https://github.com/kubernetes/kops/issues/8306 and determined that for both kiam and kube2iam, they dont support pods with [hostNetwork: true](https://github.com/kubernetes/kops/pull/8404/files#diff-cb25f4455178e7432a13960d8d342613R239) so they automatically use the EC2 instance's Instance Profile and IAM role, and giving it a role annotation won't work.

gianrubio:
you're right @rifelpet, closing the ticket now

Issue #6942, opened 2019-05-11
Allow kops to generate kube-scheduler config flags.


rralcala:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
Based on https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/ there are several flags on deprecation path, so adding support for them right now doesn't feel like the right course of action.
This request is to have kops generate the kube-scheduler config file on the node, where these flags are supported so that someone who wants to have a higher qps limit for example, can do.

Currently that flag for example is not supported by kops.
**2. Feel free to provide a design supporting your feature request.**
This feature will allow the user to set the custom kube-scheduler parameters via the kops's config or set_cluster, and those will make kops generate the config file on the node instead of using arguments when kube-scheduler runs.

ckcd:
Agree!! i also need this feature in kops 1.12.

rralcala:
This was addressed in https://github.com/kubernetes/kops/pull/8407

PR #8407, opened 2020-01-24
Support additional kube-scheduler config parameters via config file

rralcala:
Mentioned in #6942

This change allows using the --config flag and a generated configfile to set
options that were not previously supported and the use via flags is deprecated.
(https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/)

I thought that it might be better to have them in a config file to ensure
support in newer kubernetes versions.

It also makes it easy to add more.


pkg/apis/kops/componentconfig.go
```diff
@@ -609,6 +609,12 @@ type KubeSchedulerConfig struct {
 	// which has been supported as far back as Kubernetes 1.7. The default depends on the version and the cloud provider
 	// as outlined: https://kubernetes.io/docs/concepts/storage/storage-limits/
 	MaxPersistentVolumes *int32 `json:"maxPersistentVolumes,omitempty"`
+	// Qps sets the maximum qps to send to apiserver after the burst quota is exhausted
+	Qps *float32 `json:"qps,omitempty" configfile:"ClientConnection.QPS"`
+	// Burst sets the maximum qps to send to apiserver after the burst quota is exhausted
+	Burst *float32 `json:"qps,omitempty" configfile:"ClientConnection.Burst"`
```
johngmyers:
```suggestion
	Burst *float32 `json:"burst,omitempty" configfile:"ClientConnection.Burst"`
```

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,119 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
johngmyers:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

pkg/apis/kops/componentconfig.go
rralcala: Sorry, fixed

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// ClientConnectionConfig for kube-scheduler
+type ClientConnectionConfig struct {
+	Burst      *int32   `yaml:"burst,omitempty"`
+	Kubeconfig *string  `yaml:"kubeconfig"`
+	QPS        *float32 `yaml:"qps,omitempty"`
+}
+
+// SchedulerConfig used to generate the config file
```
johngmyers:
```suggestion
// SchedulerConfig is used to generate the config file.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -609,6 +609,12 @@ type KubeSchedulerConfig struct {
...
+	// Overrides the default kubeconfig path.
```
johngmyers:
```suggestion
	// KubeConfig overrides the default kubeconfig path.
```

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,119 @@
...
+type SchedulerConfig struct {
+	APIVersion         string                  `yaml:"apiVersion"`
+	Kind               string                  `yaml:"Kind"`
```
johngmyers:
```suggestion
	Kind               string                  `yaml:"kind"`
```

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,119 @@
...
+	"reflect"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+
+	"k8s.io/klog"
+	"k8s.io/kops/util/pkg/reflectutils"
+)
+
+// ClientConnectionConfig for kube-scheduler
```
johngmyers: Missing a verb here.

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,130 @@
...
+// SchedulerConfig is used to generate the config file
+type SchedulerConfig struct {
+	APIVersion         string                  `yaml:"apiVersion"`
+	Kind               string                  `yaml:"kind"`
+	BindTimeoutSeconds *int64                  `yaml:"bindTimeoutSeconds,omitempty"`
+	ClientConnection   *ClientConnectionConfig `yaml:"clientConnection,omitempty"`
+}
+
+// BuildConfigYaml reflects the options interface and extracts the parameters for the config file
+func BuildConfigYaml(options interface{}) ([]byte, error) {
```
johngmyers: This is always called with a `*KubeSchedulerConfig`, so why do you declare it as `interface{}`?

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,130 @@
...
+	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog"
+	"k8s.io/kops/util/pkg/reflectutils"
+)
+
+// ClientConnectionConfig is used by kube-scheduler to talk to the api server
+type ClientConnectionConfig struct {
+	Burst      *int32   `yaml:"burst,omitempty"`
+	Kubeconfig *string  `yaml:"kubeconfig"`
```
johngmyers: It looks like this field has changed names since Kubernetes 1.9, which is the oldest version that kops master branch supports. So it is likely that you'll have to do a Kubernetes version check of some sort, either using the config file only on new versions or changing the format of the file based on the Kubernetes version.

pkg/apis/kops/v1alpha1/componentconfig.go
```diff
@@ -609,6 +609,12 @@ type KubeSchedulerConfig struct {
 	// which has been supported as far back as Kubernetes 1.7. The default depends on the version and the cloud provider
 	// as outlined: https://kubernetes.io/docs/concepts/storage/storage-limits/
 	MaxPersistentVolumes *int32 `json:"maxPersistentVolumes,omitempty"`
+	// Qps sets the maximum qps to send to apiserver after the burst quota is exhausted
+	Qps *resource.Quantity `json:"qps,omitempty" configfile:"ClientConnection.QPS"`
+	// Burst sets the maximum qps to send to apiserver after the burst quota is exhausted
+	Burst *int32 `json:"burst,omitempty" configfile:"ClientConnection.Burst"`
+	// KubeConfig overrides the default kubeconfig path.
+	KubeConfig *string `json:"kubeConfig,omitempty" configfile:"ClientConnection.Kubeconfig"`
```
johngmyers: The `configfile` tags here and in v1alpha2 seem unnecessary.

pkg/apis/kops/componentconfig.go
```diff
@@ -609,6 +609,12 @@ type KubeSchedulerConfig struct {
 	// which has been supported as far back as Kubernetes 1.7. The default depends on the version and the cloud provider
 	// as outlined: https://kubernetes.io/docs/concepts/storage/storage-limits/
 	MaxPersistentVolumes *int32 `json:"maxPersistentVolumes,omitempty"`
+	// Qps sets the maximum qps to send to apiserver after the burst quota is exhausted
+	Qps *resource.Quantity `json:"qps,omitempty" configfile:"ClientConnection.QPS"`
+	// Burst sets the maximum qps to send to apiserver after the burst quota is exhausted
+	Burst *int32 `json:"burst,omitempty" configfile:"ClientConnection.Burst"`
```
johngmyers: This is an `int32` in k/k, so why a pointer here?

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,130 @@
...
+	schedConfig.ClientConnection = new(ClientConnectionConfig)
+
+	walker := func(path string, field *reflect.StructField, val reflect.Value) error {
+		if field == nil {
+			klog.V(8).Infof("ignoring non-field: %s", path)
+			return nil
+		}
+		tag := field.Tag.Get("configfile")
+		if tag == "" {
+			klog.V(4).Infof("not writing field with no flag tag: %s", path)
```
johngmyers:
```suggestion
			klog.V(4).Infof("not writing field with no configfile tag: %s", path)
```

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,130 @@
...
+			// We want to descend - it could be a structure containing flags
+			return nil
+		}
+		if tag == "-" {
+			klog.V(4).Infof("skipping field with %q flag tag: %s", tag, path)
```
johngmyers:
```suggestion
			klog.V(4).Infof("skipping field with %q configfile tag: %s", tag, path)
```

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, err
+	}
+
+	return configFile, nil
+}
+
+func getValueFromStruct(keyWithDots string, object *SchedulerConfig) (*reflect.Value, error) {
+	keySlice := strings.Split(keyWithDots, ".")
+	v := reflect.ValueOf(object)
+	// iterate through field names ,ignore the first name as it might be the current instance name
```
johngmyers:
```suggestion
	// iterate through field names, ignoring the first name as it might be the current instance name
```

pkg/apis/kops/componentconfig.go
```diff
@@ -609,6 +609,12 @@ type KubeSchedulerConfig struct {
...
+	// KubeConfig overrides the default kubeconfig path.
+	KubeConfig *string `json:"kubeConfig,omitempty" configfile:"ClientConnection.Kubeconfig"`
```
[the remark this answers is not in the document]
johngmyers: Why is this user configurable?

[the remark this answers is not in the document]
rralcala: You're right, I'll remove it.

nodeup/pkg/model/kube_scheduler.go
```diff
@@ -79,6 +82,23 @@ func (b *KubeSchedulerBuilder) Build(c *fi.ModelBuilderContext) error {
 		})
 	}
 
+	if b.Cluster.Spec.KubeScheduler.KubeConfig == nil {
+		b.Cluster.Spec.KubeScheduler.KubeConfig = &defaultKubeConfig
```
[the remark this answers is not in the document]
rralcala: Ok

pkg/configbuilder/buildconfigfile.go
```diff
@@ -0,0 +1,130 @@
...
+	Kubeconfig *string  `yaml:"kubeconfig"`
+	QPS        *float64 `yaml:"qps,omitempty"`
+}
+
+// SchedulerConfig is used to generate the config file
+type SchedulerConfig struct {
+	APIVersion         string                  `yaml:"apiVersion"`
+	Kind               string                  `yaml:"kind"`
+	BindTimeoutSeconds *int64                  `yaml:"bindTimeoutSeconds,omitempty"`
+	ClientConnection   *ClientConnectionConfig `yaml:"clientConnection,omitempty"`
```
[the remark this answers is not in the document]
rralcala: The pointer translates to unset and leaves the field out of the config, I'll check that's not initialized

pkg/configbuilder/buildconfigfile.go
rralcala: I want to generalize it later so that it can be used with others

pkg/configbuilder/buildconfigfile.go
rralcala: Yup, I forgot, thanks!

pkg/apis/kops/v1alpha1/componentconfig.go
rralcala: ok

pkg/apis/kops/componentconfig.go
rralcala: To allow leaving the field out of the config and let kube-scheduler just go with the default.

pkg/configbuilder/buildconfigfile.go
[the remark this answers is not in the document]
johngmyers: The field is always initialized on line 52, so can never be `nil`. I don't see the reason to make it a pointer.

pkg/configbuilder/buildconfigfile.go
johngmyers: YAGNI applies. Later code can change it when it needs to.

pkg/apis/kops/componentconfig.go
johngmyers: If it's `int32` in k/k, the default is specified by a zero value. It is not necessary to make this a pointer to allow leaving the field out of the config.

pkg/apis/kops/v1alpha2/componentconfig.go
```diff
@@ -610,6 +610,12 @@ type KubeSchedulerConfig struct {
 	// which has been supported as far back as Kubernetes 1.7. The default depends on the version and the cloud provider
 	// as outlined: https://kubernetes.io/docs/concepts/storage/storage-limits/
 	MaxPersistentVolumes *int32 `json:"maxPersistentVolumes,omitempty"`
+	// Qps sets the maximum qps to send to apiserver after the burst quota is exhausted
+	Qps *resource.Quantity `json:"qps,omitempty" configfile:"ClientConnection.QPS"`
+	// Burst sets the maximum qps to send to apiserver after the burst quota is exhausted
+	Burst *int32 `json:"burst,omitempty" configfile:"ClientConnection.Burst"`
```
johngmyers: Would it make sense to change one of the e2e tests to fill in these two values? That would make it more likely we'd detect k/k changing the name or type of these settings.

nodeup/pkg/model/kube_scheduler.go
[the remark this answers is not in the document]
rralcala: So, I looked up for this can couldn't find the default, I guess that's why the default was living in this file before the change, so I just kept it.

rralcala:
Updated based on feedback, I'm ok with the e2e test, just let me know where it should be part of this PR or a separate one, the other things are addressed, about the generalization, we were facing qps problems with another daemon so I was planning to send that PR as a followup to this one to avoid changing to many things and also propose the overall idea first. Thanks for all the input so far!!

johngmyers:
`BuildConfigYaml` seems a bit specific as it knows that a `resource.Quantity` gets converted into a `float32`.

johngmyers:
I think the e2e test would be best as part of this PR.

rralcala:
> `BuildConfigYaml` seems a bit specific as it knows that a `resource.Quantity` gets converted into a `float32`.

flagbuilder.BuildFlagsList does the same and it's used in several places, and I couldn't use floatX in KubeSchedulerConfig so that's why I went with resource.Quantity since I found it being used in other places.

nodeup/pkg/model/kube_scheduler.go
```diff
@@ -112,16 +123,30 @@ func (b *KubeSchedulerBuilder) Build(c *fi.ModelBuilderContext) error {
...
 	if err != nil {
 		return nil, fmt.Errorf("error building kube-scheduler flags: %v", err)
 	}
-	// Add kubeconfig flag
-	flags = append(flags, "--config="+"/var/lib/kube-scheduler/config")
+	if useConfigFile {
+		flags = append(flags, "--config="+"/var/lib/kube-scheduler/config.yaml")
+	} else {
+		// Add kubeconfig flag
+		flags = append(flags, "--config="+defaultKubeConfig)
```
johngmyers:
```suggestion
		flags = append(flags, "--kubeconfig="+defaultKubeConfig)
```

nodeup/pkg/model/kube_scheduler_test.go
```diff
@@ -0,0 +1,48 @@
...
+	expect := []byte(
+		`apiVersion: kubescheduler.config.k8s.io/v1alpha1
+kind: KubeSchedulerConfiguration
+clientConnection:
+  kubeconfig: /var/lib/kube-scheduler/kubeconfig
+  qps: 3.1
+`)
+	qps, _ := resource.ParseQuantity("3.1")
+
+	s := &kops.KubeSchedulerConfig{Qps: &qps}
```
johngmyers: Test `Burst` as well?

pkg/apis/kops/v1alpha2/componentconfig.go
```diff
@@ -610,12 +610,6 @@ type KubeSchedulerConfig struct {
 	// which has been supported as far back as Kubernetes 1.7. The default depends on the version and the cloud provider
 	// as outlined: https://kubernetes.io/docs/concepts/storage/storage-limits/
 	MaxPersistentVolumes *int32 `json:"maxPersistentVolumes,omitempty"`
-	// Qps sets the maximum qps to send to apiserver after the burst quota is exhausted
-	Qps *resource.Quantity `json:"qps,omitempty" configfile:"ClientConnection.QPS"`
-	// Burst sets the maximum qps to send to apiserver after the burst quota is exhausted
-	Burst *int32 `json:"burst,omitempty" configfile:"ClientConnection.Burst"`
```
johngmyers: How is the admin going to configure `Qps` and `Burst`? I would think you would need these two fields, but without the `configfile` tags.

rralcala: They are deprecated as flags already, the goal of this PR was to add the alternative of using the --config option. I took it out of v1alpha[1|2] as per previous comment.

johngmyers: The previous comment was to remove the `configfile` struct tags, not the fields. There was also a comment questioning the need for the `KubeConfig` flag.

Right now I see no way for the admin to configure the `Qps` and `Burst` settings in the `kops.KubeSchedulerConfig` struct. If they can't be configured, what is their purpose?

rralcala: But what's the point of having the field without the configfile tag since they are not going to be rendered? It's true that the fields shouldn't have been removed though, so can you explain why the tag should be removed? Like the FeatureGates or LeaderElection flags are all there for all alphas.

johngmyers: The `v1alpha1` or `v1alpha2` struct is what gets parsed from the versioned resource. That gets converted to the `kops` struct. It is the `kops` struct that gets rendered, not the `v1alpha1` or `v1alpha2` struct.

rralcala: Hmm, nice, Thanks, It confused me that all the flags were there, I'll send the update in a few

rdrgmnzs:
Hi @rralcala can you add Qps and Burst back to pkg/apis/kops/v1alpha1/componentconfig.go and pkg/apis/kops/v1alpha2/componentconfig.go, that way we can re-gen the api machinery correctly for those API versions.

rdrgmnzs:
Thanks for all the work on the PR and addressing all the comments @rralcala ! We can split the e2e tests in a separate PR, will ping you on Slack directly to get that going.

/lgtm
/approve

rifelpet:
@rralcala it looks like this may have caused a regression in our E2E suite when running k8s 1.11 clusters:

https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-k8s-1.11

@hakman and I found that the error message in the [kube-scheduler log](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-1-11/1222287351741419525/artifacts/3.104.2.211/kube-scheduler.log) is:
```
no kind "KubeSchedulerConfiguration" is registered for version "kubescheduler.config.k8s.io/v1alpha1"
```

I see in the [changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#action-required-1) that `KubeSchedulerConfiguration` used to be under `componentconfig/v1alpha1` but was moved to `kubescheduler.config.k8s.io/v1alpha1` in k8s 1.12. Rather than supporting both api groups how about we increase the version check to only use the config file in k8s 1.12?

nodeup/pkg/model/kube_scheduler.go
```diff
@@ -34,21 +35,37 @@ import (
...
 
+const defaultKubeConfig = "/var/lib/kube-scheduler/kubeconfig"
+
 // Build is responsible for building the manifest for the kube-scheduler
 func (b *KubeSchedulerBuilder) Build(c *fi.ModelBuilderContext) error {
 	if !b.IsMaster {
 		return nil
 	}
-
+	useConfigFile := b.IsKubernetesGTE("1.11")
```
rifelpet: for a followup PR:

```suggestion
	useConfigFile := b.IsKubernetesGTE("1.12")
```

PR #8408, opened 2020-01-24
Refuse to read channels with addons with unparseable versions

johngmyers:
Ensure such things get noticed early.

(Having lost half a day due to such a thing.)

### DIFF
--- a/channels/pkg/api/BUILD.bazel
+++ b/channels/pkg/api/BUILD.bazel
@@ -1,9 +1,22 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "go_default_library",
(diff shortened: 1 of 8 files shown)


justinsb:
This error is going to be buried deep in the output from channels (in protokube).  Can we / should we surface the error even earlier?

I'm working on addons anyway (thinking I might be able to show a demo in kops office hours and get some feedback) and would like to make sure that we at least fix it in the revised approach!

johngmyers:
It's at least better than installing something that will always be updated.

We could throw a check into `BootstrapChannelBuilder.buildAddons()` but that's not necessarily the only thing that creates an `Addons` slice. I couldn't come up with a good unit test for `buildAddons` because there isn't a reliable way for a test to exercise all possible code paths.

justinsb:
I believe the e2e failure is k8s / test skew, hopefully fixed by https://github.com/kubernetes/test-infra/pull/16029 , not due to this PR.

PR #8423, opened 2020-01-27
Update AWS IAM Authenticator to 0.5.0

rifelpet:
[Release Notes](https://github.com/kubernetes-sigs/aws-iam-authenticator/releases/tag/v0.5.0)

I merged changes from these manifests:

* https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/v0.5.0/deploy/example.yaml
* https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/v0.5.0/deploy/iamidentitymapping.yaml

The new version supports replacing the ConfigMap with an IAMIdentityMapping custom resource, but the `--backend-mode` command argument isnt yet exposed through the kops API, so it will still only use configmaps. We can expose a BackendMode API field in a followup PR.

Since I know you also use it...
/cc @rdrgmnzs

### DIFF
--- a/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
@@ -1,3 +1,87 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: iamidentitymappings.iamauthenticator.k8s.aws
+spec:
+  group: iamauthenticator.k8s.aws
+  version: v1alpha1
+  scope: Cluster
+  names:
+    plural: iamidentitymappings
+    singular: iamidentitymapping
+    kind: IAMIdentityMapping
+    categories:
+    - all
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          required:
+          - arn
+          - username
+          properties:
+            arn:
+              type: string
+            username:
+              type: string
+            groups:
+              type: array
+              items:
+                type: string
+
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: aws-iam-authenticator
+rules:
+- apiGroups:
+  - iamauthenticator.k8s.aws
+  resources:
+  - iamidentitymappings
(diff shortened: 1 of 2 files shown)


upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
```diff
@@ -1,3 +1,78 @@
...
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: aws-iam-authenticator
+rules:
+- apiGroups:
+  - iamauthenticator.k8s.aws
+  resources:
+  - "*"
+  verbs:
+  - "*"
```
rdrgmnzs: @rifelpet I know you just copied this from the example, however giving the authenticator all permissions could potentially be dangerous no? As in if an attacker got access to this pod, corrupted a binary or otherwise found a way to inject data at this endpoint we could potentially be allowing someone to create new credentials that an administrator may not necessarily be watching for.

Should we change the verbs for the CRD to just get/list/watch?

rifelpet: hm you're right. I opened https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/287 but anticipate it being a while before that is merged. I'll update this too and if theres a problem upstream we can update kops in a followup PR.

upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
```diff
@@ -19,6 +96,9 @@ spec:
       labels:
         k8s-app: aws-iam-authenticator
     spec:
+      # use service account with access to
```
johngmyers: The preposition `to` is missing an object.

upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
```diff
@@ -19,10 +103,8 @@ spec:
       labels:
```
johngmyers: As I've noted in #8200, the `scheduler.alpha.kubernetes.io/critical-pod` annotation is no longer supported as of k8s 1.16.

rifelpet:
/hold cancel
now that the upstream PR has merged

PR #8425, opened 2020-01-28
Update support for Amazon Linux 2

hakman:
Amazon Linux 2 support is broken with Docker packages for CentOS, because of the missing `container-selinux` dependency. Even if this is added from the CentOS repo, it is missing its own dependencies. Sadly, this is not something that can be easily fixed.

The good news is that Amazon Linux 2 has SELinux disabled by default, so the `container-selinux` package is not needed. Static binaries for containerd or Docker should run just fine.

#8199 adds support for static binaries for latest containerd and can be used to setup a Kubernetes cluster on Amazon Linux 2 with minor changes.

This PR adds the following changes:
* separates Amazon Linux from CentOS and RHEL to not require installing `container-selinux`
* updates default containerd version to 1.3.2 (noting to worry, this is targeted for 1.18)
* adds simpler aliases for well known accounts

### DIFF
--- a/nodeup/pkg/distros/distribution.go
+++ b/nodeup/pkg/distros/distribution.go
@@ -24,18 +24,19 @@ import (
 type Distribution string
 
 var (
-	DistributionJessie      Distribution = "jessie"
-	DistributionDebian9     Distribution = "debian9"
-	DistributionDebian10    Distribution = "buster"
-	DistributionXenial      Distribution = "xenial"
-	DistributionBionic      Distribution = "bionic"
-	DistributionRhel7       Distribution = "rhel7"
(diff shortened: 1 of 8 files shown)


justinsb:
Thanks @hakman - it does feel like there are a few possibly unrelated changes in this PR (introducing Amazon System Linux 2, changing the containerd version, recognizing more aliases), and in general this might in future cause problems with cherry-picks, so it's probably a good idea to split them up into separate PRs.  (They're also easier to review so we _should_ be able to merge them faster ;-) )

But in this case they're relatively related, so we should just merge:

/approve
/lgtm

hakman:
Indeed, the new aliases could have been a separate commit. Thanks @justinsb  :)

PR #8430, opened 2020-01-29
Fixes regression in e2e tests

rralcala:
As mentioned in https://github.com/kubernetes/kops/pull/8407, this change makes the feature available from kubernetes 1.12 and newer due to incompatibility with 1.11


hakman:
Very fast @rralcala. Thanks! :)

rifelpet:
Thank you!
/approve
/lgtm

PR #8431, opened 2020-01-29
Remove v1alpha1 API

johngmyers:


hakman:
@johngmyers 1.17 release is still very far away. Wouldn't this make things that modify the API harder to cherrypick to 1.17 and lower?

johngmyers:
API modifications would not normally be something one should cherrypick.

hakman:
I think it happens quite often to cherrypick API changes. I would do this change late in the release cycle to avoid such issues.

olemarkus:
There will always be versions in early release stage. An idea may be to just not update alpha1 api also for those versions where it won't be removed.

hakman:
> There will always be versions in early release stage. An idea may be to just not update alpha1 api also for those versions where it won't be removed.

@olemarkus probably won't pass the tests after that. I think I tried a while ago.

johngmyers:
It's a tradeoff of the cost of doing it at cherrypick time for cherrypicked API changes versus the cost of doing it for API changes that won't be cherrypicked. We'll always have release branches.

rifelpet:
I looked this over and it lgtm but I'll let @justinsb approve it since he is much more familiar with this area of code
/lgtm

justinsb:
As discussed in office hours ... it's time :tada: 

Thanks for driving this @johngmyers 

/approve
/lgtm

PR #8433, opened 2020-01-29
Make it possible to enable Prometheus metrics for Cilium

olemarkus:
Another bug in the new templates is the loss of prometheus metrics.

Since there are more cilium components involved here, the old config param didn't make that much sense anymore, so I rewrote that part.

Since the property is removed, this sort of breaks BC, but nothing would have happened for those that did so no loss of functionality. 

/kind bug

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -2547,6 +2547,8 @@ spec:
                       items:
                         type: string
                       type: array
+                    agentPrometheusPort:
+                      type: integer
                     allowLocalhost:
                       type: string
                     autoDirectNodeRoutes:
@@ -2593,6 +2595,8 @@ spec:
                       type: boolean
                     enablePolicy:
                       type: string
+                    enablePrometheusMetrics:
+                      type: boolean
                     enableTracing:
                       type: boolean
                     enableipv4:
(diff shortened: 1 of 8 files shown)


olemarkus:
@rifelpet anything holding this one back?

k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -2666,8 +2666,6 @@ spec:
                       type: boolean
                     prefilterDevice:
                       type: string
-                    prometheusServeAddr:
-                      type: string
```
johngmyers: Do we allow removals like this?

olemarkus: That is indeed the question. At some point we do because it is just confusing for end users if we leave it in. And no one can be using that since it doesn't do anything.

If we have a more clear deprecation guide, I can do the removal in a second PR.

johngmyers: My understanding is that we need to leave them in unused. When we create a new API version, then we can remove it from the new version. My understanding might be inaccurate and I don't know how we're tracking items to change in the next API revision.

olemarkus: I removed this bit now. Can be fixed at a later time, I guess

rifelpet:
Ideally we could have this under test [here](https://github.com/kubernetes/kops/tree/master/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/cilium) but that could be tackled in a followup PR. You'd need to add the cilium addon manifest name to [this list](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/bootstrapchannelbuilder_test.go#L46) as well.

Thanks!

/lgtm
/approve

rifelpet:
And yes we should open a GH issue that tracks any breaking changes we want to make across the entire kops API.

PR #8439, opened 2020-01-29
Cherry pick of #8200: Add missing priorityClassName for critical pods

johngmyers:


justinsb:
Thanks @johngmyers 

I probably should have waited for this before cutting 1.17.0-alpha.2.  I can always cut alpha.3, but let's hold off at least for a few days unless you think otherwise :-)

/approve
/lgtm

PR #8445, opened 2020-01-30
Cilium - Add missing Identity Allocation Mode to Operator Template

daviddyball:
In the course of debugging [this Cilium Garbage Collection Issue](https://github.com/cilium/cilium/issues/9805) it was noted that the default templates used for the `cilium-operator` `Deployment` were missing the `--identity-allocation-mode` argument.

The default configuration assumes that the Cilium version is > `1.6` with CRD backed identity management and as such the operator deployment should launch with the correct arguments for CRD-backed identity management.

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
@@ -542,9 +542,16 @@ spec:
       containers:
       - args:
         - --debug=$(CILIUM_DEBUG)
+        - --identity-allocation-mode=$(CILIUM_IDENTITY_ALLOCATION_MODE)
         command:
         - cilium-operator
         env:
+        - name: CILIUM_IDENTITY_ALLOCATION_MODE
+          valueFrom:
+            configMapKeyRef:
+              key: identity-allocation-mode
+              name: cilium-config
+              optional: true
         - name: CILIUM_K8S_NAMESPACE
           valueFrom:
             fieldRef:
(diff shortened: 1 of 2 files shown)


daviddyball:
I signed it

daviddyball:
@rifelpet This seems to be passing all checks now. It can be merged any time if you think it's good to go.

rifelpet:
So the [comments in the configmap](https://github.com/kubernetes/kops/pull/8445/files#diff-a8ab815fd6520e2f47d65a53433323f7R10-R20) suggest commenting out `identity-allocation-mode` which would cause new pods in the Deployment to fail because of the new `configMapKeyRef`. I suppose we can make that case that if users are editing the configmap in this manifest then they should also be expected to edit the Deployment.

As long as we think thats a reasonable expectation, I'm fine with getting this merged. Thoughts?

daviddyball:
@rifelpet At the moment I don't think there is really an option to use non-CRD backed configuration with Kops deployments. Perhaps we can raise an additional issue to track adding support for non-CRD backed clusters and not let that hold up this PR?

rifelpet:
Is non-CRD backed configuration commonly used for cilium? If not then we may wait until someone asks for it before doing any work to support it. If you want to open an issue to track that, feel free. I agree that doesn't need to hold this PR up.

/lgtm
/approve

Thanks!

daviddyball:
Yeh, non-CRD is quite popular, but CRD-backed is also the easiest to get started... so it's different horses for different courses really. I'll raise another issue to track support for non-CRD-backed Cilium clusters :+1:

PR #8446, opened 2020-01-31
Use PriorityClassName instead of namespace in cluster validation

johngmyers:
Use the system-{cluster,node}-critical PriorityClassName instead of the kube-system Namespace to identify pods which can fail cluster validation.

The kube-system namespace has overloaded semantics, some of which grant pods therein unnecessary privileges. This causes people to run critical pods in other namespaces. These critical pods in other namespaces should be considered in cluster validation.

The system-{cluster,node}-critical PriorityClasses are much better indicators of what cluster validation is looking for. And I will soon need cluster validation to handle cluster-critical and node-critical pods differently.

### DIFF
--- a/pkg/validation/validate_cluster.go
+++ b/pkg/validation/validate_cluster.go
@@ -229,6 +229,14 @@ func (v *ValidationCluster) collectPodFailures(client kubernetes.Interface, node
 			})
 			return nil
 		}
+		if pod.Status.Phase == v1.PodUnknown {
+			v.addError(&ValidationError{
+				Kind:    "Pod",
+				Name:    pod.Namespace + "/" + pod.Name,
+				Message: fmt.Sprintf("%s pod %q is unknown phase", priority, pod.Name),
+			})
+			return nil
+		}
 		var notready []string
 		for _, container := range pod.Status.ContainerStatuses {
 			if !container.Ready {
--- a/pkg/validation/validate_cluster.go
+++ b/pkg/validation/validate_cluster.go
@@ -17,11 +17,14 @@ limitations under the License.
 package validation
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/url"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/pager"
 	"k8s.io/kops/upup/pkg/fi"
 
 	v1 "k8s.io/api/core/v1"
@@ -195,11 +198,6 @@ func (v *ValidationCluster) collectComponentFailures(client kubernetes.Interface
 }
 
 func (v *ValidationCluster) collectPodFailures(client kubernetes.Interface, nodes []v1.Node) error {
-	pods, err := client.CoreV1().Pods("kube-system").List(metav1.ListOptions{})
-	if err != nil {
-		return fmt.Errorf("error listing Pods: %v", err)
-	}
-
 	masterWithoutManager := map[string]bool{}
 	nodeByAddress := map[string]string{}
 	for _, node := range nodes {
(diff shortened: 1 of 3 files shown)


pkg/validation/validate_cluster.go
```diff
@@ -195,22 +198,32 @@ func (v *ValidationCluster) collectComponentFailures(client kubernetes.Interface
...
-			continue
+	err := pager.New(pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
+		return client.CoreV1().Pods(metav1.NamespaceAll).List(opts)
+	})).EachListItem(context.TODO(), metav1.ListOptions{}, func(obj runtime.Object) error {
+		pod := obj.(*v1.Pod)
+		priority := pod.Spec.PriorityClassName
+		if priority != "system-cluster-critical" && priority != "system-node-critical" {
+			return nil
+		}
+		if pod.Status.Phase == v1.PodSucceeded || pod.Status.Phase == v1.PodFailed {
```
rifelpet: Why is `PodFailed` skipped? Shouldn't that be a validation error?

johngmyers: It is unlikely to change status just by waiting. Like `PodSucceeded`, the pod is spent.

It also wasn't a validation error before.

rifelpet: I created a failed pod and looking at its [status](https://gist.github.com/rifelpet/ea3938ddbfe181ab327ed08e7fd7823a) and the [original code](https://github.com/kubernetes/kops/blob/cbee4dccbaa1a6c0490a671500db921c34b521b7/pkg/validation/validate_cluster.go#L216-L228) I would expect the container's status of NotReady to cause a validation error.

johngmyers: Okay, I walked that part back

pkg/validation/validate_cluster.go
```diff
@@ -212,17 +210,32 @@ func (v *ValidationCluster) collectPodFailures(client kubernetes.Interface, node
 		}
 	}
 
-	for _, pod := range pods.Items {
+	err := pager.New(pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
```
justinsb: Fancy :-)

pkg/validation/validate_cluster.go
```diff
@@ -212,17 +210,32 @@ func (v *ValidationCluster) collectPodFailures(client kubernetes.Interface, node
...
+		return client.CoreV1().Pods(metav1.NamespaceAll).List(opts)
+	})).EachListItem(context.TODO(), metav1.ListOptions{}, func(obj runtime.Object) error {
+		pod := obj.(*v1.Pod)
+		priority := pod.Spec.PriorityClassName
+		if priority != "system-cluster-critical" && priority != "system-node-critical" {
```
justinsb: Review note (no change needed): These used to be permitted only in kube-system, but as of https://github.com/kubernetes/kubernetes/pull/76310 (i.e. k8s 1.17) that is no longer the case.

pkg/validation/validate_cluster.go
```diff
@@ -212,17 +210,32 @@ func (v *ValidationCluster) collectPodFailures(client kubernetes.Interface, node
...
-			continue
+			return nil
 		}
 		if pod.Status.Phase == v1.PodPending {
 			v.addError(&ValidationError{
 				Kind:    "Pod",
-				Name:    "kube-system/" + pod.Name,
-				Message: fmt.Sprintf("kube-system pod %q is pending", pod.Name),
+				Name:    pod.Namespace + "/" + pod.Name,
+				Message: fmt.Sprintf("%s pod %q is pending", priority, pod.Name),
```
justinsb: Nit: might be nice to print the namespace in the message, but I guess we have it in the name anyway. (Though I do now wonder why we are printing the name of the pod at all, therefore)

justinsb:
Thanks @johngmyers - this LGTM.  We should also make a pass to ensure that all our pods are marked with an critical annotation.  I actually wonder if we can do this statically by parsing the manifests...

PR #8449, opened 2020-01-31
Return more errors at once during API validation

johngmyers:


pkg/apis/kops/validation/legacy.go
```diff
@@ -37,67 +36,66 @@ import (
...
 // ValidateCluster is responsible for checking the validity of the Cluster spec
-func ValidateCluster(c *kops.Cluster, strict bool) *field.Error {
+func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 	fieldSpec := field.NewPath("spec")
-	var err error
+	allErrs := field.ErrorList{}
 
 	// kubernetesRelease is the version with only major & minor fields
-	var kubernetesRelease semver.Version
+	kubernetesRelease := semver.Version{Major: 1, Minor: 15}
```
rifelpet: why is this variable initialized to a value now? And will this version need to change on every Kops minor release?

johngmyers: We used to return early when it was unspecified. Now we need it to have some value so subsequent checks don't totally choke when they test it. It doesn't have to be changed on every minor release, but it it would be good if it were occasionally updated to still be in the supported range.

rifelpet: Ah I see. I wonder if it would be worth using the same default version that `kops create cluster` uses: https://github.com/kubernetes/kops/blob/6b4b00395470b2488a4c0bb33c361461ea38f80a/cmd/kops/create_cluster.go#L467

or if its not worth the added complexity.

johngmyers: It would need a `Channel`, which seems complex and outside the scope of apimachinery validation.

rifelpet: agreed. Perhaps a comment that explains why it is set and that it doesn't need to directly match each minor release would help future readers.

johngmyers: Added.

rifelpet:
Thanks for your work on this!
/lgtm
/approve

PR #8450, opened 2020-01-31
docs: trivial fix some typos

truongnh1992:
Although it is spelling mistakes, it might make an affects while reading.

Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>

### DIFF
--- a/docs/node_authorization.md
+++ b/docs/node_authorization.md
@@ -9,7 +9,7 @@ The [node authorization service] is an experimental service which in the absence
 - the client certificate by default is added into the system:nodes rbac group _(note, if you are using PSP this is automatically bound by kops on your behalf)_.
 - the kubelet at this point has a server certificate and the client api certificate and good to go.
 
-#### **Integretion with Kops**
+#### **Integration with Kops**
 
 The node authorization service is run on the master as a daemonset, by default dns is _node-authorizer-internal.dns_zone_:10443 and added via same mechanism at the internal kube-apiserver i.e. annotations on the kube-apiserver pods which is picked up the dns-controller and added to the dns zone.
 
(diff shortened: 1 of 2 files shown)


docs/releases.md
```diff
@@ -22,7 +22,7 @@
 
 #### 4. Release kops to homebrew
 
-* Following the [documentation](development/homebrew.md) we must release a compatible homebrew formulae with the release.
```
rifelpet: I think this is actually the correct spelling

https://formulae.brew.sh/

docs/calico-v3.md
```diff
@@ -31,7 +31,7 @@ will install v3.
 ### etcd
 Due to the etcd v3 API being a requirement of Calico Version 3
 (when using etcd as the datastore) not all Kops installations will be
-upgradable to Calico V3. Installations using etcd v2 (or earlier) will need
```
mikesplain: Not that it matters too much, but I believe both of these are correct.  Most sites say they can be used interchangeably, and there's a similiar debate with the word [upgradable](https://english.stackexchange.com/questions/56431/updatable-vs-updateable-which-is-correct) and the corresponding [ngrams](https://books.google.com/ngrams/graph?content=upgradable%2Cupgradeable&year_start=1800&year_end=2000&corpus=0&smoothing=3&share=&direct_url=t1%3B%2Cupgradable%3B%2Cc0%3B.t1%3B%2Cupgradeable%3B%2Cc0).

I think my preference goes to `upgradable`.

docs/releases.md
mikesplain: Agreed on this.

PR #8461, opened 2020-02-02
Use BindMount task for /home/kubernetes/bin

justinsb:
Without this, a reboot on COS does not remount /home/kubernetes/bin with exec permission.

### DIFF
--- a/nodeup/pkg/model/directories.go
+++ b/nodeup/pkg/model/directories.go
@@ -40,11 +40,12 @@ func (b *DirectoryBuilder) Build(c *fi.ModelBuilderContext) error {
 			Path: dirname,
 			Type: nodetasks.FileType_Directory,
 			Mode: s("0755"),
+		})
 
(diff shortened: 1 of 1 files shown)


nodeup/pkg/model/directories.go
```diff
@@ -40,11 +40,12 @@ func (b *DirectoryBuilder) Build(c *fi.ModelBuilderContext) error {
...
 			Type: nodetasks.FileType_Directory,
 			Mode: s("0755"),
+		})
 
-			OnChangeExecute: [][]string{
-				{"/bin/mount", "--bind", dirname, dirname},
-				{"/bin/mount", "-o", "remount,exec", dirname},
-			},
+		c.AddTask(&nodetasks.BindMount{
+			Source:     src,
```
johngmyers:
```suggestion
			Source:     dirname,
```
?

PR #8462, opened 2020-02-02
Tag EBS volumes when using launch templates with AWS API target

johngmyers:


rifelpet:
we can add terraform and cloudformation support in a separate PR

/approve
/lgtm

PR #8464, opened 2020-02-03
DNS: Don't try to apply empty changesets

justinsb:
For GCE this was resulting in confusing error messages.

### DIFF
--- a/dnsprovider/pkg/dnsprovider/providers/google/clouddns/rrchangeset.go
+++ b/dnsprovider/pkg/dnsprovider/providers/google/clouddns/rrchangeset.go
@@ -17,6 +17,7 @@ limitations under the License.
 package clouddns
 
 import (
+	"context"
 	"fmt"
 
 	"k8s.io/kops/dnsprovider/pkg/dnsprovider"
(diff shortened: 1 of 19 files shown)


dnsprovider/pkg/dnsprovider/providers/google/clouddns/rrchangeset.go
```diff
@@ -115,7 +119,7 @@ func (c *ResourceRecordChangeset) Apply() error {
 }
 
 func (c *ResourceRecordChangeset) IsEmpty() bool {
-	return len(c.additions) == 0 && len(c.removals) == 0
+	return len(c.additions) == 0 && len(c.removals) == 0 && len(c.upserts) == 0
```
johngmyers: Shouldn't there be a unit test for this, covering all of the implementations of the interface method? The only prior caller to this method I could find was a unit test for the case of it being empty. There should be tests for one addition, one removal, and one upsert.

justinsb: Yes, there should be.  We inherited a lot of this code from k/k, and the design is pretty different from the rest of the code (even the rest of the code in k/k).  It's hard to resist just rewriting the whole thing...

justinsb:
It should be the second contract, I believe.  I'll add a comment and fix up the others!

justinsb:
Documented the contract, changed the signature to force a code change to every call & definition, added a generic test of the contract behaviour that was problematic.

johngmyers:
Looks good, aside from the gofmt failure.

PR #8466, opened 2020-02-03
Update tags support for LaunchTemplates

hakman:
Launch Templates are the recommended way of creating an ASG in AWS and allows specifying various options that are not possible with Launch Configurations.
https://docs.aws.amazon.com/autoscaling/ec2/userguide/LaunchTemplates.html

This is a followup of #8462 with the following improvements:
* adds remaining cloud tags
* checks for changes before applying new tags

This is a bit more complicated as it has to account Launch Configurations, Mixed Instance Policies and Launch Templates at same time.
A lot of testing is also required to avoid breaking existing features.

@johngmyers as I said during office hours, I am looking into this for some time. If you plan to do any other changes in this area, we should sync so that we don't duplicate the effort.

### DIFF
--- a/pkg/model/context.go
+++ b/pkg/model/context.go
@@ -208,6 +208,10 @@ func (m *KopsModelContext) CloudTagsForInstanceGroup(ig *kops.InstanceGroup) (ma
 		}
 	}
 
+	// Add cluster and ig names
+	labels[awsup.TagClusterName] = m.ClusterName()
+	labels["Name"] = m.AutoscalingGroupName(ig)
+
 	// The system tags take priority because the cluster likely breaks without them...
 
 	if ig.Spec.Role == kops.InstanceGroupRoleMaster {
--- a/pkg/model/spotinstmodel/BUILD.bazel
+++ b/pkg/model/spotinstmodel/BUILD.bazel
@@ -13,7 +13,6 @@ go_library(
         "//pkg/model/defaults:go_default_library",
         "//upup/pkg/fi:go_default_library",
         "//upup/pkg/fi/cloudup/awstasks:go_default_library",
-        "//upup/pkg/fi/cloudup/awsup:go_default_library",
         "//upup/pkg/fi/cloudup/spotinsttasks:go_default_library",
         "//vendor/k8s.io/klog:go_default_library",
     ],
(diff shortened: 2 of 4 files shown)


johngmyers:
This would be easier to review if the separate concerns were in separate commits.

johngmyers:
Once we get Terraform and Cloudformation support, I was thinking we'd change the following setting to be false when using launch templates:

https://github.com/kubernetes/kops/blob/051ceeea707982bcb609b035065bd5bfe4e512ae/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go#L551

I don't have any changes planned in this area other than seeing if we can change the default value of the `EnableLaunchTemplates` feature flag.

hakman:
Separated the CF and TF support for a future PR. This one is pretty easy to review now.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -101,6 +101,9 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 	if err != nil {
 		return nil, fmt.Errorf("error building cloud tags: %v", err)
 	}
+	for k, v := range b.CloudTags(b.AutoscalingGroupName(ig), false) {
```
johngmyers: So why don't we put these on the ASG? Should we put this call inside `CloudTagsForInstanceGroup` so they also get applied when using launch configurations?

hakman: Sounds good to me. Will take a look to see where and how this is used.

hakman:
@rifelpet I think this is finally ready to go. Thanks!

rifelpet:
hm is `CloudTagsForInstanceGroup` not used when we target terraform? I would expect some of the terraform integration test files to require updating.

hakman:
The initial commit contained terraform and cloudformation files also, but I was asked to split them to make it easier to review.
https://github.com/kubernetes/kops/commit/f434b48f5dca4ed4ca26fbe4fbea1972fa10a77b

rifelpet:
right, but I would think that `CloudTagsForInstanceGroup` is used by both "direct" and "terraform" targets. What additional change to the terraform target needs to be made?

Ah, I see that the [terraform target doesnt mention tags at all](https://github.com/kubernetes/kops/blob/e8c107f767d00e22f6141c97abbfdb74cb3fba1c/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go) so we'll need to plumb that through before the terraform integration tests require updating. 👍  makes sense now.

/lgtm
/approve

hakman:
My general impression with this is that the `direct` target needs a lot of "love". There are warning during cluster setup (resource order), there are warnings on delete, launch templates are not deleted on cluster delete, ...

Terraform and CF targets also need a bit of tweaking, as they don't seem to work out of the box with `EnableLaunchTemplates=true`.

PR #8467, opened 2020-02-04
Stop logging to /var/log/kops-controller.log

justinsb:
Writing to a hostPath from a non-root container requires file
ownership changes, which is difficult to roll out today.  See
discussion in #8454

We were primarily using the logfile for e2e diagnostics, so we're
going to look into collecting the information via other means instead.

We also haven't yet shipped this logfile in a released version (though
we have shipped it in beta releases)


rifelpet:
I confirmed this reverted everything in https://github.com/kubernetes/kops/pull/8204 👍

PR #8468, opened 2020-02-04
Fix Github download url for nodeup

adri:
Currently this generates:
```
NODEUP_URL=https://artifacts.k8s.io/binaries/kops/1.15.1/linux/amd64/nodeup,https://github.com/kubernetes/kops/releases/download/1.15.1/linux-amd64-nodeup,https://kubeupv2.s3.amazonaws.com/kops/1.15.1/linux/amd64/nodeup
NODEUP_HASH=de4939eadb6e4d89fcf608b1f632e770bcce521d6dc5c45d76d2c4608ad23db4
```

However for the Github URL a `v` is missing in front of the version tag.

Returns a 404:
```
curl https://github.com/kubernetes/kops/releases/download/1.15.1/linux-amd64-nodeup
```

Downloads the file:
```
curl https://github.com/kubernetes/kops/releases/download/v1.15.1/linux-amd64-nodeup
```

### DIFF
--- a/tests/integration/update_cluster/additional_cidr/cloudformation.json.extracted.yaml
+++ b/tests/integration/update_cluster/additional_cidr/cloudformation.json.extracted.yaml
@@ -18,7 +18,7 @@ Resources.AWSAutoScalingLaunchConfigurationmasterustest1bmastersadditionalcidrex
   set -o nounset
   set -o pipefail
 
-  NODEUP_URL=https://artifacts.k8s.io/binaries/kops/1.15.0/linux/amd64/nodeup,https://github.com/kubernetes/kops/releases/download/1.15.0/linux-amd64-nodeup,https://kubeupv2.s3.amazonaws.com/kops/1.15.0/linux/amd64/nodeup
+  NODEUP_URL=https://artifacts.k8s.io/binaries/kops/1.15.0/linux/amd64/nodeup,https://github.com/kubernetes/kops/releases/download/v1.15.0/linux-amd64-nodeup,https://kubeupv2.s3.amazonaws.com/kops/1.15.0/linux/amd64/nodeup
   NODEUP_HASH=9604ef18267ad7b5cf4cebbf7ab64423cf5bb0342d169c608ac6376e6af26d81
 
   export AWS_REGION=us-test-1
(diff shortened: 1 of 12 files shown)


joshbranham:
Yeah the name of releases changed to prefix a v. Does this matter for older releases @rifelpet ?

rifelpet:
yes, we updated the tag in 1.15.1 to be prefixed with `v`. I'll need to investigate why this happened because we should have only updated the git tag, not the version string that is built into kops.

justinsb:
I think the underlying issue is correct - we are moving to `v` tags, which means that the github download urls now have a `v` in them.  But as you point out @rifelpet the version in kops does not have a v, hence we build the old URL.

(I assume I should be uploading releases against the `v` tags as I think eventually we want to stop tagging without the `v`)

justinsb:
I was wondering if we needed to have logic only to include the `v` if kops version <= 1.15, but actually we can achieve that with this PR and simply cherry-picking back only to the correct branches.  So this approach LGTM.

@adri if you run `hack/update-expected.sh` kops should fix the expected values for you and the tests should pass (I think).  And thanks for the PR!

rifelpet:
An alternative but potentially more invasive solution is that we include `v` in the variables themselves:

https://github.com/kubernetes/kops/blob/6c202ef3751b53b8f6200c5261f9c192a2f346e6/version.go#L24-L28

for reference, both `kubectl version`, `kubelet --version` include `v` in their output, perhaps we should follow suit?

adri:
From a user perspective: In my logs I saw issues fetching the first 2 of the 3 requests.
```
NODEUP_URL=https://artifacts.k8s.io/binaries/kops/1.15.1/linux/amd64/nodeup,https://github.com/kubernetes/kops/releases/download/1.15.1/linux-amd64-nodeup,https://kubeupv2.s3.amazonaws.com/kops/1.15.1/linux/amd64/nodeup
```

1. `curl https://artifacts.k8s.io/binaries/kops/1.15.1/linux/amd64/nodeup` fails (not sure what the right URL is
2. `curl https://github.com/kubernetes/kops/releases/download/1.15.1/linux-amd64-nodeup` fails because of the missing `v`
3. The third URL succeeds 🎉 

Whatever the version strategy will be, I think the URLs for 1. and 2. should be somehow aligned.

justinsb:
@rifelpet I don't mind changing the version number to include the `v`, except that it is much more invasive and we should probably get the existing releases out.

I do need to public these files to artifacts.k8s.io; that's more of a WIP, but having both artifacts.k8s.io and github.com not available will put more load on the 3rd bucket which I'd prefer not to use as our sole source.

I think I've just added the result of running `hack/updated-expected.sh`, so hopefully this will now pass.

justinsb:
OK this is passing tests after running the script, so proceeding so we can get the releases going.  Thanks for spotting and fixing @adri  

/approve
/lgtm

rifelpet:
I agree this looks good 👍

PR #8472, opened 2020-02-04
Prepare for future Cilium 1.7 upgrade

johngmyers:
Per https://docs.cilium.io/en/v1.7/install/upgrade/#required-changes Cilium 1.7 will remove the `kubernetes.io/cluster-service` annotation from the `DaemonSet` pod template. Remove this from the `matchLabels` early in order to prepare for this change.

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
@@ -307,7 +307,6 @@ spec:
   selector:
     matchLabels:
       k8s-app: cilium
-      kubernetes.io/cluster-service: "true"
   template:
     metadata:
       annotations:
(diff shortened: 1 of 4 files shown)


olemarkus:
1.7 also requires some additional rbac stuff. Do you want to add that to this PR? Basically adding permissions to interact with the `ciliumclusterwidenetworkpolicy` resources.

johngmyers:
The purpose of this is not to support running 1.7 but to prepare the ground so that a later upgrade won't have to be a two-step manouver.

olemarkus:
Ah. I see what you try  to do here. This will unfortunately not work. running the diff will result in 
```
The DaemonSet "cilium" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"k8s-app":"cilium"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```
The reason the annotation was removed can be seen here: https://github.com/cilium/cilium/commit/5bc434dc70c8a8219c65c9c9ad92eb67dd61992a#diff-20f9300a3a5239a4a4c9ce00951c8d5d

I don't think this is applicable to kops, so the easiest thing is to just leave the annotation be. Alternatively we need to be able to run `--force`, similar to the StorageClass issue that was discussed last week.

johngmyers:
Argh, yet another unnecessarily immutable field. I could almost see preventing additions, but removals are safe. (Except, I suppose, one wouldn't be able to roll back)

A lot of this immutability causes more problems than it solves.

PR #8493, opened 2020-02-06
GCS: Don't try to set ACLs if bucket-policy only is set

justinsb:
This means we won't be able to work unless there's a bucket permission
(which actually will typically happen if the state store is in the
same GCS project).

This is the minimal workaround for cherry-picking.

### DIFF
--- a/pkg/acls/gce/BUILD.bazel
+++ b/pkg/acls/gce/BUILD.bazel
@@ -12,5 +12,6 @@ go_library(
         "//upup/pkg/fi/cloudup/gce:go_default_library",
         "//util/pkg/vfs:go_default_library",
         "//vendor/google.golang.org/api/storage/v1:go_default_library",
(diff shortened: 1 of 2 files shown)


justinsb:
Some background: we set ACLs to enable a remote GCS bucket.  But if the bucket has [uniform bucket level access](https://cloud.google.com/storage/docs/uniform-bucket-level-access) then we simply can't set ACLs.

For the common case where the kops-state-store GCS bucket is in the same project, we likely don't need these ACLs anyway, so this is a minimal fix to unblock functionality there, that we can cherry-pick.

Future PRs will try to handle the cross-project case, but we probably wouldn't cherry-pick them.

rifelpet:
Looking forward to our [kops-gce](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-gce) job turning green :)

/lgtm

PR #8088, opened 2019-12-10
test validateCluster twice to make sure it does not flap

zetaab:
fixes #8078 

/kind bug

Issue #8078, opened 2019-12-10
rolling-update: cluster not validated correctly


zetaab:
Using 1.17.0-alpha.1 and kubernetes 1.16.2: trying to update to 1.16.3

```
kops rolling-update cluster --bastion-interval 2m --instance-group bastions,master-zone-1,nodes --validation-timeout 20m --yes

snip

I1210 10:05:51.299467 53 instancegroups.go:357] Waiting for 5s for pods to stabilize after draining.
I1210 10:05:56.299785 53 instancegroups.go:176] deleting node "master-zone-1-1-updatemajorpr-892a8a-k8s-local" from kubernetes
I1210 10:05:56.317191 53 instancegroups.go:302] Stopping instance "36a99cb5-a769-4f26-8b78-b8f91f5408ba", node "master-zone-1-1-updatemajorpr-892a8a-k8s-local", in group "updatemajorpr-892a8a.k8s.local-master-zone-1" (this may take a while).
W1210 10:05:56.317225 53 instance.go:105] This does not work without running kops update cluster --yes in another terminal
I1210 10:05:56.758342 53 instancegroups.go:189] waiting for 15s after terminating instance
I1210 10:06:11.758539 53 instancegroups.go:200] Validating the cluster.
I1210 10:07:02.275601 53 instancegroups.go:261] Cluster did not validate, will try again in "30s" until duration "20m0s" expires: error listing nodes: Get https://10.222.142.60/api/v1/nodes: unexpected EOF.
I1210 10:07:36.134273 53 instancegroups.go:268] Cluster did not pass validation, will try again in "30s" until duration "20m0s" expires: machine "f8d4ab37-f577-4e2a-8bfd-25cc7a4735f0" has not yet joined cluster.
I1210 10:08:06.597076 53 instancegroups.go:268] Cluster did not pass validation, will try again in "30s" until duration "20m0s" expires: machine "f8d4ab37-f577-4e2a-8bfd-25cc7a4735f0" has not yet joined cluster.
I1210 10:08:37.806011 53 instancegroups.go:268] Cluster did not pass validation, will try again in "30s" until duration "20m0s" expires: machine "f8d4ab37-f577-4e2a-8bfd-25cc7a4735f0" has not yet joined cluster.
[...]

PR #8515, opened 2020-02-08
Wait for validation to succeed N consecutive times

hakman:
Validation of the cluster during e2e tests is not that reliable. Nodes are expected to be in `Ready` state and pass validation once:
https://github.com/kubernetes/test-infra/blob/8eefa866327706ca2ed7048e5a53437917d92f0d/kubetest/kops.go#L444-L452

A more reliable approach would be for Kops validation to succeed N consecutive times during a time interval. This PR improves cluster validation by adding a `--count` flag that can be used together with `--wait` to achieve this.

### DIFF
--- a/cmd/kops/validate_cluster.go
+++ b/cmd/kops/validate_cluster.go
@@ -42,6 +42,7 @@ import (
 type ValidateClusterOptions struct {
 	output     string
 	wait       time.Duration
+	count      int
 	kubeconfig string
 }
 
@@ -61,7 +62,7 @@ func NewCmdValidateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 		Run: func(cmd *cobra.Command, args []string) {
 			result, err := RunValidateCluster(f, cmd, args, os.Stdout, options)
 			if err != nil {
-				exitWithError(err)
+				exitWithError(fmt.Errorf("Validation failed: %v", err))
 			}
 			// We want the validate command to exit non-zero if validation found a problem,
 			// even if we didn't really hit an error during validation.
@@ -73,6 +74,7 @@ func NewCmdValidateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVarP(&options.output, "output", "o", options.output, "Output format. One of json|yaml|table.")
 	cmd.Flags().DurationVar(&options.wait, "wait", options.wait, "If set, will wait for cluster to be ready")
+	cmd.Flags().IntVar(&options.count, "count", options.count, "If set, will validate the cluster consecutive times")
 	cmd.Flags().StringVar(&options.kubeconfig, "kubeconfig", "", "Path to the kubeconfig file")
 
 	return cmd
@@ -144,15 +146,22 @@ func RunValidateCluster(f *util.Factory, cmd *cobra.Command, args []string, out 
 		return nil, fmt.Errorf("unexpected error creating validatior: %v", err)
 	}
 
+	consecutive := 0
 	for {
+		if options.wait > 0 && time.Now().After(timeout) {
+			return nil, fmt.Errorf("wait time exceeded during validation")
+		}
+
 		result, err := validator.Validate()
 		if err != nil {
-			if time.Now().After(timeout) {
+			consecutive = 0
+			if options.wait > 0 {
+				klog.Warningf("(will retry): unexpected error during validation: %v", err)
+				time.Sleep(pollInterval)
+				continue
+			} else {
 				return nil, fmt.Errorf("unexpected error during validation: %v", err)
 			}
-			klog.Warningf("(will retry): unexpected error during validation: %v", err)
-			time.Sleep(pollInterval)
-			continue
 		}
 
 		switch options.output {
@@ -160,7 +169,6 @@ func RunValidateCluster(f *util.Factory, cmd *cobra.Command, args []string, out 
 			if err := validateClusterOutputTable(result, cluster, instanceGroups, out); err != nil {
 				return nil, err
 			}
-
 		case OutputYaml:
 			y, err := yaml.Marshal(result)
 			if err != nil {
@@ -169,7 +177,6 @@ func RunValidateCluster(f *util.Factory, cmd *cobra.Command, args []string, out 
 			if _, err := out.Write(y); err != nil {
 				return nil, fmt.Errorf("error writing to output: %v", err)
 			}
-
 		case OutputJSON:
 			j, err := json.Marshal(result)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


johngmyers:
I think we might want to replicate `RollingUpdateCluster`'s `ValidateSuccessDuration`. Only require two successful validations in a row and possibly allow adjusting the interval after the first successful validation.

From #8088 / #8078 it's `kube-controller-manager` that causes the flapping and a node tends to only flap once.

johngmyers:
...and we should default `kops validate cluster --wait` to requiring two successful validations.

hakman:
> I think we might want to replicate RollingUpdateCluster's ValidateSuccessDuration. Only require two successful validations in a row and possibly allow adjusting the interval after the first successful validation.
I don't think I want to replicate that, it is an internal feature. This would like to replace node validation on e2e tests side where things will not be predictable. This is why it requires 10 successes in a row. Maybe it can be decreased, but not to 2.

> ...and we should default `kops validate cluster --wait` to requiring two successful validations.
This would change the current behaviour of the feature. Not sure this is desired.

johngmyers:
Do we have any data on how the e2e flakiness is behaving?

I actually tend to doubt that the needs are all that different. If we're routinely getting flapping, the users of `kops validate cluster --wait` probably want it to correct for that.

For two validations to be insufficient, the cluster would have to flap at least twice and two of the flapping successes would have to line up with the checks. Our data so far show that nodes flap once upon creation. No matter what we do, we're going to have a trade-off of wall-clock time versus risk of returning success before a flap; I'd prefer we have a fair idea of what the behavior we're addressing is, rather than just throw a wildly guessed set of delays at the process.

johngmyers:
This code is waiting the whole `pollInterval` after a successful validation. In the rolling update code, we wait a much shorter amount of time after a successful validation, as flaps were much shorter.

hakman:
@johngmyers this is NOT about "known" flakiness, it is about making sure the env is stable before running e2e. You cannot predict how a bad change will affect flakiness. You are just assuming that I am expecting a regularly occurring flakiness for which we have a quick fix.

It is not a command that you will run (unless you want to), but is expected to run in e2e tests and replace this: https://github.com/kubernetes/test-infra/blob/8eefa866327706ca2ed7048e5a53437917d92f0d/kubetest/kops.go#L445.

Probably the time interval between successful checks can be reduced, but would like to get some more feedback on this before adding more changes.

johngmyers:
What is known as "flakiness" is a too high rate of intermittent false positives, when a change that is good incorrectly fails the tests. When evaluating fixes to these, one may reasonably take as an axiom that the change is good.

One does have to keep in mind the risk of increasing false negatives. Passing validation too early, however, is unlikely to cause the e2e tests to pass when they otherwise would have failed.

johngmyers:
My objections to this still stand. e2e tests should not be a special snowflake and the `kops validate cluster` command should not be so different from validation during rolling update.

hakman:
There are cases where clusters are unstable in unpredictable ways before starting e2e. I would rather wait a bit more and know in advance that the cluster is unstable, than investigate the root cause of failing tests. Here is an aexample of such an unstable cluster that passes 4 checks in a row:
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kops-aws-containerd/1227778845420032003/build-log.txt

Rolling update uses 30 seconds between unsuccessful checks and 10 seconds between successful checks.
Validate cluster uses 10 seconds between any check. It was already different before I added the consecutive validations counter.

Rolling update has to be fast. You cannot press a button and to proceed to the next node. 2 successful validations may be good enough there. For sure is better than 1. I don't want to debate statistics.
Validate cluster is a user run command when one wants to make sure the cluster is stable. Anyone can decide how many successful validation is row are good for him. Anyone can press Ctrl+C at any time and stop it.

johngmyers:
That log file does not have any output from `kops validate cluster`. What relevance does it have?

hakman:
> That log file does not have any output from `kops validate cluster`. What relevance does it have?

That is because `kops validate cluster` runs after the current simpler node validation sequence which this PR wants to improve.
Look for `success count = 4`. Should be there 2 times and might have been more if it didn't timeout.

Also you can check the comments in #kops-dev from yesterday:
> kops-controller is broken for me in master (including my typo fix PR). the cluster sporadically validates because all the kops-controllers are Running for some seconds before crashing. The error is a timeout talking to the internal API address, so it may be a cilium thing (although I cannot say why). Still, some health checking would have caught this earlier

johngmyers:
This PR is to change `kops validate cluster`. How would that improve the much more naïve and flappy sequence that e2e is using? The flapping of e2e's "validation" proves nothing about the flapping behavior of `kops validate cluster`.

If kops-controller is crashing, then improving the probability of e2e passing is undesirable.

hakman:
> How would that improve the much more naïve and flappy sequence that e2e is using?

This would replace the naive implementation e2e is using. Validating cluster N times before preceding should improve things.

> The flapping of e2e's "validation" proves nothing about the flapping behavior of kops validate cluster.

It proves that cluster can fail in unexpected ways. If you want to to only validate a cluster 2 consecutive times, it is your choice. Anyone can choose how many times to do it in its own setup. I don't see any reason to have very strong opinions here. The default should stay 0 at least until this gets tested a bit more and maybe tweak the intervals.

> If kops-controller is crashing, then improving the probability of e2e passing is undesirable.

The intention of this PR it is exactly the opposite. Validation should fail earlier.
Really don't get how it would improve the probability of e2e passing.

johngmyers:
> This would replace the naive implementation e2e is using. Validating cluster N times before preceding should improve things.

I suggest replacing the implementation e2e is using first, first trying it out in a new testgrid job. Get the data on how `kops validate cluster` is behaving, _then_ suggest improvements to it.

> It proves that cluster can fail in unexpected ways.

I don't see how that leads to the conclusion that `kops validate cluster` should pass N times before allowing e2e to start. And since `kops validate cluster` will report failure in many situations where current e2e will report success, it doesn't necessarily prove that it can fail in ways that `kops validate cluster` doesn't expect.

> The intention of this PR it is exactly the opposite. Validation should fail earlier.
Really don't get how it would improve the probability of e2e passing.

By waiting for N consecutive checks to pass before proceeding, it prevents the failures during those N times from causing the test to fail.

If, on the other hand, a failure of one of those subsequent cluster validations caused e2e to fail, that would have likely caught the change that caused kops-controller to crash.

hakman:
As discussed during office hours, validation will fail if any of the consecutive validations fails.

hakman:
@johngmyers can you check if it looks ok now?

hakman:
Fixes #8743

hakman:
Thanks @zetaab :)

PR #8516, opened 2020-02-09
[vendoring] Update google libs to latest

geojaz:
Our GCE libs are sad and old. Google recently released v0.17.0. 
This is part 1 of 2 where I will attempt to swap `compute/v0.beta` for `compute/v1` (GCE)


geojaz:
seems like a flake? since these changes should only affect gce

geojaz:
We're definitely flaking since this is a totally different failure than last run. I have added a PR to bump our `ginkgo.flakeAttempts` to 2 (from 1) on the kops e2e tests here https://github.com/kubernetes/test-infra/pull/16201 that I _hope_ will help this sort of thing...

PR #8517, opened 2020-02-09
Migrates cloudup/gce from v0.beta => v1

geojaz:
This relates to #8516 and is a first pass at modernizing our GCE integration. 

~#8516 needs to go first to get this green...~
We got the vendoring updated so when the tests are green, we can go on this one

### DIFF
--- a/upup/pkg/fi/cloudup/gce/wrappers.go
+++ b/upup/pkg/fi/cloudup/gce/wrappers.go
@@ -19,8 +19,7 @@ package gce
 import (
 	"fmt"
 
-	context "golang.org/x/net/context"
-	compute "google.golang.org/api/compute/v0.beta"
+	compute "google.golang.org/api/compute/v1"
 	"k8s.io/klog"
 )
 
@@ -86,7 +85,6 @@ func DeleteInstance(c GCECloud, instanceSelfLink string) error {
 
 // ListManagedInstances lists the specified InstanceGroupManagers in GCE
 func ListManagedInstances(c GCECloud, igm *compute.InstanceGroupManager) ([]*compute.ManagedInstance, error) {
-	ctx := context.Background()
 	project := c.Project()
 
 	zoneName := LastComponent(igm.Zone)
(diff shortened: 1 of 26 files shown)


upup/pkg/fi/cloudup/gce/wrappers.go
```diff
@@ -98,15 +96,10 @@ func ListManagedInstances(c GCECloud, igm *compute.InstanceGroupManager) ([]*com
...
 	//	)
 
-	var instances []*compute.ManagedInstance
-	err := c.Compute().InstanceGroupManagers.ListManagedInstances(project, zoneName, igm.Name).Pages(ctx,
-		func(page *compute.InstanceGroupManagersListManagedInstancesResponse) error {
-			instances = append(instances, page.ManagedInstances...)
-			return nil
-		})
+	req := c.Compute().InstanceGroupManagers.ListManagedInstances(project, zoneName, igm.Name)
+	resp, err := req.Do()
```
rifelpet: As someone unfamiliar with the GCE api, I see that theres a [PageToken](https://pkg.go.dev/google.golang.org/api/compute/v1?tab=doc#InstanceGroupManagersListManagedInstancesCall.PageToken) for ListManagedInstances but I'm having trouble finding example code that uses it so I can't tell if we should continue to use pagination or not. Do you think that pagination is no longer necessary?

geojaz: It is unlikely in most cases that it would be needed, but you're right fixing...

upup/pkg/fi/cloudup/gcetasks/instancegroupmanager.go
```diff
@@ -151,21 +151,15 @@ func (_ *InstanceGroupManager) RenderGCE(t *gce.GCEAPITarget, a, e, changes *Ins
 		}
 
 		if changes.TargetSize != nil {
-			request := &compute.InstanceGroupManagersResizeAdvancedRequest{
-				TargetSize: i.TargetSize,
-			}
-			if i.TargetSize == 0 {
-				request.ForceSendFields = append(request.ForceSendFields, "TargetSize")
-			}
-			op, err := t.Cloud.Compute().InstanceGroupManagers.ResizeAdvanced(t.Cloud.Project(), *e.Zone, i.Name, request).Do()
```
justinsb: Interesting!  The ForceSendFields was originally here because otherwise we didn't send Size == 0, but it looks like this has been fixed by using the Resize command.

upup/pkg/fi/cloudup/gce/wrappers.go
justinsb: Well this is odd - it looks like the Pages method isn't defined there?  I can try to dig around to find out why but (1) do we agree there is no Pages method and (2) did we figure out a reason why it might not be needed?

justinsb:
Ah - I'm plainly following in your footsteps @geojaz ... looks like https://github.com/googleapis/google-api-go-client/issues/451 is where you asked in the library about the missing method, and so we're likely (temporarily) blocked on this question.  I say temporarily because we could always code our own version of the method.

geojaz:
that's exactly how I got there :) I've been busy and this isn't blocking me right now....

geojaz:
/close

Now that we have the support in the client lib for that MIG listing, I'll recreate...

PR #8128, opened 2019-12-16
Add inf1 isntances

mikesplain:
Add inf1 instance type.
Currently m6g instances aren't documented, so leaving those as is.
A few other instance types were messed up so those were also regenerated.

PR #8522, opened 2020-02-11
Fix Machine types vcpus

mikesplain:
I'm not sure why this changed in https://github.com/kubernetes/kops/pull/8128 but I'm going to look into modernizing this with the new api soon.


rifelpet:
definitely looking forward to switching to ec2.DescribeInstanceTypes. We could get rid of this file entirely right? And wouldn't need to do an sdk update and kops release in order to recognize new instance types?

/lgtm

mikesplain:
Agreed, yes I believe so, if we want to trust the APIs real time.  Previously we used cluster-autoscaler as an example. I'll see if they've moved over!

PR #8497, opened 2020-02-06
Mark dns-controller and kops-controller as non-root

johngmyers:
for the benefit of admission controllers that restrict this sort of thing.

PR #8523, opened 2020-02-11
Automated cherry pick of #8497: Mark dns-controller and kops-controller as non-root

johngmyers:
Cherry pick of #8497 on release-1.17.

#8497: Mark dns-controller and kops-controller as non-root

### DIFF
--- a/upup/models/cloudup/resources/addons/dns-controller.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/dns-controller.addons.k8s.io/k8s-1.12.yaml.template
@@ -56,6 +56,8 @@ spec:
           requests:
             cpu: 50m
             memory: 50Mi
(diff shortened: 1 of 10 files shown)


rifelpet:
I don’t think the PR that changed the kops-controller user was cherrypicked to 1.17 

https://github.com/kubernetes/kops/blob/release-1.17/cmd/kops-controller/BUILD.bazel

johngmyers:
Ah, so many patches.
Do we want to cherrypick the user changes to 1.17? In any case we won't be able to cherrypick this until there is an intervening release.

rifelpet:
I have no opinion on whether we cherry-pick it. Luckily the hostPath logging work was never cherry-picked back to 1.17 or 1.16 so we don't need to worry about cherry-pick the revert if we do choose to cherry-pick the user change PR and this.

PR #8525, opened 2020-02-11
Update support for docker tgz package

hakman:
Using the Docker tgz (static) package instead of rpm/deb has some advantages:
* works on more distros, like Amazon Linux
* single package for all supported distros, so much easier to keep updated
* works with any released containerd tgz
* (in the future) will allow to specify any package URL for containerd or docker to be installed

This is similar to https://github.com/kubernetes/kops/pull/8199.

### DIFF
--- a/pkg/model/components/docker.go
+++ b/pkg/model/components/docker.go
@@ -48,7 +48,9 @@ func (b *DockerOptionsBuilder) BuildOptions(o interface{}) error {
 
 	// Set the Docker version for known Kubernetes versions
 	if fi.StringValue(clusterSpec.Docker.Version) == "" {
-		if b.IsKubernetesGTE("1.17") {
+		if b.IsKubernetesGTE("1.18") {
+			docker.Version = fi.String("19.03.7")
+		} else if b.IsKubernetesGTE("1.17") {
 			docker.Version = fi.String("19.03.4")
 		} else if b.IsKubernetesGTE("1.16") {
 			docker.Version = fi.String("18.09.9")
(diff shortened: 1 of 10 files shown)


hakman:
There is an issue with containerd builds that broke support for older distros in latest 1.2.x releases. https://github.com/containerd/containerd/issues/4008
/hold

hakman:
A new version of containerd was released and fixes the issue of running on older distros. This is ready for review.
/unhold

johngmyers:
It looks like the use of `Version` and `PackageVersion` is backwards. `PackageVersion` is what is being compared to the spec and `Version` is version of the package being installed.

Perhaps `TestDockerPackageNames` and `TestContainerdPackageNames` could assert that when `PlainBinary` is true that `Version` and `ExtraPackages` are empty?

/lgtm

pkg/model/components/docker.go
```diff
@@ -48,7 +48,9 @@ func (b *DockerOptionsBuilder) BuildOptions(o interface{}) error {
 
 	// Set the Docker version for known Kubernetes versions
 	if fi.StringValue(clusterSpec.Docker.Version) == "" {
-		if b.IsKubernetesGTE("1.17") {
+		if b.IsKubernetesGTE("1.18") {
+			docker.Version = fi.String("19.03.7")
```
justinsb: I couldn't find the note about this being the correct version.  Is this our decision or an upstream recommendation?

justinsb:
Thanks @hakman ... I'm not sure the extent to which we have a recommendation on versions before the 1.18 release (and it looks like there is now a 19.03.8), but this lgtm!

pkg/model/components/docker.go
hakman: The upstream only recommends Docker "19.03" in the release notes for 1.18. This is also why I proposed to use 19.03.8 for 1.17 in #8738.

hakman:
Thanks @justinsb. I opened https://github.com/kubernetes/kops/pull/8748 to add support for 19.03.8 in Kops and https://github.com/kubernetes/website/pull/19643 to update the website docs.
Other than that, the release notes will only mention 19.03, same as for 1.17.

PR #8538, opened 2020-02-13
Update amazon-vpc-cni-k8s to v1.6.0

hakman:
https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.6.0

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -2408,8 +2408,7 @@ spec:
                     VPC CNI networking
                   properties:
                     imageName:
-                      description: 'The container image name to use, which by default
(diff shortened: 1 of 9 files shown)


rifelpet:
we can watch for any issues in our [amazon-vpc e2e job](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-cni-amazon-vpc)

site note, I opened https://github.com/aws/amazon-vpc-cni-k8s/issues/846 to try to get rid of the wildcards in the ClusterRole but I don't think this needs to wait for that, since it the ENIConfig resources that it manages have much less impact on a cluster than say [the aws-iam-authenticator CRDs that we discussed earlier](https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/287).

/lgtm
/approve

PR #8549, opened 2020-02-13
Spotinst: Add support for blacklisting or whitelisting instance types in Ocean

liranp:
This PR adds support for blacklisting or whitelisting instance types in Ocean.

### DIFF
--- a/pkg/model/spotinstmodel/instance_group.go
+++ b/pkg/model/spotinstmodel/instance_group.go
@@ -61,6 +61,12 @@ const (
 	// Launch Spec for the Ocean cluster.
 	InstanceGroupLabelOceanDefaultLaunchSpec = "spotinst.io/ocean-default-launchspec"
 
+	// InstanceGroupLabelOceanInstanceTypes[White|Black]list are the metadata labels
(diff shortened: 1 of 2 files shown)


rifelpet:
/lgtm

@liranp on a side note, it would be great if we could setup some integration tests that use SpotInst and Terraform. You can see examples [here](https://github.com/kubernetes/kops/tree/master/tests/integration/update_cluster) and they get called [here](https://github.com/kubernetes/kops/blob/master/cmd/kops/integration_test.go). I tried creating one myself a while back but ran into issues because we dont have a "MockCloud" for SpotInst the same way that [we do for AWS](https://github.com/kubernetes/kops/blob/298f79659a2bd9c1d6548c6c95b335b5e29e6466/upup/pkg/fi/cloudup/awsup/mock_aws_cloud.go). I'm going to be making some changes to the terraform code soon and having integration tests that cover the other cloud providers would help immensely. I can open a GH issue if that helps.

liranp:
@rifelpet Sure, I‘ll take care of that as soon as possible. Thank you!

PR #8550, opened 2020-02-13
update coredns for security best practice

Ewocker:
- use non-root user
- remove container privilege
- change core port so that no need cap NET_BIND_SERVICE
- update service kube-dns to use targetPort name to be kube-dns compatible on migration

Note:
During migration when kops is patching service kube-dns, duplicate port number might cause issue that only the first port has the correct targetPort name. In this case kubectl replace needs to be use to resolve the problem. See issue https://github.com/kubernetes/kubernetes/issues/47249

### DIFF
--- a/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
@@ -182,10 +182,10 @@ spec:
           mountPath: /etc/coredns
           readOnly: true
         ports:
-        - containerPort: 53
+        - containerPort: 5353
           name: dns
           protocol: UDP
-        - containerPort: 53
+        - containerPort: 5353
           name: dns-tcp
           protocol: TCP
         - containerPort: 9153
@@ -194,10 +194,9 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:
-            add:
-            - NET_BIND_SERVICE
             drop:
             - all
+          runAsUser: 1000
           readOnlyRootFilesystem: true
         livenessProbe:
           httpGet:
@@ -246,12 +245,15 @@ spec:
   - name: dns
     port: 53
     protocol: UDP
+    targetPort: dns
   - name: dns-tcp
     port: 53
     protocol: TCP
+    targetPort: dns-tcp
   - name: metrics
     port: 9153
     protocol: TCP
+    targetPort: metrics
 
 ---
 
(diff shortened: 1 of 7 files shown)


johngmyers:
I'm not quite sure what to do with this apparent bug in the `core/v1` `ServiceSpec` API. It's not like we can neuter the `kube-dns` service and put in a `kube-dns-kops` service instead.

We're execing `kubectl apply -f`, so we don't seem to have much opportunity to recognize the case and do an update instead of a strategic patch.

johngmyers:
In researching for the kops office hours, I found that "server side apply" is the Kubernetes-level fix for the broken strategic patch. That is a beta feature right now.

The decision from office hours is that server side apply is too new and risky to take on now. The approach would be to get channels to use delete-and-replace instead of apply.

The question then becomes how to decide to do delete-and-replace. Some options I see:

- Pull the "kubectl apply -f" logic into channels so that we can add code to detect when the patch doesn't apply correctly and then choose an alternate update flow.
- Add metadata to the addon to tell channels to use the alternate flow across a particular version boundary.

We should consider whether an option can also address the immutable field problem.

rifelpet:
While this isn't directly related, I think one other issue that could benefit from adjusting how Kops applies channels is the removal of resources. If a manifest no longer defines a resource (either during an upgrade or downgrade) `kubectl apply -f` will not delete it which leaves it lingering and unmanaged in the cluster. `kubectl apply` has a prune field but I'm not sure if that is sufficient. If the new channel application strategy could handle that as well, that'd be great.

Ewocker:
> While this isn't directly related, I think one other issue that could benefit from adjusting how Kops applies channels is the removal of resources. If a manifest no longer defines a resource (either during an upgrade or downgrade) `kubectl apply -f` will not delete it which leaves it lingering and unmanaged in the cluster. `kubectl apply` has a prune field but I'm not sure if that is sufficient. If the new channel application strategy could handle that as well, that'd be great.

In this case while transitioning from kubedns to coredns, would kubedns be removed? Is that what we want, or we want the user to manually remove one of them?

rifelpet:
The current migration instructions require the user manually remove the KubeDNS deployment. It would be great if Kops could do that automatically, but in the case of KubeDNS/CoreDNS I think knowing when it is safe to remove might add significant complexity. All of the CoreDNS pods need to be Ready, the replica count would need to match, etc. I think that might be too much to strive for with this migration.

Ewocker:
@johngmyers

I am not sure what other options do we have for alternate flow other then replace (delete then create), but I can see some problems of doing that: 
- In the case of addon version change this would make sense, but from an addon to another would be a special case, ex. kubedns to coredns.
- If using `kubectl replace` but if an addon add a new resource which was not previously in the manifest, then replace will fail. Similar things will happen for first delete then create. If a resource is removed from new version, deletion will miss the removed resources (which is already a problem in kops).
- Workload can have down time because we need to first delete which might not be acceptable for some addon, ex. dns. (This might be the same in the other approach though since we are doing a single apply using a multi-resource file)

In the original comment where I was able to do this without service down time is to only replace the kube-dns service instead of all. But I am not sure what is the best way to implement this today in kops.

PR #8561, opened 2020-02-14
Replace unmaintained go-bindata

justinsb:
k/k uses go-bindata/go-bindata, so that's what we will use as well.


justinsb:
And BTW, the reason why I'm looking into this (other than that it's overdue) is that I think it's the cause of some non-determinism in our build.  I don't immediately see a fix in the diff, but first step is to get onto a maintained library.

PR #8563, opened 2020-02-14
use v0.29 in prometheus operator

coryschwartz:
there are problems without this already described in v0.29.yaml. Is there any reason v0.29 is not included in addons.yaml here?


coryschwartz:
/assign @geojaz 
Thanks for helping out with this.

I think this has context https://github.com/coreos/prometheus-operator/pull/2415/

I noticed this is already included in the prometheus operator as v0.29, but not in the addon.yaml.

johngmyers:
@geojaz I think you might want to `/approve`

PR #8568, opened 2020-02-15
makefile: use more hash helpers when building images

justinsb:
It's cleaner & more consistent.


rifelpet:
merge conflict but otherwise lgtm

PR #8570, opened 2020-02-15
Update makefile to reduce dependency on GOPATH

rifelpet:
This is one step towards removing the requirement that kops lives in the GOPATH.

* Consolidate the redundant `KOPS_ROOT` and `MAKEDIR` variables
* Remove dependency of `GOPATH_1ST` variable except for `GOPATH_1ST/bin`
* Remove broken `check-builds-in-go111` target
* Remove unused `KOPS_SERVER_TAG`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 
+# kops source root directory (without trailing /)
+KOPS_ROOT?=$(patsubst %/,%,$(abspath $(dir $(lastword $(MAKEFILE_LIST)))))
 DOCKER_REGISTRY?=gcr.io/must-override
 S3_BUCKET?=s3://must-override/
 UPLOAD_DEST?=$(S3_BUCKET)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 
+# kops source root directory (without trailing /)
+KOPS_ROOT?= $(patsubst %/,%,$(abspath $(dir $(firstword $(MAKEFILE_LIST)))))
```
johngmyers: Shouldn't this use `lastword`?

rifelpet: I copied this from L72 and had no preference about whether to use the KOPS_ROOT or MAKEDIR variable definition. Looking at the [make docs](https://ftp.gnu.org/old-gnu/Manuals/make-3.80/html_node/make_17.html) I think firstword would always use this Makefile if we were to ever import additional Makefiles, but since we don't they should be equivalent.

Unless you're thinking of the case in which third parties import the kops Makefile? I think in that case `lastword` might be more appropriate.

johngmyers: The sentence:
> This means that if the first thing a makefile does is examine the last word in this variable, it will be the name of the current makefile

suggests it should be `lastword`.

If there are Makefiles read through the MAKEFILES variable, the command line, or the default files, the first word wouldn't be this file.

rifelpet: updated 👍

PR #8571, opened 2020-02-16
Use cluster name as default subnet tag for Lyft CNI

hakman:
The Lyft CNI uses tags to determine which subnets to use when allocating new network adapters.
Current default is not matching anything, so the cluster is broken on startup. Using the cluster name for to determine the cluster subnets seems a better solution.

The the `conflist.template` was misaligned so I reformatted it.

### DIFF
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -545,7 +545,7 @@ You can specify which subnets to use for allocating Pod IPs by specifying
   networking:
     lyftvpc:
       subnetTags:
-        kubernetes_kubelet: true
+        KubernetesCluster: myclustername.mydns.io
 ```
 
 In this example, new interfaces will be attached to subnets tagged with `kubernetes_kubelet = true`.
--- a/docs/releases/1.18-NOTES.md
+++ b/docs/releases/1.18-NOTES.md
@@ -6,6 +6,8 @@
 
 * Terraform users on AWS may need to rename some resources in their state file in order to prepare for Terraform 0.12 support. See Required Actions below.
 
+* Lyft CNI plugin default subnet tags changed from from `Type: pod` to `KubernetesCluster: myclustername.mydns.io`. Subnets intended for use by the plugin will need to be tagged with this new tag and [additional tag filters](https://github.com/lyft/cni-ipvlan-vpc-k8s#other-configuration-flags) may need to be added to the cluster spec in order to achieve the desired set of subnets.
+
 * Support for Kubernetes versions prior to 1.9 has been removed.
 
 * Kubernetes 1.9 users will need to enable the PodPriority feature gate. See Required Actions below.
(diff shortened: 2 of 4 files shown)


maruina:
Looks good

rifelpet:
Hm i think technically this would be considered a breaking change. If someone's subnets were only tagged with `type: pod` and then upgraded kops, their cluster would use IPs in different subnets after a `rolling-update`. I'm wondering how we could best gate this change or otherwise have users opt in to it. Realistically I don't know how if people would really depend on `type: pod` as a tag without also specifying additional tags, but we should at the very least call it out in release notes.

Thoughts?

hakman:
@rifelpet I would say Release Notes / Breaking Changes mention should be a good idea. Kops tries to make these network plugins work out of the box as much as possible, to make it easier to get started. I also don't think the `type: pod` is something that anyone would like to use in prod, I will additionally gate it to 1.18, just to be on the safe side.

docs/releases/1.18-NOTES.md
```diff
@@ -6,6 +6,8 @@
 
 * Terraform users on AWS may need to rename some resources in their state file in order to prepare for Terraform 0.12 support. See Required Actions below.
 
+* Lyft CNI plugin default subnet tags changed from from `Type: pod` to `KubernetesCluster: myclustername.mydns.io`.
```
rifelpet: Maybe we should describe the action a user would want to take if they're affected by this.

`Subnets intended for pod IP use will need to be tagged with this new tag, and additional tag filters may need to be added to the cluster spec in order to achieve the desired set of subnets.`

or something along those lines.

hakman: I think should be ok now.

docs/networking.md
```diff
@@ -545,7 +545,7 @@ You can specify which subnets to use for allocating Pod IPs by specifying
   networking:
     lyftvpc:
       subnetTags:
-        kubernetes_kubelet: true
+        KubernetesCluster: myclustername.mydns.io
```
rifelpet: Sorry one more thing. the KubernetesCluster tag is now specified by default and can't be removed. So maybe the docs should make that clear?

> Subnets to be used for Pod IP allocation need to be tagged with `KubernetesCluster: myclustername.mydns.io`. Additional tags can be specified to further filter which subnets are used:
> 
> ```
>  networking:
>    lyftvpc:
>      subnetTags:
>        PodIPs: true
> ```

hakman: I think it can be removed and add any other tag that makes sense. This is more of a power user CNI plugin. Anyone messing with the tags will have to understand the consequences.

rifelpet: ah you're right, the `tags` variable is reassigned if the `subnetTags` field is specified, so it is possible to override the default KubernetesCluster tag. 👍

rifelpet:
One step closer to [e2e for lyft cni](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-cni-lyft)!

/lgtm
/approve

hakman:
> One step closer to [e2e for lyft cni](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-cni-lyft)!

🤞

PR #8575, opened 2020-02-16
Use UTC when calculating copyright year

johngmyers:
Avoids discrepancies between developer's laptop and CI verification in the evening of Dec 31.


justinsb:
We should have a kops NYE hackfest :-)

/approve
/lgtm

PR #8577, opened 2020-02-16
Capture logs from a kops cluster

justinsb:
This is derived from the dumping code in kubetest.  If we want to run
tests outside of kubetest (e.g. upgrade tests), we're going to need
that functionality.


johngmyers:
Need to `make gen-cli-docs`

pkg/dump/dumper.go
```diff
@@ -0,0 +1,430 @@
...
+		errors = append(errors, err)
+	}
+
+	// Capture full journal - needed so we can see e.g. disk mounts
+	// This does duplicate the other files, but ensures we have all output
+	if err := n.shellToFile(ctx, "sudo journalctl --output=short-precise", filepath.Join(n.dir, "journal.log")); err != nil {
+		errors = append(errors, err)
+	}
+
+	// Capture logs from any systemd services in our list, that are registered
```
johngmyers:
```suggestion
	// Capture logs from any systemd services in our list that are registered
```

cmd/kops/toolbox_dump.go
```diff
@@ -113,14 +127,60 @@ func RunToolboxDump(f *util.Factory, out io.Writer, options *ToolboxDumpOptions)
...
+			User: sshUser,
+			Auth: []ssh.AuthMethod{
+				ssh.PublicKeys(signer),
+			},
+			HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		}
+
+		dumper := dump.NewLogDumper(sshConfig, options.Dir)
+
+		// TODO: Should we try to get nodes?
```
johngmyers: I think you should either try to get nodes or remove the `nodes` parameter from `DumpAllNodes`

pkg/dump/dumper.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+
+	notDumped := findInstancesNotDumped(additionalIPs, dumped)
+	for _, ip := range notDumped {
+		if ctx.Err() != nil {
+			log.Printf("stopping dumping nodes: %v", ctx.Err())
+			return ctx.Err()
+		}
+
+		log.Printf("dumping node not registered in kubernetes: %s", ip)
```
johngmyers: Shouldn't log this unless there's code to get the node list.

cmd/kops/toolbox_dump.go
```diff
@@ -113,14 +127,60 @@ func RunToolboxDump(f *util.Factory, out io.Writer, options *ToolboxDumpOptions)
...
+			return fmt.Errorf("error reading private key %q: %v", privateKeyPath, err)
+		}
+
+		signer, err := ssh.ParsePrivateKey(key)
+		if err != nil {
+			return fmt.Errorf("error parsing private key %q: %v", privateKeyPath, err)
+		}
+
+		// TODO: We need to find the correct SSH user, ideally per IP
+		sshUser := "admin"
```
johngmyers: Expose as flag?

cmd/kops/toolbox_dump.go
```diff
@@ -113,14 +127,60 @@ func RunToolboxDump(f *util.Factory, out io.Writer, options *ToolboxDumpOptions)
...
+		sshConfig := &ssh.ClientConfig{
```
mitch000001, line 153: What about setups with a bastion host? Can we cover that?

cmd/kops/toolbox_dump.go
```diff
@@ -115,14 +128,60 @@ func RunToolboxDump(ctx context.Context, f *util.Factory, out io.Writer, options
...
+			return fmt.Errorf("error reading private key %q: %v", privateKeyPath, err)
+		}
+
+		signer, err := ssh.ParsePrivateKey(key)
+		if err != nil {
+			return fmt.Errorf("error parsing private key %q: %v", privateKeyPath, err)
+		}
+
+		// TODO: We need to find the correct SSH user, ideally per IP
+		sshUser := "admin"
```
hakman, line 153:
```suggestion
		sshUser := "ubuntu"
```

pkg/dump/dumper.go
```diff
@@ -0,0 +1,430 @@
...
+	if err != nil {
+		errors = append(errors, fmt.Errorf("error reading /var/log: %v", err))
+	}
+	for _, name := range n.dumper.files {
+		prefix := "/var/log/" + name + ".log"
+		for _, f := range fileList {
+			if !strings.HasPrefix(f, prefix) {
+				continue
+			}
+			if err := n.shellToFile(ctx, "sudo cat "+f, filepath.Join(n.dir, filepath.Base(f))); err != nil {
```
johngmyers, line 265: If `f` happens to have shell metacharacters then they will be interpreted as such. I believe `f` needs to be quoted for use in shell.

pkg/dump/dumper.go
```diff
@@ -262,7 +263,7 @@ func (n *logDumperNode) dump(ctx context.Context) []error {
 			if !strings.HasPrefix(f, prefix) {
 				continue
 			}
-			if err := n.shellToFile(ctx, "sudo cat "+f, filepath.Join(n.dir, filepath.Base(f))); err != nil {
+			if err := n.shellToFile(ctx, "sudo cat '"+f+"'", filepath.Join(n.dir, filepath.Base(f))); err != nil {
```
johngmyers, line 266:
```suggestion
			if err := n.shellToFile(ctx, "sudo cat '"+strings.ReplaceAll(f, "'", "'\''")+"'", filepath.Join(n.dir, filepath.Base(f))); err != nil {
```

johngmyers:
Is there a plan around fixing where it always passes in a nil `corev1.NodeList`, causing it to log a misleading `dumping node not registered in kubernetes:` for each instance?

rifelpet:
The new code now tries to get a list of nodes if possible. I also exposed the private key as a flag. I chose not to yet expose the ssh user as a flag because i think there are a few optimizations we could make here and i'm curious of other people's opinions:

Should we try to handle clusters in which there are multiple SSH users? for example, some nodes are ubuntu and others are debian. This means we'd need to know or guess the user for each instance (which could be slow for larger clusters) or for each AMI (though we don't currently expose AMI info in the Instance type that the dumper code uses so additional plumbing would be involved).

We do try to guess the ssh user [0] when we fetch the list of instances [1] but we'd still want a way to handle non-standard AMIs. We could expose a string list flag that takes a list of ssh users to loop through.

Some of this seems like premature optimization but I'd like to avoid painting us into a corner with a flag that wont be as flexible as we need in future iterations. Do we guarantee cli stability for the "toolbox" commands? If we don't I suppose we could always adjust the flags and logic later.

Thoughts?


[0] https://github.com/kubernetes/kops/blob/97181b911fb63b192c2046641d9da577d7494159/pkg/resources/aws/aws.go#L501-L521
[1] https://github.com/kubernetes/kops/blob/97181b911fb63b192c2046641d9da577d7494159/pkg/resources/aws/aws.go#L554-L560

PR #7780, opened 2019-10-14
Run node authorizer in kops-controller

justinsb:
We create a GRPC interface and use it to serve the kubelet bootstrap
certificate; we consume it directly from nodeup.

Based heavily on the existing AWS node-authorizer implementation.

PR #8580, opened 2020-02-17
WIP: Bootstrapping using client certificates

justinsb:
Similar to #7780, but using client certificates instead of bootstrap tokens.


justinsb:
Context in https://github.com/kubernetes/kops/pull/7780#issuecomment-586986004

johngmyers:
@k8s-ci-robot this needs rebase

johngmyers:
I'm still researching this, but for AWS I think we'd want the client to authenticate to the server using IAM. The server would then authenticate the caller with `sts:GetCallerIdentity`, much like aws-iam-authenticator does.

Kops would create and assign to instances one IAM role per instancegroup, so instances could not impersonate instances in higher-privilege instancegroups. Alternatively, we could create one IAM role per launchtemplate, preventing an instance with old configuration (or someone who's compromised same) from impersonating an instance with a newer spec.

johngmyers:
@justinsb can this be closed?

PR #8581, opened 2020-02-17
kubelet: map rotate-certificates options

justinsb:
Additive flag mapping, may be useful for kubelet bootstrap.


rifelpet:
looks like this needs a `make crds`

justinsb:
Thanks @rifelpet - you're quite right!

PR #8582, opened 2020-02-17
Create New Default StorageClass: kops-ssd-1-17

joshbranham:
Since kops manages the default storageclass for AWS, I feel we should add some improvements.

First, we can allow volume expansion now which is handy. https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims

Second, in order to prevent race conditions with Pods, setting the `volumeBindingMode` will only create a PV for a PVC and mount it to a node once the pod is scheduled.

https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode

Third, enable default EBS encryption. This one is debatable but I think for a production-ready kubernetes cluster in AWS this a best practice default.

This PR has the `volumeBindingMode` set as well, but I made a new PR since I am suggesting a few other changes. https://github.com/kubernetes/kops/pull/8416

### DIFF
--- a/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml
+++ b/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml
@@ -25,7 +25,7 @@ parameters:
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
-  name: gp2-kops
+  name: kops-ssd-1-17
   annotations:
     storageclass.beta.kubernetes.io/is-default-class: "true"
   labels:
--- a/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml
+++ b/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml
@@ -14,6 +14,18 @@ apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
   name: gp2
+  labels:
+    k8s-addon: storage-aws.addons.k8s.io
+provisioner: kubernetes.io/aws-ebs
+parameters:
+  type: gp2
+
+---
+
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: gp2-kops
   annotations:
     storageclass.beta.kubernetes.io/is-default-class: "true"
   labels:
@@ -21,6 +33,9 @@ metadata:
 provisioner: kubernetes.io/aws-ebs
 parameters:
   type: gp2
+  encrypted: "true"
+allowVolumeExpansion: true
+volumeBindingMode: WaitForFirstConsumer
 
 ---
 
--- a/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/amazonvpc/manifest.yaml
+++ b/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/amazonvpc/manifest.yaml
@@ -73,7 +73,7 @@ spec:
   - id: v1.15.0
     kubernetesVersion: '>=1.15.0'
     manifest: storage-aws.addons.k8s.io/v1.15.0.yaml
-    manifestHash: 23459f7be52d7c818dc060a8bcf5e3565bd87a7b
+    manifestHash: 5e829e8981470696bef2ed5e96839ea83cb36d24
     name: storage-aws.addons.k8s.io
     selector:
       k8s-addon: storage-aws.addons.k8s.io
--- a/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/cilium/manifest.yaml
+++ b/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/cilium/manifest.yaml
@@ -73,7 +73,7 @@ spec:
   - id: v1.15.0
     kubernetesVersion: '>=1.15.0'
     manifest: storage-aws.addons.k8s.io/v1.15.0.yaml
-    manifestHash: 23459f7be52d7c818dc060a8bcf5e3565bd87a7b
+    manifestHash: 5e829e8981470696bef2ed5e96839ea83cb36d24
     name: storage-aws.addons.k8s.io
     selector:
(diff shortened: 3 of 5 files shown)


joshbranham:
Looks like `volumeBindingMode` is immutable, in that case how do we end up managing these? In my opinion, it is a better default as it makes PVC more usable, but do we need to create a new storageclass and set it as the default versus modifying the old?

upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml
```diff
@@ -14,13 +14,28 @@ apiVersion: storage.k8s.io/v1
...
+provisioner: kubernetes.io/aws-ebs
+parameters:
+  type: gp2
+
+---
+
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: gp2-kops
```
hakman: Maybe add purpose instead of the generic `kops` keyword?
```suggestion
  name: gp2-wait
```

justinsb: Sorry to nit-pick... I'm wondering if we should use a consistent prefix, rather than a suffix.  We've started doing this with e.g. clusterroles, and there we've used `kops:` but I don't know if the colon is available everywhere (?).  WDYT about `kops-gp2` or `kops:gp2` in general?

justinsb: That's another good option that avoids my prefix vs suffix nit-pick :-)

joshbranham: I think that is a good idea to start with something consistent and move forward with it. Unfortunately it won't let us do `kops:gp2` but I think `kops-gp2` make sense @justinsb 

Thoughts @hakman

joshbranham: I think the purpose bit makes it too specific since I am also suggesting we set some other non-default settings

hakman: `kops-gp2-1-18`? I am sure we will change it again someday. :)

justinsb: I do think it's OK to change these going forwards; "kops-gp2" can mean "the gp2 class, but otherwise whatever kops recommends".   We should probably have a note that if people want to lock in a specific behaviour, they should probably use their own storage class (though to be honest I'm not sure how users could do that without knowing what features are coming in the future).

This is sort of philosophy with most of the kops fields - if you leave it unspecified, that means "whatever kops recommends" (I sometimes call it omakase).  But if you care, you should specify it, in which case kops honors your intention.  We can - and do - change the defaults, but if you specify a field we either honor it or it is an error.

joshbranham:
although if we are trying to target a better default for new clusters, then maybe release notes with steps to `--force` update the existing one is the better approach than modifying the name? At least in this case, it won't effect existing volumes. However I could see this pattern of `fix the default for new clusters > release notes for manual fix on existing` being harder to implement for things like deployments where labels can't change etc and `--force` just deletes the old causing potentially downtime for that workload. Hmmmmm

upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml
justinsb: (BTW, if we wanted to, we could also name this `kops-ssd`, and then offer storage classes that work across clouds, though I personally find it better to just use the default storage class in my apps)

joshbranham: > (BTW, if we wanted to, we could also name this `kops-ssd`, and then offer storage classes that work across clouds, though I personally find it better to just use the default storage class in my apps)

I do like the idea of `kops-ssd` as a starting point. Should I add a version to it as well? I would like to see this go out whenever it can, so should I target `1.16` or `1.17`? Therefore the name could be:

`kops-ssd-1-16` for instance

johngmyers:
I was under the impression that we were targeting a better default for new volumes, even on old clusters.

upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml
johngmyers: I would prefer we target 1.17. This is too featureish to go into 1.16 at this time.

joshbranham:
> I was under the impression that we were targeting a better default for new volumes, even on old clusters.

That seems like the easiest path forward. If you agree then I’ll just go with kops-ssd-1-17. The likelihood we change these settings much is low so I doubt we’ll run into multiple tweaks per k8s/kops minor version

upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml
joshbranham: Done @johngmyers @justinsb

cpanato:
@joshbranham i will close my PR

justinsb:
Thanks @joshbranham 

/approve
/lgtm

dignajar:
this PR is so strange, changing default value from the previous version (backward compatibility will fail) and adding a version number to the storage class name?

What is going to happen after we update from an early version to v1.17; We are going to have two `defaults` and if for some reason we re-deploy some application will take another type of storage class.

joshbranham:
The two defaults issue was fixed, pr is linked here. The storage class is essentially parameters to tell EBS what to use when making the volume. It has no impact on existing volumes and you can delete storage classes without affecting existing volumes. This change was just meant to make expanding volumes possible, primarily. Do you see an issue still even with the linked PRs? I will agree the defaulting behavior was a mistake, and hard to manage in long lived clusters anyways when things are always apply-only.

PR #8585, opened 2020-02-17
Staging build: don't use _GIT_TAG

justinsb:
The _GIT_TAG has a prefix vYYYYMMDD, so is not a valid semver.

Instead we use our CI=1 approach to construct a reasonable semver tag.


rifelpet:
ah thats much cleaner than what I was considering!

/lgtm

PR #8587, opened 2020-02-18
1.16 Release Notes - Add Known CoreDNS metrics regression

gjtempleton:
As per discussion at office hours.

I've followed the style and syntax as much as possible from upstream k/k given we haven't had known issues blocks since 1.6, open to feedback especially as it doesn't actually apply to all releases in the 1.16 train.


johngmyers:
/lgtm cancel
I believe you need to land this into master.

gjtempleton:
Yep, you're right, teaches me for raising these at the end of a long day.

PR #8588, opened 2020-02-19
1.16 Release Notes - Add Known CoreDNS metrics regression

gjtempleton:
Replacement for #8587 

As per discussion at office hours.

I've followed the style and syntax as much as possible from upstream k/k given we haven't had known issues blocks since 1.6, open to feedback especially as it doesn't actually apply to all releases in the 1.16 train.

### DIFF
--- a/docs/releases/1.16-NOTES.md
+++ b/docs/releases/1.16-NOTES.md
@@ -38,12 +38,19 @@ the notes prior to the release).
       featureGates:
         PodPriority: "true"
   ```
- 
+
 # Deprecations
 
 * Support for Kubernetes releases prior to 1.9 is deprecated and will be removed in kops 1.18.
 
 * The `kops/v1alpha1` API is deprecated and will be removed in kops 1.18. Users of `kops replace` will need to supply v1alpha2 resources.
+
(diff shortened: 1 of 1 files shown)


rifelpet:
Any reason this shouldn't also be mentioned in 1.17 and 1.18 release notes? Are we hoping to have upgraded to 1.6.7 in those branches?

rajansandeep:
The coredns version 1.6.7 is now made available in gcr.io (This would be the version in upstream k/k 1.18)

```
$ docker pull k8s.gcr.io/coredns:1.6.7
1.6.7: Pulling from coredns
Digest: sha256:2c8d61c46f484d881db43b34d13ca47a269336e576c81cf007ca740fa9ec0800
Status: Downloaded newer image for k8s.gcr.io/coredns:1.6.7

```

gjtempleton:
I was assuming that the 1.6.7 publish to `k8s.gcr.io` would happen before 1.17 and 1.18 hit GA. Seems it happened even faster than anticipated.

I guess in that case time to close this PR and re-pick the updates into the release branches then.

PR #8589, opened 2020-02-19
Bump Cilium to 1.7 for k8s 1.12+

olemarkus:
Cilium 1.7 requires K8s 1.12 minimum. Changed the templates so that we
can have different cilium versions for different k8s versions.

This also mean that this addon will behave similar to other addons wrt
upgrades. Cilium used to add a fixed version to the cluster spec on cluster creation so
upgrades were slightly more manual. Now, for new clusters, upgrades will
happen implicitly with kops updates unless the .Version is added
manually to the cluster spec.

### DIFF
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -593,9 +593,6 @@ func (c *Cluster) FillDefaults() error {
 	} else if c.Spec.Networking.AmazonVPC != nil {
 		// OK
 	} else if c.Spec.Networking.Cilium != nil {
-		if c.Spec.Networking.Cilium.Version == "" {
-			c.Spec.Networking.Cilium.Version = CiliumDefaultVersion
-		}
 		// OK
 	} else if c.Spec.Networking.LyftVPC != nil {
 		// OK
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
@@ -163,6 +163,14 @@ rules:
   - list
   - watch
 - apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - ""
   resources:
   - namespaces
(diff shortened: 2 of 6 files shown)


upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -191,15 +199,6 @@ rules:
...
   - patch
-- apiGroups:
-  - extensions
-  resources:
-  - ingresses
-  verbs:
-  - create
-  - get
-  - list
-  - watch
```
johngmyers: Won't removing this adversely affect version 1.6?

olemarkus:
Yes you may be right there. I'll add it back in

johngmyers:
You could put a version conditional in the template.

olemarkus:
> You could put a version conditional in the template.

I would if this was more security sensitive. But this particular change is basically just about dropping the deprecated k8s extensions apis.

pkg/apis/kops/cluster.go
```diff
@@ -593,9 +593,6 @@ func (c *Cluster) FillDefaults() error {
 	} else if c.Spec.Networking.AmazonVPC != nil {
 		// OK
 	} else if c.Spec.Networking.Cilium != nil {
-		if c.Spec.Networking.Cilium.Version == "" {
-			c.Spec.Networking.Cilium.Version = CiliumDefaultVersion
```
johngmyers: Instead of baking the logic in the templates, this could have picked the default version by doing check on the Kubernetes version. That would allow other Go code to do Cilium version checks. But we don't have any such Cilium version checks, so not a blocker.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -924,7 +924,7 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 
 	if b.cluster.Spec.Networking.Cilium != nil {
 		key := "networking.cilium.io"
-		version := "1.6.6-kops.0"
+		version := "1.7.0-kops.1"
```
hakman: Maybe use something like this here?
https://github.com/kubernetes/kops/blob/91867ce4b5f945d88a09a1424b48d22a76a4617d/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go#L861-L866

hakman:
Seems that e2e for Cilium is failing after this change. @olemarkus any thoughts?
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kops-aws-cni-cilium/1230338989827821576

olemarkus:
Yep. I'll have a PR momentarily

PR #8592, opened 2020-02-20
add s3 region

zetaab:
we are currently seeing following error in etcd-manager:

```
W0220 12:42:07.934295    8002 controller.go:149] unexpected error running etcd cluster reconciliation loop: error refreshing control store after leadership change: error reading s3://xxxx/bbb.k8s.local/backups/etcd/events/control: error listing s3://xxxx/bbb.k8s.local/backups/etcd/events/control: AuthorizationHeaderMalformed: The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'eu-north-1'
	status code: 400, request id: 9C82F7B26B72D41D, host id: bNWE4d3rO/CnkzlfpTrUcW4LcMdxuBH7M5tY6AmOivMLSvaBqrEc8PpiaOXHDKuBi5jgx6B07Cw=
```

this is because we are using custom s3 credentials and s3 region is not passed to etcd-manager container (so it defaults to us-east-1)

### DIFF
--- a/util/pkg/env/standard.go
+++ b/util/pkg/env/standard.go
@@ -42,6 +42,7 @@ func BuildSystemComponentEnvVars(spec *kops.ClusterSpec) EnvVars {
 	}
 
 	// Custom S3 endpoint
+	vars.addEnvVariableIfExist("S3_REGION")
 	vars.addEnvVariableIfExist("S3_ENDPOINT")
 	vars.addEnvVariableIfExist("S3_ACCESS_KEY_ID")
 	vars.addEnvVariableIfExist("S3_SECRET_ACCESS_KEY")


rifelpet:
How are you passing custom credentials to etcd-manager? Just curious

zetaab:
> How are you passing custom credentials to etcd-manager? Just curious

I am defining following before running kops commands like kops create cluster:
```
export S3_ACCESS_KEY_ID=foobar
export S3_SECRET_ACCESS_KEY=xxxx
export KOPS_STATE_STORE=s3://yyyy
export S3_ENDPOINT=https://s3.eu-north-1.amazonaws.com
export S3_REGION=eu-north-1
```

These environment variable are then passed to etcd manager container (or before this PR all other than S3_REGION) was passed. I have tested this PR and now everything works. Use-case for this solution is that we have shared one s3 bucket to many clusters. Of course each cluster do have own iam account that those are isolated. We are doing this because some of our clusters are in OpenStack, GCE and AWS

PR #7832, opened 2019-10-28
cilium: don't try to mount sys/fs/bpf if already mounted

justinsb:
systemd v238 already includes the mount, and the unit file fails.  We
test for the existence of the mount, rather than testing systemd
versions directly.

PR #8599, opened 2020-02-20
Cilium fix bpffs check

olemarkus:
The check introduced in #7832 is unfortunately too naïve to detect if bpffs has been mounted leading to the unit file never being installed. This is quite disastrous for cilium.

This check in this PR should be more robust.

/kind bug

### DIFF
--- a/nodeup/pkg/model/network.go
+++ b/nodeup/pkg/model/network.go
@@ -76,7 +76,7 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
 		err := unix.Statfs("/sys/fs/bpf", &fsdata)
 
 		if err != nil {
-			alreadyMounted = false
+			return fmt.Errorf("error checking for /sys/fs/bpf: %v", err)
 		} else {
 			alreadyMounted = int32(magic) == int32(fsdata.Type)
 		}
--- a/nodeup/pkg/model/network.go
+++ b/nodeup/pkg/model/network.go
@@ -18,9 +18,9 @@ package model
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 
+	"golang.org/x/sys/unix"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
 )
@@ -70,15 +70,15 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
 	if networking.Cilium != nil {
 		// systemd v238 includes the bpffs mount by default; and gives an error "has a bad unit file setting" if we try to mount it again (see mount_point_is_api)
 		var alreadyMounted bool
-		_, err := os.Stat("/sys/fs/bpf")
+		// bpffs magic number
+		magic := uint32(0xCAFE4A11)
+		var fsdata unix.Statfs_t
+		err := unix.Statfs("/sys/fs/bpf", &fsdata)
+
 		if err != nil {
-			if os.IsNotExist(err) {
-				alreadyMounted = false
-			} else {
-				return fmt.Errorf("error checking for /sys/fs/bpf: %v", err)
-			}
+			alreadyMounted = false
 		} else {
-			alreadyMounted = true
+			alreadyMounted = int32(magic) == int32(fsdata.Type)
 		}
 
 		if !alreadyMounted {
--- a/nodeup/pkg/model/network.go
+++ b/nodeup/pkg/model/network.go
@@ -68,18 +68,19 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	if networking.Cilium != nil {
-		// systemd v238 includes the bpffs mount by default; and gives an error "has a bad unit file setting" if we try to mount it again (see mount_point_is_api)
-		var alreadyMounted bool
-		// bpffs magic number
-		magic := uint32(0xCAFE4A11)
 		var fsdata unix.Statfs_t
 		err := unix.Statfs("/sys/fs/bpf", &fsdata)
 
 		if err != nil {
 			return fmt.Errorf("error checking for /sys/fs/bpf: %v", err)
-		} else {
-			alreadyMounted = int32(magic) == int32(fsdata.Type)
 		}
+
+		// systemd v238 includes the bpffs mount by default; and gives an error "has a bad unit file setting" if we try to mount it again (see mount_point_is_api)
+		var alreadyMounted bool
+		// bpffs magic number. See https://github.com/torvalds/linux/blob/v4.8/include/uapi/linux/magic.h#L80
+		magic := uint32(0xCAFE4A11)
+
+		alreadyMounted = int32(magic) == int32(fsdata.Type)
 
 		if !alreadyMounted {
 			unit := s(`
--- a/go.mod
+++ b/go.mod
@@ -124,7 +124,7 @@ require (
 	golang.org/x/crypto v0.0.0-20191202143827-86a70503ff7e
 	golang.org/x/net v0.0.0-20191126235420-ef20fe5d7933
 	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
-	golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a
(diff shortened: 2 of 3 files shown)


rifelpet:
@olemarkus I see the original PR was backported all the way to 1.15. Should this be considered a blocker for 1.16.0 stable?

olemarkus:
I don't understand why, but the 1.16 and 1.15 branch is not affected. The /sys/fs/bpf path doesn't exist at the time of the check on those branches even if I use the same ami. Another change added to kops after 1.16 branch was cut must be causing this.

hakman:
Just a quick update, the Cilium periodic e2e is working again:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kops-aws-cni-cilium/1230702775738830848

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -399,7 +399,7 @@ spec:
           value: {{ . }}
         {{ end }}
 {{ with .Networking.Cilium }}
-        image: "docker.io/cilium/cilium:{{- or .Version "v.1.7.0" }}"
```
yurrriq: oops

olemarkus:
@rifelpet unfortunately, this does indeed break cilium back to 1.15.2. It should probably block 1.16.

hakman:
@olemarkus the sorting in `vendor/modules.txt` seems off can you check that you are using go v1.13.x?

olemarkus:
I am using go1.13.4 and just running `make gomod`. When I run it again, there are no changes to the vendor/modules.txt file.

olemarkus:
For some reason `make gomod` used another binary than expected. Things looks a lot more sane now.

hakman:
Thanks for the update @olemarkus. I was just about to start checking why it worked differently on my end.

hakman:
Me and @rifelpet did some research yesterday on this topic and will add some questions.

nodeup/pkg/model/network.go
```diff
@@ -70,15 +70,15 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
...
+		magic := uint32(0xCAFE4A11)
+		var fsdata unix.Statfs_t
+		err := unix.Statfs("/sys/fs/bpf", &fsdata)
+
 		if err != nil {
-			if os.IsNotExist(err) {
-				alreadyMounted = false
-			} else {
-				return fmt.Errorf("error checking for /sys/fs/bpf: %v", err)
-			}
```
hakman: I think this is still valid. If there is an unexpected error when reading from `/sys/fs/bp`, the systemd mount would not help, right?

nodeup/pkg/model/network.go
```diff
@@ -70,15 +70,15 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
...
+			alreadyMounted = false
 		} else {
-			alreadyMounted = true
+			alreadyMounted = int32(magic) == int32(fsdata.Type)
```
hakman: What happens if the fs type is not BPF? Would it be better to return with an error also?

The Cilium code describes the logic for mounting the BPF fs and there is also a mention of a fallback location `/run/cilium/bpffs`:
https://github.com/cilium/cilium/blob/cbd63be46569751ac45060e54f6787f1286b75cd/pkg/bpf/bpffs_linux.go#L213-L224

olemarkus: If BPF is not mounted on `/sys/fs/bpf`, the path will still exist, which is why the old check fails. So most likly the fs type will not be bpf before we install the unit.

`/run/cilium/bpffs` is a fallback if the host is not mounting bpf. This will cause network disruptions every time cilium pods restart though since the bpf mounts are lost.

nodeup/pkg/model/network.go
olemarkus: This is probably still valid. Most likely this would be due to the path being absent and if the unit tries to mount it will just fail. Result will be quite similar and I suppose equally easy to detect.

nodeup/pkg/model/network.go
hakman: OK, more clear now. Thanks for clarifying the fallback part.

Regarding the check, if the path exists, `unix.Statfs` would succeed and will return `SYSFS_MAGIC=0x62656572`  or something similar. If there is an error other than the "no such file or directory", maybe it's not a good thing to continue. Same goes for a magic number that is not SYSFS_MAGIC or BPF_FS_MAGIC.

olemarkus: Shouldn't you be able to mount bpffs on whatever path that exists?

I will add back the return on error shortly. `unix.Statfs("/sys/fs/bpf", &fsdata)` would return error on file not found too, shouldn't it? So the os.Stat check would be redundant.

hakman: Just checked and you are right, if the mount path doesn't exist, the unit won't help. Probably for any type of error BPF mounting will fail.

nodeup/pkg/model/network.go
hakman: 👍

nodeup/pkg/model/network.go
hakman: This part is mostly ok. There may be a rare case where the magic number is different from `SYSFS_MAGIC=0x62656572` or `BPF_FS_MAGIC= 0xCAFE4A11` and mounting will fail. I don't think it's worth adding it for now.

nodeup/pkg/model/network.go
```diff
@@ -70,15 +70,15 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
...
+		err := unix.Statfs("/sys/fs/bpf", &fsdata)
+
 		if err != nil {
-			if os.IsNotExist(err) {
-				alreadyMounted = false
-			} else {
-				return fmt.Errorf("error checking for /sys/fs/bpf: %v", err)
-			}
+			return fmt.Errorf("error checking for /sys/fs/bpf: %v", err)
 		} else {
```
rifelpet: nit: we don't need this else statement anymore.

nodeup/pkg/model/network.go
```diff
@@ -70,15 +70,15 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
 	if networking.Cilium != nil {
 		// systemd v238 includes the bpffs mount by default; and gives an error "has a bad unit file setting" if we try to mount it again (see mount_point_is_api)
 		var alreadyMounted bool
-		_, err := os.Stat("/sys/fs/bpf")
+		// bpffs magic number
+		magic := uint32(0xCAFE4A11)
```
rifelpet: it might be worth adding this link as a comment: https://github.com/torvalds/linux/blob/v4.8/include/uapi/linux/magic.h#L80

hakman: Also would be nice to move the var and comment just before it's used.

rifelpet:
/lgtm
/approve

We increased the frequency of the cilium periodic job to run hourly, we'll keep an eye on it to confirm things are still stable before we merge cherry-picks of this 👍

justinsb:
I am a little confused ... https://k8s-testgrid.appspot.com/google-aws#kops-aws-cni-cilium seems to suggest this broke two days ago, but my original (insufficient) PR is from a month or two ago.  Do we know what caused it to break?  (Was it a newer version of cilium that makes the directory?)

olemarkus:
Just bad timing. The e2e tests broke because of the typo fixed here #8591 
The bpf mount bug this PR fixed is unrelated to the e2e tests.

justinsb:
Ah - thanks @olemarkus  (and thanks also for fixing!)

mazzy89:
@olemarkus I'm running k8s 1.15.9 with cilium from kops 1.15.2 but I'm not having this issue.Is it clear for you why this version is not affected?

olemarkus:
It is affected too. If you see the cilium agent logs, you should see a warning about bpf not already mounted on the hair

mazzy89:
I see. I'll take a look at that. Which are the effects of this? I did not notice anything weird in the cluster. Everything is operative and healthy in term of networking.

PR #8600, opened 2020-02-21
Fail cluster validation if a master missing kube-controller-manager

johngmyers:
This should fix the known cause of cluster validation flapping.


johngmyers:
@zetaab I'd appreciate it if you could give this a try.

johngmyers:
@zetaab This has been backported into 1.17.0-beta.1. I'd appreciate it if you could let me know how it is working.

PR #8610, opened 2020-02-22
Use supported kubernetes versions in tests

johngmyers:


justinsb:
Thanks - probably time to stop thinking about 1.4 :-)

/approve
/lgtm

PR #8614, opened 2020-02-22
Remove support for Canal and the vxlan Flannel backend

### LINKED ISSUE #8562: 1.17 alpha versions causing regression for kiam?
 this information.**

Any of the 1.17 alphas so far.

 a `kops` flag.**

seen in 1.17.0-rc.2 through 1.17.3. Works without issue on kops/k8s 1.15 and 1.16 built clusters. ONLY change is bump to 1.17.x.

AWS

Try to install kiam via it's included helm chart onto a kops 1.17.x built-cluster

The kiam-agent daemonsets crashloop

No crashloop.

Will follow up with this if asked for. Main thing applicable here is we are using CoreDNS

From the agent logs w/ gRPC debugging enabled:
```
kubectl -n kube-system logs kiam-agent-5z8tt
{"level":"info","msg":"started prometheus metric listener 0.0.0.0:9620","time":"2020-02-14T18:13:41Z"}
INFO: 2020/02/14 18:13:41 parsed scheme: "dns"
INFO: 2020/02/14 18:13:46 grpc: failed dns SRV record lookup due to lookup _grpclb._tcp.kiam-server on 100.64.0.10:53: dial udp 100.64.0.10:53: operation was canceled.
WARNING: 2020/02/14 18:13:46 grpc: failed dns A record lookup due to lookup kiam-server on 100.64.0.10:53: dial udp 100.64.0.10:53: operation was canceled.
INFO: 2020/02/14 18:13:46 ccResolverWrapper: got new service config: 
INFO: 2020/02/14 18:13:46 ccResolverWrapper: sending new addresses to cc: []
{"level":"fatal","msg":"error creating server gateway: error dialing grpc server: context deadline exceeded","time":"2020-02-14T18:13:46Z"}
```

Bug also posted with kiam folks here: uswitch/kiam#378

johngmyers:
Due to coreos/flannel#1243

Targeting kops 1.17

Fixes #8562

### DIFF
--- a/docs/releases/1.17-NOTES.md
+++ b/docs/releases/1.17-NOTES.md
@@ -21,6 +21,9 @@ the notes prior to the release).
 * Since 1.16, a controller is now used to apply labels to nodes.  If
   you are not using AWS, GCE or OpenStack your (non-master) nodes may
   not have labels applied correctly.
+  
+* As of Kubernetes 1.17 the Canal CNI and the "vxlan" backend of the Flannel CNI
+  are no longer supported due to [a bug in Flannel](https://github.com/coreos/flannel/issues/1243).
 
 # Required Actions
 
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -35,7 +35,7 @@ import (
 
 func newValidateCluster(cluster *kops.Cluster) field.ErrorList {
 	allErrs := validation.ValidateObjectMeta(&cluster.ObjectMeta, false, validation.NameIsDNSSubdomain, field.NewPath("metadata"))
-	allErrs = append(allErrs, validateClusterSpec(&cluster.Spec, field.NewPath("spec"))...)
+	allErrs = append(allErrs, validateClusterSpec(&cluster.Spec, cluster, field.NewPath("spec"))...)
 
 	// Additional cloud-specific validation rules
 	switch kops.CloudProviderID(cluster.Spec.CloudProvider) {
@@ -48,7 +48,7 @@ func newValidateCluster(cluster *kops.Cluster) field.ErrorList {
 	return allErrs
 }
 
-func validateClusterSpec(spec *kops.ClusterSpec, fieldPath *field.Path) field.ErrorList {
+func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	allErrs = append(allErrs, validateSubnets(spec.Subnets, fieldPath.Child("subnets"))...)
@@ -89,7 +89,7 @@ func validateClusterSpec(spec *kops.ClusterSpec, fieldPath *field.Path) field.Er
 	}
 
 	if spec.Networking != nil {
-		allErrs = append(allErrs, validateNetworking(spec, spec.Networking, fieldPath.Child("networking"))...)
+		allErrs = append(allErrs, validateNetworking(spec.Networking, c, fieldPath.Child("networking"))...)
 		if spec.Networking.Calico != nil {
 			allErrs = append(allErrs, validateNetworkingCalico(spec.Networking.Calico, spec.EtcdClusters[0], fieldPath.Child("networking", "calico"))...)
 		}
@@ -285,7 +285,7 @@ func validateKubeAPIServer(v *kops.KubeAPIServerConfig, fldPath *field.Path) fie
 	return allErrs
 }
 
-func validateNetworking(c *kops.ClusterSpec, v *kops.NetworkingSpec, fldPath *field.Path) field.ErrorList {
+func validateNetworking(v *kops.NetworkingSpec, c *kops.Cluster, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	optionTaken := false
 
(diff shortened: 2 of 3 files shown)


johngmyers:
There's activity on the underlying bug, so
/hold

joshbranham:
Yeah this thread also has some activity, the maintainer is referenced as well: https://github.com/coreos/flannel/issues/1245

jhohertz:
It would be great if the flannel folks get this fixed up, but there doesn't seem to be a lot of movement.

johngmyers:
I think I'll asses the state in 2-3 weeks. If no progress by then I'll un-hold.

This should be a 1.17 blocker.

gamer22026:
How does this affect running flannel in AWS ? Even with disabling Source/Destination Check on the nodes, I was not able to get flannel with `host-gw` backend working (k8s 1.16.7) and aws-vpc is a non-starter.

johngmyers:
I don't see any progress.
/hold cancel

johngmyers:
Decision from Kops Office Hours is to defer for two weeks.
/hold

johngmyers:
Decision of kops office hours is to proceed.
/hold cancel

andersosthus:
There's an open PR now to fix this issue at https://github.com/coreos/flannel/pull/1282.
For our part we use Flannel to run Windows Worker nodes, and Flannel is afaik the only supported overlay (https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#networking)

justinsb:
I've also (so far) only been able to reproduce the problem on rhel 7.8 images.

justinsb:
And I actually think that https://github.com/kubernetes/kops/pull/8381/files might be a fix.

justinsb:
I'm trying to get some flannel tests going on "every distro": https://github.com/kubernetes/test-infra/pull/17300/files

jhohertz:
Just a data point, I tried adding the sysctls from the CentOS7 fix to a Flatcar-based cluster and didn't see any change in flannel's behaviour.

Trying to figure out the best way to shoehorn the ethtool/checksum fix in for a test now...

hakman:
Seems that the proper  fix was accepted: https://github.com/kubernetes/kubernetes/issues/88986#issuecomment-620633097. Getting it to already released distros will probably take a long time, if it ever happens.

The [patch comment](http://patchwork.ozlabs.org/project/netfilter-devel/patch/335a95d93767f2b58ad89975e4a0b342ee00db91.1587429321.git.gnault@redhat.com/) says that `the problem seems to be there from day one, although it was probably not visible before UDP tunnels were implemented`. 

For now the best hope for the near future is the https://github.com/coreos/flannel/pull/1282, which will require a new Flannel release.

jhohertz:
I'm not sure the "proper" fix is the whole story here. I rolled a version of flatcar with the cited kernel patch and see no difference in behaviour for flannel vxlan / canal.

jhohertz:
I can however now confirm using ethtool to disable offloading works.

hakman:
@jhohertz just curious, how did you try flatcar with that patch?

jhohertz:
@hakman I forked the manifest and coreos-overlay repositories to add the patch. [Running cork create](https://docs.flatcar-linux.org/os/sdk-modifying-coreos/) for my attempt looks like: ```cork create --manifest-branch viafoura-build-2345.3.1 --manifest-url https://github.com/viafoura/flatcar-manifest.git```. There are additional steps to actually turn the result into an AMI.

The backport had one minor difference for 4.19 in call signatures but otherwise the logic was identical. There might be other differences that need accounting for, but the patch tiny... At any rate, my attempt lives here in the context of the build setup above: https://github.com/viafoura/flatcar-coreos-overlay/blob/viafoura-build-2345.3.1/sys-kernel/coreos-sources/files/4.19/z0004-netfilter-nat-never-update-the-UDP-checksum-when-it-.patch

jhohertz:
Re: the ethtool thing, I'm going to try to adapt [this](https://t.du9l.com/2020/03/kubernetes-flannel-udp-packets-dropped-for-wrong-checksum-workaround/) into a hooks yaml fragment suitable for a cluster spec. Not really used that bit of functionality before, I'll share if I get a working example.

jhohertz:
At first glance, this seems to work in a cluster spec to implement the workaround in a not terrible way. FAR from extensively tested. And of course if this is actually disabling real hardware offload, this isn't an ideal fix, but it may be the best we can get for now.

```
spec:
  hooks:
  - name: flannel-cksum-offload-disable.service
    useRawManifest: true
    roles: [Master, Node]
    manifest: |
      [Unit]
      Description=Turn off checksum offload on flannel.1
      After=sys-devices-virtual-net-flannel.1.device
      [Install]
      WantedBy=sys-devices-virtual-net-flannel.1.device
      [Service]
      Type=oneshot
      ExecStart=/sbin/ethtool -K flannel.1 tx-checksum-ip-generic off

```

hakman:
> Re: the ethtool thing, I'm going to try to adapt [this](https://t.du9l.com/2020/03/kubernetes-flannel-udp-packets-dropped-for-wrong-checksum-workaround/) into a hooks yaml fragment suitable for a cluster spec. Not really used that bit of functionality before, I'll share if I get a working example.

@jhohertz That is not such a bad idea. May be something to think about adding in Kops as a workaround until a new version of Flannel is released.

johngmyers:
Closing in favor of #9074

PR #8617, opened 2020-02-24
Validate at most one networking option chosen

johngmyers:


olemarkus:
Great work. Really appreciate these CNI specific functions.

rifelpet:
Do you think it is worth adding a test that covers specifying two CNI providers and ensuring an error is returned? otherwise it looks good to me

johngmyers:
I considered a test, but in order to be even remotely effective it would have to exercise all combinations of two providers chosen from all other than Classic. It seemed to have too little benefit for the cost, especially given the current state of the api validation code.

rifelpet:
I agree the benefit from a test is not worth the effort.

/lgtm
/approve

PR #8618, opened 2020-02-24
Update Calico and Canal to v3.12.0

hakman:
Calico 3.12 add support for autodetecting the iptables backend, quite useful for the people using Debian Buster, very soon the default base image in Kops.
https://docs.projectcalico.org/v3.12/release-notes/

This PR moves Calico and Canal to 3.12.0 and bumps the patch version for some of the previous releases.

### DIFF
--- a/pkg/apis/kops/networking.go
+++ b/pkg/apis/kops/networking.go
@@ -118,7 +118,7 @@ type CalicoNetworkingSpec struct {
 	// MajorVersion is the version of Calico to use
 	MajorVersion string `json:"majorVersion,omitempty"`
 	// IptablesBackend controls which variant of iptables binary Felix uses
-	// Default: Legacy (other options: NFT)
+	// Default: Auto (other options: Legacy, NFT)
 	IptablesBackend string `json:"iptablesBackend,omitempty"`
 	// IPIPMode is mode for CALICO_IPV4POOL_IPIP
 	IPIPMode string `json:"ipipMode,omitempty"`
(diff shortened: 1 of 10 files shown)


pkg/apis/kops/networking.go
```diff
@@ -118,7 +118,7 @@ type CalicoNetworkingSpec struct {
 	// MajorVersion is the version of Calico to use
 	MajorVersion string `json:"majorVersion,omitempty"`
 	// IptablesBackend controls which variant of iptables binary Felix uses
-	// Default: Legacy (other options: NFT)
+	// Default: Auto (other options: Legacy, NFT)
```
johngmyers: Perhaps add validations for these in `pkg/apis/kops/validation/validation.go`?

hakman:
@KashifSaadat any chance you could take a look at this? I would like to cherry-pick it in 1.16 before the release.

hakman:
No worries @KashifSaadat. I don't think there were any changes in the typha section.
Maybe I am missing something. If so, can you point me in the right direction?

pkg/apis/kops/validation/validation.go
```diff
@@ -430,25 +430,24 @@ func validateNetworkingFlannel(v *kops.FlannelNetworkingSpec, fldPath *field.Pat
...
 	}
 
-	chainInsertMode := v.ChainInsertMode
-	switch chainInsertMode {
-	case "", "insert", "append":
-	default:
-		allErrs = append(allErrs, field.NotSupported(fldPath.Child("chainInsertMode"), chainInsertMode, []string{"insert", "append"}))
+	if v.ChainInsertMode != "" {
+		valid := []string{"insert", "append"}
+		allErrs = append(allErrs, IsValidValue(fldPath, &v.ChainInsertMode, valid)...)
```
KashifSaadat: Changing from `fldPath.Child("chainInsertMode")` to `fldPath` results in the validation message being less specific, i.e.:
```
# error populating cluster spec: spec.networking.canal: Unsupported value: "bad": supported values: "insert", "append"
```

Can you correct this for all cases here.

KashifSaadat:
> No worries @KashifSaadat. I don't think there were any changes in the typha section.
> Maybe I am missing something. If so, can you point me in the right direction?

I misread my bad, it looks fine! 👍

pkg/apis/kops/validation/validation.go
hakman: Thanks for noticing this. I think it should be better now.

hakman:
Thanks for taking the time for this @KashifSaadat :)

Issue #8391, opened 2020-01-21
Cluster Rolling Update support the --ignore-daemonsets flag


denists1984:
1. **What kops version are you running?** 
Version 1.15.0

2. **What Kubernetes version are you running?**
Version:"v1.15.6"

3. **What commands did you run? What is the simplest way to reproduce this issue?**
kops rolling-update cluster ${CLUSTER_NAME} --fail-on-validate-error="false" 

4. **What happened after the commands executed?**
The rolling update was started but in the logs i see that DaemonSets will be ignored and not evicted  during the node draining. 
`Draining the node: "fluentd-logging_node".`
`WARNING: ignoring DaemonSet-managed Pods: kube-system/fluentd-logging`

5. **What did you expect to happen?**
I would expect from `kops rolling-update cluster` to support the `--ignore-daemonsets=false` flag in order to drain the node completely and gracefully terminate the fluentd-logging pod before the node termination.

My question is if there are any plans to support this option in the future releases?

mmerrill3:
+1

mmerrill3:
It looks like that option is hard coded right now.  I have run into an issue with this setting with weave, since pods remain running on the node after it is deleted from the cluster, and weave is still participating in the ipam ring, after another weave instance claimed the IP space.

This is the location where the changes would need to allow for the option to ignore daemonsets would be included.

https://github.com/kubernetes/kops/blob/cba59afac45c1f20491ddb9f89133c951907ae19/pkg/instancegroups/instancegroups.go#L390-L391

I'm thinking that a good solution would be to allow this behavior at the instance group level, not universally across all the instance groups.

johngmyers:
I suspect what might make sense is to first evict all non-daemonsets, then evict the daemonsets. But if, for example, fluentd-logging depended on weave then things would be more complicated.

/area rolling-update

johngmyers:
On second thought, this might be a duplicate of #7119. It should probably be termination that gracefully stops the DaemonSet pods, not rolling update.

mmerrill3:
[...]

Issue #7119, opened 2019-06-07
Drain nodes when they terminate


grosser:
When the ASG scales down the node should be drained and not just terminated (killing pods / not respecting poddisruptionbudget)

Either via "Amazon EC2 Auto Scaling Lifecycle Hooks" (up to 60 min) or with a termination script https://github.com/miglen/aws/blob/master/ec2/run-script-on-ec2-instance-termination.md (2 min max, so will be tight)

Flow could be Scale-Down -> SQS -> Drainer or Scale-Down -> SQS -> Node status -> https://github.com/planetlabs/draino

We might be able to contribute this, but need some "yes that's a good idea" / "yes we want this" first :)

pracucci:
I would personally suggest with the easiest approach, which looks the termination script and eventually iterate on it over the time.

so0k:
I did a detailed write up on setting up LCH / SQS / node-drainer here

https://mofoseng.github.io/posts/eks-migration/#node-maintenance-autoscaling-and-lifecyclehooks

I compared kube-aws drainer (which has a single replica deployment updating a configmap and ds on each node grepping the configmap... ) to a pure AWS lambda based approach and this SQS approach seemed the most robust

it's not kops specific but it may help adoption (I moved company and now I'm back on kops after migrating out of it at my last company)

grosser:
FYI We use a similar workflow but instead of a configmap set a label on the
node to mark it as draining ... Also looking at moving to a single
deployment to have less pods flying around.

so0k:
For now we implemented option 2: https://github.com/compareasiagroup/kube-node-lifecycle-labeller

kforsthoevel:
Another way of handling node draining on termination taken from Zalando's kubernetes-on-aws: https://github.com/zalando-incubator/kubernetes-on-aws/blob/449f8f3bf5c60e0d319be538460ff91266337abc/cluster/userdata-worker.yaml#L92-L120

kforsthoevel:
I have just implemented node drain via systemd. The systemd unit gets provisioned via kops [hooks](https://github.com/kubernetes/kops/blob/1.15.0/docs/cluster_spec.md#hooks). The kubeconfg is written to disk by a daemonset.

Works like a charm. Thanks @thomaspeitz for your support.

paalkr:
@kforsthoevel , that sounds very interesting. Do you mind sharing the implementation details?

kforsthoevel:
[...]

PR #8619, opened 2020-02-24
Enabling the ability to drain daemon set pods, with priority

mmerrill3:
This PR gives the ability to drain daemon set pods during a rolling update.  Daemon set pods will be deleted last.  This is to address issue (#8391) 

Signed-off-by: mmerrill3 <michael.merrill@vonage.com>

### DIFF
--- a/pkg/drain/drain.go
+++ b/pkg/drain/drain.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"sort"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -145,6 +146,7 @@ func (d *Helper) GetPodsForDeletion(nodeName string) (*podDeleteList, []error) {
 
 	for _, pod := range podList.Items {
 		var status podDeleteStatus
+		var priority podDeletePriorty = podDeletionPriorityHighest
 		for _, filter := range d.makeFilters() {
 			status = filter(pod)
 			if !status.delete {
@@ -153,12 +155,26 @@ func (d *Helper) GetPodsForDeletion(nodeName string) (*podDeleteList, []error) {
 				// through any additional filters
 				break
 			}
+			if status.priority < priority {
+				//a previous filter had determined that this pod should be deleted at a lower priority
+				// use that priority
+				priority = status.priority
+			}
 		}
+		status.priority = priority
 		pods = append(pods, podDelete{
 			pod:    pod,
 			status: status,
 		})
 	}
+
+	//sort the list to come up with a sensible order for deleting pods
+	//for backwards compatibility, the default priority is highest.  If a daemon-set is found
+	//and ignore-daemonsets is set to false, the priority will be the lowest.
+	//This sort will return the order of pods to delete, with the highest priority first.
+	sort.SliceStable(pods, func(i, j int) bool {
+		return pods[i].status.priority > pods[j].status.priority
+	})
 
 	list := &podDeleteList{items: pods}
 
--- a/pkg/drain/drain_test.go
+++ b/pkg/drain/drain_test.go
@@ -1,0 +1,186 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package drain
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func getTestSetup(objects []runtime.Object) *fake.Clientset {
+	k8sClient := fake.NewSimpleClientset(objects...)
+	return k8sClient
+}
+
+func dummyDaemonSet(name string) appsv1.DaemonSet {
+	return appsv1.DaemonSet{
+		ObjectMeta: v1meta.ObjectMeta{
+			Name:      name,
+			Namespace: "kube-system",
+		},
+	}
+}
+
+func dummyPod(podMap map[string]string) v1.Pod {
+	pod := v1.Pod{
+		ObjectMeta: v1meta.ObjectMeta{
+			Name:      podMap["name"],
+			Namespace: "kube-system",
+		},
+		Spec: v1.PodSpec{
+			NodeName: "node1",
+		},
+		Status: v1.PodStatus{
+			Phase: v1.PodPhase(podMap["phase"]),
+			ContainerStatuses: []v1.ContainerStatus{
+				{
+					Name:  "container1",
+					Ready: podMap["ready"] == "true",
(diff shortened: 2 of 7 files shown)


pkg/drain/drain.go
```diff
@@ -153,13 +155,27 @@ func (d *Helper) GetPodsForDeletion(nodeName string) (*podDeleteList, []error) {
...
 			pod:    pod,
 			status: status,
 		})
 	}
 
+	//sort the list to come up with a sensible order for deleting pods
+	//for backwards compatibility, the default priority is highest.  If a daemon-set is found
+	//and ignore-daemonsets is set to false, the priority will be the lowest.
+	//This sort will return the order of pods to delete, with the highest priority first.
+	sort.SliceStable(pods, func(i, j int) bool {
```
johngmyers: `evictPods()` uses goroutines to evict all pods in parallel, so sorting the list has little effect. `DaemonSet` pods will still be evicted before the workloads that depend on them.

pkg/drain/drain_test.go
```diff
@@ -0,0 +1,186 @@
...
+func makePodList(pods []map[string]string) []runtime.Object {
+	var list []runtime.Object
+	for _, pod := range pods {
+		p := dummyPod(pod)
+		list = append(list, &p)
+	}
+	return list
+}
+
+func TestRollingUpdateDaemonSetMixedPods(t *testing.T) {
```
johngmyers: This is a test of the draining code, not of rolling update.

This only tests the ordering of the list returned by `GetPodsForDeletion`. It does not test anything about how that list is handled by `DeleteOrEvictPods`, which is why it failed to catch that the sorting was ineffective.

pkg/drain/drain.go
```diff
@@ -153,13 +155,27 @@ func (d *Helper) GetPodsForDeletion(nodeName string) (*podDeleteList, []error) {
 				// through any additional filters
 				break
 			}
+			if status.priority < priority {
+				//a previous filter had determined that this pod should be deleted at a lower priority
+				// use that priority
+				priority = status.priority
+			}
 		}
+		status.priority = priority
```
johngmyers: The logic here is a little convoluted. It takes the `podDeleteStatus` of the last-called filter, except replaces the `priority` field with the lowest priority. Except if the `podDeleteStatus` indicates non-deletion, the priority of that last call is never considered, even using `podDeletionPriorityHighest` if the first filter indicated non-deletion.

I don't think the code should be modifying the returned `podDeleteStatus` structs.

pkg/drain/filters.go
```diff
@@ -161,63 +170,64 @@ func (d *Helper) daemonSetFilter(pod corev1.Pod) podDeleteStatus {
...
 	}
 
 	if !d.IgnoreAllDaemonSets {
-		return makePodDeleteStatusWithError(daemonSetFatal)
+		//updating this to set the status as OK if we intend to delete daemon sets
+		return makePodDeleteStatusOkay(podDeletionPriorityLowest)
 	}
 
-	return makePodDeleteStatusWithWarning(false, daemonSetWarning)
+	return makePodDeleteStatusWithWarning(false, daemonSetWarning, podDeletionPriorityLowest)
```
johngmyers: Why should this and the previous be `Lowest`? Better would be `podDeletionPriorityDaemonset`.

pkg/instancegroups/rollingupdate_test.go
```diff
@@ -850,6 +850,15 @@ func TestRollingUpdateMaxUnavailableAllNeedUpdateMaster(t *testing.T) {
 	concurrentTest.AssertComplete()
 }
 
+func TestRollingUpdateAllNeedUpdateDoNotIgnoreDaemonsets(t *testing.T) {
+	c, cloud, cluster := getTestSetup()
+	c.IgnoreDaemonsets = false
+	groups := make(map[string]*cloudinstances.CloudInstanceGroup)
+	makeGroup(groups, c.K8sClient, cloud, "node-1", kopsapi.InstanceGroupRoleNode, 1, 1)
+	err := c.RollingUpdate(groups, cluster, &kopsapi.InstanceGroupList{})
+	assert.NoError(t, err, "rolling update")
```
johngmyers: This isn't testing a whole lot. The test would pass even if the `IgnoreDaemonsets` field was unused.

johngmyers:
Run `hack/update-bazel.sh` and commit the results in order to pass the `pull-kops-verify-bazel` and Travis checks.

mmerrill3:
@johngmyers thanks for all the feedback and your insight.  Before I go forward, do we want to address all the use cases where pods can get drained from a node, or just the rolling update feature of kops?  To confirm, I see three use cases.  One is a user initiated rolling update, via kops.  The next use case is node deletion via an ASG, or even via the cloud providers' console or a CLI.  Finally, there's node deletion for any other reason. 

Yes, this PR only addresses the first use case, which is the user initiated rolling update command.  

To address #7119, I'm not sure if the kubelet sends a delete node request to the API server in this case.  If so, maybe we could leverage that.  That's nice since the rolling-update also sends a delete node signal to the API server, and we can come up with one solution, based around that delete node API request.

mmerrill3:
no, kubelet does not send a delete node API request when the instances are terminated via the AWS console

mmerrill3:
I feel like we are at a design impasse here, since one use case is a CLI use case where we do a rolling update, and kops is in the flow.  The other use case is an ASG or another tool terminating an instance, and kops is not in the flow.  The solutions in #7119, in particular the last one where a systemd service is registered at shutdown, are not initiated by kops.

The rolling update can be controlled by kops, so that's what this PR is for.  I don't see a way, without introducing another component (operator, systemd unit, daemon set) of handling the "cloud" termination gracefully.  
Moreover, the solutions for #7119 won't help when we want to make sure that all pods are removed before we delete the node through the k8s API server during a rolling update.  For instance, if I setup a node with a systemd stop script to gracefully remove all the pods upon node termination, it will have ran too late.  The node will already have been deleted from the API server.  My goal is to gracefully remove all pods before the calls is made to the API server to delete the node.

Maybe that's the hang up.  During a rolling update, I care when the delete request is made to the API server.  In the other cases, the delete node API call doesn't get made, and its not relevant.

johngmyers:
According to the comment in the code, which was added in e982087e3ed5434bebe8989c42853146fbe58547, rolling update deletes the node because the cloud provider could create a replacement with the same name and that could get the same node object with the cordoning in place.

One possibility would be to move the node deletion to the systemd stop script, assuming we would have enough confidence that it would reliably run.

It seems odd to me that kubelet would reuse a node resource like this. One would think that the deletion-through-console case would also suffer from such stale data in reused node resources. Perhaps we could have kops-controller delete such orphaned nodes when the replacement instance goes through bootstrapping?

@justinsb what would you say to moving the node deletion that is in rolling update to someplace later? The assertion is that Weave uses the node deletion event as a signal to deallocate that node's IP address assignments, which is a problem if that node's Weave pod is still running and communicating with other Weave pods.

johngmyers:
@mmerrill3 to answer your first question, I would prefer we spend some time exploring whether we can find a systematic solution. If we can, then work done on rolling update would likely be wasted.

If there were to be more work on rolling update, it would probably be to add a second drain call  to evict the daemonsets. I'm not sure this would need to be an option.

mmerrill3:
Thanks @johngmyers, I like your idea that the node deletion doesn't need to occur during rolling update, but rather moving the node deletion action to a workflow that handles graceful node termination, like discussed in #7119.  That would allow us to do graceful terminations the same way, for both use cases.
Nodeup could add a systemd "shutdown" unit to be invoked when the instance is being shutdown.  We can make that configurable, default turned on, as a kops configuration general option.  That shutdown systemd service needs to be able to send API commands to the API server, so we would need to maybe do  something like protokube.   As opposed to being ran to setup volumes, DNS, etc, a new "postkube" container would be ran _after_ we are done with the node as a member of the cluster, and allow us to do node deletion, removing any remaining pods if need be, gracefully close mounts, or anything else really.

mmerrill3:
this PR is going to be closed in support of a different approach.

mmerrill3:
closing to come at the from a different angle.

PR #8629, opened 2020-02-25
Add option to use etcd managed by cilium-etcd-operator as kvstore

### LINKED ISSUE #8465: CNI Improvement: Cilium KV-Store Backed Cluster Support
As of Cilium 1.6 deployments of Cilium default to using the Kubernetes CRD mechanism for storing Cilium primitives, where it previously required it's own etcd or Consul storage engine.

The only option currently configured in the Cilium manifest template is CRD-backed storage. I'd like to propose the ability to support both mechanisms should the end user desire it.

The thinking would be to change the Cilium Networking Spec to include a few options to generate alternate manifests for Kops to deploy the KV-store backed Cilium agents.

Initial discussion was posed in [this ticket](https://github.com/kubernetes/kops/pull/8445)

olemarkus:
This PR adds the possibility to use etcd as kvstore for cilium agent state. See https://docs.cilium.io/en/v1.7/gettingstarted/k8s-install-etcd-operator/ for the details.

Fixes #8465

Note that coredns 1.6+ currently has a bug that prevents etcd-operator from creating clusters, so if one wants to test this, downgrade coredns first. See https://github.com/coredns/coredns/issues/3686

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
@@ -748,11 +748,6 @@ spec:
           readOnly: true
 {{- end }}
       hostNetwork: true
-{{- if .EtcdManaged }}
-      # In managed etcd mode, Cilium must be able to resolve the DNS name of
-      # the etcd service
-{{- end }}
-
       restartPolicy: Always
       serviceAccount: cilium-operator
       serviceAccountName: cilium-operator
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
@@ -580,7 +580,7 @@ spec:
             path: etcd.config
           name: cilium-config
         name: etcd-config-path
-        # To read the k8s etcd secrets in case the user might want to use TLS
+        # To read the Cilium etcd secrets in case the user might want to use TLS
       - name: etcd-secrets
         secret:
           defaultMode: 420
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
@@ -7,6 +7,21 @@ metadata:
     role.kubernetes.io/networking: "1"
 data:
 {{ with .Networking.Cilium }}
+
+{{- if .EtcdManaged }}
+  kvstore: etcd
+  kvstore-opt: '{"etcd.config": "/var/lib/etcd-config/etcd.config", "etcd.operator": "true"}'
+
+  etcd-config: |-
+    ---
+    endpoints:
+      - https://cilium-etcd-client.kube-system.svc:2379
+
+    trusted-ca-file: '/var/lib/etcd-secrets/etcd-client-ca.crt'
+    key-file: '/var/lib/etcd-secrets/etcd-client.key'
+    cert-file: '/var/lib/etcd-secrets/etcd-client.crt'
+{{ end }}
+
   # Identity allocation mode selects how identities are shared between cilium
   # nodes by setting how they are stored. The options are "crd" or "kvstore".
   # - "crd" stores identities in kubernetes as CRDs (custom resource definition).
@@ -460,6 +475,14 @@ spec:
           name: cni-path
         - mountPath: /host/etc/cni/net.d
           name: etc-cni-netd
+{{ if .EtcdManaged }}
+        - mountPath: /var/lib/etcd-config
+          name: etcd-config-path
+          readOnly: true
+        - mountPath: /var/lib/etcd-secrets
+          name: etcd-secrets
+          readOnly: true
+{{ end }}
         - mountPath: /var/lib/cilium/clustermesh
           name: clustermesh-secrets
           readOnly: true
@@ -548,6 +571,22 @@ spec:
           type: FileOrCreate
         name: xtables-lock
         # To read the clustermesh configuration
+{{- if .Networking.Cilium.EtcdManaged }}
+        # To read the etcd config stored in config maps
+      - configMap:
+          defaultMode: 420
+          items:
+          - key: etcd-config
+            path: etcd.config
+          name: cilium-config
+        name: etcd-config-path
+        # To read the k8s etcd secrets in case the user might want to use TLS
+      - name: etcd-secrets
+        secret:
+          defaultMode: 420
+          optional: true
+          secretName: cilium-etcd-secrets
+{{- end }}
       - name: clustermesh-secrets
         secret:
           defaultMode: 420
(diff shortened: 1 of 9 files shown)


upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -698,10 +737,42 @@ spec:
...
+          name: etcd-config-path
+          readOnly: true
+        - mountPath: /var/lib/etcd-secrets
+          name: etcd-secrets
+          readOnly: true
+{{- end }}
       hostNetwork: true
+{{- if .EtcdManaged }}
+      # In managed etcd mode, Cilium must be able to resolve the DNS name of
+      # the etcd service
```
johngmyers: Did you forget to set the `dnsPolicy` here?

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -717,4 +788,242 @@ spec:
...
+              apiVersion: v1
+              fieldPath: metadata.uid
+        - name: CILIUM_ETCD_META_ETCD_AUTO_COMPACTION_MODE
+          value: "revision"
+        - name: CILIUM_ETCD_META_ETCD_AUTO_COMPACTION_RETENTION
+          value: "25000"
+        image: "cilium/cilium-etcd-operator:v2.0.7"
+        name: cilium-etcd-operator
+      dnsPolicy: ClusterFirst
+      hostNetwork: true
```
johngmyers: priorityClassName: system-cluster-critical

johngmyers:
Does cilium-etcd-operator run the etcd pods with host network? We forked etcd-operator to include coreos/etcd-operator#2094 and suspect we'll have to fork cilium-etcd-operator too.

Also, does cilium-etcd-operator run the etcd pods with the system-cluster-critical priority class?

Since etcd-operator appears to be unmaintained, I'm wondering if it would be practical to add a Cilium mode to etcd-manager.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
olemarkus: Not 100% sure this should qualify for system-cluster-critical. It technically isn't very critical even though the pods it spawns is critical.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
olemarkus: No, I removed that part on purpose. Changing the DNS policy breaks cilium because it can no longer look up the internal cluster DNS (since we use the public DNS entry for that). The change on line 14 (` "etcd.operator": "true"`) removes the need to change the DNS policy.
That particular change hasn't made it into the cilium helm templates yet, if you are comparing to what that one outputs.

johngmyers: You have a conditional wrapped around nothing but comments here. Perhaps remove the conditional and comments as well?

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
johngmyers: If it gets preempted then it can't repair any problems with the Cilium etcds, such as their losing quorum.

olemarkus:
> Does cilium-etcd-operator run the etcd pods with host network? We forked etcd-operator to include coreos/etcd-operator#2094 and suspect we'll have to fork cilium-etcd-operator too.

No, it doens't. It is unclear to me why that would be necessary.

> 
> Also, does cilium-etcd-operator run the etcd pods with the system-cluster-critical priority class?

No, I and I am not sure it is correct to do so either. The consequence of losing etcd is that state propagation will be slower, but it won't lead to immediate critical failure

> Since etcd-operator appears to be unmaintained, I'm wondering if it would be practical to add a Cilium mode to etcd-manager.

That could be interesting. Everything that talks to etcd uses hostNetworking, so etcd-manager's etcd should be reachable too. It would certainly the concerns above. Would be a pretty big change though.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
olemarkus: A thanks. I was certain I removed that one. :man_facepalming:

johngmyers:
With hundreds of nodes, "slower" may well be "completely unstable". If someone doesn't need etcd for state propagation it's likely they wouldn't enable it.

I suppose in the crd identity allocation mode an agent could bootstrap off of the apiserver, so running etcd on host network wouldn't be as important as for kvstore mode. It might even be able to do so without killing the apiserver when in a large cluster.

johngmyers:
Shouldn't we bump the version in bootstrapchannelbuilder.go?

olemarkus:
Cilium's stance on this is still that larger environments should use external etcd (https://docs.cilium.io/en/v1.7/gettingstarted/k8s-install-external-etcd/)
I think etcd-manager would qualify as that better be pretty damn stable.

Forking cilium-etcd-operator/etcd-operator etc isn't an option for me at least.

olemarkus:
> Shouldn't we bump the version in bootstrapchannelbuilder.go?

Yep. bumped.

johngmyers:
External etcd would be expensive to manage. We run in-cluster etcd on host network with forked etcd-operator. When we are able to upgrade to 1.6 we'll look at crd identity allocation mode with etcd for propagation, assuming we're willing to stay under 250 nodes.

Elsewhere they've mentioned that pure crd is only good for up to 50 nodes.

Host network is not a blocker for this PR. I'm trying to decide whether I want to hold out for priorityClassName on cilium-etcd-operator. If cilium-etcd-operator happens to not put them on the etcd pods then it'd be a moot point unless/until that could be fixed upstream.

olemarkus:
I think we should push this in without and then I can see if I can push cilium to propagate the priority class.

I agree that external etcd is too expensive, but I want to see how hard it is to add another etcd-manager cluster. That will have to come later though

johngmyers:
It looks like cilium/cilium-etcd-operator#67 is going to be a significant defect now that kops rolling update can drain nodes in parallel.

olemarkus:
Yep. I think I am going to have a look at etcd-manager sooner than I thought.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -699,10 +738,37 @@ spec:
...
+      volumes:
+      # To read the etcd config stored in config maps
+      - configMap:
+          defaultMode: 420
+          items:
+          - key: etcd-config
+            path: etcd.config
+          name: cilium-config
+        name: etcd-config-path
+        # To read the k8s etcd secrets in case the user might want to use TLS
```
johngmyers: These are the Cilium etcd secrets, not the k8s ones, right?

olemarkus: Yes. etcd-operator creates the secret `cilium-etcd-secret`.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -548,6 +571,22 @@ spec:
...
+{{- if .Networking.Cilium.EtcdManaged }}
+        # To read the etcd config stored in config maps
+      - configMap:
+          defaultMode: 420
+          items:
+          - key: etcd-config
+            path: etcd.config
+          name: cilium-config
+        name: etcd-config-path
+        # To read the k8s etcd secrets in case the user might want to use TLS
```
johngmyers:
```suggestion
        # To read the Cilium etcd secrets in case the user might want to use TLS
```

PR #8632, opened 2020-02-26
Promote alphas to stable

olemarkus:
These have been in alpha for two weeks and should be good for stable now.


justinsb:
Thanks!

/approve
/lgtm

PR #8020, opened 2019-11-27
Automatically install dependencies with local packages

hakman:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it?**
Changes the way local packages are installed from using the low level dpkg and rpm tools to their higher level alternatives apt and and yum.
This reduces the work needed to maintain the dependencies by letting the package manager take care of this task. Optional dependencies can still be provided if needed.

**Which issue(s) this PR fixes?**
May be a solution for #7878

**Special notes for your reviewer**
At the moment Kops relies on a list of dependencies for each local package that is installed.
Providing a list of dependencies is prone to errors over time and may not know exactly why each dep was added.
Dependencies are only checked if they are installed, not also if the version is the minimum required. This can lead to failure to install in some cases as discussed in the refferenced issue.
The update script would upgrade the packages anyway later, but I guess it's probably best to start with the freshes dependencies.

**Does this PR introduce a user-facing change?**
No

PR #8635, opened 2020-02-27
Install packages updates during bootstrap

hakman:
During bootstrap packages that are already installed are skipped. They are later installed by the automatic updater.

The PR checks if automatic updates are enabled and tries to install the updated version.
The package managed decides if there is an update to be installed or not, so the only thing that has to be done is try to install it again. The newer the base image, the less updates will be installed.

This is a followup of #8020 and after discussing about the issue during office hours.

### DIFF
--- a/upup/pkg/fi/nodeup/tags/tags.go
+++ b/upup/pkg/fi/nodeup/tags/tags.go
@@ -25,6 +25,12 @@ const (
 	TagOSRHEL8        = "_rhel8"
 
 	TagSystemd = "_systemd"
+
+	// Nodes with the "_automatic_upgrade" tag automatically update installed packages
+	// during bootstrapping and daily for security updates (unless this update would require
+	// a node reboot). To disable automatic node package updates, set:
+	// `Cluster.Spec.UpdatePolicy = external`
+	TagUpdatePolicyAuto = "_automatic_upgrades"
 )
 
 type HasTags interface {
--- a/docs/getting_started/arguments.md
+++ b/docs/getting_started/arguments.md
@@ -62,6 +62,8 @@ Values:
 
 * unset means to use the default policy, which is currently to apply OS security updates unless they require a reboot
 
+Required packages are also updated during bootstrapping if the value is not set. 
+
 ## out
 
 `out` determines the directory into which Kops will write the target output for Terraform and CloudFormation.  It defaults to `out/terraform` and `out/cloudformation` respectively.
--- a/upup/pkg/fi/nodeup/nodetasks/package.go
+++ b/upup/pkg/fi/nodeup/nodetasks/package.go
@@ -216,7 +216,9 @@ func (e *Package) findDpkg(c *fi.Context) (*Package, error) {
 		}
 	}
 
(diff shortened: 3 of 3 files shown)


hakman:
Kops has a feature that does automatic updates on each node. This is enabled by default and can be disabled by setting `Cluster.Spec.UpdatePolicy` to `external`:
There is some documentation for the option here:
https://github.com/kubernetes/kops/blob/master/docs/getting_started/arguments.md#updatepolicy

The `TagUpdatePolicyAuto` is an alias for `_automatic_upgrades`, which is added only when no UpdatePolicy is set:
https://github.com/kubernetes/kops/blob/77d6d381c30ec4e6d1b8e8e9563017b39a76b7e7/upup/pkg/fi/cloudup/tagbuilder.go#L90-L92

Instead of waiting for the updater to run (daily), I am trying to update the packages required for Kubernetes during bootstrap.

We discussed this during office hours about a month ago and decided that should be ok to update packages during bootstrap, as long as automatic updates are not enabled.

geojaz:
ahh! ok, thanks for connecting the dots for me- It seems that I keep asking for clarification on things that were discussed during office hours... I'll try to do better at keeping up with these :roll_eyes: 

I would still love a quick comment (in tags.go) that reminds people who are scanning the code of the implications of `_automatic_upgrades`. Something simple like: `Nodes with this tag attempt to update packages during bootstrapping instead of daily . See: ... for other options.` is that overkill? 
I'll take another look in the US daytime :)

hakman:
I am sure that you have better things to do than remember all the tiny nits from 6 weeks ago. :)
Adding comments sounds very reasonable. Will add some and continue this later.
Thanks!

hakman:
@geojaz I added the comment and an extra note in docs. I think it's ok for another look. Thanks!

upup/pkg/fi/nodeup/tags/tags.go
```diff
@@ -25,6 +25,11 @@ const (
 	TagOSRHEL8        = "_rhel8"
 
 	TagSystemd = "_systemd"
+
+	// Nodes with the "_automatic_upgrade" tag will update packages for security,
+	// but also during bootstrapping. To disable automatic updates set
+	// `Cluster.Spec.UpdatePolicy = external`
```
geojaz:
```suggestion

	// Nodes with the "_automatic_upgrade" tag automatically update packages 
	// during bootstrapping and for security updates (unless this update would require
	// a node reboot. To disable automatic node package updates, set:
	// `Cluster.Spec.UpdatePolicy = external`
```

hakman:
Can't imagine how I could have written such nonsense. Even I don't understand it anymore. :))

hakman:
Thanks @geojaz!
Btw, can you also add the approved label?

geojaz:
Teamwork! I'm just so glad that I don't have to write docs/comments in (your native language- which I don't want to guess). Ahh... I understand why you need an approve as well as a lgtm, but I don't. It's because it automatically adds approved to mine.... not fair.
/approve

PR #8639, opened 2020-02-27
Delete launch templates when using EnableLaunchTemplates

hakman:
Setting the feature flag `EnableLaunchTemplates=true` make kops use Launch Templates instead of Launch Configurations to create the ASGs.

Launch Templates implementation in Kops is derived from the Mixed Instance Policy feature, but requires a few tweaks. One of it is that security groups are declared for the primary network interface instead of the SecurityGroups field. For this reason, the delete mechanism is not finding these launch templates.

### DIFF
--- a/pkg/resources/aws/aws.go
+++ b/pkg/resources/aws/aws.go
@@ -1204,7 +1204,10 @@ func FindAutoScalingLaunchTemplateConfigurations(cloud fi.Cloud, securityGroups 
 		}
 		for _, j := range req.LaunchTemplateVersions {
 			// @check if the security group references the security group above
-			for _, y := range j.LaunchTemplateData.SecurityGroupIds {
+			var s []*string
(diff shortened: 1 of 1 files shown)


johngmyers:
The other thing this does is change from comparing against the `SecurityGroupIds` field to the `SecurityGroups` field.

hakman:
> The other thing this does is change from comparing against the `SecurityGroupIds` field to the `SecurityGroups` field.

Thanks for noticing the typo. I just pushed a fix.

rifelpet:
I created a new E2E job for launch templates: https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-launchtemplates

And confirmed that the LT is not getting deleted on cluster teardown: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-aws-misc-launchtemplates/1233770692407726083#1:build-log.txt%3A69891

so we can use this job to confirm this PR fixes the problem.

PR #8640, opened 2020-02-27
Add UT for util/pkg/proxy/proxy.go

tiendc:
Signed-off-by: Dao Cong Tien <tiendc@vn.fujitsu.com>

### DIFF
--- a/util/pkg/proxy/proxy_test.go
+++ b/util/pkg/proxy/proxy_test.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


util/pkg/proxy/proxy_test.go
```diff
@@ -0,0 +1,89 @@
...
+			expected:  []v1.EnvVar{},
+		},
+		{
+			inProxies: &kops.EgressProxySpec{
+				HTTPProxy: kops.HTTPProxy{
+					Host: "",
+					Port: 1234,
+				},
+			},
+			// TODO: The output seems a little bit weird for this case
```
justinsb: We might want to catch it in validation, but this seems a reasonable translation at this layer!

justinsb:
Thanks @tiendc - looks great!

I do wonder whether people expect https_proxy to use https, but this is a good test that shows the current state of the mapping.

/approve
/lgtm

PR #7003, opened 2019-05-16
Simplify `go test` command

justinsb:
We shouldn't have to explicitly specify packages any more (if we do,
we should exclude them individually).

PR #8647, opened 2020-02-28
Remove hack/.packages

rifelpet:
This [was introduced](https://github.com/kubernetes/kops/pull/3400/files) to speed up `go list` when passing packages to `go test` but [was removed](https://github.com/kubernetes/kops/pull/7003/files) when `go test` began ignoring the vendor directory by default.

This was the only reference I could find in the repo to `hack/.packages` so I think it is safe to remove. We'll need to remove the verify-packages prow job as well, in fact i may need to close this PR and reopen it once the job is removed. We'll see.


rifelpet:
Prow job removal: https://github.com/kubernetes/test-infra/pull/16545

rifelpet:
nevermind, this is still used by the goimports hack scripts as well. i suppose we could use `go list` for those too but i dont know if its worth the effort.

/close

PR #8649, opened 2020-02-28
Fix node-authorizer/pkg/server staticcheck failure

hakman:
Finally fixes the last exception: #7800. Thanks for doing this @tanjunchen.

```
Errors from staticcheck:
node-authorizer/pkg/server/server.go:122:16: the channel used with signal.Notify should be buffered (SA1017)
```

### DIFF
--- a/hack/.staticcheck_failures
+++ b/hack/.staticcheck_failures
@@ -1,1 +1,0 @@
-node-authorizer/pkg/server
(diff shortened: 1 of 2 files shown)


rifelpet:
Nice! We can probably remove the failures file altogether and it’s usage within the verify script
/approve

PR #8655, opened 2020-02-28
When nodeport is enabled, use strict kube-proxy-replacement to ensure…

olemarkus:
… cilium fully replaces kube-proxy

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
@@ -121,6 +121,7 @@ data:
   install-iptables-rules: "{{- if .IPTablesRulesNoinstall -}}false{{- else -}}true{{- end -}}"
   auto-direct-node-routes: "{{- if .AutoDirectNodeRoutes -}}true{{- else -}}false{{- end -}}"
   enable-node-port: "{{- if .EnableNodePort -}}true{{- else -}}false{{- end -}}"
+  kube-proxy-replacement: "{{- if .EnableNodePort -}}strict{{- else -}}disabled{{- end -}}"
   {{ with .Ipam }}
   ipam: {{ . }}
   {{ if eq . "eni" }}
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -453,7 +453,7 @@ func validateNetworkingCanal(v *kops.CanalNetworkingSpec, fldPath *field.Path) f
 func validateNetworkingCilium(c *kops.ClusterSpec, v *kops.CiliumNetworkingSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	if v.EnableNodePort && c.KubeProxy != nil && *c.KubeProxy.Enabled {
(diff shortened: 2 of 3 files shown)


olemarkus:
Without this, service forwarding fails on newer Cilium versions.

/bug

diegocn:
- Tested @olemarkus's branch and I was able to spin up a running cluster with cilium 1.7.0 and without kube-proxy. 
- I can see some logs on cilium-agents related to some sort of failure on BFP probe even with the host/node AMI with BPF enabled and mounted:
`level=warning msg="BPF system config check: NOT OK." error="CONFIG_BPF kernel parameter is required" subsys=linux-datapath`
- Before this patch kops-controller was not able to reach kubernetes-api. Now it's running as expected.

olemarkus:
See https://github.com/cilium/cilium/issues/10256 regarding the config check error. It is harmless

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -121,6 +121,9 @@ data:
   install-iptables-rules: "{{- if .IPTablesRulesNoinstall -}}false{{- else -}}true{{- end -}}"
   auto-direct-node-routes: "{{- if .AutoDirectNodeRoutes -}}true{{- else -}}false{{- end -}}"
   enable-node-port: "{{- if .EnableNodePort -}}true{{- else -}}false{{- end -}}"
+  {{ if .EnableNodePort }}
+  kube-proxy-replacement: strict
+  {{ end }}
```
hakman: Should be easier to deal with the extra white space:
```suggestion
  kube-proxy-replacement: "{{- if .EnableNodePort -}}strict{{- else -}}probe{{- end -}}"
```

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.7.yaml.template
```diff
@@ -121,6 +121,9 @@ data:
   install-iptables-rules: "{{- if .IPTablesRulesNoinstall -}}false{{- else -}}true{{- end -}}"
   auto-direct-node-routes: "{{- if .AutoDirectNodeRoutes -}}true{{- else -}}false{{- end -}}"
   enable-node-port: "{{- if .EnableNodePort -}}true{{- else -}}false{{- end -}}"
+  {{ if .EnableNodePort }}
+  kube-proxy-replacement: strict
+  {{ end }}
```
hakman:
```suggestion
  kube-proxy-replacement: "{{- if .EnableNodePort -}}strict{{- else -}}probe{{- end -}}"
```

olemarkus: I would perhaps go for `kube-proxy-replacement: disabled`. it is a bit easier to understand what is happening if one is explicit.

hakman: Sounds even better to me.

johngmyers:
@olemarkus This looks like the prime suspect in breaking the [kops-aws-cni-cilium](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-cni-cilium) e2e test. Could you take a look?

olemarkus:
yep

PR #8660, opened 2020-03-01
Add "owned" tag for volumes and instances created with launch templates

hakman:
Using launch templates for instance groups make adds the kops & k8s tags to instances and root volumes.

For volumes, the cluster delete function checks if the resource is "shared" or "owned", otherwise it displays the following warning:
```
W0229 23:33:33.176831    8549 aws.go:2148] (new) cluster tag not found on volume:vol-038ab86f6fcd93336
```

https://github.com/kubernetes/kops/blob/632b426bdfc2bd3566c6bc5a690fa1c4022d736d/pkg/resources/aws/aws.go#L2134-L2136

### DIFF
--- a/pkg/model/context.go
+++ b/pkg/model/context.go
@@ -183,7 +183,7 @@ func (m *KopsModelContext) NodeInstanceGroups() []*kops.InstanceGroup {
 
 // CloudTagsForInstanceGroup computes the tags to apply to instances in the specified InstanceGroup
 func (m *KopsModelContext) CloudTagsForInstanceGroup(ig *kops.InstanceGroup) (map[string]string, error) {
-	labels := make(map[string]string)
+	labels := m.CloudTags(m.AutoscalingGroupName(ig), false)
 
 	// Apply any user-specified global labels first so they can be overridden by IG-specific labels
 	for k, v := range m.Cluster.Spec.CloudLabels {
@@ -207,10 +207,6 @@ func (m *KopsModelContext) CloudTagsForInstanceGroup(ig *kops.InstanceGroup) (ma
 			labels[clusterAutoscalerNodeTemplateTaint+splits[0]] = splits[1]
 		}
 	}
-
-	// Add cluster and ig names
-	labels[awsup.TagClusterName] = m.ClusterName()
-	labels["Name"] = m.AutoscalingGroupName(ig)
 
 	// The system tags take priority because the cluster likely breaks without them...
 
--- a/tests/integration/update_cluster/additional_cidr/cloudformation.json
+++ b/tests/integration/update_cluster/additional_cidr/cloudformation.json
@@ -34,6 +34,11 @@
             "Key": "kops.k8s.io/instancegroup",
             "Value": "master-us-test-1b",
             "PropagateAtLaunch": true
+          },
+          {
+            "Key": "kubernetes.io/cluster/additionalcidr.example.com",
+            "Value": "owned",
+            "PropagateAtLaunch": true
           }
         ],
         "MetricsCollection": [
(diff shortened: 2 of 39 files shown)


johngmyers:
> For volumes, the cluster delete function checks if the cluster is "shared" or "owned", otherwise

Perhaps you meant that it checks if the resource is "shared" or "owned"?

hakman:
> Perhaps you meant that it checks if the resource is "shared" or "owned"?

Yes. Updated. Thanks.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -101,6 +101,8 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 	if err != nil {
 		return nil, fmt.Errorf("error building cloud tags: %v", err)
 	}
+	// TODO: move this to CloudTagsForInstanceGroup when LaunchConfigurations are removed
+	tags["kubernetes.io/cluster/"+b.ClusterName()] = "owned"
```
johngmyers: Perhaps call `b.CloudTags("", false)` ?

What would be the downside to moving that to CloudTagsFroInstanceGroup before LaunchConfigurations are removed? Why don't instances/volumes created by LaunchConfigurations throw the warning?

hakman: > What would be the downside to moving that to CloudTagsFroInstanceGroup before LaunchConfigurations are removed?

I don't know if it would affect any current workflow if this tag is applied at ASG level. Launch templates are still experimental in Kops and shouldn't be an issue.

> Why don't instances/volumes created by LaunchConfigurations throw the warning?

Instances created with launch configurations don't tag root volumes at all. Because of this, the volumes are not tracked as part of the cluster by the delete function. They are still deleted automatically when the instance is deleted.

May also be a possibility to not track root volumes at all with the deleted function, even if they have the other tags.

johngmyers: I like the idea of being consistent in what tags we apply.

I could see the `KubernetesCluster` tag being useful for cost control and leaving it out lacks consistency, so I think we should call `CloudTags` rather than setting `kubernetes.io/cluster/` directly.

For consistency, I'm in favor of moving things into `CloudTagsForInstanceGroup` sooner. Perhaps we should have `CloudTagsForInstanceGroup` call `CloudTags`. I don't know what the consequence of adding them to the template functions would be, though.

hakman: I actually tried this as the first iteration, but the changes in tests were bigger than I expected.
I don't think should affect any workflow, but preferred not to change so much.

pkg/model/context.go
```diff
@@ -183,7 +183,7 @@ func (m *KopsModelContext) NodeInstanceGroups() []*kops.InstanceGroup {
 
 // CloudTagsForInstanceGroup computes the tags to apply to instances in the specified InstanceGroup
 func (m *KopsModelContext) CloudTagsForInstanceGroup(ig *kops.InstanceGroup) (map[string]string, error) {
-	labels := make(map[string]string)
+	labels := m.CloudTags(ig.Name, false)
```
johngmyers: Line 212 is now redundant. Should pass `m.AutoscalingGroupName(ig)` instead of `ig.Name` to make line 213 redundant.

johngmyers: I'm trying to figure out how the `Name` tag is being set correctly without using `m.AutoscalingGroupName(ig)` here.

hakman: OK, changed to m.AutoscalingGroupName(ig) to not complicate things.

hakman:
The warnings are gone:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-aws-misc-launchtemplates/1234559136599904258#1:build-log.txt%3A69940

PR #8666, opened 2020-03-03
Fix SecurityGroup check when there are no Network Interfaces associated with a LaunchTemplate

KashifSaadat:


KashifSaadat:
On deleting a cluster the following runtime error occurs:
```
$ kops delete cluster --yes

panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
k8s.io/kops/pkg/resources/aws.FindAutoScalingLaunchTemplateConfigurations(0x7f7bcdc5c178, 0xc0001f91f0, 0xc00045df80, 0x7f7bcdc5c178, 0xc0001f91f0, 0x4, 0x0, 0x0)
	/go/src/k8s.io/kops/pkg/resources/aws/aws.go:1208 +0x92c
k8s.io/kops/pkg/resources/aws.ListResourcesAWS(0x476e440, 0xc0001f91f0, 0xc000052052, 0x25, 0xc0001f91f0, 0x3472200, 0xc000827928)
	/go/src/k8s.io/kops/pkg/resources/aws/aws.go:149 +0xab6
k8s.io/kops/pkg/resources/ops.ListResources(0x7f7bcdc5c178, 0xc0001f91f0, 0xc000052052, 0x25, 0x0, 0x0, 0x0, 0x10, 0x3472200)
	/go/src/k8s.io/kops/pkg/resources/ops/collector.go:41 +0x2d8
main.RunDeleteCluster(0xc00071aa20, 0x4691460, 0xc00000e018, 0xc000429440, 0x0, 0x0)
	/go/src/k8s.io/kops/cmd/kops/delete_cluster.go:133 +0x5e7
main.NewCmdDeleteCluster.func1(0xc000280a00, 0xc0000c3ae0, 0x0, 0x1)
	/go/src/k8s.io/kops/cmd/kops/delete_cluster.go:80 +0xcd
github.com/spf13/cobra.(*Command).execute(0xc000280a00, 0xc0000c3ac0, 0x1, 0x1, 0xc000280a00, 0xc0000c3ac0)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:830 +0x2aa
github.com/spf13/cobra.(*Command).ExecuteC(0x67332a0, 0x676dda8, 0x0, 0x0)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fb
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.Execute()
	/go/src/k8s.io/kops/cmd/kops/root.go:95 +0x8f
main.main()
	/go/src/k8s.io/kops/cmd/kops/main.go:25 +0x20
```

Looks like this may have been introduced in PR #8639

rifelpet:
@KashifSaadat can you explain how a LaunchTemplate would not have Network Interfaces? Just curious

hakman:
@KashifSaadat any special reason there are 2 ways of specifying the security groups here?
https://github.com/kubernetes/kops/blob/0aa97af94e2e34f4fb55df05fab8518104c98a03/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go#L100-L110

For example, for terraform we only use the `NetworkInterfaces` way.
https://github.com/kubernetes/kops/blob/298f79659a2bd9c1d6548c6c95b335b5e29e6466/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go#L158-L168

KashifSaadat:
@rifelpet : If there are no customised settings to be provided (such as `AssociatePublicIP`) then a `NetworkInterfaceSpecificationRequest` isn't created as there is no need, and so it falls back to the default, i.e. `Currently no network interface details are specified and therefore the instance will launch with the template default network interface settings`

@hakman I'm not sure if there was a specific reason for that approach, maybe just taking the stance that if there's no explicit need to define a NetworkInterface template (other than for security groups) then to avoid it to not complicate the configuration? If there are no issues either way, then that logic can be simplified and made consistent with how it's generated in the terraform output.

hakman:
Thanks @KashifSaadat. I am trying to add LaunchTemplate tags to CF and TF and noticed that there are quite a few inconsistencies. I will try to make things more uniform in a future PR.

PR #8667, opened 2020-03-03
Add more API validation for Cilium

johngmyers:


johngmyers:
@olemarkus please review

johngmyers:
@olemarkus please review

pkg/apis/kops/validation/validation.go
```diff
@@ -457,12 +457,32 @@ func validateNetworkingCilium(c *kops.ClusterSpec, v *kops.CiliumNetworkingSpec,
...
+	if v.Tunnel != "" {
+		allErrs = append(allErrs, IsValidValue(fldPath.Child("tunnel"), &v.Tunnel, []string{"vxlan", "geneve", "disabled"})...)
+	}
+
+	if v.MonitorAggregation != "" {
+		allErrs = append(allErrs, IsValidValue(fldPath.Child("monitorAggregation"), &v.MonitorAggregation, []string{"low", "medium", "maximum"})...)
+	}
+
+	if v.ContainerRuntimeLabels != "" {
+		allErrs = append(allErrs, IsValidValue(fldPath.Child("containerRuntimeLabels"), &v.ContainerRuntimeLabels, []string{"none", "containerd", "crio", "docker", "auto"})...)
```
olemarkus: Should probably warn that this is deprecated when cilium version > 1.7.0 and that setting this has no effect. The field may be removed in the future.

pkg/apis/kops/validation/validation.go
```diff
@@ -457,12 +457,32 @@ func validateNetworkingCilium(c *kops.ClusterSpec, v *kops.CiliumNetworkingSpec,
...
+	}
+
+	if v.Ipam != "" {
+		allErrs = append(allErrs, IsValidValue(fldPath.Child("ipam"), &v.Ipam, []string{"crd", "eni", "azure"})...)
```
olemarkus: We don't actually support azure, so I think perhaps remove that one. CRD mode should work just fine.

olemarkus: Should these strings be constants similar to `kops.CiliumIpamEni`?

pkg/apis/kops/validation/validation.go
johngmyers: There is no mechanism for warnings that I am aware of.

pkg/apis/kops/validation/validation.go
johngmyers: If this would be the only reference to it, I think a constant would be overkill.

pkg/apis/kops/validation/validation.go
olemarkus: I wonder if you can just print. `upup/pkg/fi/cloudup/apply_cluster.go` does that in a couple of places and validation is called from there. Or that particular validation logic could be done there (seems unclean to me though)

johngmyers: Seems hokey to me. Wouldn't work well with `kops edit cluster`. Since it would only be ignored, a warning would be low-value.

olemarkus: It would be shown on `kops update cluster` too, I think. But no problem to leave it out

PR #8670, opened 2020-03-04
Revert AWS IAM Authenticator upgrade to 0.5.0 on master

rifelpet:
After thinking through the solutions discussed during last week's office hours, I think the best and cleanest option will be to push for a 0.5.1 to fixes the regression altogether. This way we don't need to have special logic for automatically adding command arguments, or warn users to specify additional api fields when upgrading kops.

More discussion is happening [here](https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/288) and I'll see if i can open a bugfix PR upstream that could land in a 0.5.1 soon.


joshbranham:
Seems good to me as well to reduce confusion in the meantime

/lgtm

PR #8674, opened 2020-03-05
Default maxSurge to 1 on AWS

johngmyers:
On the cloud provider that supports it, a `maxSurge` of 1 (and thus `maxUnavailable` of 0) is a better default than `maxSurge` of 0 and `maxUnavailable` of 1. Nodes will still be drained one at a time, but the pods evicted from the first node won't have to be replaced on an old node and get evicted again.

The only downside I can think of would require a rather contrived situation: a workload that is deployed with a DaemonSet instead of a Deployment or StatefulSet and which cannot tolerate having the resulting surged pod.

/area rolling-update


rifelpet:
that makes sense to me 👍 
/lgtm
/approve

PR #8686, opened 2020-03-06
Add UT for util/pkg/vfs/fs.go

tiendc:
Signed-off-by: Dao Cong Tien <tiendc@vn.fujitsu.com>

### DIFF
--- a/util/pkg/vfs/fs_test.go
+++ b/util/pkg/vfs/fs_test.go
@@ -25,20 +25,18 @@ import (
 	"testing"
 )
 
-var TempDir, _ = ioutil.TempDir("", "test")
-
 func TestCreateFile(t *testing.T) {
+	var TempDir, _ = ioutil.TempDir("", "test")
+	defer os.Remove(TempDir)
 	tests := []struct {
-		path       string
-		fileExists bool
-		data       []byte
+		path string
+		data []byte
 	}{
 		{
 			path: path.Join(TempDir, "SubDir", "test1.tmp"),
 			data: []byte("test data\nline 1\r\nline 2"),
 		},
 	}
-	defer os.Remove(TempDir)
 	for _, test := range tests {
 		fspath := &FSPath{test.path}
 		// Create file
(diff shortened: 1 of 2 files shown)


util/pkg/vfs/fs_test.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path"
+	"reflect"
+	"testing"
+)
+
+var TempDir, _ = ioutil.TempDir("", "test")
```
justinsb: Should we move this inside the test function, rather than have it as a global?  (In general we try to avoid globals, and this would better match the lifetime anyway, as we do `os.Remove` the dir in the test function)

justinsb:
Thanks for this - great to see more testing!

I think we should avoid that global variable, and then I had a few suggestions but those are more tips for the future!

util/pkg/vfs/fs_test.go
```diff
@@ -0,0 +1,65 @@
...
+			data: []byte("test data\nline 1\r\nline 2"),
+		},
+	}
+	defer os.Remove(TempDir)
+	for _, test := range tests {
+		fspath := &FSPath{test.path}
+		// Create file
+		err := fspath.CreateFile(bytes.NewReader(test.data), nil)
+		if err != nil {
+			t.Errorf("Error writing file %s, error: %v", test.path, err)
```
justinsb: Nit / tip: t.Fatalf will exit the test, meaning that we won't keep going - we certainly wouldn't expect the read to succeed if the write failed, and it can be confusing to see multiple failures.

util/pkg/vfs/fs_test.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != os.ErrExist {
+			t.Errorf("Expected to get os.ErrExist, got: %v", err)
+		}
+
+		// Check file content
+		data, err := fspath.ReadFile()
+		if err != nil {
+			t.Errorf("Error reading file %s, error: %v", test.path, err)
+		}
+		if !reflect.DeepEqual(data, test.data) {
```
justinsb: Nit / tip: you could also use `bytes.Equal`, I believe

PR #8693, opened 2020-03-08
Add root volume tags for CF and TF targets

hakman:
Adding root volume tags has been requested for a long time #3358 and was implemented in #8462 and #8466, but only for the `direct` target.

This PR adds root volume tags for `terraform` and `cloudformation` targets when using launch templates or mixed instances policies.

Notes:
* launch templates were not working at all
* an autoscaling group should [use only one](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html) of launch configuration, launch template or mixed instances policy
* security groups should be specified for each network interface, instead of globally (global was meant for EC2 Classic and Default VPC)
* to enable the launch templates feature, the `EnableLaunchTemplates` [feature flag](https://github.com/kubernetes/kops/blob/master/docs/advanced/experimental.md) must be specified

Issue #8648, opened 2020-02-28
Add terraform testing


rifelpet:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
Currently Kops has integration tests that test the cluster spec -> terraform file flow. It would be great if we could take that a step further and run terraform commands against the generated tf files to ensure they are valid.

It wouldn't necessarily have to be an `apply`, perhaps just a `terraform init && terraform validate` since those don't require cloud provider credentials unlike `plan` and `apply`, but `validate` will still catch syntax and schema errors.
 
**2. Feel free to provide a design supporting your feature request.**

We'll need a way to portably run terraform. I propose we use the [official dockerhub image](https://hub.docker.com/r/hashicorp/terraform/) and a volume mount. We'll also need a way to run these tests since they wouldnt be triggered by a `go test`. I'm guessing bazel can do it but I'm not familiar enough with bazel to know for sure.

We'll need to use a combination of terraform 0.11 and 0.12 as we gradually shift kops support to terraform 0.12. We have one integration test that generates 0.12 JSON syntax and the remaining all use 0.11 HCL syntax, but that will change soon.

rifelpet:
closed via #8734 
/close

PR #8696, opened 2020-03-08
Add instance group option to use LaunchTemplate

hakman:
As agreed during Office Hours, launch templates should move from experimental to an InstanceGroup option.

At the moment, the main reason to switch to launch templates would be to get tags for root volumes and other resources. Versions should also be explored, instead of creating a launch template on each update.

This should be merged before #8693 to see how those changes affect the launch templates.


tests/integration/update_cluster/launch_templates/kubernetes.tf
```diff
@@ -0,0 +1,733 @@
...
+    key                 = "kubernetes.io/cluster/launchtemplates.example.com"
+    value               = "owned"
+    propagate_at_launch = true
+  }
+
+  metrics_granularity = "1Minute"
+  enabled_metrics     = ["GroupDesiredCapacity", "GroupInServiceInstances", "GroupMaxSize", "GroupMinSize", "GroupPendingInstances", "GroupStandbyInstances", "GroupTerminatingInstances", "GroupTotalInstances"]
+}
+
+resource "aws_autoscaling_group" "nodes-launchtemplates-example-com" {
```
rifelpet: This ASG resource doesnt have the `launch_template` property that associates it with the `aws_launch_template` defined below.

https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html#with-latest-version-of-launch-template

I know your other PR mentions that LT werent working at all, but it seems it only fixes CF, not TF. It may be out of scope for this PR but we should definitely fix that because it seems that launch templates wont work with terraform at all currently.

Side note: i'm hoping that https://github.com/kubernetes/kops/issues/8648 will prevent this from happening again in the future, since I believe this would be caught during a `terraform validate` or `terraform plan`

hakman: Yes, that was the main problem with TF. The other PR should fix it with this:
https://github.com/kubernetes/kops/blob/1294d975d478ed7ba18d9fe9d53110ab11578ddb/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go#L736-L740

johngmyers:
I don't remember that agreement; that comment must have passed under my radar.

What is the value in keeping support for launch configurations long-term? Wouldn't a better progression of the feature be to change the default of the feature flag and then later remove support for launch configurations?

With the launch templates code needing so many fixes this round, perhaps taking them out from under the feature flag should be deferred to kops 1.19.

hakman:
@johngmyers this was the agreement at the Jan 31 meeting. I asked if the flag can be removed. The answer was that yes, because people are already using it in mixed instances. But, to be safe, first change it to instancegroup option, disabled by default.

There is no value in keeping support for launch configuration long term. We cannot remove them until we are sure that the upgrade for existing clusters is without issues.

Kops feature flags are hard to use and test. This is why it's best to move it to option for now and once we are comfortable, we can set it as default.

rifelpet:
One other benefit of using an API field rather than feature flag is it allows the cluster administrator to rollout the LC -> LT migration on a per-IG basis rather than all InstanceGroups at once. The downside is that the field's relevance is relatively short. At some point we'd want to have new clusters default to using LaunchTemplates and eventually consider having existing clusters migrated by default as well.

If we're looking for precedence for making this an API field, I think a comparable field would be the [etcd Provider](https://godoc.org/k8s.io/kops/pkg/apis/kops#EtcdClusterSpec), it was primarily relevant during the etcd2/etcd3 migration but long term I don't see us needing to support the `standalone` value.

johngmyers:
The disadvantage to an API field is that it takes an extremely long time to remove.

Is there a particular reason an admin would want to rollout the migration on a per-IG basis? This doesn't seem like it should be as impactful as changing the etcd provider.

This isn't a strong objection; it just seems a bit overcomplicated and overly user visible.

hakman:
@justinsb Can you please check that this is as we agreed during office hours?

PR #8698, opened 2020-03-08
Use latest patch release for Calico, Canal and Cilium

hakman:
Seems that there will be a Kops 1.16.1 .
This PR is updating some of the network plugins to the latest patch releases:
* Calico - 3.9.5
* Canal - 3.75
* Cilium - 1.6.7

### DIFF
--- a/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
+++ b/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
@@ -828,7 +828,7 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 			"k8s-1.6":     "2.6.9-kops.1",
 			"k8s-1.7":     "2.6.12-kops.1",
 			"k8s-1.7-v3":  "3.8.0-kops.1",
-			"k8s-1.12":    "3.9.3-kops.1",
+			"k8s-1.12":    "3.9.5-kops.1",
 		}
 
 		{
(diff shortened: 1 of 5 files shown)


upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -1132,7 +1132,7 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 
 	if b.cluster.Spec.Networking.Cilium != nil {
 		key := "networking.cilium.io"
-		version := "1.6.4-kops.3"
+		version := "1.6.7-kops.1"
```
johngmyers: We're not actually changing the Cilium template, so this bump seems unnecessary.

hakman: Seems better to be consistent. I don't really like moving the version outside the template.

johngmyers: I prefer moving the version outside of the template as it allows for easier rollback. But I set the version of Cilium to 99.99.99 to prevent kops from trying to manage it.

justinsb: I see we have now embedded the "expected" version in the manifest, which is more how we've done it.

I'm very open to discussing how we do the management of these versions (and indeed I'm hoping to get addon management to a much better place generally), but this version bump is in keeping with how we've done it to date (and doesn't seem to break your workaround @johngmyers  - we have a few more years yet!), so I'm going to approve.

justinsb:
Thanks @hakman 

/approve
/lgtm

justinsb:
Which of these version bumps do we need in 1.17 @hakman?  I'm guessing for cilium we're going to backport the 1.7.1 version, but are we OK on calico & canal?

hakman:
@justinsb These are already bumped in 1.17. Calico and Canal are ready for release.

PR #8700, opened 2020-03-09
Balk at using Kubernetes versions too new to be supported

johngmyers:


johngmyers:
e2e test is failing because it is trying to use a k8s 1.19 alpha, which the added code prevents.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1053,6 +1053,29 @@ func (c *ApplyClusterCmd) validateKubernetesVersion() error {
...
 
+	kopsVersion, err := semver.Parse(kopsbase.KOPS_RELEASE_VERSION)
+	if err != nil {
+		klog.Warningf("unable to parse kops version %q", kopsVersion)
+	} else {
+		tooNewVersion := kopsVersion
+		tooNewVersion.Minor += 1
+		tooNewVersion.Pre = nil
+		tooNewVersion.Build = nil
+		if util.IsKubernetesGTE(tooNewVersion.String(), *parsed) {
```
justinsb: I wonder about GT vs GTE.  I think this means that e.g. kops 1.18 won't be able to run k8s 1.19 alpha.1 ?

It's probably fine but you know me - super careful!

Shall we discuss in office hours?

justinsb:
Another option we could do is to set these guidelines in the channel file (`channels/stable`).  Though I do agree this is a nice way to do it - thanks for starting us off here!

upup/pkg/fi/cloudup/apply_cluster.go
johngmyers: I intended for kops 1.18 to balk at running k8s 1.19.0-alpha.1 (unless the environment variable was set)

johngmyers:
Would we expect to change this guideline post-release? If not, I don't see the benefit of setting this in the channel file.

hakman:
Any special reason this is desired? I would rather not have this behaviour at all, especially not another env var. Anyone that chooses a newer Kubernetes version does this already at its own risk by setting it manually in the cluster spec.

upup/pkg/fi/cloudup/apply_cluster.go
justinsb: Yeah, I saw that after I commented :-)

It does mean that whatever tests we're still running against k8s HEAD will break at some random point in the cycle when k/k cuts the next alpha.  OTOH that might be a good promopt for us to cut the branch for the next release - though it is typically pretty early in the cycle meaning we'd have to do a _lot_ of cherry picks.

rifelpet: FWIW this test-infra PR will fix the prow tests: https://github.com/kubernetes/test-infra/pull/16678/files

johngmyers:
I've seen several issues filed by users who were using older versions of kops than the Kubernetes they were running, with no idea that wasn't supported. It would be to the users' benefit if kops declined to perform actions that are so risky to merit an explicit disclaimer of support.

hakman:
We have to find a balance between users not reading any docs and develoment pain. Having env vars required to test newer k8s versions is not pleasant. I think a good balance would be to keep only the warning.

johngmyers:
Setting an environment variable is easy, particularly compared to having to host your own assets. I believe this is a case where the needs of the users outweigh convenience to developers. The environment variable override is the pattern established by the code handling k8s versions marked as obsolete in the channel.

hakman:
> The environment variable override is the pattern established by the code handling k8s versions marked as obsolete in the channel.

This is an old piece of code that will probably be replaced by the new approach of not supporting old releases in 1.19 or 1.20.

I think this is something on which we will disagree. From my point of view, env vars are unfriendly. Even the message is annoying to see every time you run a command.

This is not a tool for the casual user. If someone doesn't spend the time to even check the compatibility matrix, there is a very slim chance that it will be able to host its own cluster.
There will be similar questions when the user will not read the release notes and so on.

olemarkus:
FWIW the place I have seen problems here is where one admin forgets to update kops locally and accidentally causes problems. Say what you want about such mistakes, but it does happen.

hakman:
> FWIW the place I have seen problems here is where one admin forgets to update kops locally and accidentally causes problems. Say what you want about such mistakes, but it does happen.

OK. Can we do it in such a way that it only applies to official releases?

johngmyers:
Heh, our wrapper around kops puts its version in an annotation on the cluster spec. It then requires a switch to run an older version on that cluster.

rifelpet:
we decided during last office hours that this is good to merge. developers can set the env var in their shell, and CI jobs will do the same according to the test-infra PR linked above.

/lgtm
/approve

Issue #8238, opened 2019-12-31
[Helped]so many TODO in kops/pkg/ , can we solve some of them?


tanjunchen:
There  are so many TODO in kops , I guess that we can solve or improve some of them. 
I open an issue to record them.
Here are TODO under /pkg folder.

If the following TODO does not exist, I hope you can reply this TODO in here, so I can update this list in time. Or you have solved one of the TODOs and hope you can associate this issue for easy tracking.

```
grep -rs --exclude-dir="vendor/"   "TODO"  | wc -l
767
```

- [ ] pkg/apis/kops/validation/legacy.go:		// TODO: Auto choose zones from region?
- [ ] pkg/apis/kops/validation/legacy.go:// @TODO semvar package doesn't appear to ignore a 'v' in v1.1.1 should could be a problem later down the line
- [ ] pkg/apis/kops/validation/aws.go:// TODO: make image validation smarter? graduate from jessie to stretch? This is quick and dirty because we keep getting reports
- [ ] pkg/apis/kops/validation/aws.go:	// TODO: how can we put this list somewhere better?
- [ ] pkg/apis/kops/util/versions.go:// TODO: Convert to our own KubernetesVersion type?
- [ ] pkg/apis/kops/v1alpha1/zz_generated.conversion.go:			// TODO: Inefficient conversion - can we improve it?
- [ ] pkg/apis/kops/v1alpha1/zz_generated.conversion.go:			// TODO: Inefficient conversion - can we improve it?
- [ ] pkg/apis/kops/v1alpha1/zz_generated.conversion.go:			// TODO: Inefficient conversion - can we improve it?
- [ ] pkg/apis/kops/v1alpha1/zz_generated.conversion.go:			// TODO: Inefficient conversion - can we improve it?
- [ ] pkg/apis/kops/v1alpha1/zz_generated.conversion.go:			// TODO: Inefficient conversion - can we improve it?
- [ ] pkg/apis/kops/v1alpha1/zz_generated.conversion.go:			// TODO: Inefficient conversion - can we improve it?
- [ ] pkg/apis/kops/v1alpha1/zz_generated.conversion.go:			// TODO: Inefficient conversion - can we improve it?
- [ ] pkg/apis/kops/v1alpha1/zz_generated.conversion.go:			// TODO: Inefficient conversion - can we improve it?
- [ ] pkg/apis/kops/v1alpha1/cluster.go:	// TODO #3070
[...]

PR #8701, opened 2020-03-09
remove // TODO path.Join here?

tanjunchen:
ref:https://github.com/kubernetes/kops/issues/8238

### DIFF
--- a/pkg/model/components/context.go
+++ b/pkg/model/components/context.go
@@ -176,8 +176,7 @@ func Image(component string, architecture string, clusterSpec *kops.ClusterSpec,
 
 	baseURL := clusterSpec.KubernetesVersion
 	baseURL = strings.TrimSuffix(baseURL, "/")
-
-	// TODO path.Join here?
+	
 	tagURL := baseURL + "/bin/linux/" + architecture + "/" + component + ".docker_tag"
 	klog.V(2).Infof("Downloading docker tag for %s from: %s", component, tagURL)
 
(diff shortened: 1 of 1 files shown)


pkg/model/components/context.go
```diff
@@ -177,8 +178,7 @@ func Image(component string, architecture string, clusterSpec *kops.ClusterSpec,
 	baseURL := clusterSpec.KubernetesVersion
 	baseURL = strings.TrimSuffix(baseURL, "/")
 
-	// TODO path.Join here?
-	tagURL := baseURL + "/bin/linux/" + architecture + "/" + component + ".docker_tag"
+	tagURL := path.Join(baseURL, "/bin/linux/", architecture, component) + ".docker_tag"
```
johngmyers: Shouldn't "/bin/linux/" omit the leading and trailing slashes?

Why is the ".docker_tag" outside the path.Join? If `component` is the empty string, these two are not equivalent.

tanjunchen: https://play.golang.org/p/EK7ti5lYdSp

tanjunchen: see https://play.golang.org/p/zrbeKMcGKg0

tanjunchen: it seems that path.join will result `"memfs://v1.10.0-download"`  to  `memfs:/v1.10.0-download/`,it will fail. It's not reasonable to use this here.@johngmyers what do you say?

johngmyers: You'd have to parse the url and use `path.Join` on the path component only.

I'm skeptical of the value of using `path.Join` in this instance. It doesn't seem to be solving a problem that simple concatenation doesn't.

tanjunchen: agree

justinsb:
/retest

prow timeout

PR #8711, opened 2020-03-10
remove // TODO move this into networking.go :(

tanjunchen:
ref:https://github.com/kubernetes/kops/issues/8238

I don't think this needs to be moved, I use it as a check function. 
Is there a better solution here? Maybe leave it here ,it is better.

### DIFF
--- a/pkg/apis/kops/channel.go
+++ b/pkg/apis/kops/channel.go
@@ -31,7 +31,6 @@ var DefaultChannelBase = "https://raw.githubusercontent.com/kubernetes/kops/mast
 
 const (
 	DefaultChannel = "stable"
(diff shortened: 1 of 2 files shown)


pkg/apis/kops/channel.go
```diff
@@ -31,7 +31,6 @@ var DefaultChannelBase = "https://raw.githubusercontent.com/kubernetes/kops/mast
 
 const (
 	DefaultChannel = "stable"
-	AlphaChannel   = "alpha"
```
tanjunchen: Unused code , so remove it .

justinsb:
Thanks @tanjunchen 

/approve
/lgtm

PR #8726, opened 2020-03-11
Switch AWS NAT Gateway creation to use tags on create

rifelpet:
AWS has been adding support for "tag on create" to many services. Typically this support also includes referencing tags in IAM policies involving these resources.

This means kops could use a more locked-down IAM policy that only allows creating NAT Gateways if they include cluster-specific tags, for example. Ideally we could use this for as many resource types as possible and have a much more locked down example IAM policy for kops users.

Marking WIP until I get a chance to test this locally since none of our e2e jobs create NAT gateways.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/natgateway.go
+++ b/upup/pkg/fi/cloudup/awstasks/natgateway.go
@@ -307,7 +307,9 @@ func (_ *NatGateway) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *NatGateway)
 
 		klog.V(2).Infof("Creating Nat Gateway")
 
-		request := &ec2.CreateNatGatewayInput{}
+		request := &ec2.CreateNatGatewayInput{
+			TagSpecifications: awsup.EC2TagSpecification(ec2.ResourceTypeNatgateway, e.Tags),
+		}
 		request.AllocationId = e.ElasticIP.ID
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/awstasks/tags.go
```diff
@@ -62,3 +62,15 @@ func intersectTags(tags []*ec2.Tag, desired map[string]string) map[string]string
 	}
 	return actual
 }
+
+func mapToTagSpecification(tags map[string]string, resourceType string) *ec2.TagSpecification {
```
hakman: I think there are a few other places where TagSpecification is used and this should be used to replace the tag conversion there.

hakman:
@rifelpet I just gave this a try and created the NAT GW with tags.

rifelpet:
I think i have to update cloudmock with equivalent logic

https://github.com/kubernetes/kops/blob/master/cloudmock/aws/mockec2/natgateway.go

I'm hoping to get to that soon.

hakman:
No rush. Just wanted to let you know that it works :).

rifelpet:
my other recent PRs have made this much simpler. This will still use ec2.CreateTags for the `AssociatedNatgateway` tag a few lines below, but we can consider updating or removing that in a separate PR.

PR #8727, opened 2020-03-11
Add support for Ubuntu 20.04 (Focal)

hakman:
Ubuntu 20.04 is close to release and will be the next Ubuntu LTS release.
From Kubernetes point of view, this will be the supported distro with the newest Kernel for some time (Debian 10 and RHEL/CentOS 8 were just released).

Once official stable images are released will add an e2e test for it.

### DIFF
--- a/nodeup/pkg/model/kubelet.go
+++ b/nodeup/pkg/model/kubelet.go
@@ -597,8 +597,10 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 
 	// In certain configurations systemd-resolved will put the loopback address 127.0.0.53 as a nameserver into /etc/resolv.conf
 	// https://github.com/coredns/coredns/blob/master/plugin/loop/README.md#troubleshooting-loops-in-kubernetes-clusters
-	if b.Distribution == distros.DistributionBionic && c.ResolverConfig == nil {
-		c.ResolverConfig = s("/run/systemd/resolve/resolv.conf")
+	if c.ResolverConfig == nil {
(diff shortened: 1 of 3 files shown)


nodeup/pkg/model/kubelet.go
```diff
@@ -597,8 +597,10 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 
 	// In certain configurations systemd-resolved will put the loopback address 127.0.0.53 as a nameserver into /etc/resolv.conf
 	// https://github.com/coredns/coredns/blob/master/plugin/loop/README.md#troubleshooting-loops-in-kubernetes-clusters
-	if b.Distribution == distros.DistributionBionic && c.ResolverConfig == nil {
-		c.ResolverConfig = s("/run/systemd/resolve/resolv.conf")
+	if c.ResolverConfig == nil {
+		if b.Distribution == distros.DistributionBionic || b.Distribution == distros.DistributionFocal {
```
justinsb: Out of interest, do you know that this was needed?  I was hoping they might have fixed it!

justinsb:
Thanks @hakman 

/approve
/lgtm

nodeup/pkg/model/kubelet.go
hakman: This is a more general issue for distros that enable DNS cache via `systemd-resolved`, so not sure it can be fixed. Without this CoreDNS keeps complaining about a DNS loop:
https://github.com/coredns/coredns/blob/master/plugin/loop/README.md#troubleshooting-loops-in-kubernetes-clusters

PR #8731, opened 2020-03-12
Allow configuration of enable-remote-node-identity

olemarkus:


hakman:
@olemarkus I think you need to run `make gofmt` to get the tests to pass.

olemarkus:
I had hoped to avoid that in order to avoid a massive conflict with the API docs PR that was about to go in. Now that it has been merged, I'll fix and add docs to this one too.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -137,6 +137,7 @@ data:
   auto-direct-node-routes: "{{- if .AutoDirectNodeRoutes -}}true{{- else -}}false{{- end -}}"
   enable-node-port: "{{- if .EnableNodePort -}}true{{- else -}}false{{- end -}}"
   kube-proxy-replacement: "{{- if .EnableNodePort -}}strict{{- else -}}partial{{- end -}}"
+  enable-remote-node-identity: "{{- if .EnableRemoteNodeIdentity -}}true{{- else -}}false{{- end -}}"
```
rifelpet: is it okay that this field is present even if running cilium <1.7.0 ?

olemarkus: Yeah. Cilium just ignores it.

PR #8732, opened 2020-03-13
Update apiVersion to resolve 'no matches for kind "Deployment" in ver…

medined:
…sion "extensions/v1beta"' message.

I am running kubectl version:
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.7", GitCommit:"be3d344ed06bff7a4fc60656200a93c74f31f9a4", GitTreeState:"clean", BuildDate:"2020-02-11T19:24:46Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```

### DIFF
--- a/addons/route53-mapper/v1.3.0.yml
+++ b/addons/route53-mapper/v1.3.0.yml
@@ -1,5 +1,5 @@
 ---
-apiVersion: app/v1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: route53-mapper
--- a/addons/route53-mapper/v1.3.0.yml
+++ b/addons/route53-mapper/v1.3.0.yml
@@ -1,5 +1,5 @@
 ---
-apiVersion: extensions/v1beta1
+apiVersion: app/v1
 kind: Deployment
 metadata:
   name: route53-mapper


addons/route53-mapper/v1.3.0.yml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: extensions/v1beta1
+apiVersion: app/v1
```
rifelpet:
```suggestion
apiVersion: apps/v1
```
I believe it should be plural

See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#deployment-v1-apps

medined: Transcription error. My apologies.

johngmyers:
Not surprising since this hasn't been touched in just shy of three years. route53-kubernetes has been officially deprecated in favor of external-dns; could we perhaps just remove this?

medined:
If it is left, please add a comment about it be superseded by external-dns. Thanks for responding so quickly.

rifelpet:
I think we can merge this and add a deprecation notice. The [instructions](https://github.com/kubernetes/kops/blob/master/docs/operations/addons.md#route53-mapper) mention to kubectl apply directly from github, which means that removing the file could break existing users that might include that command in their workflow. I'm not sure how big of a deal that is. I'll open a PR announcing the deprecation and perhaps we can discuss it at the next office hours.

Regardless, thanks for the fix @medined  

/lgtm
/approve

PR #8734, opened 2020-03-13
Add terraform testing

rifelpet:
This adds a new `verify-terraform` make target and hack script that uses the official terraform docker image to validate the .tf files in our integration tests.

This can catch issues like:
* missing required resource properties
* typos in resource property names
* invalid values for resource properties

Unfortunately the test fails unless all referenced external files are present, which means we need to check in the userdata, IAM policies, and SSH public key files used by the terraform code's `file()` references. I decided to update the regular integration tests to also validate their contents. hack/updated-expected.sh will now also create these files.

Marking WIP because some clusters didn't have all their files created, i think it might be an issue with the expected list of files vs actual list of files. also need to get this connected to a prow job.

Feel free to review the hack script now though. I chose to use docker in order to make it work across platforms more easily. I tried using bazel but only found a terraform rules repo that is two years old so I was a bit hesitant to use it. shelling out from bazel seemed messy too.
ref: #8648


rifelpet:
I think this is ready for review/merge.

There are failures with google's terraform setup. The hardcoded provider version requires terraform >= 0.12 but the remainder of the syntax requires terraform <0.12. This will be fixed as  0.12 support is added in a future PR.

#8744 will also "conflict" with this PR in the sense that tests will fail if both are merged without one of them requiring changes (I have no preference which lands first, I'll fix the other). This is because this PR adds empty bastion userdata files in order to pass the integration tests, but #8744 will assert that they don't exist.

Because of the failures, I'm making the script non-blocking: it exits 0 even in the case of failures. I'd rather get this merged early to avoid further issues with other PRs and the new data files now being checked in.

Once #8744 is merged and we either fix the google provider or rollout the 0.12 upgrade we can make the script blocking.

rifelpet:
@mikesplain you mentioned in office hours yesterday that you use Terraform. Any chance you can take a look at this?

mikesplain:
Absolutely, I’ll take a look when I’m back at my machine.

rifelpet:
I had to add empty userdata files for bastion launch configs (and launch templates) because we still compare them and ensure they exist. I actually have #8744 open which will negate that, but that requires #8737 which is ready for review/merge whenever.

Lastly theres an issue with our golden compare and line endings. the comparison strips carriage returns before comparing:

https://github.com/kubernetes/kops/blob/90af3a14b9c0c52ff50bd3726d14f76b24d8dbb8/pkg/testutils/golden/compare.go#L43-L44

but the additionalUserData field actually adds carriage returns:

https://github.com/kubernetes/kops/blob/90af3a14b9c0c52ff50bd3726d14f76b24d8dbb8/pkg/model/resources/nodeup.go#L197-L198

meaning that ./hack/updated-expected.sh wont update the file properly and the subsequent test will continue to fail. I'm still working on how to handle this, so i expect my most recent push to fail some tests.

In the mean time, #8737 is ready to go. If it causes merge conflicts with this or #8744 I'll rebase them asap while I try to work through this newline issue.

PR #8735, opened 2020-03-13
Update kube-router to v0.4.0

hakman:
Let's see if kube-router v0.4.0 fixes the issues we are seeing in e2e:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-aws-cni-kuberouter/1238479448160866306

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.kuberouter/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.kuberouter/k8s-1.12.yaml.template
@@ -1,6 +1,5 @@
-# Pulled and modified from https://github.com/cloudnativelabs/kube-router/blob/v0.3.1/daemonset/generic-kuberouter-all-features.yaml
+# Pulled and modified from https://github.com/cloudnativelabs/kube-router/blob/v0.4.0/daemonset/generic-kuberouter-all-features.yaml
 
----
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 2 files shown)


upup/models/cloudup/resources/addons/networking.kuberouter/k8s-1.12.yaml.template
```diff
@@ -1,4 +1,4 @@
-# Pulled and modified from https://github.com/cloudnativelabs/kube-router/blob/v0.3.1/daemonset/generic-kuberouter-all-features.yaml
+# Pulled and modified from https://github.com/cloudnativelabs/kube-router/blob/v0.4.0/daemonset/generic-kuberouter-all-features.yaml
 
 ---
```
rifelpet: can we remove this `---` so that the comment above doesnt get treated as a separate yaml manifest? :) I remember us removing that from all the other addons as well, even though kops can now correctly handle that situation it'd be good to not have it in the `update cluster` diffs.

hakman: No objection. Done.

upup/models/cloudup/resources/addons/networking.kuberouter/k8s-1.12.yaml.template
```diff
@@ -133,32 +128,34 @@ metadata:
...
     resources:
       - namespaces
       - pods
       - services
       - nodes
       - endpoints
     verbs:
-      - get
       - list
+  - get
```
rifelpet: indentation issue here? and the apiGroups changes above and below.

hakman: You are right. Forgot to enable white space diff and the editor went wild. Thanks :)

PR #8742, opened 2020-03-14
Fix template for kube-router v0.4.0

hakman:
#8735 uses the [kubeadm-kuberouter.yaml](https://github.com/cloudnativelabs/kube-router/blob/v0.4.0/daemonset/kubeadm-kuberouter.yaml) template as base, but it's broken. To fix it, I am trying to use the [generic-kuberouter.yaml](https://github.com/cloudnativelabs/kube-router/blob/v0.4.0/daemonset/generic-kuberouter.yaml).

The only real difference is removing `--kubeconfig=/var/lib/kube-router/kubeconfig` from args. The rest of changes are whitespace to make the diff easier next time.

My setup seems broken as I cannot get this to work on my end. Hoping that after this change, the e2e will start working again:
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-cni-kuberouter


UnderMyBed:
I see this is currently has the milestone for kops 1.18 but I am pretty sure this needs to get back ported into 1.16 and 1.17. If we don't do a backport here then upgrading the clusters is going to be quite challenging.

hakman:
Thanks for the reminder @UnderMyBed. I opened the necessary cherrypicks.

PR #8737, opened 2020-03-13
Refactor integration tests

rifelpet:
closes #8736

This refactors the integration tests so that it is easier to add new variables to tests in the future including the ability to provide default values.

PR #8744, opened 2020-03-14
Don't set terraform user_data if it is empty 

rifelpet:
Bastion hosts have no user_data by default. This is not valid in terraform, evidenced by the error messages reported [here](https://travis-ci.org/github/kubernetes/kops/jobs/662472020#L606).

The Terraform provider code says user_data is optional for both [LaunchConfigurations](https://github.com/terraform-providers/terraform-provider-aws/blob/04d24f80f3b740d0c9d74d97e818b7da15cb96c5/aws/resource_aws_launch_configuration.go#L74-L76) and [LaunchTemplates](https://github.com/terraform-providers/terraform-provider-aws/blob/04d24f80f3b740d0c9d74d97e818b7da15cb96c5/aws/resource_aws_launch_template.go#L510-L512).

This change prevents the user_data property from being added to `aws_launch_configuration` and `aws_launch_template` resources if the string is empty.

### DIFF
--- a/cmd/kops/integration_test.go
+++ b/cmd/kops/integration_test.go
@@ -61,6 +61,7 @@ type integrationTest struct {
 	lifecycleOverrides []string
 	sshKey             bool
 	jsonOutput         bool
+	bastionUserData    bool
 }
 
 func newIntegrationTest(clusterName, srcDir string) *integrationTest {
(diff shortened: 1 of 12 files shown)


rifelpet:
Ugh integration tests failing because we shouldnt [expect a bastion user_data file](https://github.com/kubernetes/kops/blob/7d030ae459751a661b83d550c3035f3b7f0f857e/cmd/kops/integration_test.go#L487) unless the additionalUserData isnt empty. adding support for that in the integration tests will be a significant change which would conflict with #8737 so i'm inclined to wait for that to get merged which might be a while.

PR #8745, opened 2020-03-14
Set kube-proxy-replacement to partial

olemarkus:
Probing for bpf-features doesn't work properly on buster images. Setting it to partial works better. Same setting is the default in the master branch already (implicitly configured by the bpf nodeport flag)


rifelpet:
this PR not being a cherry pick makes sense to me. we're not going to be cherry-picking the kube-proxy-replacement changes on master back to release-1.17.

/lgtm
/approve

rifelpet:
though we wont have any periodic e2e coverage of this change, so it would be nice to test this manually on a variety of distros.

PR #8747, opened 2020-03-15
Tweak the featureflag.GoogleCloudBucketACL.Enabled

geojaz:
This expands the code that gets turned on and off under the `GoogleCloudBucketACL` flag.
It wasn't clear to me why the flag applied to only one of the ACLs that we were setting. If this is not the intended behavior, we should document how to use the flag a bit more clearly :)


geojaz:
Hrm, welp this didn't work either. I'll have to follow up with the testing infra folks

PR #8749, opened 2020-03-15
Update bazel rules

rifelpet:
* https://github.com/bazelbuild/bazel-gazelle/releases
* https://github.com/bazelbuild/rules_docker/releases

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -7,19 +7,19 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 http_archive(
     name = "io_bazel_rules_go",
     urls = [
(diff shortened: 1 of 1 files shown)


rifelpet:
looks like we're still on [bazel 0.28.1](https://github.com/kubernetes/test-infra/blob/a0cadb92a2b48f5bbff706038ab709c9413002be/images/kubekins-e2e/variants.yaml#L6) but the latest versions of these rules requires a more recent version. I'll try to determine the latest version of the rules that can still work with bazel 0.28.1.

I'm not sure what the upgrade plan for bazel is given that we share the kubekins-e2e image with many other projects, and there is discussion of removing bazel from k/k entirely.

rifelpet:
ifconfig.co issues...
/retest

PR #8750, opened 2020-03-15
Use etcd-manager for the cilium etcd cluster

olemarkus:
This will use etcd-manager to handle the cilium etcd kvstore instead of using cilium-etcd-operator.
Just brute-forced my way until it worked, so it still needs some cleanup. But have a look if there are any concerns with this approach.

cc @johngmyers


hakman:
I think it is a good time to create a  `nodeup/pkg/model/networking/` dir, or at least prepend the file name with "networking".

olemarkus:
I wouldn't do that as part of this PR, but if this one goes in, I can do a follow-up creating that dir.

johngmyers:
How does this interact with legacy etcd mode? Do we need to add a validation to prevent this combination?

olemarkus:
Yeah. We need to ensure that the cilium etcd cluster is using etcd-manager. I will add some validation on that.

olemarkus:
@justinsb anything you think needs to be changed in this MR?

nodeup/pkg/model/cilium.go
```diff
@@ -0,0 +1,197 @@
...
+
+	humanName := dir + "/" + name
+	privateKey, err := pkiutil.NewPrivateKey()
+	if err != nil {
+		return fmt.Errorf("unable to create private key %q: %v", humanName, err)
+	}
+	privateKeyBytes := pkiutil.EncodePrivateKeyPEM(privateKey)
+
+	certConfig := &certutil.Config{
+		CommonName: "kube-apiserver",
```
rifelpet, line 166: looks like a copy/paste from [etcd_manager_tls.go](https://github.com/kubernetes/kops/blob/master/nodeup/pkg/model/etcd_manager_tls.go), should this cert have a different CN?

pkg/model/firewall.go
```diff
@@ -240,6 +240,7 @@ func (b *FirewallModelBuilder) applyNodeToMasterBlockSpecificPorts(c *fi.ModelBu
 	// Don't allow nodes to access etcd peer port
 	tcpBlocked[2380] = true
 	tcpBlocked[2381] = true
+	tcpBlocked[2382] = true
```
rifelpet, line 243: Could we conditionally block this port only if the cluster is using the etcd managed cilium? that would remove the security group changes being reported for clusters that don't use etcd-managed cilium which might be preferable. 

There are other CNIs conditionally blocking ports below, so it might be wise to do that in this case as well.

rifelpet:
Thanks for the additional tests! the new integration tests confirm the security group rule change is only present if cilium etcdManaged is enabled.

/approve
/lgtm

PR #8751, opened 2020-03-15
Fix template for kube-router v0.4.0 #2

hakman:
Followup of #8742. This time works on my test cluster also.


rifelpet:
for reference, the kubeconfig file being mounted is defined [here](https://github.com/kubernetes/kops/blob/af4df08b694e2a1f8814a7b3649060477be67c86/nodeup/pkg/model/kube_router.go)

/lgtm
/approve

hakman:
> for reference, the kubeconfig file being mounted is defined [here](https://github.com/kubernetes/kops/blob/af4df08b694e2a1f8814a7b3649060477be67c86/nodeup/pkg/model/kube_router.go)

Yes, just noticed that it was the main issue :).

PR #8753, opened 2020-03-15
Deprecate the route53-mapper addon

rifelpet:
ref: #8732 

I wanted to update the kubectl apply URLs to not use master but we'd want to wait until a stable release includes #8732 or a cherry-pick of it before referencing something besides master.


johngmyers:
Do we want to set a kops version when we'll remove it?

This brings up another topic: of the non-bootstrap addons, only prometheus-operator, metrics-server, ingress-citrix, and cluster-autoscaler have had their manifests updated in the last year. Perhaps we should deprecate some or all of the remaining non-bootstrap addons?

PR #8757, opened 2020-03-16
Update lyft CNI to 0.6.0

maruina:
This PR updates the lyft CNI to the latest version, which:

- Use the AWS API to retrieve instance network limits. This mean we can know use **any** instance type.
- Add IP allocation batching

~Marked as WIP because we want to test it first.~
As @gjtempleton said, we tested it and it's working as expected.


gjtempleton:
Can confirm that this works once the extra IAM permission is added.

PR #8758, opened 2020-03-16
Add LaunchTemplate integration tests

hakman:
Launch templates should replace launch configurations for AWS configs in a future release. More and more fixes are added to make launch templates production ready. Having integration tests should help tracking the effects of these changes.


rifelpet:
for reference, the terraform being generated in this PR is invalid because the aws_autoscaling_group doesnt reference any aws_launch_template. We're hoping that https://github.com/kubernetes/kops/pull/8693 will fix that, and we'll more clearly see the fix now that we have a terraform test for this functionality.

/lgtm
/approve

PR #8762, opened 2020-03-17
add support for OpenStack Octavia AllowedCIDRs

### LINKED ISSUE #8383: OpenStack: Octavia loadbalancer requires ".AllowedCIDRs" instead of Security Group
 this information.**

Kops 1.15 - 1.17 alpha

 a `kops` flag.**

Default included Kubernetes versions

OpenStack

kops create cluster --os-octavia=true, kops update cluster --yes

Cluster becomes healthy. Whatever I add/remove from the API (loadbalancer masters) group it does not matter: everything go's through.

Security Group Rules should be honored.

I rolled out multiple kops versions on multiple OpenStack providers, each resulting in the same wrong behaviour (no security on API LB).

zetaab:
fixes #8383

@kciredor do you have possibility to test this? Do you know how to compile kops manually, or you want that I will compile it? If yes, what OS do you use

This needs to be tested in following use-cases:
- [x] Old octavia new cluster
- [x] Old octavia update sec group rules for api
- [x] Update from old octavia to new octavia (should delete old security group)
- [x] New octavia new cluster
- [x] New octavia update sec group rules for api

/hold


kciredor:
Awesome @zetaab, moving fast!

Yes I'm used to compiling/patching kops so no problem there. Will let you know as soon as I can!

kciredor:
Works great!

I've just tested it on a new/clean cluster by adding/removing ip ranges (kops edit cluster + kops update cluster --yes) and making API calls in between.

Could this still be merged into the 1.17 release branch which is currently beta 1? That would be cool.

PR #8763, opened 2020-03-17
Fix NTP failures after #8312

hakman:
#8312 broke support for some our supported distros as seen in periodic e2e tests.
This should fix the support by using Chrony (already default in most of the distros).

### DIFF
--- a/nodeup/pkg/model/ntp.go
+++ b/nodeup/pkg/model/ntp.go
@@ -17,10 +17,6 @@ limitations under the License.
 package model
 
 import (
-	"fmt"
-	"io/ioutil"
-	"regexp"
-
 	"k8s.io/klog"
 	"k8s.io/kops/nodeup/pkg/distros"
 	"k8s.io/kops/upup/pkg/fi"
@@ -36,13 +32,6 @@ type NTPBuilder struct {
 
 var _ fi.ModelBuilder = &NTPBuilder{}
 
-type ntpDaemon string
-
-var (
-	chronyd ntpDaemon = "chronyd"
-	ntpd    ntpDaemon = "ntpd"
-)
-
 // Build is responsible for configuring NTP
 func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 	switch b.Distribution {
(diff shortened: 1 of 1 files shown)


nodeup/pkg/model/ntp.go
```diff
@@ -57,103 +46,51 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
...
-			if err != nil {
-				return err
-			}
-			c.AddTask(&nodetasks.File{
-				Path:     "/etc/ntp.conf",
-				Contents: fi.NewBytesResource(bytes),
-				Type:     nodetasks.FileType_File,
-				Mode:     s("0644"),
-			})
+		c.AddTask(&nodetasks.Package{Name: "chrony"})
```
rifelpet: This will affect all debian and ubuntu distros, not just the most recent versions, correct? Are we sure the chrony package is available in older apt/yum repos?

I see it in [ubuntu](https://launchpad.net/ubuntu/+source/chrony), [debian](https://packages.debian.org/search?keywords=chrony), and [centos](https://pkgs.org/search/?q=chrony) so maybe we're confident enough that the package will be available on any supported distros. Thoughts?

hakman: Chrony is available on all supported distros. It is installed by default on RHEL, CentOS and Debian. Ubuntu uses `systemd-timesyncd` by default though. To avoid having both running I updated the PR to configure it instead of Chrony for Ubuntu.

I tested this with the following distros and works:
* Amazon Linux 2
* CentOS Linux 7 (Core)
* Red Hat Enterprise Linux Server 7.4 (Maipo)
* Red Hat Enterprise Linux 8.1 (Ootpa)
* Debian GNU/Linux 9 (stretch)
* Debian GNU/Linux 10 (buster)
* Ubuntu 16.04.6 LTS
* Ubuntu 18.04.4 LTS
* Ubuntu Focal Fossa (development branch)

hakman: We can also keep Chrony for Ubuntu, but timesyncd should be stopped and disabled.

rifelpet: Ok, I think that is sufficient coverage 👍🏻

hakman:
Thanks @rifelpet :)

nhudson:
Is there anyway to get this merged into 1.17?  We are having a hard time keeping ec2 instances in sync using ntpd.

hakman:
@nhudson this a fairly big change to be added to an already released version, but I may add a `disable` switch for NTP conf and you could install chrony or timesyncd.
One other solution is to switch to the Ubuntu 20.04 base image which uses timesyncd.

nhudson:
I'm currently working around it using a hook to install and configure chrony, remove ntp and disable timesycd.  Didn't know if it was an easy backport to 1.17. Thanks for the response!

hakman:
The backport is easy, but with 1.17 being released it is not a good idea to change things.

PR #8691, opened 2020-03-06
Update Weave Net to version 2.6.1

hakman:
Weave 2.6.1 was (released)[https://github.com/weaveworks/weave/releases/v2.6.1] with support for Debian 10 (Buster).

Changes are based on:
https://github.com/weaveworks/weave/releases/download/v2.6.1/weave-daemonset-k8s-1.11.yaml
https://github.com/weaveworks/weave/releases/download/v2.6.1/weave-daemonset-k8s-1.8.yaml

This should be a good addition to the 1.17 release.

PR #8764, opened 2020-03-17
Update Weave Net to version 2.6.2

hakman:
Weave v2.6.1 contains a regression that is fixed in v2.6.2: https://github.com/kubernetes/kops/pull/8691#issuecomment-598111201


hakman:
/cc @rifelpet 
@murali-reddy

murali-reddy:
thanks @hakman

PR #8780, opened 2020-03-21
NodeLocal DNSCache

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

From k8s 1.18 NodeLocal DNSCache is GA.

This PR aims to implement NodeLocal DNSCache as an addon and enable it only when CoreDNS is enabled too.


pkg/model/components/kubedns.go
```diff
@@ -74,5 +74,29 @@ func (b *KubeDnsOptionsBuilder) BuildOptions(o interface{}) error {
 		clusterSpec.KubeDNS.MemoryLimit = &defaultMemoryLimit
 	}
 
+	NodeLocalDNS := clusterSpec.KubeDNS.NodeLocalDNS
+	if NodeLocalDNS == nil {
+		NodeLocalDNS.Disable = true
+	} else if !NodeLocalDNS.Disable {
+		// https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#configuration
+		NodeLocalDNS.Domain = clusterSpec.ClusterDNSDomain
+		NodeLocalDNS.LocalIP = "169.254.20.10"
```
mazzy89: Not sure here about this. Just copying here from the official doc.

Also I would need to implement this as the official doc states:

> If using kube-proxy in IPVS mode, --cluster-dns flag to kubelet needs to be modified to use <node-local-address> that NodeLocal DNSCache is listening on. Otherwise, there is no need to modify the value of the --cluster-dns flag, since NodeLocal DNSCache listens on both the kube-dns service IP as well as <node-local-address>.

mazzy89: Added in this commit https://github.com/kubernetes/kops/pull/8780/commits/b0cb7c1795a76bb1bd5d44542bdb11dae12ec92f

olemarkus: First this is made configurable by adding it to the cluster API and then it is just hardcoded to the address above unconditionally?
Isn't it better to test for this value in the template and set it to the address above if the value is empty?

pkg/model/components/kubelet.go
```diff
@@ -81,11 +81,15 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
 	}
 
 	if clusterSpec.Kubelet.ClusterDNS == "" {
-		ip, err := WellKnownServiceIP(clusterSpec, 10)
-		if err != nil {
-			return err
+		if clusterSpec.KubeProxy.ProxyMode == "ipvs" && !clusterSpec.KubeDNS.NodeLocalDNS.Disable {
+			clusterSpec.Kubelet.ClusterDNS = "169.254.20.10"
```
olemarkus: This is also a configurable field. It is better to add validation logic stating that you need to set this IP to a certain value if you enable this addon.
There is already some validation logic around this in https://github.com/kubernetes/kops/blob/b0cb7c1795a76bb1bd5d44542bdb11dae12ec92f/pkg/apis/kops/validation/legacy.go

mazzy89: Yeah make sense. I'll make wire some changes to take into account validation

pkg/model/components/kubelet.go
```diff
@@ -81,11 +81,15 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
 	}
 
 	if clusterSpec.Kubelet.ClusterDNS == "" {
-		ip, err := WellKnownServiceIP(clusterSpec, 10)
-		if err != nil {
-			return err
+		if clusterSpec.KubeProxy.ProxyMode == "ipvs" && !clusterSpec.KubeDNS.NodeLocalDNS.Disable {
```
rifelpet: The travis tests are reporting a panic here, you probably need to do additional `!= nil` checks

rifelpet:
/ok-to-test

My only thought on the new api fields is that it might be more intuitive to have an `Enabled` field rather than a `Disabled` field

upup/models/cloudup/resources/addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -0,0 +1,176 @@
+---
```
rifelpet: Can you include a commented link to where this manifest was copied from?

pkg/model/components/kubelet.go
mazzy89: Yeah need to fix that.

upup/models/cloudup/resources/addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml.template
mazzy89: Sure, make sense

mazzy89:
@rifelpet going to change the condition to `Enabled` because indeed more intutive.

pkg/model/components/kubedns.go
```diff
@@ -74,5 +74,33 @@ func (b *KubeDnsOptionsBuilder) BuildOptions(o interface{}) error {
 		clusterSpec.KubeDNS.MemoryLimit = &defaultMemoryLimit
 	}
 
+	NodeLocalDNS := clusterSpec.KubeDNS.NodeLocalDNS
+	if NodeLocalDNS == nil {
```
olemarkus: What is the benefit of doing this vs just keep checking if NodeLocalDNS is nil?

pkg/model/components/kubedns.go
```diff
@@ -74,5 +74,33 @@ func (b *KubeDnsOptionsBuilder) BuildOptions(o interface{}) error {
...
+		NodeLocalDNS = &kops.NodeLocalDNSConfig{}
+		NodeLocalDNS.Enabled = false
+	} else if NodeLocalDNS.Enabled {
+		// https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#configuration
+		NodeLocalDNS.Domain = clusterSpec.ClusterDNSDomain
+
+		if NodeLocalDNS.LocalIP == "" {
```
olemarkus: These tests are usually done inside of the templates so that the spec is close to immutable inside of kops. i.e the spec is the same as the one the users set.

pkg/model/components/kubelet.go
```diff
@@ -81,11 +81,17 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
 	}
 
 	if clusterSpec.Kubelet.ClusterDNS == "" {
-		ip, err := WellKnownServiceIP(clusterSpec, 10)
-		if err != nil {
-			return err
+		if clusterSpec.KubeDNS.NodeLocalDNS != nil {
+			if clusterSpec.KubeProxy.ProxyMode == "ipvs" && clusterSpec.KubeDNS.NodeLocalDNS.Enabled {
+				clusterSpec.Kubelet.ClusterDNS = "169.254.20.10"
```
olemarkus: Instead of unconditionally setting the ClusterDNS IP here, it is better to add a check in the validation logic and get the user to set the correct address in the cluster spec.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -451,6 +451,29 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 		}
 	}
 
+	// @check the node-local-dns has not been disabled
+	NodeLocalDNS := b.cluster.Spec.KubeDNS.NodeLocalDNS
+	if kubeDNS.Provider == "CoreDNS" && NodeLocalDNS.Enabled {
+		{
+			key := "nodelocaldns.addons.k8s.io"
+			version := "1.18.0-alpha.2"
```
olemarkus: Do you consider the addon alpha? If not, I would omit the alpha part here.
By the time this is released, I guess the manifest and such is GA from K8s side.

pkg/model/components/kubedns.go
mazzy89: You mean to assign this assignment?

```golang
NodeLocalDNS := clusterSpec.KubeDNS.NodeLocalDNS
```

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
mazzy89: It makes sense. the upstream manifest from my understanding reached GA so I think we are fine here to set it stable.

pkg/model/components/kubedns.go
mazzy89: Gotcha. It makes sense.

pkg/model/components/kubelet.go
mazzy89: But how we would lead the user to set the correct address in the cluster spec? This is the address that it is recommended upstream. I believe is not an easy thing for the user to pick it up this address. They need to dig in the NodeLocal DNSCache documentation upstream. Which is your recommendation?

pkg/model/components/kubedns.go
olemarkus: Sorry, I mean these two:
```
NodeLocalDNS = &kops.NodeLocalDNSConfig{}
NodeLocalDNS.Enabled = false
```
In general there is a lot of manipulation of the spec itself here, which would unconditionally override what the users may set. I would avoid all of these.

pkg/model/components/kubelet.go
olemarkus: The spec is validated against the rules defined in these files:
https://github.com/kubernetes/kops/tree/2e0720156ba193de4aca13fcb7c4776c04daedeb/pkg/apis/kops/validation

There should be a number of examples there that you can copy.
If any of these fail, the user will get information about what spec that needs to correct, to what, and why.

pkg/model/components/kubedns.go
mazzy89: But my understanding is that this logic runs only when a user does not set anything in `NodeLocalDNS` spec so I don't get how the override should happen. Could you elaborate please?

mazzy89: fyi I'm following the pattern here https://github.com/mazzy89/kops/blob/nodelocaldnscache/pkg/model/components/kubedns.go#L37

olemarkus: I guess that is fine when the values are empty. Although it doesn't look very common to do it this way anymore. Could be that it is done this way for kube-dns either because the addon is very old or because this addon is not optional.

Either way, I am fine with this approach as long as user values are not overridden. This does make the templates more simple.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -451,6 +451,29 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 		}
 	}
 
+	// @check the node-local-dns has not been disabled
+	NodeLocalDNS := b.cluster.Spec.KubeDNS.NodeLocalDNS
+	if kubeDNS.Provider == "CoreDNS" && NodeLocalDNS.Enabled {
```
olemarkus: I get a sigsegv here. Guessing you need to check for more `nil`.
You should also have a validation check for that if `NodeLocalDNS.Enabled` is true, then `kubeDNS.Provider` must be `CoreDNS`. If not people may enable NodeLocalDNS, but still wonder why they don't get the addon.

pkg/apis/kops/validation/legacy.go
```diff
@@ -291,6 +291,31 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 					}
 				}
 			}
+
+			// @ check that NodeLocalDNS addon is configured correctly
+			if c.Spec.KubeDNS.NodeLocalDNS != nil && c.Spec.KubeDNS.NodeLocalDNS.Enabled {
+				if c.Spec.KubeDNS.Provider != "CoreDNS" {
```
mazzy89: @olemarkus I think this would be enough right? does this mean that the user must set explicitly the `Provider` field in the spec to `CoreDNS`?

olemarkus: This should work, yes.

upup/models/cloudup/resources/addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -0,0 +1,178 @@
...
+  Corefile: |
+    {{ KubeDNS.NodeLocalDNS.Domain }}:53 {
+        errors
+        cache {
+          success 9984 30
+          denial 9984 5
+        }
+        reload
+        loop
+        bind {{ default "169.254.20.10" KubeDNS.NodeLocalDNS.LocalIP }} {{ KubeDNS.NodeLocalDNS.ServerIP }}
```
olemarkus:
```suggestion
        bind {{- or .KubeDNS.NodeLocalDNS.LocalIP "169.254.20.10"}} {{ .KubeDNS.NodeLocalDNS.ServerIP }}
```
`default` doesn't exist. The above should do what you intend here.

olemarkus:
The nodes start now, but ends up in this error:
`plugin/forward: /etc/Corefile:10 - Error during parsing: Wrong argument count or unexpected line ending after '.'`

Is there any way of testing if this works? Like prometheus counters or something?

mazzy89:
I guess you are using it in `iptables` mode. in that case then `ClusterIP` is an empty string and the template then is probably parsed with an empty space which cause the parsing failing.

I guess then we need to check if the `ClusterIP` is set.

mazzy89:
Let me wire here a change and see if it works.

olemarkus:
Now I get 
```
error building tasks: error reading manifest addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml: error opening resource: error executing resource template "addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml": error parsing template "addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml": template: addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml:75: unexpected {{else}}
```
Could you maybe try to deploy a cluster with this before submitting the next commits?

mazzy89:
> Now I get 
> 
> ```
> 
> error building tasks: error reading manifest addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml: error opening resource: error executing resource template "addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml": error parsing template "addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml": template: addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml:75: unexpected {{else}}
> 
> ```
> 
> Could you maybe try to deploy a cluster with this before submitting the next commits?

Unfortunately dont have anywhere where to test this change quickly

olemarkus:
> > Now I get
> > ```
> > 
> > error building tasks: error reading manifest addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml: error opening resource: error executing resource template "addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml": error parsing template "addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml": template: addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml:75: unexpected {{else}}
> > ```
> > 
> > 
> > Could you maybe try to deploy a cluster with this before submitting the next commits?
> 
> Unfortunately dont have anywhere where to test this change quickly

I can take a crack at this. I'll send a patch as soon as I can.
But it would be helpful to understand how I can test that this is working as intended

mazzy89:
> But it would be helpful to understand how I can test that this is working as intended

In order to verify this we need to test if a consequent request that leave a node is then served without hitting the CoreDNS pods but passing through the local.

I'm not sure if there are any Prometheus metric for that but the fastest thing it comes to my mind is enabling just for test purpose the log plugin in the `Corefile` and check if it works.

mazzy89:
@olemarkus fixed the template btw.

mazzy89:
@olemarkus As discussed, fixed the template following the upstream config in case of `iptables`

> __PILLAR__CLUSTER__DNS__ and __PILLAR__UPSTREAM__SERVERS__ will be populated by the node-local-dns pods.

Only doubt now is if `Corefile` is parsed correctly in the case of `ipvs` when `ServerIP` must be empty.

olemarkus:
I tested this running cilium. I am guessing cilium ignores the fact that nodelocaldns listens on the kube-dns IP. I had to set `kubelet.clusterDNS: 169.254.20.10` to get this to work.
I am not sure which CNIs you have tested if it works better there.

mazzy89:
> I had to set kubelet.clusterDNS: 169.254.20.10 to get this to work.

this is the case when kube-proxy is running in `ipvs` mode. is that your case?

olemarkus:
kube-proxy is running in iptables mode. but cilium is a bit special, and in many cases, it ignores kube-proxy alltogether. I did not change anything in the templates though, only set the kubelet-config.

mazzy89:
So would you recommend also to set in the validation to setup `ClusterDNS` in case that the CNI is Cilium, right?

olemarkus:
Yep. I tested with Calico and it works fine there. So adding a validation check for cilium should do the trick.

mazzy89:
Cool. It make sense. Can I also ask you if you can test ipvs mode? I'm not sure if in that case the Corefile produced is valid.

olemarkus:
With calico in and kubeproxy in iptables mode, it works as expected.
In ipvs mode, I get no warning about not having set `kubelet.clusterDNS`, so it doesn't work without further guidance. The templates looks fine, I think. At least node-local-dns isn't crashing.

I'll test with setting clusterdns tomorrow. I suspect you need to create some exceptions around line 285 in legacy.go

olemarkus:
Managed to test now after all. If one sets `kubelet.clusterDNS` then ipvs works just fine. So you should add that to the same validation logic as for cilium.

You should also bypass the featureflag if nodelocaldns is enabled, I think. It isn't that experimental then.

mazzy89:
Added validation for Cilium. Not sure why all the jobs are now failing after having rebased against master. need to find out

olemarkus:
Just run `make test` to see what fails

rifelpet:
the `pkg/apis/kops/v1alpha1` directory was removed in a PR that merged yesterday, yet your commits are adding files to that directory. remove those additions and we'll see if the jobs pass again.

pkg/apis/kops/validation/legacy.go
```diff
@@ -291,6 +291,36 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+					allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubeDNS", "provider"), "KubeDNS provider must be set to CoreDNS if NodeLocalDNS addon is enabled"))
+				}
+
+				if c.Spec.KubeDNS.NodeLocalDNS.LocalIP == "" {
+					if c.Spec.Kubelet != nil && c.Spec.Kubelet.ClusterDNS == "" {
```
olemarkus: Looks like at this point `c.Spec.Kubelet.ClusterDNS` is set to `100.64.0.10` so one will never enter this block.
I think a better check is `c.Spec.Kubelet.ClusterDNS != c.Spec.KubeDNS.NodeLocalDNS.LocalIP`.

mazzy89:
Just noticed now that I did a bad rebase.

pkg/apis/kops/validation/legacy.go
mazzy89: > Looks like at this point c.Spec.Kubelet.ClusterDNS is set to 100.64.0.10 so one will never enter this block.

Could you elaborate here the why?

Also I do already that check here https://github.com/kubernetes/kops/pull/8780/files/429f0fec9701236f79d9608506b1af91b1709a28#diff-33635946dcf9aa666832a9236f8f3349R318

olemarkus: I didn't look much into the details. I just noticed that I did not get the expected error when using cilium and not changing kubelet.ClusterDNS.

mazzy89: But yeah it makes sense somewhere `100.64.0.10` it is set up. I've run a global lookup of it in the codebase and just found that it is generated as output. So yeah somewhere is set.

mazzy89: I've updated here the logic.

pkg/apis/kops/validation/legacy.go
```diff
@@ -291,6 +291,36 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
 			}
+
+			// @ check that NodeLocalDNS addon is configured correctly
+			if c.Spec.KubeDNS.NodeLocalDNS != nil && c.Spec.KubeDNS.NodeLocalDNS.Enabled {
+				if c.Spec.KubeDNS.Provider != "CoreDNS" {
+					allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubeDNS", "provider"), "KubeDNS provider must be set to CoreDNS if NodeLocalDNS addon is enabled"))
+				}
+
+				if c.Spec.KubeDNS.NodeLocalDNS.LocalIP == "" {
+					if c.Spec.Kubelet != nil && c.Spec.Kubelet.ClusterDNS != c.Spec.KubeDNS.NodeLocalDNS.LocalIP {
```
mazzy89: @olemarkus need to change this. Just realized that this does not make the sense and breaks also the case for `iptables`.

mazzy89: Improved entire validation logic.

mazzy89: Writing test for the `validateNodeLocalDNS` func

mazzy89:
Added unit tests here. It should now be easier to tests and make sure that the different scenarios are now covered.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -97,6 +97,21 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
...
 
+	dest["NodeLocalDNSClusterIP"] = func() string {
+		if tf.cluster.Spec.KubeProxy.ProxyMode == "ipvs" {
+			return tf.cluster.Spec.KubeDNS.ServerIP
+		} else {
+			return "__PILLAR__CLUSTER__DNS__"
+		}
+	}
+	dest["NodeLocalDNSServerIP"] = func() string {
+		if tf.cluster.Spec.KubeProxy.ProxyMode == "ipvs" {
```
mazzy89: Typo here. this is supposed to be `iptables.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -97,6 +97,21 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		return tf.cluster.Spec.KubeDNS
 	}
 
+	dest["NodeLocalDNSClusterIP"] = func() string {
+		if tf.cluster.Spec.KubeProxy.ProxyMode == "ipvs" {
```
mazzy89: Typo here. This is supposed to be `iptables.`

mazzy89: sorry my bad. it is actually right.

mazzy89:
Ready here to be picked up for a review.

123BLiN:
Sorry to bother, but will it be possible to use this with Kops 1.15/1.16 and kubernetes v1.15 or this change is only for furutre 1.18 release?

mazzy89:
At the moment the change has been planned for kops 1.18

pkg/apis/kops/validation/legacy.go
```diff
@@ -277,14 +277,23 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
-						allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubeDNS", "serverIP"), "MasterKubelet ClusterDNS did not match cluster kubeDNS.serverIP"))
+					if isExperimentalClusterDNS(c.Spec.MasterKubelet, c.Spec.KubeDNS) {
+						allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("masterKubelet", "clusterDNS"), "MasterKubelet ClusterDNS did not match cluster kubeDNS.serverIP or nodeLocalDNS.localIP"))
 					}
 				}
 			}
+
+			// @ check that NodeLocalDNS addon is configured correctly
+			if c.Spec.KubeDNS.NodeLocalDNS != nil && c.Spec.KubeDNS.NodeLocalDNS.Enabled {
+				if c.Spec.KubeDNS.Provider != "CoreDNS" {
```
justinsb, line 291: Do we actually need the "raw" DNS provider to be CoreDNS?  I didn't think it was required?

pkg/apis/kops/validation/legacy.go
```diff
@@ -705,3 +714,9 @@ func validateKubelet(k *kops.KubeletConfigSpec, c *kops.Cluster, kubeletPath *fi
 	}
 	return allErrs
 }
+
+func isExperimentalClusterDNS(k *kops.KubeletConfigSpec, dns *kops.KubeDNSConfig) bool {
```
justinsb, line 718: Nit: I'm not clear why this function is called "isExperimentalClusterDNS".  Maybe `isValidClusterDNS` ?

pkg/apis/kops/cluster.go
```diff
@@ -400,6 +400,16 @@ type KubeDNSConfig struct {
...
 	// MemoryLimit specifies the memory limit of each dns container in the cluster. Default 170m.
 	MemoryLimit *resource.Quantity `json:"memoryLimit,omitempty"`
+	// NodeLocalDNS specifies the configuration for the node-local-dns addon
+	NodeLocalDNS *NodeLocalDNSConfig `json:"nodeLocalDNS,omitempty"`
+}
+
+// NodeLocalDNSConfig are options of the node-local-dns
+type NodeLocalDNSConfig struct {
+	// Disable indicates we do not wish to run the node-local-dns addon
+	Enabled bool `json:"enabled,omitempty"`
```
justinsb, line 410: I think we actually should make this `Enabled *bool`, because we want to differentiate between:

```
nodeLocalDNSConfig:
  enabled: false
  localIP: 10.0.0.1
```

and 

```
nodeLocalDNSConfig:
  localIP: 10.0.0.1
```

justinsb:
Thanks @mazzy89 

A few small nits, and the `enabled` one is actually pretty important I think.  But we can also do them in follow on PRs, so...

/approve
/lgtm

pkg/apis/kops/validation/legacy.go
olemarkus: It is just because when this one returns true, one needs the ExperimentalClusterDNS featureflag on. I am not really sure this feature flag makes sense. As in it doesn't really guard a feature.

One option would certainly be to remove the feature flag and only consider "valid" configurations.

pkg/apis/kops/cluster.go
olemarkus: I suppose this could be a change we do if we consider enabling this by default. If not, we don't really need to distinguish between the two now.

PR #8782, opened 2020-03-22
Set default log level to "info" for containerd

hakman:
"info" is the default log level for containerd and also very useful compared to "warn", but still doesn't create a huge log file.


rifelpet:
we're trying to troubleshoot some failing periodic e2e jobs, hoping the increased verbosity will reveal any issues.

PR #8783, opened 2020-03-23
Remove basic authentication support for k8s 1.19+

hakman:
Support for basic auth was removed in Kubernetes 1.19: https://github.com/kubernetes/kubernetes/pull/89069.

Kops e2e test are failing since this was merged. @rifelpet and me tracked this issue and tested that `kube-apiserver` can start now.
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-updown


rifelpet:
Adding some context:

The updown prow job was failing because the kube-apiserver pod was in a crashloop. There were no logs in /var/log/kube-apiserver.log so the job artifact was empty. I manually launched a cluster using the same `create cluster` command and `docker logs` revealed `Error: unknown flag: --basic-auth-file` which lead us to the k/k commit @hakman linked to above.

Kops creates basic auth credentials for each cluster (visible from a `kops export kubecfg`). These will no longer work in k8s 1.19.

We came up with a few options on how to proceed:

1. Switch the prow jobs to use the latest k8s 1.18 rather than 1.19. This is relevant because kops' master branch is still targeting 1.18. Once we create the release-1.18 branch, master will target 1.19 and we can merge any fixes and switch the jobs back to target k8s 1.19. This means that any fixes for this issue will land in kops 1.18 which may or may not be desirable. For precedence, we did this with the kubelet node labeling issue in 1.16 before kops-controller was introduced.
2. Merge this PR now as well as anything else required to drop basic auth support, guarding all the logic with checks for k8s >=1.19.


My main concern with this dropping of support is providing any sort of workaround for users of basic auth, if possible. Since kops creates these credentials, we'll need to make it known that they will no longer work, as well as no longer create these credentials for new clusters (or even delete them for existing clusters? possible downgrade concerns there...). The k/k commit mentions `--token-auth-file` which Kops [does support](https://github.com/kubernetes/kops/blob/62c51002ba69aaafeff38191ad02a8a7b6a2ba1d/nodeup/pkg/model/kube_apiserver.go#L295) but it seems to use a [deprecated function](https://github.com/kubernetes/kops/blob/62c51002ba69aaafeff38191ad02a8a7b6a2ba1d/nodeup/pkg/model/secrets.go#L128-L169) as well.

Anyone else have any thoughts or suggestions? This is blocking a lot of visibility into our e2e jobs. I'll add a note to discuss this at office hours as well.

rifelpet:
additionally, it'd be great to find out why the /var/log/kube-apiserver.log files were empty. My guess is that the container exited before `tee` had a chance to write the contents to the file. Maybe this is an opportunity to discontinue use of `tee` and instead just rely on `docker logs` or something else? We'd have to update kubetest to get these logs as well. I know it was discussed a while ago but I can't remember where or any decisions about it.

nodeup/pkg/model/kube_apiserver.go
```diff
@@ -294,8 +294,16 @@ func (b *KubeAPIServerBuilder) buildPod() (*v1.Pod, error) {
 	kubeAPIServer.TLSPrivateKeyFile = filepath.Join(b.PathSrvKubernetes(), "server.key")
 	kubeAPIServer.TokenAuthFile = filepath.Join(b.PathSrvKubernetes(), "known_tokens.csv")
 
-	if !kubeAPIServer.DisableBasicAuth {
-		kubeAPIServer.BasicAuthFile = filepath.Join(b.PathSrvKubernetes(), "basic_auth.csv")
+	// Support for basic auth was deprecated 1.16 and removed in 1.19
+	// https://github.com/kubernetes/kubernetes/pull/89069
+	if b.IsKubernetesLT("1.18") {
+		if (kubeAPIServer.DisableBasicAuth != nil && !*kubeAPIServer.DisableBasicAuth) || kubeAPIServer.DisableBasicAuth == nil {
```
johngmyers, line 300:
```suggestion
		if kubeAPIServer.DisableBasicAuth == nil || !*kubeAPIServer.DisableBasicAuth {
```

hakman: Thanks :)

rifelpet:
Nice, I like this compromise of disabling by default for one k8s version before removing

/lgtm
/approve

etwillbefine:
> My main concern with this dropping of support is providing any sort of workaround for users of basic auth, if possible. Since kops creates these credentials, we'll need to make it known that they will no longer work, as well as no longer create these credentials for new clusters

So I run into exactly this issue. I was using basic credentials just because kops needed to. The security docs still referring to the basic credentials. https://github.com/kubernetes/kops/blob/master/docs/security.md My kubecfg from a 1.16 cluster still contains basic credentials. It actually contains 2 users: 1 with basic auth only and one with cert-data, key and user/password. Access is denied using that kubecfg (Unauthorized). The only way I have now to get access at all to the cluster via CI is to create an OIDC user for CI. This is not my preferred way of authenticating "master CI user" as the same user database will contain the users.

PR #8787, opened 2020-03-23
Support S3 Virtual Hosted Style

johanneswuerbach:
Support S3 URLs written in Virtual Hosted Style, which seems to be only recommended notation nowadays, but isn't supported in kops.

https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#virtual-host-style-url-ex

All the other options are more or less deprecated path style https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#path-style-access, old virtual styles https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingBackwardsCompatibility

### DIFF
--- a/util/pkg/vfs/s3context.go
+++ b/util/pkg/vfs/s3context.go
@@ -42,7 +42,7 @@ var (
 	// https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region
 	// TODO: perhaps make region regex more specific, i.e. (us|eu|ap|cn|ca|sa), to prevent matching bucket names that match region format?
 	//       but that will mean updating this list when AWS introduces new regions
-	s3UrlRegexp = regexp.MustCompile(`s3([-.](?P<region>\w{2}-\w+-\d{1})|[-.](?P<bucket>[\w.\-\_]+)|)?.amazonaws.com(.cn)?(?P<path>.*)?`)
+	s3UrlRegexp = regexp.MustCompile(`(s3([-.](?P<region>\w{2}-\w+-\d{1})|[-.](?P<bucket>[\w.\-\_]+)|)?|(?P<bucket>[\w.\-\_]+).s3.(?P<region>\w{2}-\w+-\d{1})).amazonaws.com(.cn)?(?P<path>.*)?`)
 )
 
 type S3BucketDetails struct {
@@ -385,7 +385,9 @@ func VFSPath(url string) (string, error) {
 
 	captured := map[string]string{}
 	for i, value := range result {
-		captured[groupNames[i]] = value
+		if value != "" {
+			captured[groupNames[i]] = value
+		}
 	}
 	bucket := captured["bucket"]
 	path := captured["path"]
--- a/util/pkg/vfs/s3context_test.go
+++ b/util/pkg/vfs/s3context_test.go
@@ -80,6 +80,11 @@ func Test_VFSPath(t *testing.T) {
 			ExpectError:    false,
 		},
 		{
+			Input:          "https://bucket-name.s3.us-east-1.amazonaws.com/path",
+			ExpectedResult: "s3://bucket-name/path",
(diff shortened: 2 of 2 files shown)


util/pkg/vfs/s3context.go
```diff
@@ -42,7 +42,7 @@ var (
 	// https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region
 	// TODO: perhaps make region regex more specific, i.e. (us|eu|ap|cn|ca|sa), to prevent matching bucket names that match region format?
 	//       but that will mean updating this list when AWS introduces new regions
-	s3UrlRegexp = regexp.MustCompile(`s3([-.](?P<region>\w{2}-\w+-\d{1})|[-.](?P<bucket>[\w.\-\_]+)|)?.amazonaws.com(.cn)?(?P<path>.*)?`)
+	s3UrlRegexp = regexp.MustCompile(`(s3([-.](?P<region>\w{2}-\w+-\d{1})|[-.](?P<bucket>[\w.\-\_]+)|)?|(?P<bucket>[\w\-\_]+).s3.(?P<region>\w{2}-\w+-\d{1})).amazonaws.com(.cn)?(?P<path>.*)?`)
```
johngmyers: Why excluding '.' in bucket names in virtual hosted style?

johanneswuerbach: Having a dot in the bucket name causes https connection to fail:

https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules

> When you use virtual hosted–style buckets with Secure Sockets Layer (SSL), the SSL wildcard certificate only matches buckets that don't contain periods. To work around this, use HTTP or write your own certificate verification logic. We recommend that you do not use periods (".") in bucket names when using virtual hosted–style buckets.

While mirroring assets should work (as kops uses an aws sdk s3 client), the user data script is likely to fail as kops tries to fetch assets from `https://bucket`

johngmyers: That would be a reason to avoid writing out virtual hosted style URIs for such buckets, but doesn't seem a reason to avoid recognizing such URIs.

justinsb: Ah - there's actually a bug underlying this difference of opinion :-)  VFSPath is used in two different places for two quite different things.

Can I suggest that we rename VFSPath to S3URLForHTTPURL or similar? (If you can find something that doesn't look like shouting in the go naming rules, that's even better!)

I started going down this little rabbit hole and it looks like we actually need a bigger refactor... I think one of the usages (storage.go) is really `ParseS3Path` and the other one (copyfile.go) is really MapToUploadLocation (and should probably just be inlined in that function.

I'm happy to do that refactor if you want, but I don't want to "steal" the PR from you two that have actually discovered the problem :-)

johanneswuerbach: @justinsb thanks for the feedback, I did a small refactoring, but not entirely sure whether it goes in the imaging direction.

johanneswuerbach: Scratch that, I reverted my attempt as I was starting to duplicate a lot of logic and in the end those functions are fairly similar.

Instead I added the `.` as asked for by @johngmyers and as a customer your could also specify `http://my-bucket`, which would still work, but is less secure obviously.

johngmyers:
/lgtm
Someone with repo access will need to retry the failed Travis job, or you could rebase.

geojaz:
i poked travis for ya

johngmyers:
Oh, that Travis failure looks like an actual Go 1.12 incompatibility.

Please rebase, as the Travis config has changed to no longer run on Go 1.12.

johanneswuerbach:
@johngmyers rebased :-)

PR #8790, opened 2020-03-24
Upgrade to bazel 2.2.0

rifelpet:
The kubekins-e2e images [were upgraded](https://github.com/kubernetes/test-infra/pull/16910) to bazel 2.2.0 so we need to upgrade to match. This should fix the [bazel version errors](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kops/8789/pull-kops-e2e-kubernetes-aws/1242492419203665921) we're seeing in presubmit jobs

Luckily this seems uninvasive so we'll be able to cherry-pick back as necessary.

once this is merged we can update the rules to their most recent versions too.


rifelpet:
FYI the .bazelversion file wasnt added until after release-1.17 was cut, so we dont need to cherry-pick this back. Though the prow jobs of any PRs that target release branches will still use bazel 2.2.0. since there werent any other changes needed to support the new version, I dont think any action needs to be taken on the release branches.

PR #8798, opened 2020-03-25
Update default users for kubeconfig with supported distros

hakman:
Kubernetes 1.19 stoped defaulting kubeconfig to http://localhost:8080 and needs a kubeconf in the home dir for kubectl to work by default.
https://github.com/kubernetes/kubernetes/pull/86173

### DIFF
--- a/nodeup/pkg/model/kubectl.go
+++ b/nodeup/pkg/model/kubectl.go
@@ -106,8 +106,12 @@ func (b *KubectlBuilder) findKubeconfigUser() (*fi.User, *fi.Group, error) {
 	switch b.Distribution {
 	case distros.DistributionJessie, distros.DistributionDebian9, distros.DistributionDebian10:
 		users = []string{"admin", "root"}
-	case distros.DistributionCentos7:
+	case distros.DistributionXenial, distros.DistributionBionic, distros.DistributionFocal:
+		users = []string{"ubuntu"}
+	case distros.DistributionCentos7, distros.DistributionCentos8:
 		users = []string{"centos"}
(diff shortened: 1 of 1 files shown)


nodeup/pkg/model/kubectl.go
```diff
@@ -106,8 +106,12 @@ func (b *KubectlBuilder) findKubeconfigUser() (*fi.User, *fi.Group, error) {
 	switch b.Distribution {
 	case distros.DistributionJessie, distros.DistributionDebian9, distros.DistributionDebian10:
 		users = []string{"admin", "root"}
-	case distros.DistributionCentos7:
+	case distros.DistributionXenial, distros.DistributionBionic, distros.DistributionFocal:
+		users = []string{"ubuntu"}
+	case distros.DistributionCentos7, distros.DistributionCentos8:
 		users = []string{"centos"}
+	case distros.DistributionAmazonLinux2, distros.DistributionRhel7, distros.DistributionRhel8:
+		users = []string{"ec2-user"}
```
rifelpet: Hm this logic is used on every cloud provider. AL2 makes sense to use ec2-user but RHEL7 and RHEL8 might have different users depending on the cloud provider. Maybe we add a conditional for `aws` here? though it would be good to cover other cloud providers as well.

hakman: I think it has multiple choices, so should be ok. Maybe someone knows what to add for other cloud providers for centos.

rifelpet: ah you're right. we can encourage users of other cloud providers to submit additions in separate PRs

PR #8799, opened 2020-03-26
Deprecate support for Kubernetes 1.10

johngmyers:
Per our deprecation policy, I had deprecated Kubernetes 1.9 in Kops 1.18, scheduling for removal int Kops 1.19. This is the minimum necessary to avoid increasing the supported range.

This PR proposes to also deprecate Kubernetes 1.10, removing support in Kops 1.19. This narrows the supported range by one version.

### DIFF
--- a/docs/releases/1.18-NOTES.md
+++ b/docs/releases/1.18-NOTES.md
@@ -74,7 +74,7 @@
 
 # Deprecations
 
-* Support for Kubernetes version 1.10 is deprecated and will be removed in kops 1.19.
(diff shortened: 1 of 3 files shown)


rifelpet:
This is meant to be in Kops 1.19 correct? If that's the case we'll need to wait until the release-1.18 branch is created before we merge this into master

/hold

johngmyers:
@rifelpet This is needed in kops 1.18 to announce and warn about the impending removal in kops 1.19.

rifelpet:
ah, that makes sense to me now. thanks

/approve
/hold cancel

PR #8802, opened 2020-03-26
Add support for Spot block in launch template

thejasbabu:
- Launch configuration does not support the field SpotDurationInMinutes which is used to reserve the spot instances, but however Launch Template does

### DIFF
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -18,6 +18,7 @@ package validation
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -47,6 +48,8 @@ func awsValidateInstanceGroup(ig *kops.InstanceGroup) field.ErrorList {
 	allErrs = append(allErrs, awsValidateMachineType(field.NewPath(ig.GetName(), "spec", "machineType"), ig.Spec.MachineType)...)
 
 	allErrs = append(allErrs, awsValidateAMIforNVMe(field.NewPath(ig.GetName(), "spec", "machineType"), ig)...)
+
+	allErrs = append(allErrs, awsValidateSpotDurationInMinute(field.NewPath(ig.GetName(), "spec", "spotDurationInMinutes"), ig)...)
 
 	return allErrs
 }
@@ -104,6 +107,16 @@ func awsValidateAMIforNVMe(fieldPath *field.Path, ig *kops.InstanceGroup) field.
 				}
 			}
 		}
+	}
+	return allErrs
+}
+
+func awsValidateSpotDurationInMinute(fieldPath *field.Path, ig *kops.InstanceGroup) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if ig.Spec.SpotDurationInMinutes != nil {
+		validSpotDurations := []string{"60", "120", "180", "240", "300", "360"}
+		spotDurationStr := strconv.FormatInt(*ig.Spec.SpotDurationInMinutes, 10)
+		allErrs = append(allErrs, IsValidValue(fieldPath, &spotDurationStr, validSpotDurations)...)
 	}
 	return allErrs
 }
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -131,6 +131,9 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 	if ig.Spec.MixedInstancesPolicy == nil {
 		lt.SpotPrice = lc.SpotPrice
 	}
+	if ig.Spec.SpotDurationInMinutes != nil {
+		lt.SpotDurationInMinutes = ig.Spec.SpotDurationInMinutes
+	}
 	return lt, nil
 }
 
--- a/tests/integration/update_cluster/launch_templates/in-v1alpha2.yaml
+++ b/tests/integration/update_cluster/launch_templates/in-v1alpha2.yaml
@@ -71,6 +71,8 @@ spec:
   minSize: 2
   role: Node
   instanceProtection: true
+  maxPrice: "0.1"
+  spotDurationInMinutes: 120
   subnets:
   - us-test-1b
 ---
--- a/k8s/crds/kops.k8s.io_instancegroups.yaml
+++ b/k8s/crds/kops.k8s.io_instancegroups.yaml
@@ -689,6 +689,11 @@ spec:
               description: SecurityGroupOverride overrides the default security group
                 created by Kops for this IG (AWS only).
               type: string
+            spotDurationInMinutes:
+              description: SpotDurationInMinutes indicates this is a spot-block group,
+                with the specified value as the spot reservation time
+              format: int64
+              type: integer
             subnets:
               description: Subnets is the names of the Subnets (as specified in the
                 Cluster) where machines in this instance group should be placed
--- a/pkg/apis/kops/instancegroup.go
+++ b/pkg/apis/kops/instancegroup.go
@@ -121,6 +121,8 @@ type InstanceGroupSpec struct {
 	Hooks []HookSpec `json:"hooks,omitempty"`
 	// MaxPrice indicates this is a spot-pricing group, with the specified value as our max-price bid
 	MaxPrice *string `json:"maxPrice,omitempty"`
+	// SpotDurationInMinutes reserves a spot block for the period specified
+	SpotDurationInMinutes *int64 `json:"spotDurationInMinutes,omitempty"`
 	// AssociatePublicIP is true if we want instances to have a public IP
 	AssociatePublicIP *bool `json:"associatePublicIp,omitempty"`
 	// AdditionalSecurityGroups attaches additional security groups (e.g. i-123456)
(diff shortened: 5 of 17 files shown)


hakman:
@thejasbabu Thanks for spending the time to add this. Could you address a few things:
* an integration test should be added here https://github.com/kubernetes/kops/tree/master/tests/integration/update_cluster/mixed_instances_spot (you can add a new instance group or create a new test)
* I don't see any code for CloudFormation, just for Terraform

Once you add these should be easier to see if this works correctly.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -131,6 +131,9 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 	if ig.Spec.MixedInstancesPolicy == nil {
 		lt.SpotPrice = lc.SpotPrice
 	}
+	if ig.Spec.SpotDurationInMinutes != nil {
+		lt.SpotDurationInMinutes = ig.Spec.SpotDurationInMinutes
```
gjtempleton: Should we do some validation of the value provided by the user given that these values have to be a multiple of 60? 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-instancemarketoptions-spotoptions.html

thejasbabu: Added validation for this. Thanks!

thejasbabu:
@hakman Thanks for the feedback.

- I have added the code for `CloudFormation` output
- Can you please show me some docs around how to write these integration tests?

pkg/apis/kops/validation/aws.go
```diff
@@ -107,3 +109,20 @@ func awsValidateAMIforNVMe(fieldPath *field.Path, ig *kops.InstanceGroup) field.
 	}
 	return allErrs
 }
+
+func awsValidateSpotDurationInMinute(fieldPath *field.Path, ig *kops.InstanceGroup) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if ig.Spec.SpotDurationInMinutes != nil {
+		if *ig.Spec.SpotDurationInMinutes < 60 {
+			errDetails := fmt.Sprintf("cannot be lesser than 60")
```
johngmyers:
```suggestion
			errDetails := fmt.Sprintf("cannot be less than 60")
```

pkg/apis/kops/validation/aws.go
```diff
@@ -107,3 +109,20 @@ func awsValidateAMIforNVMe(fieldPath *field.Path, ig *kops.InstanceGroup) field.
...
+			allErrs = append(allErrs, field.Invalid(fieldPath, ig.Spec.SpotDurationInMinutes, errDetails))
+		} else if *ig.Spec.SpotDurationInMinutes > 360 {
+			errDetails := fmt.Sprintf("cannot be greater than 360")
+			allErrs = append(allErrs, field.Invalid(fieldPath, ig.Spec.SpotDurationInMinutes, errDetails))
+		} else if *ig.Spec.SpotDurationInMinutes%60 != 0 {
+			errDetails := fmt.Sprintf("should be multiples of 60")
```
johngmyers:
```suggestion
			errDetails := fmt.Sprintf("must be a multiple of 60")
```

pkg/apis/kops/validation/aws.go
thejasbabu: Then should I change the greater than as well? @johngmyers

pkg/model/awsmodel/autoscalinggroup.go
thejasbabu: Added the integration test to test spotblock

rifelpet:
looks like the test is failing due to whitespace issues in the integration test expected output. You can run `hack/updated-expected.sh` to fix it

pkg/apis/kops/validation/aws.go
johngmyers: "greater than" is correct. English is indeed inconsistent.

tests/integration/update_cluster/spotblock/in-v1alpha2.yaml
```diff
@@ -0,0 +1,131 @@
...
+spec:
+  associatePublicIp: true
+  image: kope.io/k8s-1.4-debian-jessie-amd64-hvm-ebs-2016-10-21
+  machineType: t3.medium
+  maxSize: 2
+  minSize: 2
+  role: Node
+  instanceProtection: true
+  maxPrice: "0.1"
+  spotDurationInMinutes: 120
```
rifelpet: Apologies for the runaround but we're doing some significant refactoring in the integration tests (see #8737) and the timing of that and some other PRs is going to cause a merge conflict here.

I think it would be simpler to add this field to the existing `launch_templates` test cluster rather than creating a new integration test cluster. You could delete `tests/integration/update_cluster/spotblock` and revert your changes in `cmd/kops/integration_test.go`, then add the new field to `tests/integration/update_cluster/launch_templates/in-v1alpha2.yaml` and rerun `./hack/update-expected.sh` once more. It will produce a much smaller PR and wont conflict with the other changes we're making.

I hope this makes sense, let me know if you have any questions.

thejasbabu: I see that #8737 is not yet merged, will the above thing work if I rebase this branch with master?

hakman: @rifelpet meant that would be easier for him to rebase and modify #8737 if you add this instance group to the existing `launch_templates` test instead of creating a new one.

hakman: Sorry for my late reply. I've been a bit busy over the last few days.

thejasbabu: There's few complication in adding the integration test as part of `launch_templates`. The `TestLaunchTemplatesASG` fails since the method `runTestAWS` is made for one type of node, now when I add a new asg with spot-node, the actual is correct but the expected is mis formed because of `runTestAWS` method.

Should I refactor the `runTestAWS` to run for multiple nodes for single kube-cluster?

thejasbabu: Also there's an assumption that the instance group name is always `nodes` 
https://github.com/kubernetes/kops/blob/master/cmd/kops/integration_test.go#L486

hakman: In this case, could you just add these to the existing IG for `launch_templates`?
```
  maxPrice: "0.1"
  spotDurationInMinutes: 120
```

thejasbabu: Seems like a quick fix, but will do.

hakman:
Seems that tests are failing because of missing of the imports order in `aws_test.go`. You should run `make goimports` to fix it.

hakman:
Considering that the list of accepted values is pretty small, maybe we should just list the accepted values instead of this logic? What do you think @gjtempleton @johngmyers ?

https://github.com/kubernetes/kops/blob/c511921e2400e58ecd45d17f62433ade02429561/pkg/apis/kops/validation/aws.go#L115-L126

johngmyers:
An explicit list checked with `validation.IsValidValue()` would indeed be better.

thejasbabu-hotstar:
@hakman @johngmyers Made the validation just a basic check of list of values. Please let me know if there are any other changes needed to be done 👍

thejasbabu:
@hakman @johngmyers Made the basic just a basic check from value list. Let me know if there's any other changes that needs to be done 👍

pkg/apis/kops/validation/aws.go
```diff
@@ -107,3 +109,15 @@ func awsValidateAMIforNVMe(fieldPath *field.Path, ig *kops.InstanceGroup) field.
...
+
+func awsValidateSpotDurationInMinute(fieldPath *field.Path, ig *kops.InstanceGroup) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if ig.Spec.SpotDurationInMinutes != nil {
+		validSpotDurations := map[int64]struct{}{60: {}, 120: {}, 180: {}, 240: {}, 300: {}, 360: {}}
+		if _, ok := validSpotDurations[*ig.Spec.SpotDurationInMinutes]; !ok {
+			errDetails := fmt.Sprintf("must be less than 360, greater than 60 and be a multiple of 60")
+			allErrs = append(allErrs, field.Invalid(fieldPath, ig.Spec.SpotDurationInMinutes, errDetails))
+		}
+	}
```
hakman: Could you try using IsValidValue like here? It should work if you convert the value to string first.
https://github.com/kubernetes/kops/blob/03adc95c4545ccd1c7c1d565d7c3a1a0b2747294/pkg/apis/kops/validation/validation.go#L440-L443

tests/integration/update_cluster/launch_templates/in-v1alpha2.yaml
```diff
@@ -69,10 +69,11 @@ spec:
   minSize: 2
   role: Node
   instanceProtection: true
+  maxPrice: "0.1"
+  spotDurationInMinutes: 120
   subnets:
   - us-test-1b
 ---
-
```
hakman: Please add back this line to reduce the diff.

pkg/apis/kops/validation/aws.go
thejasbabu: Done ✅

tests/integration/update_cluster/launch_templates/in-v1alpha2.yaml
thejasbabu: Reverted this.

hakman:
Thanks @thejasbabu. At the moment looks good to me. Please also squash the commits and rebase. It is a pretty small change to separate into multiple commits.

thejasbabu:
@hakman I have squashed all commit to a single one and rebased with master.

rifelpet:
This looks great, thanks for sticking with it!

/approve

PR #8805, opened 2020-03-27
Automated cherry pick of #8462: Tag EBS volumes when using launch templates with AWS API #8466: Update tags support for LaunchTemplates

hakman:
Cherry pick of #8462 #8466 on release-1.17.

#8462: Tag EBS volumes when using launch templates with AWS API
#8466: Update tags support for LaunchTemplates

### DIFF
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -97,6 +97,11 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 		return nil, err
 	}
 
+	tags, err := b.CloudTagsForInstanceGroup(ig)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


rifelpet:
Just to be clear and since we dont have tests that cover these, can you list out the tags that this will add and for which resources exactly?

hakman:
@rifelpet This would add all InstanceGroup tags except for "owned", but only for the direct target (not touching TF and CF)

hakman:
This is the function that creates the tags:
https://github.com/kubernetes/kops/blob/7492c10ec60746c4f35b68c8019c41a68b384cf6/pkg/model/context.go#L184-L232

rifelpet:
ok great, yea that makes sense to me and this seems pretty safe.

/lgtm
/approve

hakman:
Thanks @rifelpet

PR #8825, opened 2020-03-31
Add Terraform 0.12 support

### LINKED ISSUE #7052: terraform 0.12 support
terraform 0.12 is out, so it would be nice to have a valid output in kops.

Currently, "terraform 0.12checklist" reports one problem:

> After analyzing this configuration and working directory, we have identified some necessary steps that we recommend you take before upgrading to Terraform v0.12:
> 
> - [ ] `resource "aws_route" "0-0-0-0--0"` has a name that is not a valid identifier.
> 
> In Terraform 0.12, resource names must start with a letter. To fix this, rename the resource in the configuration and then use `terraform state mv` to mirror that name change in the state.

rifelpet:
This adds support for HCL2 via a FeatureFlag (default: enabled)

The feature flag name is `Terraform-0.12` but I'm find with using something else if people prefer.

* `kops update cluster --target terraform` will now create .tf files in Terraform 0.12 syntax.
* `KOPS_FEATURE_FLAG=-Terraform-0.12 kops update cluster --target terraform` will need to be used to preserve the old behavior.

The git changes in `tests/integration/update_cluster` demonstrate how the .tf file will change.

I've updated the `./hack/verify-terraform.sh` script to now validate .tf files with Terraform 0.12 by default and the syntax passes validation. The Google provider does report an error regarding a missing `machine_type` but that is independent of this PR (before this PR the GCE Terraform output wouldn't even work with Terraform 0.11 because of a version requirement conflict. Kops required Google Terraform provider >= 3.0 but 3.0 requires Terraform 0.12)

[HCL2](https://pkg.go.dev/github.com/hashicorp/hcl/v2@v2.3.0?tab=overview) uses the [go-cty](https://github.com/zclconf/go-cty) dynamic type library which is why the additional field tags are added to every Terraform struct in the tasks packages.

The HCL library doesn't support writing some of the syntax easily, so much of `upup/pkg/fi/cloudup/terraform/hcl2.go` implements byte-level functionality for what Kops needs.  See https://github.com/hashicorp/hcl/issues/347 and https://github.com/hashicorp/hcl/issues/356 for examples and discussion around those limitations.

closes #7052


schollii:
So what's the status right now, ie current version does not have the above PR: can we use latest official kops 1.17 with terraform 0.12: kops will generate HCL but I think terraform 0.12 should be able to handle it, has anyone tried? I'm asking here rather than hashicorp github because I think it depends on HCL that kops generates, is it fully compliant etc. One of my colleagues was having issues with the kubernetes.tf file generated.

tests/integration/update_cluster/ha_gce/kubernetes.tf
```diff
@@ -17,555 +17,341 @@ output "region" {
...
+  version {
+    instance_template = google_compute_instance_template.nodes-ha-gce-example-com.self_link
   }
-
-  target_size = 0
+  zone = "us-test1-c"
 }
 
 resource "google_compute_instance_template" "master-us-test1-a-ha-gce-example-com" {
-  can_ip_forward = true
```
rifelpet, line 397: it looks like `google_compute_instance_template` isn't getting rendered properly. My guess is because it uses type embedding, i'll investigate further.

https://github.com/kubernetes/kops/blob/3344ff46bbdc9ebcb2675a20ba6d30a416bd3891/upup/pkg/fi/cloudup/gcetasks/instancetemplate.go#L419-L422

/hold

justinsb:
I think we can merge this, and we can look at the terraform breakage in a follow on, if it's confined to that one special case.  Worst case we just expand out the struct (or write our own tf12 renderer, now that the format is a little more reasonable?)

I verified that we didn't break tf11 output by stopping at the intermediate stage and verifying that the tests still passed.

/approve
/lgtm

rifelpet:
ok, I'm fine with that. If we dont fix GCE before the next release we can always instruct GCE users to revert back to the old behavior via feature flag. I'll try to look into it soon though

/hold cancel

PR #8826, opened 2020-04-01
Remove support for the legacy etcd provider as of k8s 1.18

johngmyers:
This proposes to remove support for the legacy etcd provider for clusters running Kubernetes 1.18 or later.

### DIFF
--- a/docs/releases/1.18-NOTES.md
+++ b/docs/releases/1.18-NOTES.md
@@ -26,6 +26,8 @@
 
 * Google API client libraries updated from v0.beta to v1.
 
+* Kops does not support the "Legacy" etcd provider for Kubernetes versions 1.18 and higher. Such clusters will need to use the default "Manager" etcd provider.
+
 # Breaking changes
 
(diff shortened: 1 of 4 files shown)


olemarkus:
I am all for getting rid of the legacy etcd provider, but isn't it better to do this based on kops version instead of k8s version? There is nothing in k8s that requires dropping legacy etcd, and as a user, it is always confusing when the k8s version is what determines what kops functionality is supported. I would rather go for a deprecation in kops 1.17 and removal in kops 1.19 or something. that will also let us get rid of the code. If we base it on k8s version, we'll be still stuck with the code "forever".

johngmyers:
This is in line with the pattern of not making substantial changes in behavior to existing clusters until those clusters upgrade Kubernetes. Eventually kops will drop support for Kubernetes 1.17 and the legacy manager code will then be removed.

johngmyers:
Conflicts with #8614 and want that PR to go in first because it is going to be cherrypicked.
/hold

rifelpet:
I see `validateEtcdVersion` still defaults to `components.DefaultEtcd2Version` if the version field isn't set. I wonder if it should be updated to 3.X but that can be its own PR.

/approve
/lgtm

rifelpet:
/retest

quay...

hakman:
quay says they fixed it :)
/retest

PR #5978, opened 2018-10-20
  multiple ssh public keys for bastion, master, nodes of kubernetes 

gitlabbin:
- get public keys via vfs
    - user data script will add the keys into each nodes

The multiple keys will apply to "masters, nodes, bastion hosts"  

 Refer this issue: https://github.com/kubernetes/kops/issues/2348

-  the single key way still keep same, no change.

### How to apply multiple public keys:
-  put multiple keys into a file `multiple.keys`:
```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDrLmD61/tnBfikoSVKlU5D2sB5nG7ifJmeEIN3hWF5rYn6QlUXgmU05A3mrde1QD1WvgR7MEZOsCYehhyHOZdBGvBcMeRYUPqdXEzXwvDgBNufR58mQ4POZZKv32Ff3IrU3dTR/YfZ/e1loRQQak8zexlEs8u2pE7WN3bovcWRSKg4ez8zO1cueaxT8uiHYkoM6BB5gIzu7tF7cmbGfj9AsvmdSgtxbQbRJl1VQeTX5TUALjPK7nXddckXRJT5JT04T9AXtZ/rXhT2QCgfqFC1fCvT1Pjg6HW8WfqPqx5Sysu1YeqTD0WH/w9HH7GPTH29dSh99/KSnu4ActmOlkWt bc@MacBook-Pro-2.local
[...]

PR #8830, opened 2020-04-02
Add support for installing ec2-instance-connect

olemarkus:
Adds a field to the instance group spec that installs the ec2-instance-connect package on supported distros.
Installing the package is the only thing needed for ec2-instance-connect to work.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -20,6 +20,18 @@ To change the SSH public key on an existing cluster:
 * `kops update cluster --yes` to reconfigure the auto-scaling groups
 * `kops rolling-update cluster --name <clustername> --yes` to immediately roll all the machines so they have the new key (optional)
 
+### EC2 Instance Connect
+
+Kops clusters on AWS can configure make use of [EC2 Instance Connect](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Connect-using-EC2-Instance-Connect.html).
+
+To enable this, add the following to the Instance Group spec.
+
+```
+spec:
+    ec2InstanceConnect: true
+```
+
+
 ## Docker Configuration
 
 If you are using a private registry such as quay.io, you may be familiar with the inconvenience of managing the `imagePullSecrets` for each namespace. It can also be a pain to use [Kops Hooks](cluster_spec.md#hooks) with private images. To configure docker on all nodes with access to one or more private registries:
--- a/k8s/crds/kops.k8s.io_instancegroups.yaml
+++ b/k8s/crds/kops.k8s.io_instancegroups.yaml
@@ -97,6 +97,10 @@ spec:
                 description: DetailedInstanceMonitoring defines if detailed-monitoring
                   is enabled (AWS only)
                 type: boolean
+              ec2InstanceConnect:
+                description: EC2InstanceConnect installs the ec2-instance-connect
+                  package.
+                type: boolean
               externalLoadBalancers:
                 description: ExternalLoadBalancers define loadbalancers that should
                   be attached to the instancegroup
--- a/nodeup/pkg/model/ec2instanceconnect.go
+++ b/nodeup/pkg/model/ec2instanceconnect.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package model
+
+import (
+	"k8s.io/klog"
+	"k8s.io/kops/nodeup/pkg/distros"
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
+)
+
+// DockerBuilder install docker (just the packages at the moment)
+type EC2InstanceConnectBuilder struct {
+	*NodeupModelContext
+}
+
+var _ fi.ModelBuilder = &EC2InstanceConnectBuilder{}
+
+// Build is responsible for installing packages
+func (b *EC2InstanceConnectBuilder) Build(c *fi.ModelBuilderContext) error {
+	if !b.InstanceGroup.Spec.EC2InstanceConnect {
+		return nil
+	}
+
+	if b.Distribution.IsUbuntu() || b.Distribution == distros.DistributionAmazonLinux2 {
+		c.AddTask(&nodetasks.Package{Name: "ec2-instance-connect"})
+	} else if b.Distribution.IsDebianFamily() {
+		c.AddTask(&nodetasks.Package{
+			Name:   "ec2-instance-connect",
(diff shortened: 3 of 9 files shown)


yurrriq:
This is more compelling to me personally than #5978, though I can imagine use cases for the latter.

docs/security.md
```diff
@@ -20,6 +20,18 @@ To change the SSH public key on an existing cluster:
...
 
+### EC2 Instance Connect
+
+Kops clusters on AWS can configure make use of [EC2 Instance Connect](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Connect-using-EC2-Instance-Connect.html).
+
+To enable this, add the following to the Instance Group spec.
+
+```
+spec:
+    ec2InstanceConnect: true
```
geojaz, line 31: I don't think we're doing a wonderful job of this at present, but I think it would be sensible to adopt a standard for cloud provider specific API fields. I'm not sure we have a great way to do this right now. 

@johngmyers @justinsb @rifelpet @hakman (a random sample of kops folks that popped up in my autocomplete :) -
Do we have a way to namespace fields like this on a node or cluster spec. Would it be at all useful to do something like `nodepool.spec.aws.aws_specific_field` or should config like this live at root?

If we had a way to namespace API fields by provider, I would prefer this field to be named something like `enableInstanceConnect` to make it clear what the `true/false` value does.

k8s/crds/kops.k8s.io_instancegroups.yaml
```diff
@@ -97,6 +97,10 @@ spec:
                 description: DetailedInstanceMonitoring defines if detailed-monitoring
                   is enabled (AWS only)
                 type: boolean
+              ec2InstanceConnect:
+                description: EC2InstanceConnect installs the ec2-instance-connect
```
geojaz, line 101: We should note that this is AWS-only, either by putting this stuff in a namespace or just saying so in the description.

docs/security.md
geojaz: I realize that we have a lot of legacy properties that strictly may need to be namespaced, but I am proposing no changes to any existing properties unless others have a strong desire to port them. I was just thinking about new, clearly cloud provider specific properties, like this. 

Alternatively, this feature sounds roughly analogous to GCP/os-login. We could theoretically do a field that was like `enableCloudProviderLogin` and support these cases and potentially others under a single flag. My money is on having provider specific flags that clearly enable or disable a cloud specific feature, for simplicity, to make it obvious that something is definitely supported or unsupported, etc.

nodeup/pkg/model/ec2instanceconnect.go
```diff
@@ -0,0 +1,53 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
geojaz, line 2: I think you can 2020 here.

nodeup/pkg/model/ec2instanceconnect.go
```diff
@@ -0,0 +1,53 @@
...
+package model
+
+import (
+	"k8s.io/klog"
+	"k8s.io/kops/nodeup/pkg/distros"
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
+)
+
+// DockerBuilder install docker (just the packages at the moment)
```
geojaz, line 26: update copypasta comment

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -207,6 +207,12 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, st
 		}
 	}
 
+	if g.Spec.EC2InstanceConnect {
```
geojaz, line 210: You also only support AWS linux and ubuntu, maybe another check is useful here

johngmyers:
I'm wondering why we need an API field for this. Is there any particular reason not to install the package if on a supported distro and the provider is AWS?

geojaz:
@johngmyers i am happy to support your idea to just make these packages available by default... I think that's totally reasonable and probably what we should do. this would conveniently avoid provider-specific config and nix basically all of the issues I had. It doesn't solve my question about how to deal with provider specific options, but that may be a good thing.

olemarkus:
My reasoning for not installing this by default is that it does changes to your ssh config and it opens up an auth method that people may want to remain closed just for the principle. I personally would certainly not be against installing this by default though.

Another solution that comes to mind is a more generic way of just listing additional packages one wants installed (without config).

I also second the proposal to have a cloud specific prefix/object for these.

pkg/apis/kops/validation/instancegroup.go
olemarkus: Checking for distro during validation is a bit tricky. One doesn't really have anything that says anything about the distro.

nodeup/pkg/model/ec2instanceconnect.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Build is responsible for installing packages
+func (b *EC2InstanceConnectBuilder) Build(c *fi.ModelBuilderContext) error {
+	if !b.InstanceGroup.Spec.EC2InstanceConnect {
+		return nil
+	}
+
+	if b.Distribution.IsUbuntu() || b.Distribution == distros.DistributionAmazonLinux2 {
+		c.AddTask(&nodetasks.Package{Name: "ec2-instance-connect"})
+	} else if b.Distribution.IsDebianFamily() {
```
olemarkus, line 41: This one is a bit flaky. I need to remove support for Debian unless we host this deb somewhere else. Ubuntu is quite fond of breaking deb paths, it seems.

johngmyers:
People who run the AWS cloud provider have little choice but to trust AWS infrastructure. The only reason not to install the package on AWS (in a supported distribution) would be if it changes the SSH config as exposed to clients other than the ec2-instance-connect ones in undesired ways.

So if we do expose an option, we'd want the default to install the package for the AWS cloud provider and do nothing for other cloud providers.

We can't bake the package into our AMIs without going to cloud-provider-specific AMIs and the only other option I can think of to keep kops from installing a package would be to require folks to make a private fork. So it depends on the details of what the package does to the ssh server config.

Currently the cloud provider options are put in `spec.cloudConfig`. They're currently all mashed together. I have in mind a project to create a v1beta1 API in kops 1.20 and start refactoring things then, but that would be a different ticket.

rifelpet:
We already have cloud-provider-specific AMIs don't we? [example](https://github.com/kubernetes-sigs/image-builder/blob/01913c8ba3d6afd2ea58ae752ccd4b8105122ca4/images/kube-deploy/imagebuilder/templates/1.17-stretch.yml#L110-L113).

The [introductory blog post](https://aws.amazon.com/about-aws/whats-new/2019/06/introducing-amazon-ec2-instance-connect/) mentions there is no cost so that wouldn't be an issue.

It seems the only downside would be if the installation of the package enables a behavior that cluster admins didn't anticipate. Perhaps IAM users with permissive IAM policies could now be able to ssh into the instances even though the cluster admins didn't intend for that to be allowed previously. Maybe a prominent release note that mentions the package being installed and a review of IAM policies that allow [ec2-instance-connect:SendSSHPublicKey](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-connect-set-up.html#ec2-instance-connect-configure-IAM-role) would be sufficient.

johngmyers:
I think a prominent release note would be sufficient.

olemarkus:
I'll move this package into regular package task for only amazon linux and ubuntu then

hakman:
@olemarkus Can you check if by any chance the package is installed in Ubuntu by default? Seems to be in my 20.04 setup without doing anything.

olemarkus:
Yeah. The 20.04 images have this installed. So can probably kill this PR

PR #8833, opened 2020-04-02
Make kubelet topology-manager-policy configurable

olemarkus:


k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -2069,6 +2069,10 @@ spec:
                 tlsPrivateKeyFile:
                   description: 'TODO: Remove unused TLSPrivateKeyFile'
                   type: string
+                topologyManagerPolicy:
+                  description: TopologyManagerPolicy determines the policy for the
+                    topology manager
```
hakman, line 2074:
```suggestion
                  description: TopologyManagerPolicy determines the allocation policy for the
                    Topology Manager
```

olemarkus: Hehe. Thanks :)

pkg/apis/kops/validation/legacy.go
```diff
@@ -731,3 +671,43 @@ func DeepValidate(c *kops.Cluster, groups []*kops.InstanceGroup, strict bool) er
...
+
+		if k.BootstrapKubeconfig != "" {
+			if c.Spec.KubeAPIServer == nil {
+				allErrs = append(allErrs, field.Required(kubeletPath.Root().Child("spec").Child("kubeAPIServer"), "bootstrap token require the NodeRestriction admissions controller"))
+			}
+		}
+
+		if k.TopologyManagerPolicy != "" {
+			allErrs = append(allErrs, IsValidValue(kubeletPath.Child("topologyManagerPolicy"), &k.TopologyManagerPolicy, []string{"none", "best-effort", "restricted", "single-numa-node"})...)
+			if !c.IsKubernetesGTE("1.18") {
```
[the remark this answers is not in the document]
olemarkus: IsKubernetesLT doesn't exist, though.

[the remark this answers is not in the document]
hakman: Hmm... I thought I added it recently.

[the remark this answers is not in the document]
hakman: You are right, different code region. Not added here yet. Please ignore this.

k8s/crds/kops.k8s.io_clusters.yaml
hakman: I think this should have been changed in the 2 `componentconfig.go` files. At the moment seems reverted.

olemarkus: yeah. I made a mistake there. Should be better now.

hakman: Thanks :)

pkg/apis/kops/validation/legacy.go
```diff
@@ -731,3 +671,43 @@ func DeepValidate(c *kops.Cluster, groups []*kops.InstanceGroup, strict bool) er
...
+		if c.IsKubernetesGTE("1.10") {
+			// Flag removed in 1.10
+			if k.RequireKubeconfig != nil {
+				allErrs = append(allErrs, field.Forbidden(
+					kubeletPath.Child("requireKubeconfig"),
+					"require-kubeconfig flag was removed in 1.10.  (Please be sure you are not using a cluster config from `kops get cluster --full`)"))
+			}
+		}
+
+		if k.BootstrapKubeconfig != "" {
```
rifelpet, line 698: It seems that this was only validated on `Spec.Kubelet` but will now be validated on `Spec.MasterKubelet` as well. I think that should be safe but its probably worth double checking.

olemarkus: I should have commented on that. Given that any flag added to `kubelet` is also added to `masterKubelet` I figured that this is just the lack of copy/paste. I don't see a reason why you wouldn't want to add the flag if it was only explicitly added to masters, but not when they were added to _both_ workers and masters.

johngmyers:
Would have been easier to review if the refactor and the addition were in separate commits.

pkg/apis/kops/validation/legacy.go
```diff
@@ -731,3 +671,43 @@ func DeepValidate(c *kops.Cluster, groups []*kops.InstanceGroup, strict bool) er
...
+		{
+			// Flag removed in 1.6
+			if k.APIServers != "" {
+				allErrs = append(allErrs, field.Forbidden(
+					kubeletPath.Child("apiServers"),
+					"api-servers flag was removed in 1.6"))
+			}
+		}
+
+		if c.IsKubernetesGTE("1.10") {
```
johngmyers, line 689: This will panic if the version in the cluster spec is unparseable. Not something that should happen in the validation code. It should keep the old logic of checking `kubernetesRelease`.

olemarkus: Good catch. I think that getting this far with an unparseable kubernetes version is undefined behaviour though. While I like outputting all errors (or as many as possible) in one go, I think an unparseable kubernetes version should just return the error immediately.

johngmyers: I wouldn't agree it counts as "undefined behavior". It is, admittedly, a hard thing to unit test for. I could see us deciding that an unparseable Kubernetes version is one error we don't try to combine with others.

rifelpet: Given how much validation logic depends on a parseable Kubernetes version, I would be in favor of returning a single error early if it isn't parseable.

olemarkus: If I read the code correctly, the old code just continued on with 1.15 as the kubernetes version. Changed this to returning early now.

johngmyers: I suppose another approach would be to DeepCopy the Cluster and put 1.15 in the copy's KubernetesVersion.

pkg/apis/kops/validation/legacy.go
```diff
@@ -40,21 +40,15 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 	fieldSpec := field.NewPath("spec")
 	allErrs := field.ErrorList{}
 
-	// kubernetesRelease is the version with only major & minor fields
-	// We initialize to an arbitrary value, preferably in the supported range,
-	// in case the value in c.Spec.KubernetesVersion is blank or unparseable.
-	kubernetesRelease := semver.Version{Major: 1, Minor: 15}
-
 	// KubernetesVersion
+	// This is one case we return error because a large part of the rest of the validation logic depend on a valid kubernetes version.
```
johngmyers, line 44:
```suggestion
	// This is one case we return the error because a large part of the rest of the validation logic depends on a valid kubernetes version.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -196,6 +196,8 @@ type KubeletConfigSpec struct {
 	RegistryPullQPS *int32 `json:"registryPullQPS,omitempty" flag:"registry-qps"`
 	//RegistryBurst Maximum size of a bursty pulls, temporarily allows pulls to burst to this number, while still not exceeding registry-qps. Only used if --registry-qps > 0 (default 10)
 	RegistryBurst *int32 `json:"registryBurst,omitempty" flag:"registry-burst"`
+	//TopologyManagerPolicy determines the allocation policy for the topology manager
```
johngmyers, line 199:
```suggestion
	//TopologyManagerPolicy determines the allocation policy for the topology manager.
```

pkg/apis/kops/v1alpha2/componentconfig.go
```diff
@@ -196,6 +196,8 @@ type KubeletConfigSpec struct {
 	RegistryPullQPS *int32 `json:"registryPullQPS,omitempty" flag:"registry-qps"`
 	//RegistryBurst Maximum size of a bursty pulls, temporarily allows pulls to burst to this number, while still not exceeding registry-qps. Only used if --registry-qps > 0 (default 10)
 	RegistryBurst *int32 `json:"registryBurst,omitempty" flag:"registry-burst"`
+	//TopologyManagerPolicy determines the allocation policy for the topology manager
```
johngmyers, line 199:
```suggestion
	//TopologyManagerPolicy determines the allocation policy for the topology manager.
```

Issue #4037, opened 2017-12-11
Canal on K8s v1.8 with chain append mode no longer respects networkpolicies


KashifSaadat:
The information around this issue is documented here: https://github.com/projectcalico/canal/issues/115

Depending on the outcome from the above issue, we will likely need updates to the following files:
- [k8s-1.8.yaml.template](https://github.com/kubernetes/kops/blob/1.8.0/upup/models/cloudup/resources/addons/networking.projectcalico.org.canal/k8s-1.8.yaml.template)
- [k8s-1.6.yaml.template](https://github.com/kubernetes/kops/blob/1.8.0/upup/models/cloudup/resources/addons/networking.projectcalico.org.canal/k8s-1.6.yaml.template)
- [pre-k8s-1.6.yaml.template](https://github.com/kubernetes/kops/blob/1.8.0/upup/models/cloudup/resources/addons/networking.projectcalico.org.canal/pre-k8s-1.6.yaml.template)

---

### Versions
* Calico version: v2.6.2
* Flannel version: v0.9.1 (the issue was introduced in this version, v0.9.0 works fine)
* Kops version: v1.8.0
* Operating System and version: CoreOS stable v1520.8.0

### Cluster Spec (networking block)
```yaml
  networking:
    canal:
      chainInsertMode: append
      defaultEndpointToHostAction: RETURN
      prometheusMetricsEnabled: true
```

CC @chrislovecnm

caseydavenport:
@KashifSaadat just to confirm, this is only an issue when the `chainInsertMode`  has been set to `append` - the default `insert` continues to work as expected.

KashifSaadat:
Yes I believe so, because the flannel rules are then at the bottom of the `FORWARD` chain and so the Calico rules will be processed first.

caseydavenport:
See my comment [here](https://github.com/projectcalico/canal/issues/115#issuecomment-350808410).

Essentially, Calico is behaving as expected for the given config. There's probably room for a change in flannel to play nicer in this situation, but if possible I'd recommend using the default `chainInsertMode`.

tomdee:
@KashifSaadat Could you raise an issue against flannel to make this configurable

KashifSaadat:
@tomdee sure, I've raised the following issue to track this: https://github.com/coreos/flannel/issues/938

KashifSaadat:
/remove-lifecycle stale
/lifecycle frozen

[...]

PR #8836, opened 2020-04-03
[WIP] Set FELIX_CHAININSERTMODE default to "append"

hakman:


hakman:
@KashifSaadat do you have any thoughts on making "append" the default for `FELIX_CHAININSERTMODE` in Kops?

With the current default "insert", one of the e2e tests fails:
```
[sig-network] Services should be rejected when no endpoints exist
```

KashifSaadat:
Hey @hakman, that suggests to me an error with the e2e tests / network policy misconfiguration.

Generally the default should always be left as `insert`, otherwise with other services potentially injecting iptables rules on a node you can run into a situation where calico chains are not hit, resulting in other issues (e.g. all traffic bypassing k8s network policies and accepted by default). I actually ran into this case in #4037.

hakman:
Hey @KashifSaadat:). I saw some of the discussions related to that, but didn't know the whole context.
e2e tests pass with "append", but indeed may be risky.

For reference, there are a few issues related to this failing test and thought this may be an easy fix.
https://github.com/projectcalico/calico/issues/1055
https://github.com/kubernetes/kubernetes/issues/78994

KashifSaadat:
Aha I see, thanks for the context I'll have a read through those issues to understand it better!

hakman:
No rush. Has been like that for years :).

cmd/kops/create_cluster.go
```diff
@@ -174,7 +174,7 @@ func (o *CreateClusterOptions) InitDefaults() {
 	o.Yes = false
 	o.Target = cloudup.TargetDirect
 	o.Models = strings.Join(cloudup.CloudupModels, ",")
-	o.Networking = "kubenet"
+	o.Networking = "calico"
```
johngmyers, line 177: This seems to be an unrelated change.

hakman: Yes. Pretty much the only way to test if this fixes the failing e2e test.

hakman:
Seems that it's not worth the risk. Maybe there will be a better solution in the future.
https://github.com/kubernetes/kubernetes/issues/78994#issuecomment-611225448

PR #8842, opened 2020-04-03
Update metrics-server deployment apiVersion to apps/v1

elisiano:
Since kubernetes 1.16 does not support older deployments api versions
this needs to be updated as well.

### DIFF
--- a/addons/metrics-server/README.md
+++ b/addons/metrics-server/README.md
@@ -53,8 +53,12 @@ the top-level directory of this repository:
 # Kubernetes 1.7
 $ kubectl apply -f https://raw.githubusercontent.com/kubernetes/kops/master/addons/metrics-server/v1.7.x.yaml
 
-# Kubernetes 1.8+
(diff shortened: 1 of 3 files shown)


elisiano:
I realize there are old installations in the wild, would it be better to have a separate manifest called `v1.16.x.yaml` instead?

johngmyers:
Yes, there should be a separate manifest.

elisiano:
@johngmyers done, let me know if I should address anything ellse.

johngmyers:
I think you might need to add an `addon.yaml` to the directory, specifying the `kubernetesVersion` constraints.

elisiano:
@johngmyers I've added the `addon.yaml` and fixed the `README.md` to reflect the cluster version constraints.

Please note that I'm not a user of addons, so I would need someone to verify that the file format is correct.

justinsb:
Thanks @elisiano - this looks great!

/approve
/lgtm

PR #8846, opened 2020-04-04
Cleanup after #8855

hakman:
The repo containing the binaries for Docker 1.11, 1.12 and 1.13 was shut down on March 31. This means support for older docker versions cannot be maintained.
https://www.docker.com/blog/changes-dockerproject-org-apt-yum-repositories/
> On the 31st of March 2020, we will be shutting down the legacy APT and YUM repositories hosted at dockerproject.org and dockerproject.com. These repositories haven’t been updated with the latest releases of Docker and so the packages hosted there contain security vulnerabilities. Removing these repositories will make sure that people download the latest version of Docker ensuring their security and providing the best experience possible

This PR will need to be cherry-picked to 1.16+.


rifelpet:
With this PR what exactly will happen if a user specifies one of the impacted docker versions in the cluster manifest? Will nodeup fallback to the new default docker version of 17.03.2?

Should we add some api-level validation, so that a `kops update cluster` on a cluster that specifies one of these docker versions will fail with a useful error message?

hakman:
@rifelpet I think it will install 17.03.2, this is what I saw happening in the past when the Docker version was not found by NodeUp.
Api-level validation on the docker version would be nice to have. As long as we keep the list of supported Docker versions there, should be a big improvement.

hakman:
Validation for Docker versions is now added. Thanks for the idea @rifelpet.

nodeup/pkg/model/docker.go
```diff
@@ -850,15 +589,6 @@ var dockerVersions = []packageVersion{
...
 
-	// 19.03.7 - Linux Generic
-	{
-		PackageVersion: "19.03.7",
-		PlainBinary:    true,
-		Architectures:  []Architecture{ArchitectureAmd64},
-		Source:         "https://download.docker.com/linux/static/stable/x86_64/docker-19.03.7.tgz",
-		Hash:           "5f7199aa237cc8fa10b95ee0c06c5e9ca9ad4296",
-	},
-
```
johngmyers, line 861: Unintended removal?

nodeup/pkg/model/docker.go
```diff
@@ -43,270 +43,9 @@ var _ fi.ModelBuilder = &DockerBuilder{}
 
 // DefaultDockerVersion is the (legacy) docker version we use if one is not specified in the manifest.
 // We don't change this with each version of kops, we expect newer versions of kops to populate the field.
-const DefaultDockerVersion = "1.12.3"
+const DefaultDockerVersion = "17.03.2"
```
johngmyers, line 46: Might be better to instead require a value in the manifest.

nodeup/pkg/model/docker.go
hakman: This is an intermediary release that was added just to test the new .tgz support. It was never released so there is no need to keep it around.

nodeup/pkg/model/docker.go
hakman: This is used as a fallback when the dirstro or arch is not listed. For example, if anyone chooses to use a Debian Jessie based image with latest Docker, NodeUp will not find it and fall back to this.

nodeup/pkg/model/docker.go
johngmyers: I would have expected it to be in a separate commit then. It is a separate concern and later archaeology might benefit from the separate commit message.

hakman: Seems reasonable, will move it to a separate commit.

nodeup/pkg/model/docker.go
johngmyers: How does that work? I only see this referenced  from the `dockerVersion()` receiver of `DockerBuilder`, when `Cluster.Spec.Docker.Version` is not specified.

It seems odd to be updating a field to the oldest version we can. The comment suggests it's a backwards-compatibility strategy that is never supposed to be updated.

hakman: I looked at the code and I also don't see any other reference. It is used just if the version doesn't exist as you also noticed.

This is a very old piece of code that exists since the beginnings. No idea why it was needed then and why we still keep it. At this point there should always be a Docker version.
https://github.com/kubernetes/kops/blob/62c51002ba69aaafeff38191ad02a8a7b6a2ba1d/pkg/model/components/docker.go#L50-L62

The only explanation I have is this, since containerd support was added: https://github.com/kubernetes/kops/pull/7986#discussion_r357925590.

johngmyers: Yeah, the `components/docker.go` code that presumably ensures the field is set was added in the same commit that adds `DefaultDockerVersion`. @justinsb would have the context.

justinsb: This is indeed very very old.  The kops CLI is supposed to populate the docker version in the "expanded" form of the cluster spec (the one you can see with `--full`).  When it was first introduced, we didn't lock nodeup to the kops version, and so there was the risk of skew.  To tolerate that, we introduced a default docker version that nodeup would use, when kops didn't populate it.

At this point, it's probably easier to remove the nodeup defaulting entirely - just skip installing docker if isn't populated, I would suggest (with a warning so we can figure out what happened!).

pkg/apis/kops/validation/validation.go
```diff
@@ -600,6 +595,25 @@ func validateContainerRuntime(runtime *string, fldPath *field.Path) field.ErrorL
 	return allErrs
 }
 
+func validateDockerConfig(config *kops.DockerConfig, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if config.Version != nil {
+		valid := []string{"17.03.2", "17.09.0", "18.03.1", "18.06.1", "18.06.2", "18.06.3", "18.09.3", "18.09.9", "19.03.4", "19.03.8"}
+		allErrs = append(allErrs, IsValidValue(fldPath.Child("version"), config.Version, valid)...)
```
justinsb, line 603: This validation is the critical thing to backport, I believe.  (I think the others might have a higher risk of collisions).

Should we split this out into its own commit that we cherry-pick back?  And I'm wondering whether it's easier just to exclude the known-bad versions (1.11, 1.12 etc) with an explicit "docker removed these versions" message...  Technically I guess it doesn't really matter if we let through 19.03.8 on an older version that doesn't support it - we're no worse off than we are today.

justinsb:
Thanks for doing this @hakman .  On the cherry-pick, I think the critical thing to cherry pick is a warning/error about the no-longer available docker versions; so I suggested splitting that into its own PR.  Removal of the mappings to URLs is internal machinery and so less important to cherry-pick I suggest, and I worry that there has been some changes there that might make it harder to cherry-pick.

hakman:
Thanks @justinsb. I create a new PR that can be easily cherry-picked and will update this to remove default Docker version.

justinsb:
Thanks for splitting this out - looks great @hakman 

/approve
/lgtm

hakman:
Thanks @justinsb. Will take care of the cherry-picks for Kops 1.16 and 1.17.

PR #8847, opened 2020-04-04
Switch to using ec2.DescribeInstanceTypes for building the MachineTypes list

rifelpet:
`ec2.DescribeInstanceTypes` provides us with all of the information needed to build the machine types list. I consider this an intermediate PR: we can switch to this in the hack script and validate its output, then later switch to using the call in `upup/pkg/fi/cloudup/awsup/machine_types.go` directly, removing the hardcoded list altogether which would decouple instance type support from kops releases.

Marking WIP until I get a chance to review the differences in `upup/pkg/fi/cloudup/awsup/machine_types.go` since ideally there would be none.

In the case of the missing instance types (`cr1.8xlarge`, `hs1.8xlarge`, `m6g.*`), I'm guessing the AWS account I ran this on doesn't have those instance types available: the `cr1` and `hs1` might have been deprecated before my account was created and I know my account hasnt opted in to the `m6g` preview yet. We can figure out a short term workaround for those given that once the followup PR would land, the account running `ec2.DescribeInstanceTypes` will always be the same account the instance lives in, so I'd be confident that the instance type in use will be in the API response.

I still need to review the Memory, Disk, and ENI differences and determine which values are correct.


hakman:
This is super cool :).

upup/pkg/fi/cloudup/awsup/machine_types.go
```diff
@@ -837,7 +800,7 @@ var MachineTypes []AWSMachineTypeInfo = []AWSMachineTypeInfo{
 	{
 		Name:              "i3.metal",
 		MemoryGB:          512,
-		Cores:             64,
+		Cores:             72,
```
johngmyers, line 803: The table [here](https://aws.amazon.com/ec2/instance-types/i3/) states "i3.metal provides 72 logical processors on 36 physical cores". So this is correct.

upup/pkg/fi/cloudup/awsup/machine_types.go
```diff
@@ -1697,7 +1587,7 @@ var MachineTypes []AWSMachineTypeInfo = []AWSMachineTypeInfo{
 	// r3 family
 	{
 		Name:              "r3.large",
-		MemoryGB:          15.25,
+		MemoryGB:          15,
```
johngmyers, line 1590: The table [here](https://aws.amazon.com/ec2/previous-generation/) gives the 15.25 number.

rifelpet:
Regarding some of these differences, I wouldn't be surprised if AWS is inconsistent in their many sources of this information.

I'm tempted to round the MemoryGB fields to 2 decimal places. That would reduce the differences to just the `r3.large` and `t1.micro`. Though `grep` reveals to me that `MemoryGB` doesnt seem to be used anywhere. I suspect it could eventually be used to provide default values for kubelet's reserved resources fields but thats not the case today.

With that, I'm not concerned about changes to the MemoryGB fields. That leaves these discrepancies:

- Missing instance types: I'm tempted to hardcode the missing instance types into the hack script so that they'll be included by default, still with 0 values for ENI and IP information. 
- EphemeralDisks changes: I'm still searching but I don't believe the size is actually used anywhere, only the number of disks is used for device mapping.

If someone could confirm that MemoryGB and EphemeralDisks sizes aren't actually used, then I think this is about ready to merge. We could consider removing the fields (replacing EphemeralDisks with a single count integer rather than a slice of sizes) as well.

johngmyers:
Code crossreference shows the fields as being unreferenced, except that hack/machine_types uses MemoryGB to sort the instance types. Unless something is referenced through reflection.

I was thinking of suggesting the rounding.

justinsb:
Yes, I'm reasonably sure we aren't using Memory or EphemeralDisk sizes 👍 

This is great - thank you!

PR #8848, opened 2020-04-04
Use CNI 0.8.5 for Kubernetes 1.18+

hakman:
Equivalent of https://github.com/kubernetes/kubernetes/pull/78819.

### DIFF
--- a/upup/pkg/fi/cloudup/networking.go
+++ b/upup/pkg/fi/cloudup/networking.go
@@ -125,6 +125,10 @@ const (
 	defaultCNIAssetSHA1StringK8s1_11   = "52e9d2de8a5f927307d9397308735658ee44ab8d"
 	defaultCNIAssetSHA256StringK8s1_11 = "3ca15c0a18ee830520cf3a95408be826cbd255a1535a38e0be9608b25ad8bf64"
 
+	// defaultCNIAssetK8s1_18 is the CNI tarball for k8s >= 1.18
+	defaultCNIAssetK8s1_18             = "https://storage.googleapis.com/k8s-artifacts-cni/release/v0.8.5/cni-plugins-linux-amd64-v0.8.5.tgz"
+	defaultCNIAssetSHA256StringK8s1_18 = "bd682ffcf701e8f83283cdff7281aad0c83b02a56084d6e601216210732833f9"
+
 	// Environment variable for overriding CNI url
 	ENV_VAR_CNI_VERSION_URL       = "CNI_VERSION_URL"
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/networking.go
```diff
@@ -160,19 +164,23 @@ func findCNIAssets(c *api.Cluster, assetBuilder *assets.AssetBuilder) (*url.URL,
...
+	if util.IsKubernetesGTE("1.18", *sv) {
+		cniAsset = defaultCNIAssetK8s1_18
+		cniAssetHash = defaultCNIAssetSHA256StringK8s1_18
+		klog.V(2).Infof("Adding default CNI asset for k8s >= 1.18: %s", cniAsset)
+	} else if util.IsKubernetesGTE("1.15", *sv) {
 		// We're still on the same asset, but we use sha256
 		cniAsset = defaultCNIAssetK8s1_11
 		cniAssetHash = defaultCNIAssetSHA256StringK8s1_11
-		klog.V(2).Infof("Adding default CNI asset for k8s >= 1.11: %s", defaultCNIAssetK8s1_9)
+		klog.V(2).Infof("Adding default CNI asset for 1.18 > k8s >= 1.11: %s", cniAsset)
```
rifelpet, line 175:
```suggestion
		klog.V(2).Infof("Adding default CNI asset for 1.18 > k8s >= 1.15: %s", cniAsset)
```

upup/pkg/fi/cloudup/networking.go
```diff
@@ -160,19 +164,23 @@ func findCNIAssets(c *api.Cluster, assetBuilder *assets.AssetBuilder) (*url.URL,
...
 	} else if util.IsKubernetesGTE("1.11", *sv) {
 		cniAsset = defaultCNIAssetK8s1_11
 		cniAssetHash = defaultCNIAssetSHA1StringK8s1_11
-		klog.V(2).Infof("Adding default CNI asset for k8s >= 1.11: %s", defaultCNIAssetK8s1_9)
+		klog.V(2).Infof("Adding default CNI asset for 1.18 > k8s >= 1.11: %s", cniAsset)
```
rifelpet, line 179:
```suggestion
		klog.V(2).Infof("Adding default CNI asset for 1.15 > k8s >= 1.11: %s", cniAsset)
```

upup/pkg/fi/cloudup/networking.go
hakman: It is actually the same version, just the hash changed from sha1 to sha256, right?

rifelpet: ah you're right, nevermind

PR #8850, opened 2020-04-05
Disable static tokens by default as of Kubernetes 1.18

johngmyers:
Disables by default in k8s 1.18, can be re-enabled ~~by enabling basic auth. Removed in k8s 1.19~~ using instructions in security.md.

### DIFF
--- a/docs/releases/1.18-NOTES.md
+++ b/docs/releases/1.18-NOTES.md
@@ -42,7 +42,9 @@
 
 * Lyft CNI plugin default subnet tags changed from from `Type: pod` to `KubernetesCluster: myclustername.mydns.io`. Subnets intended for use by the plugin will need to be tagged with this new tag and [additional tag filters](https://github.com/lyft/cni-ipvlan-vpc-k8s#other-configuration-flags) may need to be added to the cluster spec in order to achieve the desired set of subnets.
 
-* Support for basic authentication has been disabled by default and will be [removed](https://github.com/kubernetes/kubernetes/pull/89069) in Kubernetes 1.19.
+* Support for basic authentication has been disabled by default for Kubernetes 1.18 and will be [removed](https://github.com/kubernetes/kubernetes/pull/89069) in Kubernetes 1.19.
+
+* Support for static tokens has been disabled by default for Kubernetes 1.18 and later. To re-enable, see the [Security Notes for Kubernetes](../security.md). We intend to remove support entirely in a future kops version, so file an issue with your use case if you need this feature. 
 
 * Support for Kubernetes versions prior to 1.9 has been removed.
 
--- a/docs/security.md
+++ b/docs/security.md
@@ -51,16 +51,23 @@ spec:
     anonymousAuth: false
 ```
 
+Clusters created with `kops create cluster` using Kubernetes 1.11 or later will have this setting in the generated cluster spec and thus have AnonymousAuth disabled.
+
 **Note** on an existing cluster with 'anonymousAuth' unset you would need to first roll out the masters and then update the node instance groups.
 
 ### API Bearer Token
 
+Static bearer tokens are disabled by default as of Kubernetes 1.18.
+In order to enable them:
+
+```YAML
+# In the cluster spec
+spec:
+  kubeAPIServer:
+    tokenAuthFile: "/srv/kubernetes/known_tokens.csv"
+```
+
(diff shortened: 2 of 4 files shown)


johngmyers:
Okay, so have to remove basic auth first.

hakman:
@johngmyers @rifelpet This looks more like complete removal of static tokens for 1.18+.
According to https://github.com/kubernetes/kubernetes/pull/89069 that remove basic auth, "Similar functionality is available via the --token-auth-file flag for development purposes".

I guess that means it's not something that should be used by people other than devs, but I am curious about the reason to remove it from Kops starting 1.18+.

johngmyers:
The removal would be for security hardening.

https://github.com/kubernetes/kops/blob/bf24a6443c270e1b4e4bc2be95ad2629c553a0bd/pkg/tokens/wellknown.go#L20

states they've been deprecated since 1.7, yet they're still provisioned always. Also

https://github.com/kubernetes/kops/blob/64ffe00b1cd27ad4c1ecd95acc1350d5eac4a1ea/pkg/apis/kops/componentconfig.go#L308

incorrectly states the field is unused.

hakman:
Seems a good idea to me. My suggestion would be to disable by default in 1.18 and remove completely in 1.19, same as with basic auth.

johngmyers:
The question would be whether it is worth adding a field to the API for re-enabling this in 1.18.

Other possibilities would be to reuse `DisableBasicAuth` in 1.18 or to go from enabled in 1.18 to removed in 1.19.

hakman:
Continuing the discussion in office hours, my vote is for adding a new flag, instead of reusing the `DisableBasicAuth` one. Best to have a name that matches the action, even if it's another flag.
Also I am all for disabling the Token Auth by default and adding this to release notes.

johngmyers:
These autogenerated static tokens have been deprecated since kops 1.8. The only thing using any of them was the basic auth support, which is controlled by the field in question.

Basic auth will not work without also enabling static tokens. Basic auth and static tokens have equivalent security properties—it is just a difference of protocol encoding. There is little to no benefit of disabling basic auth without also disabling the autogenerated static tokens.

justinsb:
I'm sort of confused about https://github.com/kubernetes/kubernetes/pull/89069 - it removes basic auth but not the seemingly equivalent token auth.

This makes it hard to find a reasonable answer, I feel.  AFAICT k8s isn't removing or deprecating token-auth, at least not in #89069.

kops can be proactive here and lock this down ahead of removal from k8s upstream - that does seem like a good idea.    One thing we could do fairly easily is to not generate auth tokens we don't ourselves use, i.e. I don't think we should still be auto-generating the tokens in GetKubernetesAuthTokens_Deprecated - with the possible exception of "kube" or "admin" which should probably be tied to basic-auth (and thus also go away in 1.19).

If anyone was relying on these tokens, we should be able to figure that out in 1.18 beta.  It's also not too hard to generate your own tokens (it looks like we don't have `kubectl create secret token`, which does suggest nobody is using them).  However, this wouldn't help people that have existing clusters, that would have these secrets.  If someone was using this, I would imagine they would be creating their own tokens.  However, we don't write any additional tokens anyway.  Because of this, I'd say it is OK simply to stop writing this file and stop passing the TokenAuthFile flag.

The fallback I imagine is that users could still directly set TokenAuthFile on the apiserver in their cluster config; so we would stop defaulting it, but if users set it we would honor it.  To introduce this gradually, we could continue to write `known_tokens.csv` but not wire it in automatically.  That gives an easy way to get back the existing behaviour (in the release note, we should also request that users open an issue to let us know, to stop us further deprecating the behaviour).

To summarize:

* I propose we stop setting `token-auth-file` on kube-apiserver automatically in kops 1.18.0-beta.1, for k8s >= 1.18.
* We add a release note explaining, and giving a workaround - set `token-auth-file`, open an issue to let us know the use-case
* I propose we continue to write `token-auth-file` for a release or two, in case of a use-case, and to provide that fallback.
* Once we're happy, we can stop generating the unused tokens / stop writing the file entirely.

I think this meets the need - we do generate and write these tokens to disk, but I think they are harmless unless we pass the flag to kube-apiserver.

johngmyers:
I'll work tonight on changing the PR to create `known_tokens.csv` but not enable them by default.

justinsb:
Thanks @johngmyers (and @hakman for the discussion!)

/approve
/lgtm

PR #8855, opened 2020-04-06
Remove support for Docker 1.11, 1.12 and 1.13

hakman:
The repo containing the binaries for Docker 1.11, 1.12 and 1.13 was shut down on March 31. This means support for older docker versions cannot be maintained.
https://www.docker.com/blog/changes-dockerproject-org-apt-yum-repositories/
> On the 31st of March 2020, we will be shutting down the legacy APT and YUM repositories hosted at dockerproject.org and dockerproject.com. These repositories haven’t been updated with the latest releases of Docker and so the packages hosted there contain security vulnerabilities. Removing these repositories will make sure that people download the latest version of Docker ensuring their security and providing the best experience possible

This PR will need to be cherry-picked to 1.16+.

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -33,14 +33,6 @@ import (
 	"k8s.io/kops/pkg/model/iam"
 )
 
-var validDockerConfigStorageValues = []string{"aufs", "btrfs", "devicemapper", "overlay", "overlay2", "zfs"}
-
-func ValidateDockerConfig(config *kops.DockerConfig, fldPath *field.Path) field.ErrorList {
-	allErrs := field.ErrorList{}
-	allErrs = append(allErrs, IsValidValue(fldPath.Child("storage"), config.Storage, validDockerConfigStorageValues)...)
-	return allErrs
-}
-
 func newValidateCluster(cluster *kops.Cluster) field.ErrorList {
 	allErrs := validation.ValidateObjectMeta(&cluster.ObjectMeta, false, validation.NameIsDNSSubdomain, field.NewPath("metadata"))
 	allErrs = append(allErrs, validateClusterSpec(&cluster.Spec, field.NewPath("spec"))...)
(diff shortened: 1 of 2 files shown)


hakman:
This should be the simplest thing that can be cherry-picked to 1.16+.
https://github.com/kubernetes/kops/pull/8846#issuecomment-609791975

pkg/apis/kops/validation/validation.go
```diff
@@ -600,6 +595,30 @@ func validateContainerRuntime(runtime *string, fldPath *field.Path) field.ErrorL
...
 
+func validateDockerConfig(config *kops.DockerConfig, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if config.Version != nil {
+		if strings.HasPrefix(*config.Version, "1.1") {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("version"), config.Version,
+				"version is no longer available: https://www.docker.com/blog/changes-dockerproject-org-apt-yum-repositories/"))
+		} else {
+			valid := []string{"17.03.2", "17.09.0", "18.03.1", "18.06.1", "18.06.2", "18.06.3", "18.09.3", "18.09.9", "19.03.4", "19.03.8"}
```
rifelpet, line 606: What used to happen if a user specified a version that was not in this list? like [18.09.8](https://docs.docker.com/engine/release-notes/18.09/#18098) for example.

hakman: Any Docker version would have been accepted and NodeUp would fail if it was not listed.
It can still fail as not all versions have support for all distros.

rifelpet: hm does that mean that if a specific version of docker was already installed in the AMI and a user specified a version not in this list, they used to have a functioning cluster with their docker version but now their `update cluster` command would fail? I'm just trying to think through any concerns with adding this validation...

hakman: If a specific Docker version is install on a AMI, `skipInstall` would have been used and the Docker version would not matter much, I guess.

PR #8856, opened 2020-04-07
Use ec2.DescribeInstanceTypes for machine type info

rifelpet:
The AWS client is not initialized by the time API validation is performed, so I had to move the  `machineType` instance group field validation to `apply_cluster.go`.  

Marking WIP because I still need to update `nodeup/pkg/model/kubelet.go`'s use of `GetMachineTypeInfo` (only relevant with the AmazonVPC CNI), as well as some tests and mocks. I also realize the validation is no longer returning multiple validation errors, I'll get that fixed.

Feedback is welcome on a better approach to support every use of GetMachineTypeInfo.


cloudmock/aws/mockec2/instances.go
```diff
@@ -49,3 +49,8 @@ func (m *MockEC2) DescribeInstancesPages(request *ec2.DescribeInstancesInput, ca
 func (m *MockEC2) DescribeInstancesPagesWithContext(aws.Context, *ec2.DescribeInstancesInput, func(*ec2.DescribeInstancesOutput, bool) bool, ...request.Option) error {
 	panic("Not implemented")
 }
+
+func (m *MockEC2) DescribeInstanceTypes(*ec2.DescribeInstanceTypesInput) (*ec2.DescribeInstanceTypesOutput, error) {
+	klog.Warningf("MockEc2::DescribeInstanceTypes is stub-implemented")
```
justinsb, line 54: If it helps, we can try to narrow the instance types we support in testing, or just support fake instance types.

pkg/apis/kops/validation/aws.go
```diff
@@ -75,20 +72,6 @@ func awsValidateAdditionalSecurityGroups(fieldPath *field.Path, groups []string)
 	return allErrs
 }
 
-func awsValidateMachineType(fieldPath *field.Path, machineType string) field.ErrorList {
-	allErrs := field.ErrorList{}
-
-	if machineType != "" {
```
justinsb, line 81: Are we removing this because we're going to do it in apply instead?

upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -1509,3 +1512,23 @@ func (c *awsCloudImplementation) zonesWithInstanceType(instanceType string) (set
...
+
+// DescribeInstanceType calls ec2.DescribeInstanceType to get information for a particular instance type
+func (c *awsCloudImplementation) DescribeInstanceType(instanceType string) (*ec2.InstanceTypeInfo, error) {
+	return describeInstanceType(c, instanceType)
+}
+
+func describeInstanceType(c AWSCloud, instanceType string) (*ec2.InstanceTypeInfo, error) {
+	req := &ec2.DescribeInstanceTypesInput{
+		InstanceTypes: aws.StringSlice([]string{instanceType}),
+		MaxResults:    aws.Int64(1),
```
justinsb, line 1524: Do we need MaxResults?  I'm worried it might mask a problem with the count?

upup/pkg/fi/cloudup/awsup/machine_types.go
```diff
@@ -58,2503 +60,37 @@ func (m *AWSMachineTypeInfo) EphemeralDevices() []*EphemeralDevice {
...
+		return nil, err
+	}
+	machine := AWSMachineTypeInfo{
+		Name:              machineType,
+		GPU:               info.GpuInfo != nil,
+		InstanceENIs:      intValue(info.NetworkInfo.MaximumNetworkInterfaces),
+		InstanceIPsPerENI: intValue(info.NetworkInfo.Ipv4AddressesPerInterface),
+	}
+	memoryGB := float64(intValue(info.MemoryInfo.SizeInMiB)) / 1024
+	machine.MemoryGB = float32(math.Round(memoryGB*100) / 100)
```
justinsb, line 76: Nit: Should we remove the rounding logic once we're happy we get matching values?  Maybe just add a todo?

upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -1509,3 +1512,23 @@ func (c *awsCloudImplementation) zonesWithInstanceType(instanceType string) (set
 
 	return zones, nil
 }
+
+// DescribeInstanceType calls ec2.DescribeInstanceType to get information for a particular instance type
+func (c *awsCloudImplementation) DescribeInstanceType(instanceType string) (*ec2.InstanceTypeInfo, error) {
+	return describeInstanceType(c, instanceType)
```
justinsb, line 1518: Idea for the future: A cache here might save us a few API calls.

mikesplain:
This is great, thanks for your work on this @rifelpet!

johngmyers:
Deferring the field validation until `cloudup` is really unfortunate.  It looks like that would mean a typoed machinetype would not be caught during `kops edit instancegroup` but would fail on the subsequent `kops update cluster`.

How bad would it be for api validation to try getting a `fi.Cloud`? If that fails it could silently skip validating the machinetypes at that point.

rifelpet:
I agree it would be unfortunate. I believe that validation is called by PopulateClusterSpec (which is called by `cmd/kops/edit_cluster.go` and `cmd/kops/create_cluster.go`). PopulateClusterSpec is called twice, once with strict=false and then again with strict=true after other fields are populated. The second call occurs [after fi.Cloud is initialized](https://github.com/kubernetes/kops/blob/a8f13e319ccce0d17e08a9f8bde52ee3029ba258/upup/pkg/fi/cloudup/populate_cluster_spec.go#L230-L330) so it might be possible to pass in the Cloud object and only validate when strict=true.

Thoughts?

pkg/apis/kops/validation/legacy.go
```diff
@@ -537,17 +537,11 @@ func DeepValidate(c *kops.Cluster, groups []*kops.InstanceGroup, strict bool) er
...
 
 	for _, g := range groups {
-		errs := CrossValidateInstanceGroup(g, c)
+		errs := CrossValidateInstanceGroup(g, c, cloud)
 
-		// Additional cloud-specific validation rules,
-		// such as making sure that identifiers match the expected formats for the given cloud
-		switch kops.CloudProviderID(c.Spec.CloudProvider) {
-		case kops.CloudProviderAWS:
-			errs = append(errs, awsValidateInstanceGroup(g)...)
```
rifelpet, line 546: `awsValidateInstanceGroup` is now called through `CrossValidateInstanceGroup` -> `ValidateInstanceGroup` so that we can get additional validation for `kops create ig` and `kops edit ig` since they only call `CrossValidateInstanceGroup` rather than `DeepValidate`.

cloudmock/aws/mockec2/instances.go
rifelpet: This function isn't actually used, it just needs to be defined in order to satisfy the mock interface. The actual stub we use in integration testing is in `upup/pkg/fi/cloudup/awsup/mock_aws_cloud.go` where I defined a reasonable mock behavior.

rifelpet:
This is ready for review.

This requires passing a cloud object in additional places throughout the validation package and originating mostly from `cmd/kops`.

This means that some kops commands now require valid cloud provider credentials, but I don't think this is an issue because the vast majority of use-cases already require the same cloud provider credentials in order to interact with the cluster's state store.

At this point my only concern is the increased AWS API usage by nodes when using the Amazon VPC CNI provider. [It uses the machine type info to set kubelet's --max-pods](https://github.com/kubernetes/kops/blob/bc876dae881439b551db94d16c2b96b887f1d607/nodeup/pkg/model/kubelet.go#L448-L463). This means any new node will make the DescribeInstanceTypes call once at startup which could run into rate limiting for large clusters. I suppose the only way around this would be to call DescribeInstanceTypes in `kops` for each instance type in the IG and pass the mapping of "instance type -> max pod" through the InstanceGroupSpec to be consumed by nodeup, but that's quite a bit more complex.

rifelpet:
As discussed during office hours, I added a release note mentioning the behavior change. Since there isn't much of a remedy we have available other than contacting AWS support for a rate limit increase, I opted to instruct the user to open a GH issue. If anyone has any better ideas please let me know.

/assign @mikesplain 
for testing

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -114,6 +115,10 @@ func ValidateInstanceGroup(g *kops.InstanceGroup) field.ErrorList {
 		allErrs = append(allErrs, validateRollingUpdate(g.Spec.RollingUpdate, field.NewPath("spec", "rollingUpdate"), g.Spec.Role == kops.InstanceGroupRoleMaster)...)
 	}
 
+	if awsCloud, ok := cloud.(awsup.AWSCloud); ok {
+		allErrs = append(allErrs, awsValidateInstanceGroup(g, awsCloud)...)
```
johngmyers, line 119: It would be better if `awsValidateInstanceGroup` were called when `cloud` is `nil` but the cloud provider is AWS.

docs/releases/1.19-NOTES.md
```diff
@@ -4,6 +4,8 @@
 
 # Significant changes
 
+* Clusters using the Amazon VPC CNI provider now perform an `ec2.DescribeInstanceTypes` call at instance  launch time. In large clusters or AWS accounts this may lead to API throttling which could delay node readiness. If this becomes a problem please open a GitHub issue.
```
johngmyers, line 7:
```suggestion
* Clusters using the Amazon VPC CNI provider now perform an `ec2.DescribeInstanceTypes` call at instance launch time. In large clusters or AWS accounts this may lead to API throttling which could delay node readiness. If this becomes a problem please open a GitHub issue.
```

pkg/apis/kops/validation/instancegroup.go
rifelpet: I'm not sure I understand. I can change this conditional to look at the cloud provider being AWS, but are you suggesting that `awsValidateInstanceGroup` should take a `fi.Cloud` instead of `awsup.AWSCloud`? Where should the `fi.Cloud` -> `awsup.AWSCloud` type assertion happen then?

johngmyers: If the type assertion fails then (if cloud provider is AWS) it should call `awsValidateInstanceGroup` with a nil `awsup.AWSCloud`.

mikesplain:
I just tried to test this out twice but both times I think there was a networking issue... kubedns/coredns failed both times.  I'm going to dig in and make sure it's not related.

Ratelimiting didn't appear to hamper things too much in one of our busiest accounts so I think we're probably safe there.

rifelpet:
sounds good, thanks for the review! thoughts on cherry-picking to 1.18? or should we let this sit for a few extra months in 1.19 pre-releases for additional testing?
/hold cancel

mikesplain:
I think it makes sense to get into 1.18, less to deal with in terms of machine updates... We could also let it bake for a few days in master if you'd like. Thanks again!

johngmyers:
I would prefer we not put larger features into 1.18 at this point. It is in beta and we should be working on getting it out the door.

olemarkus:
Second what John said.

PR #8857, opened 2020-04-07
Fix addon/ing/nginx v1.6.0 deployment apiVersion and selector

mousavian:
Deployments with `extensions/v1beta1` apiVersions are deprecated, according to [this](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/).

`spec.selector` is also required for deployment.

### DIFF
--- a/addons/ingress-nginx/v1.6.0.yaml
+++ b/addons/ingress-nginx/v1.6.0.yaml
@@ -174,7 +174,7 @@ spec:
 ---
 
 kind: Deployment
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 metadata:
   name: nginx-default-backend
   namespace: kube-ingress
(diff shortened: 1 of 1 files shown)


johngmyers:
The apps/v1 Deployment was apparently introduced in Kubernetes 1.9, which is the oldest k8s version that kops 1.18 will support. This change would not be able to be backported to kops 1.17 or 1.16 as they theoretically support k8s 1.8 and earlier. To be backported, we would need this to be a second manifest, such as was done in #8842.

The two commits here should not be separate as the first does not work without the second.

/ok-to-test

justinsb:
@johngmyers  good points. The addons folder is sort of a special case though - it's more a shared library of applications, so I think it's OK that it only works from k8s 1.9 on.  I would like to get these into being managed addons via operators, I need to circle back on these PRs once we cut 1.18.  Splitting commits is always sort of tricky - I do like when commits are individually coherent, but this often means that they won't pass testing.  Splitting into two commits is probably overkill in this case, but I'd rather we over-split than under-split our commits.

Thanks for fixing this up @mousavian 

/approve
/lgtm

PR #8868, opened 2020-04-07
Validate cluster N times in rolling-update

zetaab:
We are still seeing lots of rolling update errors in case of cluster validation after instance roll. 

Example:
```
I0407 16:41:39.366581 85 instancegroups.go:255] Cluster validated; revalidating in 10s to make sure it does not flap.
I0407 16:42:00.200357 85 instancegroups.go:271] Cluster validated.
master not healthy after update, stopping rolling-update: "cluster \"updateospr-f95a75.k8s.local\" did not pass validation: kube-system pod \"kube-apiserver-master-zone-1-1-1-updateospr-f95a75-k8s-local\" is pending"
I0407 16:42:17.290824 1 batch.go:902] error running kops rolling-update cluster --bastion-interval 2m --instance-group bastions,master-zone-1-1,master-zone-2-1,master-zone-3-1,nodes-z1,nodes-z2,nodes-z3 --validation-timeout 20m --yes
```

Disclaimer: we are running e2e tests quite heavily against kops. We are doing something like 20-50 rolling updates per day for clusters.

cc @hakman @johngmyers could you guys check this.

### DIFF
--- a/cmd/kops/rollingupdatecluster.go
+++ b/cmd/kops/rollingupdatecluster.go
@@ -121,6 +121,9 @@ type RollingUpdateOptions struct {
 	// ValidationTimeout is the timeout for validation to succeed after the drain and pause
 	ValidationTimeout time.Duration
 
+	// ValidateTimes is the amount of time that a cluster needs to be validated after single node update
+	ValidateTimes int32
+
 	// MasterInterval is the minimum time to wait after stopping a master node.  This does not include drain and validate time.
 	MasterInterval time.Duration
 
@@ -158,6 +161,7 @@ func (o *RollingUpdateOptions) InitDefaults() {
 
 	o.PostDrainDelay = 5 * time.Second
 	o.ValidationTimeout = 15 * time.Minute
+	o.ValidateTimes = 2
 }
 
 func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
@@ -177,6 +181,7 @@ func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().BoolVar(&options.CloudOnly, "cloudonly", options.CloudOnly, "Perform rolling update without confirming progress with k8s")
 
 	cmd.Flags().DurationVar(&options.ValidationTimeout, "validation-timeout", options.ValidationTimeout, "Maximum time to wait for a cluster to validate")
+	cmd.Flags().Int32Var(&options.ValidateTimes, "validate-times", options.ValidateTimes, "Amount of times that a cluster needs to be validated after single node update")
 	cmd.Flags().DurationVar(&options.MasterInterval, "master-interval", options.MasterInterval, "Time to wait between restarting masters")
 	cmd.Flags().DurationVar(&options.NodeInterval, "node-interval", options.NodeInterval, "Time to wait between restarting nodes")
 	cmd.Flags().DurationVar(&options.BastionInterval, "bastion-interval", options.BastionInterval, "Time to wait between restarting bastions")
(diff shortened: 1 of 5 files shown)


cmd/kops/rollingupdatecluster.go
```diff
@@ -333,9 +338,10 @@ func RunRollingUpdateCluster(f *util.Factory, out io.Writer, options *RollingUpd
 		ClusterName:       options.ClusterName,
 		PostDrainDelay:    options.PostDrainDelay,
 		ValidationTimeout: options.ValidationTimeout,
+		ValidateTimes:     options.ValidateTimes,
+		ValidateSucceeded: 0,
 		// TODO should we expose this to the UI?
-		ValidateTickDuration:    30 * time.Second,
-		ValidateSuccessDuration: 10 * time.Second,
```
zetaab, line 338: this was not exposed to CLI, so as I see it - its easy to delete

johngmyers:
Could we get details on why things are flapping? Did the kube-apiserver pod not exist when the cluster validated earlier? Do we need to explicitly check for it like we now do for controller-manager?

zetaab:
in my opinion checking individual pods is not maybe the best option because kube-system namespace can contain any pods which should be checked. That is why generic retry logic is better than checking individual pod

cmd/kops/rollingupdatecluster.go
```diff
@@ -177,6 +181,7 @@ func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().BoolVar(&options.CloudOnly, "cloudonly", options.CloudOnly, "Perform rolling update without confirming progress with k8s")
 
 	cmd.Flags().DurationVar(&options.ValidationTimeout, "validation-timeout", options.ValidationTimeout, "Maximum time to wait for a cluster to validate")
+	cmd.Flags().Int32Var(&options.ValidateTimes, "validate-times", options.ValidateTimes, "Amount of times that a cluster needs to be validated after single node update")
```
hakman, line 184: For validate we have `--count`. Would be nice to keep it consistent.
```suggestion
	cmd.Flags().Int32Var(&options.ValidateTimes, "validate-count", options.ValidateTimes, "Amount of times that a cluster needs to be validated after single node update")
```

zetaab:
@hakman fixed

hakman:
Cool. Thanks @zetaab.
/lgtm

pkg/instancegroups/rollingupdate.go
```diff
@@ -64,9 +64,11 @@ type RollingUpdateCluster struct {
...
 	ValidateTickDuration time.Duration
 
-	// ValidateSuccessDuration is the amount of time a cluster must continue to validate successfully
-	// before updating the next node
-	ValidateSuccessDuration time.Duration
+	// ValidateCount is the amount of time that a cluster needs to be validated after single node update
+	ValidateCount int32
+
+	// ValidateSucceeded is the amount of times that a cluster validate is succeeded already
+	ValidateSucceeded int32
```
johngmyers, line 71: Why is this public?

pkg/instancegroups/instancegroups.go
```diff
@@ -430,10 +430,12 @@ func (c *RollingUpdateCluster) validateClusterWithDuration(duration time.Duratio
 func (c *RollingUpdateCluster) tryValidateCluster(duration time.Duration) bool {
 	result, err := c.ClusterValidator.Validate()
 
-	if err == nil && len(result.Failures) == 0 && c.ValidateSuccessDuration > 0 {
-		klog.Infof("Cluster validated; revalidating in %s to make sure it does not flap.", c.ValidateSuccessDuration)
-		time.Sleep(c.ValidateSuccessDuration)
-		result, err = c.ClusterValidator.Validate()
+	if err == nil && len(result.Failures) == 0 && c.ValidateCount > 0 {
```
johngmyers, line 433: Why the `c.ValidateCount > 0` check? It seems unnecessary.

pkg/instancegroups/instancegroups.go
```diff
@@ -430,10 +430,12 @@ func (c *RollingUpdateCluster) validateClusterWithDuration(duration time.Duratio
 func (c *RollingUpdateCluster) tryValidateCluster(duration time.Duration) bool {
 	result, err := c.ClusterValidator.Validate()
 
-	if err == nil && len(result.Failures) == 0 && c.ValidateSuccessDuration > 0 {
-		klog.Infof("Cluster validated; revalidating in %s to make sure it does not flap.", c.ValidateSuccessDuration)
-		time.Sleep(c.ValidateSuccessDuration)
```
johngmyers, line 435: Why did you remove the sleep?

zetaab: when we return false, it will go back to previous function which do have sleep already

zetaab:
@johngmyers imo your comment is not true, I tested this. In case of all instancegroups it will loop through validatecount

PR #8869, opened 2020-04-08
GCE DNS Cleanup: Print record name

justinsb:
Use the GroupKey to pass the zone name, meaning the name can be the
user-facing value.

Follow-on to #8250


justinsb:
This is what I was thinking @mccare

hakman:
@justinsb should this be merged?

PR #8873, opened 2020-04-08
Update go.mod for k8s 1.17

justinsb:
These are now tagged with v0.17.4, so we don't need the trick of the
commented out imports any more!


justinsb:
BTW here's a video of how to update these, except that I don't think we need to any more because it looks like upstream has (mostly?) fixed their go.mods, so now we can just use v0.17.4 instead of kubernetes-1.17.4, and that value isn't mangled by go.mod: https://asciinema.org/a/317782

justinsb:
cc @rifelpet on the video.  The other steps are to run `make gomod` to update vendor/, and then fix up any problems - here the problem was only a new flag that GCE cloudprovider added, that otherwise would make it into all our docs, so pretty easy this time (at least so far!)

mikesplain:
Thanks @justinsb! 

/lgtm
/approve

go.mod
```diff
@@ -123,22 +97,22 @@ require (
...
-	gopkg.in/yaml.v2 v2.2.7
+	gopkg.in/yaml.v2 v2.2.8
 	honnef.co/go/tools v0.0.1-2019.2.3
-	k8s.io/api v0.0.0
-	k8s.io/apimachinery v0.0.0
-	k8s.io/cli-runtime v0.0.0
+	k8s.io/api v0.17.4
+	k8s.io/apimachinery v0.17.4
+	k8s.io/cli-runtime v0.17.4
 	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
```
rifelpet, line 113: thanks, the video was very helpful! can you explain why this version didn't (need to) change? It seems all of the k8s.io packages are covered by `replace`s above, but this, kubectl, and legacy-cloud-providers below didn't change.

justinsb: Hmmm ... that's a good point and is just my mistake!  Will hold the PR!

justinsb: Ah actually, I take it back...

The replace directives at the top take precedence, and we do have:

```
replace k8s.io/client-go => k8s.io/client-go v0.17.4
```

client-go started adopted a different versioning scheme though, and it looks like it isn't fully sorted out yet.  That confuses go.mod which then puts some random version in the directives.

I was able to override the version in go.mod manually, so it now looks intuitive, even though that version wasn't used before.

We'll likely have to do this manually going forwards (but if we forget, it shouldn't really matter)

rifelpet: ah ok. and does the same apply for kubectl and legacy-cloud-providers ?

rifelpet:
seems like the answer to my last question is yes.

/lgtm

I propose we let this sit for a bit to let periodic e2e jobs run, and then cherry-pick back to 1.17

PR #8874, opened 2020-04-08
OpenStack: possibility to detach instances from the load

zetaab:
what is changed:

1) implement detach things for instance and port.
2) it will modify instance names to be more "cattle" stylish. (it is needed if we want detach instances)

```
% kubectl get nodes
NAME                                  STATUS   ROLES    AGE     VERSION
master-zone-1-1-1-newname-k8s-local   Ready    master   2m31s   v1.17.3
master-zone-1-2-1-newname-k8s-local   Ready    master   2m25s   v1.17.3
master-zone-1-3-1-newname-k8s-local   Ready    master   2m31s   v1.17.3
nodes-1-newname-k8s-local             Ready    node     2m11s   v1.17.3
nodes-2-newname-k8s-local             Ready    node     2m8s    v1.17.3
nodes-3-newname-k8s-local             Ready    node     2m11s   v1.17.3


% kubectl get nodes
NAME                    STATUS   ROLES    AGE   VERSION
master-zone-1-1-fcnlz6  Ready    master   63m   v1.17.4
master-zone-1-2-lynpa3  Ready    master   71m   v1.17.4
master-zone-1-3-ef5w1s  Ready    master   55m   v1.17.4
nodes-3hfjy4            Ready    node     36m   v1.17.4
nodes-7fjtht            Ready    node     43m   v1.17.4
```

### DIFF
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -152,6 +152,8 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 			Metadata:         igMeta,
 			SecurityGroups:   ig.Spec.AdditionalSecurityGroups,
 			AvailabilityZone: az,
+			ClusterName:      s(b.ClusterName()),
+			GroupName:        s(ig.Name),
 		}
 		if igUserData != nil {
 			instanceTask.UserData = igUserData
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -98,13 +98,13 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 		securityGroups = append(securityGroups, b.LinkToSecurityGroup(b.Cluster.Spec.MasterPublicName))
 	}
 
+	r := strings.NewReplacer("_", "-", ".", "-")
+	groupName := r.Replace(strings.ToLower(ig.Name))
 	// In the future, OpenStack will use Machine API to manage groups,
 	// for now create d.InstanceGroups.Spec.MinSize amount of servers
 	for i := int32(0); i < *ig.Spec.MinSize; i++ {
 		// FIXME: Must ensure 63 or less characters
-		// replace all dots and _ with -, this is needed to get external cloudprovider working
-		iName := strings.Replace(strings.ToLower(fmt.Sprintf("%s-%d.%s", ig.Name, i+1, b.ClusterName())), "_", "-", -1)
-		instanceName := fi.String(strings.Replace(iName, ".", "-", -1))
+		instanceName := fi.String(r.Replace(strings.ToLower(fmt.Sprintf("%s-%d.%s", ig.Name, i+1, b.ClusterName()))))
 
 		var az *string
 		var subnets []*openstacktasks.Subnet
@@ -141,6 +141,7 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 
 		instanceTask := &openstacktasks.Instance{
 			Name:             instanceName,
+			NamePrefix:       s(groupName),
 			Region:           fi.String(b.Cluster.Spec.Subnets[0].Region),
 			Flavor:           fi.String(ig.Spec.MachineType),
 			Image:            fi.String(ig.Spec.Image),
@@ -152,8 +153,6 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 			Metadata:         igMeta,
 			SecurityGroups:   ig.Spec.AdditionalSecurityGroups,
 			AvailabilityZone: az,
-			ClusterName:      s(b.ClusterName()),
-			GroupName:        s(ig.Name),
 		}
 		if igUserData != nil {
 			instanceTask.UserData = igUserData
--- a/upup/pkg/fi/cloudup/openstack/instance.go
+++ b/upup/pkg/fi/cloudup/openstack/instance.go
@@ -65,7 +65,7 @@ func (c *openstackCloud) CreateInstance(opt servers.CreateOptsBuilder) (*servers
 	}
 }
 
-func (c *openstackCloud) ListServerFloatingIPs(instanceID string) ([]*string, error) {
+func (c *openstackCloud) ListServerIPs(instanceID string, IPType string) ([]*string, error) {
 	var result []*string
 	_, err := vfs.RetryWithBackoff(floatingBackoff, func() (bool, error) {
 		server, err := c.GetInstance(instanceID)
@@ -82,7 +82,7 @@ func (c *openstackCloud) ListServerFloatingIPs(instanceID string) ([]*string, er
 		for _, addrList := range addresses {
 			for _, props := range addrList {
 				if c.floatingEnabled {
-					if props.IPType == "floating" {
+					if props.IPType == IPType {
 						result = append(result, fi.String(props.Addr))
 					}
 				} else {
@@ -96,7 +96,7 @@ func (c *openstackCloud) ListServerFloatingIPs(instanceID string) ([]*string, er
 		return false, nil
 	})
 	if len(result) == 0 || err != nil {
-		return result, fmt.Errorf("could not find floating ip associated to server (\"%s\") %v", instanceID, err)
+		return result, fmt.Errorf("could not find %s ip associated to server (\"%s\") %v", IPType, instanceID, err)
 	}
 	return result, nil
 }
@@ -112,6 +112,7 @@ func (c *openstackCloud) DeleteInstanceWithID(instanceID string) error {
 
 // DetachInstance is not implemented yet. It needs to cause a cloud instance to no longer be counted against the group's size limits.
 func (c *openstackCloud) DetachInstance(i *cloudinstances.CloudInstanceGroupMember) error {
+	// TODO: tag port and instance with detach tag
 	klog.V(8).Info("openstack cloud provider DetachInstance not implemented yet")
 	return fmt.Errorf("openstack cloud provider does not support surging")
 }
--- a/upup/pkg/fi/cloudup/openstacktasks/instance.go
+++ b/upup/pkg/fi/cloudup/openstacktasks/instance.go
@@ -28,13 +28,13 @@ import (
 	"k8s.io/klog"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstack"
-	"k8s.io/kops/upup/pkg/fi/utils"
 )
 
 //go:generate fitask -type=Instance
 type Instance struct {
 	ID               *string
 	Name             *string
+	NamePrefix       *string
 	Port             *Port
 	Region           *string
 	Flavor           *string
@@ -47,8 +47,6 @@ type Instance struct {
 	Metadata         map[string]string
 	AvailabilityZone *string
 	SecurityGroups   []string
-	ClusterName      *string
-	GroupName        *string
 	Lifecycle        *fi.Lifecycle
 }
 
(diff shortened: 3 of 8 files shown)


zetaab:
@mitch000001 could you check this. For some time ago I did that detach thing, but now this contains also modifying the instance names. It means that now we can detach old nodes from cluster, those can still serve customers. Meanwhile in background new nodes will come up and join to cluster (that is possible because new node name is not same anymore than old one)

mitch000001:
@zetaab not necessarily related to the PR, but what happens to these detached instances?

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -97,30 +100,55 @@ func (e *Instance) Find(c *fi.Context) (*Instance, error) {
 	if e == nil || e.Name == nil {
 		return nil, nil
 	}
+	tmp := strings.Replace(strings.ToLower(fi.StringValue(e.GroupName)), "_", "-", -1)
```
mitch000001, line 103: Possibly using [strings.Replacer](https://pkg.go.dev/strings?tab=doc#Replacer) makes the code more clear and removes the need for a temporary variable

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -151,12 +179,23 @@ func (_ *Instance) ShouldCreate(a, e, changes *Instance) (bool, error) {
 	return a == nil, nil
 }
 
+func makeServerName(e *Instance) string {
+	hash := utils.RandomString(6)
+	// FIXME: Must ensure 63 or less characters
+	// replace all dots and _ with -, this is needed to get external cloudprovider working
+	stripped := strings.Replace(fi.StringValue(e.ClusterName), ".k8s.local", "", -1)
+	iName := strings.Replace(strings.ToLower(fmt.Sprintf("%s-%s.%s", fi.StringValue(e.GroupName), hash, stripped)), "_", "-", -1)
```
mitch000001, line 187: Possibly using [strings.Replacer](https://pkg.go.dev/strings?tab=doc#Replacer) makes the code more clear and removes the need for a temporary variable. As this is the second occurrence where we need that logic we could create a package var for the replacer.

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -151,12 +179,23 @@ func (_ *Instance) ShouldCreate(a, e, changes *Instance) (bool, error) {
 	return a == nil, nil
 }
 
+func makeServerName(e *Instance) string {
+	hash := utils.RandomString(6)
+	// FIXME: Must ensure 63 or less characters
+	// replace all dots and _ with -, this is needed to get external cloudprovider working
+	stripped := strings.Replace(fi.StringValue(e.ClusterName), ".k8s.local", "", -1)
```
mitch000001, line 186: As for the limit of 63 characters for the name, can't we also just remove the cluster name from the instances, as we already have the k8s tags in place?

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -151,12 +179,23 @@ func (_ *Instance) ShouldCreate(a, e, changes *Instance) (bool, error) {
...
+	iName := strings.Replace(strings.ToLower(fmt.Sprintf("%s-%s.%s", fi.StringValue(e.GroupName), hash, stripped)), "_", "-", -1)
+	instanceName := strings.Replace(iName, ".", "-", -1)
+	return instanceName
+}
+
 func (_ *Instance) RenderOpenstack(t *openstack.OpenstackAPITarget, a, e, changes *Instance) error {
 	if a == nil {
-		klog.V(2).Infof("Creating Instance with name: %q", fi.StringValue(e.Name))
-
+		e.Metadata[openstack.TagKopsName] = fi.StringValue(e.Name)
```
mitch000001, line 194: What is the difference between `e.Name` and the result of `makeServerName(e)`?

pkg/model/openstackmodel/servergroup.go
```diff
@@ -152,6 +152,8 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 			Metadata:         igMeta,
 			SecurityGroups:   ig.Spec.AdditionalSecurityGroups,
 			AvailabilityZone: az,
+			ClusterName:      s(b.ClusterName()),
```
mitch000001, line 155: Why aren't we using the new name creation in Line 106?

pkg/model/openstackmodel/servergroup.go
```diff
@@ -152,6 +152,8 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
...
+			GroupName:        s(ig.Name),
```
mitch000001, line 156: Do we need the group on the instance if we have a reference to the ServerGroup which contains the instance group name?

upup/pkg/fi/cloudup/openstack/instance.go
```diff
@@ -65,7 +65,7 @@ func (c *openstackCloud) CreateInstance(opt servers.CreateOptsBuilder) (*servers
 	}
 }
 
-func (c *openstackCloud) ListServerFloatingIPs(instanceID string) ([]*string, error) {
+func (c *openstackCloud) ListServerIPs(instanceID string, IPType string) ([]*string, error) {
```
mitch000001, line 68: I mean I like the generalization of that method, but there is no call using it without `FloatingType` in this PR.

upup/pkg/fi/cloudup/openstacktasks/instance.go
justinsb: As a style point, I try to put kops specific logic in the model, and keep the tasks generic.  So in this case I would expect to normalize the groupName in the model, and pass it in here as "namePrefix" (or something like that).

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -97,30 +100,55 @@ func (e *Instance) Find(c *fi.Context) (*Instance, error) {
...
 	serverList, err := servers.ExtractServers(serverPage)
 	if err != nil {
 		return nil, fmt.Errorf("error extracting server page: %v", err)
 	}
-	if len(serverList) == 0 {
+
+	filteredList := []servers.Server{}
+	for _, server := range serverList {
+		val, ok := server.Metadata["k8s"]
+		if !ok || val != fi.StringValue(e.ClusterName) {
```
justinsb, line 119: The "generic" way to do this would be to pass in `filterTags map[string]string`.  Or just `tags map[string]string` if they are used both for filtering and creation.

An advantage of that is that I'm not sure that these values are also set on creation.  I'm sure they are, but ... it would be more obvious if it was just `tags map[string]string` :-)

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -97,30 +100,55 @@ func (e *Instance) Find(c *fi.Context) (*Instance, error) {
...
+		}
+		_, detachTag := server.Metadata[openstack.TagNameDetach]
+		// name or metadata tag should match to instance name
+		// this is needed for backwards compatibility
+		if server.Name == fi.StringValue(e.Name) || metadataName == fi.StringValue(e.Name) {
+			if !detachTag {
+				filteredList = append(filteredList, server)
+			}
+		}
+	}
```
justinsb, line 135: Tip: you might want to either do `serverList = nil` or `serverList = filteredList` around here to prevent accidental use of serverList

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -97,30 +100,55 @@ func (e *Instance) Find(c *fi.Context) (*Instance, error) {
...
+		val, ok := server.Metadata["k8s"]
+		if !ok || val != fi.StringValue(e.ClusterName) {
+			continue
+		}
+		metadataName := ""
+		val, ok = server.Metadata[openstack.TagKopsName]
+		if ok {
+			metadataName = val
+		}
+		_, detachTag := server.Metadata[openstack.TagNameDetach]
```
justinsb, line 127: I don't know how we can make this one generic, though :-)  We _could_ pass in `ignoreMatching map[string]string` but it's starting to get a little too abstract, perhaps :-)  If you do change the others, I think it's fine to leave Detach as a special-case :-)

upup/pkg/fi/cloudup/openstacktasks/port.go
```diff
@@ -126,12 +126,19 @@ func (s *Port) Find(context *fi.Context) (*Port, error) {
 	if err != nil {
 		return nil, err
 	}
-	if rs == nil {
+	filteredList := []ports.Port{}
```
justinsb, line 129: Tip: I used to do this also (and probably still do in places), I've since learned that because a nil slice behaves (almost) identically to an empty slice, you can normally just do the shorter `var filteredList []ports.Port`.

upup/pkg/fi/utils/hash.go
```diff
@@ -32,3 +34,31 @@ func HashString(s string) (string, error) {
...
+const (
+	letterBytes   = "abcdefghijklmnopqrstuvwxyz0123456789"
+	letterIdxBits = 6                    // 6 bits to represent a letter index
+	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
+	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
+)
+
+// RandomString generates a random string of n length
+// source https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-go
+func RandomString(n int) string {
```
justinsb, line 47: upup/pkg/fi/secrets.go has:

```
func CreateSecret() (*Secret, error) {
	data := make([]byte, 128)
	_, err := crypto_rand.Read(data)
	if err != nil {
		return nil, fmt.Errorf("error reading crypto_rand: %v", err)
	}

	s := base64.StdEncoding.EncodeToString(data)
	r := strings.NewReplacer("+", "", "=", "", "/", "")
	s = r.Replace(s)
	s = s[:32]

	return &Secret{
		Data: []byte(s),
	}, nil
}
```

I know crypto_rand is overkill, but it's probably an easier API for us to use.

Also... apparently I used strings.Replacer in the past!

justinsb:
Thanks @zetaab!  Just a drive-by review with some comments on ways to structure the model vs the tasks, generally agree with the PR comments (thanks @mitch000001).

upup/pkg/fi/cloudup/openstacktasks/instance.go
zetaab: e.Name is the old stylish name for the instance. makeServerName returns the new style name. Which contains hash instead of index.

upup/pkg/fi/cloudup/openstacktasks/instance.go
zetaab: it is removed in new format

pkg/model/openstackmodel/servergroup.go
zetaab: old format uses that, but in new format the clustername needs to be passed as is to task (it is checked there)

upup/pkg/fi/cloudup/openstack/instance.go
zetaab: yeah my plan was to delete port also when deleting instance, but it was not that easy - so basically that is ready there now for next try :)

upup/pkg/fi/cloudup/openstacktasks/instance.go
zetaab: the problem in openstack is that servergroups does not support things that for instance AWS instancegroups do support. This thing is much more difficult than in AWS

zetaab:
I need test this, earliest tuesday

mitch000001:
@zetaab Do you also want to cover the detached server filtering by tests?

zetaab:
@mitch000001 I tried to look tests but could not figure where it should be written.. upup/pkg/fi/cloudup/openstacktasks? I tried to modify things under models (pkg/model/openstacktasks) but detach tags does not affect anything there

mitch000001:
@zetaab I think `upup/pkg/fi/cloudup/openstacktasks` is a good place. I mean I'd try to verify that the filter is working as expected and what happens if the detach tag is in place.
Also, in `pkg/model/openstackmodel/servergroup.go` we add a new `NamePrefix` which is not covered by the way  we introduced the test, as it is testing the fields explicitly rather than just testing all fields.

upup/pkg/fi/cloudup/openstacktasks/instance.go
mitch000001: So, in the kops world we still work with the old names whereas in Openstack we use the new ones? I tend to like it more to be consistent between those two worlds. Is that possible?

olemarkus:
Would really love to see this going in.
/remove-lifecycle stale

zetaab:
I will open new PR because this is pretty far in history. Tried to rebase but easier to make jsut new

PR #8877, opened 2020-04-09
Don't delete bindata on "make clean"

rifelpet:
Now that this is checked into version control we shouldn't be deleting it. replace it with a `bazel clean` that also removes the bazel directories.


justinsb:
bazel clean as a replacement is a nice idea - thanks @rifelpet 

/approve
/lgtm

PR #8878, opened 2020-04-09
Fix Terraform 0.12 support for GCE

rifelpet:
The HCL2 library used for Terraform 0.12 support does not handle embedded types easily. The `google_compute_instance` and `google_compute_instance_template` terraform types used embedded types for common fields. Rather than detecting embedded types with reflection I figured it was easier to just copy the shared fields directly into the two types which included creating distinct AttachedDisk types as well. This also updates the `Add*` receiver functions for the defunct InstanceCommon type to instead return the value for the field that was being modified for both `google_compute_instance` and `google_compute_instance_template`.

This also adds support for writing maps of strings to `file()` references which is used in GCE's metadata fields.

Lastly, this updates the verify-terraform script to exit non-zero on failure now that it is successfully passing for all integration test clusters.

There is a larger question of whether the `google_compute_instance` resource type is actually used. It seems synonymous with having AWS resources for both LaunchConfigurations (for use by an ASG) and individual EC2 instances, which doesn't make much sense. I'm not a GCE user but perhaps @geojaz can chime in? If we don't need to generate `google_compute_instance` resources (which the integration test outputs currently do not) then we could delete `instance.go` altogether. Thoughts?

### DIFF
--- a/hack/verify-terraform.sh
+++ b/hack/verify-terraform.sh
@@ -41,9 +41,7 @@ done 3< <(find "${KOPS_ROOT}/tests/integration/update_cluster" -type d -maxdepth
 
 if [ $RC != 0 ]; then
   echo -e "\nTerraform validation failed\n"
-  # TODO(rifelpet): make this script blocking in PRs by exiting non-zero on failure
-  # exit $RC
-  exit 0
+  exit $RC
 else
   echo -e "\nTerraform validation succeeded\n"
 fi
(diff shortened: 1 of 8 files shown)


justinsb:
Thanks @rifelpet - this looks spot on.

/approve
/lgtm

I do agree that I don't think we are using the raw gce_instance, and it could probably be deleted.  GCE does have a stronger notion of a persistent instance, more like a MIG of size 1, but ... it also has MIGs of size 1.

PR #8879, opened 2020-04-09
Update to kubernetes 1.18.1 dependencies

justinsb:


justinsb:
Adding as WIP because controller-runtime doesn't build with 1.18 libraries.

justinsb:
The latest (master) of controller-runtime actually does support 1.18, so we can update after all!

hakman:
@justinsb now that v1.18.1 was release, should we just use that instead of v.1.18.0?

justinsb:
Good call @hakman - updated to 1.18.1

hakman:
@justinsb seems that this broke 1.10 and 1.11. I am not really sure what happens with the mount part to be able to fix it.
```
W0412 18:12:27.646560    3634 volume_mounter.go:79] unable to mount master volume: "error formatting and mounting disk \"/dev/xvdu\" on \"/mnt/master-vol-00b718c718950ef02\": MountSensitive not implemented for containerized mounter"
```

https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-1-11/1249398784325062656/artifacts/3.8.78.27/protokube.log

PR #8893, opened 2020-04-10
Add go 1.14 to travis testing

justinsb:


hakman:
Nice! :)
/lgtm

hakman:
@justinsb Should we keep just 1 test for OSX?

justinsb:
@hakman probably yes, that's a good idea.  k8s 1.18 apimachinery is no longer compatible with go 1.12, so in that PR I commented out (but left ready to uncomment) the osx version exclusion.  We should add it back in once the 1.18 PR lands!

PR #8895, opened 2020-04-10
Remove duplicated docs

olemarkus:
The docs in README.md is duplicated in /docs, but we are not good at
keeping them in sync. Therefor I removed most of the information from
README.md andr refer to the docs instead.
Also fixed a few broken/missing links.

This is really more of an RFC, but I would love to get into a state where people start using https://kops.sigs.k8s.io/ over just the markdowns on github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,33 +29,9 @@ in alpha, and other platforms planned.
 </p>
 
 
-## Launching a Kubernetes cluster hosted on AWS, GCE, DigitalOcean or OpenStack
+## Installing and launching a Kubernetes cluster hosted on AWS, GCE, DigitalOcean or OpenStack
 
-To replicate the above demo, check out our [tutorial](/docs/getting_started/aws.md) for
-launching a Kubernetes cluster hosted on AWS.
-
-To install a Kubernetes cluster on GCE please follow this [guide](/docs/getting_started/gce.md).
-
-To install a Kubernetes cluster on DigitalOcean, follow this [guide](/docs/getting_started/digitalocean.md).
-
-To install a Kubernetes cluster on OpenStack, follow this [guide](/docs/getting_started/openstack.md).
-
-**For anything beyond experimental clusters it is highly encouraged to [version control the cluster manifest files](/docs/manifests_and_customizing_via_api.md) and [run kops in a CI environment](/docs/continuous_integration.md).**
-
-## Features
-
-* Automates the provisioning of Kubernetes clusters in [AWS](/docs/getting_started/aws.md), [OpenStack](/docs/getting_started/openstack.md) and [GCE](/docs/getting_started/gce.md)
-* Deploys Highly Available (HA) Kubernetes Masters
-* Built on a state-sync model for **dry-runs** and automatic **idempotency**
-* Ability to generate [Terraform](/docs/terraform.md)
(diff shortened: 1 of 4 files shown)


docs/welcome/releases.md
```diff
@@ -21,13 +21,13 @@ Kubernetes version.
 
 ## Compatibility Matrix
 
-| kops version  | k8s 1.11.x | k8s 1.12.x | k8s 1.13.x | k8s 1.14.x | k8s 1.15.x |
+| kops version  | k8s 1.12.x | k8s 1.13.x | k8s 1.14.x | k8s 1.15.x | k8s 1.16.x |
```
justinsb, line 24: I feel like the information in this chart is the biggest stumbling block for most people.  That said, the common failure mode is running a newer version of k8s than kops, and I believe we have an error message for that now, and we are hopefully about to ~ catch up, so may not be as big of a concern any more?

README.md
```diff
@@ -29,215 +29,21 @@ in alpha, and other platforms planned.
...
-__Pull Requests__
-* Read and review the code. Leave comments, questions, and critiques (`/lgtm` ).
-* Download, compile, and run the code and make sure the tests pass (make test).
-  - Also verify that the new feature seems sane, follows best architectural patterns, and includes tests.
-
-This repository uses the Kubernetes bots.  See a full list of the commands [here](
-https://go.k8s.io/bot-commands).
-
-
-## Office Hours
```
justinsb, line 190: I do think we should have a link to office hours on the "homepage"

docs/welcome/releases.md
olemarkus: I think this is less of a concern now that we have that warning.

README.md
olemarkus: I added a shorter description of office hours now, with a link to the longer one.

docs/welcome/releases.md
mikesplain: Agreed. The chart was added to help people but often confused them even more. Regardless, I feel like lack of clarity around supported versions was the real issue (which we've addresses very heavily lately).  I think we should leave this for now and come back another time if it should be removed from the releases page.

docs/index.md
```diff
@@ -36,9 +36,4 @@ in alpha, and other platforms planned.
  Manifests
 * Choose from eight different CNI [Networking](networking.md) providers out-of-the-box
 * Supports upgrading from [kube-up](upgrade_from_kubeup.md)
-* Capability to add containers, as hooks, and files to nodes via a [cluster manifest](cluster_spec.md)
-
-
-## Documentation
-
-[To check out Live documentation](https://kops.sigs.k8s.io/)
```
mikesplain, line 44: I wonder if we should move this to a comment at the top of this file to tell markdown users they should goto the live docs site?

README.md
mikesplain: Thanks for adding them, I agree it's good to have the link.

docs/index.md
olemarkus: Not sure it is easy to add this without it also being rendered on the live docs site.
Maybe something to look into later?

mikesplain: Fair enough. 👍

mikesplain:
Thanks @olemarkus!

/lgtm
/approve

PR #8896, opened 2020-04-10
Implement setter by reflection

justinsb:
This means we no longer have to individually hard-code the `kops set` fields, however we use the "language" we're now demonstrated.

We add tests to ensure we have parity with our existing (hard-coded) setter logic.


justinsb:
Removing WIP, this seems to work and I wish we had this e.g. when I was testing CoreDNS!

justinsb:
/test pull-kops-e2e-k8s-containerd

/test pull-kops-verify-staticcheck

Timeout & an old test

util/pkg/reflectutils/access.go
```diff
@@ -0,0 +1,122 @@
...
+					newV = reflect.ValueOf(b)
+				}
+
+			case "int64", "*int64", "int32", "*int32", "int", "*int":
+				v, err := strconv.Atoi(newValue)
+				if err != nil {
+					return fmt.Errorf("cannot interpret %q value as integer", newValue)
+				}
+				if strings.HasPrefix(t, "*") {
+					newV = reflect.ValueOf(&v)
```
johngmyers, line 73: Is an `*int` assignable to an `*int64` or `*int32`?

util/pkg/reflectutils/access.go
```diff
@@ -0,0 +1,122 @@
...
+				} else {
+					newV = reflect.ValueOf(v)
+				}
+
+			case "[]string":
+				newV = reflect.Append(v, reflect.ValueOf(newValue))
```
johngmyers, line 79: Do we want to split on ","?

util/pkg/reflectutils/access.go
```diff
@@ -0,0 +1,122 @@
...
+
+			default:
+				// This handles enums and other simple conversions
```
johngmyers, line 82: We could also support things that are convertible from int or bool. Also slices of things convertible from string, int, or bool.

util/pkg/reflectutils/access.go
```diff
@@ -0,0 +1,122 @@
...
+			t := v.Type().String()
+
+			var newV reflect.Value
+
+			switch v.Type().Elem().Kind() {
+			case reflect.Struct:
+				newV = reflect.New(v.Type().Elem())
+
+			default:
+				return fmt.Errorf("unhandled type %v %q", v.Type().Elem().Kind(), t)
```
johngmyers, line 110: Might be possible to recurse in order to handle the "*" cases above.

util/pkg/reflectutils/access_test.go
```diff
@@ -0,0 +1,135 @@
...
+			Value:    "hello-world",
+		},
+		{
+			Name:     "creating missing objects",
+			Input:    "{ 'spec': { 'containers': [ {} ] } }",
+			Expected: "{ 'spec': { 'containers': [ { 'policy': { 'name': 'allowed' } } ] } }",
+			Path:     "Spec.Containers[0].Policy.Name",
+			Value:    "allowed",
+		},
+		// Not sure if we should do this...
```
johngmyers, line 97: I think if it's effectively appending an element, then it's okay.

util/pkg/reflectutils/field_path.go
```diff
@@ -0,0 +1,178 @@
...
+					if err != nil {
+						return nil, fmt.Errorf("cannot parse token %q as array-index", v)
+					}
+					elements = append(elements, FieldPathElement{
+						Type:   FieldPathElementTypeArrayIndex,
+						number: n,
+					})
+
+				default:
+					return nil, fmt.Errorf("unknown token %v (%s)", tok, scan.TokenText())
```
johngmyers, line 123: More like "unexpected token"

util/pkg/reflectutils/access_test.go
johngmyers: We insert intermediate elements when they're nil pointers, which seems analogous. If it's not valid after processing all setters then API validation will catch that.

PR #8897, opened 2020-04-10
Initial validation durring rolling update does single try.

granular-ryanbonham:
When doing a rolling update, the initial validation of the cluster only does a single check. This can cause failures if pods are pending, nodes coming up, etc.. 

This MR switches the initial  validation to do multiple checks, like the rest of rolling update.

### DIFF
--- a/pkg/instancegroups/instancegroups.go
+++ b/pkg/instancegroups/instancegroups.go
@@ -91,7 +91,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(ctx context.Context, c
 	} else if c.CloudOnly {
 		klog.V(3).Info("Not validating cluster as validation is turned off via the cloud-only flag.")
 	} else {
-		if err = c.validateClusterWithDuration(c.validationTimeout); err != nil {
+		if err = c.validateClusterWithDuration(c.ValidationTimeout); err != nil {
 			if c.FailOnValidate {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


pkg/instancegroups/instancegroups.go
```diff
@@ -90,7 +90,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(cluster *api.Cluster,
 	} else if c.CloudOnly {
 		klog.V(3).Info("Not validating cluster as validation is turned off via the cloud-only flag.")
 	} else {
-		if err = c.validateCluster(); err != nil {
+		if err = c.validateClusterWithDuration(validationTimeout); err != nil {
```
zetaab, line 93: In my opinion this is good change, I was planning to make same.

zetaab:
```
pkg/instancegroups/instancegroups.go:458:32: func (*RollingUpdateCluster).validateCluster is unused (U1000)
```

that is needed to remove if we continue here. I can add lgtm, but that fix is needed first

granular-ryanbonham:
@johngmyers   This still looks like it is needed... Do you have any objections, if not I will update the MR so we can get it merged..

johngmyers:
@granular-ryanbonham looks like I have a branch I forgot to submit as a PR. Please give me some time to update and submit it.

PR #8902, opened 2020-04-11
Clean up the spec docs

olemarkus:
* try to make cluster and IG docs more consistent
* split the IG docs between the spec and common operational use cases

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -1,10 +1,12 @@
-# Description of Keys in `config` and `cluster.spec`
+# The `Cluster` resource
 
-This list is not complete but aims to document any keys that are less than self-explanatory. Our [go.dev](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops) reference provides a more detailed list of API values. [ClusterSpec](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#ClusterSpec), defined as `kind: Cluster` in YAML, and [InstanceGroupSpec](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#InstanceGroupSpec), defined as `kind: InstanceGroup` in YAML, are the two top-level API values used to describe a cluster.
+The `Cluster` resource contains the specification of the cluster itself.
 
-## spec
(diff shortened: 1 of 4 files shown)


docs/cluster_spec.md
```diff
@@ -742,9 +744,9 @@ spec:
       image: busybox
 ```
 
-### fileAssets
+## fileAssets
 
-FileAssets is an alpha feature which permits you to place inline file content into the cluster and instanceGroup specification. It's designated as alpha as you can probably do this via kubernetes daemonsets as an alternative.
+FileAssets permits you to place inline file content into the cluster and instanceGroup specification. This is useful for depoying additionl configuration files that kube components requires, such as auditlogs or admission controller configurations.
```
rifelpet, line 749:
```suggestion
FileAssets permits you to place inline file content into the cluster and instanceGroup specification. This is useful for deploying additional configuration files that kubernetes components requires, such as auditlogs or admission controller configurations.
```

docs/instance_groups.md
```diff
@@ -1,452 +1,25 @@
-# Instance Groups
+# The `InstanceGroup` resource
 
-kops has the concept of "instance groups", which are a group of similar machines.  On AWS, they map to
-an AutoScalingGroup.
+The `InstanceGroup` resource represents a group of similar machines typically provisioned in the same availability zone. On AWS, instance groups map directly to an autoscale group.
```
rifelpet, line 3:
```suggestion
The `InstanceGroup` resource represents a group of similar machines typically provisioned in the same availability zone. On AWS, instance groups map directly to an autoscaling group.
```

docs/instance_groups.md
```diff
@@ -596,3 +140,36 @@ spec:
 ```
 
 which would end up in a drop-in file on nodes of the instance group in question.
+
+## mixedInstancePolicy (AWS Only)
```
rifelpet, line 144: Perhaps an example yaml snippet that contains each of the fields mentioned below?

olemarkus: I can add that. Mind if I do it in a later PR?

rifelpet: yea that's fine with me

PR #8918, opened 2020-04-15
Exclude GO 1.14 for OSX

hakman:
Travis OSX test VMs are more rare than linux ones. No need to build on OSX for non-default GO versions.

For reference:
https://github.com/kubernetes/kops/pull/8893#issuecomment-612004667


rifelpet:
looks good, once we upgrade to go 1.14 officially we can switch this to only run osx on 1.14.

/lgtm
/approve

PR #8921, opened 2020-04-15
Bastion additionnal s gs

### LINKED ISSUE #8013: Allow to use additional security group to Bastion ELB (AWS)
Hi,

We would like to add an additional security group to the Bastion ELB (AWS).
The following tickets were created on the topic and closed by the damn lifecycle bot (one of them reopened many times, thank you @-kaazoo) but never solved.

- https://github.com/kubernetes/kops/issues/2177
- https://github.com/kubernetes/kops/issues/3350
- https://github.com/kubernetes/kops/issues/6697
- https://github.com/kubernetes/kops/issues/3937

Please read all of them.
sshAccess is not what this ticket is about.
AdditionalSecurityGroups is not what this ticket is about (https://github.com/kubernetes/kops/issues/3937#issuecomment-370851649).

The tickets I listed and comments on it list enough use-cases, so we can avoid going through them again here. People is also bumping closed ticket to raise this issue and it's clearly something that should not happen.

I want to say beforehand: I know people can do PR (me included) but that is not the point of this ticket. I also don't want to force anybody in the KOPS dev team to work on an issue that they may not think is a priority. I just want to ask you guys to protect a ticket about a functionality that some people need from being created again and again and allow people to bump it up to track the requests. Thank you for all your work on KOPS. It's a tool we use extensively and we found it really useful.

nvanheuverzwijn:
**Description**
Adds `topology.bastion.loadBalancer.additionalSecurityGroups` to the cluster spec.

It allows to add additional security groups to the bastion's ELB

fix #8013 

**Tests**
I tested this patch with version 1.16.0 but it targets the master branch.

I am not sure how I should add unit test.


nvanheuverzwijn:
So, I wasn't too sure if we wanted to add a new field to the bastion spec or not.. I noticed that some people wanted a bastion without an ELB so maybe we might add a bastion.dns field for eventual no elb bastion. However, because the bastion is an instancegroup (and an autoscaling group), I am not too sure how to implement a bastion with a direct dns record.

hakman:
Hi @nvanheuverzwijn. To fix the CI error you need to run `make crds`.
/ok-to-test

rifelpet:
This looks great, thanks for the addition!

Could you add this new field with a dummy SG value to [this](https://github.com/kubernetes/kops/blob/master/tests/integration/update_cluster/bastionadditional_user-data/in-v1alpha2.yaml) integration test? After that run `./hack/updated-expected.sh` and we should expect to see the new SG added to the ELB in [this file](https://github.com/kubernetes/kops/blob/master/tests/integration/update_cluster/bastionadditional_user-data/kubernetes.tf)

nvanheuverzwijn:
@rifelpet I modified the file you mentionned.

Should I add something in cmd/kops/integration_test.go ? There are some bastion test there as well. Thank you for your time, very appreciated.

pkg/apis/kops/bastion.go
```diff
@@ -20,4 +20,10 @@ type BastionSpec struct {
 	BastionPublicName string `json:"bastionPublicName,omitempty"`
 	// IdleTimeoutSeconds is the bastion's Loadbalancer idle timeout
 	IdleTimeoutSeconds *int64 `json:"idleTimeoutSeconds,omitempty"`
+
+	LoadBalancer *LoadBalancerSpec `json:"loadBalancer,omitempty"`
+}
+
+type LoadBalancerSpec struct {
```
rifelpet, line 27: I'm wondering about this type. We probably shouldn't reuse the LoadBalancerAccessSpec since it has fields that only make sense for the k8s API load balancer. It's probably best to have a separate type for this, in which case maybe we name the type something bastion-specific like `BastionLoadBalancerSpec`? That way we can avoid confusion with any other LoadBalancer spec in the kops API. You could keep the json field name as `loadBalancer`.

rifelpet:
no need to change anything in cmd/kops, just the integration test cluster spec and terraform files 👍

pkg/apis/kops/bastion.go
nvanheuverzwijn: Make sense, let me fix this.

nvanheuverzwijn:
@rifelpet Let me know if everything looks good. I will rebase the fixup.

rifelpet:
yea that looks good, feel free to squash the commits and then we can get this merged

nvanheuverzwijn:
@rifelpet Done!

rifelpet:
great thanks!
/lgtm
/approve

PR #8922, opened 2020-04-16
Add ARM64 build targets for kops and nodeup

hakman:
Adding the ARM64 build targets for kops and nodeup to see how build and CI is behaving.
This should be merged first to avoid issue with ARM64 assets not being found by `kops apply cluster`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -216,8 +216,19 @@ ${DIST}/linux/amd64/nodeup: ${BINDATA_TARGETS}
 	mkdir -p ${DIST}
 	GOOS=linux GOARCH=amd64 go build ${GCFLAGS} -a ${EXTRA_BUILDFLAGS} -o $@ ${LDFLAGS}"${EXTRA_LDFLAGS} -X k8s.io/kops.Version=${VERSION} -X k8s.io/kops.GitVersion=${GITSHA}" k8s.io/kops/cmd/nodeup
 
+.PHONY: ${DIST}/linux/arm64/nodeup
+${DIST}/linux/arm64/nodeup: ${BINDATA_TARGETS}
+	mkdir -p ${DIST}
+	GOOS=linux GOARCH=arm64 go build ${GCFLAGS} -a ${EXTRA_BUILDFLAGS} -o $@ ${LDFLAGS}"${EXTRA_LDFLAGS} -X k8s.io/kops.Version=${VERSION} -X k8s.io/kops.GitVersion=${GITSHA}" k8s.io/kops/cmd/nodeup
+
+.PHONY: crossbuild-nodeup-amd64
+crossbuild-nodeup-amd64: ${DIST}/linux/amd64/nodeup
+
+.PHONY: crossbuild-nodeup-arm64
+crossbuild-nodeup-arm64: ${DIST}/linux/arm64/nodeup
+
 .PHONY: crossbuild-nodeup
-crossbuild-nodeup: ${DIST}/linux/amd64/nodeup
+crossbuild-nodeup: crossbuild-nodeup-amd64 crossbuild-nodeup-arm64
 
 .PHONY: crossbuild-nodeup-in-docker
 crossbuild-nodeup-in-docker:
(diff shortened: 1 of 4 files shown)


justinsb:
This looks great.  It would be great to rationalize some of the makefile tasks (I think we have bazel and non-bazel versions for everything!), but I agree with just adding arm64 and not trying to tackle that at the same time.

I'm going to mark this for 1.19, as I worry that e.g. changing the path to nodeup will break some things, and we're about to cut 1.18.  This seems like a good reason to create the 1.18 branch also!

hakman:
Thanks @justinsb, sounds good to be a feature for 1.19!

Makefile
```diff
@@ -324,27 +355,48 @@ gen-cli-docs: ${KOPS} # Regenerate CLI docs
...
+	ssh ${TARGET} sudo cp /tmp/nodeup /var/lib/toolbox/kubernetes-install/nodeup
+	ssh ${TARGET} sudo /var/lib/toolbox/kubernetes-install/nodeup --conf=/var/lib/toolbox/kubernetes-install/kube_env.yaml --v=8
+
+# -t is for CentOS http://unix.stackexchange.com/questions/122616/why-do-i-need-a-tty-to-run-sudo-if-i-can-sudo-without-a-password
+.PHONY: push-aws-run-amd64
+push-aws-run-amd64: push-amd64
+	ssh -t ${TARGET} sudo /tmp/nodeup --conf=/opt/kops/conf/kube_env.yaml --v=8
+
 # -t is for CentOS http://unix.stackexchange.com/questions/122616/why-do-i-need-a-tty-to-run-sudo-if-i-can-sudo-without-a-password
-.PHONY: push-aws-run
```
rifelpet, line 346: I think there are a few references to these make targets in the docs directory that should get updated.

hakman: Thanks for noticing this. Done.

Makefile
```diff
@@ -324,27 +355,48 @@ gen-cli-docs: ${KOPS} # Regenerate CLI docs
...
+
+.PHONY: push-gce-dry-amd64
+push-gce-dry-amd64: push-amd64
+	ssh ${TARGET} sudo /tmp/nodeup --conf=metadata://gce/instance/attributes/config --dryrun --v=8
+
+.PHONY: push-gce-dry-arm64
+push-gce-dry-arm64: push-arm64
 	ssh ${TARGET} sudo /tmp/nodeup --conf=metadata://gce/instance/attributes/config --dryrun --v=8
 
-.PHONY: push-gce-dry
```
rifelpet, line 336: good catch :)

Makefile
```diff
@@ -265,25 +288,33 @@ kops-dist: crossbuild-in-docker
...
 	tools/sha256 ${DIST}/windows/amd64/kops.exe ${DIST}/windows/amd64/kops.exe.sha256
 
 .PHONY: version-dist
 version-dist: nodeup-dist kops-dist protokube-export utils-dist
 	rm -rf ${UPLOAD}
 	mkdir -p ${UPLOAD}/kops/${VERSION}/linux/amd64/
+	mkdir -p ${UPLOAD}/kops/${VERSION}/linux/arm64/
 	mkdir -p ${UPLOAD}/kops/${VERSION}/darwin/amd64/
 	mkdir -p ${UPLOAD}/kops/${VERSION}/images/
-	mkdir -p ${UPLOAD}/utils/${VERSION}/linux/amd64/
```
johngmyers, line 277: Looks like some of the utils.tar.gz-removal PR sneaked in here.

hakman: That this was not actually used:
https://github.com/kubernetes/kops/blob/fecec5affadc30ff269bb8f45aed74b3637ac2e2/Makefile#L321-L323

Makefile
```diff
@@ -229,6 +240,14 @@ crossbuild-nodeup-in-docker:
 	docker kill nodeup-build-${UNIQUE}
 	docker rm nodeup-build-${UNIQUE}
 
+.PHONY: nodeup-dist
+nodeup-dist: crossbuild-nodeup-in-docker
+	mkdir -p ${DIST}
```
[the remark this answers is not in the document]
hakman: I agree, this is pretty much not needed. I think it was copy/paste. In any case, seems that there are some more places where mkdirs would not be needed.

Makefile
```diff
@@ -239,14 +258,18 @@ ${DIST}/linux/amd64/kops: ${BINDATA_TARGETS}
 	mkdir -p ${DIST}
 	GOOS=linux GOARCH=amd64 go build ${GCFLAGS} -a ${EXTRA_BUILDFLAGS} -o $@ ${LDFLAGS}"${EXTRA_LDFLAGS} -X k8s.io/kops.Version=${VERSION} -X k8s.io/kops.GitVersion=${GITSHA}" k8s.io/kops/cmd/kops
 
+.PHONY: ${DIST}/linux/arm64/kops
```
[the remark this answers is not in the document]
hakman: Some cleanup in the "phony" area would be nice also. Kept it similar to the similar targets for now.

hakman:
@johngmyers any thoughts on trying to remove the `.sha1` files from the build? Do we still need those?

johngmyers:
I've been thinking of ripping out the `.sha1` files from the build. I don't think we need them ourselves, but I don't know what might be depending on them downstream.

hakman:
There shouldn't be anything downstream that cares about the upstream build files, but we can bring it up during office hours.

PR #8145, opened 2019-12-18
Enabling JSON output for Terraform instead of writing the HCL syntax …

mccare:
Enabling output of JSON instead of HCL v1, writing kubernetes.tf.json instead of kubernetes.tf. JSON can be consumed by terraform 0.12.  ( https://www.terraform.io/docs/configuration/syntax-json.html )

Tried it on AWS and GCE test cluster. 

If you enable this feature, you need to delete the previously generated kubernetes.tf file. Also, this feature will require minimum version of terraform 0.12.

PR #8926, opened 2020-04-16
Automated cherry pick of #8145: Enabling JSON output for Terraform instead of writing the HCL

johanneswuerbach:
Cherry pick of #8145 on release-1.17.

#8145: Enabling JSON output for Terraform instead of writing the HCL


rifelpet:
I think this makes sense to cherry-pick. Months ago when #8145 landed we discussed it targeting 1.18 and the native hcl2 syntax targeting 1.19. We managed to get the hcl2 syntax into 1.18, so it might be good to offer this as an intermediate step in 1.17. The renaming of resources was also cherry-picked back to 1.17, so this should be the only thing needed to support terraform 0.12 in kops 1.17.

The PR itself is very low risk: the only logic changes are behind a feature flag that defaults to disabled.

Even though we're pretty late into the 1.17 release cycle I think this is safe and would help users with their terraform 0.12 adoption. Lets do it :)

Perhaps its worthy of a mention in the 1.17 release notes? Would you be able to add a note in the Significant Changes section [here](https://github.com/kubernetes/kops/blob/master/docs/releases/1.17-NOTES.md) in a followup PR? It should probably mention the "required action" item regarding resource renaming too. If not I can do it.
 
/lgtm
/approve

PR #8928, opened 2020-04-16
Add ability to set various cilium flags through CLI

olemarkus:
Needed for a new e2e cilium test

### DIFF
--- a/pkg/commands/set_cluster_test.go
+++ b/pkg/commands/set_cluster_test.go
@@ -70,3 +70,55 @@ func TestSetClusterFields(t *testing.T) {
 
 	}
 }
+
+func TestSetCiliumFields(t *testing.T) {
+
+	grid := []struct {
+		Fields []string
+		Input  kops.Cluster
+		Output kops.Cluster
+	}{
+		{
+			Fields: []string{
+				"cluster.spec.networking.cilium.ipam=eni",
+				"cluster.spec.networking.cilium.enableNodePort=true",
+				"cluster.spec.networking.cilium.disableMasquerade=true",
+				"cluster.spec.kubeProxy.enabled=false",
+			},
+			Input: kops.Cluster{
+				Spec: kops.ClusterSpec{},
+			},
+			Output: kops.Cluster{
+				Spec: kops.ClusterSpec{
+					KubeProxy: &kops.KubeProxyConfig{
+						Enabled: fi.Bool(false),
+					},
+					Networking: &kops.NetworkingSpec{
+						Cilium: &kops.CiliumNetworkingSpec{
+							Ipam:              "eni",
+							EnableNodePort:    true,
+							DisableMasquerade: true,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for _, g := range grid {
+		var igs []*kops.InstanceGroup
+		c := g.Input
+
+		err := SetClusterFields(g.Fields, &c, igs)
+		if err != nil {
+			t.Errorf("unexpected error from setClusterFields %v: %v", g.Fields, err)
+			continue
+		}
+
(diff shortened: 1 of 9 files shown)


olemarkus:
Want to get `etcdManaged` in as well.

pkg/commands/set_cluster_test.go
```diff
@@ -70,3 +70,26 @@ func TestSetClusterFields(t *testing.T) {
...
+			Kubelet: &kops.KubeletConfigSpec{},
+		},
+	}
+
+	var igs []*kops.InstanceGroup
+
+	err := SetClusterFields(fields, cluster, igs)
+	if err != nil {
+		t.Errorf("unexpected error from setClusterFields %v: %v", cluster, err)
+	}
```
rifelpet, line 94: can you assert that the fields are correctly set in the `cluster` object?

```
if cluster.Spec.Networking.Cilium.Ipam != "eni" {
	t.Errorf("...")
}
```
etc

rifelpet:
In order to test the etcdManaged under e2e tests we'll have to figure out how to add an additional etcdCluster named `cilium` I'm struggling to come up with a simple solution for that...

pkg/commands/set_cluster_test.go
olemarkus: Yep. Just since there are pointers involved, I need to properly demarshal the spec to yaml, I guess

olemarkus:
Yeah. I am wondering if the only feasible option is a proper cli flag.

rifelpet:
/lgtm

Thoughts on the new cli flag @justinsb or anyone else?
/assign @justinsb

olemarkus:
How far away is this one from approval now?

hakman:
One thought here: instead of a new flag would it be a good idea to reuse the networking one, like `--networking=cilium-etcd`? I think there are similar flags for Flannel.

olemarkus:
I am not a big fan of adding more to `--networking`. `--networking` is a significant flag and adding something that is just to work around limitation of `kubetest` isn't that user friendly I think.

hakman:
I am debating here that it may be the easiest way to get people started with Cilium + Etcd. It is not just good for kubetest, but would generate the correct config automatically, instead of doing some steps prone to user error.
From what I see this seems to be the way to activate a particular config for a certain network plugin.

olemarkus:
It may be an idea to rethink this a bit. Nodeport and ENI is the most important features to test. For networking values, these two also makes more sense perhaps. 

etcd is less important to test. Also, if etcd-manager works well, it may also be an idea to switch that on by default, which means that cli flag is perhaps not that useful.

hakman:
My take on this is that you should have a default as close as possible to upstream that will work without issues to create smaller test clusters. Think about adding various `--networking=cilium-<type>` flags to create more complex setups.
In any case, this is mostly a nit. Most work on this is really ok. Let's see also if @johngmyers has some feedback.

johngmyers:
There appears to be precedent in `flannel-vxlan` vs. `flannel-udp` for putting it in `--networking`.  Unfortunately, Cilium appears to have three independent modes, so how would one say one wants Etcd and Nodeport, but not ENI?

We could do something like `--networking=cilium=etcd,nodeport` but that's getting unwieldy. The only other idea I have is a `--networking-option` flag of type `[]string`.

olemarkus:
With #9425 the more complex option will be on by default. ENI mode could be interesting to test still though.

olemarkus:
Rebased to #9490

hakman:
@rifelpet I think this is ready.
/lgtm

rifelpet:
ok, i took a final pass and it looks good to me and is pretty straight forward 👍 thanks for sticking with it through all the rebases.

/approve

Issue #8244, opened 2020-01-02
Import package aliases are not uniform.


yuxiaobo96:
In kops, there are some import package aliases are not uniform.
For `"k8s.io/kops/pkg/apis/kops"`, there are several aliases:
```go
"k8s.io/kops/pkg/apis/kops"
kops "k8s.io/kops/pkg/apis/kops"
api "k8s.io/kops/pkg/apis/kops"
kopsapi "k8s.io/kops/pkg/apis/kops"
```
For `"k8s.io/apimachinery/pkg/apis/meta/v1"`, there are several aliases:
```go
v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
Therefore, I suggest to update:
`"k8s.io/kops/pkg/apis/kops"` -> `kopsapi "k8s.io/kops/pkg/apis/kops"`
`v1 "k8s.io/apimachinery/pkg/apis/meta/v1"` -> `metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"`

yuxiaobo96:
/cc @hakman 
/cc @rifelpet 
Please give some suggestions.

tanjunchen:
Good suggestion.why not?

tomesm:
Hi, can I also participate. I am a newbie and this seems to be a task which I could manage. I have already signed the CLA.

I am not sure if I can open the PR and reference it to this issue without asking thats why I am asking here in advance :)

rifelpet:
I think the open PR #8256 might be addressing the remaining imports mentioned in this issue but if you find other imports with similar inconsistencies feel free to mention them here and open a PR :)

tomesm:
@rifelpet Actually, there are lots of such files:

> cmd/kops/create_ig.go
cmd/kops/export_kubecfg.go
cmd/kops/get_cluster.go
cmd/kops/rollingupdatecluster.go
cmd/kops/delete_cluster.go
cmd/kops/upgrade_cluster.go
cmd/kops/validate_cluster.go
upup/pkg/fi/cloudup/loader.go
upup/pkg/fi/cloudup/spec_builder.go
upup/pkg/fi/cloudup/networking.go
pkg/model/components/containerd_test.go
upup/pkg/fi/cloudup/deepvalidate_test.go
upup/pkg/fi/cloudup/populatecluster_test.go
upup/pkg/fi/cloudup/populate_cluster_spec.go
[...]

PR #8933, opened 2020-04-17
Import package aliases modified

tomesm:
ref: #8244 

Updated import package aliases:

`api "k8s.io/kops/pkg/apis/kops"` -> `kopsapi "k8s.io/kops/pkg/apis/kops"`


tomesm:
I've signed up the individual CLA. Please verify it.

Issue #8932, opened 2020-04-17
Add support for ARM64


hakman:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

AWS is ready to GA the [Graviton 2](https://aws.amazon.com/about-aws/whats-new/2019/12/announcing-new-amazon-ec2-m6g-c6g-and-r6g-instances-powered-by-next-generation-arm-based-aws-graviton2-processors/) instance types **(C6g, M6g, R6g)**.
Performance looks great and may become quite popular, so it would be nice if Kops could use them as **worker nodes**.

Benchmarks:
https://www.anandtech.com/show/15578/cloud-clash-amazon-graviton2-arm-against-intel-and-amd/10

**2. Feel free to provide a design supporting your feature request.**

Kops generates a list of assets that is added to the Additional User Data together as part of the NodeUp config. At the moment this only provides AMD64 assets, but could provide also other architectures like ARM64.
NodeUp should easily detect the OS architecture and pull the corresponding assets.
As long as there is a Docker version that supports ARM64, the node should start and join the cluster without any issues.

Issue #6140, opened 2018-12-01
Support of AWS a1.* instances (Arm-based)


styk-tv:
With the a1.* instances being significantly cheaper, would it make sense at support of a1.* aws instances?

Type Spec:
https://aws.amazon.com/ec2/instance-types/a1/

brandon-dacrib:
It looks like it is in master so we need to wait for a release or do a build yourself. https://github.com/kubernetes/kops/blob/b3a450119faf8678afa09d1fd76fdc2a9dec37d2/upup/pkg/fi/cloudup/awsup/machine_types.go

dobesv:
Just because it was added to the list of instance types that exist doesn't mean anything will work on it.  Probably you need an OS image specifically built for ARM and all your docker containers will have to have ARM images.

apjneeraj:
It does not work! cloud-init script fails due to exec error. I tried using arm based ubuntu ami.

> Aug  1 03:50:04 ip-172-2-96-123 cloud-init[2095]: == Downloaded https://kubeupv2.s3.amazonaws.com/kops/1.12.2/linux/amd64/nodeup (SHA1 = e7e49d1fff61badc2fff402e80858d081aa88a26) ==
Aug  1 03:50:04 ip-172-2-96-123 cloud-init[2095]: Running nodeup
Aug  1 03:50:04 ip-172-2-96-123 cloud-init[2095]: /var/lib/cloud/instance/scripts/part-001: line 133: ./nodeup: cannot execute binary file: Exec format error

brandon-dacrib:
That binary is for x86-64. You need to acquire or build one for arm64 and then modify things to pull that. I would start by sorting out what process in the project is building that binary and then submitting a PR modifying that to build for both x86-64 + aarch64.

hakman:
M6g, C6g, and R6g have been [announced](https://aws.amazon.com/about-aws/whats-new/2019/12/announcing-new-amazon-ec2-m6g-c6g-and-r6g-instances-powered-by-next-generation-arm-based-aws-graviton2-processors/) and looks like the performance is great:
https://www.anandtech.com/show/15578/cloud-clash-amazon-graviton2-arm-against-intel-and-amd/10

I will start experimenting with ARM support for Kops. Worker nodes should be pretty easy to configure.

/reopen

jsleeio:
@hakman expect pain if you try to use the `imagebuilder`/`bootstrap-vz` tooling to build Debian Stretch ARM AMIs :-(

[...]

PR #8938, opened 2020-04-18
Add initial support for ARM64

hakman:
This adds support for ARM64 **worker nodes**. Masters still require AMD64.

Fixes https://github.com/kubernetes/kops/issues/8932
Fixes https://github.com/kubernetes/kops/issues/6140


.travis.yml
```diff
@@ -10,7 +13,7 @@ go:
 go_import_path: k8s.io/kops
 
 script:
-  - GOPROXY=https://proxy.golang.org make nodeup examples test
+  - GOPROXY=https://proxy.golang.org travis_wait 20 make nodeup examples test
```
rifelpet, line 16: why is this wait required? just curious

hakman: For some reason there is no output for too much time in ARM tests and they time-out. It may be related to some changes Travis folks did, as it used to work much faster a few days ago.
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

hakman: This is one of the failed tests: https://travis-ci.org/github/kubernetes/kops/builds/676536208

hatemosphere:
M6g are in GA now, so this could bring even more value now...

hakman:
@hatemosphere should be part of Kops 1.19. At the moment, I am waiting for the branch to be created to get started with the review and various improvements.

mikesplain:
@hakman This looks pretty good, you definitely need to rebase though before we can do a final review.  This is quite a big PR... from the looks, there's not much we can do to break it up but it would be great if you thought of ways to make this more easily reviewable.

hakman:
Thanks @mikesplain. I've been also thinking of ways to break this up. I think things changed quite a bit since I worked on this PR. Still waiting for the 1.18 branch to be created to do a rebase and go ahead with the review.

upup/pkg/fi/cloudup/urls.go
```diff
@@ -118,19 +119,20 @@ func SetKopsAssetsLocations(assetsBuilder *assets.AssetBuilder) error {
...
+	if nodeUpAsset == nil {
+		nodeUpAsset = make(map[architectures.Architecture]*MirroredAsset)
+	} else if nodeUpAsset[arch] != nil {
 		// Avoid repeated logging
-		klog.V(8).Infof("Using cached nodeup location: %v", nodeUpAsset.Locations)
-		return nodeUpAsset, nil
+		klog.V(8).Infof("Using cached nodeup location: %v", nodeUpAsset[arch].Locations)
+		return nodeUpAsset[arch], nil
 	}
 	env := os.Getenv("NODEUP_URL")
```
coreypobrien, line 130: We're going to need to change this logic to support different arch urls
```suggestion
	envvar := strings.ToUpper(fmt.Sprintf("NODEUP_URL_%s", arch))
	env := os.Getenv(envvar)
```

and probably also change the messages to `return nil, fmt.Errorf("unable to parse env var %s %q as a url: %v", envvar, env, err)` and 
`klog.Warningf("Using nodeup location from %s env var: %q", envvar, u.String())` below this too

hakman: Thanks for the feedback @coreypobrien.
I think I will drop completely the requirement for 1.19+. Works pretty well with older versions.
The only supported Docker version will be 19.03.11 (or whatever the default will be in Kops 1.19). Do you see any reason to ad support for older versions?
Adding env vars for multi-arch nodeup seems like a good idea, but should be kept along with the simple one, that will work only on AMD64.

coreypobrien: Agreed, I don't know of any reason we'd have to do older docker versions. Newest on each OS seems reasonable to me.

On the variables, we need to support supplying both the amd and the arm override (as well as future architectures?) so I thought it would be clearer with `NODEUP_URL_AMD64` and `NODEUP_URL_ARM64`, but as long as we can specify one per architecture the naming doesn't matter much.

johngmyers: Perhaps allow `NODEUP_URL_AMD64`, with fallback to `NODEUP_URL` if that doesn't exist.

hakman: Yes, that is what I meant, but this will probably be the only env var for multi-arch.
For people that want more, there is `KOPS_BASE_URL` and that should be enough.

upup/pkg/fi/cloudup/urls.go
```diff
@@ -118,39 +119,44 @@ func SetKopsAssetsLocations(assetsBuilder *assets.AssetBuilder) error {
...
 		// Avoid repeated logging
-		klog.V(8).Infof("Using cached nodeup location: %v", nodeUpAsset.Locations)
-		return nodeUpAsset, nil
+		klog.V(8).Infof("Using cached nodeup location: %v", nodeUpAsset[arch].Locations)
+		return nodeUpAsset[arch], nil
+	}
+	// Use multi-arch env var, but fall back to well known env var
+	env := os.Getenv(fmt.Sprintf("NODEUP_URL_%s", strings.ToUpper(string(arch))))
+	if env == "" {
+		env = os.Getenv("NODEUP_URL")
```
johngmyers, line 133: I was thinking we'd only fall back to this for the amd64 arch.

hakman: I think it doesn't really matter much. If someone want to use if for a full ARM64 cluster, I don't have anything against it. I personally prefer the `KOPS_BASE_URL` version.

hakman:
@justinsb @mikesplain @rifelpet @johngmyers 
This is pretty much ready for review from my point of view.
Individual commits have suggestive names and are fairly simple.

The main change is to generate a multi-arch config map for assets and images in `kube_env.yaml` and let NodeUp decide what to load based on running architecture.

The `kube_env.yaml` file would change from:
```yaml
Assets:
- 769fa6399ecf541ced7ff4950974fd3d29f66f0e9d30c34a708f4c465e8e1908@https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/amd64/kubelet
- 02def7a9a6972942b76b7956f24cb09c0266f64b35710eb324bc80a7c912e306@https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/amd64/kubectl
- 994fbfcdbb2eedcfa87e48d8edb9bb365f4e2747a7e47658482556c12fd9b2f5@https://storage.googleapis.com/k8s-artifacts-cni/release/v0.8.6/cni-plugins-linux-amd64-v0.8.6.tgz
images:
- hash: ef693546847a16b083165317be4b37c98ac6c6988db3b6284971cf96014a021c
  sources:
  - https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/amd64/kube-proxy.tar
```
to:
```yaml
Assets:
  amd64:
  - 769fa6399ecf541ced7ff4950974fd3d29f66f0e9d30c34a708f4c465e8e1908@https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/amd64/kubelet
  - 02def7a9a6972942b76b7956f24cb09c0266f64b35710eb324bc80a7c912e306@https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/amd64/kubectl
  - 994fbfcdbb2eedcfa87e48d8edb9bb365f4e2747a7e47658482556c12fd9b2f5@https://storage.googleapis.com/k8s-artifacts-cni/release/v0.8.6/cni-plugins-linux-amd64-v0.8.6.tgz
  arm64:
  - d43697504aaebcb1610c7174c1ed8d732edeeeefaa490640acc0e7193e2c9c89@https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/arm64/kubelet
  - 36527838d2ce4560c10f6b726231b23b4ab1dee70126b2a759e10436b1bcb84c@https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/arm64/kubectl
  - 43fbf750c5eccb10accffeeb092693c32b236fb25d919cf058c91a677822c999@https://storage.googleapis.com/k8s-artifacts-cni/release/v0.8.6/cni-plugins-linux-arm64-v0.8.6.tgz
images:
  amd64:
  - hash: ef693546847a16b083165317be4b37c98ac6c6988db3b6284971cf96014a021c
    sources:
    - https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/amd64/kube-proxy.tar
  arm64:
  - hash: c9f2f2648039840d754c11b4c735c0b2dbf7c14486a706d12cfc4ef3fb1ed2a9
    sources:
    - https://storage.googleapis.com/kubernetes-release/release/v1.19.0-beta.2/bin/linux/arm64/kube-proxy.tar
```

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1277,67 +1310,79 @@ func (c *ApplyClusterCmd) BuildNodeUpConfig(assetBuilder *assets.AssetBuilder, i
...
 	config.InstanceGroupName = ig.ObjectMeta.Name
 
 	useGossip := dns.IsGossipHostname(cluster.Spec.MasterInternalName)
 	isMaster := role == kops.InstanceGroupRoleMaster
 
-	var images []*nodeup.Image
-
+	images := make(map[architectures.Architecture][]*nodeup.Image)
 	if components.IsBaseURL(cluster.Spec.KubernetesVersion) {
-		// When using a custom version, we want to preload the images over http
```
johngmyers, line 1293: Why remove this comment?

upup/pkg/fi/cloudup/urls.go
```diff
@@ -118,39 +119,44 @@ func SetKopsAssetsLocations(assetsBuilder *assets.AssetBuilder) error {
...
-	// Avoid repeated logging
-	if nodeUpAsset != nil {
+func NodeUpAsset(assetsBuilder *assets.AssetBuilder, arch architectures.Architecture) (*MirroredAsset, error) {
+	if nodeUpAsset == nil {
+		nodeUpAsset = make(map[architectures.Architecture]*MirroredAsset)
+	} else if nodeUpAsset[arch] != nil {
 		// Avoid repeated logging
-		klog.V(8).Infof("Using cached nodeup location: %v", nodeUpAsset.Locations)
-		return nodeUpAsset, nil
+		klog.V(8).Infof("Using cached nodeup location: %v", nodeUpAsset[arch].Locations)
```
johngmyers, line 127: Include arch in log?

pkg/model/bootstrapscript.go
```diff
@@ -39,9 +39,11 @@ import (
 
 // BootstrapScript creates the bootstrap script
 type BootstrapScript struct {
-	NodeUpSource        string
-	NodeUpSourceHash    string
-	NodeUpConfigBuilder func(ig *kops.InstanceGroup) (*nodeup.Config, error)
+	NodeUpSourceAmd64     string
+	NodeUpSourceHashAmd64 string
+	NodeUpSourceArm64     string
+	NodeUpSourceHashArm64 string
```
johngmyers, line 45: Perhaps these would be cleaner as two maps of arch to string?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1138,64 +1141,94 @@ func (c *ApplyClusterCmd) AddFileAssets(assetBuilder *assets.AssetBuilder) error
...
 		if err != nil {
-			return fmt.Errorf("unable to parse lyft-vpc URL %q", urlString)
+			return err
 		}
+		c.Assets[arch] = append(c.Assets[arch], BuildMirroredAsset(cniAsset, cniAssetHash))
+
+		if c.Cluster.Spec.Networking.LyftVPC != nil {
+			var hash *hashing.Hash
+
+			urlString := os.Getenv("LYFT_VPC_DOWNLOAD_URL")
```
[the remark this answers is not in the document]
hakman: I don't think it's worth the trouble. I don't see a good reason for any env var to exist in this case.
Most likely any test will be done on AMD64, or we can address later if there is high demand for Lyft.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1138,64 +1141,94 @@ func (c *ApplyClusterCmd) AddFileAssets(assetBuilder *assets.AssetBuilder) error
...
-	asset, err := NodeUpAsset(assetBuilder)
-	if err != nil {
-		return err
+			c.Assets[arch] = append(c.Assets[arch], BuildMirroredAsset(u, hash))
+		}
+
+		// TODO: Update Kops version in integration tests to 1.19 after 1.19 is released
+		// Integration tests fake the Kops version to 1.15 and will not be able to find NodeUp
+		kopsHasArm64 := false
+		kopsVersion, err := semver.Parse(kopsbase.Version)
```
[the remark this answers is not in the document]
hakman: The problem is that there is no released NodeUp binary to check hashes.

pkg/model/bootstrapscript.go
hakman: I thought of that, but in the bootstrap script they are separate vars (bash). This way keeps the same style.

upup/pkg/fi/cloudup/urls.go
hakman: I thought that most locations already have the arch in the name.

pkg/model/bootstrapscript.go
johngmyers: This way complicates initializing the struct. They can still be four separate template functions.

upup/pkg/fi/cloudup/apply_cluster.go
hakman: Should be added back.

pkg/bundle/builder.go
```diff
@@ -171,15 +172,19 @@ func (b *Builder) Build(cluster *kops.Cluster, ig *kops.InstanceGroup) (*Data, e
 
 		bootstrapScript := model.BootstrapScript{}
 
-		{
-			asset, err := cloudup.NodeUpAsset(assetBuilder)
-			if err != nil {
-				return nil, err
-			}
+		assetAmd64, err := cloudup.NodeUpAsset(assetBuilder, architectures.ArchitectureAmd64)
```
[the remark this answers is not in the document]
hakman: I think it still has to be there for now to pass the build and tests. But good to know. Thanks.

pkg/model/bootstrapscript.go
hakman: Done. It looks better now.

upup/pkg/fi/cloudup/apply_cluster.go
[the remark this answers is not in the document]
johngmyers: I was suggesting that the tests here instead of comparing against modern or CI versions do an inverted compare against "1.15.0".

upup/pkg/fi/cloudup/urls.go
johngmyers: Maybe, maybe not. Could be a bug where the arch in the location is wrong.

hakman: OK, done.

upup/pkg/fi/cloudup/apply_cluster.go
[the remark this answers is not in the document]
hakman: Does it look better now?

upup/pkg/fi/cloudup/apply_cluster.go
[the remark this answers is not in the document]
hakman: Sounds ok?

pkg/bundle/builder.go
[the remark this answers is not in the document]
hakman: Will rebase once #9372 is merged.

upup/pkg/fi/cloudup/apply_cluster.go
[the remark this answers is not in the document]
johngmyers: Sounds ok.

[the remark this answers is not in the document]
hakman: Thanks

hakman:
Rebased PR after https://github.com/kubernetes/kops/pull/9340 and also added some tests for side-loading.

nodeup/pkg/model/tests/golden/minimal/cluster.yaml
```diff
@@ -29,7 +29,7 @@ spec:
       backupStore: memfs://clusters.example.com/minimal.example.com/backups/etcd-events
   kubelet:
     anonymousAuth: false
-  kubernetesVersion: v1.18.0
+  kubernetesVersion: https://storage.googleapis.com/kubernetes-release/release/v1.18.0
```
johngmyers, line 32: Why the full URL?

hakman: This activates side-loading and the `-amd64`/`-arm64` suffix for the images.

johngmyers: I'm concerned about this invalidating a lot of the golden tests, as they would no longer cover the common case of no sideloading. How would our tests detect a bug that is masked by sideloading? Our e2e tests certainly wouldn't find it; the only choices would be unit, integration, or beta.

hakman: It's just an image name, nothing special. I could create a new copy of the tests for this if you feel so strongly, or just drop this commit.

johngmyers: If a new copy of the tests for this would give you adequate coverage then I'd prefer that approach. Or would you need to duplicate the whole row of golden tests to make sure each builder handles sideloading correctly?

hakman: We pretty much need to duplicate the `minimal` dir to a new one called `side-loading`.
Just pushed this as a change.

johngmyers: Hrm, this is what I meant by "need to duplicate the whole row of golden tests". I concede my concern about not covering the no-sideloading case is probably not worth the resulting cost. You can rip that out if you like. Perhaps a future PR could make the AMD/ARM/no-sideload matrix be table-driven, with text repacement to the golden file in order to form the expected output.

johngmyers:
/hold
in case you want to walk back that duplication. Feel free to cancel.

nodeup/pkg/model/tests/golden/minimal/cluster.yaml
hakman: Would it be ok to keep it as is (duplicated) for now and will try to de-duplicate it in a future PR?
Except for the fact that we get to have duplicated `cluster.yaml` file things don't change so much.

hakman:
Let's see what others think about this. The "hold" part may make it look like it's not ready for review.
/hold cancel

hakman:
Rebased PR after #9391.

mikesplain:
Thanks @hakman!

/lgtm
/approve

hakman:
Had to to one more rebase to restart the failed Netlify tests.
Thanks @mikesplain!

lubinszARM:
Hi @hakman 
A quick question.
I'm from Arm.
Currently we also want to run prow jobs on an arm64 cluster.
In the following file, I can see that you have created similar jobs.
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kops/kops-periodics-misc.yaml#L106

But  I'm curious that,  it seems gcr.io/k8s-testimages/kubekins-e2e:v20200623-2424179-master/bootstrap/sidecar/initupload/clonerefs does not support Arm64.
So,  how did you run the test on  an arm cluster?

Reference chart of our prow test for Arm64:
image

Related issue: 
https://github.com/kubernetes/test-infra/issues/16588
https://github.com/kubernetes/kubernetes/issues/86295

hakman:
@lubinszARM in this context, our setup just uses Kops to create a cluster in AWS and run tests remotely on it. Kops takes care of everything except running the test.

zhlhahaha:
Hi, @hakman 
I just read the PR https://github.com/kubernetes/test-infra/pull/18034
For the job "e2e-kops-aws-misc-arm64", it uses image "gcr.io/k8s-testimages/kubekins-e2e:v20200623-2424179-master". However, the image is not support arm64 platform yet. 
But, according to the result of the prow jobs, it looks working well. https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kops#kops-aws-arm64
How it works?

hakman:
Hi @zhlhahaha @lubinszARM.

That image is started on AMD64 setup and runs a Kops binary to create a remote cluster in AWS. When that cluster is ready the e2e tests are run on the remote cluster: https://github.com/kubernetes/kops/blob/master/docs/development/testing.md.
Kops does things a bit differently than normal tests, so it is very flexible to try multiple screnarios, (different CNI plugins, different runtimes, ...).

So, there is no need in our case for a multi-arch image for kubekins-e2e, just support in tests. As you can see, there are very few issues remaining there.

There are also some issues with the build marker that is updated before the cross-build is done and tests fail because the ARM64 CI binaries are missing. There is a discussion about this in #sig-testing and hopefully there will be a solution soon.

Thanks for the referenced issues. Just browsed them, but will make some interesting read later.

Nice to see that Arm is pushing things in this direction!

zhlhahaha:
Hi @hakman, I have read through the testing.md and it is quit interesting.
Thanks for your explaination!

atam003:
@hakman 
Will these changes work for kops for 1.18. if I try to back merge in the workspace.
I don't want to move to kops 1.19 as I see a few breaking (not backward compatible) changes (also it is a beta version).

hakman:
@atam003 these changes will not work in kOps 1.18, there are too many changes.
kOps 1.19 will be released today so, no longer beta. In case it helps, you can use k8s 1.18 with it and enable ARM64 also.

atam003:
@hakman 
I still see the kops 1.19 beta version in Github. Any idea when it will be updated on GitHub or where should I look.

PR #8939, opened 2020-04-19
Partially implement MountSensitive

justinsb:
We still don't actually support sensitive options, but it doesn't look
like we need them, MountSensitive is just called with empty
sensitiveOptions in a few places.

Issue #8908


hakman:
Thanks @justinsb.
/lgtm

hakman:
@justinsb This worked nicely for 1.11.
1.10 is still broken, this time because of the default image, based on Debian Jessie. Seems that packages already started disappearing. If 1.10 is going to work with the default image, I think it will need to be rebuilt based on Debian Stretch.

https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-1-10/1252182276532342790/artifacts/15.206.178.213/kops-configuration.log

PR #8940, opened 2020-04-20
Use latest patch version for GO in tests

hakman:
At the moment, Travis runs tests using the `.0` version of GO.
Would be better if Travis would use the latest patch version of GO, which is closer to what is used in official builds.
https://docs.travis-ci.com/user/languages/go/#specifying-a-go-version-to-use


rifelpet:
ah nice, i would've thought thats what major.minor would do automatically.

/lgtm
/approve

PR #8941, opened 2020-04-20
Add support for containerd v1.3.4

hakman:
https://github.com/containerd/containerd/releases/tag/v1.3.4

### DIFF
--- a/pkg/model/components/containerd.go
+++ b/pkg/model/components/containerd.go
@@ -53,8 +53,8 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 
 		// Set containerd based on Kubernetes version
 		if fi.StringValue(containerd.Version) == "" {
-			if b.IsKubernetesGTE("1.17") {
-				containerd.Version = fi.String("1.3.3")
+			if b.IsKubernetesGTE("1.18") {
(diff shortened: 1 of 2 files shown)


pkg/model/components/containerd.go
```diff
@@ -53,8 +53,8 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 
 		// Set containerd based on Kubernetes version
 		if fi.StringValue(containerd.Version) == "" {
-			if b.IsKubernetesGTE("1.17") {
-				containerd.Version = fi.String("1.3.3")
+			if b.IsKubernetesGTE("1.18") {
```
rifelpet, line 56: why is this version changing? i'd expect to see another `else if` condition instead of updating this one condition

hakman: We set this to 1.17 a few months ago to be able to test while 1.18 was unstable. The expected release version is 1.18.

hakman: https://github.com/kubernetes/kops/pull/8283

rifelpet: ah ok, and that PR was also only in kops 1.18. 👍

hakman:
Thanks @rifelpet :)

PR #8942, opened 2020-04-20
[WIP] Test ARM64 e2e

hakman:


hakman:
Tests work just partially because the tests pull the AMD64 image for busybox and some others. Probably they are not meant for mixed envs. It is goot to know though that **636** tests pass.

PR #8946, opened 2020-04-20
bastion: honor ig fields

### LINKED ISSUE #5242: fileAssets and hooks not honored on Bastion roles
KOPS VERSION
1.9.1

KUBERNETES VERSION
1.9.8

CLOUD PROVIDER
AWS

REPRO STEPS:
1. Set a fileAsset and hook in the cluster config and assign it to the Bastion, Node and Master roles
2. Deploy an IG with the Bastion Role
3. Deploy an IG with the Node Role
4. Deploy an IG with the Master Role

ACTUAL:
The bastion spins up, but no fileAssets are created and no hooks are issued. fileAssets and hooks are created just fine on the node and master IGs

EXPECTED:
The fileAsset would be created and the hook would be followed for the bastion, as it is on the node and master

CLUSTER MANIFEST: 
https://gist.github.com/jaumann/93a9aca4cf6fec8caa98498e7b8628e4

nvanheuverzwijn:
**Description**
This PR tries to implements all of the IG spec of bastions.
It requires a new version of nodeup.

In my test, it seems to work quite well.

fix #5242

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -244,13 +244,12 @@ func (b *PolicyBuilder) BuildAWSPolicyBastion() (*Policy, error) {
 		Version: PolicyDefaultVersion,
 	}
 
-	// Bastion hosts currently don't require any specific permissions.
-	// A trivial permission is granted, because empty policies are not allowed.
-	p.Statement = append(p.Statement, &Statement{
-		Effect:   StatementEffectAllow,
-		Action:   stringorslice.Slice([]string{"ec2:DescribeRegions"}),
-		Resource: resource,
-	})
+	addBastionEC2Policies(p, resource)
+
+	var err error
+	if p, err = b.AddS3Permissions(p); err != nil {
+		return nil, fmt.Errorf("failed to generate AWS IAM S3 access statements: %v", err)
+	}
 
 	return p, nil
 }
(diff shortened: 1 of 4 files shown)


pkg/model/iam/iam_builder.go
```diff
@@ -244,13 +244,12 @@ func (b *PolicyBuilder) BuildAWSPolicyBastion() (*Policy, error) {
...
 	}
 
-	// Bastion hosts currently don't require any specific permissions.
-	// A trivial permission is granted, because empty policies are not allowed.
-	p.Statement = append(p.Statement, &Statement{
-		Effect:   StatementEffectAllow,
-		Action:   stringorslice.Slice([]string{"ec2:DescribeRegions"}),
-		Resource: resource,
-	})
+	addNodeEC2Policies(p, resource)
```
olemarkus, line 247: Maybe rename this function to `addBasicEC2Policies` or something? I got a bit confused when bastion called this one.

pkg/model/iam/iam_builder.go
```diff
@@ -244,13 +244,12 @@ func (b *PolicyBuilder) BuildAWSPolicyBastion() (*Policy, error) {
...
+
+	var err error
+	if p, err = b.AddS3Permissions(p); err != nil {
```
olemarkus, line 250: Should Bastion get all of these permissions in s3? Letting it access config store makes sense for what is happening here, but the secrets seems like something one should avoid.

nvanheuverzwijn: I was wondering the same. Nodeup is still kind of magic to me, so i'm still researching how it works exactly.
I will try to pin point exactly what permission it needs.

pkg/model/iam/iam_builder.go
nvanheuverzwijn: I think I will create a new function called `addBastionEC2Policies` instead.

nvanheuverzwijn: @olemarkus Made a change for this function.

pkg/model/iam/iam_builder.go
nvanheuverzwijn: @olemarkus Removed S3 permission for pki sub-directory

olemarkus:
On paper this looks good to me. but more people need to weigh in here. This is a pretty big change to bastions.

/ok-to-test

nvanheuverzwijn:
@olemarkus Hey, I won't have time to work on this anymore :( Sorry about that :( Thank you for your time, I really appreciate it.

PR #8948, opened 2020-04-21
Update rotating secrets docs

olemarkus:


docs/rotate-secrets.md
```diff
@@ -10,16 +12,46 @@ kops get secrets  | grep ^Secret | awk '{print $2}' | xargs -I {} kops delete se
...
+
+```
+sudo find /mnt/ -name server.* | xargs -I {} sudo rm {}
+sudo find /mnt/ -name me.* | xargs -I {} sudo rm {}
+```
+
+You need to reboot every node (using a rolling-update). You have to use `--cloudonly` because the keypair no longer matches.
+
+```
+kops rolling-update cluster --cloudonly --force --yes
```
rifelpet, line 45: I haven't tested these instructions, but does the entire cluster need to be rolled or only the masters? Since the apiserver will essentially be down from the time you start deleting keys until the majority of them are rebooted, perhaps keeping the `--master-interval=10s` there would help minimize downtime.

If nodes do need to be replaced as well, perhaps we break it out into two rolling-update commands with only the masters using --cloudonly and the nodes being more graceful.

docs/rotate-secrets.md
```diff
@@ -35,9 +67,15 @@ for i in ${NS}; do kubectl get secrets --namespace=${i} --no-headers | grep "kub
 # Allow for new secrets to be created
 sleep 60
 
-# Bounce pods that we know use service account tokens - you will likely have to bounce more
-kubectl delete pods -lk8s-app=dns-controller --namespace=kube-system
-kubectl delete pods -lk8s-app=kube-dns --namespace=kube-system
-kubectl delete pods -lk8s-app=kube-dns-autoscaler --namespace=kube-system
+# Bounce all pods to make use of the new service tokens
 pkill -f kube-controller-manager
+kubectl delete pods --all --all-namespaces
```
rifelpet, line 72: Are the old pods still functional? I suppose any that depend on their service tokens no longer work, but other pods would still be functional. I'm mostly thinking about how we can be as least disruptive as necessary.

docs/rotate-secrets.md
olemarkus: I lost access to the API when I deleted all the secrets. I think all the nodes need to download the new certs and re-register to the cluster. I don't know if a reboot would do that, which could potentially be less disruptive.

docs/rotate-secrets.md
olemarkus: All pods that are not using the SA token is functional.

But we just rolled `--cloudonly` the entire cluster, so I figured this would be easier and faster than the user having to look for pods that may be using the API or using the token for other purposes.

docs/rotate-secrets.md
rifelpet: Ah so even after replacing just the masters, both you and the nodes still couldn't connect to the API. in that case it probably makes sense to rolling-update every master and node in the cluster all in one go.

rifelpet: Perhaps keep the `--master-interval=10s` ? no point in rolling those slowly if they're inaccessible. I do see an advantage of keeping the default node-interval, since it allows pods to be replaced gradually rather than having all of the nodes down at once. we wouldn't respect pod disruption budgets but at least 1 nodes' pods being lost at a time is better than all. Maybe we keep `--master-interval=10s` and mention that users may want to set a shorter --node-interval depending on their workloads and tolerance of downtime. Thoughts?

olemarkus: I would at least _not_ keep the master interval. The masters are not really doing anything useful: dns, api server etc will all be broken. There is nothing on the masters that work without the API and the API won't work until etcd is working again. I think you want to roll as fast as possible so new nodes can register (which requires quorum).

Masters typically take 5 min to roll. Rolling nodes faster than that won't really work since the nodes won't register with the masters until they are all up.

Nodes typically take 2 min to roll. We could say that it may be wise to wait 2-3 min to roll the nodes to minimize workload downtime.

Until we find a way to rotate this more automatic, I think one have to live with quite a bit of downtime here anyway though.

olemarkus: I think from user perspective it is easiest to either roll everything at once. Depending on workload one may want to do things differently, but you'd save 2-3 min out of maybe 10 min downtime. So I would leave it as-is in the docs.
It would be a fun exercise to look at how we can automate and do this more gracefully though.
By not rolling ca and keeping that key more safe than the other cert (maybe not even having it in S3) would save a lot of pain since the masters should accept the new certs immediately. Don't think you would have to roll all tokens this way either.

rifelpet:
ok, this looks good and if we can figure out a less invasive way to perform the rotating we can always update this later. thanks!
/lgtm
/approve

PR #8958, opened 2020-04-23
Turn on flannel sysctl fixes on rhel7 as well

justinsb:
It looks like the same problems are occuring on rhel7, so turn on the
fixes there also.


hakman:
@justinsb should this and the original be cherry-picked to 1.17 and 1.16

PR #8960, opened 2020-04-23
add some unit tests

q384566678:
Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

### DIFF
--- a/pkg/model/defaults/volumes_test.go
+++ b/pkg/model/defaults/volumes_test.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package defaults
+
+import (
+	"testing"
+
+	"k8s.io/kops/pkg/apis/kops"
+)
+
+func TestDefaultInstanceGroupVolumeSize(t *testing.T) {
+	tests := []struct {
+		role     kops.InstanceGroupRole
+		expected int32
+	}{
+		{
+			role:     "Node2",
+			expected: -1,
+		},
+	}
+	for _, test := range tests {
+		result, _ := DefaultInstanceGroupVolumeSize(test.role)
+		if test.expected != result {
+			t.Errorf("Expected %d, got %d", test.expected, result)
+		}
+	}
+}
--- a/upup/pkg/fi/utils/equals_test.go
+++ b/upup/pkg/fi/utils/equals_test.go
@@ -50,3 +50,34 @@ func TestStringSlicesEqual(t *testing.T) {
 		}
 	}
 }
+
+func TestStringSlicesEqualIgnoreOrder(t *testing.T) {
+	tests := []struct {
+		l        []string
+		r        []string
+		expected bool
+	}{
+		{
+			l:        []string{"a", "b"},
+			r:        []string{"a"},
+			expected: false,
+		},
+		{
+			l:        []string{"a", "b"},
+			r:        []string{"a", "c"},
+			expected: false,
+		},
+		{
+			l:        []string{"a", "b"},
+			r:        []string{"b", "a"},
+			expected: true,
+		},
+	}
+
+	for _, test := range tests {
+		result := StringSlicesEqualIgnoreOrder(test.l, test.r)
+		if test.expected != result {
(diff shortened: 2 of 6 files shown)


q384566678:
@geojaz @rifelpet @justinsb PTAL

q384566678:
@olemarkus PTAL

pkg/model/defaults/volumes_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+package defaults
+
+import (
+	"testing"
+
+	"k8s.io/kops/pkg/apis/kops"
+)
+
+func TestDefaultInstanceGroupVolumeSize(t *testing.T) {
```
geojaz, line 25: This test feels like it creates maintenance work rather than unit testing "kops logic". That is to say, `DefaultInstanceGroupVolumeSize` basically just returns different constants with a switch. 

IMO, changes that would break this test are likely to be intentional changes, so my vote would be to remove `volumes_test.go`.

upup/pkg/fi/utils/sanitize_test.go
```diff
@@ -0,0 +1,69 @@
...
+		s        string
+		expected string
+	}{
+		{
+			s:        "test",
+			expected: "test",
+		},
+		{
+			s:        "~/test",
+			expected: homedir.HomeDir() + "/test",
```
olemarkus, line 59: This test is pretty much the implementation of `ExpandPath`. I would change this to a hardcoded string.

upup/pkg/fi/utils/equals_test.go
```diff
@@ -0,0 +1,83 @@
...
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"testing"
+)
+
+func TestStringSlicesEqual(t *testing.T) {
```
geojaz, line 23: FYI: I think this is a much more useful unit test than the volumes_test.

upup/pkg/fi/utils/hash_test.go
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
geojaz, line 2: I think these should be bumped to 2020 now?

upup/pkg/fi/utils/hash_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"testing"
+)
+
+func TestHashString(t *testing.T) {
+	tests := []struct {
+		s           string
+		expectedStr string
+	}{
+		{
+			s:           "test",
```
geojaz, line 29: I think this is also a fairly useful test... could you add some more spicy or exciting edge cases? For these relatively simple functions, I would prefer to exercise the hard cases and assume the libraries generally do what they're supposed to.

upup/pkg/fi/utils/sanitize_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"k8s.io/client-go/util/homedir"
+)
+
+func TestSanitizeString(t *testing.T) {
+	tests := []struct {
+		s        string
+		expected string
+	}{
+		{
+			s:        "test",
```
geojaz, line 31: Again- are there any cases that better exercise the edges than these? These ones seem very easy :)

pkg/model/defaults/volumes_test.go
q384566678: Fine.

pkg/model/defaults/volumes_test.go
```diff
@@ -0,0 +1,54 @@
...
+			role:     "Node",
+			expected: 128,
+		},
+		{
+			role:     "Bastion",
+			expected: 32,
+		},
+		{
+			role:     "Node2",
+			expected: -1,
```
olemarkus, line 44: @geojaz this particular one could perhaps have some value. I suppose we want to preserve `-1` being the return value for unknown roles. The above could perhaps just test for positive integers instead.

olemarkus: If we were to add another role, it would at least be nice to see this test failing.

geojaz: Great point +1 @olemarkus

upup/pkg/fi/utils/sanitize_test.go
q384566678: Because the HomeDir obtained in different test environments is different, I think it is impossible to write a definite string.

upup/pkg/fi/utils/sanitize_test.go
q384566678: Thanks for your suggestions, I have updated the relevant tests. PATL.

upup/pkg/fi/utils/sanitize_test.go
olemarkus: The only relevant environment is linux. you can skip if the environment is something else. If not this test is rather meaningless.

q384566678: Getting the value of `homedir.HomeDir ()` in a linux environment is also different, so I will delete this test. Thanks!

PR #8965, opened 2020-04-23
Fix missing changes in Weave manifest

hakman:
Seems that these were missing compared to latest version on the weave site.

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
@@ -37,6 +37,14 @@ rules:
       - list
       - watch
   - apiGroups:
+      - extensions
+    resources:
+      - networkpolicies
+    verbs:
+      - get
+      - list
+      - watch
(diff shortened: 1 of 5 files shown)


upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
```diff
@@ -36,6 +36,14 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
```
olemarkus, line 39: Does weave actually fetch resources using the deprecated api groups? If not, this is not necessary. It doesn't do any harm either though.

hakman: Best to keep it as close as possible to their official manifest, I guess. Thanks for checking this.
https://github.com/weaveworks/weave/blob/9f00f78d3b9d5a8a31fdd90ec691095028e2690a/prog/weave-kube/weave-daemonset-k8s-1.11.yaml#L28-L35

PR #8971, opened 2020-04-24
Use correct flag --fail-on-drain-error in examples

pascal-hofmann:


rifelpet:
@pascal-hofmann thanks for the fix! can you sign the CLA according to the first comment on this PR?

Also I believe these markdown files are generated from the [.go files here](https://github.com/kubernetes/kops/blob/master/cmd/kops/rollingupdatecluster.go) so you'll need to update that too.

pascal-hofmann:
I've signed the CLA, fixed the error in the .go file and squashed my commits.

rifelpet:
there might be an issue with the git author email used in your commit. I see two GH usernames when viewing the commits for this PR. You'll need to fix that so that the author is the same email as your GH account and same as what the CLA was signed with.

pascalhofmann:
The emails match, but somehow github still thinks both my new and old github account committed. I've now signed the CLA with my old github account too.

rifelpet:
that works too 👍 
/ok-to-test

cmd/kops/rollingupdatecluster.go
```diff
@@ -92,7 +92,7 @@ var (
 		# only roll the node instancegroup,
 		# use the new drain and validate functionality.
 		kops rolling-update cluster k8s-cluster.example.com --yes \
-		  --fail-on-validate-error="false" \
+		  --fail-on-drain-error="false" \
```
johngmyers, line 95: I think the flag should be simply removed, since there is nothing in the preceding comment about not failing under a particular circumstance.

I also think you can remove the comment about "the new dran and validate functionality" since it isn't new anymore.

PR #8972, opened 2020-04-24
Migrates GCE sdk from v0.beta to v1

geojaz:
Waiting on #8969 

Bumps the gcp libraries from v0beta to v1! 
`&compute.InstanceGroupManagersResizeAdvancedRequest` is only available in the beta libraries. I've updated the IGM resize method to reflect the [docs](https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroupManagers/resize#go)


rifelpet:
you may need to restart the travis jobs through their UI, if you're able to. i can't...

geojaz:
yeah it doesn't look like i have access to restart that one, is it easy to add access for me too? :)

rifelpet:
are you logged into to travis? if you do that via your github account it may give you access to your own jobs and such. if you are already logged in and it still wont let you, i'm not sure how we can retry them other than pushing another commit :(

PR #8978, opened 2020-04-25
Don't install static utils for Flatcar

hakman:
Flatcar has just added `conntrack` to their distro and the static utils are no longer needed.
https://github.com/flatcar-linux/portage-stable/pull/48

/cc @justinsb @rifelpet


justinsb:
Thanks @hakman, excited to get rid of static utils soon :crossed_fingers: 

/lgtm
/approve

PR #8988, opened 2020-04-26
Use KOPS_RSA_PRIVATE_KEY_SIZE for more keys

johngmyers:
Also removes some redundant code.


olemarkus:
Feels like this is more about removing redundant code :)
But looks good to me if the refactor gives the same end result. We could do with more tests in this area.

PR #8989, opened 2020-04-26
Add unit test for func matchesElbTags

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/resources/aws/aws_test.go
+++ b/pkg/resources/aws/aws_test.go
@@ -23,8 +23,10 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/elb"
 	"k8s.io/kops/cloudmock/aws/mockec2"
 	"k8s.io/kops/pkg/resources"
+	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/resources/aws/aws_test.go
```diff
@@ -196,3 +197,71 @@ func TestSharedVolume(t *testing.T) {
...
+
+	for i, test := range tc {
+		got := matchesElbTags(test.tags, test.actual)
+		if got != test.expected {
+			t.Fatalf("unexpected result from testcase %d, expected %v, got %v", i, test.expected, got)
+		}
+	}
+}
+
+func strtopoint(str string) *string {
```
olemarkus, line 259: We have `fi.String` that does the same thing.

hs0210:
@olemarkus I have replaced the func strtopoint with fi.String. PTAL.

olemarkus:
Can you fix te bazel test? probably enough to run `hack/update-bazel.sh`
Looks good otherwise.

pkg/resources/aws/aws_test.go
```diff
@@ -196,3 +198,67 @@ func TestSharedVolume(t *testing.T) {
...
 }
+
+func TestMatchesElbTags(t *testing.T) {
+	tc := []struct {
+		tags     map[string]string
+		actual   []*elb.Tag
+		expected bool
+	}{
+		{
+			tags: strtomap("tagkey1", "tagvalue1"),
```
hakman, line 209: Why use `strtomap` when you could directly create a map?
```suggestion
			tags: map[string]string{"tagkey1": "tagvalue1"}
```

hs0210:
@hakman Thanks.

hs0210:
@olemarkus Sorry, can you explain a little more, I am not very clear what I need to fix.

hakman:
Happy to help @hs0210. Could you also add some tests for multiple tags? The `matchesElbTags` function should handle those also. Thanks!

@olemarkus asked that you run `hack/update-bazel.sh` and commit the changes to fix the 2 failing tests.

hs0210:
@hakman I see these tags are all in string format, don't they?

hs0210:
PTAL.

PR #8998, opened 2020-04-27
Don't force bazel to use python2

olemarkus:


rifelpet:
tested this locally and it worked great. no longer dealing with incompatibilities with pyenv. both GH issues mentioned are closed too.

/lgtm
/approve

PR #8999, opened 2020-04-27
feat(openstack): propagate cloud labels to machines

mitch000001:
This PR enables propagation of cloud labels onto the instances.


mitch000001:
Awesome, thx for the fast responses :D

PR #9004, opened 2020-04-27
GCE: needs to use v1 network resources, not beta

### LINKED ISSUE #9003: Updating to v1 of GCE libs broke network.Find()
The network self-link is [hardcoded](https://github.com/kubernetes/kops/blob/fd512ca1657ce88e3926fd32620e7de10c0fb1ba/upup/pkg/fi/cloudup/gcetasks/network.go#L79) to use `beta` resource name. When creating new clusters, this throws errors like : 

```
W0427 12:55:00.998554   49101 network.go:68] SelfLink did not match URL: "https://www.googleapis.com/compute/v1/projects/ehole-dev-abc/global/networks/default" vs "https://www.googleapis.com/compute/beta/project
s/ehole-dev-abc/global/networks/default"
```

Looks like we have another [place to update as well](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/gcetasks/subnet.go#L196).

geojaz:
closes #9003 

~~I don't think this impacts existing clusters, but am checking...~~
Looks like this is a no-op for existing clusters, we're good to go.

### DIFF
--- a/upup/pkg/fi/cloudup/gcetasks/network.go
+++ b/upup/pkg/fi/cloudup/gcetasks/network.go
@@ -76,7 +76,7 @@ func (e *Network) Find(c *fi.Context) (*Network, error) {
 
 func (e *Network) URL(project string) string {
 	u := gce.GoogleCloudURL{
-		Version: "beta",
+		Version: "v1",
 		Project: project,
 		Name:    *e.Name,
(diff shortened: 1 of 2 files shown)


rifelpet:
awesome!
/lgtm
/approve

PR #9005, opened 2020-04-27
Remove amd64 TravisCI jobs

rifelpet:
Now that we have GitHub actions I don't think we need travis anymore

/hold for input from others

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,7 @@ language: go
 arch:
   - amd64
   - arm64
-os:
-  - linux
-  - osx
+os: linux
 dist: focal
 go: 1.15rc2
 
@@ -12,13 +10,3 @@ go_import_path: k8s.io/kops
 
 script:
   - GOPROXY=https://proxy.golang.org travis_wait 30 make all examples test
-
-jobs:
-  include:
-    - name: Verify
-      arch: amd64
-      os: linux
-      dist: focal
-      go: 1.15rc2
-      script:
-        - GOPROXY=https://proxy.golang.org make travis-ci


hakman:
Will still need this for ARM64 tests, I guess.

rifelpet:
:(

https://github.community/t5/GitHub-Actions/Testing-against-multiple-architectures/m-p/40265

I'm having trouble finding out if Azure has any ARM64 instance types because google takes me to Azure Resource Manager. If they don't then I wouldnt expect support anytime soon. Maybe we run ARM64 tests on travis? that's unfortunate..

hakman:
If we keep a single ARM64 test in travis it should be ok.

johngmyers:
Do the GitHub actions block Prow from merging?

rifelpet:
Apparently not: https://prow.k8s.io/pr?query=is%3Apr%20repo%3Akubernetes%2Fkops%20author%3Arifelpet%20head%3Aghactionstest%20is%3Aopen i've mentioned it in #sig-testing so hopefully we can get that resolved soon.

irons:
Yes, there is a repo setting to require status checks to succeed before allowing a merge. 

Requiring the GA status check now does mean that existing PRs, which don't yet contain the GA workflow YAML file on master, will be prevented from merging until they pull it in. It probably won't be super obvious why the status check remains pending in those cases.

mikesplain:
We can do this with Prow, I'll open a PR and mark it as hold until others approve. 

I think by enabling prow support for this, it will likely remove Travis as blocking (unless we add it as others have).  What do we actually want to block vs not?

rifelpet:
I'm guessing that even though release-1.18's travis setup doesn't have any arm64 jobs, we shouldn't remove the file altogether because the test-infra changes might still expect the travis status contexts to exist which may block any merges to release 1.18. 

So we'll live with running both travis and GHA jobs in release-1.18, which isn't the end of the world.

hakman:
Considering that the GH workflows jobs cannot be made blocking, is this still required?
https://github.com/kubernetes/test-infra/issues/15611

johngmyers:
We need Travis for the arm64 build but we can remove the amd64 ones.

hakman:
@johngmyers And we are OK removing the TravisCI jobs because the blocking part is covered by the Prow jobs?

johngmyers:
The Travis jobs provided two things: One, they block merges, which are covered by Prow except for the arm64. Two, they provide signal before opening a PR, which is covered by the GitHub Actions.

johngmyers:
I believe the amd64 job can be removed as well.

hakman:
I think Prow build only for Bazel builds. This is why seemed a bit safer to keep the most important build still blocking.

PR #9009, opened 2020-04-28
Use a custom kubernetes client, to handle network failures

justinsb:
By setting a low HTTP timeout, and reseting connections on error, we
should be able to work around the http2 connection issues:
https://github.com/kubernetes/client-go/issues/374

I observed this bug specifically when rolling an HA control plane
without a load balancer.


pkg/k8sclient/client.go
```diff
@@ -0,0 +1,176 @@
...
+	// https://github.com/kubernetes/client-go/issues/374 We
+	// trigger a timeout error and then to recover we need to
+	// reset any existing connections
+	if config.Timeout < maxTimeout {
+		config.Timeout = maxTimeout
+	}
+
+	c, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("error creating kubernetes client: %v", err)
```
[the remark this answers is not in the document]
justinsb: Fair - done.  It's mostly habit from when calling AWS SDK when some of the functions return _very_ terse messages, so I tend to wrap-by-default.  Hopefully this all becomes easier when we start using %w - which we can start doing with go 1.13 (https://blog.golang.org/go1.13-errors) .. I'm just a little wary about backports in general.

pkg/k8sclient/client.go
```diff
@@ -0,0 +1,176 @@
...
+
+	newData, err := json.Marshal(node)
+	if err != nil {
+		return err
+	}
+
+	patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, node)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
justinsb: So the problem is that we need to call handleError.  I do agree that there's more logic in here than there should be, so I moved it to be a package-level function (which takes a k8sclient.Interface).

pkg/k8sclient/client.go
```diff
@@ -0,0 +1,176 @@
...
+const maxTimeout = 15 * time.Second
+
+// Interface is a wrapper around kubernetes.Interface, that recovers
+// better from network failures.  This error handling is only
+// available on the helper methods, but for compatability RawClient
+// exposes the underlying kubernetes.Interface
+type Interface interface {
+	// RawClient returns the current kubernetes.Interface; it should not be cached
+	// Using wrapper methods is preferrable because we can do richer retry logic and error handling.
+	RawClient() kubernetes.Interface
```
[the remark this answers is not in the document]
justinsb: My strategy is to gradually remove RawClient() - anything that uses RawClient() won't get our enhanced error handling.  I added a comment that this should be considered Deprecated.

justinsb:
You raise an excellent point on client-go wrapping and the exit strategy: I think our exit strategy is actually using less and less of client-go.  It's hard to keep track, but I _think_ the plan for client-go is to favor the [dynamic](https://godoc.org/k8s.io/client-go/dynamic) interface.

(I also marked RawClient deprecated to make it doubly clear that ideally we get to a point where we don't use it)

pkg/k8sclient/client.go
```diff
@@ -0,0 +1,177 @@
...
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+)
+
+const maxTimeout = 15 * time.Second
+
+// Interface is a wrapper around kubernetes.Interface, that recovers
+// better from network failures.  This error handling is only
+// available on the helper methods, but for compatability RawClient
```
rifelpet, line 41:
```suggestion
// available on the helper methods, but for compatibility RawClient
```

fixes the travis / GHA failure

pkg/k8sclient/fakek8sclient/fake.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/kops/pkg/k8sclient"
+)
+
+type Fake struct {
```
johngmyers, line 29: Why does this type exist? Why not use a real `client` wrapping a `fake.Clientset`?

hakman:
Stepping back a little, my understanding is that Go 1.15 was the blocker for fixing this on kubernetes/client-go side.
Is there any way to make this happen now or maybe at least the Go patch opens some other possibilities?
https://github.com/kubernetes/client-go/issues/374#issuecomment-632457187

pkg/k8sclient/client.go
```diff
@@ -0,0 +1,177 @@
...
+			klog.Warningf("client timed out, but rest client was not of expected type, was %T", restClientInterface)
+			return
+		}
+
+		httpTransport := findHTTPTransport(restClient.Client.Transport)
+		if httpTransport == nil {
+			klog.Warningf("client timed out, but http transport was not of expected type, was %T", restClient.Client.Transport)
+			return
+		}
+		httpTransport.CloseIdleConnections()
```
Abhishekvrshny, line 155: @justinsb How would this make sure that the connection over which the request just timed out is idle and hence would be closed? Is there a possibility that the connection is picked up by another request and is in use while this is getting executed? // @johngmyers

johngmyers: If the connection got picked up by another request, that's fine. When the connection does go idle the last request to complete will close it.

Abhishekvrshny: @johngmyers Sure, this may work in this particular case. Would the same approach work in general for some other high throughput clients suffering from connection staleness? Sorry to hijack this thread, but my issue is similar but related to a different system, and was wondering if the same approach can be applied there.

hakman:
@justinsb I think this should be fixed now by this https://github.com/kubernetes/kubernetes/pull/96778 via https://github.com/kubernetes/kops/pull/10385. I believe it can be closed.

PR #9011, opened 2020-04-28
Implement API load balancer class with NLB and ELB support on AWS

christianjoun:
@johngmyers


upup/pkg/fi/default_methods.go
```diff
@@ -26,6 +26,7 @@ import (
 // DefaultDeltaRunMethod implements the standard change-based run procedure:
 // find the existing item; compare properties; call render with (actual, expected, changes)
 func DefaultDeltaRunMethod(e Task, c *Context) error {
+
```
johngmyers, line 29: Unnecessary change.

pkg/commands/status_discovery.go
```diff
@@ -53,19 +61,51 @@ func (s *CloudDiscoveryStatusStore) GetApiIngressStatus(cluster *kops.Cluster) (
...
+		ELB, NLB := false, false
+
+		lbSpec := cluster.Spec.API.LoadBalancer
+		switch lbSpec.Class {
+		case kops.LoadBalancerClassClassic, "":
+			ELB = true
+		case kops.LoadBalancerClassNetwork:
+			NLB = true
+		default:
+			return nil, fmt.Errorf("Unknown Cluster.Spec.API.LoadBalancer.Class : %v", lbSpec.Class)
```
johngmyers, line 74: I think we should add an api validation in `pkg/apis/kops/validation/validation.go` that ensures the class is one of the supported values. Then this lower-level code doesn't have to worry about unsupported values.

pkg/commands/status_discovery.go
```diff
@@ -53,19 +61,51 @@ func (s *CloudDiscoveryStatusStore) GetApiIngressStatus(cluster *kops.Cluster) (
...
 		}
-		if err != nil {
-			return nil, fmt.Errorf("error looking for AWS ELB: %v", err)
+
+		var lb interface{}
+
+		if ELB {
```
johngmyers, line 79: I suggest simplifying, not having separate `ELB` and `NLB` booleans. Just check `if cluster.Spec.API.LoadBalancer == kops.LoadBalancerClassNetwork { ... } else { ... }`

Or `switch cluster.Spec.API.LoadBalancer.Class`

pkg/commands/status_discovery.go
```diff
@@ -53,19 +61,51 @@ func (s *CloudDiscoveryStatusStore) GetApiIngressStatus(cluster *kops.Cluster) (
...
+			lb, err = awstasks.FindLoadBalancerByNameTag(awsCloud, name)
+			if isNil(lb) {
```
johngmyers, line 81: Why the complicated `isNil()`? Just check `lb == nil`.

pkg/commands/status_discovery.go
```diff
@@ -53,19 +61,51 @@ func (s *CloudDiscoveryStatusStore) GetApiIngressStatus(cluster *kops.Cluster) (
...
+			if err != nil {
+				return nil, fmt.Errorf("error looking for AWS ELB: %v", err)
+			}
+		} else if NLB {
+			lb, err = awstasks.FindNetworkLoadBalancerByNameTag(awsCloud, name)
+			if isNil(lb) {
+				return nil, nil
+			}
+			if err != nil {
+				return nil, fmt.Errorf("error looking for AWS ELB: %v", err)
```
johngmyers, line 93:
```suggestion
				return nil, fmt.Errorf("error looking for AWS NLB: %v", err)
```

pkg/commands/status_discovery.go
```diff
@@ -53,19 +61,51 @@ func (s *CloudDiscoveryStatusStore) GetApiIngressStatus(cluster *kops.Cluster) (
...
-		if lb != nil {
-			lbDnsName := aws.StringValue(lb.DNSName)
+		if !isNil(lb) {
+			var lbDnsName string
+			if ELB {
+				lbDnsName = aws.StringValue(lb.(*elb.LoadBalancerDescription).DNSName)
+			} else if NLB {
+				lbDnsName = aws.StringValue(lb.(*elbv2.LoadBalancer).DNSName)
+			}
+			//lbDnsName := aws.StringValue(lb.DNSName)
```
johngmyers, line 106: Remove this comment

pkg/commands/status_discovery.go
```diff
@@ -53,19 +61,51 @@ func (s *CloudDiscoveryStatusStore) GetApiIngressStatus(cluster *kops.Cluster) (
...
+
 		}
 		var ingresses []kops.ApiIngressStatus
 
-		if lb != nil {
-			lbDnsName := aws.StringValue(lb.DNSName)
+		if !isNil(lb) {
+			var lbDnsName string
+			if ELB {
+				lbDnsName = aws.StringValue(lb.(*elb.LoadBalancerDescription).DNSName)
```
johngmyers, line 102: Probably better to pull the scope of `lb` `StringValue` call in the previous if/else blocks, so lb can be of the appropriate type. Pull `lbDnsName` to before the previous if/else block.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -47,6 +47,40 @@ var _ fi.ModelBuilder = &APILoadBalancerBuilder{}
...
+		}
+
+		cleanup := &awstasks.LoadBalancerCleanup{
+			Name:         fi.String("cleanup.api." + b.ClusterName()),
+			AgNames:      agNames,
+			Lifecycle:    b.Lifecycle, //what does this even do?
+			UseELBForAPI: fi.Bool(ELB),
+			UseNLBForAPI: fi.Bool(NLB),
+			NLBName:      fi.String("api." + b.ClusterName()),
+			ELBName:      fi.String("api." + b.ClusterName()), //TODO: think about changing their names?
```
johngmyers, line 79: I don't see any reason for the name to be different.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -47,6 +47,40 @@ var _ fi.ModelBuilder = &APILoadBalancerBuilder{}
...
+				agNames = append(agNames, b.LinkToAutoscalingGroup(ig).GetName())
+			}
+		}
+
+		cleanup := &awstasks.LoadBalancerCleanup{
+			Name:         fi.String("cleanup.api." + b.ClusterName()),
+			AgNames:      agNames,
+			Lifecycle:    b.Lifecycle, //what does this even do?
+			UseELBForAPI: fi.Bool(ELB),
+			UseNLBForAPI: fi.Bool(NLB),
```
johngmyers, line 77: Should probably pass the class instead of the two bools.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -161,18 +220,34 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
 		switch lbSpec.Type {
 		case kops.LoadBalancerTypeInternal:
 			elb.Scheme = fi.String("internal")
+			nlb.Scheme = fi.String("internal")
 		case kops.LoadBalancerTypePublic:
 			elb.Scheme = nil
+			nlb.Scheme = nil
 		default:
-			return fmt.Errorf("unknown elb Type: %q", lbSpec.Type)
+			return fmt.Errorf("unknown elb/nlb Type: %q", lbSpec.Type)
```
johngmyers, line 235:
```suggestion
			return fmt.Errorf("unknown lb Type: %q", lbSpec.Type)
```

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -161,18 +220,34 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+		}
+
+		if ELB {
```
johngmyers, line 238: Should just test the value of `b.Cluster.Spec.API.LoadBalancer`

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -232,8 +307,44 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+					FromPort:      fi.Int64(3),
+					Protocol:      fi.String("icmp"),
+					SecurityGroup: masterGroup.Task,
+					ToPort:        fi.Int64(4),
+				})
+			}
+		}
+	}
+
+	//TODO: not implemented for NLB, would the security groups go to the master nodes?
```
johngmyers, line 345: Yes, traffic needs to go to the master nodes.

johngmyers:
@christianjoun please go to https://identity.linuxfoundation.org/projects/cncf and click on the "Sign up to contribute to this project as an employee" link.

docs/cluster_spec.md
```diff
@@ -20,7 +20,7 @@ spec:
 ```
 
 
-When configuring a LoadBalancer, you can also choose to have a public ELB or an internal (VPC only) ELB.  The `type`
+When configuring a LoadBalancer, you can also choose to have a public LB or an internal (VPC only) LB.  The `type`
```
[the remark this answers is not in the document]
johngmyers: I would prefer spelling out to keeping "ELB".

upup/pkg/fi/cloudup/awstasks/dnsname.go
```diff
@@ -163,6 +204,14 @@ func (_ *DNSName) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *DNSName) error
...
 	}
 
+	if e.TargetNetworkLoadBalancer != nil {
+		rrs.AliasTarget = &route53.AliasTarget{
+			DNSName:              e.TargetNetworkLoadBalancer.DNSName,
+			EvaluateTargetHealth: aws.Bool(false),
+			HostedZoneId:         e.TargetNetworkLoadBalancer.HostedZoneId,
+		}
+	}
+
```
johngmyers, line 214: Will eventually need corresponding changes to `RenderCloudformation` and `RenderTerraform`. We should at least make sure it refuses to allow NLB with Cloudformation or Terraform rather than generating a non-working cluster.

pkg/model/dns.go
```diff
@@ -91,6 +91,29 @@ func (b *DNSModelBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	//TODO: use interface (but what about bastion?), should I create a new member instead of targetLoadBalancer for the interface?
```
johngmyers, line 94: I definitely like the idea of an interface implemented by both LoadBalancer and NetworkLoadBalancer. It will need to return at least the DNSName and HostedZoneId and would allow simplifying a bunch of logic.

cmd/kops/create_cluster.go
```diff
@@ -361,6 +364,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().StringVar(&options.MasterTenancy, "master-tenancy", options.MasterTenancy, "The tenancy of the master group on AWS. Can either be default or dedicated.")
 	cmd.Flags().StringVar(&options.NodeTenancy, "node-tenancy", options.NodeTenancy, "The tenancy of the node group on AWS. Can be either default or dedicated.")
 
+	cmd.Flags().StringVar(&options.APILoadBalancerClass, "api-loadbalancer-class", options.APILoadBalancerClass, "Currently only supported in AWS. Sets the API loadbalancer class to eiether 'classic' or 'network'")
```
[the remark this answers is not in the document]
timothyclarke: To avoid this flag I'd argue for dropping `classic` and using `network`
NLB's have fixed IP addresses. Classic loadbalancers change IP addresses.
Where I am has restricted the API to known IP addresses.  Our VPN policy is to only send the required traffic through the VPN, rather than all traffic.
Our dev team still need  to interact with the clusters. We've been busy playing wack-a-mole with the classic IP's as they change.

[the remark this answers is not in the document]
johngmyers: As mentioned in #8963 

> It would be lovely to use NLB instead of CLB, but there are some odd limitations on the > AWS side with how many NLBs that one can have per region (soft limit) and how many > NLB ENIs one can have per VPC (harder limit).

pkg/model/spotinstmodel/instance_group.go
```diff
@@ -219,6 +219,10 @@ func (b *InstanceGroupModelBuilder) buildElastigroup(c *fi.ModelBuilderContext,
 		return fmt.Errorf("error building ssh key: %v", err)
 	}
 
+	if true {
+		panic("this code is not ready for NLB")
```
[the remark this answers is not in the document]
timothyclarke: I'm probably showing ignorance and this maybe considered off topic, but what's wrong with NLB's and spots ?
Is it simply spots are considered too volatile. I've been using mixed instance policies with spots for my masters for quite some time and not encountered issues.
I assume for the API cross AZ loadbalancing would also be enabled

rifelpet:
@christianjoun any chance you'll be able to address the review comments and merge conflicts? Thanks!

rifelpet:
@christianjoun ping, just checking in to see if you'll be able to address the feedback and conflicts soon or if we can have someone else move forward with this

christianjoun:
@rifelpet I rebased master this weekend and started testing. I should be able to address the review comments by end of this weekend.

upup/pkg/fi/cloudup/awstasks/dnsname.go
christianjoun: I implemented this by panic via the following methods.
NetworkLoadBalancer.TerraformLink() 
NetworkLoadBalancer.CloudformationAttrCanonicalHostedZoneNameID() NetworkLoadBalancer.CloudformationAttrDNSName()

pkg/model/dns.go
christianjoun: There are other situations such as in spot, where an interface for either Loadbalancer or NetworkLoadBalancer would help. I wonder if DNSTarget could be re-used in all instances, or if it would be more semantically correct to have a LB interface as well separate from DNSTarget.  Methods needed in the spot instance would be getLoadBalancerName getLoadBalancerArn, getLoadBalancerType

pkg/commands/status_discovery.go
christianjoun: I think the right strategy is to search both ELB and NLB's

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -126,6 +165,26 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+		nlb = &awstasks.NetworkLoadBalancer{
+			Name:      fi.String("api." + b.ClusterName()),
+			Lifecycle: b.Lifecycle,
+
+			LoadBalancerName: fi.String(loadBalancerName),
+			Subnets:          lbSubnets,
+			Listeners:        nlbListeners,
+
+			// Configure fast-recovery health-checks
+			HealthCheck: &awstasks.NetworkLoadBalancerHealthCheck{
```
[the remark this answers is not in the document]
christianjoun: the targetGroup has a setting for health check interval however it states

	// The approximate amount of time, in seconds, between health checks of an individual
	// target. For Application Load Balancers, the range is 5 to 300 seconds. For
	// Network Load Balancers, the supported values are 10 or 30 seconds.
	//
	// With Network Load Balancers, you can't modify this setting.
	HealthCheckIntervalSeconds *int64 `min:"5" type:"integer"`

I do see the value in exposing for the ALB case....

christianjoun:
issues I am facing:

#1. the healthchecks are failing because nlb does not support https healthchecks. This is from the aws documentation.

  // The protocol the load balancer uses when performing health checks on targets.
  // The TCP protocol is supported for health checks only if the protocol of the
  // target group is TCP, TLS, UDP, or TCP_UDP. The TLS, UDP, and TCP_UDP protocols
  // are not supported for health checks.
  //
  // With Network Load Balancers, you can't modify this setting.
  HealthCheckProtocol *string `type:"string" enum:"ProtocolEnum"`

#2. i noticed that the nlb gets a ***.elb*... address. Is that ok?

#3. I need to adjust the security group on the masters to allow tcp 443 from the VPC cidr.

#4. Does anyone know if the nlb’s nodes ip’s could change?  or are they persistent?

upup/pkg/fi/cloudup/awstasks/load_balancer_cleanup.go
```diff
@@ -0,0 +1,461 @@
...
+package awstasks
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"k8s.io/klog"
```
rifelpet, line 56: We recently migrated to klog v2, it looks like this PR is adding back klog v1. This same change should be made in the other new files in this package. (run `make gomod` after updating these imports)

```suggestion
	"k8s.io/klog/v2"
```

upup/pkg/fi/cloudup/awstasks/load_balancer_cleanup.go
```diff
@@ -0,0 +1,461 @@
...
+			response, err := cloud.ELBV2().DescribeTargetGroups(request)
+			if err != nil {
+				return nil, fmt.Errorf("error querying for NLB Target groups :%v", err)
+			}
+
+			if len(response.TargetGroups) == 0 {
+				return nil, fmt.Errorf("Found no Target Groups for NLB don't think this is a normal condition :  %v", lb.LoadBalancerArn)
+			}
+
+			if len(response.TargetGroups) != 1 {
```
rifelpet, line 354: Theres a good chance that we'll use multiple target groups in the fix for https://github.com/kubernetes/kops/issues/9756 so we should handle deleting all of them.

upup/pkg/fi/cloudup/awstasks/load_balancer_cleanup.go
```diff
@@ -0,0 +1,461 @@
...
+	if lb != nil {
+
+		request := &elbv2.DeleteLoadBalancerInput{
+			LoadBalancerArn: lb.LoadBalancerArn,
+		}
+
+		removals = append(removals, &deleteNetworkLoadBalancer{request: request})
+		klog.V(2).Infof("will delete network load balancer: %v", lb.LoadBalancerName)
+
+		//TODO: Could be useful to check the autoscaling group for associated target groups and delete them as well.
```
rifelpet, line 339: the InstanceGroupSpec supports providing additional target group ARNs to be attached to the ASG, so we would need to make sure we're not deleting the attachment of one of those. I don't think that's necessary though, so we can probably remove this TODO

upup/pkg/fi/cloudup/awstasks/load_balancer_cleanup.go
```diff
@@ -0,0 +1,461 @@
...
+				removals = append(removals, &deleteTargetGroup{request: request})
+			}
+		}
+
+	}
+
+	for _, autoScalingGroupName := range e.AgNames {
+
+		request := &autoscaling.DescribeLoadBalancerTargetGroupsInput{
+			AutoScalingGroupName: autoScalingGroupName,
```
rifelpet, line 377: Ah, as I mentioned above, InstanceGroupSpec has an [ExternalLoadBalancers](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#InstanceGroupSpec) field that allows other target groups to be attached to an ASG that we shouldn't delete when deleting an NLB.

I think deleting all TGs associated with the NLB we are deleting would be sufficient, which is covered above (L340-372) so we may not need this for-loop or AgNames at all.

upup/pkg/fi/cloudup/awstasks/load_balancer_cleanup.go
```diff
@@ -0,0 +1,461 @@
...
+				return nil, fmt.Errorf("Found multiple Target groups for NLB with arn %v", lb.LoadBalancerArn)
+			}
+
+			targetGroupArn := response.TargetGroups[0].TargetGroupArn
+
+			{
+
+				klog.V(2).Infof("Deleting Target Group with arn : %v\n", targetGroupArn)
+
+				request := &elbv2.DeleteTargetGroupInput{
```
rifelpet, line 364: We should delete the target groups before deleting the NLB, that way if Kops is interrupted mid-deletion, it can successfully resume and delete the remaining resources. If the NLB is gone and we try to resume, we wont be able to find these target groups based on a `LoadBalancerArn`

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -0,0 +1,855 @@
...
+				request.SetSubnets(aws.StringSlice(append(actualSubnets, newSubnetIDs...)))
+
+				klog.V(2).Infof("Attaching Load Balancer to new subnets")
+				if _, err := t.Cloud.ELBV2().SetSubnets(request); err != nil {
+					return fmt.Errorf("Error attaching Load Balancer to new subnets: %v", err)
+				}
+			}
+		}
+
+		//TODO: decide if security groups should be applied to master nodes
```
rifelpet, line 745: We can remove all SG references in the NLB tasks. Rather than having the NLB tasks apply SGs to the masters the models will setup the master ASGs to update their SGs if we're using an NLB (which it looks like you already did in pkg/model/awsmodel/api_loadbalancer.go )

upup/pkg/fi/cloudup/awstasks/network_load_balancer_attachment.go
```diff
@@ -0,0 +1,160 @@
...
+	LoadBalancer *NetworkLoadBalancer
+
+	// LoadBalancerAttachments now support ASGs or direct instances
+	AutoscalingGroup *AutoscalingGroup
+	Subnet           *Subnet
+
+	// Here be dragons..
+	// This will *NOT* unmarshal.. for some reason this pointer is initiated as nil
+	// instead of a pointer to Instance with nil members..
+	Instance *Instance
```
rifelpet, line 44: I can't think of any reason why we'd want to be attaching individual instances to an NLB so lets remove this field and any references to it. It would be nice to figure out why it existed in the ELB tasks too, but we can do that later.

upup/pkg/fi/cloudup/awstasks/load_balancer_cleanup.go
christianjoun: should I also omit (L289-314)?

upup/pkg/fi/cloudup/awstasks/load_balancer_cleanup.go
christianjoun: if we'll use multiple target groups, network_loadbalancer.go L393 will need adjustment.

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
christianjoun: I noticed this was being done @ load_balancer.go L581-592 and I am trying to sort out why thats needed there / what edge case that is for...and if thats valid, does that also apply in the NLB case as well..

pkg/apis/kops/cluster.go
```diff
@@ -358,8 +358,27 @@ const (
...
 )
 
+// LoadBalancerClass string describes LoadBalancer classes (classic, network)
+type LoadBalancerClass string
+
+const (
+	LoadBalancerClassClassic     LoadBalancerClass = "Classic"
+	LoadBalancerClassNetwork     LoadBalancerClass = "Network"
+	LoadBalancerClassBothClassic LoadBalancerClass = "Both,Classic"
+	LoadBalancerClassBothNetwork LoadBalancerClass = "Both,Network"
```
rifelpet, line 368: I was thinking the allowed values would be
```
Classic
Classic,Network
Network
Network,Classic
```
if multiple values are specified, the first value is what the DNS will point to

johngmyers:
@christianjoun the recommendation from Kops Office Hours is to reduce the scope of this PR. It need only address the case of new clusters. Transition can be disruptive or leave orphaned resources. The transition issues can be addressed in subsequent PRs. We wish to unblock other work that is behind this.

docs/cluster_spec.md
```diff
@@ -11,7 +11,7 @@ On this page, we will expand on the more important configuration keys.
 This object configures how we expose the API:
 
 * `dns` will allow direct access to master instances, and configure DNS to point directly to the master nodes.
-* `loadBalancer` will configure a load balancer (ELB) in front of the master nodes, and configure DNS to point to the ELB.
+* `loadBalancer` will configure a load balancer (ELB or NLB) in front of the master nodes, and configure DNS to point to the LB.
```
hakman, line 14:
```suggestion
* `loadBalancer` will configure a load balancer (CLB or NLB) in front of the master nodes, and configure DNS to point to the LB.
```

rifelpet:
With @christianjoun's permission I'm rebasing this PR on top of #9794 to take advantage of the target group awstask created there. I also addressed some issues and feedback from others.

rifelpet:
I just pushed a few more commits that help separate the target group attributes from the NLB attributes. I removed the NLB and target group attributes that kops doesn't allow users to specify and therefore shouldn't need to care about. We can add them back later if we expose them through the API.

I also added terraform and cloudformation support for NLBs, listeners, and target groups so we can get a better idea of whats changing when we migrate a cluster from ELB to NLB - just look at the `tests/integration/update_cluster/complex/kubernetes.tf` diff. 

I'm noticing security group rule changes being reported that could result in temporary downtime during the migration. I know we acknowledged API downtime due to load balancer deletion and dns propagation delays and that we'll address it in a followup PR, but security group changes will also need similar attention.

rifelpet:
Specifically, we have two options for how we change the security group setup when switching between ELB and NLB:

* update the security group associated with each security group rule - any rule for the ELB SG will now be added to the master instance SG. This results in the ELB SG having its rules removed making the ELB inaccessible once "kops update cluster" runs.

* update the security groups associated with the master instances to include the ELB security group. This doesn't revoke any rules allowing the ELB to remain accessible but means that the masters won't be accessible through the NLB until after a rolling update has been performed since the SGs are a property of the launch template.

Any other ideas for how we could manage these?

We discussed for the load balancers and dns having an api field option that ensures both LBs exist and that api dns points to a specific LB. In those states we could have the SG rules exist on both the ELB SG and master instance SG. That might be the best approach.

hakman (changes requested):
I added some suggestions regarding name consistency and something that blocks export of kubecfg.

There is another small issue about adding `cloudLabels` which results in and should be fixed:
```
W1031 15:54:16.112405   88154 executor.go:131] error running task
"NetworkLoadBalancer/api-k8s-deps-test-bla-com" (9m46s remaining to succeed):
unexpected error fetching tags for resource: error listing tags on :
ValidationError: An ARN must be specified
```

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -0,0 +1,797 @@
...
+		loadBalancerName = fi.StringValue(a.LoadBalancerName)
+
+		lb, err := findNetworkLoadBalancerByLoadBalancerName(t.Cloud, loadBalancerName)
+		if err != nil {
+			return fmt.Errorf("error getting load balancer by name: %v", err)
+		}
+
+		// if lb == nil {
+		// 	return fmt.Errorf("error querying nlb: %v", err)
+		// }
```
hakman, line 552: This should address the of adding cloudLabels which errors with:
```
W1031 15:54:16.112405   88154 executor.go:131] error running task
"NetworkLoadBalancer/api-k8s-deps-test-bla-com" (9m46s remaining to succeed):
unexpected error fetching tags for resource: error listing tags on :
ValidationError: An ARN must be specified
```
```suggestion
		loadBalancerArn = fi.StringValue(lb.LoadBalancerArn)
```

hakman:
This looks ready to go. Should be a very nice addition for 1.19.
Great work both of you @christianjoun and @rifelpet!
/lgtm

rifelpet:
Agreed, thanks @christianjoun! I think this is ready to merge as is but additional reviews from other reviews would certainly be welcome. To outline the remaining work, we have the following TODO items planned for followup PRs:

### HA migration between CLB <-> NLB
- [ ] Update the API to support creating both CLB and NLB with DNS pointing to one of them. Originally we discussed a comma separated string with the first value having DNS point to it, ie `class: Network,Classic`. Update pkg/model/awsmodel to support creating both.

- [ ] Refactor the SG rule logic in pkg/model/awsmodel. We should be able to more easily define the security group rules and choose whether to add the rules to the API LB security group or the masters security group or both

- [ ] Update the NLB SG rule applied to the master SG that allows healthcheck traffic from the VPC CIDR to support VPCs with multiple CIDRs. We might be able to revert the SGR changes in this PR and have pkg/model find the VPC CIDR(s) and pass them to the SGR tasks. 

- [ ] Update `upup/pkg/fi/cloudup/awstasks/dnsname.go` to choose whichever load balancer is first in the comma separated string


### ACM Certificate on secondary NLB listener (#9756)
- [ ] Update NLB awstask to support multiple listeners and target groups
- [ ] Update the NLB awstask updating of listeners to use elbv2.ModifyListener rather than DeleteListener + CreateListener
- [ ] Reuse most of the non-loadbalancer related work from #9761


Any other followup work that should block 1.19.0 ?

pkg/apis/kops/validation/validation.go
```diff
@@ -211,6 +211,14 @@ func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *fie
 		allErrs = append(allErrs, validateRollingUpdate(spec.RollingUpdate, fieldPath.Child("rollingUpdate"), false)...)
 	}
 
+	if spec.API != nil && spec.API.LoadBalancer != nil && spec.CloudProvider == "aws" {
+		value := string(spec.API.LoadBalancer.Class)
+		allErrs = append(allErrs, IsValidValue(fieldPath.Child("class"), &value, kops.SupportedLoadBalancerClasses)...)
+		if featureflag.Spotinst.Enabled() && spec.API.LoadBalancer.Class == kops.LoadBalancerClassNetwork {
+			allErrs = append(allErrs, field.Forbidden(fieldPath, "spotinst + NLBs dissallowed"))
```
olemarkus, line 218:
```suggestion
			allErrs = append(allErrs, field.Forbidden(fieldPath, "Cannot use NLB together with spotinst"))
```

olemarkus (approved):
Looks good to me

hakman (approved):

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -200,6 +200,21 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchConfigurationTask(c *fi.ModelB
 		SecurityGroups:                []*awstasks.SecurityGroup{sgLink},
 	}
 
+	if b.APILoadBalancerClass() == kops.LoadBalancerClassNetwork {
```
seh, line 203: See #10517.

PR #9015, opened 2020-04-28
try github actions failure to see if PRs can't be merged 

rifelpet:
Ref: #9005

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,7 @@ jobs:
       - name: make nodeup examples test
         working-directory: ${{ env.GOPATH }}/src/k8s.io/kops
         run: |
-          make nodeup examples test
+          foo
 
   verify:
     strategy:
@@ -64,4 +64,4 @@ jobs:
       - name: make travis-ci
         working-directory: ${{ env.GOPATH }}/src/k8s.io/kops
         run: |
-          make travis-ci
+          foo
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,9 @@
 language: go
 arch:
-  - amd64
   - arm64
 os:
   - linux
-  - osx
 go:
-  - 1.13.x
   - 1.14.x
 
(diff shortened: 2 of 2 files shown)


rifelpet:
@mikesplain would prow merge this if it had a lgtm label? looking at the tide details, it seems like it would but i'd expect it not to.

rifelpet:
🥁 
the [tide page](https://prow.k8s.io/pr?query=is%3Apr+repo%3Akubernetes%2Fkops+author%3Arifelpet+head%3Aghactionstest) isn't mentioning the GHA failures but i guess we'll see if it is able to merge it or not. I'm guessing not given that the GHA jobs now have the `Required` label on them.

If it can't merge it, then we can close this and proceed with https://github.com/kubernetes/kops/pull/9005

rifelpet:
:(

mikesplain:
Right... I think because prow doesn't know about them... which is why https://github.com/kubernetes/test-infra/pull/17917/files#diff-00deeeaf0a7a2aba7d06744db0da6d43R535-R536 is needed.

hakman:
Do we want also to skip unknown contexts?
https://github.com/kubernetes/test-infra/blob/c68834a35762399bac39eff4bbcef65cb5b9fd12/config/prow/config.yaml#L529-L530

rifelpet:
that seems strange because the GHA jobs are now showing as "required" in the list of statuses:

Screen Shot 2020-06-17 at 11 24 52 PM

 so i would think that GH wouldn't let prow merge the PR even if it wanted to... ah well, we can resume this when you get a chance to test more on your internal prow

mikesplain:
Unknown contexts didn't work as expected when I tested it.

Tide doesn't do a github merge afaik.  It does it all via git commands essentially.  But tide blocks on jobs it knows about and, when configured, protected branch rules.

I'll revisit that PR later in the week to see if I can figure it out.  Next time I wish we could get some logs before immediately reverting.

PR #9017, opened 2020-04-28
Allow changing AZ of masters

### LINKED ISSUE #8817: unable to change zones of control plane
 this information.**

1.13

 a `kops` flag.**

1.13

aws

Create cluster:

1. kops create cluster --name ABC --dry-run with 5 master nodes in us-east-1 (zones a,b,c,d,e), and save output to kops_cluster.yaml
2. kops create -f kops_cluster.yaml
3. kops replace -f kops_cluster.yaml
4. kops update cluster ABC  --out=./ --target=terraform
5. terraform apply

Change cluster zones: 

1. edit for 3 zones (a,b,c) in kops_cluster.yaml (rename IG d and e to a-2 and b-2)
2. kops replace -f kops_cluster.yaml --force
3. kops update cluster ABC  --out=./ --target=terraform
4. terraform apply

The terraform apply gets stuck waiting for the etcd volumes to get destroyed. 
I tried manually terminating the EC2 instances using these volumes. 
This allowe a second try of change steps 2-4 to succeed but kops validate no longer validates so not possible to do a kops rolling-update. 

I expected the terraform apply to complete successfully and then the rolling-update to be possible and successful. 

olemarkus:
Fixes #8817 

I have not done any testing on terraform though.
Care still needs to be taken when deleting master IGs. If you break quorum you need to restore etcd from backup. Preventing master IGs in an unsafe way is already possible though, so I didn't do anything to prevent this here.

### DIFF
--- a/pkg/apis/kops/validation/instancegroup.go
+++ b/pkg/apis/kops/validation/instancegroup.go
@@ -189,8 +189,12 @@ func validateVolumeMountSpec(path *field.Path, spec *kops.VolumeMountSpec) field
 
 // CrossValidateInstanceGroup performs validation of the instance group, including that it is consistent with the Cluster
 // It calls ValidateInstanceGroup, so all that validation is included.
-func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, strict bool) field.ErrorList {
+func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster) field.ErrorList {
 	allErrs := ValidateInstanceGroup(g)
+
+	if g.Spec.Role == kops.InstanceGroupRoleMaster {
+		allErrs = append(allErrs, ValidateMasterInstanceGroup(g, cluster)...)
+	}
 
 	// Check that instance groups are defined in subnets that are defined in the cluster
 	{
@@ -207,6 +211,23 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, st
 		}
 	}
 
+	return allErrs
+}
+
+func ValidateMasterInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for _, etcd := range cluster.Spec.EtcdClusters {
+		hasEtcd := false
+		for _, m := range etcd.Members {
+			if fi.StringValue(m.InstanceGroup) == g.ObjectMeta.Name {
+				hasEtcd = true
+				break
+			}
+		}
+		if !hasEtcd {
+			allErrs = append(allErrs, field.NotFound(field.NewPath("spec", "etcdClusters").Key(etcd.Name), g.ObjectMeta.Name))
+		}
+	}
 	return allErrs
 }
 
--- a/pkg/apis/kops/validation/instancegroup.go
+++ b/pkg/apis/kops/validation/instancegroup.go
@@ -17,6 +17,7 @@ limitations under the License.
 package validation
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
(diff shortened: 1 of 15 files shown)


pkg/apis/kops/validation/instancegroup.go
```diff
@@ -210,6 +214,23 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, st
...
+	for _, etcd := range cluster.Spec.EtcdClusters {
+		hasEtcd := false
+		for _, m := range etcd.Members {
+			if fi.StringValue(m.InstanceGroup) == g.ObjectMeta.Name {
+				hasEtcd = true
+				break
+			}
+		}
+		if !hasEtcd {
+			allErrs = append(allErrs, field.NotFound(field.NewPath("spec", "etcdClusters").Key(etcd.Name), g.ObjectMeta.Name))
```
johngmyers, line 228:
```suggestion
			allErrs = append(allErrs, field.NotFound(field.NewPath("spec", "etcdClusters").Key(k).Key("name"), etcd.Name, "has no member for master instancegroup " + g.ObjectMeta.Name))
```

where k is the index from the range over EtcdClusters

johngmyers: Actually, the NotFound error is the other way around. It is the instancegroup's ObjectMeta.Name that wasn't found, so it should be `field.NewPath("objectMeta", "name")`

johngmyers: Actually, the NotFound error is the other way around. It is the instancegroup's Name that isn't found in the etcdcluster's members. So it should be `field.NewPath("objectMeta", "name")`

olemarkus: Thanks for spotting this.

`NotFound` unfortunately doesn't take a third description argument, so I was looking hard at how to make this more clear.

If I understand the `NotFound` description correctly, the error's field should point to the missing value, not the "cause" of the missing value (which may not even be a path). Since the validation happens as a part of a lookup in the etcdSpec, I found this direction to be correct.

Question is how to make this clear to users. The most programatically correct solution would be `field.NotFound(field.NewPath("spec", "etcdClusters").Index(k).Key("instanceGroup")` but not sure that is the most understandable for users as it won't contain the etcd cluster name.

olemarkus: I changed the path to be more correct + added a deep validation test to see how this will look like to the end user.

johngmyers: The value that's missing is the instancegroup name.

I think we need to use `field.Forbidden` here. I believe `field.NotFound` is for when there is an obvious list of possible choices for the field's value. `field.NotFound` on a path inside spec.etcdClusters would be appropriate for a validation that checks if an etcdCluster Member's InstanceGroup field value was actually a defined InstanceGroup (possibly restricting to instancegroups with role master).

johngmyers: Would it not be reasonable for a five-master-IG cluster to have a Cilium etcd cluster with only three members? Perhaps not worth supporting that edge case, but maybe this restriction should only apply to the two etcd clusters used by apiserver?

olemarkus: I think `Forbidden` is worth it just for the details string. I'll switch to that.

Having a different size for the cilium cluster could perhaps be interesting, but since we currently use the internal master DNS entry for this cluster, it won't work (or at least cilium will try to connect to members that don't exist). If someone really wants this, I can look into it later.

olemarkus: @johngmyers Which field would you say should be the target here?
There is a similar case when creating an IG with a subnet that doesn't exist. Here we use `NotFound`, but the path is much simpler.
With `Forbidden`, it makes more sense to point to `spec` itself, but not sure that makes that much sense. What do you think?

johngmyers: The field is `field.NewPath("objectMeta", "name")`, of the instancegroup. The detail would be something like `instanceGroup "eu-central-1a" must have a member in etcdCluster "main"`.

PR #9018, opened 2020-04-28
Spotinst: Support for Grace Period in Ocean Cluster

liranp:
This PR adds support for Grace Period (a period of time, in seconds, that Ocean should wait before applying instance health checks) in Ocean Cluster.


liranp:
Thank you guys for taking care of this so quickly, much appreciated!

PR #9019, opened 2020-04-28
Update the cross-account example with working policy

olemarkus:


docs/state.md
```diff
@@ -103,11 +103,10 @@ The state store can easily be moved to a different s3 bucket. The steps for a si
...
 Repeat for each cluster needing to be moved.
 
-#### Cross Account State-store (AWS S3)
+#### Cross Account State-store
 
-There are situations in which the entity executing kops to create the cluster is not in the same account as the owner of the state store bucket. In this case, you must explicitly grant the permission: `s3:getBucketLocation` to the ARN that is running kops.
-
-You can use the following policy to guide your implementation:
+Many enterprises prefer to run many AWS accounts. In these setups, having a shared cross-account S3 bucket for state may make inventory and management easier.
+In order to achieve this, you first need to let the other accounts access the s3 bucket. This is done by adding the following _bucket policy_ on the S3 bucket:
```
rifelpet, line 109: I find it easiest to describe multi-account scenarios by referring to them as Account A and Account B. Maybe we could do something similar here?

```suggestion
Consider the S3 bucket living in Account B and the kops cluster living in Account A. In order to achieve this, you first need to let Account A access the s3 bucket. This is done by adding the following _bucket policy_ on the S3 bucket:
```

and then update policy document to reference `arn:aws:iam::account-a:root` or something. and mention them again in the paragraph below.

Thoughts?

olemarkus: Yeah, it is better. That makes it more consistent with the AWS docs too, I think. Thanks.

docs/state.md
```diff
@@ -117,20 +116,26 @@ You can use the following policy to guide your implementation:
...
             "Effect": "Allow",
-            "Resource": "arn:aws:s3:::state-store-bucket",
+            "Resource": [
+                "arn:aws:s3:::<state-store-bucket>",
+                "arn:aws:s3:::<state-store-bucket>/*",
+            }
             "Principal": {
                 "AWS": [
-                    "arn:aws:iam::123456789:user/kopsuser"
+                    "arn:aws:iam::<remote-account>:root"
```
johngmyers, line 128:
```suggestion
                    "arn:aws:iam::<account-a>:root"
```

docs/state.md
```diff
@@ -117,20 +116,26 @@ You can use the following policy to guide your implementation:
...
                 ]
             }
         }
     ]
 }
 ```
 
+Kops will then use that bucket as if it was in the remote account, including creating appropriate IAM policies that limits nodes from doing bad things.
+Note that any user/role with full S3 access will be able to delete any cluster from the state store, but may not delete any instances or other things outside of S3.
```
zetaab, line 137: We are using policy like this

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "AllowRootAndHomeListingOfCompanyBucket",
            "Effect": "Allow",
            "Action": "s3:ListBucket",
            "Resource": "arn:aws:s3:::<state bucket>",
            "Condition": {
                "StringEquals": {
                    "s3:delimiter": "/",
                    "s3:prefix": ""
                }
            }
        },
        {
            "Sid": "AllowListingOfUserFolder",
            "Effect": "Allow",
            "Action": "s3:ListBucket",
            "Resource": "arn:aws:s3:::<state bucket>",
            "Condition": {
                "StringLike": {
                    "s3:prefix": "${aws:username}/*"
                }
            }
        },
        {
            "Sid": "AllowAllS3ActionsInUserFolder",
            "Effect": "Allow",
            "Action": "s3:*",
            "Resource": "arn:aws:s3:::<state bucket>/${aws:username}/*"
        }
    ]
}
```

So what we do:
1. create new IAM group and attach that policy there (this needs to be done only once)
2. When we need new cluster:
   * create new IAM user which matches to clustername, example: `mycluster.k8s.local`
   * attach IAM group to user
   * use IAM user credentials to authenticate against S3 bucket (env vars `S3_ACCESS_KEY_ID` and `S3_SECRET_ACCESS_KEY` needs to be defined)

After that each cluster has access to only own state store folder. It makes it possible to also have "root" account which has access to all clusters if needed for ops purposes

olemarkus: This is in addition to the content in the PR, right? So the group is done once per account, and the IAM user is once per cluser? You share this user among everyone that can operate that particular cluster?

Issue #8954, opened 2020-04-22
Kernel throttling bug patch in kops node images


alok87:
**What?**

Linux Kernel has a bug in which un-necessary CPU throttling happens due to the bug in Kernel.
KernelBug: https://bugzilla.kernel.org/show_bug.cgi?id=198197
KubernetesIssue: https://github.com/kubernetes/kubernetes/issues/67577

Many of us use kops default node images and have been experiencing problems due to un-necessary CPU throttling in our pods.

Please suggest if the latest [kops node images](https://github.com/kubernetes/kops/blob/master/channels/stable) has the kernel patch which has the fix for CPUThrottling. If not then can we bring this patch to these node images please.

Node Images
```
kope.io/k8s-1.17-debian-stretch-amd64-hvm-ebs-2020-01-17
kope.io/k8s-1.16-debian-stretch-amd64-hvm-ebs-2020-01-17
kope.io/k8s-1.15-debian-stretch-amd64-hvm-ebs-2020-01-17
```

cc @nuru @justinsb

alok87:
Issue is there in kops image `kope.io/k8s-1.15-debian-stretch-amd64-hvm-ebs-2019-09-26`. with Kernel `4.9.0`. Un-necessary throttling is happening. We need to path the kernel fix in node images.

```
admin@ip-10-2-21-63:~$ uname -a
Linux ip-10-2-21-63 4.9.0-11-amd64 #1 SMP Debian 4.9.189-3+deb9u1 (2019-09-20) x86_64 GNU/Linux
```

[This](https://gist.github.com/bobrik/2030ff040fad360327a5fab7a09c4ff1) was used to do the analysis.

Note: all are with 1000 iterations
We set cfs quota to 20ms and cfs period to 100ms, so we can use at most 20% of cpu during any 100ms period. If we ever go over 20ms of cpu time, we'll be throttled for the remaining of 100ms period.

| CPU Quota | CPU Period        |  Sleep between interationns |   Un-needed Throttling
-------------|-----------------|--------|-----------
|throttling disabled | throttling disabled    |       1000ms     |none (burn_took=5ms always)
| 20000 (20ms)           | 100000  (100ms)           |    100ms         |  sometimes (burn_took=5ms mostly, sometimes 90+ms)
| 20000  (20ms)         | 100000   (100ms)          |    1000ms         | frequent (burn_took=5ms always, and **many times** 80+ms due to throttling) 

[...]

PR #9029, opened 2020-04-30
Update docs on supported distros and images

hakman:
This was a bit out of date and may be easier to just point people to it than explaining the issues every time.

### DIFF
--- a/docs/operations/images.md
+++ b/docs/operations/images.md
@@ -5,8 +5,8 @@ The default images for Kops clusters are based on Debian 9 (Stretch). These imag
 You can choose a different image for an instance group by editing it with `kops edit ig nodes`. You should see an `image` field in one of the following formats:
 
 * `ami-abcdef` - specifies an AMI by id directly
-* `<owner>/<name>` specifies an AMI by its owner and name properties
-* `<alias>/<name>` specifies an AMI by its owner alias and name properties
+* `<owner>/<name>` specifies an AMI by its owner's account ID  and name properties
+* `<alias>/<name>` specifies an AMI by its owner's alias and name properties
 
 Using the AMI id is precise, but ids vary by region. It is often more convenient to use the `<owner/alias>/<name>`, if equivalent images have been copied to various regions with the same name.
 
@@ -30,13 +30,13 @@ image: ami-00579fbb15b954340
 or
 
 ```yaml
-image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64-server-20200423
+image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200423
 ```
 
 or
 
 ```yaml
-image: ubuntu/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64-server-20200423
+image: ubuntu/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200423
 ```
 
 You can find the name for an image using:
@@ -144,6 +144,8 @@ It is based on Kernel version **3.10** which has a considerable number of known 
 
 * https://github.com/kubernetes/kubernetes/issues/56903
 * https://github.com/kubernetes/kubernetes/issues/67577
+
+Before using CentOS images you must accept the agreement at https://aws.amazon.com/marketplace/pp?sku=aw0evgkw8e5c1q413zgy5pjce.
 
 Available images can be listed using:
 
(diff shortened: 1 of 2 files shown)


docs/operations/images.md
```diff
@@ -1,167 +1,235 @@
...
 
-Changing the image for an instance group
+The default images for Kops clusters are based on Debian 9 (Stretch). These images are built by the [kubernetes-sigs/image-builder](https://github.com/kubernetes-sigs/image-builder/tree/master/images/kube-deploy/imagebuilder) project for each Kubernetes release and are available for any other use under the `kope.io` account.
 
-You can choose a different AMI for an instance group.
+You can choose a different image for an instance group by editing it with `kops edit ig nodes`. You should see an `image` field in one of the following formats:
 
-If you `kops edit ig nodes`, you should see an `image` member of the spec.
+* `ami-abcdef` - specifies an AMI by id directly
+* `<owner>/<name>` specifies an AMI by its owner and name properties
```
rifelpet, line 8:
```suggestion
* `<account ID>/<name>` specifies an AMI by its owner's account ID and name properties
```

docs/operations/images.md
```diff
@@ -1,167 +1,235 @@
...
+```yaml
+image: ami-00579fbb15b954340
+```
 
-For example, to use Ubuntu 16.04, you could specify:
+or
 
-`image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20180405`
+```yaml
+image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64-server-20200423
```
rifelpet, line 33: maybe we use amd64 examples since they're more popular than arm at the moment? same with below :)

docs/operations/images.md
```diff
@@ -1,167 +1,235 @@
...
-## CentOS
+## Ubuntu 18.04 (Bionic)
 
-CentOS7 support is still experimental, but should work. Please report any issues.
+Ubuntu 18.04 is fully supported and functional.
 
-The following steps are known:
+It is based on Kernel version **4.15** which has number of known bugs that affect it and may be noticed with larger clusters:
 
-* You must accept the agreement at http://aws.amazon.com/marketplace/pp?sku=aw0evgkw8e5c1q413zgy5pjce
```
rifelpet, line 76: should we keep this mentioned somewhere?

docs/operations/images.md
hakman: The `help` command for `aws-cli` uses the "owner" word for the flag. I think a good compromise would be adding "account ID" in the description.
```
       --owners (list)
          Filters the images by the owner. Specify an  AWS  account  ID,  self
          (owner  is  the sender of the request), or an AWS owner alias (valid
          values are amazon | aws-marketplace |  microsoft  ).  Omitting  this
          option  returns  all  images  for which you have launch permissions,
          regardless of ownership.
```

docs/operations/images.md
hakman: You are right. Added back in CentOS 7 section.

docs/operations/images.md
hakman: I may have tunnel vision from all the ARM stuff these last few weeks 😄

docs/operations/images.md
rifelpet: ah, good call. that sounds good to me 👍

docs/operations/images.md
```diff
@@ -1,167 +1,237 @@
...
+  --query "sort_by(Images, &CreationDate)[*].[CreationDate,Name,ImageId]" \
+  --filters "Name=name,Values=ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-*"
+```
 
-The following steps are known:
+## Ubuntu 20.04 (Focal)
 
-* Redhat AMIs can be found using `aws ec2 describe-images --region=us-east-1 --owner=309956199498 --filters Name=virtualization-type,Values=hvm 'Name=name,Values=RHEL-*GA*' Name=architecture,Values=x86_64 --query 'sort_by(Images,&Name)'`
-* You can specify the name using the `redhat.com` owner alias, for example `redhat.com/RHEL-7.2_HVM-20161025-x86_64-1-Hourly2-GP2`
+Ubuntu 20.04 support was introduced in **Kops 1.16** and is fully functional.
```
johngmyers, line 126: @hakman In #8954 you say Ubuntu 20.04 support was added Kops 1.16.2, but this says it was added in 1.16.0. Which is it?

hakman: Was cherrypicker a bit lately. Probably the release notes has to be updated for 1..16.2. Thanks!

PR #9039, opened 2020-05-01
Update k8s.gcr.io/pause image to version 3.2

hakman:
It's been a long time since `pause` was last updated. Multi-arch support will require latest version for sure, but the signal handler to clean up orphaned zombie processes added in 3.1 will also be useful.
https://github.com/kubernetes/kubernetes/blob/master/build/pause/CHANGELOG.md


MoShitrit:
@hakman we should probably update that to point to the new url, since `k8s.gcr.io` is going away soon. https://groups.google.com/d/msg/kubernetes-dev/MkXnkTSJ_vs/7wvClbkxAgAJ

I figured it's worth noting, since you're updating that now anyway.

hakman:
@MoShitrit I think they are still figuring out what to do. I think this will be a change mostly for those that push images there, not for us that pull it. In any case, best to change all at same time when will be the case.

MoShitrit:
Makes sense. Sorry for raising a false flag! 😄

johngmyers:
Did we want to specify `pause` instead of `pause-amd64` in preparation for ARM support?

hakman:
> Did we want to specify `pause` instead of `pause-amd64` in preparation for ARM support?

For now I want to make sure it is compatible with older versions of Docker. Removing the `-amd64` suffix is next on the list.

rifelpet:
thats a lot of test references to a one line change :) 
/approve

hakman:
> thats a lot of test references to a one line change :)

Probably the most popular container image in the world 🤣

PR #9016, opened 2020-04-28
Update to etcd-manager 3.0.20200429

justinsb:
Contains the workaround for 1-year certificate expiry.

3.0.2020428 changes

* Release notes for 3.0.20200307 [#303](https://github.com/kopeio/etcd-manager/pull/303)
* Add support for etcd 3.3.17 [#304](https://github.com/kopeio/etcd-manager/pull/304)
* Adding client usage extension for server cert (#305) [#306](https://github.com/kopeio/etcd-manager/pull/306)
* Add a check to renew certificates on startup if they expire in 60 days or less [#309](https://github.com/kopeio/etcd-manager/pull/309)
* Try github actions [#310](https://github.com/kopeio/etcd-manager/pull/310)
* Upgrade bazel to 2.2.0 [#311](https://github.com/kopeio/etcd-manager/pull/311)
* Update to go 1.13.10 [#314](https://github.com/kopeio/etcd-manager/pull/314)
* Bazel: update dependency [#316](https://github.com/kopeio/etcd-manager/pull/316)
* e2e tests should wait for cluster readiness [#318](https://github.com/kopeio/etcd-manager/pull/318)
* Remove old bazel versions from travis [#317](https://github.com/kopeio/etcd-manager/pull/317)
* Always renew certificates [#313](https://github.com/kopeio/etcd-manager/pull/313)

3.0.20200429

    * Upgrade aws-sdk-go [#320](https://github.com/kopeio/etcd-manager/pull/320)
    * Release notes for 3.0.20200428 [#319](https://github.com/kopeio/etcd-manager/pull/319)

PR #9043, opened 2020-05-02
Automated cherry pick of #9016: Update to etcd-manager 3.0.20200428

hakman:
Cherry pick of #9016 on release-1.15.

#9016: Update to etcd-manager 3.0.20200428

### DIFF
--- a/pkg/model/components/etcd.go
+++ b/pkg/model/components/etcd.go
@@ -24,7 +24,7 @@ import (
 	"k8s.io/kops/upup/pkg/fi/loader"
 )
 
-const DefaultBackupImage = "kopeio/etcd-backup:3.0.20200428"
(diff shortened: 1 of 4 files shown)


hakman:
Let's get this in sooner rather than later. If there is anything that needs fixing, it can be addressed before the release.
/hold cancel

sheldonkwok:
I just noticed that the recommended [solution](https://kops.sigs.k8s.io/advisories/etcd-manager-certificate-expiration/#solution) says to use 20200428 and the MR says 20200428 but the actual commit does 20200429. Not sure if it's a big deal that it's one daily off

rifelpet:
The difference between those two versions is just an AWS SDK update so it should be trivial. You can see more release information here: https://github.com/kopeio/etcd-manager/releases

PR #9044, opened 2020-05-02
Ubuntu focal: Use generic docker & container packages

justinsb:
download.docker.com doesn't have packages for ubuntu 20.04 (focal),
use the generic tar.gz versions instead.

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -117,12 +117,12 @@ var containerdVersions = []packageVersion{
 
 	// 1.2.10 - Linux Generic
 	//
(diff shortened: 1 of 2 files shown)


justinsb:
Spotted this on the grid, and still testing, but cc @hakman as I think this makes your other PR more true :-)

hakman:
Will have to do something similar in 1.16 and 1.17 and use the 18.04 (Bionic) packages. Hope it works.

hakman:
Some better news, the previous fix for Amazon Linux 2 is working:
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-grid-calico-amzn2-k17

geojaz:
/lgtm 
:+1:

PR #9049, opened 2020-05-03
Validate more static pods on masters

johngmyers:
This should cut down on cluster validation flapping.

Not testing the etcd pods as their `k8s-app` annotation depends on whether or not the cluster is configured to use etcd-manager. (It's also unlikely kube-apiserver would be ready without them.) Not testing kube-proxy pods because whether or not they're supposed to be there depends on the cluster spec.

Suppressing the static pod checks on non-ready nodes to cut down on redundant failure messages.
/area rolling-update

### DIFF
--- a/pkg/validation/validate_cluster.go
+++ b/pkg/validation/validate_cluster.go
@@ -199,13 +199,22 @@ func (v *ValidationCluster) collectComponentFailures(ctx context.Context, client
 	return nil
 }
 
+var masterStaticPods = []string{
+	"kube-apiserver",
+	"kube-controller-manager",
+	"kube-scheduler",
+}
+
 func (v *ValidationCluster) collectPodFailures(ctx context.Context, client kubernetes.Interface, nodes []v1.Node) error {
-	masterWithoutManager := map[string]bool{}
+	masterWithoutPod := map[string]map[string]bool{}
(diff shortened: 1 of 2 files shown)


pkg/validation/validate_cluster.go
```diff
@@ -199,13 +199,22 @@ func (v *ValidationCluster) collectComponentFailures(ctx context.Context, client
 	return nil
 }
 
+var masterStaticPods = []string{
+	"kube-apiserver",
+	"kube-controller-manager",
+	"kube-scheduler",
```
hakman, line 205: Just curious, do `etcd-manager-*` startic pods get checked also?

johngmyers: No. If the cluster uses the legacy etcd provider, there won't be any `etcd-manager-*` pods. Checking the cluster's spec would add a lot of complexity. I suspect the kube-apiserver pods won't be able to go into ready state without etcd pods.

hakman: OK, probably not worth the trouble for now. Thanks.

justinsb:
Thanks @johngmyers  - looks great.  Not sure why we're not using spec.nodeName - maybe that is only populated in recent k8s versions?  (Do you remember?)

/approve
/lgtm

Issue #8834, opened 2020-04-02
Kops has a hard dependency on GitHub


rifelpet:
GitHub is currently experiencing an outage. Kops uses raw.githubusercontent.com to fetch the channels files.

During this outage, kops fails with:

```
kops update cluster
Using cluster from kubectl context: <my-cluster>
I0402 13:25:47.824966    5970 context.go:249] hit maximum retries 5 with error unexpected response code "500 Internal Server Error" for "https://raw.githubusercontent.com/kubernetes/kops/master/channels/stable": 500: Internal Server Error
error reading channel "https://raw.githubusercontent.com/kubernetes/kops/master/channels/stable": unexpected response code "500 Internal Server Error" for "https://raw.githubusercontent.com/kubernetes/kops/master/channels/stable": 500: Internal Server Error
```

We should have kops handle this more gracefully, continuing on unless thats not possible. For example, `update cluster` should work and just not report any channel updates, but `create cluster` may not since it might rely on the AMI or kubernetes version info.

geekofalltrades:
I would go a step further and say that Kops should not have a hard dependency on GitHub at all. We run Kops as part of an automated service and it's alarming to me that our production environment is regularly pulling down data from GitHub.

amadav:
Ran into this prod environment. Github seemed terribly slow and kops update cluster command kept failing to get the channels file. Is there a way to override the url at this moment before we have a better fix at this?

mariusv:
This days Github has quite few issues and this causes small "outages" due to the fact that one can't upgrade/provision clusters. Anything we can help with to get this issue going?

hakman:
Will add this on the discussion list for the next office hours and see what would be the best thing to do. :)

marek-obuchowicz:
We could store latest downloaded channel information in S3 state bucket. That could give a "last known" state of channel information to actions like update cluster

jiraguha:
This is a really bad issue and a sad issue and I have it:

```
 kops create cluster \
    --name=...
    --state=s3://$KOPS_BUCKET \
   ....
    --zones=...
    --node-count=$NODE_COUNT \
    --node-size=$NODE_SIZE \
    --master-size=$NODE_MASTER_SIZE
I0423 16:08:00.851481    3141 context.go:231] hit maximum retries 5 with error unexpected response code "500 Internal Server Error" for "https://raw.githubusercontent.com/kubernetes/kops/master/channels/stable": 500: Internal Server Error

[...]

PR #9053, opened 2020-05-03
Allow cluster maintenance when channel is unavailable

johngmyers:
Addresses some aspects of #8834, assuming no instancegroup is depending on the channel providing a default image to use.


hakman:
Really nice. I would see this cherry-picked to all live branches that will get the etcd-manager fix.
/lgtm

PR #9055, opened 2020-05-04
Move GCEServiceAccount into CloudConfig

johngmyers:
It's cloud-provider specific, so belongs in CloudConfig.

/cc @geojaz


geojaz:
love it, thank you.
/lgtm

PR #9056, opened 2020-05-04
Changing base image for node authorizer.

michalschott:
`fedora:27` is quite old

### DIFF
--- a/node-authorizer/Dockerfile
+++ b/node-authorizer/Dockerfile
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+FROM alpine:3.11 as ssl
+RUN apk --no-cache add ca-certificates && \
+    apk --no-cache upgrade
+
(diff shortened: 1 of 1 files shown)


michalschott:
Looks like there is no build pipeline set for node_authorizer.

```./pkg/model/components/node-authorizer/options.go:	return "quay.io/gambol99/node-authorizer:v0.0.4@sha256:078b948b8207e43d35885f181713de3d3c0491fe40661d198f9bc00136cff271"```

Could use some guidance to build one, and also serve image from "official" repo.

hakman:
Until this is addressed, you should build an image and test it locally. I see there is a field for that in the cluster spec:
https://github.com/kubernetes/kops/blob/318e7faa79a238faa90fa9ae1a147b42da6d5c0f/pkg/apis/kops/cluster.go#L209

rifelpet:
thanks for the update!
/approve

PR #9059, opened 2020-05-04
Update docs on supported distros and release notes

hakman:
3rd pass at this doc with very small changes.

### DIFF
--- a/docs/operations/images.md
+++ b/docs/operations/images.md
@@ -52,20 +52,20 @@ The following table provides the support status for various distros with regards
 | [Amazon Linux 2](#amazon-linux-2) | 1.10 | 1.18 | - | - |
 | [CentOS 7](#centos-7) | - | 1.5 | - | - |
 | [CentOS 8](#centos-8) | 1.15 | - | - | - |
-| [CoreOS](#coreos) | 1.6 | 1.9 | 1.17 | 1.19 |
-| [Debian 8](#debian-8-jessie) | - | 1.5 | 1.17 | 1.19 |
+| [CoreOS](#coreos) | 1.6 | 1.9 | 1.17 | 1.18 |
+| [Debian 8](#debian-8-jessie) | - | 1.5 | 1.17 | 1.18 |
 | [Debian 9](#debian-9-stretch) | 1.8 | 1.10 | - | - |
 | [Debian 10](#debian-10-buster) | 1.13 | 1.17 | - | - |
-| [Flatcar](#flatcar) | 1.15 | 1.17 | - | - |
+| [Flatcar](#flatcar) | 1.15.1 | 1.17 | - | - |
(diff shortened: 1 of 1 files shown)


johngmyers:
I had been thinking of doing the removal by Kubernetes version: have nodeup refuse to run on them for Kubernetes 1.18 and later.

Don't we want to put a version on Ubuntu 16.04?

hakman:
@johngmyers I don't think in this case there is a possibility to remove by Kubernetes version. Debian Jessie is already unusable and CoreOS will probably be soon. We can Just remove all traces of them. Ubuntu 16.04 still works, but it's just very old and buggy. Good catch that we should announce its removal in 1.20, which will be pretty close to the date it goes EOL.

johngmyers:
@hakman couldn't we do something like #9063?

hakman:
> @hakman couldn't we do something like #9063?

I don't understand why you think it's a good idea to do it based on Kubernetes version. This is not a case where these still work but we want to hide them. They are / will actually be broken and not start on new nodes.

johngmyers:
Is there something changing in kops 1.19 and 1.20 to cause these distributions to stop working?

Sure, at least CoreOS will be pretty untenable in that timeframe, but it's going to be a bit untenable in the kops 1.17 timeframe.

Generally speaking, removals by Kubernetes version can start happening sooner than removals by kops version. I believe it would be generally beneficial to have Kubernetes 1.18 on these older distributions out of the matrix.

hakman:
> Is there something changing in kops 1.19 and 1.20 to cause these distributions to stop working?
>
> Sure, at least CoreOS will be pretty untenable in that timeframe, but it's going to be a bit untenable in the kops 1.17 timeframe.

Jessie already doesn't work with anything, unless you have a private mirror.
CoreOS as you also said is gonna be gone in 3 weeks and probably everyone will notice.
So, these 2 won't work at all with any version of Kops.

> Generally speaking, removals by Kubernetes version can start happening sooner than removals by kops version. I believe it would be generally beneficial to have Kubernetes 1.18 on these older distributions out of the matrix.

For CoreOS and Jessie, I agree that 1.18 may be a better release to remove them.
For Ubuntu, I would still leave it around until 1.20, as it is still working and just old.

Removing support for these distros should be pretty simple. I would just return error in FindDistibution and say that it's no longer supported:
https://github.com/kubernetes/kops/blob/1400528fedba770cf001792796b37a19f9f8455b/nodeup/pkg/distros/identify.go#L31

johngmyers:
The "and will be removed in future versions of Kops" lines near the end of the document should list specific versions.

hakman:
> The "and will be removed in future versions of Kops" lines near the end of the document should list specific versions.

Updated. Will create a PR to remove CoreOS and Debian 8 and an issue to remind us about Ubuntu 16.04.

PR #9068, opened 2020-05-05
Added support for configuring disable-attach-detach-reconcile-sync in…

### LINKED ISSUE #9067: Add support for disable-attach-detach-reconcile-sync flag in kube-controller-manager
kubernetes/kubernetes#39551 added two flags to control the attach-detach reconcile sync in kube-controller-manager, `attach-detach-reconcile-sync-period` and `disable-attach-detach-reconcile-sync`, but only one of them got added to Kops in #1465.

We currently have the need to set `disable-attach-detach-reconcile-sync`, so I suggest adding `disable-attach-detach-reconcile-sync` as an option alongside `attach-detach-reconcile-sync-period` to Kops.

I can contribute the changes needed, if this sounds ok.

andersosthus:
… k-c-m config

Closes #9067

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -1395,6 +1395,11 @@ spec:
                     items:
                       type: string
                     type: array
+                  disableAttachDetachReconcileSync:
+                    description: DisableAttachDetachReconcileSync disables the reconcile
+                      sync loop in the attach-detach controller. This can cause volumes
+                      to become mismatched with pods
+                    type: boolean
                   enableProfiling:
                     description: EnableProfiling enables profiling via web interface
                       host:port/debug/pprof/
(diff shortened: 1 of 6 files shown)


andersosthus:
I did get an error? when running `make apimachinery crds`, but it seems to generate the CRDs properly, so not sure if it matters:

```
cannot find module providing package golang.org/x/tools/cmd/goimports: working directory is not part of a module
go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd paths=k8s.io/kops/pkg/apis/kops/v1alpha2 output:dir=k8s/crds/ crd:crdVersions=v1
```

andersosthus:
I also didn't find any good place to add tests for this, are options like this tested ?

hakman:
I think at the moment goimports needs go 1.13.
/ok-to-test

andersosthus:
> I think at the moment goimports needs go 1.13.

Ah, I run with go 1.14 locally, so that would explain it. I can switch to go 1.13 and do it again if needed.

hakman:
Let's see if tests pass. May not be needed for this small change.

PR #9069, opened 2020-05-05
kube-apiserver: healthcheck via sidecar container

justinsb:
kube-apiserver doesn't expose the healthcheck via a dedicated
endpoint, instead relying on anonyomous-access being enabled.  That
has previously forced us to enable the unauthenticated endpoint on
127.0.0.1:8080.

Instead we now run a small sidecar container, which
proxies /healthz and /readyz requests (only) adding appropriate
authentication using a client certificate.

This will also enable better load balancer checks in future, as these
have previously been hampered by the custom CA certificate.


Makefile
```diff
@@ -726,6 +728,14 @@ bazel-protokube-export:
 	cp -fp bazel-bin/images/protokube.tar.gz.sha1 ${BAZELIMAGES}/protokube.tar.gz.sha1
 	cp -fp bazel-bin/images/protokube.tar.gz.sha256 ${BAZELIMAGES}/protokube.tar.gz.sha256
 
+.PHONY: bazel-kube-apiserver-healthcheck-export
+bazel-kube-apiserver-healthcheck-export:
+	mkdir -p ${BAZELIMAGES}
+	DOCKER_REGISTRY="" DOCKER_IMAGE_PREFIX="kope/" KUBE_APISERVER_HEALTHCHECK_TAG=${KUBE_APISERVER_HEALTHCHECK_TAG} bazel build ${BAZEL_CONFIG} --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //cmd/kube-apiserver-healthcheck:image-bundle.tar.gz //cmd/kube-apiserver-healthcheck:image-bundle.tar.gz.sha1 //cmd/kube-apiserver-healthcheck:image-bundle.tar.gz.sha256
+	cp -fp bazel-bin/cmd/kube-apiserver-healthcheck/image-bundle.tar.gz ${BAZELIMAGES}/kube-apiserver-healthcheck.tar.gz
+	cp -fp bazel-bin/cmd/kube-apiserver-healthcheck/image-bundle.tar.gz.sha1 ${BAZELIMAGES}/kube-apiserver-healthcheck.tar.gz.sha1
```
johngmyers, line 736: Why do we need to create and ship SHA-1 hashes?

cmd/kube-apiserver-healthcheck/main.go
```diff
@@ -0,0 +1,172 @@
...
+		klog.Infof("error building request: %v", err)
+		http.Error(w, "internal error", http.StatusInternalServerError)
+		return
+	}
+	req.URL.RawQuery = r.URL.RawQuery
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		klog.Infof("error from %s: %v", target, err)
+		http.Error(w, "internal error", http.StatusInternalServerError)
```
johngmyers, line 96: `http.StatusBadGateway` would be more appropriate.

cmd/kube-apiserver-healthcheck/main.go
```diff
@@ -0,0 +1,172 @@
...
+// healthCheckServer is the http server
+type healthCheckServer struct {
+	tlsConfig *tls.Config
+}
+
+// handler processes a single http request
+func (s *healthCheckServer) handler(w http.ResponseWriter, r *http.Request) {
+	path := r.URL.Path
+	path = strings.TrimPrefix(path, "/")
+	tokens := strings.Split(path, "/")
```
johngmyers, line 42: My druthers would be to just `select(r.URL.Path)` and leave out any parsing of the path. There's a risk of a discrepancy between the parsing here and the parsing in `proxyRequest`, leading to request injection vulnerabilities.

cmd/kube-apiserver-healthcheck/main.go
```diff
@@ -0,0 +1,172 @@
...
+
+	klog.Infof("unknown request: %s %s", r.Method, r.URL.Path)
+	http.Error(w, "not found", http.StatusNotFound)
+}
+
+// httpClient builds an isolated http.Client
+func (s *healthCheckServer) httpClient() *http.Client {
+	transport := &http.Transport{
+		TLSClientConfig: s.tlsConfig,
+	}
```
johngmyers, line 71: It seems expensive to create a new `Transport` for each incoming query. Shouldn't this be one per `healthCheckServer`?

nodeup/pkg/model/kube_apiserver_healthcheck.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		if cert == nil {
+			return nil, nil, fmt.Errorf("certificate %q not found", signerID)
+		}
+
+		signerCertificate = cert
+	}
+
+	privateKey, err := NewPrivateKey(2048)
```
johngmyers, line 162: Please use `pki.GeneratePrivateKey()` and don't hardcode the key size.

nodeup/pkg/model/kube_apiserver_healthcheck.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, nil, err
+	}
+
+	signer, ok := signerKey.Key.(crypto.Signer)
+	if !ok {
+		return nil, nil, fmt.Errorf("private key was not a Signer: %T", signerKey.Key)
+	}
+
+	klog.Infof("signing certificate for %q", id)
+	cert, err := NewSignedClientCert(id, privateKey, signerCertificate.Certificate, signer)
```
johngmyers, line 173: Use `pki.SignNewCertificate()`

nodeup/pkg/model/kube_apiserver_healthcheck.go
```diff
@@ -0,0 +1,179 @@
...
+	{
+		c.AddTask(&nodetasks.UserTask{
+			Name:  userName,
+			UID:   userID,
+			Shell: "/sbin/nologin",
+			Home:  secretsDir,
+		})
+	}
+
+	clientKey, clientCert, err := b.buildClientKeypair(id)
```
johngmyers, line 90: Have `buildClientKeypair()` return a `pki.Certificate` and then use its `AsBytes()`. Use the pattern in `KubeletBuilder`

Makefile
mikesplain: I believe it's historical and we began migrating to sha256 about a year ago.  I'm not sure on the specifics, we can probably talk about a deprecation at some point, however I don't think this PR is the right time to do it.

justinsb: Good call - it's a new file, let's try removing the sha1, then we can find out if anyone is still using it (they shouldn't be!)

cmd/kube-apiserver-healthcheck/main.go
justinsb: Good call!

cmd/kube-apiserver-healthcheck/main.go
justinsb: Good idea - generally reworked to avoid all parsing!

cmd/kube-apiserver-healthcheck/main.go
justinsb: The fear is that they contaminate each other (e.g. cookies), but given our use case, you're right - simplifying.

nodeup/pkg/model/kube_apiserver_healthcheck.go
justinsb: I was thinking about creating simpler methods, but looking at pki, that's exactly what you've already done!

nodeup/pkg/model/kube_apiserver_healthcheck.go
justinsb: Will do!

pkg/assets/builder.go
```diff
@@ -164,6 +178,13 @@ func (a *AssetBuilder) RemapImage(image string) (string, error) {
 		}
 	}
 
+	if strings.HasPrefix(image, "kope/kube-apiserver-healthcheck:") {
+		override := os.Getenv("KUBE_APISERVER_HEALTHCHECK_IMAGE")
```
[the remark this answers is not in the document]
justinsb: Good catch!

Makefile
justinsb: Oops - our responses crossed.  The sha256 support (eca2ac6b80a2d2cb269e4494d3fa7cf23cba8a6c) should be in kops 1.15, so I am going ahead and removing the sha1.  But if it gives us trouble, let's add it back ...

nodeup/pkg/model/kube_apiserver_healthcheck.go
justinsb: Thanks - done!

Makefile
```diff
@@ -732,6 +734,13 @@ bazel-protokube-export:
 	cp -fp bazel-bin/images/protokube.tar.gz.sha1 ${BAZELIMAGES}/protokube.tar.gz.sha1
 	cp -fp bazel-bin/images/protokube.tar.gz.sha256 ${BAZELIMAGES}/protokube.tar.gz.sha256
 
+.PHONY: bazel-kube-apiserver-healthcheck-export
+bazel-kube-apiserver-healthcheck-export:
+	mkdir -p ${BAZELIMAGES}
+	DOCKER_REGISTRY="" DOCKER_IMAGE_PREFIX="kope/" KUBE_APISERVER_HEALTHCHECK_TAG=${KUBE_APISERVER_HEALTHCHECK_TAG} bazel build ${BAZEL_CONFIG} --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //cmd/kube-apiserver-healthcheck:image-bundle.tar.gz //cmd/kube-apiserver-healthcheck:image-bundle.tar.gz.sha1 //cmd/kube-apiserver-healthcheck:image-bundle.tar.gz.sha256
```
johngmyers, line 740:
```suggestion
	DOCKER_REGISTRY="" DOCKER_IMAGE_PREFIX="kope/" KUBE_APISERVER_HEALTHCHECK_TAG=${KUBE_APISERVER_HEALTHCHECK_TAG} bazel build ${BAZEL_CONFIG} --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //cmd/kube-apiserver-healthcheck:image-bundle.tar.gz //cmd/kube-apiserver-healthcheck:image-bundle.tar.gz.sha256
```

cmd/kube-apiserver-healthcheck/main.go
```diff
@@ -0,0 +1,164 @@
...
+
+// proxyRequest forwards a request, once it has been verified by handler
+func (s *healthCheckServer) proxyRequest(w http.ResponseWriter, r *http.Request) {
+	httpClient := s.httpClient()
+
+	u := &url.URL{
+		Scheme:   "https",
+		Host:     "127.0.0.1",
+		Path:     r.URL.Path,
+		RawQuery: r.URL.RawQuery,
```
johngmyers, line 72: Do you really want to pass through an unvetted `RawQuery`? It seems risky to me.

nodeup/pkg/model/kube_apiserver_healthcheck.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// https://tools.ietf.org/html/rfc5280#section-4.2.1.3
+	//
+	// Digital signature allows the certificate to be used to verify
+	// digital signatures used during TLS negotiation.
+	template.KeyUsage = template.KeyUsage | x509.KeyUsageDigitalSignature
+	// KeyEncipherment allows the cert/key pair to be used to encrypt
+	// keys, including the symmetric keys negotiated during TLS setup
+	// and used for data transfer.
+	template.KeyUsage = template.KeyUsage | x509.KeyUsageKeyEncipherment
```
johngmyers, line 191: By the way, `x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment` is the default for `pki.SignNewCertificate`.

cmd/kube-apiserver-healthcheck/main.go
justinsb: So it's not trivial.. https://github.com/kubernetes/kubernetes/blob/08e1fd3bb947faf465e8a67d5c7106dbd10840c0/cluster/gce/manifests/kube-apiserver.manifest#L39

I don't know what validation we could / should do here.  We could whitelist some queries I guess but I think that'll confuse people more.  I imagine two things are coming:

1) exposing this for use as a load balancer health-check (i.e. publicly)
2) using the richer readyz / livez queries as in the upstream manifest

Makefile
justinsb: Doh - thanks!

nodeup/pkg/model/kube_apiserver_healthcheck.go
justinsb: Ah - cool.  I suggest leaving it for now?  I think this function and the similar one to generate a kubeconfig end up being utils.

You can also see the beginnings of a thought I'm exploring here, which is putting less logic into nodeup.  The only parameter here is really the path and the CN, we could imagine this just being in the NodeUpConfig (or similar), meaning we would have more insight into exactly what nodeup does (because it would do less)

cmd/kube-apiserver-healthcheck/main.go
johngmyers: I think we need to parse the query and only pass through parameters we recognize as safe.

nodeup/pkg/model/kube_apiserver_healthcheck.go
johngmyers: Can leave for now. I'll look into extracting a function.

cmd/kube-apiserver-healthcheck/main.go
johngmyers: Looking at apiserver code, I see at least `exclude` and `verbose`, but that `http.Request` gets passed around all over the place.

So it's a risk assessment as to how likely it is that apiserver would have or later add an important authorization check on a parameter to those paths.

johngmyers: How about we parse and pass through `exclude`? I worry about information leakage from `verbose`.

justinsb: Great idea - I implemented that.  I also made it more testable and added a test!

nodeup/pkg/model/kube_apiserver_healthcheck.go
justinsb: Thanks - I don't think it's urgent, I just figured that's one way we could go, so it might not matter too much if we only have two of these (kubeconfig & "raw")

justinsb:
Cleaned up the request processing in the proxy & added some tests

cmd/kube-apiserver-healthcheck/proxy_test.go
```diff
@@ -0,0 +1,73 @@
...
+				URL:    u,
+			}
+			out := mapToProxyRequest(req)
+			actual := ""
+			if out != nil {
+				if out.Method != "GET" {
+					t.Fatalf("unexpected method %q", out.Method)
+				}
+				actual = out.URL.String()
+			}
```
johngmyers, line 63: This test wouldn't catch a bug where `mapToProxyRequest` returns an `Request` containing a URL of the empty string when it should be returning `nil`.

cmd/kube-apiserver-healthcheck/proxy_test.go
```diff
@@ -0,0 +1,73 @@
...
+				Method: "GET",
+				URL:    u,
+			}
+			out := mapToProxyRequest(req)
+			actual := ""
+			if out != nil {
+				if out.Method != "GET" {
+					t.Fatalf("unexpected method %q", out.Method)
+				}
+				actual = out.URL.String()
```
johngmyers, line 62:
```suggestion
				actual = out.URL.String()
				if actual == "" {
				        t.Fatalf("unexpected empty URL")
				}
```

justinsb: Thanks - commited, squashed, and I've put you as a co-author on the commit (as you are, at this point!)

PR #9072, opened 2020-05-06
Update instance_groups.md

ranshn:
the description and link for spotAllocationStrategy are currently taken from the Spot Fleet documentation, while kops uses EC2 Auto Scaling groups. these are two different tools/APIs, so I am proposing a fix to the readme file that would correct spotAllocationStrategy to point to the ASG documentation instead.

### DIFF
--- a/docs/instance_groups.md
+++ b/docs/instance_groups.md
@@ -143,9 +143,45 @@ which would end up in a drop-in file on nodes of the instance group in question.
 
 ## mixedInstancePolicy (AWS Only)
 
+### Example
+
+```
+apiVersion: kops.k8s.io/v1alpha2
+kind: InstanceGroup
+metadata:
+  generation: 2
+  labels:
+    kops.k8s.io/cluster: <clustername>
+  name: spot-4vcpu-16gb
+spec:
+  cloudLabels:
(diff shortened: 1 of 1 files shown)


ranshn:
assign @granular-ryanbonham

mikesplain:
Thanks @ranshn! This looks pretty good. Since you're in there, would you be up for adding an example of a complete `mixedInstancePolicy` config, as most of the other section have, so users can copy and paste and change each param? I don't have experience with them but I think that could be useful in addition to your cleanup.

/ok-to-test

ranshn:
> Thanks @ranshn! This looks pretty good. Since you're in there, would you be up for adding an example of a complete `mixedInstancePolicy` config, as most of the other section have, so users can copy and paste and change each param? I don't have experience with them but I think that could be useful in addition to your cleanup.
> 
> /ok-to-test

Done! let me know if this lgty or if you had something else in mind? maybe cut the non-mixedInstancePolicy parts out of the example?

gjtempleton:
/lgtm 
I think it probably makes most sense for users getting started as a full useful stanza that'll launch a full blown instance group.

PR #9074, opened 2020-05-06
Disable TX checksum offload for Flannel VXLAN

hakman:
Tx checksum offloading is buggy for NAT-ed VXLAN endpoints, leading to an invalid checksum sent and causing Flannel to stop to working as the traffic is being discarded by the receiver.
https://github.com/coreos/flannel/issues/1279

/cc @johngmyers @justinsb

### DIFF
--- a/nodeup/pkg/model/network.go
+++ b/nodeup/pkg/model/network.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"k8s.io/klog"
+	"k8s.io/kops/pkg/systemd"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
 )
@@ -66,6 +68,13 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	// Tx checksum offloading is buggy for NAT-ed VXLAN endpoints, leading to an invalid checksum sent and causing
+	// Flannel to stop to working as the traffic is being discarded by the receiver.
+	// https://github.com/coreos/flannel/issues/1279
+	if networking != nil && (networking.Canal != nil || (networking.Flannel != nil && networking.Flannel.Backend == "vxlan")) {
+		c.AddTask(b.buildFlannelTxChecksumOffloadDisableService())
+	}
+
 	return nil
 }
 
@@ -87,4 +96,28 @@ func (b *NetworkBuilder) addCNIBinAsset(c *fi.ModelBuilderContext, assetName str
 	})
 
 	return nil
+}
+
+func (b *NetworkBuilder) buildFlannelTxChecksumOffloadDisableService() *nodetasks.Service {
+	const serviceName = "flannel-tx-checksum-offload-disable.service"
+
+	manifest := &systemd.Manifest{}
+	manifest.Set("Unit", "Description", "Disable TX checksum offload on flannel.1")
+
+	manifest.Set("Unit", "After", "sys-devices-virtual-net-flannel.1.device")
+	manifest.Set("Install", "WantedBy", "sys-devices-virtual-net-flannel.1.device")
+	manifest.Set("Service", "Type", "oneshot")
+	manifest.Set("Service", "ExecStart", "/sbin/ethtool -K flannel.1 tx-checksum-ip-generic off")
+
+	manifestString := manifest.Render()
+	klog.V(8).Infof("Built service manifest %q\n%s", serviceName, manifestString)
+
+	service := &nodetasks.Service{
+		Name:       serviceName,
+		Definition: s(manifestString),
+	}
+
+	service.InitDefaults()
+
+	return service
 }


nodeup/pkg/model/network.go
```diff
@@ -66,6 +68,13 @@ func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	// Tx checksum offloading is buggy for NAT-ed VXLAN endpoints, leading to an invalid checksum sent and causing
+	// Flannel to stop to working as the traffic is being discarded by the receiver.
+	// https://github.com/coreos/flannel/issues/1279
+	if networking != nil && (networking.Canal != nil || (networking.Flannel != nil && networking.Flannel.Backend == "vxlan")) {
+		c.AddTask(b.builddFlannelTxChecksumOffloadDisableService())
```
rifelpet, line 75:
```suggestion
		c.AddTask(b.buildFlannelTxChecksumOffloadDisableService())
```

nodeup/pkg/model/network.go
```diff
@@ -88,3 +97,27 @@ func (b *NetworkBuilder) addCNIBinAsset(c *fi.ModelBuilderContext, assetName str
 
 	return nil
 }
+
+func (b *NetworkBuilder) builddFlannelTxChecksumOffloadDisableService() *nodetasks.Service {
```
rifelpet, line 101:
```suggestion
func (b *NetworkBuilder) buildFlannelTxChecksumOffloadDisableService() *nodetasks.Service {
```

nodeup/pkg/model/network.go
```diff
@@ -88,3 +97,27 @@ func (b *NetworkBuilder) addCNIBinAsset(c *fi.ModelBuilderContext, assetName str
...
+	const serviceName = "flannel-tx-checksum-offload-disable.service"
+
+	manifest := &systemd.Manifest{}
+	manifest.Set("Unit", "Description", "Disable TX checksum offload on flannel.1")
```
rifelpet, line 105: does canal also use the name `flannel.1` ?

hakman: yes, I tested this with Canal

nodeup/pkg/model/network.go
hakman: Thanks. Fixed now.

nodeup/pkg/model/network.go
hakman: Thanks. Fixed.

johngmyers:
Fixes #8562

hakman:
@johngmyers do we want to also bump Flannel to 0.12?
Also, if someone can test this more thoroughly would be good.

nodeup/pkg/model/network.go
joshbranham: Yeah we run `canal` and `flannel.1` is the interface name

johngmyers:
I would like to do the least amount of change needed to get 1.17 out the door. So unless bumping Flannel is necessary for 1.17, let's do that in a separate PR.

rifelpet:
/approve

looking forward to seeing how this changes our testgrid dashboard

hakman:
@johngmyers Agreed, flannel stays 0.11 for now and we will consider 0.12 for 1.18.

nodeup/pkg/model/network.go
vvbogdanov87: Not in my setup
```
admin@ip-10-63-80-157:~$ ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group
...
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc mq state UP group 
...
3: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state 
...
```
This change breaks my cluster. The service `flannel-tx-checksum-offload-disable` returns an error that prevents `kops-configuration` to finish its job.
```
admin@ip-10-63-80-157:~$ sudo /sbin/ethtool -K flannel.1 tx-checksum-ip-generic off
Cannot get device feature names: No such device
```
I'm running on AWS using canal
```
  networking:
    canal: {}
```

hakman: How does this block `kops-configuration` from finishing the job? Worst case scenario it stays as the last job.

vvbogdanov87: Sorry. My bad. Looks like flannel pod wasn't started because the node wasn't able to join the cluster. As a result, the device wasn't created and kops-configuration service log was spammed by the messages from flannel-tx-checksum-offload-disable service. I had to manually fix the issue with etcd manager certificate first(my cert was already expired and rolling upgrade to the latest kops 1.17.0 failed).

KashifSaadat:
Looks like this has now been fixed in Kubernetes (https://github.com/kubernetes/kubernetes/pull/92035), so we may be able to remove this once it has made its way into the relevant releases (k8s v1.16.13, v1.17.9, v1.18.6).

hakman:
Agreed. Those releases should be out in about 2 weeks. Probably would be good to remove it now for 1.18 and 1.19. Do we want to remove it from earlier releases also?

KashifSaadat:
For 1.16 and 1.17 I think that would make sense yeah 👍

PR #9080, opened 2020-05-06
Reduce the number of TravisCI jobs for release branch

hakman:
For release branches we only need Linux and macOS with the default GO version.
This should help reduce the load on macOS jobs and, in general, before releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ os:
   - linux
   - osx
 go:
-  # 1.12 is now best-effort; we recommend 1.13
-  - "1.12"
   - "1.13"
 
 go_import_path: k8s.io/kops


johngmyers:
Why wouldn't we need the MacOS build on release branches? The kops binaries released through brew are built from those branches, are they not?

hakman:
@johngmyers I don't think the build changes in release branches would impact macOS builds. Travis checks has nothing to do with what we publish. We have windows binaries also, but no test jobs for them at all.

johngmyers:
The Travis job ensures that commits into the release branches don't break the MacOS builds. If we don't have the checks, the MacOS build could be broken by a bad backport.

If we care about ensuring kops can build on Windows then we should have a Windows build as a CI gate.

I do agree that ensuring release branches build on older versions of Go is unimportant.

hakman:
OK, let me add back the macOS one and at least reduce this by half.

PR #9081, opened 2020-05-06
Reduce the number of TravisCI jobs for release branch

hakman:
For release branches we only need Linux and macOS with the default GO version.
This should help reduce the load on macOS jobs and, in general, before releases.


MoShitrit:
@hakman I'm curious- is the removal of OS X because we're moving to GitHub actions?
Also, 
/lgtm

PR #9091, opened 2020-05-07
Spotinst: Allow users to disable the controller add-on

liranp:
This PR adds a new feature flag, named `SpotinstController`, that allows users to toggle the installation of the Spotinst controller off. The feature flag defaults to `true` to maintain backward compatibility.

### DIFF
--- a/pkg/featureflag/featureflag.go
+++ b/pkg/featureflag/featureflag.go
@@ -76,6 +76,8 @@ var (
 	Spotinst = New("Spotinst", Bool(false))
 	// SpotinstOcean toggles the use of Spotinst Ocean instance group implementation.
 	SpotinstOcean = New("SpotinstOcean", Bool(false))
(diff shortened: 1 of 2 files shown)


rifelpet:
Out of curiosity, in what situations would a user want to disable the controller? If they want to run the controller themselves?

rifelpet:
Got it, that makes sense.

Can we add the new FF to the docs? at least `docs/advanced/experimental.md`.

I'm noticing we don't have any docs on using the Spotinst provider, but adding that or at least a link to Spotinst docs elsewhere can always happen in a separate PR :)

liranp:
Actually, I'm working on comprehensive documentation for the Spotinst provider, so I'll just add that to there and raise a separate PR really soon :)

rifelpet:
sounds great!
/lgtm
/approve

PR #9094, opened 2020-05-08
Implement VFS for vault

### LINKED ISSUE #8009: Allow vault to be a secret store for PKI information and other secrets
The high-level feature is to allow Vault to be used as the secret store in kops as opposed to always using a sub-path of the VFS state store for secrets.

The main benefit is that organizations already using Vault do not have to manage secrets in multiple places and can always rely on Vault being the source of truth for all secrets.

In addition, vault can allow secure introduction of masters and nodes using cloud integrations [example](https://www.vaultproject.io/docs/auth/aws.html#ec2-auth-method)

Multiple levels of integration are possible:

* Just replace current store as-is with a Vault KV store and just push and pull secrets from different places.
* Use Vault's PKI infrastructure to generate keys and certs on the fly. This is likely to be far more tricky than the replacement method (which itself would be tricky).

The kops codebase does have a notion of a secret store distinct from the state store so this is not a massive code refactor (unless we have to change the secret store API itself to be able to handle vault).

As of this point, this is just a wish list without a backing proposal as to how this can be accomplished. I will add a detailed document after some analysis and maybe an initial POC.

olemarkus:
This is an attempt at a VFS implementation for vault.
The goal is to be able to use as a key and secret store. It will not work to use this as a kops store/config store. Among other things, etcd-manager is not able to read VFS control file from vault.

Vault client on the cluster nodes will authenticate itself using IAM auth and expects this to be mounted on /aws. Vault client locally will expect `VAULT_TOKEN` to be set.

Vault client will use TLS by default. You can use plain text by adding "?tls=false" to the VFS url.

Tests can be run by setting running `vault server -dev` locally and with `VAULT_DEV_ROOT_TOKEN_ID` exported.

I had to hack a bit around elsewhere in the code as well to support `spec.keyStore` and `spec.secretStore` fully. These config keys were ignored in several places in the code.

To test this now, configure vault to use aws auth following https://learn.hashicorp.com/vault/identity-access-management/iam-authentication

Add this to the spec:
```
  secretStore: vault://<vault>:<port>/secret/clusters/<clustername>/secrets
  keyStore: vault://<vault>:<port>/secret/clusters/<clustername>/keys
```

When writing the ACL, do something like this _after_ running `kops update cluster --yes <clustername>`

```
vault write auth/aws/role/masters.<clustername> auth_type=iam \
              bound_iam_principal_arn=arn:aws:iam::<account>:role/masters.<clustername> policies=<my policy> max_ttl=500h
vault write auth/aws/role/nodes.<clustername> auth_type=iam \
              bound_iam_principal_arn=arn:aws:iam::<account>:role/nodes.<clustername> policies=<my policy> max_ttl=500h
```

Note that if you re-provision your cluster, you need to re-run the above in order for vault to update the role internal IDs.

Fixes #8009


util/pkg/vfs/vaultcontext.go
```diff
@@ -0,0 +1,89 @@
...
+	if token != "" {
+		client.SetToken(token)
+	} else {
+		return awsAuth(client)
+
+	}
+	return nil
+}
+
+func awsAuth(client *vault.Client) error {
```
rifelpet, line 57: I wonder if we could use fi.Cloud here? or something that would allow the aws client to be mocked more easily.

util/pkg/vfs/vaultcontext.go
```diff
@@ -0,0 +1,89 @@
...
+
+	return client, err
+}
+
+func authenticateVaultClient(client *vault.Client) error {
+	token := os.Getenv("VAULT_TOKEN")
+	if token != "" {
+		client.SetToken(token)
+	} else {
+		return awsAuth(client)
```
rifelpet, line 51: i dont have much experience with vault or other cloud providers, but is this something we could easily extend to support other cloud providers?

olemarkus: Sure. We have https://www.vaultproject.io/docs/auth/gcp.html too. It should be quite easy to extend to. Is there an easy way to know the cloud provider in that part of the code? If so, I can add some conditional here. Someone with access to GCP would have to do the actual implementation though.

util/pkg/vfs/vaultcontext.go
olemarkus: fi.Cloud is new to me. I'll have a look.

olemarkus:
Thanks for the comments. Was actually working on tightening this up now.

util/pkg/vfs/vaultcontext.go
olemarkus: Ah doh. The problem with fi.Cloud is that it causes import cycles. I cannot make use of anything from fi or from apis. So it gets a bit tricky to test for cloud provider.

util/pkg/vfs/vaultcontext.go
johngmyers: Currently `VFSContext` is implicitly initialized. It expects to be able to get credentials out of the running environment and figures out which cloud provider to use by the protocol part of the input URL. So other cloud providers would either have to use a protocol other than "vault:" or would have to key off of something environmental.

I wonder if `VFSContext` should be explicitly initialized with the `ClusterSpec`. I think the first refactor (for a different PR) would be to remove the global `vfs.Context`.

util/pkg/vfs/vaultcontext.go
johngmyers: An in-package unit test could stuff a mock inside a field added to the `VFSContext` struct.

util/pkg/vfs/vaultcontext.go
olemarkus: The way I left this now, is to not care about the error from awsauth. So one could attempt auth with gcp or similar if vault still hasn't set a token. It is not ideal, but it will work.

Refactor later on to get cloud provider passed in makes sense, but we have to sync up with etcd-manager here since they also use VFS from kops (there is even logic only used by etcd-manager in kops VFS)

pkg/apis/kops/validation/legacy.go
```diff
@@ -512,6 +516,14 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 
 	allErrs = append(allErrs, newValidateCluster(c)...)
 
+	if kops.CloudProviderID(c.Spec.CloudProvider) != kops.CloudProviderAWS {
```
johngmyers, line 519: Wouldn't this otherwise work on other cloud providers when using VAULT_TOKEN?

docs/state.md
```diff
@@ -186,3 +186,32 @@ if err != nil {
...
+
+To configure your vault server to accept IAM Auth authentication, configure vault to use aws auth following https://learn.hashicorp.com/vault/identity-access-management/iam-authentication
+
+To configure kops to use the Vault store, add this to the spec:
+
+```yaml
+spec:
+  secretStore: vault://<vault>:<port>/secret/clusters/<clustername>/secrets
+  keyStore: vault://<vault>:<port>/secret/clusters/<clustername>/keys
+```
```
johngmyers, line 204: Perhaps make it clear that the path to the secrets/keys is an example, not a requirement?

pkg/client/simple/vfsclientset/clientset.go
```diff
@@ -154,7 +172,46 @@ func DeleteAllClusterState(basePath vfs.Path) error {
...
 }
 
+func deleteAllPaths(basePath vfs.Path) error {
+	paths, err := basePath.ReadTree()
+	if err != nil {
+		return fmt.Errorf("error listing files in state store: %v", err)
+	}
+
+	for _, path := range paths {
+		err = path.Remove()
```
johngmyers, line 182: Does this need to ignore "NotFound" type errors?

pkg/client/simple/vfsclientset/clientset.go
```diff
@@ -154,7 +172,46 @@ func DeleteAllClusterState(basePath vfs.Path) error {
...
+		}
+	}
+
+	keyStore := cluster.Spec.KeyStore
+	if keyStore != "" && keyStore != secretStore {
+		path, err := vfs.Context.BuildVfsPath(keyStore)
+		if err != nil {
+			return err
+		}
+		err = deleteAllPaths(path)
```
johngmyers, line 209: Also ignore "NotFound" type errors?

util/pkg/vfs/context.go
```diff
@@ -422,3 +430,35 @@ func (c *VFSContext) buildOSSPath(p string) (*OSSPath, error) {
...
+		schema := "https://"
+		if queryValues.Get("tls") == "false" {
+			schema = "http://"
+		}
+
+		vaultClient, err := newVaultClient(schema + u.Host)
+		if err != nil {
+			return nil, err
+		}
+		err = authenticateVaultClient(vaultClient)
```
johngmyers, line 457: Perhaps `newVaultClient()` could also authenticate? What's the purpose of being able to get an unauthenticated Vault client?

util/pkg/vfs/vaultfs.go
```diff
@@ -0,0 +1,257 @@
...
+	path        string
+}
+
+var _ Path = &VaultPath{}
+
+func newVaultPath(client *vault.Client, path string) (*VaultPath, error) {
+	path = strings.TrimPrefix(path, "/")
+	dirs := strings.SplitN(path, "/", 2)
+	if len(dirs) != 2 {
+		return nil, fmt.Errorf("vault path must have both a mount point and a path. Got: %s", path)
```
johngmyers, line 45:
```suggestion
		return nil, fmt.Errorf("vault path must have both a mount point and a path. Got: %q", path)
```

util/pkg/vfs/vaultfs.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	if client == nil {
+		return nil, fmt.Errorf("vault paths needs to have a vault client")
```
johngmyers, line 49:
```suggestion
		return nil, fmt.Errorf("vault path needs to have a vault client")
```

util/pkg/vfs/vaultfs.go
```diff
@@ -0,0 +1,257 @@
...
+
+	encodedString := data["file"].(string)
+	return base64.StdEncoding.DecodeString(encodedString)
+}
+
+func (p *VaultPath) Remove() error {
+	klog.V(8).Infof("removing file %s", p)
+
+	data := map[string][]string{
+		"versions": {"1"},
```
johngmyers, line 107: This only deletes version 1? What if the data got updated?

docs/state.md
```diff
@@ -186,3 +186,32 @@ if err != nil {
 gcsClient, err := storage.New(httpClient)
 
 ```
+
+## Vault (vault://)
+
+The goal of the vault store is to be a safe storage for the kops keys and secrets store. It will not work to use this as a kops registry/config store. Among other things, etcd-manager is not able to read VFS control file from vault. Vault can also not be used as backend for etcd backups.
+
+The vault store uses IAM auth to authenticate against the vault server and expects the vault auth plugin to be mounted on /aws. Vault client locally will expect VAULT_TOKEN to be set.
```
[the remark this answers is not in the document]
olemarkus: This isn't entirely accurate.
On a server, it uses AWS auth to authenticate to Vault server.
When running kops locally, you naturally cannot use Vault server, and kops expects `VAULT_TOKEN` env variable to be set. Any authentication method using `vault` client will generate a token, so the auth method doesn't really matter.

pkg/apis/kops/validation/legacy.go
olemarkus: If you pass a static token, then I suppose this will work. Injecting a static token into each server in a safe way would be difficult though.

pkg/client/simple/vfsclientset/clientset.go
olemarkus: In the unfortunate scenarios of trying to delete these paths while something else is writing to them?
Could be. This function is a copy of the function above though, just not checking if the root is well-known or not (since we explicitly pass the secret/key store here)

util/pkg/vfs/context.go
olemarkus: Originally, it was in order to do testing. But right now, this can probably be moved in as tests use this factory function anyway.

util/pkg/vfs/vaultfs.go
olemarkus: It doesn't matter. It is more to satisfy the requirement of the destroy method. After the secret metadata is also deleted, the secret is gone.

docs/state.md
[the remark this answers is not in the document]
johngmyers: I found the sentence confusing and could not quite figure out what it was trying to convey. The term "vault client locally" is ambiguous/unclear. It seems the intent is to use `VAULT_TOKEN` when running the kops CLI and IAM during nodeup. The code simply keys off of the presence/absence of `VAULT_TOKEN`.

Getting a token to nodeup would require some sort of customization, but I could see a site injecting a script to use a (non-AWS) instance credential to obtain one.

It's not immediately obvious to me why one couldn't use IAM when running the kops CLI, aside from the code's current expectation for the plugin to be at a hardcoded `/aws`.

Perhaps it would be useful to separate the descriptions of mechanism and intent.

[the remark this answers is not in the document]
johngmyers: There was also confusion about whether VAULT_TOKEN was intended for use in integration tests against an on-box vault server.

pkg/client/simple/vfsclientset/clientset.go
johngmyers: In the scenario of trying to delete these paths when they have been deleted by a previous (failed) run of kops.

util/pkg/vfs/vaultfs.go
johngmyers: Comment resolved. I don't have detailed knowledge of the Vault API.

pkg/apis/kops/validation/legacy.go
johngmyers: I don't think it's a requirement to allow such customization, I just didn't want to foreclose it without consideration. It looks like it might be just as easy to extend kops for a new cloud provider as customize it with scripts.

pkg/client/simple/vfsclientset/clientset.go
olemarkus: `ReadTree` will return all _files_ that exists within that tree, and we only give the method the root of the secret/key store. So if a failed run of kops only delete some of the files, the rest would be automatically deleted on subsequent runs.

util/pkg/vfs/context.go
olemarkus: Moved tis in now

docs/state.md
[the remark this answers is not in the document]
olemarkus: Tried to make this a bit more clear now.

docs/state.md
olemarkus: Tried to make this more clear.

rifelpet:
do you think we should guard this behind a feature flag? I realize all of the new behavior is "opt in"  by choosing to use `vault://` in the cluster spec, but a feature flag allows us to make breaking changes more easily if any issues arise after this makes it into a stable kops release.

olemarkus:
Feature flag doesn't hurt. There is a slight chance that the VFS URL could change schema, e.g by changing how one sets the KV mount point.

util/pkg/vfs/vaultcontext.go
```diff
@@ -0,0 +1,91 @@
...
+func awsAuth(client *vault.Client) error {
+	config := aws.NewConfig()
+	config = config.WithCredentialsChainVerboseErrors(true)
+	session, err := session.NewSession(config)
+	if err != nil {
+		return err
+	}
+	svc := sts.New(session)
+	var params *sts.GetCallerIdentityInput
+	stsRequest, _ := svc.GetCallerIdentityRequest(params)
```
johngmyers, line 68: This is a verbose way of passing `nil` as the parameter.

johngmyers, line 68: Perhaps shove an `X-Vault-AWS-IAM-Server-ID` header into the request before signing it?

olemarkus: That is why that code is there. But I think this needs to be configurable as it has to match the config on the vault side, and I wasn't entirely sure how to best do this. I could do this as another query string parameter, I guess.

johngmyers: I would say pass the domain name of the Vault server.

(I really don't like how lax Vault is about authentication replay attacks.)

olemarkus: Done

util/pkg/vfs/vaultcontext.go
```diff
@@ -0,0 +1,93 @@
...
+	client, err := vault.NewClient(config)
+	if err != nil {
+		return nil, err
+	}
+
+	token := os.Getenv("VAULT_TOKEN")
+	if token == "" {
+		token, err = awsAuth(client, host)
+		if err != nil {
+			klog.Infof("Error authenticating vault to AWS: %v", err)
```
johngmyers, line 50: Shouldn't we return err here, not below? And we shouldn't both log and return an error.

olemarkus: We should return here. This was a leftover of an attempt at chaining with other auth methods.
But I think we need to move fi.Cloud to another package in order to do so, so I can use it here without running into import cycles.

docs/state.md
```diff
@@ -186,3 +186,47 @@ if err != nil {
...
+The vault store uses IAM auth to authenticate against the vault server and expects the vault auth plugin to be mounted on `/aws`.
+
+Instructions for configuring your vault server to accept IAM authentication are at https://learn.hashicorp.com/vault/identity-access-management/iam-authentication
+
+To configure kops to use the Vault store, add this to the cluster spec:
+
+```yaml
+spec:
+  secretStore: vault://<vault>:<port>/<kv2 mount>/clusters/<clustername>/secrets
+  keyStore: vault://<vault>:<port>/<kv2 mount/clusters/<clustername>/keys
```
rifelpet, line 211: the angled brackets aren't balanced on this line

docs/state.md
```diff
@@ -186,3 +186,47 @@ if err != nil {
...
+```
+
+The `secret` and `key` paths are configurable, but they have to be unique and you should not put other secrets in them.
```
rifelpet, line 214: unique per "what" exactly? per cluster? can you elaborate on that a bit

docs/state.md
```diff
@@ -186,3 +186,47 @@ if err != nil {
...
+
+After launching your cluster you need to add the cluster roles to Vault, binding them to the cluster's IAM identity and granting them access to the appropriate secrets and keys. The nodes will wait until they can authenticate before completing provisioning.
+
+```sh
+vault write auth/aws/role/masters.<clustername> auth_type=iam \
+              bound_iam_principal_arn=arn:aws:iam::<account>:role/masters.<clustername> policies=<my policy> max_ttl=500h
```
rifelpet, line 220: not knowing too much about how vault works, does kops create the vault policy automatically? or does that need to be externally managed? if so should we provide an example policy to use?

util/pkg/vfs/vaultcontext.go
```diff
@@ -0,0 +1,93 @@
...
+		HttpClient: &http.Client{},
+	}
+	client, err := vault.NewClient(config)
+	if err != nil {
+		return nil, err
+	}
+
+	token := os.Getenv("VAULT_TOKEN")
+	if token == "" {
+		token, err = awsAuth(client, host)
```
rifelpet, line 51: Maybe the docs should mention this AWS credential fallback in the `### Client configuration` section, since they seem to suggest that VAULT_TOKEN env var is required

olemarkus:
I am all for adding docs on how to test. But I am wondering where I should put it. If these things goes into docs, it can easily become messy. But I could put it in the test files.
You only need to run the vault dev server. Instructions on how to start it can be found here:  https://learn.hashicorp.com/vault/getting-started/dev-server

olemarkus:
/assign @justinsb can you look at the VFS code?

I'll do the docs improvement above meanwhile.

util/pkg/vfs/vaultcontext.go
olemarkus: awsAuth does calls to the metadata API, so it won't really work form outside of the node. hence, for clients, `VAULT_TOKEN` is needed.

docs/state.md
olemarkus: Added an example now. Could make kops create policies later, but at least we would not like to let operators have that kind of privilege in vault.

docs/state.md
olemarkus: more clear now?

olemarkus:
Just added the notes now @rifelpet

docs/state.md
```diff
@@ -186,3 +186,68 @@ if err != nil {
...
+path "kv-v2/metadata/clusters/<clustername>/*" {
+  capabilities = ["list", "read"]
+}
+
+path "kv-v2/data/clusters/<clustername>/*" {
+  capabilities = ["read"]
+}
+```
+
+Once you add this policy, you can assign them to the IAM roles like this:
```
johngmyers, line 237: This section is inconsistent in its use of number. Either use singular policy or plural policies. So either "Once you add these policies, you can assign them" or "Once you add this policy, you can assign it".

docs/state.md
```diff
@@ -186,3 +186,68 @@ if err != nil {
...
+  secretStore: vault://<vault>:<port>/<kv2 mount>/clusters/<clustername>/secrets
+  keyStore: vault://<vault>:<port>/<kv2 mount>/clusters/<clustername>/keys
+```
+
+Each of the paths specified above can be configurable, but they must be unique across all clusters. You can also not use the same path as both `stateStore` and `keyStore`.
+
+After launching your cluster you need to add the cluster roles to Vault, binding them to the cluster's IAM identity and granting them access to the appropriate secrets and keys. The nodes will wait until they can authenticate before completing provisioning.
+
+#### Vault policies
+Note that contrary to the S3 state store, kops will not provision any policies for you. You have to provide both the role for operators and the nodes.
```
johngmyers, line 219:
```suggestion
Note that contrary to the S3 state store, kops will not provision any policies for you. You have to provide roles for both operators and nodes.
```

docs/state.md
```diff
@@ -186,3 +186,68 @@ if err != nil {
...
+}
+```
+
+Once you add this policy, you can assign them to the IAM roles like this:
+
+```sh
+vault write auth/aws/role/masters.<clustername> auth_type=iam \
+              bound_iam_principal_arn=arn:aws:iam::<account>:role/masters.<clustername> policies=<my policy> max_ttl=500h
+vault write auth/aws/role/nodes.<clustername> auth_type=iam \
+              bound_iam_principal_arn=arn:aws:iam::<account>:role/nodes.<clustername> policies=<my policy> max_ttl=500h
```
johngmyers, line 243:
```suggestion
vault write auth/aws/role/masters.<clustername> auth_type=iam \
              bound_iam_principal_arn=arn:aws:iam::<account>:role/masters.<clustername> policies=<policy> max_ttl=500h
vault write auth/aws/role/nodes.<clustername> auth_type=iam \
              bound_iam_principal_arn=arn:aws:iam::<account>:role/nodes.<clustername> policies=<policy> max_ttl=500h
```

docs/state.md
```diff
@@ -186,3 +186,68 @@ if err != nil {
...
+vault write auth/aws/config/client iam_server_id_header_value="<vault server hostname>"
+```
+
+Note that if you re-provision your cluster, you need to re-run the above in order for Vault to update the role internal IDs.
+
+Vault will use TLS by default. If you want to use plaintext instead, add `?tls=false` to the url.
+
+### Client configuration
+
+The `kops` CLI only expects `VAULT_TOKEN` environment variable set to a valid token. You can use any authentication method to obtain a token and then set it manually if the authentication method does not do that automatically.
```
johngmyers, line 253:
```suggestion
The `kops` CLI only expects the `VAULT_TOKEN` environment variable to be set to a valid token. You can use any authentication method to obtain a token and then set it manually if the authentication method does not do that automatically.
```

rifelpet:
gave this another look and it looks good to me. Looking forward to getting this in and getting feedback from users, thanks!

/approve

PR #9095, opened 2020-05-08
kube-apiserver-healthcheck: actually enable on 1.17

justinsb:
We cherry picked, but now we need to activate the feature.

### DIFF
--- a/pkg/model/components/apiserver.go
+++ b/pkg/model/components/apiserver.go
@@ -296,7 +296,7 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
 		c.AnonymousAuth = fi.Bool(false)
 	}
 
-	if b.IsKubernetesGTE("1.18") {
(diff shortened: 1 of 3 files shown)


justinsb:
I expect this to fail until we merge #9096

pkg/model/components/apiserver.go
```diff
@@ -296,7 +296,7 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
 		c.AnonymousAuth = fi.Bool(false)
 	}
 
-	if b.IsKubernetesGTE("1.18") {
+	if b.IsKubernetesGTE("1.17") {
```
johngmyers, line 299: You'll want to change this on master branch as well. How about first landing to master and then cherrypicking?

mikesplain: Fair point, that's an easy follow up.

justinsb: Oops - will send PR to master now :-)

PR #9097, opened 2020-05-08
Fix port conflict on etcd-cilium vs dns-controller memberlist

justinsb:
Both were allocated port 3993; etcd-cilium is new and so it is less
impactful to renumber that.

### DIFF
--- a/pkg/wellknownports/wellknownports.go
+++ b/pkg/wellknownports/wellknownports.go
@@ -30,7 +30,7 @@ const (
 	EtcdEventsGRPC = 3997
 
 	EtcdCiliumQuarantinedClientPort = 3992
-	EtcdCiliumGRPC                  = 3993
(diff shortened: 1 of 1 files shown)


mikesplainsonos:
Thanks!

/lgtm
/approve

justinsb:
cc @olemarkus is it OK to renumber this?  I know it's not great, but I think it's less risky that renumbering dns, as etcd-cilium is still new.

(Or is there some reason I'm missing why we can share ports here?)

mikesplain:
Thanks @justinsb!

/lgtm
/approve

(used the wrong account at first)

Marking as hold until you're ready
/hold

olemarkus:
No problem at all to change the cilium etcd port

mikesplain:
Awesome, removing hold then.

/hold cancel

Issue #9099, opened 2020-05-08
containerd cannot pull kube-apiserver-healthcheck


hakman:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
```
kops create cluster \
  --name e2e-kops-aws-containerd.test-cncf-aws.k8s.io \
  --ssh-public-key /workspace/.ssh/kube_aws_rsa.pub \
  --node-count 4 \
  --node-volume-size 48 \
  --master-volume-size 48 \
  --master-count 1 \
  --zones ap-southeast-1a \
  --master-size c5.large \
  --kubernetes-version https://storage.googleapis.com/kubernetes-release/release/v1.18.2 \
  --admin-access 35.224.122.237/32 \
  --cloud aws \
  --container-runtime=containerd \
  --networking=calico \
  --image=099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200423 \
  --override cluster.spec.nodePortAccess=0.0.0.0/0
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
1.18.0-alpha.4+de7cc70c3

**3. What cloud provider are you using?**
aws

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
n/a

**5. What happened after the commands executed?**
https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-misc-containerd/1258780879564574720/artifacts/ip-172-20-48-115.ap-southeast-1.compute.internal/containerd.log
```
time="2020-05-08T15:33:38.838738725Z" level=info msg="PullImage \"kope/kube-apiserver-healthcheck:1.18.0-alpha.3\""
time="2020-05-08T15:33:43.206611091Z" level=error msg="PullImage \"kope/kube-apiserver-healthcheck:1.18.0-alpha.3\" failed" error="failed to pull and unpack image \"docker.io/kope/kube-apiserver-healthcheck:1.18.0-alpha.3\": failed to resolve reference \"docker.io/kope/kube-apiserver-healthcheck:1.18.0-alpha.3\": pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed"
```
**6. What did you expect to happen?**
Repo should be public.

**9. Anything else do we need to know?**
There are 2 issues here:
1. the `kope/kube-apiserver-healthcheck` should be public
[...]

PR #9101, opened 2020-05-08
Fix containerd image side-loading

hakman:
Imported container images have to be put into the `k8s.io` namespace to be accessible to Kubernetes. Initially I thought side-loading was just for Protokube and could be kept in a different namespace, but the mechanism is used for more than this.

In short, beginner level mistake. This fix mostly reverts an old commit of mine from https://github.com/kubernetes/kops/pull/8217.

Fixes https://github.com/kubernetes/kops/issues/9099


hakman:
Thanks @rifelpet

PR #9102, opened 2020-05-08
Update base images for TravisCI

hakman:
Travis uses pretty old distro images by default. An update may be useful.
https://docs.travis-ci.com/user/reference/linux#default
https://docs.travis-ci.com/user/reference/osx#macos-version


hakman:
Seems pretty much the same as with the default values.

PR #9107, opened 2020-05-10
Move CNI docs to their own files

olemarkus:
The networking docs were getting a bit unruly.

Moved them to their own files and made them a little bit homogenic. Also removed the calico v3 upgrade docs that describes the process from before etcd-manager.


johngmyers:
`upgrade_from_kops_1.6_to_1.7_calico_cidr_migration.md` is linked to from `docs\releases\1.7-NOTES.md`, which is why I submitted #8372.

olemarkus:
Yeah, I know I broke a link there. I don't think keeping the old release notes brings any value the way we maintain our docs. They offer no value now, or even worse, they can be very misleading. I had to actually create a calico cluster with an old version of kops and upgrade to see if those docs made sense at all.

docs/networking/lyft-vpc.md
```diff
@@ -0,0 +1,47 @@
...
+You can specify which subnets to use for allocating Pod IPs by specifying
+
+```
+  networking:
+    lyftvpc:
+      subnetTags:
+        KubernetesCluster: myclustername.mydns.io
+```
+
+In this example, new interfaces will be attached to subnets tagged with `kubernetes_kubelet = true`.
```
hakman, line 43: Leftover from https://github.com/kubernetes/kops/pull/8571
```suggestion
In this example, new interfaces will be attached to subnets tagged with `KubernetesCluster = myclustername.mydns.io`.
```

rifelpet:
this looks great, thanks! We'll have to make a decision on what to do about broken links in docs specific to kops versions like release notes. k/k seems to keep around all of their old changelogs including broken links (for example the Documentation link [here](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.2.md#v127)) if that is of any influence.

/approve
/lgtm

PR #9113, opened 2020-05-11
upgrade to use cinder v3 api

zetaab:


olemarkus:
/hold 
so you can merge at your convenience
/lgtm

zetaab:
k/k openstack csi provider has been using volume v3 api already 3 years. So I think everyone has enough new OpenStack that supports volume v3 api. 

/hold remove

PR #9115, opened 2020-05-11
Add Ambassador addon to kops

concaf:
This commit adds Ambassador (https://getambassador.io/) addon to
Kops.

Ambassador is installed via Ambassador Operator which is the
recommended way of installed Ambassador - it makes sure users
always have the latest version of Ambassador installed and takes
care of the update schedule as well.

CC: @brucehorn @inercia

### DIFF
--- a/addons/ambassador/ambassador-operator.yaml
+++ b/addons/ambassador/ambassador-operator.yaml
@@ -1,0 +1,445 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ambassador
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: ambassadorinstallations.getambassador.io
+spec:
+  additionalPrinterColumns:
+    - JSONPath: .spec.version
+      name: VERSION
+      type: string
+    - JSONPath: .spec.updateWindow
+      name: UPDATE-WINDOW
+      type: integer
+    - JSONPath: .status.lastCheckTime
+      description: Last time checked
+      name: LAST-CHECK
+      type: string
+    - JSONPath: .status.conditions[?(@.type=='Deployed')].status
+      description: Indicates if deployment has completed
+      name: DEPLOYED
+      type: string
+    - JSONPath: .status.conditions[?(@.type=='Deployed')].reason
+      description: Reason for deployment completed
+      name: REASON
+      priority: 1
+      type: string
+    - JSONPath: .status.conditions[?(@.type=='Deployed')].message
+      description: Message for deployment completed
+      name: MESSAGE
+      priority: 1
+      type: string
+    - JSONPath: .status.deployedRelease.appVersion
+      description: Deployed version of Ambassador
+      name: DEPLOYED-VERSION
+      type: string
+    - JSONPath: .status.deployedRelease.flavor
+      description: Deployed flavor of Ambassador (OSS or AES)
+      name: DEPLOYED-FLAVOR
+      type: string
+  group: getambassador.io
+  names:
+    kind: AmbassadorInstallation
+    listKind: AmbassadorInstallationList
+    plural: ambassadorinstallations
+    singular: ambassadorinstallation
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: AmbassadorInstallation is the Schema for the ambassadorinstallations
+        API
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: AmbassadorInstallationSpec defines the desired state of AmbassadorInstallation
+          properties:
+            baseImage:
+              description: An (optional) image to use instead of the image specified
+                in the Helm chart.
+              type: string
+            helmRepo:
+              description: An (optional) Helm repository.
+              type: string
+            installOSS:
+              description: 'Installs [Ambassador OSS](https://www.getambassador.io/docs/latest/topics/install/install-ambassador-oss/)
+                instead of [AES](https://www.getambassador.io/docs/latest/topics/install/).
+                Default is false which means it installs AES by default. TODO: 1.
+                AES/AOSS is not installed and the user installs using `installOSS:
+                true`, then we straightaway install AOSS. 2. AOSS is installed via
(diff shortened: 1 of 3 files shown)


olemarkus:
Hi,

Thanks for contributing this. I am curious about how you see the value of installing a single ambassador instance as an addon versus using helm or other install methods.
I support any addon that needs specific configuration of kops, and adds validation logic for that (take nodelocaldns or any CNI for example). But this addon just install basic resources without any configuration options or other forms of  kops-specific logic.

addons/ambassador/ambassador-operator.yaml
```diff
@@ -0,0 +1,361 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ambassador-operator-cluster
+rules:
+  - apiGroups: ['*']
+    resources: ['*']
+    verbs: ['*']
+  - nonResourceURLs: ['*']
+    verbs: ['*']
```
johngmyers, line 276: This is a bit overly permissive. Could it be tightened up?

brucehorn:
Ole, thanks for the comment.  We had a good discussion about this at the office hours last week.  We understand the Helm approach (we use Helm ourselves), but we think that addons have some advantages over just using Helm.  When looking for ways to provide new functionality such as ingress to kops, web search lands you right in the project at one of the addon directories, which is good--we want people to land in a useful place.  Addons provide a good UX for the end users, with a nice integration process.  One of the things that we discussed at the office hours was the fact that some of the addons seem not to have been maintained recently--we addressed that by committing to keeping our Datawire addons up to date, and there will be an effort to recruit maintainers for the existing addons.  Finally we committed to follow whatever new directions the kops team takes us to provide extension functionality in the future.

addons/ambassador/ambassador-operator.yaml
concaf: We could tighten it based on the current resources that the operator installs however the main point of concern is that since this operator uses Helm to install the later Ambassador version and keeps it up to date over time, these permissions will go stale and cause unexpected errors for our users - so we've kept the permissions relaxed intentionally.
There are other operators (specifically which perform upgrades) in the wild which do the same - one example would be https://github.com/fluxcd/helm-operator/blob/157d422bc11f127b1d9a5eba18fc8882c4b9cde9/deploy/rbac.yaml#L10-L21. WDYT?

concaf:
@johngmyers @olemarkus gentle nudge ;) Is there anything I can do to take this forward?

addons/ambassador/ambassador-operator.yaml
```diff
@@ -0,0 +1,361 @@
...
+data:
+  values.yaml: |+
+    deploymentTool: amb-oper-manifest
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ambassador-operator
+  labels:
+    getambassador.io/installer: operator
```
mikesplain, line 326: Namespace? We generally prefer explicit when possible.

addons/ambassador/ambassador-operator.yaml
```diff
@@ -0,0 +1,361 @@
...
+
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ambassador-operator-cluster
+subjects:
+  - kind: ServiceAccount
+    name: ambassador-operator
+    namespace: default
```
mikesplain, line 307: Here you mention default namespace. Is this where you want the deployment below installed?

addons/ambassador/ambassador-operator.yaml
mikesplain: I agree with John, this is pretty permissive. That said, I get that an operator can require widely scoped permissions.  

I guess my biggest concern is to whom these permissions are granted.  Currently it's a service account in the default namespace.  Personally, I consider default a dangerous place for an SA that has global permissions.  Maybe moving it to kube-system or another namespace we know exists would reduce some risk here as well?  I know we don't want to avoid deviating from your official docs but just throwing out a few ideas.

At minimum, maybe we could document the permissiveness in the readme?

johngmyers: I've been mulling my position on this. I'm having a hard time supporting such carte blanche permissions. They're not even scoped to the namespace that Ambassador would be installed in.

It seems not unreasonable to me that an addition to the functionality of a subsystem, requiring it to have new types of resources, would also require an upgrade to the operator that manages it. I'm presuming the operator provides some functionality to the subsystem beyond just calling Helm and this would be likely to need to be upgraded as well.

concaf: @mikesplain @johngmyers I've made a couple of changes as suggested. I've moved operator and Ambassador's installation to `kube-system` and documented scope of the permissions here - https://github.com/kubernetes/kops/pull/9115/files#diff-b45153696dcf1520ef38065366461babR16-R19. PTAL :+1:

addons/ambassador/ambassador-operator.yaml
concaf: I've added `namespace: kube-system` at all places.

addons/ambassador/ambassador-operator.yaml
concaf: Yep, but now I have this to `kube-system`

concaf:
Hey @rdrgmnzs @johngmyers @olemarkus @mikesplain, gentle nudge ;) Anything I can do to take this forward?

olemarkus:
I am a bit sceptical to installing this to kube-system. default seems to be the better place even though it is not ideal either.

concaf:
@olemarkus I changed the namespace to `kube-system` per suggestions by @mikesplain, see - https://github.com/kubernetes/kops/pull/9115#discussion_r427677585

> Maybe moving it to kube-system or another namespace we know exists would reduce some risk here as well? 

Would you rather have it in `default`, or should I create another namespace called `ambassador` and install it there?

olemarkus:
/assign @mikesplain 
I think Mike is better able to answer this than me.

mikesplain:
First of all, thanks for sticking with us on this one @concaf.  

@olemarkus brings up a good point.  Historically pretty much everything Kops managed has lived in kube-system but that doesn't meant it's a good practice.  That said, it looks like at least ingress-nginx [utilizes it's own namespace](https://github.com/kubernetes/kops/blob/master/addons/ingress-nginx/v1.6.0.yaml) so we should probably do that instead.  I apologize for the advice, I wasn't sure what the right answer was at the time and was trying to pose it as suggestion.  

Can we move it to an ambassador namespace? Either `ambassador` or `ambassador-operator`?  I assume the first makes the most sense but I leave that up to you.  Baring that, consider this LGTMed and ready for approval on my end.  Feel free to ping me when that change is done and I'll approve asap!

concaf:
@mikesplain yeah, makes sense. Default installation instructions for Ambassador deploy it to the `ambassador` namespace as well.
I've pushed a commit that make this change. PTAL, thanks! :)

concaf:
@mikesplain gentle nudge ;)

mikesplain:
Ahh yes thanks @concaf! This looks good, do you mind squashing into a single commit? I'll approve ASAP once that's done. Thank you so much!

concaf:
@mikesplain sure thing. I've rebased against master and folded everything in one commit. PTAL :) Thanks for sticking and helping with this.

mikesplain:
Thanks @concaf for the contribution! 

/lgtm
/approve

PR #9116, opened 2020-05-11
[Digital Ocean] Handle logic for kops edit/update cluster

srikiz:
- Added logic for FindCluster Status

- Also fixed load balancer logic to only create LB if it doesn't exist.

- Tested with kops edit cluster by adding new Instance group and was able to see the cluster updated as expected.

### DIFF
--- a/pkg/resources/digitalocean/cloud.go
+++ b/pkg/resources/digitalocean/cloud.go
@@ -203,7 +203,7 @@ func findEtcdStatus(c *Cloud, cluster *kops.Cluster) ([]kops.EtcdClusterStatus, 
 	volumes, err := getAllVolumesByRegion(c, c.RegionName)
 
 	if err != nil {
-		return nil, fmt.Errorf("error describing volumes: %v", err)
+		return nil, fmt.Errorf("failed to get all volumes by region from %s: %v", c.RegionName, err)
 	}
 
 	for _, volume := range volumes {
@@ -213,7 +213,7 @@ func findEtcdStatus(c *Cloud, cluster *kops.Cluster) ([]kops.EtcdClusterStatus, 
 		var etcdClusterSpec *etcd.EtcdClusterSpec
 
 		for _, myTag := range volume.Tags {
-			klog.V(2).Infof("findEtcdStatus status (from cloud): %v", myTag)
+			klog.V(8).Infof("findEtcdStatus status (from cloud): checking if volume with tag %q belongs to cluster", myTag)
 			// check if volume belongs to this cluster.
 			// tag will be in the format "KubernetesCluster:dev5-k8s-local" (where clusterName is dev5.k8s.local)
 			clusterName := strings.Replace(cluster.Name, ".", "-", -1)
@@ -223,16 +223,17 @@ func findEtcdStatus(c *Cloud, cluster *kops.Cluster) ([]kops.EtcdClusterStatus, 
 				// loop through the tags again and
 				for _, volumeTag := range volume.Tags {
 					if strings.Contains(volumeTag, TagKubernetesClusterIndex) {
-						if len(strings.Split(volumeTag, ":")) < 2 {
-							return nil, fmt.Errorf("error splitting the volume tag %q on volume %q", volumeTag, volume)
+						volumeTagParts := strings.Split(volumeTag, ":")
+						if len(volumeTagParts) < 2 {
+							return nil, fmt.Errorf("volume tag split failed, too few components for tag %q on volume %q", volumeTag, volume)
 						}
-						dropletIndex := strings.Split(volumeTag, ":")[1]
+						dropletIndex := volumeTagParts[1]
 						etcdClusterSpec, err = c.getEtcdClusterSpec(volume.Name, dropletIndex)
-						klog.V(10).Infof("findEtcdStatus etcdClusterSpec: %v", fi.DebugAsJsonString(etcdClusterSpec))
 						if err != nil {
 							return nil, fmt.Errorf("error parsing etcd cluster tag %q on volume %q: %v", volumeTag, volumeID, err)
 						}
 
+						klog.V(10).Infof("findEtcdStatus etcdClusterSpec: %v", fi.DebugAsJsonString(etcdClusterSpec))
 						etcdClusterName = etcdClusterSpec.ClusterKey
 						status := statusMap[etcdClusterName]
 						if status == nil {
@@ -253,7 +254,7 @@ func findEtcdStatus(c *Cloud, cluster *kops.Cluster) ([]kops.EtcdClusterStatus, 
 		}
 	}
 
-	var status []kops.EtcdClusterStatus
+	status := make([]kops.EtcdClusterStatus, 0, len(statusMap))
 	for _, v := range statusMap {
 		status = append(status, *v)
 	}
--- a/pkg/resources/digitalocean/cloud.go
+++ b/pkg/resources/digitalocean/cloud.go
@@ -200,8 +200,6 @@ func (c *Cloud) FindClusterStatus(cluster *kops.Cluster) (*kops.ClusterStatus, e
 // findEtcdStatus discovers the status of etcd, by looking for the tagged etcd volumes
 func findEtcdStatus(c *Cloud, cluster *kops.Cluster) ([]kops.EtcdClusterStatus, error) {
 	statusMap := make(map[string]*kops.EtcdClusterStatus)
-	klog.V(2).Infof("Querying DO for etcd volumes")
-
 	volumes, err := getAllVolumesByRegion(c, c.RegionName)
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/resources/digitalocean/cloud.go
```diff
@@ -175,3 +183,97 @@ func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressSta
...
+	return status, nil
+}
+
+// findEtcdStatus discovers the status of etcd, by looking for the tagged etcd volumes
+func findEtcdStatus(c *Cloud, cluster *kops.Cluster) ([]kops.EtcdClusterStatus, error) {
+	statusMap := make(map[string]*kops.EtcdClusterStatus)
+	volumes, err := getAllVolumesByRegion(c, c.RegionName)
+
+	if err != nil {
+		return nil, fmt.Errorf("error describing volumes: %v", err)
```
timoreimann, line 206: Maybe use a error message closer to the function name we called and include the region name, e.g.:

```go
return nil, fmt.Errorf("failed to get all volumes by region from %s: %v", c.RegionName, err)
```

pkg/resources/digitalocean/cloud.go
```diff
@@ -175,3 +183,97 @@ func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressSta
...
+	}
+
+	for _, volume := range volumes {
+		volumeID := volume.ID
+
+		etcdClusterName := ""
+		var etcdClusterSpec *etcd.EtcdClusterSpec
+
+		for _, myTag := range volume.Tags {
+			klog.V(2).Infof("findEtcdStatus status (from cloud): %v", myTag)
```
timoreimann, line 216: Perhaps the message should be more about the tag, for instance:

_"findEtcdStatus status (from cloud): checking if volume with tag %q belongs to cluster", myTag"_

timoreimann, line 216: Should this be at a higher log level to decrease verbosity, especially because other log statements in this function is a verbosity of 10?

pkg/resources/digitalocean/cloud.go
```diff
@@ -175,3 +183,97 @@ func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressSta
...
+			// check if volume belongs to this cluster.
+			// tag will be in the format "KubernetesCluster:dev5-k8s-local" (where clusterName is dev5.k8s.local)
+			clusterName := strings.Replace(cluster.Name, ".", "-", -1)
+			if strings.Contains(myTag, fmt.Sprintf("%s, %s", TagKubernetesClusterNamePrefix, ":", clusterName)) {
```
timoreimann: Am I misreading this one, or does the call to `fmt.Sprintf` use 3 variables but only 2 placeholders?

pkg/resources/digitalocean/cloud.go
```diff
@@ -175,3 +183,97 @@ func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressSta
...
+				klog.V(10).Infof("findEtcdStatus cluster comparison matched for tag: %v", myTag)
+				// this volume belongs to our cluster, add this to our etcdClusterSpec.
+				// loop through the tags again and
+				for _, volumeTag := range volume.Tags {
+					if strings.Contains(volumeTag, TagKubernetesClusterIndex) {
+						if len(strings.Split(volumeTag, ":")) < 2 {
+							return nil, fmt.Errorf("error splitting the volume tag %q on volume %q", volumeTag, volume)
```
timoreimann, line 227: Can we make the message more specific about the fact that we have too few components in the tag?

pkg/resources/digitalocean/cloud.go
```diff
@@ -175,3 +183,97 @@ func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressSta
...
+			if strings.Contains(myTag, fmt.Sprintf("%s:%s", TagKubernetesClusterNamePrefix, clusterName)) {
+				klog.V(10).Infof("findEtcdStatus cluster comparison matched for tag: %v", myTag)
+				// this volume belongs to our cluster, add this to our etcdClusterSpec.
+				// loop through the tags again and
+				for _, volumeTag := range volume.Tags {
+					if strings.Contains(volumeTag, TagKubernetesClusterIndex) {
+						if len(strings.Split(volumeTag, ":")) < 2 {
+							return nil, fmt.Errorf("error splitting the volume tag %q on volume %q", volumeTag, volume)
+						}
+						dropletIndex := strings.Split(volumeTag, ":")[1]
```
timoreimann, line 229: nit: if we call `strings.Split()` first and check the length on the outcome, we only need a single call to `strings.Split()`. That is,

```go
volumeTagParts := strings.Split(volumeTag, ":")
if len(volumeTagParts) < 2 {
  <error handling>
}
dropletIndex := volumeTagParts[1]
```

pkg/resources/digitalocean/cloud.go
```diff
@@ -175,3 +183,97 @@ func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressSta
...
+		etcdClusterName := ""
+		var etcdClusterSpec *etcd.EtcdClusterSpec
+
+		for _, myTag := range volume.Tags {
+			klog.V(2).Infof("findEtcdStatus status (from cloud): %v", myTag)
+			// check if volume belongs to this cluster.
+			// tag will be in the format "KubernetesCluster:dev5-k8s-local" (where clusterName is dev5.k8s.local)
+			clusterName := strings.Replace(cluster.Name, ".", "-", -1)
+			if strings.Contains(myTag, fmt.Sprintf("%s:%s", TagKubernetesClusterNamePrefix, clusterName)) {
+				klog.V(10).Infof("findEtcdStatus cluster comparison matched for tag: %v", myTag)
```
timoreimann, line 221: Wondering if 10 could be too high. AFAIK, level 8 is where Kubernetes tends to log very specific details at the request level. I'm not too familiar with how kops handles log levels in general, so not a super strong opinion here.

pkg/resources/digitalocean/cloud.go
```diff
@@ -175,3 +183,97 @@ func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressSta
...
+				// loop through the tags again and
+				for _, volumeTag := range volume.Tags {
+					if strings.Contains(volumeTag, TagKubernetesClusterIndex) {
+						if len(strings.Split(volumeTag, ":")) < 2 {
+							return nil, fmt.Errorf("error splitting the volume tag %q on volume %q", volumeTag, volume)
+						}
+						dropletIndex := strings.Split(volumeTag, ":")[1]
+						etcdClusterSpec, err = c.getEtcdClusterSpec(volume.Name, dropletIndex)
+						klog.V(10).Infof("findEtcdStatus etcdClusterSpec: %v", fi.DebugAsJsonString(etcdClusterSpec))
+						if err != nil {
```
timoreimann, line 232: The error should be handled before we log the spec in the line right above.

pkg/resources/digitalocean/cloud.go
```diff
@@ -175,3 +183,97 @@ func (c *Cloud) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressSta
...
+							Name:     memberName,
+							VolumeId: volume.ID,
+						})
+					}
+				}
+			}
+		}
+	}
+
+	var status []kops.EtcdClusterStatus
```
timoreimann, line 256: This can be improved slightly performance-wise since we know the target size of the slice:

```go
status := make([]kops.EtcdClusterStatus, 0, len(statusMap))
```

pkg/resources/digitalocean/cloud.go
srikiz: sure, updated.

pkg/resources/digitalocean/cloud.go
srikiz: updated.

srikiz: Sure, changed to verbosity of 10.

pkg/resources/digitalocean/cloud.go
srikiz: That was a miss - I updated it as part of the future commit.

pkg/resources/digitalocean/cloud.go
srikiz: sure, updated.

pkg/resources/digitalocean/cloud.go
srikiz: Good catch - thanks, updated.

pkg/resources/digitalocean/cloud.go
srikiz: I will keep it with 10 for now. There are lot places I had seen 10 before, and that's how I continued to use that. I'll keep a note to change this to 8 for all DO work as a separate PR at a later point.

pkg/resources/digitalocean/cloud.go
srikiz: Thanks, updated.

pkg/resources/digitalocean/cloud.go
srikiz: sure, updated.

rdrgmnzs:
Thanks @srikiz!

/lgtm
/approve

PR #9121, opened 2020-05-12
Add EC2 Instance LifeCycle label

### LINKED ISSUE #8923: Mixed instance types and labels
One issue we've had with using mixed instance policies has been that there doesn't seem to be a great way to tell if the node is spot or not. What would you all say about checking the instance metadata in nodeup and setting a label for it? I know we removed the automatic label setting stuff back in 1.16 because it moved to the kubelet, but this would be a different kind of thing?

atmosx:
When using a "mixed instance policy"[1] instance group spot and onDemand nodes are part of the same ASG. The ASG handles the percentage of spot vs onDemand
instances. There are no annotations, EC2 tags or labels to identify which
instances are onDemand vs spot.

This PR introduces a new label to be attached on all AWS instances.

The label is:

```
node-role.kubernetes.io/spot-worker: <value>
```

Possible values are:

* onDemand (default)
* spot

[1]: https://github.com/kubernetes/kops/blob/master/docs/instance_groups.md#mixedinstancepolicy-aws-only

### DIFF
--- a/cmd/kops-controller/controllers/node_controller.go
+++ b/cmd/kops-controller/controllers/node_controller.go
@@ -119,6 +119,15 @@ func (r *NodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, fmt.Errorf("unable to build config for node %s: %v", node.Name, err)
 	}
 
+	lifecycle, err := r.getInstanceLifecycle(ctx, node)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to get instance lifecycle %s: %v", node.Name, err)
+	}
+
+	if len(lifecycle) > 0 {
+		labels[fmt.Sprintf("node-role.kubernetes.io/%s-worker", lifecycle)] = "true"
+	}
+
 	updateLabels := make(map[string]string)
 	for k, v := range labels {
 		actual, found := node.Labels[k]
@@ -186,6 +195,17 @@ func (r *NodeReconciler) getClusterForNode(node *corev1.Node) (*kops.Cluster, er
 		return nil, err
 	}
 	return cluster, nil
+}
+
+// getInstanceLifecycle returns InstanceLifecycle string object
+func (r *NodeReconciler) getInstanceLifecycle(ctx context.Context, node *corev1.Node) (string, error) {
+
+	identity, err := r.identifier.IdentifyNode(ctx, node)
+	if err != nil {
+		return "", fmt.Errorf("error identifying node %q: %v", node.Name, err)
+	}
+
+	return identity.InstanceLifecycle, nil
 }
 
 // getInstanceGroupForNode returns the kops.InstanceGroup object for the node
--- a/pkg/nodeidentity/aws/identify.go
+++ b/pkg/nodeidentity/aws/identify.go
@@ -102,6 +102,11 @@ func (i *nodeIdentifier) IdentifyNode(ctx context.Context, node *corev1.Node) (*
 		return nil, fmt.Errorf("found instance %q, but state is %q", instanceID, instanceState)
 	}
 
+	lifecycle := ""
+	if instance.InstanceLifecycle != nil {
+		lifecycle = *instance.InstanceLifecycle
+	}
+
 	// TODO: Should we traverse to the ASG to confirm the tags there?
 	igName := getTag(instance.Tags, CloudTagInstanceGroupName)
 	if igName == "" {
@@ -110,6 +115,7 @@ func (i *nodeIdentifier) IdentifyNode(ctx context.Context, node *corev1.Node) (*
 
 	info := &nodeidentity.Info{}
 	info.InstanceGroup = igName
+	info.InstanceLifecycle = lifecycle
 
 	return info, nil
 }
--- a/pkg/nodeidentity/interfaces.go
+++ b/pkg/nodeidentity/interfaces.go
@@ -27,5 +27,6 @@ type Identifier interface {
 }
 
 type Info struct {
-	InstanceGroup string
+	InstanceGroup     string
+	InstanceLifecycle string
 }


atmosx:
Some tests fail from what I see. I'll fix them ASAP. Unfortunately I didn't came across the links shared by the bot while reading the contributing guide.

johngmyers:
The presence of the `node-role.kubernetes.io/spot-worker` annotation indicates the node is spot. The annotation should not be placed on onDemand instances.

johngmyers:
@atmosx You need to edit the message on the first commit to remove the issue reference.

atmosx:
The PR and logic are fairly small. I squashed the commits for easier review.

hakman:
@atmosx this is unexpectedly small. Very nice! :)

pkg/nodeidentity/interfaces.go
```diff
@@ -27,5 +27,6 @@ type Identifier interface {
 }
 
 type Info struct {
-	InstanceGroup string
+	InstanceGroup     string
+	InstanceLifecycle string
```
hakman, line 31: In this case, a string it's not really needed to check the value. I think it is important only if it is a spot instance or not? 
```suggestion
	InstanceSpot bool
```

atmosx: > In this case, a string it's not really needed to check the value. I think it is important only if it is a spot instance or not?

True! The string type made sense in the first approach. I'll update the PR when I find time.

atmosx:
We’re adding a [a new field](https://github.com/kubernetes/kops/pull/9121/files#diff-358c9355d1fc10a80d4713f505e64f14R31) `InstanceSpot` to the `Info` struct. The field is `bool`, hence defaults to `false`. We don't define the `InstanceSpot` field in `do` and `gce` [packages](https://github.com/kubernetes/kops/tree/master/pkg/nodeidentity).

I wonder if that's a problem or not.  :confused:

pkg/nodeidentity/aws/identify.go
```diff
@@ -101,6 +102,9 @@ func (i *nodeIdentifier) IdentifyNode(ctx context.Context, node *corev1.Node) (*
 	if instanceState != ec2.InstanceStateNameRunning {
 		return nil, fmt.Errorf("found instance %q, but state is %q", instanceID, instanceState)
 	}
+	if instance.InstanceLifecycle != nil {
+		spot = true
+	}
```
hakman, line 107: Seems more like the previous blocks this way.
```suggestion

	spot := false
	if instance.InstanceLifecycle != nil {
		spot = true
	}
```

hakman:
@atmosx I think doing this for AWS is a good start. Not sure if the label is even appropriate for GCE and DO.

cmd/kops-controller/controllers/node_controller.go
```diff
@@ -119,6 +119,15 @@ func (r *NodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, fmt.Errorf("unable to build config for node %s: %v", node.Name, err)
 	}
 
+	spot, err := r.getInstanceLifecycle(ctx, node)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to get instance lifecycle %s: %v", node.Name, err)
+	}
+
+	if spot {
+		labels["node-role.kubernetes.io/spot-worker"] = "spot"
```
hakman, line 128: The documented value seems to be "true":
https://github.com/pusher/k8s-spot-termination-handler/blob/master/README.md#requirements
```suggestion
		labels["node-role.kubernetes.io/spot-worker"] = "true"
```

pkg/nodeidentity/aws/identify.go
```diff
@@ -102,6 +102,11 @@ func (i *nodeIdentifier) IdentifyNode(ctx context.Context, node *corev1.Node) (*
 		return nil, fmt.Errorf("found instance %q, but state is %q", instanceID, instanceState)
 	}
 
+	spot := false
+	if instance.InstanceLifecycle != nil {
```
johngmyers, line 106: Shouldn't we also check that `*instance.InstanceLifecycle == "spot"`?

atmosx: > Shouldn't we also check that `*instance.InstanceLifecycle == "spot"`?

I don't think so. The field `InstanceLifecycle` is available only in spot instances, onDemand instances do not have this field.

johngmyers: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Instance.html states that this field can have two values: `spot` and `scheduled`

atmosx: > https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Instance.html states that this field can have two values: `spot` and `scheduled`

I missed that, thanks for bringing this to my attention! 

Given that there are two values there, I believe makes sense to revert to `string` type and pass the type of instance as an argument.

Instead of `node-role.kubernetes.io/spot-worker` maybe we should go with `node-role.kubernetes.io/instance-lifecycle: <spot|scheduled>`.

What do you think?

cc @hakman

johngmyers: The `node-role.kubernetes.io/spot-worker` label is exposed through kubectl:

```
$ kubectl get nodes
NAME                            STATUS   ROLES              AGE    VERSION
ip-172-20-33-222.ec2.internal   Ready    master             147m   v1.17.5
ip-172-20-36-39.ec2.internal    Ready    node,spot-worker   125m   v1.17.5
ip-172-20-47-204.ec2.internal   Ready    node,spot-worker   123m   v1.17.5
```
Having that report `node,instance-lifecycle` would be less useful.

johngmyers: I think a string type `InstanceLifecycle` would work well. Then we could also add a `node-role.kubernetes.io/scheduled` annotation.

cmd/kops-controller/controllers/node_controller.go
```diff
@@ -119,6 +119,15 @@ func (r *NodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, fmt.Errorf("unable to build config for node %s: %v", node.Name, err)
 	}
 
+	spot, err := r.getInstanceLifecycle(ctx, node)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to get instance lifecycle %s: %v", node.Name, err)
+	}
+
+	if spot {
+		labels["node-role.kubernetes.io/spot-worker"] = "true"
```
johngmyers, line 128:
```suggestion
		labels[fmt.Sprintf("node-role.kubernetes.io/%s-worker", lifecycle)] = "true"
```

pkg/nodeidentity/aws/identify.go
hakman: I think I saw mentions of `node-role.kubernetes.io/worker` label for the worker nodes, not "scheduled-worker". I did not mention this because I don't see much value in having an extra label for regular worker nodes. I don't mind if it exists though.

https://github.com/pusher/k8s-spot-rescheduler/blob/master/README.md#requirements

johngmyers: "scheduled-worker" would be for scheduled instances, not onDemand.

atmosx: I updated the implementation according to @johngmyers suggestions. AWS has three types of instances:

1. OnDemand: no `InstanceLifecycle` field, no label attached
2. Spot: `InstanceLifecycle` field is set to `spot`, label attached is `node-role.kubernetes.io/spot-worker: true`
3. Scheduled: `InstanceLifecycle` field set to `scheduled` label attached is `node-role.kubernetes.io/scheduled-worker: true`

The current approach IMO is complete in comparison to previous ones as it covers all cases.

hakman: Never knew about the scheduled ones. Thanks :).

hakman:
I think this may be good also for 1.17 as it's a fairly small change.
/lgtm

rifelpet:
We don't have any e2e tests that use spot instances so we wont be able to test this change in our periodic jobs beyond checking the node objects do not contain the new label since they are all on-demand instances.

That said, this looks good to me. thanks!

/approve

atmosx:
> I think this may be good also for 1.17 as it's a fairly small change.
> /lgtm

I would like that as we're planning to install v1.17 and we can use this feature.

atmosx:
resolves #8923

PR #9124, opened 2020-05-13
Don't put bastions in the utility subnets

johngmyers:
There's no reason to put bastions in utility subnets as their securitygroup only allows incoming traffic from the load balancer.


rifelpet:
This seems reasonable to me. Should we add a release note that mentions the change in behavior for new clusters?

/approve
/hold for release note if we decide to add it

johngmyers:
I assessed it as being below the threshold for a relnote. Do others think this needs to be called out?

olemarkus:
Since this is only for new clusters, I don't think this is very important to call out. Those who manage many clusters hopefully use templating and will also not be affected by this.

We should probably update https://kops.sigs.k8s.io/bastion/ too though.

rifelpet:
I'm fine with not having a release note. feel free to cancel the hold when you're ready.

PR #9126, opened 2020-05-13
Fail cluster validation if too few nodes for ig's target size

johngmyers:
If the ASG has scaled above the min, rolling updates can start draining the next node before the ASG replaces a terminated one.


rdrgmnzs:
Thanks for the fix @johngmyers

/lgtm
/approve

Issue #9110, opened 2020-05-10
Certificate issuance cleanup and hardening


johngmyers:
I'd like to clean up and consolidate our certificate issuance code. It would be nice to harden things in the process. I'd like some input on the requirements.

Currently, most certificates and private keys are created out of `Keystore.CreateKeypair()`, called from `upup/pkg/fi/fitasks/Keypair`. All of these pairs are per-cluster, stored in the keystore in the VFS. They're also all issued with 10 year lifetimes.

The exceptions are created at nodeup time and are one per node:

* The master kubelet client certificate is created by `KubeletBuilder.buildMsterKubeletKubeconfig()`. It also has a 10 year lifetime.

* The kube-apiserver-healthcheck client cert is created by `KubeAPIServerBuilder.buildClientKeypair()`. It also has a 10 year lifetime.

* The kube-apiserver client cert for authenticating to etcd is created by `EtcdManagerTLSBuilder.buildKubeAPIServerKeypair()`. It has a 1 year lifetime.

* The Cilium client cert for agents authenticating to the cilium etcd is created by `CiliumBuilder.buildCiliumEtcdSecrets`. It has a 1 year lifetime.

My first question is about the discrepancy between 1 year and 10 year lifetimes. Is there a requirement to support masters or node instances with lifetimes greater than one year? If not, then I'd like to reduce the lifetimes of all non-CA certificates to 1 year, to reduce the number of places that 10-year keys are left around. If so, then those existing 1 year lifetimes are a problem.

Perhaps we'd want a field in the cluster spec for the lifetime of non-CA certificates, so sites that update their nodes every 30 days or so could drop the lifetime down to something like the 60-day range?

I think for cleanup I'd create a nodeup task for creating a keypair and transition all the non-CA keypairs (or at least all master non-CA keypairs) from `fitasks.Keypair` to it. An alternative would be to move all those that happen after apiserver and kops-controller are up to using a `certificatesigningrequest` API.

@justinsb what are your thoughts?

johngmyers:
Progress:

* The certificates that are specific to the master, excluding ones for the legacy etcd manager and node-authorizer, have been migrated to being issued by nodeup.
* I do not plan to migrate the certificates for the legacy etcd manager or node-authorizer. I expect those features to deprecate out.
* Migrating the the "kubecfg" admin cert to be issued at export time is #9593.
* Migrating the kubelet certificate is #9708.
* The remaining certificates used by worker nodes will follow #9470.

[...]

PR #9130, opened 2020-05-15
Refactor cert issuance code

johngmyers:
See #9110

Pull cert issuance code up into `Keypair` then push it down into `pki` where it can be shared by nodeup tasks.

Add up to 30 days skew to the lifetimes of nodeup-generated certs, to try to cause nodes generated around the same time to expire on different days.

Decrease the lifetime of the kube-apiserver-healthcheck and master kubelet certificates from 10 years to 455 days plus skew. Increases the lifetime of the etcd-clients-ca-cilium and apiserver-to-etcd client certs from 1 year to 455 days plus skew.

Make `pki.SignNewCertificate()` private.


olemarkus:
This looks pretty good to me.
/lgtm

nodeup/pkg/model/kubelet.go
```diff
@@ -597,54 +594,15 @@ func (b *KubeletBuilder) buildMasterKubeletKubeconfig() (*nodetasks.File, error)
...
-	}
-
-	template := &x509.Certificate{
-		BasicConstraintsValid: true,
-		IsCA:                  false,
+	req := &pki.IssueCertRequest{
+		Name:         "master-kubelet",
+		Signer:       fi.CertificateId_CA,
+		Type:         "client",
+		Subject:      fmt.Sprintf("cn=system:node:%s,o=%s", nodeName, rbac.NodesGroup),
```
johngmyers, line 601: This sort of thing is making me reconsider specifying the Subject as a string. One hopes there's no node name or nodeGroup with a comma in it. I'm considering changing it so that Subject has to be specified as a `pkix.Name` by all callers.

rdrgmnzs:
Code looks good to me, I'd like @justinsb to take a peek at it just to ensure there's nothing I'm unaware of that may be affected by decreasing the lifetime of the masters kubelet cert.

/lgtm
/assign @justinsb

pkg/pki/issue.go
```diff
@@ -0,0 +1,169 @@
...
+type Keystore interface {
+	// FindKeypair finds a cert & private key, returning nil where either is not found
+	// (if the certificate is found but not keypair, that is not an error: only the cert will be returned).
+	// This func returns a cert, private key and a bool.  The bool value is whether the keypair is stored
+	// in a legacy format. This bool is used by a keypair
+	// task to convert a Legacy Keypair to the new Keypair API format.
+	FindKeypair(name string) (*Certificate, *PrivateKey, bool, error)
+}
+
+func IssueCert(request *IssueCertRequest, keystore Keystore) (*Certificate, *PrivateKey, *Certificate, error) {
```
rifelpet, line 68: Can this function get a comment? Mostly because it has two identical return types

pkg/pki/issue.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+
+	if request.MinValidDays != 0 {
+		hash := fnv.New32()
+		addrs, err := net.InterfaceAddrs()
+		sort.Slice(addrs, func(i, j int) bool {
+			return addrs[i].String() < addrs[j].String()
+		})
+		if err == nil {
```
justinsb, line 151: Nit: we should probably warn in this error case.  It might also be easier to understand if the random generation was split into its own function.  Finally, I do wonder if it's worth it, vs a simple random number?

justinsb:
Drive-by nit, but this is a big simplification of the code - thank you @johngmyers 

/approve
/lgtm

pkg/pki/issue.go
johngmyers: We want all certs generated on the same node to have the same validity period, so everything on the node breaks at the same time and there's more time until the next node breaks. A simple random number would have different things on the node breaking at different times and different nodes having components breaking at times closer to each other.

johngmyers: Conflicted, so rebased and addressed those items.

PR #9131, opened 2020-05-16
Remove support for reading legacy-format keypairs

johngmyers:


upup/pkg/fi/vfs_castore.go
```diff
@@ -147,24 +145,24 @@ func (c *VFSCAStore) buildPrivateKeyPath(name string, id string) vfs.Path {
...
+		return nil, false, fmt.Errorf("object was not a keyset, was a %T", object)
 	}
 
 	if gvk == nil {
-		return nil, "", fmt.Errorf("object did not have GroupVersionKind: %q", keyset.Name)
+		return nil, false, fmt.Errorf("object did not have GroupVersionKind: %q", keyset.Name)
 	}
 
-	return keyset, KeysetFormat(gvk.Version), nil
+	return keyset, gvk.Version != KeysetFormatV1Alpha2, nil
```
rifelpet, line 165: this seems to be a pretty important use of the constant that is buried in the middle of this file. Any thoughts on having something like
```go
const KeysetFormatLatest = KeysetFormatV1Alpha2
```

defined at the top of the file and referencing it here? I'm debating if that would improve readability or hinder it...

upup/pkg/fi/vfs_castore.go
```diff
@@ -931,6 +809,7 @@ func (c *VFSCAStore) storeCertificate(name string, ki *keysetItem) error {
 		}
 	}
 
+	// TODO remove this
```
rifelpet, line 812: Is this referring to the code block below? when should (or could) it get removed?

upup/pkg/fi/vfs_castore.go
rifelpet: In fact this might be the only use of `KeysetFormatV1Alpha2` now, so maybe it makes sense to get rid of it and just have this in ca.go:

```go
const (
  KeysetFormatV1Latest KeysetFormat = "v1alpha2"`
)

Comment (johngmyers) [reply upup/pkg/fi/vfs_castore.go:165]: I've been considering inlining the const since this is the only remaining reference. It certainly doesn't need to be public anymore.

File: upup/pkg/fi/vfs_castore.go (hunk 2 above)
Comment (johngmyers) [reply upup/pkg/fi/vfs_castore.go:812]: Yes. To remove writing of the legacy format I'd have to  fix the unit test that relies on them, verify there is nothing else left reading them, and write code to delete them from upgraded keystores. To allow kops version rollback this should only be done in a subsequent kops release.

Comment (rifelpet) [reply upup/pkg/fi/vfs_castore.go:812]: ok, can we update the comment to reflect that?

```suggestion
	// TODO remove this when we drop support for rolling back to kops versions that support legacy-format keypairs
```

upup/pkg/fi/vfs_castore.go
rifelpet: ok, i just want to make sure i understood what was happening here. Inlining the const could happen in a followup PR 👍

PR #9133, opened 2020-05-17
Map kube-apiserver service-account-jwks-uri flag

justinsb:
This is needed/useful for identity federation to AWS.


mikesplain:
Thanks @justinsb, looks like you need to update CRDs as well.  
/lgtm
/approve
/retest

PR #9135, opened 2020-05-17
GCE: don't rely on hostname being correct

justinsb:
Distros that use systemd for DHCP often don't have the hostname
correct, due to e.g. the requirement for policy kit.

We don't rely on it being set correctly on other clouds; no real
reason to require it on GCP either!

Example issue: https://github.com/systemd/systemd/issues/13501

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -357,7 +357,7 @@ push: crossbuild-nodeup
 
 .PHONY: push-gce-dry
 push-gce-dry: push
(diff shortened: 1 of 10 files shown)


rdrgmnzs:
Looks good to me besides the tests that need fixing.

geojaz:
+1 will lgtm with green tests @justinsb

PR #9137, opened 2020-05-17
Move networking in nodeup to dedicated subpackage

olemarkus:
As discussed in a previous PR, the networking builders should be in its own package. It went a bit further than intended when trying to clean up the NetworkBuilder. The CNI plugin installations were a bit messy. If a CNI requires kops to install plugins, it will be handled in specific builders instead.


johngmyers:
This is turning into a real pain to review. It would have been much easier to cross-check if there were a commit per CNI builder.

nodeup/pkg/model/network.go
```diff
@@ -1,123 +0,0 @@
...
-var _ fi.ModelBuilder = &NetworkBuilder{}
-
-// Build is responsible for configuring the network cni
-func (b *NetworkBuilder) Build(c *fi.ModelBuilderContext) error {
-	var assetNames []string
-
-	// @TODO need to clean up this code, it isn't the easiest to read
-	networking := b.Cluster.Spec.Networking
-	if networking == nil || networking.Classic != nil {
-	} else if networking.Kubenet != nil || networking.GCE != nil {
```
johngmyers, line 43: Where did the asset adding for GCE go? Is `KubenetBuilder` supposed to also act for `networking.GCE`?

nodeup/pkg/model/network.go
```diff
@@ -1,123 +0,0 @@
...
-		assetNames = append(assetNames, "bridge", "host-local", "loopback")
-	} else if networking.External != nil {
```
johngmyers, line 45: Where did the asset adding for External go? Is KubenetBuilder supposed to also act for networking.External?

upup/pkg/fi/nodeup/command.go
```diff
@@ -244,62 +242,23 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
...
 	loader.Builders = append(loader.Builders, &model.SecretBuilder{NodeupModelContext: modelContext})
 	loader.Builders = append(loader.Builders, &model.FirewallBuilder{NodeupModelContext: modelContext})
-	loader.Builders = append(loader.Builders, &model.NetworkBuilder{NodeupModelContext: modelContext})
 	loader.Builders = append(loader.Builders, &model.SysctlBuilder{NodeupModelContext: modelContext})
 	loader.Builders = append(loader.Builders, &model.KubeAPIServerBuilder{NodeupModelContext: modelContext})
 	loader.Builders = append(loader.Builders, &model.KubeControllerManagerBuilder{NodeupModelContext: modelContext})
 	loader.Builders = append(loader.Builders, &model.KubeSchedulerBuilder{NodeupModelContext: modelContext})
 	loader.Builders = append(loader.Builders, &model.EtcdManagerTLSBuilder{NodeupModelContext: modelContext})
+	loader.Builders = append(loader.Builders, &model.KubeProxyBuilder{NodeupModelContext: modelContext})
+	loader.Builders = append(loader.Builders, &model.EtcdTLSBuilder{NodeupModelContext: modelContext})
```
johngmyers, line 251: `EtcdTLSBuilder` needs to be changed to not act when networking is not Calico.

nodeup/pkg/model/network.go
```diff
@@ -1,123 +0,0 @@
...
-		// external is based on kubenet
-		assetNames = append(assetNames, "bridge", "host-local", "loopback")
-
-	} else if networking.CNI != nil || networking.Weave != nil || networking.Flannel != nil || networking.Calico != nil || networking.Canal != nil || networking.Kuberouter != nil || networking.Romana != nil || networking.AmazonVPC != nil || networking.Cilium != nil {
```
johngmyers, line 49: My crosschecks were unable to find corresponding asset adding for CNI, Weave, Calico, Romana, AmazonVPC, or Cilium.

johngmyers:
That review took way too long. For future refactors, please submit a series of commits, each designed to permit a reviewer to see when logic changes and when it does not. Please see #9130 as an example.

upup/pkg/fi/nodeup/command.go
johngmyers: Perhaps rename it to `CalicoBuilder`.

nodeup/pkg/model/networking/flannel.go
```diff
@@ -0,0 +1,75 @@
...
+	*model.NodeupModelContext
+}
+
+var _ fi.ModelBuilder = &FlannelBuilder{}
+
+// Build is responsible for configuring the network cni
+func (b *FlannelBuilder) Build(c *fi.ModelBuilderContext) error {
+	networking := b.Cluster.Spec.Networking
+
+	if networking.Flannel != nil {
```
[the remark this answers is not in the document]
olemarkus: Calico daemonset installs all the necessary plugins for canal. ptp is not referenced in the CNI config file.

nodeup/pkg/model/networking/kube_router.go
```diff
@@ -40,9 +44,11 @@ func (b *KubeRouterBuilder) Build(c *fi.ModelBuilderContext) error {
 			Path:     "/var/lib/kube-router/kubeconfig",
 			Contents: fi.NewStringResource(kubeconfig),
 			Type:     nodetasks.FileType_File,
-			Mode:     s("0400"),
+			Mode:     fi.String("0400"),
 		})
 	}
 
+	b.AddCNIBinAssets(c, []string{"loopback", "host-local", "bridge", "portmap"})
```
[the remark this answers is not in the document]
olemarkus: Yes. "ptp" is not used by kube-router

nodeup/pkg/model/networking/kubenet.go
```diff
@@ -0,0 +1,40 @@
...
+
+var _ fi.ModelBuilder = &KubenetBuilder{}
+
+// Build is responsible for configuring the kube-router
+func (b *KubenetBuilder) Build(c *fi.ModelBuilderContext) error {
+	if b.Cluster.Spec.Networking.Kubenet == nil && b.Cluster.Spec.Networking.Kopeio == nil {
+		return nil
+	}
+
+	b.AddCNIBinAssets(c, []string{"loopback", "host-local", "bridge", "portmap"})
```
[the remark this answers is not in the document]
olemarkus: I thought it was convenient to have, but it actually isn't used when kubelet is in kubenet mode. Will remove.

upup/pkg/fi/nodeup/command.go
olemarkus: Good catch. Will look into this

nodeup/pkg/model/network.go
olemarkus: All of these install the CNI plugins themselves. I think it is better to just let them do so as the appropriate plugin set is more correct then. For example cilium only requires loopback (because kubernetes itself does), but got a ton installed with it.

nodeup/pkg/model/network.go
olemarkus: Thanks. I didn't realise that GCE was kubenet.

nodeup/pkg/model/network.go
olemarkus: External is a bit special. Kops doesn't do anything here, but expects a daemonset to handle networking. But it is a good point that we should probably just install the plugins required by kubenet anyway.

upup/pkg/fi/nodeup/command.go
olemarkus: This one is a bit misleading. It creates certs named calico, but it is used by any provider that accesses legacy etcd with TLS enabled. For Romana, this seems to be broken, for cilium, I have removed support for this. Calico _could_ probably work if the user works really hard to to configure this. Would have to use calico V2 ++. I think we can just leave this as is and rather work on deprecating legacy etcd further.

nodeup/pkg/model/network.go
olemarkus: This is replaced with a generic check for kubenet-based providers now.

nodeup/pkg/model/network.go
olemarkus: This is replaced with a generic check for kubenet-based providers now.

nodeup/pkg/model/networking/flannel.go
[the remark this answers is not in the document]
johngmyers: Okay, so intended. I suggest making such logic changes in a separate commit than the one that moves the code. It makes it clearer to reviewers and code archaeologists that the change was intentional (and that there was, in fact, a change).

nodeup/pkg/model/networking/kube_router.go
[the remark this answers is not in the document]
johngmyers: Comment resolved.

nodeup/pkg/model/network.go
johngmyers: The comment immediately below says "external is based on kubenet". `pkg/model/components.UsesKubenet()` returns `true` for External.

The lack of documentation makes it unclear to me what External is for. I'm not confident one could, for example, run a Cilium daemonset with networking set to External.

nodeup/pkg/model/network.go
johngmyers: If they're all intentional removals, then comment resolved.

nodeup/pkg/model/networking/kubenet.go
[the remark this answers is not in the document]
johngmyers: Comment resolved by subsequent commit.

nodeup/pkg/model/networking/flannel.go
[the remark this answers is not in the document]
johngmyers: ...and comment resolved

nodeup/pkg/model/network.go
johngmyers: Comment resolved by subsequent commit.

nodeup/pkg/model/network.go
johngmyers: Comment resolved by subsequent commit (as `UsesKubenet()` returns `true` for External).

upup/pkg/fi/nodeup/command.go
johngmyers: I do not want to start creating that cert when it previously was not. The appending of `EtcdTLSBuilder` was previously conditional on networking being set to Calico. So either the appending should remain conditional or the builder should only act for Calico.

olemarkus: Agreed.

nodeup/pkg/model/network.go
olemarkus: So external baffled me a bit as well.
Kublet has a flag `network-plugin` that can for all intents and purposes only be set to `kubenet` or `cni`. Cilium, calico etc are CNI plugins. Kubenet is an older implementation where kubernetes handled more stuff internally, but you could still provide an additional daemonset to create overlay networks and such (e.g kopeio vxlan).

on kops `--networking cni` means bring your own cni plugin, while `--networking external` means bring your own kubenet "plugin". So it basically just determines what value is passed to kublet.

I plan on cleaning up our docs wrt this in a later PR.

olemarkus:
Sorry for putting all of this into one commit. It was quite a journey cleaning all of this up and the original commits reflected that.
I could/should have rewritten the changes from scratch and commit patches more easier to review.

nodeup/pkg/model/network.go
johngmyers: For this future doc cleanup:

When I removed Classic from `docs/networking.md`, that reduced the first section to stating that there are 3 networking modes. I think that needs to be taken down to 2 modes: "Kubenet" and "CNI". External and CNI should be put into their own consolidated "BYO" section.

That doc doesn't adequately mention the Kubenet-based variants and I don't have the knowledge to add them. It's also not clear to me whether the Kubenet info in that doc applies to all Kubenet variants or just to the Kubenet provider itself.

upup/pkg/fi/nodeup/command.go
johngmyers: Comment resolved by subsequent commit.

nodeup/pkg/model/networking/flannel.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"k8s.io/klog"
+	"k8s.io/kops/nodeup/pkg/model"
+	"k8s.io/kops/pkg/systemd"
+
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
+)
+
+// CiliumBuilder writes Cilium's assets
```
rdrgmnzs, line 28: NIT: Function description name doesn't match function name.

nodeup/pkg/model/networking/lyft.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/kops/nodeup/pkg/model"
+	api "k8s.io/kops/pkg/apis/kops"
+	"k8s.io/kops/upup/pkg/fi"
+)
+
+type LyftVPCBuilder struct {
+	*model.NodeupModelContext
+}
+
+var _ fi.ModelBuilder = &CiliumBuilder{}
```
rdrgmnzs, line 40:
```suggestion
var _ fi.ModelBuilder = &LyftVPCBuilder{}
```

nodeup/pkg/model/networking/kubenet.go
```diff
@@ -0,0 +1,45 @@
...
+
+package networking
+
+import (
+	"k8s.io/kops/nodeup/pkg/model"
+	"k8s.io/kops/pkg/model/components"
+	"k8s.io/kops/upup/pkg/fi"
+)
+
+// KuberouterBuilder installs kube-router
```
rdrgmnzs, line 25: NIT: Descriptions function name and description does not match function.

olemarkus:
Thanks for spotting those @rdrgmnzs

nodeup/pkg/model/networking/kubenet.go
```diff
@@ -22,7 +22,7 @@ import (
 	"k8s.io/kops/upup/pkg/fi"
 )
 
-// KuberouterBuilder installs kube-router
+// Kubenet installs kube-router
```
johngmyers, line 25:
```suggestion
// KubenetBuilder installs kube-router.
```

nodeup/pkg/model/networking/kubenet.go
```diff
@@ -0,0 +1,42 @@
...
+
+package networking
+
+import (
+	"k8s.io/kops/nodeup/pkg/model"
+	"k8s.io/kops/pkg/model/components"
+	"k8s.io/kops/upup/pkg/fi"
+)
+
+// KubenetBuilder installs Kubenet networking provider
```
johngmyers, line 25: nit:
```suggestion
// KubenetBuilder installs the Kubenet networking provider.
```

rdrgmnzs:
Thanks for all the work you put into this @olemarkus!

/lgtm
/approve

PR #9140, opened 2020-05-18
Add table of networking providers and their status

olemarkus:
As discussed on slack, we should have a table here to guide users at what kind of support one can expect.

There is a quite a bit to debate here.
The question marks is where I believe they have this status, but no idea since when. Predates the release logs so we can just default to 1.6 here if no one minds.

Romana is dead and I think this one is a no-brainer. Starting a new cluster with Romana hasn't worked since the introduction of etcd-manager.
Classic doesn't work. If you try to create the cluster, you get a validation error about classic not being a valid provider.
Kopeio is undocumented and not frequently modify. We could keep it at experimental. It seems to work when I install it.
Flannel has had its issues lately, but I am not too familiar with the status. My guess is experimental.

### DIFF
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -6,19 +6,18 @@ The following table provides the support status for various networking providers
 
 | Network provider | Experimental | Stable | Deprecated | Removed | 
 | ------------ | -----------: | -----: | ---------: | ------: |
-| AWS VPC | ? | ? | - | - |
+| AWS VPC | 1.9 | - | - | - |
 | Calico | 1.6 | 1.11 | - | - |
 | Canal  | 1.12 | - | - | - |
 | Cilium | 1.9 | 1.15 | - | - |
-| Classic | ? | ? | 1.17 | 1.18 |
-| Flannel udp | ? | - | - | - |
-| Flannel vxlan | ? | - | - | - |
-| Kopeio | ? | - | 1.17 | 1.18 |
+| Flannel udp | 1.5.2 | - | - | - |
+| Flannel vxlan | 1.8.0 | - | - | - |
+| Kopeio | 1.5 | - | 1.17 | 1.18 |
 | Kube-router | 1.6.2 | - | - | - |
-| Kubenet | ? | ? | - | - |
+| Kubenet | 1.5 | 1.5 | - | - |
 | Lyft VPC | 1.11 | - | - | - |
-| Romana | ? | - | 1.17 | 1.18 |
-| Weave | ? | ? | - | - |
+| Romana | 1.8 | - | 1.17 | 1.18 |
+| Weave | 1.5 | - | - | - |
 
 
 The networking options determines how the pod and service networking is implemented and managed.
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -1,4 +1,25 @@
-## Kubernetes Networking Options
+# Kubernetes Networking Options
+
+## Supported networking options
+
+The following table provides the support status for various networking providers with regards to Kops version: 
+
+| Network provider | Experimental | Stable | Deprecated | Removed | 
+| ------------ | -----------: | -----: | ---------: | ------: |
+| AWS VPC | ? | ? | - | - |
+| Calico | 1.6 | 1.11 | - | - |
+| Canal  | 1.12 | - | - | - |
+| Cilium | 1.9 | 1.15 | - | - |
+| Classic | ? | ? | 1.17 | 1.18 |
+| Flannel udp | ? | - | - | - |
+| Flannel vxlan | ? | - | - | - |
+| Kopeio | ? | - | 1.17 | 1.18 |
+| Kube-router | 1.6.2 | - | - | - |
+| Kubenet | ? | ? | - | - |
+| Lyft VPC | 1.11 | - | - | - |
+| Romana | ? | - | 1.17 | 1.18 |
+| Weave | ? | ? | - | - |
+
 
 The networking options determines how the pod and service networking is implemented and managed.
 
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -12,11 +12,11 @@ The following table provides the support status for various networking providers
 | Cilium | 1.9 | 1.15 | - | - |
 | Flannel udp | 1.5.2 | - | - | - |
 | Flannel vxlan | 1.8.0 | - | - | - |
-| Kopeio | 1.5 | - | 1.17 | 1.18 |
+| Kopeio | 1.5 | - | - | - |
 | Kube-router | 1.6.2 | - | - | - |
 | Kubenet | 1.5 | 1.5 | - | - |
 | Lyft VPC | 1.11 | - | - | - |
-| Romana | 1.8 | - | 1.17 | 1.18 |
+| Romana | 1.8 | - | 1.18 | 1.19 |
 | Weave | 1.5 | - | - | - |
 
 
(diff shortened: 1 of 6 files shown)


joshbranham:
We have been using Canal since 1.11 I believe without issues (outside of the calico CRD migration downtime, where we just built new clusters). What is the criteria for stable, outside of a hunch based on users anecdotes and GH issues etc?

Overall this looks good, I think its helpful to steer people towards the `happy path` configurations.

olemarkus:
> We have been using Canal since 1.11 I believe without issues (outside of the calico CRD migration downtime, where we just built new clusters). What is the criteria for stable, outside of a hunch based on users anecdotes and GH issues etc?
> 
> Overall this looks good, I think its helpful to steer people towards the `happy path` configurations.

Canal should actually have been 1.11 because of the v3 change. It is maybe somewhat arbitrary, but it is sufficiently far into the past too.

docs/networking.md
```diff
@@ -1,4 +1,25 @@
...
+
+The following table provides the support status for various networking providers with regards to Kops version: 
+
+| Network provider | Experimental | Stable | Deprecated | Removed | 
+| ------------ | -----------: | -----: | ---------: | ------: |
+| AWS VPC | ? | ? | - | - |
+| Calico | 1.6 | 1.11 | - | - |
+| Canal  | 1.12 | - | - | - |
+| Cilium | 1.9 | 1.15 | - | - |
+| Classic | ? | ? | 1.17 | 1.18 |
```
johngmyers, line 13: Classic has been unsupported as of Kubernetes 1.4.0

johngmyers: I think Classic should simply be omitted from the table.

docs/networking.md
```diff
@@ -1,4 +1,25 @@
-## Kubernetes Networking Options
+# Kubernetes Networking Options
+
+## Supported networking options
+
+The following table provides the support status for various networking providers with regards to Kops version: 
+
+| Network provider | Experimental | Stable | Deprecated | Removed | 
+| ------------ | -----------: | -----: | ---------: | ------: |
+| AWS VPC | ? | ? | - | - |
```
johngmyers, line 9: Added in kops 1.9

docs/networking.md
```diff
@@ -1,4 +1,25 @@
...
+| Calico | 1.6 | 1.11 | - | - |
+| Canal  | 1.12 | - | - | - |
+| Cilium | 1.9 | 1.15 | - | - |
+| Classic | ? | ? | 1.17 | 1.18 |
+| Flannel udp | ? | - | - | - |
```
johngmyers, line 14: Added in kops 1.5.2

docs/networking.md
```diff
@@ -1,4 +1,25 @@
...
+| Flannel vxlan | ? | - | - | - |
```
johngmyers, line 15: Added in kops 1.8.0

docs/networking.md
```diff
@@ -1,4 +1,25 @@
...
+| Kopeio | ? | - | 1.17 | 1.18 |
```
johngmyers, line 16: Added in kops 1.5.0

docs/networking.md
```diff
@@ -1,4 +1,25 @@
...
+| Kube-router | 1.6.2 | - | - | - |
+| Kubenet | ? | ? | - | - |
```
johngmyers, line 18: Added in kops 1.5.0

docs/networking.md
```diff
@@ -1,4 +1,25 @@
...
+| Lyft VPC | 1.11 | - | - | - |
+| Romana | ? | - | 1.17 | 1.18 |
```
johngmyers, line 20: Added in kops 1.8.0

docs/networking.md
```diff
@@ -1,4 +1,25 @@
...
+| Weave | ? | ? | - | - |
```
johngmyers, line 21: Added in kops 1.5.0

johngmyers:
Add GCE, which was added in 1.15, to the table?

olemarkus:
> I don't see either of the other deprecations/removals in the release notes or code. I don't object to deprecating them, but they do need to be announced and implemented.
> 

Let's agree on the status first. Then I can implement deprecations in code/docs shortly after

> Would we not remove network providers based on Kubernetes version? The table suggests it would be by kops version.

Both images and cloud providers use kops version for lifecycle milestones.To me it makes sense to stick with that.

> 
> Also is Weave considered stable? That hasn't been my experience.

I am happy to make it experimental.

Thanks a lot for digging up the versions above. I never expected anyone to do that :)

docs/networking.md
olemarkus: Classic is mentioned in the docs, API, and kops help, so we need to handle that as well.

olemarkus:
> Add GCE, which was added in 1.15, to the table?

Thanks. Will add that as well.

johngmyers:
I'm more inclined to merge the deprecations/removals before merging the addition of a table. I've been having trouble getting the already agreed upon 1.17 deprecations merged.

Addition, stability, and deprecation are inherently by kops version. It does not make sense to conflate the choice of removal edge with those. Have we removed a cloud provider?

Support for the Classic provider was removed based on Kubernetes version, so there's that precedent.

Removal by kops version makes more sense when the thing being removed is already dead, as is the case for the deprecated images. Removal by Kubernetes version allows the removal to start earlier.

Romana appears to be already dead, so appropriate to be removed by kops version. I presume Kopeio is as well?

olemarkus:
> I'm more inclined to merge the deprecations/removals before merging the addition of a table. I've been having trouble getting the already agreed upon 1.17 deprecations merged.
> 

Absolutely. I am not removing the WIP label before everything is in order. It is just easier to discuss around this table than around the code that would implement this.

> Addition, stability, and deprecation are inherently by kops version. It does not make sense to conflate the choice of removal edge with those. Have we removed a cloud provider?
> 

Not yet. But I would like to see vmware removed. It looks like a vendor dump a couple of years ago and nothing has happened since. So it is a matter of time.

> Support for the Classic provider was removed based on Kubernetes version, so there's that precedent.
> 
> Removal by kops version makes more sense when the thing being removed is already dead, as is the case for the deprecated images. Removal by Kubernetes version allows the removal to start earlier.
> 
> Romana appears to be already dead, so appropriate to be removed by kops version. I presume Kopeio is as well?

The ones I marked as both removed and deprecated are projects that look dead to me. So I think we can just process these asap so that users are aware.

If we decide to drop something like weave, we need a longer process and probably a migration path. I think that would be a different deprecation process entirely.

olemarkus:
On classic, I will remove it from the table and add a new PR removing this from docs etc.

docs/networking.md
johngmyers: I removed it from this doc, at least.

olemarkus:
Not much feedback here so I assume we are good. I will bump the deprecation versions by one and add some warnings.

johngmyers:
1.17 shipped without deprecating either Romana or Kopeio. The earliest we could deprecate would be 1.18. We should commit a deprecation announcement to the 1.18 release notes, possibly through a separate PR.

Kopeio has gotten commits this year, so it's not completely dead.

olemarkus:
Yep. What I meant with bumping deprecation versions was to move 1.17 -> 1.18. Can leave kopio as experimental too. Better to be conservative here.

olemarkus:
Will remove Romana from master in another PR shortly

rifelpet:
/approve

We've been using the AWS VPC CNI provider since 1.11 without any issue but the criteria for "stable" may be vague, so we could address that table cell in another PR 👍

PR #9142, opened 2020-05-19
Remove Classic networking from docs

johngmyers:
Classic networking support was dropped as of Kubernetes 1.4.


hakman:
Cleanup is nice!
/lgtm

PR #9154, opened 2020-05-21
Add support for encryption in Cilium 

### LINKED ISSUE #9031: Add ability to set encryption flags for Cilium
When using the CNI for Cilium my team would like to enable the ipsec settings for pod-to-pod and node traffic.

MoShitrit:
Fixes #9031 
Summary of the PR:

1. Update the networking modules and CRDs (including auto generated) to support `enableEncryption` flag, along with creating an encryption key, along with relevant documentation.

2. Add a subcommand `ciliumpassword` for `kops create secret`. This will generate the encryption key according to the provided instructions in the [official cilium docs](https://docs.cilium.io/en/v1.7/gettingstarted/encryption/#generate-import-the-psk).


johngmyers:
Weave added a `create secret weavepassword` subcommand. I think I'd prefer a new subcommand of `create secret` than an option on `create secret encryptionconfig`.

Alternatively we could rename `create secret weavepassword` to `create secret network` and use that secret for any networking provider that supports encryption.

MoShitrit:
> Weave added a `create secret weavepassword` subcommand. I think I'd prefer a new subcommand of `create secret` than an option on `create secret encryptionconfig`.
> 
> Alternatively we could rename `create secret weavepassword` to `create secret network` and use that secret for any networking provider that supports encryption.

Yeah, my initial thought was to create a separate subcommand for cilium, but then I figured it will end up being too much code replication, just for the sake of having a different command.
As for renaming the weave script, I wasn't sure if that would be a good practice, since doing that would break backwards compatibility - how do we usually feel about this kind of this?

Which eventually brought me to think that adding on top of the existing one without breaking any existing compatibility would work best, for future cases (and weave would be the single exception).

WDYT?

johngmyers:
I think the user interface of an option on a subcommand is worse than the code duplication. The code duplication could probably be addressed with a refactor.

Cilium should either use the secret named "weavepassword" or there should be auto-migration code. For the Cobra subcommand we could make "weavepassword" an alias for "network" or whatever we choose.

docs/networking/cilium.md
```diff
@@ -85,6 +85,23 @@ Note that since Cilium Operator is the entity that interacts with the EC2 API to
 
 Also note that this feature has only been tested on the default kops AMIs.
 
+#### Enabling Encryption in Cilium
+Starting kops verion 1.18, it is possible to enable encryption for Cilium agent.
```
olemarkus, line 89: I think this feature will rather go into 1.19 due to the larger feature set we have already put into 1.18

pkg/apis/kops/networking.go
```diff
@@ -257,6 +257,9 @@ type CiliumNetworkingSpec struct {
 	EnableTracing bool `json:"enableTracing,omitempty"`
 	// EnablePrometheusMetrics enables the Cilium "/metrics" endpoint for both the agent and the operator.
 	EnablePrometheusMetrics bool `json:"enablePrometheusMetrics,omitempty"`
+	// EncryptionEnabled enables Cilium Encryption.
+	// Default: false
+	EncryptionEnabled bool `json:"encryptionEnabled"`
```
olemarkus, line 262: To be more inline with the keys above, I suggest renaming to `EnableEncryption`

MoShitrit:
> I think the user interface of an option on a subcommand is worse than the code duplication. The code duplication could probably be addressed with a refactor.
> 
> Cilium should either use the secret named "weavepassword" or there should be auto-migration code. For the Cobra subcommand we could make "weavepassword" an alias for "network" or whatever we choose.

Gotcha, @johngmyers. Point taken about the user interface. 😄 
So since weave seems less strict about the key requirements than Cilium, I think it'll make most since to separate it to a new subcommand (`kops create secret ciliumpassword`). Uniting both of them under the same file would still mean they need separate functions, so at this point might as well just create a new subcommand.
I'll get to work on it

docs/networking/cilium.md
MoShitrit: Fixed in https://github.com/kubernetes/kops/pull/9154/commits/41c76dabaad8ffcbef6a4b19dfdf1174325bd8f8

pkg/apis/kops/networking.go
MoShitrit: Fixed in https://github.com/kubernetes/kops/pull/9154/commits/41c76dabaad8ffcbef6a4b19dfdf1174325bd8f8

MoShitrit:
@johngmyers @olemarkus 
I've addressed all comments and suggestions, and squashed my commits to a single one so it'll be cleaner.
Let me know your thoughts.
Thanks! 😄

MoShitrit:
@johngmyers @olemarkus
I've addressed all comments and suggestions, and squashed my commits to a single one so it'll be cleaner.
Let me know your thoughts.
Thanks! 😄

docs/cli/kops_create_secret.md
```diff
@@ -52,6 +52,7 @@ Create a secret
 ### SEE ALSO
 
 * [kops create](kops_create.md)	 - Create a resource by command line, filename or stdin.
+* [kops create secret ciliumpassword](kops_create_secret_ciliumpassword.md)	 - Create a cilium encryption config.
```
olemarkus, line 55: Nit, but I think "encryption key" is more correct than "encryption config"

MoShitrit: Agreed, fixing!

MoShitrit: @olemarkus - Fixed and pushed. Looks like that caused your `/lgtm` to be discarded.
Would you mind re-adding that if it looks okay to you?
Thanks!

MoShitrit:
@olemarkus now that 1.18 was branched-off, are we clear to remove the hold?

olemarkus:
Yep.

/hold cancel

KashifSaadat:
Looks great, thanks @MoShitrit for the contribution!

/approve

PR #9155, opened 2020-05-21
Update alpha channels with May updates

MoShitrit:


rifelpet:
I think we'll want to add 1.18 to the alpha channel at some point. I don't recall exactly when in the kops release process that gets added though.

/approve

MoShitrit:
> I think we'll want to add 1.18 to the alpha channel at some point. I don't recall exactly when in the kops release cycle that gets added though.
> 
> /approve

@rifelpet Yeah, I thought about that too, but since I didn't see a release branch for 1.18, I wasn't sure if there's a specific time in which we add the new version to play.
If you think now is a good time- I can add it through a new PR since this one is already merged 😄

PR #9156, opened 2020-05-21
Add unit test case for pkg/apis/kops/util/versions_test.go

Hellcatlk:
Signed-off-by: ZouYu <zouy.fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/apis/kops/util/versions_test.go
+++ b/pkg/apis/kops/util/versions_test.go
@@ -17,26 +17,79 @@ limitations under the License.
 package util
 
 import (
+	"fmt"
+	"reflect"
 	"testing"
+
+	"github.com/blang/semver"
 )
 
 func Test_ParseKubernetesVersion(t *testing.T) {
-	grid := map[string]string{
-		"1.3.7":         "1.3.7",
-		"v1.4.0-beta.8": "1.4.0-beta.8",
-		"1.5.0":         "1.5.0",
-		"https://storage.googleapis.com/kubernetes-release-dev/ci/v1.4.0-alpha.2.677+ea69570f61af8e/": "1.4.0",
+	cases := []struct {
+		version       string
+		expected      *semver.Version
+		expectedError error
+	}{
+		{
+			version: "1.3.7",
+			expected: &semver.Version{
+				Major: 1,
+				Minor: 3,
+				Patch: 7,
+			},
+		},
+		{
+			version: "v1.4.0-beta.8",
+			expected: &semver.Version{
+				Major: 1,
+				Minor: 4,
+				Patch: 0,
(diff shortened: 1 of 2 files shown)


johngmyers:
I'd prefer the grid to explicitly state whether the test case is expecting success or an error.
/ok-to-test

Hellcatlk:
@johngmyers , done.

pkg/apis/kops/util/versions_test.go
```diff
@@ -21,21 +21,54 @@ import (
...
+			if c.expectedError {
+				continue
+			}
+			t.Errorf("ParseKubernetesVersion error parsing %q: %v", c.version, err)
 		}
 
 		actual := sv.String()
-		if actual != expected {
-			t.Errorf("version mismatch: %q -> %q but expected %q", version, actual, expected)
+		if actual != c.expected {
```
johngmyers, line 70: This will incorrectly pass if the test case expects an error but the function under test returns success and an empty string.

I usually make `expectedError` of type `string` and put the expected error message in it.

pkg/apis/kops/util/versions_test.go
```diff
@@ -21,21 +21,54 @@ import (
...
+			expectedError: false,
+		},
+		{
+			version:       "https://storage.googleapis.com/kubernetes-release-dev/ci/v1.4.0-alpha.2.677+ea69570f61af8e/",
+			expected:      "1.4.0",
+			expectedError: false,
+		},
+		{
+			version:       "",
+			expected:      "",
```
johngmyers, line 51: You don't need to explicitly list `expected: ""` for cases that expect an error.

pkg/apis/kops/util/versions_test.go
johngmyers: `Test_TemplateFunctions_CloudControllerConfigArgv` shows a pattern where `expectedError` is an `error`. That appears to be the current pattern in kops.

Hellcatlk: Thanks for your comment, But I can't find where the `Test_TemplateFunctions_CloudControllerConfigArgv` is.

pkg/apis/kops/util/versions_test.go
Hellcatlk: done

pkg/apis/kops/util/versions_test.go
johngmyers: upup/pkg/fi/cloudup/template_functions_test.go

Hellcatlk: Thx, done.

johngmyers:
Thanks!
/lgtm

PR #9157, opened 2020-05-21
Networking cleanup

olemarkus:


pkg/apis/kops/v1alpha2/networking.go
```diff
@@ -38,21 +38,21 @@ type NetworkingSpec struct {
 }
 
 // ClassicNetworkingSpec is the specification of classic networking mode, integrated into kubernetes
```
rifelpet, line 40:
```suggestion
// ClassicNetworkingSpec is the specification of classic networking mode, integrated into kubernetes.
```

rifelpet: should these changes also be made to `pkg/apis/kops/networking.go`?

olemarkus: They should :)

pkg/apis/kops/networking.go
```diff
@@ -37,22 +37,22 @@ type NetworkingSpec struct {
 	GCE        *GCENetworkingSpec        `json:"gce,omitempty"`
 }
 
-// ClassicNetworkingSpec is the specification of classic networking mode, integrated into kubernetes
+// ClassicNetworkingSpec is the specification of classic networking mode, integrated into kubernetes.
+// Support been removed since kops 1.4.
```
johngmyers, line 41: No, it was Kubernetes 1.4. I don't think this information is relevant.

pkg/apis/kops/networking.go
```diff
@@ -37,22 +37,22 @@ type NetworkingSpec struct {
...
 }
 
 // KubenetNetworkingSpec is the specification for kubenet networking, largely integrated but intended to replace classic
 type KubenetNetworkingSpec struct {
 }
 
-// ExternalNetworkingSpec is the specification for networking that is implemented by a Daemonset
-// It also uses kubenet
+// ExternalNetworkingSpec is the specification for networking that is implemented by a Daemonset.
+// Uses the kubenet networking provider.
```
johngmyers, line 50: I think "networking provider" is misleading as it doesn't use the kops Kubenet provider. It uses the Kubernetes kubenet networking code.

pkg/apis/kops/networking.go
olemarkus: Thanks. There are quite a few issues about when features are added/removed from kops. And given that we refer users to the spec docs, I think this is relevant to many. At least it doesn't hurt.

pkg/apis/kops/networking.go
olemarkus: Amended.

johngmyers: Resolved

pkg/apis/kops/networking.go
johngmyers: Resolved

cmd/kops/create_cluster.go
```diff
@@ -1111,7 +1109,7 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 
 	case api.TopologyPrivate:
 		if !supportsPrivateTopology(cluster.Spec.Networking) {
-			return fmt.Errorf("Invalid networking option %s. Currently only '--networking kopeio-vxlan (or kopeio)', '--networking weave', '--networking flannel', '--networking calico', '--networking canal', '--networking kube-router', '--networking romana', '--networking amazon-vpc-routed-eni', '--networking cilium', '--networking lyftvpc', '--networking cni' are supported for private topologies", c.Networking)
+			return fmt.Errorf("invalid networking option %s. Kubenet does not support private topology", c.Networking)
```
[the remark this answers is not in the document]
olemarkus: That is a good point. `supportsPrivateTopology` is also only called from this location, so given the simplification, the function call can just be replaced with an `kubenet != nil`.

cmd/kops/create_cluster.go
```diff
@@ -1454,11 +1452,7 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 }
 
 func supportsPrivateTopology(n *api.NetworkingSpec) bool {
-
-	if n.CNI != nil || n.Kopeio != nil || n.Weave != nil || n.Flannel != nil || n.Calico != nil || n.Canal != nil || n.Kuberouter != nil || n.Romana != nil || n.AmazonVPC != nil || n.Cilium != nil || n.LyftVPC != nil || n.GCE != nil {
-		return true
-	}
-	return false
+	return n.Kubenet == nil
```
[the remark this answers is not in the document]
olemarkus: I think true is more correct. Given that both GCP and Kopeio supports private topology, it is up to the user to configure things correctly if they choose to use external.

docs/cli/kops_create_cluster.md
```diff
@@ -94,7 +94,7 @@ kops create cluster [flags]
       --master-zones strings             Zones in which to run masters (must be an odd number)
       --model string                     Models to apply (separate multiple models with commas) (default "proto,cloudup")
       --network-cidr string              Set to override the default network CIDR
-      --networking string                Networking mode to use.  kubenet (default), classic, external, kopeio-vxlan (or kopeio), weave, flannel-vxlan (or flannel), flannel-udp, calico, canal, kube-router, romana, amazon-vpc-routed-eni, cilium, cni. (default "kubenet")
+      --networking string                Networking mode to use.  kubenet (default), external, weave, flannel-vxlan (or flannel), flannel-udp, calico, canal, kube-router, amazon-vpc-routed-eni, cilium, cni, lyftvpc. (default "kubenet")
```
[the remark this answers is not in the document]
olemarkus: Yep.

pkg/apis/kops/networking.go
```diff
@@ -37,22 +37,22 @@ type NetworkingSpec struct {
...
 }
 
 // KubenetNetworkingSpec is the specification for kubenet networking, largely integrated but intended to replace classic
 type KubenetNetworkingSpec struct {
 }
 
-// ExternalNetworkingSpec is the specification for networking that is implemented by a Daemonset
-// It also uses kubenet
+// ExternalNetworkingSpec is the specification for networking that is implemented by a Daemonset.
+// Uses the kubenet kubelet networking plugin.
```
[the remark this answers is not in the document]
olemarkus: Sounds good

cmd/kops/create_cluster.go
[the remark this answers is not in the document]
johngmyers: Resolved

pkg/apis/kops/networking.go
```diff
@@ -37,22 +37,20 @@ type NetworkingSpec struct {
...
 type ClassicNetworkingSpec struct {
 }
 
 // KubenetNetworkingSpec is the specification for kubenet networking, largely integrated but intended to replace classic
 type KubenetNetworkingSpec struct {
 }
 
-// ExternalNetworkingSpec is the specification for networking that is implemented by a Daemonset
-// It also uses kubenet
+// ExternalNetworkingSpec is the specification for networking that is implemented by a user-provided Daemonset that uses the kubelet networking plugin.
```
johngmyers, line 49:
```suggestion
// ExternalNetworkingSpec is the specification for networking that is implemented by a user-provided Daemonset that uses the Kubenet kubelet networking plugin.
```

cmd/kops/create_cluster.go
```diff
@@ -330,7 +330,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVar(&options.Image, "image", options.Image, "Image to use for all instances.")
 
-	cmd.Flags().StringVar(&options.Networking, "networking", options.Networking, "Networking mode to use.  kubenet (default), classic, external, kopeio-vxlan (or kopeio), weave, flannel-vxlan (or flannel), flannel-udp, calico, canal, kube-router, romana, amazon-vpc-routed-eni, cilium, cni.")
```
rifelpet, line 333: is there a reason `kopeio-vxlan (or kopeio), ` was removed? I _think_ that is [Justin's CNI](https://github.com/kopeio/networking) and if it still works I think we should keep it in the CLI docs, and if it doesn't we should discuss next steps but either way I think that would be outside the scope of this PR.

thoughts?

olemarkus: kopeio works. But there is no documentation of that CNI in the kops docs, so it isn't very helpful to have users consider that option. I could even see us even narrow the list to only listing stable CNIs following #9140 (that would be the next steps, in my optinion).

If it is a blocker for this PR, I will add it back in though.

rifelpet: ah, I suppose that makes sense. That sounds good to me now, thanks!

PR #9162, opened 2020-05-22
Revert "Add EC2 instance lifecycle label to nodes"

johngmyers:
This reverts commit 7f2f195a cherrypicked in through #9152.

Concern was raised during Kops Office Hours as to the naming of the label. I propose we revert this change from 1.17 to give us until 1.18 to resolve the naming decision.

### DIFF
--- a/cmd/kops-controller/controllers/node_controller.go
+++ b/cmd/kops-controller/controllers/node_controller.go
@@ -118,15 +118,6 @@ func (r *NodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, fmt.Errorf("unable to build config for node %s: %v", node.Name, err)
 	}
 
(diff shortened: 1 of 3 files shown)


granular-ryanbonham:
We currently handle this with our own controller that adds a "node-labeler.granular.ag/spot" label to the nodes.. I can see where kops would want to add other node labels in the future.. Should use do something like "node-labeler.kops.k8s.io/spot-worker".. That would avoid the reuse of the "node-role.kubernetes.io" which was raised as a concern and establish a good pattern if we need to add more.. P.S. Would love to see this in 1.17....

chrisz100:
SSL error sounds super odd...

/retest
/lgtm

rifelpet:
GH outage...

/retest

hakman:
Not sure if the decision was actually to revert it. I see in the notes:
> Decision: used by k8s spot termination handler tool, so probably not renaming label

Is it such a big deal if 1.17 ships with this as it was?

PR #9165, opened 2020-05-23
Try validating multiple times before updating instancegroup

johngmyers:
My alternative to #8897

### DIFF
--- a/pkg/instancegroups/instancegroups.go
+++ b/pkg/instancegroups/instancegroups.go
@@ -88,16 +88,8 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(ctx context.Context, c
 
 	if isBastion {
 		klog.V(3).Info("Not validating the cluster as instance is a bastion.")
-	} else if c.CloudOnly {
-		klog.V(3).Info("Not validating cluster as validation is turned off via the cloud-only flag.")
-	} else {
-		if err = c.validateCluster(); err != nil {
-			if c.FailOnValidate {
-				return err
-			}
-			klog.V(2).Infof("Ignoring cluster validation error: %v", err)
-			klog.Info("Cluster validation failed, but proceeding since fail-on-validate-error is set to false")
-		}
+	} else if err = c.maybeValidate("", 1); err != nil {
+		return err
 	}
 
 	if !c.CloudOnly {
@@ -156,7 +148,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(ctx context.Context, c
 					klog.Infof("waiting for %v after detaching instance", sleepAfterTerminate)
 					time.Sleep(sleepAfterTerminate)
 
-					if err := c.maybeValidate(c.ValidationTimeout, "detaching"); err != nil {
+					if err := c.maybeValidate(" after detaching instance", c.ValidateCount); err != nil {
 						return err
 					}
 					noneReady = false
@@ -185,7 +177,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(ctx context.Context, c
 			return waitForPendingBeforeReturningError(runningDrains, terminateChan, err)
 		}
 
-		err = c.maybeValidate(c.ValidationTimeout, "removing")
+		err = c.maybeValidate("after terminating instance", c.ValidateCount)
 		if err != nil {
 			return waitForPendingBeforeReturningError(runningDrains, terminateChan, err)
 		}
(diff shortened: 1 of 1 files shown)


pkg/instancegroups/instancegroups.go
```diff
@@ -88,16 +88,8 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(ctx context.Context, c
...
-		klog.V(3).Info("Not validating cluster as validation is turned off via the cloud-only flag.")
-	} else {
-		if err = c.validateCluster(); err != nil {
-			if c.FailOnValidate {
-				return err
-			}
-			klog.V(2).Infof("Ignoring cluster validation error: %v", err)
-			klog.Info("Cluster validation failed, but proceeding since fail-on-validate-error is set to false")
-		}
+	} else if err = c.maybeValidate("", 1); err != nil {
```
granular-ryanbonham, line 91: What is reason for hard coding validateCount to 1 vs using c.validationCount which is 2 by default? It seems like a tiny speed gain for added code complexity passing around the value, also may not be what user expects if they have set a higher validationCount.

johngmyers: The code hasn't recently made a change to the cluster, so it's not likely to flap from validated to failing. We don't want to hang around before we start waiting for it to inexplicably flap.

I doubt users who have set a higher validationCount would expect to have that count doubled between instance groups.

granular-ryanbonham: Addon manifest will have been updated in S3, so CNIs for example may be updating on the existing nodes. On large clusters this can take a bit... That said I think this is something we can tweak/revisit if needed. I will LGTM for now if it is a problem we can address.

granular-ryanbonham:
@johngmyers I can no longer approve, so we will need additional eyes on this..

pkg/instancegroups/instancegroups.go
```diff
@@ -185,7 +177,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(ctx context.Context, c
 			return waitForPendingBeforeReturningError(runningDrains, terminateChan, err)
 		}
 
-		err = c.maybeValidate(c.ValidationTimeout, "removing")
+		err = c.maybeValidate("after terminating instance", c.ValidateCount)
```
rifelpet, line 180:
```suggestion
		err = c.maybeValidate(" after terminating instance", c.ValidateCount)
```

pkg/instancegroups/instancegroups.go
```diff
@@ -88,16 +88,8 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(ctx context.Context, c
...
-	} else {
-		if err = c.validateCluster(); err != nil {
-			if c.FailOnValidate {
-				return err
-			}
-			klog.V(2).Infof("Ignoring cluster validation error: %v", err)
-			klog.Info("Cluster validation failed, but proceeding since fail-on-validate-error is set to false")
-		}
+	} else if err = c.maybeValidate("", 1); err != nil {
+		return err
```
rifelpet, line 92: Why are we now validating even if CloudOnly is true?

pkg/instancegroups/instancegroups.go
```diff
@@ -383,40 +375,40 @@ func (c *RollingUpdateCluster) drainTerminateAndWait(ctx context.Context, u *clo
...
+			klog.Warningf("Cluster validation failed%s, proceeding since fail-on-validate is set to false: %v", operation, err)
 		}
 	}
 	return nil
 }
 
 // validateClusterWithDuration runs validation.ValidateCluster until either we get positive result or the timeout expires
-func (c *RollingUpdateCluster) validateClusterWithDuration(validationTimeout time.Duration) error {
-	ctx, cancel := context.WithTimeout(context.Background(), validationTimeout)
+func (c *RollingUpdateCluster) validateClusterWithDuration(validateCount int) error {
```
rifelpet, line 399: should this function be renamed now that it uses a count parameter rather than a duration parameter?

pkg/instancegroups/instancegroups.go
johngmyers: `maybeValidate()` doesn't validate if CloudOnly is true. This is removing redundant code

pkg/instancegroups/instancegroups.go
johngmyers: What the function does is impose the timeout. I combined it and `tryValidateCluster`

PR #9170, opened 2020-05-24
Remove loader support for nodeup tasks not used in models

johngmyers:
/kind cleanup


upup/pkg/fi/nodeup/nodetasks/service.go
```diff
@@ -87,22 +86,6 @@ func (s *Service) String() string {
 	return fmt.Sprintf("Service: %s", s.Name)
 }
 
-func NewService(name string, contents string, meta string) (fi.Task, error) {
```
olemarkus, line 90: Could you leave this one in?
I am using this one in #9137

johngmyers: Where? I don't see that PR adding a file to the model directory. If you're creating a `nodetasks.Service` you should just use a literal, not call `NewService` with a string.

justinsb:
I'm surprised this one isn't used; I'm going to dig into why it isn't / where it came from.

On removing the serialization, I do think it's a good idea to clean up the custom deserialization.  I do have a longer term idea that we should put less logic into nodeup, and pass it more of these tasks (as I think the logic in nodeup is much more opaque, harder to understand, harder to change).  But I don't think that is incompatible with removing the "legacy" deserialization.  (And we should just rely on yaml/json deserialization if we start passing raw tasks, IMO)

justinsb:
Yup, you're right ... unused legacy code.  Thanks @johngmyers!

/approve
/lgtm

johngmyers:
I find tasks in Go code to be easier to understand and change than those in YAML. My IDE can cross-reference the Go code.

PR #9174, opened 2020-05-25
Remove unused VFSScan

johngmyers:


olemarkus:
So I also went down this path. The code removed here is the only place using `WriterTo` interface that `VFSPath` has to implement. 

While this code is indeed dead in kops, it is used in etcd-manager: https://github.com/kopeio/etcd-manager/blob/master/pkg/backup/vfs.go#L44 As far as I can see, it is only `CopyFile` though, so the rest of the code can probably be removed.

johngmyers:
Perhaps it would be better if etcd-manager pulled `CopyFile` into its own repo? I don't see any particular reason that function has to be in kops.

johngmyers:
Even better would be to replace the `WriterTo` requirement with something that allows direct copy to another `Path`.

johngmyers:
Created kopeio/etcd-manager#324 which removes the dependency on `CopyFile` and the extra temporary file it creates.

justinsb:
Yeah, we should remove this if it's not used.  I think it was used for mirroring keys, but we ended up with an approach at the higher (object level) of abstraction IIRC.

/approve
/lgtm

PR #9177, opened 2020-05-25
Remove vsphere cloud provider

### LINKED ISSUE #9166: Remove vsphere support
(Almost) all of the code related to vsphere was added in #2401. After that most changes were to keep things compiling. It has always lived behind a feature flag and it requires a lot of hacks to get running.

I don't see this going anywhere, especially now that VMWare is pushing Tanzu for all it's worth. So I suggest we just punt it.

olemarkus:
Fixes #9166


hakman:
@olemarkus Could you split the vendor code into a separate commit? Should make it easier to review.

olemarkus:
Certainly. Better now?

hakman:
Yes, looks better. Thanks!

upup/pkg/fi/cloudup/dns.go
```diff
@@ -166,22 +166,6 @@ func precreateDNS(ctx context.Context, cluster *kops.Cluster, cloud fi.Cloud) er
 	}
 
 	recordsMap := make(map[string]dnsprovider.ResourceRecordSet)
-	// vSphere provider uses CoreDNS, which doesn't have rrs.List() function supported.
-	// Thus we use rrs.Get() to check every dnsHostname instead
-	if cloud.ProviderID() != kops.CloudProviderVSphere {
```
johngmyers, line 171: This is a `!=` comparison, so the body should become unconditional

rifelpet:
I posted in the issue, but i think its worth waiting a few days for responses before we continue with the removal just to be sure.

https://github.com/kubernetes/kops/issues/9166#issuecomment-633723137

olemarkus:
/hold

If all looks good, it would be nice to get approval. 
I'll remove the hold if there is no response in #9166 after the weekend.

rifelpet:
👍 
/approve

PR #9178, opened 2020-05-26
WIP Introduce v1beta1 API

johngmyers:
Introduce the v1beta1 API as a copy of v1alpha2. Subsequent PRs during the kops 1.19 development period are expected to change it.

/milestone v1.19


rifelpet:
Great idea! I wonder if its worth having an umbrella issue to track and discuss desired changes from a high level? before the work is done to implement the changes into a PR. I know I have a few in the back of my head too.

olemarkus:
I guess this new API version will be implemented over a number of PRs based on the various changes we'd like to put into them. I suppose some will be more global and requires broader agreement, while some are more local (such as removing all the unused fields in Cilium).

Starting with an issue on what we'd like to achieve with this API version would be good.

How large are the changes we see here? Is beta the next step or will we change some parts of the spec so much that v1alpha3 makes more sense?

Regardless, I certainly welcome a new API version and I think 1.19 is a good time to add this.

rifelpet:
Heres some from my list, feel free to paste into an issue. Most are around consistency and some could be done without bumping api versions but I never got around to it.

* `KubeAPIServerConfig` and `KubeProxyConfig` both have CPU and Memory resource fields of type `string` rather than `*resource.Quantity`
* `AccessSpec.DNS` is unused
* `LoadBalancerAccessSpec.UseForInternalAPI` could have a better name
* Any field that represents a docker image should have a consistent name. We currently have a mix of `Image` and `ImageName`
* Rename `KubeDNSSpec` given that it is also used by CoreDNS
* `BastionSpec.BastionPublicName` -> `BastionSpec.PublicName`
* Any fields related to security groups should be made more consistent or consolidated: 
    * `BastionLoadBalancerSpec.AdditionalSecurityGroups`
    * `CloudConfiguration.ElbSecurityGroup`
    * `InstanceGroupSpec.AdditionalSecurityGroups`
    * `InstanceGroupSpec.SecurityGroupOverride`
    * `LoadBalancerAccessSpec.AdditionalSecurityGroups`
    * `LoadBalancerAccessSpec.SecurityGroupOverride`
* type `LoadBalancer` could use a more specific name
* In general we could look at each field with `additional` or `override` in the name and figure out if we can merge it with the "original" data, making it a list. `NetworkCIDR` and `AdditionalNetworkCIDRs` for example.

olemarkus:
* Cilium
  * remove all unused fields.
  * Some fields should also be renamed to be similar to upstream, e.g `IPTablesRulesNoinstall -> InstallIptablesRules`
* Consistency way of 
 * repository/image:version is specify
 * enable prometheus for addons

How easy would it be to change the structs? Some of those we have today is overly flat and use prefixes, where their helm counterpart is fairly nested. Nesting could make things more user-friendly and easier to reason about.

rifelpet:
`InstanceGroupSpec`'s `RootVolume*` fields would be another good candidate for a nested struct

johngmyers:
It is an apimachinery requirement that new fields be added to the old apiversions. This is so they round-trip.

I have not found apimachinery rules on removing unsupported fields. I believe fields that are checked by api validation should remain in the unversioned API with a `TODO: remove with v1alpha2` comment. Fields that are simply ignored I suspect we can remove from the unversioned API.

The v1alpha2 API has been around so long I believe we should treat it as requiring at least the beta deprecation period of 3 releases/9 months. We might even treat it as stable and give it 12 months before removing. I don't think the new API should have less than a beta deprecation period either.

olemarkus:
So if I understand this correctly, we can either offer a v1alpha3 where we can break whatever, but create script/subcommand that converts the compatible parts from v1alpha2 to v1alpha3, or we can introduce v1beta1, but then have to keep v1alpha2 in sync. Most of the stuff in @rifelpet's first post can be done because it is just renaming the json keys, while the structs can remain as they are today.

johngmyers:
@olemarkus I don't think you understand correctly. The difference between alpha and beta is only in the minimum length of the deprecation period before we remove the API. For beta it is 3 releases or 9 months, whichever is longer. For alpha it is 0 releases. I don't think either the old or new APIs should be removed immediately.

For APIs that still exist we need to maintain round-trip semantics, so new fields have to be added to old api versions and moved fields have to have conversion code for old apis. This is independent on whether they are alpha or beta.

rifelpet:
hm, didn't we determine that v1alpha1 was missing some fields before we removed it? was that because the prow jobs weren't enforcing round-trip semantics?

olemarkus:
@johngmyers I am not sure we disagree here. I am suggesting some changes that breaks round-trip. But if we can avoid this, we definitely should. I don't want to be in a situation where those fields are with us forever though.

But if we remove e.g the deprecated cilium fields from v1beta1, can we have conversion logic that sets the unversioned api values to blank? Or is it possible to remove the deprecated fields from the unversioned API and leave them only in v1alpha1, then have conversion logic that sets these values to blank in v1alpha1?

johngmyers:
I'm not entirely sure what the exact rules are for removing fields: the apimachinery documentation is extremely hard to find and the few times I did find it I didn't have time left to read it.

As I said before, I think we can simply remove unused, unvalidated fields from the unversioned API. That would allow them to be specified, but they'd be immediately dropped.

olemarkus:
Doesn't seem too hard. When one make incompatible changes one have to implement manual conversion, as expected. apimachinery makes some boilerplate functions to make life easier (everything that is still mapped 1:1). See https://github.com/johngmyers/kops/compare/v1beta1...olemarkus:api-v1beta1-cilium for my attempt at doing various changes to the API.

olemarkus:
> hm, didn't we determine that v1alpha1 was missing some fields before we removed it? was that because the prow jobs weren't enforcing round-trip semantics?

@rifelpet think this happen a while ago before we had `make verify-staticcheck` or something. Have some faint memories of that.

johngmyers:
One wrinkle is that kops writes the unversioned API to "cluster.spec" in the state store. We won't be able to make any incompatible changes to the unversioned API.

I'm working on a PR to start migrating to a versioned form of the full cluster spec, also writing the unversioned API to "cluster.spec" but not reading it. Perhaps we do that migration in 1.19 and introduce v1beta1 in 1.20.

olemarkus:
I don't know if we should do truly incompatible changes. The ones I did will mean some values will get lost, but only those that were ignored already. So while conversion isn't entirely 1:1, the effect on the cluster is the same.

It may be a bit confusing to the users though.

Another question: Do we want to change some defaults when we create new api version? E.g flip some of the flags we already tell the user to do because of security.

rifelpet:
It would be nice to have roundtrip tests, given a v1alpha2 manifest, convert it to v1beta1 and back and ensure its equal. and do the same for v1beta to v1alpha2 and back. It seems like this would be a good place for that:

https://github.com/kubernetes/kops/tree/master/tests/integration/conversion

and we should make sure the manifests include any fields that we're modifying in the new apiVersion. This could be in a followup PR though.

johngmyers:
We're going to need a whole kops minor version with the #9229 fix or we're going to be significantly constrained in how we can change the unversioned API. So I think this project will have to wait until kops 1.20.
/milestone v1.20

olemarkus:
Just out of curiosity, what is an incompatible change? Would any of the changes in my branch be incompatible (the round-trip is not entirely complete for some as the deprecated values will be nulled even if the user set a value. The effect on the cluster is identical though)

hakman:
@johngmyers Can this be closed now that #12406 was merged?

johngmyers:
This has lists of things to change in the API, so I'm keeping it open for that.

PR #9179, opened 2020-05-26
Update DigitalOcean cloud-controller-manager to v0.1.24

timoreimann:
This changes bumps DigitalOcean's cloud-controller-manager (CCM) to version 0.1.24 which brings a number of new features and bug fixes.

It seems like there are now two occurrences of the CCM manifest: one as YAML and one in-code. I wasn't entirely sure if the image versions needed to be updated in both places, which I did for now. Please let me know in case that's not needed.

Ideally, the change should be cherry-picked into 1.17 subsequently.

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -1767,7 +1767,7 @@ spec:
           operator: Exists
           tolerationSeconds: 300
       containers:
-      - image: digitalocean/digitalocean-cloud-controller-manager:v0.1.20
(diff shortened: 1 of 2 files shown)


hakman:
> Ideally, the change should be cherry-picked into 1.17 subsequently.

Would this be an issue from your point of view @zetaab ?

timoreimann:
If no one objects then I'll move forward with the cherry pick. Let me know if there are any concerns, though.

timoreimann:
Nevermind, I see it got already taken care of. 🎉

Issue #9180, opened 2020-05-26
Ubuntu 20.04 - Nodes sometime fails to come up due to package install issues?


andersosthus:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
1.17-beta2

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
1.17.5

**3. What cloud provider are you using?**
AWS

Monday 25th of May we started noticing problems with new nodes added to our clusters. When adding nodes, some nodes comes up just fine, while some doesn't come up at all.

When investigating the failed nodes, we see what appears to be the `kops-configuration.service` being restarted in the middle of an `apt-get install` command, resulting in a bad `dpkg` state that needs to be resolved with `dpkg --configure -a`.

The timeline on a nodeup looks like this:
1. Nodeup starts
1. When starting to install the required packages, nodeup gets restarted (`kops-configuration.service: Main process exited, code=killed, status=15/TERM`)
1. Nodeup then starts up again. Depending on when the restart occured (either during a package install or just inbetween two package installs) determines if the nodeup command completes successfully. Also note that it's not the same package this occurs on every time, but it's always on one of the initial packages.
1. If the restart of nodeup occured during a package install, nodeup will halt when it tries to install packages again due to a bad dpkg state

Right now it seems random if a node comes up or not.

Excerpt of the `kops-configuration.service` around the time the restart happens:
```
May 26 12:10:20 ip-10-9-155-142.eu-west-1.compute.internal nodeup[1992]: I0526 12:10:20.562898    1992 executor.go:176] Executing task "Package/nfs-common": Package: nfs-common
May 26 12:10:20 ip-10-9-155-142.eu-west-1.compute.internal nodeup[1992]: I0526 12:10:20.562926    1992 package.go:142] Listing installed packages: dpkg-query -f ${db:Status-Abbrev}${Version}\n -W nfs-common
May 26 12:10:20 ip-10-9-155-142.eu-west-1.compute.internal nodeup[1992]: I0526 12:10:20.562981    1992 executor.go:176] Executing task "File//var/lib/kubelet/kubeconfig": File: "/var/lib/kubelet/kubeconfig"
May 26 12:10:20 ip-10-9-155-142.eu-west-1.compute.internal nodeup[1992]: I0526 12:10:20.563068    1992 files.go:50] Writing file "/var/lib/kubelet/kubeconfig"
[...]

PR #9182, opened 2020-05-26
Use systemd-timesyncd for Ubuntu 20.04

hakman:
`systemd-timesyncd` is the default service for clock synchronisation in Ubuntu 20.04. Installing `ntp` seems to make `kops-configuration` crash as reported in #9180.

We already use `systemd-timesyncd` for Ubuntu in 1.18 and this issue is not noticeable. Using it  in 1.17 and 1.16 instead of `ntp` seems the best way to go.

https://github.com/kubernetes/kops/blob/82257362184894e882bdbba79e79840c2b89b340/nodeup/pkg/model/ntp.go#L60-L64

### DIFF
--- a/nodeup/pkg/model/ntp.go
+++ b/nodeup/pkg/model/ntp.go
@@ -46,7 +46,10 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 		return nil
 	}
 
-	if b.Distribution.IsDebianFamily() {
+	if b.Distribution == distros.DistributionFocal {
+		c.AddTask(&nodetasks.Package{Name: "systemd-timesyncd"})
+		c.AddTask((&nodetasks.Service{Name: "systemd-timesyncd"}).InitDefaults())
+	} else if b.Distribution.IsDebianFamily() {
 		c.AddTask(&nodetasks.Package{Name: "ntp"})
 		c.AddTask((&nodetasks.Service{Name: "ntp"}).InitDefaults())
 	} else if b.Distribution.IsRHELFamily() {


nodeup/pkg/model/ntp.go
```diff
@@ -46,7 +46,10 @@ func (b *NTPBuilder) Build(c *fi.ModelBuilderContext) error {
 		return nil
 	}
 
-	if b.Distribution.IsDebianFamily() {
+	if b.Distribution == distros.DistributionFocal {
```
johngmyers, line 49:
```suggestion
	if b.Distribution.IsUbuntu() && b.Distribution != distros.DistributionXenial && b.Distribution != distros.DistributionBionic {
```

hakman: Would it be ok to leave it as is? Seems simpler.

johngmyers: The issue would be whether a new major version of Ubuntu is used with kops 1.17. But I suppose code would have to be added to 1.17 to support it and that's not likely.

hakman: Exactly. I don't think there is any plan to support any other distro in the near future. Thanks!

justinsb:
Sounds good to me; I'm also wondering whether we should install this in more distros, but thinking about e.g. debian buster.

In any case, this is a good change - thanks @hakman, and @johngmyers for reviewing!

/approve
/lgtm

johngmyers:
This probably undid the effect of #8312 for Focal. We should see about configuring systemd-timesyncd with any cloud provider supplied time source.

hakman:
@johngmyers what do you mean? I think that is only in 1.18+.

johngmyers:
I mean the corresponding change in 1.18.

hakman:
I think my change in 1.18 kept the spirit and also fixed what was broken in that PR:
https://github.com/kubernetes/kops/commit/5003bd2f16aa206568d6ca76dbcc3711d9e02cb6#diff-ae9bef714f1963c836e9427f8dfea1f2
https://github.com/kubernetes/kops/commit/9ae94664bf3d831f19378b6a94a72a6cc48136b0#diff-ae9bef714f1963c836e9427f8dfea1f2

PR #9184, opened 2020-05-27
Update etcd-manager to 3.0.20200527

justinsb:
Changes since 3.0.20200429:

* Use env vars to customize backup retention
* Use next attachment point when device already in use
* Simplify uploading backups


hakman:
@justinsb we may want to do another release once we have a solution for https://github.com/kubernetes/kops/pull/9171.
Thanks for pushing a new release 😄 
/lgtm

PR #9188, opened 2020-05-27
Fix nits for removal of S3 file versions

hakman:
@justinsb noticed a few minor issues with https://github.com/kubernetes/kops/pull/9171.

### DIFF
--- a/util/pkg/vfs/s3fs.go
+++ b/util/pkg/vfs/s3fs.go
@@ -106,13 +106,13 @@ func (p *S3Path) Remove() error {
 	return nil
 }
 
-func (p *S3Path) RemoveAll() error {
+func (p *S3Path) RemoveAllVersions() error {
 	client, err := p.client()
 	if err != nil {
 		return err
 	}
 
-	klog.V(8).Infof("removing file %s", p)
+	klog.V(8).Infof("removing all versions of file %s", p)
 
 	request := &s3.ListObjectVersionsInput{
 		Bucket: aws.String(p.bucket),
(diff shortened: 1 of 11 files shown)


util/pkg/vfs/s3fs.go
```diff
@@ -106,13 +106,13 @@ func (p *S3Path) Remove() error {
...
 
-func (p *S3Path) RemoveAll() error {
+func (p *S3Path) RemoveAllVersions() error {
 	client, err := p.client()
 	if err != nil {
 		return err
 	}
 
-	klog.V(8).Infof("removing file %s", p)
+	klog.V(8).Infof("removing  all versions of file %s", p)
```
johngmyers, line 115:
```suggestion
	klog.V(8).Infof("removing all versions of file %s", p)
```

util/pkg/vfs/vfs.go
```diff
@@ -59,8 +59,8 @@ type Path interface {
 	// Remove deletes the file
 	Remove() error
 
-	// RemoveAll completely deletes the file (with all its versions and markers)
-	RemoveAll() error
+	// RemoveAllVersions completely deletes the file (with all its versions and markers)
```
johngmyers, line 62:
```suggestion
	// RemoveAllVersions completely deletes the file (with all its versions and markers).
```

hakman: This would make it inconsistent with the other comments around it.

johngmyers: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

hakman: I agree with the principle. Fixed.

PR #9189, opened 2020-05-27
Remove WriterTo from VFS interface

olemarkus:
And remove/don't export WriteTo in VFS implementations

With  #9174 merged, the WriterTo interface is no longer used in VFS.

### DIFF
--- a/util/pkg/vfs/BUILD.bazel
+++ b/util/pkg/vfs/BUILD.bazel
@@ -22,7 +22,6 @@ go_library(
     importpath = "k8s.io/kops/util/pkg/vfs",
     visibility = ["//visibility:public"],
     deps = [
-        "//pkg/try:go_default_library",
         "//util/pkg/hashing:go_default_library",
         "//vendor/github.com/aws/aws-sdk-go/aws:go_default_library",
         "//vendor/github.com/aws/aws-sdk-go/aws/awserr:go_default_library",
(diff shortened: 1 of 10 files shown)


justinsb:
We're using this in etcd-manager, right?

We better be anyway ... loading etc backups into a []byte would be very expensive! :-)

justinsb:
Ah I see what happened...

https://github.com/kopeio/etcd-manager/blob/18116e4a53ca4bcfa9f656ec9ee840087ce0202a/pkg/backup/vfs.go#L152

We do want the ability to write to a VFSPath via streaming, which is what WriteTo gives us that we can't otherwise get.  I probably shouldn't have merged the removal of CopyFile until we had the replacement in etcd-manager; I think that replacement is going to use WriteTo

Does that sound right @johngmyers and @olemarkus?

olemarkus:
@justinsb can we remove `WriterTo` from the VFS interface and rather implement them just in the VFS implementations that etcd-manager will use? That way we don't need dummy implementations of `writeTo` for the VFS that doesn't support streaming anyway (like the Vault implementation in #9094).

johngmyers:
I filed etcd-manager/issues#326, which I can take on.

johngmyers:
I think we want to replace `WriteTo` with a variant of `ReadFile` that returns an `io.ReadCloser` and `error`. The caller would be expected to always call `Close` when the `error` is `nil`.

johngmyers:
Almost all the existing `ReadFile` implementations are implemented in terms of `WriteTo`. It would be trivial to implement `WriteTo` in terms of `ReadFile`.

PR #9196, opened 2020-05-28
Update Bazel rules for Docker to v0.14.2

hakman:
This version contains a fix required for multi architecture support when creating container images:
https://github.com/bazelbuild/rules_docker/releases/tag/v0.14.2

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -6,20 +6,20 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 http_archive(
     name = "io_bazel_rules_go",
+    sha256 = "dbde93957f2a96e6c97edb74e29107b7de1b2b57608b8304401e10326e02c66d",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.22.2/rules_go-v0.22.2.tar.gz",
-        "https://github.com/bazelbuild/rules_go/releases/download/v0.22.2/rules_go-v0.22.2.tar.gz",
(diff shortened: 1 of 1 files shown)


WORKSPACE
```diff
@@ -6,20 +6,20 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 http_archive(
     name = "io_bazel_rules_go",
+    sha256 = "dbde93957f2a96e6c97edb74e29107b7de1b2b57608b8304401e10326e02c66d",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.22.2/rules_go-v0.22.2.tar.gz",
```
rifelpet, line 10: any reason not to bump this to the latest v0.23.1 ? I see [architecture related improvements in v0.23.0](https://github.com/bazelbuild/rules_go/releases/tag/v0.23.0)

hakman: Output paths change in v0.23.0:
https://github.com/bazelbuild/rules_go/releases/tag/v0.23.0
> Output file paths no longer include extra configuration data. For example, //tests/core/go_binary:hello creates the file bazel-bin/tests/core/go_binary/hello_/hello. Note that output file paths are still not stable and should not be depended on unless the "out" attribute is set explicitly (for rules that support it).

rifelpet: Looking at every use of `go_binary` we either reference it immediately in a `container_image` or don't do anything with it, in which case our make targets copy them elsewhere, so its possibly only the Makefile would need updating.

we can do that version bump and test it out in a separate PR 👍

rifelpet:
travis might be stuck :(

/lgtm
/approve

hakman:
Yup, noticed in a few PRs...
Nothing on https://www.traviscistatus.com though.

PR #9197, opened 2020-05-28
Remove all traces of utils.tar.gz

hakman:
With support for CoreOS gone, utis.tar.gz is not needed anywhere anymore. See also:
* https://github.com/kubernetes/kops/pull/8978
* https://github.com/kubernetes/kops/pull/9065


johngmyers:
Does ContainerOS not need this either?

Please also remove from docs/aws-china.md

hakman:
> Does ContainerOS not need this either?

ContainerOS has the binaries bundled, I did a quick test o GCP. No idea if there is a reference with the installed packages there.

hakman:
Rebased after https://github.com/kubernetes/kops/pull/8922

PR #9200, opened 2020-05-28
Add olemarkus as reviewer

hakman:
@olemarkus has been an active contributor and reviewer for some time now and I think he would make a great addition to the Kops reviewers.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -20,5 +20,6 @@ reviewers:
   - joshbranham
   - kashifsaadat
   - mikesplain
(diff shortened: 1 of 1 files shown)


johngmyers:
Technically, @olemarkus isn't eligible until Jun 24.

hakman:
> Technically, @olemarkus isn't eligible until Jun 24.

Not that far away though. If this is a hard requirement, I guess we can hold for another 3 weeks.

PR #9204, opened 2020-05-29
Use kubescheduler.config.k8s.io/v1beta1 for Kubernetes 1.19

hakman:
https://github.com/kubernetes/kubernetes/pull/91420


hakman:
I think this is the thing that broke the build for `release/latest`.
/cc @rifelpet @johngmyers

PR #9205, opened 2020-05-29
Allow listing versions for objects in the S3 bucket

hakman:
`s3:ListBucketVersions` is required for permanently deleting files from the S3 state store (for now intended just for backups).
Followup for https://github.com/kubernetes/kops/pull/9171#discussion_r431245812.


rifelpet:
The api action is ListObjectVersions but the iam action is ListBucketVersions...

Confirmed here https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazons3.html

/approve

PR #9208, opened 2020-05-29
Update adding_a_feature.md with more modern example

johngmyers:
Move away from nodeup model files. This one is light on testing and the history is a bit revisionist, but there weren't a lot of examples that touched api, validation, and nodeup.

/cc @olemarkus

### DIFF
--- a/docs/development/adding_a_feature.md
+++ b/docs/development/adding_a_feature.md
@@ -1,84 +1,142 @@
 This is an overview of how we added a feature:
 
-To make auto-upgrading on the nodes an option.
+To add an option for Cilium to use the ENI IPAM mode.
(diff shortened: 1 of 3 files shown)


olemarkus:
I think it is a huge improvement over what was already there. And an extra test to boot!

At some point we need to tie some of those development pages more together. E.g with testing.md.

/lgtm

johngmyers:
Unfortunately, API validation tests have about the least value to effort. They test internal APIs and fail to test the description of the error messages.

It probably would be worthwhile to add an integration test for ENI mode and document the process.

Is there a good way to test nodeup model code other than the e2e testgrid?

PR #9214, opened 2020-05-30
Upgrde amazon vpc cni to 1.6.2

MoShitrit:
Two changes in this PR:
1. Upgrading image to version 1.6.2 following [release](https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.6.2)
2. Applying changes to node selectors, following [this update](https://github.com/aws/amazon-vpc-cni-k8s/pull/937) to the upstream manifest

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -3926,15 +3926,15 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
-              - key: "beta.kubernetes.io/os"
+              - key: "kubernetes.io/os"
                 operator: In
                 values:
                 - linux
-              - key: "beta.kubernetes.io/arch"
+              - key: "kubernetes.io/arch"
                 operator: In
                 values:
                 - amd64
-              - key: eks.amazonaws.com/compute-type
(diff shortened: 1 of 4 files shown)


upup/models/bindata.go
```diff
@@ -3926,15 +3926,15 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
-              - key: "beta.kubernetes.io/os"
+              - key: "kubernetes.io/os"
```
gjtempleton, line 3929: I don't think we want to do this just yet as we are only just removing support for k8s 1.9 as of the 1.18 release, and these stable labels were only added as of k8s 1.14.

Looks like we'll need to fix it and all the other uses through the codebase of the beta labels before 1.19 though: https://github.com/kubernetes/kubernetes/pull/91046

hakman: @gjtempleton This change is applied only to the 1.16+ template, so should be ok.

MoShitrit: Yep, I’m not planning to backport it to earlier versions since it doesn’t seem to contain any critical fixes, just slight improvements. So if we merge it today it’ll probably go to 1.18 which should be fine.

PR #9216, opened 2020-05-30
Prepare Kops for multi-architecture support

hakman:
The multi-architecture (ARM64) support will be a larger change. Sorting out these things in advance will make it simpler. https://github.com/kubernetes/kops/pull/8938#issuecomment-631159009

Each commits takes care of a certain aspect.

/cc @justinsb @mikesplain @rifelpet


util/pkg/architectures/architectures.go
```diff
@@ -0,0 +1,59 @@
...
+func FindArchitecture() (Architecture, error) {
+	switch runtime.GOARCH {
+	case "amd64":
+		return ArchitectureAmd64, nil
+	default:
+		return "", fmt.Errorf("unsupported arch: %q", runtime.GOARCH)
+	}
+}
+
+func GetArchitectures() []Architecture {
```
johngmyers, line 41: `GetSupportedArchitectures()` ?

pkg/model/components/context.go
```diff
@@ -176,13 +169,13 @@ func Image(component string, architecture string, clusterSpec *kops.ClusterSpec,
 	//
 	// But ... this is only the case from 1.16 on...
 	if kubernetesVersion.IsGTE("1.16") {
-		imageName += "-" + architecture
+		imageName += "-amd64"
```
johngmyers, line 172: Why hardcode amd64 here and below?

util/pkg/architectures/architectures.go
hakman: I agree, but as it is not used for now, I just removed it.
I will use the new name when I add the ARM64 pice.

pkg/model/components/context.go
hakman: This is already hardcoded to `amd64`:
https://github.com/kubernetes/kops/blob/bfd65ae589af979f0df9252b5269dcb5b3eb700b/pkg/model/components/context.go#L57-L59

Even more, this part of the code is used only for side-loading during CI, which works only with `amd64` images.

The approach taken for multi-arch support is to provide the list of assets for all supported architectures and let NodeUp decide which ones to download. At this point, it wouldn't really be possible to know which architecture would be used.

johngmyers: It's not immediately obvious to me why sideloading would only work with amd64. I presume making `LoadImageTask` not sideload the amd64 images on arm would be a later PR, but I don't see it in the WIP PR.

I would have thought we'd want to support arm-only CI builds, in which case we'd get `OptionsContext.Architecture()` to return the target architecture.

hakman: At the moment there are not CI binaries for arm64.
AMD64:
https://storage.googleapis.com/kubernetes-release-dev/ci/v1.19.0-beta.0.313+46d08c89ab9f55/bin/linux/amd64/kube-apiserver.tar
ARM64 (doesn't exists):
https://storage.googleapis.com/kubernetes-release-dev/ci/v1.19.0-beta.0.313+46d08c89ab9f55/bin/linux/arm64/kube-apiserver.tar

hakman: It will work though, by using `imageName += "-{{arch}}"` and later replacing it in NodeUp with the detected architecture.

johngmyers: But the different arch images will have different hashes.

johngmyers: It looks like this isn't for sideloading but for filling in the various `Image` fields in the full cluster spec. That entire concept is going to have to be adjusted, at least in the case of kube-proxy, for multi-arch.

johngmyers:
If it weren't for kube-proxy, we could keep a concept of "the arch for the masters", but as it is we'll have to figure out how to pick the right kube-proxy image for the node's arch.

/lgtm

hakman:
Rebasing after https://github.com/kubernetes/kops/pull/9258.

hakman:
Thanks @johngmyers !

johngmyers:
Now that I think about it, a package is likely to be good for either one architecture or all architectures. A slice doesn't make much sense.

hakman:
@johngmyers This may explain better my changes for the loading of multi-arch images:
https://github.com/hakman/kops/commit/74e51d6395c2036c1e9d02c0bb64538985b8aa96

At this point Kops will pull hashes for both AMD and ARM and forward them to NodeUp and NodeUp will choose which ones to load based on the "detected" architecture.

johngmyers:
Ah, `packageVersion.Architectures` is already a slice. I was saying it would make more sense if it were a single value with a reserved value for "all architectures".

hakman:
For distros we use the empty / null value as "all". I think we can live with it as is for now and optimise later more. I also doubt that the slice part will help much also.

rifelpet:
looks good!
/approve

hakman:
Thanks a lot for going through this @rifelpet 😄

PR #9217, opened 2020-05-30
Refactor and improve API validation

johngmyers:
/kind cleanup


olemarkus:
`validation.go` migth be a bit large too now, but can clean that up further later on.

johngmyers:
Probably split `validateNetworking()` and its dependencies into a separate file.

PR #9221, opened 2020-05-31
Spotinst: Allow a user specifiable node draining timeout

liranp:
This PR adds support for node draining timeout for both Elastigroup and Ocean.


MoShitrit:
/cc timoreimann
Just to make sure it looks good from DO perspective 😄

timoreimann:
@MoShitrit looks like this is about Spot Ocean? That'd be different waters than what I am used to swim in (DigitalOcean). 🙂

MoShitrit:
@timoreimann There are too many oceans around here 🤣 
Sorry for the false ping and thanks for letting me know!

timoreimann:
No worries at all! Glad we got it sorted out. 🙂👍

PR #9229, opened 2020-05-31
Put versioned API of cluster into state store

johngmyers:
This addresses two problems:

1) The "completed" cluster spec, stored in the file `cluster.spec` in the state store, is stored using the unversioned API. This is contrary to apimachinery rules, which state that APIs should be persisted in a versioned form of the API, and limits our ability to change the unversioned API.

2) The `cluster.spec` file is written at `kops edit` (etc.) time, causing some aspects of the cluster's operation to change prior to running `kops update cluster --yes`.

This PR starts writing the versioned API to a new `cluster-completed.spec`. This new file is only written at `kops update cluster --yes` time.

This PR fixes readers of the completed cluster spec to deserialize as the versioned API.

It also does not address storage of `cluster-completed.spec` through a k8s state store.

I am considering a future PR that would write the `cluster-completed.spec` file as a `ManagedFile`, replacing the VFS API method added here.


johngmyers:
This will have some effect on Terraform users, who currently expect to run `kops update cluster` without `--yes`. Unless we can get Terraform to update the state store we'll need to update the instructions.

johngmyers:
My plan to address the issue of the instancegroup specs being written prior to `kops update cluster --yes`, at least as far as nodeup, is to remove the `InstanceGroup` field from `NodeupModelContext`, moving all necessary information to the `NodeupConfig`.

johngmyers:
I'm putting a hold on this due to the Terraform issue. For now I'll pursue moving state into the `NodeupConfig`.

rifelpet:
> Unless we can get Terraform to update the state store we'll need to update the instructions.

Would managing the cluster and instance group manifests via terraform fix the issue? TF does have the [aws_s3_bucket_object](https://www.terraform.io/docs/providers/aws/r/s3_bucket_object.html) resource type. Rather than kops uploading the object to s3 directly it could write the object to the .tf files and have terraform manage them. Would that be sufficient? We wouldn't use it for any of the PKI secrets, just the manifests

johngmyers:
@rifelpet that would probably address the issue, modulo race conditions during the update itself.

johngmyers:
The current status of this project:

* Changing the readers of the completed cluster spec to a new file that is written at "update cluster --yes" time instead of "update cluster" time would adversely affect Terraform users, who don't run "update cluster --yes". It also exposes the problem that managed files get updated too early for all users. Fixing this is #9621, which is mired in an epic set of import cycles created by trying to push Terraform logic into VFS. I think the path there is to create a `TerraformWriter` type in a new package.
* There are currently no "completed instancegroup" specs. Instead of creating a set of these, I'm trying to remove the readers of these.
    * I've taken out most references from nodeup, leaving in #9294 the two things that might run up against the userdata size restrictions. I think the path there is to introduce a managedfile per launchtemplate version, so this also depends on resolving #9621 
    * Removing the reference from the node controller is #9575

olemarkus:
This PR looks largely good to me. We should get this in.

johngmyers:
This PR depends on #9294

olemarkus:
Still looks good to me

/lgtm

olemarkus (approved):

PR #9230, opened 2020-06-01
Clean up wording in releases.md

johngmyers:


docs/welcome/releases.md
```diff
@@ -31,15 +25,16 @@ Kubernetes version.
 
 
 Use the latest version of kops for all releases of Kubernetes, with the caveat
-that higher versions of Kubernetes are not _officially_ supported by kops. Releases who are ~~crossed out~~ _should_ work but we suggest should be upgraded soon.
+that higher versions of Kubernetes are not _officially_ supported by kops.
+Releases who are ~~crossed out~~ _should_ work, but we suggest they be upgraded soon.
```
MoShitrit, line 29: Would "Releases _which_ are..." suit better here?

docs/welcome/releases.md
```diff
@@ -1,23 +1,17 @@
...
+do our best to support previous releases for some period.
 
-One exception, in regard to compatibility, kops supports the equivalent
-Kubernetes minor release number.  A minor version is the second digit in the
-release number.  kops version 1.16.0 has a minor version of 16. The numbering
-follows the semantic versioning specification, MAJOR.MINOR.PATCH.
+Kops does not, however, support Kubernetes releases with a greater minor release number
+than that of kops. A minor version is the second digit in the
+release number: kops version 1.16.0 has a minor version of 16. The numbering
+follows the semantic versioning specification: MAJOR.MINOR.PATCH.
```
gjtempleton, line 11: Worth linking out to the semver spec here?

johngmyers: The semver spec would be highly parenthetical. The key point here is that it is the second part that counts.

Though now I notice it's not the second _digit_ and, pedantically, the major version number also matters.

PR #9215, opened 2020-05-30
Upgrade docker/containerd/containeros hashes to SHA256

johngmyers:

PR #9234, opened 2020-06-01
Add packages hashes verification for containerd and Docker

hakman:
Prow should verify package hashes for containerd and Docker.
Reference: https://github.com/kubernetes/kops/pull/9215#issuecomment-636327311

/cc @rifelpet

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -553,6 +553,10 @@ verify-terraform:
 verify-bindata:
 	./hack/verify-bindata.sh
 
+.PHONY: verify-hashes
+verify-hashes:
+	./hack/verify-hashes.sh
+
 # ci target is for developers, it aims to cover all the CI jobs
 # verify-gendocs will call kops target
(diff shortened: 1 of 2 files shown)


olemarkus:
Nice!
/lgtm

rifelpet:
it looks like it worked:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kops/9235/pull-kops-verify-hashes/1267438545979576320

based on the [prow docs](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md) i'm wondering if we could remove `optional: true` on the job and set `run_if_changed: nodeup/pkg/model/.*` ? that way it would always run if there are changes to that package

I think we'd also want `skip_report: false` to make it a blocking job if it is run for a PR, which is what [these docs](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#requiring-jobs-for-auto-merge-through-tide) indicate.

PR #9235, opened 2020-06-01
Don't build site when docs are unchanged

hakman:
https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,4 +2,4 @@
 [build]
 publish = "site"
 command = "make build-docs-netlify"
-ignore = "git diff --quiet HEAD^ HEAD docs/"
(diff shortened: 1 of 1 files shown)


hakman:
I did some quick tests and seems to work as expected.

rifelpet:
there might be changes outside of `docs/` that we'd want to trigger a netlify build: 

* `netlify.toml`
* `Makefile`
* `mkdocs.yml`
* `images/mkdocs/`

Any chance we can include those in the ignore condition?

rifelpet:
hijacking your PR to test your verify-hashes job :)

/test pull-kops-verify-hashes

rifelpet:
this PR does now build the docs site since it changes netlify.toml, so I think this is good to merge now 👍  thanks!

/approve

PR #9236, opened 2020-06-01
Fix some go-lint warning

Hellcatlk:
Signed-off-by: ZouYu <zouy.fnst@cn.fujitsu.com>

### DIFF
--- a/upup/pkg/fi/cloudup/dns.go
+++ b/upup/pkg/fi/cloudup/dns.go
@@ -35,7 +35,7 @@ import (
 )
 
 const (
-	// This IP is from TEST-NET-3
+	// PlaceholderIP is from TEST-NET-3
 	// https://en.wikipedia.org/wiki/Reserved_IP_addresses
 	PlaceholderIP  = "203.0.113.123"
 	PlaceholderTTL = 10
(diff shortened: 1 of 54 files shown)


upup/pkg/fi/cloudup/dns.go
```diff
@@ -35,11 +35,11 @@ import (
 )
 
 const (
-	// This IP is from TEST-NET-3
+	// PlaceholderIP is from TEST-NET-3
 	// https://en.wikipedia.org/wiki/Reserved_IP_addresses
 	PlaceholderIP  = "203.0.113.123"
 	PlaceholderTTL = 10
-	// DigitalOcean's DNS servers require a certain minimum TTL (it's 30), keeping 60 here.
+	// PlaceholderTTLDigitialOcean digitalOcean's DNS servers require a certain minimum TTL (it's 30), keeping 60 here.
```
johngmyers, line 42: These two don't correctly express the intent.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -61,7 +61,7 @@ type TemplateFunctions struct {
 	tags           sets.String
 }
 
-// This will define the available functions we can use in our YAML models
+// AddTo will define the available functions we can use in our YAML models
```
johngmyers, line 64:
```suggestion
// AddTo defines the available functions we can use in our YAML models.
```

Hellcatlk: Thanks, done.

upup/pkg/fi/cloudup/dns.go
```diff
@@ -35,7 +35,7 @@ import (
 )
 
 const (
-	// This IP is from TEST-NET-3
+	// PlaceholderIP is from TEST-NET-3
```
johngmyers, line 38: I intended the earlier comment about comments not correctly expressing the intent to also apply here. It seems minor enough, so I won't block on it.

PR #9248, opened 2020-06-02
More nodeup golden tests

justinsb:
Add tests for kube-controller-manager, kube-proxy, kube-scheduler

### DIFF
--- a/nodeup/pkg/model/kube_apiserver_test.go
+++ b/nodeup/pkg/model/kube_apiserver_test.go
@@ -21,13 +21,10 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/flagbuilder"
-	"k8s.io/kops/pkg/pki"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
-	"k8s.io/kops/util/pkg/vfs"
 )
 
 func Test_KubeAPIServer_Builder(t *testing.T) {
@@ -42,7 +39,9 @@ func Test_KubeAPIServer_Builder(t *testing.T) {
 		t.Fatalf("error loading model %q: %v", basedir, err)
 		return
 	}
-	nodeUpModelContext.KeyStore = &fakeKeyStore{T: t}
+	keystore := &fakeCAStore{}
+	keystore.T = t
+	nodeUpModelContext.KeyStore = keystore
 
 	builder := KubeAPIServerBuilder{NodeupModelContext: nodeUpModelContext}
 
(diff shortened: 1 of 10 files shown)


nodeup/pkg/model/kubelet_test.go
```diff
@@ -245,9 +269,26 @@ func RunGoldenTest(t *testing.T, basedir string, key string, builder func(*Nodeu
...
-	nodeupModelContext.KeyStore = &fakeKeyStore{T: t}
+
+	keystore := &fakeCAStore{}
+	keystore.T = t
+	keystore.privateKeys = map[string]*pki.PrivateKey{
+		"ca":                      mustParsePrivateKey(dummy_key),
+		"apiserver-aggregator-ca": mustParsePrivateKey(dummy_key),
+		"kube-controller-manager": mustParsePrivateKey(dummy_key),
+		"kube-proxy":              mustParsePrivateKey(dummy_key),
+		"kube-scheduler":          mustParsePrivateKey(dummy_key),
```
johngmyers, line 281: Would be slightly better to use different certs/keys to catch cases where the code used the wrong CA, but then it's not likely a human would catch the wrong key when reviewing the expected output.

justinsb: I don't disagree; I debated actually removing the keys entirely before comparison, but I figured that would be more confusing.

It's the flags and other pod settings (e.g. resources) that I think are particularly valuable in these tests.

hakman:
https://github.com/kubernetes/test-infra/pull/17805 should fix e2e.
/retest

nodeup/pkg/model/kubelet_test.go
johngmyers: I suppose I'd need to access `pki.IssueCert` through an interface that the tests would mock out.

nodeup/pkg/model/kube_apiserver_test.go
```diff
@@ -70,53 +71,63 @@ type fakeKeyStore struct {
...
 
 func (k fakeKeyStore) MirrorTo(basedir vfs.Path) error {
-	panic("implement me")
+	panic("fakeKeyStore does not implement MirrorTo")
 }
 
-func (k fakeKeyStore) FindCertificatePool(name string) (*fi.CertificatePool, error) {
-	panic("implement me")
+type fakeCAStore struct {
+	fakeKeyStore
```
johngmyers, line 93: Since you're changing all existing references to `fakeKeyStore` to `fakeCAStore`, why keep `fakeKeyStore` separate?

nodeup/pkg/model/kube_apiserver_test.go
```diff
@@ -70,53 +71,63 @@ type fakeKeyStore struct {
 	T *testing.T
 }
 
+var _ fi.Keystore = &fakeKeyStore{}
```
johngmyers, line 74: Not actually necessary, as this file is assigning it to a field of type `CAStore` and a `CAStore` embeds `Keystore`.

But perhaps this type should be pulled out into a separate file, since it is referenced from tests other than kube_apiserver?

nodeup/pkg/model/kubelet_test.go
justinsb: Ah - I have two ideas on that front then. 

* If you've put cert generation into a task (which isn't a bad idea, you might be doing that already) then I think we'll be testing the right thing - the task invocation arguments, not the filesystem write calls.

* Otherwise we can post-process the serialized yaml and either remove the keys or replace them with their key characteristics (e.g. "CN=foo, Validity=Client")

Are you planning on putting cert generation into a task?

nodeup/pkg/model/kube_apiserver_test.go
```diff
@@ -43,7 +42,9 @@ func Test_KubeAPIServer_Builder(t *testing.T) {
 		t.Fatalf("error loading model %q: %v", basedir, err)
 		return
 	}
-	nodeUpModelContext.KeyStore = &fakeKeyStore{T: t}
+	keystore := &fakeCAStore{}
+	keystore.T = t
+	nodeUpModelContext.KeyStore = keystore
```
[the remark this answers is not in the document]
justinsb: Good question - I assume it was left over from some meandering along the way.  Cleaned up!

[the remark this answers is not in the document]
justinsb: Edit: because of the embedded type, it would be `&fakeCAStore{fakeKeyStore: fakeKeyStore{t}}`

nodeup/pkg/model/kube_apiserver_test.go
justinsb: I'm not entirely happy with the dual CAStore / KeyStore structure, and I'm hoping we can clean it up over time.  I didn't want to add another thing that would need to be untangled later.

nodeup/pkg/model/kube_apiserver_test.go
justinsb: I like to leave the assignment because it documents that the struct should implement the interface (i.e. it is for reading).  That could reveal my language history though ;-)

Agree that it can be pulled out into a separate file - did that, and added some comments, including on the split.

nodeup/pkg/model/kubelet_test.go
johngmyers: #9172

nodeup/pkg/model/kube_apiserver_test.go
johngmyers: The only thing that takes just a `KeyStore` are `BuildKubecfg` and `fi.Context`. One thing using the field in `fi.Context` upcasts it to a `VFSCAStore`. (The other is `fitasks.Keypair`.) I don't think the interface split is well-designed.

nodeup/pkg/model/kubelet_test.go
justinsb: I'll take a look at #9130, to get closer to #9172.  Your approach looks good.  I don't know exactly how the TaskDependentResource will get serialized, but I imagine we can make it work nicely!

We can always defer this until #9172 has landed ... we've gotten by without it so far!

johngmyers:
/hold
in case you want another round. Feel free to cancel.

nodeup/pkg/model/kube_apiserver_test.go
justinsb: I do agree with that - it has resisted my attempts at making them cleaner (thus far)

johngmyers: It comes down to sharing `fi.Context` between nodeup and cloudup. Two of the three methods in `KeyStore` are only needed by specific cloudup tasks.

PR #9250, opened 2020-06-02
Update channels for 1.18

mikesplain:


channels/stable
```diff
@@ -55,6 +55,9 @@ spec:
     networking:
       kubenet: {}
   kubernetesVersions:
+  - range: ">=1.18.0-alpha.1
+    recommendedVersion: 1.18.0-beta.1
+    requiredVersion: 1.18.0-alpha.1
```
rifelpet, line 60: should these be:
```suggestion
    recommendedVersion: 1.18.3
    requiredVersion: 1.18.0
```
?

mikesplain: Doh, thank you

mikesplain: Fixed thanks. Too much on the clipboard.

PR #9252, opened 2020-06-03
Remove redundant ValidateInstanceGroup call

rifelpet:
The `cloudup.PopulateInstanceGroupSpec` directly after this calls `ValidateInstanceGroup` so this first call is redundant.

https://github.com/kubernetes/kops/blob/7b067983dfab5e78148dc9906242bed29efed455/upup/pkg/fi/cloudup/populate_instancegroup_spec.go#L61-L66

This is minor cleanup to help simplify the aws instance type validation PR


johngmyers:
Indeed.
/lgtm

PR #9255, opened 2020-06-03
Remove romana support

olemarkus:


docs/releases/1.19-NOTES.md
```diff
@@ -6,6 +6,8 @@
 
 # Breaking changes
 
+* Removed support for Romana networking provider.
```
johngmyers, line 9:
```suggestion
* Support for the Romana networking provider has been removed.
```

pkg/apis/kops/validation/validation.go
```diff
@@ -367,6 +367,7 @@ func validateNetworking(c *kops.ClusterSpec, v *kops.NetworkingSpec, fldPath *fi
 	}
 
 	if v.Romana != nil {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("romana"), "support for Romana has been removed"))
 		if optionTaken {
```
johngmyers, line 371: This if block can be removed

olemarkus: Can it? If I remove that block, kops creates my cluster.

johngmyers: A validation error should prevent creation. How is that happening?

olemarkus: Sorry, I thought you were referring to the outer if. Yeah, this one can technically be removed.

rifelpet:
this looks good to me too. i'm curious if/when we'd want to remove the `RomanaNetworkingSpec` type from the API altogether. If we remove it now, kops will error saying the field isnt recognized. I suppose having an explicit validation error is clearer to the user, but it might be a good idea to come up with a plan to remove the type altogether.

olemarkus:
We can remove the struct from the next API version, but I think we should keep the unversioned struct for as long as any versioned API has the struct so that people run into the validation errors.

justinsb:
Agree with keeping the API fields, to explicitly mark it deprecated and help users.  Also makes it easier if we change our mind and re-add support!

PR #9256, opened 2020-06-03
Use CNI 0.8.6 for Kubernetes 1.15+

hakman:
Due to security advisory https://github.com/kubernetes/kubernetes/issues/91507, k/k upgraded the CNI binaries to 0.8.6 for 1.16+. We still support 1.15, so we should do it for it also.
Refs:
* https://github.com/kubernetes/kubernetes/pull/91370
* https://github.com/kubernetes/release/pull/1309

### DIFF
--- a/upup/pkg/fi/cloudup/networking.go
+++ b/upup/pkg/fi/cloudup/networking.go
@@ -39,13 +39,12 @@ const (
 	defaultCNIAssetHashStringK8s1_9 = "d595d3ded6499a64e8dac02466e2f5f2ce257c9f"
 
 	// defaultCNIAssetK8s1_11 is the CNI tarball for k8s >= 1.11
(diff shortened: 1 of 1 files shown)


rifelpet:
hm, so existing k8s 1.15-1.17 clusters will be upgraded from 0.7.5 to 0.8.6 when they upgrade to the version of Kops that includes this? Are there any concerns with that upgrade process?

hakman:
@rifelpet The idea was to migrate as much as possible to 0.8.6, probably including Kops 1.16.
I see that k/k set it as the default for 1.16+ so probably will not hurt 1.15 either.

If we merge it, we can see how the weekly e2e are behaving with older versions also.

rifelpet:
ok, yes we can merge this and keep an eye on the 1.15 job before we release the next kops version that includes this 👍 

https://testgrid.k8s.io/kops-versions#kops-aws-k8s-1.15


/approve

PR #9257, opened 2020-06-03
Add unit test for func isKopsManagedDNSName in pkg/resources/gce/gce.go

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/resources/gce/gce_test.go
+++ b/pkg/resources/gce/gce_test.go
@@ -62,3 +62,35 @@ func TestNameMatch(t *testing.T) {
 		}
 	}
 }
+
+func TestIsKopsManagedDNSName(t *testing.T) {
+	grid := []struct {
+		name     string
+		expected bool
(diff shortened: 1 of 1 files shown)


johngmyers:
I see two problems with this:

1) The grid does not do a good job of covering the combinatorics. While it hits all the possibilities for both prefix and suffix, it does not cover the possibility of `bastion` missing from the prefix list. It is not necessary to cover every possible combination, but each axis should have some coverage that is independent of other variables.

2) This is testing a private interface. That is somewhat justifiable if testing the public interface is unwieldy or impractical, but in this case even if testing through `gce.ListResourcesGCE()` is impractical then testing through `listGCEDNSZone()` doesn't appear to be. (A quick look suggests to me that testing through `gce.ListResourcesGCE()` would be practical.)

/ok-to-test

PR #9258, opened 2020-06-03
Use Docker 19.03.11 for Kubernetes 1.18+

hakman:
Due to security advisory kubernetes/kubernetes#91507, Docker 19.03.11 was released.
We may want to add and set this version as the default in 1.17 and 1.16 also.

Refs:
* https://github.com/docker/docker-ce/releases/v19.03.11

### DIFF
--- a/docs/releases/1.18-NOTES.md
+++ b/docs/releases/1.18-NOTES.md
@@ -4,7 +4,7 @@
 
 # Significant changes
 
-* The default Docker version has been changed to 19.03.8.
+* The default Docker version has been changed to 19.03.11.
 
 * Support for [RHEL 8](../operations/images.md#rhel-8) and [CentOS 8](../operations/images.md#centos-8) has been added.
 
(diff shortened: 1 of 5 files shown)


rifelpet:
Is there a k/k PR that matches this? I couldn't fine one.

just curious

hakman:
@rifelpet k/k didn't upgrade docker, but they mentioned in the advisory that this is the only unaffected version.
They only reference the major.minor version and not the patch one.

rifelpet:
Ah thats right, I forgot k/k only deals with minor versions. Maybe Kops should consider doing something similar? Allow the `dockerVersion` field to only reference `major.minor` and under the hood it always uses the latest `patch`? Something to think about, perhaps for the v1beta1 apiVersion...

/approve

rifelpet:
or the `dockerVersion` field supports both `major.minor.patch` and `major.minor` (in which case we always use the latest patch) and we somehow encourage users to use `major.minor`. That would avoid making it a breaking change.

hakman:
We don't really break anything, we only set the default version and that should be backwards compatible at least with those with same major & minor.
We still provide the older versions, but the user has to set it manually if he doesn't want what we choose.

rifelpet:
True. Maybe we add support for `major.minor` in the dockerVersion field and have kops always use the latest patch version in that case. Does `create cluster` populate the field? If it does we could have it only use `major.minor` too.

hakman:
Create cluster does not populate the version by default. It is up to us to set the default.
I like the idea though of major and minor instead of patch for the future.

PR #9265, opened 2020-06-03
Validate IG RootVolumeType

### LINKED ISSUE #5807: Add a warning message for unsupported volume types (AWS) for boot volumes
While cluster creation, if we specify boot volume types as `st1` or `sc1` for nodes, the cluster creation continues but doesn't create desired nodes as `st1` and `sc1` are not supported by AWS for creating boot volumes.

In such cases, a friendly warning message mentioning this fact would be nice so that a user can understand that these volume types cannot be used as boot volumes and users won't keep wondering why kops is not able to create required nodes.

Same goes while updating instance groups' volume types from `gp2` -> `st1`. `kops` continues with this change and even shows something like this:

```
Will modify resources:
  LaunchConfiguration/nodes.xyz.k8s.local
  	RootVolumeSize      	 128 -> 500
  	RootVolumeType      	 gp2 -> st1
```
And if at this point, a user is unaware of the above-mentioned fact, he'll likely continue with `kops update cluster --yes` which also wouldn't show anything suspicious. The error comes only when we do a rolling update.

Instead if we can have a warning at the very beginning, it will be very helpful and will save some time as well.

For more info, see #4256

olemarkus:
Fixes #5807

### DIFF
--- a/pkg/apis/kops/validation/instancegroup.go
+++ b/pkg/apis/kops/validation/instancegroup.go
@@ -65,6 +65,10 @@ func ValidateInstanceGroup(g *kops.InstanceGroup) field.ErrorList {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "rootVolumeIops"), g.Spec.RootVolumeIops, "RootVolumeIops must be greater than 0"))
 	}
 
+	if g.Spec.RootVolumeType != nil {
+		allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "rootVolumeType"), g.Spec.RootVolumeType, []string{"gp2", "io1"})...)
+	}
+
 	// @check all the hooks are valid in this instancegroup
 	for i := range g.Spec.Hooks {
 		allErrs = append(allErrs, validateHookSpec(&g.Spec.Hooks[i], field.NewPath("spec", "hooks").Index(i))...)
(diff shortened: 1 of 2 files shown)


rifelpet:
I wonder if this build off of https://github.com/kubernetes/kops/pull/8856 to validate that the field matches the `SupportedRootDeviceTypes` field from DescribeInstanceTypes for the specified instance type. It might get messy with mixedInstancesPolicy though.

olemarkus:
It is a thought, but compared to instance types and such, EBS types are more manageable. They have proven to not change too often.

rifelpet:
thats true 👍

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -65,6 +65,10 @@ func ValidateInstanceGroup(g *kops.InstanceGroup) field.ErrorList {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "rootVolumeIops"), g.Spec.RootVolumeIops, "RootVolumeIops must be greater than 0"))
 	}
 
+	if g.Spec.RootVolumeType != nil {
```
rifelpet, line 68: do we know if this field is only used by AWS? if there are other cloud providers with different allowed values, we might want to only perform this validation if this is an AWS IG.

rdrgmnzs: +1 a quick search shows that aliclound and gcp both use the field.

olemarkus: Good catch. Ali also uses this as well.

PR #9266, opened 2020-06-03
Update channel 1.15 k8s recommendation to 1.15.12

jeffb4:
The latest patch release in the 1.15 series is 1.15.12, released 29 days ago, with plenty of bugfixes


johngmyers:
Usually we update the alpha channel first, then stable a week or two later.

jeffb4:
@johngmyers reverted the stable change

PR #9267, opened 2020-06-04
Don't make dev uploads public

johngmyers:


johngmyers:
Hrm, kops is trying to pull the assets over the unauthenticated public URL. I guess I can't test nodeup changes.

hakman:
Would be nice to have at least the S3 target without `--public-read`. I use private with CloudFront in front for some that need public http access and this pretty much breaks the upload.

https://github.com/kubernetes/kops/blob/5476fc871d84a7a36d7927e669e6e99a8d96de72/Makefile#L324

johngmyers:
Taking public-read off of the S3 target resulted in inability to download, at least for the bucket I tried it with. Perhaps I need different settings to allow https access only from within the account?

justinsb:
Yes, nodeup needs to be public so it can be pulled from the bash script we use on boot.  It would be really nice if we could figure out how to mirror everything to a private S3 bucket; we could try using the aws cli (`aws s3 cp`) instead of wget (we'd also need IAM permissions at the node level).  Another option is a signed S3 url, but that is only valid for a week so probably not a good option for production.

I suspect that unless you're working on something super-secret, that just pushing to a random prefix is reasonable?  (e.g. `s3://mybucket/the-combination-is-1-2-3-4-5/`)

johngmyers:
I need to go through a complicated policy exception process in order to get permission to write to a public bucket.

rdrgmnzs:
I had the same issue, I just ended up modifying the upload script internally to not push publicly and adding a proxy in front of S3 so that we could keep all the objects in the bucket private.

PR #9268, opened 2020-06-04
Update Calico and Canal for CVE-2020-13597

hakman:
Due to security advisory kubernetes/kubernetes#91507, new versions of Calico were released.

Ref: https://www.projectcalico.org/security-bulletins/

/assign @KashifSaadat


hakman:
Thanks @KashifSaadat. I will create the cherry-picks once this is merged.
/retest

Issue #8264, opened 2020-01-03
Feature - Support AWS IAM Roles for Service Accounts


rifelpet:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

AWS added support for IAM Roles using K8s service accounts, and I think it would be great if Kops could handle configuring this automatically. See the [blog post](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) and [eks docs](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html)

Kops can essentially follow the instructions here:

https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/SELF_HOSTED_SETUP.md

I have a WIP branch that will expose this through a set of API fields, but there are a few design decisions we'll need to make:

1. The public signing key needs to be publicly accessible through a URL. Options:
   - Upload it to the S3 state store bucket with an ACL allowing public read access. Having specific files be publicly accessible in a bucket that also contains sensitive objects could be concerning for users, or even disallowed by security requirements.
   - Have kops create a second bucket that is publicly accessible only used for hosting this file. This will require adding a new S3 bucket task to cloudup.
   - Host it elsewhere - API Gateway? (more new cloudup tasks) Or provide the file to the user for them to host however they want? The user would need to provide the resulting URL in the cluster spec so that Kops can add it to kube-apiserver's `--service-account-issuer`.

2. I think it would be good for Kops to create the OIDC provider for the cluster, since according to [this comment](https://github.com/aws/amazon-eks-pod-identity-webhook/issues/23#issuecomment-565109745) an OIDC provider can only be used with one cluster, so in most cases it would need to be created regardless. If a user wants to provide their own OIDC provider we could support that as well and optionally have Kops not create one. In my case, we'll be using an OIDC provider in a different AWS account as the cluster, but I dont think Kops needs to be concerned about that at all. 

3. It's not immediately clear to me, but I think the keypair that needs to be generated needs signing permissions, so should we reuse the existing cluster CA or create a new one just for this purpose?

4. This is minor, but I'm not sure exactly what we should name the API fields because I'm not entirely sure what the name of the feature is :) I see mentions of `IAM Roles for Service Accounts`, `Pod Identity Webhook`, I think the former is the more official name but might be tricky to fit in an API field name.

[...]

PR #9271, opened 2020-06-04
[WIP] Add API fields for AWS OIDC Provider

rifelpet:
This PR allows Kops to create the IAM OIDC Provider resource, and discovery.json and keys.json files necessary to setup IAM Roles for Service Accounts. As discussed in the linked issue, initial implementation will create the json files and require that the user copies them to their chosen url (specifics don't matter: could be a public s3 bucket, static site, etc.)

The rough workflow will be:
1. User chooses a url to host the json files and determines the CA Thumbprint for that url
2. Add thumbprint and url to cluster spec
3. Run `kops update cluster --yes`
4. Copy resulting json files from state store to their url
5. Update any IAM roles' assume role policies to allow assumption via the new OIDC Provider

TODO for this PR:
* Create keys.json
* Pull the service-account-signer keypair down on masters and pass via flags to kube-controller-manager
* Fix integration test (it runs both terraform and cloudformation but this feature isnt supported by cloudformation)
* Add docs
* Figure out how we can have the Lister and Deleter work in pkg/resources/aws - these are used during cluster tear down. We're only given a cluster name (not the full cluster spec) and oidc providers don't support naming nor tagging, so theres no way for us to attribute an oidc provider back to a cluster without knowing an issuer url or thumbprint.

Roadmap for future PRs:
* Add the admission webhook to bootstrapchannelbuilder.go - ~Currently there is [no public docker image](https://github.com/aws/amazon-eks-pod-identity-webhook/issues/5#issuecomment-592787878) of the webhook.~ EDIT: [images are now available here](https://hub.docker.com/r/amazon/amazon-eks-pod-identity-webhook). There may also be hurdles with the webhook's CA bundle - This would be the first occurrence of a webhook in bootstrapchannelbuilder.go and we have to figure out how to generate the CA bundle while also making it available to boostrapchannelbuilder.go which I think means we can't use fitasks.Keypair. Alternatively [we use a CSR and have kops-controller approve it](https://github.com/kubernetes/kops/issues/8264#issuecomment-620398192).
* Determine a reasonable way to publicly host the json files. This would simplify the workflow by not having the user copy them to a url themselves and also not having to determine the CA thumbprint.

ref: https://github.com/kubernetes/kops/issues/8264


pkg/model/awsmodel/oidc_provider.go
```diff
@@ -0,0 +1,110 @@
...
+	discoveryContents := fmt.Sprintf(discoveryJSON, issuerURL, issuerURL)
+	discoveryFile := &fitasks.ManagedFile{
+		Contents:  fi.WrapResource(fi.NewStringResource(discoveryContents)),
+		Lifecycle: b.Lifecycle,
+		Location:  fi.String("discovery.json"),
+		Name:      fi.String("discovery.json"),
+	}
+	c.AddTask(discoveryFile)
+
+	// TODO create keys.json from https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/hack/self-hosted/main.go
```
justinsb, line 86: The other thing we can do here is to get them from the apiserver.  The big advantage of this is that I understand we're (eventually) planning on rotating the keys in apiserver, and it would be good to coordinate with that!

pkg/resources/aws/aws.go
```diff
@@ -2102,6 +2103,68 @@ func ListIAMInstanceProfiles(cloud fi.Cloud, clusterName string) ([]*resources.R
...
+		for _, provider := range response.OpenIDConnectProviderList {
+			arn := provider.Arn
+			descReq := &iam.GetOpenIDConnectProviderInput{
+				OpenIDConnectProviderArn: arn,
+			}
+			_, err := c.IAM().GetOpenIDConnectProvider(descReq)
+			if err != nil {
+				return nil, fmt.Errorf("error getting IAM OIDC Provider: %v", err)
+			}
+			// TODO: only delete oidc providers if they're owned by the cluster.
```
justinsb, line 2125: If the issuer url was 1:1 with the cluster though, I think we could be reasonably sure?

pkg/resources/aws/aws.go
```diff
@@ -2102,6 +2103,68 @@ func ListIAMInstanceProfiles(cloud fi.Cloud, clusterName string) ([]*resources.R
...
+			// We need to figure out how we can determine that given only a cluster name.
```
justinsb, line 2126: If the issuer url was 1:1 with the cluster though, I think we could be reasonably sure?

pkg/resources/aws/aws.go
rifelpet: the issuer url is 1:1 with the cluster, but we only have access to the cluster name here, and the initial plan was to require the user host the json files themselves at the url, so the url may be totally unrelated to the cluster name even if they are 1:1.

I was imagining a common pattern would be to use an externally-managed public s3 bucket, making the issuer url something like `https://s3.amazonaws.com/peters-oidc-provider-bucket/cluster-foobar/`. It's unique to the cluster but if we only have the cluster name i dont think we can guarantee that we've identified a provider to delete unless we document additional requirements like "the url's path must contain the full cluster name".

I proposed some alternative implementations we could explore down the road in #8264. I'm aware of https://github.com/kubernetes/enhancements/issues/1393 and i think it would be good to integrate, perhaps in this case kops could setup an `oidc` dns record within the cluster's hosted zone that points to a dedicated load balancer targeting the new apiserver endpoints. If we can use that as the issuer url, it not only lets us identify the issuer purely from the cluster name but also lets us compute both the issuer url and the CA thumbprint ourselves so the user doesn't need to provide either. This setup would require an internet-accessible load balancer and the cluster using a public hosted zone, but I think thats a reasonable requirement. Is this what you had in mind?

pkg/apis/kops/validation/validation.go
```diff
@@ -855,3 +860,40 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
...
+		allErrs = append(allErrs, field.Forbidden(path, "serviceOIDCProvider is supported only in AWS"))
+	}
+
+	if provider.Issuer == "" {
+		allErrs = append(allErrs, field.Required(path.Child("issuer"), ""))
+	} else {
+		// Based on these requirements https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html#manage-oidc-provider-console
+		url, err := url.ParseRequestURI(provider.Issuer)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(path.Child("issuer"), provider.Issuer, "must be a URL"))
```
johngmyers, line 879: Include `err.Error()` in detail? (or make the detail `err.Error()`)

pkg/apis/kops/validation/validation.go
```diff
@@ -855,3 +860,40 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
...
+		}
+		if url.Scheme != "https" {
```
johngmyers, line 881: Scheme and port should be in an else block. They should not cascade if the URL didn't parse.

pkg/apis/kops/validation/validation_test.go
```diff
@@ -801,3 +801,69 @@ func Test_Validate_NodeLocalDNS(t *testing.T) {
...
+			CloudProvider:  "gce",
+			ExpectedErrors: []string{"Forbidden::testField"},
+		},
+		{
+			Input: kops.ServiceOIDCProviderSpec{},
+			ExpectedErrors: []string{
+				"Required value::testField.issuer",
+				"Required value::testField.issuerCAThumbprints",
+			},
+		},
```
johngmyers, line 834: Add case for malformed URL

pkg/model/awsmodel/oidc_provider.go
```diff
@@ -0,0 +1,110 @@
...
+
+	saSigner := &fitasks.Keypair{
+		Name:      fi.String("service-account-signer"),
+		Lifecycle: b.Lifecycle,
+		Subject:   "cn=service-account-signer",
+		Type:      "ca",
+	}
+	c.AddTask(saSigner)
+
+	discoveryContents := fmt.Sprintf(discoveryJSON, issuerURL, issuerURL)
```
johngmyers, line 77: What if issuerURL has a `\` or `"` in it? Instead of using a `discoveryJSON` string, I prefer to build some Go object then serialize it to JSON.

pkg/model/awsmodel/oidc_provider.go
```diff
@@ -0,0 +1,110 @@
...
+	if provider == nil {
+		return nil
+	}
+
+	issuerURL := provider.Issuer
+
+	saSigner := &fitasks.Keypair{
+		Name:      fi.String("service-account-signer"),
+		Lifecycle: b.Lifecycle,
+		Subject:   "cn=service-account-signer",
```
johngmyers, line 72: `cn=service-oidc-ca`?

pkg/model/awsmodel/oidc_provider.go
```diff
@@ -0,0 +1,110 @@
...
+func (b *OIDCProviderBuilder) Build(c *fi.ModelBuilderContext) error {
+	provider := b.Cluster.Spec.ServiceOIDCProvider
+	if provider == nil {
+		return nil
+	}
+
+	issuerURL := provider.Issuer
+
+	saSigner := &fitasks.Keypair{
+		Name:      fi.String("service-account-signer"),
```
johngmyers, line 70: `service-oidc-ca`? I'd prefer all CA's have names ending with `-ca`.

pkg/resources/aws/aws.go
```diff
@@ -2102,6 +2103,68 @@ func ListIAMInstanceProfiles(cloud fi.Cloud, clusterName string) ([]*resources.R
...
+func DeleteIAMOIDCProvider(cloud fi.Cloud, r *resources.Resource) error {
+	c := cloud.(awsup.AWSCloud)
+	arn := r.Obj.(*string)
+	{
+		klog.V(2).Infof("Deleting IAM OIDC Provider %v", arn)
+		request := &iam.DeleteOpenIDConnectProviderInput{
+			OpenIDConnectProviderArn: arn,
+		}
+		_, err := c.IAM().DeleteOpenIDConnectProvider(request)
+		if err != nil {
```
johngmyers, line 2160: Return `nil` if error was "NoSuchEntity".

pkg/resources/aws/aws.go
johngmyers: I think you have to match on the URL, possibly in combination with the ClientIDs.

pkg/apis/kops/validation/validation.go
```diff
@@ -855,3 +860,41 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
...
 	return allErrs
 }
+
+func validateServiceOIDCProvider(c *kops.ClusterSpec, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	thumbprintFormat := regexp.MustCompile(`^[a-fA-F0-9]{40}$`)
+	provider := c.ServiceOIDCProvider
+
+	if kops.CloudProviderID(c.CloudProvider) != kops.CloudProviderAWS {
+		allErrs = append(allErrs, field.Forbidden(path, "serviceOIDCProvider is supported only in AWS"))
```
johngmyers, line 870: Should probably return here

pkg/apis/kops/cluster.go
```diff
@@ -191,6 +191,8 @@ type ClusterSpec struct {
 	SysctlParameters []string `json:"sysctlParameters,omitempty"`
 	// RollingUpdate defines the default rolling-update settings for instance groups
 	RollingUpdate *RollingUpdate `json:"rollingUpdate,omitempty"`
+	// ServiceOIDCProvider defines the OIDC provider setup for the cluster (AWS only)
+	ServiceOIDCProvider *ServiceOIDCProviderSpec `json:"serviceOIDCProvider,omitempty"`
```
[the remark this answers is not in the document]
justinsb: Yeah - naming for this is going to be tough.

* It's not technically specific to AWS; if we publish the JWKS doc from e.g. an on-metal cluster we should be able to authenticate to AWS.
* Similarly GCE has some support for this now also, in the form of GKE workload identity, so it's not just _to_ AWS.
* It's actually even broader than this; we can also authenticate service accounts across clusters etc.

The word that keeps sticking in my mind is "discovery", and this is specifically "identity discovery".  I think it's also possible to do cross-cluster discovery using that same bucket (and those files wouldn't necessarily have to be public, either)

Maybe something like

```
discovery:
  store: s3://some-other-bucket/clusters
  identity: true
```

I think technically we could even reuse the same S3 bucket, I can't imagine it's worth it from an isolation point of view though :-)

justinsb:
FYI I just sent #9341 which extends this PR, adding the idea of automatically publishing the JWKS document to a bucket.  (I expect it to be a different bucket, because the JWKS documents themselves need to be publicly readable)

This has clarified things a bit in my mind; there really are two things going on here:

1) Publishing the JWKS document publicly, which enables consumers to trust serviceaccounts
2) Registering the cluster/jwks with consumers, like AWS, so that they recognize k8s service accounts.

I do think we want to do both, but it can make sense to do the second even when not running on AWS, or to register more sources (GCP, Azure, maybe other services in future like github?)

In #9341 I defaulted the AWS registration, BTW (when a feature flag UsePodIAM is passed).  We might therefore be able to get away without any API at all in this PR (though we do want it in the end, we might be able to separate the discussions).

pkg/apis/kops/cluster.go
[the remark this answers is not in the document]
johngmyers: I don't know how things other than AWS consume OIDC, but I would hope the `IssuerCAThumbprints` field wouldn't be copied by other cloud providers. It would be a lot more sane if one could just configure AWS with the identity document directly, rather than faffing about with serving it on the public Internet. If AWS ever changes the CA they use for their S3 buckets, everyone's IRSA will likely break immediately.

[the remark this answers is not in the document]
justinsb: It's a good point on the rotations - I presume this means that S3 effectively can't change CAs quickly, though I don't know what happens in 2025 when the current cert expires, presumably some pre-publication.  (I think the cert fingerprint includes the expiration date?)

@rifelpet and I chatted on Friday, and we had an idea about an approach that might be a nice way to get this merged: we define a feature flag (or two) that effectively does all this automagically in a (very) limited set of circumstances - currently you must use a public ELB, with a real DNS name, likely others we don't even know about.  The featureflag enables the apiserver with anonymous-access turned on (which normally we don't like to do in kops, particularly for public ELBs), we also grant the anonymous-access user permission to read the JWKS documents directly.

We should need to find a way to not encourage this as a real scenario (e.g. I don't want to preclude private ELBs).  But it means we don't need an API surface in this PR and we don't need kops-controller to publish to an S3 bucket in the one I added.  That PR is https://github.com/kubernetes/kops/pull/9352

We can then discuss e.g. whether to have a public bucket, how to name the clusters, what the API should be for enabling AWS integration in a more lesiurely way.  But we don't block working on e.g. per-pod roles or figuring out different ways to publish the JWKS doc (e.g. publish from kops CLI vs publish from kops-controller).

This also means that upstream (k/k) can start to see some real test coverage of the feature, and get it to GA.

johngmyers:
I think creating the JWKS document and publishing it publicly are separable. While publishing to an S3 bucket is convenient (and, incidentally, is what we do) there are other possible mechanisms.

rifelpet:
rebased with https://github.com/kubernetes/kops/pull/9375 merged, though I'm not sure how applicable much of this will be if we choose to have kops-controller publish the jwks documents to an s3 bucket rather than having the kops cli do it.

rifelpet:
to be replaced by #10756

PR #9274, opened 2020-06-05
Ensure we have IAM bucket permissions to other S3 buckets

justinsb:
If we are expected to write to other buckets, we need to have suitable
permissions to e.g. determine their location.


justinsb:
This will be useful both if we want to split the etcd backup bucket or if we want to auto-upload the JWKS documents to a public bucket as in #9271 (cc @rifelpet )

johngmyers:
Would have been good to add a test that would have failed without the code change.

PR #9279, opened 2020-06-05
Allow failure of the ARM64 job in TravisCI

hakman:
The ARM64 test in TravisCI seems flaky and should be allowed to fail for now until we have a chance to see if anything can be improved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ os:
   - linux
   - osx
 go:
-  # 1.12 is no longer supported by apimachinery; we recommend 1.13
   - 1.13.x
   - 1.14.x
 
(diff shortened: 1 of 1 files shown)


rifelpet:
is there an example failure you can link to? just curious

johngmyers:
@rifelpet https://travis-ci.org/github/kubernetes/kops/builds/694863847

johngmyers:
/lgtm
(Though I don't think removing the OSX comment is necessary.)

hakman:
Also this: https://travis-ci.org/github/kubernetes/kops/builds/694878314
Had to restart this test multiple times this morning to get it to pass. May be just a problem with TravisCI infra, but not that important for not to block merging.

PR #9280, opened 2020-06-05
Specify user on export kubecfg

### LINKED ISSUE #8017: Environment variable to control exported user
It would be great to have an environment variable to be able to configure the user kops exports.

We have RBAC setup with Google OIDC for auditing purposes and when we run commands like `kops replace` or `kops export kubecfg`, it overwrites the current kube config context with the default certificate based admin user. Then any commands we run with kubectl show up as the admin user rather than the individual users.

In this case, it would be great to have kops not add the admin user to the kubecfg file and just use the value from the environment variable in the context block. The environment variable could be something like `KOPS_KUBECFG_USER`

olemarkus:
* Don't export admin user by default.
* Allow specifying existing user when exporting context.

Fixes #8017

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -1360,13 +1360,11 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 		updateClusterOptions.Target = c.Target
 		updateClusterOptions.Models = c.Models
 		updateClusterOptions.OutDir = c.OutDir
+		updateClusterOptions.admin = true
+		updateClusterOptions.CreateKubecfg = true
 
 		// SSHPublicKey has already been mapped
 		updateClusterOptions.SSHPublicKey = ""
-
-		// No equivalent options:
-		//  updateClusterOptions.MaxTaskDuration = c.MaxTaskDuration
-		//  updateClusterOptions.CreateKubecfg = c.CreateKubecfg
 
 		_, err := RunUpdateCluster(ctx, f, cluster.Name, out, updateClusterOptions)
 		if err != nil {
@@ -1388,7 +1386,7 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 				fmt.Fprintf(&sb, " * edit your master instance group: kops edit ig --name=%s %s\n", cluster.Name, masters[0].ObjectMeta.Name)
 			}
 			fmt.Fprintf(&sb, "\n")
-			fmt.Fprintf(&sb, "Finally configure your cluster with: kops update cluster --name %s --yes\n", cluster.Name)
+			fmt.Fprintf(&sb, "Finally configure your cluster with: kops update cluster --name %s --yes --admin\n", cluster.Name)
 			fmt.Fprintf(&sb, "\n")
 
 			_, err := out.Write(sb.Bytes())
--- a/cmd/kops/export_kubecfg.go
+++ b/cmd/kops/export_kubecfg.go
@@ -34,14 +34,17 @@ import (
 
 var (
 	exportKubecfgLong = templates.LongDesc(i18n.T(`
-	Export a kubecfg file for a cluster from the state store. The configuration
-	will be saved into a users $HOME/.kube/config file.
-	To export the kubectl configuration to a specific file set the KUBECONFIG
-	environment variable.`))
+	Export a kubecfg file for a cluster from the state store. By default the configuration
+	will be saved into a users $HOME/.kube/config file. Kops will respect the KUBECONFIG environment variable
+	if the --kubeconfig flag is not set.
+	`))
 
 	exportKubecfgExample = templates.Examples(i18n.T(`
-	# export a kubecfg file
-	kops export kubecfg kubernetes-cluster.example.com
+	# export a kubeconfig file with the cluster admin user (make sure you keep this user safe!)
+	kops export kubecfg kubernetes-cluster.example.com --admin
+
+	# export using a user already existing in the kubeconfig file
+	kops export kubecfg kubernetes-cluster.example.com --user my-oidc-user
 		`))
 
 	exportKubecfgShort = i18n.T(`Export kubecfg.`)
@@ -50,6 +53,8 @@ var (
 type ExportKubecfgOptions struct {
 	KubeConfigPath string
 	all            bool
+	admin          bool
+	user           string
 }
 
 func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
@@ -69,8 +74,10 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&options.KubeConfigPath, "kubeconfig", options.KubeConfigPath, "The location of the kubeconfig file to create.")
+	cmd.Flags().StringVar(&options.KubeConfigPath, "kubeconfig", options.KubeConfigPath, "the location of the kubeconfig file to create.")
 	cmd.Flags().BoolVar(&options.all, "all", options.all, "export all clusters from the kops state store")
+	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "export the cluster admin user and add it to the context")
+	cmd.Flags().StringVar(&options.user, "user", options.user, "add an existing user to the cluster context")
 
 	return cmd
 }
@@ -80,12 +87,17 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 	if err != nil {
 		return err
 	}
+	if options.all {
+		if len(args) != 0 {
+			return fmt.Errorf("cannot use both --all flag and positional arguments")
+		}
+	}
+	if options.admin && options.user != "" {
+		return fmt.Errorf("cannot use both --admin and --user")
+	}
 
 	var clusterList []*kopsapi.Cluster
 	if options.all {
-		if len(args) != 0 {
-			return fmt.Errorf("Cannot use both --all flag and positional arguments")
-		}
 		list, err := clientset.ListClusters(ctx, metav1.ListOptions{})
 		if err != nil {
 			return err
@@ -116,7 +128,7 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 			return err
 		}
 
-		conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, buildPathOptions(options))
+		conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, buildPathOptions(options), options.admin, options.user)
 		if err != nil {
 			return err
 		}
--- a/cmd/kops/update_cluster.go
+++ b/cmd/kops/update_cluster.go
@@ -52,7 +52,7 @@ var (
 
 	updateClusterExample = templates.Examples(i18n.T(`
 	# After cluster has been edited or upgraded, configure it with:
-	kops update cluster k8s-cluster.example.com --yes --state=s3://kops-state-1234 --yes
+	kops update cluster k8s-cluster.example.com --yes --state=s3://kops-state-1234 --yes --admin
 	`))
 
 	updateClusterShort = i18n.T("Update a cluster.")
@@ -65,7 +65,10 @@ type UpdateClusterOptions struct {
 	OutDir          string
 	SSHPublicKey    string
 	RunTasksOptions fi.RunTasksOptions
-	CreateKubecfg   bool
+
+	CreateKubecfg bool
+	admin         bool
+	user          string
 
 	Phase string
 
@@ -80,7 +83,7 @@ func (o *UpdateClusterOptions) InitDefaults() {
 	o.Models = strings.Join(cloudup.CloudupModels, ",")
 	o.SSHPublicKey = ""
 	o.OutDir = ""
-	o.CreateKubecfg = true
+	o.CreateKubecfg = false
 	o.RunTasksOptions.InitDefaults()
 }
 
@@ -115,6 +118,8 @@ func NewCmdUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().StringVar(&options.SSHPublicKey, "ssh-public-key", options.SSHPublicKey, "SSH public key to use (deprecated: use kops create secret instead)")
 	cmd.Flags().StringVar(&options.OutDir, "out", options.OutDir, "Path to write any local output")
 	cmd.Flags().BoolVar(&options.CreateKubecfg, "create-kube-config", options.CreateKubecfg, "Will control automatically creating the kube config file on your local filesystem")
+	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "Also export the admin user. Implies --create-kube-config")
+	cmd.Flags().StringVar(&options.user, "user", options.user, "Existing user to add to the cluster context. Implies --create-kube-config")
 	cmd.Flags().StringVar(&options.Phase, "phase", options.Phase, "Subset of tasks to run: "+strings.Join(cloudup.Phases.List(), ", "))
 	cmd.Flags().StringSliceVar(&options.LifecycleOverrides, "lifecycle-overrides", options.LifecycleOverrides, "comma separated list of phase overrides, example: SecurityGroups=Ignore,InternetGateway=ExistsAndWarnIfChanges")
 	viper.BindPFlag("lifecycle-overrides", cmd.Flags().Lookup("lifecycle-overrides"))
@@ -136,6 +141,24 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, clusterName string, 
 
 	isDryrun := false
 	targetName := c.Target
+
+	if c.admin && c.user != "" {
+		return nil, fmt.Errorf("cannot use both --admin and --user")
+	}
+
+	if c.CreateKubecfg && !c.admin && c.user == "" {
+		return nil, fmt.Errorf("--create-kube-config requires that either --admin or --user is set")
+	}
+
+	if c.admin && !c.CreateKubecfg {
+		klog.Info("--admin implies --create-kube-config")
+		c.CreateKubecfg = true
+	}
+
+	if c.user != "" && !c.CreateKubecfg {
+		klog.Info("--user implies --create-kube-config")
+		c.CreateKubecfg = true
+	}
 
 	// direct requires --yes (others do not, because they don't do anything!)
 	if c.Target == cloudup.TargetDirect {
@@ -282,10 +305,11 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, clusterName string, 
 		}
 		if kubecfgCert != nil {
 			klog.Infof("Exporting kubecfg for cluster")
-			conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, clientcmd.NewDefaultPathOptions())
+			conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, clientcmd.NewDefaultPathOptions(), c.admin, c.user)
 			if err != nil {
 				return nil, err
 			}
+
 			err = conf.WriteKubecfg()
 			if err != nil {
 				return nil, err
--- a/docs/cli/kops_export_kubecfg.md
+++ b/docs/cli/kops_export_kubecfg.md
@@ -7,7 +7,7 @@ Export kubecfg.
 
 ### Synopsis
 
-Export a kubecfg file for a cluster from the state store. The configuration will be saved into a users $HOME/.kube/config file. To export the kubectl configuration to a specific file set the KUBECONFIG environment variable.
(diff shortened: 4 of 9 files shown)


cmd/kops/export_kubecfg.go
```diff
@@ -116,7 +120,12 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 			return err
 		}
 
-		conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, buildPathOptions(options))
+		conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, buildPathOptions(options), options.admin)
+		if options.admin {
```
rifelpet, line 124: Given that `--admin` and `--user` are mutually exclusive, we should probably return an error if both are specified.

cmd/kops/export_kubecfg.go
```diff
@@ -71,6 +73,8 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVar(&options.KubeConfigPath, "kubeconfig", options.KubeConfigPath, "The location of the kubeconfig file to create.")
 	cmd.Flags().BoolVar(&options.all, "all", options.all, "export all clusters from the kops state store")
+	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "also export the admin user")
```
rifelpet, line 76:
```suggestion
	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "export the admin user")
```

```suggestion
	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "export the admin user and use in the cluster context")
```

maybe one of these? I'll keep thinking of how this could be worded...

olemarkus:
Thanks for looking at this @rifelpet . The current staet is pretty much a brain dump of how I want this to work. It is in no way close to merge :) I just wanted some comments that more people agree to this approach. The final solution will probably need a bit of refactor etc, so I didn't want to do too much up front :)

Should have said so initially.Apologies.

cmd/kops/export_kubecfg.go
```diff
@@ -34,10 +34,10 @@ import (
 
 var (
 	exportKubecfgLong = templates.LongDesc(i18n.T(`
-	Export a kubecfg file for a cluster from the state store. The configuration
-	will be saved into a users $HOME/.kube/config file.
-	To export the kubectl configuration to a specific file set the KUBECONFIG
-	environment variable.`))
+	Export a kubecfg file for a cluster from the state store. By default, the configuration
+	will be saved into a users $HOME/.kube/config file. Kops will respect KUBECONFIG environment variable
+	if --kubeconfig flag is not set.
```
johngmyers, line 39:
```suggestion
	Export a kubecfg file for a cluster from the state store. By default the configuration
	will be saved into a users $HOME/.kube/config file. Kops will respect the KUBECONFIG environment variable
	if the --kubeconfig flag is not set.
```

cmd/kops/export_kubecfg.go
johngmyers: "include the admin user and credentials"

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -85,56 +85,57 @@ func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.Se
...
 
-	// add the CA Cert to the kubeconfig only if we didn't specify a SSL cert for the LB
-	if cluster.Spec.API == nil || cluster.Spec.API.LoadBalancer == nil || cluster.Spec.API.LoadBalancer.SSLCertificate == "" {
-		cert, _, _, err := keyStore.FindKeypair(fi.CertificateId_CA)
-		if err != nil {
-			return nil, fmt.Errorf("error fetching CA keypair: %v", err)
-		}
-		if cert != nil {
-			b.CACert, err = cert.AsBytes()
+	if admin {
```
johngmyers, line 90: Why omit the certificate-authority-data when `admin` is false?

pkg/kubeconfig/create_kubecfg_test.go
```diff
@@ -225,7 +225,7 @@ func TestBuildKubecfg(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := BuildKubecfg(tt.args.cluster, tt.args.keyStore, tt.args.secretStore, tt.args.status, tt.args.configAccess)
+			got, err := BuildKubecfg(tt.args.cluster, tt.args.keyStore, tt.args.secretStore, tt.args.status, tt.args.configAccess, true)
```
johngmyers, line 228: Should add a test for `false`

pkg/kubeconfig/kubecfg_builder.go
```diff
@@ -116,25 +117,15 @@ func (b *KubeconfigBuilder) WriteKubecfg() error {
...
+		cluster.CertificateAuthorityData = b.CACert
 
 		if config.Clusters == nil {
 			config.Clusters = make(map[string]*clientcmdapi.Cluster)
 		}
 		config.Clusters[b.Context] = cluster
 	}
 
-	{
+	if b.KubeUser == "admin" {
```
johngmyers, line 128: Why only write out the client cert if there is a plaintext password?

pkg/kubeconfig/kubecfg_builder.go
```diff
@@ -186,7 +181,9 @@ func (b *KubeconfigBuilder) WriteKubecfg() error {
 		}
 
 		context.Cluster = b.Context
-		context.AuthInfo = b.Context
```
johngmyers, line 189: I'm not quite sure what the reason for a `--user` flag is, but if a user specified `--admin` I'd think they'd want the `AuthInfo` set to the exported user.

cmd/kops/export_kubecfg.go
olemarkus: Yep. Done.

pkg/kubeconfig/create_kubecfg.go
olemarkus: Yeah. That wasn't intentional. Fixed.

pkg/kubeconfig/create_kubecfg_test.go
olemarkus: And fixed (which caught the above-mentioned error)

pkg/kubeconfig/kubecfg_builder.go
olemarkus: `--user` is to add an existing user to a new context. Since we use OIDC users, we just re-use the same user block for every cluster rather than specifying one user block per cluster.

If we allow both `--admin` and `--user` at the same time, and that this means "write the exported admin credentials to this user block" at least kops would not be able to delete that user when deleting the cluster. I am not sure what the use case would be for having a very specific name for the "kops managed" user block either. If we do want this feature, I'd rather have a new `--admin-user-name` flag or something.

pkg/kubeconfig/kubecfg_builder.go
olemarkus: It is a bit of a hack. I changed the check to something matching kops assumptions on user names.

pkg/kubeconfig/kubecfg_builder.go
johngmyers: I would think that if just `--admin` was given then the expectation would be the current behavior of setting `AuthInfo` to the newly created context.

olemarkus: Yes. That is the intention. The newly created context is always set to the cluster name, and here I set the user to be the same as the cluster name as well: https://github.com/kubernetes/kops/pull/9280/files#diff-2b7a768c601c36223ba9f3bafca41a31R137
So the effect should be the same.

cmd/kops/update_cluster.go
```diff
@@ -138,6 +143,22 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, clusterName string,
 	isDryrun := false
 	targetName := c.Target
 
+	if c.admin && c.user != "" {
+		return nil, fmt.Errorf("cannot use both --admin and --user")
+	}
+
+	if c.CreateKubecfg && !c.admin && c.user == "" {
+		return nil, fmt.Errorf("--create-kube-config requires that either --admin or --user is set")
```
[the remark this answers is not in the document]
olemarkus: To some extent code related. It is not straightforward in the current code to see if there is already a context for the cluster early in the `update` code. I didn't want to let the user run `update` to provision a new cluster, only for the command to fail at the very end when exporting the user and _then_ complain about `--admin` or `--user` not being set.

So we assume that the user only adds if they want to add the new context `--create-kube-config`. That is also why I defaulted this flag to false rather than true now since usually one only wants this if one runs `update` to create the cloud resources the first time.

The other use case I can think of is when one rotates certs, and at least if one uses the admin user, one would need to add `--admin` anyway.

cmd/kops/update_cluster.go
```diff
@@ -138,6 +143,22 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, clusterName string,
...
+	}
+
+	if c.admin && !c.CreateKubecfg {
+		return nil, fmt.Errorf("--admin requires that --create-kube-config is set")
+	}
+
+	if c.user != "" && !c.CreateKubecfg {
+		return nil, fmt.Errorf("--user requires that --create-kube-config is set")
+	}
```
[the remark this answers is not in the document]
olemarkus: That is a good idea! I'll give that a go.

cmd/kops/update_cluster.go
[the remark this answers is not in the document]
johngmyers: I believe it makes sense to export a context for the cluster without either adding admin credentials or setting the context to a precreated singular user.

Our authentication setup has per-cluster tokens which are obtained by an exec hook. That means a single user shared across multiple clusters would not work.

olemarkus:
The error from the AWS job is expected since it is not setting the `--admin` user on `kops update`. Would be easier if it just ran `kops create cluster --yes`.

I am strongly for requiring the `--admin` user flag which means either fixing test-infra code or the flags we set on the jobs (can we set flags on the update command?).

rifelpet:
kubetest currently only adds args to `create cluster`: https://github.com/kubernetes/test-infra/blob/4572dc3bf92e70f572e55e7ac1be643bdf6b2566/kubetest/kops.go#L465-L470

olemarkus:
If you agree with me, I can look at a PR to kubetest to just add the admin flag

rifelpet:
hm it'd be great if we could make the transition a bit smoother. I realize this is a breaking change regardless of how we implement it, but maybe we could do something like add the new cli flags in release-1.18 with defaults to the existing behavior, and then change the defaults in master. This way kubetest could test both master and release-1.18 kops builds.  Thoughts?

olemarkus:
We'd have to go further back than 1.18, wouldn't we? If we don't change kubetest, we'd have to backport noops of these flags as far back as we want to do any patch releases.

olemarkus:
`kops create cluster --yes` will default to `--admin`. So the easiest way to solve this is to just add `--yes` to `--kops-args` in the test grid. Test infra will do a redundant `kops update` after, but I can do a PR to get that command removed eventually.

olemarkus:
When https://github.com/kubernetes/test-infra/pull/17890 goes in, this should pass.

cmd/kops/create_cluster.go
```diff
@@ -1376,7 +1374,7 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 				fmt.Fprintf(&sb, " * edit your master instance group: kops edit ig --name=%s %s\n", clusterName, masters[0].ObjectMeta.Name)
 			}
 			fmt.Fprintf(&sb, "\n")
-			fmt.Fprintf(&sb, "Finally configure your cluster with: kops update cluster --name %s --yes\n", clusterName)
+			fmt.Fprintf(&sb, "Finally configure your cluster with: kops update cluster --name %s --yes --admin --create-kube-config\n", clusterName)
```
johngmyers, line 1377: Perhaps leave off the `--create-kube-config` since it's implied?

cmd/kops/export_kubecfg.go
```diff
@@ -34,14 +34,17 @@ import (
...
+	if the --kubeconfig flag is not set.
+	`))
 
 	exportKubecfgExample = templates.Examples(i18n.T(`
-	# export a kubecfg file
-	kops export kubecfg kubernetes-cluster.example.com
+	# export a kubeconfig file with the cluster admin user (make sure you keep this user safe!)
+	kops export kubecfg kubernetes-cluster.example.com --admin
+
+	# exports a kubeconfig file with an existing user
```
johngmyers, line 46:
```suggestion
	# export using a user already existing in the kubeconfig file
```

"exports" is inconsistent with the previous imperative "export".

cmd/kops/update_cluster.go
```diff
@@ -53,7 +53,7 @@ var (
 
 	updateClusterExample = templates.Examples(i18n.T(`
 	# After cluster has been edited or upgraded, configure it with:
-	kops update cluster k8s-cluster.example.com --yes --state=s3://kops-state-1234 --yes
+	kops update cluster k8s-cluster.example.com --yes --state=s3://kops-state-1234 --yes --admin --create-kube-config
```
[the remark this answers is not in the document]
johngmyers: Should remove `--create-kube-config` as it's implied.

cmd/kops/update_cluster.go
```diff
@@ -66,7 +66,10 @@ type UpdateClusterOptions struct {
 	OutDir          string
 	SSHPublicKey    string
 	RunTasksOptions fi.RunTasksOptions
-	CreateKubecfg   bool
+
+	CreateKubecfg bool
+	admin         bool
+	user          string
```
johngmyers, line 72: These two should probably be public.

cmd/kops/update_cluster.go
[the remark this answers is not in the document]
johngmyers: The logging probably isn't necessary.

docs/cli/kops_export_kubecfg.md
```diff
@@ -7,7 +7,7 @@ Export kubecfg.
 
 ### Synopsis
 
-Export a kubecfg file for a cluster from the state store. The configuration will be saved into a users $HOME/.kube/config file. To export the kubectl configuration to a specific file set the KUBECONFIG environment variable.
+Export a kubecfg file for a cluster from the state store. By default the configuration will be saved into a users $HOME/.kube/config file. Kops will respect the KUBECONFIG environment variable if the --kubeconfig flag is not set.
```
johngmyers, line 10:
```suggestion
Export a kubecfg file for a cluster from the state store. By default the configuration will be saved into a user's $HOME/.kube/config file. Kops will respect the KUBECONFIG environment variable if the --kubeconfig flag is not set.
```

cmd/kops/update_cluster.go
olemarkus: I see these are made public a bit inconsistently. In my opinion, probably all of these should not be exposed (like in create_cluster). I just prefer to leave them unexposed and then unexpose the existing flag at some later time.

johngmyers: Different issue, but some of the logic in create_cluster needs to be moved into pkg so it can be called as imported code. I haven't gone through update_cluster to see if there's logic that should be exposed in pkg.

olemarkus:
Looks like tests are passing and all now!

/assign @rifelpe

cmd/kops/update_cluster.go
johngmyers: There's only one block of logic in `RunUpdateCluster` that should be pushed down into `ApplyClusterCmd`, so consider this resolved.

cmd/kops/update_cluster.go
```diff
@@ -306,10 +329,17 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, clusterName string,
...
+			conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, clientcmd.NewDefaultPathOptions(), c.admin)
 			if err != nil {
 				return nil, err
 			}
+
+			if c.admin {
+				conf.User = cluster.ObjectMeta.Name
+			} else {
+				conf.User = c.user
+			}
```
johngmyers, line 341: I believe this logic should be pushed down into `kubeconfig.BuildKubecfg()`.

cmd/kops/update_cluster.go
```diff
@@ -306,10 +329,17 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, clusterName string,
 		}
 		if kubecfgCert != nil {
 			klog.Infof("Exporting kubecfg for cluster")
-			conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, clientcmd.NewDefaultPathOptions())
+			conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, clientcmd.NewDefaultPathOptions(), c.admin)
 			if err != nil {
 				return nil, err
 			}
+
+			if c.admin {
```
johngmyers, line 337:
```suggestion
			if c.user == "" {
```

cmd/kops/update_cluster.go
olemarkus: Not sure why I did not do this earlier. Much better and the logic is now tested properly too.

johngmyers:
Review comments on create_cluster.go and update_cluster.go still unaddressed.

cmd/kops/update_cluster.go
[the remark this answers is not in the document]
olemarkus: I am a bit concerned that if one sets `--create-kube-config`, but does not catch that one _probably should_ also set `--admin` or `--user` one will get a broken config. So I do agree that there are use cases for not doing this, I am wondering if this is worth it usabilitywise.

For your use case, you have to remove the user from the context once and then kops will at least leave the context alone and not keep re-adding the admin user, which is a good benefit on its own I should think. I assume that typically you would only export the cluster config once per operator per cluster so this isn't too much hassle.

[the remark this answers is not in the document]
johngmyers: Since `--admin` and `--user` imply `--create-kube-config` it is unlikely that `--create-kube-config` would be explicitly set unless the invoker is doing something special.

Our wrapper around kops has two forms of `export kubecfg`, one which corresponds to `--admin` and one which adds a per-cluster user with our local exec hook added. Since the admin credential is so powerful, I would prefer not to have that second case extract the credential, write it to disk, and then try to remove it.

I suppose I could let the current PR through and then address this case in a followup PR. I am planning on reducing the default lifetime of the admin credential to 18 hours.

docs/releases/1.19-NOTES.md
```diff
@@ -4,6 +4,16 @@
 
 # Significant changes
 
+## Changes to kubernetes config export
+
+Kops will no longer automatically export the kubernetes config on `kops update cluster`. In order to export the config on cluster update, you need to either add the `--user <user>` to reference an existing user, or `--admin` to export the cluster admin user. If neither flag is passed, the kubernetes config will not be modified. This makes it easier to reuse user definitions across clusters should you for example use OIDC for authentication.
```
johngmyers, line 9:
```suggestion
Kops will no longer automatically export the kubernetes config on `kops update cluster`. In order to export the config on cluster update, you need to either add the `--user <user>` to reference an existing user, or `--admin` to export the cluster admin user. If neither flag is passed, the kubernetes config will not be modified. This makes it easier to reuse user definitions across clusters should you, for example, use OIDC for authentication.
```

docs/releases/1.19-NOTES.md
```diff
@@ -4,6 +4,16 @@
...
+
+Similarly, `kops export kubecfg` will also require that either the `--admin` or `--user` flag is passed if the context does not already exist.
```
johngmyers, line 11:
```suggestion
Similarly, `kops export kubecfg` will also require passing either the `--admin` or `--user` flag if the context does not already exist.
```

docs/releases/1.19-NOTES.md
```diff
@@ -4,6 +4,16 @@
...
+
+`kops create cluster --yes` implies exporting the admin user together with rest of the cluster config equal to existing behaviour.
```
johngmyers, line 13:
```suggestion
`kops create cluster --yes` exports the admin user along with rest of the cluster config, as is existing behaviour.
```

cmd/kops/update_cluster.go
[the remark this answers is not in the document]
olemarkus: I think that would be helpful. You have a much better idea of how this should work for your use case than I do.

I do agree that passing only `--create-kube-config` would be something special. If you log a warning about creating a context without user, I think that should be fine. It would at least give a hint to the user about why the credentials aren't working.

cmd/kops/update_cluster.go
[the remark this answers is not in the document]
olemarkus: I think they could be helpful and I guess common for CLI tools. But we also do document that it is implied. If you think they should go, I can remove it.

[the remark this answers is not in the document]
johngmyers: I think most users aren't going to keep `--create-kube-config` in mind. They're going to know either `--admin` or `--user`, whichever fits their use case. So I think this would just be log chatter.

johngmyers:
/lgtm
Could remove the log chatter either now or in a followup PR.

rifelpet:
this lgtm as well but since it's a breaking change and rather user-facing, i'd prefer if justin gives the final approval. if it doesnt get merged before next office hours we can discuss it then.

/lgtm
/assign @justinsb

rifelpet:
No objections during office hours
/approve

PR #9282, opened 2020-06-05
Create nodetasks.IssueCert()

johngmyers:


upup/pkg/fi/nodeup/nodetasks/issue_cert.go
```diff
@@ -0,0 +1,140 @@
...
+func (i *IssueCert) GetResources() (certResource, keyResource, caResource *fi.TaskDependentResource) {
+	if i.cert == nil {
+		i.cert = &fi.TaskDependentResource{Task: i}
+		i.key = &fi.TaskDependentResource{Task: i}
+		i.ca = &fi.TaskDependentResource{Task: i}
+	}
+	return i.cert, i.key, i.ca
+}
+
+func (i *IssueCert) AddFileTasks(c *fi.ModelBuilderContext, dir string, name string, caName string, owner *string) {
```
justinsb, line 69: Not something to address now, but we might want to use a struct just so we can name these fields.  (Or we could have 3 methods that each build one of the file tasks, then it's easy to set e.g. the owner on them.)

upup/pkg/fi/nodeup/nodetasks/issue_cert.go
```diff
@@ -0,0 +1,140 @@
...
+	certResource, keyResource, caResource := i.GetResources()
+	c.AddTask(&File{
```
justinsb, line 71: FYI we have EnsureTask which can cope with duplicates (and enforces that the duplicates should be the same ... I guess we're not hitting it though!)

upup/pkg/fi/resources.go
```diff
@@ -248,3 +248,22 @@ func WrapResource(r Resource) *ResourceHolder {
...
+	Resource Resource
+	Task     Task
+}
+
+var _ Resource = &TaskDependentResource{}
+var _ HasDependencies = &TaskDependentResource{}
+
+func (r *TaskDependentResource) Open() (io.Reader, error) {
+	if r.Resource == nil {
+		return nil, fmt.Errorf("resource opened before it is ready")
```
justinsb, line 262: Nit: we might want to put the task name in here to help us debug if we do hit it - but I'm guessing we don't expect to hit this.

upup/pkg/fi/nodeup/nodetasks/issue_cert.go
```diff
@@ -0,0 +1,140 @@
...
+	caResource.Resource = &asBytesResource{caCertificate}
+
+	return nil
+}
+
+type hasAsBytes interface {
+	AsBytes() ([]byte, error)
+}
+
+type asBytesResource struct {
```
justinsb, line 130: We do have `fi.NewBytesResource` - I feel like we might be able to reuse that?  (But what you're doing here with the late binding is pretty clever; I might be missing a subtlety!)

justinsb:
This LGTM - I had some comments / thoughts which I want to make sure you see @johngmyers , but you can choose whether they're worth resolving.  So I'm going to add `hold`, but feel free to `hold cancel`.

/approve
/lgtm
/hold

upup/pkg/fi/nodeup/nodetasks/issue_cert.go
johngmyers: I'm trying to defer calling `AsBytes()` until the caller invokes `Open()`.

upup/pkg/fi/nodeup/nodetasks/issue_cert.go
johngmyers: That would require two Builders to write the same cert to the same file. It would be a bad practice for two independent things to share the same cert. I might use EnsureTask for the CA file, though.

upup/pkg/fi/resources.go
johngmyers: We don't expect to hit this: it would require a bug where the consumer of the `TaskDependentResource` fails to declare it as a dependency. I'm thinking of turning this into a panic() as the retrying behavior of the `Task` scheduler would mask any such bug.

upup/pkg/fi/nodeup/nodetasks/issue_cert.go
justinsb: Sorry - I was really thinking about the directory!

upup/pkg/fi/resources.go
justinsb: :+1: either way works.  `klog.Fatalf` == panic but makes it easier to add more information!

upup/pkg/fi/nodeup/nodetasks/issue_cert.go
johngmyers: Most Builders that use certs write them as three files. A few write as two files, omitting the CA. A very few, like the kubelet builder, need to  include them in a configuration file. 

I do want to keep the common case succinct. I'll see what using the struct looks like.

johngmyers:
/hold cancel
I'm going to wait for #9248 to land, then see about converting one of those client certs to IssueCert, adding the testing infrastructure around IssueCert users.

justinsb:
Sounds great, thanks @johngmyers  - should I rebase #9248 now or are there more we should get in first?  (I'm happy to add the testing for this task in #9248, I'm guessing it'll be required now anyway!)

johngmyers:
Please proceed with #9248 now. I was planning on doing further cert work on top of both. I also have one refactor depending on this and `NodeupContext` changes that could benefit from having golden tests.

PR #9284, opened 2020-06-06
Don't export basic auth credentials if basic auth is disabled

johngmyers:


hakman:
e2e test failure is related to https://github.com/kubernetes/test-infra/pull/17864

PR #9286, opened 2020-06-06
Copy "portmap" to /opt/cni/bin for Weave

hakman:
Weave requires `portmap` that seems to be missing at the moment:
https://www.weave.works/docs/net/latest/kubernetes/kube-addon/#-installation

e2e tests are failing because of the missing portmap plugin:
https://testgrid.k8s.io/kops-network-plugins#kops-aws-cni-weave

```
E0608 09:35:30.294544    5002 kubelet.go:2187] Container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: cni config uninitialized
I0608 09:35:33.143148    5002 kubelet.go:1952] SyncLoop (PLEG): "weave-net-6v74s_kube-system(14526285-9260-43e3-8c5e-62eb340b2dc1)", event: &pleg.PodLifecycleEvent{ID:"14526285-9260-43e3-8c5e-62eb340b2dc1", Type:"ContainerStarted", Data:"6cea0ffccafd5578fc62e565b78258d4d10bc6fc11be27b5e1990fbc0b40cf2e"}
W0608 09:35:34.281513    5002 cni.go:202] Error validating CNI config list {
    "cniVersion": "0.3.0",
    "name": "weave",
    "plugins": [
        {
            "name": "weave",
            "type": "weave-net",
            "hairpinMode": true
        },
        {
            "type": "portmap",
            "capabilities": {"portMappings": true},
            "snat": true
        }
    ]
}
: [failed to find plugin "portmap" in path [/opt/cni/bin/]]
W0608 09:35:34.282448    5002 cni.go:237] Unable to update cni config: no valid networks found in /etc/cni/net.d/
```

PR #9287, opened 2020-06-06
Install loopback CNI plugin when using amazon vpc cni

olemarkus:


olemarkus:
similar to #9286, but different plugin

nodeup/pkg/model/networking/amazonvpc.go
```diff
@@ -0,0 +1,39 @@
...
+
+package networking
+
+import (
+	"k8s.io/kops/nodeup/pkg/model"
+
+	"k8s.io/kops/upup/pkg/fi"
+)
+
+// FlannelBuilder writes Flannel's assets
```
MoShitrit, line 25: Should that comment be updated to `// AmazonVPCBuilder writes Amazon VPC's assets`?

olemarkus: Thanks for spotting

hakman:
@olemarkus @rifelpet how do you feel about adding back the `NetworkBuilder` and always extract *all* CNI binaries to `/opt/cni/bin`?

olemarkus:
The loopback plugin is fine, maybe on the standard ones, I'd prefer to avoid those that are nearly never used (like flannel)

hakman:
Sounds good for me. I guess we will add flannel just for Flannel, Calico and Canal.

olemarkus:
Superseded by #9310 
/close

PR #9288, opened 2020-06-06
Fix mismatch in SecurityGroups handling with launch templates

johngmyers:
When launch templates are enabled, they will be unnecessarily updated when they have a security group and are in a private subnet.

/kind bug
Targeted for 1.18

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
+++ b/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
@@ -203,15 +203,10 @@ func (t *LaunchTemplate) Find(c *fi.Context) (*LaunchTemplate, error) {
 	for _, x := range lt.LaunchTemplateData.NetworkInterfaces {
 		if aws.BoolValue(x.AssociatePublicIpAddress) {
 			actual.AssociatePublicIP = fi.Bool(true)
-			// @note: not sure i like this https://github.com/hashicorp/terraform/issues/2998
-			for _, id := range x.Groups {
-				actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: id})
-			}
 		}
-	}
-	// @step: add at the security groups
-	for _, id := range lt.LaunchTemplateData.SecurityGroupIds {
-		actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: id})
+		for _, id := range x.Groups {
+			actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: id})
+		}
 	}
 	sort.Sort(OrderSecurityGroupsById(actual.SecurityGroups))
 


hakman:
@johngmyers This change was intentional, as best practice. I think for Mixed instance policies is the only possibility.
Does this happen only first time or during subsequent deployments also?

johngmyers:
@hakman Every `kops update cluster --yes` updates the launch template, as it thinks it needs to add the `SecurityGroup` back in. It causes the lifecycle tests to fail when I enable launch templates by default.

johngmyers:
@hakman Notice I'm changing `Find()`, not `RenderAWS()`.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -203,15 +203,10 @@ func (t *LaunchTemplate) Find(c *fi.Context) (*LaunchTemplate, error) {
...
 			actual.AssociatePublicIP = fi.Bool(true)
-			// @note: not sure i like this https://github.com/hashicorp/terraform/issues/2998
-			for _, id := range x.Groups {
-				actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: id})
-			}
 		}
-	}
-	// @step: add at the security groups
-	for _, id := range lt.LaunchTemplateData.SecurityGroupIds {
-		actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: id})
```
hakman, line 214: I think we kept this in case someone migrated from an older version of the launch template.

hakman:
> @hakman Notice I'm changing `Find()`, not `RenderAWS()`.

Thanks. Looks pretty good. Let me know what you think about the other comment and we can get this in.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
johngmyers: To upgrade them correctly it should ensure that the returned actual will not match the expected.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -203,15 +203,10 @@ func (t *LaunchTemplate) Find(c *fi.Context) (*LaunchTemplate, error) {
...
+		for _, id := range x.Groups {
+			actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: id})
+		}
 	}
```
hakman, line 210:
```suggestion
	}

	// In older Kops versions, security groups were added to LaunchTemplateData.SecurityGroupIds
	for _, id := range lt.LaunchTemplateData.SecurityGroupIds {
		actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: id})
	}
```

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -208,6 +208,10 @@ func (t *LaunchTemplate) Find(c *fi.Context) (*LaunchTemplate, error) {
 			actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: id})
 		}
 	}
+	// In older Kops versions, security groups were added to LaunchTemplateData.SecurityGroupIds
+	for _, id := range lt.LaunchTemplateData.SecurityGroupIds {
+		actual.SecurityGroups = append(actual.SecurityGroups, &SecurityGroup{ID: fi.String("legacy-" + *id)})
```
hakman, line 213: Why the "legacy-" prefix?
I thought we just want to move the SGs from `LaunchTemplateData.SecurityGroupIds` to `LaunchTemplateData.NetworkInterfaces[0].Groups`.

johngmyers: `Find()` builds a representation of the "actual" object, the state as it currently exists. The caller then diffs that against the representation of the "expected" object, the state that was built from the config in the state store. Only if those two differ does it call `RenderAWS`, which would build a new launch template to attach to the ASG.

If we did not add the "legacy-" prefix then the actual could equal the expected and thus it would believe that no changes were necessary.

hakman: In this particular case, if SG is in the legacy place, I think the LT should not change. Only on next change the SG will be moved together with something else to the new location.

I guess it is debatable if we want this or not. Your choice here.

PR #9293, opened 2020-06-07
Adding recent releases to docs site

MoShitrit:
I've noticed that we don't have recent releases in the nav pane (the most recent is 1.16).
image

I added 1.17 release and 1.18 (marked as beta)

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -145,6 +145,8 @@ nav:
     - Vendoring: "development/vendoring.md"
     - Ports: "development/ports.md"
   - Releases:
+    - 1.18-beta: releases/1.18-NOTES.md
+    - 1.17: releases/1.17-NOTES.md
     - 1.16: releases/1.16-NOTES.md
     - 1.15: releases/1.15-NOTES.md
     - 1.14: releases/1.14-NOTES.md


mkdocs.yml
```diff
@@ -145,6 +145,9 @@ nav:
     - Vendoring: "development/vendoring.md"
     - Ports: "development/ports.md"
   - Releases:
+    - 1.19: releases/1.19-NOTES.md
```
olemarkus, line 148: Perhaps not list 1.19 and 1.18 just yet?

MoShitrit: Yeah, I wasn't sure about 1.19 since it's not really released yet. 
But since the[ file is already there](https://github.com/kubernetes/kops/blob/master/docs/releases/1.19-NOTES.md) and notes that right now just state that it hasn't been released yet, I figured we _could_ just add it and whoever clicks it will see that it's not released yet.
I'm okay with taking it out if you think we shouldn't include it yet, let me know 😄

olemarkus: Remove the 1.19 one. I see that we link to 1.18 from the releases page, but perhaps add "(beta)" to it?

MoShitrit: Done 😄

hakman:
@MoShitrit for now, e2e tests are broken. Probably will take another 1-2 days to fix.

MoShitrit:
Thanks for the update @hakman !
Yeah I was wondering why it keeps failing.
I'll hang tight then 😄

PR #9295, opened 2020-06-08
Validate cilium version

olemarkus:
It is a bit too easy for Cilium users to configure unsupported combinations of cilium version and features. This PR ensures that users only set a supported version. Plan on follow-ups that will validate various features as well.

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -673,7 +673,7 @@ func validateNetworkingCilium(cluster *kops.Cluster, v *kops.CiliumNetworkingSpe
 			allErrs = append(allErrs, field.Invalid(versionFld, v.Version, "Only versions 1.6 and 1.7 are supported"))
 		}
 
-		if !cluster.IsKubernetesGTE("1.12.0") && version.GTE(v7) {
+		if !cluster.IsKubernetesGTE("1.12") && version.GTE(v7) {
 			allErrs = append(allErrs, field.Forbidden(versionFld, "Version >= 1.7 requires kubernetesVersion 1.12 or higher"))
 		}
 	}
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -658,6 +658,9 @@ func validateNetworkingCilium(cluster *kops.Cluster, v *kops.CiliumNetworkingSpe
 	if v.Version != "" {
 		versionFld := fldPath.Child("version")
 		version, err := semver.ParseTolerant(v.Version)
+
+		version.Pre = nil
+		version.Build = nil
 		if err != nil {
 			allErrs = append(allErrs, field.Invalid(versionFld, v.Version, "Could not parse as semantic version"))
 		}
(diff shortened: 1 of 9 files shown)


pkg/apis/kops/validation/validation.go
```diff
@@ -459,9 +460,30 @@ func validateNetworkingCanal(v *kops.CanalNetworkingSpec, fldPath *field.Path) f
...
+
+		v8, _ := semver.Parse("1.8.0")
+		v7, _ := semver.Parse("1.7.0")
+		v6, _ := semver.Parse("1.6.0")
+
+		if !(version.GTE(v6) && version.LT(v8)) {
+			allErrs = append(allErrs, field.Invalid(versionFld, v.Version, "Only versions 1.6 and 1.7 are supported"))
+		}
+
+		if !cluster.IsKubernetesGTE("1.12.0") && version.GTE(v7) {
```
johngmyers, line 482: This would permit "1.7.0-rc1" on Kubernetes 1.11. You might want to use `util.IsKubernetesGTE()`

pkg/apis/kops/validation/validation.go
```diff
@@ -459,9 +460,30 @@ func validateNetworkingCanal(v *kops.CanalNetworkingSpec, fldPath *field.Path) f
...
+			allErrs = append(allErrs, field.Invalid(versionFld, v.Version, "Version >= 1.7 requires kubernetesVersion 1.12 or higher"))
```
johngmyers, line 483: I believe this should be `field.Forbidden`

pkg/apis/kops/validation/validation.go
olemarkus: Just ignoring the pre/build parts similar to what `util.IsKubernetesGTE()` does.

pkg/apis/kops/validation/validation.go
olemarkus: Fixed

pkg/apis/kops/validation/validation.go
johngmyers: The usual pattern is to use just `"1.12"`

johngmyers:
Just the one remaining nit for me.

pkg/model/components/cilium.go
```diff
@@ -0,0 +1,84 @@
...
+	c := clusterSpec.Networking.Cilium
+	if c == nil {
+		return nil
+	}
+
+	if c.Version == "" {
+		if b.Context.IsKubernetesLT("1.12.0") {
+			c.Version = "v1.6.9"
+		} else {
+			c.Version = "v1.7.4"
```
rifelpet, line 42: Should this be bumped to 1.7.5 ?

upup/models/bindata.go
```diff
@@ -4587,7 +4587,7 @@ spec:
               key: wait-bpf-mount
               name: cilium-config
               optional: true
-        image: "docker.io/cilium/cilium:{{- or .Version "v1.7.3" }}"
+        image: "docker.io/cilium/cilium:{{ "v1.7.3" }}"
```
rifelpet, line 4590: also wondering about this version. it'd be nice if we could reduce the number of places  these are defined.

rifelpet: (i really should be commenting on the yaml but bindata.go sorts first when viewing GH diffs so I see this before the yaml... ☹️  )

pkg/model/components/cilium.go
olemarkus: 1.7.5 wasn't released when I did this PR, and now I'd like to get this in before doing more follow-up work :)

upup/models/bindata.go
olemarkus: Yep. I am working on a follow-up PR to clean up these templates.

pkg/model/components/cilium.go
rifelpet: got it 👍

PR #9309, opened 2020-06-09
Upgrade mkdocs to latest versions

rifelpet:
The biggest improvement I see is much snappier search results.

followed the material theme upgrade instructions [here](https://squidfunk.github.io/mkdocs-material/releases/5/)

### DIFF
--- a/images/mkdocs/Dockerfile
+++ b/images/mkdocs/Dockerfile
@@ -14,7 +14,7 @@
 
 FROM alpine:3.10
 
-RUN apk update && apk add --no-cache \
+RUN apk add --no-cache \
   bash \
   git \
   git-fast-import \
@@ -22,6 +22,7 @@ RUN apk update && apk add --no-cache \
   python3 \
   python3-dev \
   curl \
+  build-base \
   && python3 -m ensurepip \
   && rm -r /usr/lib/python*/ensurepip \
   && pip3 install --upgrade pip setuptools \
--- a/images/mkdocs/.dockerignore
+++ b/images/mkdocs/.dockerignore
@@ -1,0 +1,1 @@
+Dockerfile
--- a/images/mkdocs/requirements.txt
+++ b/images/mkdocs/requirements.txt
@@ -1,5 +1,4 @@
-mkdocs-material~=4.4.0
-mkdocs~=1.0.4
-# mkdocs-awesome-pages-plugin~=1.2.0
-pymdown-extensions==6.2.1
-pygments~=2.3.1
+mkdocs-material~=5.2.3
+mkdocs~=1.1.2
+pymdown-extensions==7.1
+pygments~=2.6.1
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -15,10 +15,11 @@ markdown_extensions:
       permalink: ' ¶'
 theme:
   name: material
-  feature:
-    tabs: true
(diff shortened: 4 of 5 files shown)


rifelpet:
this may need to wait until netlify updates the image tag used in their builds: https://github.com/netlify/build-image/releases/tag/v3.3.15 in order to support PYTHON_VERSION env var (its defaulting to 2.7 which doesn't have the newest modules we're upgrading to)

build log shows:

```
11:02:56 PM: build-image tag: v3.3.14
```

They push tags every couple weeks and the newest tag with PYTHON_VERSION support is only 6 days old, so hopefully it won't be long before builds start using it. alternatively we create a [runtime.txt](https://docs.netlify.com/configure-builds/manage-dependencies/#python) in the repo root but that's not very clean. I'd rather wait for the env var support.

/hold

mikesplain:
Looks good, only holding an lgtm since we can’t see a preview, hopefully they’ll get that updated soon!

Thanks!

rifelpet:
netlify updated their build environment to 3.3.15:

https://app.netlify.com/sites/kubernetes-kops/deploys/5ee9819edcd1790007bb6543

which supports specifying the python version by env var, as you can see it is now using python 3.7. The rest of the build succeeds and you can check out the deploy preview as normal.

I think this is ready for review/merge

images/mkdocs/Dockerfile
```diff
@@ -14,14 +14,15 @@
 
 FROM alpine:3.10
 
-RUN apk update && apk add --no-cache \
+RUN apk add --no-cache \
```
hakman, line 17: Why remove the `apk update` part?

rifelpet: `add`'s `--no-cache` performs an `update` under the hood but doesnt cache the results. you can see the indexes being downloaded twice:

```
docker run -it --rm alpine
/ # apk update && apk add --no-cache bash
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz
v3.12.0-74-gb8f92cf12a [http://dl-cdn.alpinelinux.org/alpine/v3.12/main]
v3.12.0-76-ge6ea27f814 [http://dl-cdn.alpinelinux.org/alpine/v3.12/community]
OK: 12734 distinct packages available
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz
(1/4) Installing ncurses-terminfo-base (6.2_p20200523-r0)
(2/4) Installing ncurses-libs (6.2_p20200523-r0)
(3/4) Installing readline (8.0.4-r0)
(4/4) Installing bash (5.0.17-r0)
Executing bash-5.0.17-r0.post-install
Executing busybox-1.31.1-r16.trigger
OK: 8 MiB in 18 packages
```

also see [SO](https://stackoverflow.com/questions/49118579/alpine-dockerfile-advantages-of-no-cache-vs-rm-var-cache-apk)

hakman: Thanks. Nice trick.

PR #9310, opened 2020-06-09
Install common CNI plugin binaries for all network plugins

hakman:
Kops should control the version of the common CNI plugin binaries that is used.
This is needed for example for https://github.com/kubernetes/kops/pull/9256.

Ref: https://github.com/kubernetes/kops/pull/9287#issuecomment-640705152


nodeup/pkg/model/networking/canal.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"k8s.io/klog"
+	"k8s.io/kops/nodeup/pkg/model"
+	"k8s.io/kops/pkg/systemd"
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
+)
+
+// CanalBuilder writes Flannel's assets
```
olemarkus, line 27: Nit. I guess you meant Canal here

nodeup/pkg/model/networking/canal.go
```diff
@@ -0,0 +1,71 @@
...
+	b.AddCNIBinAssets(c, []string{"flannel"})
+	b.buildFlannelTxChecksumOffloadDisableService(c)
+
+	return nil
+}
+
+// Tx checksum offloading is buggy for NAT-ed VXLAN endpoints, leading to an invalid checksum sent and causing
+// Flannel to stop to working as the traffic is being discarded by the receiver.
+// https://github.com/coreos/flannel/issues/1279
+func (b *CanalBuilder) buildFlannelTxChecksumOffloadDisableService(c *fi.ModelBuilderContext) {
```
olemarkus, line 51: Is this one copied between Canal and Flannel now? 
Since `*CanalBuilder` isn't used here, you could just make this a helper function rather than to make it part of the `CanalBuilder` type.

hakman: Good idea. Done.

nodeup/pkg/model/networking/common.go
```diff
@@ -21,20 +21,25 @@ import (
 	"k8s.io/kops/upup/pkg/fi"
 )
 
-// WeaveBuilder installs weave
-type WeaveBuilder struct {
+// CommonBuilder runs common tasks
+type CommonBuilder struct {
```
johngmyers, line 25: `NetworkingBuilder`?

nodeup/pkg/model/networking/common.go
```diff
@@ -21,20 +21,25 @@ import (
...
 	*model.NodeupModelContext
 }
 
-var _ fi.ModelBuilder = &WeaveBuilder{}
+var _ fi.ModelBuilder = &CommonBuilder{}
```
hakman, line 29: @johngmyers I just realised that the name was actually `networking.CommonBuilder`.
Should be better like this than `networking.NetworkingBuilder`.

PR #9313, opened 2020-06-09
Issue kube-scheduler and kube-controller-manager certs in nodeup

johngmyers:
Need to research how to make the golden output more readable/useful.

/cc @justinsb

### DIFF
--- a/upup/pkg/fi/nodeup/nodetasks/issue_cert.go
+++ b/upup/pkg/fi/nodeup/nodetasks/issue_cert.go
@@ -30,6 +30,7 @@ import (
 
 // PKIXName is a simplified form of pkix.Name, for better golden test output
 type PKIXName struct {
+	fi.NotADependency
 	CommonName string
 }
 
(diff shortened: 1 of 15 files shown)


justinsb:
This actually looks really good to me.  It doesn't have the generated keys themselves, which was always going to be the concern :-)

I did try two things: I wrapped pkix.Name so we could avoid all the nils e.g. `PostalCode: null` , and then I added some `omitempty` directives to TaskDependentResource just to clean up a few more nils:
https://github.com/justinsb/kops/commit/bd51fb426ce21047ac7886c64a97b261f35cba89

Personally I think this looks really good at this point - very clear.   Do you think we need to do more here @johngmyers ?

upup/pkg/fi/nodeup/nodetasks/issue_cert.go
```diff
@@ -28,13 +28,25 @@ import (
 	"k8s.io/kops/upup/pkg/fi"
 )
 
+// PKIXName is a simplified form of pkix.Name, for better golden test output
+type PKIXName struct {
+	fi.NotADependency
```
justinsb, line 33: nice idea :+1:

justinsb:
Thanks @johngmyers - this is looking really good

/approve
/lgtm

PR #9317, opened 2020-06-09
Use Docker 19.03.11 for Kubernetes 1.17+

hakman:
As discussed during Office Hours, the default Docker version for Kubernetes 1.17+ should be 19.03.11.


nodeup/pkg/model/docker.go
```diff
@@ -298,252 +298,24 @@ var dockerVersions = []packageVersion{
...
-				Version: "5:19.03.4~3-0~debian-stretch",
-				Source:  "https://download.docker.com/linux/debian/dists/stretch/pool/stable/amd64/docker-ce-cli_19.03.4~3-0~debian-stretch_amd64.deb",
-				Hash:    "f0f3c9c91a9482b0fe120cd9e404c3ade342ce01d0d98a7f6bce3e16b7c57a11",
-			},
-		},
-	},
-
-	// 19.03.4 - Debian Buster
-	{
-		PackageVersion: "19.03.4",
```
johngmyers, line 459: I don't think we should be removing these, since 19.03.4 is still a supported option in validation.go

hakman: We don't remove it, it remains as static package for all distros.

nodeup/pkg/model/docker.go
```diff
@@ -298,252 +298,24 @@ var dockerVersions = []packageVersion{
...
 		Hash:           "82a362af7689038c51573e0fd0554da8703f0d06f4dfe95dd5bda5acf0ae45fb",
 	},
 
-	// 19.03.4 - k8s 1.17 - https://github.com/kubernetes/kubernetes/pull/84476
-
-	// 19.03.4 - Linux Generic
-	//
-	// * AmazonLinux2: the Centos7 package depends on container-selinux, but selinux isn't used on amazonlinux2
-	// * UbuntuFocal: no focal version available at download.docker.com
+	// 19.03.8 - Linux Generic
```
johngmyers, line 310:
```suggestion
	// 19.03.4 - Linux Generic
```

hakman: Thanks. Fixed.

hakman:
Same treatment for containerd versions.

nodeup/pkg/model/containerd.go
```diff
@@ -51,103 +51,15 @@ var containerdVersions = []packageVersion{
...
 	{
 		PackageVersion: "1.2.10",
 		PlainBinary:    true,
-		Distros:        []distros.Distribution{distros.DistributionAmazonLinux2, distros.DistributionFocal},
 		Architectures:  []architectures.Architecture{architectures.ArchitectureAmd64},
 		Source:         "https://storage.googleapis.com/cri-containerd-release/cri-containerd-1.2.10.linux-amd64.tar.gz",
 		Hash:           "9125a6ae5a89dfe9403fea7d03a8d8ba9fa97b6863ee8698c4e6c258fb14f1fd",
 	},
 
-	// 1.2.11 - Linux Generic
```
johngmyers, line 133: I presume you're planning on removing support for these four versions from 1.18 as well.

I notice we don't have API validation for the containerd version.

hakman: Yes, will cherry-pick to 1.18 also. These were only used during development and current versions will be the final ones.

hakman: The version validation for Docker was a bit rushed up when 1.x support was removed.
Thanks for the reminder to add validation for containerd also.

Issue #9336, opened 2020-06-11
New Kops 1.17 AWS StorageClass Wreaks Havoc with Existing Defaults and Provides Uncertain Upgrade Path


LeeHampton:
[This commit](https://github.com/kubernetes/kops/pull/8582/files#diff-efbe8714d84ff556a917cc97b0e2e75d) introduces a new StorageClass called `kops-ssd-1-17`. There seem to be a couple of issues with this. 

First, even though it is no longer explicitly declared in [this file](https://github.com/joshbranham/kops/blob/9f033f8e999b52ee7bd7dc7b5661f62cc62a706f/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml), the `gp2` StorageClass is set to the default volume in existing pre 1.17 clusters (i.e., it has the `"storageclass.beta.kubernetes.io/is-default-class":"true"` annotation). That's because it used to be set to the default, as seen in [this commit](https://github.com/joshbranham/kops/commit/9f2b671a9a7475b861e604c8694c6b6f129b591f).  For whatever reason, this annotation does not get undone during the upgrade to 1.17. Now that `kops-ssd-1-17` has that annotation explicitly set, what I ended up with after upgrading from 1.16 to 1.17 is 2 default StorageClasses, which leads to the following error when attempting to create a PV/PVC:

```Internal error occurred: 2 default StorageClasses were found```

Second, tying the name of this new StorageClass to the kops version seems pretty problematic. In my case, I'm going to need to remove the annotation on this StorageClass to get my existing infra working again. To do so, I'll add a simple `kubectl patch` operation to my init/upgrade scripts. However, with this naming schema I'm assuming subsequent versions of kops will change the name of the storage class to e.g. `kops-ssd-1-18`, which means I'm always going to be playing wack-a-mole with these version specific names. Why can't we just use a generic name like `ssd-ebs` (or even just still use the gp2 name, since it's still gp2)? 

Is there any way to work around this behavior that I'm missing, or is `kubectl patch` my best bet for now? And does anyone have thoughts on what a good patch for 1.17 would be to fix this cluster-breaking behavior? Change the name to something more generic and remove the default status of this new StorageClass for now? Or ensure that the default status from gp2 is removed -- (more risky for existing clusters IMO)? Happy to take this on if there's some consensus around it.

joshbranham:
[...]

PR #9337, opened 2020-06-11
Bug: Explicitly set default StorageClass to support upgrades

joshbranham:
Dropping the annotation versus setting it to `false` breaks cluster upgrades.

Set the old `gp2` StorageClass explicitly not the default

Fixes https://github.com/kubernetes/kops/issues/9336

### DIFF
--- a/docs/releases/1.17-NOTES.md
+++ b/docs/releases/1.17-NOTES.md
@@ -71,6 +71,8 @@
 
 * Support for the "Legacy" etcd provider has been deprecated. It will not be supported for Kubernetes 1.18 or later. To migrate to the default "Manager" etcd provider see the [etcd migration documentation](../etcd3-migration.md).
 
+* The default StorageClass `gp2` prior to Kops 1.17.0 is no longer the default, replaced by StorageClass `kops-ssd-1-17`.
+
 # Known Issues
 
 * Kops 1.17.0-beta.1 included an update for AWS IAM Authenticator to 0.5.0.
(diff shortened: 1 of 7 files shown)


hakman:
Any thoughts on using `storageclass.kubernetes.io/is-default-class` instead of the `beta` version?

joshbranham:
Does 1.17 honor that annotation as well?

hakman:
The beta one seems deprecated since 1.6:
https://github.com/kubernetes/kubernetes/pull/40088

joshbranham:
> The beta one seems deprecated since 1.6:
> [kubernetes/kubernetes#40088](https://github.com/kubernetes/kubernetes/pull/40088)

Yeah we can change the `kops-ssd-1-17` to use the non-beta label but we will need to use the beta label (and maybe if we want the non-beta) on the `gp2` in order to have the desired behavior.

Otherwise, we go back to the original point of how to make upgrades work on immutable fields, where if we just made `gp2` have these updated settings alongside a breaking change of `you have to run this with --force to upgrade` 🤷

johngmyers:
Since we've released 1.17.0, I don't think we can change the name of the annotation. Otherwise clusters updated with 1.17.0 will encounter this bug when we need to make `kops-ssd-1-17` non-default.

joshbranham:
So I think the only other thing would be to maybe just add some wording to release notes in case someone runs into a conflict on upgrade, its easy enough and zero risk to fix themselves. We can even provide a few one-liners. Thoughts?

olemarkus:
+1 from me!

hakman:
+1 from me also

johngmyers:
The only reason we left the `gp2` StorageClass in the manifest was to remove the annotation upon upgrade. It seems a poor user experience to require a manual step on upgrade. What is the downside of explicitly setting the annotation to `"false"`?

johngmyers:
Ah, that release note commit was added; not a replacement for the first one.
/lgtm

rifelpet:
I agree that this sounds good to me. We can get this merged now to get it under e2e test coverage, and open a cherry-pick to release-1.17.
I presume we wont have a 1.17.1 release before office hours on friday, so we can take any final feedback there.

/approve

PR #9340, opened 2020-06-11
Store terraform launchtemplate userdata in plaintext rather than b64

rifelpet:
AWS launch templates require their userdata be submitted in base64. terraform has a [filebase64 function](https://www.terraform.io/docs/configuration/functions/filebase64.html) for this purpose, in fact their [aws_launch_template example](https://www.terraform.io/docs/providers/aws/r/launch_template.html#example-usage) uses it. Using it allows us to store the userdata contents in plaintext and terraform performs the encoding, rather than storing it encoded. This makes it easier to grok the userdata contents and any changes.

I'm not super happy with changing the AddFile function signature and updating every reference to it in cloudup.

WIP while I think of better alternatives (feedback welcome)


rifelpet:
`make verify-terraform` failed because filebase64 isnt in terraform 0.11, so this may need to wait until we drop 0.11 support or we make it handle both cases :(

hakman:
When is the plan to drop terraform 0.11? I kind of like this more 😄 .

rifelpet:
I'd love for us to drop terraform 0.11 support asap, but we should probably add deprecation warnings and announce it according to our deprecation policy. Perhaps another office hours topic?

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
```diff
@@ -212,11 +210,10 @@ func (t *LaunchTemplate) RenderTerraform(target *terraform.TerraformTarget, a, e
 		if err != nil {
 			return err
 		}
-		b64d := base64.StdEncoding.EncodeToString(d)
-		if b64d != "" {
-			b64UserDataResource := fi.WrapResource(fi.NewStringResource(b64d))
+		if d != nil {
+			userDataResource := fi.WrapResource(fi.NewBytesResource(d))
```
rifelpet, line 214: I suppose i could check the value of the terraform feature flag here and conditionally encode the userdata contents, but that feels ugly. any other ideas?

rifelpet: or add a new `target.AddBase64File` function that wraps `target.AddFile`

hakman: Seems like it's in the nature of the feature flag to be ugly. A nice reminder that has to go away soon. 😄
I would go with the flag.

rifelpet: done 👍  `hack/verify-terraform.sh` passes too. I'm not sure why there isnt a prow job for it, i'll look into creating one.

hakman: LGTM, but still WIP and also the new job will need some work to pass.

hack/verify-terraform.sh
```diff
@@ -36,7 +36,7 @@ while IFS= read -r -d '' -u 3 test_dir; do
   cluster=$(basename "${test_dir}")
   kube::util::array_contains "${cluster}" "${CLUSTERS_0_11[@]}" && tag=$TAG_0_11 || tag=$TAG_0_12
 
-  docker run --rm -it -v "${test_dir}":"${test_dir}" -w "${test_dir}" --entrypoint=sh hashicorp/terraform:$tag -c '/bin/terraform init >/dev/null && /bin/terraform validate' || RC=$?
+  docker run --rm -v "${test_dir}":"${test_dir}" -w "${test_dir}" --entrypoint=sh hashicorp/terraform:$tag -c '/bin/terraform init >/dev/null && /bin/terraform validate' || RC=$?
```
hakman, line 39: Any reason the tag also contains patch version?
For 0.11 I not much will change and for 0.12 I think we want to know if something breaks.

rifelpet: Hashicorp doesn't have major.minor tags: https://hub.docker.com/r/hashicorp/terraform/tags

I suppose we could keep the `0.11.14` tag and use `latest` for 0.12+ ? I may address that in a followup PR though.

hakman: 👍

PR #9341, opened 2020-06-11
WIP: Autopublish IAM Roles for Service Accounts JWKS doc

justinsb:
This extends #9271 to automatically publish the JWKS document to a "discovery bucket", copying it from the apiserver.  We also use this for dns-controller, as a first pod, and start refactoring the IAM code to generate both node or pod roles.

Though this is "only" 2000 lines, this is an overall PR and I plan on sending smaller PRs.  But this should give the big picture!


justinsb:
cc @rifelpet

pkg/model/components/discovery.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if len(kubeAPIServer.APIAudiences) == 0 {
+		kubeAPIServer.APIAudiences = []string{"kubernetes.svc.default"}
+	}
+
+	if kubeAPIServer.ServiceAccountIssuer == nil {
+		var publicURL string
+		switch p := p.(type) {
+		case *vfs.S3Path:
+			publicURL = "https://" + p.Bucket() + ".s3.amazonaws.com/" + p.Path()
```
johngmyers: Presumably this needs to be pushed down into the vfs interface as a new method.

pkg/apis/kops/v1alpha2/cluster.go
```diff
@@ -191,6 +191,12 @@ type ClusterSpec struct {
 	RollingUpdate *RollingUpdate `json:"rollingUpdate,omitempty"`
 	// ServiceOIDCProvider defines the OIDC provider setup for the cluster (AWS only)
 	ServiceOIDCProvider *ServiceOIDCProviderSpec `json:"serviceOIDCProvider,omitempty"`
+
+	Discovery *DiscoverySpec `json:"discovery,omitempty"`
+}
+
+type DiscoverySpec struct {
+	Base string `json:"base,omitempty"`
```
johngmyers, line 199: Should document this as being a VFS path.

justinsb:
I scanned and it looks like all the functionality here is now present :-)

/close

PR #9334, opened 2020-06-11
Add "--selinux-enabled" flag for Docker

hakman:
https://success.docker.com/article/how-to-set-selinux-file-contexts-when-using-a-custom-docker-data-root

I don't thin the current solution does what is expected based on the Docker docs: https://docs.docker.com/engine/reference/commandline/dockerd/#docker-runtime-execution-options

This PR should work better for the desired use case. On my ubuntu box I see this message:
```
Jun 11 10:21:46 ip-10-4-211-22 dockerd[4928]: time="2020-06-11T10:21:46.664002518Z"
level=warning msg="Docker could not enable SELinux on the host system"
```

Ref: https://github.com/kubernetes/test-infra/pull/17902

/cc @bertinatto @rifelpet

PR #9346, opened 2020-06-12
Temporarily use containerd from Docker packages

hakman:
Docker builds the containerd binaries differently than containerd does.
This makes possible some things like enabling SELinux, which otherwise as described in https://github.com/kubernetes/kops/pull/9334#issuecomment-642601044.

Using containerd from the docker packages should ensure better compatibility.

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -69,11 +69,11 @@ var containerdVersions = []packageVersion{
 		PackageVersion: "1.2.13",
 		PlainBinary:    true,
 		Architectures:  []architectures.Architecture{architectures.ArchitectureAmd64},
-		Source:         "https://storage.googleapis.com/cri-containerd-release/cri-containerd-1.2.13.linux-amd64.tar.gz",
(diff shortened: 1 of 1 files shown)


justinsb:
My knee-jerk reaction is that we should stick to the community packages; I was very happy with the work you did that reduced our dependencies on the docker.com packages, as they've bitten us repeatedly in the past.

Is there a containerd issue tracking the selinux issue?

hakman:
@justinsb I was quite happy with that also and would like to keep the same direction in the future.
I think the issue here is https://github.com/containerd/containerd/pull/4183, which was not cherry-picked to release branches.

hakman:
@justinsb I created some PRs to address this in future releases, and change again to the community packages. Do you think it's ok to switch to Docker packages until then?
https://github.com/containerd/containerd/pull/4318
https://github.com/containerd/containerd/pull/4319

hakman:
Both PRs for containerd were accepted and next release will be built with SELinux support.
As is, this change will just use Docker temporarily and should be reverted once a new version of containerd is released.

rifelpet:
As discussed during office hours, we decided this is acceptable for a temporary solution until the next containerd release 👍 
/lgtm
/approve

PR #9347, opened 2020-06-12
Issue kubecfg and kops certs in nodeup

johngmyers:


justinsb:
This LGTM - thanks for doing this.

At some point, we're going to hit the default credentials used by the automatic kubeconfig.  I'd like for us to keep making that flow easy, though the whole topic of authentication is probably worth a discussion to find out what people are using.

/approve
/lgtm

Test failure looks like a flake around IAM rate limiting:

/retest

johngmyers:
@justinsb I actually ran into that on this PR. The e2e tests didn't get too far without the kubecfg cert generated by pki.go. For now, the masters have their own, separate, kubecfg certs generated by nodeup.

I do think we should significantly reduce the lifetime of the cert put into the generated kubeconfig, but that's a separate PR.

PR #9348, opened 2020-06-12
Create separate field for disabling rolling updates

### LINKED ISSUE #7685: Prevent instance group from being managed by rolling-update

johngmyers:
On further reflection, I think my idea of having rolling updates for an instancegroup be disabled by setting both `maxSurge` and `maxUnavailable` to zero is too indirect. This PR makes that behavior be controlled by a new setting.

Fixes #7685 
Intended to be backported to 1.18.


gjtempleton:
/lgtm

I definitely agree with this being explicit rather than implicit.

johngmyers:
One issue might be naming. It doesn't disable rolling updates entirely, it just disables the drain and terminate behavior. Perhaps `DrainAndDelete`?

rdrgmnzs:
After thinking about it I actually like the current naming, I think `DrainAndDelete` may be even more ambivalent from a users perspective.

johngmyers:
Let's just check on field naming in Office Hours
/hold

pkg/apis/kops/cluster.go
```diff
@@ -695,11 +695,13 @@ type DNSControllerGossipConfig struct {
 }
 
 type RollingUpdate struct {
+	// DrainAndTerminate enables draining and terminating nodes during rolling updates.
+	// Defaults to true.
+	DrainAndTerminate *bool `json:"enabled,omitempty"`
```
justinsb, line 700: Looks like the json value needs updating, but I like the new name!

pkg/apis/kops/validation/validation.go
```diff
@@ -1025,8 +1025,10 @@ func validateDockerConfig(config *kops.DockerConfig, fldPath *field.Path) field.
 
 func validateRollingUpdate(rollingUpdate *kops.RollingUpdate, fldpath *field.Path, onMasterInstanceGroup bool) field.ErrorList {
 	allErrs := field.ErrorList{}
+	var err error
+	unavailable := 1
```
justinsb, line 1029: Nit: `maxUnavailable`?  (though I realize this is pre-existing, so /shrug)

pkg/apis/kops/validation/validation.go
```diff
@@ -1046,6 +1048,9 @@ func validateRollingUpdate(rollingUpdate *kops.RollingUpdate, fldpath *field.Pat
 		} else if surge < 0 {
 			allErrs = append(allErrs, field.Invalid(fldpath.Child("maxSurge"), rollingUpdate.MaxSurge, "Cannot be negative"))
 		}
+		if unavailable == 0 && surge == 0 {
+			allErrs = append(allErrs, field.Forbidden(fldpath.Child("maxSurge"), "Cannot be zero if maxUnavailable is zero"))
```
justinsb, line 1052: This is what we told users to do before, IIUC.  Should we have a hint like "set drainAndTerminate: false to disable rolling updates" ?

But ... I don't think we can do this, if the object was previously valid I think our apimachinery requires that it still be valid.  We should see what happens, but I suspect a warning is going to be easier for everyone here (at least for this api version, in the next api version we can map 0/0 to drainAndTerminate=false, I think)

pkg/instancegroups/instancegroups.go
```diff
@@ -108,11 +108,6 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(ctx context.Context, c
 	}
 	maxConcurrency := maxSurge + settings.MaxUnavailable.IntValue()
 
-	if maxConcurrency == 0 {
-		klog.Infof("Rolling updates for InstanceGroup %s are disabled", group.InstanceGroup.Name)
-		return nil
```
justinsb, line 113: I think apimachinery rules would allow us to fail the rolling update with 0/0 - as long as we don't make valid objects suddenly invalid.

Because we've shipped 0/0 as meaning "no rolling updates", I do think it's more user-friendly here also to treat 0/0 as not enabled, but log a warning / deprecate it.

justinsb:
Thanks @johngmyers ... one leftover json field name, but also if we told users to set 0/0, we might not be able to now mark that invalid, and also I think we should probably continue to honor 0/0 as meaning "off", at least for some deprecation period.

johngmyers:
@justinsb 0/0 hasn't been in a stable release yet, so I think we can skip a deprecation period.

justinsb:
Ah - that's good news.  Fine to disallow 0/0 then :-)

pkg/apis/kops/cluster.go
johngmyers: Fixed in subsequent commit.

pkg/apis/kops/validation/validation.go
johngmyers: I had taken the "max" off of both `MaxUnavailable` and `MaxSurge` here for brevity.

justinsb:
Thanks @johngmyers - doubly important to get this in so we don't have to deprecate 0/0 :-)

/approve
/lgtm

Issue #9251, opened 2020-06-02
master unavailable after update from 1.7.4 to 1.9.6


ssdowdavalon:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
1.16.2

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
1.18.2 (and 1.7.4 to talk to the 1.7.4 cluster)

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
kops rolling-update cluster --master-interval=10m --node-interval=10m --yes

**5. What happened after the commands executed?**
master not healthy after update, stopping rolling-update: "error validating cluster after removing a node: cluster did not validate within a duration of \"15m0s\""

**6. What did you expect to happen?**
I expected nodes to be upgraded to 1.9.6.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**
```
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2017-10-05T18:47:09Z"
  generation: 5
  name: k8s-it.DO.MAIN
spec:
  api:
    loadBalancer:
      type: Public
  authorization:
    alwaysAllow: {}
  channel: stable
  cloudProvider: aws
  configBase: s3://k8s-it.DO.MAIN-state-store/k8s-it.DO.MAIN
  dnsZone: k8s-it.DO.MAIN
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-us-east-1c
      name: c
    name: main
  - etcdMembers:
    - instanceGroup: master-us-east-1c
      name: c
    name: events
  iam:
    legacy: true
  kubernetesApiAccess:
  - 0.0.0.0/0
  kubernetesVersion: 1.9.6
  masterInternalName: api.internal.k8s-it.DO.MAIN
  masterPublicName: api.k8s-it.DO.MAIN
  networkCIDR: 172.20.0.0/16
  networking:
    calico: {}
  nonMasqueradeCIDR: 100.64.0.0/10
  sshAccess:
  - 0.0.0.0/0
  subnets:
  - cidr: 172.20.32.0/19
    name: us-east-1c
    type: Private
    zone: us-east-1c
  - cidr: 172.20.0.0/22
[...]

PR #9352, opened 2020-06-12
Simplified form of IAM Roles for ServiceAccounts

justinsb:
We create a featureflag "PublicJWKS" which runs in a simpler
configuration, where we expose the JWKS directly from the apiserver.

This has some limitations:

* Only works when we have a public ELB
* Requires enabling anonymous-auth on the apiserver

But it is a simpler configuration, and likely we can pare down the
code and get this under e2e before tackling some of the more
complicated scenarios.

Edit:

To try this out (but note that this isn't really a production config yet because of anonymous auth):

```
export KOPS_FEATURE_FLAGS="UseServiceAccountIAM,PublicJWKS"
kops create cluster foo.example.com --zones us-east-2a --api-loadbalancer-type public
```


rifelpet:
So we think this will be the PR that we'll land first between this, #9341 and #9251 

Based on this comment: https://github.com/kubernetes/kops/pull/9271#discussion_r439759704 the remaining items for this PR are:

* Remove the API changes. `pkg/model/awsmodel/oidc_provider.go` will only run if the feature flag is enabled. Also remove the creation of the jwks `fitasks.ManagedFile`s.
* Add validation if the feature flag is enabled:
  * cluster cannot use gossip, must use public DNS
  * must have an internet-accessible api load balancer
  * must not use an ACM certificate on the load balancer ( i think this is required? or we'd have to reference the ACM CA for thumbprints. I'm not sure we'd want to handle that complexity for initial support of this feature)
  * there may be additional requirement's well need to check here...
* Hardcode issuer URL in AWS OIDC Provider to use the cluster's api domain name and cluster's CA thumbprint
* Ensure unauthenticated requests can reach the apiserver's jwks endpoints
* Update `pkg/resources/aws/aws.go` to look for OIDC Providers with issuer URLs that match the cluster's api domain name

I'm sure theres stuff I'm missing but hopefully that makes sense so far.

justinsb:
/retest

prow picked up totally the wrong base

pkg/model/iam/types.go
```diff
@@ -29,3 +31,22 @@ func ParseStatements(policy string) ([]*Statement, error) {
...
+	AWSAccountID string
+
+	Cluster *kops.Cluster
+}
+
+// IAMNameForPodRole determines the name of the IAM Role and Instance Profile to use for the pod role
+func (b *IAMModelContext) IAMNameForPodRole(role PodRole) string {
+	serviceAccount := ServiceAccountForPodRole(role)
+
+	return b.ClusterName() + ".sa." + serviceAccount.Namespace + "." + serviceAccount.Name
```
rifelpet, line 46: our existing roles are named:
* `masters.$clustername`
* `nodes.$clustername`

so what if we used one of these formats?
* `pods.$clustername.$namespace.$name`
* `serviceaccounts.$clustername.$namespace.$name`
* `sa.$clustername.$namespace.$name`

Roles also support paths which are included in the ARN but are separate fields from the role name itself, so maybe we could use those as well.

justinsb: A good point!  If we're putting the prefix before the cluster name, would `$name.$namespace.sa.$clustername` be a good choice?

justinsb:
Changed the per-service-account role name to $name.$namespace.sa.$clustername per your nudge @rifelpet ... we can easily change to other values also.  Using the path could also be a good option - I guess that would be $name.$namespace.sa but we could put it in the $clustername path?

Also removing WIP - this seems to work well now, though it's behind a featureflag.  We can get it under some e2e tests...

pkg/kubemanifest/manifest.go
```diff
@@ -120,3 +121,58 @@ func (m *Object) APIVersion() string {
...
+	}
+
+	if err := yaml.Unmarshal(b, obj); err != nil {
+		return fmt.Errorf("error unmarshaling subobject %s: %v", humanFields, err)
+	}
+
+	return nil
+}
+
+// Set parses a subfield from an object
```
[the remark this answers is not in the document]
justinsb: Fixed (and cleaned up generally to be more self-documenting) - thanks!

cmd/kops/integration_test.go
```diff
@@ -52,17 +52,21 @@ import (
...
-	private             bool
-	zones               int
-	expectPolicies      bool
+	clusterName    string
+	srcDir         string
+	version        string
+	private        bool
+	zones          int
+	expectPolicies bool
+	// expectPodRoles is true if we expect to assign per-pod IAM roles (instead of just using the node roles)
```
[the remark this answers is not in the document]
justinsb: Thanks - renamed!

nodeup/pkg/model/secrets.go
```diff
@@ -120,6 +120,10 @@ func (b *SecretBuilder) Build(c *fi.ModelBuilderContext) error {
 			Subject:        nodetasks.PKIXName{CommonName: "kubernetes-master"},
 			AlternateNames: alternateNames,
 		}
+
+		// Including the CA certificate is more correct, and is needed for e.g. AWS WebIdentity federation
+		issueCert.WriteCertificateChain = true
```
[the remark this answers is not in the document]
justinsb: Good suggestion - done!

pkg/featureflag/featureflag.go
```diff
@@ -92,6 +92,10 @@ var (
 	Terraform012 = New("Terraform-0.12", Bool(true))
 	// LegacyIAM will permit use of legacy IAM permissions.
 	LegacyIAM = New("LegacyIAM", Bool(false))
+	// UsePodIAM controls whether we use pod-level IAM permissions for our system pods.
+	UsePodIAM = New("UsePodIAM", Bool(false))
```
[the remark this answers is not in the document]
justinsb: Renamed to UseServiceAccountIAM

justinsb:
Thanks for the review @johngmyers ... made the suggested changes.  Going to carve off the functionality from this one to make it smaller.

#9812 is the PR to include the CA certificate (i.e. the first commit)

justinsb:
Exposing JWKS via a feature-flag is #9813

justinsb:
Refactoring the Context is #9814

pkg/model/iam/types.go
rifelpet: Regarding naming format, it'd be nice if they sorted such that all of a cluster's roles would grouped together.

We kinda have that currently with our roles beginning with `masters` and `nodes` so that all of the kops roles are near eachother in a list of dozens or hundreds of roles, but adding `dns-controller` or other names would spread them around a bit more.

Just a nit if we have the opportunity to choose, not a big deal either way.

pkg/model/iam.go
```diff
@@ -48,6 +49,20 @@ const RolePolicyTemplate = `{
   ]
 }`
 
+const PodRolePolicyTemplate = `{
```
johngmyers, line 52: ServiceAccountRolePolicyTemplate?

pkg/model/iam/pod_roles.go
```diff
@@ -19,10 +19,40 @@ package iam
...
+
+const (
+	// PodRoleEmpty indicates we are not using a PodRole
+	PodRoleEmpty PodRole = ""
+	// PodRoleDNSController is the role used by the dns-controller pods
+	PodRoleDNSController PodRole = "dns-controller"
+)
+
+// PodOrNodeRole is used to specify generation of permissions for either a pod or a node
+type PodOrNodeRole struct {
```
johngmyers, line 40: ServiceAccountOrNodeRole?

pkg/model/iam/pod_roles.go
```diff
@@ -19,10 +19,40 @@ package iam
...
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/featureflag"
+	"k8s.io/kops/pkg/wellknownusers"
 )
 
+// PodRole holds a pod-specific IAM policy
+type PodRole string
```
johngmyers, line 30: ServiceAccountRole?

pkg/model/iam.go
justinsb: Did a bulk rename - thanks!

pkg/model/iam/pod_roles.go
johngmyers: ...or is just `Role` possible?

johngmyers:
Comments before top-level declarations are supposed to end with a period. Many of those added here are missing that.

pkg/model/iam.go
```diff
@@ -72,33 +87,57 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
-		err := b.buildIAMTasks(igRole, iamName, c, false)
+		err := b.buildIAMTasks(role, iamName, c, false)
 		if err != nil {
 			return err
 		}
 	}
 
+	// Generate IAM tasks for service-account roles
+	if featureflag.UseServiceAccountIAM.Enabled() {
+		serviceAccountRoles := []iam.ServiceAccountRole{iam.ServiceAccountRoleDNSController}
```
johngmyers, line 114: This doesn't seem a good place for determining which service account roles need to be created. For example, if dns-controller is disabled then we should not build its service account role. That information is more properly in bootstrapchannelbuilder or the respective addon manifest templates.

I'm thinking that BootstrapChannelBuilder could build the IAM tasks, either directly or as a consequence of the manifest template calling a template function with the name of the role as a parameter.

pkg/model/iam/iam_builder.go
```diff
@@ -140,13 +141,40 @@ func (b *PolicyBuilder) BuildAWSPolicy() (*Policy, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate AWS IAM Policy for Master Instance Group: %v", err)
+		}
+
 	}
 
 	return p, nil
 }
 
+// BuildAWSPolicyServiceAccountRole generates a custom policy for a pod IAM role
```
johngmyers, line 162:
```suggestion
// BuildAWSPolicyServiceAccountRole generates a custom policy for a ServiceAccount IAM role.
```

pkg/model/components/addonmanifests/dnscontroller/remap.go
```diff
@@ -0,0 +1,100 @@
...
+	addonsapi "k8s.io/kops/channels/pkg/api"
+	"k8s.io/kops/pkg/featureflag"
+	"k8s.io/kops/pkg/kubemanifest"
+	"k8s.io/kops/pkg/model"
+	"k8s.io/kops/pkg/model/iam"
+)
+
+// Remap remaps the dns-controller addon
+func Remap(context *model.KopsModelContext, addon *addonsapi.AddonSpec, objects []*kubemanifest.Object) error {
+	if !featureflag.UseServiceAccountIAM.Enabled() {
```
johngmyers, line 33: Consider coding this as `context.UseServiceAccountIAM()`, to ease later transition from feature flag to field in cluster spec.

pkg/model/iam/iam_builder.go
```diff
@@ -571,13 +602,22 @@ func addECRPermissions(p *Policy) {
 }
 
 func (b *PolicyBuilder) addRoute53Permissions(p *Policy) {
```
johngmyers, line 604: I think this should be called `addDNSControllerPermissions()`. What needs the permissions is more important than what the permissions give access to.

pkg/model/iam/iam_builder.go
```diff
@@ -571,13 +602,22 @@ func addECRPermissions(p *Policy) {
...
+			return
+		}
+	} else if b.Role.ServiceAccountRole != ServiceAccountRoleDNSController {
+		// Only the dns-controller gets route53 permissions
 		return
 	}
 
 	// Legacy IAM permissions for node roles
-	if b.Cluster.Spec.IAM.Legacy {
+	if b.Cluster.Spec.IAM.Legacy && b.Role.IsNodeRole() {
```
johngmyers, line 620: I would suggest prohibiting having both LegacyIAM and UseServiceAccountIAM enabled at the same time.

pkg/model/iam/iam_builder.go
justinsb: I like that distinction that we can make now :-)

pkg/model/iam.go
```diff
@@ -48,6 +49,20 @@ const RolePolicyTemplate = `{
...
+    {
+      "Effect": "Allow",
+      "Principal": { "Federated": "{{ FederatedPrincipal }}" },
+      "Action": "sts:AssumeRoleWithWebIdentity",
+      "Condition": {
+        "StringEquals": { "{{OIDCProvider}}:sub": "{{ OIDCSub }}" }
+      }
+    }
+  ]
+}`
```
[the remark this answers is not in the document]
justinsb: Done for ServiceAccount roles.  For Node roles it would change the whitespace for existing cluster policies, so I left it using the string method - but I removed the templating as it's not really needed.

pkg/model/components/addonmanifests/dnscontroller/remap.go
justinsb: Done - good call!

pkg/model/iam/iam_builder.go
justinsb: Done - good call!

pkg/model/iam/pod_roles.go
```diff
@@ -19,10 +19,50 @@ package iam
...
+
+const (
+	// ServiceAccountRoleEmpty indicates we are not using a ServiceAccountRole
+	ServiceAccountRoleEmpty ServiceAccountRole = ""
+	// ServiceAccountRoleDNSController is the role used by the dns-controller pods
+	ServiceAccountRoleDNSController ServiceAccountRole = "dns-controller"
+)
+
+// PodOrNodeRole is used to specify generation of permissions for either a pod or a node
+type PodOrNodeRole struct {
```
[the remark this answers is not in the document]
justinsb: I gave it a go, the problem was that because go interfaces aren't explicit, I found the end result to be really unintuitive and unobvious.  You end up with these weird interfaces that are basically `void *`.  I suggest we try it again in go2.

Here's the core interface I tried - maybe you had a different one in mind?

```
// IAMRole is implemented by our IAM scoping objects: ServiceAccountRole and InstanceGroupRole.
type IAMRole interface {
	// HasServiceAccountRole returns true if we include the specified ServiceAccountRole.
	HasServiceAccountRole(role ServiceAccountRole) bool

	// HasNodeRole returns true if we include the specified NodeRole.
	HasNodeRole(role kops.InstanceGroupRole) bool
}
```

pkg/model/components/addonmanifests/dnscontroller/remap.go
```diff
@@ -0,0 +1,100 @@
...
+	corev1 "k8s.io/api/core/v1"
+	addonsapi "k8s.io/kops/channels/pkg/api"
+	"k8s.io/kops/pkg/featureflag"
+	"k8s.io/kops/pkg/kubemanifest"
+	"k8s.io/kops/pkg/model"
+	"k8s.io/kops/pkg/model/iam"
+)
+
+// Remap remaps the dns-controller addon
+func Remap(context *model.KopsModelContext, addon *addonsapi.AddonSpec, objects []*kubemanifest.Object) error {
```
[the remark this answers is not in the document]
justinsb: The challenge is that it's going to be different on each cloud, and we're going to want to do this for several pods, so this is only going to get more verbose.  I also see this as an implementation detail - we could have an Route53 proxy, for example, so even on one cloud we might end up with multiple overlays.

pkg/model/iam.go
justinsb: Good idea - moved to BootstrapChannelBuilder.

I'm generally trying to avoid magic templates, so I made it explicit.

I would like in future to have the addon able to declare the actual IAM permissions it needs, ideally by having a CRD that represented the IAM policy ... but that would make this change look small!

justinsb:
/hold

just chasing done one last test case

pkg/model/iam.go
johngmyers: ACK is unfortunately a little underbaked as of yet.

pkg/model/components/addonmanifests/dnscontroller/remap.go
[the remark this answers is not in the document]
johngmyers: I wonder if we could have a template function wrapped around the PodSpec or PodTemplateSpec which injects the things. What I'm most concerned about is having to have a per-addon block of code that knows details of the manifest such as it's a Deployment, which Deployment it is, etc. It's also invisible when looking at the manifest.

As an aside, this is unfortunately not something that Helm can do, which is likely the reason people write mutating admission hooks for this sort of thing. One disadvantage I've found with mutating admission hooks, other than their complexity, is that the pre-injection template still has to pass API validation.

pkg/model/components/addonmanifests/dnscontroller/remap.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	if addon.KubernetesVersion != "" {
+		versionRange, err := semver.ParseRange(addon.KubernetesVersion)
+		if err != nil {
+			return fmt.Errorf("cannot parse KubernetesVersion=%q", addon.KubernetesVersion)
+		}
+
+		if !kubernetesRangesIntersect(versionRange, semver.MustParseRange(">= 1.19.0")) {
+			// Skip; this is an older manifest
```
johngmyers, line 43: This logic needs to be in sync with the iam_builder logic that decides whether to give the privileges to the service account or node

pkg/model/iam/iam_builder.go
```diff
@@ -570,14 +606,29 @@ func addECRPermissions(p *Policy) {
...
```
johngmyers, line 631: No longer need this change. Legacy now implies all roles are node roles.

pkg/model/iam/pod_roles.go
```diff
@@ -19,10 +19,50 @@ package iam
...
+	return r.ServiceAccountRole != ServiceAccountRoleEmpty
+}
+
+// IsNodeRole is true if this is a node-scoped role
+func (r *NodeOrServiceAccountRole) IsNodeRole() bool {
+	return r.NodeRole != ""
+}
+
+// ServiceAccountForServiceAccountRole returns the kubernetes service account used by pods with the specified role
+func ServiceAccountForServiceAccountRole(serviceAccountRole ServiceAccountRole) types.NamespacedName {
```
johngmyers, line 58: Should this be a receiver?

pkg/model/iam/pod_roles.go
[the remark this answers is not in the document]
johngmyers: I was thinking of pushing the relevant `buildAWSIAMRolePolicy()` logic under an interface method. And the `HasServiceAccountRole` method would return the name of any service account.

Basically I'm applying the principle that code that does type assertions (or tests for existence of union fields) suggests that the branches under the test might be better as an interface method.

pkg/model/components/addonmanifests/dnscontroller/remap.go
justinsb: I _think_ this is OK;  this is saying "only try to patch the latest versions of the manifests".  There is a deeper issue though, which is how to enable this on an existing cluster ... I talk about that in the top level comment.

justinsb:
Thanks for explaining the interface comment @johngmyers  - I implemented that and it seems to work really well.  I think we could go even further with this reorganization, but there's likely a lot of moving around of code that would make this PR unreadably large again.

Your first comment (about keeping the logic in sync) gets at a deeper point  - we don't have a great way to enable IRSA on an existing cluster right now.  I suspect it'll actually work out, but it's possible we could hit a circular dependency while trying to publish the jwks document.

I think the answer will likely be:

* The production flow will involve kops CLI publishing the JWKS document to an S3 bucket (i.e. no dependency on apiserver running nor on kops-controller having permissions to publish to an S3 bucket) 
* We'll therefore be able to enable IRSA and use it in one step; we might have some pods that temporarily don't have valid credentials but it should be a matter of a few seconds until everything converges.
* When we graduate the feature-flag we'll have to include kubernetes-versions selection for whatever version we enable it on.

I don't particularly want to tackle that in this PR though - my goal here is to get test coverage on IRSA so we can get the feature enabled in upstream kubernetes (and demonstrate that kops is a good place to build test coverage of new functionality!).

pkg/model/components/addonmanifests/dnscontroller/remap.go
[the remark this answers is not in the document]
justinsb: My hope (over say the 2 year horizon) is that this is just an implementation detail, and that it gets easier to use kubernetes service accounts everywhere.  For example the AWS SDK could do the service account token exchange directly; or kubernetes could issue client-certificates to pods, rendering the whole per-audience service-account concept moot.

We could mark the pod template in some way, but because we end up targeting multiple places I expect it would end up as boilerplate wrapping the template that gets ignored.  I don't mind doing that - it's likely a good idea because our current logic can't cope with multiple containers - we just don't need it right now so it's difficult to have an intuition for what to do.

I agree it would be nice to be able to see the final yaml more easily.  I think it's visible in our diffing engine, but we could render the final controllers to let people see it more clearly, or something along those lines.

pkg/model/components/addonmanifests/dnscontroller/remap.go
johngmyers: If the Kubernetes version is 1.11 but the feature flag is on, then iam_builder will withhold the permissions from the node role, yet this won't inject the projected volume. So dns-controller won't have the permissions.

johngmyers:
If the plan is to condition IRSA support on the manifest version, then different service accounts are likely to have different Kubernetes-version cutoffs.

Now when we drop support for k8s 1.11 in kops 1.20 a lot of variant manifests will drop away. But there are a few CNIs that vary over 1.16 and spotinst varies over 1.14.

pkg/model/context.go
```diff
@@ -408,3 +408,8 @@ func (m *KopsModelContext) NodePortRange() (utilnet.PortRange, error) {
 
 	return defaultServiceNodePortRange, nil
 }
+
+// UseServiceAccountIAM returns true if we are using service-account bound IAM roles.
+func (m *KopsModelContext) UseServiceAccountIAM() bool {
+	return featureflag.UseServiceAccountIAM.Enabled()
```
johngmyers, line 414:
```suggestion
	return featureflag.UseServiceAccountIAM.Enabled() && m.IsKubernetesGTE("1.12")
```

pkg/model/components/addonmanifests/dnscontroller/remap.go
justinsb: Right - there is "caveat emptor" with feature flags, I see you've got it with your suggested change though so I'm going to incorporate that!

pkg/model/iam/pod_roles.go
justinsb: Done!

justinsb:
Merged your suggested change @johngmyers and then passed down `UseServiceAccountIAM` to avoid duplicating the logic.  I guess in some version (e.g. 1.21) we'll enable a "real" flag/field for enabling this; my hope is that we won't have to tweak the manifests at that stage, but if we need to we can cut a new version of the manifests to address any incompatibilities.

I've also been thinking more about your point about functions/templates etc... there's actually a ton of overlap with the idea of operators and their ability to do transforms nicely, but the challenge of the opacity of the resulting manifest.  I don't really have a fully formed answer yet, but there's something there I think...

pkg/model/iam/pod_roles.go
justinsb: It's now `iam.Subject` now that it's an interface - I think that's the correct term!

justinsb:
(Had to add another commit to workaround AWS IAM vs Golang JSON restrictions on the generated IAM policy)

pkg/model/components/addonmanifests/dnscontroller/remap.go
johngmyers: This check can probably be removed now.

pkg/model/components/addonmanifests/dnscontroller/remap.go
```diff
@@ -0,0 +1,100 @@
...
+		return fmt.Errorf("failed to parse spec.template.spec from Deployment: %v", err)
+	}
+
+	containers := podSpec.Containers
+	if len(containers) != 1 {
+		return fmt.Errorf("expected exactly one container in dns-controller Deployment, found %d", len(containers))
+	}
+	container := &containers[0]
+
+	if context.UseServiceAccountIAM() {
```
johngmyers, line 74: Cannot be false here.

pkg/model/iam/iam_builder.go
```diff
@@ -78,9 +77,110 @@ type Condition map[string]interface{}
...
+	if err != nil {
+		j.err = err
+		return
+	}
+	j.WriteLiteral(b)
+}
+
+// MarshalJSON formats the IAM statement for the AWS IAM restrictions.
+// For example, `Resource: []` is not allowed, but golang would force us to use pointers.
+func (s *Statement) MarshalJSON() ([]byte, error) {
```
johngmyers, line 144: We might want unit tests for this.

pkg/model/iam/pod_roles.go
```diff
@@ -19,10 +19,68 @@ package iam
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/featureflag"
+	"k8s.io/kops/pkg/wellknownusers"
 )
 
+// Subject represents an IAM identity, to which permissions are granted.
+// It is implemented by NodeRole objects and per-ServiceAccount objects.
+type Subject interface {
```
johngmyers, line 31: Perhaps rename this file subject.go

pkg/model/iam/pod_roles.go
```diff
@@ -31,3 +89,74 @@ func ServiceAccountIssuer(clusterName string, clusterSpec *kops.ClusterSpec) (st
...
+
+// AddServiceAccountRole adds the appropriate mounts / env vars to enable a pod to use a service-account role
+func AddServiceAccountRole(context *IAMModelContext, podSpec *corev1.PodSpec, container *corev1.Container, serviceAccountRole Subject) error {
+	cloudProvider := kops.CloudProviderID(context.Cluster.Spec.CloudProvider)
+
+	switch cloudProvider {
+	case kops.CloudProviderAWS:
+		return addServiceAccountRoleForAWS(context, podSpec, container, serviceAccountRole)
+	default:
+		return fmt.Errorf("Pod-level IAM is not yet supported on cloud %T", cloudProvider)
```
johngmyers, line 101:
```suggestion
		return fmt.Errorf("ServiceAccount-level IAM is not yet supported on cloud %T", cloudProvider)
```

johngmyers (commented):
It looks like there's only minor stuff left. I think it's really close.

There are some parameters of type `iam.Subject` that are named `role`; perhaps they should be named `subject`.

pkg/model/iam/iam_builder.go
justinsb: Good point - we had a unit test and added test cases to cover the new cases (e.g. no Resources).

pkg/model/iam/pod_roles.go
justinsb: Good idea - done!

pkg/model/components/addonmanifests/dnscontroller/remap.go
justinsb: Thanks - removed!

PR #9354, opened 2020-06-13
Continue refactoring certs into nodeup

johngmyers:
Finally got the api-server server cert. It took a lot of refactoring to get here.


rifelpet:
/lgtm
/approve

When someone upgrades to a version of Kops that includes these refactors, they'll be left with orphaned certificates in their state store correct? Perhaps we'll want to include a release note on cleaning them up after upgrading.

johngmyers:
They will, but then they'll also have the CA private keys in the state store. This would be more of a concern for keys that can be accessed from worker nodes.

rifelpet:
ah thats true. anything that currently has read access to the orphaned certs and private keys also has read access to the CA private keys 👍

PR #9355, opened 2020-06-13
Move host-network services off of port 8080

johngmyers:
I am unable to test the node-local-dns change.

### DIFF
--- a/cmd/kube-apiserver-healthcheck/main.go
+++ b/cmd/kube-apiserver-healthcheck/main.go
@@ -28,6 +28,7 @@ import (
 	"os"
 
 	"k8s.io/klog"
+	"k8s.io/kops/pkg/wellknownports"
 )
 
 // healthCheckServer is the http server
@@ -129,7 +130,7 @@ func (s *healthCheckServer) proxyRequest(w http.ResponseWriter, forwardRequest *
 }
 
 func run() error {
-	listen := ":8080"
+	listen := fmt.Sprintf(":%d", wellknownports.KubeAPIServerHealthCheck)
 
 	clientCert := ""
 	clientKey := ""
--- a/pkg/model/components/kubeapiserver/model.go
+++ b/pkg/model/components/kubeapiserver/model.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/kops/pkg/assets"
 	"k8s.io/kops/pkg/k8scodecs"
 	"k8s.io/kops/pkg/model"
+	"k8s.io/kops/pkg/wellknownports"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/fitasks"
 )
@@ -100,7 +101,7 @@ spec:
         # The sidecar serves a healthcheck on the same port,
         # but with a .kube-apiserver-healthcheck prefix
         path: /.kube-apiserver-healthcheck/healthz
-        port: 8080
+        port: %d
         host: 127.0.0.1
       initialDelaySeconds: 5
       timeoutSeconds: 5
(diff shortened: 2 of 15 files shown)


cmd/kube-apiserver-healthcheck/main.go
```diff
@@ -129,7 +130,7 @@ func (s *healthCheckServer) proxyRequest(w http.ResponseWriter, forwardRequest *
 }
 
 func run() error {
-	listen := ":8080"
+	listen := fmt.Sprintf(":%d", wellknownports.KubeAPIServerHealthCheck)
```
olemarkus, line 133: Just to make sure, does the health check have to listen to all ips? I did a quick check to see where reqs were coming from and seemed to be all localhost. It was a very quick test though.

johngmyers: This PR is not changing that aspect.

olemarkus: Fair enough

upup/models/bindata.go
```diff
@@ -14233,7 +14233,7 @@ data:
           force_tcp
         }
         prometheus :9253
-        health {{ KubeDNS.NodeLocalDNS.LocalIP }}:8080
+        health {{ KubeDNS.NodeLocalDNS.LocalIP }}:{{ NodeLocalDNSHealthCheck }}
```
olemarkus, line 14236: I was wondering if this should be configurable in the cluster spec. Then you can use the well-known port as default in `components/kubedns.go` instead of adding the template function. Just a thought.

olemarkus:
Can confirm that this fixes #9245

upup/models/bindata.go
johngmyers: What is the use case for making this port configurable?

Incidentally, I wonder if we could have kops automatically default the value of `clusterDNS` in `kubelet` and `masterKubelet` to `169.254.20.10` when needed rather than just mentioning it in documentation.

olemarkus: Well. Both can conflict with other things, so it isn't unreasonable to have this configurable, I think.

johngmyers: Our approach is to reserve ports in `wellknownports`. If anything else runs in host network and conflicts with those, the other things are what have to move.

pkg/model/components/kubeapiserver/model.go
```diff
@@ -126,7 +127,7 @@ spec:
 func (b *KubeApiserverBuilder) buildHealthcheckSidecar() (*corev1.Pod, error) {
 	// TODO: pull from bundle
 	bundle := "(embedded kube-apiserver-healthcheck manifest)"
-	manifest := []byte(defaultManifest)
+	manifest := []byte(fmt.Sprintf(defaultManifest, wellknownports.KubeAPIServerHealthCheck))
```
justinsb, line 130: Nit: I normally try to avoid using Sprintf with our templates, favoring `{{ HealthCheckPort }}` and `strings.ReplaceAll`,  because the `%d` is harder to read, and has more gotchas (repeated placeholders don't work, it's vaguely possible that a `%` placeholder could appear 'accidentally')

upup/models/bindata.go
justinsb: It is still a breaking change in kops.  That said, it is something we need to do, so maybe we should just reserve a range of ports for kops usage.

I think it's a good idea also to make it as easy as possible to turn on NodeLocalDNS as possible.  Can we change the defaulting logic when nodelocal dns is enabled, but continue to honor an explicit value if the user sets one?

justinsb:
Looks good to me; one nit that `%d` in a template is maybe hard to use; one idea as to whether we should pre-reserve a range of ports to avoid collisions.

/approve
/lgtm

johngmyers:
@justinsb I thought `pkg/wellknownports` was us reserving a range of ports for kops. We perhaps might be more explicit about reserving ~3900-4099 for kops use.

I do think we should enable NodeLocalDNS by default when using CoreDNS.

PR #9356, opened 2020-06-13
Issue kubelet-api cert in nodeup

johngmyers:


pkg/bundle/builder.go
```diff
@@ -239,37 +239,18 @@ func (b *Builder) buildPKIFiles(cluster *kops.Cluster, ig *kops.InstanceGroup, k
 	keys = append(keys, "kube-proxy")
 
 	if ig.IsMaster() {
```
johngmyers, line 241: The fact that kubelet-api was not in here to be removed suggests that baremetal is not being used.

hakman: I think @olemarkus was thinking of removing baremetal.

johngmyers: Filed #9360

justinsb:
Thanks @johngmyers 

/approve
/lgtm

PR #9359, opened 2020-06-14
Fix kube-apiserver-healthcheck image

coreypobrien:
Use a [valid tag](https://hub.docker.com/r/kope/kube-apiserver-healthcheck/tags) for the kube-apiserver-healthcheck image


johngmyers:
This was changed in release-1.18 branch, but we haven't done a release off of master since branching release-1.18.

PR #9360, opened 2020-06-14
Remove the baremetal cloud provider

### LINKED ISSUE #9357: Removing baremetal cloud provider
Following #9166, does anyone mind removing baremetal? Looks like some basic work was done on this about three years ago, but that it was never completed. Later commits are more about general changes to cloud provides.

@justinsb any plans on doing more work on this? With where kops is today, a baremetal provider perhaps overlaps more than usual with kubeadm.

johngmyers:
I found with #9356 that the baremetal provider has been broken for k8s >= 1.7 unless anonymous auth was enabled.

The baremetal provider, being an extremely special case, is hindering development. I believe that
baremetal would be better managed under a cloud provider such as OpenStack.

Fixes #9357


justinsb:
I'm OK to remove this and we can add it back in.  I guess I should get my WIP PRs out of WIP!

/approve
/lgtm

PR #9362, opened 2020-06-14
Require extra flag when updating cluster with downgraded kops version

johngmyers:


pkg/apis/kops/registry/registry.go
```diff
@@ -29,6 +29,8 @@ const (
 	PathCluster = "config"
 	// Path for completed cluster spec in the state store
 	PathClusterCompleted = "cluster.spec"
+	// PathKopsVersion is the path for the version of kops last used to apply the cluster.
+	PathKopsVersionApplied = "kops-version-applied.txt"
```
hakman, line 33: I think this would be closer to `cluster.spec`, `master-a`, `config` that are in the bucket root.
```suggestion
	PathKopsVersionApplied = "kops.version"
```

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -234,6 +239,34 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
...
+				fmt.Printf("\n")
+				fmt.Printf("The cluster was last updated by kops version %s\n", kopsVersionApplied)
+				fmt.Printf("To permit updating by the older version %s, run with the --allow-kops-downgrade flag\n", kopsbase.Version)
+				fmt.Printf("\n")
+				fmt.Printf("%s\n", starline)
+				fmt.Printf("\n")
+				return fmt.Errorf("kops version older than that last used to apply cluster")
+			}
+		} else if err != os.ErrNotExist {
+			return fmt.Errorf("error reading last kops version applied: %v", err)
```
hakman, line 266: This file will not exist before applying this fix the fist time. Maybe should just ignore it?
```suggestion
			klog.Infof("Last used kops version not found; will assume older version")
```

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -234,6 +239,34 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("error parsing config base %q: %v", cluster.Spec.ConfigBase, err)
+	}
+
+	if !c.AllowKopsDowngrade {
+		kopsVersionApplied, err := configBase.Join(registry.PathKopsVersionApplied).ReadFile()
+		if err == nil {
+			version, err := semver.Parse(strings.TrimSpace(string(kopsVersionApplied)))
+			if err != nil {
+				return fmt.Errorf("error parsing last kops version applied: %v", err)
```
hakman, line 252:
```suggestion
				return fmt.Errorf("error parsing last kops version used: %v", err)
```

pkg/apis/kops/registry/registry.go
hakman, line 33:
```suggestion
	// PathKopsVersionUsed is the path for the version of kops last used to apply the cluster.
	PathKopsVersionUsed = "kops.version"
```

upup/pkg/fi/cloudup/apply_cluster.go
johngmyers: The `!=` comparison causes it to ignore `os.ErrNotExist`

upup/pkg/fi/cloudup/apply_cluster.go
johngmyers: It's not the last version used. One can use a downrev or uprev kops to edit the cluster or ig specs; it only matters which version of kops last did an `apply_cluster`.

upup/pkg/fi/cloudup/apply_cluster.go
hakman: Ah, missed that.

pkg/apis/kops/registry/registry.go
johngmyers: Again, it's not the last kops version used.

johngmyers: Having the well-known `.txt` extension helps with troubleshooting.

upup/pkg/fi/cloudup/apply_cluster.go
hakman: I understand this, but it's more from the developer's point of view. "Used" is more user oriented.
In any case, even using "last kops version used to apply changes" would be a bit better.

johngmyers: Perhaps "updated", since the CLI command is `kops update cluster --yes`. The `apply_cluster.go` is mixing terminology a bit.

hakman: That would be even better.

johngmyers: Went with the "updated" terminology.

hakman: Probably you want to update this message also.

pkg/apis/kops/registry/registry.go
```diff
@@ -29,6 +29,8 @@ const (
 	PathCluster = "config"
 	// Path for completed cluster spec in the state store
 	PathClusterCompleted = "cluster.spec"
+	// PathKopsVersionUpdated is the path for the version of kops last used to apply the cluster.
+	PathKopsVersionUpdated = "kops-version-updated.txt"
```
hakman, line 33: Should be enough without updated.
```suggestion
	PathKopsVersionUpdated = "kops-version.txt"
```

rifelpet:
Agreed during office hours that this is good to merge as-is
/approve

PR #9363, opened 2020-06-14
Prefer the GA label for node zone

johngmyers:


rifelpet:
looks like the beta labels were deprecated in 1.17: https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#failure-domainbetakubernetesiozone

/approve

PR #9367, opened 2020-06-15
Bump cilium to 1.7.5

olemarkus:


rifelpet:
is there a reason this is targeting release-1.18 and not master? I don't see this change in master nor an open PR for it.

olemarkus:
I am redoing the templates and config on master. See #9295 etc. Aim to use the 1.8 line on master if cilium release it in time (they are in RC phase now)

rifelpet:
ahh got it, thanks for the clarification.

/approve

rifelpet:
though unfortunately we wont have much test coverage of this since the cilium periodic job(s) use the master branch...

olemarkus:
@rifelpet Wouldn't this do? https://testgrid.k8s.io/kops-k8s-1.18#kops-grid-cilium-u2004-k18
doesn't run that frequent though.

olemarkus:
Or they perhaps uses kops master, but that specific release branch?

hakman:
> Or they perhaps uses kops master, but that specific release branch?

👍

hakman:
@rifelpet should we create optional e2e tests for the popular CNIs that we could run on any PR?

rifelpet:
Ah i like that idea a lot. we could even have them run automatically based on changes to each provider's yaml directory.

hakman:
Right, that hashes test should be a pretty good template for this.

rifelpet:
anyone want to volunteer setting those up? We might be able to use them across multiple branches (master and release-*) too.

hakman:
Let's start with master. I think I may have some time tomorrow to copy / paste these jobs, unless someone else beats me to it.

PR #8119, opened 2019-12-16
Addons: Support arbitrary additional objects

justinsb:
We will be managing cluster addons using CRDs, and so we want to be
able to apply arbitrary objects as part of cluster bringup.

Start by allowing (behind a feature-flag) for arbitrary objects to be
specified.

PR #9374, opened 2020-06-16
First addon operator integration: CoreDNS

justinsb:
Hidden behind a feature-flag, but when the UseAddonOperators flag is set,
we now use the cluster-addons CoreDNS operator instead of our built-in
manifests.

### DIFF
--- a/channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
+++ b/channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
@@ -1,0 +1,266 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.5
+  creationTimestamp: null
+  name: coredns.addons.x-k8s.io
+spec:
+  group: addons.x-k8s.io
+  names:
+    kind: CoreDNS
+    listKind: CoreDNSList
+    plural: coredns
+    singular: coredns
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: CoreDNS is the Schema for the coredns API
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: CoreDNSSpec defines the desired state of CoreDNS
+          properties:
+            channel:
+              description: 'Channel specifies a channel that can be used to resolve a specific addon, eg: stable It will be ignored if Version is specified'
+              type: string
+            corefile:
+              type: string
+            dnsDomain:
+              type: string
+            dnsIP:
+              type: string
+            patches:
+              items:
+                type: object
+              type: array
+            version:
+              description: Version specifies the exact addon version to be deployed, eg 1.2.3 It should not be specified if Channel is specified
+              type: string
+          type: object
+        status:
+          description: CoreDNSStatus defines the observed state of CoreDNS
+          properties:
+            errors:
+              items:
+                type: string
+              type: array
+            healthy:
+              type: boolean
+          required:
+          - healthy
+          type: object
+      type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
+
+---
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: coredns-system
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    k8s-app: coredns-operator
+  name: coredns-operator
+  namespace: coredns-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: coredns-operator
+rules:
+- apiGroups:
+  - addons.x-k8s.io
+  resources:
+  - coredns
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - addons.x-k8s.io
+  resources:
+  - coredns/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - serviceaccounts
+  - services
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resourceNames:
+  - coredns
+  resources:
+  - configmaps
+  - serviceaccounts
+  - services
+  verbs:
+  - delete
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - serviceaccounts
+  - services
+  verbs:
+  - create
+- apiGroups:
+  - apps
+  - extensions
+  resources:
+  - deployments
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - apps
+  - extensions
+  resourceNames:
+  - coredns
+  resources:
+  - deployments
+  verbs:
+  - delete
+  - patch
+  - update
+- apiGroups:
+  - apps
+  - extensions
+  resources:
+  - deployments
+  verbs:
+  - create
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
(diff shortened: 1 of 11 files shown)


justinsb:
Builds on #8119

pkg/client/simple/vfsclientset/addons.go
```diff
@@ -0,0 +1,96 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
johngmyers, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

justinsb: Sadly not ;-)

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
```diff
@@ -0,0 +1,340 @@
...
+    k8s-app: kube-dns
+  name: coredns-operator-leader-election-rolebinding
+  namespace: kube-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: coredns-operator-leader-election-role
+subjects:
+- kind: ServiceAccount
+  name: default
```
johngmyers, line 257: The `default` service account? Those are some really powerful rights being granted here and below. The effort of separating roles is wasted by not separating out service accounts.

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
```diff
@@ -0,0 +1,340 @@
...
+metadata:
+  labels:
+    control-plane: controller-manager
+    k8s-app: kube-dns
+  name: coredns-operator-controller-manager-metrics-service
+  namespace: kube-system
+spec:
+  ports:
+  - name: https
+    port: 8443
```
johngmyers, line 301: Why not 443?

johngmyers:
I am alarmed by the sheer amount of insecurity and risk that this approach appears to be leading to. Code running with almost unbounded privileges and not a whole lot of segregation. And the point appears to be to download and run code directly from external sources, which is diametrically opposed to what my employer wants.

justinsb:
So I think we've advanced *part* of the security posture here.  I've hacked up the coredns operator a bit, but if we agree then I think we can upstream those changes.

The idea is that we can (and should) pre-create the RBAC objects that the addon is going to use: the ServiceAccount, the ClusterRole, the ClusterRoleBinding.  Because ClusterRoles have to be inherited by the operator if they're going to be created, this doesn't actually cost us anything in terms of updates - updates that need new permissions would always require an operator update; that operator update now includes the new permissions.

This makes the operator now into an application-aware health-checker and configurer for the addon itself (primarily the deployment), which is really the goal.

I personally think this makes the addon operator story good from the point of view of "is every operator running with cluster-admin permissions" - the permissions look pretty reasonable to me... but please take a look @johngmyers !

Now I think there are two things that are _not_ solved, but I think they might be better solved more generally (i.e. I think we should solve them, I'm not sure we should solve them for cluster-addons as they might apply everywhere):

* How do we restrict usage of the ServiceAccount?  i.e. if I can create a pod in kube-system, then can I use any ServiceAccount I find there, and act as it.  I suspect I'm just missing something here.  In general, it's an argument for more namespaces, we might have problems with DNS specifically because I imagine lots of people expect it to be called `kube-dns` in `kube-system`.

* How do we prevent "non approved" images from running.  Here I think kubernetes has a strong answer in ImagePolicyWebhook, but I don't know to what extent this is actually used.

It would be intriguing to actually combine the two, I think:  I probably want much stricter controls on things running in kube-system than I do in some `myapp` application namespace.

I would propose that we proceed with the RBAC pre-creation, with this behind a feature-flag.  I think if we treat the ability to modify `CoreDNS` objects as equivalent to being able to create Pods in kube-system then I don't think we're changing the security stance (?).  CoreDNS objects are only mutable by cluster-admin, by default.


To try this out:

```
export KOPS_FEATURE_FLAGS="ClusterAddons,UseOperators"

kops create cluster addons.k8s.local --zones us-east-2a --channel file://channels/stable
```

`kops edit cluster addons.k8s.local` and add to spec:
```
  kubeDNS:
    provider: CoreDNS
```

```
kops update cluster addons.k8s.local --yes
kops export kubecfg addons.k8s.local --admin
kops validate cluster addons.k8s.local --wait=5m
```

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
justinsb: This was done upstream - moved to 443 with seemingly no consequence so I'll take it up there along with other changes!

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
justinsb: Agreed - I'll take it up upstream along with the idea of precreating RBAC

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
```diff
@@ -0,0 +1,418 @@
...
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - list
```
johngmyers, line 173: Sure, it's only list, but the `apiGroups` and `resources` are remarkably broad.

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
```diff
@@ -0,0 +1,418 @@
...
+  resources:
+  - coredns
+  verbs:
+  # TODO: do we need create & delete?
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
```
johngmyers, line 185: Is patch/update needed on `coredns` or is it sufficient to only have it on `coredns/status`? Does the operator really need to be able to write its own instructions?

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
justinsb: Concur... I'll track it down upstream

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
justinsb: Good question - that would be a nice restriction

justinsb:
> The need for the kube-dns Service in kube-system does make CoreDNS particularly troublesome. To site it in a different namespace would require the operator to manually sync the Endpoints from the other namespace (but it would only need rights to Service and Endpoints in kube-system).

I agree, and maybe we could overcome that (including as you described).  I definitely think it's worth trying to really lock down the permissions as much as possible here, both because it'll be a bigger project to move it out of kube-system (not that I'm ruling it out!) and because we're likely to copy these permissions to other operators.

> Running an admission controller to limit the image registries is pretty easy. Would the image references in these manifests go through remapping in the assets phase or is this an area where lack of templating would cause a problem?

So we can make sure that the operator manifests go through remapping (if they're not doing so already) but for the manifests the operators deploy, by default I'd imagine they are not going to be sourced from something kops-controlled.  We could make them kops-controlled (indeed, that might not be a bad thing from a reliability and security point of view).  I'm proposing here that we would mirror upstream manifests, not that we would ourselves take on the definitions of those manifests.  I'm not sure if we would do that mirroring in the kops repo (or some other similar place) or whether we would have the CLI mirror on the fly.  I'm leaning towards the mirror-on-the-fly approach.

The other option for image mirroring is that we use the image mirroring functionality that is in containerd.  Last I checked this would still go to the primary location on a mirror "cache miss", which was not really what I wanted.

justinsb:
cc @johngmyers I've updated this so that the operator has much fewer permissions.... using a combination of name filtering and pre-creationg of RBAC.  (Sadly RBAC name filtering doesn't work on create).

There's also #10381 which proposes a mode where we offer client-side expansion of these operators (i.e. there's no operator running on the server, but we're still able to pull manifests out of the kops binary)

justinsb:
So I'd like to suggest that we merge this and continue to improve it behind the feature-flag.  It's particularly difficult to make changes here because we're working across multiple repos.  As it stands, everything is behind the UseAddonOperators feature flag

Before removing it from the feature-flag, I/we will:

* Split out the RBAC roles for the operator so that the operator has vastly fewer permissions
* For other operators, try to run in dedicated namespaces (coredns is tricky because of the kube-dns service)
* Support an optional no-operator mode, using the client-side operator evaluation technique in #10381.
* Look into whether it's possible to produce a sandbox that doesn't require docker in the #10381 approach.  Likely Linux only, but in particular would address the "run in a container" problem.  (Maybe not a blocker for removing the feature-flag?)

What do you think @johngmyers ?   (I _think_ others are on board...)

justinsb:
In office hours on Friday, @johngmyers kindly agreed that we should move forward.

That said, I think i'm going to target this to the 1.22 branch.  I don't think that's as far away as it sounds, with 1.20 likely going out in the next few days (#11021 being the last known blocker)

If anyone disagrees though, that gives people a bit more time!

/milestone v1.22

channels/operators/coredns.addons.x-k8s.io/0.1.0-kops.1/manifest.yaml
justinsb: (I will take this up in the addon operator after this merges)

hakman (approved):
:)

PR #9375, opened 2020-06-16
Add support for AWS OIDC Provider

rifelpet:
This is a subset of https://github.com/kubernetes/kops/pull/9271 without any API field changes or actual usage of the awstasks. This will make it easier to rebase the changes in https://github.com/kubernetes/kops/pull/9352 based on these steps: https://github.com/kubernetes/kops/pull/9352#issuecomment-644193397


upup/pkg/fi/cloudup/awstasks/iamoidcprovider.go
```diff
@@ -0,0 +1,168 @@
...
+	"k8s.io/kops/upup/pkg/fi/cloudup/terraform"
+)
+
+//go:generate fitask -type=IAMOIDCProvider
+type IAMOIDCProvider struct {
+	Lifecycle *fi.Lifecycle
+
+	ARN         *string
+	ClientIDs   []*string
+	Thumbprints []*string
```
justinsb, line 38: FYI in a later PR I changed this to be `[]fi.Resource`, so we can defer evaluation

upup/pkg/fi/cloudup/awstasks/iamoidcprovider.go
```diff
@@ -0,0 +1,168 @@
...
+			return fi.RequiredField("URL")
+		}
+		if e.ClientIDs == nil {
+			return fi.RequiredField("ClientIDs")
+		}
+		if len(e.Thumbprints) == 0 {
+			return fi.RequiredField("Thumbprints")
+		}
+	} else {
+		if changes.ClientIDs == nil {
```
justinsb, line 97: This (and the next one) should be `!=`, I think

justinsb:
I think some of the `==` should be `!=`, but otherwise LGTM

upup/pkg/fi/cloudup/awstasks/iamoidcprovider.go
rifelpet: I left this as-is so we can keep your commit that changes it, since your commit also touches some other packages and I'd like to keep this PR small in scope

justinsb: :+1:

justinsb:
This LGTM and doesn't wire anything in yet (AFAICT) - thanks @rifelpet 

/approve
/lgtm

pkg/resources/aws/aws.go
```diff
@@ -2102,6 +2103,73 @@ func ListIAMInstanceProfiles(cloud fi.Cloud, clusterName string) ([]*resources.R
...
+			}
+			_, err := c.IAM().GetOpenIDConnectProvider(descReq)
+			if err != nil {
+				return nil, fmt.Errorf("error getting IAM OIDC Provider: %v", err)
+			}
+			// TODO: only delete oidc providers if they're owned by the cluster.
+			// We need to figure out how we can determine that given only a cluster name.
+			// Providers dont support tagging or naming.
+
+			// providers = append(providers, arn)
```
johngmyers, line 2129: Would have been better to comment out the deletion code, since this way the render code will try to create instead of update.

PR #9377, opened 2020-06-16
Don't start kubelet in protokube

justinsb:
Previously as an optimization we would start the kubelet from
protokube, after we had mounted the disks.  This helped avoid e.g. the
apiserver going into backoff waiting for etcd.

However, this no longer achieves anything with etcd-manager - nothing
happens on this front until after we start the kubelet anyway.

Doing this both takes protokube out of the dependency sequence here
(slightly faster boot time), but also removes the systemd dependency
from the protokube image.  (So we can get a smaller image, perhaps
even distroless)


hakman:
This would make things much simpler for ARM64 images. Nice!

images/protokube/Dockerfile
```diff
@@ -17,7 +17,7 @@ FROM k8s.gcr.io/debian-base-amd64:0.3
 # ca-certificates: Needed to talk to EC2 API
 # e2fsprogs: Needed to mount / format ext4 filesytems
 RUN apt-get update && apt-get install --yes \
-  bash ca-certificates e2fsprogs systemd \
+  bash ca-certificates e2fsprogs \
```
hakman, line 20: Anything similar that should be done one the Bazel side?
https://github.com/kubernetes/kops/blob/afd9aa1a4de0ba413bb303c70667cd2e8d519ee6/images/BUILD.bazel#L10-L13

hakman:
The mounter also uses systemd to mount volumes, but has a fallback. Is this acceptable?
https://github.com/kubernetes/kops/blob/53f16fd9efea78e933f053257ddc282dd9319f81/protokube/pkg/hostmount/nsenter_linux.go#L97
From one of my test boxes:
```bash
ubuntu@ip-10-4-211-228:~$ systemctl status mnt-master\\x2dvol\\x2d04957ee21a920b3db.mount 
● mnt-master\x2dvol\x2d04957ee21a920b3db.mount - /mnt/master-vol-04957ee21a920b3db
     Loaded: loaded (/proc/self/mountinfo)
     Active: active (mounted) since Mon 2020-06-15 13:53:10 UTC; 1 day 7h ago
      Where: /mnt/master-vol-04957ee21a920b3db
       What: /dev/nvme1n1
      Tasks: 0 (limit: 2310)
     Memory: 0B
     CGroup: /system.slice/mnt-master\x2dvol\x2d04957ee21a920b3db.mount
```

justinsb:
Great points both @hakman ... I was thinking that we can also remove the mounting, but I guess we should wait until we require etcd-manager - I know @johngmyers is planning on doing that!

At that point, I think we might even be able to go to distroless.  For now...

/hold

hakman:
@justinsb I am not really getting something. Protokube mounts the host root volume and runs all commands from there, including those related to systemd. Does it really need to have it in the image?

hakman:
> At that point, I think we might even be able to go to distroless. For now...

Please see how well distroless works with some tweaks in #9403 .
😄

hakman:
@justinsb The relevant changes from this PR were part of https://github.com/kubernetes/kops/pull/9403. I think it can be closed.

justinsb:
Agreed, thanks @hakman!

/close

PR #9378, opened 2020-06-17
Issue aws-iam-authenticator cert in nodeup

johngmyers:


nodeup/pkg/model/kube_apiserver.go
```diff
@@ -245,46 +245,31 @@ func (b *KubeAPIServerBuilder) writeAuthenticationConfig(c *fi.ModelBuilderConte
...
+				Signer:  fi.CertificateIDCA,
+				Type:    "server",
+				Subject: nodetasks.PKIXName{CommonName: id},
+				AlternateNames: []string{
+					"localhost",
+					"127.0.0.1",
+				},
 			}
+			c.AddTask(issueCert)
+			certificate, privateKey, _ := issueCert.GetResources()
```
justinsb, line 259: Oh that's cool - it's a future, but we don't have to treat it specially :-)

upup/pkg/fi/nodeup/nodetasks/issue_cert.go
```diff
@@ -120,11 +124,28 @@ func (i *IssueCert) AddFileTasks(c *fi.ModelBuilderContext, dir string, name str
 }
 
 func (e *IssueCert) Run(c *fi.Context) error {
+	// Skew the certificate lifetime by up to 30 days based on information about the generating node.
+	// This is so that different nodes created at the same time have the certificates they generated
+	// expire at different times, but all certificates on a given node expire around the same time.
```
justinsb, line 129: Oh!  Now I get why you used the network interfaces - so that the skew is constant for a given machine!

justinsb:
Thanks @johngmyers 

/approve
/lgtm

nodeup/pkg/model/kube_apiserver.go
johngmyers: At least not for `File`. The consumer does need to know to check it in its `GetDependencies()`.

Issue #3223, opened 2017-08-17
Kubelet Unable To Apply Reserved Cgroup Limits because Cgroup does not exist


ivelichkovich:
I have been trying to reserve compute resources following https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/

However, when I deploy an instance group with kops that instance group comes up as NotReady and has the following errors:

  4m		4m		1	kubelet, ip-172-20-180-67.ec2.internal			Normal		Starting				Starting kubelet.
  4m		4m		1	kubelet, ip-172-20-180-67.ec2.internal			Warning		FailedNodeAllocatableEnforcement	Failed to enforce System Reserved Cgroup Limits on "system.slice": "system.slice" cgroup does not exist
  4m		4m		1	kubelet, ip-172-20-180-67.ec2.internal			Warning		KubeletSetupFailed			Failed to start ContainerManager Failed to enforce System Reserved Cgroup Limits on "system.slice": "system.slice" cgroup does not exist
  4m		4m		1	kubelet, ip-172-20-180-67.ec2.internal			Normal		NodeAllocatableEnforced			Updated Node Allocatable limit across pods
  4m		1m		18	kubelet, ip-172-20-180-67.ec2.internal			Normal		NodeHasSufficientDisk			Node ip-172-20-180-67.ec2.internal status is now: NodeHasSufficientDisk
  4m		1m		18	kubelet, ip-172-20-180-67.ec2.internal			Normal		NodeHasSufficientMemory			Node ip-172-20-180-67.ec2.internal status is now: NodeHasSufficientMemory
  4m		1m		18	kubelet, ip-172-20-180-67.ec2.internal			Normal		NodeHasNoDiskPressure			Node ip-172-20-180-67.ec2.internal status is now: NodeHasNoDiskPressure

I can't seem to figure out how to point to the right cgroups.

Here is my instance group spec:

apiVersion: kops/v1alpha2
kind: InstanceGroup
metadata:
  labels:
    kops.k8s.io/cluster: xxxxxxxxx
  name: nodes
spec:
  kubelet:
    kubeReserved:
        cpu: "200m"
        memory: "200Mi"
        storage: "2Gi"
    kubeReservedCgroup: "/kube.slice"
    systemReserved:
        cpu: "200m"
        memory: "200Mi"
        storage: "2Gi"
    systemReservedCgroup: "/system.slice"
    enforceNodeAllocatable: "pods,system-reserved,kube-reserved"
[...]

PR #9380, opened 2020-06-17
Updating cluster_spec doc.

michalschott:
Updating docs with two extra flags which allowed me finally to enforce resource allocation for kubelet on flatcar linux, altho might be valid for other distros too.

Potentially fixes/closes https://github.com/kubernetes/kops/issues/3223 ?

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -659,6 +659,8 @@ spec:
         memory: "100Mi"
         ephemeral-storage: "1Gi"
     kubeReservedCgroup: "/kube-reserved"
+    kubeletCgroups: "/kube-reserved"
+    runtimeCgroups: "/kube-reserved"
     systemReserved:
         cpu: "100m"
         memory: "100Mi"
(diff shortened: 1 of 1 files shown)


docs/cluster_spec.md
```diff
@@ -667,7 +669,7 @@ spec:
     enforceNodeAllocatable: "pods,system-reserved,kube-reserved"
 ```
 
-Will result in the flag `--kube-reserved=cpu=100m,memory=100Mi,ephemeral-storage=1Gi --kube-reserved-cgroup=/kube-reserved --system-reserved=cpu=100m,memory=100Mi,ephemeral-storage=1Gi --system-reserved-cgroup=/system-reserved --enforce-node-allocatable=pods,system-reserved,kube-reserved`
+Will result in the flag `--kube-reserved=cpu=100m,memory=100Mi,ephemeral-storage=1Gi --kube-reserved-cgroup=/kube-reserved --kubelet-cgroups=/kube-reserved --runtime-cgroups=/kube-reserved --system-reserved=cpu=100m,memory=100Mi,ephemeral-storage=1Gi --system-reserved-cgroup=/system-reserved --enforce-node-allocatable=pods,system-reserved,kube-reserved`
```
gjtempleton, line 672: As we're adding this should we also add a new link out to detailed k8s level docs on the cgroups flags, as the existing link to reserving compute resources doesn't cover this?

michalschott: Pushed an update.

PR #9387, opened 2020-06-17
Clean up the HA docs

### LINKED ISSUE #8769: Not clear when/how to configure more than 3 masters
 this information.**

1.13.0

 a `kops` flag.**

1.13.10

AWS

1. Create a cluster with 3 master nodes
2. Decide to increase the number of master nodes
3. Find the page https://github.com/kubernetes/kops/blob/master/docs/operations/high_availability.md
4. Determine you're not sure: if you use 5 master nodes, should you use 5 zones, or is 3 zones ok? The docs seem to say that more than one master in a zone is actually less HA, but not sure why. What if you want 11 master nodes in a region that has only 5 zones, you'll have 4 zones with 2 nodes each, and one zone with 3 nodes. 

Basically what's not clear is, does it even make sense to have # masters > # zones? If it does make sense, are there rules about the number of zones vs masters. Eg if you have 6 masters, in 3 zones, an one zone fails, you're down to 4 masters; why is that worse than having a 5 master cluster in 3 zones, and the one-master zone fails, then you're also down to 4. This needs to be explained better. 

It seems very likely that as time goes, your requirements are going to increase in services and hence number of pods and hence demand on master nodes will increase, and I imagine at some point you will notice the CPU on masters become "too high too often". At that point, you will need to know the above. 

n/a

Have a clear idea of whether increasing master node count from 3 to 5 required that I also increase the # of zones, or if I could stay at 3 zones, and overall I was expecting to have a better understanding of when and how to scale the control plane of a kops-based kubernetes cluster. 

Probably n/a but let me know if applicable

Probably not applicable 

Not for now

olemarkus:
Fixes #8769

### DIFF
--- a/docs/operations/high_availability.md
+++ b/docs/operations/high_availability.md
@@ -1,96 +1,59 @@
-High Availability (HA)
-======================
+# High Availability (HA)
 
-Introduction
--------------
+## Introduction
 
-Kubernetes has two strategies for high availability:
+For testing purposes, kubernetes works just fine with a single master. However, when the master becomes unavailable, for example due to upgrade or instance failure, the kubernetes API will be unavailable. Pods and services that are running on the continues to operate as long as they do not depend on interacting with the API, but operations such as adding nodes, scaling pods, replacing terminated pods will not work. Running kubectl will also not work. 
 
-* Run multiple independent clusters and combine them behind one management plane: [federation](https://kubernetes.io/docs/user-guide/federation/)
-* Run a single cluster in multiple cloud zones, with redundant components
+kops runs each master in a dedicated autoscaling groups (ASG) and stores data on EBS volumes. That way, if a master node is terminated the ASG will launch a new master instance with the master's volume. Because of the dedicated EBS volumes, each master is bound to a fixed Availability Zone (AZ). If the AZ becomes unavailable, the master instance in that AZ will also become unavailable.
 
-kops has good support for a cluster than runs
-with redundant components.  kops is able to create multiple kubernetes masters, so in the event of
-a master instance failure, the kubernetes API will continue to operate.
+For production use, you therefore want to run kubernetes in a HA setup with multiple masters. With multiple master nodes, you will be able both to do graceful (zero-downtime) upgrades and you will be able to survive AZ failures.
 
-However, when running kubernetes with a single master, if the master fails, the kubernetes API will be unavailable, but pods and services that are running on the (unaffected) nodes should continue to operate.  In this situation, we won't be able to do anything that involves the API (adding nodes, scaling pods, replacing
-terminated pods), and kubectl won't work.  However your application should continue to run, and most applications
-could probably tolerate an API outage of an hour or more.
+Very few regions offer less than 3 AZs. In this case, running multiple masters in the same AZ is an option. If the AZ with multiple masters becomes unavailable you will still have downtime with this configuration. But regular changes to master nodes such as upgrades will be graceful and without downtime.
 
-Moreover, kops runs the masters in an automatic replacement mode.  Masters are run in auto-scaling groups, with
-the data on an EBS volume.  If a master node is terminated, the ASG will launch a new master instance, and kops
-will mount the master volume and replace the master.
+If you already have a single-master cluster you would like to convert to a multi-master cluster, read the [single to multi-master](../single-to-multi-master.md) docs.
 
-In short:
+Note that running clusters spanning several AZs is more expensive than running clusters spanning one or two AZs. This happens not only because of the master EC2 cost, but also because you have to pay for cross-AZ traffic. Depending on your workload you may therefore also want to consider running worker nodes only in two AZs. As long as your application do not rely on quorum, you will still have AZ fault tolerance.
 
-* A single master kops cluster is still reasonably available; if the master instance terminates it will be automatically
-  replaced.  But the use of EBS binds us to a single AZ, and in the event of a prolonged AZ outage, we might experience
-  downtime.
-* A multi-node kops cluster can tolerate the outage of a single AZ
+## Creating a HA cluster
 
+### Example 1: public topology
 
-Using Kops HA
--------------
-
-We can create HA clusters using kops, but only it's important to note that migrating from a single-master
-cluster to a multi-master cluster is a complicated operation (described [here](../single-to-multi-master.md)).
-If possible, try to plan this at time of cluster creation.
-
-When you first call `kops create cluster`, you specify the `--master-zones` flag listing the zones you want your masters
-to run in, for example:
+The simplest way to get started with a HA cluster is to run `kops create cluster` as shown below. The `--master-zones` flag lists the zones you want your masters
+to run in. By default, kops will create one master per AZ. Since the kubernetes etcd cluster runs on the master nodes, you have to specify an odd number of zones in order to obtain quorum.
 
 ```
 kops create cluster \
     --node-count 3 \
     --zones us-west-2a,us-west-2b,us-west-2c \
     --master-zones us-west-2a,us-west-2b,us-west-2c \
-    --node-size t2.medium \
-    --master-size t2.medium \
-    --topology private \
-    --networking kopeio-vxlan \
     hacluster.example.com
 ```
 
-Kubernetes relies on a key-value store called "etcd", which uses the Quorum approach to consistency,
-so it is available if 51% of the nodes are available.
+## Example 2: private topology
 
-As a result there are a few considerations that need to be taken into account when using kops with HA:
-
-* Only odd number of masters instances should be created, as an even number is likely _less_ reliable than the lower odd number.
-* Kops has experimental support for running multiple masters in the same AZ, but it should be used carefully.
-  If we create 2 (or more) masters in the same AZ, then failure of the AZ will likely cause etcd to lose quorum
-  and stop operating (with 3 nodes).  Running in the same AZ therefore increases the risk of cluster disruption,
-  though it can be a valid scenario, particularly if combined with [federation](https://kubernetes.io/docs/user-guide/federation/).
-
-
-Advanced Example
-----------------
-
-Another example `create cluster` invocation for HA with [a private network topology](../topology.md):
+Create a cluster using [private network topology](../topology.md):
 
 ```
 kops create cluster \
     --node-count 3 \
     --zones us-west-2a,us-west-2b,us-west-2c \
     --master-zones us-west-2a,us-west-2b,us-west-2c \
-    --dns-zone example.com \
-    --node-size t2.medium \
-    --master-size t2.medium \
-    --node-security-groups sg-12345678 \
-    --master-security-groups sg-12345678,i-abcd1234 \
     --topology private \
-    --networking weave \
-    --cloud-labels "Team=Dev,Owner=John Doe" \
-    --image 293135079892/k8s-1.4-debian-jessie-amd64-hvm-ebs-2016-11-16 \
+    --networking <provider> \
     ${NAME}
 ```
 
-Notes (Best Practice)
-----
-* In regions with 2 Availability Zones, deploy the 3 masters in one zone and the nodes can be distributed between the 2
-zones. This can be done by specifying the flags:
+Note that the default networking provider (kubenet) does not support private topology.
+
+## Example 3: multiple masters in the same AZ
+
+If necessary, for example in regions with less than 3 AZs, you can launch multiple masters in the same AZ.
+
 ```
-     --master-count=3
-     --master-zones=$MASTER_ZONE
-     --zones=$NODE_ZONES
+kops create cluster \
+    --node-count 3 \
+    --master-count 3 \
+    --zones cn-north-1a,cn-north-1b \
+    --master-zones cn-north-1a,cn-north-1b \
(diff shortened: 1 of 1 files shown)


docs/operations/high_availability.md
```diff
@@ -1,96 +1,41 @@
...
-    --dns-zone example.com \
-    --node-size t2.medium \
-    --master-size t2.medium \
-    --node-security-groups sg-12345678 \
-    --master-security-groups sg-12345678,i-abcd1234 \
     --topology private \
-    --networking weave \
-    --cloud-labels "Team=Dev,Owner=John Doe" \
-    --image 293135079892/k8s-1.4-debian-jessie-amd64-hvm-ebs-2016-11-16 \
+    --networking cilium \
```
hakman, line 39: I wouldn't mention Cilium here. If you want, maybe Calico or Weave which seem simpler to digest for beginners (less options).

olemarkus: Weave is listed as experimental on our networking page, which is why I am changing weave for something else when I can. But I am happy with using calico.

hakman: Should be better as I think is the default in other places, like Docker Desktop and Enterprise.

hakman: Not sure how experimental Weave is. I don't use it, but kept it up to date and seems pretty stable in tests.

olemarkus: Weave is experimental per https://kops.sigs.k8s.io/networking/ :) I think I actually had weave a stable initially, but someone said it probably wasn't and I decided to be conservative.

docs/operations/high_availability.md
```diff
@@ -1,96 +1,41 @@
-High Availability (HA)
-======================
+# High Availability (HA)
 
-Introduction
--------------
+## Introduction
 
-Kubernetes has two strategies for high availability:
+For testing purposes, kubernetes works just fine with a single master. However, when the master becomes unavailable, for example due to upgrade or instance failure, the kubernetes API will be unavailable. Pods and services that are running on the continues to operate as long as they do not depend on interacting with the API, but operations such as adding nodes, scaling pods, replacing terminated pods will not work. Running kubectl will also not work.
```
justinsb, line 5: Missing word (maybe "nodes"?): "running on the continues to operate"

docs/operations/high_availability.md
```diff
@@ -1,96 +1,41 @@
...
-Introduction
--------------
+## Introduction
 
-Kubernetes has two strategies for high availability:
+For testing purposes, kubernetes works just fine with a single master. However, when the master becomes unavailable, for example due to upgrade or instance failure, the kubernetes API will be unavailable. Pods and services that are running on the continues to operate as long as they do not depend on interacting with the API, but operations such as adding nodes, scaling pods, replacing terminated pods will not work. Running kubectl will also not work. 
 
-* Run multiple independent clusters and combine them behind one management plane: [federation](https://kubernetes.io/docs/user-guide/federation/)
-* Run a single cluster in multiple cloud zones, with redundant components
+kops runs each master in a dedicated autoscaling groups (ASG) and stores data on ESB volumes. That way, if a master node is terminated the ASG will launch a new master instance with the master's volume. Because of the dedicated ESB volumes, each master is bound to a fixed Availability Zone (AZ). If the AZ becomes unavailable, the master instance in that AZ will also become unavailable.
```
justinsb, line 7: Nit: s/groups/group  (or remove "a" from "a dedicated autoscaling group")

Nit: s/ESB/EBS

We should probably try to call them "control plane nodes" instead of "master" also.

Nits aside, love this paragraph - it explains a tricky subject well!

docs/operations/high_availability.md
```diff
@@ -1,96 +1,41 @@
...
 
-kops has good support for a cluster than runs
-with redundant components.  kops is able to create multiple kubernetes masters, so in the event of
-a master instance failure, the kubernetes API will continue to operate.
+For production use, you therefor want to run kubernetes in a HA setup with multiple masters. With multiple master nodes, you will be able both to do graceful, zero-down time upgrades, and you will be able to survive AZ failures.
```
justinsb, line 9: Nit: therefor -> therefore

docs/operations/high_availability.md
```diff
@@ -1,96 +1,41 @@
...
-Using Kops HA
--------------
-
-We can create HA clusters using kops, but only it's important to note that migrating from a single-master
-cluster to a multi-master cluster is a complicated operation (described [here](../single-to-multi-master.md)).
-If possible, try to plan this at time of cluster creation.
-
-When you first call `kops create cluster`, you specify the `--master-zones` flag listing the zones you want your masters
-to run in, for example:
+The simplest way to get started with a HA cluster is to run `kops create cluster` as shown below. The `--master-zones` flag listing the zones you want your masters
```
justinsb, line 18: s/listing/lists

docs/operations/high_availability.md
```diff
@@ -1,96 +1,41 @@
...
     hacluster.example.com
 ```
 
-Kubernetes relies on a key-value store called "etcd", which uses the Quorum approach to consistency,
-so it is available if 51% of the nodes are available.
-
-As a result there are a few considerations that need to be taken into account when using kops with HA:
-
-* Only odd number of masters instances should be created, as an even number is likely _less_ reliable than the lower odd number.
-* Kops has experimental support for running multiple masters in the same AZ, but it should be used carefully.
```
justinsb, line 60: I feel like this is an important sentence to keep  - the idea that you can repeat the zone.  If you only have 2 AZs in a region, whether it's better to run in 2 AZs or all in one is debatable, I guess - we seem to have suggested at the end that running in one region is best.

(Agree with dropping the federation stuff though!)

justinsb:
A few nits, and we should probably prefer the term "control plane" over "master", but I guess that is confusing because of the flags also.

docs/operations/high_availability.md
johngmyers: After getting a ten-hour 3AM call about Weave, I'm reluctant to call it stable. I'm given to understand it has at its core an algorithm with polynomial complexity.

hakman: OK 😄 
Do we all agree that Calico is stable and the easiest one to get started with?

olemarkus: I don't :) 
After some sleep, I am wondering if our examples should just put "<provider>" placeholder instead of a specific one. Or we need to have an "official" opinion on what users should go with. Most of the docs tries to be neutral here.
If we do go for e.g calico as the recommended provider, we should probably also use that one as default instead of kubenet, since you typically don't want to use kubenet.
This is probably worth an issue on its own though.

docs/operations/high_availability.md
olemarkus: Thanks for the nits.
I am all for changing "master" to "control plane nodes" (something a bit less verbose would be nice though) if that is what k/k is doing as well. But we should have a plan on changing this everywhere. It would probably be a good idea to do this as part of #9178

docs/operations/high_availability.md
hakman: Neutral is good enough for me :)

docs/operations/high_availability.md
olemarkus: I thought there was no longer any region with 2 zones anymore, but turns out it is one: Beijing. I'll add a version of that back in.

hakman:
I would also add somewhere to be mindful of how many AZs are used for HA. Transferring data between AZs can be expensive. This is why, maybe limit to 2 AZs for worker nodes.

olemarkus:
Do we really want to recommend running workers in two AZs? That would at least be a "you better know what you are doing" disclaimer with that as if you run apps that requires quorum, you'll have downtime should the wrong AZ fail.

I am considering a warning about running e.g 5 AZs though. It gives you higher fault tolerance, but in most cases, a bit too much.

hakman:
Apps with quorum is a totally different story. To get to those you actually have to get past the beginner status. Also have to understand how pod scheduling works because you may only think it works and instead all your quorum pods are in same AZ.

I don't say we recommend 2 AZs. I just mean to phrase it in a way that explains that inter-AZ traffic costs in most cases. You need 2+ AZ for HA, but depends on use case how many.

olemarkus:
Something like this?

docs/operations/high_availability.md
```diff
@@ -1,96 +1,59 @@
...
 
-kops has good support for a cluster than runs
-with redundant components.  kops is able to create multiple kubernetes masters, so in the event of
-a master instance failure, the kubernetes API will continue to operate.
+For production use, you therefore want to run kubernetes in a HA setup with multiple masters. With multiple master nodes, you will be able both to do graceful, zero-down time upgrades, and you will be able to survive AZ failures.
 
-However, when running kubernetes with a single master, if the master fails, the kubernetes API will be unavailable, but pods and services that are running on the (unaffected) nodes should continue to operate.  In this situation, we won't be able to do anything that involves the API (adding nodes, scaling pods, replacing
-terminated pods), and kubectl won't work.  However your application should continue to run, and most applications
-could probably tolerate an API outage of an hour or more.
+A few regions offers less than 3 AZs. In this case, running multiple masters in the same AZ is an option. If the AZ with multiple masters in becomes unavailable, you will still have downtime, but upgrades will be graceful and without downtime.
```
hakman, line 11:
```suggestion
Very few regions offer less than 3 AZs. In this case, running multiple masters in the same AZ is an option. If the AZ with multiple masters becomes unavailable, you will still have downtime, but upgrades will be graceful and without downtime.
```

hakman: Also, this is a bit confusing `you will still have downtime, but upgrades will be graceful and without downtime`.

docs/operations/high_availability.md
```diff
@@ -1,96 +1,59 @@
...
 
-Moreover, kops runs the masters in an automatic replacement mode.  Masters are run in auto-scaling groups, with
-the data on an EBS volume.  If a master node is terminated, the ASG will launch a new master instance, and kops
-will mount the master volume and replace the master.
+If you already have a single-master cluster you would like to convert to a multi-master cluster, read the [single to multi-master](../single-to-multi-master.md) docs.
 
-In short:
+Note that running clusters spanning several AZs are more expensive than clusters spanning one or two AZs. Not only because of the master EC2 cost, but also because you have to pay for cross-AZ traffic. Depending on your workload you may therefore also want to consider running worker nodes only in two AZs.
```
hakman, line 15:
```suggestion
Note that running clusters spanning several AZs is more expensive than running clusters spanning one or two AZs. This happens not only because of the master EC2 cost, but also because you have to pay for cross-AZ traffic. Depending on your workload you may therefore also want to consider running worker nodes only in two AZs.
```

docs/operations/high_availability.md
```diff
@@ -1,96 +1,59 @@
...
 
-* Run multiple independent clusters and combine them behind one management plane: [federation](https://kubernetes.io/docs/user-guide/federation/)
-* Run a single cluster in multiple cloud zones, with redundant components
+kops runs each master in a dedicated autoscaling groups (ASG) and stores data on EBS volumes. That way, if a master node is terminated the ASG will launch a new master instance with the master's volume. Because of the dedicated EBS volumes, each master is bound to a fixed Availability Zone (AZ). If the AZ becomes unavailable, the master instance in that AZ will also become unavailable.
 
-kops has good support for a cluster than runs
-with redundant components.  kops is able to create multiple kubernetes masters, so in the event of
-a master instance failure, the kubernetes API will continue to operate.
+For production use, you therefore want to run kubernetes in a HA setup with multiple masters. With multiple master nodes, you will be able both to do graceful, zero-down time upgrades, and you will be able to survive AZ failures.
```
hakman, line 9:
```suggestion
For production use, you therefore want to run kubernetes in a HA setup with multiple masters. With multiple master nodes, you will be able both to do graceful (zero-downtime) upgrades and you will be able to survive AZ failures.
```

hakman:
> Something like this?

Yes, sounds pretty good.
There are a few more small nits, but other than that it's a nice change.

docs/operations/high_availability.md
rifelpet: 👍  Some regions have more than two AZs but only two of them are effectively usable due to instance type availability: us-west-1 for example.

docs/operations/high_availability.md
```diff
@@ -1,96 +1,59 @@
-High Availability (HA)
-======================
+# High Availability (HA)
 
-Introduction
--------------
+## Introduction
 
-Kubernetes has two strategies for high availability:
+For testing purposes, kubernetes works just fine with a single master. However, when the master becomes unavailable, for example due to upgrade or instance failure, the kubernetes API will be unavailable. Pods and services that are running in the cluster continues to operate as long as they do not depend on interacting with the API, but operations such as adding nodes, scaling pods, replacing terminated pods will not work. Running kubectl will also not work.
```
hakman, line 5:
```suggestion
For testing purposes, kubernetes works just fine with a single master. However, when the master becomes unavailable, for example due to upgrade or instance failure, the kubernetes API will be unavailable. Pods and services that are running in the cluster continue to operate as long as they do not depend on interacting with the API, but operations such as adding nodes, scaling pods, replacing terminated pods will not work. Running kubectl will also not work. 
```

PR #9389, opened 2020-06-18
Run "go mod vendor" in verify-gomod

rifelpet:
This will actually catch vendoring issues.

also removing it from `make travis-ci` because `make gomod` depends on bazel which isnt available in travis. `make gomod` has its own prow job though, so we still get sufficient coverage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -564,7 +564,7 @@ ci: govet verify-gofmt verify-generate verify-gomod verify-goimports verify-boil
 # verify-gofmt: uses bazel, covered by pull-kops-verify
 # govet needs to be after verify-goimports because it generates bindata.go
 .PHONY: travis-ci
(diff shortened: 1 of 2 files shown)


justinsb:
Whoops!  Thanks for catching this mistake I made :-)

I know it's WIP, but looks good so far!

rifelpet:
It looks like go is still downloading the modules into the mod cache rather than using the vendor directory. I'm sure theres an env var that we can set in the Makefile to tell go to use the vendor directory every time, and that this default behavior changed between 1.13 and 1.14. k/k is still on 1.13 but [is hoping to upgrade to 1.14 for k8s 1.19](https://github.com/kubernetes/kubernetes/pull/88638#issuecomment-639890325), so maybe we just follow suit once that lands and our jobs should start using the vendor directory automatically.

regardless, that can be in a followup PR

justinsb:
Thanks @rifelpet .  I believe the flag is `mod=vendor`, and it can be passed to each `go` invocation or simply set with `export GOFLAGS=-mod=vendor`.

PR #9391, opened 2020-06-18
Start moving InstanceGroup data to NodeupConfig

johngmyers:
This is the first few commits of #9294 submitted separately in attempt to make it more reviewable. It stops short of the first commit that has userdata size limit concerns.

This lays down some groudwork I'll need for other refactoring efforts.


justinsb:
Thanks - great idea to break it into smaller steps.

I also had an idea - we could have a "compatability" mode where we have a field `FileAssets: true` or `Hooks: true`.  In that case we would continue to load the instance group from the yaml file (it would have non-ideal semantics, but it wouldn't be a regression).  We could even do something like allow for (say) 8KB of data inline, but if we go over then we log a warning and only in that case would we set the flag e.g. `LoadExternalAssets: true`

/approve
/lgtm

PR #9396, opened 2020-06-18
Use -mod=vendor for most go commands

rifelpet:
Without this set, go 1.13 was redownloading all dependencies into the module cache, effectively ignoring the vendor directory.

This instructs go 1.13 to always use (and verify the contents of) the vendor directory.

See an example prow job [here](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kops/9389/pull-kops-verify-govet/1273460433402990595). All of the `go: downloading` and `go: extracting` lines indicate its downloading those dependencies rather than using the vendor directory.

Unsetting it for when we actually update the vendor directory with `make gomod`.

I'm not sure if/how this affects bazel

Ref: https://github.com/kubernetes/kops/pull/9389#issuecomment-645757357

/assign @justinsb

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -463,9 +463,8 @@ gomod-prereqs:
 	(which bazel > /dev/null) || (echo "gomod requires that bazel is installed"; exit 1)
 
 .PHONY: gomod
-gomod: export GOFLAGS=
 gomod: gomod-prereqs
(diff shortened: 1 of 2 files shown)


rifelpet:
/unassign @justinsb

while i work out some issues

PR #9400, opened 2020-06-18
Stop creating SHA1 hashes for build artifacts

hakman:
Kops moved to SHA256 and SHA1 are no longer used.


hakman:
Thanks @rdrgmnzs 😄

PR #9402, opened 2020-06-19
Prepatory refactoring of BootstrapScript

johngmyers:
In order to issue the `kubernetes-master` server cert out of nodeup, I'm going to need to pass the loadbalancer IP in userdata. That means the code that is currently in `BootstrapScript.ResourceNodeUp()` is going to have to run later, after the load balancer IPs have been determined.

This refactoring is in preparation for that. It pulls code out of `BootstrapScript.NodeupConfigBuilder` that probably should not be run later.


olemarkus:
Could perhaps be an opportunity to split some methods, but overall looks good to me.

/lgtm

PR #9403, opened 2020-06-19
Use distroless image as base for Protokube

hakman:
This builds on #9377 and moves Protokube to the distroless base image.


hakman:
@justinsb as you can see, distroless Protokube works without issues.
It uses the systemd binaries from the hosts and mounts are still handled with systemd.

hakman:
As you can see, distroless Protokube works without issues.
It uses the systemd binaries from the hosts and mounts are still handled with systemd.
/cc @justinsb

hakman:
@justinsb any chance you could review this?

hakman:
This doesn't work for etcd legacy provider. It is fixable, if we want to keep it though. We would just have to run the commands in chroot.

johngmyers:
Kops 1.19 currently supports k8s 1.11, which defaults to legacy provider.

hakman:
> Kops 1.19 currently supports k8s 1.11, which defaults to legacy provider.

This is how I found out it doesn't work 😁

hakman:
Simplest solution for legacy provider is to mount `/bin` and `/sbin` from the host.
This provides access to all the binaries required by `utils/mount` and `utils/nsenter`.

hakman:
Current version of the PR works with both `Manager` and `Legacy` etcd providers.

johngmyers:
Perhaps add TODO comments on the bits that can be removed with legacy etcd provider.

hakman:
> Perhaps add TODO comments on the bits that can be removed with legacy etcd provider.

I think it is a bit premature to add a TODO for that. The `mount` and `nsenter` libs could be used for any number of other reasons by the time the Legacy etcd provider is removed. The comments for the mounts should be sufficient to remind about that.

hakman:
I cleaned up the initial commit to only move the kubelet startup to nodeup.
That being said, this is good to go from my point of view.
/hold cancel

olemarkus:
Looks good to me as well.
/lgtm

justinsb:
Thanks for doing this, and for fixing the legacy issue.

I'm OK with trying this for 1.19.  I propose we not backport to 1.18 though, as we want plenty of time for this to "bake".

/approve
/lgtm

hakman:
Thanks for reviewing this @justinsb :).
I agree that 1.18 should stay as is. I think we are all very eager to see it released.

PR #9407, opened 2020-06-19
Add master and node image options when creating a cluster

hakman:
As discussed during office hours, we decided to add these new options to help test ARM64 deployments.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -84,6 +84,8 @@ type CreateClusterOptions struct {
 	ContainerRuntime     string
 	OutDir               string
 	Image                string
+	NodeImage            string
+	MasterImage          string
 	VPCID                string
 	SubnetIDs            []string
 	UtilitySubnetIDs     []string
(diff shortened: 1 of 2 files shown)


olemarkus:
I wonder if there should be some documentation or some indication of what happens if both `--image` and the new flags are set. What happens now is intuitive, but it doesn't hurt to be explicit.

/assign @rifelpet
/assign @justinsb too as there are new flags involved.

cmd/kops/create_cluster.go
```diff
@@ -299,8 +301,14 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVar(&sshPublicKey, "ssh-public-key", sshPublicKey, "SSH public key to use (defaults to ~/.ssh/id_rsa.pub on AWS)")
 
-	cmd.Flags().StringVar(&options.NodeSize, "node-size", options.NodeSize, "Set instance size for nodes")
+	cmd.Flags().Int32Var(&options.MasterCount, "master-count", options.MasterCount, "Set number of masters. Defaults to one master per master-zone")
+	cmd.Flags().Int32Var(&options.NodeCount, "node-count", options.NodeCount, "Set number of nodes")
+
+	cmd.Flags().StringVar(&options.Image, "image", options.Image, "Set image for all instances.")
+	cmd.Flags().StringVar(&options.NodeImage, "node-image", options.NodeImage, "Set image for nodes")
+	cmd.Flags().StringVar(&options.MasterImage, "master-image", options.MasterImage, "Set image for masters")
```
rifelpet, line 309: Maybe something like this?

```suggestion
	cmd.Flags().StringVar(&options.NodeImage, "node-image", options.NodeImage, "Set image for nodes. Takes precedence over --image")
	cmd.Flags().StringVar(&options.MasterImage, "master-image", options.MasterImage, "Set image for masters. Takes precedence over --image")
```

PR #9409, opened 2020-06-19
Fold multiple integration test cases into the complex test case

rifelpet:
Each integration test case adds many LoC and some overhead in running the integration tests.

Since many of the tests are only testing a specific feature, it would be simpler to combine all of the non-mutually exclusive features into the single complex case. We could probably do this to more of the test cases but this was my first high level pass.

~WIP while I figure out this test failure. hack/verify-expected.sh reports that it found a difference and wrote it to the files but there aren't any actual changes to the file afterwards...~


rifelpet:
oh, i remember this happening before. its because we run 3 test cases for `complex` and one of them changes the file to one thing and a later test (looks like the legacy apiVersion test) changes it back. end result is the same incorrect file. ill see if i can split that up somehow.

PR #9412, opened 2020-06-20
Rolling update instance groups in consistent order

johngmyers:


mikesplain:
Good idea, thanks @johngmyers. 

/lgtm
/approve

PR #9413, opened 2020-06-20
Start pushing create_cluster logic into pkg

johngmyers:
This is initial work in pushing some of the logic that is in create_cluster into pkg for use as an import.

My intent for `NewCluster()` is for it to initialize the cluster and instance group specs, but not query or update the cluster state in the state store.

We try to approximate GitOps and want to avoid configuration drift between "defaults for new clusters" and "defaults or existing clusters". Our existing update cluster code always goes through a `kops create cluster --dry-run` for that reason and we would like to instead call this `NewCluster()` code.


rifelpet:
I believe the staticcheck failure is because our prow job image got upgraded to 1.14: https://github.com/kubernetes/test-infra/pull/18047 so we'll need to fix that before any PRs can be merged to master

PR #9416, opened 2020-06-21
Update terraform.md

ozbillwang:
the codes generated by `kops --target terraform` are not for latest terraform v0.12+

So we have to upgrade it first.

### DIFF
--- a/docs/terraform.md
+++ b/docs/terraform.md
@@ -79,6 +79,14 @@ $ kops update cluster \
   --target=terraform
 ```
 
+Upgrade to terraform v0.12+
+```
+$ terraform 0.12upgrade
+
+# Need init again
+$ terraform init
+```
+
 Then apply your changes after previewing what changes will be applied:
 
 ```


ozbillwang:
I have signed

hakman:
Thanks @ozbillwang !
/ok-to-test
/check-cla
/assign @rifelpet

rifelpet:
Thanks for this!

Terraform support with kops is as follows:
* Terraform 0.12 HCL generated automatically with Kops 1.18 (to be released soon)
* `terraform 0.12upgrade` is needed with Kops 1.17
* `terraform 0.12upgrade` may not work with Kops <1.17 due to resource naming restrictions in TF 0.12 which were addressed in Kops 1.17.

So I'm wondering if we could add some clarification here. Maybe we add a note at the top of the document:

> Terraform 0.12 support requires Kops 1.17 and above.

and then your current addition could be:

---

If using Kops 1.17, upgrade to terraform v0.12+ manually:
```
$ terraform 0.12upgrade

# Need init again
$ terraform init
```

---

Thoughts?

ozbillwang:
If `0.12upgrade` is only for kubernetes version 1.17, maybe we needn't do this PR. Things move on quickly.

PR #9418, opened 2020-06-21
Make dns pods work on arm64 clusters

olemarkus:
* Use multiarch where possible. Remove hardcoded use of amd64-only images.
 * Remove Arch templating function as an approach using this will not
   work anyway
 * No multiarch version of dns autoscaler, so force this one to run on
   master for now.

### DIFF
--- a/upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io/k8s-1.12.yaml.template
@@ -53,7 +53,7 @@ spec:
     spec:
       containers:
       - name: autoscaler
-        image: k8s.gcr.io/cluster-proportional-autoscaler-{{Arch}}:1.4.0
+        image: k8s.gcr.io/cluster-proportional-autoscaler-amd64:1.4.0
         resources:
             requests:
                 cpu: "20m"
@@ -69,10 +69,14 @@ spec:
           - --default-params={"linear":{"coresPerReplica":256,"nodesPerReplica":16,"preventSinglePointFailure":true}}
           - --logtostderr=true
           - --v=2
+      nodeSelector:
+        kubernetes.io/arch: amd64
       priorityClassName: system-cluster-critical
       tolerations:
       - key: "CriticalAddonsOnly"
         operator: "Exists"
+      - key: node-role.kubernetes.io/master
+        operator: Exists
       serviceAccountName: kube-dns-autoscaler
 
 ---
@@ -130,7 +134,7 @@ spec:
 
       containers:
       - name: kubedns
-        image: k8s.gcr.io/k8s-dns-kube-dns-{{Arch}}:1.14.13
+        image: k8s.gcr.io/k8s-dns-kube-dns:1.14.13
         resources:
           # TODO: Set memory limits when we've profiled the container for large
           # clusters, then set request = limit to keep this container in
@@ -182,7 +186,7 @@ spec:
           mountPath: /kube-dns-config
 
       - name: dnsmasq
-        image: k8s.gcr.io/k8s-dns-dnsmasq-nanny-{{Arch}}:1.14.13
+        image: k8s.gcr.io/k8s-dns-dnsmasq-nanny:1.15.13
         livenessProbe:
           httpGet:
             path: /healthcheck/dnsmasq
@@ -224,7 +228,7 @@ spec:
           mountPath: /etc/k8s/dns/dnsmasq-nanny
 
       - name: sidecar
-        image: k8s.gcr.io/k8s-dns-sidecar-amd64:1.14.13
+        image: k8s.gcr.io/k8s-dns-sidecar:1.14.13
         livenessProbe:
           httpGet:
             path: /metrics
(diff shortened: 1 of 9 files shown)


hakman:
This should get the ARM64 tests rolling.
/lgtm
/assign @rifelpet

upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -53,11 +53,13 @@ spec:
...
       containers:
       - name: autoscaler
-        image: k8s.gcr.io/cluster-proportional-autoscaler-{{Arch}}:1.4.0
+        image: k8s.gcr.io/cluster-proportional-autoscaler-amd64:1.4.0
         resources:
             requests:
                 cpu: "20m"
                 memory: "10Mi"
+            limits:
+                memory: "10Mi"
```
rifelpet, line 62: Any reason this memory limit is being added for kubedns but not coredns?

upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -130,7 +136,7 @@ spec:
 
       containers:
       - name: kubedns
-        image: k8s.gcr.io/k8s-dns-kube-dns-{{Arch}}:1.14.13
+        image: k8s.gcr.io/k8s-dns-kube-dns:1.15.13
```
rifelpet, line 139: bootstrapchannelbuilder.go should be updated to match

https://github.com/kubernetes/kops/blob/77663acec2626a275d7bbd396b52331fac1d4ab3/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go#L229

(but only for the k8s 1.12 manifest, not the 1.6 manifest since you're not bumping the version there)

upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io/k8s-1.12.yaml.template
olemarkus: I'll move this and the kube-dns bump to a different PR, I think.

upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io/k8s-1.12.yaml.template
olemarkus: Same as above

johngmyers:
Now that I think about it, it would probably be easier to get the masters running on ARM than to support clusters without AMD worker nodes. We have more control over the images on the master.

I also don't like moving workloads to the masters that don't need to be on the masters, to not increase the attack surface of the high-privilege nodes. The DNS autoscaler admittedly doesn't take a whole lot of input, though.

johngmyers:
Is there an issue asking for a multiarch dns autoscaler?

olemarkus:
No issue on it that I could find. We need to find a better solution to the dns autoscaler before 1.19 GA, I think. But right now we need this to get a working cluster.

hakman:
The main blockers for ARM64 masters are:
* https://github.com/kopeio/etcd-manager/pull/315
* https://github.com/kubernetes/kops/pull/9403

I don't think this change for the DNS autoscaler is such a big deal. Most clusters could probably live without it too.

johngmyers:
I'm thinking separate `MasterArch` and `NodesArch` templating functions could work, except for `DaemonSet`s that have to run across everything.

hakman:
I think we are complicating things for no good reason. This is pretty much the only core thing we need now. Hopefully it won't take long to get things working on masters too.

rifelpet:
Looking at the autoscaler project's [Makefile](https://github.com/kubernetes-sigs/cluster-proportional-autoscaler/blob/master/Makefile) it seems already be building images [for each architecture](https://github.com/kubernetes-sigs/cluster-proportional-autoscaler/releases/tag/1.8.1) so the only change needed is to push a [single multi-arch manifest](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) maybe we open a PR for them to support that, but in the meantime is there any objection to merging this as-is?

johngmyers:
No objection. It enables tests and can be walked back later.

johngmyers:
I can see `NodeArch` being useful to push the addons onto cheaper nodes in multi-arch clusters.

rifelpet:
👍 

/approve

hakman:
@johngmyers maybe would be best to make it generic and tied to an instance group, not necessarily tied to arch.

rifelpet:
FYI i opened this issue, feel free to add context to it or open a PR: https://github.com/kubernetes-sigs/cluster-proportional-autoscaler/issues/89

hakman:
Thanks @rifelpet. 
Probably we'll have to create or ping some issues in other repos, like multi-arch image for Flannel.

johngmyers:
@olemarkus Looks like this change broke k8s <= 1.13

olemarkus:
I'll have a look

johngmyers:
The `kubernetes.io/arch` label was added in k8s 1.14 with kubernetes/kubernetes@04575f01b0426f8494d17dc85d47200fcdef3a5e

olemarkus:
We may have to split the manifest then as the beta labels were removed in 1.18

olemarkus:
Seems easier to wait and see if the autoscaler can support multiarch. Or temporarily remove the use of the autoscaler for the master branch.

PR #9425, opened 2020-06-22
Enable nodeport by default

olemarkus:
Upstream default since cilium 1.7 and perhaps the primary reason why people pick cilium.


rifelpet:
should this be enabled for a minimum k8s version only?

johngmyers:
Yeah, probably shouldn't default to running this way with Cilium 1.6.9, which is the default of k8s 1.11.

I tend to prefer changing this to be the default for k8s >= 1.19. But what is the rule we use for when defaults depend on new cluster or not versus k8s version?

olemarkus:
That's true.
I'll see if I can get a version check in there.

PR #9428, opened 2020-06-23
Update bazel rules versions

rifelpet:
In preparation for updating to go 1.14, we need to at least update rules_go. I updated the other rules while I was at it.

* [rules_go](https://github.com/bazelbuild/rules_go/releases/tag/v0.23.3)
* [rules_docker](https://github.com/bazelbuild/rules_docker#setup) 
* [bazel-gazelle](https://github.com/bazelbuild/bazel-gazelle/releases/tag/v0.21.1)
* [bazel-toolchains](https://github.com/bazelbuild/bazel-toolchains/releases/tag/2.2.0) to match the .bazelversion file

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -6,19 +6,19 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "dbde93957f2a96e6c97edb74e29107b7de1b2b57608b8304401e10326e02c66d",
+    sha256 = "e0d2e3d92ef8b3704f26ac19231ef9aba66c8a3bdec4aca91a22ad7d6e6f3ef7",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.22.5/rules_go-v0.22.5.tar.gz",
-        "https://github.com/bazelbuild/rules_go/releases/download/v0.22.5/rules_go-v0.22.5.tar.gz",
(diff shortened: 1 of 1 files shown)


rifelpet:
looks like rules_go updated the output path of binaries: https://github.com/bazelbuild/rules_go/releases/tag/v0.23.0 from `bazel-bin/cmd/kops/linux_amd64_pure_stripped/kops` to `bazel-bin/cmd/kops/kops_/kops`. We're now encouraged to use "[configuration transitions](https://github.com/bazelbuild/rules_go/pull/2249)" for multi-arch builds, which I'm still reading up on. If someone knows what changes need to be made feel free to suggest them.

rifelpet:
bazel make targets are passing but i'm not happy with how we're copying from bazel-bin to BAZELUPLOAD after each bazel build command.

Maybe theres a better way?
/assign @justinsb 
/hold for comment

Also noticed that [this](https://github.com/bazelbuild/rules_go/issues/1351) issue is resolved so we might be able to take advantage of that too.

justinsb:
Thanks for looking at this @rifelpet - it looks like we'll have to tackle this as part of rules-go 0.23.  It looks like k/kubernetes and even k/test-infra are holding on 0.22.6 (which adds support for the newer go versions).  I suspect we don't want to be an early adopter here - should we hold this one until e.g. test-infra shows us how to go to 0.23, but in the meantime we can go to 0.22.6?

rifelpet:
Ah good eye. I just updated this to 0.22.6 to match upstream. I believe that k/k doesn't use bazel to build release binaries, only `go test -c -o` https://github.com/kubernetes/kubernetes/blob/master/hack/lib/golang.sh#L677 so perhaps we consider doing that too.

Once this lands we can consider upgrading to go 1.14 and updating k/k dependencies to 1.19

PR #9431, opened 2020-06-23
Fix dns selectors for older k8s

olemarkus:
Use `kubernetes.io/role` instead of `kubernetes.io/arch` as nodeselector. This will work for all supported versions of K8s.
ref #9418 

/cc @hakman @johngmyers

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -1281,7 +1281,7 @@ spec:
           - --logtostderr=true
           - --v=2
       nodeSelector:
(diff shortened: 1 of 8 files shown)


hakman:
Thanks @olemarkus !
/lgtm

johngmyers:
@olemarkus k8s 1.11 is still broken. It's scheduling the pods but the containers aren't becoming ready. Perhaps walk back the 1.6 template to simply directly use the amd64 images? I don't think we're likely to support ARM on k8s 1.11.

PR #9435, opened 2020-06-24
Remove kube-discovery

johngmyers:
Appears to be related to baremetal.


hakman:
As discussed during last office hours, this should also be removed.
/lgtm

PR #9443, opened 2020-06-25
Refactor more cluster creation code into NewCluster()

johngmyers:


olemarkus:
/lgtm

Assume the renaming is to match the kops API

johngmyers:
Yes, but I'm wondering what I should do about `OpenstackNetworkID`, which appears to be separate only so the flag can have a different name. Perhaps I should map that to `NetworkID` in create_cluster.

olemarkus:
I'd map it. 

/cc @zetaab

johngmyers:
I mapped it.

pkg/apis/kops/channel.go
```diff
@@ -84,6 +84,10 @@ type ChannelImageSpec struct {
 
 // LoadChannel loads a Channel object from the specified VFS location
 func LoadChannel(location string) (*Channel, error) {
+	if location == "none" {
```
rifelpet, line 87: why is this needed? I don't see `none` referenced anywhere else

johngmyers: This allows for a simple way to opt out of channels.

rifelpet: ah ok. It would be helpful to document the different values for `--channel` in the flag's help text (`stable`, `alpha`, `none`, or a URL? any others?) but that could be done in a separate PR.

rifelpet: and consider adding `none` here: https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/channel.go#L33 and using it in the string comparison. I'd be hesitant to add an `alpha` const because we don't actually reference the specific value anywhere, we just treat it as a relative path appended to the channel base URL. But maybe it would be helpful for the flag's help text.

PR #9444, opened 2020-06-25
Update Calico to v3.15.0 for k8s 1.16+

hakman:
https://docs.projectcalico.org/archive/v3.15/release-notes/

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
@@ -1,4 +1,4 @@
-# Pulled and modified from: https://docs.projectcalico.org/v3.13/manifests/calico-typha.yaml
+# Pulled and modified from: https://docs.projectcalico.org/v3.15/manifests/calico-typha.yaml
 
 ---
 # Source: calico/templates/calico-config.yaml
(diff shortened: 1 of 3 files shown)


upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -61,232 +65,3131 @@ data:
 ---
 # Source: calico/templates/kdd-crds.yaml
 
-apiVersion: apiextensions.k8s.io/v1beta1
+
+---
```
johngmyers, line 69: Empty resources have caused us problems in the past. Are we stripping all these out now?

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -61,232 +65,3131 @@ data:
...
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: (devel)
```
johngmyers, line 74: Seems suspicious to have a version of "(devel)"

hakman: Yeah, seems to be a small mess in the official templates.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
hakman: We are stripping them now. I thought is best to keep them as close as possible to the official manifest, even if not as clean.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
hakman: Calico is moving to install via operator but still provides the manifests. Probably this is why the mess. At some point we should look into that and see if we can convert our templates also.

rdrgmnzs:
Thanks @hakman !

/lgtm
/approve

PR #9458, opened 2020-06-28
Add unit test for func VersionedJSON

Hellcatlk:
Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/kopscodecs/codecs_test.go
+++ b/pkg/kopscodecs/codecs_test.go
@@ -76,6 +76,40 @@ func TestToVersionedYaml(t *testing.T) {
 
 }
 
+func TestToVersionedJSON(t *testing.T) {
+	grid := []struct {
+		obj      runtime.Object
+		expected string
+	}{
+		{
+			obj: &v1alpha2.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					CreationTimestamp: testTimestamp,
+					Name:              "hello",
+				},
+				Spec: v1alpha2.ClusterSpec{
+					KubernetesVersion: "1.2.3",
+				},
+			},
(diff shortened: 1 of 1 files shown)


pkg/kopscodecs/codecs_test.go
```diff
@@ -76,6 +76,41 @@ func TestToVersionedYaml(t *testing.T) {
...
+			obj: &v1alpha2.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					CreationTimestamp: testTimestamp,
+					Name:              "hello",
+				},
+				Spec: v1alpha2.ClusterSpec{
+					KubernetesVersion: "1.2.3",
+				},
+			},
+			expected: heredoc.Doc(`{"kind":"Cluster","apiVersion":"kops.k8s.io/v1alpha2","metadata":{"name":"hello","creationTimestamp":"2017-01-01T00:00:00Z"},"spec":{"kubernetesVersion":"1.2.3"}}`),
```
olemarkus, line 94: I don't think you need to use `heredoc.Doc` here. In the test above, the json is formatted, while here it is not

pkg/kopscodecs/codecs_test.go
```diff
@@ -76,6 +76,41 @@ func TestToVersionedYaml(t *testing.T) {
...
+	}
+	for _, g := range grid {
+		actualBytes, err := ToVersionedJSON(g.obj)
+		if err != nil {
+			t.Errorf("error from ToVersionedJSON: %v", err)
+			continue
+		}
+		actual := string(actualBytes)
+		actual = strings.TrimSpace(actual)
+		expected := strings.TrimSpace(g.expected)
```
olemarkus, line 105: Given my comment above, you also do not need to trim spaces (there are no spaces in the string even).

pkg/kopscodecs/codecs_test.go
Hellcatlk: Thanks, done.

PR #9465, opened 2020-06-30
Update the service manifest for Docker

hakman:
Updating based on latest Docker service manifest:
https://github.com/docker/docker-ce-packaging/blob/360f755c768ebb06c9974705674302098bb9c398/systemd/docker.service

This ended up being a bit bigger than expected, but mostly because a new version of `github.com/blang/semver/v4` was needed to parse the docker version, which contains leading `0`s.

Summary (with similar commit names):
* add cgroup related dependencies
* use "infinity" for LimitNOFILE and LimitNPROC for less accounting overhead
* remove workaround for old unsupported Docker versions
* remove workaround for socket activation on CentOS/RHEL distros
* rearrange manifest to match official manifest
* add containerd service dependency for newer Docker versions (with new `semver` as prerequisite)


johngmyers:
Appears to also remove code for unsupported old docker versions and remove a RHEL workaround.

hakman:
> Appears to also remove code for unsupported old docker versions and remove a RHEL workaround.

That is the intent. The unsupported old docker versions were already removed and the RHEL workaround should no longer be needed.

rifelpet:
overall this looks good to me. regarding the removed docker versions, is there any api validation that we should do there to alert users to upgrade if their manifest is still specifying an older docker version? or what exactly will the user experience be in that case?

hakman:
@rifelpet The validation was introduced in a previous commit:
https://github.com/kubernetes/kops/blob/39db604e0c8b5657c8655bb6086bd866adf846cc/pkg/apis/kops/validation/validation.go#L1005-L1013

PR #9467, opened 2020-07-01
Move more cluster creation code to NewCluster()

johngmyers:


cmd/kops/create_cluster.go
```diff
@@ -400,17 +388,19 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
...
+				return err
+			}
 		}
-	}
 
-	if cluster != nil {
-		return fmt.Errorf("cluster %q already exists; use 'kops update cluster' to apply changes", c.ClusterName)
+		if cluster != nil {
+			return fmt.Errorf("cluster %q already exists; use 'kops update cluster' to apply changes", c.ClusterName)
+		}
```
hakman, line 403: Any special reason why this gets its own code block?

johngmyers: To take `cluster` out of scope from subsequent code.

hakman: 👍

PR #9471, opened 2020-07-01
Create one nodes instance group per zone

### LINKED ISSUE #8152: Proposal: Have new multi-zone AWS clusters create node instance groups per zone

Currently `kops create cluster --zones us-east-1a,us-east-1b,us-east-1c` will create a single instance group (autoscaling group) that spans all 3 zones. This is problematic for autoscaling because any autoscaler can't choose which zone a new instance will be launched in. This impacts pod scheduling for StatefulSets that use EBS Volumes as well as various affinity/antiaffinity rules. 

We can avoid these limitations by having new clusters create 3 instance groups, one for each zone. Kops users often discover these limitations after having created the cluster and adjust their InstanceGroups as described, so if we can address it automatically, it is one less hurdle for them to deal with.

This would only impact the `kops create cluster` command. Anyone providing InstanceGroup manifests directly would be unaffected.

The downside of this is that there are more AWS resources which may be overwhelming for new users experimenting with Kops.

I dont know if other cloud providers have similar limitations but we could extend this to other clouds if appropriate.

johngmyers:
Creates one nodes instance group per zone, restricted to that zone. Changes the default number of nodes to the number of zones. If fewer nodes than zones, only creates one instance group per node.

Fixes #8152

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -215,7 +215,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().StringVar(&sshPublicKey, "ssh-public-key", sshPublicKey, "SSH public key to use (defaults to ~/.ssh/id_rsa.pub on AWS)")
 
 	cmd.Flags().Int32Var(&options.MasterCount, "master-count", options.MasterCount, "Set number of masters. Defaults to one master per master-zone")
-	cmd.Flags().Int32Var(&options.NodeCount, "node-count", options.NodeCount, "Set number of nodes")
+	cmd.Flags().Int32Var(&options.NodeCount, "node-count", options.NodeCount, "Set number of nodes. Defaults to one node per zone")
 
 	cmd.Flags().StringVar(&options.Image, "image", options.Image, "Set image for all instances.")
 	cmd.Flags().StringVar(&options.NodeImage, "node-image", options.NodeImage, "Set image for nodes. Takes precedence over --image")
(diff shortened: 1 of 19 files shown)


hakman:
> If fewer nodes than zones, only creates one instance group per node.

I would go with changing what number of nodes means and applying it to each instance group.

olemarkus:
I would also change the meaning of `node-count` to be number of nodes per AZ. Isn't that what happens as the implementation is exactly now?

hakman:
@olemarkus at the moment an ASG can span multiple AZs so, you just set a fixed number of nodes that are randomly assigned to an AZ.

olemarkus:
Sorry, I missed a change in `new_cluster.go`.

I would either change this to nodes per AZ or documenting that this is the total number of nodes in the cluster. If we go for nodecount remaining the total number of nodes in the cluster, how we distribute the nodes across zones should be mentioned in the docs too.

johngmyers:
`--master-count` is the total number of master nodes, not the number of master nodes per IG. Making `--nodes-count` be the number of nodes per IG would be inconsistent.

The documentation says "Set number of nodes. Defaults to one node per zone", which is precisely the behavior implemented.

cmd/kops/create_cluster.go
```diff
@@ -215,7 +215,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().StringVar(&sshPublicKey, "ssh-public-key", sshPublicKey, "SSH public key to use (defaults to ~/.ssh/id_rsa.pub on AWS)")
 
 	cmd.Flags().Int32Var(&options.MasterCount, "master-count", options.MasterCount, "Set number of masters. Defaults to one master per master-zone")
-	cmd.Flags().Int32Var(&options.NodeCount, "node-count", options.NodeCount, "Set number of nodes")
+	cmd.Flags().Int32Var(&options.NodeCount, "node-count", options.NodeCount, "Set number of nodes. Defaults to one node per zone")
```
geojaz, line 218: Could you clarify "Set total number of nodes"?

johngmyers: Done

rdrgmnzs:
Thanks @johngmyers !

/lgtm
/approve

PR #9473, opened 2020-07-02
Add unit test for util/pkg/ui/user.go

Hellcatlk:
Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>


johngmyers:
Thank you, but this is not a useful unit test. It only tests the mode of `GetConfirm()` that stubs out the confirmation for use in tests. A test that exercised the user-promting code would be useful.

PR #9476, opened 2020-07-02
[Digital Ocean] Implement KOPS validate cluster

srikiz:
Implementation of KOPS validate cluster for Digital Ocean KOPS provider.
Tested with `./kops validate cluster` and it returns the state of your KOPS cluster as expected.

### DIFF
--- a/pkg/resources/digitalocean/cloud.go
+++ b/pkg/resources/digitalocean/cloud.go
@@ -324,9 +324,7 @@ func getCloudGroups(c *Cloud, cluster *kops.Cluster, instancegroups []*kops.Inst
 // FindInstanceGroups finds instance groups matching the specified tags
 func FindInstanceGroups(c *Cloud, clusterName string) ([]DOInstanceGroup, error) {
 	var result []DOInstanceGroup
-	var masterDroplets []string
-	var workerDroplets []string
-	nodeExists := false
+	instanceGroupMap := make(map[string][]string) // map of instance group name with droplet ids
 
 	clusterTag := "KubernetesCluster:" + strings.Replace(clusterName, ".", "-", -1)
 	droplets, err := getAllDropletsByTag(c, clusterTag)
@@ -341,36 +339,14 @@ func FindInstanceGroups(c *Cloud, clusterName string) ([]DOInstanceGroup, error)
 			return nil, fmt.Errorf("get droplets Instance group for tags %v returned error. Error=%v", droplet.Tags, err)
 		}
 
-		if strings.Contains(doInstanceGroup, "master") {
-			// This is a master instance group, get the index.
-			k8sIndex, err := getDropletIndexFromTag(droplet.Tags)
-			if err != nil {
-				klog.Warningf("tag check returned err = %v", err)
-			}
-			instanceGroupName = fmt.Sprintf("%s-%s-%s", clusterName, doInstanceGroup, k8sIndex)
-
-			result = append(result, DOInstanceGroup{
-				InstanceGroupName: instanceGroupName,
-				GroupType:         "master",
-				ClusterName:       clusterName,
-				Members:           append(masterDroplets, strconv.Itoa(droplet.ID)),
-			})
-
-		} else if strings.Contains(doInstanceGroup, "nodes") {
-			// This is a node (add this instance group only once)
-			nodeExists = true
-			workerDroplets = append(workerDroplets, strconv.Itoa(droplet.ID))
-		}
-	}
-
-	if nodeExists {
-		instanceGroupName = fmt.Sprintf("%s-%s", clusterName, "nodes")
+		instanceGroupName = fmt.Sprintf("%s-%s", clusterName, doInstanceGroup)
+		instanceGroupMap[instanceGroupName] = append(instanceGroupMap[instanceGroupName], strconv.Itoa(droplet.ID))
 
 		result = append(result, DOInstanceGroup{
 			InstanceGroupName: instanceGroupName,
-			GroupType:         "node",
+			GroupType:         instanceGroupName,
 			ClusterName:       clusterName,
-			Members:           workerDroplets,
+			Members:           instanceGroupMap[instanceGroupName],
 		})
 	}
 
@@ -391,22 +367,7 @@ func getDropletInstanceGroup(tags []string) (string, error) {
 		}
 	}
 
-	return "", fmt.Errorf("Didn't find k8s-index for tag %v", tags)
-}
-
-func getDropletIndexFromTag(tags []string) (string, error) {
-	for _, tag := range tags {
-		klog.V(8).Infof("Check tag = %s", tag)
-		if strings.Contains(strings.ToLower(tag), TagKubernetesClusterIndex) {
-			tagParts := strings.Split(tag, ":")
-			if len(tagParts) < 2 {
-				return "", fmt.Errorf("tag split failed, too few components for tag %q", tag)
-			}
-			return tagParts[1], nil
-		}
-	}
-
-	return "", fmt.Errorf("Didn't find k8s-index for tag %v", tags)
+	return "", fmt.Errorf("Didn't find k8s-instancegroup for tag %v", tags)
 }
 
 // matchInstanceGroup filters a list of instancegroups for recognized cloud groups
--- a/pkg/resources/digitalocean/cloud.go
+++ b/pkg/resources/digitalocean/cloud.go
@@ -88,8 +88,7 @@ func NewCloud(region string) (*Cloud, error) {
 
 // GetCloudGroups is not implemented yet, that needs to return the instances and groups that back a kops cluster.
 func (c *Cloud) GetCloudGroups(cluster *kops.Cluster, instancegroups []*kops.InstanceGroup, warnUnmatched bool, nodes []v1.Node) (map[string]*cloudinstances.CloudInstanceGroup, error) {
-	klog.V(8).Info("digitalocean cloud provider GetCloudGroups not implemented yet")
-	return nil, fmt.Errorf("digital ocean cloud provider does not support getting cloud groups at this time")
+	return nil, nil
 }
 
 // DeleteGroup is not implemented yet, is a func that needs to delete a DO instance group.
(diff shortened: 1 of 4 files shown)


pkg/resources/digitalocean/cloud.go
```diff
@@ -278,3 +287,172 @@ func (c *Cloud) getEtcdClusterSpec(volumeName string, dropletName string) (*etcd
...
+		Raw:           g,
+		MinSize:       int(fi.Int32Value(ig.Spec.MinSize)),
+		TargetSize:    int(fi.Int32Value(ig.Spec.MinSize)),
+		MaxSize:       int(fi.Int32Value(ig.Spec.MaxSize)),
+	}
+
+	for _, member := range g.Members {
+
+		// DO doesn't have a notion of Auto Scaling Group - use the same config Type for both current config and new config.
+		err := cg.NewCloudInstanceGroupMember(member, g.GroupType, g.GroupType, nodeMap)
```
johngmyers, line 451: This means DO won't be able to know when to do a rolling update. All rolling updates would  have to use `--force`.

rifelpet: Looking at the [DO Cloud](https://github.com/kubernetes/kops/tree/master/upup/pkg/fi/cloudup/do), it doesn't (yet) appear to support deleting instances, so I'm guessing rolling updates don't work at all. Perhaps @srikiz can confirm?

pkg/resources/digitalocean/cloud.go
```diff
@@ -38,6 +39,8 @@ import (
 
 const TagKubernetesClusterIndex = "k8s-index"
 const TagKubernetesClusterNamePrefix = "KubernetesCluster"
+const TagKubernetesInstanceGroup = "kops-instancegroup"
+const DOInstanceGroupConfig = "do-instancegroup-config"
```
johngmyers, line 43: Where is this referenced?

hakman:
Let's wait a bit before merging this.
/lgtm cancel

pkg/resources/digitalocean/cloud.go
```diff
@@ -38,6 +39,8 @@ import (
 
 const TagKubernetesClusterIndex = "k8s-index"
 const TagKubernetesClusterNamePrefix = "KubernetesCluster"
+const TagKubernetesInstanceGroup = "kops-instancegroup"
```
johngmyers, line 42: Why have a duplicate of `do.TagKubernetesClusterInstanceGroupPrefix`?

johngmyers:
The DO code is a bit oversimplified, apparently because DO doesn't support autoscaling groups. But kops doesn't care (much) about the autoscaling aspect of autoscaling groups.

The existing `DropletBuilder` tags all instances that don't have master role as being in the "nodes" instance group. This ignores the possibility of the admin creating additional instance groups and causes this code to always identify non-master nodes as being in the "nodes" instance group, regardless of which instance group actually created them.

For validation, `GetCloudGroups` needs to identify:
* The instance groups deployed (from whatever ASG-equivalent concept the provider code maps instance groups to)
* The target size of the ASG-equivalent
* The extant instances, and for each:
** Which instance group caused the instance to be created
** Which k8s node the instance corresponds to, if the instance has joined the cluster

For rolling update, it needs to identify:
** Whether the instance was created with the version of the spec that is the most current for the ASG-equivalent
** Whether the instance is detached (if the provider supports surging)

For rolling update, it also needs `DeleteInstance` to terminate the instance and cause a replacement instance to be created. (Except, if it supports surging, to not replace the instance if it was previously detached.)

johngmyers:
To further clarify what I wrote, validation will throw failures if there aren't enough instances in a group to cover what that group reports as its target size or if any instances don't have a corresponding node (haven't joined the cluster). Additionally it will throw a failure if `GetCloudGroups` fails to return a group for an extant InstanceGroup.

Rolling update needs `GetCloudGroups` to identify which instances are out of date, which `NewCloudInstanceGroupMember` determines by whether the second and third parameters compare equal. It also needs to be able to terminate instances.  It needs validation to eventually succeed after termination, which means that after termination the group needs to eventually have enough instances to match its reported target size.

pkg/resources/digitalocean/cloud.go
srikiz: yes @rifelpet - that is something I want to check next.

johngmyers: Yeah, looks like DO doesn't support rolling updates. So that's not a blocker for this PR.

The comment is a bit off-point—the second and third parameters just need to be unequal whenever any field of the `godo.DropletCreateRequest` used to create the instance was different than what would be used to create a new instance. This can be implemented as either a generation id or a hash of the fields.

srikiz: sure, thanks @johngmyers  - I'll keep a note of this when I implement rolling update feature.

pkg/resources/digitalocean/cloud.go
srikiz: thanks, removed DOInstanceGroupConfig

pkg/resources/digitalocean/cloud.go
srikiz: @johngmyers  - I think this needs refactoring. I tried to use it and it looks like we have a cyclic dependency here. Ideally, I should have kept this logic (and few other methods) in the do package, but that's a little more work that I thought of.
Would it be okay if we get this merged, and I take up the clean up activity as a separate PR? Please suggest.

srikiz:
> To further clarify what I wrote, validation will throw failures if there aren't enough instances in a group to cover what that group reports as its target size or if any instances don't have a corresponding node (haven't joined the cluster). Additionally it will throw a failure if `GetCloudGroups` fails to return a group for an extant InstanceGroup.
> 
> Rolling update needs `GetCloudGroups` to identify which instances are out of date, which `NewCloudInstanceGroupMember` determines by whether the second and third parameters compare equal. It also needs to be able to terminate instances. It needs validation to eventually succeed after termination, which means that after termination the group needs to eventually have enough instances to match its reported target size.

- Thanks, this helps. I need to discuss with @timoreimann if we can support Rolling update for DO. I'll be doing that next week. In the meanwhile, I would like to get this current PR going, if you think it's okay. Below is the validate cluster response -

`Validating cluster dev5.k8s.local

INSTANCE GROUPS
NAME            ROLE    MACHINETYPE     MIN     MAX     SUBNETS
master-tor1-1   Master  s-2vcpu-4gb     1       1       tor1
master-tor1-2   Master  s-2vcpu-4gb     1       1       tor1
master-tor1-3   Master  s-2vcpu-4gb     1       1       tor1
nodes           Node    s-2vcpu-4gb     2       2       tor1

NODE STATUS
NAME            ROLE    READY
10.137.176.248  master  True
10.137.240.102  master  True
10.137.240.248  master  True
10.137.240.3    node    True
10.137.8.136    node    True

Your cluster dev5.k8s.local is ready`

pkg/resources/digitalocean/cloud.go
johngmyers: Perhaps move `TagKubernetesClusterInstanceGroupPrefix` here?

pkg/resources/digitalocean/cloud.go
```diff
@@ -278,3 +286,172 @@ func (c *Cloud) getEtcdClusterSpec(volumeName string, dropletName string) (*etcd
...
+		InstanceGroup: ig,
+		Raw:           g,
+		MinSize:       int(fi.Int32Value(ig.Spec.MinSize)),
+		TargetSize:    int(fi.Int32Value(ig.Spec.MinSize)),
+		MaxSize:       int(fi.Int32Value(ig.Spec.MaxSize)),
+	}
+
+	for _, member := range g.Members {
+
+		// DO doesn't have a notion of Auto Scaling Group - use the same config Type for both current config and new config.
```
johngmyers, line 449:
```suggestion
		// TODO use a hash of the godo.DropletCreateRequest fields for second and third parameters.
```

johngmyers:
@srikiz Rolling update support does not block this PR. In my mind the thing that blocks this PR is its identifying all non-master instances as being in the "nodes" instance group. The droplet creation code should tag instances with the name of the instance group that created them so this code can assign them to the right group.

The code for assigning the group of the masters seems unnecessarily complex. If there's a reason the masters need to have special case string manipulation, please explain.

johngmyers:
I believe there is scope to share code between this and `dotasks.Droplet.Find()`. `dotasks.Droplet` is really a `DropletGroup`.

pkg/resources/digitalocean/cloud.go
srikiz: sure, updated.

srikiz:
> @srikiz Rolling update support does not block this PR. In my mind the thing that blocks this PR is its identifying all non-master instances as being in the "nodes" instance group. The droplet creation code should tag instances with the name of the instance group that created them so this code can assign them to the right group.
> 
> The code for assigning the group of the masters seems unnecessarily complex. If there's a reason the masters need to have special case string manipulation, please explain.

Thanks for the inputes - I updated code to also check for instance group called "nodes" if it doesn't match the master. While creating the droplet, we currently tag the master with 3 different tags.

- kubernetescluster: {clusterName}
- Kops-instancegroup: master-{region}
- k8s-index: 1 (will always be 1 for a single master cluster). For a multi-master node, it can be K8s-Index:2, or k8s-index:3
The instance group itself is a combination of kops-instancegroup and the k8s-index. 
I will keep a note to find if I could update kops-instancegroup itself to the correct value (instead of concatenating with k8s-index).

For the worker nodes we have the below tags.

- Kubernetescluster: {clusterName}
- Kops-instancegroup: nodes

I updated code to also check if the matching "kops-instancegroup: nodes" tag is available on the droplet.

johngmyers:
Why don't you put the instancegroup's `ObjectMeta.Name` in kops-instancegroup?

It looks like the code will only return at most one `CloudInstanceGroup` of non-master type. That means cluster validation will always fail if there are two or more non-master instance groups.

srikiz:
> Why don't you put the instancegroup's `ObjectMeta.Name` in kops-instancegroup?
> 
> It looks like the code will only return at most one `CloudInstanceGroup` of non-master type. That means cluster validation will always fail if there are two or more non-master instance groups.

I don't know if there is a way in DO to add more non-master instance groups. DO doesn't support any new instance group creation. I am currently managing with the help of tags. I tag droplets that act as worker nodes with "kops-instancegroup:nodes" tag. If there are more than 1 DO KOPS cluster, I currently retrieve all droplets that match the KOPS cluster name here - https://github.com/kubernetes/kops/pull/9476/files#diff-09d5ce1959bcc4dffcb0d3d88a1ffb55R332

srikiz:
> I believe there is scope to share code between this and `dotasks.Droplet.Find()`. `dotasks.Droplet` is really a `DropletGroup`.

Yeah. I moved the consts from the current package (upup/pkg/fi/cloudup/do) into pkg/resources/digitialocean. It I had to modify droplets.go, api_loadbalancer.go, master_volumes.go, etc.  I thought it would be safer to make this change as a separate PR. Hope that's okay.

johngmyers:
`dotasks.Droplet` is essentially a (non-automatic) scaling group. Its `RenderDO()` handles instance groups by creating enough droplets to satisfy the instancegroup's expected count. I don't see anything that would prevent `Droplet` from creating droplets for multiple worker instance groups.

`Droplet` puts an ASG name, which is derived from the instancegroup and cluster names, into the `Name` field of each droplet it creates. This can be used to determine which instancegroup a given droplet belongs to. `Droplet.Find()` appears to essentially do this, though it filters for the instancegroup it was instantiated for.

srikiz:
> `dotasks.Droplet` is essentially a (non-automatic) scaling group. Its `RenderDO()` handles instance groups by creating enough droplets to satisfy the instancegroup's expected count. I don't see anything that would prevent `Droplet` from creating droplets for multiple worker instance groups.
> 
> `Droplet` puts an ASG name, which is derived from the instancegroup and cluster names, into the `Name` field of each droplet it creates. This can be used to determine which instancegroup a given droplet belongs to. `Droplet.Find()` appears to essentially do this, though it filters for the instancegroup it was instantiated for.

I see, got it. I will put Objectmeta.Name to the kops-instancegroup tag and verify it. Will update the PR shortly.

pkg/resources/digitalocean/cloud.go
srikiz: I removed TagKubernetesClusterInstanceGroupPrefix and instead used the above const.

srikiz:
@johngmyers - I have incorporated as per your comments - that really helped !
Please do have a look when you get a chance. Thanks !

pkg/resources/digitalocean/cloud.go
johngmyers: Resolved

pkg/resources/digitalocean/cloud.go
```diff
@@ -62,6 +64,13 @@ type Cloud struct {
 	RegionName string
 }
 
+type DOInstanceGroup struct {
+	ClusterName       string
+	InstanceGroupName string
+	GroupType         string   // will be either "master" or "worker"
```
johngmyers, line 70: Use `pkg/apis/kops/InstanceGroupRole` as the type?

srikiz: I'll keep a note to update this when I touch the rolling update feature. I'll add a new tag when I create a droplet that holds this information, so I can extract it from the droplet tag.
Thanks for your help in reviewing this @johngmyers !

johngmyers: You can get the group role from the instancegroup spec of the ig that was in the tag.

PR #9477, opened 2020-07-02
Dependencies for Toolbox Instance Selector

bwagner5:
This PR updates dependencies for the new `kops toolbox instance-selector` command. 

It adds an AWS go pkg `github.com/aws/amazon-ec2-instance-selector` to go.mod and vendors the pkg.


bwagner5:
I signed it

hakman:
@bwagner5 I think deps and code should be in same PR, just different commits.

bwagner5:
gotcha, I misread this as two PRs:

"Open a pull request with these changes separately from other work so that it is easier to review."

I'll close this PR in favor of the other that does include both deps and feature.

hakman:
I think the docs are now quite up to date there. 😄 
Thanks!

PR #9491, opened 2020-07-03
Default ClusterDNS appropriately when NodeLocalDNS is enabled

johngmyers:
/cc @olemarkus

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -583,7 +583,7 @@ spec:
 
 ## Node local DNS cache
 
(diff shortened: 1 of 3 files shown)


olemarkus:
Thanks for this. I think this does make things simpler for users.
If this goes in, we'd want this backported to 1.18 too.

/lgtm

johngmyers:
One question is whether `ClusterDNS` should default to `NodeLocalDNS.LocalIP` for all cases when `NodeLocalDNS` is enabled, not just the ones where it doesn't work otherwise.

olemarkus:
An earlier PR I had of this did that. I suspect there may be other cases where you need localIP. For example calico in BPF (kubeproxy-free) mode.

johngmyers:
/assign @rifelpet 
It would be good to get this into 1.18 branch soon.

rifelpet:
👍 
/approve

PR #9492, opened 2020-07-04
Remove support for legacy IAM permissions

johngmyers:


hakman:
Any reason for not removing this completely? Seems that they are unofficially deprecated since v1.8.0.
https://github.com/kubernetes/kops/blob/99adb5690b8b0c2e1a363312f2af58ca01060c37/docs/iam_roles.md
> On provisioning a new cluster with Kops v1.8.0 or above, by default you will be using the new stricter IAM policies. Upgrading an existing cluster will use the legacy IAM privileges to reduce risk of potential regression.

johngmyers:
We could probably be more aggressive, depreciating in kops 1.18 and removing in 1.19.

hakman:
> We could probably be more aggressive, depreciating in kops 1.18 and removing in 1.19.

👍

rifelpet:
I agree with the proposed timeline but I think we should get input from others before doing so. Perhaps an agenda item for the next office hours? Unless we think Kops 1.18 will be released before the next office hours, in which case we could at least get something into the release notes for now.

rifelpet:
Perhaps the release notes should mention a migration strategy or at least the impact of this:

> Cluster operators should ensure any pods using the EC2 instances' IAM instance profiles have the necessary permissions, either with the cluster's `spec.additionalPolicies` or by providing AWS credentials through different means.

johngmyers:
`kops create cluster` as of kops 1.8 has explicitly disabled legacy IAM, so I don't expect much of anything to be affected. If the cluster was created without an `iam` field, however, kops would have defaulted to legacy IAM. Is this likely?

johngmyers:
Per kops office hours, need to put in a escape hatch for 1.19, deferring code removal to 1.20

hakman:
I think the Cilium test is broken and those tests should be added to the skip list.

johngmyers:
@hakman I agree those should be added to the skip list. Do we have a tracking issue so we can stop skipping once they start passing?

hakman:
> @hakman I agree those should be added to the skip list. Do we have a tracking issue so we can stop skipping once they start passing?

We don't, but we will have to re-review skipped test on each Kubernetes minor release.

johngmyers:
Hrm, looks like that test is just flaky in kops-aws-cni-cilium. But a similar, consistently failing test is cilium/cilium#12489

PR #9494, opened 2020-07-05
Release 1.19.0-alpha.1

justinsb:
As this is the first release of the 1.19.0 alpha series, I think we
can try releasing from the automated build.  I _believe_ that tags
will now be automatically be built by CI.  So (if this works) the
process will be:

* We merge the "release" PR, but it's not yet tagged

* It merges and prow and cloudbuild build it like any other PR, with a
"CI" version (i.e. not the correct version number, yet).

* Our normal e2e tests will run against it (and our normal PR tests
  run against it).

* We then tag that version (manually, currently).  This should trigger
  a build with the version set to that tag - though it still goes to
  our staging area.

* We can perform additional validation against that build (including,
  in future, automated validation against the "last tag"?)

* We then promote that build to the release GCS / GCR / other registry
  / github areas.  Ideally we use a PR, wg-k8s-infra has some
  mechanics here, but we'll likely need our own tooling for github
  (until we can contribute those upstream).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,11 +56,11 @@ unexport KOPS_BASE_URL KOPS_CLUSTER_NAME KOPS_RUN_OBSOLETE_VERSION KOPS_STATE_ST
 unexport SKIP_REGION_CHECK S3_ACCESS_KEY_ID S3_ENDPOINT S3_REGION S3_SECRET_ACCESS_KEY
 
 # Keep in sync with upup/models/cloudup/resources/addons/dns-controller/
-DNS_CONTROLLER_TAG=1.18.0-alpha.3
+DNS_CONTROLLER_TAG=1.19.0-alpha.1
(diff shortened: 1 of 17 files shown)


johngmyers:
Not quite sure how to code review this. All the changes look reasonable, but I'd be unlikely to catch if one of the versions didn't get updated.

Since this isn't updating the change list in docs/releases/1.19-NOTES.md I presume there'd be a post-release PR as well?

/lgtm

rifelpet:
looks like the postsubmit job may need a longer timeout?

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/kops-postsubmit/1279603648619352064

I already increased it recently: https://github.com/kubernetes/test-infra/pull/18132 but we may need to increase it again or allocate more resources to the job.

PR #8882, opened 2020-04-09
Update to golang 1.13.9 to match 1.18.1

justinsb:

PR #9499, opened 2020-07-05
Upgrade to go 1.14.4

rifelpet:
k/k 1.19 is using go 1.14.4 so this updates kops to match. https://github.com/kubernetes/kubernetes/pull/88638

Using these PRs as a reference: https://github.com/kubernetes/kops/pull/8882 https://github.com/kubernetes/kops/pull/8886

Depends on https://github.com/kubernetes/kops/pull/9428

also reverts https://github.com/kubernetes/kops/pull/9396 since it is no longer needed (go 1.14's [default behavior changed](https://golang.org/doc/go1.14#go-command))

### DIFF
--- a/hack/make-apimachinery.sh
+++ b/hack/make-apimachinery.sh
@@ -30,16 +30,15 @@ mkdir -p "${WORK_DIR}/go/"
 cp -R "${GOPATH}/src/k8s.io/kops/vendor/" "${WORK_DIR}/go/src"
 
 unset GOBIN
-unset GOFLAGS
 
-env GOBIN="${WORK_DIR}/go/bin" GOPATH="${WORK_DIR}/go/" go install -v k8s.io/code-generator/cmd/conversion-gen/
+env GOBIN="${WORK_DIR}/go/bin" GOPATH="${WORK_DIR}/go/" go install -mod=mod -v k8s.io/code-generator/cmd/conversion-gen/
 cp "${WORK_DIR}/go/bin/conversion-gen" "${GOPATH}/bin/"
 
-env GOBIN="${WORK_DIR}/go/bin" GOPATH="${WORK_DIR}/go/" go install k8s.io/code-generator/cmd/deepcopy-gen/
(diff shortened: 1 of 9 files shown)


rifelpet:
We may have to update the required statuses in test-infra :/ 

https://github.com/kubernetes/test-infra/blob/1771aee48e8d2e88811e71d50b18994d79e01f53/config/prow/config.yaml#L195-L202

johngmyers:
Using "primary"/"secondary" would have the consequence of allowing a merge based on a GHA job that ran with an older go version. That might be okay, since it's only macos where we don't have Prow coverage.

rifelpet:
Yea that's true. I suppose we had the same limitation with Travis before GHA. Regardless I think we'll need to remove at least some of the GHA jobs from the required statuses and we'll need to wait some time before readding the new jobs' names in order to minimize disruption of other open PRs.

johngmyers:
Certainly allowing merges based on runs with older Go versions is better than allowing merges with failed runs.

rifelpet:
With #9552 merged this has been simplified a bit and is ready for review

hack/make-apimachinery.sh
```diff
@@ -30,16 +30,15 @@ mkdir -p "${WORK_DIR}/go/"
 cp -R "${GOPATH}/src/k8s.io/kops/vendor/" "${WORK_DIR}/go/src"
 
 unset GOBIN
-unset GOFLAGS
 
-env GOBIN="${WORK_DIR}/go/bin" GOPATH="${WORK_DIR}/go/" go install -v k8s.io/code-generator/cmd/conversion-gen/
+env GOBIN="${WORK_DIR}/go/bin" GOPATH="${WORK_DIR}/go/" go install -mod=mod -v k8s.io/code-generator/cmd/conversion-gen/
```
justinsb, line 34: Not sure whether this should be mod or vendor?  But I don't think it's a blocker.

justinsb:
This LGTM, but I'm afraid there's a conflict on the makefile

/approve
/lgtm

hack/make-apimachinery.sh
rifelpet: We're not vendoring k8s.io/code-generator so I believe this was always installing from upstream or the module cache. It might be a good idea to vendor it by adding it to tools.go but that can happen in a followup PR.

PR #9508, opened 2020-07-06
Update CoreDNS to v1.7.0

rajansandeep:
```
ACTION REQUIRED : In CoreDNS v1.7.0, [metrics names have been changed](https://github.com/coredns/coredns/blob/master/notes/coredns-1.7.0.md#metric-changes) which will be backward incompatible with existing reporting formulas that use the old metrics' names. Adjust your formulas to the new names before upgrading. 

Some of the major changes in CoreDNS v1.7.0 include:
-  Fixed a bug that could cause CoreDNS to stop updating service records.
-  Fixed a bug in the forward plugin where only the first upstream server is always selected no matter which policy is set.
-  Remove already deprecated options `resyncperiod` and `upstream` in the Kubernetes plugin.
-  Includes Prometheus metrics name changes (to bring them in line with standard Prometheus metrics naming convention). They will be backward incompatible with existing reporting formulas that use the old metrics' names.
-  The federation plugin (allows for v1 Kubernetes federation) has been removed.
More details are available in https://coredns.io/2020/06/15/coredns-1.7.0-release/
```

I've also updated the default Corefile according to the latest release.
 - Removes the already deprecated `upstream` option from the Kubernetes plugin
 - Adds the new `max_concurrent` option to forward plugin to mitigate query spikes that might otherwise cause CoreDNS to OOM.

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -1232,11 +1232,12 @@ data:
         }
         kubernetes {{ KubeDNS.Domain }}. in-addr.arpa ip6.arpa {
           pods insecure
-          upstream
           fallthrough in-addr.arpa ip6.arpa
         }
         prometheus :9153
-        forward . /etc/resolv.conf
+        forward . /etc/resolv.conf {
+          max_concurrent 1000
+        }
         loop
         cache 30
         loadbalance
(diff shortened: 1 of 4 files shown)


joshbranham:
I think you need to run `./hack/update-expected.sh`

hakman:
Ping @rajansandeep ^^^. Thanks!

rifelpet:
This has been superseded by #9541 so we can close this out. thanks again!

/close

Issue #7871, opened 2019-11-04
Use kops to setup k8s cluster in AWS China region to use the Route 53 DNS


liangruibupt:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
Now Route 53 can be preview on AWS China region, but whitelist user can not directly use the it. Because below code do not specify the Region and Endpoint for new Route 53 instance
https://github.com/kubernetes/kops/blob/f9f80f0b26e718e87c08716a2d4f9a7eaf3a2964/dnsprovider/pkg/dnsprovider/providers/aws/route53/route53.go#L60

**2. Feel free to provide a design supporting your feature request.**
1. aws.NewConfig() can specify region and endpoint, for example:
aws.NewConfig().WithRegion("cn-northwest-1").WithEndpoint(route53.amazonaws.com.cn)

2. Update the below guide to add route 53 support in AWS China region
https://github.com/kubernetes/kops/blob/cb66ae7af90f3af0cafea47706fa622c68e9eca9/docs/aws-china.md

liangruibupt:
any update ?

jecnua:
Still valid, please remove the label stale.

Issue #9138, opened 2020-05-18
Update AWS China documentation with Route53 Support


rifelpet:
One of the AWS China regions now supports Route53: https://aws.amazon.com/about-aws/whats-new/2020/05/amazon-route-53-is-now-available-in-AWS-china-region/

If someone with an AWS China account can try creating a cluster using Route53 rather than gossip k8s.local and update docs/aws-china.md with their findings that would be much appreciated.

I'd be particularly curious if one can use Route53 in the other region as well.

hakman:
After 1.18 is released, maybe we should bump the `aws-sdk-go` version to pick-up the update for Route53 in China. Should be pretty safe, as it is just a patch release:
https://github.com/aws/aws-sdk-go/pull/3321/files#diff-fe3b36d57dc4378fa7b13c85492f4119R6241-R6252

hakman:
No complaints about China for a very long time.
/close

PR #9510, opened 2020-07-07
Update aws-sdk-go to v1.32.13

hakman:
Newer versions of aws-sdk-go include the AWS China endpoints. Kops should work in China now without modifications.

Ref: #9138
Fixes: #7871


rifelpet:
Nice, [this version](https://github.com/aws/aws-sdk-go/releases/tag/v1.32.13) also adds "tag on create" support for more ec2 resources. I'm hoping to get our awstasks switched over to that soon which could help reduce the IAM permissions needed for the kops cli even further.

PR #9511, opened 2020-07-07
Update go.mod's gazelle to 0.21.1

rifelpet:
I had updated the gazelle version in WORKSPACE earlier but it turns out it is defined twice, and the hack/update-bazel.sh script uses the go module copy, so this updates the go.mod instance to match the WORKSPACE instance.

https://github.com/kubernetes/kops/commit/03e620bf074b8fb406466f0e6ac5d94053bfb6ba#diff-09498dbadf45966909850dc8a47ebb13L16-L23


rifelpet:
@hakman determined this depends on rules_go 0.23 which will require larger work around arch-specific builds, so we may defer until k/k upgrades to rules_go to update this as well.
/close

PR #9513, opened 2020-07-07
Move remaining new cluster setup to pkg

johngmyers:
This is intended to be the last PR in the series.

The intent is to leave settings with trivial logic in cmd: the assumption is that external callers could just as easily implement that themselves.


rifelpet:
looks great, thanks for all your hard work on this!
/approve

PR #9520, opened 2020-07-08
Cloudmock cleanup - preparation for EC2 tag-on-create

rifelpet:
In order to add "tag-on-create" support for ec2 resources, cloudmock first needs to support recognizing and storing tags provided in `ec2.Create*` requests. This way a `DescribeTags` call will return the tags provided either by the resource's `Create` call or in a `CreateTags` call.

Additionally this sets up launch templates to use their `lt-` ID prefix so that we can get/set their tags more easily (now that #9519 is merged).

### DIFF
--- a/cloudmock/aws/mockec2/dhcpoptions.go
+++ b/cloudmock/aws/mockec2/dhcpoptions.go
@@ -126,9 +126,10 @@ func (m *MockEC2) CreateDhcpOptions(request *ec2.CreateDhcpOptionsInput) (*ec2.C
 	}
 
 	n := len(m.DhcpOptions) + 1
+	id := fmt.Sprintf("dopt-%d", n)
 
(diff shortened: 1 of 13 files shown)


cloudmock/aws/mockec2/dhcpoptions.go
```diff
@@ -147,8 +149,11 @@ func (m *MockEC2) CreateDhcpOptions(request *ec2.CreateDhcpOptionsInput) (*ec2.C
 	}
 	m.DhcpOptions[*dhcpOptions.DhcpOptionsId] = dhcpOptions
 
+	for _, tag := range tags {
+		m.addTag(id, tag)
```
johngmyers, line 153: Perhaps make `MockEC2.addTag()` be variadic?

rifelpet: Thanks, updated

PR #7476, opened 2019-08-26
fix(upup/models/cloudup/resources/addons/kube-dns.addons.k8s.io) missing resourceVersion

phspagiari:
This PR fixes a protokube error while applying the `kube-dns` addon.  [We can see the same behavior (already fixed) on kubeadm.](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/addons/dns/manifests.go#L194-L196)

## Error logs

```
I0826 22:05:34.677310   14969 addon.go:130] Applying update from "s3://BUCKET-STATE-STORE/CLUSTER-NAME/addons/kube-dns.addons.k8s.io/k8s-1.12.yaml"
I0826 22:05:34.677339   14969 s3fs.go:220] Reading file "s3://BUCKET-STATE-STORE/CLUSTER-NAME/addons/kube-dns.addons.k8s.io/k8s-1.12.yaml"
I0826 22:05:34.693912   14969 apply.go:67] Running command: kubectl apply -f /tmp/channel974681140/manifest.yaml
I0826 22:05:34.853468   14969 apply.go:70] error running kubectl apply -f /tmp/channel974681140/manifest.yaml
I0826 22:05:34.853492   14969 apply.go:71] deployment.apps/kube-dns-autoscaler unchanged
deployment.apps/kube-dns configured
serviceaccount/kube-dns-autoscaler unchanged
clusterrole.rbac.authorization.k8s.io/kube-dns-autoscaler unchanged
clusterrolebinding.rbac.authorization.k8s.io/kube-dns-autoscaler unchanged
The Service "kube-dns" is invalid: metadata.resourceVersion: Invalid value: "": must be specified for an update
Error: error updating "kube-dns.addons.k8s.io": error applying update from "kube-dns.addons.k8s.io/k8s-1.12.yaml": error running kubectl
Usage:
  channels apply channel [flags]

Flags:
  -f, --filename strings   Apply from a local file
  -h, --help               help for channel
      --yes                Apply update

Global Flags:
      --alsologtostderr                  log to standard error as well as files
      --config string                    config file (default is $HOME/.channels.yaml)
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --log_file string                  If non-empty, use this log file
      --log_file_max_size uint           Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
[...]

PR #9521, opened 2020-07-08
Fix KubeDNS missing resourceVersion

hakman:
Another attempt at https://github.com/kubernetes/kops/pull/7476.


olemarkus:
/lgtm
Assume the test will eventually pass

PR #9533, opened 2020-07-09
Add tagging support for AWS Keypairs

rifelpet:
AWS recently added tagging support for keypairs as well as autogenerated IDs. This adds the usual cloud tags to keypairs as seen in the integration test outputs.

WIP while I figure out how to handle keys that are shared with the cluster rather than created by kops- would we tag it with "shared" like with do VPCs and Subnets or do we not tag it at all?

Also need to double check how kops deletes keypairs during a `delete cluster` and make sure these tags don't affect that process.

ref: #9640

At this point we could probably drop the cluster name prefix from key names but that's probably not necessary.

### DIFF
--- a/cloudmock/aws/mockec2/keypairs.go
+++ b/cloudmock/aws/mockec2/keypairs.go
@@ -50,18 +50,25 @@ func (m *MockEC2) ImportKeyPair(request *ec2.ImportKeyPairInput) (*ec2.ImportKey
 		return nil, err
 	}
 
+	n := len(m.KeyPairs) + 1
+	id := fmt.Sprintf("key-%d", n)
+
 	kp := &ec2.KeyPairInfo{
 		KeyFingerprint: aws.String(fp),
 		KeyName:        request.KeyName,
+		KeyPairId:      aws.String(id),
 	}
 	if m.KeyPairs == nil {
 		m.KeyPairs = make(map[string]*ec2.KeyPairInfo)
 	}
-	m.KeyPairs[aws.StringValue(request.KeyName)] = kp
+	m.KeyPairs[id] = kp
 	response := &ec2.ImportKeyPairOutput{
 		KeyFingerprint: kp.KeyFingerprint,
 		KeyName:        kp.KeyName,
 	}
+
+	m.addTags(id, tagSpecificationsToTags(request.TagSpecifications, ec2.ResourceTypeKeyPair)...)
+
 	return response, nil
 }
 func (m *MockEC2) CreateKeyPairRequest(*ec2.CreateKeyPairInput) (*request.Request, *ec2.CreateKeyPairOutput) {
@@ -122,6 +129,7 @@ func (m *MockEC2) DescribeKeyPairs(request *ec2.DescribeKeyPairsInput) (*ec2.Des
 		}
 
 		copy := *keypair
+		copy.Tags = m.getTags(ec2.ResourceTypeKeyPair, *copy.KeyPairId)
 		keypairs = append(keypairs, &copy)
 	}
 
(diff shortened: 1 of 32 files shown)


cloudmock/aws/mockec2/keypairs.go
```diff
@@ -50,18 +50,25 @@ func (m *MockEC2) ImportKeyPair(request *ec2.ImportKeyPairInput) (*ec2.ImportKey
 		return nil, err
 	}
 
+	n := len(m.DhcpOptions) + 1
```
justinsb, line 53: Do you want len(m.KeyPairs) ?

cloudmock/aws/mockec2/keypairs.go
```diff
@@ -138,12 +146,17 @@ func (m *MockEC2) DeleteKeyPair(request *ec2.DeleteKeyPairInput) (*ec2.DeleteKey
 
 	klog.Infof("DeleteKeyPair: %v", request)
 
-	id := aws.StringValue(request.KeyName)
-	o := m.KeyPairs[id]
-	if o == nil {
-		return nil, fmt.Errorf("KeyPairs %q not found", id)
+	name := aws.StringValue(request.KeyName)
```
justinsb, line 149: Can we / should we now delete by id?

tests/integration/update_cluster/existing_iam/kubernetes.tf
```diff
@@ -318,6 +318,11 @@ resource "aws_internet_gateway" "existing-iam-example-com" {
 resource "aws_key_pair" "kubernetes-existing-iam-example-com-c4a6ed9aa889b9e2c39cd663eb9c7157" {
   key_name   = "kubernetes.existing-iam.example.com-c4:a6:ed:9a:a8:89:b9:e2:c3:9c:d6:63:eb:9c:71:57"
   public_key = file("${path.module}/data/aws_key_pair_kubernetes.existing-iam.example.com-c4a6ed9aa889b9e2c39cd663eb9c7157_public_key")
+  tags = {
+    "KubernetesCluster"                              = "existing-iam.example.com"
+    "Name"                                           = "existing-iam.example.com"
+    "kubernetes.io/cluster/existing-iam.example.com" = "owned"
```
justinsb, line 324: Yes, I think this should be shared, because we don't delete it (we couldn't safely delete it before); so it does behave more like "shared"

justinsb: Actually, scratch that - I'm wrong!  We _do_ delete the keypairs, which is why we have the cluster name in the prefix, so `owned` is correct.

justinsb:
Generally LGTM.  I was a little confused as to whether we delete keys or not; because/if we do, it should be 'owned', which is what you've done :+1: 

 > At this point we could probably drop the cluster name prefix from key names but that's probably not necessary.

Can we have duplicate SSH key names now?  I assume we can't, so I would imagine if we dropped the prefix that we might get collisions if someone used the same SSH key for two clusters in the same region?

rifelpet:
This is ready for review.

I had to make the SSHKey task aware of whether it is Shared or not. The intent is to not add any tags to keys that are specified in the ClusterSpec.sshKeyName. These tags aren't consumed by anything else (unlike shared VPC and Subnet tags) and may collide with tags from other clusters that also use the same sshKeyName, so I thought it was better to just not add the tags at all. In order to do that, we have to specify whether the key is shared or not. Previously we had the same task behavior regardless of whether a key was defined via sshKeyName or via a SSHPublicKey secret.

I tested the following scenarios:


1. create new cluster with old kops and local ssh key file specified
2. update cluster with new kops: correctly applies tags
3. update cluster with new kops: no changes reported

---

1. create new cluster with new kops and local ssh key file specified: correctly applies tags
2. update cluster with new kops: no changes reported

---

1. create new cluster with old kops and existing AWS sshKeyName specified in ClusterSpec: 
2. update cluster with new kops: no changes reported (key remains untagged)

---

1. create new cluster with new kops and existing AWS sshKeyName specified in ClusterSpec:
2. update cluster with new kops: no changes reported (key remains untagged)

pkg/model/sshkey.go
```diff
@@ -41,6 +41,8 @@ func (b *SSHKeyModelBuilder) Build(c *fi.ModelBuilderContext) error {
 	t := &awstasks.SSHKey{
 		Name:      s(name),
 		Lifecycle: b.Lifecycle,
+		Tags:      b.CloudTags(b.ClusterName(), false),
+		Shared:    b.Cluster.Spec.SSHKeyName != nil && *b.Cluster.Spec.SSHKeyName != "",
```
hakman, line 45: May be better:
```suggestion
		Shared:    fi.StringValue(b.Cluster.Spec.SSHKeyName) != "",
```
https://github.com/kubernetes/kops/blob/1c37a328e1a76563f3fb2a41b9353675653650b4/upup/pkg/fi/values.go#L25-L30

PR #9534, opened 2020-07-09
Use a stable key for signing service account tokens

### LINKED ISSUE #9531: Kops master branch: multiple masters fail to cluster together
When there are multiple master nodes, the API servers fail to cluster together and worker nodes fail to register.

The second api-server fails with:

```
W0708 19:38:50.391903       1 lease.go:224] Resetting endpoints for master service "kubernetes" to [172.20.40.133 172.20.70.113]
I0708 19:38:50.392785       1 controller.go:606] quota admission added evaluator for: endpoints
I0708 19:38:50.399710       1 controller.go:606] quota admission added evaluator for: endpointslices.discovery.k8s.io
I0708 19:38:56.175853       1 controller.go:606] quota admission added evaluator for: leases.coordination.k8s.io
I0708 19:39:42.672783       1 controller.go:606] quota admission added evaluator for: serviceaccounts
E0708 19:39:43.684332       1 authentication.go:53] Unable to authenticate the request due to an error: [invalid bearer token, square/go-jose: error in cryptographic primitive]
E0708 19:39:43.692574       1 authentication.go:53] Unable to authenticate the request due to an error: [invalid bearer token, square/go-jose: error in cryptographic primitive]
```

This appears to have been caused by #9354, which changed from using a single, ten-year TLS server certificate to per-node TLS server certificates with shorter lifetimes.

Initial investigation suggests that this is caused by kops configuring kube-controller-manager to reuse the api-server's TLS private key as the service account private key. Now that the nodes have different TLS private keys, service tokens created on one master node would not be recognized on the other master nodes.

johngmyers:
Service account tokens were signed by the private key of the api-server TLS certificate. When #9354 changed the latter to being issued out of nodeup, multiple masters would fail to verify each other's service account tokens. Single masters would work for a while, but would invalidate all extant service account tokens whenever the master was replaced.

This PR stops reusing the api-server TLS private key as the service account signing key, instead using a key generated by cloudup. In order to keep clusters created by kops 1.18 and earlier working, this key is named "master" so that such clusters will continue to use the (previous) api-server TLS private key as the service account signing key.

Fixes #9531

### DIFF
--- a/nodeup/pkg/model/kube_apiserver.go
+++ b/nodeup/pkg/model/kube_apiserver.go
@@ -287,6 +287,8 @@ func (b *KubeAPIServerBuilder) writeAuthenticationConfig(c *fi.ModelBuilderConte
 // buildPod is responsible for generating the kube-apiserver pod and thus manifest file
 func (b *KubeAPIServerBuilder) buildPod() (*v1.Pod, error) {
 	kubeAPIServer := b.Cluster.Spec.KubeAPIServer
+	// TODO pass the public key instead. We would first need to segregate the secrets better.
+	kubeAPIServer.ServiceAccountKeyFile = append(kubeAPIServer.ServiceAccountKeyFile, filepath.Join(b.PathSrvKubernetes(), "service-account.key"))
 	kubeAPIServer.ClientCAFile = filepath.Join(b.PathSrvKubernetes(), "ca.crt")
 	kubeAPIServer.TLSCertFile = filepath.Join(b.PathSrvKubernetes(), "server.crt")
 	kubeAPIServer.TLSPrivateKeyFile = filepath.Join(b.PathSrvKubernetes(), "server.key")
--- a/nodeup/pkg/model/secrets.go
+++ b/nodeup/pkg/model/secrets.go
@@ -128,6 +128,10 @@ func (b *SecretBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	if err := b.BuildPrivateKeyTask(c, "master", "service-account.key"); err != nil {
+		return err
+	}
+
 	// Support for basic auth was deprecated 1.16 and removed in 1.19
 	// https://github.com/kubernetes/kubernetes/pull/89069
 	if b.IsKubernetesLT("1.19") && b.SecretStore != nil {
--- a/pkg/model/pki.go
+++ b/pkg/model/pki.go
@@ -171,6 +171,18 @@ func (b *PKIModelBuilder) Build(c *fi.ModelBuilderContext) error {
 		c.AddTask(aggregatorCA)
 	}
 
+	{
+		serviceAccount := &fitasks.Keypair{
+			// We only need the private key, but it's easier to create a certificate as well.
+			// The strange name is because Kops prior to 1.19 used the api-server TLS key for this.
+			Name:      fi.String("master"),
+			Lifecycle: b.Lifecycle,
+			Subject:   "cn=service-account",
+			Type:      "ca",
+		}
+		c.AddTask(serviceAccount)
+	}
+
 	// @TODO this is VERY presumptuous, i'm going on the basis we can make it configurable in the future.
 	// But I'm conscious not to do too much work on bootstrap tokens as it might overlay further down the
 	// line with the machines api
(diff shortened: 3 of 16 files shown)


johngmyers:
@olemarkus if there's a question, I can't see it.

olemarkus:
It was a review comment that apparently didn't get submitted. 
Kubernetes lists a list of recommended paths here: https://kubernetes.io/docs/setup/best-practices/certificates/#certificate-paths
Was just wondering if we would use those. I.e sa.key/pub instead of service-account.key

johngmyers:
We're not kubeadm and there are issues with some of those locations. One issue (that we should fix) is that putting them all in the same directory and then having the containers mount that directory means that the containers have access to secrets they don't need.

olemarkus:
Just a thought anyway.
/lgtm

nodeup/pkg/model/kube_apiserver.go
```diff
@@ -287,6 +287,8 @@ func (b *KubeAPIServerBuilder) writeAuthenticationConfig(c *fi.ModelBuilderConte
 // buildPod is responsible for generating the kube-apiserver pod and thus manifest file
 func (b *KubeAPIServerBuilder) buildPod() (*v1.Pod, error) {
 	kubeAPIServer := b.Cluster.Spec.KubeAPIServer
+	// TODO pass the public key instead. We would first need to segregate the secrets better.
```
justinsb, line 290: Yeah, this would be great.  I think we've started doing things under `/etc/kubernetes`; the two suggestions I've heard/seen are `/etc/kubernetes/secrets/<component>` and `/etc/kubernetes/pki/<component>`  Or maybe `/etc/kubernetes/<component>/secrets` :-)  Regardless of the naming nit-pick, I think splitting by component would be good, particularly in conjunction with the idea you're championing of having things run under dedicated non-root unix accounts.

nodeup/pkg/model/secrets.go
```diff
@@ -128,6 +128,10 @@ func (b *SecretBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	if err := b.BuildPrivateKeyTask(c, "master", "service-account.key"); err != nil {
```
justinsb, line 131: I was thinking about strategies if we ever did want to rename this from master -> service-account (or similar) without breaking things.  I think it's possible, but it's not entirely trivial.   I suggest we live with it until the OIDC signing thing evolves (e.g. rotation of these signing keys is coming at some stage, I think)

pkg/model/pki.go
```diff
@@ -171,6 +171,18 @@ func (b *PKIModelBuilder) Build(c *fi.ModelBuilderContext) error {
 		c.AddTask(aggregatorCA)
 	}
 
+	{
+		serviceAccount := &fitasks.Keypair{
+			// We only need the private key, but it's easier to create a certificate as well.
+			// The strange name is because Kops prior to 1.19 used the api-server TLS key for this.
```
justinsb, line 177: :+1: for comment!

justinsb:
Thanks @johngmyers - and I agree with the naming strategy

/approve
/lgtm

nodeup/pkg/model/kube_apiserver.go
johngmyers: I think we want `/etc/kubernetes/<component>/` with nothing mounting `/etc/kubernetes/`

nodeup/pkg/model/secrets.go
johngmyers: I was just thinking of renaming this and adding a `kops rotate secret service-account` command. There would be three secrets: `service-account`, `service-account-previous`, and `service-account-next`.

We could do this at the VFS layer, having a read of the `service-account` keypair falling back to `master` and returning it with the `legacyFormat` bit set.

adammw:
just a bump that this (or the change to server.key rather) should have been explicitly called out in the 1.19 changelog that user action was required if previously setting `serviceAccountKeyFile` or `serviceAccountSigningKeyFile` as recommended in the docs (updated in #10712).

agilgur5:
For reference, https://github.com/kubernetes/kops/pull/11390 changed this `master` key to be named `service-account`, which is why one doesn't see `master.key` in any config anymore.

Also +++ to the above comment, would've caught that much earlier if we saw that

Issue #9526, opened 2020-07-08
pull-kubernetes-e2e-kops-aws failed for 1.19


jqmichael:
Hey kops experts,

Could anyone help me understand why the presubmit job below consistantly failed for 1.19?

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/91513/pull-kubernetes-e2e-kops-aws/1280849041231450112/

The build job output seemed to suggest kube-apiserver isn't up running for some reason? If so, any hints where I could go find kube-apiserver/kubelet logs?
```
W0708 13:12:49.909] cannot determine hash for "https://storage.googleapis.com/kubernetes-release-pull/ci/pull-kubernetes-e2e-kops-aws/v1.19.0-beta.2.793+01d6753b5c7271/bin/linux/arm64/kubelet" (have you specified a valid file location?)

W0708 13:12:49.939] 2020/07/08 13:12:49 process.go:155: Step '/workspace/kops create cluster --name e2e-bd9ab6df6f-dba53.test-cncf-aws.k8s.io --ssh-public-key /workspace/.ssh/kube_aws_rsa.pub --node-count 4 --node-volume-size 48 --master-volume-size 48 --master-count 1 --zones sa-east-1c --master-size c5.large --kubernetes-version https://storage.googleapis.com/kubernetes-release-pull/ci/pull-kubernetes-e2e-kops-aws/v1.19.0-beta.2.793+01d6753b5c7271 --admin-access 35.226.123.140/32 --cloud aws --override cluster.spec.nodePortAccess=0.0.0.0/0 --yes' finished in 9.795259329s

W0708 13:12:49.945] 2020/07/08 13:12:49 process.go:153: Running: kubectl -n kube-system get pods -ojson -l k8s-app=kops-controller

W0708 13:12:50.739] The connection to the server localhost:8080 was refused - did you specify the right host or port?

W0708 13:12:50.759] 2020/07/08 13:12:50 kubernetes.go:117: kubectl get pods failed: error during kubectl -n kube-system get pods -ojson -l k8s-app=kops-controller: exit status 1

W0708 13:12:50.802] W0708 13:12:50.802076    7513 vfs_castore.go:604] CA private key was not found
W0708 13:12:50.803] 
W0708 13:12:50.803] cannot find CA certificate
```

Thanks

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

**2. What Kubernetes version are you running? `kubectl version` will print the
[...]

PR #9535, opened 2020-07-09
Force single arch support via env var

hakman:
Force AMD64 or ARM64 only support via env var to address the issues with Kubernetes PR binaries being AMD64 only.

Fixes: #9526

### DIFF
--- a/util/pkg/architectures/architectures.go
+++ b/util/pkg/architectures/architectures.go
@@ -18,6 +18,7 @@ package architectures
 
 import (
 	"fmt"
+	"os"
 	"runtime"
 
 	"k8s.io/klog"
@@ -42,6 +43,12 @@ func FindArchitecture() (Architecture, error) {
 }
 
 func GetSupprted() []Architecture {
+	// Force support for AMD64 only if env var is set
+	arch := os.Getenv("KOPS_ARCH_AMD64")
+	if arch != "" {
+		return []Architecture{ArchitectureAmd64}
+	}
+
 	return []Architecture{
 		ArchitectureAmd64,
 		ArchitectureArm64,
--- a/util/pkg/architectures/architectures.go
+++ b/util/pkg/architectures/architectures.go
@@ -43,10 +43,14 @@ func FindArchitecture() (Architecture, error) {
 }
 
 func GetSupported() []Architecture {
-	// Force support for AMD64 only if env var is set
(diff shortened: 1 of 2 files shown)


olemarkus:
Feels like this should be a list of archs to support, but then I don't really think we'll support more archs than these two anytime soon.

/lgtm

olemarkus:
Thanks :)
/lgtm

util/pkg/architectures/architectures.go
```diff
@@ -41,7 +42,13 @@ func FindArchitecture() (Architecture, error) {
 	}
 }
 
-func GetSupprted() []Architecture {
+func GetSupported() []Architecture {
+	// Force support for AMD64 only if env var is set
+	arch := os.Getenv("KOPS_ARCH_AMD64")
+	if arch != "" {
+		return []Architecture{ArchitectureAmd64}
+	}
```
jqmichael, line 50: Nice! Appreciate the fix!

Any idea whether we need to update the documentation?

hakman: Np. This is for internal use at the moment, so I don't see much value in adding docs for it.

rifelpet: I wonder if we could setup this env var so its a bit more flexible in the future if we do ever support additional architectures? Perhaps move the architecture into the env var value:

`KOPS_ARCH=amd64`
`KOPS_ARCH=amd64,arm64`

hakman: I think in the future we may want to make it a spec option or even a create cluster flag.
At the moment, there is very little value to allow various values.

This function really should be used for anything cpu arch related in case there will be something like that, and it touches cni plugins and maybe other types of assets.

If we ever want to, we could add also `KOPS_ARCH_ARM64` if we only want ARM64 support, but I don't have a better idea either.

jqmichael: Another dump question. :)

I assume it was too late to rely on `FindArchitecture()` to provide the correct arch at runtime probably because we want to download binary/images upfront to save time?

hakman: You would have to know what arch is on each node. That is not possible or unreliable when the cluster is created.
We only retrieve the hashes, not the binaries. Each node decides on its own what arch it needs.

PR #9536, opened 2020-07-09
Validate cloud labels

olemarkus:
Since moving to LaunchTemplates, cloudLabels cannot have empty values anymore. If one tries to, kops will just loop endlessly with AWS errors when trying to create the launch template.

/cc @rifelpet

### DIFF
--- a/pkg/apis/kops/validation/instancegroup.go
+++ b/pkg/apis/kops/validation/instancegroup.go
@@ -119,6 +119,8 @@ func ValidateInstanceGroup(g *kops.InstanceGroup, cloud fi.Cloud) field.ErrorLis
 		allErrs = append(allErrs, awsValidateInstanceGroup(g, cloud.(awsup.AWSCloud))...)
 	}
 
+	allErrs = append(allErrs, validateCloudLabels(g.Spec.CloudLabels, field.NewPath("spec", "cloudLabels"))...)
+
 	return allErrs
 }
 
@@ -240,6 +242,19 @@ func validateInstanceProfile(v *kops.IAMProfileSpec, fldPath *field.Path) field.
 		if err != nil || !strings.HasPrefix(parsedARN.Resource, "instance-profile") {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("profile"), instanceProfileARN,
 				"Instance Group IAM Instance Profile must be a valid aws arn such as arn:aws:iam::123456789012:instance-profile/KopsExampleRole"))
+		}
+	}
+	return allErrs
+}
+
+func validateCloudLabels(labels map[string]string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for tag, val := range labels {
+		if tag == "" {
+			allErrs = append(allErrs, field.Invalid(fldPath, tag, "cloud labels cannot be cannot be empty strings"))
+		}
+		if val == "" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child(tag), val, "cloud labels cannot have empty values"))
 		}
 	}
 	return allErrs
--- a/pkg/apis/kops/validation/instancegroup.go
+++ b/pkg/apis/kops/validation/instancegroup.go
@@ -251,7 +251,7 @@ func validateCloudLabels(labels map[string]string, fldPath *field.Path) field.Er
 	allErrs := field.ErrorList{}
 	for tag, val := range labels {
 		if tag == "" {
-			allErrs = append(allErrs, field.Invalid(fldPath, tag, "cloud labels cannot be cannot be empty strings"))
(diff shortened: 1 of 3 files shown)


pkg/apis/kops/validation/instancegroup_test.go
```diff
@@ -221,3 +221,19 @@ func TestValidBootDevice(t *testing.T) {
 		testErrors(t, g.volumeType, errs, g.expected)
 	}
 }
+
+func TestValidateCloudLabels(t *testing.T) {
+	labels := map[string]string{
+		"nonEmptyLabel": "value",
+		"emptyValue":    "",
+		"":              "emptyTag",
+	}
```
hakman, line 230: Please ad a test for empty key.

olemarkus: Doesn't this test both empty key (tag) and empty value?

hakman: You are right, didn't notice.

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -244,3 +246,16 @@ func validateInstanceProfile(v *kops.IAMProfileSpec, fldPath *field.Path) field.
 	}
 	return allErrs
 }
+
+func validateCloudLabels(labels map[string]string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for tag, val := range labels {
+		if tag == "" {
+			allErrs = append(allErrs, field.Invalid(fldPath, tag, "cloud labels cannot be cannot be empty strings"))
```
rifelpet, line 254:
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath, tag, "cloud labels cannot be empty strings"))
```

rifelpet:
This issue: https://github.com/terraform-providers/terraform-provider-aws/issues/11164 makes me wonder if it is specifically "tag on create" that doesn't allow empty values, and if `ec2.CreateTags` with a launch template resourceID would succeed :( I haven't tested it yet but I'll try to today unless anyone else has already.

If thats the case I can open a support ticket to get more info to confirm this is expected behavior and maybe we can add a workaround in Kops (only pass nonempty-value tags to `ec2.CreateLaunchTemplate()` and let `AddAWSTags()`'s `ec2.CreateTags()` add the remaining tags)

rifelpet:
hm so this works for me:

(replace sg-12345 with a valid one)
`
aws ec2 create-launch-template --launch-template-name TemplateForAutoScaling --version-description AutoScalingVersion1 --launch-template-data '{"NetworkInterfaces":[{"DeviceIndex":0,"AssociatePublicIpAddress":true,"Groups":["sg-12345"],"DeleteOnTermination":true}],"ImageId":"ami-b42209de","InstanceType":"m4.large","TagSpecifications":[{"ResourceType":"instance","Tags":[{"Key":"foo-instance","Value":""}]},{"ResourceType":"volume","Tags":[{"Key":"foo-volume","Value":""}]},{"ResourceType":"launch-template","Tags":[{"Key":"foo-lt","Value":""]],"BlockDeviceMappings":[{"DeviceName":"/dev/sda1","Ebs":{"VolumeSize":100}}]}' --debug
`

this creates a launch template with value-less tags assigned to the instance, volume, and launch template itself.

rifelpet:
Ah, i was able to repro the issue, turns out i was specifying the tag in the wrong place.

`aws ec2 create-launch-template               --launch-template-name TemplateForAutoScaling3               --version-description AutoScalingVersion1               --launch-template-data '{"NetworkInterfaces":[{"DeviceIndex":0,"AssociatePublicIpAddress":true,"Groups":["sg-12345"],"DeleteOnTermination":true}],"ImageId":"ami-b42209de","InstanceType":"m4.large","BlockDeviceMappings":[{"DeviceName":"/dev/sda1","Ebs":{"VolumeSize":100}}]}' --region us-east-1 --debug --tag-specifications 'ResourceType=launch-template,Tags=[{Key=lt-tag,Value=""}]'`

`
An error occurred (InvalidParameterValue) when calling the CreateLaunchTemplate operation: Tag keys and tag values cannot be empty. Ensure that each tag has a key and value pair and try again.
`

I'll open a support case and find out if that's intended.

hakman:
I didn't notice this before the change to "tag on create.
It is a PITA, as it will affect all people using cluster autoscaler.

rifelpet:
I contacted AWS support and they're raising an issue with an internal team. No ETA on more info or a fix, so we may want to consider reverting the tag-on-create for now.

hakman:
> I contacted AWS support and they're raising an issue with an internal team. No ETA on more info or a fix, so we may want to consider reverting the tag-on-create for now.

I think that would be a good idea for now.

olemarkus:
I suppose this one should just be closed then?

hakman:
AWS already deployed the fix in `us-east-1` and will continue the rollout to the rest of the regions this week or the next one.

PR #9537, opened 2020-07-09
Promote alphas to stable

olemarkus:
Been in alpha for 12 days. Also worth doing given  https://github.com/kubernetes/kubernetes/issues/92315


hakman:
@olemarkus Could you also update the Ubuntu images?

olemarkus:
@hakman alpha didn't bump image either. Don't we want to run the new image through alpha first?

hakman:
In current situation, no. We did not release 1.18 yet so should be ok.
From now on, alphas bump should include image bumps also.

olemarkus:
I'll do another PR to bump the ubuntu image in both channels then.
/retest

olemarkus:
Or on second thoughs. I'll just do another commit.

PR #9539, opened 2020-07-09
Add lyft hash environment variable

hintofbasil:
When setting the `LYFT_VPC_DOWNLOAD_URL` environment variable kops uses an empty hash in the user data.  This causes new nodes to fail to launch as the hash is unable to be verified.

Adding a new environment variable `LYFT_VPC_DOWNLOAD_HASH` allows a user to set the hash manually.


gjtempleton:
I think it's probably worth adding this [to the docs](https://github.com/kubernetes/kops/blob/master/docs/networking/lyft-vpc.md) at the same time.

There's currently no mention of the existing `LYFT_VPC_DOWNLOAD_URL` env var, which may be partly why it seems this code path's not actually been hit before.

hintofbasil:
Thanks @gjtempleton, I've added a small section there.

PR #9552, opened 2020-07-12
Use stable names for GH workflow jobs

hakman:
We will need to pass the job names to Tide to make them blocking, so stable names would be required.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 ---
-name: build
+name: CI
 
 'on':
   - push
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,27 +10,14 @@ env:
   GOPATH: ${{ github.workspace }}/go
 
 jobs:
-  build:
-    strategy:
-      # Note, the following build matrix generates jobs that are blocking via prow.
-      # When changing the versions or images below, please update prow config too:
-      # https://github.com/kubernetes/test-infra/blob/566fea31f63670f583dde93b55fcb0a13bdc2562/config/prow/config.yaml
-      matrix:
-        os: [ubuntu-18.04, macos-10.15]
-        go: [1.13, 1.14]
-        exclude:
-          - os: macos-10.15
-            go: 1.14
-      fail-fast: true
-
-    runs-on: ${{ matrix.os }}
-
+  build-linux-amd64:
+    runs-on: ubuntu-20.04
     steps:
       - name: Set up go
         uses: actions/setup-go@v1
         with:
-          go-version: ${{ matrix.go }}
-        id: go
+          go-version: 1.14
+          stable: true
 
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -41,21 +28,32 @@ jobs:
...
+
+    - uses: actions/checkout@v2
+      with:
+        path: ${{ env.GOPATH }}/src/k8s.io/kops
 
-    runs-on: ${{ matrix.os }}
+    - name: make nodeup examples test
+      working-directory: ${{ env.GOPATH }}/src/k8s.io/kops
+      run: |
+        make nodeup examples test
```
johngmyers, line 47: We can probably skip making nodeup on MacOS.

johngmyers:
This removes Go 1.13. Is that because it's currently redundant with the Prow jobs and we expect to drop it as a requirement when we move Prow to 1.14?

hakman:
> This removes Go 1.13. Is that because it's currently redundant with the Prow jobs and we expect to drop it as a requirement when we move Prow to 1.14?

Yes. Also, when Go 1.15 is released we can add an optional build-next job.

.github/workflows/main.yml
hakman: Good point. Done.

hakman: Any idea why we only build `nodeup` and not `all` ?

justinsb: I suspect this is just historical about the way these tests were introduced (they're still fairly new); I suspect we were just trying to replace tests in travis.

From that point of view I really like splitting out the tests explicitly.  What we really want on macos is coverage of the kops CLI (I don't think we want nodeup tests to run on macos; there's a case for windows in the future but I don't know of anyone trying to get kubelet to run on macos)

For linux, I think trying everything seems reasonable!  We should probably be trending towards running tests in bazel, as more and more of our release pipeline is using bazel (I have some exciting news on this front for our next office hours :-) )

I think it's fine to rearrange what these do a bit in a separate PR.

.github/workflows/main.yml
```diff
@@ -41,21 +28,32 @@ jobs:
...
+      with:
+        go-version: 1.14
+        stable: true
+
+    - uses: actions/checkout@v2
+      with:
+        path: ${{ env.GOPATH }}/src/k8s.io/kops
 
-    runs-on: ${{ matrix.os }}
+    - name: make nodeup examples test
```
justinsb, line 44: Can we remove nodeup here also?  Otherwise it's going to confuse me every time :-)

hakman: Good point. Done.

.github/workflows/main.yml
hakman: Looking forward to that :).

rifelpet:
this looks good to me. I agree we can have optional jobs for the next or previous go versions. I know i've asked this before, but to confirm here, are we safe to merge this with prow configured as-is and will it affect other open PRs? and what will the rollout to enforce these jobs look like exactly?

hakman:
@rifelpet open PRs would behave like current release branches, expecting these tests, but will not block for now.
Having stable names, means that we can backport these to release branches also and satisfy Prow and Tide requirements.
Once we are happy with how things work, we can see what should be done to make them blocking and remove the equivalent TravisCI jobs.

justinsb:
/approve
/lgtm

Github testing isn't entirely happy, so this is also a nice test of whether we're going to break other PRs (I think!)

PR #9554, opened 2020-07-12
Openstack fixes

olemarkus:
The refactoring done lately have created cyclic dependencies in cloudup tasks when provisioning clusters that use floating or fixed ips for accessing the master API.
This PR breaks this cycle by provisioning floating IPs before the instances.

Using fixed IPs for the API is probably not possible to fix since they are created together with the instance, so one will always have the bootstrapscript -> instance -> bootstrapscript cycle. I think that should be removed in a later PR.

/cc @zetaab

### DIFF
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -174,7 +174,7 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 					instanceTask.FloatingIP = t
 				}
 			default:
-				if !b.UsesSSHBastion() {
+				if b.Cluster.Spec.Topology.Nodes == "public" {
 					t := &openstacktasks.FloatingIP{
 						Name:      fi.String(fmt.Sprintf("%s-%s", "fip", *instanceTask.Name)),
 						Lifecycle: b.Lifecycle,
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -174,7 +174,7 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 					instanceTask.FloatingIP = t
 				}
 			default:
-				if b.Cluster.Spec.Topology.Nodes == "public" {
+				if b.Cluster.Spec.Topology == nil || b.Cluster.Spec.Topology.Nodes == "public" {
 					t := &openstacktasks.FloatingIP{
 						Name:      fi.String(fmt.Sprintf("%s-%s", "fip", *instanceTask.Name)),
 						Lifecycle: b.Lifecycle,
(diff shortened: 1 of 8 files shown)


pkg/model/openstackmodel/servergroup.go
```diff
@@ -159,28 +159,28 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
...
-						Server:    instanceTask,
 						Lifecycle: b.Lifecycle,
 					}
 					c.AddTask(t)
 					b.associateFIPToKeypair(t)
+					instanceTask.FloatingIP = t
 				}
 			default:
-				if !b.UsesSSHBastion() {
+				if b.Cluster.Spec.Topology.Nodes == "public" {
```
zetaab, line 177: is this change backwards compatible to old clusters?

olemarkus: The reason I added this was that I got floating IPs on all nodes otherwise. Even with private topology.
There is a lot of validation missing for configuring an openstack cluster, where e.g private topology on masters + floating ips on masters should be conflicting. In addition there is `ig.spec.associatePublicIP` that comes into play too, where floating IPs should be added regardless.

Happy to amend this any other way.

olemarkus:
Thanks for the early feedback. Yeah, I am hoping to get a cluster with Octavia to test this as well. Multi-master clusters were broken until just now, so didn't have chance to test that, but single master works just fine.

johngmyers:
If the instance can figure out its fixed IP, then nodeup can add that IP into the cert.

zetaab:
fixed ip is available in OpenStack metadata API in each Instance (and that also should match to interface ip address in that instance)

olemarkus:
That should work.
The HasAddress implementation can also be moved from instance then.

I think that is a largely independent change from this one, so I can see if I can do a parallel PR on that.

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -45,6 +46,7 @@ type Instance struct {
 	Metadata         map[string]string
 	AvailabilityZone *string
 	SecurityGroups   []string
+	FloatingIP       *FloatingIP
```
johngmyers, line 49: `Find()` needs to fill in this field.

olemarkus: I agree that at some point Find should fill in this. But right now, there is a `CheckChanges` that only returns true when the instance doesn't exist. So an instance will basically never be updated once created. That is, if I understand fi.task correctly.

One can easily run into an issue where floating IP and instance is created, but association fails for some reason, and it would be nice to have kops detect and amend that. But that's the case for a number of things here. Had a plan to clean that part up later as well.

johngmyers:
Looks okay, but I don't know OpenStack.

zetaab:
I will test this when I get internet to my apartment (next week)

zetaab:
I cannot get this PR working

https://gist.github.com/zetaab/e0990214a939830a36c16c022fcb600b

zetaab:
@olemarkus thanks a lot! This will fix all other cases than LB use-case.

Anyway lets make separate PR from that.

/approve
/lgtm

PR #9555, opened 2020-07-12
Adding feature stability table to docs and including one example for encryption support in cilium

### LINKED ISSUE #9151: [Documentation Clarity] Please clarify when specific fields are added to the configuration schemas
 this information.**

1.16.2

 a `kops` flag.**

1.16.9

AWS

N/A

Silent dropping of configuration. See below

At least an error message

See below

N/A

We're in the middle of a cluster upgrade to Kops 1.16.2 and nearly suffered an outage because we tried to add fields like `InstanceGroup.spec.sysctlParameters` and `Cluster.spec.networking.amazonvpc.envVars` that are documented in the `master` branch, not accepted in `1.16.2`, and are silently dropped by `1.16.2` instead of throwing an error about unknown configuration.

I understand that `master` reflects the newest state of development rather than the version of the documentation that is tied to a specific release, and that version-controlling the documentation in this way is beneficial precisely because it does allow me to see what the precise state of the documentation is for the version that I'm installing. Nevertheless, it is *extremely easy* to forget to set the tag on GitHub when viewing the documentation and presume that I am looking at the "correct" version of the documentation for what is the latest stable / certified / production-ready version of Kops. In many projects, the latest recommended version for installation is generally sufficiently close to if not identical to `master` that the documentation drift does not present a problem in practice; in Kops, this is not the case, particularly as the manifest version is set to `api/v1alpha2` and not changed when fields are added or changed.

My recommendation is to adopt a similar standard as that found in the main Kubernetes user-facing documentation, which, even for the latest versions, clearly indicates in which version various features were made available. This ensures, for instance, that Kubernetes 1.16 users do not attempt to work with [EndpointSlices](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/). As such, as a user, I would be able to see from reading the docs in `master` that various features are not yet available for the version of Kops that I am trying to install.

MoShitrit:
Fixes #9151 
To get started, I also added this table to the `cilium` docs for a feature I previously added. This can provide an example for additional features to which we would like to add this table.
Example output of the table when rendered:

image

/cc @rifelpet @olemarkus

### DIFF
--- a/docs/extra.css
+++ b/docs/extra.css
@@ -11,3 +11,7 @@
 .hidden {
 	display: none;
 }
+
+.kops_feature_table .md-typeset__table table{
+	max-width: fit-content;
+}
--- a/docs/networking/cilium.md
+++ b/docs/networking/cilium.md
@@ -121,6 +121,8 @@ Note that since Cilium Operator is the entity that interacts with the EC2 API to
 Also note that this feature has only been tested on the default kops AMIs.
 
 #### Enabling Encryption in Cilium
+{{ kops_feature_table(kops_added_default='1.19', k8s_min='1.17') }}
+
 As of Kops 1.19, it is possible to enable encryption for Cilium agent.
 In order to enable encryption, you must first generate the pre-shared key using this command:
 ```bash
--- a/hack/__init__.py
+++ b/hack/__init__.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def main():
+    pass
+
+
+if __name__ == "__main__":
+    main()
--- a/hack/mkdocs_macros/__init__.py
+++ b/hack/mkdocs_macros/__init__.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from .feature_stability_table import define_env
+
+
+def main():
+    pass
+
+
+if __name__ == "__main__":
+    main()
--- a/hack/mkdocs_macros/feature_stability_table.py
+++ b/hack/mkdocs_macros/feature_stability_table.py
@@ -55,7 +55,9 @@ def define_env(env):
             separators,
             values
         ]
-        return '\n'.join(table)
+        table = '\n'.join(table)
+        table = f'<div class="kops_feature_table">{table}</div>'
+        return table
 
 
 def main():
--- a/hack/mkdocs_macros/feature_stability_table.py
+++ b/hack/mkdocs_macros/feature_stability_table.py
@@ -14,8 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# -*- coding: utf-8 -*-
-
 
 def define_env(env):
     """Hook function"""
@@ -64,5 +62,5 @@ def main():
     pass
 
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     main()
--- a/hack/update-header.sh
+++ b/hack/update-header.sh
@@ -17,7 +17,7 @@
 . "$(dirname "${BASH_SOURCE[0]}")/common.sh"
 
 BAD_HEADERS=$((${KOPS_ROOT}/hack/verify-boilerplate.sh || true) | awk '{ print $7}')
-FORMATS="sh go Makefile Dockerfile"
+FORMATS="sh go Makefile Dockerfile py"
 
 YEAR=`date -u +%Y`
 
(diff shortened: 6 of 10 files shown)


MoShitrit:
I might be missing something, but following the failure output of `hack/verify-boilerplate.sh` I ran `hack/update-header.sh` but nothing happens, it doesn't seem to change anything in `mkdocs_macro.py`.
If I read that script correctly, I _think_ it looks for the copyright block, so I tried adding that manually and it still didn't work.

What am I missing?
```
mosheshitrit:kops mosheshitrit$ hack/verify-boilerplate.sh
FAIL: Boilerplate header is wrong for: /Users/mosheshitrit/src/kubernetes/kops/src/k8s.io/kops/mkdocs_macros.py
FAIL: Please execute ./hack/update-header.sh
mosheshitrit:kops mosheshitrit$ hack/update-header.sh
mosheshitrit:kops mosheshitrit$ hack/verify-boilerplate.sh
FAIL: Boilerplate header is wrong for: /Users/mosheshitrit/src/kubernetes/kops/src/k8s.io/kops/mkdocs_macros.py
FAIL: Please execute ./hack/update-header.sh
```

MoShitrit:
OK found it.
Apparently `hack/update_header.sh` doesn't have a config for `.py` files even though it _does_ exist under `hack/boilerplate` folder. Adding a commit to address that.

olemarkus:
/lgtm

Why did the netlify deploy cancel though?

hakman:
> Why did the netlify deploy cancel though?

@olemarkus Most likely because last commit was on a something not tracked by config file:
https://github.com/kubernetes/kops/blob/8c3b4e4f430914b8a95bfa2f571dc273979f6d1d/netlify.toml#L5

@MoShitrit any chance `mkdocs_macros.py` can be moved to the `scripts` dir?

MoShitrit:
> > Why did the netlify deploy cancel though?
> 
> @olemarkus Most likely because last commit was on a something not tracked by config file:
> 
> https://github.com/kubernetes/kops/blob/8c3b4e4f430914b8a95bfa2f571dc273979f6d1d/netlify.toml#L5
> 
> @MoShitrit any chance `mkdocs_macros.py` can be moved to the `scripts` dir?

@hakman Yeah, sure. I think it'll involve some additional config to the plugin but I can do that.
By the `scripts` dir- do you mean I should create one and put it there? Or is it the `hack` dir? 
I tried to find a dir named `scripts` but couldn't find any.

hakman:
@MoShitrit you are right, meant `hack`. Thanks!

MoShitrit:
@hakman 
Okay, so after re-reading their [docs about the location of the module](https://mkdocs-macros-plugin.readthedocs.io/en/latest/python/#location-of-the-module), I think it might work out better to have a dedicated python package under the root directory, if we don't want to have the file located in the same path as `mkdocs.yml`. So I'm thinking probably creating a python package called `mkdocs_macros/` and moving the `.py` file to this folder. Then we'll have to change the plugin config to point to that directory and that should do it.
WDYT? Is it worth the effort? Or should we just keep it as it is now?
If we _do_ want to use the `hack` dir, we'll have to convert it to a python package (by adding a `__init__.py` file as well there. I'm okay with taking that approach as well, but I'm not sure if we _should_ do it, since most of the files there aren't python and aren't really used by `mkdocs` at all.

hakman:
Thanks for checking @MoShitrit . I don't think it's worth the extra work, but let's see if @rifelpet and @mikesplain have some feedback also.

rifelpet:
I think I'd lean towards moving mkdocs_macros.py into the hack directory and add a `hack/__init__.py` that way we avoid yet another file in the repo root. I think if you squash your commits, the netlify job will not be cancelled and we'll be able to check it out in the browser.

MoShitrit:
Created a new package under `hack` called `mkdocs_macros` and updated `mkdocs.yml` to point to the new package as the module for `Mkdocs-macros` plugin.
I see that the netlify deploy preview worked this time, so we can see an example of the table [here](https://deploy-preview-9555--kubernetes-kops.netlify.app/networking/cilium/#enabling-encryption-in-cilium)

mikesplain:
This looks great, thanks so much for your work on this @MoShitrit!

My only comment is more of a nit, looks like the table goes full with rather than ending at the second column.  Any way we can clean this up?

Screen Shot 2020-07-14 at 9 20 46 AM

Other than that, I think this is a great start!

MoShitrit:
Thanks, @mikesplain !
So I tried to look into that table alignment/padding thing, but I think it's more of how Mkdocs renders every custom block (tables, code etc.). I have a strong feeling this comes from the `Mkdocs-material` theme we're using.
I confirmed that by finding another table in our docs which doesn't use the `macros` plugin at all, and it looks exactly the same - https://kops.sigs.k8s.io/welcome/releases/#compatibility-matrix
I tried to do some html styling to change that, but it kinda broke it. 😖 
I tried to look a little bit a the [material theme docs](https://squidfunk.github.io/mkdocs-material/) but couldn't find a straightforward answer on how to change that (we can maybe apply some external css, but I feel like it could be a little bit out of scope for this PR and it could take me good bit of time to figure this out, lol)
WDYT?

mikesplain:
Hmm gotcha, you should be able to change `max-width: 100%;` to `max-width: fit-content;` on the table object, which worked for me at least.  You would likely need to wrap the table in a div with a new class like `class=kops_feature_table`, then add an override css to [extra.css](https://github.com/kubernetes/kops/blob/master/docs/extra.css) like 

```
.kops_feature_table table {
    max-width: fit-content;
}``` 

But to your point, this is easily fixed in a separate PR if someone spends the time on it. I'm fine without it for now. Thanks for the hard work on this, it's great!

MoShitrit:
> Hmm gotcha, you should be able to change `max-width: 100%;` to `max-width: fit-content;` on the table object, which worked for me at least. You would likely need to wrap the table in a div with a new class like `class=kops_feature_table`, then add an override css to [extra.css](https://github.com/kubernetes/kops/blob/master/docs/extra.css) like
> 
> ```
> .kops_feature_table table {
>     max-width: fit-content;
> }
> ```
> 
> But to your point, this is easily fixed in a separate PR if someone spends the time on it. I'm fine without it for now. Thanks for the hard work on this, it's great!

Yas!!! Got it now. I had to do it with a subclass, but your pointer took me to where I needed 😄 

image
 
Pushing up the commit now so you can see the code change.

Thanks for that, @mikesplain ! 🙇‍♂️

MoShitrit:
Updated example with the fixed table width: https://deploy-preview-9555--kubernetes-kops.netlify.app/networking/cilium/#enabling-encryption-in-cilium

MoShitrit:
Looking to get a 👍 whenever anyone has a free cycle

MoShitrit:
Looking to get a 👍 whenever anyone has a free cycle
@mikesplain @rifelpet

hakman:
@MoShitrit could you also add the `hack/` dir to this list to trigger docs updates on modules changes?
https://github.com/kubernetes/kops/blob/8c3b4e4f430914b8a95bfa2f571dc273979f6d1d/netlify.toml#L5

MoShitrit:
@hakman done! 🚀 😄  https://github.com/kubernetes/kops/pull/9555/commits/d43c9b5169aab580feef0456c87134343cf9141b

rifelpet:
looks great, thanks again! hopefully we can get this setup with all the other feature docs too.

/approve

PR #9559, opened 2020-07-13
When channel is unavailable, don't try to validate things from it

olemarkus:
Github was returning 500s on the channel request earlier today and I ran into nil pointers around this part of the code.

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -863,6 +863,11 @@ func (c *ApplyClusterCmd) validateKopsVersion() error {
 		return nil
 	}
 
+	if c.channel == nil {
+		klog.Warning("channel unavailable, skipping version validation")
+		return nil
+	}
+
 	versionInfo := kops.FindKopsVersionSpec(c.channel.Spec.KopsVersions, kopsVersion)
 	if versionInfo == nil {
 		klog.Warningf("unable to find version information for kops version %q in channel", kopsVersion)
@@ -974,6 +979,11 @@ func (c *ApplyClusterCmd) validateKubernetesVersion() error {
 
 	// TODO: make util.ParseKubernetesVersion not return a pointer
 	kubernetesVersion := *parsed
+
+	if c.channel == nil {
+		klog.Warning("unable to load channel, skipping kubernetes version recommendation/requirements checks")
+		return nil
+	}
 
 	versionInfo := kops.FindKubernetesVersionSpec(c.channel.Spec.KubernetesVersions, kubernetesVersion)
 	if versionInfo == nil {


upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -222,9 +222,13 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
...
 
-	err = c.validateKubernetesVersion()
-	if err != nil {
-		return err
+	if c.channel != nil {
+		err = c.validateKubernetesVersion()
+		if err != nil {
+			return err
+		}
+	} else {
```
rifelpet, line 230: in addition to nil, c.channel can also point to an empty `Channel{}`:
 https://github.com/kubernetes/kops/blob/781279e268847579b47a87600380675317233b58/pkg/apis/kops/channel.go#L88

maybe we should handle that case as well?

justinsb: Looking at the code for validateKubernetesVersion, it does seem like (1) there are some extra checks in validateKubernetesVersion and (2) validateKubernetesVersion does tolerate the idea that the kubernetes version might not be listed in the channel.

So I suggest we tolerate a nil channel in validateKubernetesVersion, as we do empty kubernetesVersions, but we should always call into validateKubernetesVersion.

olemarkus: Returning nil if the channel is nil just before the version recommendation/requirements checks now

rifelpet:
This looks good now 👍 I'm wondering if it should behave differently if the channel is a URL rather than the normal `stable` or `alpha` values. If a user has provided their own channel URL, would they still want these failures to be treated as a warning?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -976,6 +981,11 @@ func (c *ApplyClusterCmd) validateKubernetesVersion() error {
 	// TODO: make util.ParseKubernetesVersion not return a pointer
 	kubernetesVersion := *parsed
 
+	if c.channel == nil {
+		klog.Warning("unable to load channel. Skipping kubernetes version recommendation/requirements checks", kubernetesVersion)
+		// Not a hard-error
+		return nil
+	}
```
hakman, line 988:
```suggestion
		return nil
	}

```

hakman:
> I'm wondering if it should behave differently if the channel is a URL rather than the normal `stable` or `alpha` values. If a user has provided their own channel URL, would they still want these failures to be treated as a warning?

Should be ok to keep it as is. If channel is down, you need to provide a version anyway.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -976,6 +981,11 @@ func (c *ApplyClusterCmd) validateKubernetesVersion() error {
 	// TODO: make util.ParseKubernetesVersion not return a pointer
 	kubernetesVersion := *parsed
 
+	if c.channel == nil {
+		klog.Warning("unable to load channel. Skipping kubernetes version recommendation/requirements checks", kubernetesVersion)
```
hakman, line 985: Missed this the first time...
```suggestion
		klog.Warning("unable to load channel, skipping kubernetes version recommendation/requirements checks")
```

hakman:
Please also squash the commits.

PR #9560, opened 2020-07-13
Determine fixedip during nodeup directly

olemarkus:
As discussed in #9554 

This will break the task dependency cycle. It is a bit hackish with directly using protokube-code. That part should be moved to something shared, but perhaps best done in a followup PR?

### DIFF
--- a/pkg/model/openstackmodel/network.go
+++ b/pkg/model/openstackmodel/network.go
@@ -50,6 +50,10 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	needRouter := true
+	//Do not need router if there is no external network
+	if b.Cluster.Spec.CloudConfig.Openstack.Router.ExternalNetwork == nil {
+		needRouter = false
+	}
 	routerName := strings.Replace(clusterName, ".", "-", -1)
 	for _, sp := range b.Cluster.Spec.Subnets {
 		// assumes that we do not need to create routers if we use existing subnets
--- a/pkg/model/openstackmodel/network.go
+++ b/pkg/model/openstackmodel/network.go
@@ -49,9 +49,11 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
 		c.AddTask(t)
 	}
 
+	osSpec := b.Cluster.Spec.CloudConfig.Openstack
+
 	needRouter := true
 	//Do not need router if there is no external network
-	if b.Cluster.Spec.CloudConfig.Openstack.Router.ExternalNetwork == nil {
+	if osSpec.Router == nil || osSpec.Router.ExternalNetwork == nil {
 		needRouter = false
 	}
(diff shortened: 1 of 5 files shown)


johngmyers:
First commit message has typo "defauls"

pkg/model/openstackmodel/servergroup.go
```diff
@@ -164,6 +161,9 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 				}
 				c.AddTask(t)
 			case kops.InstanceGroupRoleMaster:
+				if b.Cluster.Spec.Topology != nil && b.Cluster.Spec.Topology.Masters == kops.TopologyPrivate {
+					continue
+				}
```
johngmyers, line 166: This looks like a different concern. Perhaps it should be in a different commit?

pkg/model/openstackmodel/servergroup.go
```diff
@@ -183,23 +186,12 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
...
 	return nil
 }
 
-func (b *ServerGroupModelBuilder) associateFixedIPToKeypair(fipTask *openstacktasks.Instance) {
-	// Ensure the floating IP is included in the TLS certificate,
-	// if we're not going to use an alias for it
-	fipTask.ForAPIServer = true
-}
-
 func (b *ServerGroupModelBuilder) associateFIPToKeypair(fipTask *openstacktasks.FloatingIP) {
```
johngmyers, line 195: I suspect you'll want to remove this as well, but that can be a separate PR.

pkg/model/openstackmodel/servergroup.go
olemarkus: Yep. Done.

pkg/model/openstackmodel/servergroup.go
olemarkus: Yep. Either in #9554 or a follow-up.

olemarkus:
/hold
Suspect @zetaab wants to review this one as well.

zetaab:
I cannot get this PR working

https://gist.github.com/zetaab/a878b0f4f6f7cd18dc174231d1731eda

rifelpet:
Retrying the MacOS GHA job fixed the failure but there's now a merge conflict

pkg/model/openstackmodel/network.go
```diff
@@ -50,6 +50,10 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	needRouter := true
+	//Do not need router if there is no external network
+	if b.Cluster.Spec.CloudConfig.Openstack.Router.ExternalNetwork == nil {
```
rifelpet, line 54: I think you'll want to do additional checking here to avoid a nil pointer dereference

olemarkus: Yep. Thanks for catching.

PR #9561, opened 2020-07-13
Don't try to delete terraform providers

justinsb:
Because we are running it in docker, I was hitting permission issues
during the cleanup step of verify-terraform.


rifelpet:
👍 this will also cache providers between script invocations which is a big improvement too.

/lgtm

PR #9562, opened 2020-07-13
File permission test: clear umask before testing

justinsb:
If the umask isn't 022, directories may not be created in mode 0755,
even if we ask them to be created as such.

umask problems seem unlikely to affect real world usage (negatively)


justinsb:
Maybe we should warn in e.g. nodeup if the umask is "surprising"?

PR #9565, opened 2020-07-13
Upgrade kubernetes dependencies to 1.19.0-rc.4

rifelpet:
This is a find/replace of `v0.18.1` to `v0.19.0-rc.2` in go.mod and then running `make gomod`


hakman:
@rifelpet tomorrow should be a .rc1 release. Should we wait for it. Many things were blocked from entering .rc0.

rifelpet:
sounds good to me. ill use this as a sandbox to get the builds working with the updated deps and then update it to rc.1 when it is released.

hakman:
Sounds perfect

rifelpet:
some troubleshooting notes:

There are failures anytime the apimachinery code is exercised:

>     TestCreateClusterWithINGWSpecified: create_cluster_integration_test.go:191: error running create cluster: error writing updated configuration: error writing Cluster "private.example.com": error marshaling object: error encoding object: converting (*kops.EtcdClusterSpec) to (*v1alpha2.EtcdClusterSpec): unknown conversion

The error is always for EtcdClusterSpec (though its possible if we fix it then we'll get an error for the next impacted field). This seems to be related to https://github.com/kubernetes/kubernetes/pull/90018

---

We're also getting interface type errors for logr which is used by klog. This seems to fix it but will need to be merged and propagated https://github.com/go-logr/logr/pull/20 otherwise we're in a bit of a dependency mess until all our deps update to the newer version of logr.

---

lastly theres a github.com/googleapis/gnostic case change that was a pain to update on my case-insensitive mac: https://github.com/googleapis/gnostic/issues/157

rifelpet:
I think we're blocked on https://github.com/kubernetes-sigs/controller-runtime/pull/1068 to get entirely off of logr v0.1.0 and waiting on https://github.com/kubernetes/cloud-provider-openstack to update to 1.19 k8s dependencies

justinsb:
Thanks for doing this @rifelpet 

Updating go-logr to ee2de8da5be655b0a1755b264167fcca1d3193f5 i.e. `github.com/go-logr/logr v0.2.1-0.20200730175230-ee2de8da5be6` in go.mod should get past that problem (i.e. https://github.com/go-logr/logr/pull/20 was merged )

It looks like there will be two other problems - one a deprecation of a proto library (can either exclude or just change the import) and the other a yaml format issue which I think just needs a regeneration of the expected values.

rifelpet:
Thanks @justinsb the jobs seem to be much better now. Do you know which proto library needs to be excluded or have its import updated to fix `error marshaling object: error encoding object: converting (*kops.EtcdClusterSpec) to (*v1alpha2.EtcdClusterSpec): unknown conversion` ?

rifelpet:
Regarding the api conversion errors, I was able to fix it by editing the generated conversions file in https://github.com/kubernetes/kops/pull/9565/commits/fa68d634c0438be309cf1ce166145f1b2ce87a5c. This isn't an acceptable fix because this file is automatically generated, so we'll have to figure out how to have the code generators make these equivalent changes or I believe we can move the conversion functions into a non-generated file - the code generator will see the functions already exist and not try to duplicate them but we would need to keep those functions up to date whenever the API types change.

rifelpet:
@justinsb any chance you could take a look at this? I forgot to bring it up at office hours today. my last comment above describes the current state of this PR

/assign @justinsb

go.mod
```diff
@@ -2,58 +2,57 @@ module k8s.io/kops
...
-replace k8s.io/cli-runtime => k8s.io/cli-runtime v0.18.6
+replace k8s.io/cli-runtime => k8s.io/cli-runtime v0.19.0-rc.4
 
-replace k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.18.6
+replace k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.19.0-rc.4
 
-replace k8s.io/code-generator => k8s.io/code-generator v0.18.6
+replace k8s.io/code-generator => k8s.io/code-generator v0.19.0-rc.4
 
 replace github.com/gophercloud/gophercloud => github.com/gophercloud/gophercloud v0.9.0
```
olemarkus, line 49: I wonder if we can just punt this one. If not you should probably change it to match the version below.

rifelpet:
🎉 
/test pull-kops-e2e-k8s-containerd

rifelpet:
/hold cancel

this is ready to go

rifelpet:
Just in time for the 1.19.0 tomorrow :)  so much for the head start

PR #9566, opened 2020-07-13
Add ARM64 support for masters

hakman:
This PR adds build and side-loading support for ARM64 masters:
* build multi-arch images (the AMD64 image is pushed twice, with and without the `-amd64` repo suffix.
* add support for side-loading the multi-arch images
* disable the Protokube checks in mock

The Protokube check has to be disabled in mock to allow changing the image names change to multi-arch from:
```
dns-controller.tar.gz
kops-controller.tar.gz
kube-apiserver-healthcheck.tar.gz
protokube.tar.gz
```
to:
```
dns-controller-amd64.tar.gz
dns-controller-arm64.tar.gz
kops-controller-amd64.tar.gz
kops-controller-arm64.tar.gz
kube-apiserver-healthcheck-amd64.tar.gz
kube-apiserver-healthcheck-arm64.tar.gz
protokube-amd64.tar.gz
protokube-arm64.tar.gz
```

TODO (in future PRs):
* re-enable the Protokube checks in mock after the next release
* decide on how to handle multi-arch manifests for the _kops_ images
* remove the `-plain` bazel image targets
* build etcd-manager multi-arch images https://github.com/kopeio/etcd-manager/pull/334


johngmyers:
This supports a mixed-arch control plane, which seems unlikely. Perhaps we should have a cluster spec field specifying the arch of the control plane, allowing apply_cluster to save a bunch of asset management.

hakman:
> This supports a mixed-arch control plane, which seems unlikely. Perhaps we should have a cluster spec field specifying the arch of the control plane, allowing apply_cluster to save a bunch of asset management.

@johngmyers I thought about that also, but I decided on not complicating the spec even more. Current implementation allows the node to autodetect whatever OS arch the node has and keeps the logic consistent in the asset management code.

I am in four of extending this to a cluster spec option though:
https://github.com/kubernetes/kops/pull/9535

olemarkus:
/lgtm

This gives me a valid cluster bar the autoscaler pods that comes in #9744

PR #9211, opened 2020-05-29
feat(openstack): propagate cloud labels to machines

mitch000001:
Fixes #9086

PR #9575, opened 2020-07-15
Take node labels from cloud tags on AWS

### LINKED ISSUE #9856: kops-controller can't determine AWS region
 this information.**

```shell
➜ kops version
Version 1.18.0 (git-698bf974d8)
```

 a `kops` flag.**

```shell
➜ kubectl version --short
Client Version: v1.19.0
Server Version: v1.18.8
```

AWS GovCloud

This was done as part of an upgrade from Kubernetes v1.17.10 created using kops v1.17.1 to Kubernetes v1.18.8 using kops v1.18.0. We upgraded the cluster and removed the pinned Docker version (originally added to fix Docker issues on Amazon Linux). So it went something like the following:

```shell
kops upgrade cluster --yes
kops edit cluster # removed the docker stuff
kops update cluster --yes
kops rolling-update cluster --yes
```

Everything came up again smoothly except the `role` labels weren't being added to the worker nodes. Looking into the `kops-controller`, we were getting errors like this:

```
I0901 17:32:15.966765       1 s3context.go:325] unable to read /sys/devices/virtual/dmi/id/product_uuid, assuming not running on EC2: open /sys/devices/virtual/dmi/id/product_uuid: permission denied
I0901 17:32:15.967033       1 s3context.go:170] defaulting region to "us-east-1"
I0901 17:32:16.311839       1 s3context.go:191] unable to get bucket location from region "us-east-1"; scanning all regions: InvalidToken: The provided token is malformed or otherwise invalid.
     status code: 400, request id: 4Z4P1MCS0S6K8P0R, host id: HOST-ID
 E0901 17:32:16.647080       1 controller.go:258] controller-runtime/controller "msg"="Reconciler error" "error"="unable to load cluster object for node ip-172-20-40-221.us-gov-west-1.compute.internal: error loading Cluster \"s3://cipher
 morph-com-k8s-local-kops/ciphermorph.com.k8s.local/cluster.spec\": Unable to list AWS regions: AuthFailure: AWS was not able to validate the provided access credentials\n\tstatus code: 401, request id: 870266f7-a2ac-417e-9fb3-e7496e5ba6
 29"  "controller"="node" "request"={"Namespace":"","Name":"ip-172-20-40-221.us-gov-west-1.compute.internal"}
```

I expected the `kops-controller` to be able to determine I'm running in by reading that file and not error out. Therefore, the `kops-controller` would be able to proceed and add the labels to the node.

```yaml
kind: Cluster
metadata:
  creationTimestamp: "2020-04-11T19:12:40Z"
  generation: 7
  name: our.cluster.k8s.local
spec:
  additionalPolicies:
    node: |
      [
        {"Effect":"Allow","Action":["autoscaling:DescribeAutoScalingGroups","autoscaling:DescribeAutoScalingInstances","autoscaling:DescribeLaunchConfigurations","autoscaling:DescribeTags","autoscaling:SetDesiredCapacity","autoscaling:TerminateInstanceInAutoScalingGroup"],"Resource":"*"}
      ]
  api:
    loadBalancer:
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudProvider: aws
  configBase: s3://our-bucket/our.cluster.k8s.local
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-us-gov-west-1a
      name: a
    memoryRequest: 100Mi
    name: main
  - cpuRequest: 100m
    etcdMembers:
    - instanceGroup: master-us-gov-west-1a
      name: a
    memoryRequest: 100Mi
    name: events
  iam:
    allowContainerRegistry: true
    legacy: false
  kubelet:
    anonymousAuth: false
  kubernetesApiAccess:
  - 0.0.0.0/0
  kubernetesVersion: 1.18.8
  masterInternalName: api.internal.our.cluster.k8s.local
  masterPublicName: api.our.cluster.k8s.local
  networkCIDR: 172.172.0.0/16
  networking:
    kubenet: {}
  nonMasqueradeCIDR: 127.0.0.1/10
  sshAccess:
  - 0.0.0.0/0
  subnets:
  - cidr: 172.172.172.0/19
    name: us-gov-west-1a
    type: Public
    zone: us-gov-west-1a
  topology:
    dns:
      type: Public
    masters: public
    nodes: public
```

This is running in AWS GovCloud region `us-gov-west-1`.

Digging into the code, I found [the place where this check it happening](https://github.com/kubernetes/kops/blob/86cbd296327cea20b84927b29bbdabb160b34459/util/pkg/vfs/s3context.go#L157-L158). By setting the `AWS_REGION` environment variable in the Daemonset, I was able to get things working. However, this won't be a long-term solution since I assume this will get overwritten when we do another upgrade.

I looked at the file `/sys/devices/virtual/dmi/id/product_uuid` directly on the node since it is the file the controller can't open and saw that it is owned by `root`. So Initially I was thinking maybe Amazon Linux makes `root ` own it but it isn't on other distros. This is incorrect. I spun up an Ubuntu instance, and it too had that file owned by root.

I also tried updating to the most recent AMI of Amazon Linux, but that didn't fix anything either.

johngmyers:
On AWS, refactors the kops-controller to source the node labels from the instance's cloud tags. This is so kops-controller doesn't read the (possibly unapplied) instancegroup spec from the state store.

The other cloud providers are strongly encouraged to refactor to this design.

~~Reverts #9519 due to its inability to handle tags with empty values (as mentioned in #9536).~~

Fixes #9856

### DIFF
--- a/cmd/kops-controller/controllers/legacy_node_controller.go
+++ b/cmd/kops-controller/controllers/legacy_node_controller.go
@@ -18,15 +18,12 @@ package controllers
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
@@ -87,7 +84,7 @@ type LegacyNodeReconciler struct {
 }
 
 // +kubebuilder:rbac:groups=,resources=nodes,verbs=get;list;watch;patch
-// Reconciler is the main reconciler function that observes node changes
+// Reconcile is the main reconciler function that observes node changes.
 func (r *LegacyNodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
 	_ = r.log.WithValues("nodecontroller", req.NamespacedName)
@@ -138,7 +135,7 @@ func (r *LegacyNodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) 
 		return ctrl.Result{}, nil
 	}
 
-	if err := r.patchNodeLabels(ctx, node, updateLabels); err != nil {
+	if err := patchNodeLabels(r.coreV1Client, ctx, node, updateLabels); err != nil {
 		klog.Warningf("failed to patch node labels on %s: %v", node.Name, err)
 		return ctrl.Result{}, err
 	}
@@ -150,37 +147,6 @@ func (r *LegacyNodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Node{}).
 		Complete(r)
-}
-
-type nodePatch struct {
-	Metadata *nodePatchMetadata `json:"metadata,omitempty"`
-}
-
-type nodePatchMetadata struct {
-	Labels map[string]string `json:"labels,omitempty"`
-}
-
-// patchNodeLabels patches the node labels to set the specified labels
-func (r *LegacyNodeReconciler) patchNodeLabels(ctx context.Context, node *corev1.Node, setLabels map[string]string) error {
-	nodePatchMetadata := &nodePatchMetadata{
-		Labels: setLabels,
-	}
-	nodePatch := &nodePatch{
-		Metadata: nodePatchMetadata,
-	}
-	nodePatchJson, err := json.Marshal(nodePatch)
-	if err != nil {
-		return fmt.Errorf("error building node patch: %v", err)
-	}
-
-	klog.V(2).Infof("sending patch for node %q: %q", node.Name, string(nodePatchJson))
-
-	_, err = r.coreV1Client.Nodes().Patch(ctx, node.Name, types.StrategicMergePatchType, nodePatchJson, metav1.PatchOptions{})
-	if err != nil {
-		return fmt.Errorf("error applying patch to node: %v", err)
-	}
-
-	return nil
 }
 
 // getClusterForNode returns the kops.Cluster object for the node
--- a/cmd/kops-controller/controllers/legacy_node_controller.go
+++ b/cmd/kops-controller/controllers/legacy_node_controller.go
@@ -40,9 +40,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-// NewNodeReconciler is the constructor for a NodeReconciler
-func NewNodeReconciler(mgr manager.Manager, configPath string, identifier nodeidentity.Identifier) (*NodeReconciler, error) {
-	r := &NodeReconciler{
+// NewLegacyNodeReconciler is the constructor for a LegacyNodeReconciler
+func NewLegacyNodeReconciler(mgr manager.Manager, configPath string, identifier nodeidentity.LegacyIdentifier) (*LegacyNodeReconciler, error) {
+	r := &LegacyNodeReconciler{
 		client:     mgr.GetClient(),
 		log:        ctrl.Log.WithName("controllers").WithName("Node"),
 		identifier: identifier,
@@ -64,9 +64,9 @@ func NewNodeReconciler(mgr manager.Manager, configPath string, identifier nodeid
 	return r, nil
 }
 
-// NodeReconciler observes Node objects, and labels them with the correct labels for the instancegroup
+// LegacyNodeReconciler observes Node objects, and labels them with the correct labels for the instancegroup
 // This used to be done by the kubelet, but is moving to a central controller for greater security in 1.16
-type NodeReconciler struct {
+type LegacyNodeReconciler struct {
 	// client is the controller-runtime client
 	client client.Client
 
@@ -77,7 +77,7 @@ type NodeReconciler struct {
 	coreV1Client *corev1client.CoreV1Client
 
 	// identifier is a provider that can securely map node ProviderIDs to InstanceGroups
-	identifier nodeidentity.Identifier
+	identifier nodeidentity.LegacyIdentifier
 
 	// configBase is the parsed path to the base location of our configuration files
 	configBase vfs.Path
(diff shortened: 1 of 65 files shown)


pkg/nodeidentity/aws/identify.go
```diff
@@ -35,6 +35,7 @@ import (
 // CloudTagInstanceGroupName is a cloud tag that defines the instance group name
 // This is used by the aws nodeidentifier to securely identify the node instancegroup
 const CloudTagInstanceGroupName = "kops.k8s.io/instancegroup"
+const CloudTagLabelPrefix = "kops.k8s.io/label/"
```
johngmyers, line 38: We might want to use `"k8s.io/cluster-autoscaler/node-template/label/"` instead, per https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-scale-a-node-group-to-0

johngmyers:
/approve cancel
Would like @justinsb review on this design change.

hakman:
I think @rifelpet was saying that AWS will fix soon the inability to handle tags with empty values. Maybe best to exclude from this PR.

johngmyers:
As can be seen from the expected test results, this PR will try to push `k8s.io/cluster-autoscaler/node-template/label/node-role.kubernetes.io/node` tags with empty values. So it cannot go in without either reverting #9519 or waiting for AWS to fix.

hakman:
For Cluster Autoscaler it does not matter what value the tags have. Could set the value to "1" or "true" for now and remove it when the AWS fix is fully deployed. It is already available in `us-east-1` I think, right @rifelpet ?

rifelpet:
I believe you can remove the revert commit now, the AWS fix has been rolled out.

johngmyers:
@olemarkus that Test_ServerGroupModelBuilder looks difficult to update. A good project would be to refactor it to use golden output files like the AWS tests use.

johngmyers:
It's #9211 that's causing `Test_ServerGroupModelBuilder` to interact badly with this change. Now every time there's a change to `model.CloudTagesForInstanceGroup()` a whole lot of manual changes need to be made to the expected output it has encoded in test case structs.

I'm inclined to disable that test until it can be recoded to use `model.RunGoldenTest()`.

cmd/kops-controller/main.go
```diff
@@ -179,16 +181,26 @@ func addNodeController(mgr manager.Manager, opt *config.Options) error {
 		return fmt.Errorf("identifier for cloud %q not implemented", opt.Cloud)
 	}
 
-	if opt.ConfigBase == "" {
-		return fmt.Errorf("must specify configBase")
-	}
+	if identifier != nil {
```
justinsb, line 184: Just a suggestion - the other way to structure this is to handle the "new" cases first, then fall-through to the other case.  I find it tends to be more intuitive to read.

cmd/kops-controller/controllers/legacy_node_controller.go
```diff
@@ -0,0 +1,233 @@
...
+	"k8s.io/kops/pkg/nodelabels"
+	"k8s.io/kops/upup/pkg/fi/utils"
+	"k8s.io/kops/util/pkg/vfs"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// NewLegacyNodeReconciler is the constructor for a LegacyNodeReconciler
+func NewLegacyNodeReconciler(mgr manager.Manager, configPath string, identifier nodeidentity.LegacyIdentifier) (*LegacyNodeReconciler, error) {
```
justinsb, line 41: Ugh ... github messed up this diff!  I try to create a new file for the new code, rather than renaming, just to keep the diff simpler (if I remember!).  This could be `CloudLabelNodeReconciler` or something.

FYI cluster-api may have yet another way ... if we create the AWS instance ourselves, we can feed through a per-instance identifier that gives us even more confidence in the mapping.

cmd/kops-controller/controllers/node_controller.go
```diff
@@ -104,26 +85,12 @@ func (r *NodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-
-	labels := nodelabels.BuildNodeLabels(cluster, ig)
-
-	lifecycle, err := r.getInstanceLifecycle(ctx, node)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to get instance lifecycle %s: %v", node.Name, err)
-	}
-
-	if len(lifecycle) > 0 {
-		labels[fmt.Sprintf("node-role.kubernetes.io/%s-worker", lifecycle)] = "true"
```
justinsb, line 125: Did we decide what to do about this node-role?  (It looks like we're no longer going to set it?)

justinsb:
I'm trying to understand the underlying problem.  We update the instancegroup to add/remove labels, and then they get applied immediately?

I'd like to get the nodes able to get their configuration directly from kops-controller rather than requiring it to be on S3 (and giving the nodes S3 permission); that does require all the information in the instancegroup.yaml.  The PoC is in #8198 .  So the issue is that if we head in that direction, we likely won't be able to put all the information into the instance tags (?).  I feel that the problem likely still exists when booting the nodes as well, as they read the latest cluster & instancegroup from S3.  So I'm trying to understand the problem & the direction here :-)

The changes in here to fix the labels are good though - we should have been using BuildNodeLabels - that looks like a bug.

cmd/kops-controller/controllers/node_controller.go
johngmyers: I moved this logic into the cloud-provider-specific code. It just includes this label in the list of labels it returns.

cmd/kops-controller/controllers/legacy_node_controller.go
johngmyers: I spliced in a rename commit, so the diff should be clearer now.

johngmyers:
I've listed the dependency graph in #9229. The underlying problem is indeed that changes to labels get applied too soon. There's also the issue that the instancegroup spec is written as the versioned API but read as the unversioned API.

The direction I'm trying to take is refactoring away from reading things that are updated at edit time towards things that are updated only on a "kops update cluster --yes" or "terraform apply".

Not that I understand why people want kops to generate Terraform output.

olemarkus:
> @olemarkus that Test_ServerGroupModelBuilder looks difficult to update. A good project would be to refactor it to use golden output files like the AWS tests use.

Yeah it is a beast, but it has also been very helpful. I'll about converting to golden output after I am done with the PR/branches I have in queue now.

johngmyers:
@justinsb I'm having some difficulty teasing out the remaining intent of #8198, especially since my recent kops-controller worker-node bootstrap work addresses the problem of getting the keypairs.

We are currently advertising a model where edits to the cluster and instancegroup specs don't have an immediate effect on the cluster. In this model the admin can get a preview of the cloud-provider-level changes and the changes are applied as a set only when the admin takes a confirming action. After that point new instances get the new specification, whereas existing nodes tend to keep their existing configuration.

Reality is a lot more fuzzy, with some changes taking effect at edit and/or preview time. I expect admins would find such discrepancies undesirably surprising.

justinsb:
Thanks @johngmyers ... I think my objections were around whether there were going to be more things to solve down the road, which is a silly thing for me to block on ... let's try it and see!

/approve
/lgtm

We can discuss today whether we want this in 1.19

PR #9576, opened 2020-07-15
Remove unused and unmaintained Docker build targets

hakman:
As discussed during the June 19th Office Hours, the Kops Docker images should be build with Bazel.
Pretty much any image built with Docker is unmaintained.


Makefile
```diff
@@ -661,7 +571,7 @@ bazel-crossbuild-protokube:
 
 .PHONY: bazel-crossbuild-protokube-image
 bazel-crossbuild-protokube-image:
-	bazel build ${BAZEL_CONFIG} --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //images:protokube.tar
+	bazel build ${BAZEL_CONFIG} --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //protokube/cmd/protokube:image-bundle.tar
```
johngmyers, line 574: This and the following hunk appear to be a separate concern, moving the location of the bazel build of protokube.

hakman: I can move it in a separate PR if too much for this one. It is cleanup in the same department, removing everything from `/images`.

johngmyers: A separate commit/explanation would have been good. I don't think it needs to be a separate PR.

hakman: I can do that. Thanks :).

hakman: Done.

PR #9585, opened 2020-07-17
Docs - Add the new feature table to various feature sections

rifelpet:
Also cleaning up some creationTimestamps from yaml examples.

It'd be great if someone could double check these versions or even suggest any minimum k8s versions that I left out here.

followup to #9555

### DIFF
--- a/docs/bastion.md
+++ b/docs/bastion.md
@@ -34,7 +34,6 @@ You should now be able to edit and configure your bastion instance group.
 apiVersion: kops.k8s.io/v1alpha2
 kind: InstanceGroup
 metadata:
-  creationTimestamp: "2017-01-05T13:37:07Z"
   name: bastions
 spec:
   associatePublicIp: true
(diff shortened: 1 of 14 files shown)


rifelpet:
FYI @MoShitrit I see that the table's shadow is still the full width on FireFox so the fix you provided might have only worked in Chrome

image

MoShitrit:
> FYI @MoShitrit I see that the table's shadow is still the full width on FireFox so the fix you provided might have only worked in Chrome
> 
> image

Huh, you're right!
Apparently, seems like it's the only browser that doesn't handle that well. I checked in Safari and Chrome and both show it properly. I wonder why is FF different. 
Any idea, @mikesplain ?

mikesplain:
Interesting. I'll open a up a PR fix.

mikesplain:
I believe this should take care of that issue: https://github.com/kubernetes/kops/pull/9592

PR #9592, opened 2020-07-17
Remove min width on tables

mikesplain:
Fixes table issue in firefox:

Before:
Screen Shot 2020-07-17 at 4 36 38 PM
After:
Screen Shot 2020-07-17 at 4 36 16 PM

At first I was going to fix it simply for the new feature boxes however I think this `min-width` is to blame for a number of issues that occur cross all browsers such as on the networking page (chrome): https://kops.sigs.k8s.io/networking/
Screen Shot 2020-07-17 at 4 38 42 PM

Clearly that's not displaying properly so I'm going to remove the min-width param.  We can figure out the right way to set min-width on tables in a follow up, but this should clean up a number of things for now at least.

More context see https://github.com/kubernetes/kops/pull/9585#issuecomment-660244678

### DIFF
--- a/docs/extra.css
+++ b/docs/extra.css
@@ -1,7 +1,3 @@
-.md-typeset__table {
-    min-width: 100%;
-}
-
 @media only screen and (min-width: 768px){
 	td:nth-child(1){
 		white-space: nowrap;


MoShitrit:
/lgtm
Thanks for fixing that! 😄

mikesplain:
Looks better to me: https://5f120ca947720400088f5699--kubernetes-kops.netlify.app/networking/cilium/#enabling-encryption-in-cilium

MoShitrit:
Looks good on my end as well!
Confirmed on Firefox, Chrome and Safari

mikesplain:
Great thanks!

PR #9593, opened 2020-07-18
Reduce the lifetime of exported kubecfg credentials

johngmyers:
/approve cancel

### DIFF
--- a/cmd/kops/export_kubecfg.go
+++ b/cmd/kops/export_kubecfg.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -53,7 +54,7 @@ var (
 type ExportKubecfgOptions struct {
 	KubeConfigPath string
 	all            bool
-	admin          bool
+	admin          time.Duration
 	user           string
 }
 
@@ -76,7 +77,8 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVar(&options.KubeConfigPath, "kubeconfig", options.KubeConfigPath, "the location of the kubeconfig file to create.")
 	cmd.Flags().BoolVar(&options.all, "all", options.all, "export all clusters from the kops state store")
-	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "export the cluster admin user and add it to the context")
+	cmd.Flags().DurationVar(&options.admin, "admin", options.admin, "export a cluster admin user credential with the given lifetime and add it to the cluster context")
+	cmd.Flags().Lookup("admin").NoOptDefVal = kubeconfig.DefaultKubecfgAdminLifetime.String()
 	cmd.Flags().StringVar(&options.user, "user", options.user, "add an existing user to the cluster context")
 
 	return cmd
@@ -92,7 +94,7 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 			return fmt.Errorf("cannot use both --all flag and positional arguments")
 		}
 	}
-	if options.admin && options.user != "" {
+	if options.admin != 0 && options.user != "" {
 		return fmt.Errorf("cannot use both --admin and --user")
 	}
 
--- a/pkg/kubeconfig/create_kubecfg.go
+++ b/pkg/kubeconfig/create_kubecfg.go
@@ -17,18 +17,24 @@ limitations under the License.
 package kubeconfig
 
 import (
+	"crypto/x509/pkix"
 	"fmt"
 	"sort"
+	"time"
 
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/apis/kops/util"
 	"k8s.io/kops/pkg/dns"
+	"k8s.io/kops/pkg/pki"
+	"k8s.io/kops/pkg/rbac"
 	"k8s.io/kops/upup/pkg/fi"
 )
 
-func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.SecretStore, status kops.StatusStore, configAccess clientcmd.ConfigAccess, admin bool, user string) (*KubeconfigBuilder, error) {
+const DefaultKubecfgAdminLifetime = 18 * time.Hour
+
+func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.SecretStore, status kops.StatusStore, configAccess clientcmd.ConfigAccess, admin time.Duration, user string) (*KubeconfigBuilder, error) {
 	clusterName := cluster.ObjectMeta.Name
 
 	master := cluster.Spec.MasterPublicName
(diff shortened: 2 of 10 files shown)


pkg/kubeconfig/create_kubecfg.go
```diff
@@ -104,30 +110,29 @@ func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.Se
...
-				}
-			} else {
-				return nil, fmt.Errorf("cannot find kubecfg key")
-			}
+	if admin != 0 {
+		req := pki.IssueCertRequest{
+			Signer: fi.CertificateIDCA,
+			Type:   "client",
+			Subject: pkix.Name{
+				CommonName:   "kubecfg",
```
johngmyers, line 118: Perhaps use `user.Current().Username` instead?

cmd/kops/export_kubecfg.go
```diff
@@ -76,7 +77,8 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVar(&options.KubeConfigPath, "kubeconfig", options.KubeConfigPath, "the location of the kubeconfig file to create.")
 	cmd.Flags().BoolVar(&options.all, "all", options.all, "export all clusters from the kops state store")
-	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "export the cluster admin user and add it to the context")
+	cmd.Flags().DurationVar(&options.admin, "admin", options.admin, "export a cluster admin user credential with the given lifetime and add it to the cluster context")
+	cmd.Flags().Lookup("admin").NoOptDefVal = kubeconfig.DefaultKubecfgAdminLifetime.String()
 	cmd.Flags().StringVar(&options.user, "user", options.user, "add an existing user to the cluster context")
```
hakman, line 82: Is this in any way useful for "user" also?

johngmyers: "user" doesn't export any credentials, so no.

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -17,18 +17,24 @@ limitations under the License.
...
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/apis/kops/util"
 	"k8s.io/kops/pkg/dns"
+	"k8s.io/kops/pkg/pki"
+	"k8s.io/kops/pkg/rbac"
 	"k8s.io/kops/upup/pkg/fi"
 )
 
-func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.SecretStore, status kops.StatusStore, configAccess clientcmd.ConfigAccess, admin bool, user string) (*KubeconfigBuilder, error) {
+const DefaultKubecfgAdminLifetime = 18 * time.Hour
```
hakman, line 35: Could you explain the 18 hours default?
It is a pretty big change from previous versions.

cmd/kops/export_kubecfg.go
hakman: 👍

pkg/kubeconfig/create_kubecfg.go
johngmyers: Ten years is an unreasonably long time to have an unrevokable credential.

18 hours lets an admin go a working day without having to obtain a new credential, yet makes it likely that their next day starts off with an expired credential. It is more convenient to have to reacquire a credential at the start of a task than mid-way through it.

hakman: Seems like a good explanation for the purpose of strong security.

IMHO, this is a step in the right direction, but may be a bit too extreme. I don't think there was a strong request to reduce the lifetime of the credentials so much. It may be a bit excessive for people with multiple clusters to do this every day, or for every test cluster. I would suggest to change the default to something like 7-30 days that would at least not require new exports for short lived clusters.

Also, an env var may be desirable to change the default to something more suitable for each operator.

johngmyers: 7-30 days is too long for a default lifetime. Fixed-lifetime credentials that aren't themselves protected by other credentials typically have a lifetime no longer than a day. For example, temporary IAM credentials have a default lifetime of 12 hours.

Do we have a precedent/pattern for env vars providing defaults for flags? The env vars I can think of don't have corresponding flags. Do we want to adopt something like viper?

cmd/kops/export_kubecfg.go
```diff
@@ -76,7 +77,8 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVar(&options.KubeConfigPath, "kubeconfig", options.KubeConfigPath, "the location of the kubeconfig file to create.")
 	cmd.Flags().BoolVar(&options.all, "all", options.all, "export all clusters from the kops state store")
-	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "export the cluster admin user and add it to the context")
+	cmd.Flags().DurationVar(&options.admin, "admin", options.admin, "export a cluster admin user credential with the given lifetime and add it to the cluster context")
+	cmd.Flags().Lookup("admin").NoOptDefVal = kubeconfig.DefaultKubecfgAdminLifetime.String()
```
justinsb, line 81: OK, now that is cool.

cmd/kops/export_kubecfg.go
```diff
@@ -92,7 +94,7 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 			return fmt.Errorf("cannot use both --all flag and positional arguments")
 		}
 	}
-	if options.admin && options.user != "" {
+	if options.admin != 0 && options.user != "" {
 		return fmt.Errorf("cannot use both --admin and --user")
```
justinsb, line 98: Should we allow setting the duration when `--user` is specified?  (--user is not an option I use myself, so I don't really know)

pkg/kubeconfig/create_kubecfg.go
justinsb: Specifically on the viper point, we do have limited precedent for --state KOPS_STATE_STORE and we even have a config file.  However, I couldn't figure out how to get viper to just do this automatically - the state store configuration is a non-trivial amount of repetition.  I'm not sure if we're "holding it wrong" but it wasn't very clear.

Multiple clusters (as hakman points out) is indeed going to be more annoying.  I'm wondering if we should think about a helper process for login (that's essentially how GKE gets around this problem), or even nudging users to using a different authenticator, reserving the admin kubeconfig for "break glass".

I'm wondering if we could use a helper process to generate this on the fly, effectively chaining off your AWS credentials, so we'd inherit the same validity period.

pkg/kubeconfig/create_kubecfg.go
justinsb: Yeah, that's not a bad change to make at the same time.  Or maybe "admin-kubecfg-${USER}", just in case of usage inside a container where I expect most of these will end up as "root" or "jenkins" which might be confusing.

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -17,18 +17,24 @@ limitations under the License.
...
+
+func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.SecretStore, status kops.StatusStore, configAccess clientcmd.ConfigAccess, admin time.Duration, user string) (*KubeconfigBuilder, error) {
```
justinsb, line 37: Nit: I might consider splitting admin into `buildAdmin` and `duration`; keep this function logical and not too tightly bound to the flags.  Although admin=false/0 is weird... it looks like the kubeconfig won't work?

justinsb:
Thanks for driving this discussion @johngmyers ... I am going to think more about whether a helper authn process will help us out here, using something like https://kubernetes.io/docs/reference/access-authn-authz/authentication/#input-and-output-formats

cmd/kops/export_kubecfg.go
johngmyers: `--user` doesn't generate credentials, so a duration isn't relevant.

pkg/kubeconfig/create_kubecfg.go
johngmyers: A `duration` without a `buildAdmin` wouldn't make sense. Built credentails with `duration` 0 would not be useful. `buildAdmin` without `duration` would be pushing the defaulting logic down, making it less visible to the command line.

pkg/kubeconfig/create_kubecfg.go
johngmyers: A kubectl credential plugin is an interesting idea.

For production clusters it would be better to nudge users to use a different authenticator.

olemarkus:
FWIW I think that either way we should merge this PR so that those who provision the static certs really know what they are doing if they are producing long-lived ones.

johngmyers:
@justinsb can we land this now there's a kubectl plugin in the works?

PR #9609, opened 2020-07-22
Prefer nodes with "master" role for Calico Typha pods

### LINKED ISSUE #9608: Calico's Typha pods should prefer to run on masters, but tolerate running elsewhere
Following up from https://github.com/kubernetes/kops/issues/9240#issuecomment-661903467, we'd like to [run more Calico Typha pods than we have master nodes](https://docs.projectcalico.org/getting-started/kubernetes/self-managed-onprem/onpremises#install-calico-on-nodes), for clusters with 600 nodes or more. Unfortunately, we can't get more Typha pods scheduled than we have master nodes, even when we have other non-master nodes that could host Typha as well.

1. **What `kops` version are you running?**  
  Version 1.18.0-beta.2 (git-63f9ae1099)

1. **What Kubernetes version are you running?**  
  - _**Client Version:** version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-16T00:04:31Z", GoVersion:"go1.14.4", Compiler:"gc", Platform:"darwin/amd64"}_
  - _**Server Version:** version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}_

3. **What cloud provider are you using?**  
  AWS EC2

1. **What commands did you run?  What is the simplest way to reproduce this issue?**  
  - Create a _kops_ cluster [using the Calico CNI implementation](https://github.com/kubernetes/kops/blob/master/docs/networking/calico.md).
  - [Enable Typha.](https://github.com/kubernetes/kops/blob/master/docs/networking/calico.md#configuring-calico-to-use-typha)
  - In a cluster using three master machines, configure Typha's _Deployment_ to use at least one replica more.  
  ```yaml
networking:
  calico:
    typhaReplicas: 4
```
  - Observe that three of the pods created by the _Deployment_'s _ReplicaSet_ get scheduled—one per master node—but the fourth one remains pending.
  - Observe that the pending pod's spec contains a "spec.nodeSelector" field.
```yaml
spec:
  # ...
  nodeSelector:
    kubernetes.io/os: linux
    kubernetes.io/role: master
```

5. **What happened after the commands executed?**  
  One of the Typha pods remains pending, not placed by the Kubernetes scheduler because each node can host at most one such pod, and there are fewer master nodes that Typha pods.

1. **What did you expect to happen?**
  The first three Typha pods would be scheduled on the master nodes, but the fourth Typha pod would be scheduled on a different non-master node.

1. **Anything else do we need to know?**  
  Using [a node affinity rule](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity) in the Typha _Deployment_'s pod spec template would allow us to express a _preference_ for Typha running on a master node, but not a strict requirement. In this case, we'd use a `preferredDuringSchedulingIgnoredDuringExecution` entry with a match expression adapted from the existing node selector, and we'd remove the node selector.
```yaml
spec:
  affinity:
    nodeAffinity:
        nodeSelectorTerms:
        - matchExpressions:
          - key: kubernetes.io/os
            operator: In
            values:
            - linux
      preferredDuringSchedulingIgnoredDuringExecution:
      - preference:
          matchExpressions:
          - key: kubernetes.io/role
            operator: In
            values:
            - master
       weight: 75
```

hakman:
Fixes #9608 
/hold

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -12170,15 +12170,22 @@ spec:
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict: 'true'
     spec:
+      affinity:
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            preference:
+              matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
       nodeSelector:
         kubernetes.io/os: linux
-        kubernetes.io/role: master
       hostNetwork: true
       tolerations:
         # Mark the pod as a critical add-on for rescheduling.
         - key: CriticalAddonsOnly
           operator: Exists
-        - key: "node-role.kubernetes.io/master"
+        - key: node-role.kubernetes.io/master
           effect: NoSchedule
       # Since Calico can't network a pod until Typha is up, we need to run Typha itself
       # as a host-networked pod.
--- a/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
@@ -3497,15 +3497,22 @@ spec:
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict: 'true'
     spec:
+      affinity:
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            preference:
+              matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
       nodeSelector:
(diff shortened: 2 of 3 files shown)


hakman:
Thanks @gjtempleton :)

rifelpet:
/lgtm
as well.

tmjd:
What is the use case for pushing Typha toward nodes with the "master" role?

PR #9610, opened 2020-07-23
add os.RemoveAll err verification

zhouhao3:
Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/render_test.go
+++ b/upup/pkg/fi/cloudup/awstasks/render_test.go
@@ -41,7 +41,12 @@ func doRenderTests(t *testing.T, method string, cases []*renderTest) {
 		t.Errorf("failed to create local render directory: %s", err)
 		t.FailNow()
 	}
-	defer os.RemoveAll(outdir)
+	defer func() {
+		err := os.RemoveAll(outdir)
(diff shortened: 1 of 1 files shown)


johngmyers:
@zhouhau3 Thank you for your PR. I appreciate efforts to improve the tests.

Unfortunately, this particular change does not bring a lot of value. The failure it is checking for neither indicates a problem in the code under test nor would it invalidate a successful test result. It adds the cost of code complexity to negligible benefit.

I would suggest that your time and the time of the reviewers would be better spent on a different class of improvement. Perhaps you could pick an issue from https://github.com/kubernetes/kops/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

zhouhao3:
@johngmyers Okay, I will focus on these issues you mentioned in the future, thank you for your suggestion.

justinsb:
Going to merge this one; these sorts of things can be a good way to get into the codebase - thanks @zhouhao3 and looking forward to seeing what you tackle next :-)  I always find fixing problems you're encountering yourself is a good way to proceed, particularly if you're trying to do something a little different which other people might not yet have encountered.

/approve
/lgtm

PR #9621, opened 2020-07-24
Render managed files with Terraform

johngmyers:
This is an attempt to have the `ManagedFile` task render through Terraform instead of directly to the state store. This would make them be updated at `terraform apply` time, not during (dryrun) `kops update cluster`.

This is an attempt to move forward with #9229. A followup PR would make the completed cluster spec a `ManagedFile`.

Some issues (beyond the fact this is my first time dealing with Terraform):

* I believe GCE also works with Terraform, so this would probably need to be implemented against `GSPath` as well.
* The integration tests use Terraform with `MemFSPath`. I'm not quite sure what the code should do in this case.


rifelpet:
Could we update the integration test manifests to use s3:// instead of memfs:// ? Some of them test cloudformation as well which doesn't have an s3 object resource, so we may need to create separate manifests or test cases for cloudformation vs terraform.

johngmyers:
It doesn't look like the integration tests have a way to mock the `S3Context` currently. So that would be an effort in itself.

rifelpet:
A few things:
* I'm wondering if we should put this behind a feature flag given the potential impact
* We'll want to document that downgrading kops (or disabling the feature flag) requires the objects be manually removed from the terraform state with `terraform state rm` otherwise terraform will delete the objects on the subsequent apply.
* Just to be clear, the PKI objects in the keyStore aren't being defined in terraform because they're not susceptible to the original issue of storing `cluster-completed.spec` in the state store, is that correct?

johngmyers:
We might need to document getting Terraform to adopt the resources upon upgrade.

We don't want to put secrets in the Terraform spec. As a project dependent on this I plan to move references to the certificates/public keys to userdata or derived managed files so they only change upon node update. Secrets updating before apply could be a problem, but impact should be limited given proper preparation of verifiers before changing the primary secret.

rifelpet:
/lgtm

Perhaps we can bring this up at office hours tomorrow to see if anyone has any concerns, otherwise I think this is ready to merge.

johngmyers:
Decision from Office hours is to hold this a bit until we get further along with the #9294 series. I will look further into per-launchtemplate-version files.

johngmyers:
It's time for this to go in. Currently node boot breaks between the time one does a `kops update cluster --target terraform` and when instances are created using the new spec. This closes the gap to between the time one does a `terraform apply` and when instances are created using the new spec.

To cut down on churn we might want to stop having golden files for the `kops-version.txt` files and many of the addon manifests. But that can be in a subsequent PR.

rifelpet:
What is the status on the per-LaunchTemplateVersion files idea?

johngmyers:
I believe per-LT files is dependent on this. Writing that code will be easier if both non-Terraform and Terraform have the per-LT files cleaned up through reconciliation.

rifelpet (approved):

rifelpet:
I was looking into adding support for GCS objects but given that the integration tests use memfs which use an s3 object definition, it will be difficult to add integration test coverage for GCS. Any suggestions on how to achieve that?

johngmyers:
Since the integration tests use memfs, there's no integration test coverage for S3 either. I don't think that's a problem. What is needed is unit tests for gsfs.

PR #9622, opened 2020-07-24
Delete all security groups when VPC is "owned"

hakman:
Sometimes, tools like [AWS Firewall Manager](https://aws.amazon.com/firewall-manager) add security groups that prevent the VPC from being deleted. Kops should remove them also before attempting to delete the VPC.


hakman:
The required changes are larger than expected. They way resources are discovered has to be redone.

PR #9625, opened 2020-07-24
default to external ccm in OpenStack

### LINKED ISSUE #9619: Use OpenStack external cloud controller by default

in-tree legacy OpenStack cloudprovider is deprecated. This is used by default currently in kops. The recommended action is to move to use external cloud controller manager (external ccm)

when creating kops cluster using OpenStack we need to do currently following if we want use external ccm:

0) `export KOPS_FEATURE_FLAGS=EnableExternalCloudController`
1) kops create cluster ...
2) kops edit cluster
3) add following to spec:
```
cloudControllerManager: {}
```
4) kops update cluster like usually

So we need somehow add that spec modification automatically to OpenStack clusters

zetaab:
fixes #9619 

By default new clusters will be created with external cloud controller manager which is the recommended setup.

Is it valid to remove featureflag OR should I modify it? I could add featureflag or cloudprovider == openstack to most of the places. However, not sure is cluster spec available everywhere


zetaab:
/hold

hold until someone from approvers will comment about featureflag

hakman:
@zetaab I think you need to rebase to make the tests pass.

hakman:
The cilium test is generally broken, don't spend too much time on it. Will have to do something to skip the failed tests until @olemarkus is back from vacation.

rifelpet:
is not using external ccm still a valid configuration for openstack? if we need to support both external and not-external, we should keep the feature flag in. adjusting the feature flag's default value is a bit tricky because we'd only want to adjust the default for openstack and not all cloud providers, and the feature flag package doesnt differentiate between true or false vs unset. It seems like we'd want to have:

| provider | external FF | resulting behavior |
| -- | -- | -- |
| openstack | unset | use external ccm |
| openstack | true | use external ccm | 
| openstack | false | do not use external ccm | 
| any other provider | unset | do not use external ccm |

which isnt possible to express with the current feature flag package on its own. maybe a new function that returns whether the flag is set or not would be acceptable.

johngmyers:
@rifelpet the feature flag does not cause an external CCM to be used, it just allows the `ExternalCloudControllerManager` field in the cluster spec to have a non-`nil` value. Whether an external CCM is used (and how it is configured) is controlled by the field in the cluster spec.

So for OpenStack it makes sense for the feature flag to have no effect. For AWS we probably still want the feature flag, as that mode is not tested.

Currently the feature flag is tested in eight different places. But it would be simpler and just as effective to only test it inside of API validation (and not if the provider is OpenStack).

mitch000001:
The rational behind most of the checks I introduced for the warnings was that it was possible to configure a cluster to use external CCM by defining one but the feature flag would cause that the nodes are not setup appropriately. So, if we choose to only switch to external CCM and configure a default CCM config I think it's fine to drop the feature flag.

mitch000001:
Following up on my pervious post. I saw in the discussion on Slack and when revisiting it that this feature flag is a global one. So possibly it's not as easy to remove it. However we should drop the checks for it if we choose to add Openstack CCM per default.

cmd/kops/create_cluster.go
```diff
@@ -479,6 +479,11 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 		cluster.Spec.MasterPublicName = c.MasterPublicName
 	}
 
+	// in OpenStack we default to external cloud controller manager because legacy in-tree ccm is deprecated
```
johngmyers, line 482: This code should go inside `cloudup.NewCluster()`

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -1053,25 +1053,9 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
...
-
-				addons.Spec.Addons = append(addons.Spec.Addons, &channelsapi.AddonSpec{
-					Name:              fi.String(key),
-					Version:           fi.String(version),
-					Manifest:          fi.String(location),
-					Selector:          map[string]string{"k8s-addon": key},
-					KubernetesVersion: "<1.13.0",
-					Id:                id,
-				})
-			}
```
johngmyers, line 1074: Why is this being deleted?

johngmyers:
Consider moving the image-defaulting code in the `OpenStackCCM` template function to `cloudup.PerformAssignments()`. That will give the image a chance to be remapped to a local registry if one is defined.

If the in-tree CCM is unsupported for k8s 1.19 and higher, that would be a good place to ensure that `ExternalCloudControllerManager` does not remain nil for upgraded clusters.

PR #9637, opened 2020-07-27
Cleanup AWS EC2 eventual consistency warnings

hakman:
When a cluster is created, there are warnings such as:
```
W0727 09:19:16.521016    5204 executor.go:128] error running task "AutoscalingGroup/nodes-ap-northeast-2c.e2e-8e9eb293c0-ff1eb.test-cncf-aws.k8s.io" (9m58s remaining to succeed): error creating AutoscalingGroup: ValidationError: You must use a valid fully-formed launch template. Value (nodes.e2e-8e9eb293c0-ff1eb.test-cncf-aws.k8s.io) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
	status code: 400, request id: 56bfbee6-2dad-4c52-902e-5a8ddf9f5734
```

This happens because of AWS EC2 eventual consistency issues:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html#eventual-consistency

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
+++ b/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
@@ -299,7 +299,13 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 
 		// @step: attempt to create the autoscaling group for us
 		if _, err := t.Cloud.Autoscaling().CreateAutoScalingGroup(request); err != nil {
-			return fmt.Errorf("error creating AutoscalingGroup: %v", err)
+			code := awsup.AWSErrorCode(err)
+			message := awsup.AWSErrorMessage(err)
+			if code == "ValidationError" && strings.Contains(message, "Invalid IAM Instance Profile name") {
+				klog.V(4).Infof("error creating AutoscalingGroup: %s", message)
+				return fi.NewTryAgainLaterError("waiting for the IAM Instance Profile to be propagated")
+			}
+			return fmt.Errorf("error creating AutoScalingGroup: %s", message)
 		}
 
 		// @step: attempt to enable the metrics for us
--- a/upup/pkg/fi/cloudup/awstasks/launchconfiguration.go
+++ b/upup/pkg/fi/cloudup/awstasks/launchconfiguration.go
@@ -22,10 +22,8 @@ import (
 	"math"
 	"sort"
 	"strings"
-	"time"
 
 	"k8s.io/kops/pkg/apis/kops"
-
 	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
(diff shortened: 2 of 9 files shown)


johngmyers:
Good question whether we prefer the individual tasks to sleep or letting the task scheduler retry after a delay.

hakman:
> Good question whether we prefer the individual tasks to sleep or letting the task scheduler retry after a delay.

The task already retries until it succeeds, but the problem is the useless warnings. Unless there is a way to tell the task just to retry and not print a warning, seems best to keep this logic in each individual task.

johngmyers:
We could create a new type which, if embedded in an `error`, causes `fi.RunTasks()` to print at info level instead of warning.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -142,31 +141,8 @@ func (t *LaunchTemplate) RenderAWS(c *awsup.AWSAPITarget, a, ep, changes *Launch
...
-				klog.V(4).Infof("ErrorCode=%q, Message=%q", awsup.AWSErrorCode(err), awsup.AWSErrorMessage(err))
-			}
-		}
-
-		return err
-	}()
-	if err != nil {
-		return fmt.Errorf("failed to create aws launch template: %s", err)
+	if _, err = c.Cloud.EC2().CreateLaunchTemplate(input); err != nil {
+		return fmt.Errorf("error creating LaunchTemplate: %v", err)
```
hakman, line 145: I never seen these warnings in my tests. Would like to see if this workaround is still needed.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -142,31 +141,8 @@ func (t *LaunchTemplate) RenderAWS(c *awsup.AWSAPITarget, a, ep, changes *Launch
...
 	// @step: attempt to create the launch template
-	err = func() error {
-		for attempt := 0; attempt < 10; attempt++ {
-			if _, err = c.Cloud.EC2().CreateLaunchTemplate(input); err == nil {
-				return nil
-			}
-
-			if awsup.AWSErrorCode(err) == "ValidationError" {
-				message := awsup.AWSErrorMessage(err)
-				if strings.Contains(message, "not authorized") || strings.Contains(message, "Invalid IamInstance") {
```
hakman, line 153: I never seen these warnings in my tests. Would like to see if the handling for these errors is still needed.

hakman:
> We could create a new type which, if embedded in an `error`, causes `fi.RunTasks()` to print at info level instead of warning.

I think it should be better now.

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -299,7 +299,13 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 
 		// @step: attempt to create the autoscaling group for us
 		if _, err := t.Cloud.Autoscaling().CreateAutoScalingGroup(request); err != nil {
-			return fmt.Errorf("error creating AutoscalingGroup: %v", err)
+			code := awsup.AWSErrorCode(err)
+			message := awsup.AWSErrorMessage(err)
+			if code == "ValidationError" && strings.Contains(message, "Invalid IAM Instance Profile name") {
+				klog.V(4).Infof("error creating AutoscalingGroup: %s", message)
+				return fi.NewTryAgainLaterError("Waiting for the IAM Instance Profile to be propagated")
```
johngmyers, line 306:
```suggestion
				return fi.NewTryAgainLaterError("waiting for the IAM Instance Profile to be propagated")
```

upup/pkg/fi/cloudup/awstasks/launchconfiguration.go
```diff
@@ -353,32 +351,14 @@ func (_ *LaunchConfiguration) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *La
...
-		klog.V(8).Infof("AWS CreateLaunchConfiguration %s", aws.StringValue(request.LaunchConfigurationName))
-		_, err = t.Cloud.Autoscaling().CreateLaunchConfiguration(request)
-		if err == nil {
-			break
+	if _, err = t.Cloud.Autoscaling().CreateLaunchConfiguration(request); err != nil {
+		code := awsup.AWSErrorCode(err)
+		message := awsup.AWSErrorMessage(err)
+		if code == "ValidationError" && strings.Contains(message, "Invalid IamInstanceProfile") {
+			klog.V(4).Infof("error creating LaunchConfiguration: %s", message)
+			return fi.NewTryAgainLaterError("Waiting for the IAM Instance Profile to be propagated")
```
johngmyers, line 359:
```suggestion
			return fi.NewTryAgainLaterError("waiting for the IAM Instance Profile to be propagated")
```

upup/pkg/fi/cloudup/awstasks/route.go
```diff
@@ -175,7 +171,13 @@ func (_ *Route) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Route) error {
 
 		response, err := t.Cloud.EC2().CreateRoute(request)
 		if err != nil {
-			return fmt.Errorf("error creating Route: %v", err)
+			code := awsup.AWSErrorCode(err)
+			message := awsup.AWSErrorMessage(err)
+			if code == "InvalidNatGatewayID.NotFound" {
+				klog.V(4).Infof("error creating Route: %s", message)
+				return fi.NewTryAgainLaterError("Waiting for the NAT Gateway to be created")
```
johngmyers, line 178:
```suggestion
				return fi.NewTryAgainLaterError("waiting for the NAT Gateway to be created")
```

upup/pkg/fi/cloudup/awstasks/route.go
```diff
@@ -204,9 +202,14 @@ func (_ *Route) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Route) error {
...
 
-		_, err := t.Cloud.EC2().ReplaceRoute(request)
-		if err != nil {
-			return fmt.Errorf("error updating Route: %v", err)
+		if _, err := t.Cloud.EC2().ReplaceRoute(request); err != nil {
+			code := awsup.AWSErrorCode(err)
+			message := awsup.AWSErrorMessage(err)
+			if code == "InvalidNatGatewayID.NotFound" {
+				klog.V(4).Infof("error creating Route: %s", message)
+				return fi.NewTryAgainLaterError("Waiting for the NAT Gateway to be created")
```
johngmyers, line 210:
```suggestion
				return fi.NewTryAgainLaterError("waiting for the NAT Gateway to be created")
```

PR #9645, opened 2020-07-29
Refactor openstackCloud to be mockable, add a MockCloud

rifelpet:
This is the first step towards an Openstack implementation of cloudmock, allowing us to have Openstack integration tests.

This separates the majority of the logic from the live implementation of the OpenstackCloud interface, allowing us to reuse the logic in openstack.MockCloud. 

This is the same pattern followed in the AWS package. example:

https://github.com/kubernetes/kops/blob/7f7b5b1782f7b0ac8aabd6581ea8d2b95bf82d1d/upup/pkg/fi/cloudup/awsup/aws_cloud.go#L1038-L1059

https://github.com/kubernetes/kops/blob/7f7b5b1782f7b0ac8aabd6581ea8d2b95bf82d1d/upup/pkg/fi/cloudup/awsup/mock_aws_cloud.go#L161-L163

### DIFF
--- a/upup/pkg/fi/cloudup/openstack/BUILD.bazel
+++ b/upup/pkg/fi/cloudup/openstack/BUILD.bazel
@@ -12,6 +12,7 @@ go_library(
         "instance.go",
         "keypair.go",
         "loadbalancer.go",
+        "mock_cloud.go",
         "network.go",
         "port.go",
         "router.go",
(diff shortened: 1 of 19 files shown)


rifelpet:
/cc @olemarkus 

I have a followup PR that creates a cloudmock implementation for openstack and sets up an integration test but I'd like to get this reviewed and merged first.

johngmyers:
Were you planning on making that OpenStack integration test merge-blocking? I'm not sure how I would have managed to get that cert refactor accomplished with a blocking OpenStack test, given I have neither sufficient OpenStack expertise nor a test environment.

olemarkus:
I don't think I am a huge fan of this pattern, but at least it is consistent :)
/lgtm

PR #9646, opened 2020-07-29
Remove dead code from tasks

johngmyers:
/kind cleanup

### DIFF
--- a/upup/pkg/fi/assettasks/copydockerimage_fitask.go
+++ b/upup/pkg/fi/assettasks/copydockerimage_fitask.go
@@ -21,31 +21,10 @@ limitations under the License.
 package assettasks
 
 import (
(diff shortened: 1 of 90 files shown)


hakman:
Should "upup/pkg/fi/cloudup/awstasks/launchtemplate_fitask.go" be part of the "make codegen" commit?

johngmyers:
launchtemplate_fitask.go isn't actually automatically generated.

hakman:
> launchtemplate_fitask.go isn't actually automatically generated.

Anything that can be done about it, or we should leave it as is?

johngmyers:
It looks like it would be possible to make it autogenerated in a follow-up PR.

hakman:
Cool. Thanks!
/hold cancel

PR #9650, opened 2020-07-29
Remove tags from NodeupConfig

johngmyers:
/kind cleanup


hakman:
Cool. I think we can get rid of the other tags also at some point.
/lgtm

PR #7758, opened 2019-10-08
Add a BAZEL_CONFIG Makefile arg to bazel commands

fejta:
Intend to change CI jobs from `make bazel-build` to `make bazel-build BAZEL_CONFIG=--config=ci`

https://github.com/kubernetes/test-infra/blob/4db0b31df68e72c01edfe6eeb450d6cccd4f1615/config/jobs/kubernetes/kops/kops-config.yaml#L22-L23

Note that `bazel test --config=ci` is required (`bazel --config=ci test # unrecognized flag`)

PR #9651, opened 2020-07-30
Use BAZEL_OPTIONS  for bazel build

srikiz:
Included BAZEL_OPTIONS pre-fixed for bazel build.

Initial changes to https://github.com/kubernetes/kops/pull/7758 added BAZEL_CONFIG but that doesn't work for few arguments like below.
`--host_jvm_args=-Djavax.net.ssl.trustStore=/usr/lib/jvm/java-11-openjdk-amd64/lib/security/cacerts  --host_jvm_args=-Djavax.net.ssl.trustStorePassword=changeit --server_javabase=/usr/lib/jvm/java-11-openjdk-amd64 `

With this change, you can continue to add these JVM args as build options while continue to use other bazel configuration for build and test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -340,7 +340,7 @@ ${NODEUP}: ${BINDATA_TARGETS}
 
 .PHONY: bazel-crossbuild-dns-controller
 bazel-crossbuild-dns-controller:
-	bazel ${BAZEL_OPTIONS} build  ${BAZEL_CONFIG} --features=pure --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //dns-controller/...
+	bazel ${BAZEL_OPTIONS} build ${BAZEL_CONFIG} --features=pure --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //dns-controller/...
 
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -340,7 +340,7 @@ ${NODEUP}: ${BINDATA_TARGETS}
 
 .PHONY: bazel-crossbuild-dns-controller
 bazel-crossbuild-dns-controller:
-	bazel build ${BAZEL_CONFIG} --features=pure --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //dns-controller/...
+	bazel ${BAZEL_OPTIONS} build  ${BAZEL_CONFIG} --features=pure --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //dns-controller/...
```
johngmyers, line 343: nit:
```suggestion
	bazel ${BAZEL_OPTIONS} build ${BAZEL_CONFIG} --features=pure --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //dns-controller/...
```

and so on.

srikiz: sure, updated.

PR #9653, opened 2020-07-31
Bootstrap worker nodes using kops-controller

### LINKED ISSUE #1231: Update our kubelet (and other) certs for RBAC
We need to do the equivalent of https://github.com/kubernetes/kubernetes/pull/39020/commits

And check for any other changes

johngmyers:
The idea is for nodeup to generate private keys, send the public keys to kops-controller, and have kops-controller issue the necessary certificates.

~~This is far enough along to prove the concept. Posting to get early review (especially from @justinsb) while I plumb in the remaining pieces.~~

On AWS for Kubernetes 1.19 and later, worker nodes bootstrap by generating a private key and authenticating with a cloud-provider-specific mechanism to kops-controller. Kops-controller then verifies the authentication and issues a node-specific kubelet certificate.

This new bootstrap mechanism replaces both the mechanisms of reading a cluster-wide kubelet cert from the state store and node-authorizer. There is no option to use either of these previous mechanisms on Kubernetes 1.19+ in AWS.

Additionally, enables Kubernetes `Node` authorization for Kubernetes 1.19 and later, on all cloud providers.

Cribs from #8580. The authentication mechanism for AWS is based on that used by Vault. The code is structured to allow other cloud providers to provide their own authentication mechanisms.

Followup PRs will switch other worker node certs to use this mechanism.

Closes #1231


cmd/kops-controller/main.go
```diff
@@ -81,6 +85,29 @@ func main() {
 	}
 
 	ctrl.SetLogger(klogr.New())
+	if opt.Server != nil {
+		var verifier fi.Verifier
+		var err error
+		switch opt.Server.Provider {
+		case kops.CloudProviderAWS:
+			verifier, err = awsup.NewAWSVerifier()
```
justinsb, line 93: We do have a node identifier in kops-controller already.  We do also have the node verifier in `node-authorizer`.

Not a blocker, just that we can probably look at that code.

cmd/kops-controller/main.go
```diff
@@ -81,6 +85,29 @@ func main() {
...
+			if err != nil {
+				setupLog.Error(err, "unable to create verifier")
+				os.Exit(1)
+			}
+		}
+
+		srv, err := server.NewServer(&opt, verifier)
+		if err != nil {
+			setupLog.Error(err, "unable to create server")
+			os.Exit(1)
```
justinsb, line 103: I've been finding recently that I've been rewriting the kubebuilder main to call into a `func run() error` function immediately, just to avoid having to log & exit every time!

cmd/kops-controller/pkg/server/server.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return s, nil
+}
+
+func (s *Server) Start() error {
+	return s.server.ListenAndServeTLS(s.opt.Server.ServerCertificatePath, s.opt.Server.ServerKeyPath)
+}
+
+func (s *Server) bootstrap(w http.ResponseWriter, r *http.Request) {
+	if r.Body == nil {
```
justinsb, line 66: As we control both client & server ... how do you feel about GRPC?

nodeup/pkg/model/bootstrap_client.go
```diff
@@ -0,0 +1,59 @@
...
+package model
+
+import (
+	"k8s.io/kops/pkg/apis/kops"
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
+	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
+)
+
+// SecretBuilder writes secrets
```
justinsb, line 26: Nit: out of date comment

nodeup/pkg/model/bootstrap_client.go
```diff
@@ -0,0 +1,59 @@
...
+	if b.IsMaster || !b.UseKopsControllerForNodeBootstrap() {
+		return nil
+	}
+
+	var authenticator fi.Authenticator
+	var err error
+	switch kops.CloudProviderID(b.Cluster.Spec.CloudProvider) {
+	case kops.CloudProviderAWS:
+		authenticator, err = awsup.NewAWSAuthenticator()
+	}
```
justinsb, line 41: Probably should check `default` in the switch or otherwise handle the `authenticator == nil` case

nodeup/pkg/model/kops_controller.go
```diff
@@ -0,0 +1,85 @@
...
+		UID:   wellknownusers.KopsControllerID,
+		Shell: "/sbin/nologin",
+	})
+
+	issueCert := &nodetasks.IssueCert{
+		Name:           "kops-controller",
+		Signer:         fi.CertificateIDCA,
+		Type:           "server",
+		Subject:        nodetasks.PKIXName{CommonName: "kops-controller"},
+		AlternateNames: []string{b.Cluster.Spec.MasterInternalName},
```
justinsb, line 64: BTW I don't know if you've seen the golang 1.15 change - AIUI it will no longer accept the hostname in CN; it must be in AlternateNames.

pkg/apis/kops/model/features.go
```diff
@@ -0,0 +1,26 @@
...
+
+package model
+
+import (
+	"k8s.io/kops/pkg/apis/kops"
+)
+
+// UseKopsControllerForNodeBootstrap is true if nodeup should use kops-controller for bootstrapping.
+func UseKopsControllerForNodeBootstrap(cluster *kops.Cluster) bool {
+	return kops.CloudProviderID(cluster.Spec.CloudProvider) == kops.CloudProviderAWS && cluster.IsKubernetesGTE("1.19")
```
justinsb, line 25: I guess this is why we don't need to check the default/nil case; I think it's probably still a good idea there though!

upup/pkg/fi/cloudup/awsup/aws_authenticator.go
```diff
@@ -0,0 +1,62 @@
...
+	stsRequest, _ := a.sts.GetCallerIdentityRequest(nil)
+
+	stsRequest.HTTPRequest.Header.Add("X-Kops-Request-SHA", base64.RawStdEncoding.EncodeToString(sha[:]))
+	err := stsRequest.Sign()
+	if err != nil {
+		return "", err
+	}
+
+	headers, _ := json.Marshal(stsRequest.HTTPRequest.Header)
+	return AWSAuthenticationTokenPrefix + base64.StdEncoding.EncodeToString(headers), nil
```
justinsb, line 61: Would it be easier just to put this in to an "evidence" section of the request, rather than encoding them?  It would also be more extensible, I think.

upup/pkg/fi/cloudup/awsup/aws_verifier.go
```diff
@@ -0,0 +1,149 @@
...
+	RequestId string `xml:"RequestId"`
+}
+
+func (a awsVerifier) VerifyToken(token string, body []byte) (string, error) {
+	if !strings.HasPrefix(token, AWSAuthenticationTokenPrefix) {
+		return "", fmt.Errorf("incorrect authorization type")
+	}
+	token = strings.TrimPrefix(token, AWSAuthenticationTokenPrefix)
+
+	// We rely on the client and server using the same version of the same STS library.
```
justinsb, line 96: Should we instead send the body of the STS request?

upup/pkg/fi/cloudup/awsup/aws_verifier.go
```diff
@@ -0,0 +1,149 @@
...
+		return "", fmt.Errorf("incorrect SHA")
+	}
+
+	requestBytes, _ := ioutil.ReadAll(stsRequest.Body)
+	_, _ = stsRequest.Body.Seek(0, io.SeekStart)
+	if stsRequest.HTTPRequest.Header.Get("Content-Length") != strconv.Itoa(len(requestBytes)) {
+		return "", fmt.Errorf("incorrect content-length")
+	}
+
+	// TODO - implement retry?
```
justinsb, line 124: I suspect our client needs to retry anyway, so there might not be much reason for us to retry here.

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+	httpReq.Header.Set("Authorization", token)
+
+	resp, err := b.client.Do(httpReq)
+	if err != nil {
+		return err
+	}
+
```
justinsb, line 117: I'm likely missing something ... should we just `defer resp.Body.Close()` here?

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
```diff
@@ -0,0 +1,142 @@
...
+	if resp.StatusCode != http.StatusOK {
+		detail := ""
+		if resp.Body != nil {
+			scanner := bufio.NewScanner(resp.Body)
```
justinsb, line 121: Any reason not to just `ioutil.ReadAll` ?

justinsb:
Really great stuff.  Mostly minor feedback.

Topics that may be more interesting:
* Can we / should we reuse any of the existing validation code
* GRPC vs REST
* Should we make the request more extensible, with a flexible "evidence" section that we basically throw lots of fields into.  This is the approach taken by node-authorizer, which supports some interesting verification policies.  But generally it might just be easier anyway!

cmd/kops-controller/pkg/server/server.go
johngmyers: I don't see the value of adding the complexity.

upup/pkg/fi/cloudup/awsup/aws_authenticator.go
johngmyers: There's a secure hash of the body going into the credential, so we can't put the credential in the body.

upup/pkg/fi/cloudup/awsup/aws_verifier.go
johngmyers: Vault does that, but they have to support arbitrary clients. I took the ability to require the client to use the same version of the AWS library to reduce the risk of allowing attackers to use kops-controller as a pass-through proxy to STS.

upup/pkg/fi/cloudup/awsup/aws_verifier.go
johngmyers: Nodeup will retry, it's just a question of how long transient errors delay node startup.

cmd/kops-controller/main.go
johngmyers: The verifier needs to authenticate the request data as having come from an instance that was created using the spec of one of the cluster's instancegroups. Bonus points for authenticating the particular instance group or even the particular instance.

`node-authorizer` makes two checks, neither of which are sufficient, even in combination. First, it verifies an instance identity document provided by the instance. This document is long-lived and replayable: it is not tied to any data in the rest of the request. The only thing tying the data in the request to the instance is a check that the source IP address of the incoming request matches that of the instance, which is not a particularly strong mechanism.

nodeup/pkg/model/kops_controller.go
johngmyers: Since we aren't putting hostnames in the CN, this doesn't seem to be a problem.

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
johngmyers: We only want the first line.

nodeup/pkg/model/bootstrap_client.go
johngmyers: fixed

nodeup/pkg/model/bootstrap_client.go
johngmyers: fixed

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
johngmyers: fixed

johngmyers:
* aws_verifier might want to verify the received arn's partition matches that of its own identity.
* aws_verifier assumes the arn it receives is for an assumed-role. It does in my (limited) testing, but in case there's some region or situation where it isn't we might want to have an escape hatch to the ten-year-kubelet-cert mechanism.
* I haven't tested non-`nil` values of `Spec.IAM.Profile`. I doubt we have coverage of that feature. What is the use case for that feature? Where should we document that anything sharing such such a profile would be able to register as a node?
* I think we should remove support for node-authorizer in k8s 1.19+ for the non-AWS cloud providers as well.

johngmyers:
I wonder if now there's a circular dependency for most CNIs now, with CNIs depending on a worker node joining and worker node joining depending on the CNI.
/retest

olemarkus:
I don't think so. If you install a CNI manually, you wait for the masters to be available, but NOT READY, and then apply the CNI spec. The worker nodes join once the masters are ready. Just looking quickly at this, it looks like dns-controller is not updating the DNS and dns-controller uses host network.

cmd/kops-controller/pkg/server/server.go
justinsb: Fair enough ... it can be helpful in that it effectively takes care of the JSON marshalling/unmarshalling, along with authentication.  But I think it's reasonable to argue that it's just not that hard, and we couldn't use the built-in authentication mechanisms.  (And the really great features come with bigger scale than we're talking about for this API.)

cmd/kops-controller/pkg/server/server.go
```diff
@@ -0,0 +1,193 @@
...
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"hash/fnv"
+	"io/ioutil"
+	"net/http"
+	"runtime/debug"
+	"time"
+
+	"github.com/gorilla/mux"
```
justinsb, line 32: If we're on the topic of unneeded complexity ... ;-)

nodeup/pkg/model/kops_controller.go
justinsb: Phew!

pkg/apis/nodeup/bootstrap.go
```diff
@@ -0,0 +1,33 @@
...
+package nodeup
+
+const BootstrapAPIVersion = "bootstrap.kops.k8s.io/v1alpha1"
+
+// BootstrapRequest is a request from nodeup to kops-controller for bootstrapping a node.
+type BootstrapRequest struct {
+	// APIVersion defines the versioned schema of this representation of a request.
+	APIVersion string `json:"apiVersion"`
+	// Certs are the requested certificates and their respective public keys.
+	Certs map[string]string `json:"certs"`
```
justinsb, line 26: You _might_ want to make this `Certs map[string]*CertRequest` or `Certs []*CertRequest` just to give you more extensibility

pkg/apis/nodeup/bootstrap.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// BootstrapRespose is a response to a BootstrapRequest.
+type BootstrapResponse struct {
+	// Certs are the issued certificates.
+	Certs map[string]string
```
justinsb, line 32: Ditto here - `map[string]*Cert` doesn't cost much now, might avoid version contortions later

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
justinsb: OK; not a problem.  Unless there's something that we actively don't want to log, I'd personally just log the whole thing - little cost, potential benefit for debugging.  But your call.

upup/pkg/fi/cloudup/awsup/aws_authenticator.go
```diff
@@ -0,0 +1,62 @@
...
+		sts: sts.New(sess),
+	}, nil
+}
+
+func (a awsAuthenticator) CreateToken(body []byte) (string, error) {
+	sha := sha256.Sum256(body)
+
+	stsRequest, _ := a.sts.GetCallerIdentityRequest(nil)
+
+	stsRequest.HTTPRequest.Header.Add("X-Kops-Request-SHA", base64.RawStdEncoding.EncodeToString(sha[:]))
```
justinsb, line 54: Ah I see - this is how we sign the body data!  I think having a comment here would help future readers to understand.

upup/pkg/fi/cloudup/awsup/aws_verifier.go
justinsb: Ah - interesting point.  Does this introduce a challenge for upgrading the AWS SDK though?  If there's master vs node skew then I am worried that either old-version nodes or new nodes won't be able to join?

upup/pkg/fi/authenticate.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Authenticator generates authentication credentials for requests.
+type Authenticator interface {
+	CreateToken(body []byte) (string, error)
+}
+
+// VerifyResult is the result of a successfully verified request.
+type VerifyResult struct {
+	// Instance is the hostname of the instance.
+	Instance string
```
justinsb, line 27: We could name this `NodeName`, as I think that's what we've actually verified - it's approved to use this node name but not others.

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
```diff
@@ -0,0 +1,193 @@
...
+		Host:   net.JoinHostPort(c.Cluster.Spec.MasterInternalName, strconv.Itoa(wellknownports.KopsControllerPort)),
+		Path:   "/bootstrap",
+	}
+	httpReq, err := http.NewRequest("POST", bootstrapUrl.String(), bytes.NewReader(reqBytes))
+	if err != nil {
+		return nil, err
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+
+	token, err := b.Authenticator.CreateToken(reqBytes)
```
justinsb, line 156: Thanks for explaining the signature challenge.  I do understand, but ... can we implement this on GCE / OpenStack / Bare-Metal?  I know it's not _as_ simple, but could we do something like this:

```
type SignedBootstrapRequest struct {
  Request *BootstrapRequest `json:"request"` 
  AWSSecureTokenService *AWSSecureTokenService `json:"awsSecureTokenService"`
  // GCESignature *GCESignature
}

type AWSSecureTokenService struct {
  Headers map[string]string `json:"headers"`
}

```

It feels more extensible (for other clouds, possible for AWS also)

justinsb:
So I like this PR a lot!  My only reservation is around making the requests more structured, to allow for evolution and support for other clouds.

re deprecating node-authorizer, let's get this going first, then check with users!

cmd/kops-controller/pkg/server/server.go
johngmyers: Cribbed from node-authorizer. I can take it out.

upup/pkg/fi/cloudup/awsup/aws_verifier.go
johngmyers: I didn't think supporting version skew was a requirement. Old nodes would only be an issue if they were created before the ASG's spec was updated yet didn't call `kops-controller` until after the first master was updated. The old node would be cleaned out by rolling update (though the cluster validation failure would impede rolling update from starting). New nodes would fail only until the first master is updated.

I do have an idea of making rolling update ignore non-ready nodes that are in nodes-role instance groups other than the one being updated.

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
johngmyers: I believe this interface is implementable by other clouds, assuming those clouds have some way to authenticate an instance. The interface I defined seems to me to have been reduced to the essentials: the provider-specific client gets the serialized body and produces an authenticator. The common code transfers the authenticator in the `Authorization:` header. The provider-specific server gets the serialized body and the authenticator and gets to say what the node name is or reject the request.

I'm presuming that multi-provider clusters are well out of scope.

cmd/kops-controller/main.go
johngmyers: The structured logging, such as in the handling of `addNodeController`, makes it tricky to use that pattern.

cmd/kops-controller/pkg/server/server.go
johngmyers: Removed

upup/pkg/fi/cloudup/awsup/aws_authenticator.go
johngmyers: Comment added.

upup/pkg/fi/authenticate.go
johngmyers: Renamed field.

pkg/apis/nodeup/bootstrap.go
johngmyers: It depends on whether we need to support bootstrapping nodes created with a different version, which I discuss below.

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
justinsb: Agree that multi-provider clusters should be out of scope.  Even if we have a central management cluster, I think we would want node bring-up to be done relatively locally.  I could imagine forwarding the request to another service, but I imagine that would essentially be wrapping the current request anyway.

I can see the authorization header as isomorphic to what I described - we're encoding the signature data (with the body) into json, base64 encoding it, and then putting that in the authorization header.  I'm really not sure if that is simpler, but it's likely tied up in the skew problem ... I'll ponder!

pkg/apis/nodeup/bootstrap.go
justinsb: Did you post this discussion?  I'm not seeing it.

johngmyers: It's in the block of comments on aws_verifier.go line 129.

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
johngmyers: Putting the token in the header simplifies the task of getting the signature to include the request, since you can then just take a secure hash of the entire body. Putting it into the `Authorization:` header means anything in the middle is likely to treat it as confidential.

johngmyers:
Okay, the e2e failure https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/9653/pull-kops-e2e-kubernetes-aws/1292531001418846211 is interesting. Per https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/9653/pull-kops-e2e-kubernetes-aws/1292531001418846211/artifacts/ip-172-20-33-248.ap-northeast-2.compute.internal/kops-controller-nfrbv.log STS didn't like the ap-northeast-2 region.

upup/pkg/fi/cloudup/awsup/aws_authenticator.go
```diff
@@ -0,0 +1,64 @@
...
+const AWSAuthenticationTokenPrefix = "x-aws-sts "
+
+type awsAuthenticator struct {
+	sts *sts.STS
+}
+
+var _ fi.Authenticator = &awsAuthenticator{}
+
+func NewAWSAuthenticator() (fi.Authenticator, error) {
+	config := aws.NewConfig().WithCredentialsChainVerboseErrors(true)
```
rifelpet, line 39: I wonder if the cluster's region needs to be explicitly provided here. Either that or there might be a setting that needs to be adjusted in the prow AWS accounts? https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html

johngmyers: Looks like that might have been it.

upup/pkg/fi/cloudup/awsup/aws_verifier.go
justinsb: So I think the scenario we be most mindful of is the master being ahead of the nodes.  i.e. we update the AWS SDK and somehow the body changes, we roll that out for the control-plane, but the nodes aren't yet updated.  As you point out, old node configurations would fail to join in this scenario.  Kubernetes itself tries very hard to ensure that the masters can be a minor version ahead of the nodes (indefinitely); we don't _have_ to support that but we should be aware of this.

Do we have an example of what's in the body?  How likely is it to change?

johngmyers: At the time we run `kops update cluster --yes` (or `terraform apply`) the node ASGs will be updated. All new instances created at that point will be on the new version of the library. Similarly, all nodes that have joined by the time the first master is updated will continue to be in the cluster. The only issue would be a worker instance that is created before the `kops update cluster --yes` but takes longer to join the cluster than the time it takes for the first control plane node to get updated and apply the new kops-controller manifest.

johngmyers: The body usually has `QWN0aW9uPUdldENhbGxlcklkZW50aXR5JlZlcnNpb249MjAxMS0wNi0xNQ==` which is the
  base64 encoding of `Action=GetCallerIdentity&Version=2011-06-15`. It doesn't seem likely to change often

justinsb: OK, that sounds reasonable.  We could create a unit test to check that the body is this value; if the unit test ever failed we could deal with it (e.g constructing all allowed forms and seeing if any of them match the sha), but as you point out it seems unlikely.

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
justinsb: I'm OK to proceed on this basis;  if we do find that on another cloud it's much trickier to put it into the header (e.g. GCE's TPM flow) then we might have to put it into the body for that cloud - is that reasonable?

johngmyers: I think that's reasonable. If that were to happen, we'd probably do something like change the interfaces to return a modified body.

johngmyers:
I think I'm about ready to remove the WIP on this. I'm going to bring it up in the next Office Hours since it's a pretty significant change.

justinsb:
Yes - I agree with removing WIP & bringing it up it office hours.

justinsb:
/retest

One prow networking failure before we even touch our code, the other test looks like it actually passed

justinsb:
This LGTM ... I think we agreed to merge and aim for 1.19 (instead of 1.20), so...

/approve
/lgtm

PR #9663, opened 2020-08-01
Prefer nodes with "master" role for Canal Typha pods

hakman:
Adding the fix from #9609 to Canal also.


hakman:
Thanks @johngmyers

PR #9666, opened 2020-08-02
Add sslPolicy for NLB to change listener's security policy

### LINKED ISSUE #9633: Able to change api aws elb policy with annotation
Please allow to change api's elb policy with annotation. Aws uses too old policy as default with TLS 1.0 allowed.

It would be good if we could change this policy with  service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy

FrankYang0529:
closes  #9633

### DIFF
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -389,6 +389,8 @@ type LoadBalancerAccessSpec struct {
 	UseForInternalApi bool `json:"useForInternalApi,omitempty"`
 	// SSLCertificate allows you to specify the ACM cert to be used the LB
 	SSLCertificate string `json:"sslCertificate,omitempty"`
+	// SSLPolicy allows you to overwrite the LB listener's Security Policy
+	SSLPolicy *string `json:"sslPolicy,omitempty"`
 	// CrossZoneLoadBalancing allows you to enable the cross zone load balancing
 	CrossZoneLoadBalancing *bool `json:"crossZoneLoadBalancing,omitempty"`
 }
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -34,6 +34,7 @@ func awsValidateCluster(c *kops.Cluster) field.ErrorList {
 	if c.Spec.API != nil {
 		if c.Spec.API.LoadBalancer != nil {
 			allErrs = append(allErrs, awsValidateAdditionalSecurityGroups(field.NewPath("spec", "api", "loadBalancer", "additionalSecurityGroups"), c.Spec.API.LoadBalancer.AdditionalSecurityGroups)...)
+			allErrs = append(allErrs, awsValidateSSLPolicy(field.NewPath("spec", "api", "loadBalancer", "sslPolicy"), c.Spec.API.LoadBalancer)...)
 		}
 	}
 
@@ -141,4 +142,19 @@ func awsValidateMixedInstancesPolicy(path *field.Path, spec *kops.MixedInstances
 	errs = append(errs, IsValidValue(path.Child("spotAllocationStrategy"), spec.SpotAllocationStrategy, kops.SpotAllocationStrategies)...)
 
 	return errs
+}
+
+func awsValidateSSLPolicy(fieldPath *field.Path, spec *kops.LoadBalancerAccessSpec) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if spec.SSLPolicy != nil {
+		if spec.Class != kops.LoadBalancerClassNetwork {
+			allErrs = append(allErrs, field.Forbidden(fieldPath, "sslPolicy should be specified with Network Load Balancer"))
+		}
+		if spec.SSLCertificate == "" {
+			allErrs = append(allErrs, field.Forbidden(fieldPath, "sslPolicy should not be specified without SSLCertificate"))
+		}
+	}
+
+	return allErrs
 }
--- a/pkg/model/awsmodel/api_loadbalancer.go
+++ b/pkg/model/awsmodel/api_loadbalancer.go
@@ -119,11 +119,16 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 		if lbSpec.SSLCertificate != "" {
 			listeners["443"].SSLCertificateID = lbSpec.SSLCertificate
 			nlbListeners[0].Port = 8443
-			nlbListeners = append(nlbListeners, &awstasks.NetworkLoadBalancerListener{
+
+			nlbListener := &awstasks.NetworkLoadBalancerListener{
 				Port:             443,
 				TargetGroupName:  b.NLBTargetGroupName("tls"),
 				SSLCertificateID: lbSpec.SSLCertificate,
-			})
+			}
+			if lbSpec.SSLPolicy != nil {
+				nlbListener.SSLPolicy = *lbSpec.SSLPolicy
+			}
+			nlbListeners = append(nlbListeners, nlbListener)
 		}
 
 		if lbSpec.SecurityGroupOverride != nil {
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -2,7 +2,7 @@
 
 The `Cluster` resource contains the specification of the cluster itself.
 
-The complete list of keys can be found at the [Cluster](https://pkg.go.dev/k8s.io/kops/pkg/apis/kOps#ClusterSpec) reference page.
+The complete list of keys can be found at the [Cluster](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#ClusterSpec) reference page.
 
 On this page, we will expand on the more important configuration keys.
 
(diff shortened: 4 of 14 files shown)


rifelpet:
/ok-to-test

Thanks for adding this! I haven't looked at the ELB API behavior closely but we should make sure that when someone removes listenerPolicies from their ClusterSpec that the API ELB is updated to use AWS' default listener policy. We'll also want to add terraform and cloudformation support but that can happen in followup PRs.

FrankYang0529:
@rifelpet Thanks for reviewing! I have two questions about generating terraform and cloudformation.

* For terraform, it uses another resource name for [listener's policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/load_balancer_listener_policy). Is it correct to only add `RenderTerraform` and ` TerraformLink` functions for `LoadBalancerListener`, so we can have corresponding terraform resource?
* For cloudformation, I am not sure why we only generate [TCP listeners](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/awstasks/load_balancer.go#L861-L866)? Cloudformation has `PolicyNames` field in [listener](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-elb-listener.html#cfn-ec2-elb-listener-protocol). So if we distinguish whether load balancer port is 443 in `RenderCloudformation`, we can add corresponding policy names.

rifelpet:
* We should be able to render two terraform resources from one RenderTerraform function, [heres](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/alitasks/eip_natgateway_association.go#L148-L161) an example. You'll just need an additional terraform type with all the fields necessary for the additional resource type. 

* Regarding cloudformation, my guess is that Kops' ACM Certificate support wasnt extended to include cloudformation, so we may be hardcoding the TCP listener there. If that's the case then it would be good to support both ACM certs and this new security policy in cloudformation but again, this can be a followup PR.

Also we'll want to add this to an integration test which will test both the terraform and cloudformation outputs. Add the new field to the ClusterSpec in the two `in-*.yaml` files [here](https://github.com/kubernetes/kops/tree/master/tests/integration/update_cluster/complex), and run `./hack/update-expected.sh` and you should see the terraform and CF files in the same directory be updated to match.

olemarkus:
Small one from me: Add a line about this in the cluster spec docs (`docs/cluster_spec.md`)

FrankYang0529:
Updated it. Thank you!

upup/pkg/fi/cloudup/awstasks/load_balancer.go
```diff
@@ -529,6 +536,24 @@ func (_ *LoadBalancer) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *LoadBalan
 			return fmt.Errorf("Unable to find newly created ELB %q", loadBalancerName)
 		}
 		e.HostedZoneId = lb.CanonicalHostedZoneNameID
+
+		// Set ELB listener's Security Policy
+		if l, ok := e.Listeners["443"]; ok && len(l.PolicyNames) != 0 {
```
rifelpet, line 541: Rather than looking for the listener on port 443 perhaps it would be more flexible to perform this logic on any listener that has a Protocol of `SSL`?

FrankYang0529: We only apply security policies when specifying an ACM certificate. However, we set the listener with 443 port when the ACM certificate is existent. So I only set security policies for the listener with 443 port here.

https://github.com/kubernetes/kops/blob/d90c90cfa41bc8b161dc5475c3280fb0f133c12e/pkg/model/awsmodel/api_loadbalancer.go#L112-L114

FrankYang0529:
> I'm wondering if Kops should try to reconcile this as well, creating the load-balancer-policy if it does not already exist, based on the API Field being used as the Rerence-Security-Policy. Once it is ensured to exist for the API ELB then Kops can `SetLoadBalancerPoliciesOfListener`. Thoughts?


For PolicyNotFound error, I am not sure why we can't set predefined policies on AWS. There is the same issue at https://github.com/boto/boto3/issues/1438.

Probably, it's better to let users create policies by themselves. Because we can't even distinguish whether policies are existent by `aws elb describe-load-balancer-policies`. We can note this in the document.

BTW, I added API validation for `listenerPolicies`.

rifelpet:
The downside of requiring that users define the policies themselves is that setting up a new cluster with a custom policy becomes a 3-step process involving kops and an external tool (aws cli for example):
1. `kops create cluster --yes` to create the API ELB
2. `aws elb create-load-balancer-policy`
3. update the ClusterSpec to reference the new policy and `kops update cluster --yes`

It would be great if Kops could handle it automatically so that a user could create a new cluster with a custom policy using 1 kops command.

Stepping back, we have an [open PR](https://github.com/kubernetes/kops/pull/9011) to support NLBs for the api server and setting the listener security policies on NLBs is [much more straight forward](https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#CreateListenerInput). Maybe we wait for that PR to land and only add support for setting policies on NLBs? That would simplify the new API fields and keep cluster creation a 1-step process. At the very least we should ensure that the new API fields will easily support defining policies for both NLBs and ELBs.

FrankYang0529:
Sounds good! Let's wait for https://github.com/kubernetes/kops/pull/9011 to land and only add support for setting policies on NLBs.

rifelpet:
@FrankYang0529 The NLB PR has been merged, are you still willing to update this to support NLB Listener SSLPolicies?

FrankYang0529:
@rifelpet Yes, I will update it this week.

FrankYang0529:
@rifelpet , I have updated it. We can support NLB Listener SSL Policy now.

pkg/apis/kops/cluster.go
```diff
@@ -389,6 +389,8 @@ type LoadBalancerAccessSpec struct {
 	UseForInternalApi bool `json:"useForInternalApi,omitempty"`
 	// SSLCertificate allows you to specify the ACM cert to be used the LB
 	SSLCertificate string `json:"sslCertificate,omitempty"`
+	// SSLPolicy allows you to overwrite ELB listener's Security Policy
```
rifelpet, line 392: along with the same change in `v1alpha2/cluster.go` and `make crds`:
```suggestion
	// SSLPolicy allows you to overwrite the LB listener's Security Policy
```

pkg/apis/kops/validation/aws.go
```diff
@@ -142,3 +143,13 @@ func awsValidateMixedInstancesPolicy(path *field.Path, spec *kops.MixedInstances
 
 	return errs
 }
+
+func awsValidateSSLPolicy(fieldPath *field.Path, spec *kops.LoadBalancerAccessSpec) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if spec.SSLPolicy != nil && spec.SSLCertificate == "" {
+		allErrs = append(allErrs, field.Forbidden(fieldPath, "sslPolicy should not be specified without SSLCertificate"))
```
rifelpet, line 151: given that we're only using sslPolicy on network load balancers, we should also ensure that `spec.Class` is set to `kops.LoadBalancerClassNetwork`

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -117,6 +117,9 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 		if lbSpec.SSLCertificate != "" {
 			listeners["443"].SSLCertificateID = lbSpec.SSLCertificate
 			nlbListeners["443"].SSLCertificateID = lbSpec.SSLCertificate
+			if lbSpec.SSLPolicy != nil {
```
rifelpet, line 120: Just a heads up that this will conflict with 
https://github.com/kubernetes/kops/pull/10157/files#diff-c72a0990bb603f634a945a4410886578bb02cbd7db602682f1c5ada942d6f012R121

I'm not sure which we'll merge first but resolving the conflict shouldn't be a big deal regardless.

rifelpet (commented):
This is much simpler now that it supports NLBs rather than classic load balancers :) nice work!

We'll want to add terraform and cloudformation support for this too. It should just be a matter of adding the new field here and setting its value below:

https://github.com/kubernetes/kops/blob/ecea47790b359069ff6446410a7d98db8c24e388/upup/pkg/fi/cloudup/awstasks/network_load_balancer.go#L645-L651

and cloudformation has its own struct later in the file too. (docs for [terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener#argument-reference) and [cloudformation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listener.html#aws-resource-elasticloadbalancingv2-listener-syntax))


If you want I can do that in a followup PR.

pkg/apis/kops/cluster.go
FrankYang0529: Updated it. Thank you.

pkg/model/awsmodel/api_loadbalancer.go
FrankYang0529: Resolved it. Thank you.

pkg/apis/kops/validation/aws.go
FrankYang0529: Added it. Thank you.

FrankYang0529:
@rifelpet Thanks for reviewing. I updated the code and added `SSLPolicy` in terraform and cloudformation.

rifelpet:
looks good, one last suggestion:

Can you add an example `sslPolicy`to the two `tests/integration/update_cluster/complex/in-*.yaml` files? Then run `./hack/update-expected.sh` and the kubernetes.tf and cloudformation.json in the same directory should get updated.

FrankYang0529:
@rifelpet, I have added `sslPolicy` in two yaml files. However, I am not sure why I can't run tests successfully for `k8s.io/kops/cmd/kops`.  Could you show me how to fix the test error? Thank you.

rifelpet:
The verify-cloudformation failure is unrelated and was fixed by #10256. This PR's next invocation of that job should pass.

The TestLifecycleComplex failures are because the network load balancer task isn't recognizing the new field when describing the existing listeners in AWS.

Fixing it should just be a matter of setting the `actualListener`'s value here, like we do with `Port`:

https://github.com/kubernetes/kops/blob/a0c89250e43a27460e6f5e1af2a805c9f5a5300a/upup/pkg/fi/cloudup/awstasks/network_load_balancer.go#L360-L372

FrankYang0529:
@rifelpet Thank you for your elaboration. I know how it works now! Also, I have updated the code and I can run tests successfully.

rifelpet:
looks great, thanks for sticking with this!

/lgtm
/approve

PR #9667, opened 2020-08-02
Support authentication helper for kubectl

justinsb:
We create a simple exec plugin command which can create and renew
short-lived admin credentials on the fly, essentially leveraging the
security of the underlying cloud credentials.

### DIFF
--- a/cmd/kops/export_kubecfg.go
+++ b/cmd/kops/export_kubecfg.go
@@ -144,7 +144,6 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 			keyStore,
 			secretStore,
 			&commands.CloudDiscoveryStatusStore{},
-			buildPathOptions(options),
 			options.admin,
 			options.user,
 			options.internal,
@@ -154,7 +153,7 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 			return err
 		}
 
-		if err := conf.WriteKubecfg(); err != nil {
+		if err := conf.WriteKubecfg(buildPathOptions(options)); err != nil {
 			return err
 		}
 	}
--- a/cmd/kops/export_kubecfg.go
+++ b/cmd/kops/export_kubecfg.go
@@ -60,6 +60,9 @@ type ExportKubecfgOptions struct {
 	admin          time.Duration
 	user           string
 	internal       bool
+
+	// UseKopsAuthenticationPlugin controls whether we should use the kops auth helper instead of a static credential
+	UseKopsAuthenticationPlugin bool
 }
 
 func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
@@ -85,6 +88,7 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().Lookup("admin").NoOptDefVal = kubeconfig.DefaultKubecfgAdminLifetime.String()
 	cmd.Flags().StringVar(&options.user, "user", options.user, "add an existing user to the cluster context")
 	cmd.Flags().BoolVar(&options.internal, "internal", options.internal, "use the cluster's internal DNS name")
+	cmd.Flags().BoolVar(&options.UseKopsAuthenticationPlugin, "auth-plugin", options.UseKopsAuthenticationPlugin, "use the kops authentication plugin")
 
 	return cmd
 }
@@ -135,7 +139,17 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 			return err
 		}
 
-		conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, buildPathOptions(options), options.admin, options.user, options.internal)
+		conf, err := kubeconfig.BuildKubecfg(
+			cluster,
+			keyStore,
+			secretStore,
+			&commands.CloudDiscoveryStatusStore{},
+			buildPathOptions(options),
+			options.admin,
+			options.user,
+			options.internal,
+			f.KopsStateStore(),
+			options.UseKopsAuthenticationPlugin)
 		if err != nil {
 			return err
 		}
--- a/pkg/commands/helpers/kubectl_auth.go
+++ b/pkg/commands/helpers/kubectl_auth.go
@@ -33,7 +33,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/util/homedir"
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
 	"k8s.io/kops/cmd/kops/util"
 	"k8s.io/kops/pkg/commands/commandutils"
 	"k8s.io/kops/pkg/pki"
--- a/pkg/commands/helpers/kubectl_auth.go
+++ b/pkg/commands/helpers/kubectl_auth.go
@@ -1,0 +1,280 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helpers
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"crypto/x509/pkix"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/util/homedir"
+	"k8s.io/klog"
+	"k8s.io/kops/cmd/kops/util"
+	"k8s.io/kops/pkg/commands/commandutils"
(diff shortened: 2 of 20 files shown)


justinsb:
cc @johngmyers

pkg/commands/helpers/kubectl_auth.go
```diff
@@ -0,0 +1,270 @@
...
+	if cluster == nil {
+		return nil, fmt.Errorf("cluster not found %q", options.ClusterName)
+	}
+
+	keyStore, err := clientset.KeyStore(cluster)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get cluster keystore: %v", err)
+	}
+
+	req := pki.IssueCertRequest{
```
justinsb, line 241: This is based on #9593

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -127,7 +127,17 @@ func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.Se
...
+	}
 
+	if useKopsAuthenticationPlugin {
+		b.AuthenticationExec = []string{
+			"kops",
+			"helpers",
+			"kubectl-auth",
+			"--cluster=" + clusterName,
+			"--state=" + kopsStateStore,
+			"--api-version=" + "v1alpha1",
```
justinsb, line 139: We might be able to default to v1beta1 now

pkg/commands/helpers/kubectl_auth.go
```diff
@@ -0,0 +1,270 @@
...
+	"k8s.io/kops/cmd/kops/util"
+	"k8s.io/kops/pkg/commands/commandutils"
+	"k8s.io/kops/pkg/pki"
+	"k8s.io/kops/pkg/rbac"
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kubectl/pkg/util/i18n"
+)
+
+var (
+	kubectlAuthShort = i18n.T(`Helper for kubectl authentication integration`)
```
johngmyers, line 45:
```suggestion
	kubectlAuthShort = i18n.T(`kubectl authentication plugin`)
```
?

pkg/commands/helpers/kubectl_auth.go
```diff
@@ -0,0 +1,270 @@
...
+			err := RunKubectlAuthHelper(ctx, f, out, options)
+			if err != nil {
+				commandutils.ExitWithError(err)
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&options.APIVersion, "api-version", options.APIVersion, "version of client.authentication.k8s.io schema in use")
+	cmd.Flags().StringVar(&options.ClusterName, "cluster", options.ClusterName, "cluster to target")
+	cmd.Flags().DurationVar(&options.Duration, "validity", options.Duration, "attempt to issue a credential with this validity")
```
johngmyers, line 85:
```suggestion
	cmd.Flags().DurationVar(&options.Duration, "validity", options.Duration, "lifetime of the credential to issue")
```

pkg/commands/helpers/kubectl_auth.go
```diff
@@ -0,0 +1,270 @@
...
+		execCredential.APIVersion = "client.authentication.k8s.io/v1beta1"
+
+	default:
+		return fmt.Errorf("api-version %q is not supported", options.APIVersion)
+	}
+
+	cacheFilePath := cacheFilePath(f.KopsStateStore(), options.ClusterName)
+	cached, err := loadCachedExecCredential(cacheFilePath)
+	if err != nil {
+		klog.Warningf("cached credential %q was not valid: %v", cacheFilePath, err)
```
johngmyers, line 115: Not sure this and the wrong API version count as warnings.

pkg/commands/helpers/kubectl_auth.go
```diff
@@ -0,0 +1,270 @@
...
+		return fmt.Errorf("error writing to stdout: %v", err)
+	}
+
+	if !isCached {
+		if err := os.MkdirAll(filepath.Dir(cacheFilePath), 0755); err != nil {
+			klog.Warningf("failed to make cache directory for %q: %v", cacheFilePath, err)
+		}
+		if err := ioutil.WriteFile(cacheFilePath, b, 0600); err != nil {
+			klog.Warningf("failed to write cache file %q: %v", cacheFilePath, err)
+		}
```
johngmyers, line 151: Ah, so you have to cache these in a separate file. Our authentication plugin tries to cache them in the kubecfg, which seems to result in failure whenever it has to update the token.

pkg/commands/helpers/kubectl_auth.go
johngmyers: looks good. Might want to put what we think the userid is in the `CommonName`.

pkg/commands/helpers/kubectl_auth.go
rifelpet: I'd try to come up with a better flag name than `--validity`. validity on its own doesn't suggest a duration, merely that something is valid. Maybe just `--duration` ?

johngmyers: `--lifetime` would be my suggestion

pkg/commands/helpers/kubectl_auth.go
justinsb: Yes.  I thought it was safer to put them in another file - one of the things I don't like about the GCE plugin is that your token ends up in your kubeconfig and is easily accidentally shared.  Putting them in another file also means that (in theory) we could also store them in a more secure location (e.g. https://github.com/zalando/go-keyring).  But it is another file which means a bigger surface we now have to think about.

How does your authentication plugin write back to the kubeconfig?  Does it use the exec plugin mechanism, and is there a trick I'm missing?

johngmyers: It uses `k8s.io/client-go/tools/clientcmd/api/ModifyConfig()` to write the new token and expiration to the kubeconfig. That unfortunately doesn't affect the current kubectl process, so the first kubectl command usually fails after updating the kubecfg and the next one works.

cmd/kops/export_kubecfg.go
```diff
@@ -78,6 +81,7 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().BoolVar(&options.all, "all", options.all, "export all clusters from the kops state store")
 	cmd.Flags().BoolVar(&options.admin, "admin", options.admin, "export the cluster admin user and add it to the context")
 	cmd.Flags().StringVar(&options.user, "user", options.user, "add an existing user to the cluster context")
+	cmd.Flags().BoolVar(&options.UseKopsAuthenticationPlugin, "use-plugin", options.UseKopsAuthenticationPlugin, "use kops authentication plugin, instead of exporting a static credential")
```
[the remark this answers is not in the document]
justinsb: Thanks!

cmd/kops/main.go
```diff
@@ -28,6 +27,5 @@ func main() {
 // exitWithError will terminate execution with an error result
 // It prints the error to stderr and exits with a non-zero exit code
 func exitWithError(err error) {
```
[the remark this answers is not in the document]
justinsb: Agreed, I just wanted to keep the PRs thematically separate

pkg/commands/helpers/kubectl_auth.go
justinsb: Yes, you're right!

pkg/commands/helpers/kubectl_auth.go
justinsb: Will use the help text and switch to `lifetime`

BTW these flags are not very visible; the whole helpers command subtree is hidden.  But it is visible in your kubeconfig if you're using the plugin, so it's worth having them be clear.

pkg/commands/helpers/kubectl_auth.go
justinsb: I believe this is when the credential is an error, not just expired, but you're right - going with Infof

pkg/commands/helpers/kubectl_auth.go
justinsb: I'll try to use user.Current() (which I believe is cgo safe now), doing `kubecfg-$name`  Happy to change to something better!

justinsb:
This seems to work, marking as no longer WIP.  Thanks for the feedback @johngmyers and @rifelpet 

Also added a test!

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -152,6 +150,18 @@ func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.Se
...
 
+	if useKopsAuthenticationPlugin {
+		b.AuthenticationExec = []string{
+			"kops",
+			"helpers",
+			"kubectl-auth",
+			"--cluster=" + clusterName,
+			"--state=" + kopsStateStore,
+			"--lifetime=1h",
+			"--api-version=" + "v1beta1",
```
johngmyers, line 161: Perhaps we should let lifetime and api-version default instead of freezing these values in the user's kubeconfig.

justinsb: The api-version is an internal detail I think, at least until kubectl introduces v1.  I would hope that would be a gradual transition and not something we need to bother the user with.

For the lifetime, my plan is to get this in there and then look at all the changes we've made.  For example - should this be the same as the admin lifetime?  We've taken a usability hit by not exporting the kubecfg, because it breaks the "infer the current cluster" experience.

I moved the variables to the top of the block and added a comment and a TODO.

johngmyers: If we write them to the kubecfg then they will stay fixed for that user unless/until that user takes an explicit action to update them. If we omit the flags from the kubecfg and let cobra default them then upon kops upgrade any new default values will automatically take effect.

justinsb: Oh I get it now!  That's a good point :-)   I'll remove the defaults!

PR #9672, opened 2020-08-03
[Digital Ocean] Upload binaries to DO Spaces

srikiz:
Using s3cmd to upload KOPS binaries to DO spaces, helps in development when working on nodeup and other binaries to use custom images from DO Spaces.

### DIFF
--- a/hack/upload
+++ b/hack/upload
@@ -35,5 +35,11 @@ if [[ "${DEST:0:6}" == "oss://" ]]; then
   exit 0
 fi
 
+if [[ "${DEST:0:5}" == "do://" ]]; then
+  DO_BUCKET=`echo "${DEST}" | cut -c 6-`
+  s3cmd put "*" ${SRC}/ s3://$DO_BUCKET --recursive ${PUBLIC:+--acl-public}
+  exit 0
+fi
+
 echo "Unsupported destination - supports s3://, gs:// and oss:// urls: ${DEST}"
 exit 1


johngmyers:
@srikiz Could you perhaps add this logic to `hack/upload`?

srikiz:
@johngmyers - I have incorporated your comments. Please have a look when you get a chance, thanks !

hack/upload
```diff
@@ -35,5 +35,11 @@ if [[ "${DEST:0:6}" == "oss://" ]]; then
   exit 0
 fi
 
+if [[ "${DEST:0:5}" == "do://" ]]; then
+  DO_BUCKET=`echo "${DEST}" | cut -c 6-`
+  s3cmd put "*" ${SRC}/ s3://$DO_BUCKET --recursive --acl-public
```
johngmyers, line 40: Can this use `PUBLIC` like the others?

johngmyers:
With the change to `hack/upload` is the `Makefile` change still needed? Can people not use `make dev-upload`?

hack/upload
srikiz: I have set the --acl-public flag set, allowing read access for everyone.

srikiz:
yeah, make dev-upload is good. I reverted the change in Makefile. Thanks @johngmyers !

hack/upload
johngmyers: The point is to allow callers to omit the `--acl-public` flag.

johngmyers:
Also, please squash commits.

hack/upload
srikiz: Ah, I got that now. I didn't realize parameter substitution was happening.. I updated it and squashed the commits as well.

PR #9677, opened 2020-08-04
Add missing locking in MockEC2

johngmyers:


johngmyers:
Argh, recursive locking.

PR #9691, opened 2020-08-06
Add an initial structure for openstack cloudmock

rifelpet:
More info in the docs changes.

This adds stubbed HTTP handlers for every resource type used by Kops. Currently they only return an empty 200 response but followup PRs will implement the relevant CRUD logic.


olemarkus:
Approach looks good to me.
/lgtm

PR #9699, opened 2020-08-06
Spotinst: Selecting a default Instance Group in Spot Ocean should be optional

liranp:
This PR changes the way the default Instance Group for Spot Ocean is selected to allow users to create new clusters with multiple Instance Groups (#9471) without having to generate their manifests first to select the default by adding the metadata label (`spotinst.io/ocean-default-launchspec`).

### DIFF
--- a/pkg/model/spotinstmodel/instance_group.go
+++ b/pkg/model/spotinstmodel/instance_group.go
@@ -323,7 +323,7 @@ func (b *InstanceGroupModelBuilder) buildOcean(c *fi.ModelBuilderContext, igs ..
 	{
 		// Single instance group.
 		if len(igs) == 1 {
-			ig = igs[0]
+			ig = igs[0].DeepCopy()
 		}
 
 		// Multiple instance groups.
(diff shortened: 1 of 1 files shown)


pkg/model/spotinstmodel/instance_group.go
```diff
@@ -343,22 +343,25 @@ func (b *InstanceGroupModelBuilder) buildOcean(c *fi.ModelBuilderContext, igs ..
...
-					InstanceGroupLabelOceanDefaultLaunchSpec)
+				ig = igs[0].DeepCopy()
 			}
 		}
 
 		klog.V(4).Infof("Detected default launch spec: %q", b.AutoscalingGroupName(ig))
 	}
 
+	// Rename the instance group to avoid duplicate tasks with same name.
+	ig.Name = fi.StringValue(ocean.Name)
```
povils, line 363: HEy @liranp !
Could it be that this line introduced a bug I might be facing?

By overriding ig.Name it messes up userdata for default nodes:
ocean.UserData, err = b.BootstrapScriptBuilder.ResourceNodeUp(c, ig)

Because now nodeUp tries to fetch instance group config with name "ocean.Name" which ofcourse it doesn;t exist

liranp: @povils, thanks for reporting this. I've raised another PR (#10630) to fix it. Sorry about that.

PR #9700, opened 2020-08-06
Add cinder plugin

olemarkus:
This will install the cinder plugin allowing dynamic provisioning of volumes.
Since kops already depend on cinder and since most people need this anyway, I add this unconditionally.
If there are needs for disabling this, I can always add it. But we add storageclass etc unconditionally for AWS and GCE too.

### DIFF
--- a/docs/releases/1.19-NOTES.md
+++ b/docs/releases/1.19-NOTES.md
@@ -12,6 +12,12 @@ Similarly, `kops export kubecfg` will also require passing either the `--admin` 
 
 `kops create cluster --yes` exports the admin user along with rest of the cluster config, as is existing behaviour.
 
+## OpenStack Cinder plugin
+
+Kops will install the Cinder plugin for kops running kubernetes 1.16 or newer. If you already have this plugin installed you should remove it before upgrading.
+
+If you already have a default `StorageClass`, you should set `cloudConfig.Openstack.BlockStorage.CreateStorageClass: false` to prevent kops from installing one.
+
 ## Other significant changes
 
 * New clusters will now have one nodes group per zone. The number of nodes now defaults to the number of zones.
@@ -36,6 +42,8 @@ has been updated by a newer version of kops unless it is given the `--allow-kops
 * Support for legacy IAM permissions has been removed. This removal may be temporarily deferred to kops 1.20 by setting the `LegacyIAM` feature flag.
 
 # Required Actions
+
+* See note about Openstack Cinder plugin above.
 
 # Deprecations
 
--- a/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
+++ b/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
@@ -1053,9 +1053,26 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 		}
 	}
 
-	if b.Cluster.Spec.ExternalCloudControllerManager != nil {
-		// cloudprovider specific out-of-tree controller
-		if kops.CloudProviderID(b.Cluster.Spec.CloudProvider) == kops.CloudProviderOpenstack {
+	if kops.CloudProviderID(b.Cluster.Spec.CloudProvider) == kops.CloudProviderOpenstack {
+		{
+			key := "storage-openstack.addons.k8s.io"
+			version := "1.18.0-kops.1"
+
+			id := "k8s-1.16"
+			location := key + "/" + id + ".yaml"
+
+			addons.Spec.Addons = append(addons.Spec.Addons, &channelsapi.AddonSpec{
+				Name:              fi.String(key),
+				Version:           fi.String(version),
+				Manifest:          fi.String(location),
+				Selector:          map[string]string{"k8s-addon": key},
+				KubernetesVersion: ">=1.15.0",
+				Id:                id,
+			})
+		}
+
+		if b.Cluster.Spec.ExternalCloudControllerManager != nil {
+			// cloudprovider specific out-of-tree controller
 			{
 				key := "openstack.addons.k8s.io"
 				version := "1.11.0"
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -222,6 +222,10 @@ spec:
                         properties:
                           bs-version:
                             type: string
+                          createStorageClass:
+                            description: CreateStorageClass provisions a default class
+                              for the Cinder plugin
+                            type: boolean
                           ignore-volume-az:
                             type: boolean
                           override-volume-az:
--- a/pkg/apis/kops/componentconfig.go
+++ b/pkg/apis/kops/componentconfig.go
@@ -684,6 +684,8 @@ type OpenstackBlockStorageConfig struct {
 	Version    *string `json:"bs-version,omitempty"`
 	IgnoreAZ   *bool   `json:"ignore-volume-az,omitempty"`
 	OverrideAZ *string `json:"override-volume-az,omitempty"`
+	// CreateStorageClass provisions a default class for the Cinder plugin
+	CreateStorageClass *bool `json:"createStorageClass,omitempty"`
 }
 
 // OpenstackMonitor defines the config for a health monitor
--- a/pkg/apis/kops/v1alpha2/componentconfig.go
+++ b/pkg/apis/kops/v1alpha2/componentconfig.go
@@ -685,6 +685,8 @@ type OpenstackBlockStorageConfig struct {
 	Version    *string `json:"bs-version,omitempty"`
 	IgnoreAZ   *bool   `json:"ignore-volume-az,omitempty"`
 	OverrideAZ *string `json:"override-volume-az,omitempty"`
+	// CreateStorageClass provisions a default class for the Cinder plugin
+	CreateStorageClass *bool `json:"createStorageClass,omitempty"`
 }
 
 // OpenstackMonitor defines the config for a health monitor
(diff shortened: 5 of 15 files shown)


upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.15.yaml
```diff
@@ -0,0 +1,477 @@
...
+metadata:
+  name: csi-cinder-controller-service
+  namespace: kube-system
+  labels:
+    app: csi-cinder-controllerplugin
+spec:
+  selector:
+    app: csi-cinder-controllerplugin
+  ports:
+    - name: dummy
```
johngmyers, line 226: Not a very good name

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.15.yaml
```diff
@@ -0,0 +1,477 @@
...
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      app: csi-cinder-nodeplugin
+  template:
+    metadata:
+      labels:
+        app: csi-cinder-nodeplugin
+    spec:
```
johngmyers, line 377: Shouldn't this have tolerations?

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.15.yaml
olemarkus: good catch

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.15.yaml
olemarkus: Yes it should. Fixed

docs/releases/1.19-NOTES.md
```diff
@@ -27,6 +27,8 @@ Similarly, `kops export kubecfg` will also require passing either the `--admin`
 * The `kops update cluster` command will now refuse to run on a cluster that
 has been updated by a newer version of kops unless it is given the `--allow-kops-downgrade` flag.
 
+* Kops will install the Cinder plugin and `StorageClass` on OpenStack clusters.
```
zetaab, line 30: I do not see the logic of StorageClass in this PR?

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -1053,9 +1053,26 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
...
+
+			id := "k8s-1.15"
+			location := key + "/" + id + ".yaml"
+
+			addons.Spec.Addons = append(addons.Spec.Addons, &channelsapi.AddonSpec{
+				Name:              fi.String(key),
+				Version:           fi.String(version),
+				Manifest:          fi.String(location),
+				Selector:          map[string]string{"k8s-addon": key},
+				KubernetesVersion: ">=1.15.0",
```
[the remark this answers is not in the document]
olemarkus: I tried to find info about that. But https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-cinder-csi-plugin.md#compatibility stops at 1.15 ...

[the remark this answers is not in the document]
zetaab: there was some feature that is not supported in 1.15, anyway if we want be safe here is that we start from 1.16

docs/releases/1.19-NOTES.md
olemarkus: Yeah, this one somehow escaped the final template. Added back in.

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.15.yaml
```diff
@@ -0,0 +1,516 @@
...
+                  fieldPath: spec.nodeName
+          imagePullPolicy: "IfNotPresent"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+            - name: registration-dir
+              mountPath: /registration
+        - name: cinder-csi-plugin
+          securityContext:
+            privileged: true
```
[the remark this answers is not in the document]
olemarkus: Fixed.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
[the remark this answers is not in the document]
olemarkus: Sounds good. Also added that to the release notes.

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml
```diff
@@ -0,0 +1,531 @@
...
+              mountPath: /registration
+        - name: cinder-csi-plugin
+          securityContext:
+            privileged: true
+            runAsNonRoot: false
+            runAsUser: 0
+            capabilities:
+              add: ["SYS_ADMIN"]
+            allowPrivilegeEscalation: true
+          image: docker.io/k8scloudprovider/cinder-csi-plugin:v1.18.0
```
zetaab, line 450: This is kind of problem when we hardcode this? What happens when 1.19 plugin is released? Is this 1.18 really backward compatible to 1.16 :)

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml
```diff
@@ -0,0 +1,531 @@
...
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: default
+  annotations:
+    storageclass.beta.kubernetes.io/is-default-class: "true"
+  labels:
+    k8s-addon: storage-openstack.addons.k8s.io
+provisioner: cinder.csi.openstack.org
```
zetaab, line 529: we do have storageclass currently which have two parameters: `type` and `availability`. We cannot use cinder without those. So I suggest of somehow adding possible values for those

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml
zetaab: okay it might be that some other setting in manifest is also changing, which forces us to update the manifest

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml
```diff
@@ -0,0 +1,531 @@
...
+            - "--csi-address=$(ADDRESS)"
+          env:
+            - name: ADDRESS
+              value: /var/lib/csi/sockets/pluginproxy/csi.sock
+          imagePullPolicy: "IfNotPresent"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /var/lib/csi/sockets/pluginproxy/
+        - name: cinder-csi-plugin
+          image: docker.io/k8scloudprovider/cinder-csi-plugin:v1.18.0
```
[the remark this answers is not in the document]
olemarkus: How about modifying the OpenStackCCM template function to just care about the tag, not the image repository and just reuse that one for cinder as well? I think those versions just follow each other right?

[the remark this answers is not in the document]
zetaab: yeah those follow each other, and those are released at the same time. So that makes sense, lets do that

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml
olemarkus: This is a tricky one. With the other cloud providers, we don't allow modifications of the SC either. I wonder if the best thing to do is to just keep the one you have and set it specifically on your PVCs. I realise this is not all that nice for end-users, but it is consistent with other providers.

zetaab: one possible thing is that we could disable provisioning the storageclass (that can be applied to other clouds also if needed)? The default behaviour could be that it will provision it

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml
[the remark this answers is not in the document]
olemarkus: I assume they are also released in concert with k8s so we don't run into an issue with the new tag not being available once users upgrade k8s?

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml
olemarkus: Something like this will work at least.

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml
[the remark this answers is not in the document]
zetaab: we do have that problem, if people tries like k8s alpha/beta versions. Openstack cloudprovider components are released something like week after the final k8s stable is released.

olemarkus:
Should have merged this before I lgtm'ed the integration tests :)

PR #9701, opened 2020-08-06
Respect Topology when assigning floating ips or not

olemarkus:
This is a commit that got lost from #9560, I think. Without this, kops will unconditionally add floating IPs even though openstack setup doesn't support it resulting in an npr.

/kind bug


pkg/apis/kops/validation/openstack.go
```diff
@@ -0,0 +1,35 @@
...
+
+package validation
+
+import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/kops/pkg/apis/kops"
+)
+
+func openstackValidateCluster(c *kops.Cluster) (errList field.ErrorList) {
+	if c.Spec.CloudConfig.Openstack.Router != nil && c.Spec.CloudConfig.Openstack.Router.ExternalNetwork == nil {
```
johngmyers, line 25: But public topology doesn't require an external network when there's no router?

olemarkus: Public topology means using floating IPs. In order to use floating IPs you need both an external network and you need a router (which handles traffic between internal and external network). So public topology is not possible without one.

johngmyers: But this validation code skips the topology check if `c.Spec.CloudConfig.Openstack.Router` is `nil`

olemarkus: Ah. That is quite a bug. Fixed now. And added some tests as well.

PR #9702, opened 2020-08-06
Adds support for using OS application credentials

olemarkus:
Application credentials allows you to export a purpose-specific set of
credentials for a user instead of exposing user login credentials.
Especially useful when using LDAP or similar for Openstack users.
Also lets you rotate credentials more easily since multiple application
credentials can be provisioned per user.

Depends on https://github.com/kopeio/etcd-manager/pull/336 as well.

### DIFF
--- a/nodeup/pkg/model/cloudconfig.go
+++ b/nodeup/pkg/model/cloudconfig.go
@@ -92,6 +92,8 @@ func (b *CloudConfigBuilder) Build(c *fi.ModelBuilderContext) error {
 			fmt.Sprintf("tenant-name=\"%s\"", tenantName),
 			fmt.Sprintf("domain-name=\"%s\"", os.Getenv("OS_DOMAIN_NAME")),
 			fmt.Sprintf("domain-id=\"%s\"", os.Getenv("OS_DOMAIN_ID")),
+			fmt.Sprintf("application-credential-id=\"%s\"", os.Getenv("OS_APPLICATION_CREDENTIAL_ID")),
+			fmt.Sprintf("application-credential-secret=\"%s\"", os.Getenv("OS_APPLICATION_CREDENTIAL_SECRET")),
 			"",
 		)
 
--- a/pkg/model/bootstrapscript.go
+++ b/pkg/model/bootstrapscript.go
@@ -108,17 +108,30 @@ func (b *BootstrapScript) buildEnvironmentVariables(cluster *kops.Cluster) (map[
 		env["S3_SECRET_ACCESS_KEY"] = os.Getenv("S3_SECRET_ACCESS_KEY")
 	}
 
-	// Pass in required credentials when using user-defined swift endpoint
-	if os.Getenv("OS_AUTH_URL") != "" {
-		for _, envVar := range []string{
-			"OS_TENANT_ID", "OS_TENANT_NAME", "OS_PROJECT_ID", "OS_PROJECT_NAME",
-			"OS_PROJECT_DOMAIN_NAME", "OS_PROJECT_DOMAIN_ID",
-			"OS_DOMAIN_NAME", "OS_DOMAIN_ID",
+	osEnvs := []string{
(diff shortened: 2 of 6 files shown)


nodeup/pkg/model/cloudconfig.go
```diff
@@ -92,7 +92,8 @@ func (b *CloudConfigBuilder) Build(c *fi.ModelBuilderContext) error {
 			fmt.Sprintf("tenant-name=\"%s\"", tenantName),
 			fmt.Sprintf("domain-name=\"%s\"", os.Getenv("OS_DOMAIN_NAME")),
 			fmt.Sprintf("domain-id=\"%s\"", os.Getenv("OS_DOMAIN_ID")),
-			"",
```
hakman, line 95: Intended to remove the "\n"?

pkg/model/bootstrapscript.go
```diff
@@ -108,17 +108,30 @@ func (b *BootstrapScript) buildEnvironmentVariables(cluster *kops.Cluster) (map[
...
-			"OS_DOMAIN_NAME", "OS_DOMAIN_ID",
+	osEnvs := []string{
+		"OS_TENANT_ID", "OS_TENANT_NAME", "OS_PROJECT_ID", "OS_PROJECT_NAME",
+		"OS_PROJECT_DOMAIN_NAME", "OS_PROJECT_DOMAIN_ID",
+		"OS_DOMAIN_NAME", "OS_DOMAIN_ID",
+		"OS_AUTH_URL",
+		"OS_REGION_NAME",
+	}
+
+	if os.Getenv("OS_APPLICATION_CREDENTIAL_ID") != "" {
```
hakman, line 119:
```suggestion
	if os.Getenv("OS_APPLICATION_CREDENTIAL_ID") != "" && os.Getenv("OS_APPLICATION_CREDENTIAL_SECRET") {
```

util/pkg/env/standard.go
```diff
@@ -58,6 +58,8 @@ func BuildSystemComponentEnvVars(spec *kops.ClusterSpec) EnvVars {
 	vars.addEnvVariableIfExist("OS_DOMAIN_ID")
 	vars.addEnvVariableIfExist("OS_USERNAME")
 	vars.addEnvVariableIfExist("OS_PASSWORD")
+	vars.addEnvVariableIfExist("OS_APPLICATION_CREDENTIAL_ID")
+	vars.addEnvVariableIfExist("OS_APPLICATION_CREDENTIAL_SECRET")
```
hakman, line 62: Maybe move at the end of the list, same as in `nodeup/pkg/model/protokube.go` ?

nodeup/pkg/model/cloudconfig.go
```diff
@@ -92,7 +92,8 @@ func (b *CloudConfigBuilder) Build(c *fi.ModelBuilderContext) error {
...
+			fmt.Sprintf("application-credential-id=\"%s\"", os.Getenv("OS_APPLICATION_CREDENTIAL_ID")),
+			fmt.Sprintf("application-credential-secret=\"%s\"", os.Getenv("OS_APPLICATION_CREDENTIAL_SECRET")),
```
hakman, line 96: Should these be added even when `OS_APPLICATION_CREDENTIAL_ID` and `OS_APPLICATION_CREDENTIAL_SECRET` are not set?

olemarkus: Everything else is unconditionally added. Any key is empty by default anyway, so from config consumer perspective it is the same.

hakman: 👍

util/pkg/env/standard.go
hakman: Duplicate now?

olemarkus: Noticed that. Already pushed an amendment. thanks

PR #9704, opened 2020-08-07
Add aws-cloud-controller-manager config to addons

nckturner:
- Added addon for aws-cloud-controller (aws-cloud-controller.addons.k8s.io) 
- Initial manifest: `upup/models/cloudup/resources/addons/aws-cloud-controller.addons.k8s.io/k8s-1.18.yaml.template`
- `--external-cloud-volume-plugin` flag for KCM
- AWSCCMTag to hold the logic to determine the CCM tag.  This will change as soon as we have stable releases (soon).

### DIFF
--- a/pkg/model/components/kubecontrollermanager.go
+++ b/pkg/model/components/kubecontrollermanager.go
@@ -98,6 +98,14 @@ func (b *KubeControllerManagerOptionsBuilder) BuildOptions(o interface{}) error 
 
 	if clusterSpec.ExternalCloudControllerManager != nil {
 		kcm.CloudProvider = "external"
+
+		// External cloud provider disables KCM volume controllers, so
+		// most users would want to either install CSI or pass
+		// --external-cloud-volume-plugin to the KCM, which runs the
+		// KCM volume controllers.
+		if kcm.ExternalCloudVolumePlugin == "" {
+			klog.Infof("An external cloud controller manager is configured, but ExternalCloudVolumePlugin is not configured for the KCM.  This means a CSI plugin must be installed by the user or else volume management might not work.")
+		}
 	}
 
 	kcm.LogLevel = 2
--- a/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
+++ b/upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
@@ -67,6 +67,7 @@ func (b *BootstrapChannelBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 		name := b.Cluster.ObjectMeta.Name + "-addons-" + key
 		manifestPath := "addons/" + *a.Manifest
+		klog.V(4).Infof("Addon %q", name)
 
 		manifestResource := b.templates.Find(manifestPath)
 		if manifestResource == nil {
@@ -978,6 +979,28 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*chann
 					Selector: map[string]string{"k8s-addon": key},
 					Manifest: fi.String(location),
 					Id:       id,
+				})
+			}
+		}
+	}
+
+	if kops.CloudProviderID(b.Cluster.Spec.CloudProvider) == kops.CloudProviderAWS {
+		key := "aws-cloud-controller.addons.k8s.io"
+
+		if b.Cluster.Spec.ExternalCloudControllerManager != nil {
+			// Version refers to the addon configuration.  The CCM tag is given by
+			// the template function AWSCCMTag()
+			version := "1.18.0-kops.1"
+			{
+				id := "k8s-1.18"
+				location := key + "/" + id + ".yaml"
+				addons.Spec.Addons = append(addons.Spec.Addons, &channelsapi.AddonSpec{
+					Name:              fi.String(key),
+					Version:           fi.String(version),
+					Manifest:          fi.String(location),
+					Selector:          map[string]string{"k8s-addon": key},
+					KubernetesVersion: ">=1.18.0",
+					Id:                id,
 				})
 			}
 		}
--- a/upup/pkg/fi/cloudup/template_functions.go
+++ b/upup/pkg/fi/cloudup/template_functions.go
@@ -122,6 +122,7 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 
 	// will return openstack external ccm image location for current kubernetes version
 	dest["OpenStackCCMTag"] = tf.OpenStackCCMTag
+	dest["AWSCCMTag"] = tf.AWSCCMTag
 	dest["ProxyEnv"] = tf.ProxyEnv
 
 	dest["KopsSystemEnv"] = tf.KopsSystemEnv
@@ -530,6 +531,23 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
 		}
 	}
 	return tag
+}
+
+// AWSCCMTag returns the correct tag for the cloud controller manager based on
+// the Kubernetes Version
+func (tf *TemplateFunctions) AWSCCMTag() (string, error) {
+	var tag string
+	parsed, err := util.ParseKubernetesVersion(tf.Cluster.Spec.KubernetesVersion)
+
+	// Update when we have stable releases
+	if err != nil {
+		return "", fmt.Errorf("failed to parse Kubernetes version from cluster spec: %q", err)
+	} else if parsed.Minor == 18 {
+		tag = "v1.18.0-alpha.1"
+	} else if parsed.Minor == 19 {
+		tag = "v1.19.0-alpha.1"
+	}
+	return tag, nil
 }
 
 // GetNodeInstanceGroups returns a map containing the defined instance groups of role "Node".
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -1223,6 +1223,9 @@ spec:
                   experimentalClusterSigningDuration:
                     description: ExperimentalClusterSigningDuration is the duration that determines the length of duration that the signed certificates will be given. (default 8760h0m0s)
                     type: string
+                  externalCloudVolumePlugin:
+                    description: ExternalCloudVolumePlugin is a fallback mechanism that allows a legacy, in-tree cloudprovider to be used for volume plugins even when an external cloud controller manager is being used.  This can be used instead of installing CSI.  The value should be the same as is used for the --cloud-provider flag, i.e. "aws".
+                    type: string
                   featureGates:
                     additionalProperties:
                       type: string
--- a/pkg/apis/kops/componentconfig.go
+++ b/pkg/apis/kops/componentconfig.go
@@ -608,6 +608,10 @@ type KubeControllerManagerConfig struct {
 	AuthorizationKubeconfig string `json:"authorizationKubeconfig,omitempty" flag:"authorization-kubeconfig"`
 	// AuthorizationAlwaysAllowPaths is the list of HTTP paths to skip during authorization
 	AuthorizationAlwaysAllowPaths []string `json:"authorizationAlwaysAllowPaths,omitempty" flag:"authorization-always-allow-paths"`
+	// ExternalCloudVolumePlugin is a fallback mechanism that allows a legacy, in-tree cloudprovider to be used for volume plugins
+	// even when an external cloud controller manager is being used.  This can be used instead of installing CSI.  The value should
+	// be the same as is used for the --cloud-provider flag, i.e. "aws".
+	ExternalCloudVolumePlugin string `json:"externalCloudVolumePlugin,omitempty" flag:"external-cloud-volume-plugin"`
 
 	// EnableProfiling enables profiling via web interface host:port/debug/pprof/
 	EnableProfiling *bool `json:"enableProfiling,omitempty" flag:"profiling"`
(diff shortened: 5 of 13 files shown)


upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -1088,6 +1088,22 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
...
+
+			location := key + "/v1.18.0.yaml"
+			id := "k8s-1.18-ccm"
+
+			addons.Spec.Addons = append(addons.Spec.Addons, &channelsapi.AddonSpec{
+				Name:              fi.String(key),
+				Version:           fi.String(version),
+				Manifest:          fi.String(location),
+				Selector:          map[string]string{"k8s-addon": key},
+				KubernetesVersion: ">=1.18.0",
```
johngmyers, line 1103: What is the correct behavior for k8s < 1.18? Previously it would install `core.addons.k8s.io`.

upup/models/cloudup/resources/addons/aws-cloud-controller.addons.k8s.io/v1.18.0.yaml
```diff
@@ -0,0 +1,39 @@
...
+      tolerations:
+      - key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
+        effect: NoSchedule
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
+      serviceAccountName: cloud-controller-manager
+      containers:
+        - name: aws-cloud-controller-manager
+          image: gcr.io/k8s-staging-provider-aws/cloud-controller-manager:v1.18.0-alpha.1
```
johngmyers, line 31: Does the version need to track that of Kubernetes? The cloud-provider-aws repo doesn't say what the version compatibility policy is.

olemarkus: You can most likely just copy the OpenStackCCMTag function in template_functions.go.

olemarkus:
Need any help to progress on this?

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
nckturner: Yeah, we should probably keep that behavior, as we don't have any published images pre-1.18.

nckturner:
@olemarkus so I ran into a ton of failures running e2e tests on clusters created with this change, and finally had a chance to dig in and discovered that the issue is that aws volumes need kubelet to start up with `--cloud-provider=aws` in order to pass.  Otherwise, the volumes fail to be mounted.  

`Error: "MountVolume.SetUp failed for volume \"aws-j6cxz\" (UniqueName: \"kubernetes.io/aws-ebs/ddc37e26-a4ea-4800-b9cf-89f4c4ba2e06-aws-j6cxz\") pod \"aws-injector\" (UID: \"ddc37e26-a4ea-4800-b9cf-89f4c4ba2e06\") : mount failed: exit status 32\nMounting command: systemd-run\nMounting arguments: --description=Kubernetes transient mount for /var/lib/kubelet/pods/ddc37e26-a4ea-4800-b9cf-89f4c4ba2e06/volumes/kubernetes.io~aws-ebs/aws-j6cxz --scope -- mount  -o bind /var/lib/kubelet/plugins/kubernetes.io/aws-ebs/mounts/aws/us-west-2a/vol-0927658c7047d003e/var/lib/kubelet/pods/ddc37e26-a4ea-4800-b9cf-89f4c4ba2e06/volumes/kubernetes.io~aws-ebs/aws-j6cxz\nOutput: Running scope as unit: run-r0506619461a2436f9b6de03226e313b4.scope\nmount: /var/lib/kubelet/pods/ddc37e26-a4ea-4800-b9cf-89f4c4ba2e06/volumes/kubernetes.io~aws-ebs/aws-j6cxz: special device /var/lib/kubelet/plugins/kubernetes.io/aws-ebs/mounts/aws/us-west-2a/vol-0927658c7047d003e does not exist.\n"`

Interestingly, with `--cloud-provider=external` set, code paths using `pkg/volume/awsebs/aws_ebs.go` were being triggered, which was unexpected to me. 

This issue looks like the same: https://github.com/kubernetes/kubernetes/issues/71018.  So possibly the options as the code stands today are either 1. use CSI or 2. run kubelet with `--cloud-provider=aws`?  Note that I tested with 1.18.6.

olemarkus:
I think it makes sense to use CSI. It is something that would be interesting to support anyway.

nckturner:
@olemarkus I'd actually like to allow for both, as one use case for this change is testing upgrade scenarios.  Right now this is configurable with:
```
  kubelet:
    cloudProvider: aws
```

nckturner:
The other KCM flag that I've exposed in this PR is the `--external-cloud-volume-plugin` which allows the KCM to run the existing cloud volume control loops even when the other cloud loops are disabled due to setting `--cloud-provider=external`.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -1088,6 +1088,25 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
...
+				Id:                id,
+			})
+
+			// Ensure ExternalCloudVolumePlugin is set to AWS
+			b.Cluster.Spec.KubeControllerManager.ExternalCloudVolumePlugin = "aws"
```
olemarkus, line 1108: This type of logic can be added in `./pkg/model/components`. That way changes done are also subject to validation and such. There is already something similar in place for openstack.

olemarkus:
Sounds good. It may be an idea to add a flag to `CloudConfig` controlling if the CSI addon is installed or not, and then ensure users are configuring the flags you mention correctly through validation.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
nckturner: Thanks, that's very helpful.  Will add some validation here.

nckturner:
* I've added a template function for the CCM tag: `AWSCCMTag`
* I've added a propsed CSI API so that we can detect and at least log a message when the user has a configuration that will break volumes.  I'm not sure if we want to overwrite the user's value for `--external-cloud-volume-plugin` or not, I think we shouldn't do it in this PR because CSI is not yet implemented.  (I can also remove the proposed CSI API from this PR if we want to discuss that separately, I'm not planning on implementing it in this PR).

pkg/apis/kops/cluster.go
```diff
@@ -199,6 +199,10 @@ type ClusterSpec struct {
 
 	// ClusterAutoscaler defines the cluster autoscaler configuration.
 	ClusterAutoscaler *ClusterAutoscalerConfig `json:"clusterAutoscaler,omitempty"`
+
+	// CSI defines container storage interface configuration.
+	// +optional
+	CSI *CSIConfiguration `json:"csi,omitempty"`
```
olemarkus, line 205: For other providers we have similar types of configuration under a struct in `CloudConfiguration`. AWS doesn't have its own struct here; now may be the time to add it.

nckturner: I thought about that, but CloudConfig seems to be tightly mapped to the actual config file that is placed on disk, which is deprecated according to documentation I found.  Shouldn't we move on from that object?  Also, it kindof combines/conflates various addons, like CCM, CSI, etc. (even though, yes they are cloud related, there are many combinations that users might want to use and combining all of the configuration makes it quite difficult to understand what configuration is configuring which addon).

nckturner:
I think I missed the versioned v1alpha2 API, will update that now.

olemarkus:
And `make apimachinery` it seems. Running `make verify && ./hack/verify-staticcheck.sh` locally may be a faster feedback loop.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -538,6 +539,23 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
...
 
+// AWSCCMTag returns the correct tag for the cloud controller manager based on
+// the Kubernetes Version
+func (tf *TemplateFunctions) AWSCCMTag() string {
+	var tag string
+	parsed, err := util.ParseKubernetesVersion(tf.Cluster.Spec.KubernetesVersion)
+
+	// Update when we have stable releases
+	if err != nil {
+		tag = "v1.19.0-alpha.1"
```
ayberk, line 550: Should we log an info message here to inform that parsing failed and we're defaulting to this tag?

nckturner: Yeah I'll add that.

olemarkus:
Do you want to add any docks on how to use external AWS CCM?

nckturner:
Good point.  I was planning on adding them to the CCM repo, but they should definitely be here as well.

justinsb:
This looks really good.  We discussed that probably the CSI fields are better in a separate PR, and that we can just leave the warning when running without CSI & without the CCM flag in the code for now but commented out (maybe just the CSI bit commented out?) until the CSI PR lands separately.

There won't be much in the codepath then unless the feature-flag is set, so I'm in favor of merging and getting this under testing.

Thanks @nckturner for this :-)

pkg/apis/kops/cluster.go
```diff
@@ -744,3 +748,34 @@ type RollingUpdate struct {
...
+	// +optional
+	LivenessProbeImage string `json:"livenessprobeImage,omitempty"`
+	// NodeDriverRegistrarImage is the CSI Node Driver registrar image
+	// https://github.com/kubernetes-csi/node-driver-registrar
+	// +optional
+	NodeDriverRegistrarImage string `json:"nodeDriverRegistrarImage,omitempty"`
+
+	// EBSCSIConfiguration defines AWS EBS CSI configuration
+	// +optional
+	EBS *EBSCSIConfiguration `json:"ebs,omitempty"`
```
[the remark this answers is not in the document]
nckturner: Going to move the proposed CSI API into a separate PR per our convo.

pkg/model/components/kubecontrollermanager.go
```diff
@@ -98,6 +98,14 @@ func (b *KubeControllerManagerOptionsBuilder) BuildOptions(o interface{}) error
 
 	if clusterSpec.ExternalCloudControllerManager != nil {
 		kcm.CloudProvider = "external"
+
+		// External cloud provider disables KCM volume controllers, so
+		// most users would want to either install CSI or pass
+		// --external-cloud-volume-plugin to the KCM, which runs the
+		// KCM volume controllers.
+		if clusterSpec.CSI == nil && kcm.ExternalCloudVolumePlugin == "" {
+			klog.Infof("An external cloud controller manager is configured, but CSI is not configured and --external-cloud-volume-plugin is not configured for the KCM.  Volume management might not work.")
```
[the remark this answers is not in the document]
nckturner: This has become:

               // External cloud provider disables KCM volume controllers, so
               // most users would want to either install CSI or pass
               // --external-cloud-volume-plugin to the KCM, which runs the
               // KCM volume controllers.
               if kcm.ExternalCloudVolumePlugin == "" {
                       klog.Infof("An external cloud controller manager is configured, but ExternalCloudVolumePlugin is not configured for the KCM.  This means a CSI plugin must be installed by the user or else volume management might not work.")
               }

Once kops has the ability to install CSI we can modify this.

upup/models/cloudup/resources/addons/aws-cloud-controller.addons.k8s.io/k8s-1.18.yaml
```diff
@@ -0,0 +1,154 @@
...
+subjects:
+- apiGroup: ""
+  kind: ServiceAccount
+  name: cloud-controller-manager
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:cloud-controller-manager
```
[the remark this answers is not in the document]
nckturner: I left it as is, since there isn't multiple as of now, and this matches what we have in the aws-cloud-provider repo.  Happy to revisit in the future.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -538,6 +539,24 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
...
+		klog.V(4).Infof("Failed to parse Kubernetes version from cluster spec.  Defaulting tag of AWS cloud controller manager to: %s", tag)
+	} else if parsed.Minor == 18 {
+		tag = "v1.18.0-alpha.1"
```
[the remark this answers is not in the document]
nckturner: Agreed, this is ugly but shouldn't last long.

nckturner:
Updated the PR per @justinsb's comments.  I decided to leave out docs at his suggestion since its early/feature gate protected, we will add them in a follow on PR.  We can start with some basic usage docs in the aws-cloud-provider repository.

nckturner:
Fixing bug
/hold

nckturner:
@justinsb ready for another look

nckturner:
/hold cancel

Any chance we can merge this @justinsb?  It will make testing the CCM easier..

justinsb:
Sorry about the delay here, and thanks for doing this @nckturner !

This looks to be entirely additive and feature flagged, and I believe the current target for external cloud-controller-manager is 1.21, so I think it's important that we start to get this under testing (not least because my understanding is that it doesn't yet entirely work!).

/approve
/lgtm

Issue #7200, opened 2019-06-28
metrics-server: enable authenticationTokenWebhook


sdarwin:
Per https://github.com/kubernetes-incubator/metrics-server/issues/133 , the metrics-server will not function on kops without these fixes.

1. --kubelet-insecure-tls added to metrics-server
2. authenticationTokenWebhook set on the kubelet  

kops edit cluster

```
kubelet:
     anonymousAuth: false
     authenticationTokenWebhook: true
     authorizationMode: Webhook
```
Although, minimalistically this is sufficient:
```
kubelet:
     anonymousAuth: false
     authenticationTokenWebhook: true
```

Since running a metrics server is pretty common, eliminating the extra steps would be convenient. 

Could the settings of "authenticationTokenWebhook: true" and "authorizationMode: Webhook" be configured by default on kops, so the end-user doesn't encounter unexpected errors when deploying metrics?

sdarwin:
Investigating the issue a bit more, a question could be "Do other k8s installers also have this same problem? How do they solve it?"

1. Kubespray

Kubespray enables both Webhook authentication and authorization for the kubelet. Here's the kubelet-config.


root@node1:~# cat /etc/kubernetes/kubelet-config.yaml
```
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
nodeStatusUpdateFrequency: "10s"
failSwapOn: True
authentication:
  anonymous:
    enabled: false
  webhook:
    enabled: True
  x509:
    clientCAFile: /etc/kubernetes/ssl/ca.crt
authorization:
  mode: Webhook
staticPodPath: /etc/kubernetes/manifests
cgroupDriver: cgroupfs
maxPods: 110
address: 172.31.6.69
readOnlyPort: 0
kubeletCgroups: /systemd/system.slice
clusterDomain: cluster.local
clusterDNS:
- 169.254.25.10
kubeReserved:
  cpu: 200m
  memory: 512M
```

2. eksctl

eksctl also enables Webhook authentication and authorization for the kubelet. Here's the kubelet-config.

root@node1:~# cat /etc/eksctl/kubelet.yaml
```
address: 0.0.0.0
apiVersion: kubelet.config.k8s.io/v1beta1
authentication:
  anonymous:
    enabled: false
  webhook:
    cacheTTL: 2m0s
    enabled: true
  x509:
    clientCAFile: /etc/eksctl/ca.crt
authorization:
  mode: Webhook
  webhook:
[...]

PR #9718, opened 2020-08-09
Default kubelet authorization-mode to Webhook for k8s 1.19+

johngmyers:
Kubelet defaults the authorization mode to `Webhook` when the `--config` flag is provided, but it defaults to the insecure `AlwaysAllow` mode when the `--config` flag is not provided.

Since kops uses (the deprecated) flags instead of a config file, this change causes us to get the more secure default.

### DIFF
--- a/nodeup/pkg/model/kubelet.go
+++ b/nodeup/pkg/model/kubelet.go
@@ -513,6 +513,10 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 		c.NodeLabels = nil
 	}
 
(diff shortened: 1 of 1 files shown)


hakman:
Maybe you should add a short description?
Seems similar to https://github.com/kubernetes/kops/issues/7200.

johngmyers:
It might make sense to change the  default of `AuthenticationTokenWebhook` at the same time, but kubelet does not change its default based on whether or not there is a config file.

johngmyers:
On second thought, `AuthenticationTokenWebhook` is orthogonal. Instead of disabling insecure features it is enabling additional functionality. A change to its default should be a separate PR.

hakman:
Cool. Got it now.
/lgtm

PR #9719, opened 2020-08-10
Update shared-VPC documentation

### LINKED ISSUE #232: Support for existing subnets and route tables on AWS
For a simple setup it may be suitable for kops/kubernetes to setup subnets and route tables. More complicated setups may require custom routes, ingress/egress rules, etc

Support for private subnets will require specifying a custom default route. We use the managed GW service now but many will be using NAT boxes, direct connects, VPN's, etc.

In any case it may be desirable, and certainly is in our case, to have the all VPC networking managed in the same place.

johngmyers:
Clean up and remove outdated information.

Closes #232

### DIFF
--- a/docs/run_in_existing_vpc.md
+++ b/docs/run_in_existing_vpc.md
@@ -1,8 +1,8 @@
 ## Running in a shared VPC
 
-When launching into a shared VPC, the VPC & the Internet Gateway will be reused. If you are not using an internet gateway
- or NAT gateway you can tell _kops_ to ignore egress. By default we create a new subnet per zone, and a new route table, 
- but you can also use a shared subnet (see [below](#shared-subnets)).
+When launching into a shared VPC, kops will reuse the VPC and Internet Gateway. If you are not using an Internet Gateway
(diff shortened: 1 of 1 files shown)


docs/run_in_existing_vpc.md
```diff
@@ -1,8 +1,8 @@
 ## Running in a shared VPC
 
-When launching into a shared VPC, the VPC & the Internet Gateway will be reused. If you are not using an internet gateway
- or NAT gateway you can tell _kops_ to ignore egress. By default we create a new subnet per zone, and a new route table, 
- but you can also use a shared subnet (see [below](#shared-subnets)).
+When launching into a shared VPC, _kops_ will reuse the VPC and Internet Gateway. If you are not using an Internet Gateway
```
olemarkus, line 3: Nit: do we need to emphasise kops?
If so, we should make that a convention for all our docs.

johngmyers: I was trying to be consistent within the doc, but better to be consistent with the other docs. I took the italics out.

Issue #9698, opened 2020-08-06
KubeAPIErrorBudgetBurn warnings fire after upgrade to 1.18.6


demisx:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Version 1.18.0

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-16T00:04:31Z", GoVersion:"go1.14.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

Cluster upgrade from 1.16 to 1.18

**5. What happened after the commands executed?**
The following warning alert started to fire (see below). I am not sure where to look for the problem and how to fix it.

```
alert: KubeAPIErrorBudgetBurn
expr: sum(apiserver_request:burnrate3d)
  > (1 * 0.01) and sum(apiserver_request:burnrate6h) > (1 * 0.01)
for: 3h
labels:
  severity: warning
annotations:
  message: The API server is burning too much error budget
  runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubeapierrorbudgetburn
```

**6. What did you expect to happen?**
No warning alerts

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2019-10-22T17:23:28Z"
  generation: 10
  name: prod.dl.org
spec:
  api:
    loadBalancer:
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudProvider: aws
[...]

PR #9722, opened 2020-08-10
Implement Openstack cloudmock, add integration test

rifelpet:
This implements basic CRUD logic for all of the openstack resources, a followup to #9698 

Most of the resource types are pretty straight forward except:
* zones and recordsets live under the same /zones path prefix
* loadbalancer responses have their pools and listener IDs embedded in them
* routers and routerinterfaces live under the same /routers path prefix
* the security group rule's `remote_group_id` query param behaves differently than other filters - if it isn't provided that indicates the rules should not have a remote_group_id field, rather than not filtering on the field at all.
 
Next steps:
* server creation just uses a hardcoded IP address. I can update this to use sequential IPs in their subnet in a followup PR 
* loadbalancers and pools dont have their /members subresource yet. 
* I believe loadbalancer creation creates a VIP automatically but I need to look into the docs more
* I'll need to update this for octavia support
* We'll want a "complex_openstack" integration test that has a more complete `spec.cloudConfig.openstack` 

This is a large PR but most of the file changes are all independent so the "context depth" for reviewing this shouldn't be too bad, but if you want me to break up the first commit further I can do that.


olemarkus:
The test and the clusterspec looks good. I didn't go through all the mock implementations in detail though. I suspect tweaks are needed anyway once we try to add a more complex spec.

There is no way to go in and verify that e.g dns servers was correctly set on a subnet after the tests has run right? One would need something like Heat or Terraform support to validate that? Which doesn't even guarantee that the direct output works as expected.

/lgtm

rifelpet:
Right, this wont verify that values are set properly. Terraform support would take us a step further since it has input validation, but neither would catch issues like "the wrong IP was used" or "there's a typo in userdata". Those could only be caught by having e2e jobs which would definitely be my preferred route but we've discussed that a few times :) hopefully we can get the infra in place for that sometime.

In the mean time, this integration test will catch issues like:
* nil pointer dereference in openstack models or openstacktasks
* openstacktasks dependency cycles
* inaccurate openstacktasks Find logic
* missing resource types in pgk/resources/openstack

xoxbet:
> This implements basic CRUD logic for all of the openstack resources, a followup to #9698

Hello @rifelpet, are you sure you've mentioned correct issue here?

rifelpet:
@xoxbet oops you're right, that was supposed to be https://github.com/kubernetes/kops/pull/9691. Thanks for catching that

PR #9732, opened 2020-08-11
Add --internal flag for export kubecfg that targets the internal dns name

### LINKED ISSUE #800: Support internal DNS names when exporting kubecfg
When I created my cluster, I specified the `--admin-access` flag to the CIDR of a management subnet in my VPC. I connect to the management subnet using a VPN server, which gives me network access to the machines in my cluster.

When I run a `kops export kubecfg`, it populates my `~/.kube.config` using the public DNS entry for my cluster instead, which isn't reachable from my workstation. If I change the `server` for the cluster in my `~/.kube/config` file to the "internal" DNS name that was create for the cluster, everything works. Unfortuneatly, every time I update the cluster, the `~/.kube/config` file resets the `server` attribute to the public DNS name.

Would it make sense to have a flag to support using the internal hostname in `~/.kube/config`?

If so, I'd be happy to give implementing it a shot, although I may need a few hints on implementation ideas. 

If not, is there a recommended way to setup kubectl when accessing the cluster over a private network? I'd be happy to submit a PR with some documentation updates if someone wants to help me explain the approach.

### LINKED ISSUE #8009: Allow vault to be a secret store for PKI information and other secrets

rifelpet:
Kops creates an "api.internal.$clustername" dns A record that points to the master IP(s)

This adds a flag that will use that name and force the CA cert to be included.
This is a workaround for client certificate authentication not working on API ELBs with ACM certificates.
The ELB has a TLS listener rather than TCP, so the client certificate is not passed through to the apiserver.
Using --internal will bypass the API ELB so that the client certificate will be passed directly to the apiserver.

This also requires that the masters' security groups allow 443 access from the client which this does not handle automatically. I suppose if sslCertificateId is set, kops could automatically open 443 on the masters to the same sources that the ELB's 443 allows but I could see that being frowned upon for security reasons.

Fixes #800
(wow)

/hold for comment

### DIFF
--- a/cmd/kops/export_kubecfg.go
+++ b/cmd/kops/export_kubecfg.go
@@ -46,6 +46,9 @@ var (
 
 	# export using a user already existing in the kubeconfig file
 	kops export kubecfg kubernetes-cluster.example.com --user my-oidc-user
+
+	# export using the internal DNS name, bypassing the cloud load balancer
+	kops export kubecfg kubernetes-cluster.example.com --internal
 		`))
 
 	exportKubecfgShort = i18n.T(`Export kubecfg.`)
@@ -56,6 +59,7 @@ type ExportKubecfgOptions struct {
 	all            bool
 	admin          time.Duration
 	user           string
+	internal       bool
 }
 
 func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
@@ -80,6 +84,7 @@ func NewCmdExportKubecfg(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().DurationVar(&options.admin, "admin", options.admin, "export a cluster admin user credential with the given lifetime and add it to the cluster context")
 	cmd.Flags().Lookup("admin").NoOptDefVal = kubeconfig.DefaultKubecfgAdminLifetime.String()
 	cmd.Flags().StringVar(&options.user, "user", options.user, "add an existing user to the cluster context")
+	cmd.Flags().BoolVar(&options.internal, "internal", options.internal, "use the cluster's internal DNS name")
 
 	return cmd
 }
@@ -130,7 +135,7 @@ func RunExportKubecfg(ctx context.Context, f *util.Factory, out io.Writer, optio
 			return err
 		}
 
-		conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, buildPathOptions(options), options.admin, options.user)
+		conf, err := kubeconfig.BuildKubecfg(cluster, keyStore, secretStore, &commands.CloudDiscoveryStatusStore{}, buildPathOptions(options), options.admin, options.user, options.internal)
 		if err != nil {
 			return err
 		}
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -48,7 +48,9 @@ spec:
       idleTimeoutSeconds: 300
 ```
 
-You can use a valid SSL Certificate for your API Server Load Balancer. Currently, only AWS is supported:
+You can use a valid SSL Certificate for your API Server Load Balancer. Currently, only AWS is supported.
+
+Note that when using `sslCertificate`, client certificate authentication, such as with the credentials generated via `kops export kubecfg`, will not work through the load balancer. As of Kops 1.19, a `kubecfg` that bypasses the load balancer may be created with the `--internal` flag to `kops update cluster` or `kops export kubecfg`. Security groups may need to be opened to allow access from the clients to the master instances' port TCP/443, for example by using the `additionalSecurityGroups` field on the master instance groups.
 
 ```yaml
 spec:
--- a/cmd/kops/update_cluster.go
+++ b/cmd/kops/update_cluster.go
@@ -70,6 +70,7 @@ type UpdateClusterOptions struct {
 	CreateKubecfg bool
 	admin         time.Duration
 	user          string
+	internal      bool
 
 	Phase string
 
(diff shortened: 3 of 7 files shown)


cmd/kops/export_kubecfg.go
```diff
@@ -45,6 +45,9 @@ var (
 
 	# export using a user already existing in the kubeconfig file
 	kops export kubecfg kubernetes-cluster.example.com --user my-oidc-user
+
+	# export using the internal DNS name which bypasses the cloud load balancer
+	kops export kubecfg kubernetes-cluster.example.com --internal
```
olemarkus, line 50: Maybe add a quick note here that you need an additional SG?

cmd/kops/export_kubecfg.go
```diff
@@ -45,6 +45,9 @@ var (
 
 	# export using a user already existing in the kubeconfig file
 	kops export kubecfg kubernetes-cluster.example.com --user my-oidc-user
+
+	# export using the internal DNS name which bypasses the cloud load balancer
```
johngmyers, line 49:
```suggestion
	# export using the internal DNS name, bypassing the cloud load balancer
```

docs/cluster_spec.md
```diff
@@ -48,7 +48,9 @@ spec:
       idleTimeoutSeconds: 300
 ```
 
-You can use a valid SSL Certificate for your API Server Load Balancer. Currently, only AWS is supported:
+You can use a valid SSL Certificate for your API Server Load Balancer. Currently, only AWS is supported.
+
+Note that when using `sslCertificate`, client certificate authentication will no longer work through the load balancer, for example credentials generated via `kops export kubecfg`. This can be bypassed with the `--internal` flag which will target the master instances directly rather than the load balancer. Security groups may need to be opened to allow access from the clients to the master instances' port TCP/443, for example by using the `additionalSecurityGroups` field on the master instance groups.
```
johngmyers, line 53:
```suggestion
Note that when using `sslCertificate`, client certificate authentication, such as with the credentials generated via `kops export kubecfg`, will not work through the load balancer. A `kubecfg` that bypasses the load balancer may be created with the `--internal` flag to `kops update cluster` or `kops export kubecfg`. Security groups may need to be opened to allow access from the clients to the master instances' port TCP/443, for example by using the `additionalSecurityGroups` field on the master instance groups.
```

cmd/kops/export_kubecfg.go
hakman: Or maybe should add by default 443 when `sslCertificateId` is enabled and export "internal"?

olemarkus: I think that is a good idea!
Non-internal endpoints doesn't work anyway with this setup, right?

rifelpet: the problem with modifying security groups is that `kops export kubecfg` is decoupled from where SG definitions are managed in `kops update cluster`. At `update cluster` time we don't know if the cluster operator will ever be running `kops export kubecfg --internal`. while we do generate the kubecfg in `update cluster --yes`, I suppose we could add the security group rule if the `--internal` flag is passed to `update cluster` but that could lead to confusion. Users running `kops export kubecfg --internal` would need to know to include `--internal` in their `update cluster` commands as well. I don't know if we have precedent for making a "pair" of commands require similar flags like that.

hakman: Can we just add the port if sslCertificateId is enabled?
The export part can be explained in docs or printed at update cluster summary.

rifelpet:
The specific scenario in which we're relying on kubernetes api access with `export kubecfg` is our CI pipeline that runs `rolling-update cluster`. My concern with this approach is that as the masters are rolled, we're now more reliant on DNS TTLs to maintain API access which could be problematic. I haven't had the opportunity to test this yet (we just reenabled basic auth to unblock our upgrades), I'm hoping to do that before office hours.

cmd/kops/export_kubecfg.go
rifelpet: I think always adding the rule if sslCertificateId is present might be reasonable. It assumes that anyone using an ACM cert will want the ability to use client cert auth for api access. I'm failing to think of scenarios in which that wouldn't be the case. I haven't looked into the code but I'm assuming we can use the same sources (CIDRs or SG IDs) we use on the API ELB for the instance 443 as well. I don't see a need to split that out into its own flag or diverge those settings.

One could go a step further and imply the `--internal` flag's behavior when exporting kubecfg if `sslCertificateId` and `--admin` are set. In fact 1.19 might be the best time to do that because the --admin flag is being added, so we could document this change in behavior along side the new flag.

hakman: 👍

johngmyers: I could see a site with an alternate authentication provider not wanting the ability to use client cert auth from outside the cluster. But I see little reason to restrict anything that can hit the API ELB from hitting the instance's port 443.

I agree that `sslCertificateId` and `--admin` should imply the `--internal` flag's behavior.

rifelpet: After more thought, creating the additional SG rules when `kops update cluster --internal` would be problematic for organizations in which multiple people run their kops commands manually. If someone were to forget to include `--internal` then kops would revoke the SG rules.

It would be much preferable to add the SG rules based on fields in the ClusterSpec. Either `sslCertificateId` being set or we add an additional field that allows users to toggle the behavior. Which we choose depends on if we think anyone with `sslCertifiateId` set would _not_ want the SG rules added.

Regardless, I think adding `--internal` while also implying `--internal` when `sslCertificateId` and `--admin` are used is a good idea.

rifelpet:
I think this is ready for review. It isn't the ideal solution for the ACM certificate issue but the original ask was independent of that. We can go ahead and support this even with a better cert solution later on at which point we can update the cluster_spec.md text.

PR #9737, opened 2020-08-12
Remove unused Tags and FSRoot from NodeUp

hakman:
/kind cleanup


nodeup/pkg/bootstrap/install.go
```diff
@@ -33,21 +32,17 @@ import (
...
 type Installation struct {
-	FSRoot          string
 	CacheDir        string
 	RunTasksOptions fi.RunTasksOptions
 	Command         []string
 }
 
 func (i *Installation) Run() error {
-	distribution, err := distros.FindDistribution(i.FSRoot)
+	_, err := distros.FindDistribution("/")
```
rifelpet, line 41: I think we could get rid of the string argument in `FindDistribution()` since it is only ever passed `"/"`

hakman: Maybe we want to add some tests for it in the future, or use in some container?

cmd/nodeup/main.go
```diff
@@ -101,7 +101,6 @@ func main() {
 			i := bootstrap.Installation{
 				CacheDir: flagCacheDir,
 				Command:  command,
-				FSRoot:   flagRootFS,
```
johngmyers, line 104: You should probably also remove the `rootfs` flag.

hakman: Good point. Forgot about it.

PR #9739, opened 2020-08-12
Add an integration test for openstack floating ip

olemarkus:
* Integration test for floatingip cluster
* Implements mocking of floatingIP (only list for now)
* Expands various cloudmocks
* Fixes an NPR in openstack validation
* Fixes a bug where kops tries to use DNS even if the cluster is gossip

/cc @rifelpet 

/area provider/openstack


cloudmock/openstack/mocknetworking/floatingips.go
```diff
@@ -0,0 +1,76 @@
...
+		defer m.mutex.Unlock()
+
+		w.Header().Add("Content-Type", "application/json")
+
+		floatingIPID := re.ReplaceAllString(r.URL.Path, "")
+		switch r.Method {
+		case http.MethodGet:
+			if floatingIPID == "" {
+				r.ParseForm()
+				m.listFloatingIPs(w, r.Form)
```
rifelpet, line 47: This mock of floating IPs only has "list" support. How do they get created? does kops create them (in which case we need to look for create requests) or does the creation of another resource cascade to floating IPs? Or are they expected to already exist, in which case we should either create them in integrationtestharness.go or create them in mockFloatingIPs like I did for availability zones

olemarkus: The floating IP is implicitly created when the port is added to the external network, I believe.

There is something similar with routerinterface too. A router interface is just a port that is implicitly created when calling add_router_interface. Looks like openstack has a lot of these implicit creations.

rifelpet: I also see them being created here: https://github.com/kubernetes/kops/blob/fefe747f83e4c5d9fec331e6e5fb244e9b184cf0/upup/pkg/fi/cloudup/openstacktasks/floatingip.go#L210-L236

so this might need to do support creation requests if this kops code is exercised in our integration tests

olemarkus: Yep. We need to add more integration tests that exercise other configurations.

rifelpet:
glad to see the cloudmock catching bugs already 👍 
/lgtm
/approve

Issue #5070, opened 2018-04-26
Need to expose Kubelet --config flag in KubeletConfigSpec


tvi:
We are running 1.10.1 cluster and our logs contain following log line
```
kubelet[8594]: Flag --max-pods has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
```
The problem is that we cannot specify --config file and KubeletConfiguration spec.

ihoegen:
I'm not seeing it exposed in 

https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/componentconfig.go

so it'll have to be added.

It's super easy to expose, see:

https://github.com/kubernetes/kops/blob/master/docs/development/api_updates.md

Add the struct value, generate api machinery and you'll be good to

sdanbury:
/remove-lifecycle rotten

I am currently hitting this. This is what I see when I `ps aux | grep kubelet` on one of the worker nodes, so I can confirm kops is setting the flag correctly:
```
/opt/kubernetes/bin/kubelet --feature-gates=ExperimentalCriticalPodAnnotation=true...
```

But then in the logs you see the same messages shown above, one of those being:
```
Flag --feature-gates has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/
```

Did anybody get anywhere on this or should I take a look?

evanmcclure:
/remove-lifecycle rotten

Any updates?

mars64:
/remove-lifecycle rotten

I think this is still a desirable feature.

rifelpet:
Agreed, given that kubelet is deprecating the old style of providing its configuration I definitely think we should migrate to defining the config file automatically.

/lifecycle frozen

PR #9749, opened 2020-08-13
Map --config flag in KubeletConfigSpec

mdanielolsson:
Most config parameters passed as flags on the kubelet has been deprecated in favor of the `--config` parameter which expects a file as input. (https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/)

I read issue #5070 that the suggestion is to migrate away from the old style of migrating configuration to the kubelet, so I think this would be a good first step to just expose the flag. That way users can start migrating configuration to file before they are forced to do so.

### DIFF
--- a/pkg/apis/kops/componentconfig.go
+++ b/pkg/apis/kops/componentconfig.go
@@ -198,13 +198,13 @@ type KubeletConfigSpec struct {
 	RegistryBurst *int32 `json:"registryBurst,omitempty" flag:"registry-burst"`
 	//TopologyManagerPolicy determines the allocation policy for the topology manager.
 	TopologyManagerPolicy string `json:"topologyManagerPolicy,omitempty" flag:"topology-manager-policy"`
-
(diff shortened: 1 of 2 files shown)


johngmyers:
I don't see how exposing the flag is helpful. There would need to be code in kops for creating the config file that the flag points to and the settings exposed in the spec would need to be mapped to settings in that file instead of to flags.

olemarkus:
The correct thing to do would be to parse our API to configfile params and compile the configfile. We already do this for some flags on KubeScheduler.

johngmyers:
Converting our API struct to a `KubeletConfiguration` would be more complicated than the current code. Given the `KubeletConfiguration` format is beta, it would also increase our exposure to the risk of format changes.

hakman:
We should probably move to `KubeletConfiguration` in 1.20. I doubt the format will change much. Maybe the API version from time to time, but we can handle it.

hakman:
Closing, per the above comments.

PR #9754, opened 2020-08-14
Move kubernetes 1.18 from alpha to stable

mariusv:
Signed-off-by: Marius Voila <myself@mariusv.com>

### DIFF
--- a/channels/stable
+++ b/channels/stable
@@ -58,7 +58,7 @@ spec:
     networking:
       kubenet: {}
   kubernetesVersions:
-  - range: ">=1.18.0-alpha.1"
+  - range: ">=1.18.0"
     recommendedVersion: 1.18.6
     requiredVersion: 1.18.0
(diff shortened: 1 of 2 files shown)


hakman:
Thanks for fixing these @mariusv!
/ok-to-test

channels/stable
```diff
@@ -58,7 +58,7 @@ spec:
     networking:
       kubenet: {}
   kubernetesVersions:
-  - range: ">=1.18.0-alpha.1"
+  - range: ">=1.18.0"
```
hakman, line 61: Please do this change in the alpha channel also.

mariusv: sure thing, in the same PR or should I open a new PR for those?

hakman: Same PR would be ok.

mariusv: done

Issue #9750, opened 2020-08-14
KOPS (1.19 master) aws-iam-authenticator - clusterID value defaults to 'MasterPublicName' instead of 'clusterID' from the config map (mountedFile)


fred-vogt:
@rifelpet - FYI

kubectl error:
```
error: You must be logged in to the server (Unauthorized)
```

aws-iam-authenticator server error:
```
could not verify token: sts getCallerIdentity failed: error from AWS (expected 200, got 403). 
Body:
{
  "Error": {
    "Code": "SignatureDoesNotMatch",
    "Message": "The request signature we calculated does not match the signature you provided. 
                Check your AWS Secret Access Key and signing method. 
                Consult the service documentation for details.",
    "Type": "Sender"
  },
  "RequestId": "uuid"
}
```

The only possible way to get this error message on the server is if different values for `x-k8s-aws-id` were used between token creation versus verification.

[upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template](https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template)
```yaml
{{- if or .Authentication.Aws.ClusterID (not (contains "MountedFile" .Authentication.Aws.BackendMode))}}
- --cluster-id={{ or .Authentication.Aws.ClusterID .MasterPublicName }}
{{- end }}
```

The [docs](https://kops.sigs.k8s.io/authentication/) use an example of:
```yaml
# cluster spec
spec:
  authentication:
    aws: {}

# CM
data:
  config.yaml: |
    clusterID: my-dev-cluster.example.com
```
However `clusterID: my-dev-cluster.example.com` would be ignored and `-i api.clusterID: my-dev-cluster.example.com` would be required in kube-config.

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
- Unreleased master version:  `Version 1.19.0-alpha.2 (git-cdbb780)`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
N/A.

**3. What cloud provider are you using?**
AWS.

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
Spec:
```yaml
spec:
  authentication:
    aws: {}
```
aws-iam-authenticator CM:
```yaml
data:
  config.yaml: |
    clusterID: <cluster-fqdn>
```

**5. What happened after the commands executed?**
[...]

PR #9755, opened 2020-08-14
Fix backwards compatibility when backend mode isn't set

WarpRat:
This will address the issue opened here: https://github.com/kubernetes/kops/issues/9750

I changed the test to confirm that the gotemplate or function worked as expected. I ran the test multiple times locally with a different mix of configurations to confirm that this somewhat ugly chain of if/or/not works as intended and will address the issue.


upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
```diff
@@ -131,8 +131,8 @@ spec:
         {{- if or (not .Authentication.Aws.BackendMode) (contains "MountedFile" .Authentication.Aws.BackendMode) }}
         - --config=/etc/aws-iam-authenticator/config.yaml
         {{- end }}
-        {{- if or .Authentication.Aws.ClusterID (not (contains "MountedFile" .Authentication.Aws.BackendMode))}}
-        - --cluster-id={{ or .Authentication.Aws.ClusterID .MasterPublicName }}
+        {{- if or .Authentication.Aws.ClusterID (or (not .Authentication.Aws.BackendMode) (not (contains "MountedFile" .Authentication.Aws.BackendMode)))}}
```
johngmyers, line 134: Can this be simplified?

Can this be simplified?

```suggestion
        {{- if or .Authentication.Aws.ClusterID (not .Authentication.Aws.BackendMode) (not (contains "MountedFile" .Authentication.Aws.BackendMode))}}
```

fred-vogt:
@WarpRat - Excellent work my friend. Thanks for jumping on this so quickly.

I'm still building an integration around the newly available backendMode(s).

WarpRat:
@fred-vogt - no problem, feel foolish that I missed this in the first place. I still need to make a minor update to bring the docs in line with the new default cluster id which @rifelpet suggested and the minor simplification that @johngmyers suggested above, but I should be able to handle that in an hour or two once I'm off work.

upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
```diff
@@ -131,8 +131,8 @@ spec:
         {{- if or (not .Authentication.Aws.BackendMode) (contains "MountedFile" .Authentication.Aws.BackendMode) }}
         - --config=/etc/aws-iam-authenticator/config.yaml
         {{- end }}
-        {{- if or .Authentication.Aws.ClusterID (not (contains "MountedFile" .Authentication.Aws.BackendMode))}}
-        - --cluster-id={{ or .Authentication.Aws.ClusterID .MasterPublicName }}
+        {{- if or .Authentication.Aws.ClusterID (not .Authentication.Aws.BackendMode) (not (contains "MountedFile" .Authentication.Aws.BackendMode))}}
```
rifelpet, line 134: I'm not sure i understand the logic change here.

The old logic would add `--cluster-id` if any of the following:
* ClusterID API field is set
* BackendMode does not contain MountedFile (including if BackendMode is not set)

This new logic will add `--cluster-id` if any of the following:
* ClusterID API field is set
* BackendMode API field is not set
* BackendMode API field does not contain MountedFile

Wouldn't we want to _not_ set --cluster-id if BackendMode is not set in order to preserve original behavior? 

```suggestion
        {{- if or .Authentication.Aws.ClusterID (and .Authentication.Aws.BackendMode (not (contains "MountedFile" .Authentication.Aws.BackendMode)))}}
```

The above would add `--cluster-id` if any of the following:
* ClusterID API field is set
* BackendMode API field is set and does not contain MountedFile

upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
WarpRat: Much clearer. Thanks!

upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
WarpRat: Yeah, you're right. In my local tests I saw the `--cluser-id` flag get replaced with the `--config` and didn't even notice the clusterID flag was just a few lines lower. I've tested all the different possibilities you mentioned here and with your suggested change everything is working as expected. Thanks for catching that.

rifelpet:
👍 
/lgtm
/approve

PR #9761, opened 2020-08-15
Setup a second ELB listener when an AWS ACM certificate is used 

rifelpet:
This is the alternative solution to #9756

When `sslCertificate` is set, the API ELB will also have a TCP listener on 8443 that will pass through the TLS session to the apiserver pods, allowing client certificate auth.

When `--admin` is specified in `kops update cluster` or `kops export kubecfg`, the kubeconfig file will use this port and will include the cluster's CA.

If anyone has a preference for port number other than 8443 I'm happy to change it.

### DIFF
--- a/pkg/kubeconfig/create_kubecfg.go
+++ b/pkg/kubeconfig/create_kubecfg.go
@@ -99,11 +99,17 @@ func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.Se
 
 	b := NewKubeconfigBuilder()
 
+	// Use the secondary load balancer port if a certificate is on the primary listener
+	if admin != 0 && cluster.Spec.API != nil && cluster.Spec.API.LoadBalancer != nil && cluster.Spec.API.LoadBalancer.SSLCertificate != "" {
+		server = server + ":8443"
+	}
+
 	b.Context = clusterName
 	b.Server = server
 
-	// add the CA Cert to the kubeconfig only if we didn't specify a SSL cert for the LB or are targeting the internal DNS name
-	if cluster.Spec.API == nil || cluster.Spec.API.LoadBalancer == nil || cluster.Spec.API.LoadBalancer.SSLCertificate == "" || internal {
+	// add the CA Cert to the kubeconfig only if we didn't specify a certificate for the LB
+	//  or if we're using admin credentials and the secondary port
+	if admin != 0 || cluster.Spec.API == nil || cluster.Spec.API.LoadBalancer == nil || cluster.Spec.API.LoadBalancer.SSLCertificate == "" || internal {
 		cert, _, _, err := keyStore.FindKeypair(fi.CertificateIDCA)
 		if err != nil {
 			return nil, fmt.Errorf("error fetching CA keypair: %v", err)
--- a/pkg/model/awsmodel/api_loadbalancer.go
+++ b/pkg/model/awsmodel/api_loadbalancer.go
@@ -111,6 +111,7 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 
 		if lbSpec.SSLCertificate != "" {
 			listeners["443"] = &awstasks.LoadBalancerListener{InstancePort: 443, SSLCertificateID: lbSpec.SSLCertificate}
+			listeners["8443"] = &awstasks.LoadBalancerListener{InstancePort: 443}
 		}
 
 		if lbSpec.SecurityGroupOverride != nil {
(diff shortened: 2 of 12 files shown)


pkg/kubeconfig/create_kubecfg.go
```diff
@@ -91,6 +91,11 @@ func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.Se
 
 	b := NewKubeconfigBuilder(configAccess)
 
+	// Use the secondary load balancer port if an ACM certificate is on the primary listener
+	if admin && cluster.Spec.API != nil && cluster.Spec.API.LoadBalancer != nil && cluster.Spec.API.LoadBalancer.SSLCertificate != "" {
+		server = server + ":8443"
+	}
```
johngmyers, line 97: nit: belongs in the other commit.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -174,11 +175,15 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 	// Create security group for API ELB
 	var lbSG *awstasks.SecurityGroup
 	{
+		removeExtraRules := []string{"port=443"}
+		if lbSpec.SSLCertificate != "" {
+			removeExtraRules = append(removeExtraRules, "port=8443")
```
johngmyers, line 180: We should probably do this even when `lbspec.SSLCertificate` is empty. Otherwise will disabling the option clear the extra rule for 8443?

rifelpet: good call, I've updated it to always remove extra rules on 8443 too.

pkg/kubeconfig/create_kubecfg.go
rifelpet: thanks, fixed.

johngmyers:
Looks like the Terraform provider needs to consistently sort the listeners on an ELB.

rifelpet:
We might need to go back to the drawing board on this one :( I finally had a chance to test it out:

```
error running task "LoadBalancer/api.foo.k8s.local" (9m54s remaining to succeed): error creating LoadBalancerListeners: InvalidConfigurationRequest: Listeners can't talk to InstancePort 443 with secure and insecure protocols at the same time
```

johngmyers:
Might be worth an AWS ticket to let them know of the use case, but that probably won't help us.

hakman:
What if also open port 8443 on host?

rifelpet:
Opening a second port on the instances would require we forward traffic on that port to apiserver. I suppose the apiserver health check container is performing a similar function right now so we could extend it and give it a more generic name. That will vastly increase the scope of this change though, so it'd be nice to get agreement on that prior to implementation.

rifelpet:
Ok I tried to use a second hostPort in the kube-apiserver pod spec and pointed the second ELB listener at the second port but that didn't work because `hostNetwork: true` requires matching `containerPort` and `hostPort` (of course)

`Aug 20 21:14:54 ip-XX-XX-XX-XX kubelet[6631]: E0820 21:14:54.409160    6631 file.go:187] Can't process manifest file "/etc/kubernetes/manifests/kube-apiserver.manifest": invalid pod: [spec.containers[0].ports[1].containerPort: Invalid value: 443: must match `hostPort` when `hostNetwork` is true]`

So it seems like if we want to keep pursuing the second listener option we'd need to have something listening on a second instance port (hijacking the healthcheck container is sounding better and better...)

Beyond that, I did notice when upgrading an existing cluster, we'll need to remind users not to use `update cluster --admin` before `rolling-update cluster` or else the generated kubeconfig file will be pointing to the new port which wont work until at least one master has been replaced, causing rolling-update to fail. Therefor we should recommend `kops update cluster --yes; kops rolling-update cluster --yes; kops update cluster --yes --admin`

rifelpet:
I wonder if NLBs don't have the duplicate instancePort limitation that ELBs do. We have an open PR to add NLB support, maybe it's worth focusing on getting that merged for 1.19. That would be much more elegant than our other proposals.

I'll test an NLB tomorrow unless someone else gets a chance first.

seh:
One note on NLBs, in case you consider their use for from the control plane components as well: They only support hairpin connections from EC2 instances registered as targets **by their IP address**, and **not by their instance ID**. That precludes registering the instances automatically by virtue of their management by an ASG, as that registration method registers them by instance ID.

Instead, you have register and deregister the EC2 instances when they boot or shut down (such as via systemd units), but of course it's possible for the machines to die without the deregistration procedure running. A controller running elsewhere would be more resilient. though more complicated.

rifelpet:
Superseded by #10157 
/close

PR #9762, opened 2020-08-15
Add command for listing cloud instances

### LINKED ISSUE #7902: How to display which instances are in NeedsUpdate state?

When running `kops rolling-update cluster` I might see output like:

```
NAME                    STATUS          NEEDUPDATE      READY   MIN     MAX     NODES
bastions                Ready           0               1       1       1       0
master-us-east-1a       Ready           0               1       1       1       1
master-us-east-1b       Ready           0               1       1       1       1
master-us-east-1c       Ready           0               1       1       1       1
nodes                   NeedsUpdate     2               1       2       40      2

Must specify --yes to rolling-update.
```

This is great, and if I want to continue and actually roll the cluster, this is enough.

However, I have a use-case where instead of rolling one of the InstanceGroups, I want to first see _which_ nodes specifically need updating.

A `kops rolling-update cluster -o json` (or `-o yaml`) may be enough - I would envision it returning the full details of _which_ nodes need to be updated, in a format that I can easily post-process with other tooling.

olemarkus:
I find myself often either using the AWS api or `kops toolbox dump` to either get an instance id for `mssh`/ec2-instance-connect or just listing nodes that have spawned before they join the cluster. Often also useful way of listing nodes during development when masters/nodes fail to start for whatever reasons.

This PR implements `kops get nodes`, which would be the kops equivalent to `kubectl get nodes`. 

A few things still missing, but I wanted to get some feedback on this before cleaning up, providing docs etc.

Fixes  #7902


cmd/kops/get_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf("cannot build kubernetes api client for %q: %v", contextName, err)
+	}
+	nodeList, err := k8sClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		klog.V(2).Infof("error listing nodes: %v", err)
+	}
+	nodes := cloudinstances.GetNodeMap(nodeList.Items, cluster)
+
+	var instances []*resources.Instance
+	switch cloud.ProviderID() {
```
johngmyers, line 122: Would it make sense to use `fi.Cloud.GetCloudGroups()` instead?

olemarkus: I was thinking of using `GetCloudGroups` to output `NeedsUpdate` on instance level (also requested in #7902).
I don't think that function works without access to the API though, so it doesn't replace this switch block.

olemarkus:
I think this now does what I want it to do. There are many other things that one could add to a command like this. Instance age, image, instance group etc, but suggest that is done in a follow-up if wanted.

olemarkus:
I think this now does what I want it to do. There are many other things that one could add to a command like this. Instance age, image, instance group etc, but suggest that is done in a follow-up if wanted.

pkg/resources/aws/aws.go
```diff
@@ -530,16 +530,34 @@ func DumpInstance(op *resources.DumpOperation, r *resources.Resource) error {
...
+		Name: r.Name,
 	}
 	for _, networkInterface := range ec2Instance.NetworkInterfaces {
 		if networkInterface.Association != nil {
 			publicIP := aws.StringValue(networkInterface.Association.PublicIp)
 			if publicIP != "" {
 				i.PublicAddresses = append(i.PublicAddresses, publicIP)
 			}
+			i.PrivateAddress = aws.StringValue(networkInterface.PrivateIpAddress)
+
```
hakman, line 560: Private IP is not related to Association.
```suggestion
		i.PrivateAddress = aws.StringValue(networkInterface.PrivateIpAddress)
		if networkInterface.Association != nil {
			publicIP := aws.StringValue(networkInterface.Association.PublicIp)
			if publicIP != "" {
				i.PublicAddresses = append(i.PublicAddresses, publicIP)
			}
```

cmd/kops/get_nodes.go
johngmyers: Where does that function access the API? It's not even given an API client.

You can pass it a slice of Node, but instances don't need to be in that slice.

johngmyers: This switch block is only useful if there's a bug in `GetCloudGroups` that causes it to not return instances that this block would.

cmd/kops/get_instances.go
```diff
@@ -0,0 +1,242 @@
...
+	default:
+		return fmt.Errorf("Unsupported output format: %q", options.output)
+	}
+}
+
+func instanceOutputTable(instances []*resources.Instance, status map[string]string, out io.Writer) error {
+	t := &tables.Table{}
+	t.AddColumn("NAME", func(i *resources.Instance) string {
+		return i.Name
+	})
```
hakman, line 161: In my case, all instances of the same IG are named the same. Easier than just having an IG column instead?
Also, maybe order should be consistent with table columns order.

cmd/kops/get_instances.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+	case kops.CloudProviderOpenstack:
+		rs, _ := openstack.ListResources(cloud.(osCloudup.OpenstackCloud), options.clusterName)
+		for _, r := range rs {
+			if r.Type == "Instance" {
+				instances = append(instances, openstack.GetInstanceFromResource(r))
+			}
+		}
+	default:
+		return fmt.Errorf("cloud not supported")
```
hakman, line 146:
```suggestion
		return fmt.Errorf("cloud provider not supported")
```

cmd/kops/get_nodes.go
olemarkus: The two functions return different representation of the instance though (we may have too many different representations of an instance in kops ..). The `CloudInstanceMember` type doesn't have much of the data that is useful for this command, so we either way need to use an additional abstraction in addition.

cmd/kops/get_instances.go
olemarkus: In AWS you typically have the same name for all nodes, but OS does not. So there are some value in the name column there. IG could be useful, but not that easy to get to. Could perhaps add it to `resources.Instance`. Easy enough to extract it from tags there.

hakman: Let's leave as is. Thanks.

pkg/resources/aws/aws.go
```diff
@@ -542,26 +558,24 @@ func DumpInstance(op *resources.DumpOperation, r *resources.Resource) error {
...
+		if key == "kops.k8s.io/instancegroup" {
+			i.InstanceGroup = fi.StringValue(tag.Value)
+		}
 		if !strings.HasPrefix(key, awsup.TagNameRolePrefix) {
 			continue
 		}
 		role := strings.TrimPrefix(key, awsup.TagNameRolePrefix)
 		i.Roles = append(i.Roles, role)
+		if key == "kops.k8s.io/instancegroup" {
+			klog.Info("huh")
```
hakman, line 573: Guess this is not supposed to be here.

olemarkus: absolutely not. Thanks for catching

pkg/resources/aws/aws.go
```diff
@@ -542,26 +558,23 @@ func DumpInstance(op *resources.DumpOperation, r *resources.Resource) error {
...
+			i.InstanceGroup = fi.StringValue(tag.Value)
+		}
 		if !strings.HasPrefix(key, awsup.TagNameRolePrefix) {
 			continue
 		}
 		role := strings.TrimPrefix(key, awsup.TagNameRolePrefix)
 		i.Roles = append(i.Roles, role)
+		if key == "kops.k8s.io/instancegroup" {
+			i.InstanceGroup = fi.StringValue(tag.Value)
+		}
```
johngmyers, line 574: Duplicate with block above

pkg/validation/node_conditions.go
```diff
@@ -39,11 +39,11 @@ func findNodeCondition(node *v1.Node, conditionType v1.NodeConditionType) *v1.No
 	return nil
 }
 
-// isNodeReady returns if a Node is considered ready.
+// IsNodeReady returns if a Node is considered ready.
 // It is considered ready if:
 // 1) its Ready condition is set to true
 // 2) doesn't have NetworkUnavailable condition set to true
-func isNodeReady(node *v1.Node) bool {
+func IsNodeReady(node *v1.Node) bool {
```
johngmyers, line 46: Note that is can return `true` when the node can't actually be productive. A system-node-critical pod, such as the CNI, might not be ready or a static pod, such as kube-proxy, might be absent.

cmd/kops/get_nodes.go
johngmyers: It might be reasonable to add the fields to `CloudInstanceMember`. The AWS ones don't look too expensive to always do for cluster validation.

The switch block is a code smell. If not `GetCloudGroups` it should probably be a new method on `fi.Cloud`.

olemarkus:
/hold
Should be reimplemented using the work from #9799

johngmyers:
Obvious extentions would be to add options for filtering the list, including the `--instance-group` and `--instance-group-roles` options of `kops rolling-update cluster`.

/lgtm
/approve

hakman:
Nicely simplified from previous implementation. Looking forward to use this!

PR #9766, opened 2020-08-16
Use /etc/os-release to identify the distribution

hakman:
All supported distributions have the `/etc/os-release` file describing the distro info:
http://0pointer.de/blog/projects/os-release


justinsb:
OK this is a good simplification.  Pondering whether we should fallback to the existing logic for a period... how did you get the example os-release files that you use in the tests?

hakman:
@justinsb I got the files from starting various VMs for the respective distros and copying the `/etc/*-release` files.

hakman:
If a fallback is added, will be very hard to catch any corner cases. 
The chage should be verified quite easily in the distros periodic tests, but also there is enough time until 1.19 to address any feedback from users.

PR #9768, opened 2020-08-17
WIP: Refactor apply cluster to split out task building

justinsb:
This will enable reuse in cluster-api, but also starts to better
organize what has become a very intricate function.

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -134,11 +134,11 @@ type ApplyClusterCmd struct {
 	TaskMap map[string]fi.Task
 }
 
-func (c *ApplyClusterCmd) Run(ctx context.Context) error {
+func (c *ApplyClusterCmd) buildLoader(ctx context.Context, cloud fi.Cloud, keyStore fi.CAStore, secretStore fi.SecretStore, sshCredentialStore fi.SSHCredentialStore, assetBuilder *assets.AssetBuilder) (*Loader, error) {
 	if c.InstanceGroups == nil {
 		list, err := c.Clientset.InstanceGroupsFor(c.Cluster).List(ctx, metav1.ListOptions{})
 		if err != nil {
-			return err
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -675,12 +585,89 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
...
+	}
+
+	return l, nil
+}
+
+func (c *ApplyClusterCmd) Run(ctx context.Context) error {
+	assetBuilder := assets.NewAssetBuilder(c.Cluster, string(c.Phase))
+
+	// Note that upgradeSpecs writes back to c.Cluster
+	if err := c.upgradeSpecs(assetBuilder); err != nil {
```
johngmyers, line 605: Should probably be renamed to `populateSpecs`

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -675,12 +585,89 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	configBase, err := vfs.Context.BuildVfsPath(cluster.Spec.ConfigBase)
+	if err != nil {
+		return fmt.Errorf("error parsing config base %q: %v", cluster.Spec.ConfigBase, err)
+	}
+
+	if !c.AllowKopsDowngrade {
```
johngmyers, line 641: In the existing code this and `configBase` came before `BuildCloud()`, etc. Why move it later?

olemarkus:
I think that all the validation checks that takes place here could be moved to a VFS/cloud-aware validation package. That would also be nice to have rather than relying on failed fi.tasks with good error messages. When we have started running tasks it is so easy to end up with a half-provisioned cloud because of e.g typo in the subnet id or similar. Doing some checks up front would make provisioning take longer, but give a better user experience.

PR #9771, opened 2020-08-17
Issue more certs out of kops-controller

johngmyers:


pkg/model/iam/iam_builder.go
```diff
@@ -453,25 +453,28 @@ func ReadableStatePaths(cluster *kops.Cluster, role kops.InstanceGroupRole) ([]s
...
 			"/instancegroup/*",
 			"/pki/issued/*",
-			"/pki/private/kube-proxy/*",
 			"/pki/ssh/*",
 			"/secrets/dockerconfig",
 		)
 
 		// @check if bootstrap tokens are enabled and if so enable access to client certificate
 		if model.UseKopsControllerForNodeBootstrap(cluster) {
 			// no additional permissions
```
rifelpet, line 462: Is there a potential race when enabling kops-controller bootstrapping?

1. ASG launches an instance
2. `kops update cluster --yes` revokes these s3 permissions from the node IAM policy
3. instance tries to pull certs from s3 and fails

Issue #7286, opened 2019-07-19
Support permission boundaries for AWS IAM Roles


dvianello:
A while ago AWS released [Permissions Boundaries](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_boundaries.html) to allow Admins to delegate creation of Roles to users while retaining control on the maximum set of permissions that can be delegated. This is needed to avoid situations where users would be able to escalate privileges attaching a policy with admin permission to a role they created & then assume it. We're extensively using this in our low-tier environments to allow users to play with everything they want while ensuring they won't be able to do things such as removing key resources that we use for monitoring.

However, all of this breaks when trying to use `kops` since it tries to create roles straight away without a boundary. While we could create roles and instance profiles upfront and feed them to `kops` as described [here](https://github.com/kubernetes/kops/blob/master/docs/iam_roles.md#use-existing-aws-instance-profiles), but this will quickly become a problem to manage should kops change the required permissions. Also, I assume `additionalPolicies` in the spec will stop working the moment we start using 

```--lifecycle-overrides IAMRole=ExistsAndWarnIfChanges,IAMRolePolicy=ExistsAndWarnIfChanges,IAMInstanceProfileRole=ExistsAndWarnIfChanges``` to prevent `kops` from creating IAM roles.

What we would really need is `kops` being able to apply a permission boundary on role creation starting from a variable defined in the spec.

We don't have a lot of Go experience in house, but we'd be happy to help if we'd be able to get some guidance on what / how to touch. Our assumption is that most of the changes should be in [iam.go](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/awstasks/iamrole.go),
but we're not completely sure of how to do this cleanly.

Thanks for the help!

rifelpet:
can you describe exactly what needs to change in order to support permission boundaries? From my limited understanding of permission boundaries it seems like the only change is in the policy attached to the IAM entity running the kops command, not any of the policies that kops creates.

If you can paste an AWS error message or suggest how the IAM policies need to change in order to support this, that would be appreciated.

And yes depending on what needs to change exactly, it would probably either be in the iam.go you linked to or [iam_builder.go](https://github.com/kubernetes/kops/blob/af4df08b694e2a1f8814a7b3649060477be67c86/pkg/model/iam/iam_builder.go)

dvianello:
Hi @rifelpet,

[...]

PR #9773, opened 2020-08-17
Adding support for permission boundaries for AWS IAM Roles

victorfrancax1:
Refs #7286 
Also adding Terrraform and Cloudformation support.


rifelpet:
overall this looks good. Can you add an integration test case for this? It should just involve setting the new api field with a dummy value in the two `in-*.yaml` files [here](https://github.com/kubernetes/kops/tree/master/tests/integration/update_cluster/complex) and running `./hack/update-expected.sh` and the terraform and cloudformation files in the same directory should be updated to match. 

My only question is around the API design. Would it make sense to ever have separate permission boundaries for the master role vs the node role? If so we'd want the API to be able to support that.

victorfrancax1:
Hey @rifelpet , thanks for reviewing! I added an integrated test case as requested.

About the API design: although being possible, it's seems very unlikely to me that people would need to set permission boundaries per instance role, since they were designed mainly to prevent privilege escalation. The furthest I would go personally is configuring different boundaries for each cluster, which is supported by this implementation. As stated by the author of the original issue (can't tag him here for some reason), should an edge case arise, we could suggest that they provide their own roles to `kops`. What do you think?

upup/pkg/fi/cloudup/awstasks/iamrole.go
```diff
@@ -147,6 +148,10 @@ func (_ *IAMRole) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *IAMRole) error
 		request.AssumeRolePolicyDocument = aws.String(policy)
 		request.RoleName = e.Name
 
+		if e.PermissionsBoundary != nil {
+			request.PermissionsBoundary = e.PermissionsBoundary
```
rifelpet, line 152: Up in the Find function above (~ line 74) you'll need to read the PermissionsBoundary field from the AWS SDK's Role and set it on the `awstasks.IAMRole`, similar to name and id fields on those lines.

victorfrancax1: Fixed. Thanks for pointing that out :)

victorfrancax1: I realized that I wasn't addressing possible PermissionsBoundary updates in the `RenderAws` function. Added that logic too.

pkg/model/iam.go
```diff
@@ -110,6 +110,11 @@ func (b *IAMModelBuilder) buildIAMTasks(igRole kops.InstanceGroupRole, iamName s
 				RolePolicyDocument: fi.WrapResource(rolePolicy),
 				ExportWithID:       s(strings.ToLower(string(igRole)) + "s"),
 			}
+
+			if b.Cluster.Spec.IAM.PermissionsBoundary != nil {
```
rifelpet, line 114: I think we'll need an extra nil check here
```suggestion
			if b.Cluster.Spec.IAM != nil && b.Cluster.Spec.IAM.PermissionsBoundary != nil {
```

rifelpet:
one last comment and I think this will be good to merge :) thanks for adding this!

pkg/model/iam.go
victorfrancax1: Fixed :)

rifelpet:
/lgtm
/approve

This will land in Kops 1.19. I believe there will be another alpha release soon if you want to use it in an official build.

Thanks again!

rifelpet:
One more thing @victorfrancax1 would you mind adding some docs [here](https://github.com/kubernetes/kops/blob/master/docs/iam_roles.md)? This can be done in a separate PR.

PR #9783, opened 2020-08-19
Update Weave Net to v2.7.0

hakman:
https://github.com/weaveworks/weave/releases/tag/v2.7.0

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
@@ -1,3 +1,5 @@
+# Pulled and modified from: https://github.com/weaveworks/weave/releases/download/v2.7.0/weave-daemonset-k8s-1.11.yaml
+
 {{- if WeaveSecret }}
 apiVersion: v1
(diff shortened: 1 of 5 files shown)


upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
```diff
@@ -253,8 +253,6 @@ spec:
           operator: Exists
         - effect: NoExecute
           operator: Exists
-        - key: CriticalAddonsOnly
```
rifelpet, line 256: Why the removal of this toleration?

Also maybe we should add a comment at the top that links to where this manifest (loosely) came from: https://github.com/weaveworks/weave/blob/9f00f78d3b9d5a8a31fdd90ec691095028e2690a/prog/weave-kube/weave-daemonset-k8s-1.11.yaml

hakman: Some confusion from me. Added back and also the header.
One other thing, had to move from the 1.8 manifest (that doesn't exist anymore) to the 1.19 one.

PR #9784, opened 2020-08-19
Add kops delete instance command

olemarkus:
This command makes it possible to delete a single instance.
It is implemented mostly reusing rollingupdate code. This meant opening more of the API than I liked, but given the intricate way that `RollingUpdate` is implemented, it was hard to get around. Ideally some of the logic could be moved to `cloudinstances` but quickly ran into `fi` circular imports when going down that route.


cmd/kops/delete_instance.go
```diff
@@ -0,0 +1,277 @@
...
+	deleteIgExample = templates.Examples(i18n.T(`
+		# Delete an instance from the currently active cluster.
+		kops delete instance i-0a5ed581b862d3425
+
+		# Delete an instance from the currently active cluster without
+		validation or draining.
+		kops delete instance --cloudonly i-0a5ed581b862d3425
+		`))
+
+	deleteIgShort = i18n.T(`Delete instancegroup`)
```
rifelpet, line 98: there are some copy/paste references to `instancegroup` including `Ig` in variable names.

olemarkus: Should be cleaned up now.

olemarkus:
Moved some stuff into rollingupgrade and made the functions there private again. Feels cleaner this way.

pkg/instancegroups/instancegroups.go
```diff
@@ -561,3 +561,21 @@ func (c *RollingUpdateCluster) deleteNode(ctx context.Context, node *corev1.Node
 
 	return nil
 }
+
+func (c *RollingUpdateCluster) RollSingleInstance(ctx context.Context, cloudMember *cloudinstances.CloudInstanceGroupMember, detach bool) error {
```
johngmyers, line 565:
```suggestion
func (c *RollingUpdateCluster) UpdateSingleInstance(ctx context.Context, cloudMember *cloudinstances.CloudInstanceGroupMember, detach bool) error {
```

pkg/instancegroups/instancegroups.go
```diff
@@ -561,3 +561,21 @@ func (c *RollingUpdateCluster) deleteNode(ctx context.Context, node *corev1.Node
...
+	if detach {
+		if cloudMember.CloudInstanceGroup.InstanceGroup.IsMaster() {
+			return fmt.Errorf("cannot detatch master instances. Run with --detatch=false")
```
johngmyers, line 568: I think it would be better to just ignore `detach` in this case. Having to reissue the command with a difficult to type flag is not the best user experience.

pkg/instancegroups/instancegroups.go
```diff
@@ -561,3 +561,21 @@ func (c *RollingUpdateCluster) deleteNode(ctx context.Context, node *corev1.Node
...
+
+		}
+		err := c.detachInstance(cloudMember)
+		if err != nil {
+			return fmt.Errorf("failed to detach instance: %v", err)
```
johngmyers, line 573: The most likely cause of this would be that the cloud provider doesn't support surging. It might be better to default `detach` to false for such cloud providers.

pkg/instancegroups/instancegroups.go
```diff
@@ -561,3 +561,21 @@ func (c *RollingUpdateCluster) deleteNode(ctx context.Context, node *corev1.Node
...
+		}
+		if err := c.maybeValidate(" after detaching instance", c.ValidateCount); err != nil {
+			return err
+		}
+	}
+
+	return c.drainTerminateAndWait(ctx, cloudMember, false, 0)
```
johngmyers, line 580: If the instance is a bastion, this will incorrectly try to drain it.

If the instance is a master, you might want to `maybeValidate` afterwards.

pkg/instancegroups/instancegroups.go
```diff
@@ -561,3 +561,21 @@ func (c *RollingUpdateCluster) deleteNode(ctx context.Context, node *corev1.Node
 
 	return nil
 }
+
```
johngmyers, line 564: Should have doc comment. Should have added unit test(s).

cmd/kops/delete_instance.go
```diff
@@ -0,0 +1,263 @@
...
+				break groupsloop
+			}
+		}
+		for _, r := range group.NeedUpdate {
+			if r.ID == options.InstanceID {
+				cloudMember = r
+				break groupsloop
+			}
+		}
+	}
```
johngmyers, line 227: Possibly also match against r.Node.Name (when Node not nil)?

pkg/instancegroups/instancegroups.go
olemarkus: I missed that this was a param passed on. Changed the functions to figure out the instance group role on its own instead.

hakman:
@olemarkus to keep things consistent with other commands, this one needs `--yes` also.

Issue #9661, opened 2020-07-31
Add flag to skip the NTP task in NodeUp


hakman:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

There are case in which operators don't want the settings provided by Kops and there are easy ways to include it in the base image or do something similar via hooks. One of the cases is: https://github.com/kubernetes/kops/issues/9307#issuecomment-643766267.

**2. Feel free to provide a design supporting your feature request.**

Main question is what flag to use for that:
* skipInstall - as for docker
* disabled
* unmanaged
* ...

johngmyers:
Generally, I'd prefer we consistently use either `enabled: false` or `enableFoo: false`.

hakman:
> Generally, I'd prefer we consistently use either `enabled: false` or `enableFoo: false`.

Good point, but `enabled` implies management of some sort, so operators might expect `enabled: false` to disable `Foo`.

PR #9787, opened 2020-08-20
Implement cluster autoscaler as bootstrap addon

olemarkus:
We see that when managing a larger number of clusters, getting the cluster autoscaler version correct is a bit of a hassle. We also need to take care to properly update CAS after cluster upgrades etc. So adding CAS as a configurable bootstrap addon seems like a good idea to make this more manageable.

If no one objects to this, I will finish this and add documentation and validation.


MoShitrit:
That's a great idea, thanks for adding that @olemarkus !!

MoShitrit:
BTW, I guess it would be also worth adding some documentation for it? Or will that be done through a separate PR?

olemarkus:
Docs will come as part of this PR. Just want to ensure there are no objections before starting on that.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -444,18 +444,39 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 	if kubeDNS.Provider == "CoreDNS" && NodeLocalDNS != nil && fi.BoolValue(NodeLocalDNS.Enabled) {
 		{
 			key := "nodelocaldns.addons.k8s.io"
-			version := "1.18.0"
+			version := "1.19.0"
```
johngmyers, line 447: This looks like an unrelated concern.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -126,6 +126,7 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 
 	// will return openstack external ccm image location for current kubernetes version
 	dest["OpenStackCCMTag"] = tf.OpenStackCCMTag
+	dest["CasTag"] = tf.CasTag
```
johngmyers, line 129: This seems a bit terse. Perhaps `ClusterAutoscalerTag`

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -526,3 +527,23 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
...
+		case 19:
+			tag = "v1.19.0"
+		case 18:
+			tag = "v1.18.2"
+		case 17:
+			tag = "v1.17.3"
+		case 16:
+			tag = "v1.16.6"
+		case 15:
+			tag = "v1.15.7"
```
johngmyers, line 545: I think I'd prefer an error for the default case. Otherwise we might forget to update it for a new Kubernetes minor version.

pkg/model/components/clusterautoscaler.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	if cas.Expander == nil {
+		cas.Expander = fi.String("least-waste")
+	}
+	if cas.ScaleDownUtilizationThreshold == nil {
+		cas.ScaleDownUtilizationThreshold = fi.String("0.5")
+	}
+	if cas.SkipNodesWithLocalStorage == nil {
+		cas.SkipNodesWithLocalStorage = fi.Bool(true)
```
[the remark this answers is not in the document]
olemarkus: 100%. It is a bad default from upstream.

docs/cluster_spec.md
```diff
@@ -616,6 +616,20 @@ spec:
...
+## Cluster autoscaler
+
+{{ kops_feature_table(kops_added_default='1.19', k8s_min='1.15') }}
+
+Cluster autoscaler can be enabled to automatically adjust the size of the kubernetes cluster.
+
+```yaml
+spec:
+  cluster:
+    clusterAutoscaler: {}
```
johngmyers, line 628: Would be better to give an example that disables the two `SkipNodesWith` options.

pkg/apis/kops/cluster.go
```diff
@@ -191,6 +191,9 @@ type ClusterSpec struct {
 	SysctlParameters []string `json:"sysctlParameters,omitempty"`
 	// RollingUpdate defines the default rolling-update settings for instance groups
 	RollingUpdate *RollingUpdate `json:"rollingUpdate,omitempty"`
+
+	// ClusterAutoscaler defines the cluaster autoscaler configuration
```
johngmyers, line 195:
```suggestion
	// ClusterAutoscaler defines the cluaster autoscaler configuration.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -755,6 +755,26 @@ type CloudConfiguration struct {
 	Openstack *OpenstackConfiguration `json:"openstack,omitempty"`
 }
 
+// ClusterAutoscalerConfig determines the cluster autoscaler configuration
+type ClusterAutoscalerConfig struct {
+	// Expander determines the strategy for which instance group gets expanded
+	// Supported values: least-waste, most-pods, random
+	// Default: least-waste
```
johngmyers, line 762:
```suggestion
	// Expander determines the strategy for which instance group gets expanded.
	// Supported values: least-waste, most-pods, random.
	// Default: least-waste
```

Otherwise the generated docs look odd.

pkg/apis/kops/componentconfig.go
```diff
@@ -755,6 +755,26 @@ type CloudConfiguration struct {
...
+	Expander *string `json:"expander,omitempty"`
+	// BalanceSimilarNodeGroups makes cluster autoscaler treat similar node groups as one
```
johngmyers, line 764:
```suggestion
	// BalanceSimilarNodeGroups makes cluster autoscaler treat similar node groups as one.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -755,6 +755,26 @@ type CloudConfiguration struct {
...
+	// Default: false
+	BalanceSimilarNodeGroups *bool `json:"balanceSimilarNodeGroups,omitempty"`
+	// ScaleDownUtilizationThreshold determines the utilization threshold for node scale-down
```
johngmyers, line 767:
```suggestion
	// ScaleDownUtilizationThreshold determines the utilization threshold for node scale-down.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -755,6 +755,26 @@ type CloudConfiguration struct {
...
+	// Default: 0.5
+	ScaleDownUtilizationThreshold *string `json:"scaleDownUtilizationThreshold,omitempty"`
+	// SkipNodesWithSystemPods makes cluster autoscaler skip scale-down of nodes with system pods
```
johngmyers, line 770:
```suggestion
	// SkipNodesWithSystemPods makes cluster autoscaler skip scale-down of nodes with non-DaemonSet pods in the kube-system namespace.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -755,6 +755,26 @@ type CloudConfiguration struct {
...
+	// Default: true
+	SkipNodesWithSystemPods *bool `json:"skipNodesWithSystemPods,omitempty"`
+	// SkipNodesWithLocalStorage makes cluster autoscaler skip scale-down of nodes with local storage
```
johngmyers, line 773:
```suggestion
	// SkipNodesWithLocalStorage makes cluster autoscaler skip scale-down of nodes with local storage.
```

pkg/apis/kops/v1alpha2/cluster.go
```diff
@@ -189,6 +189,9 @@ type ClusterSpec struct {
 	SysctlParameters []string `json:"sysctlParameters,omitempty"`
 	// RollingUpdate defines the default rolling-update settings for instance groups
 	RollingUpdate *RollingUpdate `json:"rollingUpdate,omitempty"`
+
+	// ClusterAutoscaler defines the cluaster autoscaler configuration
```
johngmyers, line 193:
```suggestion
	// ClusterAutoscaler defines the cluaster autoscaler configuration.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -755,6 +755,26 @@ type CloudConfiguration struct {
 	Openstack *OpenstackConfiguration `json:"openstack,omitempty"`
 }
 
+// ClusterAutoscalerConfig determines the cluster autoscaler configuration
```
johngmyers, line 758:
```suggestion
// ClusterAutoscalerConfig determines the cluster autoscaler configuration.
```

pkg/apis/kops/v1alpha2/componentconfig.go
```diff
@@ -756,6 +756,26 @@ type CloudConfiguration struct {
...
+	// BalanceSimilarNodeGroups makes cluster autoscaler treat similar node groups as one
+	// Default: false
+	BalanceSimilarNodeGroups *bool `json:"balanceSimilarNodeGroups,omitempty"`
+	// ScaleDownUtilizationThreshold determines the utilization threshold for node scale-down
+	// Default: 0.5
+	ScaleDownUtilizationThreshold *string `json:"scaleDownUtilizationThreshold,omitempty"`
+	// SkipNodesWithSystemPods makes cluster autoscaler skip scale-down of nodes with system pods
+	// Default: true
+	SkipNodesWithSystemPods *bool `json:"skipNodesWithSystemPods,omitempty"`
+	// SkipNodesWithLocalStorage makes cluster autoscaler skip scale-down of nodes with local storage
```
johngmyers, line 774: Same comment fixups as for the non-versioned API.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -447,15 +447,36 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 			version := "1.18.0"
 
 			{
-				location := key + "/k8s-1.12.yaml"
-				id := "k8s-1.12"
+				location := key + "/k8s-1.15.yaml"
+				id := "k8s-1.15"
```
johngmyers, line 451: This still looks like an unrelated change.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -526,3 +528,34 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
 	}
 	return tag
 }
+
+// GetInstanceGroup returns the instance group with the specified name
```
johngmyers, line 532:
```suggestion
// GetInstanceGroups returns a map containing the defined instance groups.
```

k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -335,6 +335,28 @@ spec:
               cloudProvider:
                 description: The CloudProvider to use (aws or gce)
                 type: string
+              clusterAutoscaler:
+                description: ClusterAutoscaler defines the cluaster autoscaler configuration.
```
gjtempleton, line 339:
```suggestion
                description: ClusterAutoscaler defines the cluster autoscaler configuration.
```

pkg/apis/kops/cluster.go
```diff
@@ -191,6 +191,9 @@ type ClusterSpec struct {
 	SysctlParameters []string `json:"sysctlParameters,omitempty"`
 	// RollingUpdate defines the default rolling-update settings for instance groups
 	RollingUpdate *RollingUpdate `json:"rollingUpdate,omitempty"`
+
+	// ClusterAutoscaler defines the cluaster autoscaler configuration.
```
gjtempleton, line 195:
```suggestion
	// ClusterAutoscaler defines the cluster autoscaler configuration.
```

pkg/model/components/clusterautoscaler.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (b *ClusterAutoscalerOptionsBuilder) BuildOptions(o interface{}) error {
+	clusterSpec := o.(*kops.ClusterSpec)
+	cas := clusterSpec.ClusterAutoscaler
+	if cas == nil || !fi.BoolValue(cas.Enabled) {
+		return nil
+	}
+
+	if cas.Expander == nil {
+		cas.Expander = fi.String("least-waste")
```
gjtempleton, line 40: Is there a reason for deviating from the CA and setting `least-waste` as the default expander rather than `random`?

k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -335,6 +335,28 @@ spec:
...
+                properties:
+                  balanceSimilarNodeGroups:
+                    description: 'BalanceSimilarNodeGroups makes cluster autoscaler treat similar node groups as one. Default: false'
+                    type: boolean
+                  enabled:
+                    description: 'Enabled enables the cluster autoscaler. Default: false'
+                    type: boolean
+                  expander:
+                    description: 'Expander determines the strategy for which instance group gets expanded. Supported values: least-waste, most-pods, random. Default: least-waste'
```
gjtempleton, line 348: Is there a reason for not supporting `priority` and `price` expanders? 

Obviously would require some more docs/links out to CA docs in both cases.

pkg/model/components/clusterautoscaler.go
olemarkus: Think this was the default in the example I sourced from. No other specific reason.

k8s/crds/kops.k8s.io_clusters.yaml
olemarkus: Only because of simplicity. Shouldn't be a problem to add this later if anyone needs this. I'd rather not have that block this PR at least

gjtempleton: That seems reasonable enough.

pkg/apis/kops/validation/validation.go
```diff
@@ -1113,3 +1117,17 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
 
 	return allErrs
 }
+
+func validateClusterAutoscaler(cluster *kops.Cluster, spec *kops.ClusterAutoscalerConfig, fldPath *field.Path) (allErrs field.ErrorList) {
+	if !cluster.IsKubernetesGTE("1.15") || cluster.IsKubernetesGTE("1.20") {
+		allErrs = append(allErrs, field.Forbidden(fldPath, "Cluster autoscaler requires kubernetesVersion between 1.11 and 1.19"))
```
johngmyers, line 1123: I don't think there's any other validation that puts an upper bound on the Kubernetes version for a reason other than removal of support. Cilium, for example, allows the highest known Cilium version be used for arbitrary newer Kubernetes versions. So I don't think we should cap this either.

pkg/apis/kops/validation/validation.go
```diff
@@ -1113,3 +1117,17 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
...
+	}
+
+	allErrs = append(allErrs, IsValidValue(fldPath.Child("expander"), spec.Expander, []string{"least-waste", "random", "most-pods"})...)
+
+	if kops.CloudProviderID(cluster.Spec.CloudProvider) == kops.CloudProviderOpenstack {
+		allErrs = append(allErrs, field.Forbidden(fldPath, "Cluster autoscaler is not supported on OpenStack"))
```
johngmyers, line 1129: I suspect it might not be supported on Spotinst either.

pkg/model/components/clusterautoscaler.go
johngmyers: We should probably take the default from cluster-autoscaler itself. That appears to be `random`.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -526,3 +528,34 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
 	}
 	return tag
 }
+
+// GetNodeInstanceGroups returns a map containing the defined instance groups.
```
johngmyers, line 532:
```suggestion
// GetNodeInstanceGroups returns a map containing the defined instance groups of role "Node".
```

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -526,3 +528,34 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
...
+func (tf *TemplateFunctions) GetNodeInstanceGroups() map[string]kops.InstanceGroupSpec {
+	nodegroups := make(map[string]kops.InstanceGroupSpec)
+	for _, ig := range tf.KopsModelContext.InstanceGroups {
+		if ig.Spec.Role == kops.InstanceGroupRoleNode {
+			nodegroups[ig.ObjectMeta.Name] = ig.Spec
+		}
+	}
+	return nodegroups
+}
+
```
johngmyers, line 542:
```suggestion

// ClusterAutoscalerTag returns the image tag for cluster-autoscaler.
```

upup/pkg/fi/cloudup/template_functions.go
johngmyers: Please respond to this review comment.

pkg/apis/kops/validation/validation.go
olemarkus: The reason for this one was "remembering to add support for newer autoscaler". Original suggestion was to panic in the template function. Third option is to use the latest tag when we have an unknown/unsupported K8s version.

I suspect we'll bump the upper bound long before the first kops alpha that supports the new k8s version.

upup/pkg/fi/cloudup/template_functions.go
olemarkus: So this is why I had the validation logic instead. I just felt it nicer should we ever forget to add support here.

pkg/apis/kops/validation/validation.go
olemarkus: I suspect the same. They still roll their own, it seems

olemarkus: Doesn't seem like we treat spotinst fully as a cloudprovider. And cluster autoscaler will work for those IGs not managed by Spotinst. So I think it is best to leave it as it is for now. Spotinst is also behind a feature flag, so users know to take care.

johngmyers: Fair enough

pkg/apis/kops/validation/validation.go
hakman: Could 1.13 and 1.14 also be supported, this would cover all supported versions? Also validation lower limit is for 1.15 and message says 1.11.
I also think that shouldn't be an upper limit and just use the one for the previous version.

olemarkus: I just added the versions that had a release the last 6 months or so. Can add more, but I am not 100% sure the manifests will work that far back. 

My initial PR used latest tag for higher versions. But using latest release works for me too. Probably doesn't break anything.

hakman: I am ok with latest as the default tag, as long as we add the rest of the images. Probably the tag should be changed to full image name to allow older image names.
The Helm chart seems to work for Kubernetes 1.8+ with the same manifest, so should be ok.

pkg/model/components/clusterautoscaler.go
```diff
@@ -0,0 +1,75 @@
...
+			case 17:
+				tag = "v1.17.3"
+			case 16:
+				tag = "v1.16.6"
+			case 15:
+				tag = "v1.15.7"
+			}
+		}
+
+		cas.Image = fi.String("k8s.gcr.io/autoscaling/cluster-autoscaler:" + tag)
```
hakman, line 58:
```suggestion
		var image string
		v, err := util.ParseKubernetesVersion(clusterSpec.KubernetesVersion)
		if err == nil {
			switch v.Minor {
			case 19:
				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.19.0"
			case 18:
				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.18.2"
			case 17:
				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.17.3"
			case 16:
				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.16.6"
			case 15:
				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.15.7"
			case 14:
				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.14.8"
			case 13:
				image = "gcr.io/google-containers/cluster-autoscaler:v1.13.9"
			case 12:
				image = "gcr.io/google-containers/cluster-autoscaler:v1.12.8"
			default:
				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:latest"
			}

			cas.Image = fi.String(image)
```

pkg/apis/kops/validation/validation.go
hakman: Now that you added the image override, I don't mind if we have the upper limit for this as long as it is ignored when the image is overridden. Your call.

hakman:
@johngmyers leaving the approval to you, in case there is anything left from your point of view.
/lgtm

pkg/model/components/clusterautoscaler.go
```diff
@@ -0,0 +1,82 @@
...
+func (b *ClusterAutoscalerOptionsBuilder) BuildOptions(o interface{}) error {
+	clusterSpec := o.(*kops.ClusterSpec)
+	cas := clusterSpec.ClusterAutoscaler
+	if cas == nil || !fi.BoolValue(cas.Enabled) {
+		return nil
+	}
+
+	if cas.Image == nil {
+
+		image := "k8s.gcr.io/autoscaling/cluster-autoscaler:latest"
```
johngmyers, line 42: I'm not a big fan of mutable tags such as "latest"; I'd prefer using "v1.19.0" (or whatever the highest known version is) for newer Kubernetes versions.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -521,6 +521,27 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 		}
 	}
 
+	if b.Cluster.Spec.ClusterAutoscaler != nil && fi.BoolValue(b.Cluster.Spec.ClusterAutoscaler.Enabled) {
```
johngmyers, line 524: Perhaps we should default `Enabled` to true if `ClusterAutoscaler` is non-`nil`

johngmyers (commented):
Neither of the review comments is blocking. Holding to give opportunity to address.

/approve
/hold

pkg/model/components/clusterautoscaler.go
hakman: I don't have any preference here.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
hakman: We agreed to keep it disabled by default at last office hours.
Maybe in 1.20 we materialise it into to the cluster spec so that people know it's enabled.

johngmyers: But default is to have a `nil` `ClusterAutoscaler` field. Why would someone specify a `ClusterAutoscaler` field yet want `Enabled` to default to `false`?

olemarkus:
I am a bit back and forth on what I prefer on the above. So I think I'll just let this one go in and then do a follow-up. Especially if cluster-autoscaler should prove to be slower with their support for newer k8s in the future.
/hold cancel

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
hakman: I agree that it may be a good idea in general, just don't know how used this pattern is in kops.
Sometimes when I test or just want to disable something I just comment the `enabled: true` line, without actually but keep all other setting to make it easy to reenable.

One of the reasons for starting the discussion on https://github.com/kubernetes/kops/issues/9661 was to establish some guidelines for enabled/disabled and managed/unmanaged, but seems not very popular topic. Maybe should open a more generic one.

paalkr:
Thanks for contributing with this addition, that makes deploying the CAS easier. We have a use case that uses a mix of different cluster AutoScaling methods (external and CAS) for various node pools. I think it would be vice to add a keyword to the instance group spec to either include or exclude the group with CAS.

olemarkus:
Thanks. I'd add this as a separate issue. I may not be able to implement this in the near future, but it should be a pretty easy "good first issue".

avdhoot:
Might late to join the party. Thanks @olemarkus for this feature. It will remove extra step to manage  `autoscaler` version separately .  But it will be good if we add support for `priority`  expander.  When you have multiple instance group mix of on-demand & spot instance group. `priority`  expander help to set priority to spot instance group.

PR #9790, opened 2020-08-20
Remove compute floating ip extension

olemarkus:
Managing floating IPs through the compute API has been deprecated and removed in newer versions of the compute API. Migrate all to use the networking API instead.


zetaab (approved):
/approve
/lgtm

PR #9792, opened 2020-08-21
Don't use nova for glance mocks

olemarkus:
Fetching images through nova is deprecated and removed in newer versions
of the compute API. Mocks now reflect this behavior.

### DIFF
--- a/upup/pkg/fi/cloudup/openstacktasks/instance.go
+++ b/upup/pkg/fi/cloudup/openstacktasks/instance.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
+
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/bootfromvolume"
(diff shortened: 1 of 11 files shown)


upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -176,10 +178,17 @@ func (_ *Instance) ShouldCreate(a, e, changes *Instance) (bool, error) {
 func (_ *Instance) RenderOpenstack(t *openstack.OpenstackAPITarget, a, e, changes *Instance) error {
 	if a == nil {
 		klog.V(2).Infof("Creating Instance with name: %q", fi.StringValue(e.Name))
+		cloud := t.Cloud.(openstack.OpenstackCloud)
+
+		imageName := fi.StringValue(e.Image)
+		imageRef, err := images.IDFromName(cloud.ImageClient(), imageName)
```
olemarkus, line 184: Just a note on this change: If you pass image name to create server, gophercloud will do the nova equivalent for this `IDFromName` lookup. So instead, we explicitly use the glance version.
As a bonus, the error message for users should they have a typo in the image definition, is much more clear.

PR #9799, opened 2020-08-23
Cloudinstances refactor

olemarkus:
@johngmyers what do you think about an approach like this?

### DIFF
--- a/pkg/cloudinstances/cloud_instance.go
+++ b/pkg/cloudinstances/cloud_instance.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cloudinstances
+
+import v1 "k8s.io/api/core/v1"
+
+// CloudInstanceStatusDetached means the instance needs update and has been detached.
+const CloudInstanceStatusDetached = "Detached"
+
+// CloudInstanceStatusNeedsUpdate means the instance has joined the cluster, is not detached, and needs to be updated.
+const CloudInstanceStatusNeedsUpdate = "NeedsUpdate"
+
+// CloudInstanceStatusReady means the instance has joined the cluster, is not detached, and is up to date.
+const CloudInstanceStatusUpToDate = "UpToDate"
+
+// CloudInstance describes an instance in a CloudInstanceGroup group.
+type CloudInstance struct {
+	// ID is a unique identifier for the instance, meaningful to the cloud
+	ID string
+	// Node is the associated k8s instance, if it is known
+	Node *v1.Node
+	// CloudInstanceGroup is the managing CloudInstanceGroup
+	CloudInstanceGroup *CloudInstanceGroup
+	// Status indicates if the instance has joined the cluster and if it needs any updates.
+	Status string
+	// Roles are the roles the instance have.
+	Roles []string
+	// MachineType is the hardware resource class of the instance.
+	MachineType string
+	// Private IP is the private ip address of the instance.
+	PrivateIP string
+}
--- a/pkg/cloudinstances/cloud_instance_group.go
+++ b/pkg/cloudinstances/cloud_instance_group.go
@@ -30,8 +30,8 @@ type CloudInstanceGroup struct {
 	// HumanName is a user-friendly name for the group
 	HumanName     string
 	InstanceGroup *kopsapi.InstanceGroup
-	Ready         []*CloudInstanceGroupMember
-	NeedUpdate    []*CloudInstanceGroupMember
+	Ready         []*CloudInstance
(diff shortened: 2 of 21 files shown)


pkg/cloudinstances/cloud_instance_group.go
```diff
@@ -51,45 +51,22 @@ func (c *CloudInstanceGroup) NewCloudInstance(instanceId string, newGroupName st
...
+		cm.Status = status
 	} else {
 		klog.V(8).Infof("unable to find node for instance: %s", instanceId)
 		cm.Status = CloudInstanceStatusNotJoined
 	}
 	return nil
 }
 
-// NewDetachedCloudInstance creates a new CloudInstance for a detached instance
-func (c *CloudInstanceGroup) NewDetachedCloudInstance(instanceId string, nodeMap map[string]*v1.Node) error {
```
johngmyers: This change is incorrect. It means a detached instance that hasn't joined won't be recognized as detached.

pkg/cloudinstances/cloud_instance.go
```diff
@@ -40,4 +40,8 @@ type CloudInstance struct {
 	CloudInstanceGroup *CloudInstanceGroup
 	// Status indicates if the instance has joined the cluster and if it needs any updates.
 	Status string
+	// Roles is the roles the instance has
```
johngmyers, line 43:
```suggestion
	// Roles are the roles the instance have.
```

pkg/cloudinstances/cloud_instance.go
```diff
@@ -40,4 +40,8 @@ type CloudInstance struct {
...
+	Roles []string
+	// Machine type is the hardware resource class of the instance
```
johngmyers, line 45:
```suggestion
	// MachineType is the hardware resource class of the instance.
```

pkg/cloudinstances/cloud_instance.go
```diff
@@ -0,0 +1,43 @@
...
+const CloudInstanceStatusDetached = "Detached"
+
+// CloudInstanceStatusNeedsUpdate means the instance needs update
+const CloudInstanceStatusNeedsUpdate = "NeedsUpdate"
+
+// CloudInstanceStatusReady means the instance has joined the cluster and is up to date
+const CloudInstanceStatusReady = "Ready"
+
+// CloudInstanceStatusNotJoined means the instance has not yet joined the cluster
+const CloudInstanceStatusNotJoined = "NotJoined"
```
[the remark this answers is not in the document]
olemarkus: I suppose you are right. It gets a bit awkward either way. If we were to remove this, it would mean that we'd have ready nodes that have not yet joined the cluster. But not being able to detect detached nodes that are absent from the cluster would be worse.
Clould be that "ready" is not the right term for this status. Maybe it should be "UpToDate" or "Running" or something is better.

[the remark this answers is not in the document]
johngmyers: I think you're right that "Ready" is the wrong term for this. I believe with one of your other changes it could be in this state even when it's pending.

I think I have a slight preference for "UpToDate". Another possibility might be "Current".

Also, consider whether `instancegroups.AdjustNeedUpdate()` needs to change the `Status` field.

olemarkus:
Only adds the data I wanted for aws for now. Can add for openstack in #9762. That would require some other PRs to go in first anyway.

/retest

zetaab (approved):
from my point of view this looks good.

could @johngmyers check that everything is in place

/lgtm

pkg/instancegroups/rollingupdate_test.go
```diff
@@ -1397,8 +1397,13 @@ func TestRollingUpdateMaxSurgeAllNeedUpdateMaxAlreadyDetached(t *testing.T) {
 
 	groups := make(map[string]*cloudinstances.CloudInstanceGroup)
 	makeGroup(groups, c.K8sClient, cloud, "node-1", kopsapi.InstanceGroupRoleNode, 7, 7)
+	groups["node-1"].NeedUpdate[0].Status = cloudinstances.CloudInstanceStatusDetached
```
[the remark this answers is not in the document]
olemarkus: The name of the test is `TestRollingUpdateMaxSurgeAllNeedUpdateMaxAlreadyDetached` so I assumed all needs updates and two were already detached.
What would be the correct values here then?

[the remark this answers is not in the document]
johngmyers: 0 should be `CloudInstanceStatusNeedsUpdate`

pkg/instancegroups/instancegroups.go
```diff
@@ -321,7 +321,11 @@ func (c *RollingUpdateCluster) patchTaint(ctx context.Context, node *corev1.Node
 	return err
 }
 
+<<<<<<< HEAD
```
johngmyers, line 324: Unresolved merge conflict

olemarkus: Bah. Yesterday's push got stuck. Pushed now.

pkg/cloudinstances/cloud_instance.go
```diff
@@ -0,0 +1,46 @@
...
+// CloudInstance describes an instance in a CloudInstanceGroup group.
+type CloudInstance struct {
+	// ID is a unique identifier for the instance, meaningful to the cloud
+	ID string
+	// Node is the associated k8s instance, if it is known
+	Node *v1.Node
+	// CloudInstanceGroup is the managing CloudInstanceGroup
+	CloudInstanceGroup *CloudInstanceGroup
+	// Status indicates if the instance has joined the cluster and if it needs any updates.
+	Status string
```
johngmyers, line 39: Perhaps this should be its own type `CloudInstanceStatus`?

PR #9802, opened 2020-08-23
Update API slice fields to not use pointers

rifelpet:
A slice of pointers causes problems with the Kubernetes 1.19 code-generator.

A nil entry in these slices isn't valid anyways, so this should have no impact short of being a breaking change for anyone importing these API types.

### DIFF
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -231,7 +231,8 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchConfigurationTask(c *fi.ModelB
 	}
 
 	// @step: add any additional block devices to the launch configuration
-	for _, x := range ig.Spec.Volumes {
+	for i := range ig.Spec.Volumes {
+		x := &ig.Spec.Volumes[i]
 		if x.Type == "" {
 			x.Type = DefaultVolumeType
 		}
--- a/pkg/model/components/etcd.go
+++ b/pkg/model/components/etcd.go
@@ -50,7 +50,8 @@ const (
 func (b *EtcdOptionsBuilder) BuildOptions(o interface{}) error {
 	spec := o.(*kops.ClusterSpec)
 
-	for _, c := range spec.EtcdClusters {
+	for i := range spec.EtcdClusters {
+		c := &spec.EtcdClusters[i]
 		if c.Provider == "" {
 			if b.IsKubernetesGTE("1.12") {
 				c.Provider = kops.EtcdProviderTypeManager
@@ -110,11 +111,8 @@ func (b *EtcdOptionsBuilder) BuildOptions(o interface{}) error {
 				c.EnableTLSAuth = true
 			}
 		}
-	}
 
-	// Remap the well known images
-	for _, c := range spec.EtcdClusters {
-
+		// Remap the well known images
 		// We remap the etcd manager image when we build the manifest,
 		// but we need to map the standalone images here because protokube launches them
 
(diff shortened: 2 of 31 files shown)


pkg/model/components/etcdmanager/options.go
```diff
@@ -38,34 +38,35 @@ var _ loader.OptionsBuilder = &EtcdManagerOptionsBuilder{}
 // BuildOptions generates the configurations used to create etcd manager manifest
 func (b *EtcdManagerOptionsBuilder) BuildOptions(o interface{}) error {
 	clusterSpec := o.(*kops.ClusterSpec)
+	etcdClusters := clusterSpec.EtcdClusters
 
-	for _, etcdCluster := range clusterSpec.EtcdClusters {
-		if etcdCluster.Provider != kops.EtcdProviderTypeManager {
+	for i := range etcdClusters {
+		if etcdClusters[i].Provider != kops.EtcdProviderTypeManager {
```
hakman, line 44:
```suggestion
	for i := range etcdClusters {
		etcdCluster := &etcdClusters[i]
		if etcdCluster.Provider != kops.EtcdProviderTypeManager {
```

pkg/model/components/etcd.go
```diff
@@ -50,41 +50,42 @@ const (
 func (b *EtcdOptionsBuilder) BuildOptions(o interface{}) error {
 	spec := o.(*kops.ClusterSpec)
 
-	for _, c := range spec.EtcdClusters {
-		if c.Provider == "" {
+	etcdClusters := spec.EtcdClusters
+	for i := range etcdClusters {
+		if etcdClusters[i].Provider == "" {
```
hakman, line 55:
```suggestion
	for i := range spec.EtcdClusters {
		c := &spec.EtcdClusters[i]
 		if c.Provider == "" {
```

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -229,17 +229,19 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchConfigurationTask(c *fi.ModelB
 	}
 
 	// @step: add any additional block devices to the launch configuration
-	for _, x := range ig.Spec.Volumes {
-		if x.Type == "" {
-			x.Type = DefaultVolumeType
+	for i := range ig.Spec.Volumes {
+		if ig.Spec.Volumes[i].Type == "" {
+			ig.Spec.Volumes[i].Type = DefaultVolumeType
```
hakman, line 234:
```suggestion
		x := &ig.Spec.Volumes[i]
		if x.Type == "" {
 			x.Type = DefaultVolumeType
```

hakman (changes requested):
I think using a pointer to a slice member would reduce the number of changes.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -776,8 +776,8 @@ func setupNetworking(opt *NewClusterOptions, cluster *api.Cluster) error {
 		if errList := validation.ValidateEtcdVersionForCalicoV3(cluster.Spec.EtcdClusters[0], cluster.Spec.Networking.Calico.MajorVersion, field.NewPath("spec", "networking", "calico")); len(errList) != 0 {
 
 			// This is not a special version but simply of the 3 series
-			for _, etcd := range cluster.Spec.EtcdClusters {
-				etcd.Version = components.DefaultEtcd3Version_1_11
+			for i := range cluster.Spec.EtcdClusters {
+				cluster.Spec.EtcdClusters[i].Version = components.DefaultEtcd3Version_1_11
 			}
 		}
```
hakman, line 782: This should not exist at all. Removing in https://github.com/kubernetes/kops/pull/9803.

rifelpet:
Updated to use pointers when modifying elements in a for loop. diff looks much cleaner now 👍 
I'll wait for #9803 to merge before rebasing once more. just to confirm the containerd e2e job still passes.

PR #9804, opened 2020-08-24
Bump stable versions from alpha

MoShitrit:
11 days have passed since these version were pushed to alpha ([ref](https://github.com/kubernetes/kops/commit/9f768c127511b30c0094d2f2a7df3fd7da688ae7#diff-37c59376fa8938637370b21d9540af69)).
We should be good bumping to stable at this point.


mikesplain:
Thanks @MoShitrit!

/lgtm
/approve

PR #9808, opened 2020-08-25
Pull images from k8s.gcr.io/kops instead of docker.io/kope

hakman:
xRef: https://github.com/kubernetes/k8s.io/pull/1175


zetaab (approved):
/approve
/lgtm

PR #9811, opened 2020-08-25
Add all images to GH release

hakman:
/cc @rifelpet @justinsb


rifelpet:
I'm a fan of moving `nodeup` to the beginning of the asset names, it improves the sorting when the list becomes as long as it is 👍 
 
/lgtm
/approve

hakman:
@rifelpet I think this is just a step forward, but would rather see in that list just `kops-*` and `offline-*` that includes everything, with dir structure.

PR #9812, opened 2020-08-25
Support writing a full certificate chain

justinsb:
This means that our https endpoint will serve the ca.crt as well.


justinsb:

PR #9813, opened 2020-08-25
Expose JWKS via a feature-flag

justinsb:
When the PublicJWKS feature-flag is set, we expose the apiserver JWKS
document publicly (including enabling anonymous access).  This is a
stepping stone to a more hardened configuration where we copy the JWKS
document to S3/GCS/etc.


justinsb:

pkg/model/components/discovery.go
```diff
@@ -0,0 +1,83 @@
...
+		kubeAPIServer.ServiceAccountJWKSURI = &jwksURL
+	}
+
+	if kubeAPIServer.ServiceAccountSigningKeyFile == nil {
+		s := "/srv/kubernetes/server.key"
+		kubeAPIServer.ServiceAccountSigningKeyFile = &s
+	}
+
+	if len(kubeAPIServer.ServiceAccountKeyFile) == 0 {
+		kubeAPIServer.ServiceAccountKeyFile = []string{"/srv/kubernetes/server.key"}
```
rifelpet, line 79: would it be better to append this filepath if ServiceAccountKeyFile is non-empty?

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -141,6 +141,25 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 		}
 	}
 
+	if featureflag.PublicJWKS.Enabled() {
+		key := "anonymous-access.addons.k8s.io"
```
johngmyers, line 145: Perhaps call it "anonymous-issuer-discovery"?

upup/pkg/fi/cloudup/bootstrapchannelbuilder_test.go
```diff
@@ -50,6 +51,20 @@ func TestBootstrapChannelBuilder_BuildTasks(t *testing.T) {
 	runChannelBuilderTest(t, "awsiamauthenticator", []string{"authentication.aws-k8s-1.12"})
 }
 
+func TestBootstrapChannelBuilder_JWKS(t *testing.T) {
```
johngmyers, line 54:
```suggestion
func TestBootstrapChannelBuilder_PublicJWKS(t *testing.T) {
```

upup/pkg/fi/cloudup/bootstrapchannelbuilder_test.go
```diff
@@ -50,6 +51,20 @@ func TestBootstrapChannelBuilder_BuildTasks(t *testing.T) {
...
+	defer h.Close()
+
+	h.SetupMockAWS()
+
+	featureflag.ParseFlags("+PublicJWKS")
+	unsetFeaureFlag := func() {
+		featureflag.ParseFlags("-PublicJWKS")
+	}
+	defer unsetFeaureFlag()
+	runChannelBuilderTest(t, "jwks", []string{"dns-controller.addons.k8s.io-k8s-1.12", "kops-controller.addons.k8s.io-k8s-1.16", "anonymous-access.addons.k8s.io-k8s-1.16"})
```
johngmyers, line 65:
```suggestion
	runChannelBuilderTest(t, "public-jwks", []string{"dns-controller.addons.k8s.io-k8s-1.12", "kops-controller.addons.k8s.io-k8s-1.16", "anonymous-access.addons.k8s.io-k8s-1.16"})
```

(with corresponding directory rename)

upup/pkg/fi/fitasks/keypair.go
```diff
@@ -240,6 +248,17 @@ func (_ *Keypair) Render(c *fi.Context, a, e, changes *Keypair) error {
 		klog.Infof("updated Keypair %q to new format", name)
 	}
 
+	// Load the keypair for dependent resources
+	{
+		cert, _, _, err := c.Keystore.FindKeypair(name)
```
johngmyers, line 253: You _could_ optimize this out...

upup/pkg/fi/fitasks/keypair.go
```diff
@@ -44,6 +45,9 @@ type Keypair struct {
 	Type string `json:"type"`
 	// LegacyFormat is whether the keypair is stored in a legacy format.
 	LegacyFormat bool `json:"oldFormat"`
+
+	certificate                *fi.TaskDependentResource
```
johngmyers, line 49: What is this for? It doesn't appear to be read by anything.

upup/pkg/fi/fitasks/keypair.go
```diff
@@ -268,3 +287,33 @@ func parsePkixName(s string) (*pkix.Name, error) {
...
+	if e.certificate == nil {
+		e.certificate = &fi.TaskDependentResource{Task: e}
+	}
+	if e.certificateSHA1Fingerprint == nil {
+		e.certificateSHA1Fingerprint = &fi.TaskDependentResource{Task: e}
+	}
+}
+
+func (e *Keypair) setResources(cert *pki.Certificate) error {
+	e.ensureResources()
```
johngmyers, line 301: I prefer to do stuff like this more lazily, not calculating the sha1 unless some previous ModelBuilder has called `CertificateSHA1Fingerprint()`

upup/pkg/fi/resources.go
```diff
@@ -30,6 +30,11 @@ type Resource interface {
 	Open() (io.Reader, error)
 }
 
+// HasIsReady is implemented by Resources that are derived (and thus may not be ready at comparison time)
+type HasIsReady interface {
```
johngmyers, line 34: Seems like it might be useful, but I don't see where it's being referenced.

justinsb: Good point - an artifact of trying to resplit the commits.  Removed from this commit.

justinsb: And sent https://github.com/kubernetes/kops/pull/9837

upup/pkg/fi/cloudup/bootstrapchannelbuilder_test.go
justinsb: Done!

upup/pkg/fi/cloudup/bootstrapchannelbuilder_test.go
justinsb: Done!

upup/pkg/fi/fitasks/keypair.go
justinsb: It just felt more intuitive to store it, as we're also storing the fingerprint.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
justinsb: Done!

pkg/model/components/discovery.go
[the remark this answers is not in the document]
justinsb: Moved to nodeup, and mirrored the existing logic!

I think cert rotation is going to be difficult, but you're right - I don't need to make it harder :-)

justinsb: So I moved this to nodeup at @johngmyers 's suggestion below, and the public key is already done there, so I was able to simply drop this code.  It is indeed done the way you (quite rightly) suggested :-)

pkg/model/components/discovery.go
```diff
@@ -0,0 +1,83 @@
...
+	if kubeAPIServer.ServiceAccountJWKSURI == nil {
+		jwksURL := *kubeAPIServer.ServiceAccountIssuer
+		jwksURL = strings.TrimSuffix(jwksURL, "/") + "/openid/v1/jwks"
+
+		kubeAPIServer.ServiceAccountJWKSURI = &jwksURL
+	}
+
+	if kubeAPIServer.ServiceAccountSigningKeyFile == nil {
+		s := "/srv/kubernetes/server.key"
+		kubeAPIServer.ServiceAccountSigningKeyFile = &s
```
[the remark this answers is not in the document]
justinsb: Moved to nodeup!

upup/pkg/fi/fitasks/keypair.go
justinsb: This may have been an attempted workaround for the problem that was actually solved by IsReady in the other PR (now).  It doesn't seem to be needed yet, so I removed from this PR.

upup/pkg/fi/fitasks/keypair.go
justinsb: Fair - as it is so cheap and to avoid the ordering requirement (must have called CertificateSHA1Fingerprint before setResources), is it OK to leave as is?

johngmyers: It's cheap enough to leave as-is.

upup/pkg/fi/cloudup/bootstrapchannelbuilder_test.go
```diff
@@ -50,6 +51,20 @@ func TestBootstrapChannelBuilder_BuildTasks(t *testing.T) {
...
+	h := testutils.NewIntegrationTestHarness(t)
+	defer h.Close()
+
+	h.SetupMockAWS()
+
+	featureflag.ParseFlags("+PublicJWKS")
+	unsetFeaureFlag := func() {
+		featureflag.ParseFlags("-PublicJWKS")
+	}
+	defer unsetFeaureFlag()
```
johngmyers, line 64:
```suggestion
	unsetFeatureFlag := func() {
		featureflag.ParseFlags("-PublicJWKS")
	}
	defer unsetFeatureFlag()
```

upup/pkg/fi/fitasks/keypair.go
johngmyers: I tend to be more YAGNI on these sorts of things. Anything that's going to need the cert will probably just pull it (and possibly the private key) out of the keystore. But I suppose it doesn't block.

rifelpet:
Fixing e2e tests here: https://github.com/kubernetes/test-infra/pull/19056

PR #9814, opened 2020-08-25
Refactor: KopsModelContext embeds IAMModelContext

justinsb:
go syntax makes this an annoying change, unfortunately.


justinsb:

PR #9820, opened 2020-08-26
Remove unknown rules from managed security groups on openstack

### LINKED ISSUE #9716: Have kops ensure the correct rules for the master/node security groups

Right now, kops does not clean up old security group rules or rules added manually. There is no reason why kops shouldn't do this since we support adding additional security groups to be used for whatever other purposes.

The larger, but perhaps better way of solving this is to make security group rules a property of the securitygroup task. Then `Find()` can just get the current set of rules and ensure they match expected. There are some rule deletions that take place there already, for example if one enable bastions. But there is very specific logic around those few ports.

The more direct way of doing this is implement deletions in the securitygrouprule task. E.g add a Delete property to the task, create a task for every existing rule in the security group. If not an identical rule is created in `firewall.go`, set the Deletion property to true. Then one could also remove the `FindDeletions` logic for rules in the securitygroup task. 

Any thoughts?

olemarkus:
Fixes #9716 for openstack.

### DIFF
--- a/pkg/model/openstackmodel/firewall.go
+++ b/pkg/model/openstackmodel/firewall.go
@@ -17,7 +17,8 @@ limitations under the License.
 package openstackmodel
 
 import (
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
+	"fmt"
+
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstacktasks"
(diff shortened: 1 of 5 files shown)


pkg/model/openstackmodel/firewall.go
```diff
@@ -465,15 +474,90 @@ func (b *FirewallModelBuilder) addProtokubeRules(c *fi.ModelBuilderContext, sgMa
...
+	for sgName, sgt := range sgMap {
+
+		sgs, err := osCloud.ListSecurityGroups(sg.ListOpts{
+			Name: sgName,
+		})
+		if err != nil {
+			return err
+		}
+		if len(sgs) != 1 {
+			continue
```
zetaab, line 501: if we have two sgs with same name? in that case imo we should print log about that

zetaab (changes requested):
did you test this against running openstack kops cluster? if yes, then it looks quite good

pkg/model/openstackmodel/firewall.go
olemarkus: Erroring out now if that is the case. Thanks for catching.

olemarkus:
Yeah, tested that extra rules are detected and removed. And that at least on my test setups, there are no unwanted removals.

zetaab (approved):
thanks!

/approve
/lgtm

PR #9821, opened 2020-08-26
Reconcile ports and floating ips

olemarkus:
Will recreate lost resources or resources that didn't apply for whatever reason on first run.


zetaab (approved):
/approve
/lgtm

Issue #1647, opened 2017-01-27
add healthz health check on elb in private topo


chrislovecnm:
yo do it :)

kris-nova:
@chrislovecnm I am going to wordsmith this issue - I hope you don't mind!

kris-nova:
Do we want to get this into 1.5.0?

@justinsb?

justinsb:
Inclined to put this into 1.5.1 but I'll take a PR if someone sends it in time.

kris-nova:
### What we have

1. Healthcheck on TCP 443

<img width="1191" alt="screen shot 2017-02-11 at 11 11 21 pm" src="https://cloud.githubusercontent.com/assets/13757818/22860072/7c93bbf6-f0af-11e6-9404-aa5e02dde90e.png">
<img width="756" alt="screen shot 2017-02-11 at 11 11 09 pm" src="https://cloud.githubusercontent.com/assets/13757818/22860073/81244690-f0af-11e6-96d7-5f7fd3ad83a0.png">

### What we want

1. Healthcheck on TCP 443 (As it exists today)
2. Healthcheck(s) on Kubernetes (probably something that will validate against the `/healthz` endpoint(s))

### Where should you start looking

The existing ELB healthcheck code is [here](https://github.com/kubernetes/kops/blob/master/pkg/model/api_loadbalancer.go#L94)

A great starting point would be to add more health check tasks that are conducive to the kubernetes way of verifying cluster health!


### FYI

I think @seanknox is interested in taking a look here, I have volunteered to help him out in any way I can! Yay!

chrislovecnm:
I just looked at this, and how are we going to do auth?  I am seeing that /healthz is asking for auth.

chrislovecnm:
So I am wondering if we can use rbac to allow for anon access on the healthz endpoint.

chrislovecnm:
This is impacting rolling-update as well, as the health check comes back too quick.

GarethLewin:
Any progress on this?

ms4720:
interested too, this looks related https://github.com/kubernetes/kubernetes/issues/45746 
SSL checks would be better than tcp

ahharu:
any update?

PR #9825, opened 2020-08-27
WIP Improve API ELB Healthchecking

rifelpet:
This is untested and still WIP but I'd like to get others views on this general change.
As of k8s 1.17 we now have a healthcheck container that proxies HTTP requests for the healthcheck endpoints to authenticated HTTPS requests to kube-apiserver. This PR extends that functionality to also be used by the API ELB for a more accurate healthcheck.

[These docs](https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health) mention that `/healthz` is deprecated in 1.16 so the livenessProbe now uses `/livez` and the ELB healthcheck uses `/readyz`. While in the apiserver docs I noticed a [--shutdown-delay-duration](https://github.com/kubernetes/kubernetes/pull/74416/files) flag that was also introduced in 1.16 which will fail the `/readyz` endpoint for a period of time between the SIGINT and SIGKILL. This seems beneficial for rolling updates to reduce lost requests while a master is being replaced. I chose 30s because it matches the default `terminationGracePeriodSeconds` value and aligns with the ELB's health check settings of `unhealthyThreshold=2` and `Interval=10s`.

Does the shutdown delay make sense?  Maybe we add an equivalent sleep in the etcd preStop lifecycle so that etcd is available for as long as apiserver stays up. I'm also wondering exactly how etcd and apiserver are shutdown during the drain and terminate steps and if they will be able to have this additional time to shutdown gracefully.

Fixes https://github.com/kubernetes/kops/issues/1647

### DIFF
--- a/nodeup/pkg/model/kube_apiserver.go
+++ b/nodeup/pkg/model/kube_apiserver.go
@@ -397,11 +397,15 @@ func (b *KubeAPIServerBuilder) buildPod() (*v1.Pod, error) {
 		},
 	}
 
+	if b.IsKubernetesGTE("1.16") {
+		flags = append(flags, "--shutdown-delay-duration=30s")
+		pod.Spec.TerminationGracePeriodSeconds = fi.Int64(35)
+	}
+
 	useHealthcheckProxy := b.findHealthcheckManifest() != nil
 
 	probeAction := &v1.HTTPGetAction{
 		Host: "127.0.0.1",
-		Path: "/healthz",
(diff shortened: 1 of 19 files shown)


johngmyers:
I think you'd want the shutdown delay to be shorter than the termination grace period, to allow time for the server to actually shut down.

etcd and apiserver are voluntarily evicted during rolling update, so will be able to take advantage of the termination grace period.

PR #8104, opened 2019-12-13
Update Calico to v3.10.2

hakman:
Calico v3.10 is here with some nice new stuff like bird/bird6 liveness checks.
Canal is already at 3.10 but Calico should be upgraded too.

This PR addresses also some other issues:
* make Calico and Canal templates easier to compare, by moving around some fields
* make Canal easier to compare to official manifest
* add role.kubernetes.io/networking labels to Canal

Commits should be reviewed one by one, as it will be much easier to compare each step.

PR #7917, opened 2019-11-12
Canal v3.10 manifest for k8s v1.15+

KashifSaadat:
From Canal v3.8 onwards, a new init container is introduced and included in the manifests by default, which installs a Flex Volume Driver. This requires having a writeable path to install the driver in, which by default is `/usr/libexec/kubernetes/kubelet-plugins/volume/exec/`.

The default flexvol directory above is not writeable for OSs such as CoreOS & ContainerOS, so the volume plugin directory must be specified in the Cluster Spec like below:
```yml
spec:
  kubelet:
    volumePluginDirectory: /var/lib/kubelet/volumeplugins/
```

Typha support is now available for Canal deployments. This can be configured as follows:
```yml
spec:
  canal:
    typhaReplicas: 3
```

Other available options are: `typhaPrometheusMetricsEnabled`, `typhaPrometheusMetricsPort`

PR #9826, opened 2020-08-27
remove nodeAffinity from typha

ozdanborne:
This PR removes the unnecessary nodeAffinity from the Typha deployment in calico and canal manifests.

**Some background / history on how we got to this point:**
1. `nodeSelector: {kubernetes.io/role: master}` is added to Canal in https://github.com/kubernetes/kops/pull/7917/
1. This `nodeSelector` change is "copied" into the Calico manifest in #8104 .
1. @seh  points out that the node selector "makes it much harder to run more than three Typha replicas. It would be better to have node affinity that expresses a preference, but not a requirement for Typha pods to run on the masters." https://github.com/kubernetes/kops/pull/8104#discussion_r458153054, and raises https://github.com/kubernetes/kops/issues/9608 to track.
1. Nodeselector is converted to an Affinity as part of https://github.com/kubernetes/kops/pull/9609

**Why it was added in the first place**

I believe that the nodeSelector should never have been added in the first place. @tmjd has been asking questions on those threads and as best as we can tell, the node affinity / selectors were added to mitigate an issue where Typha was not being evicted from a node when the node group was being scaled down to zero. As per @seh:

>I've seen cases where the cluster autoscaler can't drop a worker node because a Typha pod is running on it. That Typha pod could run on one of the master nodes that are not subject to the autoscaler's control. The safe-to-evict annotation didn't work as expected. We wound up paying to run that worker node unnecessarily. [Note that] I wasn't using kops when I saw that problem with the cluster autoscaler failing to evict Typha's pods.
>Source:  https://github.com/kubernetes/kops/issues/9608#issuecomment-676520700

As per @KashifSaadat 
>I vaguely recall this was more specific to how we performed Kops node rolling upgrades and manifest updates. I believe as part of a Kops upgrade from v1.14 to v1.15 (including both Kubernetes and the Canal manifest update), after rolling the first master nodes the new Canal manifest was applied and Typha enabled (if specified in the cni spec), but Typha couldn't be scheduled on the old v1.14 nodes. At this point the canal pods have also begun rolling (as the new manifest was applied) and would fail to start because the Typha service is unavailable.
>I'd have to re-run the upgrade path to confirm the exact issue I encountered, but fairly certain the nodeSelector should no longer be required any more so we can look to remove it in a future update.
>Source: https://github.com/kubernetes/kops/pull/7917#issuecomment-677776726

I don't disagree that something may have been funky with evictions. But I do disagree with using affinity to avoid it. We should fix that eviction issue instead! Even after pouring through these issues, It's not clear to me what the eviction problem is, though it sounds like we might not be expecting it anymore?

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -12233,14 +12233,6 @@ spec:
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict: 'true'
     spec:
-      affinity:
-        nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 100
-            preference:
-              matchExpressions:
-              - key: node-role.kubernetes.io/master
-                operator: Exists
       nodeSelector:
         kubernetes.io/os: linux
       hostNetwork: true
@@ -16226,14 +16218,6 @@ spec:
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict: 'true'
     spec:
-      affinity:
-        nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 100
-            preference:
-              matchExpressions:
-              - key: node-role.kubernetes.io/master
-                operator: Exists
       nodeSelector:
         kubernetes.io/os: linux
       hostNetwork: true
--- a/upup/models/cloudup/resources/addons/networking.projectcalico.org.canal/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.projectcalico.org.canal/k8s-1.16.yaml.template
@@ -531,14 +531,6 @@ spec:
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict: 'true'
     spec:
-      affinity:
-        nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 100
-            preference:
-              matchExpressions:
-              - key: node-role.kubernetes.io/master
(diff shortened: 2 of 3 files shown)


seh:
I believe the problem lies with the autoscaler's treatment of pod priority, as [described in its FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-cluster-autoscaler-work-with-pod-priority-and-preemption). The priority cutoff was zero as recently as version 1.11, but they've since dropped it to -10. I suspect that the Typha pods hadn't been preventing scale downs because their priority was zero. Now that their priority is significantly higher, well above the default -10 cutoff, they prevent the autoscaler from scaling down.

The "cluster-autoscaler.kubernetes.io/safe-to-evict" is supposed to allow overriding that policy, so long as the Typha pods don't violate any of the other requirements expressed [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node). In some circumstances we can't move a Typha pod due its use of host ports, precluding multiple running on the same node. We should be able to move them back onto master machines, though.

As for why I had seen these pods keeping worker nodes alive when it appeared the pods could have run elsewhere instead, I can't go back to recover that environment today to investigate further.

hakman:
@ozdanborne Thanks for doing this. I agree with the approach. Let's see what happens when this is removed and fix it. Btw, nice job with Calico 3.16. Looking forward to give it a try!
/lgtm
/approve

PR #9827, opened 2020-08-27
GCE - Set Bastion InstanceGroup zone

rifelpet:
GCE uses Spec.Zones rather than Spec.Subnets because subnets are regional rather than zonal.
This sets the Zones field for bastion IGs in GCE, avoiding an index out of range panic during cluster creation.

Master and node IG creation uses the same logic:

https://github.com/kubernetes/kops/blob/55f33c68b36a91d213687bf79ec003fd71195d35/upup/pkg/fi/cloudup/new_cluster.go#L627-L629

Also cleaning up some old v1alpha1 integration test files that are no longer referenced.

Fixes https://github.com/kubernetes/kops/issues/8626#issuecomment-680915891


MoShitrit:
@rifelpet that worked!
The cluster is now spinning up, fingers crossed that everything's going to work.
I LGTMd the PR since I confirmed it solved my issue.
I'll keep you posted.
Thanks a lot!! 😄

PR #9829, opened 2020-08-28
Update Calico to v3.16.0 for k8s 1.16+

hakman:
https://docs.projectcalico.org/archive/v3.16/release-notes/

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
@@ -3709,7 +3709,7 @@ spec:
         # It can be deleted if this is a fresh installation, or if you have already
         # upgraded to use calico-ipam.
         - name: upgrade-ipam
-          image: calico/cni:v3.16.0
+          image: calico/cni:v3.16.0-2-gf6793aa
           command: ["/opt/cni/bin/calico-ipam", "-upgrade"]
           envFrom:
           - configMapRef:
@@ -3736,7 +3736,7 @@ spec:
         # This container installs the CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: calico/cni:v3.16.0
+          image: calico/cni:v3.16.0-2-gf6793aa
           command: ["/opt/cni/bin/install"]
           envFrom:
           - configMapRef:
--- a/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
@@ -1,4 +1,4 @@
-# Pulled and modified from: https://docs.projectcalico.org/v3.15/manifests/calico-typha.yaml
+# Pulled and modified from: https://docs.projectcalico.org/v3.16/manifests/calico-typha.yaml
 
 ---
 # Source: calico/templates/calico-config.yaml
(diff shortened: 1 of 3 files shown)


upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -407,6 +466,15 @@ spec:
...
             properties:
+              awsSrcDstCheck:
+                description: 'Set source-destination-check on AWS EC2 instances. Accepted
+                  value must be one of "DoNothing", "Enabled" or "Disabled". [Default:
+                  DoNothing]'
+                enum:
+                - DoNothing
+                - Enable
+                - Disable
+                type: string
```
gjtempleton, line 477: Do we want to move the logic from https://github.com/kubernetes/kops/blob/fc2e70e1e8e8901b6713ac930dded00dff077516/upup/models/bindata.go#L8618 for installing the `k8s-ec2-srcdst` pod to here as this is now the supported path?

I think allowing users to enable this will require them to add permissions to worker nodes as we currently only give the required `ec2:ModifyInstanceAttribute` permissions to masters, at a minimum I think we need to add something to the docs, and make clear that enabling this alongside the existing logic for installing `k8s-ec2-srcdst` is an untested path.

hakman: Yes, we do, and also enable WireGuard. But best to do it in some follow-up PRs.

hakman:
@lwr20 I am a bit lost here and really puzzled about why I cannot make 3.16.0 work. Any idea what I am doing wrong? 😄

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
seh: Does _calico-node_ talk to the AWS API for this, or is it the _calico-kube-controllers_ pod? If it's the latter, we could optionally confine it to run only on master nodes, and grant the IAM permissions only to the instance profile used by the masters.

_[Time passes...]_

[It's part of Felix,](https://github.com/projectcalico/felix/blob/master/aws/ec2.go) and therefore [part of _calico-node_](https://docs.projectcalico.org/reference/node/configuration).

fasaxc:
@hakman is there a way to get the actual error that caused validation to fail?

hakman:
@fasaxc I am not really sure where the issue comes from from. I also tried using https://docs.projectcalico.org/manifests/calico.yaml and still same issue.

You can find the logs of the master node from the test here:
https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kops/9829/pull-kops-e2e-cni-calico/1299270983718277120/artifacts/ip-172-20-39-18.ap-northeast-2.compute.internal/

seh:
Looking at the controller manager logs, I see lots of errors that match what I saw when I tried upgrading _etcd_ directly from version 3.2.24 to version 3.4.3.

hakman:
@seh I got some help from @fasaxc and tracked the issue to https://github.com/projectcalico/cni-plugin/pull/942.
Looks like the pre-release `calico/cni:v3.16.0-2-gf6793aa` image addresses the issue and the tests can run now.

Thanks again for all your help @fasaxc :).

hakman:
Works pretty well to use during the net few weeks until a new release is available.
/hold cancel

PR #9831, opened 2020-08-28
cloudbuild: allow CI env var to be specified

justinsb:
This should allow us to build our tagged builds as _non_ CI builds.


justinsb:
There's going to be more to this one, because it turns out that prow reserves CI for "reasons", but this actually stands alone.

/hold cancel

PR #9836, opened 2020-08-29
Always use OpenStack Swift reauthentication

justinsb:
If we were using credentials from env vars, we would not do
reauthentication with Swift.


justinsb:
Issue #9730

kciredor:
Thanks, did not realise there's vendoring involved. Hope this makes it into kops 1.18.x when merged.

PR #9837, opened 2020-08-29
TaskDependentResource: support preview when the task isn't ready

justinsb:
This is needed because otherwise if we try to diff a computed field,
we can't read the value.


justinsb:
/test pull-kops-verify-gomod

looks like a network flake

PR #9840, opened 2020-08-30
Add nil support for setting fields via reflection

rifelpet:
Extends #8896

This has an unfortunate side-effect of being unable to set `*string` fields to `""` since the pointer would instead be set to `nil`. If we think setting an empty string could be desirable we could instead have a special string, perhaps `"nil"` or `"null"`.

/hold for comment

### DIFF
--- a/pkg/commands/set_cluster_test.go
+++ b/pkg/commands/set_cluster_test.go
@@ -296,6 +296,55 @@ func TestSetClusterFields(t *testing.T) {
 				},
 			},
 		},
+		{
+			Fields: []string{
+				"spec.docker.userNamespaceRemap=",
+			},
+			Input: kops.Cluster{
+				Spec: kops.ClusterSpec{
+					Docker: &kops.DockerConfig{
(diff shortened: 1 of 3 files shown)


johngmyers:
What is the use case for setting a field to `nil`? I'm definitely wary of the inability to set to `""`.

justinsb:
Yes, I'd like to hear some example use case(s).

We had a mini-debate about `[]string` - should it append or should it overwrite?  But we have all this feature-flagged precisely so we can figure out the correct syntax here, so it's good that we're discussing it now :-)

Two main options that come to mind:

* Use some DSL, e.g. `set spec.example+=foo`, though not clear what nil would be (`set -spec.example`?)
* Create multiple commands: set, clear, append are the obvious ones.

justinsb:
In our hacktoberfest chat today we found this: https://tools.ietf.org/html/rfc6902#section-4 and we thought the first 3 (add, remove, replace) might be a good set of well-defined operations to add/support.  Possibly via sub-flags or some syntax.

PR #9841, opened 2020-08-31
Remove more machinery for file-based cloudup models

johngmyers:
/kind cleanup


olemarkus:
Holding in case there is need for more reviewers, but this does look dead/unused to me.
/hold
/lgtm

Issue #8723, opened 2020-03-10
Support Bottlerocket In AWS


rifelpet:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

AWS just announced a new container-focused Linux distribution. We should consider adding support for it in Kops

https://aws.amazon.com/bottlerocket/

**2. Feel free to provide a design supporting your feature request.**

jmcdonagh:
This would be great

rifelpet:
/remove-lifecycle stale

I'm picking this back up and getting the PR ready for testing again. The current blocker is that because bottlerocket's node bootstrapping relies on the CSR flow, we need a controller to approve the CSRs for kubelet server certificates which [can't be auto-approved](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/#certificate-rotation) in core components.

oded-dd:
This PR is really important for us. Can it be merged?

olemarkus:
This is far from a working state. I would be interested in why this is important to you.

The main challenge is how we can support kops features on a distro we cannot really configure/install things on.

olemarkus:
Why would it be problematic? Can you elaborate?

oded-dd:
bottlerocket supports FIPS 140-2. This is a security standard which becomes more and more common and is a requirement for SaaS providers (like us) by many customers (mostly Enterprise). bottlerocket also is about to become the official image used by AWS instead of Amazon linux2.

olemarkus:
I don't feel you are answering the question. The default kOps image is Ubuntu LTS, which is also FIPS 140-2 certified.
Why would also the kOps project care about what image EKS is using? EKS has always had the strategy of using EKS specific images, while kOps project has moved away from that. You may also conflate "default" with "supported". I doubt AL2 or ubuntu images for EKS is going away either. There are too many companies needing more customisation than what bottlerocket allows for.

drpdishant:
[...]

PR #9855, opened 2020-09-01
WIP Initial support for Bottlerocket OS

rifelpet:
https://github.com/bottlerocket-os/bottlerocket

ref: https://github.com/kubernetes/kops/issues/8723

The userdata needs to be entirely different (in fact the userdata is just settings defined in TOML) so Kops needs to be aware of the AMI's OS at the time it creates userdata, so I added a new `ImageFamily` API field (following the name from [eksctl](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sample-eksctl.yaml#L14) but I'm happy to change it). I figured the new field would also be useful for Windows support, since that will need different userdata too.

This also requires that the Userdata model now depends on the CA keypair. I wasn't able to get that setup easily in unit testing, but the integration test covers it well.

for this PR:
- [x] make tests results consistent - I'm assuming i'll use the same CA key strategy used in the new PublicJWKS integration test
- [x] warn or fail API validation if certain IG fields are set that are incompatible with bottlerocket. gossip clusters wont be supported yet either since that would require access to the host's /etc/hosts. aws-iam-authenticator is also required to be enabled and setup to bind the nodes role with rbac permissions to join the cluster
- [x] images.md

for followup PRs:
- [ ] add SSH support
- [ ] add support for additionalUserdata so that users can provide [additional bottlerocket settings](https://github.com/bottlerocket-os/bottlerocket#description-of-settings). Or come up with a more integrated way of specifying these.

Limitations:
* kubelet serving certificate will not work for metrics server


rifelpet:
```
 curl -fsS --retry 3 -o /workspace/kops https://storage.googleapis.com/kops-ci/pulls/pull-kops-e2e-kubernetes-aws/pull-337b921200/linux/amd64/kops
curl: (56) OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104 
```
🤔

rifelpet:
apparently the bottlerocket AMIs are not published in every region, which doesn't play nicely with prow and boskos' randomly chosen regions.

mikesplain:
Awesome, glad to hear you started on this!

rifelpet:
Quick update before I head out for vacation. I was able to get a bottlerocket node joined to the cluster.

* They currently only have AMIs for up to k8s 1.17. I expect 1.18 and 1.19 AMIs to be introduced alongside EKS' support for them
* kubelet uses aws-iam-authenticator for authentication (see [here](https://github.com/bottlerocket-os/bottlerocket/blob/aaa2fe2a416dddbeeca02a25cd5994c9684dd840/packages/kubernetes-1.17/kubelet-kubeconfig)) so this requires that the aws-iam-authenticator server component is enabled on the cluster.
* re: RBAC, the groups included in the KubernetesNode mapping in [this example](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/2c936a2108a739244311ac979eafaac2de142164/README.md#full-configuration-format) were not sufficient, there were additional permissions needed around services, csidrivers, and runtimeclasses that were needed. I didn't have a chance to scope out the exact clusterrole we'll need to add. Users will also need to ensure the mapping is present in the aws-iam-authenticator config. New versions of the authenticator support CRDs, so Kops could theoretically add the mapping.
* The kubelet API endpoints werent working, so logs and exec werent functional. I didn't have a chance to investigate this further.
  kubectl:
  ```
  Error from server: Get "https://172.20.52.72:10250/containerLogs/kube-system/kube-dns-696cb84c7-b4nr4/kubedns?follow=true": remote error: tls: internal error
  ```
  kubelet:
  ```
  I0903 01:54:45.012939    8643 log.go:181] http: TLS handshake error from 172.20.55.160:39782: no serving certificate available for the kubelet
  ```
* The first `create cluster` or `update cluster` without `--yes` fails because of the way I fetch the CA certificate to add to the userdata. This loops forever until kops gives up:
  ```
  W0901 23:14:34.469620   48299 executor.go:131] error running task "BootstrapScript/nodes-us-west-2a" (9m57s remaining to succeed): error opening resource: resource opened before it is ready (task=*fitasks.Keypair {"Name":"ca","alternateNames":null,"Lifecycle":"Sync","Signer":null,"subject":"cn=kubernetes","type":"ca","oldFormat":false})
  ```
  With `--yes` or once the CA exists, it runs just fine.
  I have to figure out how to reference it in a deferred way so that the CA will be created and then the userdata can be created. I thought adding the CA task to dependencies would be sufficient but apparently not.
* TOML serializing doesn't handle node labels containing periods correctly so we'll have to handle that ourselves. I believe eksctl solves this already. It was rendering them as this:
  ```toml
      [settings.kubernetes.node-labels]

      [settings.kubernetes.node-labels.kops]

        [settings.kubernetes.node-labels.kops.k8s]
          io/instancegroup = "nodes-us-west-2a"
  ```
  instead of this:
  ```toml
      [settings.kubernetes.node-labels]
          "kops.k8s.io/instancegroup" = "nodes-us-west-2a"
  ```
  so I disabled setting node labels for now.
If anyone has suggestions or wants to pick up any of these tasks feel free, otherwise I'll resume when I'm back.

tjkirch:
> TOML serializing doesn't handle node labels containing periods correctly so we'll have to handle that ourselves.

Just a quick note in case this helps with implementation options - in Bottlerocket settings TOML, to handle label names with characters like `.`, you can either quote the entire label name as a key, like you showed above:

>     [settings.kubernetes.node-labels]
>        "kops.k8s.io/instancegroup" = "nodes-us-west-2a"

That's typically what we do, and allows for easy specification of multiple labels in the `node-labels` table.

You can also quote the label name inside the TOML key name, like `settings.kubernetes.node-labels."kops.k8s.io/instancegroup" = "nodes-us-west-2a"`

(The quoting is there is so the "kops" and "k8s" portions aren't interpreted as deeper Bottlerocket settings.)

rifelpet:
@tjkirch thanks! I found that our go-toml library actually has a `QuoteMapKeys` encoder option which solves that issue for us. While in the rust code I noticed that node taints are supported as well so I plumbed that in too.

I think we're close to adding alpha support for this. I'm considering gating the new API field with a feature flag in case we want to change the design later.

rifelpet:
Just a quick update, almost everything is ready to go except:

* kubelet api (logs/exec) - likely a cert issue. I haven't had a chance to look into this, hopefully soon
* Having kops create the IAMIdentityMappings automatically so that kubelet will be able to authenticate and join the cluster. Without this the cluster will fail validation until the mappings are added externally.

The problem I'm running into with this is that the mappings need to use the IG's role ARN whereas the instance groups deal with either a role name (if using the built-in roles) or an instance profile ARN (if using an external role). Additionally, the role ARN [must have any paths removed](https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/268).

I'm hesitant to just do string substitution on the instance profile ARN because instance profiles can be associated with roles of any name, so they don't necessarily match.

I can try to get the role ARN from the role name but that may require using an `awsup.AWSCloud` which gets complicated if I were to just include these `IAMIdentityMapping` manifests in the bootstrapchannelbuilder via template functions. Perhaps this is an opportunity for the addon operator instead.

rifelpet:
Oh and I updated the API to use a struct rather than a string. I figured that would be more flexible if we end up needing bottlerocket-specific settings (or more likely, windows-specific settings when we add support down the road)

old:
```yaml
spec:
  imageFamily: bottlerocket
```

new:
```yaml
spec:
  imageFamily:
    bottlerocket: {}
```

Its also guarded behind a feature flag as mentioned in the docs update.

johngmyers:
I presume there will be other restrictions due to the fact that this can't request certs through kops-controller. For example, running Kuberouter or Cilium in EtcdManaged mode.

rifelpet:
Yes, certain networking setups won't work. I added those to the documentation and API validation.

I was able to successfully create a cluster following the documentation's instructions and using `cilium: hubble: {}`. The kubelet server certificate issue I was running into was because each node was creating a CSR that I needed to manually approve with `kubectl certificate approve`. Once the cert was approved, `kubectl logs` and `kubectl exec` to pods on the bottlerocket nodes worked fine.

I may need assistance in figuring out how this should be handled properly. Here are the [kubelet config](https://github.com/bottlerocket-os/bottlerocket/blob/1ecb62d48a75a4874efae2b4965f184c3102dfd4/packages/kubernetes-1.17/kubelet-config) and [kubelet flags](https://github.com/bottlerocket-os/bottlerocket/blob/1ecb62d48a75a4874efae2b4965f184c3102dfd4/packages/kubernetes-1.17/kubelet.service) if that helps. Reading through the [k8s docs](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/#approval) and these [cluster role bindings](https://github.com/kubernetes/kops/blob/ef30628f95107855d7c4ebf8cddc3be205dd921e/upup/models/cloudup/resources/addons/node-authorizer.addons.k8s.io/k8s-1.12.yaml.template#L31-L63), I think node-authorizor might help here? or KCM directly?

Additionally, the only other bug I've found is that `kops update cluster` without `--yes` on a brand new cluster fails because the bootstrapscript's contents is now depending on the CA. I have to figure out how to defer that dependency:

```
I0917 17:25:21.784422   15434 executor.go:103] Tasks: 0 done / 82 total; 47 can run
W0917 17:25:22.336037   15434 vfs_castore.go:604] CA private key was not found
I0917 17:25:22.347677   15434 executor.go:103] Tasks: 47 done / 82 total; 17 can run
W0917 17:25:22.635369   15434 executor.go:131] error running task "BootstrapScript/nodes-us-west-2a" (9m59s remaining to succeed): error opening resource: resource opened before it is ready (task=*fitasks.Keypair {"Name":"ca","alternateNames":null,"Lifecycle":"Sync","Signer":null,"subject":"cn=kubernetes","type":"ca","oldFormat":false})
I0917 17:25:22.635412   15434 executor.go:103] Tasks: 63 done / 82 total; 16 can run
W0917 17:25:22.883242   15434 executor.go:131] error running task "BootstrapScript/nodes-us-west-2a" (9m59s remaining to succeed): error opening resource: resource opened before it is ready (task=*fitasks.Keypair {"Name":"ca","alternateNames":null,"Lifecycle":"Sync","Signer":null,"subject":"cn=kubernetes","type":"ca","oldFormat":false})
```

Any assistance on either of these problems would be appreciated.

johngmyers:
`BootstrapScript`'s `GetDependencies()` has to return the "ca" `Keypair` task as a dependency when it is configured to use it.

This is a highly divergent case, as the Bottlerocket instances don't even run nodeup. So nothing that needs code in nodeup is likely to work on them, they're going to have a different mechanism for joining the cluster than everything else. It seems sort of equivalent to having kops support using EKS as a control plane.

Node-authorizer is unlikely to help as kops won't even run it on AWS for k8s 1.19+.

pkg/model/bootstrapscript.go
```diff
@@ -374,6 +395,49 @@ func (b *BootstrapScript) Run(c *fi.Context) error {
...
+	for _, task := range c.AllTasks() {
+		if t, ok := task.(*fitasks.Keypair); ok && fi.StringValue(t.Name) == fi.CertificateIDCA {
+			caTask = task.(*fitasks.Keypair)
+		}
+	}
+	if caTask == nil {
+		return fmt.Errorf("CA task not found")
+	}
+	caResource := caTask.Certificate()
+	ca, err := fi.ResourceAsBytes(caResource)
```
rifelpet, line 409: I do have the CA task being added to `GetDependencies` (see L233 above) but the problem is that on this line we're needing to read the contents of the CA even during the dryrun process in order to print the userdata that would be created.

I'm trying to figure out why the OIDC provider and its CA dependency do not share this same problem. It might be because the OIDC provider doesn't fetch the CA thumbprints until `RenderAWS()` whereas this is fetched in `Run()`. I'm wondering if we need to break the bootstrapscript out into its own Render function and use `DefaultDeltaRunMethod` like the cloud-specific tasks?

@justinsb any thoughts here?

rifelpet:
This is _very_ close to working. The remaining gaps are around the node bootstrapping process. It uses the built-in CSR method including the kube-controller-manager auto-approval documented [here](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/#approval). I know we have very similar functionality being added to kops-controller so I want to assess whether it would make sense to be supporting both methods or if we can have kops-controller handle the bottlerocket bootstrapping as well.

In  the [most recent e2e job](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/9855/pull-kops-e2e-kubernetes-aws/1339382731091480576) the nodes join the cluster but for some reason the kube-dns pod never becomes  ready even though it is scheduled to one of the nodes. Unfortunately `kubectl logs` doesn't work because bottlerocket's kubelet doesn't have a valid server certificate. The k8s documentation linked above mentions this should be explicitly handled by an external process (such as kops-controller) so we may need to build the support into kops-controller regardless.

Any feedback here would be appreciated. Bottlerocket support is in the top 10 kops issues so it'd be nice to get this added.

olemarkus:
#10321 has the code for signing cert-manager CSRs. I suspect you can copy most of that to use with the K8s CSRs as well.

I wonder how sustainable it really is to support a distro without nodeup, but I don't have any big problems with what has been done here.

rifelpet:
During the provider-aws meeting yesterday I brought up the need for a CSR approving controller, and the AWS folks are going to discuss potentially open-sourcing their controller used in EKS and possibly integrating it into aws-cloud-controller-manager. see the meeting notes [here](https://docs.google.com/document/d/1-i0xQidlXnFEP9fXHWkBxqySkXwJnrGJP9OGyP2_P14/edit#heading=h.d2u2axvps5x1).

rifelpet:
[bottlerocket 1.0.6](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.0.6) added support for [TLS auth](https://github.com/bottlerocket-os/bottlerocket/pull/1338) instead of AWS auth, and [static pods](https://github.com/bottlerocket-os/bottlerocket/pull/1317). This may simplify how we go about supporting bottlerocket.

oded-dd:
Great Work! Any update on that?

PR #9857, opened 2020-09-02
Detect AWS region for S3 inside containers

hakman:
AWS detection assumes that `/sys/devices/virtual/dmi/id/product_uuid` is readable. This is not the case inside containers like `kops-controller`. If there is an error reading the file, AWS api should be queried anyway as a fallback.

Fixes: #9856

### DIFF
--- a/util/pkg/vfs/s3context.go
+++ b/util/pkg/vfs/s3context.go
@@ -155,13 +155,16 @@ func (s *S3Context) getDetailsForBucket(bucket string) (*S3BucketDetails, error)
 	}
 
 	awsRegion := os.Getenv("AWS_REGION")
-	if awsRegion == "" && isRunningOnEC2() {
-		region, err := getRegionFromMetadata()
-		if err != nil {
-			klog.V(2).Infof("unable to get region from metadata:%v", err)
-		} else {
-			awsRegion = region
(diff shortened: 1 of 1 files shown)


johngmyers:
Another possibility would be to set `AWS_REGION` in the manifest.

hakman:
> Another possibility would be to set `AWS_REGION` in the manifest.

This is a more generic fix that will work in any container and I don't think it has any downside.

johngmyers:
Some security systems block access to AWS metadata from pods in order to deny them use of the node's IAM roles.

hakman:
> Some security systems block access to AWS metadata from pods in order to deny them use of the node's IAM roles.

Ok, I would still leave this as is and create a new PR that adds AWS_REGION to env vars. Sounds reasonable?

hakman:
@johngmyers any thoughts on merging this?

ryan-dyer-sp:
Can we get this back ported to 1.18?  without this change the ability to use kops to deploy to aws govcloud regions is broken; at least without needing to manually update the kops-controller DS to include the AWS_REGION env.

hakman:
@ryan-dyer-sp already cherry-picked into 1.18, will take another 2-4 for a new release. I hope this is ok for you.

PR #9860, opened 2020-09-02
Add instance info to detached nodes

olemarkus:


upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -787,18 +787,9 @@ func awsBuildCloudInstanceGroup(c AWSCloud, cluster *kops.Cluster, ig *kops.Inst
...
-			key := aws.StringValue(tag.Key)
-			if !strings.HasPrefix(key, TagNameRolePrefix) {
-				continue
-			}
-			role := strings.TrimPrefix(key, TagNameRolePrefix)
-			cm.Roles = append(cm.Roles, role)
-			cm.PrivateIP = aws.StringValue(instance.PrivateIpAddress)
-		}
+
+		addCloudInstanceData(cm, instance)
```
hakman, line 792:
```suggestion
		addCloudInstanceData(cm, instances[id])
```

hakman (changes requested):

upup/pkg/fi/cloudup/awsup/aws_cloud.go
olemarkus: like this?

hakman: yes

PR #9864, opened 2020-09-03
Update installation

inductor:
- OSX was renamed to macOS a long time ago
- Homebrew has supported Linux for a long time
- brew install is much easier and better for version management on Linux too


hakman:
Thanks for updating this. It was long due.

PR #9865, opened 2020-09-03
WIP: OpenStack: possibility to detach instances from the load

zetaab:
original PR #8874 

This is under construction. The idea here is to make detachInstance working also with OpenStack. It will speed up rolling updates.


upup/pkg/fi/cloudup/openstacktasks/servergroup.go
```diff
@@ -133,14 +133,33 @@ func (_ *ServerGroup) RenderOpenstack(t *openstack.OpenstackAPITarget, a, e, cha
...
-			instances, err := t.Cloud.ListInstances(opts)
+			// TODO: we could maybe use some filter here
+			allInstances, err := t.Cloud.ListInstances(servers.ListOpts{})
 			if err != nil {
 				return fmt.Errorf("error fetching instance list: %v", err)
 			}
 
+			instances := []servers.Server{}
+			for _, server := range allInstances {
+				val, ok := server.Metadata["k8s"]
```
mitch000001, line 144: Gophercloud also supports tags in ListOpts: https://pkg.go.dev/github.com/gophercloud/gophercloud@v0.13.0/openstack/compute/v2/servers#ListOpts Can we use that at line 137 to already limit the instances?

mitch000001:
@zetaab I was wondering if we could extract the bits in here related to instance naming, as this is already valuable on its own. Also, it shrinks the actual PR.

zetaab:
closed, see #10557

Issue #9852, opened 2020-09-01
kops fatal error: found duplicate tasks with name "IAMRole"


kesor:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

`v1.17.0` , `v1.17.1` , `v1.18.0`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

`v1.17.11`

**3. What cloud provider are you using?**

Amazon AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

```shell
kops create cluster
kops create ig second
# add `iam:\nprofile: arn:aws:iam::123456789012:instance-profile/nodes.mycluster.domain.com`
# the IAM profile used for second IG is the same as `nodes`
kops update cluster
```

**5. What happened after the commands executed?**

A fatal error appears:

```
F0901 14:58:45.758233   63950 task.go:59] found duplicate tasks with name "IAMRole/nodes.mycluster.domain.com": *awstasks.IAMRole {"ID":null,"Lifecycle":"Sync","Name":"nodes.mycluster.domain.com","RolePolicyDocument":{"Name":"","Resource":{}},"ExportWithID":"nodes"} and *awstasks.IAMRole {"ID":null,"Lifecycle":"Sync","Name":"nodes.mycluster.domain.com","RolePolicyDocument":{"Name":"","Resource":{}},"ExportWithID":"nodes"}
```

**6. What did you expect to happen?**

To be able having a mix of explicit `iam: profile:` instance groups, and some that are implicit. At the moment `kops` allows to share IAM InstanceProfile across multiple instance groups, but either ALL have to have `iam: profile:` be explicit, or ALL have to have `iam: profile:` be implicit.

**7. Please provide your cluster manifest. Execute `kops get --name my.example.com -o yaml` to display your cluster manifest. You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2020-09-01T11:33:17Z"
  generation: 1
  name: mycluster.domain.com
spec:
  api:
    loadBalancer:
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudLabels:
    k8sClusterName: mycluster
  cloudProvider: aws
  configBase: s3://mybucket.domain.com/mycluster.domain.com
  dnsZone: domain.com
  etcdClusters:
[...]

PR #9867, opened 2020-09-03
Only apply external policies when these are defined

kesor:
We experience fatal errors with this new feature (that we are not using) because some of the instance groups in our clusters include `spec.iam.profile` while others use the managed `nodes` role/profile. The `ExternalPolicies` feature is broken, and this makes it go away when not specifically asked for in the cluster configuration.

At the moment, without this patch, the `kops update cluster` operation results in -

```
F0903 17:45:44.372520   48506 task.go:59] found duplicate tasks with name "IAMRolePolicy/node-policyoverride": *awstasks.IAMRolePolicy {"ID":null,"Lifecycle":"Sync","Name":"node-policyoverride","Role":{"ID":null,"Lifecycle":"Sync","Name":"nodes.mycluster.domain.com","RolePolicyDocument":{"Name":"","Resource":{}},"ExportWithID":"nodes"},"PolicyDocument":null,"ExternalPolicies":null,"Managed":true} and *awstasks.IAMRolePolicy {"ID":null,"Lifecycle":"Sync","Name":"node-policyoverride","Role":{"ID":null,"Lifecycle":"Sync","Name":"secondgroup.mycluster.domain.com","RolePolicyDocument":{"Name":"","Resource":{}},"ExportWithID":"nodes"},"PolicyDocument":null,"ExternalPolicies":null,"Managed":true}
```

With this patch, everything works correctly without fatal errors.

### DIFF
--- a/pkg/model/iam.go
+++ b/pkg/model/iam.go
@@ -242,7 +242,7 @@ func (b *IAMModelBuilder) buildIAMTasks(role iam.Subject, iamName string, c *fi.
 		}
 
 		// Create External Policy tasks
-		{
+		if !shared {
 			var externalPolicies []string
 
 			if b.Cluster.Spec.ExternalPolicies != nil {


johngmyers:
The usual process is to PR into master branch, then cherry-pick into a release branch. Please send a PR into master branch instead.

hakman:
@johngmyers I don't think this PR is correct, or in any case has nothing to do with the original issue.
@kesor I am still trying to reproduce this and I am having a hard time as the initial issue doesn't have the IGs in the cluster manifest.

kesor:
> The usual process is to PR into master branch, then cherry-pick into a release branch. Please send a PR into master branch instead.

rebased on master

kesor:
> @johngmyers I don't think this PR is correct, or in any case has nothing to do with the original issue.

Which "original issue"? The one I describe in the PR description?
I have been chasing for the reason why trying to upgrade our clusters to 1.18 fails, and eventually foundthat `found duplicate tasks with name "IAMRolePolicy/node-policyoverride"` is directly related to the `externalPolicies` feature which we are not using, while the clusters work just fine with 1.17 as-is right now.

> @kesor I am still trying to reproduce this and I am having a hard time as the initial issue doesn't have the IGs in the cluster manifest.

Which initial issue?

hakman:
@kesor I thought it's related to https://github.com/kubernetes/kops/issues/9852.

I am trying to understand how to reproduce this to see if there is anything else that has to be considered.

kesor:
> @kesor I thought it's related to #9852.
> 
> I am trying to understand how to reproduce this to see if there is anything else that has to be considered.

This is unrelated. The bug in #9852 talks about the issue of shared `iam.profile` not working. This is kind of mentioned in the documentation for IAM in a "NOTE" quote.

This issue is talking about, if and when `iam.profile` is "correctly" used, then the new 1.18 feature `externalPolicies` is actually in conflict with it and causes a bug that can be avoided by "closing off" this `externalPolicies` new feature completely, and not scheduling its `AddTask` unless there were actually some `externalPolicies` present in the cluster spec.

hakman (changes requested):
In general it would be helpful if you would open first an issue with an easy way to reproduce. 

For this case, the proposed change indeed skips the `AddTask` for `externalPolicies` which seems reasonable at first sight. Except that the pattern used by kops for tasks is to reuse them for deleting and updating the resources and skipping the task will make it impossible to disable the external policies once they are applied.

IMHO the correct solution would be to check if the IAM profile is not shared (meaning is overridden in the IG spec) and add the task only for the kops managed profile.

pkg/model/iam.go
```diff
@@ -171,13 +171,11 @@ func (b *IAMModelBuilder) buildIAMTasks(igRole kops.InstanceGroupRole, iamName s
 		}
 
 		// Create External Policy tasks
-		{
+		if b.Cluster.Spec.ExternalPolicies != nil {
```
hakman, line 174:
```suggestion
		if !shared {
```

kesor: This also works.

hakman: Thanks for verifying @kesor. Can you update the PR with the suggested change and rebase to remove the "contains-merge-commits" label?

kesor: Done! Any chance to see this in `v1.18.1`?

hakman: Unfortunately a bit late for [v1.18.1](https://github.com/kubernetes/kops/releases/tag/v1.18.1), but should be part of v1.18.2.

hakman:
One more thing, would you mind adding the same condition for the next code block? I think there is a similar issue there.
https://github.com/kubernetes/kops/blob/036ea6952554d86fd452f244127957beceb841c5/pkg/model/iam.go#L265-L267

kesor:
> One more thing, would you mind adding the same condition for the next code block? I think there is a similar issue there.
> https://github.com/kubernetes/kops/blob/036ea6952554d86fd452f244127957beceb841c5/pkg/model/iam.go#L265-L267

We are using `additionalPolicies` in our clusters, and it works fine. So adding a condition here, will actually break things! So no ... `addtionalPolicies` and `externalPolicies` are apparently different enough that its a bad idea to do it.

hakman:
From my point of view, additional policies should not be combined with existing IAM profiles.
kops should not modify the existing IAM profiles, because those could be used in multiple clusters, with different additional policies.

Could you maybe explain what would be the use case for it?

kesor:
> From my point of view, additional policies should not be combined with existing IAM profiles.
> kops should not modify the existing IAM profiles, because those could be used in multiple clusters, with different additional policies.
> 
> Could you maybe explain what would be the use case for it?

Maybe it shouldn't ... but it does, and that means that closing it off behind a condition will break things.

The idea is that even though the IAM profile was created before the IG, it doesn't mean that the person managing the cluster should loose the ability to affect ALL IAM ROLES OF ALL IGs from one central location - which is the cluster `additionalPolicy.node:` location.

In any case, its another issue and doesn't really have to do with `externalPolicies` breaking 1.18.

hakman:
I understand your point of view, but the external/shared policies should only be modified externally.
I agree that it is a similar, but different issue.
/lgtm
/approve

PR #9869, opened 2020-09-04
Update alpha channel k8s 1.16 to 1.16.15 and Ubuntu image to latest available

MoShitrit:
https://github.com/kubernetes/kubernetes/releases/tag/v1.16.15

### DIFF
--- a/channels/alpha
+++ b/channels/alpha
@@ -65,7 +65,7 @@ spec:
     recommendedVersion: 1.17.11
     requiredVersion: 1.17.0
   - range: ">=1.16.0"
-    recommendedVersion: 1.16.14
+    recommendedVersion: 1.16.15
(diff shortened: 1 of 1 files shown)


hakman:
@MoShitrit could you also update the ubuntu image?

MoShitrit:
> @MoShitrit could you also update the ubuntu image?

Sure @hakman , 
So [1.16 image](https://github.com/kubernetes/kops/blob/master/channels/alpha#L44) should be changed to this as well? 
`099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200716`

hakman:
Meant that ubuntu has a newer image which we should update for 1.18 :).

PR #9872, opened 2020-09-05
Use root volume encryption flag for LaunchConfiguration with TF and CF

hakman:
#9793 missed the config flags for LaunchConfiguration with Terraform and CloudFormation.
This will be useful in kops 1.18.1.


hakman:
Thanks @olemarkus :)

PR #9874, opened 2020-09-05
Openstack golden servergrouptests

olemarkus:
Openstack servergroup test had a rather intricate method for building tasks and diffing them. Instead of using that specific code, the test now just generates golden outputs based on the same spec, which makes the tests a whole lot more manageable, as well as easier to figure out what would cause failures.

I did some rudimentary bit flipping in `servergroup.go` to see that this indeed causes test failures.


olemarkus:
/cc @johngmyers 
Something like this you asked for in #9575 ?

PR #9885, opened 2020-09-08
Errors when encryptionConfig is enabled, but no encryptionconfig secret

olemarkus:
When encryptionConfig is enabled, but the secret is missing, there is no visible errors anywhere. kube-apiserver just goes into a crashloop without any complains. This PR adds warnings both on the client side and through nodeup.

### DIFF
--- a/nodeup/pkg/model/kube_apiserver.go
+++ b/nodeup/pkg/model/kube_apiserver.go
@@ -71,8 +71,8 @@ func (b *KubeAPIServerBuilder) Build(c *fi.ModelBuilderContext) error {
 			}
 
 			key := "encryptionconfig"
-			encryptioncfg, _ := b.SecretStore.Secret(key)
-			if encryptioncfg != nil {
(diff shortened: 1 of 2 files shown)


nodeup/pkg/model/kube_apiserver.go
```diff
@@ -71,7 +71,7 @@ func (b *KubeAPIServerBuilder) Build(c *fi.ModelBuilderContext) error {
 			}
 
 			key := "encryptionconfig"
-			encryptioncfg, _ := b.SecretStore.Secret(key)
+			encryptioncfg, err := b.SecretStore.Secret(key)
 			if encryptioncfg != nil {
```
hakman, line 75: This seems odd to read the error, but condition based on the `encryptioncfg` value.

olemarkus: Yeah, you are right. If the secret is not set, `Secret` will return an error too. So I guess we can trust we have a secret if `err` is `nil`. Amended.

PR #9900, opened 2020-09-09
Continue if asg instance is unknown

olemarkus:
Most likely this comes from ASG thinking the instance is inService, but it is terminating when we run describe from EC2.

`kops get instances` should maybe also show instance status: terminating at some point, but that would also require changes in the rolling update code as it now assumes we filter terminating/terminated nodes.

### DIFF
--- a/upup/pkg/fi/cloudup/awsup/aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/aws_cloud.go
@@ -778,6 +778,9 @@ func awsBuildCloudInstanceGroup(c AWSCloud, cluster *kops.Cluster, ig *kops.Inst
 			klog.Warningf("ignoring instance as it is terminating: %s in autoscaling group: %s", id, cg.HumanName)
 			continue
 		}
+		if instances[id] == nil {
+			continue
+		}
 		currentConfigName := findInstanceLaunchConfiguration(i)
 		status := cloudinstances.CloudInstanceStatusUpToDate
 		if newConfigName != currentConfigName {


upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -768,6 +768,9 @@ func awsBuildCloudInstanceGroup(c AWSCloud, cluster *kops.Cluster, ig *kops.Inst
 
 	for _, i := range g.Instances {
 		id := aws.StringValue(i.InstanceId)
+		if instances[id] == nil {
```
johngmyers, line 771: Perhaps put after the following if block, as `instances[""]` is unlikely to be non-`nil`.

It should probably be after the `LifecycleStateTerminating` check as well.

Would we want to `klog.Warningf` for this as well?

olemarkus: Makes sense to move the function. Done.

This scenario is so common that I would find warnings to be noisy. It would happen consistently through a rolling update for example.

johngmyers: We consistently get the "ignoring instance as it is terminating" warning during rolling updates. I consider it more informative than noisy, but to be fair it should be info instead of warning.

I suppose in this case we don't have anything informative to log. This is a semantic change: ignoring these instances will cause them to not contribute to satisfying the "enough instances in group" cluster validation check (which might be good) but it will also prevent them from being updated if they're stuck in that state (and on an old spec).

johngmyers:
I'll let this proceed on the assumption that it isn't likely that instances would get stuck in this state.
/lgtm
/approve

By the way, I noticed that `addCloudInstanceData()` won't set `PrivateIP` if it can't find a role tag. Perhaps that assignment should be pulled out of the for loop?

PR #9907, opened 2020-09-09
Remove constraint of setting volume type for OS

### LINKED ISSUE #9893: PV creation fails on openstack
 this information.**
Version 1.18.0 (git-698bf974d8)

 a `kops` flag.**
v1.19.0

openstack

kops create cluster --cloud openstack \
     	--name my-cluster.k8s.local   \
	--state ${KOPS_STATE_STORE} \
     	--zones nova \
     	--network-cidr 10.0.0.0/24 \
      	--image Ubuntu-20.04-focal \
      	--master-count=1 \
      	--node-count=1 \
      	--node-size 8124120 \
      	--master-size 8124120 \
        --etcd-storage-type default \
      	--topology private \
      	--bastion \
      	--networking calico \
      	--api-loadbalancer-type public \
    	--ssh-public-key ~/sshkey.pub \
     	--os-ext-net ext-net-vlan52 \
        --os-kubelet-ignore-az=true 	

The VMs get created on openstack, but the creation of the Persistent Volume fails with the following error message:

W0908 09:56:31.548562   10816 executor.go:128] error running task "Volume/a.etcd-main.my-cluster.k8s.local" (37s remaining to succeed): error creating PersistentVolume: error creating volume {20 nova   map[KubernetesCluster:my-cluster.k8s.local k8s.io/etcd/main:a/a k8s.io/role/master:1] a.etcd-main.my-cluster.k8s.local      default false}: Resource not found

The k8s cluster comes up 

```apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2020-09-08T16:46:21Z"
  name: my-cluster.k8s.local
spec:
  api:
    loadBalancer:
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudConfig:
    openstack:
      blockStorage:
        bs-version: v3
        ignore-volume-az: true
      loadbalancer:
        floatingNetwork: ext-net-vlan52
        floatingNetworkID: d286fe57-c256-4686-b00f-d45fe8d4eebe
        method: ROUND_ROBIN
        provider: haproxy
        useOctavia: false
      monitor:
        delay: 1m
        maxRetries: 3
        timeout: 30s
      router:
        externalNetwork: ext-net-vlan52
  cloudProvider: openstack
  configBase: swift://ois/my-cluster.k8s.local
  containerRuntime: docker
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-nova
      name: a
      volumeType: default
    memoryRequest: 100Mi
    name: main
    version: 3.2.24
iam:
    allowContainerRegistry: true
    legacy: false
  kubelet:
    anonymousAuth: false
  kubernetesApiAccess:
  - 0.0.0.0/0
  kubernetesVersion: 1.18.8
  masterPublicName: api.my-cluster.k8s.local
  networkCIDR: 10.0.0.0/24
  networking:
    calico:
      majorVersion: v3
  sshAccess:
  - 0.0.0.0/0
  subnets:
  - cidr: 10.0.0.32/27
    name: nova
    type: Private
    zone: nova
  - cidr: 10.0.0.0/30
    name: utility-nova
    type: Utility
    zone: nova
  topology:
    dns:
      type: Public
    masters: private
    nodes: private
---
kind: InstanceGroup
```

```
error running tasks: deadline exceeded executing task Volume/a.etcd-events.my-cluster.k8s.local. Example error: error creating PersistentVolume: error creating volume {20 nova   map[KubernetesCluster:my-cluster.k8s.local k8s.io/etcd/events:a/a k8s.io/role/master:1] a.etcd-events.my-cluster.k8s.local      default false}: Resource not found
```
This is being run against an openstack cluster where kubespray seems to work. I am trying this out to evaluate if this is a better option, as it seems to offer a few more options out of the box.

olemarkus:
There is no real reason to do this. In some cases this may even prevent clusters from starting where there is no explicit volume type defined in cinder.

fixes #9893

### DIFF
--- a/pkg/model/master_volumes.go
+++ b/pkg/model/master_volumes.go
@@ -247,9 +247,6 @@ func (b *MasterVolumeBuilder) addGCEVolume(c *fi.ModelBuilderContext, name strin
 
 func (b *MasterVolumeBuilder) addOpenstackVolume(c *fi.ModelBuilderContext, name string, volumeSize int32, zone string, etcd kops.EtcdClusterSpec, m kops.EtcdMemberSpec, allMembers []string) error {
 	volumeType := fi.StringValue(m.VolumeType)
-	if volumeType == "" {
-		return fmt.Errorf("must set ETCDMemberSpec.VolumeType on Openstack platform")
-	}
 
 	// The tags are how protokube knows to mount the volume and use it for etcd
 	tags := make(map[string]string)


olemarkus:
/cc @zetaab 
Can you think of any reason why this constraint was added?

pkg/model/master_volumes.go
```diff
@@ -247,9 +247,6 @@ func (b *MasterVolumeBuilder) addGCEVolume(c *fi.ModelBuilderContext, name strin
 
 func (b *MasterVolumeBuilder) addOpenstackVolume(c *fi.ModelBuilderContext, name string, volumeSize int32, zone string, etcd kops.EtcdClusterSpec, m kops.EtcdMemberSpec, allMembers []string) error {
 	volumeType := fi.StringValue(m.VolumeType)
-	if volumeType == "" {
```
zetaab, line 250: if you remove this, then we should not either post volumetype to openstack APIs. https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/openstacktasks/volume.go#L134

I know that openstack uses default type if you do not define anything. However, this modification is not enough - we should not post anything to openstack api if we do not have volumetype defined

zetaab (changes requested):

pkg/model/master_volumes.go
olemarkus: My reasoning for only changing this particular code is:
`cinderv3.CreateOpts.VolumeType` is not a pointer so we don't change anything by conditionally setting that. The serialisation in gophercloud also uses `omitempty` so empty strings will not be serialised to the request payload.

zetaab (approved):
/approve
/lgtm

Issue #9806, opened 2020-08-24
Increased RateLimit (RequestLimitExceeded) running kops update at DescribeLaunchTemplates step


phspagiari:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
`Version 1.17.1 (git-4dfd4779fe)`
**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
`Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.11", GitCommit:"ea5f00d93211b7c80247bf607cfa422ad6fb5347", GitTreeState:"clean", BuildDate:"2020-08-13T15:11:47Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}`

**3. What cloud provider are you using?**
`AWS`

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
`kops -v10 update cluster $NAME 2>&1 | tee -a log`

**5. What happened after the commands executed?**
We see kops Executing Tasks getting LaunchTemplates and returing the results but in the sequence we can see a lot of requests without response and a lot of `RequestLimitExceeded` from `ec2/DescribeLaunchTemplateVersions`. We are getting RateLimit for around 10 minutes to run a simple update.

**6. What did you expect to happen?**
Run kops update without `RateLimit` / `RequestLimitExceeded` from AWS.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: null
  generation: 6
  name: CLUSTER-NAME
spec:
  additionalPolicies:
    node: |
      [
        {
          "Effect": "Allow",
          "Action": ["sts:AssumeRole"],
          "Resource": ["*"]
        },
        {
          "Effect": "Allow",
          "Action": ["autoscaling:DescribeAutoScalingGroups","autoscaling:DescribeAutoScalingInstances","autoscaling:DescribeTags","autoscaling:SetDesiredCapacity","autoscaling:TerminateInstanceInAutoScalingGroup"],
          "Resource": ["*"]
        }
      ]
  api:
    loadBalancer:
      type: Internal
      useForInternalApi: true
  authorization:
    rbac: {}
  channel: stable
[...]

PR #9909, opened 2020-09-10
Get launch template versions after filtering templates

johngmyers:
Should at least partially remediate #9806 by filtering the launch templates before enumerating their versions.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
+++ b/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
@@ -301,68 +301,53 @@ func (t *LaunchTemplate) findAllLaunchTemplates(c *fi.Context) ([]*ec2.LaunchTem
 	}
 }
 
-// findAllLaunchTemplateVersions returns all the launch templates versions for us
-func (t *LaunchTemplate) findAllLaunchTemplatesVersions(c *fi.Context) ([]*ec2.LaunchTemplateVersion, error) {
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -301,68 +301,53 @@ func (t *LaunchTemplate) findAllLaunchTemplates(c *fi.Context) ([]*ec2.LaunchTem
...
-					LaunchTemplateName: x.LaunchTemplateName,
-					NextToken:          next,
-				})
-				if err != nil {
-					return err
-				}
-				list = append(list, resp.LaunchTemplateVersions...)
-				if resp.NextToken == nil {
-					return nil
+		if strings.HasPrefix(aws.StringValue(x.LaunchTemplateName), prefix) {
```
justinsb, line 323: Totally out of scope for this PR, but apparently filters can also support wildcards, e.g. `fmt.Sprintf("%s-*", name)`.  I certainly didn't know that - so it wouldn't surprise me if we had multiple places where we could pre-filter by prefix.

Source: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html#Filtering_Resources_CLI

justinsb:
LGTM - will approve but hold for other reviewers.  Feel free to hold cancel after a few hours!

/approve
/lgtm

PR #9911, opened 2020-09-10
Allow the BootstrapClient task to run after Protokube

hakman:
Fixes: #9896

/cc @johngmyers

### DIFF
--- a/pkg/model/iam.go
+++ b/pkg/model/iam.go
@@ -22,6 +22,7 @@ import (
 
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
+	"k8s.io/kops/pkg/dns"
 	"k8s.io/kops/pkg/model/iam"
 	"k8s.io/kops/pkg/util/stringorslice"
 	"k8s.io/kops/upup/pkg/fi"
@@ -164,11 +165,13 @@ func (b *IAMModelBuilder) buildIAMRolePolicy(role iam.Subject, iamName string, i
 		},
 	}
 
-	// This is slightly tricky; we need to know the hosted zone id,
-	// but we might be creating the hosted zone dynamically.
-	// We create a stub-reference which will be combined by the execution engine.
-	iamPolicy.DNSZone = &awstasks.DNSZone{
-		Name: fi.String(b.NameForDNSZone()),
+	if !dns.IsGossipHostname(b.Cluster.ObjectMeta.Name) {
+		// This is slightly tricky; we need to know the hosted zone id,
+		// but we might be creating the hosted zone dynamically.
+		// We create a stub-reference which will be combined by the execution engine.
+		iamPolicy.DNSZone = &awstasks.DNSZone{
+			Name: fi.String(b.NameForDNSZone()),
+		}
 	}
 
 	t := &awstasks.IAMRolePolicy{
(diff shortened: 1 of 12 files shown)


pkg/model/iam.go
```diff
@@ -164,11 +165,13 @@ func (b *IAMModelBuilder) buildIAMRolePolicy(role iam.Subject, iamName string, i
...
-	// We create a stub-reference which will be combined by the execution engine.
-	iamPolicy.DNSZone = &awstasks.DNSZone{
-		Name: fi.String(b.NameForDNSZone()),
+	if !dns.IsGossipHostname(b.Cluster.ObjectMeta.Name) {
+		// This is slightly tricky; we need to know the hosted zone id,
+		// but we might be creating the hosted zone dynamically.
+		// We create a stub-reference which will be combined by the execution engine.
+		iamPolicy.DNSZone = &awstasks.DNSZone{
+			Name: fi.String(b.NameForDNSZone()),
+		}
```
johngmyers, line 174: This brings up an interesting question: should we be disabling the dns-controller addon when using gossip? I looks like we might be running it with IAM over no DNS zones?

upup/pkg/fi/nodeup/nodetasks/service.go
```diff
@@ -67,11 +72,22 @@ func (p *Service) GetDependencies(tasks map[string]fi.Task) []fi.Task {
...
-		case *File, *Package, *UpdatePackages, *UserTask, *GroupTask, *Chattr, *BindMount, *Archive:
+		switch v := v.(type) {
+		case *Package, *UpdatePackages, *UserTask, *GroupTask, *Chattr, *BindMount, *Archive:
 			deps = append(deps, v)
-		case *Service, *LoadImageTask, *IssueCert:
+		case *Service, *LoadImageTask, *IssueCert, *BootstrapClient:
 			// ignore
+		case *File:
+			// Skip adding kubeconfig files as dependencies for services other than kubelet
+			if strings.HasSuffix(v.Path, "/kubeconfig") && p.Name != kubeletService {
```
johngmyers, line 82: I have a preference for something more explicit and closer to the Builders than looking at the path suffix. One idea would be to add a `BeforeService string` field to `File` and populate it for all the kubeconfigs that can derive from `BootstrapClient`. The value `""` would make all `Service`s depend on it.

upup/pkg/fi/nodeup/nodetasks/service.go
```diff
@@ -67,11 +72,22 @@ func (p *Service) GetDependencies(tasks map[string]fi.Task) []fi.Task {
...
+				continue
+			}
+			deps = append(deps, v)
+		case *KubeConfig:
+			// Only add kubeconfigs as dependencies for kubelet
```
johngmyers, line 87: It's not really necessary to add this as a dependency, since the `File` that `kubeletService` depends on will pull this in as a transitive dependency.

pkg/model/iam.go
hakman: I created https://github.com/kubernetes/kops/issues/9915 so that we can discuss about it.

upup/pkg/fi/nodeup/nodetasks/service.go
hakman: Agreed, moved to the ignore list.

mikesplain:
I was bit by this a few days ago for some internal work.  Glad to see you're working on a fix!  Just tested and looks good to me!

/lgtm
/hold in case @johngmyers has more comments

upup/pkg/fi/nodeup/nodetasks/service.go
hakman: I think the latest changes should address this.

hakman:
Thanks @mikesplain will wait for the OK from @johngmyers and will cut the alpha.4 once this is merged.

PR #9917, opened 2020-09-11
Remove force_tcp flag for nodelocalcache dot zone

astrikos:
This PR is proposing a change in the nodelocalcache confimap. 
The change is removing the `force_tcp` flag from the `.` zone that handles all external DNS requests.
Using the current nodelocalcache configuration we have experienced increased latencies for our DNS requests that needs to go to our upstream DNS server (which in our case is AWS VPC's DNS servers).
Forcing all DNS communication go through TCP can lead to increased latency and slower DNS responses since a lot of upstream DNS servers are not optimized for TCP traffic.
An example is AWS's VPC DNS server that sending all the DNS traffic through TCP cause it to increase the latencies for this requests.
The kubernetes version has already changed this configmap to skip `force_tcp` flag for the . zone, so it might make sense to follow their example.

Reference https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L100

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -18594,9 +18594,7 @@ data:
         reload
         loop
         bind {{ KubeDNS.NodeLocalDNS.LocalIP }}{{ if NodeLocalDNSServerIP }} {{ NodeLocalDNSServerIP }}{{ end }}
-        forward . __PILLAR__UPSTREAM__SERVERS__ {
-          force_tcp
-        }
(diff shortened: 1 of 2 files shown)


olemarkus:
/ok-to-test

Are you able to sign the CLA?

astrikos:
>Are you able to sign the CLA?
I signed it

/check-cla

olemarkus:
/lgtm

Thanks for this!

PR #9918, opened 2020-09-11
Validate labels

olemarkus:
This fixes a couple of issues with labels reported by users:
* A mixup between cloudLabels (which allows many /) and nodeLabels (which don't) ended up in issues with kops-controller who couldn't set labels with a bad value
* Someone setting the kops IG node cloud label manually to something else than the IG name ended up in issues with node identity.

We should validate more, but that can come as future PRs.


mikesplain:
Very nice, I've noticed this as well, glad to get more validation going on it. Thanks!

/lgtm
/approve

PR #9921, opened 2020-09-11
Add addon for aws node termination handler

olemarkus:


pkg/apis/kops/cluster.go
```diff
@@ -158,6 +158,9 @@ type ClusterSpec struct {
 	CloudConfig                    *CloudConfiguration           `json:"cloudConfig,omitempty"`
 	ExternalDNS                    *ExternalDNSConfig            `json:"externalDns,omitempty"`
 
+	// Components contains the configuration for kops components.
+	Components *ComponentConfig `json:"components,omitempty"`
```
johngmyers, line 162: What is the definition of a "component"? What determines if something is a "component" versus an "addon" or just has its own subfield in the cluster spec?

pkg/apis/kops/v1alpha2/componentconfig.go
```diff
@@ -760,6 +760,28 @@ type CloudConfiguration struct {
...
+// ComponentConfig contains the configuration for kops components.
+type ComponentConfig struct {
+	// NodeTerminationHandlerConfig determines the cluster autoscaler configuration.
+	NodeTerminationHandler *NodeTerminationHandlerConfig `json:"nodeTerminationHandler,omitempty"`
+}
+
+// NodeTerminationHandlerConfig determines the cluster autoscaler configuration.
+type NodeTerminationHandlerConfig struct {
+	// Enabled enables the node termination handler.
+	// Default: false
```
johngmyers, line 772: Should default to `true` when there is a non-`nil` `nodeTerminationHandler`

pkg/model/components/nodeterminationhandler.go
```diff
@@ -0,0 +1,49 @@
...
+
+package components
+
+import (
+	"k8s.io/kops/pkg/apis/kops"
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/loader"
+)
+
+// NodeTerminationHandlerOptionsBuilder adds options for cluster autoscaler to the model
```
johngmyers, line 25:
```suggestion
// NodeTerminationHandlerOptionsBuilder adds options for the node termination handler to the model.
```

pkg/model/components/nodeterminationhandler.go
```diff
@@ -0,0 +1,49 @@
...
+type NodeTerminationHandlerOptionsBuilder struct {
+	*OptionsContext
+}
+
+var _ loader.OptionsBuilder = &NodeTerminationHandlerOptionsBuilder{}
+
+func (b *NodeTerminationHandlerOptionsBuilder) BuildOptions(o interface{}) error {
+	clusterSpec := o.(*kops.ClusterSpec)
+	if clusterSpec.Components == nil || clusterSpec.Components.NodeTerminationHandler == nil {
```
johngmyers, line 34: We should probably enable node termination handler by default for k8s 1.19+.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -542,6 +542,32 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
 		}
 	}
 
+	components := b.Cluster.Spec.Components
+
+	if components != nil {
+		nth := components.NodeTerminationHandler
+
+		if nth != nil && fi.BoolValue(nth.Enabled) {
```
johngmyers, line 550:
```suggestion
		if fi.BoolValue(nth.Enabled) {
```

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -542,6 +542,32 @@ func (b *BootstrapChannelBuilder) buildAddons() *channelsapi.Addons {
...
+			{
+				location := key + "/k8s-1.12.yaml"
+				id := "k8s-1.12"
+
+				addons.Spec.Addons = append(addons.Spec.Addons, &channelsapi.AddonSpec{
+					Name:              fi.String(key),
+					Version:           fi.String(version),
+					Selector:          map[string]string{"k8s-addon": key},
+					Manifest:          fi.String(location),
+					KubernetesVersion: ">=1.12.0",
```
johngmyers, line 564: If this is going into kops 1.19, you'll need to do something for k8s 1.11. Prohibiting it through API validation would be fine.

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.12.yaml.template
```diff
@@ -0,0 +1,176 @@
...
+            value: "true"
+          - name: ENABLE_PROMETHEUS_SERVER
+            value: "{{ .EnablePrometheusMetrics }}"
+          resources:
+            limits:
+              memory: 128Mi
+            requests:
+              cpu: 50m
+              memory: 64Mi
+      nodeSelector:
```
johngmyers, line 172: I think it would make sense to at least have the option of adding a `nodeSelector` that restricts it to "node-role.kubernetes.io/spot-worker": "true"

It might make sense to make that the default when scheduled event draining is disabled.

johngmyers (changes requested):

pkg/apis/kops/componentconfig.go
```diff
@@ -759,6 +759,28 @@ type CloudConfiguration struct {
...
+}
+
+// NodeTerminationHandlerConfig determines the cluster autoscaler configuration.
+type NodeTerminationHandlerConfig struct {
+	// Enabled enables the node termination handler.
+	// Default: false
+	Enabled *bool `json:"enabled,omitempty"`
+	// EnableSpotInterruptionDraining makes node termination handler drain nodes when spot interruption termination notice is received.
+	// Default: true
+	EnableSpotInterruptionDraining *bool `json:"enableSpotInterruptionDraining,omitempty"`
```
johngmyers, line 775: I'm not quite sure why someone would want to turn this off.

pkg/apis/kops/cluster.go
olemarkus: The reason for adding a new struct was to not further "bloat" the cluster spec. Was briefly mentioned on slack.

Addons was already used for those where you provide manifest urls and such. Didn't quite seem to quite fit what I wanted to do.
The reason for calling it component is that we the API spec for this is in componentconfig.go and that the options builders are also called components.

If this part goes in, I want to move CAS in there as well. In future spec versions, other components like kubedns, kubeproxy etc could go in there as well.

pkg/apis/kops/componentconfig.go
olemarkus: Someone asked for this on slack. They wanted just the cordon.

pkg/apis/kops/v1alpha2/componentconfig.go
olemarkus: Thanks. Will fix that.

pkg/model/components/nodeterminationhandler.go
olemarkus: I am not against that. But I suggest testing how this works in one version of kops first. So maybe enable it for 1.20+.

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.12.yaml.template
olemarkus: Not against this one either. But perhaps as a follow-up?

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
olemarkus: My mistake. I thought perhaps it was 1.12. I didn't actually try this on any older versions if there are manifest incompatibilities. I can quickly test that part.

pkg/model/components/nodeterminationhandler.go
hakman: Please don't enable by default.

pkg/apis/kops/v1alpha2/componentconfig.go
hakman: Do we have this behaviour anywhere else? Should we document and agree about it first at next office hours?

pkg/apis/kops/cluster.go
hakman: I am ok with having a new structure to keep new addons separated. Depending on the agreed purpose, the name is not quite right.
I don't think kubedns, kubeproxy, ... can be moved here, because the have some established sections already.

pkg/apis/kops/v1alpha2/componentconfig.go
olemarkus: CAS did not have this. I know kubeproxy does, but that one is a bit special. Either way is fine for me.

pkg/apis/kops/cluster.go
olemarkus: Just to be clear: I am thinking of when we do a full new API version.

Naming things is hard, but it was the only one I could think of besides "addons" that was already in use.

hakman: I agree about the "hard" part :). I would see kube-proxy, kube-... in an Kubernetes struct, not in the "addons" one that should be optional managed things.

pkg/apis/kops/v1alpha2/componentconfig.go
hakman: kube-proxy has this behaviour because it is (historically) a required component.

pkg/apis/kops/cluster.go
olemarkus: So I could always add this to the addons struct instead, but it does feel a bit strange to me.

pkg/apis/kops/v1alpha2/componentconfig.go
olemarkus: Precedence would say that it should be disabled by default even if the struct is non-nil. I'd also think that is the most intuitive default.

pkg/apis/kops/cluster.go
johngmyers: Adding a new level to the v1alpha2 API makes it even more inconsistent than it already is. It might make sense to reorganize things when we can do a v1beta1 API, but adding a new level at this point is not going to make things any less confusing.

This is fundamentally an AWS-specific feature and should be off of the AWS-specific cloud provider settings. The fact that it is implemented as a bootstrap addon is an implementation detail.

hakman: Sounds good to me. So, same way as autoscaler was done.

olemarkus: Moving it to the AWS struct sounds fine for me. But autoscaler stays where it is then.

pkg/apis/kops/v1alpha2/componentconfig.go
johngmyers: I would disagree that it is the most intuitive default. Why would anyone specify a non-nil struct if they don't want that thing enabled? With the enabled field defaulting to false, specifying a non-nil struct and any other subfield will have no effect unless they also specify an explicit value for `enabled`.

Numerous existing things, such as `cloudControllerManager`, the authentication, authorization, networking, and API access options, etc. are enabled by specifying non-nil structs.

The point of an `enabled` field is to be able to disable a whole subcomponent/feature which has multiple config settings.

hakman: OK, but let's document somewhere for the future that we expect this behaviour for new components.

PR #9924, opened 2020-09-12
Only add additional policies to kops managed IAMRoles

hakman:
While reviewing https://github.com/kubernetes/kops/pull/9867 I noticed that kops modifies existing/shared IAM instance profiles when setting additional policies.

In my opinion, kops should only modify the resources it creates. If changes are applied on external resources shared between multiple clusters, the consequences would be unpredictable.


olemarkus:
100% agree with @hakman 
/lgtm

PR #9936, opened 2020-09-15
use subnet also when creating floatingip

zetaab:


upup/pkg/fi/cloudup/openstacktasks/floatingip.go
```diff
@@ -209,40 +209,32 @@ func (f *FloatingIP) RenderOpenstack(t *openstack.OpenstackAPITarget, a, e, chan
...
-				FloatingNetworkID: external.ID,
-				Description:       fi.StringValue(e.Name),
-			})
-			if err != nil {
-				return fmt.Errorf("failed to create floating IP: %v", err)
-			}
-			e.ID = fi.String(fip.ID)
-			e.IP = fi.String(fip.FloatingIP)
-
+		// instance floatingips are also coming from same subnet than kubernetes API floatingip
```
olemarkus, line 222: Nit, but I had to read this one a few times
```suggestion
		// instance floatingips comes from the same subnet as the kubernetes API floatingip
```

PR #9938, opened 2020-09-15
Improve kops get instances when api is unavailable

olemarkus:
When the api is unavailable, kops will say all the nodes have not yet joined the cluster. That is not the case simply because e.g the admin credentials have been expired. This PR makes it a bit more clear that we cannot know the node name when the API is unavailable.

### DIFF
--- a/cmd/kops/get_instances.go
+++ b/cmd/kops/get_instances.go
@@ -110,7 +110,7 @@ func RunGetInstances(ctx context.Context, f *util.Factory, out io.Writer, option
 
 	nodeList, err := k8sClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
 	if err != nil {
-		klog.V(2).Infof("error listing nodes: %v", err)
+		klog.Warningf("cannot list node names. Kubernetes API unavailable: %v", err)
 	}
 
 	igList, err := clientset.InstanceGroupsFor(cluster).List(ctx, metav1.ListOptions{})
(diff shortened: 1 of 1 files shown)


cmd/kops/get_instances.go
```diff
@@ -108,9 +108,11 @@ func RunGetInstances(ctx context.Context, f *util.Factory, out io.Writer, option
 		return err
 	}
 
+	emptyNodeName := "NotJoinedYet"
 	nodeList, err := k8sClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
 	if err != nil {
-		klog.V(2).Infof("error listing nodes: %v", err)
+		klog.Warningf("error listing nodes. Cannot determine node status: %v", err)
+		emptyNodeName = "APIUnavailable"
```
rdrgmnzs, line 115: NIT: Should we make this a bit more clear that it's the k8s API that is unavailable since we normally deal with both the k8s api and cloud provider API?

rdrgmnzs (commented):
LGTM only have 1 nit.

cmd/kops/get_instances.go
olemarkus: Any suggestion on the wording?  don't think we can make the node name more verbose at least.

rifelpet:
Maybe it makes more sense for the NODE-NAME column to use an empty string or `N/A` for every value? and the `klog.Warningf` should make it clear that the API is unavailable if `err` doesn't provide enough info itself. I think that would be more intuitive than having a column of `APIUnavailable` values for the NODE-NAME column. If the API server is unavailable then we will definitely not know any node names... maybe we remove the column altogether if thats feasible.

PR #9940, opened 2020-09-15
Set minimum Terraform version to 0.12.26/0.13.0

bmelbourne:
As Terraform 0.11 is in the later stages of end-of-life (EOL), the Terraform build/test templates have been updated to be fully compatible with Terraform `0.12.26+` and `0.13.0+`.

https://www.hashicorp.com/blog/deprecating-terraform-0-11-support-in-terraform-providers

Minimum Terraform AWS provider is set to `v3.7.0` and Google provider is set to `v3.39.0`.

N/A

The `KOPS_FEATURE_FLAGS=Terraform-0.12` will be deprecated in Kops v0.20.

Terraform reference notes regarding 0.12/0.13 provider compatibility.

https://www.terraform.io/docs/configuration/provider-requirements.html#v0-12-compatible-provider-requirements
https://www.terraform.io/upgrade-guides/0-13.html#explicit-provider-source-locations


hakman:
Seems pretty cool @bmelbourne :). Would you mind restaging into 2+ commits to make it easier to review? At least the generated output should be a separate commit.
/ok-to-test
/cc @rifelpet

bmelbourne:
> Seems pretty cool @bmelbourne :). Would you mind restaging into 2+ commits to make it easier to review? At least the generated output should be a separate commit.
> /ok-to-test
> /cc @rifelpet

@hakman not sure I understand `the generated output` part...also I've always been told by the K8s community to squash all commits into a single commit prior to PR review...I'm not familiar with the steps to creating 2+ commits once I have a tested single commit?

upup/pkg/fi/cloudup/terraform/target_hcl2.go
```diff
@@ -87,7 +84,22 @@ func (t *TerraformTarget) finish012(taskMap map[string]fi.Task) error {
...
+	terraformBody.SetAttributeValue("required_version", cty.StringVal(">= 0.12.26"))
+
+	requiredProvidersBlock := terraformBody.AppendNewBlock("required_providers", []string{})
+	requiredProvidersBody := requiredProvidersBlock.Body()
+
+	if t.Cloud.ProviderID() == kops.CloudProviderGCE {
+		requiredProviderBlockGoogle := requiredProvidersBody.AppendNewBlock("google = ", []string{})
+		requiredProviderBodyGoogle := requiredProviderBlockGoogle.Body()
+		requiredProviderBodyGoogle.SetAttributeValue("source", cty.StringVal("hashicorp/google"))
+		requiredProviderBodyGoogle.SetAttributeValue("version", cty.StringVal(">= 3.38.0"))
```
rifelpet, line 96: I think rather than making the block name include an equals sign it might be more consistent to use the `writeMap` function in hcl2.go

(untested)
```suggestion
		writeMap(requiredProvidersBody, "google", map[string]cty.Value{
			"source": cty.StringVal("hashicorp/google"),
			"version": cty.StringVal(">= 3.38.0"),
		})
```

rifelpet (commented):
Thanks for this cleanup! one nit but otherwise looks great. I appreciate the file renaming from 0_12 to hcl2 too.

My only concern is around timing and whether we want this in 1.19 or 1.20. I think the term we'd want to use is that the feature flag will be _removed_, in which case we'd want to announce the deprecation ahead of time (ideally in the previous minor release). Its a bit late into the 1.19 release cycle and 1.18.0 is already out, so we may want to defer this to 1.20 in which case we should have a separate PR that adds the deprecation announcement to the 1.19 release notes and then this PR will need to update the table in docs/terraform.md to use `>= 1.20`.

Anyone else have thoughts on timing? Implementation-wise I think this is ready to go.

bmelbourne:
> Thanks for this cleanup! one nit but otherwise looks great. I appreciate the file renaming from 0_12 to hcl2 too.
> 
> My only concern is around timing and whether we want this in 1.19 or 1.20. I think the term we'd want to use is that the feature flag will be _removed_, in which case we'd want to announce the deprecation ahead of time (ideally in the previous minor release). Its a bit late into the 1.19 release cycle and 1.18.0 is already out, so we may want to defer this to 1.20 in which case we should have a separate PR that adds the deprecation announcement to the 1.19 release notes and then this PR will need to update the table in docs/terraform.md to use `>= 1.20`.
> 
> Anyone else have thoughts on timing? Implementation-wise I think this is ready to go.

@rifelpet 
Notifying the users of the depreciation of the feature flag `Terraform-0.12` in the v1.19 release absolutely makes sense. If the other members agree, I'll update the `terraform.md` to reflect the delay to the v1.20 release. Do you have an example of a previous depreciation announcement which I can use for the new PR?

Also, I'll take a look at the recommended code changes.
I'm not too familiar with Go, hence why the code hasn't been optimised.

upup/pkg/fi/cloudup/terraform/target_hcl2.go
bmelbourne: I've tested the new block but it creates invalid Terraform HCL2 code as it encloses double quotes around both `source` and `version` keys. Is there a `cty` function to create this string as a key without the double quotes?

```
        terraform {
          required_version = ">= 0.12.26"
          required_providers {
            aws = {
              "source"  = "hashicorp/aws"
              "version" = ">= 3.6.0"
            }
          }
        }
```

rifelpet: I think the double quotes are still valid, I tested it locally with 0.12 and 0.13 and init seemed to work fine:

```hcl
        terraform {
          required_version = ">= 0.12.26"
          required_providers {
            aws = {
              "source"  = "hashicorp/aws"
              "version" = ">= 3.6.0"
            }
          }
        }
```
```
providertest $ terraform0.12 init

Initializing the backend...

Initializing provider plugins...
- Checking for available provider plugins...
- Downloading plugin for provider "aws" (hashicorp/aws) 3.6.0...


Warning: Provider source not supported in Terraform v0.12

  on main.tf line 4, in terraform:
   4:             aws = {
   5:               "source"  = "hashicorp/aws"
   6:               "version" = ">= 3.6.0"
   7:             }

A source was declared for provider aws. Terraform v0.12 does not support the
provider source attribute. It will be ignored.

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.
providertest $ terraform0.13 init

Initializing the backend...

Initializing provider plugins...
- Finding hashicorp/aws versions matching ">= 3.6.0"...
- Installing hashicorp/aws v3.6.0...
- Installed hashicorp/aws v3.6.0 (signed by HashiCorp)

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.

```

bmelbourne:
> > Thanks for this cleanup! one nit but otherwise looks great. I appreciate the file renaming from 0_12 to hcl2 too.
> > My only concern is around timing and whether we want this in 1.19 or 1.20. I think the term we'd want to use is that the feature flag will be _removed_, in which case we'd want to announce the deprecation ahead of time (ideally in the previous minor release). Its a bit late into the 1.19 release cycle and 1.18.0 is already out, so we may want to defer this to 1.20 in which case we should have a separate PR that adds the deprecation announcement to the 1.19 release notes and then this PR will need to update the table in docs/terraform.md to use `>= 1.20`.
> > Anyone else have thoughts on timing? Implementation-wise I think this is ready to go.
> 
> @rifelpet
> Notifying the users of the depreciation of the feature flag `Terraform-0.12` in the v1.19 release absolutely makes sense. If the other members agree, I'll update the `terraform.md` to reflect the delay to the v1.20 release. Do you have an example of a previous depreciation announcement which I can use for the new PR?
> 
> Also, I'll take a look at the recommended code changes.
> I'm not too familiar with Go, hence why the code hasn't been optimised.

@rifelpet 
Code changes have been tested and committed, as well as a few other commits to update the TF provider versions.

I'll squash these once the review is complete prior to merging for the v1.20 release.

I'll raise the PR for the deprecation announcement in the near future.

hakman:
@bmelbourne @rifelpet now that we switched to 1.20, should this be merged?

johngmyers:
I am splitting out removing `TerraformJSON`'s dependency on disabling `Terraform-0.12` in #10106 as that could then be backported to 1.19 branch.

hakman:
@bmelbourne I think you deleted `vendor/github.com/BurntSushi/toml/session.vim` by mistake.

bmelbourne:
> @bmelbourne I think you deleted `vendor/github.com/BurntSushi/toml/session.vim` by mistake.

It's much worse than that...it appears that Git and my WSL Ubuntu machine decided to have a party in my PR branch, which seems have corrupted various files, commits, etc. So the only option was to delete my forked repo and re-clone the kops repo.

I'll have to raise a new PR, as there is no way to link this PR back to my new forked repo 😒

hakman:
You have an edit repo on top. That should work to set a new PR.

bmelbourne:
FYI
Seems Virgin Media Web Safe settings were primarily responsible for my GitHub branch sync issues as it was terminating https connections abruptly and blocking ssh connections 😞

New PR #10109 raised (from new branch) with same changes and latest master rebase.

PR #9945, opened 2020-09-15
Support ChainInsertMode config option for Calico Networking

asmith030:
Add support for the ChainInsertMode configuration option when using Calico networking.

From https://docs.projectcalico.org/reference/felix/configuration:

Controls whether Felix hooks the kernel’s top-level iptables chains by inserting a rule at the top of the chain or by appending a rule at the bottom. Insert is the safe default since it prevents Calico’s rules from being bypassed. If you switch to Append mode, be sure that the other rules in the chains signal acceptance by falling through to the Calico rules, otherwise the Calico policy will be bypassed. [Default: Insert]

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -12837,6 +12837,9 @@ spec:
               value: "true"
 
             # kops additions
+            # Controls whether Felix inserts rules to the top of iptables chains, or appends to the bottom
+            - name: FELIX_CHAININSERTMODE
+              value: "{{- or .Networking.Calico.ChainInsertMode "insert" }}"
             # Set Felix iptables binary variant, Legacy or NFT
             - name: FELIX_IPTABLESBACKEND
               value: "{{- or .Networking.Calico.IptablesBackend "Auto" }}"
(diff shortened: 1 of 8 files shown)


upup/models/bindata.go
```diff
@@ -12828,6 +12828,9 @@ spec:
             # Set Felix endpoint to host default action to ACCEPT.
             - name: FELIX_DEFAULTENDPOINTTOHOSTACTION
               value: "ACCEPT"
+            # Controls whether Felix inserts rules to the top of iptables chains, or appends to the bottom
+            - name: FELIX_CHAININSERTMODE
+              value: "{{- or .Networking.Calico.ChainInsertMode "insert" }}"
```
KashifSaadat, line 12833: There's a section further below ` # kops additions`, can you move it to there?

hakman (changes requested):
Please also add validation for the new option, similar to the Canal one. Other than that, looks good.
https://github.com/kubernetes/kops/blob/036ea6952554d86fd452f244127957beceb841c5/pkg/apis/kops/validation/validation.go#L657-L660

KashifSaadat (approved):
Thanks for the contribution @asmith030 !

/ok-to-test
/lgtm
/approve

PR #9952, opened 2020-09-16
[Networking][Cilium] Allow to configure enable-endpoint-routes without enable Ipam

jorge07:


olemarkus:
Could you explain how you plan on using this? If you enable endpoint routes, you will need something else to route pod traffic to other nodes. At least if destination is in another subnet.

javipolo:
Hello. I'm working with @jorge07 in this issue

We are setting up a cluster and when there was only one node and one master, everything seemed to work properly:
- traefik as ingress controller with proxy-protocol enabled
- one pod with one application
- whitelisting in ingress

This set up seemed to work, but as soon as we launched another node, we started to have problems with whitelisting

The issue is that when the AWS loadbalancer sends the request to the node that has traefik, the client IP address is preserved and thus whitelisting works, but when the request is sent to the other node, instead of the client IP address traefik receives the cillium_host IP address of the other node as the client IP, and so whitelisting stops working

We discovered that if we enabled the setting enable-endpoint-routes, all the requests arrived with the proper Client IP address, and our problem was received ... but as @olemarkus points out, by doing this we've screwed the comunication with pods that are in different nodes

Which is the correct way to set this up? We "just" need to get the client ip address properly in our ingress controller ...

olemarkus:
Outgoing traffic from the cluster to an ELB will be NATed behind the node IP. Maybe that is what you are seeing? If this is the case, you most likely need a separate service for in-cluster and external traffic.

Proxy-protocol will work just fine for any traffic originating outside of the cluster.

This is more of a generic K8s issue though. You may want to read this one: https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip

jorge07:
The problem were facing is that the  application Ina different node than traefik is receiving cillium_host ip a private one instead the real IP of the user. So we thought that enabling entry point routing will solve that. It does but bring way bigger problems

jorge07:
`externalTrafficPolicy: Cluster` ->`externalTrafficPolicy: Local` 
🤦 
sorry

PR #9958, opened 2020-09-17
Use all kops mirrors to determine artifacts hashes

hakman:
Current implementation only looks for hash files in the S3 mirror and, if that is down, kops commands will fail with:
```
cannot determine hash for "https://kubeupv2.s3.amazonaws.com/kops/1.19.0-alpha.4/linux/amd64/nodeup"
(have you specified a valid file location?)
```

The code in the assets remapping and mirroring is quite complex and even if this looks a bit less elegant than I would like, it works pretty well.

Fixes: #9951

### DIFF
--- a/pkg/assets/builder.go
+++ b/pkg/assets/builder.go
@@ -318,11 +318,7 @@ func (a *AssetBuilder) findHash(file *FileAsset) (*hashing.Hash, error) {
 				b, err := vfs.Context.ReadFile(hashURL, vfs.WithBackoff(backoff))
 				if err != nil {
 					// Try to log without being too alarming - issue #7550
-					if ext == ".sha256" {
-						klog.V(2).Infof("Unable to read new sha256 hash file (is this an older/unsupported kubernetes release?) %q: %v", hashURL, err)
-					} else {
-						klog.V(2).Infof("Unable to read hash file %q: %v", hashURL, err)
-					}
+					klog.V(2).Infof("Unable to read hash file %q: %v", hashURL, err)
 					continue
 				}
 				hashString := strings.TrimSpace(string(b))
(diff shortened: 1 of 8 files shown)


zetaab (approved):
/lgtm

rifelpet:
This new logic and the constants look very similar to https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/urls.go would it make sense to consolidate them?

hakman:
> This new logic and the constants look very similar to https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/urls.go would it make sense to consolidate them?

It would make sense to move this to a function somewhere they both could use, but no idea where. It can also be done in a followup PR once we agree where to extract it.

pkg/assets/builder.go
```diff
@@ -311,27 +312,30 @@ func (a *AssetBuilder) findHash(file *FileAsset) (*hashing.Hash, error) {
...
-				} else {
-					klog.V(2).Infof("unable to read hash file %q: %v", hashURL, err)
+			for _, mirror := range mirrors.FindUrlMirrors(u.String()) {
+				hashURL := mirror + ext
+				klog.V(3).Infof("Trying to read hash fie: %q", hashURL)
+				b, err := vfs.Context.ReadFile(hashURL, vfs.WithBackoff(backoff))
+				if err != nil {
+					// Try to log without being too alarming - issue #7550
+					if ext == ".sha256" {
+						klog.V(2).Infof("Unable to read new sha256 hash file (is this an older/unsupported kubernetes release?) %q: %v", hashURL, err)
```
johngmyers, line 322: Should log this only after trying all mirrors.

johngmyers (changes requested):

pkg/assets/builder.go
hakman: Good idea. Done.

zetaab (approved):
/lgtm

PR #9961, opened 2020-09-17
AWS add cluster tag to detachinstances/findinstances filters

zetaab:


hakman:
Hmm, maybe squash the commits?

olemarkus:
Tested by running rolling-update in parallel. Without this patch, rogue detached instances show up during validation, with this patch, there are none.

/lgtm

PR #9965, opened 2020-09-17
Fix fip description

### LINKED ISSUE #9957: Failed to associate floatingip to bastion
 this information.**

1.19 alpha 4

 a `kops` flag.**

1.19.1

OpenStack

I updated cluster that was created with kops 1.18 to kops 1.19 alpha 4

when running kops update cluster --yes

W0916 20:17:53.089607 197 executor.go:131] error running task "Instance/bastions-1-updateospr-ece461-k8s-local" (8m20s remaining to succeed): failed to associated floating IP to instance bastions-1-updateospr-ece461-k8s-local: Expected HTTP response code [200] when accessing [PUT https://xxx:13696/v2.0/floatingips/29f31140-6a9d-45e3-9cda-75b57e6f44db], but got 409 instead
663

The problem in this is that floatingips before kops 1.19 was created without `Description` field. Now with kops 1.19 it tries to search floatingips using that Description field. Because it does not find floatingip with correct description, it creates new floatingip and tries to attach that to existing bastion. However, the bastion do have old floatingip and new cannot attached.

Workaround: detach existing floatingip from bastion (and delete that its not needed anymore). The kops update cluster will create new floatingip and attach that to bastion, everything will work after that.

/kind bug

/cc @olemarkus

olemarkus:
Set description field on existing floating ips

fixes #9957 

/cc @zetaab

### DIFF
--- a/upup/pkg/fi/cloudup/openstacktasks/floatingip.go
+++ b/upup/pkg/fi/cloudup/openstacktasks/floatingip.go
@@ -18,7 +18,10 @@ package openstacktasks
 
 import (
 	"fmt"
+	"strings"
 	"time"
+
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 
 	l3floatingip "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/openstacktasks/floatingip.go
```diff
@@ -131,34 +134,29 @@ func (e *FloatingIP) Find(c *fi.Context) (*FloatingIP, error) {
 	}
 	cloud := c.Cloud.(openstack.OpenstackCloud)
 	if e.LB != nil && e.LB.PortID != nil {
-		// Layer 3
-		fips, err := cloud.ListL3FloatingIPs(l3floatingip.ListOpts{
-			PortID: fi.StringValue(e.LB.PortID),
-		})
+		fip, err := findFipByPortID(cloud, fi.StringValue(e.LB.PortID))
```
zetaab, line 137: I do not see how this is going to fix anything to instances which have floatingip attached? This applies only to loadbalancers

zetaab (changes requested):

upup/pkg/fi/cloudup/openstacktasks/floatingip.go
```diff
@@ -171,11 +169,59 @@ func (e *FloatingIP) Find(c *fi.Context) (*FloatingIP, error) {
 			e.IP = actual.IP
 			return actual, nil
 		}
+
 	}
 
+	if len(fips) == 0 {
+		//If we fail to find an IP address we need to look for IP addresses attached to a port with similar name
+		//TODO: remove this in kops 1.21 where we can expect that the description field has been added
```
zetaab, line 177: But what will add it? does user manually add that? or should our code do that?

olemarkus: The PR was missing the change to the RenderOpenstack function. Fixed that now. Hope that makes it more clear.

zetaab (approved):
/approve
/lgtm

I tested this and it works like should, we just need remember to remove those part of the code in 1.21

PR #9967, opened 2020-09-18
Cilium hubble pointer

### LINKED ISSUE #9966: Inaccurate instructions for enabling Cilium
 this information.**
master

 a `kops` flag.**
1.19.1

AWS

```
kops create cluster --name foo.k8s.local --node-count 1 --master-count 1 --zones us-west-2a --cloud aws
kops edit cluster --name foo.k8s.local # see details below
```
`edit cluster` failed
`edit cluster` to succeed

```yaml
kind: Cluster
metadata:
  creationTimestamp: "2020-09-17T20:04:28Z"
  generation: 2
  name: foo.k8s.local
spec:
  api:
    dns: {}
  authentication:
    aws:
      backendMode: CRD
  authorization:
    rbac: {}
  channel: stable
  cloudProvider: aws
  configBase: s3://tune-kops-state-us-west-2/foo.k8s.local
  containerRuntime: docker
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-us-west-2a
      name: a
    memoryRequest: 100Mi
    name: main
  - cpuRequest: 100m
    etcdMembers:
    - instanceGroup: master-us-west-2a
      name: a
    memoryRequest: 100Mi
    name: events
  iam:
    allowContainerRegistry: true
    legacy: false
  kubelet:
    anonymousAuth: false
  kubernetesApiAccess:
  - 67.4.208.203/32
  kubernetesVersion: 1.19.0-rc.1
  masterInternalName: api.internal.foo.k8s.local
  masterPublicName: api.foo.k8s.local
  networkCIDR: 172.20.0.0/16
  networking:
    cilium:
      hubble: {}
  sshAccess:
  - 67.4.208.203/32
  subnets:
  - cidr: 172.20.32.0/19
    name: us-west-2a
    type: Public
    zone: us-west-2a
  topology:
    dns:
      type: Public
    masters: public
    nodes: public
```

The [docs](https://kops.sigs.k8s.io/networking/cilium/#installing-cilium-on-a-new-cluster) mention that enabling cilium just requires:

```yaml
spec:
  networking:
    cilium: {}
```

but the following error is reported when using this cluster spec:
```
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
# Found fields that are not recognized
# ...
#     networkCIDR: 172.20.0.0/16
#     networking:
# -     cilium:
# +     cilium: {}
# -       hubble: {}
#     nonMasqueradeCIDR: 100.64.0.0/10
#     sshAccess:
# ...
```

It seems like we should either make the hubble field optional by making it a pointer, or update the docs to include `hubble: {}`

olemarkus:
Fixes #9966 

/cc @hakman @rifelpet

### DIFF
--- a/pkg/apis/kops/networking.go
+++ b/pkg/apis/kops/networking.go
@@ -441,7 +441,7 @@ type CiliumNetworkingSpec struct {
 	// Default: false
 	EnableRemoteNodeIdentity *bool `json:"enableRemoteNodeIdentity,omitempty"`
 	// Hubble configures the Hubble service on the Cilium agent.
-	Hubble HubbleSpec `json:"hubble,omitempty"`
+	Hubble *HubbleSpec `json:"hubble,omitempty"`
 
 	// RemoveCbrBridge is not implemented and may be removed in the future.
 	// Setting this has no effect.
--- a/pkg/apis/kops/v1alpha2/networking.go
+++ b/pkg/apis/kops/v1alpha2/networking.go
@@ -439,7 +439,7 @@ type CiliumNetworkingSpec struct {
 	// Default: false
 	EnableRemoteNodeIdentity *bool `json:"enableRemoteNodeIdentity,omitempty"`
 	// Hubble configures the Hubble service on the Cilium agent.
-	Hubble HubbleSpec `json:"hubble,omitempty"`
+	Hubble *HubbleSpec `json:"hubble,omitempty"`
 
 	// RemoveCbrBridge is not implemented and may be removed in the future.
 	// Setting this has no effect.
--- a/pkg/apis/kops/v1alpha2/zz_generated.conversion.go
+++ b/pkg/apis/kops/v1alpha2/zz_generated.conversion.go
@@ -1474,8 +1474,14 @@ func autoConvert_v1alpha2_CiliumNetworkingSpec_To_kops_CiliumNetworkingSpec(in *
 	out.EnableNodePort = in.EnableNodePort
 	out.EtcdManaged = in.EtcdManaged
 	out.EnableRemoteNodeIdentity = in.EnableRemoteNodeIdentity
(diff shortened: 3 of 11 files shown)


hakman:
Would it be a better idea to flatten the struct? Seems that the Hubble thing is the only one.
Something like: https://github.com/hakman/kops/commit/490da2046650637f40f5ebd186f6a07b5c67eb6c

olemarkus:
> Would it be a better idea to flatten the struct? Seems that the Hubble thing is the only one.
> Something like: [hakman@490da20](https://github.com/hakman/kops/commit/490da2046650637f40f5ebd186f6a07b5c67eb6c)

You would need both HubbleEnable and HubbleEnableMetrics.
The primary use case for hubble is to attach hubble-ui or similar to it and view things like service graphs and such. EnableMetrics just enables some prometheus metrics in addition like number of drops per service etc.

hakman:
HubbleMetrics is an array, enable/disable can be based on number of items. In any case, was just a suggestion. Structs seems to complicate templates.

olemarkus:
So HubbleMetrics already determines if _prometheus_ metrics should be enabled or not by the number of items. But one may want to enable hubble without any prometheus metrics enabled.

hakman:
OK, I understand. It was not meant as an alternate PR, just a suggestion.
The point was not to complicate the code unnecessarily with extra structs and pointer validations. I don't mind either way.

olemarkus:
Yep. Thanks for the suggestion :) 
As long as one uses options builders, one can get away with quite a bit of pointer checks. More annoying that any non-nil bool pointer is evaluated as true by sprig ...

PR #9973, opened 2020-09-19
Build cloud outside of PerformAssignments

olemarkus:
We tend to build cloud, call some method, and then build cloud over
again. It would be easier to just pass the first one along.

Passing along cloud would also make it easier to mock cloud.


upup/pkg/fi/cloudup/defaults.go
```diff
@@ -21,6 +21,8 @@ import (
 	"net"
 	"strings"
 
+	"k8s.io/kops/upup/pkg/fi"
+
```
hakman, line 25: Should be moved a the the `k8s.io/kops` group.

PR #9979, opened 2020-09-22
Add missing permissions for cilium-operator

codablock:
These became required a few cilium versions ago.

This basically fixes the same as reported in kubespray: https://github.com/kubernetes-sigs/kubespray/issues/6681, which was fixed in https://github.com/kubernetes-sigs/kubespray/pull/6683

I encountered exactly the same error with kops and cilium. A follow up error is that the cilium agents fail and keep restarting due to the cilium-operator not starting up and not issuing heart beats.

EDIT: I'm not sure I understand the folder structure in `upup/models/cloudup/resources/addons/networking.cilium.io`. I already figured out that I have to run a local build first to update bindata.go...but I don't understand which of the 3 templates needs the changes.

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -4735,6 +4735,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - create
+  - get
+  - update
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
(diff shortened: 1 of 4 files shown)


olemarkus:
/ok-to-test


I think the current default cilium version doesn't use 1.8.3 yet, which is why this hasn't been added. But thanks for catching this anyway.

codablock:
@olemarkus I'm using kops v1.19.0-alpha.4 and from what I can see, it has deployed cilium 1.8.3 to my cluster. I assume https://github.com/kubernetes/kops/pull/9871 is the reason.

olemarkus:
Then I missed that part. Thanks!

PR #9991, opened 2020-09-25
Clarified S3 state store encryption default

### LINKED ISSUE #9978: aws getting started docs no client side encryption
The getting started docs for the AWS cloud provider seem to suggest that kops uses client side encryption when it's unable to verify encryption configuration of the state store S3 bucket. Instead, I see from the source code that kops asks S3 to use the default AES256 server side encryption.

CrossRoast:
Fixes #9978


CrossRoast:
Oops, a different email was registered on my account. Is that an issue?

rifelpet:
/lgtm
/approve

Thanks for updating this!

PR #9995, opened 2020-09-26
Fix small typo in create cluster help output

erismaster:
Fixes typo in cmd/kops/create_cluster.go and
docs/cli/kops_create_cluster.md where example output had filename.yamlh,
changed to filename.yaml

Resolves https://github.com/kubernetes/kops/issues/9994


rifelpet:
thanks for the fix!
/lgtm
/approve

PR #9998, opened 2020-09-28
Add label to prevent kops-controller from running on old nodes

### LINKED ISSUE #9953: not possible to rolling update cluster
 this information.**

1.19 alpha 4

 a `kops` flag.**

1.19.1

openstack / aws

I am updating clusters from 1.17 / 1.18 -> 1.19.1

kops update cluster --yes && kops rolling-update --yes

Bastion is rotated fine. However, after the bastion is rotated something is executing new manifests for kops-controller. This will lead to following situation:

```
kops-controller-68fc4                                        1/1     Running             2          20d
kops-controller-9hq4s                                        1/1     Running             0          15d
kops-controller-b7jrx                                        0/1     ContainerCreating   0          20m
```

```
% kubectl describe pod kops-controller-b7jrx

  Warning  FailedMount  14s (x18 over 20m)   kubelet, master-zone-1-1-1-jannem-k8s-local  MountVolume.SetUp failed for volume "kops-controller-pki" : hostPath type check failed: /etc/kubernetes/kops-controller/ is not a directory
```

So the new kops-controller manifest is NOT backwards compatible and this means that only way to update kops cluster masters currently is roll ALL masters at once (or skip cluster validation or modify kops controller manifest and be fast). The folder does not exist (yet) in old masters installed by older kops version. This folder is available only in newest kops version. So it should not update the manifest before the folder really exists.

I expect that rolling update should work usually

johngmyers:
Adds a label and nodeSelector to prevent kops-controller from running on nodes that were created with an older version of kops (and thus don't have the necessary PKI files provisioned by nodeup).

Fixes #9953 

Upgrading a 3-node control plane from kops 1.18.1 appears to be working.

### DIFF
--- a/protokube/pkg/protokube/labeler.go
+++ b/protokube/pkg/protokube/labeler.go
@@ -40,7 +40,8 @@ func bootstrapMasterNodeLabels(ctx context.Context, kubeContext *KubernetesConte
 	}
 
 	labels := map[string]string{
-		"node-role.kubernetes.io/master": "",
+		"node-role.kubernetes.io/master":  "",
+		"kops.k8s.io/kops-controller-pki": "",
 	}
 
 	shouldPatch := false
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -2751,6 +2751,7 @@ spec:
         operator: Exists
       nodeSelector:
         node-role.kubernetes.io/master: ""
+        kops.k8s.io/kops-controller-pki: ""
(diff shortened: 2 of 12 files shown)


zetaab:
nodeselector is added to kops-controller fine. However, when I rolling update nodes in openstack the new nodes will not get needed labels -> kops-controller is not provisioned at all

johngmyers:
The label should be applied by protokube. Is the process different for OpenStack?

PR #10005, opened 2020-09-30
Add missing closing inline code character

mbacchi:


rifelpet:
Thanks for catching this!
/lgtm
/approve

PR #10011, opened 2020-10-01
Spotinst: Upgrade the Spot Cluster Controller to version 1.0.67

liranp:
This PR upgrades the Spot Cluster Controller to version 1.0.67 and stores the user credentials in a `Secret` instead of a `ConfigMap`.

### DIFF
--- a/upup/pkg/fi/cloudup/template_functions.go
+++ b/upup/pkg/fi/cloudup/template_functions.go
@@ -28,6 +28,7 @@ When defining a new function:
 package cloudup
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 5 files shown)


upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -139,6 +140,8 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		if creds, err := spotinst.LoadCredentials(); err == nil {
 			dest["SpotinstToken"] = func() string { return creds.Token }
 			dest["SpotinstAccount"] = func() string { return creds.Account }
```
hakman, line 142: I don't think these are used anywhere anymore.
Also, not quite sure how `SPOTINST_ACCOUNT_LEGACY` and `SPOTINST_TOKEN_LEGACY` are populated.

liranp: Both are used [in the older manifest](https://github.com/kubernetes/kops/blob/15cc0fefae658bffa0dca7216d054918d70ebb8e/upup/models/cloudup/resources/addons/spotinst-kubernetes-cluster-controller.addons.k8s.io/v1.9.0.yaml.template#L10-L11) that points to an older version of the controller. I've kept the legacy variables just to maintain backward compatibility with the `ConfigMap`.

hakman: Got it. Thanks!

liranp:
@hakman Thank you for reviewing this!

Issue #10014, opened 2020-10-03
Add Calico implementation for disabling AWS source/destination checks


hakman:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

Calico 3.16 adds a new option to disable source/destination checks: https://github.com/projectcalico/felix/pull/2381.
https://docs.projectcalico.org/reference/resources/felixconfig#spec

This can replace the current implementation that uses https://github.com/ottoyiu/k8s-ec2-srcdst.

**2. Feel free to provide a design supporting your feature request.**

At the moment, this is enabled automatically when `spec.networking.calico.crossSubnet: true`.
https://kops.sigs.k8s.io/networking/calico/#enable-cross-subnet-mode-in-calico-aws-only

The manifest flag that can be used to enable/disable this feature in Calico is `FELIX_AWSSRCDSTCHECK`. It should be added as a separate option, but also keep the previous setting for backwards compatibility.

IAM Policies may need updating also to allow this to be set from any node.

monicagangwar:
@hakman I'd like to give this a try, can you please assign this issue to me ?

hakman:
Thanks for giving this a try @monicagangwar ! 😄 
/assign @monicagangwar

monicagangwar:
@hakman a question here, if the option to provide the manifest flag `FELIX_AWSSRCDSTCHECK` is added along with the existing option `spec.networking.calico.crossSubnet` 
what should be the behaviour if the options are conflicting. eg: FELIX_AWSSRCDSTCHECK is set to `Disable` or `DoNothing` but `spec.networking.calico.crossSubnet` is set to true ?

hakman:
I think the behaviour should be to enable the deployment for `k8s-ec2-srcdst`, but set the number of replicas to 0.
Unfortunately, kops cannot remove `k8s-ec2-srcdst` once it's deployed, so best to just to the above and document that if one wants to switch to `AwsSrcDstCheck`, should delete the `k8s-ec2-srcdst` deployment manually.

monicagangwar:
@hakman correct me if i am wrong but why do we need to deploy k8s-ec2-srcdst altogether ? if `spec.networking.calico.crossSubnet` is set to true we can simply add the flag `FELIX_AWSSRCDSTCHECK` to `Disable` source destination checks, right ?

hakman:
[...]

PR #10019, opened 2020-10-06
[calico] awsSrcDstCheck to disable src/dest checks in AWS

monicagangwar:
using calico config `awsSrcDstCheck` and flag `FELIX_AWSSRCDSTCHECK` instead of [k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst).  

for backward compatibility where in [k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst) resources are already deployed following the below mentioned rules

crossSubnet | awsSrcDestCheck | Result
-- | -- | --
false | Enabled/Disabled/DoNothing | k8s-ec2-srcdst deployment doesn't exist
true | Enabled/Disabled/DoNothing | k8s-ec2-srcdst deployment has 0 replicas, to effectively disable it

Fixes: #10014

### DIFF
--- a/docs/networking/calico.md
+++ b/docs/networking/calico.md
@@ -55,11 +55,18 @@ To enable this mode in a cluster, add the following to the cluster spec:
     calico:
       crossSubnet: true
 ```
-
 In the case of AWS, EC2 instances have source/destination checks enabled by default.
-When you enable cross-subnet mode in kops, an addon controller ([k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst))
+When you enable cross-subnet mode in kops 1.19+, it is equivalent to: 
+```yaml
+  networking:
+    calico:
+      awsSrcDstCheck: Disable
+      IPIPMode: CrossSubnet
+```
+An IAM policy will be added to all nodes to allow Calico to execute `ec2:DescribeInstances` and `ec2:ModifyNetworkInterfaceAttribute`, as required when [awsSrcDstCheck](https://docs.projectcalico.org/reference/resources/felixconfig#spec) is set.
+For older versions of kops, an addon controller ([k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst))
 will be deployed as a Pod (which will be scheduled on one of the masters) to facilitate the disabling of said source/destination address checks.
-Only the masters have the IAM policy (`ec2:*`) to allow k8s-ec2-srcdst to execute `ec2:ModifyInstanceAttribute`.
+Only the control plane nodes have an IAM policy to allow k8s-ec2-srcdst to execute `ec2:ModifyInstanceAttribute`.
 
 ### Configuring Calico MTU
 
--- a/pkg/apis/kops/networking.go
+++ b/pkg/apis/kops/networking.go
@@ -108,6 +108,9 @@ type CalicoNetworkingSpec struct {
 	CPURequest *resource.Quantity `json:"cpuRequest,omitempty"`
 	// CrossSubnet enables Calico's cross-subnet mode when set to true
 	CrossSubnet bool `json:"crossSubnet,omitempty"`
+	// AwsSrcDstCheck enables/disables source/destination checks (AWS only)
+	// Options: "DoNothing" (default) , "Enable" or "Disable"
+	AwsSrcDstCheck string `json:"awsSrcDstCheck,omitempty"`
 	// LogSeverityScreen lets us set the desired log level. (Default: info)
 	LogSeverityScreen string `json:"logSeverityScreen,omitempty"`
 	// MTU to be set in the cni-network-config for calico.
--- a/pkg/apis/kops/v1alpha2/networking.go
+++ b/pkg/apis/kops/v1alpha2/networking.go
@@ -108,6 +108,9 @@ type CalicoNetworkingSpec struct {
 	CPURequest *resource.Quantity `json:"cpuRequest,omitempty"`
 	// CrossSubnet enables Calico's cross-subnet mode when set to true
 	CrossSubnet bool `json:"crossSubnet,omitempty"`
+	// AwsSrcDstCheck enables/disables source/destination checks (AWS only)
+	// Options: "DoNothing" (default) , "Enable" or "Disable"
+	AwsSrcDstCheck string `json:"awsSrcDstCheck,omitempty"`
 	// LogSeverityScreen lets us set the desired log level. (Default: info)
 	LogSeverityScreen string `json:"logSeverityScreen,omitempty"`
 	// MTU to be set in the cni-network-config for calico.
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -275,6 +275,10 @@ func (r *NodeRoleMaster) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 		addCiliumEniPermissions(p, resource, b.Cluster.Spec.IAM.Legacy)
 	}
 
+	if b.Cluster.Spec.Networking != nil && b.Cluster.Spec.Networking.Calico != nil && b.Cluster.Spec.Networking.Calico.AwsSrcDstCheck != "" {
+		addCalicoSrcDstCheckPermissions(p)
+	}
+
 	return p, nil
 }
 
@@ -308,6 +312,10 @@ func (r *NodeRoleNode) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 
 	if b.Cluster.Spec.Networking != nil && b.Cluster.Spec.Networking.LyftVPC != nil {
 		addLyftVPCPermissions(p, resource, b.Cluster.Spec.IAM.Legacy, b.Cluster.GetName())
+	}
+
+	if b.Cluster.Spec.Networking != nil && b.Cluster.Spec.Networking.Calico != nil && b.Cluster.Spec.Networking.Calico.AwsSrcDstCheck != "" {
+		addCalicoSrcDstCheckPermissions(p)
 	}
 
 	return p, nil
@@ -662,6 +670,17 @@ func addECRPermissions(p *Policy) {
 			"ecr:DescribeRepositories",
 			"ecr:ListImages",
 			"ecr:BatchGetImage",
+		),
+		Resource: stringorslice.Slice([]string{"*"}),
+	})
+}
+
+func addCalicoSrcDstCheckPermissions(p *Policy) {
+	p.Statement = append(p.Statement, &Statement{
+		Effect: StatementEffectAllow,
+		Action: stringorslice.Of(
+			"ec2:DescribeInstances",
+			"ec2:ModifyNetworkInterfaceAttribute",
 		),
 		Resource: stringorslice.Slice([]string{"*"}),
 	})
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -12873,6 +12873,9 @@ spec:
             # Enable Prometheus process metrics collection
             - name: FELIX_PROMETHEUSPROCESSMETRICSENABLED
               value: "{{- or .Networking.Calico.PrometheusProcessMetricsEnabled "true" }}"
+            # Enable / Disable source/destination checks in AWS
+            - name: FELIX_AWSSRCDSTCHECK
+              value: "{{- if and (eq .CloudProvider "aws") (.Networking.Calico.CrossSubnet) -}}Disable{{- else -}} {{- or .Networking.Calico.AwsSrcDstCheck "DoNothing" -}} {{- end -}}"
           securityContext:
             privileged: true
           resources:
@@ -13042,6 +13045,7 @@ metadata:
 # pod) may not match the receiving machine's address.
 #
 # This only applies for AWS environments.
+# This is a deprecated setting, use awsSrcDstCheck instead
 ---
 
 kind: ClusterRole
@@ -13099,7 +13103,7 @@ metadata:
     k8s-app: k8s-ec2-srcdst
     role.kubernetes.io/networking: "1"
 spec:
-  replicas: 1
+  replicas: 0
   selector:
     matchLabels:
       k8s-app: k8s-ec2-srcdst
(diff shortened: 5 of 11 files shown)


monicagangwar:
@hakman can you please review

hakman:
@monicagangwar please sign the CLA as described above.

monicagangwar:
> @monicagangwar please sign the CLA as described above.

@hakman done

hakman:
Thanks @monicagangwar. To fix the failing test you need to run `make crds` and commit the result.
/ok-to-test

pkg/apis/kops/networking.go
```diff
@@ -108,6 +108,9 @@ type CalicoNetworkingSpec struct {
 	CPURequest *resource.Quantity `json:"cpuRequest,omitempty"`
 	// CrossSubnet enables Calico's cross-subnet mode when set to true
 	CrossSubnet bool `json:"crossSubnet,omitempty"`
+	// AwsSrcDstCheck enables/disables source/destination checks in aws only
```
hakman, line 111:
```suggestion
	// AwsSrcDstCheck enables/disables source/destination checks (AWS only)
```

pkg/apis/kops/v1alpha2/networking.go
```diff
@@ -108,6 +108,9 @@ type CalicoNetworkingSpec struct {
 	CPURequest *resource.Quantity `json:"cpuRequest,omitempty"`
 	// CrossSubnet enables Calico's cross-subnet mode when set to true
 	CrossSubnet bool `json:"crossSubnet,omitempty"`
+	// AwsSrcDstCheck enables/disables source/destination checks in aws only
```
hakman, line 111:
```suggestion
	// AwsSrcDstCheck enables/disables source/destination checks (AWS only)
```

pkg/model/iam/iam_builder.go
```diff
@@ -310,6 +310,10 @@ func (r *NodeRoleNode) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 		addLyftVPCPermissions(p, resource, b.Cluster.Spec.IAM.Legacy, b.Cluster.GetName())
 	}
 
+	if b.Cluster.Spec.Networking != nil && b.Cluster.Spec.Networking.Calico != nil && b.Cluster.Spec.Networking.Calico.AwsSrcDstCheck != "" {
```
hakman, line 313: Should this be done on Masters also?

hakman (changes requested):
I added some suggestions and comments that should be addressed.
Also please add an entry to the docs section:
https://github.com/kubernetes/kops/blob/master/docs/networking/calico.md#enable-cross-subnet-mode-in-calico-aws-only

pkg/model/iam/iam_builder.go
monicagangwar: yes, should be done on masters as well, adding

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -837,7 +837,7 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*chann
 			"k8s-1.7":    "2.6.12-kops.1",
 			"k8s-1.7-v3": "3.8.0-kops.2",
 			"k8s-1.12":   "3.9.6-kops.1",
-			"k8s-1.16":   "3.16.1-kops.2",
+			"k8s-1.16":   "3.16.2-kops.2",
```
hakman, line 840:
```suggestion
			"k8s-1.16":   "3.16.1-kops.3",
```

monicagangwar: @hakman can you tell me what does `-kops.x` suffix means here ? also, i've changes the version of calico images from 3.16.1 to 3.16.2

hakman: Yes, please revert the Calico version change.
`-kops.x` means the manifest version. It is used internally to determine if the manifest changed and needs to be reapplied.

monicagangwar:
@hakman all the suggested changes are done

hakman:
Thanks @monicagangwar. I will take another look in the evening, maybe come up with a few more suggestions.
As is now, it looks good from my point of view. Great work!

docs/networking/calico.md
```diff
@@ -54,13 +54,18 @@ To enable this mode in a cluster, add the following to the cluster spec:
...
 
+> This is a deprecated behavior, to disable source/destination address checks use networking.calico.awsSrcDstCheck  
+> If crossSubnet was set to true, ([k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst)) resources exists on master nodes which can be removed manually  
+
 In the case of AWS, EC2 instances have source/destination checks enabled by default.
 When you enable cross-subnet mode in kops, an addon controller ([k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst))
 will be deployed as a Pod (which will be scheduled on one of the masters) to facilitate the disabling of said source/destination address checks.
 Only the masters have the IAM policy (`ec2:*`) to allow k8s-ec2-srcdst to execute `ec2:ModifyInstanceAttribute`.
 
+
```
hakman, line 68:
````suggestion
```

In the case of AWS, EC2 instances have source/destination checks enabled by default.
When you enable cross-subnet mode in kops 1.19+, it is equivalent to: 
```yaml
  networking:
    calico:
      awsSrcDstCheck: Disable
      IPIPMode: CrossSubnet
```
An IAM policy (`ec2:*`) will be added to all nodes to allow Calico to execute `ec2:DescribeInstances` and `ec2:ModifyNetworkInterfaceAttribute`, as required when [awsSrcDstCheck](https://docs.projectcalico.org/reference/resources/felixconfig#spec) is set.

For older versions of kops, an addon controller ([k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst))
will be deployed as a Pod (which will be scheduled on one of the masters) to facilitate the disabling of said source/destination address checks.
Only the control plane nodes have the IAM policy (`ec2:*`) to allow k8s-ec2-srcdst to execute `ec2:ModifyInstanceAttribute`.
````

**Review by hakman — CHANGES_REQUESTED** [review]

File: upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template [hunk 6]
```diff
@@ -3893,6 +3893,9 @@ spec:
             # Enable Prometheus process metrics collection
             - name: FELIX_PROMETHEUSPROCESSMETRICSENABLED
               value: "{{- or .Networking.Calico.PrometheusProcessMetricsEnabled "true" }}"
+            # Enable / Disable source/destination checks in aws
+            - name: FELIX_AWSSRCDSTCHECK
+              value: "{{- or .Networking.Calico.AwsSrcDstCheck "DoNothing" -}}"
```
hakman, line 3898: This way CrossSubnet will works same as it does now, but using the new setting.
```suggestion
            # Enable / Disable source/destination checks in AWS
            - name: FELIX_AWSSRCDSTCHECK
              value: "{{- if and (eq .CloudProvider "aws") (.Networking.Calico.CrossSubnet) -}}Disable{{- else -}} {{- or .Networking.Calico.AwsSrcDstCheck "DoNothing" -}} {{- end -}}"
```

hakman (changes requested):

docs/networking/calico.md
monicagangwar: @hakman the policy getting added for awsSrcDstCheck is not `ec2:*` its `ec2:DescribeInstances` and `ec2:ModifyNetworkInterfaceAttribute`

hakman: True, I copy/pasted from below. Updated.

docs/networking/calico.md
```diff
@@ -57,10 +57,17 @@ To enable this mode in a cluster, add the following to the cluster spec:
...
 In the case of AWS, EC2 instances have source/destination checks enabled by default.
-When you enable cross-subnet mode in kops, an addon controller ([k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst))
+When you enable cross-subnet mode in kops 1.19+, it is equivalent to: 
+```yaml
+  networking:
+    calico:
+      awsSrcDstCheck: Disable
+      IPIPMode: CrossSubnet
+```
+An IAM policy (`ec2:*`) will be added to all nodes to allow Calico to execute `ec2:DescribeInstances` and `ec2:ModifyNetworkInterfaceAttribute`, as required when [awsSrcDstCheck](https://docs.projectcalico.org/reference/resources/felixconfig#spec) is set.
```
hakman, line 67:
```suggestion
An IAM policy will be added to all nodes to allow Calico to execute `ec2:DescribeInstances` and `ec2:ModifyNetworkInterfaceAttribute`, as required when [awsSrcDstCheck](https://docs.projectcalico.org/reference/resources/felixconfig#spec) is set.
```

docs/networking/calico.md
```diff
@@ -57,10 +57,17 @@ To enable this mode in a cluster, add the following to the cluster spec:
...
+For older versions of kops, an addon controller ([k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst))
 will be deployed as a Pod (which will be scheduled on one of the masters) to facilitate the disabling of said source/destination address checks.
-Only the masters have the IAM policy (`ec2:*`) to allow k8s-ec2-srcdst to execute `ec2:ModifyInstanceAttribute`.
-
+Only the control plane nodes have the IAM policy (`ec2:*`) to allow k8s-ec2-srcdst to execute `ec2:ModifyInstanceAttribute`.
```
hakman, line 70:
```suggestion
Only the control plane nodes have the IAM policy required to allow `k8s-ec2-srcdst` to execute `ec2:ModifyInstanceAttribute`.
```

hakman:
Thanks @monicagangwar. This should be it. Please squash the commits before merging.
Do you need the `hacktoberfest` tag added to this?

monicagangwar:
@hakman sure, you can add `hacktoberfest-accepted` label to this PR

monicagangwar:
@hakman squashed the commits and rebased my branch as well

hakman (approved):
Thanks for your contribution @monicagangwar. Looking forward to see more :).
/lgtm
/approve

pkg/model/iam/iam_builder.go
```diff
@@ -275,6 +275,10 @@ func (r *NodeRoleMaster) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 		addCiliumEniPermissions(p, resource, b.Cluster.Spec.IAM.Legacy)
 	}
 
+	if b.Cluster.Spec.Networking != nil && b.Cluster.Spec.Networking.Calico != nil && b.Cluster.Spec.Networking.Calico.AwsSrcDstCheck != "" {
+		addCalicoSrcDstCheckPermissions(p)
```
seh, line 279: Why do we add the permission regardless of the value of the "AwsSrcDstCheck" field? If it's "DoNothing," we wouldn't need this permission, right?

pkg/model/iam/iam_builder.go
seh: Likewise here, shouldn't we check to see if the "AwsSrcDstCheck" field is not "DoNothing?"

upup/models/bindata.go
```diff
@@ -12873,6 +12873,9 @@ spec:
             # Enable Prometheus process metrics collection
             - name: FELIX_PROMETHEUSPROCESSMETRICSENABLED
               value: "{{- or .Networking.Calico.PrometheusProcessMetricsEnabled "true" }}"
+            # Enable / Disable source/destination checks in AWS
+            - name: FELIX_AWSSRCDSTCHECK
+              value: "{{- if and (eq .CloudProvider "aws") (.Networking.Calico.CrossSubnet) -}}Disable{{- else -}} {{- or .Networking.Calico.AwsSrcDstCheck "DoNothing" -}} {{- end -}}"
```
seh, line 12878: Here we tell Calico to take control of the ENI source/destination checks if "CrossSubnet" is in effect, but we don't check whether "AwsSrcDstCheck" was "DoNothing." If the user sets "CrossSubnet" but fails to set "AwsSrcDstCheck," then Calico attempts to reconfigure the ENI, but its governing IAM role lacks permission to do so.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -3893,6 +3893,9 @@ spec:
             # Enable Prometheus process metrics collection
             - name: FELIX_PROMETHEUSPROCESSMETRICSENABLED
               value: "{{- or .Networking.Calico.PrometheusProcessMetricsEnabled "true" }}"
+            # Enable / Disable source/destination checks in AWS
+            - name: FELIX_AWSSRCDSTCHECK
+              value: "{{- if and (eq .CloudProvider "aws") (.Networking.Calico.CrossSubnet) -}}Disable{{- else -}} {{- or .Networking.Calico.AwsSrcDstCheck "DoNothing" -}} {{- end -}}"
```
seh, line 3898: Likewise, it looks like we should take the value of "AwsSrcDstCheck into account here.

pkg/model/iam/iam_builder.go
seh: Reading the documentation in file _calico.md_, it suggests that if "CrossSubnet" is true, then we behave as if "AwsSrcDstCheck" is set to "Disable," even if the latter field is not set at all. However, here, we're only adding this permission of "AwsSrcDstCheck" is nonempty. Did you intend to set the "AwsSrcDstCheck" field earlier? Is the documentation making the wrong promise, such that we do intend to force the user to set the "AwsSrcDstCheck" field?

docs/networking/calico.md
```diff
@@ -55,11 +55,18 @@ To enable this mode in a cluster, add the following to the cluster spec:
...
       crossSubnet: true
 ```
-
 In the case of AWS, EC2 instances have source/destination checks enabled by default.
-When you enable cross-subnet mode in kops, an addon controller ([k8s-ec2-srcdst](https://github.com/ottoyiu/k8s-ec2-srcdst))
+When you enable cross-subnet mode in kops 1.19+, it is equivalent to: 
+```yaml
+  networking:
+    calico:
+      awsSrcDstCheck: Disable
```
seh, line 63: But what if the user set "crossSubnet" to true, but hasn't set this field at all? Should we behave as if it had been set to "Disable?"

docs/networking/calico.md
```diff
@@ -55,11 +55,18 @@ To enable this mode in a cluster, add the following to the cluster spec:
...
+      IPIPMode: CrossSubnet
```
DOboznyi, line 64: It should be ipipMode

PR #10022, opened 2020-10-07
Kubelet serving certificate and metrics server addon

### LINKED ISSUE #6879: [METRICS-SERVER] New kubelet-certificate-authority flag available (v0.3.2)
Don't use `kubelet-insecure-tls` flag by default in https://github.com/kubernetes/kops/blob/master/addons/metrics-server/v1.8.x.yaml because of security reasons.

Maybe using the new **--kubelet-certificate-authority** flag this could be done.

This should be changed here https://github.com/kubernetes/kops/blob/master/addons/metrics-server/v1.8.x.yaml

olemarkus:
fixes #6879 

This PR adds metrics-server as a configurable addon deprecating the manifest-based one.
More importantly, it uses kops-controller to provision the kubelet _serving_ certificate and makes kubelet use it. This lets us avoid using the `--kubelet-insecure-tls` flag.


upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -563,6 +563,27 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*chann
 		}
 	}
 
+	if b.Cluster.Spec.MetricsServer != nil && fi.BoolValue(b.Cluster.Spec.MetricsServer.Enabled) {
+		{
+			key := "metrics-server.addons.k8s.io"
+			version := "1.19.0"
```
hakman, line 569: Maybe meant this?
```suggestion
			version := "0.3.7"
```

addons/metrics-server/README.md
```diff
@@ -1,5 +1,7 @@
 # Kubernetes Metrics Server
 
+**This addon is deprecated. Set `spec.metricsServer.enabled: true` instead**
```
agilgur5, line 3: Wouldn't it be good to also update the manifest's line anyway? https://github.com/kubernetes/kops/blob/174ba0e32338c828f1706e6845cf7da1c14d8707/addons/metrics-server/v1.16.x.yaml#L144 

Might need to add a new manifest though for a new version, not sure if this is getting backported

I would think that may make it easier for gradual transitions or just understanding that there's a difference between the two variants -- an important security one at that. Otherwise may make it seem like they're otherwise equivalent. I was using the `metrics-server` Helm Chart instead of the add-on as well, but looking at the add-on made me realize they both shared a security risk.

agilgur5 (commented):
Thanks for adding this @olemarkus ! Since `metrics-server` is basically a core component necessary for HPA / VPA (and with little in the form of competition), this resolves a big security risk in getting it set-up, which, as you mentioned in https://github.com/kubernetes/kops/issues/6879#issuecomment-705162948, may not be truly possible to resolve without this change.

I just had two questions and two comments -- don't remotely have the depth to actually review this. Apologies if these are an inconvenience to answer given my lack of depth!

1. Am I correct in understanding that this makes it so each kubelet uses the `kops` CA to provision its serving cert during node bootstrap (instead of self-signing as before)?
I'm curious if there are more potential use-cases for this type of workflow for other core services that require certs. E.g. in one of my comments I mention the use-case of `metrics-server`'s own cert.

2. Based on https://github.com/kubernetes/kops/issues/6879#issuecomment-705195803, am I correct in understanding that the configurable add-on change is not necessary to resolve this problem? Like, based on your comment, I got the impression that the manifest or any other type of `metrics-server` deploy (I'm using the Helm Chart) should no longer need `--kublet-insecure-tls` with the cert piece alone, right?
Which would mean the configurable version just simplifies deploy, but does not resolve the security risk in isolation

cmd/kops-controller/pkg/server/server.go
```diff
@@ -175,6 +175,11 @@ func (s *Server) issueCert(name string, pubKey string, id *fi.VerifyResult, vali
 			CommonName:   fmt.Sprintf("system:node:%s", id.NodeName),
 			Organization: []string{rbac.NodesGroup},
 		}
+	case "kubelet-server":
+		issueReq.Subject = pkix.Name{
+			CommonName: id.NodeName,
+		}
```
johngmyers, line 181:
```suggestion
		}
		issueReq.AlternateNames = []string{id.NodeName}
```

Accepting hostnames in the CommonName is deprecated.

nodeup/pkg/model/kubelet.go
```diff
@@ -226,6 +232,11 @@ func (b *KubeletBuilder) buildSystemdEnvironmentFile(kubeletConfig *kops.Kubelet
 		}
 	}
 
+	if b.UseKopsControllerForNodeBootstrap() {
+		flags += " --tls-cert-file " + b.PathSrvKubernetes() + "/kubelet-server.crt"
+		flags += " --tls-private-key-file " + b.PathSrvKubernetes() + "/kubelet-server.key"
```
johngmyers, line 237:
```suggestion
		flags += " --tls-cert-file=" + b.PathSrvKubernetes() + "/kubelet-server.crt"
		flags += " --tls-private-key-file=" + b.PathSrvKubernetes() + "/kubelet-server.key"
```

nodeup/pkg/model/kubelet.go
```diff
@@ -538,3 +549,49 @@ func (b *KubeletBuilder) buildMasterKubeletKubeconfig(c *fi.ModelBuilderContext)
...
+			})
+
+		} else {
+			issueCert := &nodetasks.IssueCert{
+				Name:   name,
+				Signer: signer,
+				Type:   "server",
+				Subject: nodetasks.PKIXName{
+					CommonName: nodeName,
+				},
```
johngmyers, line 589:
```suggestion
				},
				AlternateNames: []string{id.NodeName},

```

johngmyers (changes requested):

johngmyers:
I believe the 1.19 manifest will still need to set `--kubelet-insecure-tls` on cloud providers that don't support bootstrap using kops-controller.

There's still `insecureSkipTLSVerify: true` on the APIService, but I suppose that's a different issue.

upup/models/bindata.go
```diff
@@ -3838,6 +3840,365 @@ func cloudupResourcesAddonsMetadataProxyAddonsK8sIoV0112Yaml() (*asset, error) {
...
+kind: APIService
+metadata:
+  name: v1beta1.metrics.k8s.io
+spec:
+  service:
+    name: metrics-server
+    namespace: kube-system
+  group: metrics.k8s.io
+  version: v1beta1
+  insecureSkipTLSVerify: true
```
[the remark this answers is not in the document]
olemarkus: Providing a kops-signed certificate to metrics-server is a whole different story.
With kubelet and other system components, what we really validate is the nodes themselves. We don't have any way of validating pods running in the node.
One way of solving this could be that kops-controller knows that we are deploying metrics-server and provisions secrets for this, but that would be a much larger change.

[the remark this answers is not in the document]
agilgur5: > One way of solving this could be that kops-controller knows that we are deploying metrics-server and provisions secrets for this, but that would be a much larger change.

Yea that's exactly what I was thinking. Thought it wouldn't be much larger than the current change, but I don't have the depth to know what else that would involve

addons/metrics-server/README.md
olemarkus: This is a new feature, so it won't be backported. Adding a new manifest would also only be for 1.19+, so that kind of defeats the purpose of the deprecation notice.

But what I can do is add a note to the release notes. Should have done so anyway.

olemarkus:
> I believe the 1.19 manifest will still need to set `--kubelet-insecure-tls` on cloud providers that don't support bootstrap using kops-controller.
> 

That is true. Using UseKopsControllerForNodeBootstrap instead of k8s version now. Cleaner than duplicating the whole manifest anyway.

> There's still `insecureSkipTLSVerify: true` on the APIService, but I suppose that's a different issue.

Yep. Need a different mechanism for that. Maybe kops-controller could create TLS secrets for known addons. It may be more in the cluster-addons space too.

addons/metrics-server/README.md
agilgur5: Gotcha, makes sense. I was looking back into this due to some security settings (`kube-apiserver` not setting `--kubelet-certificate-authority`) and the release notes for #10836 , and it seems like I forgot to ever respond here. 

The [release](https://github.com/kubernetes/kops/releases/tag/v1.19.0) note on `metrics-server` doesn't say that the new add-on is more secure -- which is a really key piece. The wording makes it sound like the add-on is just done via the new approach now, so other than deprecation it doesn't motivate one to move to the new method as the security piece isn't explicitly mentioned. I was previously using `metrics-server` directly via its Helm chart, so I also wouldn't know that I could now remove `--kubelet-insecure-tls` from my config.
I think it would be useful to mention that it is more secure than the now deprecated variant and that users of direct `metrics-server` should remove the `--kubelet-insecure-tls` flag when upgrading to kOps 1.19+

olemarkus: Good suggestions! Thanks! Would you be able to add a PR for this?

agilgur5: It only took me 2 years to finally get the PR out (sorry -- but I did get several other PRs I had in my backlog out as well at least!) but finally made it: https://github.com/kubernetes/kops/pull/15327

Issue #9598, opened 2020-07-19
Improve E2E test coverage


rifelpet:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
Kops has many periodic prow jobs that run the kubernetes e2e suite against various combinations of k8s versions, OS, CNI, cloud provider, and cpu architecture.

However there is a lot of kops functionality not covered by e2e tests:

- Upgrading & downgrading Kops
- Upgrading & downgrading Kubernetes / etcd / CNI providers
- Applying & upgrading & downgrading all of the miscelaneous "bootstrapchannelbuilder" manifests
- Terraform & Cloudformation targets
- Rolling Updates
- Almost all feature flags
- Many API fields that aren't settable via `kops create cluster`, requiring a manifest file and `kops replace -f`

Adding test coverage for these scenarios would require significant changes to kubetest (our e2e runner) which is in [maintenance mode](https://github.com/kubernetes/test-infra/tree/master/kubetest#deprecation-notice) so I'm proposing that we reevaluate how we run our e2e jobs. https://github.com/kubernetes/kops/pull/8605 was open as an experiment but I think we should settle on a design proposal before we merge anything.
 

The kubetest logic that we currently use involves:

1. download or build the appropriate kops version
2. download and extract the appropriate k8s version
3. choose a region
4. run the `kops create cluster` with the appropriate flags and wait for validation to pass
5. download and run the appropriate k/k e2e test suite
6. dump node logs and troubleshooting info
7. tear down the cluster
8. publish the kops "version marker" if applicable

Of the above steps, less than half are also used by other projects. In order to cover the areas mentioned above we would need to add a significant amount of kops-specific testing logic. I'm wondering if it makes more sense to have our test runner code live in the kops repo rather than test-infra. We could consider vendoring the necessary kubetest (or kubetest2) logic for some of the above steps, but keeping most of the development in the kops repo allows us to iterate faster by not requiring approvals from outside our OWNERS. It also reduces potential impact or compromises necessary for a "shared" tool like kubetest2.

**2. Feel free to provide a design supporting your feature request.**

Some concerns:
* Do we continue to use go for consistency? It would allow us to vendor certain logic from test-infra. If we choose a different language or technology like Jupyter it would involve additional effort to get working in prow jobs.
* Does each job compile the runner code each time or do we publish a runner artifact/image similar to our existing kops version markers?

Anyone have any other thoughts or suggestions?

johngmyers:
[...]

PR #10031, opened 2020-10-08
Initial kubetest2 structure for e2e testing 

rifelpet:
ref: #9598
This creates a new go module for the e2e code and the kubetest2 skeleton.
Most of the kubetest2 code was copied from [sigs.k8s.io/kubetest2/kubetest2-gke](https://github.com/kubernetes-sigs/kubetest2/tree/master/kubetest2-gke/deployer).

Currently only building (`make gcs-publish-ci`) is in place.
I used [test-infra/scenarios/kubernetes_e2e.py](https://github.com/kubernetes/test-infra/blob/b63443db54d6877eaaaa7410d66f47178b2fadd2/scenarios/kubernetes_e2e.py#L255-L267) as reference, removing environment and make variables that are no longer needed.

Notes:
* This is a new go module so that anyone importing kops does not end up with transitive dependencies used only by e2e testing. (`gopls` users may need to [update to 0.5.1](https://github.com/golang/tools/releases/tag/gopls%2Fv0.5.1) for multi-module workspace support)
* I elected not to vendor the dependencies as suggested by Justin
* I envision kops-specific e2e tests could live in `tests/e2e/` as well as Cluster and IG template files from which we can launch clusters. (we'll need to templatize them to adjust the cluster name, region and other fields which are specific to each job run)
* For presubmits we'll compile kubetest2-kops with the same commit being tested, and install kubetest2 with the version defined in go.mod. Because periodics don't clone the kops repo we'll need a way to use a specific kubetest2-kops binary. A few ideas for this:
  * Build and publish the kubetest2 and kubetest2-kops binaries on postsubmit along with the other kops artifacts. This may get into a chicken/egg problem - periodics would need to pull the kops artifacts down before running any kubetest2 commands (I was originally assuming kubetest2-kops would download the artifacts)
  * Build and promote an image that we'll use for prow jobs - a postsubmit job will build & push the image and we'll manually bump the image tags in the test-infra job definitions whenever we want periodics to use a new version of kubetest2-kops.
  * Have the periodic jobs clone the kops repo using [extra_refs](https://github.com/kubernetes/test-infra/blob/0e1693104626508f41155d9aea9e69c27e68fdff/prow/pod-utilities.md#writing-a-prowjob-that-uses-pod-utilities) and build kubetest2 and kubetest2-kops the same way that presubmits would.
  
  We'll need to keep in mind release branching and which versions of kubetest2-kops we use to test which release branches of kops.

Next Steps:
* Get our CI / verify jobs to play nicely with the new go module (i was envisioning we could eventually move tools.go into its own go module too)
* Setup an optional presubmit prow job that performs the instructions below
* Implement Up() and Down() for AWS with just `kops create cluster` and `kops delete cluster`. This will require determining all the environment variables and file paths that are in place for the prow jobs, including setting up the boskos client.

Instructions:
```
cd tests/e2e
go install sigs.k8s.io/kubetest2
go install ./kubetest2-kops
# runs `make gcs-publish-ci` and exits
kubetest2 kops --build --stage-location=gs://foobar/ --kops-root=../../
```

Help text:
```
kubetest2 kops --help
Usage:
  kubetest2 kops [Flags] [DeployerFlags] -- [TesterArgs]

Flags:
      --artifacts string   top-level directory to put artifacts under for each kubetest2 run, defaulting to "${ARTIFACTS:-./_artifacts}". If using the ginkgo tester, this must be an absolute path. (default "src/k8s.io/kops/tests/e2e/_artifacts")
      --build              build kubernetes
      --down               tear down the test cluster
  -h, --help               display help
      --run-id string      unique identifier for a kubetest2 run (default "78e3e602-c079-403f-a0cd-8319b7993cc2")
      --test string        test type to run, if unset no tests will run
      --up                 provision the test cluster

DeployerFlags(kops):
      --add_dir_header                   If true, adds the file directory to the header of the log messages
      --alsologtostderr                  log to standard error as well as files
      --kops-root string                 Path to root of the kops repo. Used with --build.
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --log_file string                  If non-empty, use this log file
      --log_file_max_size uint           Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --logtostderr                      log to standard error instead of files (default true)
      --skip_headers                     If true, avoid header prefixes in the log messages
      --skip_log_headers                 If true, avoid headers when opening log files
      --stage-location string            Storage location for kops artifacts. Only gs:// paths are supported.
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
  -v, --v Level                          number for the log level verbosity
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
```


hakman:
I am interested also :)
/cc @hakman

olemarkus:
This is a great start!

I wonder if we should think of doing update tests too. Spin up a cluste with spec A, apply spec B, run some e2e. To test that e.g ELB -> NLB migration ends up in a working cluster. This test suite probably don't have to run that often, maybe just pre-release.

tests/e2e/kubetest2-kops/deployer/common.go
```diff
@@ -0,0 +1,35 @@
...
+	var err error
+	d.doInit.Do(func() { err = d.initialize() })
+	return err
+}
+
+// initialize should only be called by init(), behind a sync.Once
+func (d *deployer) initialize() error {
+	if d.commonOptions.ShouldBuild() {
+		if err := d.verifyBuildFlags(); err != nil {
+			return fmt.Errorf("init failed to check build flags: %s", err)
```
justinsb, line 31: %w or %v I think

justinsb:
A great start - let's merge and iterate!

/approve
/lgtm

PR #10032, opened 2020-10-09
Open etcd port only when Calico uses "etcd" datastore

hakman:
For _kubernetes_ 1.12+, _kops_ hardcodes `DATASTORE_TYPE` to "kubernetes".
https://github.com/kubernetes/kops/blob/5cc1b5ad8ef074b86a61db80a8b24aca1f6dbb53/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.12.yaml.template#L36
https://github.com/kubernetes/kops/blob/5cc1b5ad8ef074b86a61db80a8b24aca1f6dbb53/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.12.yaml.template#L763-L764

### DIFF
--- a/nodeup/pkg/model/networking/calico.go
+++ b/nodeup/pkg/model/networking/calico.go
@@ -39,7 +39,7 @@ func (b *CalicoBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	// @check if tls is enabled and if so, we need to download the client certificates
-	if !b.UseEtcdManager() && b.UseEtcdTLS() {
+	if b.IsKubernetesLT("1.12") && !b.UseEtcdManager() && b.UseEtcdTLS() {
 		name := "calico-client"
 		dirname := "calico"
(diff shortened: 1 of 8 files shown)


olemarkus:
/lgtm
Always wondered about why we kept those ports open.

hakman:
> Always wondered about why we kept those ports open.

Me too...
Next step would be to fix the SG reconciling logic.

olemarkus:
Yeah. I fixed that for openstack, but that may have been easier than it is for AWS. I wanted to see how it goes with OS anyway before attempting the same on AWS.

hakman:
After this change I ended up with 2 rules:
```
Custom TCP | TCP | 2382 - 4000 | sg-052348d0f129fc923 (nodes.test.local)
Custom TCP | TCP | 2382 - 4001 | sg-052348d0f129fc923 (nodes.test.local)
```

PR #10034, opened 2020-10-09
kops/cmd: add sshAccess to be recognized for the kops set cluster command

cpanato:
Some fields you cannot change using the `kops set cluster` command because the validation only parses a few.

Adding the `sshAccess` to be in the list for the release-1.18 branch.
On the main branch this is fixed and you can change all the fields, but I cannot use master right now for some cases :)

I propose to add this tiny change and I'm willing to add more in this PR if needed.
I know that might have another release for this branch so will be good to have this.

thanks!

/cc @hakman


hakman:
Thanks for adding this @cpanato .
/lgtm
/approve

Issue #10021, opened 2020-10-07
kops upgrade image requires kope.io


antoinedeschenes:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

1.18.1

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

1.18.9

**3. What cloud provider are you using?**

aws

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

`kops upgrade cluster` on a kops >=1.18.0 cluster

**5. What happened after the commands executed?**

```
I1007 16:04:36.843987  775287 upgrade_cluster.go:198] Custom image (099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907) has been provided for Instance Group "nodes"; not updating image
```

**6. What did you expect to happen?**

kops should handle upgrades from images listed in the channels

**7. Please provide your cluster manifest. Execute

**8. Please run the commands with most verbose logging by adding the `-v 10` flag.
  Paste the logs into this report, or in a gist and provide the gist link here.**

**9. Anything else do we need to know?**

Message is triggered when the image name doesn't contain "kope.io". It should probably check for the images listed in the channels/stable file instead

See:
https://github.com/kubernetes/kops/blob/56f3cb33481ab7c1d073ddaac6e4873b0b264249/cmd/kops/upgrade_cluster.go#L184

stable channel now uses the 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907 image, which triggers the warning
https://github.com/kubernetes/kops/blob/56f3cb33481ab7c1d073ddaac6e4873b0b264249/channels/stable

nvanheuverzwijn:
To handle this properly I see these solutions :

1. Add a flag to explicitely *not* update the image such as `kops upgrade cluster --yes --skip-image-update`.
2. Add a field other than image indicating we are not using channel's images.
3. Assume that all ubuntu image will be auto-updated.
4. Keep track of older image that went into the channel.

PR #10023, opened 2020-10-08
upgrade-cluster: Use correct test mechanism when auto-updating image name

nvanheuverzwijn:
When automatically managing image name of cluster, make sure we check that the image name is in the channel list instead of verifying that kope.io is in the name.

Fixes #10021

**Sample output BEFORE this patch**
```
$ kops upgrade cluster
Using cluster from kubectl context: test

I1008 00:38:57.241945   25556 upgrade_cluster.go:198] Custom image (099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907) has been provided for Instance Group "backups"; not updating image
I1008 00:38:57.242064   25556 upgrade_cluster.go:198] Custom image (099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907) has been provided for Instance Group "datagateways"; not updating image
I1008 00:38:57.242085   25556 upgrade_cluster.go:198] Custom image (099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907) has been provided for Instance Group "master-ca-central-1a"; not updating image
I1008 00:38:57.242100   25556 upgrade_cluster.go:198] Custom image (099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907) has been provided for Instance Group "nodes"; not updating image
I1008 00:38:57.242118   25556 upgrade_cluster.go:198] Custom image (099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907) has been provided for Instance Group "ops"; not updating image
I1008 00:38:57.242144   25556 upgrade_cluster.go:198] Custom image (099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907) has been provided for Instance Group "workers"; not updating image
ITEM	PROPERTY		OLD	NEW
Cluster	KubernetesVersion	1.18.9	1.19.2

Must specify --yes to perform upgrade
```

**Sample output BEFORE this patch**
```
$ kops upgrade cluster
Using cluster from kubectl context: ca-accp.k8s.equisoft.io

ITEM	PROPERTY		OLD	NEW
Cluster	KubernetesVersion	1.18.9	1.19.2

Must specify --yes to perform upgrade

```

[...]

PR #10035, opened 2020-10-09
Allow upgrades when IG image is Ubuntu 20.04

hakman:
Ref: https://github.com/kubernetes/kops/issues/10021
/cc @rifelpet @justinsb


rifelpet:
for reference: this is only for 1.18, #10023 will have a more flexible solution going forward
/lgtm

hakman:
Thanks @rifelpet 🙂

Issue #9751, opened 2020-08-14
Improvement suggestion for ADDING A BASTION HOST TO OUR CLUSTER documentation


vkotheka:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Version 1.18.0

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
This is a documentation issue on https://github.com/kubernetes/kops/blob/master/docs/examples/kops-tests-private-net-bastion-host.md#adding-a-bastion-host-to-our-cluster

**5. What happened after the commands executed?**
No commands executed as this is a doc issue

**6. What did you expect to happen?**
The doc provides example to ssh assuming the user = admin
Which is not always the case because if it's an ubuntu image underlying then it will be ubuntu user. That makes sense and is obvious.

However, if we don't explicitly mention that please use the appropriate user for ssh then a reader may think:
- admin user is created by KOPS behind the scene on bastion and cluster nodes.

It's not a big issue but just adding that line of caution will help if somebody runs into this trap.

The problem is when I ssh using admin@LB-bastion, it doesn't say invalid user - it gives an error 
**admin@LB-bastion: Permission denied (publickey).**
which unfortunately is misleading

h3poteto:
Hi, I will take this issue.

/assign

PR #10044, opened 2020-10-12
Rewrite ssh user to ubuntu in bastions docs

h3poteto:
Refs: #9751 
I only update a doc `kops-tests-private-net-bastion-host`.

### DIFF
--- a/docs/examples/kops-tests-private-net-bastion-host.md
+++ b/docs/examples/kops-tests-private-net-bastion-host.md
@@ -63,8 +63,8 @@ kops create cluster \
 --node-count=2 \
 --topology private \
 --networking kopeio-vxlan \
---node-size=t2.micro \
---master-size=t2.micro \
+--node-size=t3.micro \
+--master-size=t3.micro \
 ${NAME}
 ```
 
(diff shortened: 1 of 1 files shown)


hakman:
@h3poteto Hello again. In this case, I think we may want to update all docs from `admin@` to `ubuntu@`. There maybe be some other references to Debian in those places that should be updated also.
Would this be ok for you?

h3poteto:
Hi, OK, I see.
I will update all user from `admin@` to `ubuntu@`.

hakman:
Thanks @h3poteto

h3poteto:
@hakman I updated `kops-tests-private-net-bastion-host`.
Should I rewrite ssh user in other documents as well?
I confirm that `admin` user is used for ssh in some documents, but I haven't confirmed if they are the correct information yet.

docs/examples/kops-tests-private-net-bastion-host.md
```diff
@@ -148,8 +148,8 @@ kind: InstanceGroup
 metadata:
   name: bastions
 spec:
-  image: kope.io/k8s-1.7-debian-jessie-amd64-hvm-ebs-2017-07-28
-  machineType: t2.micro
+  image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20200907
```
hakman, line 151: There is another reference to a `kope.io` image in the file. Please update that one too.
Also, please change `t2.micro` to `t3.micro` everywhere.

hakman (changes requested):
Let's keep this commit focused on this file.The rest of the `admin@` can be updated in a followup PR.

docs/examples/kops-tests-private-net-bastion-host.md
h3poteto: I changed it in https://github.com/kubernetes/kops/pull/10044/commits/aff8c0af0090bb8e61c9c11a55c418547b95082b

hakman:
Great, thanks @h3poteto !
/lgtm
/approve

PR #10045, opened 2020-10-12
Fix bug in MergeAddons and test

nckturner:
I was a little bit confused by the desired behavior of the AddonMenu merge in channels so I decided to write a test to better understand it.  Can someone help me understand the expected behavior after a merge of two AddonMenus, like what happens in `apply_channel.go`?  Unless my understanding is incorrect, (or I've got a bug in the test) it seems like the merge behavior is incorrect, or maybe just confusing.  

```
--- FAIL: Test_MergeAddons (0.00s)
    addons_test.go:223: Unexpected AddonMenu merge result,
        Merged:
        AddonMenu{
          Addon{Name: a, Version: 1.0.0, KubernetesVersion: >=1.18.0, Id: k8s-1.18},
        }
        
        Expected:
        AddonMenu{
          Addon{Name: a, Version: 1.0.1, KubernetesVersion: >=1.18.0, Id: k8s-1.18},
        }
        
    addons_test.go:223: Unexpected AddonMenu merge result,
        Merged:
        AddonMenu{
          Addon{Name: a, Version: 1.0.0, KubernetesVersion: >=1.18.0, Id: k8s-1.18},
        }
        
        Expected:
        AddonMenu{
          Addon{Name: a, Version: 1.0.1, KubernetesVersion: >=1.18.0, Id: k8s-1.18},
        }
        
FAIL
FAIL	k8s.io/kops/channels/pkg/channels	0.016s
FAIL
```

This test passes with a change in addon.go from:
```
if existing.ChannelVersion().replaces(v.ChannelVersion()) {
```
to:
```
if v.ChannelVersion().replaces(existing.ChannelVersion()) {
```
So either my intuitive understanding of the MergeAddons function is off or this is a bug?

### DIFF
--- a/channels/pkg/channels/addons_test.go
+++ b/channels/pkg/channels/addons_test.go
@@ -17,7 +17,9 @@ limitations under the License.
 package channels
 
 import (
+	"fmt"
 	"net/url"
+	"reflect"
 	"testing"
 
 	"github.com/blang/semver/v4"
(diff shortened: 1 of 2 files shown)


ayberk:
This looks like an actual bug because [replaces](https://github.com/kubernetes/kops/blob/56f3cb33481ab7c1d073ddaac6e4873b0b264249/channels/pkg/channels/channel_version.go#L105) has to be called from the new version by passing the existing as a parameter. It should be `if v.ChannelVersion().replaces(existing.ChannelVersion())` indeed.

rifelpet:
Reading over the [ChannelVersion.replaces](https://github.com/kubernetes/kops/blame/master/channels/pkg/channels/channel_version.go#L105) code, I'm inclined to agree that [this](https://github.com/kubernetes/kops/blame/master/channels/pkg/channels/addon.go#L62) should be flipped around. Given that, I don't understand how new Kops versions could still be applying addons with updated versions. Most of this code hasn't been touched in over a year, some of it 4 years. How has this bug not been noticed until now?

/cc @justinsb for input

nckturner:
I've added a commit that modifies the check, as it does seem to make sense as the correct merge behavior, but agree with @rifelpet that its strange this has never been detected.  I think its because the usages I see for MergeAddons() are where we merge the addons from the virtual file system (i.e. S3) with those from what appears to be a local filesystem location.  I'm not sure this code is even used during normal operations.  Further down in apply_channel.go there is additional update code, which I think is the main update logic: https://github.com/kubernetes/kops/blob/master/channels/pkg/cmd/apply_channel.go#L136-L202

channels/pkg/channels/addons_test.go
```diff
@@ -197,6 +199,68 @@ func Test_UnparseableVersion(t *testing.T) {
...
+	}
+	return addonMenu
+}
+
+func addonString(addon *Addon) string {
+	return fmt.Sprintf("  Addon{Name: %s, Version: %s, KubernetesVersion: %s, Id: %s},\n", *addon.Spec.Name, *addon.Spec.Version, addon.Spec.KubernetesVersion, addon.Spec.Id)
+}
+
+func addonMenuString(addonMenu *AddonMenu) string {
+	addonMenuString := fmt.Sprintf("AddonMenu{\n")
```
hakman, line 261:
```suggestion
	addonMenuString := "AddonMenu{\n"
```

channels/pkg/channels/addons_test.go
```diff
@@ -197,6 +199,68 @@ func Test_UnparseableVersion(t *testing.T) {
...
+	for _, addon := range addonMenu.Addons {
+		addonMenuString += addonString(addon)
+	}
+	return addonMenuString + fmt.Sprintf("}\n")
```
hakman, line 265:
```suggestion
	return addonMenuString +"}\n"
```

hakman (changes requested):
These should make `staticcheck` happy.

nckturner:
@hakman thanks for cleaning that up

channels/pkg/channels/addons_test.go
```diff
@@ -197,6 +199,68 @@ func Test_UnparseableVersion(t *testing.T) {
...
+func addonString(addon *Addon) string {
+	return fmt.Sprintf("  Addon{Name: %s, Version: %s, KubernetesVersion: %s, Id: %s},\n", *addon.Spec.Name, *addon.Spec.Version, addon.Spec.KubernetesVersion, addon.Spec.Id)
+}
+
+func addonMenuString(addonMenu *AddonMenu) string {
+	addonMenuString := "AddonMenu{\n"
+	for _, addon := range addonMenu.Addons {
+		addonMenuString += addonString(addon)
+	}
+	return addonMenuString +"}\n"
```
hakman, line 265: My bad, now `gofmt` complains.
```suggestion
	return addonMenuString + "}\n"
```

hakman:
> @hakman thanks for cleaning that up

No pb. Let's mention this next office hours and maybe get a quick review from @justinsb .

justinsb:
Great find & fix - thanks @nckturner 

I think we've got away with this in the past because historically we've only normally had a single manifest with our addons, so we likely haven't really gone down this codepath.

/approve
/lgtm

PR #10048, opened 2020-10-14
Install container runtime packages as assets

hakman:
The main purpose for this PR is to move the logic of choosing the container runtime (Docker or containerd) package from noddup to the assets building phase, when running `kops update cluster`. An additional benefit is that it's easy to add new versions and added all existing Docker packages to the supported list.

Feature related changes are quite small and limited to the "Main" and "Misc" commits.
Tests changes are more extensive to be able to do changes with confidence and are separated in the "Tests" commit.
Integration tests changes are about adding the new asset to `conf/kube_env.yaml`, but touches all these tests. Fortunately changes are separated in the "Integration" commit.


upup/pkg/fi/cloudup/containerd.go
```diff
@@ -0,0 +1,173 @@
...
+		"1.3.4": "4616971c3ad21c24f2f2320fa1c085577a91032a068dd56a41c7c4b71a458087",
+		"1.4.0": "b379f29417efd583f77e095173d4d0bd6bb001f0081b2a63d152ee7aef653ce1",
+		"1.4.1": "757efb93a4f3161efc447a943317503d8a7ded5cb4cc0cba3f3318d7ce1542ed",
+	}
+
+	return hashes
+}
+
+func findAllContainerdDockerMappings() map[string]string {
+	versions := map[string]string{
```
johngmyers, line 164: So:
* drops 1.2.4
* adds 1.2.6 and 1.2.12
* changes 1.2.10 from containerd to docker
* changes 1.2.13 from docker 19.03.11 to 19.03.12

Shouldn't at least the dropping of 1.2.4 be mentioned in release notes?

upup/pkg/fi/cloudup/docker.go
```diff
@@ -0,0 +1,239 @@
...
+	dockerVersionUrlArm64 = "https://download.docker.com/linux/static/stable/aarch64/docker-%s.tgz"
+	// Docker legacy AMD64 packages URLs for v17.03.0 to v18.06.3
+	dockerLegacyUrlAmd64 = "https://download.docker.com/linux/static/stable/x86_64/docker-%s-ce.tgz"
+	// Docker legacy ARM64 packages URLs for v17.09.0 to v18.06.3
+	dockerLegacyUrlArm64 = "https://download.docker.com/linux/static/stable/aarch64/docker-%s-ce.tgz"
+	// Docker version that is available for both AMD64 and ARM64, used in case the selected version is too old and not available for ARM64
+	dockerFallbackVersion = "17.09.0"
+)
+
+func findDockerAssets(c *kops.Cluster, assetBuilder *assets.AssetBuilder, arch architectures.Architecture) (*url.URL, *hashing.Hash, error) {
```
johngmyers, line 44: `findDockerAsset()`?

upup/pkg/fi/cloudup/docker.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	return h, nil
+}
+
+func findAllDockerHashesAmd64() map[string]string {
+	hashes := map[string]string{
+		"17.03.0":  "aac08524db82d3fdc8fc092f495e1174f5e1dd774b95a6d081544997d34b4855",
+		"17.03.1":  "3e070e7b34e99cf631f44d0ff5cf9a127c0b8af5c53dfc3e1fce4f9615fbf603",
+		"17.03.2":  "183b31b001e7480f3c691080486401aa519101a5cfe6e05ad01b9f5521c4112d",
```
johngmyers, line 138: So moving all distributions to the "static" variant. Is the removal of the docker-ce-selinux package on RHEL-7 and derived distributions not a problem?

(This is a change for 18.06.3 and earlier, with the exception of 17.03.2 on Ubuntu 18.04.)

nodeup/pkg/model/docker.go
```diff
@@ -435,76 +114,6 @@ func (b *DockerBuilder) Build(c *fi.ModelBuilderContext) error {
...
-			}
-
-			// As a mitigation for CVE-2019-5736 (possibly a fix, definitely defense-in-depth) we chattr docker-runc to be immutable
-			for _, f := range dv.MarkImmutable {
-				c.AddTask(&nodetasks.Chattr{
-					File: f,
-					Mode: "+i",
-					Deps: []fi.Task{packageTask},
-				})
-			}
```
johngmyers, line 494: Do we not need this mitigation anymore?

nodeup/pkg/model/context.go
```diff
@@ -598,29 +599,31 @@ func (c *NodeupModelContext) GetPrivateKey(name string) ([]byte, error) {
 
 func (b *NodeupModelContext) AddCNIBinAssets(c *fi.ModelBuilderContext, assetNames []string) error {
 	for _, assetName := range assetNames {
-		if err := b.addCNIBinAsset(c, assetName); err != nil {
+		re, err := regexp.Compile(fmt.Sprintf("^%s$", assetName))
```
johngmyers, line 602: I'm a little worried about regexp metacharacters, such as ".", in the assetName.

Perhaps use `regexp.Compile(fmt.Sprintf("^%s$", regexp.QuoteMeta(assetName)))` instead?

nodeup/pkg/model/context.go
```diff
@@ -598,29 +599,31 @@ func (c *NodeupModelContext) GetPrivateKey(name string) ([]byte, error) {
...
-	asset, err := b.Assets.Find(assetName, assetPath)
-	if err != nil {
-		return fmt.Errorf("error trying to locate asset %q: %v", assetName, err)
-	}
-	if asset == nil {
-		return fmt.Errorf("unable to locate asset %q", assetName)
+func (b *NodeupModelContext) addCNIBinAsset(c *fi.ModelBuilderContext, assetPath *regexp.Regexp) error {
+	a := b.Assets.FindMatches(assetPath)
+	if len(a) != 1 {
+		return fmt.Errorf("unable to locate asset %q", assetPath.String())
```
johngmyers, line 616: "unable to locate" is misleading if `len(a) > 1`.

johngmyers (changes requested):

upup/pkg/fi/cloudup/docker.go
hakman: @justinsb also noticed this, so will try to explain below.

Kops support for SELinux was added in 1.19 (mostly experimental from my point of view) and could not be enabled in docker config in the past. Contents of the packages show that it contains only the SELinux profiles:
```bash
> rpm -qlp ./docker-ce-selinux-17.03.2.ce-1.el7.centos.noarch.rpm 
/usr/share/doc/docker-ce-selinux-17.03.2.ce
/usr/share/doc/docker-ce-selinux-17.03.2.ce/LICENSE
/usr/share/selinux/devel/include/services/docker.if
/usr/share/selinux/packages/docker.pp.bz2
```

The `docker-ce-selinux` package is distributed only for 17.03.x and later versions switch to the `container-selinux` dependency, distributed with RHEL/CentOS.

Considering that SELinux could not be used in the past and we already install by default a replacement package, I don't think this is an issue.

upup/pkg/fi/cloudup/containerd.go
```diff
@@ -0,0 +1,173 @@
...
+const (
+	// containerd packages URLs for v1.4.x+
+	containerdVersionUrlAmd64 = "https://github.com/containerd/containerd/releases/download/v%s/cri-containerd-cni-%s-linux-amd64.tar.gz"
+	// containerd legacy packages URLs for v1.2.x and v1.3.x
+	containerdLegacyUrlAmd64 = "https://storage.googleapis.com/cri-containerd-release/cri-containerd-%s.linux-amd64.tar.gz"
+	// containerd version that is available for both AMD64 and ARM64, used in case the selected version is not available for ARM64
+	containerdFallbackVersion = "1.2.13"
+)
+
+func findContainerdAssets(c *kops.Cluster, assetBuilder *assets.AssetBuilder, arch architectures.Architecture) (*url.URL, *hashing.Hash, error) {
```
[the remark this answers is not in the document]
hakman: Wanted to keep the same naming convention as for `findCNIAssets` and `findLyftVPCAssets`.
I also think the singular is the better solution. Will change.

upup/pkg/fi/cloudup/docker.go
hakman: Wanted to keep the same naming convention as for findCNIAssets and findLyftVPCAssets.
I also think the singular is the better solution. Will change.

nodeup/pkg/model/docker.go
hakman: I saw this done only for 17.03.x and didn't seem like something that important anymore considering the number of security issues with that version. Anyone interested in security should really use a newer Docker version.

nodeup/pkg/model/context.go
hakman: Sounds reasonable. Thanks for the suggestion.

nodeup/pkg/model/context.go
hakman: Agreed. Will change.

upup/pkg/fi/cloudup/containerd.go
hakman: Containerd support is a bit of a puzzle here. Will have to think about it.

nodeup/pkg/model/docker.go
johngmyers: It doesn't have value for newer versions, but I don't think we should reintroduce a significant security vulnerability for 17.03.

Perhaps we should work on our deprecation policy for container runtimes.

upup/pkg/fi/cloudup/containerd.go
johngmyers: I would be okay with only moving the newer versions of containerd to assets phase.

hakman: The problem comes mostly from packaging bugs in containerd and package URL changes and lack of ARM builds...
So, to the point:
* drops 1.2.4 - adding back pretty easy to do
* adds 1.2.6 and 1.2.12 - it is not a big deal, added them just because it was easy to map to Docker packages
* changes 1.2.10 from containerd to docker - intended, shouldn't make any difference
* changes 1.2.13 from docker 19.03.11 to 19.03.12 - reverted this

nodeup/pkg/model/docker.go
hakman: Good call on digging a bit more. Seems that this is wider than initially thought and affects about anything < 18.09.2.
Re-added the mitigation and updated the tests output.

> Perhaps we should work on our deprecation policy for container runtimes.
+1 from me 😄

hakman:
@johngmyers I think I addressed all the review comments in my latest commit.

justinsb:
Extracted out the package scripts for a few representative packages, the only thing I saw was adding the docker group and some apparmor (and we didn't support apparmor for these older versions).

Thanks for doing this @hakman !

hakman:
Thanks for checking @justinsb 😄

PR #10064, opened 2020-10-17
Update Flannel CNI to v0.13.0

hakman:
https://github.com/coreos/flannel/releases/tag/v0.13.0

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.flannel/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.flannel/k8s-1.12.yaml.template
@@ -18,6 +18,7 @@ spec:
   - emptyDir
   - hostPath
   allowedHostPaths:
+  - pathPrefix: "/dev/net"
   - pathPrefix: "/etc/cni/net.d"
   - pathPrefix: "/etc/kube-flannel"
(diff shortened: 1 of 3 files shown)


upup/models/cloudup/resources/addons/networking.flannel/k8s-1.12.yaml.template
```diff
@@ -86,37 +140,43 @@ data:
...
   namespace: kube-system
   labels:
-    k8s-app: flannel
+    tier: node
+    app: flannel
     role.kubernetes.io/networking: "1"
 spec:
   selector:
     matchLabels:
-      tier: node
```
rifelpet, line 100: I think the selector field of a daemonset is immutable which would cause applying this manifest on an existing cluster to fail.

hakman: Right as usual:
```
 The DaemonSet "kube-flannel-ds" is invalid: spec.selector: Invalid value:
v1.LabelSelector{MatchLabels:map[string]string{"app":"flannel"},
MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

Issue #10009, opened 2020-10-01
Rolling update ignore unready nodes in inapplicable instance groups


johngmyers:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

Currently when a rolling update is performing cluster validation and an unrelated nodes instancegroup scales up, the rolling update will wait for that node to pass validation. It would be desirable if the rolling update could proceed even when nodes in unrelated instance groups are unhealthy.

The exception is that failures in instance groups of role "master" should probably block the progression of rolling update in other instance groups.

**2. Feel free to provide a design supporting your feature request.**

The cluster validation API will need to be changed such that failures that are specific to a node are associated with that node. Examples of such failures would be node readiness and unready pods with system-node-critical priority. The "not enough nodes in instance group" failure would need to be associated with the group. Rolling update would then need to ignore validation failures that are in irrelevant groups.

bharath-123:
Hey @johngmyers ! I would like to pick this issue up and submit a PR. 
Any steps to reproduce this issue or any suggestions for me before setting out to resolve this issue would be highly appreciated :)

johngmyers:
@bharath-123 to reproduce this, start with a cluster that has at least two instance groups of role "Nodes". Perform a rolling update on one of the instance groups, using `--force` if necessary to get it to update the group's nodes. While the a new node is coming up and rolling update is validating, terminate a node in the other instance group. You will notice that the validation phase will wait for the other instance group to become completely ready before proceeding to drain another node in the instance group it is updating.

A PR for this will have at least two commits. One commit will update the `validation.ClusterValidator` interface so that a caller will be able to tell when a returned `ValidationError` is specific to a node or group and be able to determine which instance group that applies to. I'm being a bit nonspecific on how it should return this information as that is part of designing the change, but I can give suggestions if needed.

The change will also have to correspondingly mark the group-specific `ValidationError`s as such. Those failures are:

* In `collectPodFailures`, any of the `Kind: "Pod"` failures when the priority is `"system-node-critical"`. These are things like CNI DaemonSet pods. (When the priority is "system-cluster-critical", that is a cluster-wide failure that should always block a rolling update.)
* All of the failures returned by `validateNodes`.

My view is that an unhealthy control plane should block a rolling update, so the `masterWithoutPod` failures could be considered to be cluster-wide. It would be acceptable to consider them specific to their instance group.

[...]

PR #10065, opened 2020-10-17
Avoid waiting on validation during rolling update for inapplicable instance groups

bharath-123:
When an error occurs during cluster validation during a rolling update, the validation phase waits for a certain period of time before continuing. 

We can avoid this wait for certain errors inapplicable to an instance group whose rolling update occurring. eg: When a node in a different instance group is terminated leading to a error regarding instance group target size being lesser than expected. For such errors we do not want rolling update on the instance group to wait for errors on inapplicable instance groups.

Fixes [https://github.com/kubernetes/kops/issues/10009](https://github.com/kubernetes/kops/issues/10009)

Really appreciate the review comments and discussions. Working on this PR so far has been a great experience for me.

### DIFF
--- a/pkg/instancegroups/instancegroups.go
+++ b/pkg/instancegroups/instancegroups.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/klog/v2"
 	api "k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/cloudinstances"
+	"k8s.io/kops/pkg/validation"
 	"k8s.io/kubectl/pkg/drain"
 )
 
@@ -92,7 +93,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(group *cloudinstances.
 
 	if isBastion {
 		klog.V(3).Info("Not validating the cluster as instance is a bastion.")
-	} else if err = c.maybeValidate("", 1); err != nil {
+	} else if err = c.maybeValidate("", 1, group); err != nil {
 		return err
 	}
 
@@ -147,7 +148,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(group *cloudinstances.
 					klog.Infof("waiting for %v after detaching instance", sleepAfterTerminate)
 					time.Sleep(sleepAfterTerminate)
 
-					if err := c.maybeValidate(" after detaching instance", c.ValidateCount); err != nil {
+					if err := c.maybeValidate(" after detaching instance", c.ValidateCount, group); err != nil {
 						return err
 					}
 					noneReady = false
@@ -181,7 +182,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(group *cloudinstances.
 			return waitForPendingBeforeReturningError(runningDrains, terminateChan, err)
 		}
 
-		err = c.maybeValidate(" after terminating instance", c.ValidateCount)
+		err = c.maybeValidate(" after terminating instance", c.ValidateCount, group)
 		if err != nil {
 			return waitForPendingBeforeReturningError(runningDrains, terminateChan, err)
 		}
@@ -227,7 +228,7 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(group *cloudinstances.
 			}
 		}
 
-		err = c.maybeValidate(" after terminating instance", c.ValidateCount)
+		err = c.maybeValidate(" after terminating instance", c.ValidateCount, group)
 		if err != nil {
 			return err
 		}
@@ -410,14 +411,14 @@ func (c *RollingUpdateCluster) reconcileInstanceGroup() error {
 
 }
 
-func (c *RollingUpdateCluster) maybeValidate(operation string, validateCount int) error {
+func (c *RollingUpdateCluster) maybeValidate(operation string, validateCount int, group *cloudinstances.CloudInstanceGroup) error {
 	if c.CloudOnly {
 		klog.Warningf("Not validating cluster as cloudonly flag is set.")
 
 	} else {
 		klog.Info("Validating the cluster.")
 
-		if err := c.validateClusterWithTimeout(validateCount); err != nil {
+		if err := c.validateClusterWithTimeout(validateCount, group); err != nil {
 
 			if c.FailOnValidate {
 				klog.Errorf("Cluster did not validate within %s", c.ValidationTimeout)
@@ -431,7 +432,7 @@ func (c *RollingUpdateCluster) maybeValidate(operation string, validateCount int
 }
 
 // validateClusterWithTimeout runs validation.ValidateCluster until either we get positive result or the timeout expires
-func (c *RollingUpdateCluster) validateClusterWithTimeout(validateCount int) error {
+func (c *RollingUpdateCluster) validateClusterWithTimeout(validateCount int, group *cloudinstances.CloudInstanceGroup) error {
 	ctx, cancel := context.WithTimeout(context.Background(), c.ValidationTimeout)
 	defer cancel()
 
@@ -445,7 +446,7 @@ func (c *RollingUpdateCluster) validateClusterWithTimeout(validateCount int) err
 	for {
 		// Note that we validate at least once before checking the timeout, in case the cluster is healthy with a short timeout
 		result, err := c.ClusterValidator.Validate()
-		if err == nil && len(result.Failures) == 0 {
+		if err == nil && !hasFailureRelevantToGroup(result.Failures, group) {
 			successCount++
 			if successCount >= validateCount {
 				klog.Info("Cluster validated.")
@@ -477,12 +478,33 @@ func (c *RollingUpdateCluster) validateClusterWithTimeout(validateCount int) err
 		// Reset the success count; we want N consecutive successful validations
 		successCount = 0
 
-		// Wait before retrying
+		// Wait before retrying in some cases
 		// TODO: Should we check if we have enough time left before the deadline?
 		time.Sleep(c.ValidateTickDuration)
 	}
 
 	return fmt.Errorf("cluster did not validate within a duration of %q", c.ValidationTimeout)
+}
+
+// checks if the validation failures returned after cluster validation are relevant to the current
+// instance group whose rolling update is occurring
+func hasFailureRelevantToGroup(failures []*validation.ValidationError, group *cloudinstances.CloudInstanceGroup) bool {
+	// Ignore non critical validation errors in other instance groups like below target size errors
+	for _, failure := range failures {
+		// Determining InstanceGroups for certain resources like Pods, ComponentStatus is not straightforward.
+		// Till we are able to determine the InstanceGroups for these resources without ambiguity, the
+		// InstanceGroup field of the ValidationErrors for these resources will be nil
+		if failure.InstanceGroup == nil {
+			return true
+		}
+
+		// if there is a failure in the same instance group or a failure which has cluster wide impact
+		if (failure.InstanceGroup.IsMaster()) || (failure.InstanceGroup == group.InstanceGroup) {
+			return true
+		}
+	}
+
+	return false
 }
 
 // detachInstance detaches a Cloud Instance
@@ -608,7 +630,7 @@ func (c *RollingUpdateCluster) UpdateSingleInstance(cloudMember *cloudinstances.
 		if err != nil {
 			return fmt.Errorf("failed to detach instance: %v", err)
 		}
-		if err := c.maybeValidate(" after detaching instance", c.ValidateCount); err != nil {
+		if err := c.maybeValidate(" after detaching instance", c.ValidateCount, cloudMember.CloudInstanceGroup); err != nil {
 			return err
 		}
 	}
--- a/pkg/instancegroups/rollingupdate_test.go
+++ b/pkg/instancegroups/rollingupdate_test.go
@@ -105,6 +105,24 @@ func (*erroringClusterValidator) Validate() (*validation.ValidationCluster, erro
 	return nil, errors.New("testing validation error")
 }
 
+// instanceGroupNodeSpecificErrorClusterValidator simulates failures in a specific node group in the map of instance groups
+type instanceGroupNodeSpecificErrorClusterValidator struct {
+	InstanceGroup *kopsapi.InstanceGroup
+}
+
+func (igErrorValidator *instanceGroupNodeSpecificErrorClusterValidator) Validate() (*validation.ValidationCluster, error) {
+	return &validation.ValidationCluster{
+		Failures: []*validation.ValidationError{
+			{
+				Kind:          "testing",
+				Name:          "testing failure",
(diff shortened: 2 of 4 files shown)


pkg/instancegroups/instancegroups.go
```diff
@@ -477,14 +478,32 @@ func (c *RollingUpdateCluster) validateClusterWithTimeout(validateCount int) err
...
 		successCount = 0
 
-		// Wait before retrying
-		// TODO: Should we check if we have enough time left before the deadline?
-		time.Sleep(c.ValidateTickDuration)
+		// Wait before retrying in some cases
+		// TODO: Should we check if we have enough time left before the deadline? 
+		if shouldWaitBeforeRetry(result.Failures, group) {
+			time.Sleep(c.ValidateTickDuration)
+		}
```
johngmyers, line 485: There's a logic error here: when shouldWaitBeforeRetry fails, instead of letting the validation succeed and proceeding with updating the next node it instead goes into a fast spinloop.

When all failures are inapplicable it should go through the code block that increments `successCount`.

pkg/validation/validate_cluster_test.go
```diff
@@ -154,22 +155,24 @@ func Test_ValidateCloudGroupMissing(t *testing.T) {
 			Kind:    "InstanceGroup",
 			Name:    "node-1",
 			Message: "InstanceGroup \"node-1\" is missing from the cloud provider",
+			InstanceGroup: instanceGroup,
```
johngmyers, line 158: Nit: you could have made a smaller change by using `instanceGroups[0]`.

bharath-123:
Seems like some static checks and format tests are failing. Should have run gofmt and a bunch of other static checks before raising the PR. Will do so now.

johngmyers (changes requested):
You'll need to run `make gofmt` and fix that duplicate import in order to fix those failed tests.

A useful unit test for rolling update would be to mock in a validation failure for a nodes group that is not selected for update and verify that the rolling update completes regardless.

The validation unit tests should be updated to assert that the instanceGroup field is filled in appropriately.

In `collectPodFailures`, failures for pods of priority `system-node-critical` should be marked as being specific to the pod's node's group. These are typically things like CNI DaemonSets. Pods that are `system-cluster-critical`, on the other hand, are cluster-level failures.

johngmyers:
If the `system-node-critical` feature is too complicated, it can be done as a follow-up PR.

bharath-123:
@johngmyers I was not able to figure out how to infer the InstanceGroup for a pod in collectPodFailures? The pod has a NodeName field in it's podspec. But I was not really sure how we could infer the InstanceGroup for a pod from that?

bharath-123:
I will follow this up with the unit test and the code changes that you have suggested. Thank you.

johngmyers:
It would be better to edit the commit stream so that the formatting and staticcheck issues were fixed in their original commits.

I like the separation between the first and second commits. The third commit should be combined with the second as it is part of the same logical change.

bharath-123:
> If the `system-node-critical` feature is too complicated, it can be done as a follow-up PR.

Shouldn't be difficult if we can figure out how to infer the instanceGroup from a pod given its structure.

bharath-123:
> It would be better to edit the commit stream so that the formatting and staticcheck issues were fixed in their original commits.
> 
> I like the separation between the first and second commits. The third commit should be combined with the second as it is part of the same logical change.

yes I agree with this. I will anyways be doing a whole bunch of changes now. So will revise the commit stream.

pkg/instancegroups/instancegroups.go
bharath-123: One question, even if the failures are inapplicable and we skip the wait period. We definitely want to log the failures right? I think we would need to change the log messages here if I am right. If the validation fails and the failures are inapplicable, showing the failure log messages and a cluster validated messages would be very unintuitive for the sys admins. 
I am planning to copy this block of code and add in the else stmt if shouldWaitBeforeRetry fails. I will change up the log messages to indicate that there **were** failures but they were ignored because they are in a different instance group.
```
			successCount++
			if successCount >= validateCount {
				klog.Info("Cluster validated.")
				return nil
			}
			klog.Infof("Cluster validated; revalidating in %s to make sure it does not flap.", c.ValidateSuccessDuration)
			time.Sleep(c.ValidateSuccessDuration)
			continue
```

johngmyers:
`validateNodes` could construct a map of nodes to groups. `collectPodFailures` already constructs a map of IP address to node, so it could then use that map of nodes to groups to go from pod to group.

johngmyers:
The rolling update code should probably remove the inapplicable failures from what it logs.

pkg/validation/validate_cluster.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/pager"
+	kopsapi "k8s.io/kops/pkg/apis/kops"
```
justinsb, line 28: This should already be imported as kops below, I believe

johngmyers:
Okay, not sure why kube-apiserver defaults that off. An admission controller would be more in line with current practice and would allow policies more nuanced than "prohibit everything".

bharath-123:
@johngmyers I was working on this? Hard to manage this and office work, so I mostly work on the weekends :/ Any reason why this was closed?

bharath-123:
I did not understand your last comment?

johngmyers:
Sorry, wrong PR.

pkg/validation/validate_cluster.go
bharath-123: yes @justinsb will be doing this.

bharath-123:
> `validateNodes` could construct a map of nodes to groups. `collectPodFailures` already constructs a map of IP address to node, so it could then use that map of nodes to groups to go from pod to group.

Great idea. I ll do this in a follow up PR tho. All the pod failures caught during validate cluster phase are system critical anyways. By not adding the instanceGroup field to them, I am not ignoring the errors.

bharath-123:
> The rolling update code should probably remove the inapplicable failures from what it logs.

Makes sense. I am working on a different implementation of this feature now.

bharath-123:
@johngmyers @justinsb @hakman @olemarkus @zetaab do have a look at this. This is a complete implementation with unit tests added for rolling update. Do review this.

bharath-123:
There are a couple of unused imports and unrelated comments which I had added when I was trying a different implementation. I ll remove them now.

bharath-123:
gentle ping @johngmyers @justinsb @hakman

rifelpet:
@bharath-123 Feel free to remove `[WIP]` from the PR title if this is ready for review.

bharath-123:
> @bharath-123 Feel free to remove `[WIP]` from the PR title if this is ready for review.

done

pkg/instancegroups/rollingupdate_test.go
```diff
@@ -105,6 +105,30 @@ func (*erroringClusterValidator) Validate() (*validation.ValidationCluster, erro
...
+// simulates failures in a specific node group in the map of instance groups
+type instanceGroupNodeSpecificErrorClusterValidator struct {
+	Groups    map[string]*cloudinstances.CloudInstanceGroup
+	NodeGroup string
+}
+
+func (igErrorValidator *instanceGroupNodeSpecificErrorClusterValidator) Validate() (*validation.ValidationCluster, error) {
+	instanceGroup := igErrorValidator.Groups[igErrorValidator.NodeGroup].InstanceGroup
+	if instanceGroup.IsMaster() || instanceGroup.IsBastion() {
+		instanceGroup = nil
```
johngmyers, line 117: This is testing the mock, not the code under test.

johngmyers (changes requested):
I don't see where in the code failures specific to a master IG cause rolling update to be blocked from proceeding.

From comments it seems you intended to put this logic in the validation code. I think it might be simpler to put it in the rolling update code that examines the returned list of failures.

pkg/instancegroups/instancegroups.go
```diff
@@ -477,14 +478,26 @@ func (c *RollingUpdateCluster) validateClusterWithTimeout(validateCount int) err
...
-		// Wait before retrying
+		// Wait before retrying in some cases
 		// TODO: Should we check if we have enough time left before the deadline?
 		time.Sleep(c.ValidateTickDuration)
 	}
 
 	return fmt.Errorf("cluster did not validate within a duration of %q", c.ValidationTimeout)
 }
 
+func isNotRelatedInstanceGroupError(failures []*validation.ValidationError, group *cloudinstances.CloudInstanceGroup) bool {
```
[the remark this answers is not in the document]
bharath-123: yes. makes sense. Will do that.

pkg/instancegroups/instancegroups.go
```diff
@@ -445,7 +446,7 @@ func (c *RollingUpdateCluster) validateClusterWithTimeout(validateCount int) err
 	for {
 		// Note that we validate at least once before checking the timeout, in case the cluster is healthy with a short timeout
 		result, err := c.ClusterValidator.Validate()
-		if err == nil && len(result.Failures) == 0 {
+		if err == nil && (len(result.Failures) == 0 || isNotRelatedInstanceGroupError(result.Failures, group)) {
```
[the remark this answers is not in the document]
bharath-123: yes. Will remove it.

pkg/validation/validate_cluster.go
```diff
@@ -49,6 +49,10 @@ type ValidationError struct {
 	Kind    string `json:"type,omitempty"`
 	Name    string `json:"name,omitempty"`
 	Message string `json:"message,omitempty"`
+	// optional field to indicate which instance group this validation error is coming from
+	// if InstanceGroup is nil, then the validation error is either a cluster wide error
+	// or an error in the master instancegroup which indicates a cluster wide error
```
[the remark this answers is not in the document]
bharath-123: Acked these comment suggestions. They are indeed more helpful.

pkg/instancegroups/rollingupdate_test.go
```diff
@@ -105,6 +105,30 @@ func (*erroringClusterValidator) Validate() (*validation.ValidationCluster, erro
...
 }
 
+// simulates failures in a specific node group in the map of instance groups
+type instanceGroupNodeSpecificErrorClusterValidator struct {
+	Groups    map[string]*cloudinstances.CloudInstanceGroup
+	NodeGroup string
+}
+
+func (igErrorValidator *instanceGroupNodeSpecificErrorClusterValidator) Validate() (*validation.ValidationCluster, error) {
+	instanceGroup := igErrorValidator.Groups[igErrorValidator.NodeGroup].InstanceGroup
```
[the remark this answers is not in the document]
bharath-123: alright. Will do that. It makes sense.

pkg/validation/validate_cluster.go
```diff
@@ -238,6 +243,9 @@ func (v *ValidationCluster) collectPodFailures(ctx context.Context, client kuber
 		if pod.Status.Phase == v1.PodSucceeded {
 			return nil
 		}
+
+		// pod validationErrors usually do not have instanceGroup field as all pod errors caught
+		// are system critical pods.
```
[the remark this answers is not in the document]
bharath-123: I inferred this from line 240 containing the following code:
```
		if priority != "system-cluster-critical" && priority != "system-node-critical" {
			return nil
		}
```
But I think I wrongly read system-node-critical. Need to brainstorm on how to get the InstanceGroup for the pods which doesn't seem to be straightforward.

bharath-123:
> I don't see where in the code failures specific to a master IG cause rolling update to be blocked from proceeding.
> 
> From comments it seems you intended to put this logic in the validation code. I think it might be simpler to put it in the rolling update code that examines the returned list of failures.



> I don't see where in the code failures specific to a master IG cause rolling update to be blocked from proceeding.
> 
> From comments it seems you intended to put this logic in the validation code. I think it might be simpler to put it in the rolling update code that examines the returned list of failures.

In the ValidationError struct, If the InstanceGroup field is nil then this failure is a cluster wide failure for which rolling update should indeed be blocked. For all validation failures in master IG, the InstanceGroup field is nil. isNotRelatedInstanceGroupError ( Which I will change the name of ) returns false if InstanceGroup field is nil which causes rolling update to be blocked. 

It would definitely be cleaner if we can have InstanceGroup field filled up for all failures. I mainly did this because it would be a bit difficult to identify the groups for the pods. I ll think more about this and get back.

pkg/instancegroups/rollingupdate_test.go
bharath-123: Would it be better if we include the Validate logic in the Validate mock? I thought just mocking the failures should have been enough to test this. Since we are mostly concerned about the failures return by Validate() and not how it returns it. Am I understanding something wrongly?

pkg/validation/validate_cluster.go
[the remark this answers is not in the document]
johngmyers: As mentioned previously, this can be deferred to a later PR.

pkg/instancegroups/rollingupdate_test.go
johngmyers: If it was the rolling update code that was responsible for ensuring failures for IGs of role master still blocked rolling update, then testing that in the rolling updates code would make sense. In the current design, where it is the validation code that is responsible for that logic, that logic should be tested by the validation unit tests, not the rolling update unit tests.

johngmyers:
> For all validation failures in master IG, the InstanceGroup field is nil.

I did not see code for ensuring the `InstanceGroup` field is `nil` for failures in IGs of role master.

pkg/validation/validate_cluster.go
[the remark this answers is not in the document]
bharath-123: Alright. So we can leave the InstanceGroup field as nil for validation error for kind Pods.

pkg/instancegroups/rollingupdate_test.go
bharath-123: Fair enough. This makes sense. I ll push all logic to determine whether we should wait or not for validation to the rolling update code rather than the validation code. It would make things much easier and simpler

bharath-123:
> > For all validation failures in master IG, the InstanceGroup field is nil.
> 
> I did not see code for ensuring the `InstanceGroup` field is `nil` for failures in IGs of role master.

Will be pushing all the logic to rolling update code now.

bharath-123:
@johngmyers working on refining this. Just to confirm, for Pods and ComponentStatuses it's not straightforward to get the InstanceGroup. This will anyways be deferred to a future PR. 

For these cases, in the rolling update code we will be considering cases where `InstanceGroup == nil` to be a relevant InstanceGroup failure. Would I need to assert these cases in the validate_cluster tests? In the rolling update test, I ll include a test to fail rolling update if a validation error is returned if InstanceGroup is nil.

bharath-123:
gentle ping on this @johngmyers . Thank you for your help and patience on this PR.

PR #10069, opened 2020-10-19
Add eBPF dataplane support for Calico CNI

hakman:
https://docs.projectcalico.org/about/about-ebpf
https://docs.projectcalico.org/maintenance/enabling-bpf
https://docs.projectcalico.org/reference/felix/configuration#ebpf-dataplane-configuration

### DIFF
--- a/docs/networking/calico.md
+++ b/docs/networking/calico.md
@@ -93,6 +93,23 @@ It is possible to configure Calico to use Typha by editing a cluster and adding 
       typhaReplicas: 3
 ```
 
+### Configuring the eBPF dataplane
+{{ kops_feature_table(kops_added_default='1.19', k8s_min='1.16') }}
+
+Calico supports using an [eBPF dataplane](https://docs.projectcalico.org/about/about-ebpf) as an alternative to the standard Linux dataplane (which is iptables based). While the standard dataplane focuses on compatibility by inter-working with kube-proxy, and your own iptables rules, the eBPF dataplane focuses on performance, latency and improving user experience with features that aren’t possible in the standard dataplane. As part of that, the eBPF dataplane replaces kube-proxy with an eBPF implementation. The main “user experience” feature is to preserve the source IP of traffic from outside the cluster when traffic hits a NodePort; this makes the server-side logs and network policy much more useful on that path.
+
+For more details on enabling the eBPF dataplane please refer the [Calico Docs](https://docs.projectcalico.org/maintenance/ebpf/enabling-bpf).
+
+```yaml
+  kubeProxy:
+    enabled: false
+  networking:
+    calico:
+      bpfEnabled: true
+      bpfExternalServiceMode: Tunnel
+      bpfLogLevel: Info
+```
+
 ### Configuring WireGuard
 {{ kops_feature_table(kops_added_default='1.19', k8s_min='1.16') }}
 
--- a/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
@@ -1,4 +1,6 @@
-# Pulled and modified from: https://docs.projectcalico.org/v3.16/manifests/calico-typha.yaml
+# Pulled and modified from:
+#   https://docs.projectcalico.org/v3.16/manifests/calico-bpf.yaml
+#   https://docs.projectcalico.org/v3.16/manifests/calico-typha.yaml
 
 ---
 # Source: calico/templates/calico-config.yaml
@@ -11,6 +13,12 @@ metadata:
   labels:
     role.kubernetes.io/networking: "1"
 data:
+  {{- if .Networking.Calico.BPFEnabled }}
+  # Set these to the IP and port of your API server; In BPF mode, we need to connect directly to the
+  # API server because we take over kube-proxy's role.
+  kubernetes_service_host: "{{ .MasterInternalName }}"
+  kubernetes_service_port: "443"
+  {{- end }}
   # You must set a non-zero value for Typha replicas below.
   typha_service_name: "{{- if .Networking.Calico.TyphaReplicas -}}calico-typha{{- else -}}none{{- end -}}"
   # Configure the backend to use.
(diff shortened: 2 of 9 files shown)


upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -179,6 +179,12 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		dest["CiliumSecret"] = func() string { return ciliumsecretString }
 	}
 
+	if cluster.Spec.Networking != nil && cluster.Spec.Networking.Calico != nil {
```
olemarkus, line 182: You don't need this one. You can just the value directly. Already doing the same for Cilium.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -769,6 +769,10 @@ func setupNetworking(opt *NewClusterOptions, cluster *api.Cluster) error {
 		cluster.Spec.Networking.Calico = &api.CalicoNetworkingSpec{
 			MajorVersion: "v3",
 		}
+		if cluster.Spec.KubeProxy == nil {
+			cluster.Spec.KubeProxy = &api.KubeProxyConfig{}
+		}
+		cluster.Spec.KubeProxy.Enabled = fi.Bool(false)
```
olemarkus, line 775: I assume this is just for e2e testing

hakman: Yeah. Checking failures, which are many.

upup/pkg/fi/cloudup/template_functions.go
hakman: Ah got it. Thanks for the tip.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -3873,7 +3914,8 @@ spec:
               value: "{{- or .Networking.Calico.IPv6AutoDetectionMethod "first-found" }}"
             # Enable IPIP
             - name: CALICO_IPV4POOL_IPIP
-              value: "{{- if and (eq .CloudProvider "aws") (.Networking.Calico.CrossSubnet) -}}CrossSubnet{{- else -}} {{- or .Networking.Calico.IPIPMode "Always" -}} {{- end -}}"
+              value: "Never"
```
fasaxc, line 3917: Think you need to remove this diff; BPF mode support IPIP mode in v3.16 and it's needed unless your nodes all live in one subnet.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -3836,6 +3863,20 @@ spec:
               name: kubernetes-services-endpoint
```
fasaxc, line 3863: You could use the `kubernetes-services-endpoint` config map that we now have for KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT.  All the components are alrady configured to pick that up.

hakman: Wouldn't this require first to move to the Operator first?

docs/networking/calico.md
```diff
@@ -93,6 +93,23 @@ It is possible to configure Calico to use Typha by editing a cluster and adding
...
+
+For more details on enabling the eBPF dataplane please refer the [Calico Docs](https://docs.projectcalico.org/maintenance/ebpf/enabling-bpf).
+
+```yaml
+  kubeProxy:
+    enabled: false
+  networking:
+    calico:
+      bpfEnabled: true
+      bpfExternalServiceMode: DSR
```
olemarkus, line 109: Does DSR actually work in AWS? Maybe where NLB is used?

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
olemarkus: Not sure you need to do this conditionally. For cilium I chose to just always do this.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -3916,6 +3963,18 @@ spec:
               value: "true"
 
             # kops additions
+            # Enable source/destination checks for AWS
+            - name: FELIX_AWSSRCDSTCHECK
+              value: "{{- if and (eq .CloudProvider "aws") (.Networking.Calico.CrossSubnet) -}}Disable{{- else -}} {{- or .Networking.Calico.AwsSrcDstCheck "DoNothing" -}} {{- end -}}"
```
olemarkus, line 3968: You could consider using a options builder for calico to avoid this much logic in the template.

hakman: Good idea, will think of doing it in a future PR.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
hakman: Because we merge logic from different manifests, it will be easier later to track why these were added.

docs/networking/calico.md
hakman: Did not run e2e with this so, not really sure, but you are right, Tunnel should be safer.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
fasaxc: > Wouldn't this require first to move to the Operator first?

No, we have the same config map in both install methods (but for operator it needs to be in the operator's namespace, not kube-system).  If you look a few lines above you can see the config map ref in your manifest ;-)

hakman: Got it. Thanks! :)

PR #10073, opened 2020-10-19
Allow more volume types

### LINKED ISSUE #10072: Root volume type 'standard' is no longer accepted since kops 1.19
 this information.**

Version 1.19.0-alpha.5 (git-921f3e7109a4871ec1cb0b0e512785533a24dbce)

 a `kops` flag.**

1.19.3 but should be reproducible with any version

AWS

Use a template with values

```
  etcdClusters:
  - etcdMembers:
      volumeType: standard
```

then create or update a cluster

spec.rootVolumeType: Unsupported value: "standard": supported values: "gp2", "io1"

Expected 'standard' to be supported as this value works in kops 1.18.x
The main issue for us is that we cannot upgrade our clusters created with 'standard' volumes if we change the spec to 'gp2':

error: error replacing cluster: [spec.etcdClusters[main].etcdMembers[etcd-eu-central-1a].volumeType: Forbidden: volumeType cannot be cha
nged, spec.etcdClusters[events].etcdMembers[etcd-eu-central-1a].volumeType: Forbidden: volumeType cannot be changed]

This is caused by https://github.com/kubernetes/kops/pull/9265 as it implements the validation but does not allow for 'standard' type

I should also mention that this issue was once fixed in https://github.com/kubernetes/kops/issues/4256

olemarkus:
Fixes #10072

### DIFF
--- a/pkg/apis/kops/validation/instancegroup.go
+++ b/pkg/apis/kops/validation/instancegroup.go
@@ -190,7 +190,7 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, cl
 	}
 
 	if g.Spec.RootVolumeType != nil && kops.CloudProviderID(cluster.Spec.CloudProvider) == kops.CloudProviderAWS {
-		allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "rootVolumeType"), g.Spec.RootVolumeType, []string{"gp2", "io1"})...)
+		allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "rootVolumeType"), g.Spec.RootVolumeType, []string{"standard", "gp2", "io1", "io2"})...)
 	}
 
 	return allErrs
--- a/pkg/apis/kops/validation/instancegroup_test.go
+++ b/pkg/apis/kops/validation/instancegroup_test.go
@@ -194,6 +194,9 @@ func TestValidBootDevice(t *testing.T) {
 		expected   []string
 	}{
 		{
+			volumeType: "standard",
+		},
+		{
 			volumeType: "gp2",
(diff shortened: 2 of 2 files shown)


rifelpet:
I'm looking at the latest version of the aws sdk and there's a function that returns a list of all EBS volume types, maybe we should use that? I don't see the function in our vendored version though, so we'll need to update it first.

(too big to link to the line # directly)
https://raw.githubusercontent.com/aws/aws-sdk-go/master/service/ec2/api.go

```go
// VolumeType_Values returns all elements of the VolumeType enum
func VolumeType_Values() []string {
	return []string{
		VolumeTypeStandard,
		VolumeTypeIo1,
		VolumeTypeIo2,
		VolumeTypeGp2,
		VolumeTypeSc1,
		VolumeTypeSt1,
	}
}
```

rifelpet:
The SDK has been updated and I updated some similar references in https://github.com/kubernetes/kops/pull/10075

olemarkus:
Note that the original reason for the validation was that `sc1` and `st1` _is not_ supported boot volumes. So I could take that slice and filter known unsupported types or I can explicitly list known good types. The `standard` type was missed from my side because most of the AWS docs doesn't mention it [example 1](https://aws.amazon.com/ebs/features/) [example 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html).

Given that setting a bad value here ends up in an unusable cluster, I'd rather do an allow list than a deny list.

rifelpet:
Ah, I missed that earlier. That makes more sense, I think we can keep it as this list of strings for now. I was wondering if we can get this info from ec2.DescribeInstanceTypes but we can explore that separately.

/approve
/lgtm

PR #10076, opened 2020-10-21
WIP - generate an acm cert in the e2e job and use it in the cluster

rifelpet:
not intending this to merge this, i'm just testing the ACM permissions on the prow infra. Long term we'll want to setup an e2e test that uses an ACM cert once we come up with a strategy for requesting them. I'm imagining kubetest2 can request the cert and pass it to the cluster via flags or manifest.


rifelpet:
hmm I was expecting either the cluster validation or the e2e tests to fail because of #9756. I'll keep looking into this tomorrow

rifelpet:
Ok, that last error looks to be what I'd expect. We can use this commit to confirm the upcoming NLB + second listener for ACM fixes work as expected.

PR #10077, opened 2020-10-21
Rebrand kops to kOps

hakman:
This should make the purpose of the project more clear.
https://deploy-preview-10077--kubernetes-kops.netlify.app

/cc @justinsb @mikesplain 
/hold for feedback


olemarkus:
This didn't immediately resonate well with me. Does this also change the pronunciation to k-ops?

hakman:
> This didn't immediately resonate well with me. Does this also change the pronunciation to k-ops?

I would say so, but not mandatory.

johngmyers:
I showed the old and suggested logo to someone with graphic design experience and got this feedback:

* The O does not read as capital.
* What is the purpose of the chevron? In the current logo it appears to incorporate a K, but that is not necessary in the new logo.
* What is the purpose of the shading?
* The kerning needs to be adjusted.

What is the typeface? Are there any particular color or design standards we are trying to follow? What are we trying to achieve (other then de-emphasizing the K)?

randomvariable:
+100 on the idea

olemarkus:
After giving this some more thought, I am also 100% on-board with this.

hakman:
Thanks @olemarkus :).
/assign @justinsb

geojaz:
@olemarkus Thanks for confirming, initially i had a similar reaction, but coming back to it, feel like this is the right direction. +1 from me too

geojaz:
@hakman are you specifically waiting for justin's approval? after the discussion in office hours, i'm happy to do it

hakman:
Thanks @geojaz. I think we all agree, so probably should merge and see how the users receive it.
/hold cancel

geojaz (approved):
I think we're ready to do this one, so i'll push the button... With the understanding that we're all trying to do the best we can with the world we live in. And if and when we need to iterate on this, we will.
/lgtm
Thanks all

PR #10080, opened 2020-10-22
WIP: Don't force APIServer --allow-privileged flag to true for k8s 1.19+

johngmyers:
We should use defaults that are less secure than upstream Kubernetes defaults, much less force settings to such less secure values.


johngmyers:
Okay, not sure why kube-apiserver defaults that off. An admission controller would be more in line with current practice and would allow policies more nuanced than "prohibit everything".

PR #10091, opened 2020-10-24
[WIP] Simplify the kops logo

hakman:
Ref: https://github.com/kubernetes/kops/pull/10077


johngmyers:
I showed the old and suggested logo to someone with graphic design experience and got this feedback:

* The O does not read as capital.
* What is the purpose of the chevron? In the current logo it appears to incorporate a K, but that is not necessary in the new logo.
* What is the purpose of the shading?
* The kerning needs to be adjusted.

What is the typeface? Are there any particular color or design standards we are trying to follow? What are we trying to achieve (other then de-emphasizing the K)?

johngmyers:
Perhaps we could use as a design element a more modern helm, without protruding spokes?

hakman:
@johngmyers anything that looks better?
kOps-logos

olemarkus:
If you want to emphasise the ops bit, maybe stylise it kOPS.

hakman:
> If you want to emphasise the ops bit, maybe stylise it kOPS.

That would look like this:
kOPS

johngmyers:
The feedback I got was:

* The kerning is much improved.
* The helm/gear graphics still aren't reading as O. There is a strong tendency to scan as one syllable and the busy graphics distract from the intent to break that.
* The gear element is appropriate for "operations"

I suggested using the gear as an O and got "...maybe."

I suggested using a punctuation graphic between the k and the O and got a better reception. "Perhaps a nonagon dot"

hakman:
Thanks for the feedback @johngmyers. Will do one more iteration.

hakman:
kOps-logo

hakman:
@justinsb will look into this, but will reopen if ever needed.
/close

Issue #10081, opened 2020-10-22
Cluster not validated when cilium.enablePrometheusMetrics is set to true


linecolumn:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Version 1.19.0-alpha.5 (git-921f3e7109a4871ec1cb0b0e512785533a24dbce)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

v1.18.9

**3. What cloud provider are you using?**

AWS

I'm testing KOPS 1.19 alpha5 and cluster is not validated when I set enablePrometheusMetrics under cilium sections:

```
  networking:
    cilium:
       enablePrometheusMetrics: true
```

As soon as I remove `enablePrometheusMetrics` KOPS can validate cluster.

I'm generating Terraform file(s) and then apply it for cluster creation.

Terraform v0.12.24

olemarkus:
Hi,

Can you please follow the issue template here: https://github.com/kubernetes/kops/blob/master/.github/ISSUE_TEMPLATE/bug_report.md

In particular, we would need to know what validation errors you are getting.

linecolumn:
Here are addition information.

Validating cluster test2.redacted.com

```
INSTANCE GROUPS
NAME			ROLE	MACHINETYPE	MIN	MAX	SUBNETS
test-nodes		Node	t3.medium	1	7	eu-central-1a
master-eu-central-1a	Master	t3.medium	1	1	eu-central-1a
test2-nodes		Node	t3.medium	1	7	eu-central-1a
spot-ig			Node	t3.medium	1	3	eu-central-1a

NODE STATUS
NAME						ROLE	READY
ip-172-17-35-232.eu-central-1.compute.internal	node	False
ip-172-17-37-37.eu-central-1.compute.internal	master	False
ip-172-17-42-176.eu-central-1.compute.internal	node	False
ip-172-17-51-31.eu-central-1.compute.internal	node	False

VALIDATION ERRORS
KIND	NAME						MESSAGE
Node	ip-172-17-35-232.eu-central-1.compute.internal	node "ip-172-17-35-232.eu-central-1.compute.internal" is not ready
Node	ip-172-17-37-37.eu-central-1.compute.internal	master "ip-172-17-37-37.eu-central-1.compute.internal" is not ready
[...]

PR #10094, opened 2020-10-24
Fix cilium template by specifying boolean as a string for enable-metrics

h3poteto:
Fixes: #10081 

Cilium ConfigMap template was wrong, so I fixed it.
I confirmed that kops can successfully create clusters using this change when `enablePrometheusMetrics` is specified  for Cilium.

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -4646,8 +4646,8 @@ data:
   # NOTE that this will open the port on ALL nodes where Cilium pods are
   # scheduled.
   prometheus-serve-addr: ":{{ .AgentPrometheusPort }}"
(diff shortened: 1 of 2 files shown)


olemarkus:
You can add
Fixes #10081 
to your initial comment to properly link and auto-close the issue upon merge.

olemarkus:
Thanks for catching this. Could now replicate and confirm the bug. I changed the config on an already-working cluster, which is why I couldn't reproduce earlier.

/lgtm

hakman:
Thanks @h3poteto .
@olemarkus Could you cherry-pick this?
/approve

h3poteto:
Thank you for reviewing.

PR #10098, opened 2020-10-24
[Digital Ocean] Use Debian10 as default image

srikiz:
Cilium leverages and builds on the kernel BPF functionality as well as various subsystems which integrate with BPF. Therefore, host systems are required to run Linux kernel version 4.9.17 or later to run a Cilium agent.

Using Debian10 should ensure we have the latest kernel that Cilium expects.

### DIFF
--- a/upup/pkg/fi/cloudup/populate_instancegroup_spec.go
+++ b/upup/pkg/fi/cloudup/populate_instancegroup_spec.go
@@ -43,7 +43,7 @@ const (
 	defaultMasterMachineTypeDO  = "s-2vcpu-4gb"
 	defaultMasterMachineTypeALI = "ecs.n2.medium"
 
-	defaultDONodeImage  = "debian-9-x64"
+	defaultDONodeImage  = "debian-10-x64"
 	defaultALINodeImage = "centos_7_04_64_20G_alibase_201701015.vhd"
 )
 
--- a/upup/pkg/fi/cloudup/populate_instancegroup_spec.go
+++ b/upup/pkg/fi/cloudup/populate_instancegroup_spec.go
@@ -43,7 +43,7 @@ const (
 	defaultMasterMachineTypeDO  = "s-2vcpu-4gb"
 	defaultMasterMachineTypeALI = "ecs.n2.medium"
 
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/populate_instancegroup_spec.go
```diff
@@ -43,7 +43,7 @@ const (
 	defaultMasterMachineTypeDO  = "s-2vcpu-4gb"
 	defaultMasterMachineTypeALI = "ecs.n2.medium"
 
-	defaultDONodeImage  = "debian-9-x64"
+	defaultDONodeImage  = "debian-10-x64"
```
hakman, line 46: One of the big moves in Kops 1.18 was switching to Ubuntu 20.04 as default image for the foreseeable future. Would it be a good idea to do the same for DO?
```suggestion
	defaultDONodeImage  = "ubuntu-20-04-x64"
```

olemarkus:
Especially since citing cilium as the reason, I would go for Ubuntu. Debian uses nftables by default and cilium doesn't support that. Users need various workarounds for switching to iptables legacy to avoid having both legacy and nftables rules at play.

upup/pkg/fi/cloudup/populate_instancegroup_spec.go
srikiz: @timoreimann - Is it okay to switch to ubuntu 20.04 as default image for DO?

timoreimann: Yeah totally fine with me, thanks for asking.

hakman: Thanks @timoreimann. @srikiz could you change the PR and cherrypick to 1.19 branch? There is a beta release for 1.19 coming soon.

hakman:
Thanks @srikiz.
/lgtm
/approve

srikiz:
Thanks @hakman for cherry-picking this into release-1.19 !

hakman:
😉

PR #10101, opened 2020-10-25
Prepare for version 1.20

johngmyers:


permalinks/upgrade_k8s.md
```diff
@@ -8,6 +8,7 @@ Kops will drop support for Kubernetes versions as follows:
 | 1.18         | 1.8 and below                        |
 | 1.19         | 1.9 and 1.10                         |
 | 1.20         | 1.11 and 1.12                        |
+| 1.21         | 1.13 and 1.14                        |
```
hakman, line 11: Maybe we should add another column "Deprecates support for Kubernetes version" and use "Removes" instead of "Drops" in first, to match the release notes wording.

hakman:
Looks fine to me. @rifelpet, maybe you want to take a look also.

rifelpet:
The 1.19 release notes currently mention that only `Terraform-0.12` will be removed, it doesn't mention that `TerraformJSON` will be removed, so we'd need to keep `target_0_11.go` and the `MinimalJSON` integration test unless we decide to also deprecate and remove `TerraformJSON`.

johngmyers:
@rifelpet `TerraformJSON` does nothing unless `Terraform-0.12` is off. So it doesn't make sense to keep `TerraformJSON` as a no-op.

rifelpet:
I was thinking `TerraformJSON` would imply the current `TerraformJSON,-Terraform-0.12` behavior but `-Terraform-0.12` on its own would no longer be valid. #9940 implements much of this so it might be better to defer those changes to that PR. I agree that removing a feature flag that defaults to off shouldn't require a deprecation notice but it might still be helpful to get user feedback, given that the Terraform JSON output is still a valid use-case for Terraform 0.12+.

johngmyers:
I created #10106 to effect that change in the `TerraformJSON` semantics.

rifelpet:
I believe this is ready for a rebase and merge.

PR #10102, opened 2020-10-25
Remove hack/workaround from etcd-manager certificate expiration advisory

hakman:
The hack/workaround works only on some versions so best not to have it at all. Also, I think `--cloudonly` is needed for the rolling-update to work.
We could add some hack/workaround in the issue, if we want to share the link with someone in Slack at some point.

https://deploy-preview-10102--kubernetes-kops.netlify.app/advisories/etcd-manager-certificate-expiration/#solution

/cc @rifelpet @olemarkus


olemarkus (approved):

rifelpet:
Yes I agree its probably best for us to suggest the longer term solution

/lgtm

PR #10111, opened 2020-10-26
Add nodeLocalDNSCache.kubeDnsOnly option

javipolo:
## Add a kubeDnsOnly option for nodeLocalDNScache

Add ability to use kube-dns as upstream on node-local-dns, instead of using systems default in resolv.conf

With this flag enabled, even if we use the node-local-dns to cache responses, we use as upstream the main kube-dns server for all types of requests, and not only the ones of *.cluster.local

This way we can customize the main kube-dns for external domains and make this changes also effective to the node-local-dns nodes

### Example use case
In our case, we're using rewrites on coredns to easily route some requests to internal services without exiting the cluster, and without modifying the application. With this flag we can keep this behaviour and still use node-local-dns:

```
rewrite name exact queue.myapp.com queue-system.myapp.svc.cluster.local
```

A dns request of `queue.myapp.com` to coredns will return the internal IP address of the service `queue-system` in `myapp` namespace

So far when enabling node-local-dns in kops this behaviour dissapeared and `queue.myapp.com` was resolving to an external internet IP address.
With this patch we can decide if to route all traffic to internal kube-dns, or keep the actual behaviour

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -603,6 +603,8 @@ NodeLocal DNSCache can be enabled if you are using CoreDNS. It is used to improv
 
 `memoryRequest` and `cpuRequest` for the `node-local-dns` pods can also be configured. If not set, they will be configured by default to `5Mi` and `25m` respectively.
 
+If `forwardToKubeDNS` is enabled, kubedns will be used as a default upstream
+
 ```yaml
 spec:
   kubeDNS:
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -418,6 +418,8 @@ type NodeLocalDNSConfig struct {
 	Enabled *bool `json:"enabled,omitempty"`
 	// Local listen IP address. It can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP.
 	LocalIP string `json:"localIP,omitempty"`
+	// If enabled, nodelocal dns will use kubedns as a default upstream
+	ForwardToKubeDNS *bool `json:"forwardToKubeDNS,omitempty"`
 	// MemoryRequest specifies the memory requests of each node-local-dns container in the daemonset. Default 5Mi.
 	MemoryRequest *resource.Quantity `json:"memoryRequest,omitempty"`
 	// CPURequest specifies the cpu requests of each node-local-dns container in the daemonset. Default 25m.
--- a/pkg/apis/kops/v1alpha2/cluster.go
+++ b/pkg/apis/kops/v1alpha2/cluster.go
@@ -420,6 +420,8 @@ type NodeLocalDNSConfig struct {
 	Enabled *bool `json:"enabled,omitempty"`
 	// Local listen IP address. It can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP.
 	LocalIP string `json:"localIP,omitempty"`
+	// If enabled, nodelocal dns will use kubedns as a default upstream
+	ForwardToKubeDNS *bool `json:"forwardToKubeDNS,omitempty"`
 	// MemoryRequest specifies the memory requests of each node-local-dns container in the daemonset. Default 5Mi.
 	MemoryRequest *resource.Quantity `json:"memoryRequest,omitempty"`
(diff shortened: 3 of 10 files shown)


pkg/apis/kops/cluster.go
```diff
@@ -419,6 +419,8 @@ type NodeLocalDNSConfig struct {
 	// Local listen IP address. It can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP.
 	LocalIP string `json:"localIP,omitempty"`
 	// MemoryRequest specifies the memory requests of each node-local-dns container in the daemonset. Default 5Mi.
```
olemarkus, line 421: Copy pasted a bit too much?

upup/models/bindata.go
```diff
@@ -19193,7 +19193,13 @@ data:
         reload
         loop
         bind {{ KubeDNS.NodeLocalDNS.LocalIP }}
+        {{- if KubeDNS.NodeLocalDNS.KubeDnsOnly }}
```
olemarkus, line 19196: Instead of this, you can probably simplify the config by consolidating "." with the two sections above. I.e you only have "." and cluster domain in the end.

docs/cluster_spec.md
```diff
@@ -603,6 +603,8 @@ NodeLocal DNSCache can be enabled if you are using CoreDNS. It is used to improv
 
 `memoryRequest` and `cpuRequest` for the `node-local-dns` pods can also be configured. If not set, they will be configured by default to `5Mi` and `25m` respectively.
 
+If `kubeDnsOnly` is enabled, it will not use reslov.conf as fallback for domains outside of the cluster and will always query kube-dns
```
olemarkus, line 606: I am not that happy about the description here. It is more correct to say nodelocal dns is using kubedns as default upstream.

javipolo: Happy to change it

upup/models/bindata.go
javipolo: Sure, I wanted to make "a small change" but I totally see your point

javipolo:
@olemarkus I did the proposed changes :+1:

pkg/apis/kops/cluster.go
```diff
@@ -418,6 +418,8 @@ type NodeLocalDNSConfig struct {
 	Enabled *bool `json:"enabled,omitempty"`
 	// Local listen IP address. It can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP.
 	LocalIP string `json:"localIP,omitempty"`
+	// If enabled, nodelocal dns will use kubedns as a default upstream
+	KubeDnsOnly *bool `json:"kubeDnsOnly,omitempty"`
```
johngmyers, line 422:
```suggestion
	KubeDNSOnly *bool `json:"kubeDNSOnly,omitempty"`
```

pkg/apis/kops/v1alpha2/cluster.go
```diff
@@ -420,6 +420,8 @@ type NodeLocalDNSConfig struct {
 	Enabled *bool `json:"enabled,omitempty"`
 	// Local listen IP address. It can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP.
 	LocalIP string `json:"localIP,omitempty"`
+	// If enabled, nodelocal dns will use kubedns as a default upstream
+	KubeDnsOnly *bool `json:"kubeDnsOnly,omitempty"`
```
johngmyers, line 424:
```suggestion
	KubeDNSOnly *bool `json:"kubeDNSOnly,omitempty"`
```

johngmyers (changes requested):
Go convention is to have acronyms in all-caps.

hakman:
@johngmyers Not quite sure the field name is appropriate. Any thoughts?

johngmyers:
You have a point. It's more like `ForwardToKubeDNS`.

javipolo:
I saw "Dns" somewhere in the code so I thought it was OK
Anyways I'm pretty happy to change both the acronym and the name of the field, to be honest finding a proper name is the hardest task haha

Here's the new PR

jorge07:
Thanks for the work Javi and folks

PR #10115, opened 2020-10-27
Update Calico config for eBPF mode

hakman:
Based on discussion with @fasaxc in Slack I changed the Calico manifest to use an optional ConfigMap to set the k8s endpoints.
Additionally I added an option for the cleanup of iptables rules (in case it's need for migration) and updated the version to 3.16.4.

/cc @seh @olemarkus


seh:
Thank you for this one, @hakman. Is it correct that for now, even when Calico's eBPF mode is enabled, we still run _kube-proxy_, or do we automatically disable those static pods when we're in that mode?

hakman:
For now `kube-proxy` is always enabled unless `spec.kubeProxy.enabled: false` and rolling-update is done.

seh:
Do you envision ever tying these settings together in _kops_ automatically? That is, if Calico is in eBPF mode, then the default value for "spec.kubeProxy.enabled" would become false?

hakman:
Can't say never, but you probably understand the complications of doing this.
Cilium can coexist with `kube-proxy`, so there is no problem there. Not sure if Calico can do that also in the future.

Issue #10108, opened 2020-10-26
[v1.18.2] Release notes reference non-existant Rolling Updates link 


timothyclarke:
**1. What `kops` version are you running? **
v1.18.2

**2. What Kubernetes version are you running?**
 N/A

**3. What cloud provider are you using?**
N/A

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
Browse to [v1.18.2 release notes](https://github.com/kubernetes/kops/releases/tag/v1.18.2) Then follow documentation link for Rolling updates now support surging and parallelism within an instance group. For details see [the documentation](https://github.com/kubernetes/kops/blob/operations/rolling-update.md).

**5. What happened after the commands executed?**
Github 404

**6. What did you expect to happen?**
The documentation to be distplayed

**7. Please provide your cluster manifest.**
 N/A

**8. Please run the commands with most verbose logging by adding the `-v 10` flag.
  Paste the logs into this report, or in a gist and provide the gist link here.**
N/A

**9. Anything else do we need to know?**

hakman:
Thanks for reporting this @timothyclarke. I updated the release notes in GitHub with proper links to the docs site.

timothyclarke:
Thank you

PR #10118, opened 2020-10-27
Fix release notes links to point to https://kops.sigs.k8s

hakman:
Fixes: #10108


rifelpet:
ah, since release notes get included in GH releases in addition to the docs site, its probably best to always have links in the release notes be full URLs

/lgtm

PR #10124, opened 2020-10-27
Update AWS VPC CNI to 1.7.5

MoShitrit:
In this PR:
- Update AWS VPC image on main and init container to v.1.7.5
- Adding `apps` `apiGroup` to handle `list` and `watch` `daemonsets`, since `extensions` is no more (we should remove `extensions` at some point, but I figured I would still leave it here since the [original manifest](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.7/aws-k8s-cni-cn.yaml#L47) still contains it).
- Adding `xtables-lock` volume and volumeMount which were introduced in [v1.7.1](https://github.com/aws/amazon-vpc-cni-k8s/pull/1187) as a fix for `iptables`
- Adding `DISABLE_TCP_EARLY_DEMUX` toggle as introduced in [v1.7.3](https://github.com/aws/amazon-vpc-cni-k8s/pull/1222)

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -4379,7 +4379,7 @@ rules:
   resources:
   - nodes
   verbs: ["list", "watch", "get", "update"]
-- apiGroups: ["extensions"]
+- apiGroups: ["extensions", "apps"]
   resources:
   - daemonsets
   verbs: ["list", "watch"]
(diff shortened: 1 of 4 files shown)


hakman:
Thanks @MoShitrit !
/ok-to-test
/cc @rifelpet

rifelpet:
hm IAM issues?
/retest

hakman:
Seems like we have a winner here.
/lgtm
/approve

Issue #10119, opened 2020-10-27
official manual has mistakes in syntax bash 


ttsrg:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Version 1.18.1 (git-453d7d96be)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
v1.19.3
**3. What cloud provider are you using?**
aws
**4. What commands did you run?  What is the simplest way to reproduce this issue?**
kops create cluster --help and I've run:
 export NODE_SIZE=${NODE_SIZE:-m4.large}
  export MASTER_SIZE=${MASTER_SIZE:-m4.large}
  export ZONES=${ZONES:-"us-east-1d,us-east-1b,us-east-1c"}

```
export NODE_SIZE=${NODE_SIZE:-m4.large}
echo $NODE_SIZE
m4.large
```

then I've tried change data:
```
export NODE_SIZE=${NODE_SIZE:-t3.small}
echo $NODE_SIZE
m4.large
```

**5. What happened after the commands executed?**
variables in export have had  constant values

**6. What did you expect to happen?**
I've got variables without redefine values.
I propose to use classic syntax :
export ZONES="us-east-1d,us-east-1b,us-east-1c"  instead of the export ZONES=${ZONES:-"us-east-1d,us-east-1b,us-east-1c"}

and change information in official docs about non-use such construction.

hakman:
@ttsrg This is related to Bash version (or other shell in general). See https://tldp.org/LDP/abs/html/parameter-substitution.html for explanation of how `${parameter:-default}` works.
What shell version and OS are you using?

ttsrg:
Hello.
I'm using Ubuntu 20.04.1 LTS and bash, version 5.0.17(1).

hakman:
I have no idea why was done like that 4 years ago. Maybe BASH evolved also and TBH, it's not the most stable syntax over time. It is a good idea to simplify things though.

ttsrg:
Then I propose to apply my proposal and I guess  it  improves official --help.
Thanks

PR #10128, opened 2020-10-28
Update shell style for CLI docs for better compatibility

hakman:
Relevant changes are in the `Update shell style for CLI docs for better compatibility` commit.
Also makes cluster and state store names more uniform.

Fixes: #10119

### DIFF
--- a/cmd/kops/update.go
+++ b/cmd/kops/update.go
@@ -32,7 +32,7 @@ var (
 
 	updateExample = templates.Examples(i18n.T(`
 		# After cluster has been created, configure it with:
-		kops update cluster k8s.cluster.site --yes --state=s3://kops-state-1234
+		kops update cluster k8s-cluster.example.com --yes --state=s3://my-state-store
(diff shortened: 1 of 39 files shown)


cmd/kops/update.go
```diff
@@ -32,7 +32,7 @@ var (
 
 	updateExample = templates.Examples(i18n.T(`
 		# After cluster has been created, configure it with:
-		kops update cluster k8s.cluster.site --yes --state=s3://kops-state-1234
+		kops update cluster k8s-aws-cluster.cluster.site --yes --state=s3://my-state-store
```
rifelpet, line 35: might as well make this consistent with every other command's examples
```suggestion
		kops update cluster k8s-cluster.example.com --yes --state=s3://my-state-store
```

hakman: Thought I did :)

hakman: Done.

PR #10138, opened 2020-10-29
ELB/TargetGroup/ASG attachment fixes

rifelpet:
ref: https://github.com/kubernetes/kops/pull/10134#issuecomment-718399054

This addresses three issues:
* External TargetGroupARNs werent being attached to ASGs at creation-time with `--target direct`
* Kops was passing the name of the API ELB task to attach to the ASG rather than the resulting AWS LoadBalancerName which gets truncated and suffixed. This was resulting in AWS errors when creating the ASG, reporting that the ELB is not found.
* Subsequent `update cluster`s were reporting changes because when `Find`ing the LBs attached to the ASG, we couldn't tell which was the API ELB and which were external ELBs passed in through the InstanceGroupSpec. Find now looks for the API ELB to decide whether an ELB attached to the ASG should be the API ELB task or an external ELB task.

### DIFF
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -364,7 +364,10 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 
 	for _, extLB := range ig.Spec.ExternalLoadBalancers {
 		if extLB.LoadBalancerName != nil {
-			t.LoadBalancers = append(t.LoadBalancers, &awstasks.LoadBalancer{Name: extLB.LoadBalancerName})
+			t.LoadBalancers = append(t.LoadBalancers, &awstasks.LoadBalancer{
+				Name:             extLB.LoadBalancerName,
+				LoadBalancerName: extLB.LoadBalancerName,
+			})
 
 			c.AddTask(&awstasks.LoadBalancer{
 				Name:   extLB.LoadBalancerName,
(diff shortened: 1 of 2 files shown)


rifelpet:
@hakman can you test this commit out?

hakman:
> @hakman can you test this commit out?

No luck. Same thing.

rifelpet:
@hakman did some troubleshooting and discovered that we used to be correctly passing the AWS LoadBalancerName when attaching LBs to ASGs. the LoadBalancerName is the truncated name we pass in CreateLoadBalancer and gets a random suffix.

https://github.com/kubernetes/kops/blob/f3ea21a8c6110d52d6e92f5a139511dcc689a648/upup/pkg/fi/cloudup/awstasks/load_balancer_attachment.go#L64-L66

#9794 changed that such that we now try to pass the original Name (the kops task name) to the Attach call which fails.

https://github.com/kubernetes/kops/blob/b7f66a6d98acf7f37b5d45780b2b2fedbc577436/pkg/model/awsmodel/autoscalinggroup.go#L356-L358

https://github.com/kubernetes/kops/blob/b7f66a6d98acf7f37b5d45780b2b2fedbc577436/pkg/model/names.go#L88-L91

https://github.com/kubernetes/kops/blob/b7f66a6d98acf7f37b5d45780b2b2fedbc577436/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go#L279-L281

So we need to update the ASG task to try to use the LoadBalancerName field of the LB task, fetching it from AWS if we don't yet know it. My last commit in this PR aims to accomplish that.

rdrgmnzs:
/lgtm

Thanks for the fix @rifelpet and for catching the issue @hakman

PR #10142, opened 2020-10-30
WIP Open ELB to kops-controller port when using it for internal API

### LINKED ISSUE #10139: Enabling "spec.api.loadBalancer.useForInternalApi" requires access to kops controller port through API load balancer
 this information.**

Version 1.19.0-alpha.5 (git-921f3e7109a4871ec1cb0b0e512785533a24dbce)

 a `kops` flag.**

1.19.3

AWS

1. _kops edit cluster_
1. Set the _Cluster_ manifest's ["spec.api.loadBalancer.useForInternalApi" field](https://github.com/kubernetes/kops/blob/v1.18.2/pkg/apis/kops/cluster.go#L363-L364) to true in an earlier version of _kops_, such as version 1.18.2.
1. Use _kops update cluster_ and _kops rolling-update cluster_ to update the DNS record in Route 53 and get all the Kubernetes cluster's nodes to use it.
1. Upgrade the master machines to _kops_ version 1.19.0-alpha.5.
1. Create a new EC2 instance, intending for it to join the Kubernetes cluster as a **worker node** (not a master node).
1. Observe that the _kops-configuration_ systemd service (or the _nodeup_ program) can't finish its work, because it can't connect to the _kops_ controller on the masters on TCP port 3988.  
  The error message looks like this:
  `error running task “BootstrapClient/BootstrapClient” (3m27s remaining to succeed): Post “https://api.internal.redacted-cluster-name::3988/bootstrap”: dial tcp 3.20.73.237:3988: connect: connection timed out`
1. Observe that the new machine can't join the cluster as a node.

_kops_ updated its machine bootstrapping scripts to attempt to connect to any of the master machines by way of the "internal" API server DNS record on port 3988. However, our API server load balancer (for now, a Classic ELB) has only one listener on port 443. It rejects inbound connections on port 3988.

Past that, even if you add another ELB listener on TCP port 3988, forwarding on to port 3988 on the target instances, you then run into the firewalls defined by our security group rules. The masters accept inbound traffic on port 3988 from the worker nodes in the cluster, but not from the load balancer.

I added a rule to the security group used by the master machines to allow traffic in from the load balancer on port 3988, but that still wasn't enough. The bootstrap procedure on the new worker machine continued trying to connect to the masters on port 3988 through the load balancer, but it never succeeded.

Either setting the _Cluster_ manifest's "spec.api.loadBalancer.useForInternalApi" field to true would induce _kops_ to adjust the API server load balancer's listeners and related security group rules to allow this traffic through or, better yet, use a different DNS record for this internal traffic on ports other than 443 (or whichever port we choose to serve the Kubernetes API).

<details>
<summary>Cluster manifest</summary>

```yaml
apiVersion: kops/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2020-10-29T20:45:51Z"
  name: redacted-cluster-name
spec:
  additionalSans:
  - api.internal.redacted-internal-domain
  api:
    loadBalancer:
      additionalSecurityGroups:
      - sg-0ef9de1ab14bee565
      crossZoneLoadBalancing: true
      idleTimeoutSeconds: 3600
      type: Public
      useForInternalApi: true
  authorization:
    rbac: {}
  channel: stable
  cloudConfig:
    disableSecurityGroupIngress: true
  cloudProvider: aws
  configBase: s3://redacted-kops-state/redacted-cluster-name
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-us-east-2a
      name: a
    - instanceGroup: master-us-east-2b
      name: b
    - instanceGroup: master-us-east-2c
      name: c
    name: main
  - etcdMembers:
    - instanceGroup: master-us-east-2a
      name: a
    - instanceGroup: master-us-east-2b
      name: b
    - instanceGroup: master-us-east-2c
      name: c
    name: events
  iam:
    allowContainerRegistry: true
    legacy: false
  kubeAPIServer:
    featureGates:
      EphemeralContainers: "true"
  kubeProxy:
    proxyMode: ipvs
  kubelet:
    anonymousAuth: false
    authenticationTokenWebhook: true
    authorizationMode: Webhook
    featureGates:
      EphemeralContainers: "true"
  kubernetesVersion: 1.19.3
  masterInternalName: api.internal.redacted-cluster-name
  masterPublicName: api.redacted-cluster-name
  metricsServer:
    enabled: true
  networkCIDR: 10.3.0.0/16
  networkID: vpc-077913405bb66aac2
  networking:
    calico:
      crossSubnet: true
      typhaReplicas: 3
  subnets:
  - cidr: 10.3.100.0/22
    id: subnet-081bbc33a2014ac94
    name: utility-us-east-2a
    type: Utility
    zone: us-east-2a
  - cidr: 10.3.104.0/22
    id: subnet-028e96931b1a7e0db
    name: utility-us-east-2b
    type: Utility
    zone: us-east-2b
  - cidr: 10.3.108.0/22
    id: subnet-07d72675b6de4221d
    name: utility-us-east-2c
    type: Utility
    zone: us-east-2c
  - cidr: 10.3.0.0/22
    egress: nat-003b528448366f901
    id: subnet-08e5f928a1aa92d13
    name: us-east-2a
    type: Private
    zone: us-east-2a
  - cidr: 10.3.4.0/22
    egress: nat-04cbc0b316dc7407f
    id: subnet-0ba5ffa2b5a4383a4
    name: us-east-2b
    type: Private
    zone: us-east-2b
  - cidr: 10.3.8.0/22
    egress: nat-017dff05abb092a02
    id: subnet-0b3b58fecd5dd7373
    name: us-east-2c
    type: Private
    zone: us-east-2c
  topology:
    dns:
      type: Public
    masters: private
    nodes: private
```
</details>

Beyond the error message above found in the _journalctl_ output, I don't have anything else to share here.

There are some serious gymnastics I have to go through to roll out the change the "spec.api.loadBalancer.useForInternalApi" field from false to true. It involves generating the Terraform configuration, importing the existing "internal" DNS record as a Terraform resource, tainting that new resource, applying the configuration, but then while _kops rolling-update cluster_ is running, in the background I have to run _terraform apply -target \<internal dns record address\>_ repeatedly until _kops rolling-update cluster_ finishes, because _kops_ will keep updating the DNS record with new IP addresses of the master servers as they arrive.

I can share the shell code fragments for this if you want more detail. Though I've gotten this to work, it's fragile during the upgrade. I think this feature may be too hard to use, especially if not enabled at cluster creation time.

johngmyers:
Fixes #10139

### DIFF
--- a/pkg/model/awsmodel/api_loadbalancer.go
+++ b/pkg/model/awsmodel/api_loadbalancer.go
@@ -21,6 +21,8 @@ import (
 	"sort"
 	"time"
 
+	"k8s.io/kops/pkg/wellknownports"
+
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
@@ -110,6 +112,10 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 
 		if lbSpec.SSLCertificate != "" {
 			listeners["443"] = &awstasks.LoadBalancerListener{InstancePort: 443, SSLCertificateID: lbSpec.SSLCertificate}
+		}
+
+		if b.UseKopsControllerForNodeBootstrap() && b.Cluster.Spec.API != nil && b.Cluster.Spec.API.LoadBalancer != nil && b.Cluster.Spec.API.LoadBalancer.UseForInternalApi {
+			listeners[fmt.Sprintf("%d", wellknownports.KopsControllerPort)] = &awstasks.LoadBalancerListener{InstancePort: wellknownports.KopsControllerPort}
 		}
 
 		if lbSpec.SecurityGroupOverride != nil {
@@ -262,6 +268,40 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 				SecurityGroup: masterGroup.Task,
 				SourceGroup:   lbSG,
 				ToPort:        fi.Int64(443),
+			})
+		}
+	}
+
+	// Allow Kops-controller from nodes to ELB, thence to the master instances.
+	if b.UseKopsControllerForNodeBootstrap() && b.Cluster.Spec.API != nil && b.Cluster.Spec.API.LoadBalancer != nil && b.Cluster.Spec.API.LoadBalancer.UseForInternalApi {
+		nodeGroups, err := b.GetSecurityGroups(kops.InstanceGroupRoleNode)
+		if err != nil {
+			return err
+		}
+
+		for _, nodeGroup := range nodeGroups {
+			suffix := nodeGroup.Suffix
+			c.AddTask(&awstasks.SecurityGroupRule{
+				Name:          fi.String(fmt.Sprintf("kops-controller-node%s-to-elb", suffix)),
+				Lifecycle:     b.SecurityLifecycle,
+				FromPort:      fi.Int64(wellknownports.KopsControllerPort),
+				Protocol:      fi.String("tcp"),
+				SecurityGroup: lbSG,
+				SourceGroup:   nodeGroup.Task,
+				ToPort:        fi.Int64(wellknownports.KopsControllerPort),
+			})
+		}
+
+		for _, masterGroup := range masterGroups {
+			suffix := masterGroup.Suffix
+			c.AddTask(&awstasks.SecurityGroupRule{
+				Name:          fi.String(fmt.Sprintf("kops-controller-elb-to-master%s", suffix)),
+				Lifecycle:     b.SecurityLifecycle,
+				FromPort:      fi.Int64(wellknownports.KopsControllerPort),
+				Protocol:      fi.String("tcp"),
+				SecurityGroup: masterGroup.Task,
+				SourceGroup:   lbSG,
+				ToPort:        fi.Int64(wellknownports.KopsControllerPort),
 			})
 		}
 	}
--- a/pkg/model/awsmodel/api_loadbalancer.go
+++ b/pkg/model/awsmodel/api_loadbalancer.go
@@ -19,6 +19,7 @@ package awsmodel
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"time"
 
 	"k8s.io/kops/pkg/wellknownports"
@@ -115,7 +116,7 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 
 		if b.UseKopsControllerForNodeBootstrap() && b.Cluster.Spec.API != nil && b.Cluster.Spec.API.LoadBalancer != nil && b.Cluster.Spec.API.LoadBalancer.UseForInternalApi {
-			listeners[fmt.Sprintf("%d", wellknownports.KopsControllerPort)] = &awstasks.LoadBalancerListener{InstancePort: wellknownports.KopsControllerPort}
+			listeners[strconv.Itoa(wellknownports.KopsControllerPort)] = &awstasks.LoadBalancerListener{InstancePort: wellknownports.KopsControllerPort}
 		}
 
 		if lbSpec.SecurityGroupOverride != nil {
--- a/pkg/model/awsmodel/BUILD.bazel
+++ b/pkg/model/awsmodel/BUILD.bazel
@@ -18,6 +18,7 @@ go_library(
         "//pkg/model/defaults:go_default_library",
         "//pkg/model/iam:go_default_library",
         "//pkg/model/spotinstmodel:go_default_library",
+        "//pkg/wellknownports:go_default_library",
         "//upup/pkg/fi:go_default_library",
         "//upup/pkg/fi/cloudup/awstasks:go_default_library",
         "//upup/pkg/fi/fitasks:go_default_library",


seh:
I built _kops_ including this patch, and it worked as intended: my worker machines were able to contact _kops-controller_ on port 3988 and finish their bootstrap procedure.

seh:
> I built _kops_ including this patch, and it worked as intended: my worker machines were able to contact _kops-controller_ on port 3988 and finish their bootstrap procedure.

I spoke too soon. It seemed that this worked earlier, but then failed again in my current test. What I observed was that opening up the ELB's security group to allow ingress on port 3988 from the "nodes" security group wasn't good enough. Opening ingress on port 3988 from **everywhere** did work.

I don't understand why this is so. The source machine—one of my worker machines—is a member of the "nodes" security group. It also a member of another security group, but—not that I expected this to work—when I also tried allowing ingress to the ELB from that other security group, it didn't change the outcome. Only opening up the ELB to ingress from **all sources** has worked so far.

I can't tell if there's some SNAT going on here that's confusing AWS's ability to tell that the incoming traffic (that is, from a worker machine to the ELB) is coming from a blessed security group.

seh:
I enabled access logs on the ELB, but the client IP addresses only show the IP addresses of the ELB listeners, which isn't helpful. There are a few clients with an address like _18.188.58.156_ which does not look like one from any of my VPCs.

I tried replacing the ELB security group rule allowing ingress from anywhere with one allowing ingress from the same ELB security group itself, just to see if this firewall enforcement matches what the ELB logs show. That didn't work, though: traffic was still blocked from the worker machines.

I'm going to take a break and get some sleep, and see if reasonable explanation comes to me—as usually occurs the moment I turn off the computer.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -266,6 +272,40 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+		if err != nil {
+			return err
+		}
+
+		for _, nodeGroup := range nodeGroups {
+			suffix := nodeGroup.Suffix
+			c.AddTask(&awstasks.SecurityGroupRule{
+				Name:          fi.String(fmt.Sprintf("kops-controller-node%s-to-elb", suffix)),
+				Lifecycle:     b.SecurityLifecycle,
+				FromPort:      fi.Int64(wellknownports.KopsControllerPort),
```
hakman, line 287: Any reason why the source port is set? I would think this one would be random.

seh: I think "from" and "to" here just means "beginning" and "end" of the port range, inclusive on both ends. It's confusing phrasing, no doubt.

hakman: Damn... I always miss this part. Thanks for the reminder @seh .

hakman:
Your problem seems to be that the ELB is public and the SGs will not help too much with that. Probably `18.188.58.156` is VPC NAT GW addrress.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -112,6 +114,10 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 			listeners["443"] = &awstasks.LoadBalancerListener{InstancePort: 443, SSLCertificateID: lbSpec.SSLCertificate}
 		}
 
+		if b.UseKopsControllerForNodeBootstrap() && b.Cluster.Spec.API != nil && b.Cluster.Spec.API.LoadBalancer != nil && b.Cluster.Spec.API.LoadBalancer.UseForInternalApi {
+			listeners[fmt.Sprintf("%d", wellknownports.KopsControllerPort)] = &awstasks.LoadBalancerListener{InstancePort: wellknownports.KopsControllerPort}
```
seh, line 118: Consider using `strconv.FormatInt` instead of `fmt.Sprintf` here.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -266,6 +272,40 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+
+		for _, nodeGroup := range nodeGroups {
+			suffix := nodeGroup.Suffix
+			c.AddTask(&awstasks.SecurityGroupRule{
+				Name:          fi.String(fmt.Sprintf("kops-controller-node%s-to-elb", suffix)),
+				Lifecycle:     b.SecurityLifecycle,
+				FromPort:      fi.Int64(wellknownports.KopsControllerPort),
+				Protocol:      fi.String("tcp"),
+				SecurityGroup: lbSG,
+				SourceGroup:   nodeGroup.Task,
```
seh, line 290: Can we check if the API server load balancer is public, and if so, use a more lax source range here? As @hakman realized, for a public ELB, the traffic comes in from the nodes via a NAT gateway with a source address that's not even within any of the VPC's CIDR blocks. (The NAT gateways each have a private IP address within the VPC, but the corresponding "Elastic IP address" is outside the VPC.)

seh: Per https://github.com/kubernetes/kops/issues/10139#issuecomment-719978826, I now think we're fixing the wrong problem here.

johngmyers:
Looks like the choices might be to expose the port externally or pay for a second, internal, load balancer.

pkg/model/awsmodel/api_loadbalancer.go
hakman: If someone asks for Public LB and UseForInternalApi, I think should allow access from 0.0.0.0/0, same as for API.
Or maybe not allow UseForInternalApi in this case, as it won't work anyway.
I don't have a strong preference in general here, so feel free to ignore my comment.

seh: I want to make sure I understand your second proposal: Could _kops_ reject "useForInternalApi" as invalid when the load balancer is public? I think that's the best choice, if it's feasible.

seh: I liked the option—clearly, as I tried to enable it—for accessing the Kubernetes API servers, not really thinking about whether the load balancer was public or not. At that point, though, I didn't realize that it would be used for other things like this node bootstrapping.

And did this other use start in _kops_ 1.19? I didn't experience this problem with _kops_ 1.18.2.

johngmyers:
Going with a separate dns-controller-managed domain instead, in #10239

PR #10144, opened 2020-10-30
allow reauth for openstack client

zetaab:


util/pkg/vfs/swiftfs.go
```diff
@@ -51,8 +51,6 @@ func NewSwiftClient() (*gophercloud.ServiceClient, error) {
 		return nil, err
 	}
 
-	authOption.AllowReauth = true
```
zetaab, line 54: this is already defined if we add allowreauth to GetCredential

PR #10151, opened 2020-11-02
Use LaunchTemplate versions instead of timestamped LaunchTemplates

hakman:
Historically, kOps used LaunchConfigurations for configuring AutoScalingGroups with the IG name as prefix and creation timestamp as suffix (`nodes-a.k8s.example.com-20200806091534`).

Newer versions of kOps added the possibility of using LaunchTemplates for configuring AutoScalingGroups, as some features are not available for when using LaunchConfigurations. Even though LaunchTemplates allow versioning, kOps uses the same naming convention as for LaunchConfigurations, but having to deal with LaunchTemplateVersions also and generating many requests to the AWS API, as seen in https://github.com/kubernetes/kops/issues/9806.

kOps 1.19 introduces LaunchTemplates as the default AutoScalingGroups configuration method and switching to versions should simplify things:
* the LaunchTemplate would have a fixed name identical to the IG name
* ASGs will use the `$Latest` LaunchTemplate version
* all versions will be kept (for now) as there AWS limit is 10.000 versions for each LaunchTemplate
* all previous LaunchTemplates with timestamp prefixed names will be deleted, reducing the clutter in the LaunchTemplate list

CloudFormation output will not be affected by these changes.

Terraform output will change to using `name` instead of `name_prefix` (another leftover from the LaunchConfigurations days).

### DIFF
--- a/pkg/resources/aws/aws.go
+++ b/pkg/resources/aws/aws.go
@@ -145,7 +145,7 @@ func ListResourcesAWS(cloud awsup.AWSCloud, clusterName string) (map[string]*res
 		if err != nil {
 			return nil, err
 		}
-		lts, err := FindAutoScalingLaunchTemplateConfigurations(cloud, securityGroups)
+		lts, err := FindAutoScalingLaunchTemplates(cloud, clusterName, securityGroups)
 		if err != nil {
 			return nil, err
 		}
@@ -1179,47 +1179,34 @@ func ListAutoScalingGroups(cloud fi.Cloud, clusterName string) ([]*resources.Res
 	return resourceTrackers, nil
 }
 
-// FindAutoScalingLaunchTemplateConfigurations finds any launch configurations which reference the security groups
-func FindAutoScalingLaunchTemplateConfigurations(cloud fi.Cloud, securityGroups sets.String) ([]*resources.Resource, error) {
-	var list []*resources.Resource
-
+// FindAutoScalingLaunchTemplates finds any launch configurations which reference the security groups
+func FindAutoScalingLaunchTemplates(cloud fi.Cloud, clusterName string, securityGroups sets.String) ([]*resources.Resource, error) {
 	c, ok := cloud.(awsup.AWSCloud)
 	if !ok {
 		return nil, errors.New("expected a aws cloud provider")
 	}
 	klog.V(2).Infof("Finding all Autoscaling LaunchTemplates associated to security groups")
 
-	resp, err := c.EC2().DescribeLaunchTemplates(&ec2.DescribeLaunchTemplatesInput{MaxResults: fi.Int64(100)})
+	output, err := c.EC2().DescribeLaunchTemplates(&ec2.DescribeLaunchTemplatesInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("tag:KubernetesCluster"),
+				Values: []*string{aws.String(clusterName)},
+			},
+		},
+	})
 	if err != nil {
-		return list, nil
+		return nil, nil
 	}
 
-	for _, x := range resp.LaunchTemplates {
-		// @step: grab the actual launch template
-		req, err := c.EC2().DescribeLaunchTemplateVersions(&ec2.DescribeLaunchTemplateVersionsInput{
-			LaunchTemplateName: x.LaunchTemplateName,
+	var list []*resources.Resource
+	for _, lt := range output.LaunchTemplates {
+		list = append(list, &resources.Resource{
+			Name:    aws.StringValue(lt.LaunchTemplateName),
+			ID:      aws.StringValue(lt.LaunchTemplateId),
+			Type:    TypeAutoscalingLaunchConfig,
+			Deleter: DeleteAutoScalingGroupLaunchTemplate,
 		})
-		if err != nil {
-			return list, err
-		}
-		for _, j := range req.LaunchTemplateVersions {
-			// @check if the security group references the security group above
-			var s []*string
-			for _, ni := range j.LaunchTemplateData.NetworkInterfaces {
-				s = append(s, ni.Groups...)
-			}
-			s = append(s, j.LaunchTemplateData.SecurityGroupIds...)
-			for _, y := range s {
-				if securityGroups.Has(fi.StringValue(y)) {
-					list = append(list, &resources.Resource{
-						Name:    aws.StringValue(x.LaunchTemplateName),
-						ID:      aws.StringValue(x.LaunchTemplateName),
-						Type:    TypeAutoscalingLaunchConfig,
-						Deleter: DeleteAutoScalingGroupLaunchTemplate,
-					})
-				}
-			}
-		}
 	}
 
 	return list, nil
@@ -1377,7 +1364,7 @@ func DeleteAutoScalingGroupLaunchTemplate(cloud fi.Cloud, r *resources.Resource)
 	klog.V(2).Infof("Deleting EC2 LaunchTemplate %q", r.ID)
 
 	if _, err := c.EC2().DeleteLaunchTemplate(&ec2.DeleteLaunchTemplateInput{
-		LaunchTemplateName: fi.String(r.ID),
+		LaunchTemplateId: fi.String(r.ID),
 	}); err != nil {
 		return fmt.Errorf("error deleting ec2 LaunchTemplate %q: %v", r.ID, err)
 	}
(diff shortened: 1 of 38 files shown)


pkg/resources/aws/aws.go
```diff
@@ -1179,47 +1179,34 @@ func ListAutoScalingGroups(cloud fi.Cloud, clusterName string) ([]*resources.Res
 	return resourceTrackers, nil
 }
 
-// FindAutoScalingLaunchTemplateConfigurations finds any launch configurations which reference the security groups
-func FindAutoScalingLaunchTemplateConfigurations(cloud fi.Cloud, securityGroups sets.String) ([]*resources.Resource, error) {
-	var list []*resources.Resource
-
+// FindAutoScalingLaunchTemplates finds any launch configurations which reference the security groups
```
johngmyers, line 1182:
```suggestion
// FindAutoScalingLaunchTemplates finds any launch templates which reference the security groups.
```

pkg/resources/aws/aws.go
```diff
@@ -1179,47 +1179,34 @@ func ListAutoScalingGroups(cloud fi.Cloud, clusterName string) ([]*resources.Res
...
+func FindAutoScalingLaunchTemplates(cloud fi.Cloud, clusterName string, securityGroups sets.String) ([]*resources.Resource, error) {
 	c, ok := cloud.(awsup.AWSCloud)
 	if !ok {
 		return nil, errors.New("expected a aws cloud provider")
 	}
 	klog.V(2).Infof("Finding all Autoscaling LaunchTemplates associated to security groups")
 
-	resp, err := c.EC2().DescribeLaunchTemplates(&ec2.DescribeLaunchTemplatesInput{MaxResults: fi.Int64(100)})
+	output, err := c.EC2().DescribeLaunchTemplates(&ec2.DescribeLaunchTemplatesInput{
```
johngmyers, line 1190: Do we not have to worry about paging results here?

pkg/resources/aws/aws.go
```diff
@@ -1179,47 +1179,34 @@ func ListAutoScalingGroups(cloud fi.Cloud, clusterName string) ([]*resources.Res
...
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("tag:KubernetesCluster"),
```
johngmyers, line 1193: Should we be using the legacy tag? Perhaps use the `kubernetes.io/cluster/` one instead?

pkg/resources/aws/aws.go
```diff
@@ -1179,47 +1179,34 @@ func ListAutoScalingGroups(cloud fi.Cloud, clusterName string) ([]*resources.Res
...
+		return nil, nil
 	}
 
-	for _, x := range resp.LaunchTemplates {
-		// @step: grab the actual launch template
-		req, err := c.EC2().DescribeLaunchTemplateVersions(&ec2.DescribeLaunchTemplateVersionsInput{
-			LaunchTemplateName: x.LaunchTemplateName,
+	var list []*resources.Resource
+	for _, lt := range output.LaunchTemplates {
+		list = append(list, &resources.Resource{
```
johngmyers, line 1204: Where is it filtering for referencing `securityGroups`? I don't see where `securityGroups` is referenced.

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -363,8 +366,8 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 				},
 				LaunchTemplate: &autoscaling.LaunchTemplate{
 					LaunchTemplateSpecification: &autoscaling.LaunchTemplateSpecification{
-						LaunchTemplateName: e.LaunchTemplate.ID,
-						Version:            aws.String("1"),
+						LaunchTemplateName: e.LaunchTemplate.Name,
```
johngmyers, line 369: Perhaps specify the `LaunchTemplateID` instead?

upup/pkg/fi/cloudup/awstasks/launchtemplate.go
```diff
@@ -154,22 +148,17 @@ func (t *LaunchTemplate) CheckChanges(a, e, changes *LaunchTemplate) error {
...
-	if len(configurations) <= RetainLaunchConfigurationCount() {
-		return nil, nil
-	}
-
-	configurations = configurations[:len(configurations)-RetainLaunchConfigurationCount()]
-
+	prefix := fmt.Sprintf("%s-", fi.StringValue(t.Name))
 	for _, configuration := range configurations {
-		removals = append(removals, &deleteLaunchTemplate{lc: configuration})
+		if strings.HasPrefix(aws.StringValue(configuration.LaunchTemplateName), prefix) {
```
johngmyers, line 158: Why is this necessary? Wouldn't all the old launch templates be deleted as owned by the cluster but unreferenced by an ASG?

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -138,18 +126,40 @@ func (t *LaunchTemplate) RenderAWS(c *awsup.AWSAPITarget, a, ep, changes *Launch
...
+				},
+			},
+		}
+		output, err := c.Cloud.EC2().CreateLaunchTemplate(input)
+		if err != nil || output.LaunchTemplate == nil {
+			return fmt.Errorf("error creating LaunchTemplate %q: %v", fi.StringValue(t.Name), err)
+		}
+		e.ID = output.LaunchTemplate.LaunchTemplateId
+	} else {
+		// TODO: Update the LaunchTemplate tags
```
johngmyers: I think implementing this might block merging.

upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -694,52 +694,44 @@ func findAutoscalingGroupLaunchConfiguration(c AWSCloud, g *autoscaling.Group) (
...
+		}
+		output, err := c.EC2().DescribeLaunchTemplates(input)
+		if err != nil {
+			return "", fmt.Errorf("error describing launch templates: %q", err)
+		}
+		if len(output.LaunchTemplates) == 0 {
+			return "", fmt.Errorf("error finding launch template by name: %q", name)
+		}
+		launchTemplate := output.LaunchTemplates[0]
+		if version == "" || version == "$Default" {
```
johngmyers, line 726: I think it would be simpler to test for `version == "$Latest"` instead.

johngmyers (changes requested):

pkg/resources/aws/aws.go
hakman: Good point, will change.

pkg/resources/aws/aws.go
hakman: I doubt anyone will have so many LTs for a single cluster, so probably we're ok without paging.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
hakman: Hmm, already implemented it in https://github.com/kubernetes/kops/pull/10151/commits/852c0e2a8f4c376909d413db500fa798807239f2 and seems to work without issues. Nothing actually reads these tags after creation, they are just used for deleting the LT when needed.

upup/pkg/fi/cloudup/awsup/aws_cloud.go
hakman: Agreed. Didn't want to make the diff even bigger.

pkg/resources/aws/aws.go
hakman: Just forgot to remove it. Should be addressed in the next commit.

upup/pkg/fi/cloudup/awstasks/launchtemplate.go
hakman: There can be N previous LTs for each ASG. already unreferenced that should be deleted.
Maybe there's some mechanism that I'm missing.

pkg/resources/aws/aws.go
hakman: Updated to use `DescribeLaunchTemplatesPages`. Probably at some point we should try and clean up the code in AWS and use the same style of pagination and naming convention.

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
hakman: Updated, should make things clearer in case of mismatches.

upup/pkg/fi/cloudup/awstasks/launchtemplate.go
johngmyers: Looks like I misunderstood the `Find` / `FindDeletions` mechanisms. I thought it would look for resources tagged as owned by the cluster and delete them if they're not expected.

hakman:
Thanks @johngmyers :)

PR #10156, opened 2020-11-02
Add ACM cert permalink

rifelpet:
The ACM bugfix will fail API validation with k8s >=1.19 + `sslCertificate` + `class: Classic`.

The error message will contain this permalink, so I'd like to get eyes on this early.

If/when we handle migration without downtime, we can update these instructions to match.

ref: https://github.com/kubernetes/kops/issues/9756


rifelpet:
/hold cancel

we can iterate on this in followup PRs if there's any feedback, but I wanted to get the link out there to be able to reference in the #10157 code

PR #10157, opened 2020-11-02
 Setup a second NLB listener when an AWS ACM certificate is used

rifelpet:
This uses commits from https://github.com/kubernetes/kops/pull/9761 combined with the new NLB support to provision a second NLB listener. It requires converting the `Listeners` and `TargetGroups` NLB fields to slices in which each listener should forward traffic to the target group of the same list index. I couldn't use a map of port number to TargetGroup because fi/loader.go uses reflection to set values of fields that are Tasks, and map values aren't addressable which resulted in a panic.

I also temporarily pulled in the commit from https://github.com/kubernetes/kops/pull/10076/files to run the e2e presubmit job with an ACM cert. Before we merge this I'll remove the commit.

ref: https://github.com/kubernetes/kops/issues/9756

### DIFF
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -371,6 +371,9 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 		if b.UseLoadBalancerForAPI() && ig.Spec.Role == kops.InstanceGroupRoleMaster {
 			if b.UseNetworkLoadBalancer() {
 				t.TargetGroups = append(t.TargetGroups, b.LinkToTargetGroup("api"))
+				if b.Cluster.Spec.API.LoadBalancer.SSLCertificate != "" {
+					t.TargetGroups = append(t.TargetGroups, b.LinkToTargetGroup("tcp"))
+				}
 			} else {
 				t.LoadBalancers = append(t.LoadBalancers, b.LinkToCLB("api"))
 			}
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -364,6 +364,8 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 
 	t.InstanceProtection = ig.Spec.InstanceProtection
 
+	t.TargetGroups = []*awstasks.TargetGroup{}
+
 	// When Spotinst Elastigroups are used, there is no need to create
 	// a separate task for the attachment of the load balancer since this
 	// is already done as part of the Elastigroup's creation, if needed.
@@ -384,7 +386,7 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 		}
 	}
 
-	for _, extLB := range ig.Spec.ExternalLoadBalancers {
+	for i, extLB := range ig.Spec.ExternalLoadBalancers {
 		if extLB.LoadBalancerName != nil {
 			lb := &awstasks.ClassicLoadBalancer{
 				Name:             extLB.LoadBalancerName,
@@ -397,7 +399,7 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 
 		if extLB.TargetGroupARN != nil {
 			tg := &awstasks.TargetGroup{
-				Name:   extLB.TargetGroupARN,
+				Name:   fi.String(fmt.Sprintf("external-tg-%d", i)),
 				ARN:    extLB.TargetGroupARN,
 				Shared: fi.Bool(true),
 			}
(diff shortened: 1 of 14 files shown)


upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -317,6 +336,7 @@ func (e *NetworkLoadBalancer) Find(c *fi.Context) (*NetworkLoadBalancer, error)
 	actual.Scheme = lb.Scheme
 	actual.VPC = &VPC{ID: lb.VpcId}
 	actual.Type = lb.Type
+	actual.TargetGroups = make([]*TargetGroup, 0)
```
hakman, line 339: Is this really needed?

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -343,36 +363,44 @@ func (e *NetworkLoadBalancer) Find(c *fi.Context) (*NetworkLoadBalancer, error)
 			return nil, fmt.Errorf("error querying for NLB listeners :%v", err)
 		}
 
-		actual.Listeners = make(map[string]*NetworkLoadBalancerListener)
+		actual.Listeners = make([]*NetworkLoadBalancerListener, 0)
```
hakman, line 366: Is this still needed?

upup/pkg/fi/cloudup/awstasks/targetgroup.go
```diff
@@ -191,6 +194,18 @@ func (_ *TargetGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *TargetGrou
...
+
+func (a OrderTargetGroupsByPort) Len() int      { return len(a) }
+func (a OrderTargetGroupsByPort) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a OrderTargetGroupsByPort) Less(i, j int) bool {
+	if a[i].ARN != nil || a[j].ARN != nil {
+		return fi.StringValue(a[i].ARN) < fi.StringValue(a[j].ARN)
+	}
+	return fi.StringValue(a[i].Name) < fi.StringValue(a[j].Name)
+}
+
```
hakman, line 208: Not quite by port. Name should be enough?

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -343,36 +363,44 @@ func (e *NetworkLoadBalancer) Find(c *fi.Context) (*NetworkLoadBalancer, error)
...
+					descResp, err := cloud.ELBV2().DescribeTargetGroups(&elbv2.DescribeTargetGroupsInput{
+						TargetGroupArns: []*string{targetGroupARN},
+					})
+					if err != nil {
+						return nil, fmt.Errorf("error querying for NLB listener target groups: %v", err)
+					}
+					if len(descResp.TargetGroups) != 1 {
+						return nil, fmt.Errorf("unexpected DescribeTargetGroups response: %v", descResp)
+					}
+					actualListener.TargetGroupName = aws.StringValue(descResp.TargetGroups[0].TargetGroupName)
```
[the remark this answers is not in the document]
rifelpet: Not sure I understand, which field would we use with "compare by ID" ? We need a field that can be both provided by the model and something we can look up in `Find()`. It seems like our only options for that are TargetGroupName or tags.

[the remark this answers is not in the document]
hakman: I meant that at the time this runs there is also full info on target groups populated in `e.TargetGroups`, so you can just get the name from there without an extra call to `DescribeTargetGroups`.
Though this is an optimisation that can be done later.

upup/pkg/fi/cloudup/awstasks/targetgroup.go
hakman: Still curious, why compare both by Name and ARN? Names should be unique.

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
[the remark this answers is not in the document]
rifelpet: If we're removing a listener (and therefore target group), `e.TargetGroups` won't have the target group being removed so we wont be able to lookup its name from `e.TargetGroups`. We'd have to `DescribeTargetGroups` to find it.

Though maybe the value of `actualListener.TargetGroupName` doesn't actually matter for a listener being deleted? When we add target group deletion support we probably won't be using the `listener.TargetGroupName` to determine what to delete, instead just use `NLBTargetGroupName()` to determine all the possible TG names and look for any that exist in AWS that shouldn't and delete them. just thinking out loud here...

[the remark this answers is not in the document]
hakman: If delete is done by ARN, probably no describing will be needed. I agree that the compare part should check which ones are removed and just delete them.

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -163,12 +163,11 @@ func (e *AutoscalingGroup) Find(c *fi.Context) (*AutoscalingGroup, error) {
 	}
```
hakman, line 164:
```suggestion

	actual.TargetGroups = []*TargetGroup{}
```

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -370,7 +370,10 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 	if !featureflag.Spotinst.Enabled() {
```
hakman, line 370:
```suggestion
	t.TargetGroups = []*awstasks.TargetGroup{}
	if !featureflag.Spotinst.Enabled() {
```

hakman (changes requested):

PR #10154, opened 2020-11-02
Automated cherry pick of #10140 and #9011: Implement API load balancer class with NLB and ELB support on AWS

rifelpet:
Cherry pick of #10140 #9011 on release-1.19.

#10140: Create cloudmock implementations for elbv2 API calls used
#9011: Implement API load balancer class with NLB and ELB support on AWS

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

PR #10158, opened 2020-11-03
Fix NLB naming for terraform and cloudformation targets

rifelpet:
This ensures we use the LoadBalancerName field rather than the name of the NLB task itself. The `direct` target is already using LoadBalancerName:

https://github.com/kubernetes/kops/blob/e91ed114499294f3e4242c6b23e7150a5c7262ec/upup/pkg/fi/cloudup/awstasks/network_load_balancer.go#L503-L504

This will need to be cherry-picked to release-1.19.

ref: https://github.com/kubernetes/kops/pull/10154#issuecomment-720820551

### DIFF
--- a/tests/integration/update_cluster/complex/cloudformation.json
+++ b/tests/integration/update_cluster/complex/cloudformation.json
@@ -1135,7 +1135,7 @@
     "AWSElasticLoadBalancingV2LoadBalancerapicomplexexamplecom": {
       "Type": "AWS::ElasticLoadBalancingV2::LoadBalancer",
       "Properties": {
-        "Name": "api-complex-example-com",
+        "Name": "api-complex-example-com-vd3t5n",
         "Scheme": "internet-facing",
         "Subnets": [
           {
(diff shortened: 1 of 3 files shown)


seh:
Thank you, @rifelpet. When I went to test this, I got further, but then ran into this next error from AWS:
```
Error: Error updating Autoscaling group: ValidationError: You must use a valid fully-formed launch template. The security group 'nlb-sg-0ef9de1ab14bee565' does not exist in VPC 'vpc-077913405bb66aac2'
	status code: 400, request id: 6e6096b8-a3a7-427a-a846-5272aa46858f

  on kubernetes.tf line 123, in resource "aws_autoscaling_group" "master-us-east-2a-masters-redacted-cluster-name":
 123: resource "aws_autoscaling_group" "master-us-east-2a-masters-redacted-cluster-name" {
```
In the Terraform `aws_launch_template` resource, the "network_interfaces.security_groups" attribute contains two entries:
- `aws_security_group.masters-redacted-cluster-name.id`
- `"nlb-sg-0ef9de1ab14bee565"`

The second security group is not defined as a Terraform resource.

That error repeats similarly for for all of the ASGs in the cluster.

PR #10161, opened 2020-11-03
Move NLB's VPC CIDR security group rule logic into model

rifelpet:
This way the security group rule task doesn't need to be aware of VPCs, since we should always know the VPC CIDR ahead of time via cluster spec.

This also fixes the terraform and cloudformation rendering of this rule (see the added cidr block in the integration test outputs)

These rules are for NLB's health checks. The AWS docs recommend allowing access from the entire VPC CIDRs
Also add rules for additionalNetworkCIDRs, supporting VPCs with multiple CIDR blocks.

ref: https://github.com/kubernetes/kops/pull/9011#issuecomment-720143425

### DIFF
--- a/pkg/model/awsmodel/api_loadbalancer.go
+++ b/pkg/model/awsmodel/api_loadbalancer.go
@@ -357,8 +357,19 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 				Protocol:      fi.String("tcp"),
 				SecurityGroup: masterGroup.Task,
 				ToPort:        fi.Int64(443),
-				VPC:           b.LinkToVPC(),
+				CIDR:          fi.String(b.Cluster.Spec.NetworkCIDR),
 			})
+			for _, cidr := range b.Cluster.Spec.AdditionalNetworkCIDRs {
+				c.AddTask(&awstasks.SecurityGroupRule{
+					Name:          fi.String(fmt.Sprintf("https-lb-to-master%s-%s", suffix, cidr)),
+					Lifecycle:     b.SecurityLifecycle,
+					FromPort:      fi.Int64(443),
+					Protocol:      fi.String("tcp"),
+					SecurityGroup: masterGroup.Task,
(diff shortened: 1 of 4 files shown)


pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -357,8 +357,19 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 				Protocol:      fi.String("tcp"),
 				SecurityGroup: masterGroup.Task,
 				ToPort:        fi.Int64(443),
-				VPC:           b.LinkToVPC(),
+				CIDR:          &b.Cluster.Spec.NetworkCIDR,
```
hakman, line 360:
```suggestion
				CIDR:          fi.String(b.Cluster.Spec.NetworkCIDR),
```

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -357,8 +357,19 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
 			})
+			for _, cidr := range b.Cluster.Spec.AdditionalNetworkCIDRs {
+				c.AddTask(&awstasks.SecurityGroupRule{
+					Name:          fi.String(fmt.Sprintf("https-lb-to-master%s-%s", suffix, cidr)),
+					Lifecycle:     b.SecurityLifecycle,
+					FromPort:      fi.Int64(443),
+					Protocol:      fi.String("tcp"),
+					SecurityGroup: masterGroup.Task,
+					ToPort:        fi.Int64(443),
+					CIDR:          &cidr,
```
hakman, line 370:
```suggestion
					CIDR:          fi.String(cidr),
```

seh:
I tried a build based on commit b2728989f0351290cd4359ec822bccb05eda9ea9 and it failed the same way it did before, but now I see we have f08284834e4d4630d93d3cb9853c86822f55d432.

PR #10164, opened 2020-11-04
Fix output for CF and TF

hakman:
This addresses a few issues that makes `update cluster` show that there are unapplied changes:
* add missing resource names for CF
* add missing instance monitoring for CF and TF
* Ignore tags added by CF
* order policy document sections alphabetically

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/iaminstanceprofilerole.go
+++ b/upup/pkg/fi/cloudup/awstasks/iaminstanceprofilerole.go
@@ -126,14 +126,14 @@ func (_ *IAMInstanceProfileRole) RenderTerraform(t *terraform.TerraformTarget, a
 }
 
 type cloudformationIAMInstanceProfile struct {
-	//Path  *string              `json:"name"`
-	Roles []*cloudformation.Literal `json:"Roles"`
+	InstanceProfileName *string                   `json:"InstanceProfileName"`
+	Roles               []*cloudformation.Literal `json:"Roles"`
 }
 
 func (_ *IAMInstanceProfileRole) RenderCloudformation(t *cloudformation.CloudformationTarget, a, e, changes *IAMInstanceProfileRole) error {
 	cf := &cloudformationIAMInstanceProfile{
-		//Path:  e.InstanceProfile.Name,
-		Roles: []*cloudformation.Literal{e.Role.CloudformationLink()},
+		InstanceProfileName: e.InstanceProfile.Name,
+		Roles:               []*cloudformation.Literal{e.Role.CloudformationLink()},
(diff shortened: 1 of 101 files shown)


upup/pkg/fi/cloudup/awstasks/iaminstanceprofilerole.go
```diff
@@ -126,12 +126,14 @@ func (_ *IAMInstanceProfileRole) RenderTerraform(t *terraform.TerraformTarget, a
...
 
 type cloudformationIAMInstanceProfile struct {
+	InstanceProfileName *string `json:"InstanceProfileName"`
 	//Path  *string              `json:"name"`
 	Roles []*cloudformation.Literal `json:"Roles"`
 }
 
 func (_ *IAMInstanceProfileRole) RenderCloudformation(t *cloudformation.CloudformationTarget, a, e, changes *IAMInstanceProfileRole) error {
 	cf := &cloudformationIAMInstanceProfile{
+		InstanceProfileName: e.Name,
```
rifelpet, line 136: Terraform uses `e.InstanceProfile.Name` (Line 121) perhaps we should be consistent?

upup/pkg/fi/cloudup/awstasks/iamrolepolicy.go
```diff
@@ -115,7 +114,19 @@ func (e *IAMRolePolicy) Find(c *fi.Context) (*IAMRolePolicy, error) {
 		if err != nil {
 			return nil, fmt.Errorf("error parsing PolicyDocument for IAMRolePolicy %q: %v", aws.StringValue(e.Name), err)
 		}
-		actual.PolicyDocument = fi.WrapResource(fi.NewStringResource(policy))
+
+		// Reformat the PolicyDocument by unmarshaling and re-marshaling to JSON.
```
rifelpet, line 118: why is this necessary exactly?

hakman: Good point. Will update.

hakman: The document is just a JSON that is stored in AWS. In Direct and TF case, what you send is what you get back.
In the case of CloudFormation.... you get back a nice "optimised" JSON, that is no longer matching what was sent. Obviously, `kops update cluster` sees this as a change and tries to fix it.

upup/pkg/fi/cloudup/awstasks/iaminstanceprofilerole.go
hakman: Good point. Updated.

PR #10167, opened 2020-11-04
Make it possible to use OnDelete update strategy on addon daemonset

### LINKED ISSUE #7971: Bootstrap channel DaemonSets have updateStrategy: RollingUpdate
 this information.**

Version 1.14.1 (git-b7c25f9a9)

 a `kops` flag.**

Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"archive", BuildDate:"2019-09-21T09:43:39Z", GoVersion:"go1.12.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.10", GitCommit:"e3c134023df5dea457638b614ee17ef234dc34a6", GitTreeState:"clean", BuildDate:"2019-07-08T03:40:54Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}

AWS

`kops update cluster --target cloudformation --out /tmp` using kops 1.14.1, on a cluster using weave for networking that was last updated with kops 1.12.3.

As soon as the new version of weave was installed, the cluster initiated a rolling restart of the weave pods, which caused a brief but visible network interruption on each node.

I would not expect any weave pods to be replaced until I ask kops to perform a rolling update, given that they are essential for the pods on a node to function correctly.

https://gist.github.com/stevenjm/c7eccd8b3fc20c675df61dfa3d6e9ecc

In the case of weave, this is an [explicit setting](https://github.com/kubernetes/kops/blob/1.14.1/upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template#L257) that appears to have been inherited from upstream. I suspect this makes sense for weave upstream, as they can't assume the cluster is managed in such a way that nodes can easily be updated in place, but kops allows us to make this assumption and set updatePolicy to OnDelete. This will let a `kops rolling-update` replace these pods for us.

For this to work correctly, the node's user data must include a checksum of the bootstrap channel so that kops knows that nodes needs to be replaced when weave is updated. See also my suggested approach to fixing https://github.com/kubernetes/kops/issues/7970.

olemarkus:
Fixes #7971

Or at least makes it possible to fix it for the addons we need.

Works by adding another element to the addon spec. If it is set, will make channels patch an annotation on the relevant nodes that rolling update uses to determine if an instance needs updating.

### DIFF
--- a/channels/pkg/api/channel.go
+++ b/channels/pkg/api/channel.go
@@ -62,6 +62,11 @@ type AddonSpec struct {
 	// version of the software we are packaging.  But we always want to reinstall when we
 	// switch kubernetes versions.
 	Id string `json:"id,omitempty"`
+
+	// NeedUpdate determines if we should mark nodes as needing an update.
+	// Legal values are control-plane, workers, and all
+	// Empty value means no update needed
+	NeedUpdate string `json:"needsRollingUpdate,omitempty"`
 }
 
 func (a *Addons) Verify() error {
--- a/channels/pkg/api/channel.go
+++ b/channels/pkg/api/channel.go
@@ -63,10 +63,10 @@ type AddonSpec struct {
 	// switch kubernetes versions.
 	Id string `json:"id,omitempty"`
 
-	// NeedUpdate determines if we should mark nodes as needing an update.
+	// NeedsRollingUpdate determines if we should mark nodes as needing an update.
 	// Legal values are control-plane, workers, and all
 	// Empty value means no update needed
-	NeedUpdate string `json:"needsRollingUpdate,omitempty"`
+	NeedsRollingUpdate string `json:"needsRollingUpdate,omitempty"`
 }
 
 func (a *Addons) Verify() error {
--- a/channels/pkg/channels/addon.go
+++ b/channels/pkg/channels/addon.go
@@ -149,7 +149,7 @@ func (a *Addon) EnsureUpdated(ctx context.Context, k8sClient kubernetes.Interfac
 		return nil, fmt.Errorf("error applying update from %q: %v", manifestURL, err)
 	}
 
-	if a.Spec.NeedUpdate != "" {
+	if a.Spec.NeedsRollingUpdate != "" {
 		err = a.AddNeedsUpdateLabel(ctx, k8sClient)
 		if err != nil {
 			return nil, fmt.Errorf("error adding needs-update label: %v", err)
@@ -166,9 +166,9 @@ func (a *Addon) EnsureUpdated(ctx context.Context, k8sClient kubernetes.Interfac
 }
 
 func (a *Addon) AddNeedsUpdateLabel(ctx context.Context, k8sClient kubernetes.Interface) error {
-	klog.Infof("addon %v wants to roll %v nodes", a.Name, a.Spec.NeedUpdate)
+	klog.Infof("addon %v wants to update %v nodes", a.Name, a.Spec.NeedsRollingUpdate)
 	selector := ""
-	switch a.Spec.NeedUpdate {
+	switch a.Spec.NeedsRollingUpdate {
 	case "control-plane":
 		selector = "node-role.kubernetes.io/master="
 	case "worker":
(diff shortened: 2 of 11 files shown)


channels/pkg/channels/addon.go
```diff
@@ -152,3 +164,36 @@ func (a *Addon) EnsureUpdated(ctx context.Context, k8sClient kubernetes.Interfac
 
 	return required, nil
 }
+
+func (a *Addon) AddNeedsUpdateLabel(ctx context.Context, k8sClient kubernetes.Interface) error {
+	klog.Infof("addon %v wants to roll %v nodes", a.Name, a.Spec.NeedUpdate)
+	selector := ""
+	switch a.Spec.NeedUpdate {
+	case "masters":
```
johngmyers, line 172: Should probably use "control-plane" for newly added fields.

channels/pkg/channels/addon.go
```diff
@@ -152,3 +164,36 @@ func (a *Addon) EnsureUpdated(ctx context.Context, k8sClient kubernetes.Interfac
...
+		selector = "node-role.kubrnetes.io/master="
+	case "nodes":
```
johngmyers, line 174: Consider "worker"

channels/pkg/api/channel.go
```diff
@@ -62,6 +62,11 @@ type AddonSpec struct {
 	// version of the software we are packaging.  But we always want to reinstall when we
 	// switch kubernetes versions.
 	Id string `json:"id,omitempty"`
+
+	// NeedUpdate determines if we should mark nodes as needing an update.
```
johngmyers, line 66:
```suggestion
	// NeedUpdate specifies which node roles to mark as needing an update when installing or updating the addon.
```

channels/pkg/channels/addon.go
johngmyers: Hrm, I wonder if we should introduce a mix of old and new terminology now or instead wait for a new apiVersion to change all the terminology used in the API at once.

johngmyers: I'm willing to let this proceed with the legacy terminology.

olemarkus: This isn't really exposed to users, so I am all for changing the names.

channels/pkg/api/channel.go
```diff
@@ -62,6 +62,11 @@ type AddonSpec struct {
...
+	// Legal values are control-plane, workers, and all
+	// Empty value means no update needed
+	NeedUpdate string `json:"needsRollingUpdate,omitempty"`
```
johngmyers, line 69: Please make the struct field name and json name consistent. I think `NeedsRollingUpdate` would be a better struct name.

channels/pkg/channels/addon.go
```diff
@@ -152,3 +164,36 @@ func (a *Addon) EnsureUpdated(ctx context.Context, k8sClient kubernetes.Interfac
 
 	return required, nil
 }
+
+func (a *Addon) AddNeedsUpdateLabel(ctx context.Context, k8sClient kubernetes.Interface) error {
+	klog.Infof("addon %v wants to roll %v nodes", a.Name, a.Spec.NeedUpdate)
```
johngmyers, line 169:
```suggestion
	klog.Infof("addon %v wants to update %v nodes", a.Name, a.Spec.NeedUpdate)
```

johngmyers (changes requested):
Just a few nits.

channels/pkg/channels/addon.go
johngmyers: This appears to have been dropped without change or comment.

olemarkus: Forgot to pull in this change before force pushing ...

johngmyers: Figured it was something like that.

PR #10178, opened 2020-11-06
OpenStack Reset deviceID status if needed

zetaab:
Because we changed the behaviour how kops rolling update works, we are all the time randomly seeing following error:

```
{"conflictingRequest": {"message": "Port e3a94980-6b13-4ed1-8188-0c3634964915 is still in use.", "code": 409}}
```

The problem here is that we delete the instance that used port x, and then we are creating new instance using port x. Seems that this behaviour is now too fast at least for our OpenStack. Still after one hour the port is attached to deleted instance. This seems OpenStack bug, but because its quite easy to fix in kops side - I will do it.

With this code we are able to run rolling updates also in these cases.

/cc @olemarkus


olemarkus:
This looks good. Thanks!

/lgtm

PR #10179, opened 2020-11-06
Consistent naming of security group rules

olemarkus:
In order to let kops fully control the rules for each security group we need to be able to generate names from the info in AWS. This is similar to the approach we used for openstack.

This only converts the naming of some of the rules. Since this affects CF/TF resource names, I want to make sure this doesn't cause any concerns.


pkg/model/firewall.go
```diff
@@ -521,3 +403,44 @@ func JoinSuffixes(src SecurityGroupInfo, dest SecurityGroupInfo) string {
...
+	if o.SourceGroup != nil {
+		target = fi.StringValue(o.SourceGroup.Name)
+	} else if o.CIDR != nil && fi.StringValue(o.CIDR) != "" {
+		target = fi.StringValue(o.CIDR)
+	} else {
+		target = "0.0.0.0/0"
+	}
+
+	if o.Protocol == nil || fi.StringValue(o.Protocol) == "" {
+		proto = "AllProtos"
```
hakman, line 429: Should be enough to say "all".
```suggestion
		proto = "all"
```

pkg/model/firewall.go
```diff
@@ -521,3 +403,44 @@ func JoinSuffixes(src SecurityGroupInfo, dest SecurityGroupInfo) string {
...
+		src = target
+		dst = fi.StringValue(o.SecurityGroup.Name)
+	} else {
+		direction = "egress"
+		dst = target
+		src = fi.StringValue(o.SecurityGroup.Name)
+	}
+
+	return fmt.Sprintf("%v-%v-from-%v-to-%v-%v-%v", direction,
+		proto, src, dst, fi.Int64Value(o.FromPort), fi.Int64Value(o.ToPort))
```
hakman, line 445: Any special reason for using "%v"?
```suggestion
	return fmt.Sprintf("%s-%s-%d-%d-from-%s-to-%s", direction,
		proto, fi.Int64Value(o.FromPort), fi.Int64Value(o.ToPort), src, dst)
```

pkg/model/firewall.go
olemarkus: all what? :) I used allprotos because for descriptiveness when one reads the names in tasks, logs etc.
But I don't have a strong opinion.

hakman: I guess it could be replaced with a `-` or something else, but the location will be used for protocol everywhere.

pkg/model/firewall.go
hakman: When we have `Ingress`, do we still need `To`, or is it intended that it refers to current SG?
Same for `Egress` and `From`.

olemarkus: We need to mention both src and dst to avoid duplications. or are you referring to the word "from" itself?

rifelpet:
I think this renaming could result in short periods of connectivity loss. We could add another release note with the 'terraform state mv' commands to make terraform report a no-op.

PR #10186, opened 2020-11-07
Compare KubernetesAPIAccess to OpenStack allowedCIDRs deterministically

### LINKED ISSUE #10173: Kops on OpenStack AllowedCIDRs should ignore specific ordering
Kops 1.18 supports AllowedCIDRs which is the modern Octavia way to whitelist API access on a loadbalancer instead of using a Security Group.
So far so good.
Problem is: kops update cluster wants the ordering of the whitelisted ip's returned by OpenStack's api to exactly match the ordering of kops state bucket (e.g. kops get cluster -o yaml). This means a dry run of kops update cluster will always mention this change, even though it's not an actual change.

 this information.**
1.18.2

 a `kops` flag.**
1.18.9

OpenStack

kops update cluster

The diff shows 2 lists. The current state (the allowed-cidrs array on the openstack loadbalancer listener) and the desired state (from kops edit cluster: kubernetesApiAccess).

They are the same, just having a different ordering of the list. So you can keep on applying (kops update cluster --yes), but this will keep coming back as a change that needs to be applied.

The ordered list comparison should be an unordered list comparison. It's about the contents of the arrays being compared, ordering is irrelevant.

havulv:
This fixes #10173 by sorting the Allowed CIDRs and KubernetesAPIAccess, which will result in tasks comparing deterministic slices.

The issue in question mentions comparing the unordered contents against each other, but the change required for that might be considerable.

https://github.com/kubernetes/kops/blob/3941bd507d62a9ef7bb36f91b45478120f840d9b/upup/pkg/fi/changes.go#L181 compares indexed values directly against each other, so I think the implicit assumption is that all comparison slices should be ordered.

### DIFF
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -19,6 +19,7 @@ package openstackmodel
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


olemarkus:
Thanks @havulv 
/lgtm

PR #10188, opened 2020-11-07
Use consistent naming for the remaining SGRs part two

olemarkus:
Continuation of #10179

### DIFF
--- a/pkg/model/firewall.go
+++ b/pkg/model/firewall.go
@@ -442,6 +442,6 @@ func generateName(o *awstasks.SecurityGroupRule) string {
 		src = fi.StringValue(o.SecurityGroup.Name)
 	}
 
-	return fmt.Sprintf("%s-%s-%s-%dto%d-%s", src, direction,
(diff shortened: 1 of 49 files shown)


pkg/model/firewall.go
```diff
@@ -441,6 +441,6 @@ func generateName(o *awstasks.SecurityGroupRule) string {
 		src = fi.StringValue(o.SecurityGroup.Name)
 	}
 
-	return fmt.Sprintf("%s-%s-%s-%dto%d-%s", src, direction,
+	return fmt.Sprintf("from-%s-%s-%s-%dto%d-%s", src, direction,
```
rdrgmnzs, line 444: @olemarkus please correct me if I'm wrong, but won't changing the existing resource name for the terraform resources cause all resources to be re-created, effectively causing downtime while the changes are applied?

cc: @rifelpet

olemarkus:
Yeah, I think so. I mentioned that concern and @rifelpet commented here:
https://github.com/kubernetes/kops/pull/10179#issuecomment-723444677

Issue #7946, opened 2019-11-18
Make etcd-manager LogVerbosity configurable 


elblivion:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

etcd-manager's `LogVerbosity` is currently hard-coded and we can't opt-out of very noisy logging: https://github.com/kubernetes/kops/blob/baa5ae3934139aecaf1d0a55620bcb02e617a1a4/pkg/model/components/etcdmanager/model.go#L341

Before etcd-manager:

* The etcd-server container stdout was very minimal, less than 5 messages an hour about snapshots.
* etcd logged more verbosely to etcd.log, which is picked up in standard Fluentd logging pipelines.

With etcd-manager we see an explosion in logging to stdout (all the etcd log output plus etcd-manager's own logging), and duplicate because of our Fluentd picking up etcd.log. We can and will stop importing the hosts' etcd.log into our logging pipeline, but we lack a way of telling etcd-manager to be less verbose. 

This has led to two orders of magnitude more log volume for this component.

**2. Feel free to provide a design supporting your feature request.**

Something like this in the cluster.spec would feel similar to other components managed by Kops: 
```
  etcdManager:
    LogLevel: N
```

rifelpet:
Kops 1.18 will support setting environment variables for etcd (#8402), so you should be able to set the `ETCD_LOG_LEVEL` env var. see the example [here](https://github.com/kubernetes/kops/blob/b129a83638f9c51a3981d907741ed65efc955ff1/pkg/model/components/etcdmanager/tests/overwrite_settings/cluster.yaml#L17-L20)

olemarkus:
Looks like we can close this one
/close

PR #10194, opened 2020-11-07
Make etcd-manager log verbosity configurable

elblivion:
This PR addresses #7946, which I believed was closed in error - the PR referenced as a fix allows the etcd log level to be changed, but etcd-manager has its own klog-based log level which was hardcoded to a quite verbose level in Kops itself, even reducing the log level of etcd you still have to deal with a lot of noise which then has to be filtered out in fluentd or whatever you use to collect/aggregate container logs.

I took https://github.com/kubernetes/kops/pull/8402 as a guide, please let me know what you think.

### DIFF
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -512,6 +512,9 @@ type EtcdManagerSpec struct {
 	// This allows etcd setting to be overwriten. No config validation is done.
 	// A list of etcd config ENV vars can be found at https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/configuration.md
 	Env []EnvVar `json:"env,omitempty"`
+	// LogLevel allows the klog library verbose log level to be set for etcd-manager. The default is 6.
+	// https://github.com/google/glog#verbose-logging
+	LogLevel *int32 `json:"logLevel,omitempty"`
 }
 
 // EtcdMemberSpec is a specification for a etcd member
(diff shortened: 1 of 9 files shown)


elblivion:
I signed it

pkg/apis/kops/cluster.go
```diff
@@ -512,6 +512,9 @@ type EtcdManagerSpec struct {
 	// This allows etcd setting to be overwriten. No config validation is done.
 	// A list of etcd config ENV vars can be found at https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/configuration.md
 	Env []EnvVar `json:"env,omitempty"`
+	// LogVerbosity allows the klog library verbose log level to be set. The default is 6.
+	// https://github.com/google/glog#verbose-logging
+	LogVerbosity *int32 `json:"logVerbosity,omitempty"`
```
rdrgmnzs, line 517: Can we rename this to LogLevel to follow the pattern used in other components.

https://github.com/kubernetes/kops/search?q=logLevel

rdrgmnzs (commented):
Hi @elblivion thanks for the PR! Things look great, I left one comment/NIT once that's taken care of I'll go ahead and approve the PR.

pkg/apis/kops/cluster.go
elblivion: Sure thing - also, to avoid further confusion it might help to clarify further that this is the etcd-manager log level, not the etcd it manages; any suggestions on that? e.g. a clearer comment (which I saw is put into the CRD), but maybe also in the property name?

rdrgmnzs: I think just a comment mentioning this is for etcd-manager would be enough.

elblivion:
Thanks @rdrgmnzs! Would this be accepted for cherry-picking into 1.19 beta? Or is it too late?

PR #10196, opened 2020-11-08
WIP: HA migration between CLB <-> NLB

christianjoun:


rifelpet:
We discussed this today on a kops call and arrived at the following ideas:
* We don't need to support HA migration via API fields (`class: Network,Classic`, etc.) yet. it could possibly be added in a future kops minor version but a very low priority
* We'll need to support cleanup of both CLBs and NLBs in order to support migrating back and forth
* Regarding cleanup tasks: rather than having a dedicated cleanup task type, we could setup deletion support for each LB type in the opposite type's task. so CLB's `fi.Deletion` code will cleanup NLBs and vice versa.

* elbv2 (NLB) has a [WaitUntilLoadBalancerAvailable](https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#ELBV2.WaitUntilLoadBalancerAvailable) that we could block on before deleting the CLB. I dont know if this will have a material affect on apiserver downtime because DNS will still be updated immediately. We might be able to have the DNS update block until the NLB is available but that would be outside the scope of this PR. CLBs don't have anything like this so we wouldn't have the same availability guarantee when migrating from NLB -> CLB.

upup/pkg/fi/cloudup/awstasks/classic_load_balancer.go
```diff
@@ -80,6 +84,114 @@ func (e *ClassicLoadBalancer) CompareWithID() *string {
...
+
+	awsTarget, ok := t.(*awsup.AWSAPITarget)
+	if !ok {
+		return fmt.Errorf("unexpected target type for deletion: %T", t)
+	}
+
+	klog.V(2).Infof("Calling Nlb DeleteTargetGroup for %s", aws.StringValue(d.request.TargetGroupArn))
+
+	if _, err := awsTarget.Cloud.ELBV2().DeleteTargetGroup(d.request); err != nil {
+		return fmt.Errorf("error Deleting TargetGroup from NLB: %v", err)
```
rifelpet, line 112: We should ignore "target group not found" errors here, just like we do in FindDeletions

upup/pkg/fi/cloudup/awstasks/classic_load_balancer.go
```diff
@@ -80,6 +84,114 @@ func (e *ClassicLoadBalancer) CompareWithID() *string {
...
+
+	awsTarget, ok := t.(*awsup.AWSAPITarget)
+	if !ok {
+		return fmt.Errorf("unexpected target type for deletion: %T", t)
+	}
+
+	klog.V(2).Infof("Calling elb DeleteLoadBalancer for %s", aws.StringValue(d.request.LoadBalancerArn))
+
+	if _, err := awsTarget.Cloud.ELBV2().DeleteLoadBalancer(d.request); err != nil {
+		return fmt.Errorf("error deleting nlb %s: %v", aws.StringValue(d.request.LoadBalancerArn), err)
```
rifelpet, line 147: ignore "load balancer not found" errors

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -76,6 +78,67 @@ func (e *NetworkLoadBalancer) CompareWithID() *string {
...
+
+	awsTarget, ok := t.(*awsup.AWSAPITarget)
+	if !ok {
+		return fmt.Errorf("unexpected target type for deletion: %T", t)
+	}
+
+	klog.V(2).Infof("Calling elb DeleteClassicLoadBalancer for %s", aws.StringValue(d.request.LoadBalancerName))
+
+	if _, err := awsTarget.Cloud.ELB().DeleteLoadBalancer(d.request); err != nil {
+		return fmt.Errorf("error deleting clb %s: %v", aws.StringValue(d.request.LoadBalancerName), err)
```
rifelpet, line 106: ignore "load balancer not found" errors

upup/pkg/fi/cloudup/awstasks/classic_load_balancer.go
```diff
@@ -80,6 +84,114 @@ func (e *ClassicLoadBalancer) CompareWithID() *string {
...
+
+	cloud := c.Cloud.(awsup.AWSCloud)
+
+	for _, nlbName := range e.NLBNamesToDelete {
+		if lb, err := findNetworkLoadBalancerByLoadBalancerName(cloud, nlbName); err != nil {
+			if aerr, ok := err.(awserr.Error); ok && aerr.Code() != elbv2.ErrCodeLoadBalancerNotFoundException {
+				return nil, err
+			}
+		} else if lb != nil {
+			// Deleting a load balancer also deletes its listeners. NLB must be deleted first to delete TG.
```
rifelpet, line 168: I don't know how deletion tasks are processed: if they're ran in the order that they're returned in the slice and/or if they're retried multiple times until they succeed. I suppose testing will need to confirm this.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -176,7 +178,9 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 				IdleTimeout: fi.Int64(int64(idleTimeout.Seconds())),
 			},
 
-			Tags: tags,
+			Tags:             tags,
+			NLBNamesToDelete: []string{b.NLBName("api")},
+			TGNamesToDelete:  []string{b.NLBTargetGroupName("tcp"), b.NLBTargetGroupName("tls")},
```
hakman, line 183: I would prefer that these are not passed like this and just hardcoded in `FindDeletions`.

hakman (changes requested):

pkg/model/awsmodel/api_loadbalancer.go
hakman: @rifelpet any thoughts?

rifelpet: we explored this, but it would require either refactoring these methods and the methods they depend on to not depend on the model and move them to a common utility package that could be imported by both awsmodel and awstasks, or duplicate the logic which is pretty significant (NLBTargetGroupName calls [this](https://github.com/kubernetes/kops/blob/e91ed114499294f3e4242c6b23e7150a5c7262ec/pkg/model/context.go#L57-L99) for example)

Given that these fields wont ever be shown in the dryrun output I don't think its that bad of a solution, but maybe the refactoring would be a better idea.

upup/pkg/fi/cloudup/awstasks/classic_load_balancer.go
christianjoun: The order they are processed is the order that they're returned in the slice, and it appears they are retried multiple times until they succeed.

pkg/model/awsmodel/api_loadbalancer.go
hakman: I think would be best either to refactor or to just make a copy of `GetELBName32()` in https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/awsup/aws_utils.go.

Also, not sure if we should detect the ELB by name or by tags, but name should be ok for now.

PR #10187, opened 2020-11-07
Fix auto scaling group changes when using spot instances

hakman:
Fixes: #10082
Fixes: #10104

PR #10198, opened 2020-11-08
Fix disabling spot instances when using launch templates

hakman:
Some extra fixes after #10187, to be able to remove the SpotPrice when already set.

/cc @rifelpet

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
+++ b/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
@@ -456,8 +456,8 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 
 		if changes.LaunchTemplate != nil {
 			spec := &autoscaling.LaunchTemplateSpecification{
-				LaunchTemplateId: changes.LaunchTemplate.ID,
-				Version:          &launchTemplateVersion,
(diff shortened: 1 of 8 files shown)


upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -456,8 +456,8 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 
 		if changes.LaunchTemplate != nil {
 			spec := &autoscaling.LaunchTemplateSpecification{
-				LaunchTemplateId: changes.LaunchTemplate.ID,
-				Version:          &launchTemplateVersion,
+				LaunchTemplateName: changes.LaunchTemplate.ID,
```
rifelpet, line 459: I'm just curious, why is this change necessary?

hakman: There is some strange mix in the current code as you can see in other sections of the file:
https://github.com/kubernetes/kops/blob/b91a40c52aa67d10774d041754b74cc503513a44/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go#L379

This will be changed in https://github.com/kubernetes/kops/pull/10151 where I tested the fix initially and this is why the current "mismatch".

PR #10204, opened 2020-11-09
Allow custom cilium registry images

jorge07:
# Motivation

Prepare for new Docker hub restrictions 

# Description

Allow custom images for cilium

### DIFF
--- a/docs/networking/cilium.md
+++ b/docs/networking/cilium.md
@@ -138,6 +138,16 @@ Once the secret has been created, encryption can be enabled by setting `enableEn
       enableEncryption: true
 ```
 
+#### Custom images
+
+If you want to use other image repositories than the default, you can specify them like this:
+ ```yaml
+   networking:
+    cilium:
+      agentRepository: registry.dummy.com/cillium/cillium
+      operatorRepository: registry.dummy.com/cillium/operator
+      hubbleRelayRepository: registry.dummy.com/cillium/hubble-relay
+ ```
 
 ## Getting help
 
--- a/pkg/apis/kops/networking.go
+++ b/pkg/apis/kops/networking.go
@@ -244,7 +244,12 @@ const CiliumIpamEni = "eni"
 type CiliumNetworkingSpec struct {
 	// Version is the version of the Cilium agent and the Cilium Operator.
 	Version string `json:"version,omitempty"`
-
+	// AgentRepository specifies the repository for cilium agent
+	AgentRepository string `json:"agentRepository,omitempty"`
+	// OperatorRepository specifies the repository for cilium operator
+	OperatorRepository string `json:"operatorRepository,omitempty"`
+	// HubbleRelayRepository specifies the repository for cilium relay
+	HubbleRelayRepository string `json:"hubbleRelayRepository,omitempty"`
 	// AccessLog is not implemented and may be removed in the future.
 	// Setting this has no effect.
 	AccessLog string `json:"accessLog,omitempty"`
--- a/pkg/apis/kops/v1alpha2/networking.go
+++ b/pkg/apis/kops/v1alpha2/networking.go
@@ -242,7 +242,12 @@ type AmazonVPCNetworkingSpec struct {
 type CiliumNetworkingSpec struct {
 	// Version is the version of the Cilium agent and the Cilium Operator.
 	Version string `json:"version,omitempty"`
-
+	// AgentRepository specifies the repository for cilium agent
+	AgentRepository string `json:"agentRepository,omitempty"`
+	// OperatorRepository specifies the repository for cilium operator
+	OperatorRepository string `json:"operatorRepository,omitempty"`
+	// HubbleRelayRepository specifies the repository for cilium relay
+	HubbleRelayRepository string `json:"hubbleRelayRepository,omitempty"`
 	// AccessLog is not implemented and may be removed in the future.
 	// Setting this has no effect.
 	AccessLog string `json:"accessLog,omitempty"`
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.8.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.8.yaml.template
@@ -475,7 +475,7 @@ spec:
           value: {{ . }}
         {{ end }}
 {{ with .Networking.Cilium }}
-        image: "docker.io/cilium/cilium:{{ .Version  }}"
+        image: "{{ .AgentRepository }}:{{ .Version }}"
         imagePullPolicy: IfNotPresent
         lifecycle:
           postStart:
@@ -596,7 +596,7 @@ spec:
               key: wait-bpf-mount
               name: cilium-config
               optional: true
-        image: "docker.io/cilium/cilium:{{ .Version }}"
+        image: "{{ .AgentRepository }}:{{ .Version }}"
 ## end of `with .Networking.Cilium`
 #{{ end }}
         imagePullPolicy: IfNotPresent
@@ -761,7 +761,7 @@ spec:
         - name: KUBERNETES_SERVICE_PORT
           value: "443"
 {{ with .Networking.Cilium }}
-        image: "docker.io/cilium/operator:{{ .Version }}"
+        image: "{{ .OperatorRepository }}:{{ .Version }}"
         imagePullPolicy: IfNotPresent
         name: cilium-operator
         {{ if .EnablePrometheusMetrics }}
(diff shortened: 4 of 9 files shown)


olemarkus:
Technically the registry proxy feature is more applicable for the use case mentioned. However allowing to specify the image registry brings cilium addon inline with other addons that allows configuring this.

/ok-to-test

jorge07:
@olemarkus the goal it's to be able to run you own images. Like in other projects, we sometimes need to contribute and this require put things in production even when are not stable.

docs/networking/cilium.md
```diff
@@ -138,6 +138,16 @@ Once the secret has been created, encryption can be enabled by setting `enableEn
       enableEncryption: true
 ```
 
+#### Custom images
+
+`docker.io/cilium/cilium` it's the default image but there're cases, after Docker Hub T&C changes, where requirements needs another registry. You can set you own setting `image`, `operatorImage` or, `hubbleRelayImage`:
```
olemarkus, line 143: I would not refer to the Docker TOS here.

pkg/apis/kops/networking.go
```diff
@@ -244,7 +244,12 @@ const CiliumIpamEni = "eni"
 type CiliumNetworkingSpec struct {
 	// Version is the version of the Cilium agent and the Cilium Operator.
 	Version string `json:"version,omitempty"`
-
+	// Image specifies the registry and the image name
+	Image string `json:"image,omitempty"`
```
olemarkus, line 248: What you actually specify here is the image repository, not the image (as image also includes the tag)
I would also perhaps use `AgentRepository` to make what this is for more explicit

jorge07: Way better. Will change

docs/networking/cilium.md
jorge07: Will remove, no problem

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12.yaml.template
```diff
@@ -443,7 +443,7 @@ spec:
           value: {{ . }}
         {{ end }}
 {{ with .Networking.Cilium }}
-        image: "docker.io/cilium/cilium:{{ .Version  }}"
+        image: "{{- if .Image -}}{{ .Image }}{{- else -}}docker.io/cilium/cilium{{- end -}}:{{ .Version }}"
```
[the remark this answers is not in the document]
jorge07: Much better. Will Refactor

jorge07:
Thanks for the feedback @olemarkus, all applied

docs/networking/cilium.md
```diff
@@ -138,6 +138,16 @@ Once the secret has been created, encryption can be enabled by setting `enableEn
       enableEncryption: true
 ```
 
+#### Custom images
+
+`docker.io/cilium/cilium` it's the default agent image but there're cases where requirements needs another registry. You can set you own registry for `AgentRepository`, `OperatorRepository` or, `HubbleRelayRepository`:
```
olemarkus, line 143:
```suggestion
If you want to use other image repositories than the default, you can specify them like this:
```

pkg/apis/kops/v1alpha2/networking.go
```diff
@@ -242,7 +242,12 @@ type AmazonVPCNetworkingSpec struct {
 type CiliumNetworkingSpec struct {
 	// Version is the version of the Cilium agent and the Cilium Operator.
 	Version string `json:"version,omitempty"`
-
+	// AgentRepository specifies the registry for cilium agent
```
olemarkus, line 245:
```suggestion
	// AgentRepository specifies the repository for cilium agent
```

pkg/apis/kops/v1alpha2/networking.go
```diff
@@ -242,7 +242,12 @@ type AmazonVPCNetworkingSpec struct {
...
+	AgentRepository string `json:"agentRepository,omitempty"`
+	// OperatorRepository specifies the registry for cilium operator
```
olemarkus, line 247:
```suggestion
	// OperatorRepository specifies the repository for cilium operator
```

pkg/apis/kops/v1alpha2/networking.go
```diff
@@ -242,7 +242,12 @@ type AmazonVPCNetworkingSpec struct {
...
+	OperatorRepository string `json:"operatorRepository,omitempty"`
+	// HubbleRelayRepository specifies the registry for cilium relay
```
olemarkus, line 249:
```suggestion
	// HubbleRelayRepository specifies the repository for cilium relay
```

pkg/apis/kops/networking.go
```diff
@@ -244,7 +244,12 @@ const CiliumIpamEni = "eni"
 type CiliumNetworkingSpec struct {
 	// Version is the version of the Cilium agent and the Cilium Operator.
 	Version string `json:"version,omitempty"`
-
+	// AgentRepository specifies the registry for cilium agent
```
olemarkus, line 247:
```suggestion
	// AgentRepository specifies the repository for cilium agent
```

pkg/apis/kops/networking.go
```diff
@@ -244,7 +244,12 @@ const CiliumIpamEni = "eni"
...
+	AgentRepository string `json:"agentRepository,omitempty"`
+	// OperatorRepository specifies the registry for cilium operator
```
olemarkus, line 249:
```suggestion
	// OperatorRepository specifies the repository for cilium operator
```

pkg/apis/kops/networking.go
```diff
@@ -244,7 +244,12 @@ const CiliumIpamEni = "eni"
...
+	OperatorRepository string `json:"operatorRepository,omitempty"`
+	// HubbleRelayRepository specifies the registry for cilium relay
```
olemarkus, line 251:
```suggestion
	// HubbleRelayRepository specifies the repository for cilium relay
```

johngmyers:
Why the per-component duplication of the `spec.assets.containerRegistry` field? If you want to use a private registry that should be configured across the cluster, not on a component-by-component basis.

jorge07:
That sounds ideal but doesn't cover many things. 
Projects can move between registries, and they do, and have it global means all of them should be living in the same one. 
Another consideration are bug fixing that requires run a custom version for some time, this ones will be living elsewhere and you should be able to do it safely, as with the current implementation you can only do it manually and a rolling update overwrites your changes.

olemarkus:
I agree with @jorge07 on this one. usually you would want a custom agent, but leave the other components as is. Also, if you'd like  to use `cilium/cilium-dev` (which you have to in order to use arm right now) you have to specify repository like this.

olemarkus:
/lgtm

But approvers may want you to squash your commits.

jorge07:
Squashed and renamed the original commit to use the proper terms

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.8.yaml.template
```diff
@@ -886,7 +886,7 @@ spec:
             topologyKey: "kubernetes.io/hostname"
       containers:
         - name: hubble-relay
-          image: "docker.io/cilium/hubble-relay:{{ .Networking.Cilium.Version }}"
+          image: "{{ .HubbleRelayRepository }}:{{ .Networking.Cilium.Version }}"
```
hakman, line 889: Out of curiosity, any reason to use `.Networking.Cilium.Version` and not `.Version` here also?

jorge07: I did a find and replace and missed that, thanks, done

johngmyers:
If you want a custom image then you will likely need to specify a tag as well, not just a repository. The feature is not fit for that use case.

jorge07:
With the current implementation I can specify the tag/version. True that should be the same for the 3 but I can deal with that. Repositories are hardcoded, so that's the main reason. If you see more useful a version for component I can add it too, sounds very reasonable.

johngmyers:
Feedback from the informal Kops meeting of November 12:

* `spec.assets.containerRegistry` and the assets phase are the correct mechanism for addressing the use case of mirroring to a local registry, for example to avoid Docker Hub rate limits.
* Existing API fields for supporting custom images take the entire image path, including image name and tag. If supporting custom images is a requirement, any new fields should take the entire image path for consistency.
* A future API version might replace such image paths with a structure.
* One suggested alternative would be to add a general feature for replacing arbitrary images with custom images. Such a feature should not be conflated with assets phase, which should only address mirroring.

jorge07:
I'll try to get that suggestion done but I'll probably need some guidelines.
Context [in this pr](https://github.com/cilium/cilium/pull/14299)

olemarkus:
I suppose we can close this particular PR now?

We can certainly help out with the alternative suggestions

jorge07:
Let's address this in another PR as will not be even close to this one original idea.

PR #10217, opened 2020-11-11
Fix various nits

hakman:
* Validate external IAM policies
* Print changes also for consistency errors
* Handle cluster cleanup more gracefully

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -23,21 +23,21 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/blang/semver/v4"
 	"golang.org/x/net/ipv4"
 	"golang.org/x/net/ipv6"
-	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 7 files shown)


pkg/apis/kops/validation/validation.go
```diff
@@ -794,32 +800,52 @@ func validateNetworkingGCE(c *kops.ClusterSpec, v *kops.GCENetworkingSpec, fldPa
 	return allErrs
 }
 
-func validateAdditionalPolicy(role string, policy string, fldPath *field.Path) field.ErrorList {
-	errs := field.ErrorList{}
+func validateAdditionalPolicies(role string, policy string, fldPath *field.Path) field.ErrorList {
```
johngmyers, line 803: It's validating a singular policy, so the original name is correct.

johngmyers (changes requested):

pkg/apis/kops/validation/validation.go
hakman: Wanted to name it as the field it validates, but can change back.

hakman: Done.

PR #10221, opened 2020-11-11
Fix cluster validation dependency on local kubeconfig

eddycharly:
This PR fixes cluster validation dependency on a local kubeconfig file.

Cluster validation requires a local kubeconfig file with a context having the same name as the cluster.
This PR replaces this requirement by providing a `rest.Config`, the same that was used to build the `k8sclient` used to communicate with the target cluster.

### DIFF
--- a/pkg/validation/validate_cluster.go
+++ b/pkg/validation/validate_cluster.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/pager"
 	"k8s.io/kops/upup/pkg/fi"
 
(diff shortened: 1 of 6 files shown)


eddycharly:
Don't know why travis fails, it works fine locally. Is there a way to retrigger the job ?

johngmyers:
I wonder if we want to pass the whole config or just the Host.

eddycharly:
@johngmyers i wondered about that too. Maybe you’re right, passing only the Host will be simpler.

Thanks, I ’ll change that.

pkg/validation/validate_cluster.go
```diff
@@ -133,9 +126,7 @@ func (v *clusterValidatorImpl) Validate() (*ValidationCluster, error) {
 
 	// Do not use if we are running gossip
 	if !dns.IsGossipHostname(clusterName) {
```
eddycharly, line 128: Not sure if we should use `host` instead of `clusterName` here ?

olemarkus: If we are using gossip, won't `host` be the load balancer DNS?

eddycharly: You’re probably right, I don’t much about gossip mode.

PR #10222, opened 2020-11-11
Don't install the misc packages for k8s 1.20+

johngmyers:
I intend to make the boundary be 1.20, but the e2e tests are currently configured with 1.19.


hakman (commented):
> I intend to make the boundary be 1.20, but the e2e tests are currently configured with 1.19.

Any special reason not to start with 1.19?

I would move the "nfs-*" packages to `packages.go` and skip `miscutils.go` completely.

johngmyers:
It didn't seem that the benefit from this merited backporting to the 1.19 branch.

hakman:
Sounds good to me. Thanks.
/lgtm

Issue #10183, opened 2020-11-06
Identify associated instance group for all validation error during cluster validation


bharath-123:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
In validate_cluster, ValidationError describes any sort of failures that have occured during cluster validation. A recent PR was merged where we don't wait on rolling update due to validation errors in an unrelated instance group. This feature requires us to know the instance group of the validation error. 

Currently, it is not a straightforward task to know the instance group in which certain failures occur. Work would need to be done to identify the instance groups for the remaining validation errors in validate_cluster.

**2. Feel free to provide a design supporting your feature request.**

Would need to explore the code more myself to see if there are any cases where a mapping of Nodes to Instancegroup is already being done. and also pods to instance group.

bharath-123:
@johngmyers @hakman @justinsb Any suggestions or ideas which I can use here will be super helpful.

johngmyers:
`validateNodes` could construct and return a map of nodes to groups. `collectPodFailures` already constructs a map of IP address to node, so it could then use that map of nodes to groups to go from pod to group.

PR #10237, opened 2020-11-14
Associate instance group to pod validation failures in cluster validation.

bharath-123:
This PR fixes https://github.com/kubernetes/kops/issues/10183


johngmyers:
I believe that failures of pods of priorityClass "system-cluster-critical" are cluster-wide, so should not be marked as specific to their pod's instanceGroup.

Failures of pods of priorityClass "system-node-critical" are specific to the node, so should be marked as specific to their pods instanceGroup.

bharath-123:
Yes your right. Will take care of this. Thanks for pointing this out!

johngmyers:
Thanks!
/lgtm
/approve

PR #10241, opened 2020-11-15
Add HPA Flags for `horizontal-pod-autoscaler-initial-readiness-delay` & `horizontal-pod-autoscaler-cpu-initialization-period`

JoelBCarter:
Flags described in Kubernetes docs [here](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#algorithm-details)


bharath-123:
Was about to suggest to order the flags in alphabetical order but it looks like the entire file is not in alphabetical order. Shoud be fine but.

/lgtm

JoelBCarter:
> Was about to suggest to order the flags in alphabetical order but it looks like the entire file is not in alphabetical order. Shoud be fine but.
> 
> /lgtm

@bharath-123 I started down that same path and came to the same conclusion.

PR #10251, opened 2020-11-17
Delete security group rules that kops do not know about

olemarkus:
This PR will make kops remove all rules that _should_ be safe to remove. It will still ignore rules that belong to CCM or that links to other security groups.

The RemoveRules struct on SG tasks should also be redundant now.

### DIFF
--- a/pkg/model/awsmodel/firewall.go
+++ b/pkg/model/awsmodel/firewall.go
@@ -19,10 +19,14 @@ package awsmodel
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"k8s.io/kops/pkg/apis/kops"
+	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awstasks"
+	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
 
 	"k8s.io/klog/v2"
 )
@@ -295,6 +299,129 @@ func (b *FirewallModelBuilder) buildMasterRules(c *fi.ModelBuilderContext, nodeG
 	}
 
 	return masterGroups, nil
+}
+
+func (b *FirewallModelBuilder) FindDeletions(context *fi.ModelBuilderContext, cloud fi.Cloud) error {
+	awsCloud := cloud.(awsup.AWSCloud)
+
+	sgs, err := b.getClusterSecurityGroups(awsCloud)
+	if err != nil {
+		return err
+	}
+	for _, sg := range sgs {
+		name := fi.StringValue(sg.GroupName)
+		sgTask := context.Tasks["SecurityGroup/"+name].(*awstasks.SecurityGroup)
+
+		klog.V(4).Infof("found group %q with id %s", name, fi.StringValue(sg.GroupId))
+		// We assume that if the security group name ends with the cluster name it is owned by kOps.
+		// We must ignore security groups about which we don't know anything, as we cannot make tasks of them.
+		if !strings.HasSuffix(name, "."+b.ClusterName()) {
+			klog.V(4).Infof("skipping EC2 security group %q", name)
+			continue
+		}
+		for _, rule := range sg.IpPermissions {
+			ts := b.createRulesFromPermissions(sgTask, rule, false, sgs)
+			for _, task := range ts {
+				if _, ok := context.Tasks["SecurityGroupRule/"+*task.Name]; !ok {
+					if !featureflag.DeleteUnknownSGRules.Enabled() {
+						klog.Warningf("found unknown security group permission: %v", fi.DebugAsJsonString(task))
+						klog.Warning("you can delete this rule by using the DeleteUnknownSGRules feature gate")
+						klog.Warning(("a future version of kOps will remove this automatically"))
+					} else {
+						context.AddTask(task)
+					}
+				}
+			}
+		}
+		for _, rule := range sg.IpPermissionsEgress {
+			ts := b.createRulesFromPermissions(sgTask, rule, true, sgs)
+			for _, task := range ts {
+				if _, ok := context.Tasks["SecurityGroupRule/"+*task.Name]; !ok {
+					if !featureflag.DeleteUnknownSGRules.Enabled() {
+						klog.Warningf("found unknown security group permission: %v", fi.DebugAsJsonString(task))
+						klog.Warning("you can delete this rule by using the DeleteUnknownSGRules feature gate")
+						klog.Warning(("a future version of kOps will remove this automatically"))
+					} else {
+						context.AddTask(task)
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func (b *FirewallModelBuilder) getClusterSecurityGroups(cloud awsup.AWSCloud) (map[string]*ec2.SecurityGroup, error) {
+	sgs := make(map[string]*ec2.SecurityGroup)
+
+	request := &ec2.DescribeSecurityGroupsInput{
+		Filters: []*ec2.Filter{
+			awsup.NewEC2Filter("tag:KubernetesCluster", b.ClusterName()),
+		},
+	}
+
+	response, err := cloud.EC2().DescribeSecurityGroups(request)
+	if err != nil {
+		return nil, fmt.Errorf("error listing EC2 security groups: %w", err)
+	}
+
+	for _, sg := range response.SecurityGroups {
+		sgs[fi.StringValue(sg.GroupId)] = sg
+	}
+	return sgs, nil
+
+}
+
+func (b *FirewallModelBuilder) createRulesFromPermissions(sg *awstasks.SecurityGroup, rule *ec2.IpPermission, egress bool, sgs map[string]*ec2.SecurityGroup) (tasks []*awstasks.SecurityGroupRule) {
+	for _, cidr := range rule.IpRanges {
+		// CIDRs with description starting with kubernetes.io are owned by CCM
+		if strings.HasPrefix(fi.StringValue(cidr.Description), "kubernetes.io") {
+			continue
+		}
+		rule := b.createBaseRuleForDeletion(sg, rule, egress)
+		rule.Egress = fi.Bool(egress)
+		rule.CIDR = cidr.CidrIp
+
+		rule.Name = fi.String(generateName(rule))
+		tasks = append(tasks, rule)
+	}
+	for _, p := range rule.UserIdGroupPairs {
+		rule := b.createBaseRuleForDeletion(sg, rule, egress)
+		rule.Egress = fi.Bool(egress)
+		pGroup := sgs[fi.StringValue(p.GroupId)]
+		groupName := fi.StringValue(pGroup.GroupName)
+		// We assume that if the security group name ends with the cluster name it is owned by kOps.
+		// We must ignore security groups about which we don't know anything, as we cannot make tasks of them.
+		if !strings.HasSuffix(groupName, b.ClusterName()) {
+			klog.V(4).Infof("Skipping rule; target EC2 security group %q not owned by kOps", groupName)
+			continue
+		}
+		source := &awstasks.SecurityGroup{Name: pGroup.GroupName}
+
+		rule.SourceGroup = source
+
+		rule.Name = fi.String(generateName(rule))
+		tasks = append(tasks, rule)
+	}
+	return tasks
+}
+
+func (b *FirewallModelBuilder) createBaseRuleForDeletion(sg *awstasks.SecurityGroup, rule *ec2.IpPermission, egress bool) *awstasks.SecurityGroupRule {
+
+	task := &awstasks.SecurityGroupRule{
+		SecurityGroup: sg,
+		FromPort:      rule.FromPort,
+		ToPort:        rule.ToPort,
+		Protocol:      rule.IpProtocol,
+		Egress:        fi.Bool(egress),
+		Lifecycle:     sg.Lifecycle,
+		Delete:        fi.Bool(true),
+	}
+	if fi.StringValue(task.Protocol) == "-1" {
+		task.Protocol = nil
+	}
+
+	return task
 }
 
 type SecurityGroupInfo struct {
--- a/tests/integration/update_cluster/bastionadditional_user-data/kubernetes.tf
+++ b/tests/integration/update_cluster/bastionadditional_user-data/kubernetes.tf
@@ -942,6 +942,15 @@ resource "aws_security_group" "nodes-bastionuserdata-example-com" {
   vpc_id = aws_vpc.bastionuserdata-example-com.id
 }
 
+resource "aws_security_group_rule" "from-0-0-0-0--0-ingress-icmp-3to4-api-elb-bastionuserdata-example-com" {
+  cidr_blocks       = ["0.0.0.0/0"]
+  from_port         = 3
+  protocol          = "icmp"
+  security_group_id = aws_security_group.api-elb-bastionuserdata-example-com.id
+  to_port           = 4
+  type              = "ingress"
+}
+
 resource "aws_security_group_rule" "from-0-0-0-0--0-ingress-tcp-22to22-bastion-elb-bastionuserdata-example-com" {
   cidr_blocks       = ["0.0.0.0/0"]
   from_port         = 22
@@ -976,6 +985,15 @@ resource "aws_security_group_rule" "from-api-elb-bastionuserdata-example-com-egr
   security_group_id = aws_security_group.api-elb-bastionuserdata-example-com.id
   to_port           = 0
   type              = "egress"
+}
+
+resource "aws_security_group_rule" "from-api-elb-bastionuserdata-example-com-ingress-tcp-443to443-masters-bastionuserdata-example-com" {
+  from_port                = 443
+  protocol                 = "tcp"
+  security_group_id        = aws_security_group.masters-bastionuserdata-example-com.id
+  source_security_group_id = aws_security_group.api-elb-bastionuserdata-example-com.id
+  to_port                  = 443
+  type                     = "ingress"
 }
 
 resource "aws_security_group_rule" "from-bastion-bastionuserdata-example-com-egress-all-0to0-0-0-0-0--0" {
@@ -1138,24 +1156,6 @@ resource "aws_security_group_rule" "from-nodes-bastionuserdata-example-com-ingre
   source_security_group_id = aws_security_group.nodes-bastionuserdata-example-com.id
   to_port                  = 65535
   type                     = "ingress"
-}
-
-resource "aws_security_group_rule" "https-elb-to-master" {
-  from_port                = 443
-  protocol                 = "tcp"
-  security_group_id        = aws_security_group.masters-bastionuserdata-example-com.id
-  source_security_group_id = aws_security_group.api-elb-bastionuserdata-example-com.id
-  to_port                  = 443
-  type                     = "ingress"
-}
-
-resource "aws_security_group_rule" "icmp-pmtu-api-elb-0-0-0-0--0" {
-  cidr_blocks       = ["0.0.0.0/0"]
-  from_port         = 3
-  protocol          = "icmp"
-  security_group_id = aws_security_group.api-elb-bastionuserdata-example-com.id
-  to_port           = 4
-  type              = "ingress"
 }
 
 resource "aws_subnet" "us-test-1a-bastionuserdata-example-com" {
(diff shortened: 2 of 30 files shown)


pkg/model/firewall.go
```diff
@@ -36,12 +39,25 @@ const (
...
 func (b *FirewallModelBuilder) Build(c *fi.ModelBuilderContext) error {
+
+	tasks, err := b.getExistingRulesFromCloud()
+	if err != nil {
+		return err
+	}
+
+	finalRules := make(map[string]*awstasks.SecurityGroupRule)
+	for _, task := range tasks {
+		klog.Infof("found rule %s", fi.StringValue(task.Name))
```
olemarkus, line 57: Should add `V(8)` to these info lines

olemarkus:
Tested this on a couple of our older clusters that have gone through various changes. It wants to delete rules that are no longer needed, and it keeps the rules that should be kept.

docs/releases/1.20-NOTES.md
```diff
@@ -4,6 +4,8 @@
 
 # Significant changes
 
+* kOps will reconsile security group rules for groups that kOps owns. If you have manually added rules to these groups, you need to move these to a new securiy group and add this group using `spec.additionalSecurityGroups` on the relevant Instance Groups.
```
seh, line 7:
```suggestion
* kOps will reconcile security group rules for groups that kOps owns. If you have manually added rules to these groups, you need to move these to a new security group and add this group using `spec.additionalSecurityGroups` on the relevant Instance Groups.
```

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
 }
+
+func (b *FirewallModelBuilder) getExistingRulesFromCloud() ([]*awstasks.SecurityGroupRule, error) {
+	tasks := []*awstasks.SecurityGroupRule{}
+	sgs, err := b.getClusterSecurityGroups()
+	if err != nil {
+		return nil, err
+	}
+	for _, sg := range sgs {
+
```
seh, line 481: Nit: This blank line is distracting.

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+		}
+		for _, rule := range sg.IpPermissions {
+			ts := b.createRulesFromPermissions(sg, rule, false, sgs)
+			tasks = append(tasks, ts...)
+		}
+		for _, rule := range sg.IpPermissionsEgress {
+			ts := b.createRulesFromPermissions(sg, rule, true, sgs)
+			tasks = append(tasks, ts...)
+		}
+
```
seh, line 499: Nit: This blank line is distracting.

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+	if err != nil {
+		return nil, err
+	}
+	for _, sg := range sgs {
+
+		name := fi.StringValue(sg.GroupName)
+
+		klog.V(4).Infof("found group %s, %s", name, fi.StringValue(sg.GroupId))
+		// sec groups we don't know anything about must be ignored as we cannot make tasks of them
+		// and we assume that if the SG ends with the clustername it is owner by kOps.
```
seh, line 486:
```suggestion
		// We assume that if the security group name ends with the cluster name it is owned by kOps.
		// We must ignore security groups about which we don't know anything, as we cannot make tasks of them.
```

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+		if !strings.HasSuffix(name, "."+b.ClusterName()) {
+			klog.V(4).Infof("skipping group %s", name)
```
seh, line 488:
```suggestion
			klog.V(4).Infof("skipping EC2 security group %q", name)
```

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+	response, err := b.Cloud.EC2().DescribeSecurityGroups(request)
+	if err != nil {
+		return nil, fmt.Errorf("error listing SecurityGroup: %v", err)
+	}
+
+	for _, sg := range response.SecurityGroups {
+		sgs[fi.StringValue(sg.GroupId)] = sg
+	}
+	return sgs, nil
+
```
seh, line 522: Nit: This blank line is distracting.

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+	request := &ec2.DescribeSecurityGroupsInput{
+		Filters: []*ec2.Filter{
+			awsup.NewEC2Filter("tag:KubernetesCluster", b.ClusterName()),
+		},
+	}
+
+	response, err := b.Cloud.EC2().DescribeSecurityGroups(request)
+	if err != nil {
+		return nil, fmt.Errorf("error listing SecurityGroup: %v", err)
```
seh, line 515:
```suggestion
		return nil, fmt.Errorf("error listing EC2 security groups: %w", err)
```

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+	for _, sg := range response.SecurityGroups {
+		sgs[fi.StringValue(sg.GroupId)] = sg
+	}
+	return sgs, nil
+
+}
+
+func (b *FirewallModelBuilder) createRulesFromPermissions(sg *ec2.SecurityGroup, rule *ec2.IpPermission, egress bool, sgs map[string]*ec2.SecurityGroup) (tasks []*awstasks.SecurityGroupRule) {
+	for _, cidr := range rule.IpRanges {
+
```
seh, line 527: Nit: This blank line is distracting.

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+		rule.Name = fi.String(generateName(rule))
+		tasks = append(tasks, rule)
+	}
+	for _, p := range rule.UserIdGroupPairs {
+		rule := b.createBaseRule(sg, rule, egress)
+		rule.Egress = fi.Bool(egress)
+		pGroup := sgs[fi.StringValue(p.GroupId)]
+		groupName := fi.StringValue(pGroup.GroupName)
+		// sec groups we don't know anything about must be ignored as we cannot make tasks of them
+		// and we assume that if the SG ends with the clustername it is owner by kOps.
```
seh, line 545:
```suggestion
		// We assume that if the security group name ends with the cluster name it is owned by kOps.
		// We must ignore security groups about which we don't know anything, as we cannot make tasks of them.
```

pkg/model/firewall.go
```diff
@@ -442,6 +467,108 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+		if !strings.HasSuffix(groupName, b.ClusterName()) {
+			klog.V(4).Infof("Skipping rule %s, target group not owned by kops", groupName)
```
seh, line 547:
```suggestion
			klog.V(4).Infof("Skipping rule; target EC2 security group %q not owned by kOps", groupName)
```

upup/pkg/fi/cloudup/awstasks/securitygrouprule.go
```diff
@@ -47,10 +47,11 @@ type SecurityGroupRule struct {
 	SourceGroup *SecurityGroup
 
 	Egress *bool
+
+	Delete *bool
 }
 
 func (e *SecurityGroupRule) Find(c *fi.Context) (*SecurityGroupRule, error) {
-	cloud := c.Cloud.(awsup.AWSCloud)
```
seh, line 53: Nit: Consider deleting the blank line that follows the line you deleted, so the method body starts without a blank line as before.

upup/pkg/fi/cloudup/awstasks/securitygrouprule.go
```diff
@@ -394,3 +409,17 @@ func (_ *SecurityGroupRule) RenderCloudformation(t *cloudformation.Cloudformatio
...
+		return nil, nil
+	}
+
+	return []fi.Deletion{&deleteSecurityGroupRule{
+		groupID:    e.SecurityGroup.ID,
+		permission: e.createIpPermission(),
+		egress:     fi.BoolValue(e.Egress),
+	},
+	}, nil
+
```
seh, line 424: Nit: This blank line is distracting.

pkg/model/firewall.go
```diff
@@ -36,12 +39,25 @@ const (
...
 type FirewallModelBuilder struct {
 	*KopsModelContext
+	Cloud     awsup.AWSCloud
 	Lifecycle *fi.Lifecycle
 }
 
 var _ fi.ModelBuilder = &FirewallModelBuilder{}
 
 func (b *FirewallModelBuilder) Build(c *fi.ModelBuilderContext) error {
+
```
seh, line 49: Nit: The blank line at the beginning of the method body looks odd to me.

pkg/model/firewall.go
```diff
@@ -36,12 +39,25 @@ const (
...
+	tasks, err := b.getExistingRulesFromCloud()
+	if err != nil {
+		return err
+	}
+
+	finalRules := make(map[string]*awstasks.SecurityGroupRule)
```
seh, line 55:
```suggestion
	finalRules := make(map[string]*awstasks.SecurityGroupRule, len(tasks))
```
Though I suppose that should really be the maximum of `len(tasks)` and `len(b.SecurityGroupRules)`.

pkg/model/firewall.go
```diff
@@ -59,6 +75,14 @@ func (b *FirewallModelBuilder) Build(c *fi.ModelBuilderContext) error {
 	// b.applyNodeToMasterAllowSpecificPorts(c)
 	b.applyNodeToMasterBlockSpecificPorts(c, nodeGroups, masterGroups)
 
+	for _, task := range b.SecurityGroupRules {
+		finalRules[fi.StringValue(task.Name)] = task
```
seh, line 79: It looks like the intention here is to replace an existing rule that might already be in the map per the call to `(*FirewallModelBuilder).getExistingRulesFromCloud` above.

pkg/model/firewall.go
```diff
@@ -408,10 +432,11 @@ func JoinSuffixes(src SecurityGroupInfo, dest SecurityGroupInfo) string {
 func (b *KopsModelContext) AddDirectionalGroupRule(c *fi.ModelBuilderContext, t *awstasks.SecurityGroupRule) {
 
 	name := generateName(t)
+	b.SecurityGroupRules[name] = t
 	t.Name = fi.String(name)
+	t.Delete = fi.Bool(false)
 
-	klog.V(8).Infof("Adding rule %v", name)
-	c.AddTask(t)
+	klog.V(4).Infof("Adding rule %v", name)
```
seh, line 439:
```suggestion
	klog.V(4).Infof("Adding rule %q", name)
```
Consider moving this line up to between lines 434 and 435, once we have the generated name in hand.

pkg/model/firewall.go
```diff
@@ -36,12 +39,25 @@ const (
...
 func (b *FirewallModelBuilder) Build(c *fi.ModelBuilderContext) error {
+
+	tasks, err := b.getExistingRulesFromCloud()
+	if err != nil {
+		return err
+	}
+
+	finalRules := make(map[string]*awstasks.SecurityGroupRule)
+	for _, task := range tasks {
+		klog.V(4).Infof("found rule %s", fi.StringValue(task.Name))
```
seh, line 57:
```suggestion
		klog.V(4).Infof("Found rule %q", fi.StringValue(task.Name))
```

pkg/model/firewall.go
```diff
@@ -445,3 +470,102 @@ func generateName(o *awstasks.SecurityGroupRule) string {
 	return fmt.Sprintf("from-%s-%s-%s-%dto%d-%s", src, direction,
 		proto, fi.Int64Value(o.FromPort), fi.Int64Value(o.ToPort), dst)
 }
+
+func (b *FirewallModelBuilder) getExistingRulesFromCloud() ([]*awstasks.SecurityGroupRule, error) {
+	tasks := []*awstasks.SecurityGroupRule{}
```
seh, line 475: Consider moving this line down to just before line 480, then:
```suggestion
	// Estimate the number of rules per security group.
	tasks := make([]*awstasks.SecurityGroupRule{}, 0, len(sgs)*4)
```

pkg/model/firewall.go
```diff
@@ -445,3 +470,102 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+	sgs, err := b.getClusterSecurityGroups()
+	if err != nil {
+		return nil, err
+	}
+	for _, sg := range sgs {
+		name := fi.StringValue(sg.GroupName)
+
+		klog.V(4).Infof("found group %s, %s", name, fi.StringValue(sg.GroupId))
```
seh, line 483:
```suggestion
		klog.V(4).Infof("Found group %q (with ID %s)", name, fi.StringValue(sg.GroupId))
```
I'm not sure of our convention for capitalizing the fist word in these log messages; I see variation throughout these files.

pkg/model/firewall.go
```diff
@@ -445,3 +470,102 @@ func generateName(o *awstasks.SecurityGroupRule) string {
...
+		for _, rule := range sg.IpPermissionsEgress {
+			ts := b.createRulesFromPermissions(sg, rule, true, sgs)
+			tasks = append(tasks, ts...)
+		}
+	}
+	return tasks, nil
+}
+
+func (b *FirewallModelBuilder) getClusterSecurityGroups() (map[string]*ec2.SecurityGroup, error) {
+	sgs := make(map[string]*ec2.SecurityGroup)
```
seh, line 503: Consider moving this line down after line 514, then:
```suggestion
	sgs := make(map[string]*ec2.SecurityGroup, len(response.SecurityGroups))
```

upup/pkg/fi/cloudup/awstasks/securitygroup.go
```diff
@@ -268,34 +269,39 @@ type deleteSecurityGroupRule struct {
...
+			request.IpPermissions = []*ec2.IpPermission{d.permission}
 
-		klog.V(2).Infof("Calling EC2 RevokeSecurityGroupIngress")
-		_, err := awsTarget.Cloud.EC2().RevokeSecurityGroupIngress(request)
-		if err != nil {
-			return fmt.Errorf("error revoking SecurityGroupIngress: %v", err)
+			klog.V(2).Infof("Calling EC2 RevokeSecurityGroupEgress")
+			_, err := awsTarget.Cloud.EC2().RevokeSecurityGroupEgress(request)
+			if err != nil {
+				return fmt.Errorf("error revoking SecurityGroupEgress: %v", err)
```
seh, line 292: Consider wrapping the error:
```suggestion
				return fmt.Errorf("error revoking security group egress: %w", err)
```

upup/pkg/fi/cloudup/awstasks/securitygroup.go
```diff
@@ -268,34 +269,39 @@ type deleteSecurityGroupRule struct {
...
+		} else {
+			request := &ec2.RevokeSecurityGroupIngressInput{
+				GroupId: d.groupID,
+			}
+			request.IpPermissions = []*ec2.IpPermission{d.permission}
+
+			klog.V(2).Infof("Calling EC2 RevokeSecurityGroupIngress")
+			_, err := awsTarget.Cloud.EC2().RevokeSecurityGroupIngress(request)
+			if err != nil {
+				return fmt.Errorf("error revoking SecurityGroupIngress: %v", err)
```
seh, line 303: Consider wrapping the error:
```suggestion
				return fmt.Errorf("error revoking security group ingress: %w", err)
```

upup/pkg/fi/cloudup/awstasks/securitygrouprule.go
```diff
@@ -228,37 +238,42 @@ func (e *SecurityGroupRule) Description() string {
...
+		CIDR := e.CIDR
+		// Default to 0.0.0.0/0 ?
+		ipPermission.IpRanges = []*ec2.IpRange{
+			{CidrIp: CIDR},
 		}
+	}
 
-		description := e.Description()
+	return ipPermission
+
```
seh, line 268: Nit: Delete blank line.

pkg/model/firewall.go
olemarkus: Correct

pkg/model/firewall.go
olemarkus: We don't have a convention, but I grew up following `fmt.errorf` convention.

seh: Thank you for clarifying. Please ignore my earlier suggestions to capitalize the first word in these messages.

olemarkus: Will do. I think I fixed the remaining bits

johngmyers:
I'm not a fan of keeping a side list of `SecurityGroupRule` tasks. The `ModelBuilderContext.Tasks` map is the canonical set of expected tasks. Keeping a side list invites errors where code a `SecurityGroupRule` to `Tasks` but not the side list.

johngmyers:
Consider refactoring to use the new `fi.HasDeletions` interface of #11497.

olemarkus:
Now refactored to use the `fi.HasDeletions` interface.

rifelpet:
I'm wondering if this could be simplified if we relied on the recently-introduced SecurityGroupRuleIds. I started to migrate our SecurityGroupRule task to use IDs but got caught up in integration test failures, likely due to cloudmock: https://github.com/kubernetes/kops/compare/master...rifelpet:sg-rule-ids if we can get that merged then we could rely on the ID field of the SecurityGroupRule tasks in the `RevokeSecurityGroup*` calls

pkg/model/awsmodel/firewall.go
```diff
@@ -297,6 +301,129 @@ func (b *FirewallModelBuilder) buildMasterRules(c *fi.ModelBuilderContext, nodeG
...
+	}
+	return nil
+}
+
+func (b *FirewallModelBuilder) getClusterSecurityGroups(cloud awsup.AWSCloud) (map[string]*ec2.SecurityGroup, error) {
+	sgs := make(map[string]*ec2.SecurityGroup)
+
+	request := &ec2.DescribeSecurityGroupsInput{
+		Filters: []*ec2.Filter{
+			awsup.NewEC2Filter("tag:KubernetesCluster", b.ClusterName()),
```
rifelpet, line 359: maybe we filter on the owned tag too? that might be specific enough to not need to check the SG name's suffix matching the cluster name above.

https://github.com/kubernetes/kops/blob/fc91d0d459d28630157fd49a370525b482153482/tests/integration/update_cluster/complex/kubernetes.tf#L803

pkg/model/awsmodel/firewall.go
```diff
@@ -297,6 +301,129 @@ func (b *FirewallModelBuilder) buildMasterRules(c *fi.ModelBuilderContext, nodeG
...
+		sgs[fi.StringValue(sg.GroupId)] = sg
+	}
+	return sgs, nil
+
+}
+
+func (b *FirewallModelBuilder) createRulesFromPermissions(sg *awstasks.SecurityGroup, rule *ec2.IpPermission, egress bool, sgs map[string]*ec2.SecurityGroup) (tasks []*awstasks.SecurityGroupRule) {
+	for _, cidr := range rule.IpRanges {
+		// CIDRs with description starting with kubernetes.io are owned by CCM
+		if strings.HasPrefix(fi.StringValue(cidr.Description), "kubernetes.io") {
```
rifelpet, line 378: what about other controllers like LBC? or if someone has a custom controller that is managing SG rules, how could they opt out of this behavior once we enable the FF by default?

tests/integration/update_cluster/bastionadditional_user-data/kubernetes.tf
```diff
@@ -942,6 +942,15 @@ resource "aws_security_group" "nodes-bastionuserdata-example-com" {
   vpc_id = aws_vpc.bastionuserdata-example-com.id
 }
 
+resource "aws_security_group_rule" "from-0-0-0-0--0-ingress-icmp-3to4-api-elb-bastionuserdata-example-com" {
```
rifelpet, line 945: these renames are unfortunate. not only because the apply might cause a temporary loss in network connectivity, but because `terraform` may run into consistency conflicts since we're deleting and creating tf resources that represent the same underlying AWS resource. I've had this happen many times, albeit not with SG rules specifically:

* terraform attempts to delete the rule (old resource) and create the same rule (new resource)
* creation fails because the rule already exists
* deletion succeeds
* terraform doesn't retry the creation because it isnt an error that would normally succeed if retried
* `terraform apply` aborts
* rerun `terraform apply` a second time, the plan reports _only_ creating the new rule. the creation succeeds 

We can always suggest yet another `terraform state mv` (seems to be a regular occurrence with every kOps release these days)

olemarkus:
Maybe I'll pick this up later again. But this PR has drifted quite far from master, and I'll probably redo it from scratch.

PR #10258, opened 2020-11-18
Add support of Azure Blob storage to VFS

kenji-cloudnatix:
The schema is "azureblob".

azureClient provides two ways to set up credential. One approach is to
use an account key stored in env var. This approach is used when
accessing Blob from kops CLI. The second approach is to retrieve
credentials from Instance Metadata Service. This works only when
azureClient is created on a VM that has sufficient privilege to access
a specified blob. This approach is used from nodeup, etcd-manager,
etc.


util/pkg/vfs/azureblob.go
```diff
@@ -0,0 +1,245 @@
...
+
+// AzureBlobPath is a path in the VFS space backed by Azure Blob.
+type AzureBlobPath struct {
+	client    *azureClient
+	container string
+	key       string
+}
+
+var _ Path = &AzureBlobPath{}
+var _ HasHash = &S3Path{}
```
hakman, line 43: I guess it won't be too different from the S3 one, but still best to keep consistent.
```suggestion
var _ HasHash = &AzureBlobPath{}
```

util/pkg/vfs/azureblob.go
```diff
@@ -0,0 +1,245 @@
...
+	"io"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/Azure/azure-storage-blob-go/azblob"
+)
+
+const (
+	azureBlobSchema = "azureblob"
```
hakman, line 32: Pretty much every other supported storage has this inlined. I think would be best to inline this also and add constants for all in a different PR.

hakman (changes requested):
Looks like a great start for merging the big PR. Just a few nits from me.

util/pkg/vfs/azureblob.go
kenji-cloudnatix: Sure thing! Will follow the convention.

kenji-cloudnatix (commented):
Thanks for the quick review, @hakman ! Changed `AzureBlobPath` to implement `HashHash`. Also inlined the cost as suggested.

util/pkg/vfs/azureclient.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, err
+	}
+	// Create a ContainerURL object that wraps the container URL and a request
+	// pipeline to make requests.
+	cURL := azblob.NewContainerURL(*u, c.p)
+	return &cURL, nil
+}
+
+func newAzureClient() (*azureClient, error) {
+	accountName := os.Getenv("AZURE_STORAGE_ACCOUNT")
```
rifelpet, line 54: These environment variable names seem unusual to me but I confirmed the azure-cli uses them so it's best to be consistent :)

rifelpet (changes requested):

util/pkg/vfs/azureblob.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+		return 0, err
+	}
+	return io.Copy(w, resp.Body(azblob.RetryReaderOptions{MaxRetryRequests: 10}))
+}
+
+// CreateFile writes the file contents only if the file does not already exist.
+//
+// TODO(kenji): Handle a case where CreateFile is called concurrently to the same blob.
+func (p *AzureBlobPath) CreateFile(data io.ReadSeeker, acl ACL) error {
```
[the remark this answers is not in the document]
kenji-cloudnatix: Done!

util/pkg/vfs/azureblob.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return err
+	}
+	// Use block blob. Other options are page blobs (optimized for
+	// random read/write) and append blob (optimized for append).
+	_, err = cURL.NewBlockBlobURL(p.key).Upload(
+		context.TODO(),
+		data,
+		azblob.BlobHTTPHeaders{
+			ContentType: "text/plain",
```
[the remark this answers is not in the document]
kenji-cloudnatix: I see. Sure, updated!

util/pkg/vfs/azureclient.go
kenji-cloudnatix: Ah, yes.. :) Any suggestion appreciated here. I decided to use the env var as we need to pass the information to nodeup, protokube, and all other components that need to access a cluster config.

kenji-cloudnatix (commented):
Thanks for the review, @rifelpet !

rifelpet:
This looks good to me but I'll wait for input from others before merging. We can bring it up at the Kops office hours tomorrow
/lgtm

hakman:
I am also OK with this.
/lgtm

kenji-cloudnatix:
Thanks, both! How should I merge? Rebase and squash all the commits into one?

hakman:
We will have a short review on Friday during office hours and probably the final LGTM.
I think would be a good idea to squash, so that it can be merged if no other issues are found.

kenji-cloudnatix:
Sounds good! Squashed. Will also join the office hour.

hakman:
> Sounds good! Squashed. Will also join the office hour.

Perfect. Thanks!

util/pkg/vfs/context.go
```diff
@@ -486,3 +491,42 @@ func (c *VFSContext) buildVaultPath(p string) (*VaultPath, error) {
...
 }
+
+func (c *VFSContext) buildAzureBlobPath(p string) (*AzureBlobPath, error) {
+	u, err := url.Parse(p)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse %q: %s", p, err)
+	}
+
+	if u.Scheme != "azureblob" {
+		return nil, fmt.Errorf("invalid Azure Blob schem: %q", p)
```
justinsb, line 502: Nit: scheme

justinsb:
This looks great - so excited to see this - thanks @kenji-cloudnatix and all the reviewers :-)

/approve
/lgtm

util/pkg/vfs/context.go
kenji-cloudnatix: Ah thanks for catching! Will create a small PR to fix.

PR #10261, opened 2020-11-18
Explicitly set the target environment

hakman:
TravisCI uses AMD64 silently instead of ARM64 at the moment.

/cc @olemarkus


hakman:
Closing in favor of #10262.

PR #10240, opened 2020-11-15
Update Bazel rules for Go to v0.24.7

hakman:
/cc @justinsb @mikesplain @rifelpet

PR #10267, opened 2020-11-18
Optimize Bazel builds by os and arch

hakman:
Latest Bazel changes invalidate the build cache when switching the arch, so build is slower. Grouping artifacts by arch speeds things up. This did not happen before https://github.com/kubernetes/kops/pull/10240.

https://testgrid.k8s.io/kops-misc#kops-postsubmit-push-to-staging

/cc @rifelpet


rifelpet:
this took a while to trace through all the changes but it makes sense to me
/lgtm

hakman:
Thanks a lot @rifelpet :)

Issue #10268, opened 2020-11-18
Can't generate terraform when using externalLoadBalancers in two IGs


kuzaxak:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

```
kops version
Version 1.19.0-beta.1 (git-6fa37fc5bdc2eedd7f0c696dd8155ac267b310fd)
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

1.18

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

`kops update cluster --target=terraform  --out=update --yes`

Create two IG with externalLoadBalancers -> targetGroupArn

**5. What happened after the commands executed?**

Fatal error:

```
F1118 23:54:29.146441   64478 task.go:59] found duplicate tasks with name "TargetGroup/external-tg-0": *awstasks.TargetGroup {"Name":"external-tg-0","Lifecycle":null,"VPC":null,"Tags":null,"Port":null,"Protocol":null,"ARN":"arn:aws:elasticloadbalancing:eu-west-1:*******:targetgroup/nodes-ingress-http/55f391*******","Shared":true,"HealthyThreshold":null,"UnhealthyThreshold":null} and *awstasks.TargetGroup {"Name":"external-tg-0","Lifecycle":null,"VPC":null,"Tags":null,"Port":null,"Protocol":null,"ARN":"arn:aws:elasticloadbalancing:eu-west-1:*******:targetgroup/jenkins-nlb/dac3*******","Shared":true,"HealthyThreshold":null,"UnhealthyThreshold":null}
```

**6. What did you expect to happen?**

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2018-10-15T17:39:23Z"
  generation: 10
  name: %%%%%%%%
spec:
  api:
    loadBalancer:
      class: Classic
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudProvider: aws
  configBase: s3://XXXXXXXX/%%%%%%%%
  dnsZone: %%%%%%%%
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-eu-west-1a
[...]

PR #10276, opened 2020-11-20
Parse TargetGroup names from ARNs

hakman:
Fixes: #10268 
Summary of fixes:
1. duplicate tasks names because the names for external TGs are the same for each IG
2. model names differ from actual names so always show changes
3. ordering is required to not show false changes

### DIFF
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -18,8 +18,11 @@ package awsmodel
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/featureflag"
@@ -28,8 +31,6 @@ import (
 	"k8s.io/kops/pkg/model/spotinstmodel"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awstasks"
-
-	"github.com/aws/aws-sdk-go/service/ec2"
 )
 
 const (
@@ -388,7 +389,7 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 		}
 	}
 
-	for i, extLB := range ig.Spec.ExternalLoadBalancers {
+	for _, extLB := range ig.Spec.ExternalLoadBalancers {
 		if extLB.LoadBalancerName != nil {
 			lb := &awstasks.ClassicLoadBalancer{
 				Name:             extLB.LoadBalancerName,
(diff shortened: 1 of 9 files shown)


hakman:
This and other may need back-porting to 1.18 also now.
/cc @rifelpet @rdrgmnzs

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -400,15 +401,24 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
...
+			if err != nil {
+				return nil, fmt.Errorf("error parsing target grup ARN: %v", err)
+			}
+			resource := strings.Split(parsed.Resource, "/")
+			if len(resource) != 3 || resource[0] != "targetgroup" {
+				return nil, fmt.Errorf("error parsing target grup ARN resource: %q", parsed.Resource)
+			}
 			tg := &awstasks.TargetGroup{
-				Name:   fi.String(fmt.Sprintf("external-tg-%d", i)),
+				Name:   fi.String(resource[1]),
```
rifelpet, line 413: This assumes that the ARN field after `targetgroup/` is always equal to the target group's name and that AWS never modifies that in anyway. is that a safe assumption?

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -523,9 +533,11 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 		var attachLBRequest *autoscaling.AttachLoadBalancersInput
 		var detachLBRequest *autoscaling.DetachLoadBalancersInput
 		if changes.LoadBalancers != nil {
-			attachLBRequest = &autoscaling.AttachLoadBalancersInput{
-				AutoScalingGroupName: e.Name,
-				LoadBalancerNames:    e.AutoscalingLoadBalancers(),
+			if e != nil && len(e.TargetGroups) > 0 {
```
rifelpet, line 536: I believe you want this?
```suggestion
			if e != nil && len(e.LoadBalancers) > 0 {
```

rifelpet (commented):
is there anything in the integration tests that we could add that would expose this bug and confirm this fix?

pkg/model/awsmodel/autoscalinggroup.go
hakman: I think for now should be ok. AWS will break pretty much anything it wants :).

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
hakman: Thanks for catching this :). Fixed.

pkg/model/awsmodel/autoscalinggroup.go
hakman: For now, in case of TGs, the resource ID is always `tg-name/big-number`. This PR validates this format so people will know in advance if something is changed in the future.

Technically, tasks are compared by name and ARN. The name is also used for ordering. We could potentially add some logic that detects the external TGs and generates the name is a way that doesn't assume `tg-name/big-number`, but would complicate things and add assume some fixed names as belonging to kOps. Seems safer to just assume AWS will not change the ARN format for TGs. For sure AWS will not change the ARNs of existing TGs.

PR #10278, opened 2020-11-20
Recognize ubuntu 20.10

justinsb:
Teach nodeup about ubuntu 20.10, including the unusual
/etc/resolv.conf configuration.

### DIFF
--- a/nodeup/pkg/model/miscutils.go
+++ b/nodeup/pkg/model/miscutils.go
@@ -48,8 +48,15 @@ func (b *MiscUtilsBuilder) Build(c *fi.ModelBuilderContext) error {
 			packages = append(packages, "curl")
 			packages = append(packages, "wget")
 			packages = append(packages, "perl")
-			packages = append(packages, "python-apt")
 			packages = append(packages, "apt-transport-https")
+
+			// TODO: Do we really need python-apt?
+			switch b.Distribution {
+			case distributions.DistributionUbuntu2010:
+				// python-apt not available (though python3-apt is)
+			default:
+				packages = append(packages, "python-apt")
+			}
 		}
 	} else if b.Distribution.IsRHELFamily() {
 		// TODO: These packages have been auto-installed for a long time, and likely we don't need all of them any longer
--- a/util/pkg/distributions/identify_test.go
+++ b/util/pkg/distributions/identify_test.go
@@ -100,6 +100,11 @@ func TestFindDistribution(t *testing.T) {
 			expected: DistributionUbuntu2004,
 		},
 		{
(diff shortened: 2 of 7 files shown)


justinsb:
I'm actually not sure if we want to recognize / support non-LTS ubuntu versions.  It's pretty low-effort now that we're installing docker / containerd from tar.gz packages.

hakman:
> I'm actually not sure if we want to recognize / support non-LTS ubuntu versions. It's pretty low-effort now that we're installing docker / containerd from tar.gz packages.

As long as we agree that we will treat these non-LTS as LATEST and always replace it when a new one is released, should be fine.

util/pkg/distributions/identify_test.go
```diff
@@ -99,6 +99,11 @@ func TestFindDistribution(t *testing.T) {
 			err:      nil,
 			expected: DistributionUbuntu2004,
 		},
+		{
+			rootfs:   "ubuntu2010",
+			err:      nil,
+			expected: DistributionUbuntu2010,
+		},
```
hakman, line 106: The tests expect the `/etc/os-release` from an Ubuntu 20.10 to be copied to the `tests` dir.

nodeup/pkg/model/miscutils.go
```diff
@@ -48,8 +48,15 @@ func (b *MiscUtilsBuilder) Build(c *fi.ModelBuilderContext) error {
 			packages = append(packages, "curl")
 			packages = append(packages, "wget")
 			packages = append(packages, "perl")
-			packages = append(packages, "python-apt")
 			packages = append(packages, "apt-transport-https")
+
+			// TODO: Do we really need python-apt?
```
johngmyers, line 53: I doubt we need any of these, which is why I'm phasing them out for k8s 1.20+

justinsb:
> As long as we agree that we will treat these non-LTS as LATEST and always replace it when a new one is released, should be fine.

I'm going to put this on the agenda for today.  OTOH I agree that otherwise it could become a lot of work, OTOH we don't really want a kops upgrade to break people's working configurations.

What motivated this was that the latest ubuntu image on GCP is not LTS (the LTS is also available, it just wasn't the first that caught my eye).  So it was a little annoying to have the cluster fail to start there.

hakman:
Cool! :)
/lgtm

MoShitrit:
@hakman Does this mean we can start updating the ami with 20.10 versions on a more frequent basis from now on?

olemarkus:
@MoShitrit If you refer to the channels, they should remain with the LTS versions.

MoShitrit:
Thanks for clarifying @olemarkus !

Issue #10202, opened 2020-11-09
Reduce boilerplate in AWS LaunchTemplate


djablonski-moia:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

In Kops 1.19.0-alpha.5, we can't start a new cluster anymore because of the size limitation of user data in AWS launch templates, which is 16k. We have approximately 3k of our own user data, and for the master nodes, deployment fails with

```
error running task "LaunchTemplate/master-eu-central-1b.masters.cluster.<domain>" (26s remaining to succeed): error creating LaunchTemplate: InvalidUserData.Malformed: User data is limited to 16384 bytes.
```

Looking into the Launch templates for the non-master instance groups, I see that Kops own part already takes up 10k for them, and most likely even more than 13k for masters, since our own user data is the same for them.

I'd like to see the boilerplate in the user data being reduces as much as possible, leaving more space for custom user data. For example, just the license header in the embedded `nodeup.sh` is taking up 607 bytes.
Additionally, the documentation should warn about the limited space for custom user data, so users can mitigate before running into this issue.

**2. Feel free to provide a design supporting your feature request.**

E.g. remove the license header from the embedded `nodeup.sh` script, and also other comments, unccessary blank lines etc.

djablonski-moia:
After removing most of our custom user data, I can confirm that Kops already uses up 15k of the user data on master nodes! 😳

olemarkus:
The typical master user data should be around 12-13k. But since user data contains the cluster spec and the spec also contains additionalUserData, 3k will result in 6k added data.

3k of additional user data is quite sizeable though. Maybe some of this could be moved to a hook instead?

djablonski-moia:
@olemarkus Sorry, I was talking about just the userData in the cluster spec, there is no duplication of our userData I can see (after reducing it significantly). But our list of `fileAssets` is also adding 1.5k.
I'll try to reduce our userData and fileAssets as much as possible by shifting stuff into the AMI, but I still think the Kops part should also be minimized.

DOboznyi:
We also can't update kops in clusters from 1.18 to 1.19 because of file assets. Can it be decreased?

Nilubkal:
We are experiencing the same issues trying to create a new cluster via `kops-1.21-alpha-3`.

d-m:
I'm also running into this updating a cluster from 1.19 to 1.20

Nilubkal:
[...]

PR #10280, opened 2020-11-20
Build artifacts only for the used CPU architectures

hakman:
This PR reverts the previously added behaviour that was adding both AMD64 and ARM64 assets to each IG user data. The main reason for it is the increase in the size of the user data, which generated issues for users like #10202.

The CPU architecture of each IG can be changed using a new IG field named `Architecture`.
https://github.com/kubernetes/kops/blob/43b9d253ae09c757d4f0c0606385776bd114ccf6/pkg/apis/kops/instancegroup.go#L93-L94

It defaults to AMD64, but there is also a simple check for `-arm64` in the image name that should cover most cases for auto-detecting ARM64:
https://github.com/kubernetes/kops/blob/599fb7d993cf4355f8a5ef9aad5b9472b80641d9/upup/pkg/fi/cloudup/apply_cluster.go#L167-L174

/cc @justinsb @johngmyers @olemarkus @rdrgmnzs


olemarkus:
It feels like there are better ways to optimise the template files. E.g zipping + base64 encode the strings yield 25-50% fewer bytes depending on the length of the spec.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -163,6 +163,24 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
 		}
 	}
 
+	for _, ig := range c.InstanceGroups {
+		if ig.Spec.Architecture == "" {
+			// TODO: Find CPU arch by using the image/instance info
+			if strings.Contains(ig.Spec.Image, "-arm64") {
+				ig.Spec.Architecture = string(architectures.ArchitectureAmd64)
```
rdrgmnzs, line 170: Did you mean ArchitectureArm64 here?

rdrgmnzs (commented):
LGTM, besides the one question I have. Thanks @hakman!

IMO including the hashes of binaries that will never be used doesn't make that much sense given the user data size constraints.

rdrgmnzs:
One issue I just hit while testing out gziping the user-data is that facter (puppet) does not handle gzip user-data well https://tickets.puppetlabs.com/browse/FACT-1354, so it may cause issues for any users (like my self) that have puppet installed on their hosts.

olemarkus:
Yep. Hence the gzip+base64 of just those heredoc blocks.

rdrgmnzs:
Ah, sorry I misunderstood then.

hakman:
Closing in favor of #10357

PR #10281, opened 2020-11-20
[addons/CA] Add support for specifying resources and metrics

dntosas:
In this PR, we make some additions on Cluster Autoscaler Addon Specs

#### Resources
We enable users to set their desired capacity for cluster-autoscaler addon.

There are edge cases, especially in big clusters, where autoscaler needs
to reconcile a large number of objects thus may need increased memory to
avoid OOMkills or increased cpu to avoid saturation.

#### Metrics
Cluster autoscaler provides valuable insights for monitoring capacity
allocation and scheduling aspects of a cluster. In this commit, we
enable users to add proper annotation on deployment to scrape metrics
via Prometheus.

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -710,6 +710,23 @@ spec:
 
 For more details on `horizontalPodAutoscaler` flags see the [official HPA docs](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) and the [kOps guides on how to set it up](horizontal_pod_autoscaling.md).
 
+## Cluster autoscaler
+{{ kops_feature_table(kops_added_default='1.19', k8s_min='1.15') }}
+
+Cluster autoscaler can be enabled to automatically adjust the size of the kubernetes cluster.
+
+```yaml
+spec:
+  clusterAutoscaler:
+    enabled: true
+    skipNodesWithLocalStorage: true
+    skipNodesWithSystemPods: true
+    cpuRequest: "100m"
+    memoryRequest: "300Mi"
+```
+
+Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
+
 ###  Feature Gates
 
 Feature gates can be configured on the kubelet.
--- a/pkg/model/components/clusterautoscaler.go
+++ b/pkg/model/components/clusterautoscaler.go
@@ -44,13 +44,13 @@ func (b *ClusterAutoscalerOptionsBuilder) BuildOptions(o interface{}) error {
 		if err == nil {
 			switch v.Minor {
 			case 19:
-				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.19.0"
+				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.19.1"
 			case 18:
-				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.18.2"
+				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.18.3"
 			case 17:
-				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.17.3"
+				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.17.4"
 			case 16:
-				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.16.6"
+				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.16.7"
 			case 15:
 				image = "k8s.gcr.io/autoscaling/cluster-autoscaler:v1.15.7"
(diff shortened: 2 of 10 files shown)


upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -134,6 +134,12 @@ spec:
     metadata:
       labels:
         app: cluster-autoscaler
+      {{ if .Metrics }}
```
olemarkus, line 137: These annotations are not much in use anymore and we tend not to set them. Instead we expect users to create pod/service monotors themselves.

olemarkus:
I am not sure we should set any limits. It is rather dangerous to have CAS crashing due to oom. Maybe it is better to remove limits alltoghether.

dntosas:
> I am not sure we should set any limits. It is rather dangerous to have CAS crashing due to oom. Maybe it is better to remove limits alltoghether.

Will do that ^

Two more thoughts to consider:
- Do you think is worth make `replicas` editable?
- Should we also bump container image patch versions?
- If yes, could these two included on this PR or we need separate ones?

Sorry for the many questions :yum:

olemarkus:
Bumping patch versions is good. I didn't do that as I suspect more patch versions will be released before kops 1.20. but feel free to add to this PR.

I don't think making replicas configurable is that useful. But if someone makes a good case for it, I will certainly consider it.

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
dntosas: There are cases like [coreDNS](https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template#L231-L233) where these annotations seem useful. Do you suggest completely drop from here or migrating to a ServiceMonitor? I am cool with both options ^^

olemarkus: Note that since Metrics is a pointer you need to use `WithDefaultBool`.

olemarkus:
@dntosas do you have time to look at wrapping this one up?

dntosas:
> @dntosas do you have time to look at wrapping this one up?

sure ^^ i let it stale because i didn't understand if we wanted to close this PR as not needed

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -134,6 +134,12 @@ spec:
     metadata:
       labels:
         app: cluster-autoscaler
+      {{ if WithDefaultBool .Metrics false }}
```
olemarkus, line 137: If the only thing this flag does is add the annotation, it is not that useful. You could just add the annotations without the flag. Typically such a flag would be used for enabling/disabling CAS serving metrics in the first place.

pkg/model/components/clusterautoscaler.go
```diff
@@ -83,6 +83,9 @@ func (b *ClusterAutoscalerOptionsBuilder) BuildOptions(o interface{}) error {
 	if cas.NewPodScaleUpDelay == nil {
 		cas.NewPodScaleUpDelay = fi.String("0s")
 	}
+	if cas.Metrics == nil {
```
olemarkus, line 86: You already default this one to false in the manifest. So you do not need this bit

olemarkus (changes requested):
Sorry for the confusion. My intention was to only provide feedback on improving this.

pkg/model/components/clusterautoscaler.go
dntosas: removed it ^^

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
dntosas: removed flag also ^^

docs/cluster_spec.md
```diff
@@ -710,6 +710,24 @@ spec:
...
+
+```yaml
+spec:
+  clusterAutoscaler:
+    enabled: true
+    skipNodesWithLocalStorage: true
+    skipNodesWithSystemPods: true
+    cpuRequest: "100m"
+    memoryRequest: "300Mi"
+    metrics: true
```
olemarkus, line 726: If you remove this one, we should be good to go

olemarkus:
Cool. Thanks
/lgtm
/approve

PR #10283, opened 2020-11-20
Add additional policies support at instance group level

eddycharly:
This PR adds support for additional inline and external IAM policies at the instance group level.

It does so by creating an instance profile per instance group and one per node role (Master, Node and Bastion).
Then, when linking launch configuration/template, the instance profile is chosen for a given instance group depending on the presence of specific policies. If there's no specific policy specified, it uses the per role instance profile, else it picks up the instance group instance profile.

Two new fields, `AdditionalPolicy` and `ExternalPolicies` were added to the `InstanceGroupSpec` struct.
These two new fields are used to setup the instance groups / roles in `pkg/model/iam.go` (combined with the policies declared at the cluster level).

I also added integration testing support for additional policies, both at the cluster and instance group level.

**NOTE**: This PR changes a lot of files but it is mostly files used by integration tests, code changes are pretty small.

I need to test it on a real cluster, and check that migrating an existing cluster works as expected (it should trigger a rolling update).

I will also update the docs once i'm sure the PR is going to be accepted.

Any feedback welcome, i can provide infos on the use case that led me to propose this change if needed. I discussed it with @olemarkus on Slack here https://app.slack.com/client/T09NY5SBT/C3QUFP0QM/thread/C3QUFP0QM-1605543979.445000


eddycharly:
That's a lot of integration tests to fix, is the terraform/cloudformation code in the `tests/integration` forlder authored by hand ?

eddycharly:
Ah, `update-expected.sh` is my friend.

eddycharly:
Tests are passing ! 🎉

eddycharly:
This should be ready for review 🙏 
I will test it for real tomorrow and will let you know 🤞

olemarkus:
Overall this looks fairly okay to me. But given the larger change here, I suspect the approvers will want to discuss this in office hours.

eddycharly:
@olemarkus Thanks, i will join the next meeting. From what i see in the docs, it occurs every Friday ?
I can be joined on slack too until then.

It still needs some testing and i guess i should delete the dedicated profile/role when `kops delete instancegroup` is invoked too ?

olemarkus:
You will also need to add some docs, and some of the struct comments could be more clear (expects names vs arns, for example). But it may be worth polishing this after discussing it in the office hours. 

The docs are up to date wrt the office hours :)

pkg/model/names.go
```diff
@@ -125,14 +125,14 @@ func (b *KopsModelContext) NameForDNSZone() string {
...
 	switch role {
 	case kops.InstanceGroupRoleMaster:
-		return "masters." + b.ClusterName()
+		return igName + "." + b.ClusterName()
 	case kops.InstanceGroupRoleBastion:
-		return "bastions." + b.ClusterName()
+		return igName + "." + b.ClusterName()
 	case kops.InstanceGroupRoleNode:
-		return "nodes." + b.ClusterName()
+		return igName + "." + b.ClusterName()
```
rdrgmnzs, line 135: Can we test how existing clusters will react to this name change? 

I'm specially worried about Terraform, if I remember correctly the name change would cause the old resources to be garbage collected by Terraform. However this would most likely fail as there's still hosts using the old IAM roles, pretty much causing Terraform based clusters to be stuck in a place where they can't be updated.

eddycharly: I will try to test it tomorrow, i didn't have the time do it today.

It's true that it could cause troubles  with Terraform based clusters. Luckily with Terraform we can always manipulate the state before upgrading.

Thanks for raising the issue, i will bring it when discussing this in the office hours.

rifelpet: I think the `direct` target would also orphan the old IAM roles with existing clusters - the new roles will be created but the old roles will never be deleted, which might be preferred (I don't recall the behavior of deleting a role still attached to an instance)

Either way, changing the role name means any "downstream" IAM roles assumed by these roles might need their assume role policies updated.

I'm definitely in favor of supporting IG-level IAM roles but we'll have to think through the consequences of a change like this. Perhaps we can make it opt-in via an InstanceGroupSpec field? Otherwise fallback to the old roles and role names.

olemarkus: Opt-in for existing clusters and opt-out for new clusters makes sense to me. 

We should also tag our roles so that kops can better reconcile and delete them later.

eddycharly: Can we keep the per role roles ? It would allow for a smooth transition for Terraform based clusters until the cluster is rolled out.

> Either way, changing the role name means any "downstream" IAM roles assumed by these roles might need their assume role policies updated.

True, this definitely needs to be well documented.

> Perhaps we can make it opt-in via an InstanceGroupSpec field? Otherwise fallback to the old roles and role names.

If we keep the old roles and role names, we could use the old roles when no additional policy is specified, thus making the `AdditionalPolicy` and `ExternalPolicies` act as an opt-in flag ?
I mean, if a user starts using `AdditionalPolicy` and/or `ExternalPolicies` in an instance group, he probably expects that the role will be different.

eddycharly: I changed the implementation, it now creates both per role instance profiles AND per instance group instance profiles.

Then, when linking the launch configuration/launch template with the instance profile, if there's no policy specified for the given instance group it uses the profile associated with the instance group role. If the instance group has additional policy specified it chooses the instance group profile.

This should solve both the terraform based clusters issue and the issue with "downstream" IAM roles assuming.

Except that it will create additional resources (one instance profile/role per instance group), as long as one doesn't specify instance group policy, it should behave exactly the same as before.

cc @rifelpet @rdrgmnzs

eddycharly:
Shall i add this to the today agenda myself or it will be added during the meeting ?

rifelpet:
@eddycharly feel free to add it yourself

eddycharly:
@rifelpet yeah, thanks. i didn't get that it needed to be accepted, it's all good now.

eddycharly:
Would this make sense to put this behind a feature flag ?

PR #10284, opened 2020-11-21
Set --service-account-issuer for k8s 1.20+

### LINKED ISSUE #10279: Clusters created with k8s v1.20 don't start
e2e tests for k8s v1.20 fail to start:
https://testgrid.k8s.io/kops-versions#kops-aws-k8s-latest

Me and @rifelpet tracked this to https://github.com/kubernetes/kubernetes/pull/95896, which introduces some new mandatory flags:
```
TokenRequest and TokenRequestProjection are now GA features. The following flags are required by the API server:
- `--service-account-issuer`, should be set to a URL identifying the API server that will be stable over the cluster lifetime.
- `--service-account-key-file`, set to one or more files containing one or more public keys used to verify tokens.
- `--service-account-signing-key-file`, set to a file containing a private key to use to sign service account tokens. Can be the same file given to `kube-controller-manager` with `--service-account-private-key-file`.
```

johngmyers:
Kubernetes 1.20+ requires the `--service-account-issuer` flag be passed to kube-apiserver

Fixes #10279 

/cc @justinsb

### DIFF
--- a/pkg/model/iam/subject.go
+++ b/pkg/model/iam/subject.go
@@ -81,8 +81,8 @@ func BuildNodeRoleSubject(igRole kops.InstanceGroupRole) (Subject, error) {
 }
 
 // ServiceAccountIssuer determines the issuer in the ServiceAccount JWTs
-func ServiceAccountIssuer(clusterName string, clusterSpec *kops.ClusterSpec) (string, error) {
-	return "https://api." + clusterName, nil
+func ServiceAccountIssuer(clusterName string, clusterSpec *kops.ClusterSpec) string {
+	return "https://api." + clusterName
 }
 
 // AddServiceAccountRole adds the appropriate mounts / env vars to enable a pod to use a service-account role
--- a/pkg/model/iam/subject.go
+++ b/pkg/model/iam/subject.go
@@ -22,7 +22,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kops/pkg/apis/kops"
-	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/pkg/wellknownusers"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/model/iam/subject.go
```diff
@@ -83,11 +82,7 @@ func BuildNodeRoleSubject(igRole kops.InstanceGroupRole) (Subject, error) {
 
 // ServiceAccountIssuer determines the issuer in the ServiceAccount JWTs
 func ServiceAccountIssuer(clusterName string, clusterSpec *kops.ClusterSpec) (string, error) {
-	if featureflag.PublicJWKS.Enabled() {
-		return "https://api." + clusterName, nil
-	}
-
-	return "", fmt.Errorf("ServiceAcccountIssuer not (currently) supported without PublicJWKS")
+	return "https://api." + clusterName, nil
```
johngmyers, line 85: Somewhat unrelated to what this PR is trying to fix, if the admin has set `spec.kubeAPIServer.serviceAccountIssuer` explicitly, this should probably return that value instead.

johngmyers:
This is at least sufficient to get k8s 1.20 to start, but I'm not entirely confident about setting that issuer without enabling discovery. I would appreciate review from @justinsb or @rifelpet.
/approve cancel

olemarkus:
Service account issuer discovery was promoted to beta in 1.20 so this feature should be enabled by default. I don't think we need to do anything in particular around that feature.

pkg/model/iam/subject.go
rifelpet: I agree we should respect the API field but that could be done in a separate PR.

This function no longer needs to return an error

rifelpet (commented):
this looks good to me but I'd like Justin's opinion on potential impact

/cc @justinsb

justinsb:
This looks good.  One thing we'll have to verify is that upgrading works!

/approve
/lgtm

PR #9769, opened 2020-08-17
WIP: Remove modelStore - no longer needed

justinsb:
We've replaced templates with code.

PR #10297, opened 2020-11-22
Remove dead code

hakman:
Follow-up of #9769 and #9841.

/cc @johngmyers


johngmyers:
Had incorrectly thought that was used by bootstrap addons.
/lgtm

PR #10308, opened 2020-11-24
Move tools into separate `hack` go module

rifelpet:
This follows a pattern used in [other projects](https://github.com/kubernetes-sigs/kubetest2/tree/master/hack/tools) moving imports used only in development tools into a separate go module. The new module does not use vendoring. all scripts and make targets that use the tools will now build them into `_output/bin/`.

I updated the goimports python scripts to rely less on GOPATH but the `goimports` tool itself still relies on it.

EDIT: i opted to keep go-bindata in the root module because bazel depends on it and I haven't yet figured out how we could best handle that without vendoring go-bindata somewhere:

https://github.com/kubernetes/kops/blob/c62181968f525cfeace516d5891059fd28e848d1/upup/models/BUILD.bazel#L23-L30

also [update-machinery.sh](https://github.com/kubernetes/kops/blob/master/hack/update-apimachinery.sh) still relies on GOPATH, we can update this in a followup PR


rifelpet:
I'll need to figure out how to reference the new go-bindata location within this bazel rule

https://github.com/kubernetes/kops/blob/e789c24c3a712d7b7af454867151a5898de2b8a9/upup/models/BUILD.bazel#L23-L33

rifelpet:
this is ready for review. as mentioned in the original comment I had to back out some of the moves, we can address them in followup PRs.

olemarkus:
This looks good! Works as intended locally.

/lgtm

PR #10312, opened 2020-11-24
[Digital Ocean] Promote to Beta

srikiz:
Hi,
I wanted thoughts on promoting Digital Ocean to Beta. Currently, a good amount of KOPS functionality is in place for DO, and I wanted to check with the KOPS maintainers if you are okay with this promotion.

Tested with both single cluster and multi-master HA setup
Tested kubernetes upgrade with kops update feature (ex: from v1.19.3 to v1.19.4)
Tested kops validate command on cluster creation
Tested creating multiple instance groups and managing & updating & deleting it.
Tested using the cluster with a DO load balancer when a new service with type: LoadBalancer is used.
Tested creating new PV with CSI drivers for managing DO volumes.

FYI - @timoreimann

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -72,7 +72,7 @@ const (
 )
 
 var (
-	// AlphaAllowDO is a feature flag that gates DigitalOcean support while it is alpha
+	// AlphaAllowDO is a feature flag that gates cloudprovider support while it is alpha
 	AlphaAllowDO = featureflag.New("AlphaAllowDO", featureflag.Bool(false))
 	// AlphaAllowGCE is a feature flag that gates GCE support while it is alpha
 	AlphaAllowGCE = featureflag.New("AlphaAllowGCE", featureflag.Bool(false))
@@ -392,10 +392,6 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
 
 	case kops.CloudProviderDO:
 		{
-			if !AlphaAllowDO.Enabled() {
-				return fmt.Errorf("DigitalOcean support is currently (very) alpha and is feature-gated. export KOPS_FEATURE_FLAGS=AlphaAllowDO to enable it")
-			}
-
 			if len(sshPublicKeys) == 0 && (c.Cluster.Spec.SSHKeyName == nil || *c.Cluster.Spec.SSHKeyName == "") {
 				return fmt.Errorf("SSH public key must be specified when running with DigitalOcean (create with `kops create secret --name %s sshpublickey admin -i ~/.ssh/id_rsa.pub`)", cluster.ObjectMeta.Name)
 			}
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -72,8 +72,6 @@ const (
 )
 
 var (
-	// AlphaAllowDO is a feature flag that gates cloudprovider support while it is alpha
-	AlphaAllowDO = featureflag.New("AlphaAllowDO", featureflag.Bool(false))
 	// AlphaAllowGCE is a feature flag that gates GCE support while it is alpha
 	AlphaAllowGCE = featureflag.New("AlphaAllowGCE", featureflag.Bool(false))
 	// AlphaAllowALI is a feature flag that gates aliyun support while it is alpha
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -1,4 +1,5 @@
 # cloud.google.com/go v0.54.0
+## explicit
 cloud.google.com/go/compute/metadata
 # github.com/Azure/azure-pipeline-go v0.2.3
 ## explicit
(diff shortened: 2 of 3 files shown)


olemarkus:
I guess a concern from my side is who will own this going forward. I think at least one person who contribute outside of only this provider should commit to also maintain this. Beta is fairly close to stable in our world.

@srikiz the stuff you have been doing here is pretty solid. I don't know if you are interested in contributing to kops as a whole?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -72,7 +72,7 @@ const (
 )
 
 var (
-	// AlphaAllowDO is a feature flag that gates DigitalOcean support while it is alpha
+	// AlphaAllowDO is a feature flag that gates cloudprovider support while it is alpha
 	AlphaAllowDO = featureflag.New("AlphaAllowDO", featureflag.Bool(false))
```
rifelpet, line 76: We can just remove this feature flag altogether right? Below was the only place its used so enabling DO support by default should involve removing this FF.

rifelpet (commented):
I would definitely like to see DO promoted to beta. I think its worth clarifying exactly what the criteria is for beta and stable cloud providers. Perhaps we discuss it at the next office hours?

I think it would be great to have automated e2e tests for DO. I believe [cluster-api](https://github.com/kubernetes-sigs/cluster-api-provider-digitalocean) has [conformance tests](https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-digitalocean#periodic-conformance-v1alpha3) and presubmits running, so I would think much of the setup work is already taken care of if we can get permission to share the DO account with Cluster API. I'm happy to assist with setting up the prow jobs in test-infra once we're ready. I think just periodic tests (similar to GCP's) would be fine.

upup/pkg/fi/cloudup/apply_cluster.go
srikiz: I know there was another PR on Azure support to DO, so was not sure if it would get into the Alpha state. I can remove this flag if you think we should.

srikiz:
> I guess a concern from my side is who will own this going forward. I think at least one person who contribute outside of only this provider should commit to also maintain this. Beta is fairly close to stable in our world.
> 
> @srikiz the stuff you have been doing here is pretty solid. I don't know if you are interested in contributing to kops as a whole?

Thanks @olemarkus - I would like to contribute to kops as a whole, but currently can't commit to it. May be we can discuss more during our next KOPS meeting.

timoreimann:
👍  Timo here from DigitalOcean. I've been supporting @srikiz over the last few months as far as support for DigitalOcean in kops is concerned.

We raised the topic on what it takes for a provider to be promoted to beta in a prior office hours meeting on 13. April 2020. Unfortunately, not too many points on that discussion were apparently noted down in the [meeting notes](https://docs.google.com/document/d/12QkyL0FkNbWPcLFxxRGSPt_tNPBHbmni3YLY-lHny7E/edit#heading=h.b4uuvuwdqfkr). However, I did take notes myself back then and wrote down the following requirements as discussed during the office hours:

- Cluster does not need to be destroyed anymore for rolling upgrade
- Implement validation for DO executed during kops create / update
- HA support

To my understanding, these TODOs are now completed. (@srikiz, correct me if I'm wrong here.)

I _think_ we explicitly excluded test integration at the time. Not exactly sure about the reasons anymore, perhaps the level of support for / ease of adding tests was different back then?
That said, if test integration is expected today from all supported clouds, I'd be happy to talk about it. Let me know what you think, thanks!

olemarkus:
From my perspective, testing is super nice, but it is not a hard requirement. Timely fixes, keeping up with the development of kops overall etc is more important. 

My fear is that DO goes beta, a lot of people starts using it, and then something breaks. Or that we need to update some deps, this breaks the provider, and there is no one around to fix it.

Now @srikiz almost immediately fixed a case of the latter issue earlier today, so I have full confidence that this is not a problem today. I just, perhaps, need a little bit more confidence that this won't be a problem next year.

If DO softly commits to support kops, even support progress DO to stable, then I will happily support DO going to beta in its current state.

rifelpet:
I tracked down a recording of the office hours in which we discussed DO being promoted to beta: https://youtu.be/A6DXuR4Nu30?t=2584

Justin's points for reaching beta status:
* API Stability
* Going to GA won't require a cluster to be torn down and recreated

@srikiz mentioned an issue around using an internal load balancer for kubelets to reach the apiserver, possibly the `useForInternalApi` ClusterSpec field. Has that been addressed?

I agree that e2e testing needn't be a hard requirement for beta. I think if we fix the conflicts here and address any review comments then we can bring it up on Friday's office hours call for any final objections and plan to merge it then.

upup/pkg/fi/cloudup/apply_cluster.go
rifelpet: I think if we're removing any use of the flag, effectively making its default behavior to being enabled, then we should just remove the flag altogether given that explicitly disabling the use of a cloud provider feature flag doesn't make sense.

srikiz:
@rifelpet - I didn't use the ```useForInternalAPI``` behavior. As I understand, currently KOPS injects /etc/hosts file entries of all the master hosts and kubelet uses one of those host entries for communicating to api server. DO uses the same behavior in this case.

rifelpet:
@srikiz ok that makes sense. Feel free to fix the conflicts here and we can discuss and hopefully merge this on Friday.

upup/pkg/fi/cloudup/apply_cluster.go
srikiz: Removed.

vendor/modules.txt
```diff
@@ -1408,4 +1408,4 @@ sigs.k8s.io/yaml
 # k8s.io/cli-runtime => k8s.io/cli-runtime v0.20.0-beta.2
 # k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.20.0-beta.2
 # k8s.io/code-generator => k8s.io/code-generator v0.20.0-beta.2
-# github.com/gophercloud/gophercloud => github.com/gophercloud/gophercloud v0.11.0
```
rifelpet, line 1411: i think this change should be reverted in order to make the verify-gomod job pass

rifelpet (changes requested):

rifelpet:
hm not sure why that job failure is still in the GH comment if the job itself passed...

/test pull-kops-verify-packages

rifelpet:
Everyone on the call today was happy to see DO get promoted to Beta so I'm going to go ahead and merge this 👍🏻 congrats!

/lgtm
/approve

timoreimann:
@rifelpet thanks a lot for your help on driving this one over the finishing line, really appreciate it. Excited to hopefully see more adoption of kops in DigitalOcean over time now that stability / maturity has improved.

timoreimann:
I have also filed #10386 to track the work needed to get proper e2e testing in place. Sri and I are happy to move that one forward as well.

PR #10318, opened 2020-11-26
Add minimal cert-manager addon

olemarkus:
There are several images involved here, so dropped the usual `Image` field until we have a better understanding how to handle this. Making at minimum the version configurable can be done well in time for 1.20 alpha anyway.

Adding this PR is a prereq for some other addons + the kops-controller issuer.


hakman:
I am a bit concerned about the 1.6 MB template that is being pulled into bindata.go also.

olemarkus:
Pretty huge CRDs. But can we do much about that?

hakman:
Not sure what can be done about it, but probably something should be done.

olemarkus:
https://github.com/jetstack/cert-manager/issues/3299

They are what they are. I can't think of a reason this would cause any harm.

upup/pkg/fi/cloudup/bootstrapchannelbuilder.go
```diff
@@ -488,6 +488,26 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*chann
 		}
 	}
 
+	if b.Cluster.Spec.CertManager != nil && fi.BoolValue(b.Cluster.Spec.CertManager.Enabled) {
+		{
+			key := "certmanager.io"
+			version := "1.0.4"
```
hakman, line 494:
```suggestion
			version := "1.1.0"
```

hakman (changes requested):
A few Q:
1. Why switch from the `cert-manager` namespace to `kube-system`?
2. Why limit running on master nodes?

olemarkus:
>     Why switch from the cert-manager namespace to kube-system?

Kops could technically have its own namespace for addons that doesn't strictly need to be in kube-system. But we don't have anything like this. So when it comes to PSPs and such, we don't have much choice but to deploy it to kube-system.

> Why limit running on master nodes?

A lot of users run cert-manager on masters to leverage DNS ACME authentication. If/when we get IRSA more mature (and if this doesn't end up in a circular dependency) it could run on normal nodes using a dedicated IAM role.

hakman:
Thanks for the explanation.
/lgtm
/approve

PR #10321, opened 2020-11-26
Adds controller to that signs CRs with kops CA

olemarkus:
* watches only cert-manager CRs in kube-system.
* doesn't do any particular validation of the CR, but assumes that RBAC prevents someone from creating unwanted certs
* as mentioned in office hours one could fortify things a bit by only allowing known CR names to be signed based on what addons has been added.

Still a question if we want to do solution or if we want to just provision a new CA issuer with cert-manager.


olemarkus:
/close

Superseded by #10545

PR #10328, opened 2020-11-28
Don't try to detach masters

olemarkus:
detatchInstance used to return if executed against master IG. This is no longer the case, so kops tries to detach master nodes when deleting instance.

/kind bug


hakman:
Thanks @olemarkus. Any cherry-picks needed?
/lgtm
/approve

PR #10335, opened 2020-12-01
Allow attaching same external target group to multiple instance groups

hakman:
/cc @rifelpet @rdrgmnzs


upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -172,15 +171,15 @@ func (e *AutoscalingGroup) Find(c *fi.Context) (*AutoscalingGroup, error) {
 			if len(resource) != 3 || resource[0] != "targetgroup" {
 				return nil, fmt.Errorf("error parsing target grup ARN resource: %q", parsed.Resource)
 			}
-			actualTGs = append(actualTGs, &TargetGroup{ARN: aws.String(*tg), Name: aws.String(resource[1])})
-		}
-		targetGroups, err := ReconcileTargetGroups(c.Cloud.(awsup.AWSCloud), actualTGs, e.TargetGroups)
-		if err != nil {
-			return nil, err
+			clusterName := fi.StringValue(g.AutoScalingGroupName)
+			if resource[1] != awsup.GetResourceName32(clusterName, "tcp") && resource[1] != awsup.GetResourceName32(clusterName, "tls") {
```
rifelpet, line 175: I think if we're using this new GetResourceName32 function in the Find task then we should also use it in pkg/model/awsmodel

upup/pkg/fi/cloudup/awsup/aws_utils.go
```diff
@@ -197,3 +200,28 @@ func EC2TagSpecification(resourceType string, tags map[string]string) []*ec2.Tag
 
 	return []*ec2.TagSpecification{specification}
 }
+
+// GetResourceName32 will attempt to calculate a meaningful name for a resource given a prefix
+// Will never return a string longer than 32 chars
+func GetResourceName32(cluster string, prefix string) string {
```
rifelpet, line 206: this isn't necessarily specific to AWS so i'm wondering if it would better belong in `pkg/util` or `util/pkg` (why do we have both??)

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
hakman: Updated `GetELBName32` to use the new function.

upup/pkg/fi/cloudup/awsup/aws_utils.go
hakman: For now the 32 chars name is limited to AWS and the other packages don't seem to be suitable for this. I think it's good enough for now.

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -172,15 +171,15 @@ func (e *AutoscalingGroup) Find(c *fi.Context) (*AutoscalingGroup, error) {
 			if len(resource) != 3 || resource[0] != "targetgroup" {
 				return nil, fmt.Errorf("error parsing target grup ARN resource: %q", parsed.Resource)
 			}
-			actualTGs = append(actualTGs, &TargetGroup{ARN: aws.String(*tg), Name: aws.String(resource[1])})
-		}
-		targetGroups, err := ReconcileTargetGroups(c.Cloud.(awsup.AWSCloud), actualTGs, e.TargetGroups)
-		if err != nil {
-			return nil, err
+			clusterName := fi.StringValue(g.AutoScalingGroupName)
```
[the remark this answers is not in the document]
hakman: Indeed, fixed.

olemarkus (approved):
This looks good to me

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -98,7 +98,7 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 	var clb *awstasks.ClassicLoadBalancer
 	var nlb *awstasks.NetworkLoadBalancer
 	{
-		loadBalancerName := b.GetELBName32("api")
+		loadBalancerName := b.CLBName32("api")
```
rifelpet, line 101: we use `loadBalancerName` in both the CLB and NLB tasks, so maybe the `CLBName32` function name could be updated to reflect that?

hakman: Done. Renamed to `LBName32`.

Issue #9982, opened 2020-09-24
panic: Terraform resource names cannot start with a digit. This is a bug in Kops, please report this in a GitHub Issue. Name: 1.etcd-events.k8s-cs.domain.net


dimitrez:
Version 1.18.1 (git-453d7d96be)

AWS

```
kops create cluster \
--cloud=aws \
--zones=us-west-2a \
--name=k8s-cs.domain.net \
--state=s3://domaininfrastructurestate/kops/cs/ \
--dns-zone=k8s-cs.domain.net \
--node-size=t3a.small \
--node-tenancy=default \
--node-volume-size=50 \
--node-count=3 \
--master-size=t3a.small \
--master-volume-size=50 \
--master-zones=us-west-2a \
--master-count=3 \
--out=. \
--target=terraform \
--yes \
--image=ami-028e52edfeb33adb2
```

```
W0924 16:33:48.893149   30386 create_cluster.go:771] Running with masters in the same AZs; redundancy will be reduced
I0924 16:33:50.377267   30386 subnets.go:184] Assigned CIDR 172.20.32.0/19 to subnet us-west-2a
I0924 16:33:54.964514   30386 create_cluster.go:1537] Using SSH public key: /home/demi4/.ssh/id_rsa.pub
I0924 16:34:08.385316   30386 executor.go:103] Tasks: 0 done / 95 total; 47 can run
I0924 16:34:08.397555   30386 dnszone.go:242] Check for existing route53 zone to re-use with name "k8s-cs.domain.net"
I0924 16:34:09.249425   30386 dnszone.go:249] Existing zone "k8s-cs.domain.net." found; will configure TF to reuse
I0924 16:34:10.792010   30386 vfs_castore.go:590] Issuing new certificate: "apiserver-aggregator-ca"
I0924 16:34:10.831807   30386 vfs_castore.go:590] Issuing new certificate: "etcd-clients-ca"
I0924 16:34:10.900833   30386 vfs_castore.go:590] Issuing new certificate: "ca"
I0924 16:34:10.919639   30386 vfs_castore.go:590] Issuing new certificate: "etcd-manager-ca-main"
I0924 16:34:11.035820   30386 vfs_castore.go:590] Issuing new certificate: "etcd-peers-ca-events"
I0924 16:34:11.106013   30386 vfs_castore.go:590] Issuing new certificate: "etcd-peers-ca-main"
[...]

PR #10361, opened 2020-12-03
Prefix etcd cluster names with letters

hakman:
Refs: #9982

/cc @rifelpet

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -692,6 +692,13 @@ func trimCommonPrefix(names []string) []string {
 		}
 	}
 
+	for i := range names {
+		_, err := strconv.Atoi(names[i])
+		if err == nil {
+			names[i] = "etcd-" + names[i]
+		}
+	}
+
 	return names
 }
 
--- a/tests/integration/create_cluster/ha_shared_zone/expected-v1alpha2.yaml
+++ b/tests/integration/create_cluster/ha_shared_zone/expected-v1alpha2.yaml
@@ -16,21 +16,21 @@ spec:
   - cpuRequest: 200m
     etcdMembers:
     - instanceGroup: master-us-test-1a-1
-      name: "1"
+      name: etcd-1
     - instanceGroup: master-us-test-1a-2
-      name: "2"
+      name: etcd-2
     - instanceGroup: master-us-test-1a-3
-      name: "3"
+      name: etcd-3
     memoryRequest: 100Mi
     name: main
   - cpuRequest: 100m
     etcdMembers:
     - instanceGroup: master-us-test-1a-1
-      name: "1"
+      name: etcd-1
     - instanceGroup: master-us-test-1a-2
-      name: "2"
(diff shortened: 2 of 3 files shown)


upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -692,6 +692,13 @@ func trimCommonPrefix(names []string) []string {
 		}
 	}
 
+	for i := range names {
+		_, err := strconv.Atoi(names[i])
+		if err == nil {
+			names[i] = "e" + names[i]
+		}
```
rifelpet, line 699: why `e`? I think it might be confusing to someone glancing at the cluster spec that kops just created for them.

```yaml
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-us-west-2a-1
      name: e1
    - instanceGroup: master-us-west-2a-2
      name: e2
    - instanceGroup: master-us-west-2a-3
      name: e3
    memoryRequest: 100Mi
    name: main
```
How about `etcd-` or `etcd` ?

```yaml
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-us-west-2a-1
      name: etcd-1
    - instanceGroup: master-us-west-2a-2
      name: etcd-2
    - instanceGroup: master-us-west-2a-3
      name: etcd-3
    memoryRequest: 100Mi
    name: main
```

hakman: I didn't have better inspiration at that time :).

rifelpet:
this looks good to me though it doesn't fix existing clusters being upgraded to Terraform 0.12.

hakman:
Not sure if we can have more than a release note for those running Terraform 0.12. At least with this new clusters could be created.

rifelpet:
what would the release note say? It seems like we should either use a different terraform resource name if the original name starts with a number and have the release notes instruct users to `terraform state mv`, or walk users through changing the etcd cluster member names which isn't clearly understood or documented, and likely involves control plane downtime.

Users will experience this issue when upgrading existing clusters, so we should be prepared to handle that.

hakman:
I think this is a start. If someone bumps into this issue with an existing cluster, will be a pain, but I don't see any easy solution. It's not just about renaming resources. If the community will find some way to fix things manually and documents the steps, I would be happy to look at that.

rifelpet:
i think the easiest solution is to have kops change the terraform resource name that it generates when it would otherwise start with a number in this situation. Then we know that anyone that had their resource name change would have otherwise ran into this issue and they'll need to `terraform state mv` the resource to its new name. The fix only impacts users that are directly affected by the issue, and the fix is not disruptive to the cluster itself.

hakman:
That would require more work and testing.

justinsb:
I find it hard to reason about whether this is safe.  Can we first create a test case that hits the numbers problem (in the create_cluster case), and then rebase this on that so we can see the change?

hakman:
@justinsb I added the test as part of https://github.com/kubernetes/kops/pull/10365.

Issue #10327, opened 2020-11-28
CentOS 8: as base dist


sergey-safarov:
could you add CenOS8 support for AWS installation?
Base AMI now [available](https://aws.amazon.com/marketplace/pp/B08KYLN2CG?qid=1606550294161&sr=0-6&ref_=srh_res_product_title#pdp-pricing).

sergey-safarov:
Also available images from CentOS
You can check the output of the command:
```
aws ec2 describe-images \
--region us-east-1 \
--output table \
--owners 679593333241 \
--query "sort_by(Images, &CreationDate)[*].[CreationDate,Name,ImageId]" \
--filters "Name=name,Values=CentOS-8.*x86_64*"
```

hakman:
Very good news. Thanks @sergey-safarov!
/assign

PR #10368, opened 2020-12-05
Update docs for CentOS 8

hakman:
Fixes: #10327

### DIFF
--- a/docs/operations/images.md
+++ b/docs/operations/images.md
@@ -74,21 +74,29 @@ CentOS 7 is based on Kernel version **3.10** which has a considerable number of 
 * [kubernetes/kubernetes#56903](https://github.com/kubernetes/kubernetes/issues/56903)
 * [kubernetes/kubernetes#67577](https://github.com/kubernetes/kubernetes/issues/67577)
 
-Before using CentOS images you must accept the agreement at https://aws.amazon.com/marketplace/pp?sku=aw0evgkw8e5c1q413zgy5pjce.
-
 The minimum supported version is **7.4**. Available images can be listed using:
 
 ```bash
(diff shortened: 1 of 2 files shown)


docs/operations/images.md
```diff
@@ -87,8 +87,18 @@ aws ec2 describe-images --region us-east-1 --output table \
...
-Please [report](https://github.com/kubernetes/kops/issues/new/choose) any changes.
+CentOS 8 is based on Kernel version **4.18** which fixes some of the bugs present in RHEL/CentOS 7 and effects are less visible.
+
+One notable change is the addition of `iptables` NFT, which is the only iptables backend available. This may not be supported by some CNI plugins and should be used with care.
+
+Available images can be listed using:
+
+```bash
+aws ec2 describe-images --region us-east-1 --output table \
+  --owners 679593333241 \
```
rifelpet, line 98: So this is the account ID that is kops' `WellKnownAccountCentOS` but the account ID in the CentOS wiki is different: https://wiki.centos.org/Cloud/AWS It would be nice to figure out why they're different and if we should switch to the one mentioned upstream.

hakman: This is the AWS Marketplace account. Changed it a bit and now references the `AWS Provided Marketplace Images with Updates`:
https://wiki.centos.org/Cloud/AWS#AWS_Provided_Marketplace_Images_with_Updates

We could move to their new owner ID and images, but don't know if there's a lot of value and we would have to change the WellKnownAccountCentOS also.

hakman: Thought about it a  bit more and just changed it to the new CentOS repo. Easier to understand for anyone and probably will get the latest updates.

PR #10369, opened 2020-12-05
Template functions for recommended kubernetes versions

olemarkus:
This PR implements two template functions for setting recommended k8s version based on info from the given channel.

### DIFF
--- a/pkg/util/templater/template_functions.go
+++ b/pkg/util/templater/template_functions.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package templater
+
+import (
+	"text/template"
+
+	"github.com/Masterminds/sprig/v3"
+	"k8s.io/kops"
+	kopsapi "k8s.io/kops/pkg/apis/kops"
+)
+
+// templateFuncsMap returns a map if the template functions for this template
+func (r *Templater) templateFuncsMap(tm *template.Template) template.FuncMap {
+	// grab the template functions from sprig which are pretty awesome
+	funcs := sprig.TxtFuncMap()
+
+	funcs["indent"] = indentContent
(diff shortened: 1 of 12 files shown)


pkg/util/templater/template_functions.go
```diff
@@ -0,0 +1,66 @@
...
+	funcs["include"] = func(name string, context map[string]interface{}) string {
+		content, err := includeSnippet(tm, name, context)
+		if err != nil {
+			panic(err.Error())
+		}
+
+		return content
+	}
+
+	funcs["ChannelRecommendedKubernetesUpgradeVersion"] = func(version string) string {
```
rifelpet, line 44: These function names are pretty long.

This is essentially the recommended patch version for a given kubernetes version and the function below is the recommended minor+patch version for a given kops version.

How about:
```
ChannelRecommendedKubernetesPatchVersion
ChannelRecommendedKubernetesMinorVersion
```

or even just
```
ChannelKubernetesPatchVersion
ChannelKubernetesMinorVersion
```
?

olemarkus: Not the easiest functions to name :)

I mainly based the function names after the primary underlying function. I don't mind long function names; it boils down to what most intuitive for users.

olemarkus:
I realise that channel is actually something we pass in as a value.
I could pass channel name as a param to these template functions, but I think the overall code would be more awkward.
A better alternative may be to make the value of the channel CLI arg available as a function too.

Any thoughts?

rifelpet:
Yea, I don't have a strong opinion about how these should work or their name though I do see the value in having them. Perhaps your original names are sufficient.

I do think its worthy of some documentation so that users don't have to dig into the source code to find the function names.

I'm also wondering if not specifying --channel should use the default channel or not load any channel and have those functions fail somehow. I suppose it only really impacts airgapped environments since `toolbox template` used to not make any network requests and now it does by default. If we go with the default channel then we should consider a release note mentioning the change in behavior.

olemarkus:
My documentation got lost somewhere ... I had actually written that.

Those who run kops in airgapped envs could point to their local channel files. I would guess they have one in order to update their clusters as well.

I'll dig up those docs again, add a bit of release notes, but otherwise keep this PR as is. There is sufficient time to change anything later on. And I have some more similar functions I want to add such as recommended image.

Recommended image would also require cloud to be passed though. Which means either passing that as a variable or as another CLI arg similar to the channel discussion above.

rifelpet:
Yes, I think all of that sounds good 👍🏻 

Airgapped users can use their local channel file but they'll need to _know_ to add the `--channel` flag pointing to their local file when upgrading to kops 1.20, which is what I think the release notes could help with.

olemarkus:
Actually saving files to disk seems to make a difference ... docs pushed.

rifelpet:
This LGTM. Once the conflict is resolved we can get it merged

Issue #10356, opened 2020-12-02
Docker should not be the default runtime


AngerM-DD:
With k8s 1.20 Docker support as the CRI is being deprecated. The default should probably be changed to containerd and a warning placed for users creating clusters with docker that it will not be supported in the future.

see: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#deprecation

hakman:
Thanks for the news. We will discuss how to notify the users and switch to containerd by default in the near future.

PR #10370, opened 2020-12-05
Set default container runtime to containerd

bmelbourne:
As Kubernetes will deprecate Docker runtime support from version 1.20 and will be removed in a future release (currently planned for the 1.22 release in late 2021), set the default container runtime to `containerd` from kOps 1.20.

https://github.com/kubernetes/kubernetes/tree/master/CHANGELOG/CHANGELOG-1.20.md#deprecation

#10356 

Merge this PR in to kOps 1.20 release milestone.


olemarkus:
This only changes the default for new clusters. It does not change anything for existing.

I would think that you need a check for k8s version. We only support/test containerd as of k8s 1.18. I suspect we want the default to change only for 1.20+ clusters.

bmelbourne:
> This only changes the default for new clusters. It does not change anything for existing.
> 
> I would think that you need a check for k8s version. We only support/test containerd as of k8s 1.18. I suspect we want the default to change only for 1.20+ clusters.

I'm working on exactly that scenario and testing various cluster configs including Terraform to ensure the default container runtime remains as `docker` for kOps version up to 1.19.

olemarkus:
Right now, it looks like we are setting defaults in two places. The one in `create_cluster.go` that you already changed, and in `pkg/model/components/defaults.go`. I would completely remove the one in `create_cluster.go` and rely on the other one instead. Here you have better access to the kubernetes version.

hakman:
The point of setting the container runtime in 2 places was to be able to change the default only for newly created clusters.
At the moment, there is a pending discussion in the Office Hours meeting about when would be the right time and how to approach this change in kOps.

That being said, I think getting this to ready is a good step and I think 1.20 is a good goal to change the default for newly created clusters.

/ok-to-test
/hold until agreement is reached

olemarkus:
It makes sense to perhaps to make the change only for new clusers, perhaps not. At some point we need to change the default for everyone anyway.

You can have a look at `new_cluster.go` then. There are some options there that are set based on k8s version.

hakman:
I think for now it's ok to not think about k8s version. I think we will switch it on for all new clusters.
To make this work, this will have to be relaxed and not require a containerd version and run `hack/update-expected.sh`:
https://github.com/kubernetes/kops/blob/c86e509dbc394ff7228d07cbc5f68fb16c30e2c5/pkg/model/components/containerd.go#L55-L61

bmelbourne:
> It makes sense to perhaps to make the change only for new clusers, perhaps not. At some point we need to change the default for everyone anyway.
> 
> You can have a look at `new_cluster.go` then. There are some options there that are set based on k8s version.

My initial testing had been based on new code introduced in `new_cluster.go` but with the deprecation release note, I thought this would result in some potentially confusing scenarios for users creating new and updating existing clusters, so I moved the code to `defaults.go` where it could be better maintained going forward.

If users then choose to continue using `Docker` as their (unsupported) container runtime beyond K8s and kOps versions 1.20+, then they would need to override the default setting using either `--container-runtime=docker` or `containerRuntime: docker`. From an Operations perspective, this is very clear and obvious to all support engineers maintaining K8s/kOps clusters.

pkg/model/components/defaults.go
```diff
@@ -37,7 +37,11 @@ func (b *DefaultsOptionsBuilder) BuildOptions(o interface{}) error {
 	}
 
 	if options.ContainerRuntime == "" {
-		options.ContainerRuntime = "docker"
+		if b.Context.IsKubernetesGTE("1.20") {
```
hakman, line 40: Should also check that `options.Docker` is not set, otherwise we will probably break the cluster.

tests/integration/create_cluster/complex/expected-v1alpha2.yaml
```diff
@@ -11,7 +11,6 @@ spec:
   channel: stable
   cloudProvider: aws
   configBase: memfs://tests/complex.example.com
-  containerRuntime: docker
```
hakman, line 14: Should use `1.20.0` as k8s version in these tests, to see how things actually change.

hakman (changes requested):
This looks pretty good. Just a few tweaks remaining.

tests/integration/create_cluster/complex/expected-v1alpha2.yaml
bmelbourne: I'm considering updating all of the `options.yaml` under `tests\integration\create_cluster` to use K8s v1.20.1 and running `hack\update-expected.sh` as the K8s versions are inconsistent across the test yaml scripts, but I'm not sure what the overal purpose of the tests is.

Is it to test the creation of a kOps cluster using the earliest supported K8s version or the latest K8s version?

bmelbourne: This error comes up quite often if I don't update the K8s version with `Networking: calico` option as well.

```
--- FAIL: TestCreateClusterSharedVPC (0.13s)
    create_cluster_integration_test.go:201: error running create cluster: --networking with CNI plugin is required for containerd
```

bmelbourne: If we stick with the unchanged K8s versions, I can just set `containerRuntime: docker` in the `options.yaml` files to overcome this error.

hakman: containerd requires setting a CNI like calico, so all tests should have it. Can you check if this fixes things?

bmelbourne: Yes, I've added that to `tests\integration\create_cluster\complex\options.yaml` and it resolves the issue for this specific test, but should I update all tests to k8s `v1.20.1` or set `ContainerRuntime: docker` for the other tests?

```
Networking: calico
KubernetesVersion: v1.20.1
```

It seems strange to have K8s beta versions e.g. `v1.15.6-beta.1` defined in the integration tests, so I'm considering updating all _create_cluster_ tests to K8s v1.20.1.

hakman: 👍 from me to update all these tests to v1.20.1.

bmelbourne: thanks for confirming @hakman, I believe it's the correct approach to be testing against the latest K8s version.

pkg/model/components/defaults.go
bmelbourne: My new code is not quite working as expected during the kops `edit` and `update` and needs some further debugging.

```
	if options.ContainerRuntime == "" {
		if b.Context.IsKubernetesGTE("1.20") {
			options.ContainerRuntime = "containerd"

			if options.Docker != nil {
				klog.Errorf("Docker options are invalid with the default container runtime, containerd")
			}
		} else {
			options.ContainerRuntime = "docker"
		}
	}
```

bmelbourne: @hakman 
It's working as expected, I've forgotten to remove `ContainerRuntime: containerd` from the cluster spec.

```
$ kops edit cluster --state ${AWS_KOPS_STATE_STORE} --name ${KOPS_CLUSTER_NAME}
E0113 22:24:30.445745   24928 defaults.go:45] Docker options are invalid with the default container runtime, containerd
```

hakman:
@bmelbourne after discussing a bit, decided to set the k8s version from integration tests to "1.20.0" and thought it was best to do it before this PR, to see changes. Also, the networking is set to "cni" to avoid extra customizations for that.
Does this sound good to you? If so, can you rebase and update the PR? I plan to take a look at it this weekend anyway.

bmelbourne:
> @bmelbourne after discussing a bit, decided to set the k8s version from integration tests to "1.20.0" and thought it was best to do it before this PR, to see changes. Also, the networking is set to "cni" to avoid extra customizations for that.
> Does this sound good to you? If so, can you rebase and update the PR? I plan to take a look at it this weekend anyway.

That was good timing, I'm just in the process of rebasing the other `containerd` changes now.
Once I've completed my own integration testing, I should have everything committed sometime tomorrow.

hakman:
Thanks! No worries, I can look at the code without the tests part too.

pkg/model/components/containerd.go
```diff
@@ -44,6 +45,10 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 	containerd := clusterSpec.Containerd
 
 	if clusterSpec.ContainerRuntime == "containerd" {
+		if b.IsKubernetesLT("1.18") {
+			klog.Warningf("kubernetes %s is untested with containerd", clusterSpec.KubernetesVersion)
+		}
+
```
hakman, line 51: Is there any technical reason to add this?

pkg/model/components/defaults.go
```diff
@@ -37,7 +38,15 @@ func (b *DefaultsOptionsBuilder) BuildOptions(o interface{}) error {
...
+			options.ContainerRuntime = "containerd"
+
+			if options.Docker != nil {
+				klog.Errorf("Docker options are invalid with the default container runtime, containerd")
+			}
+		} else {
+			options.ContainerRuntime = "docker"
+		}
```
hakman, line 49:
```suggestion
		if b.Context.IsKubernetesLT("1.20") || options.Docker != nil {
			options.ContainerRuntime = "docker"
		} else {
			options.ContainerRuntime = "containerd"
		}
```

upup/pkg/fi/cloudup/containerd_test.go
```diff
@@ -33,20 +33,6 @@ func TestContainerdVersionUrlHash(t *testing.T) {
...
-			hash:    "4616971c3ad21c24f2f2320fa1c085577a91032a068dd56a41c7c4b71a458087",
-			err:     nil,
-		},
-		{
-			arch:    architectures.ArchitectureArm64,
-			version: "1.3.4",
-			url:     "https://download.docker.com/linux/static/stable/aarch64/docker-20.10.0.tgz",
-			hash:    "6e3f80e8451ecbe7b3559247721c3e226be6b228acaadee7e13683f80c20e81c",
-			err:     nil,
-		},
```
hakman, line 49: I am a bit puzzled about the changes in this file. Versions shouldn't matter that much. I would just leave it as it was, but would like to hear your thoughts on this.

docs/releases/1.20-NOTES.md
```diff
@@ -4,6 +4,8 @@
 
 # Significant changes
 
+* Default container runtime set to `containerd` for K8s versions >= 1.20.0.
```
hakman, line 7:
```suggestion
* Default container runtime is now set to `containerd` for new clusters running Kubernetes 1.20.0+.
```

docs/cluster_spec.md
```diff
@@ -937,7 +937,10 @@ spec:
 ## containerRuntime
 {{ kops_feature_table(kops_added_default='1.18', k8s_min='1.11') }}
 
-Alternative [container runtimes](https://kubernetes.io/docs/setup/production-environment/container-runtimes/) can be used to run Kubernetes. Docker is still the default container runtime, but [containerd](https://kubernetes.io/blog/2018/05/24/kubernetes-containerd-integration-goes-ga/) can also be selected.
+Alternative [container runtimes](https://kubernetes.io/docs/setup/production-environment/container-runtimes/) can be used to run Kubernetes. The default container runtime is [containerd](https://containerd.io/docs/).
+
+Docker can still be selected but be aware that Kubernetes is deprecating support for the Docker runtime from version 1.20 and will be removed in a future release (currently planned for the 1.22 release in late 2021).
+@see [Don't Panic: Kubernetes and Docker](https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/)
```
hakman, line 943:
```suggestion
As of Kubernetes 1.20, the default [container runtime](https://kubernetes.io/docs/setup/production-environment/container-runtimes) is containerd. Previously, the default container runtime was Docker.

Docker can still be used as container runtime with Kubernetes 1.20+,  but be aware that Kubernetes is [deprecating](https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker) support for it and will be removed in Kubernetes 1.22.
```

hakman (changes requested):
As promised, I made another pass over the PR. Looks pretty well, just small changes needed.
Please rebase and use the k8s versions and network settings that were updated in https://github.com/kubernetes/kops/pull/10591. The changes to the integration tests should be smaller like this.

pkg/model/components/containerd.go
bmelbourne: Ole Markus requested this change back in 6th Dec.

hakman: I think we decided later during one of the calls that is not necessary anymore. Should be ok to remove as we are testing all supported versions with containerd.

upup/pkg/fi/cloudup/containerd_test.go
bmelbourne: I've focused the testing on the latest versions containerd `1.3.x` and `1.4.x`.

I can put the code back in if there's a specific reason for testing `1.3.4`?

hakman: These are more like unit tests that go through various functionalities of the module. Has not nothing to do with latest versions. I would prefer not to change the file at all unless there is something that changes there.

bmelbourne: I've reverted the deletion of the earlier versions and added `1.4.3` unit tests.

bmelbourne:
> > @bmelbourne after discussing a bit, decided to set the k8s version from integration tests to "1.20.0" and thought it was best to do it before this PR, to see changes. Also, the networking is set to "cni" to avoid extra customizations for that.
> > Does this sound good to you? If so, can you rebase and update the PR? I plan to take a look at it this weekend anyway.
> 
> That was good timing, I'm just in the process of rebasing the other `containerd` changes now.
> Once I've completed my own integration testing, I should have everything committed sometime tomorrow.

I've completing the rebasing of the latest `master` changes and local integration testing.

I've also updated `tests\integration\create_cluster\minimal\options.yaml` to K8s v1.20.0 as I believe this was missed in previous commits.

hakman:
This looks perfect now @bmelbourne. Thanks!
/retest
/lgtm
/approve

PR #10371, opened 2020-12-05
Add K8s Docker runtime support deprecation release note

bmelbourne:
As Kubernetes will deprecate Docker runtime support from version 1.20 and will be removed in a future release (currently planned for the 1.22 release in late 2021), notify users that the default container runtime will be set to `containerd` from kOps 1.20.

https://github.com/kubernetes/kubernetes/tree/master/CHANGELOG/CHANGELOG-1.20.md#deprecation

#10356 

PR #10370 raised to set default container runtime to `containerd` in kOps 1.20.

### DIFF
--- a/docs/releases/1.19-NOTES.md
+++ b/docs/releases/1.19-NOTES.md
@@ -113,10 +113,9 @@ has been updated by a newer version of kOps unless it is given the `--allow-kops
 
 # Deprecations
 
-* Support for Kubernetes versions 1.11 and 1.12 are deprecated and will be removed in kOps 1.20.
+* Support for Kubernetes versions 1.11 and 1.12 has been deprecated and will be removed in kOps 1.20.
 
-* As of kOps 1.20, the default container runtime will be `containerd` instead of Docker for new clusters. For existing clusters the default behaviour can be overridden using either the `--container-runtime` flag or add `spec.containerRuntime` to the cluster spec. This change closely tracks the upstream deprecation of Docker in Kubernetes 1.20 and will be removed in a future release.
-  For more information see [this page](https://github.com/kubernetes/kubernetes/tree/master/CHANGELOG/CHANGELOG-1.20.md#deprecation).
+* Support for Docker container runtime has been deprecated and will be replaced with containerd for new clusters as of kOps 1.20. For existing clusters the default behaviour can be overridden by setting `spec.containerRuntime: containerd` in the cluster spec. This change closely tracks the upstream [deprecation of Docker](https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker).
 
 * Support for Terraform version 0.11 has been deprecated and will be removed in kOps 1.20.
 
--- a/docs/releases/1.19-NOTES.md
+++ b/docs/releases/1.19-NOTES.md
@@ -115,6 +115,9 @@ has been updated by a newer version of kOps unless it is given the `--allow-kops
 
 * Support for Kubernetes versions 1.11 and 1.12 are deprecated and will be removed in kOps 1.20.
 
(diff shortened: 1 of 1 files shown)


olemarkus:
Any reason not to do this as part of #10370 ?

Also, I think we should follow the k8s deprecation here. Meaning that kOps will still use docker as default prior to k8s 1.20.

bmelbourne:
> Any reason not to do this as part of #10370 ?
> 
> Also, I think we should follow the k8s deprecation here. Meaning that kOps will still use docker as default prior to k8s 1.20.

That's essentially the purpose of this deprecation release note (once merged in KOps 1.19)...we'll be notifying users that `containerd` will become the default container runtime from kOps 1.20 (as covered in #10370).

olemarkus:
Ah, it is for the 1.19 notes.

/cc @hakman

geojaz (commented):
@bmelbourne Thanks! This lgtm, but I think it could use just a quick iteration.

I think we should call out that kOps won't do anything (automatically) to existing clusters running on Docker. And as long you're iterating, I would prefer the note to be something like:
`As of kOps 1.20, new clusters will default to use containerd instead of Docker. Existing clusters can define their upgrade path manually. This change in default behavior is in line with the upstream deprecation of Docker in K8s 1.20.`

I like this because it calls out the action item clearly at the beginning and leaves the explanation and further details for those who are interested. Also, this way, we're not trying to predict the future and possibly having to fix later :)

MoShitrit:
Hey @bmelbourne 
It looks like the `CI/verify` test is still failing, complains that it needs you to run `make apimachinery`.
My guess is that since your branch is 27 days old, it might've been forked off one with failing tests? 🤷🏼‍♂️ 
Either way, I could suggest that you try and pull `master` into your branch and see if it helps, or alternatively run `make apimachinery`, even though I'm not sure why would a release notes change cause that to fail.

rifelpet:
The failing test is due to https://github.com/kubernetes/kops/issues I'd like to address that once and for all rather than punting this to next year. I'm hoping we can get that fixed soon so I wouldn't worry about the failing test for now

geojaz:
@bmelbourne I think the other stuff is ready now so we can do one last rebase and `make apimachinery` :)

docs/releases/1.19-NOTES.md
```diff
@@ -115,6 +115,9 @@ has been updated by a newer version of kOps unless it is given the `--allow-kops
 
 * Support for Kubernetes versions 1.11 and 1.12 are deprecated and will be removed in kOps 1.20.
 
+* As of kOps 1.20, the default container runtime will be `containerd` instead of Docker for new clusters. For existing clusters the default behaviour can be overridden using either the `--container-runtime` flag or add `spec.containerRuntime` to the cluster spec. This change closely tracks the upstream deprecation of Docker in Kubernetes 1.20 and will be removed in a future release.
+  For more information see [this page](https://github.com/kubernetes/kubernetes/tree/master/CHANGELOG/CHANGELOG-1.20.md#deprecation).
```
hakman, line 119:
```suggestion
* Support for Kubernetes versions 1.11 and 1.12 has been deprecated and will be removed in kOps 1.20.

* Support for Docker container runtime has been deprecated and will be replaced with containerd for new clusters as of kOps 1.20. For existing clusters the default behaviour can be overridden by setting `spec.containerRuntime: containerd` in the cluster spec. This change closely tracks the upstream [deprecation of Docker](https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker).
```

hakman:
Considering that most periodic tests are passing using containerd, we decided in yesterday's Office Hours to go ahead with the deprecation as described here. Thanks @bmelbourne, I will look at the other PR that changes the default later this week.
/hold cancel
/lgtm
/approve

geojaz (approved):
/lgtm

PR #10378, opened 2020-12-07
Rename duplicate ci target to quick-ci

hakman:
This will avoid some make warnings.

/cc @olemarkus


hakman:
Small mistake in how this should be handled, but fixed now.
/hold cancel

hakman:
Thanks @olemarkus

PR #10381, opened 2020-12-07
WIP: Demonstrate client-side mode for CoreDNS operator

justinsb:
This PR (the second commit) demonstrates the possibility of a mode where we client-side expand operators.

The trick is that we pass a `--dry-run` flag to the operator, which takes the CR and "related objects" on stdin, and outputs the manifest on stdout.  We run the operator in docker today, which would be a new dependency for this mode (we could eventually support our own sandboxing on Linux, but docker feels like the cross-OS sandbox today).

The "clever bit" is those other objects - that is how we pass e.g. the kube-dns ClusterIP.  The operator then runs in a fake API environment, and is able to observe any objects we pass in.  I imagine we'll end up with a longer list of objects we pass in, but this feels like a sustainable approach.

I think we can use an image label to "know" how to invoke an operator in dry-run mode (cf https://github.com/kubernetes/enhancements/pull/1481)


justinsb:
cc @johngmyers I think this might satisfy your use-case where you want less server-side logic, but still allow us to extract these binaries from kops?  Two gotchas I can think of are:

1. that we are executing containers client-side (both a docker dependency and maybe a security risk?  But we can do things like disabling networking)
1. how can we trust the operator images?  I think we could address this by ensuring that our core operators are under the control of the kubernetes-sigs release practices.

fabriziopandini:
I like the idea of having the operator (the addon) to be the owner of all the domain specific bits of the provisioning, but the dependency on docker is kind of tricky if you plan to run your tool on actual Kubernetes nodes like we do in kubeadm.

However, if we can define a contract around this idea that includes "standard" way to interact for managing the lifecycle of the addon, this could be a interesting element to build on top.

I can see other applications for this ideas, op top of my mind:
- upgrading component config files
- generated artefacts for managing the lifecycle of operator-less addon

justinsb:
I just closed the other rollup PR, but this one has the client-side mode we need for people that don't want to run operators.

PR #10396, opened 2020-12-09
Mount /lib64 for Protokube only on AMD64

hakman:
The /lib64 dir doesn't exist on ARM64, at least not for Ubuntu 20.04.
/cc @rifelpet @olemarkus


nodeup/pkg/model/protokube.go
```diff
@@ -214,13 +215,18 @@ func (t *ProtokubeBuilder) ProtokubeContainerRunCommand() (string, error) {
...
 			"--volume /bin:/bin:ro",
 			"--volume /lib:/lib:ro",
-			"--volume /lib64:/lib64:ro",
 			"--volume /sbin:/sbin:ro",
 			"--volume /usr/bin:/usr/bin:ro",
 			"--volume /var/run/dbus:/var/run/dbus",
 			"--volume /run/systemd:/run/systemd",
 		}...)
 
+		if t.Architecture == architectures.ArchitectureAmd64 {
```
rifelpet, line 224: would it be better to just check if the directory exists rather than keying off of architecture?

hakman: This is a short term solution until we get the same image as the one used in etcd-manager.
I think for now this is the simplest solution. Usually these checks are not done outside NodeUp tasks.

PR #10397, opened 2020-12-09
Add control-plane node role label to cp nodes

olemarkus:
Following kubeadm and others, we should change the control plane label and taints.

Personally I want to minimise the time we use both labels, so the timeline is agressive. But consider the timeline more a suggestion open for discussion.


docs/releases/1.20-NOTES.md
```diff
@@ -14,6 +16,8 @@
 
 # Required Actions
 
+* If your workload targets control plane nodes, you need to change them to select the `node-role.kubernetes.io/control-plane=""` label. You should also add the `node-role.kubernetes.io/control-plane:NoSchedule` taint to these workloads. This taint will not be added to control plane nodes before kOps 1.21.
```
johngmyers, line 19:
```suggestion
* If your workload targets control plane nodes, you need to change them to select the `node-role.kubernetes.io/control-plane=""` label. You should also add the `node-role.kubernetes.io/control-plane:NoSchedule` toleration to these workloads. This taint will not be added to control plane nodes before kOps 1.21.
```

pkg/nodelabels/builder.go
```diff
@@ -22,23 +22,25 @@ import (
 )
 
 const (
-	RoleLabelName15        = "kubernetes.io/role"
-	RoleLabelName16        = "kubernetes.io/role"
-	RoleMasterLabelValue15 = "master"
-	RoleNodeLabelValue15   = "node"
+	RoleLabelName15              = "kubernetes.io/role"
+	RoleLabelName16              = "kubernetes.io/role"
+	RoleControlPlaneLabelValue15 = "master"
```
johngmyers, line 27: This change strikes me as being misleading.

pkg/nodelabels/builder.go
```diff
@@ -22,23 +22,25 @@ import (
...
+	RoleNodeLabelValue15         = "node"
 
-	RoleLabelMaster16 = "node-role.kubernetes.io/master"
-	RoleLabelNode16   = "node-role.kubernetes.io/node"
+	RoleLabelControlPlane16 = "node-role.kubernetes.io/master"
```
johngmyers, line 30: Also misleading.

pkg/nodelabels/builder_test.go
```diff
@@ -61,12 +61,13 @@ func TestBuildNodeLabels(t *testing.T) {
...
-				RoleLabelName15:   RoleMasterLabelValue15,
-				"master1":         "master1",
-				"master2":         "master2",
-				"node1":           "override1",
-				"node3":           "override3",
+				RoleLabelControlPlane16: "",
+				RoleLabelControlPlane20: "",
+				RoleLabelName15:         RoleControlPlaneLabelValue15,
+				"master1":               "master1",
+				"master2":               "master2",
```
johngmyers, line 68: Might as well change these two.

johngmyers (changes requested):
I think 1.21 is far too aggressive for removal of the legacy labels.

pkg/nodelabels/builder.go
olemarkus: You want to keep the master name in this case?

pkg/nodelabels/builder_test.go
olemarkus: Makes sense

pkg/nodelabels/builder.go
johngmyers: Because the value will have to continue to be "master", yes.

docs/releases/1.20-NOTES.md
```diff
@@ -20,6 +22,8 @@
 
 * If you are running `kops toolbox template` in an airgapped environment, you have to set `--channel` to point to a local channel file.
 
+* If your workload targets control plane nodes, you need to change them to select the `node-role.kubernetes.io/control-plane=""` label. You should also add the `node-role.kubernetes.io/control-plane:NoSchedule` toleration to these workloads. This taint will not be added to control plane nodes before kOps 1.22.
```
ddelange, line 25: Hi 👋 

I arrived here from 1.23 release notes: https://github.com/kubernetes/kops/blob/v1.23.0/docs/releases/1.23-NOTES.md?plain=1#L48

I'm running kOps v1.22.2 and can confirm there is no matching taint as suggested in this PR. So when I replace my `node-role.kubernetes.io/master:NoSchedule` toleration as suggested here, my pod won't schedule.

Is it possible someone forgot to add the new taint? Or, if was it added in between v1.22 and v1.23, can someone point me to the source code?

Many thanks!

olemarkus: The notes say you need both taints to be forward proof

PR #10404, opened 2020-12-10
Calico: Allow operators to choose which encapsulation mode to use

### LINKED ISSUE #10168: Allow Calico to use VXLAN backend (instead of BIRD)
Calico—in particular, its "calico-node" daemon—has a configurable "networking backend" that accepts one of three values:
- _bird_  
  Run the BIRD BGP daemon, for use with IP-in-IP or VXLAN encapsulation.
- _vxlan_  
  Disable BIRD, and allow use of VXLAN encapsulation.
- _none_  
  Disable BIRD, and use Calico only for network policy enforcement.

We can specify this value via [the "calico_backend" field in the "calico-node" _ConfigMap_](https://docs.projectcalico.org/getting-started/kubernetes/installation/config-options#other-configuration-options) or [the "CALICO_NETWORKING_BACKEND" environment variable](https://docs.projectcalico.org/reference/node/configuration#other-environment-variables).

At present, _kops_ forces Calico to use the BIRD backend unconditionally. When using only VXLAN encapsulation in our IP pools, though, we don't need BGP, and so we don't need BIRD. It would be helpful if we could customize Calico via the _Cluster_ manifest's "networking.calico" field to designate use of the VXLAN backend instead. _kops_ would then need to amend the "calico-node" _DaemonSet_ manifest to [remove the `-bird-live` and `-bird-ready` command-line flags from the "calico-node" pod's liveness and readiness probes](https://docs.projectcalico.org/getting-started/kubernetes/installation/config-options#switching-from-ip-in-ip-to-vxlan), respectively.

Note that I tried using _kubectl patch_ to adjust this configuration myself in a _kops_-managed cluster. While that works to a point, it's a fragile arrangement; any subsequent invocation of _kops update cluster_ will apply _kops_'s manifests again, updating the Calico configuration back to using the BIRD backend.

seh:
Introduce a new "encapsulationMode" field in Calico's portion of the _Cluster_ specification to allow switching between the the IP-in-IP and VXLAN encapsulation protocols. For now, we accept the values "ipip" and "vxlan," and forgo a possible "none" value that would disable encapsulation altogether (at least for the default Calico IP pool).

Augment the default-populating procedure for Calico to take this field into account when deciding both which networking backend to use and whether to use IP-in-IP or VXLAN encapsulation for the default IP pool. Note that these values supplied for [the "CALICO_IPV4POOL_IPIP" and "CALICO_IPV4POOL_VXLAN" environment variables](https://docs.projectcalico.org/reference/node/configuration#configuring-the-default-ip-pools) in the "calico-node" _DaemonSet_ pod spec only matter for creating the "default" _IPPool_ pool object when no such objects already exist.

Generalize the documentation for the "crossSubnet" field to cover environments more broad than just AWS, as Calico can employ this selective encapsulation in any environment in which it can detect boundaries between subnets.

Fixes #10168.

### DIFF
--- a/docs/networking/calico.md
+++ b/docs/networking/calico.md
@@ -30,21 +30,41 @@ kops create cluster \
 
 ## Configuring
 
-### Enable Cross-Subnet mode in Calico (AWS only)
+### Select an Encapsulation Mode
 
-Calico supports a new option for IP-in-IP mode where traffic is only encapsulated
-when it’s destined to subnets with intermediate infrastructure lacking Calico route awareness
-– for example, across heterogeneous public clouds or on AWS where traffic is crossing availability zones/ regions.
+In order to send network traffic to and from Kubernetes pods, Calico can use either of two networking encapsulation modes: [IP-in-IP](https://tools.ietf.org/html/rfc2003)  or [VXLAN](https://tools.ietf.org/html/rfc7348). Though IP-in-IP encapsulation uses fewer bytes of overhead per packet than VXLAN encapsulation, [VXLAN can be a better
+choice when used in concert with Calico's eBPF dataplane|https://docs.projectcalico.org/maintenance/troubleshoot/troubleshoot-ebpf#poor-performance]. In particular, eBPF programs can redirect packets between Layer 2 devices, but not between devices at Layer 2 and Layer 3, as is required to use IP-in-IP tunneling.
 
-With this mode, IP-in-IP encapsulation is only [performed selectively](https://docs.projectcalico.org/v3.10/networking/vxlan-ipip#configure-ip-in-ip-encapsulation-for-only-cross-subnet-traffic).
-This provides better performance in AWS multi-AZ deployments, and in general when deploying on networks where
-pools of nodes with L2 connectivity are connected via a router.
+kOps chooses the IP-in-IP encapsulation mode by default, it still being the Calico project's default choice, which is equivalent to writing the following in the cluster spec:
+```yaml
+  networking:
+    calico:
+      encapsulationMode: ipip
+```
+To use the VXLAN encapsulation mode instead, add the following to the cluster spec:
+```yaml
+  networking:
+    calico:
+      encapsulationMode: vxlan
+```
 
-Note that Calico by default, routes between nodes within a subnet are distributed using a full node-to-node BGP mesh.
+As of Calico version 3.17, in order to use IP-in-IP encapsulation, Calico must use its BIRD networking backend, in which it runs the BIRD BGP daemon in each "calico-node" container to distribute routes to each machine. With the BIRD backend Calico can use either IP-in-IP or VXLAN encapsulation between machines. For now, IP-in-IP encapsulation requires maintaining the routes with BGP, whereas VXLAN encapsulation does not. Conversely, with the VXLAN backend, Calico does not run the BIRD daemon and does not use BGP to maintain routes. This rules out use of IP-in-IP encapsulation, and allows only VXLAN encapsulation. Calico may remove this need for BGP with IP-in-IP encapsulation in the future.
+
+### Enable Cross-Subnet mode in Calico
+
+Calico supports a new option for both of its IP-in-IP and VXLAN encapsulation modes where traffic is only encapsulated
+when it’s destined to subnets with intermediate infrastructure lacking Calico route awareness—for example, across
+heterogeneous public clouds or on AWS where traffic is crossing availability zones.
+
+With this mode, encapsulation is only [performed selectively](https://docs.projectcalico.org/v3.10/networking/vxlan-ipip#configure-ip-in-ip-encapsulation-for-only-cross-subnet-traffic).
+This provides better performance in AWS multi-AZ deployments, or those spanning multiple VPC subnets within a single AZ, and in general when deploying on networks where pools of nodes with L2 connectivity are connected via a router.
+
+Note that by default with Calico—when using its BIRD networking backend—routes between nodes within a subnet are
+distributed using a full node-to-node BGP mesh.
 Each node automatically sets up a BGP peering with every other node within the same L2 network.
 This full node-to-node mesh per L2 network has its scaling challenges for larger scale deployments.
 BGP route reflectors can be used as a replacement to a full mesh, and is useful for scaling up a cluster. [BGP route reflectors are recommended once the number of nodes goes above ~50-100.](https://docs.projectcalico.org/networking/bgp#topologies-for-public-cloud)
-The setup of BGP route reflectors is currently out of the scope of kops.
+The setup of BGP route reflectors is currently out of the scope of kOps.
 
 Read more here: [BGP route reflectors](https://docs.projectcalico.org/reference/architecture/overview#bgp-route-reflector-bird)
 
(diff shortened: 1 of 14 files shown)


pkg/apis/kops/networking.go
```diff
@@ -103,7 +103,10 @@ type FlannelNetworkingSpec struct {
 
 // CalicoNetworkingSpec declares that we want Calico networking
 type CalicoNetworkingSpec struct {
-	// BpfEnabled enables the eBPF dataplane mode.
+	// AWSENISrcDstCheck enables/disables ENI source/destination checks (AWS only)
+	// Options: DoNothing (default), Enable, or Disable
+	AWSENISrcDstCheck string `json:"awsSrcDstCheck,omitempty"`
```
hakman, line 108: This name comes from [libcalico-go](https://github.com/projectcalico/libcalico-go/blob/728c26dd8282e7ee95b09ea1f9c5ccee116d814b/lib/apis/v3/felixconfig.go#L355-L357) and I would like to keep it.
I pulled some of the changes related to ordering of fields and comments to https://github.com/kubernetes/kops/pull/10408 to reduce the change list, I hope you don't mind

upup/models/bindata.go
```diff
@@ -33358,7 +33358,7 @@ data:
   # You must set a non-zero value for Typha replicas below.
   typha_service_name: "{{- if .Networking.Calico.TyphaReplicas -}}calico-typha{{- else -}}none{{- end -}}"
   # Configure the backend to use.
-  calico_backend: "bird"
+  calico_backend: {{ .Networking.Calico.Backend }}
```
hakman, line 33361: I wonder why exactly we need `Backend`...
```suggestion
  {{- if ne CalicoIPv4PoolIPIPMode "Never" }}
  calico_backend: "bird"
  {{- else }}
  calico_backend: "vxlan"
  {{- end }}
```

upup/models/bindata.go
```diff
@@ -37284,7 +37284,9 @@ spec:
               command:
               - /bin/calico-node
               - -felix-live
+              {{- if eq .Networking.Calico.Backend "bird" }}
```
hakman, line 37287:
```suggestion
              {{- if ne CalicoIPv4PoolIPIPMode "Never" }}
```

upup/models/bindata.go
```diff
@@ -37293,7 +37295,9 @@ spec:
               command:
               - /bin/calico-node
               - -felix-ready
+              {{- if eq .Networking.Calico.Backend "bird" }}
```
hakman, line 37298:
```suggestion
              {{- if ne CalicoIPv4PoolIPIPMode "Never" }}
```

hakman (changes requested):
I wonder if we really need Backend added. What do you think?

pkg/apis/kops/networking.go
olemarkus: I tend to ignore upstream if it conflicts with kops consistency, but I don't mind if you insist on keeping it. This is cosmetic and is not exposed to users.

hakman: If you mean the case part, I agree was an oversight. But adding ENI, I am not a big fan.

upup/models/bindata.go
seh: One thing that I think we’d get wrong is that the “Always” and “Never” values only impact at most one _IPPool_ object, but it’s possible to operate Calico with several such _IPPools_, each configured differently. An operator could decide that the default _IPPool_ should use VXLAN, but another pool should use IP-in-IP always, and another pool should use IP-in-IP only across subnet boundaries.

With that arrangement, we’d set the backend to “vxlan,” which would preclude the operator from creating those other _IPPools_ like that using IP-in-IP encapsulation.

Those two “mode” variables only induce Calico to create an _IPPool_ like that if no other _IPPools_ exist. But even after that, an operator can edit that default pool, create more of them, and even delete the default _IPPool_.

hakman: Thanks for the explanation.

pkg/apis/kops/networking.go
seh: I included "ENI" just because these checks are actually defined on the ENI, not on the instance per se, but then again we didn't name the field "AWSEC2InstanceSrcDstCheck," so my pedantry is not warranted here. I'll change it per your recommendation.

seh:
Should we still [allow IP-in-IP traffic in the security rules](https://github.com/kubernetes/kops/blob/e43efbe102528355ffeaca38de02b72b9110d91f/pkg/model/firewall.go#L134-L136) even if the Calico backend is VXLAN?

rifelpet:
is migrating from one backend to another a supported operation for existing clusters? If it is, then modifying security group rule definitions could mean that `kops update cluster` revokes a security group rule while the backend that depended on that access is still in use by a portion of the cluster.

seh:
> Is migrating from one backend to another a supported operation for existing clusters?

Yes, it is, though at present _kOps_ would not orchestrate this all on its own. The reason is that we'd wind up changing the requisite field in the "calico-node" _ConfigMap_, but we wouldn't automatically restart all the pods that need to consume that updated field value in order for the change to actually take effect. After running _kops update cluster_ and waiting for the updated Calico manifests to be applied, an operator would have to run the following command:
```shell
kubectl --namespace=kube-system rollout restart calico-node
```
I am not sure if it's necessary to restart the pods managed by the "calico-kube-controllers" _Deployment_. I will ask the Calico maintainers about that.

seh:
Please hold pending [discussion with the Calico maintainers](https://calicousers.slack.com/archives/C1J26M73P/p1607951261079700).

docs/networking/calico.md
```diff
@@ -55,37 +75,45 @@ To enable this mode in a cluster, add the following to the cluster spec:
...
 ```
-In the case of AWS, EC2 instances have source/destination checks enabled by default.
-When you enable cross-subnet mode in kOps 1.19+, it is equivalent to:
+In the case of AWS, EC2 instances' ENIs have source/destination checks enabled by default.
+When you enable cross-subnet mode in kOps 1.19+, it is equivalent to either:
 ```yaml
   networking:
     calico:
       awsSrcDstCheck: Disable
+      encapsulationMode: ipip
```
hakman, line 84: Maybe would be better to remove this for simplicity as it should be the default.
```suggestion
      encapsulationMode: ipip
```

hakman (changes requested):
Thanks @seh. This looks mostly good. Will take another look tomorrow and give it a try.

seh:
> This looks mostly good.

Thank you for the prompt review. I'll follow up with corrections tomorrow morning (EST).

docs/networking/calico.md
```diff
@@ -55,37 +75,44 @@ To enable this mode in a cluster, add the following to the cluster spec:
...
       IPIPMode: CrossSubnet
 ```
-An IAM policy will be added to all nodes to allow Calico to execute `ec2:DescribeInstances` and `ec2:ModifyNetworkInterfaceAttribute`, as required when [awsSrcDstCheck](https://docs.projectcalico.org/reference/resources/felixconfig#spec) is set.
+or
+```yaml
+  networking:
+    calico:
+      awsSrcDstCheck: Disable
+      encapsulationMode: vxlan
+      IPIPMode: Never
```
hakman, line 92: Still needed, I think this change deprecates it?

seh: I was just trying to show that with the VXLAN encapsulation mode, the effective value of "IPIPMode" becomes "Never," but I guess it doesn't really matter; the main point of these examples is to show that "awsSrcDstCheck" becomes "Disable" automatically.

I'll edit this section again.

docs/networking/calico.md
```diff
@@ -106,8 +133,6 @@ For more details on enabling the eBPF dataplane please refer the [Calico Docs](h
   networking:
     calico:
       bpfEnabled: true
-      bpfExternalServiceMode: Tunnel
-      bpfLogLevel: Info
```
hakman, line 110: Any reason to remove these? They were meant to show that there are some optional parameters there.

seh: I thought it was confusing, as they're not influenced by the choose of cross-subnet. Could we show these in the section on the eBPF dataplane instead?

hakman: Thought this was the eBPF section.

seh: Could we comment them, to indicate that "Tunnel" is the default, but "Info" is not?

hakman:
This looks 👍 . Thanks @seh !
/ok-to-test
/lgtm

PR #10409, opened 2020-12-11
Move bootstrapchannelbuilder to a dedicated package

olemarkus:
This builder got annoyingly long, so I moved it to a dedicated subpackage where we can split it out into multiple files.
Maybe later on these could in turn be moved closer to other addon logic.

/kind cleanup


johngmyers:
We should consider moving the manifests closer to bootstrapchannelbuilder. They don't belong under upup/models anymore.

olemarkus:
I agree. Also consider some of pkg/model/components. Could be an idea to move these to pkg/model/addons/<addon> at some point.

PR #10414, opened 2020-12-12
Expose metrics port when PrometheusMetricsEnabled set to true in Calico

avdhoot:
Problem:
Currently `PrometheusMetricsEnabled` only start servings metrics on `PrometheusMetricsPort` . But Prometheus can not scrape it. This PR will expose that service so Prometheus can scrape it. 
Sample config for Prometheus-operator
```apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  labels:
    k8s-app: calico-node
    monitoredBy: prometheus-infra
  name: calico-node
  namespace: kube-system
spec:
  podMetricsEndpoints:
  - interval: 30s
    path: /metrics
    port: metrics
  namespaceSelector:
    matchNames:
    - kube-system
  selector:
    matchLabels:
      k8s-app: calico-node

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -36961,6 +36961,11 @@ spec:
         - containerPort: 5473
           name: calico-typha
           protocol: TCP
+        {{- if .Networking.Calico.TyphaPrometheusMetricsEnabled }}
+        - containerPort: {{- or .Networking.Calico.TyphaPrometheusMetricsPort "9093" }}
+          name: metrics
+          protocol: TCP
+        {{- end }}
         envFrom:
         - configMapRef:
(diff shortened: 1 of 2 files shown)


hakman:
@avdhoot In the Calico docs I see that the ports are exposed with a service. Should this be added here also?
https://docs.projectcalico.org/maintenance/monitor/monitor-component-metrics#creating-a-service-to-expose-felix-metrics

avdhoot:
@hakman we are using Prometheus [pod](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#pod)  discovery mechanism. I do not think we should create service only to expose metrics.

hakman:
Thanks for the explanation and the contribution. I think will be appreciated by users.
/lgtm
/approve

Issue #10372, opened 2020-12-05
Use systemd as the cgroup driver for kubelet and CRI


bharath-123:
Currently in kops, kubelet and the container runtimes use cgroupfs driver as the cgroup-manager. Kubernetes recommends using systemd as the default cgroup-manager for kubelet and the container runtimes. Some projects like kubeadm and minikube have moved to set systemd as the default cgroup-driver.

https://kubernetes.io/docs/setup/production-environment/container-runtimes/

https://github.com/kubernetes/minikube/pull/6651
https://github.com/freenas/freenas/pull/5263
https://github.com/kubernetes/kubernetes/pull/73837

bharath-123:
Was having a discussion with @bmelbourne on this referring to PR [9879](https://github.com/kubernetes/kops/pull/9879). It seems like there was a consensus that moving to systemd cgroup driver would be a big breaking change for users. 

I do agree that this can be a breaking change. Defaulting to systemd for kubelet and container runtimes is pretty straightforward, we only have to set a couple of default options nothing too much. 

@hakman @olemarkus @bmelbourne would love to know your thoughts on this.

olemarkus:
One can already set systemd as driver for kubelet, right? So what is missing is setting the appropriate driver for containerd.
For me that sounds straightforward enough.

I am not sure what the breaking change would be. From what I can tell, there are mostly benefits from this change.

Compared to the change done for docker, I would really like to see some validation logic ensuring that the containerd and kubelet config match.

hakman:
I would like to not have so much logic for a setting that most people won't notice or change. Defaulting to this for 1.20+ should be ok. Validating the kubelet and container runtime match seems overkill to me. Documentation should be good enough.
Don't remember the discussion about the breaking changes, but may be related to old Debian image.

hakman:
Had a more detailed discussion with @olemarkus and agreed on to do here:
1. doc explaining how to set the cgroup driver and that needs to be done in both k8s and cri
2. doc note for contained configOverride that using it has to be done with care, as the default config may change between releases and can lead to incompatibilities
3. print a warning when configOverride is used that the cluster is using non default containerd config
4. set default to systemd only for k8s >= 1.20 for kubelet, docker and containerd

bharath-123:
alright got these points. Will raise a PR with code keeping all of these in mind. Thanks!

hakman:
Thanks also for doing this.

hakman:
[...]

PR #10419, opened 2020-12-13
Default cgroup driver to systemd from k8s 1.20

bharath-123:
Currently, kOps uses cgroupfs cgroup driver for the kubelet and CRIs. This PR defaults
the cgroup driver to systemd for clusters created with k8s versions >= 1.20.

Using systemd as the cgroup-driver is the recommended way as per
https://kubernetes.io/docs/setup/production-environment/container-runtimes/

It would be really great if this can be tested by others :) !! I have tested this by bring up and tearing down a whole bunch of clusters :) .

I verified that systemd was being used by checking the memory cgroup hierarchy. I was able to see the top level kubepod.slice hierarchy. If cgroupfs driver is used then the directory would be named kubepod instead of kubepod.slice. 

Please do let me know if I have missed anything major here. Getting containerd working was very tricky since we don't have a TOML parser in kOps.

Fixes: https://github.com/kubernetes/kops/issues/10372

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -932,6 +932,7 @@ spec:
 ### Configuration
 
 It is possible to override the [containerd](https://github.com/containerd/containerd/blob/master/README.md) daemon options for all the nodes in the cluster. See the [API docs](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#ContainerdConfig) for the full list of options.
+Overriding the configuration of containerd has to be done with care as the default config may change with new releases and can lead to incompatibilities.
 
 ```yaml
 spec:
@@ -1178,3 +1179,30 @@ spec:
 ```
 
 which would end up in a drop-in file on all masters and nodes of the cluster.
+
+## cgroupDriver
+
+As of Kubernetes 1.20, kOps will default the cgroup driver of the kubelet and the container runtime to use systemd as the default cgroup driver
+as opposed to cgroup fs.
+
+It is important to ensure that the kubelet and the container runtime are using the same cgroup driver. Below are examples showing
+how to set the cgroup driver for kubelet and the container runtime. 
+
+
+Setting kubelet to use cgroupfs
+```yaml
+spec:
+  kubelet:
+    cgroupDriver: cgroupfs
+```
+
+Setting Docker to use cgroupfs
+```yaml
+spec:
+  docker:
+    execOpt:
+      - native.cgroupdriver=cgroupfs
+```
+
+In the case of containerd, the cgroup-driver is dependant on the cgroup driver of kubelet. To use cgroupfs, just update the 
+cgroupDriver of kubelet to use cgroupfs.
--- a/pkg/model/components/containerd.go
+++ b/pkg/model/components/containerd.go
@@ -61,6 +61,7 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 			for name, endpoints := range containerd.RegistryMirrors {
 				config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "registry", "mirrors", name, "endpoint"}, endpoints)
 			}
+			config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", "runc", "runtime_type"}, "io.containerd.runc.v2")
 			containerd.ConfigOverride = fi.String(config.String())
 		}
 
--- a/nodeup/pkg/model/docker_test.go
+++ b/nodeup/pkg/model/docker_test.go
@@ -97,6 +97,10 @@ func TestDockerBuilder_BuildFlags(t *testing.T) {
 			kops.DockerConfig{Bridge: fi.String("br0")},
 			"--bridge=br0",
 		},
+		{
+			kops.DockerConfig{ExecOpt: []string{"native.cgroupdriver=systemd"}},
+			"--exec-opt=native.cgroupdriver=systemd",
+		},
 	}
 
 	for _, g := range grid {
--- a/pkg/model/components/docker.go
+++ b/pkg/model/components/docker.go
@@ -73,5 +73,23 @@ func (b *DockerOptionsBuilder) BuildOptions(o interface{}) error {
 	// and it is an error to specify the flag twice.
 	docker.Storage = fi.String("overlay2,overlay,aufs")
 
+	// Set systemd as the default cgroup driver in docker from k8s 1.20.
+	if b.IsKubernetesGTE("1.20") && getDockerCgroupDriver(docker.ExecOpt) == "" {
+		docker.ExecOpt = append(docker.ExecOpt, "native.cgroupdriver=systemd")
+	}
+
 	return nil
+}
+
+// checks if cgroup-driver is configured or not for docker or not.
+func getDockerCgroupDriver(execOpts []string) string {
+	for _, value := range execOpts {
+		if value == "native.cgroupdriver=systemd" {
+			return "systemd"
+		} else if value == "native.cgroupdriver=cgroupfs" {
+			return "cgroupfs"
+		}
+	}
+
+	return ""
 }
--- a/pkg/model/components/kubelet.go
+++ b/pkg/model/components/kubelet.go
@@ -213,5 +213,10 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
 		}
 	}
 
+	// Set systemd as the default cgroup driver for kubelet from k8s 1.20
+	if b.IsKubernetesGTE("1.20") && clusterSpec.Kubelet.CgroupDriver == "" {
+		clusterSpec.Kubelet.CgroupDriver = "systemd"
+	}
+
 	return nil
 }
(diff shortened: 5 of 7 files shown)


bharath-123:
cc: @hakman @bmelbourne @olemarkus

bharath-123:
as usual, i have forgotten to run make gofmt and static checks. Need to add them to my pre push commits :)

bharath-123:
I think this can reviewed. Would love to get some feedback on this. Made this a WIP mainly so that we can test this out since this can break kOps clusters if not done right.

hakman:
@bharath-123 Thanks for digging into this. It looks mostly good at first sight. Will need some more time to review it and not sure if I can get to it until next week.

One thought is that kOps can use at the moment containerd 1.2+ and this PR introduces 2 things that may be incompatible with older containerd versions:
* config version 2
* runtime version 2 (not sure when it was introduced, but was made default only in 1.4)

I am considering dropping support for 1.2, as it was never advertised as supported or used anywhere.

bmelbourne:
> @bharath-123 Thanks for digging into this. It looks mostly good at first sight. Will need some more time to review it and not sure if I can get to it until next week.
> 
> One thought is that kOps can use at the moment containerd 1.2+ and this PR introduces 2 things that may be incompatible with older containerd versions:
> 
> * config version 2
> * runtime version 2 (not sure when it was introduced, but was made default only in 1.4)
> 
> I am considering dropping support for 1.2, as it was never advertised as supported or used anywhere.

@hakman @bharath-123 
I've actually started working on a similar update for Kubespray where the `containerd` version is being bumped to `1.4.3`, and could incorporate the containerd config v2 into PR #10370...just let me know your preference?

bharath-123:
Seems I don't get emails anymore regarding github! Missed these convos!

@hakman I think we can drop support for containerd 1.2. It has reached EOL https://containerd.io/releases/

@bmelbourne will have a look at your PR!

pkg/model/components/containerd.go
```diff
@@ -61,26 +61,58 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
...
-			containerd.ConfigOverride = fi.String(contents)
+
+			// Use systemd as the default cgroup-driver for containerd for clusters running k8s 1.20
+			if b.IsKubernetesGTE("1.20") {
+				systemdConfig := []string{
+					"  [plugins.\"io.containerd.grpc.v1.cri\".containerd]",
+					"    [plugins.\"io.containerd.grpc.v1.cri\".containerd.runtimes]",
+					"      [plugins.\"io.containerd.grpc.v1.cri\".containerd.runtimes.runc]",
+					"        runtime_type = \"io.containerd.runc.v2\"",
+					"          [plugins.\"io.containerd.grpc.v1.cri\".containerd.runtimes.runc.options]",
```
bmelbourne, line 91: As you're switching to the default CRI plugin runtime `io.containerd.runc.v2` for Containerd 1.4, the `SystemdCgroup` option can be removed as it's only relevant for runtime `io.containerd.runtime.v1.linux`.

https://github.com/containerd/containerd/blob/master/pkg/cri/config/config.go#L213-L216

pkg/model/components/containerd.go
```diff
@@ -61,26 +61,58 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
...
+				"[plugins]",
+			}
+			// Use systemd as the default cgroup-driver for containerd for clusters running k8s 1.20
+			if b.IsKubernetesGTE("1.20") {
+				systemdConfig := []string{
+					"  [plugins.\"io.containerd.grpc.v1.cri\".containerd]",
+					"    [plugins.\"io.containerd.grpc.v1.cri\".containerd.runtimes]",
+					"      [plugins.\"io.containerd.grpc.v1.cri\".containerd.runtimes.runc]",
+					"        runtime_type = \"io.containerd.runc.v2\"",
+					"          [plugins.\"io.containerd.grpc.v1.cri\".containerd.runtimes.runc.options]",
```
bmelbourne, line 108: As you're switching to the default CRI plugin runtime `io.containerd.runc.v2` for Containerd 1.4, the `SystemdCgroup` option can be removed as it's only relevant for `io.containerd.runtime.v1.linux`.

https://github.com/containerd/containerd/blob/master/pkg/cri/config/config.go#L213-L216

pkg/model/components/containerd.go
bharath-123: got it! Thanks for this @bmelbourne !

hakman:
@bharath-123 @bmelbourne sorry for not answering sooner. I spent some time during the holidays to simplify the code and remove the unsupported parts like kubenet and 1.2.x.
Also had some time to think about this and the settings. We should still keep `io.containerd.runc.v2` to set this for 1.3.x.
@bharath-123 if you could update the PR, I can take a look at it and review it these days.

bharath-123:
Sure @hakman. I ll update this PR by tommorow for a review.

bharath-123:
Thanks @olemarkus ! @hakman do have a look at this when you get the time :)

olemarkus:
/lgtm

You can probably remove the WIP bit

bharath-123:
Thanks @olemarkus ! I will remove the WIP

hakman (changes requested):

pkg/model/components/containerd.go
```diff
@@ -61,7 +63,16 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 			for name, endpoints := range containerd.RegistryMirrors {
 				config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "registry", "mirrors", name, "endpoint"}, endpoints)
 			}
+
+			//default cgroup-driver to systemd from k8s 1.20 onwards
+			if b.IsKubernetesGTE("1.20") {
```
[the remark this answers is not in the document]
bharath-12345: That sounds like a good idea. We would need to add a new api field and just add an extra check in the containerd code. Will do this.

pkg/model/components/containerd.go
```diff
@@ -19,6 +19,8 @@ package components
 import (
 	"fmt"
 
+	"k8s.io/klog/v2"
+
```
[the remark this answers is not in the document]
bharath-12345: will do

bharath-123:
I think we may not require any code in containerd for v1.4. It seems that the containerd cgroup-driver depends on the kubelet cgroup-driver based on my tests. I have removed all cgroup-driver related code in containerd and kept toggling the kubelet cgroup-driver between cgroupfs and systemd. I have noticed that the pods created depend solely on the cgroup driver of kubelet. I ll do some more research on this as this can avoid any extra containerd which will be great

hakman:
@bharath-123 still need to support 1.3.x though, not just 1.4.x.

bharath-123:
@hakman i assumed that k8s 1.20 on kOps will use only containerd 1.4. I based that off this https://github.com/kubernetes/kops/blob/master/pkg/model/components/containerd.go#L49 but now to think about it, this is only if the user does not specify the containerd version right?

hakman:
Exactly. Also to make it easier, we should still use runtime v2 by default to allow kubelet to decide what cgroup-driver to use. Any thoughts?

bharath-123:
I am not sure if this would break existing clusters running on runc runtime type v1 since containerd v < 1.4 run on runtime v1?  I can't confirm that kubelet decides which cgroup-driver is used tbh. I have put a message in #sig-node and #containerd to confirm from the cri and containerd maintainers.

bharath-123:
Based on discussions on slack, we will be defaulting to runtime v2 on containerd. Have pushed the latest code. Would love a round of review :) @bmelbourne @hakman @olemarkus

docs/cluster_spec.md
```diff
@@ -932,6 +932,7 @@ spec:
 ### Configuration
 
 It is possible to override the [containerd](https://github.com/containerd/containerd/blob/master/README.md) daemon options for all the nodes in the cluster. See the [API docs](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#ContainerdConfig) for the full list of options.
+Overriding the configuration of containerd has to be done with care as the default config may change with releases which can lead to incompatibilities.
```
hakman, line 935:
```suggestion
Overriding the configuration of containerd has to be done with care as the default config may change with new releases and can lead to incompatibilities.
```

hakman (changes requested):

pkg/model/components/containerd.go
```diff
@@ -61,9 +61,11 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 			for name, endpoints := range containerd.RegistryMirrors {
 				config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "registry", "mirrors", name, "endpoint"}, endpoints)
 			}
+
+			config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", "runc", "runtime_type"}, "io.containerd.runc.v2")
+
 			containerd.ConfigOverride = fi.String(config.String())
 		}
-
```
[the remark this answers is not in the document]
bharath-123: formatting issues :(

hakman:
Time to see it in action. Thanks @bharath-123!
/lgtm
/approve

PR #10421, opened 2020-12-14
Bump go-bindata and use go module

olemarkus:


rifelpet:
This likely needs updating

https://github.com/kubernetes/kops/blob/52ea87d6ba38b18a3ff3e547f0b0a8c951299d21/upup/models/BUILD.bazel#L24

Its unfortunate that we define the go-bindata command in both bazel and Makefile. Maybe we can consolidate somehow.

olemarkus:
Nice catch. Thanks.

PR #10425, opened 2020-12-15
Bump etcd client to 3.4.13. Use go modules

olemarkus:


hakman:
Looks pretty nice. Maybe you can look etcd-manager also when you have some time.
/lgtm
/approve

PR #10434, opened 2020-12-16
Update aws_launch_template_nodes.sharedvpc.example.com_user_data

wangxy518:
update an url


rifelpet:
Hi @wangxy518 this file is actually generated automatically based on this source:

https://github.com/kubernetes/kops/blob/d248e3ed13d8007abe200f527de9cb0ffe83229a/pkg/model/resources/nodeup.go#L151

So if you update that .go file and run `./hack/update-expected.sh` it should regenerate this and many other files to include the change.

PR #7919, opened 2019-11-12
Make CoreDNS the default DNS server

rajansandeep:
This will install CoreDNS as the default DNS server instead of kube-dns

Fixes https://github.com/kubernetes/kops/issues/6500

PR #10435, opened 2020-12-16
Delay defaulting to CoreDNS to k8s v1.20

hakman:
1.19 may be a bit too late in the game.
/cc @olemarkus @rifelpet

### DIFF
--- a/upup/pkg/fi/cloudup/template_functions.go
+++ b/upup/pkg/fi/cloudup/template_functions.go
@@ -538,15 +538,20 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
 func (tf *TemplateFunctions) AWSCCMTag() (string, error) {
 	var tag string
 	parsed, err := util.ParseKubernetesVersion(tf.Cluster.Spec.KubernetesVersion)
-
-	// Update when we have stable releases
 	if err != nil {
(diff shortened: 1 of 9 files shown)


upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -538,15 +538,20 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
...
+
+	// Update when we have stable releases
+	switch parsed.Minor {
+	case 18:
 		tag = "v1.18.0-alpha.1"
-	} else if parsed.Minor == 19 {
+	case 19:
 		tag = "v1.19.0-alpha.1"
+	default:
+		tag = "latest"
```
rifelpet, line 552: Why is this changing to latest? That might cause undesired behavior

rifelpet (commented):
also we should add a release note for https://github.com/kubernetes/kops/pull/7919

upup/pkg/fi/cloudup/template_functions.go
hakman: Changing this because there was no tag for 1.20, and we need one for the tests. I think there should be a default value here somehow.

hakman:
> also we should add a release note for #7919

Agreed, will add a release note soon.

PR #10437, opened 2020-12-16
IAM ServiceAccount Roles: truncate name at 64 characters

justinsb:
The maximum IAM role name length is 64 characters, which we hit much
more often now that we are constructing complex names.  Use our normal
strategy of adding a hash when we truncate.

This is not a breaking change, because these names were not valid
previously.

### DIFF
--- a/pkg/model/iam/BUILD.bazel
+++ b/pkg/model/iam/BUILD.bazel
@@ -16,6 +16,7 @@ go_library(
         "//pkg/wellknownusers:go_default_library",
         "//upup/pkg/fi:go_default_library",
         "//upup/pkg/fi/cloudup/awstasks:go_default_library",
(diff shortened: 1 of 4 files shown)


justinsb:
This is intended to fix the scenario test: https://k8s-testgrid.appspot.com/google-aws#kops-grid-scenario-public-jwks

rifelpet:
/lgtm
for reference, this will truncate the offending iam role name from

```
dns-controller.kube-system.sa.e2e-kops-scenario-public-jwks.test-cncf-aws.k8s.io
```

to

```
dns-controller.kube-system.sa.e2e-kops-scenario-public-jw-ucdio5
```

which i think provides enough of the cluster name to still be useful to users viewing an account's list of IAM roles.

PR #10443, opened 2020-12-17
Enable Go Race Detector in tests

hakman:
https://blog.golang.org/race-detector

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,7 +180,7 @@ hooks: # Install Git hooks
 
 .PHONY: test
 test: ${BINDATA_TARGETS}  # Run tests locally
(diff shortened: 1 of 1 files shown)


justinsb:
Nice idea @hakman - and thanks for extracting the problems @rifelpet 

#10553 is a fix for OpenStack, #10552 is a fix for AWS.

Both are real errors I believe - the AWS one is trickier, because we should be running the SecurityGroup rule deletion logic after everything else; I proposed a second phase of execution where we run the "FindDeletion" operations.

hakman:
Sounds good. Thanks @justinsb.

PR #10444, opened 2020-12-17
Spotinst: Schedule Ocean Controller to Linux nodes only

liranp:
This PR adds a node affinity to the Ocean Controller deployment to schedule it to Linux nodes only.


hakman:
Thanks @liranp 
/lgtm
/approve

PR #10452, opened 2020-12-17
Spotinst: Support for multiple subnets per zone

liranp:
This PR adds support for multiple subnets per zone by configuring subnets without specifying zone names (according to the Spot API specification).


vmasule:
@here Is there documentation for this, I have a scenario where multiple private subnets should be created in one AZ. Current I can't understand how that can be achieved?

PR #10461, opened 2020-12-18
Make it possible to change the etcd volume type and iops

olemarkus:


hakman:
This is pretty cool.
Only one Q, should this be merged after https://github.com/kubernetes/kops/pull/10453, to add also the throughout field?

olemarkus:
That makes sense. 
/hold

PR #10463, opened 2020-12-18
Update k8s and AMI versions from alpha

MoShitrit:
It's been 9 days since we updated these versions on `alpha` channel, seems like we're good to move to `stable`.


MoShitrit:
Actually, I just noticed [this thread](https://groups.google.com/g/kubernetes-dev/c/dNH2yknlCBA/m/7jq8KMcgAgAJ) that went out yesterday and seems like there's going to be a newer patch release soon with fixing of the tags, so better wait for the corrected versions. We can keep these versions running in `alpha` for now.
/close

PR #10467, opened 2020-12-18
AWS CSI driver

olemarkus:


pkg/apis/kops/validation/aws.go
```diff
@@ -38,9 +38,25 @@ func awsValidateCluster(c *kops.Cluster) field.ErrorList {
...
 	return allErrs
 }
 
+func awsValidateExternalCloudControllerManager(c kops.ClusterSpec) (allErrs field.ErrorList) {
+
+	if c.ExternalCloudControllerManager != nil {
+		if c.KubeControllerManager == nil || c.KubeControllerManager.ExternalCloudVolumePlugin != "aws" {
+			if c.CloudConfig == nil || c.CloudConfig.AWSEBSCSIDriver == nil || !fi.BoolValue(c.CloudConfig.AWSEBSCSIDriver.Enabled) {
+				allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "externalCloudControllerManager"),
+					"AWS external CCM cannot be used without enabling spec.cloudConfig.AWSEBSCSIDriver or setting spec.kubeControllerManaager.externalCloudVolumePlugin set to `aws`"))
```
rifelpet, line 52: I'm wondering if this is needs to be a requirement. Users may have clusters that don't run any persistent volumes, for example.

olemarkus: The challenge right now is that everyone else need to actively set one of these values to get a working cluster. I think that when we by default enable either using external CCM for AWS and/or external CSI driver we can relax the relax the requirement.

I suspect that the CSI driver will be more popular than the external CCM for the time being anyway. CSI driver does bring in additional features.

olemarkus:
Not sure how much we really want to be configurable here. 
In particular, there is a large amount of images involved and it could get messy allowing each of them to be configured.

upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
```diff
@@ -0,0 +1,591 @@
...
+      containers:
+        - name: ebs-plugin
+          image: k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v0.8.0
+          imagePullPolicy: IfNotPresent
+          args:
+            - controller
+            - --endpoint=$(CSI_ENDPOINT)
+            - --logtostderr
+            - --k8s-tag-cluster-id={{ ClusterName }}
+            - --extra-volume-tags=KubernetesCluster={{ ClusterName }}
```
rifelpet, line 452: This flag is deprecated in favor of `--extra-tags`:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/f54138034204850e54c26f67dde3f8217339c09c/cmd/options/controller_options.go#L39-L40

Also we could consider populating this with the standard set of tags we include on resources, including `ClusterSpec.CloudLabels`

rifelpet (commented):
Looking at the [helm chart's values file](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/v0.8.0/charts/aws-ebs-csi-driver/values.yaml), I don't see any parameters that would be particularly important to expose as api fields for an initial implementation.

I think there are additional IAM permissions needed. The docs mention [ec2:CreateSnapshot](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/f54138034204850e54c26f67dde3f8217339c09c/docs/example-iam-policy.json#L8) which neither the control plane nor node roles have, and the ebs-snapshot-controller StatefulSet isn't currently restricted to just the control plane nodes. Perhaps this is a good opportunity to use the [UseServiceAccountIAM](https://github.com/kubernetes/kops/pull/9352) feature flag functionality.

upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
olemarkus: Good catch.

olemarkus:
> I think there are additional IAM permissions needed. The docs mention [ec2:CreateSnapshot](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/f54138034204850e54c26f67dde3f8217339c09c/docs/example-iam-policy.json#L8) which neither the control plane nor node roles have, and the ebs-snapshot-controller StatefulSet isn't currently restricted to just the control plane nodes. Perhaps this is a good opportunity to use the [UseServiceAccountIAM](https://github.com/kubernetes/kops/pull/9352) feature flag functionality.

I like the IRSA concept, and I think strategically we should move in that direction, but is it mature enough to start depending on it now?

olemarkus:
Reading https://github.com/kubernetes-csi/external-snapshotter in more detail, it seems worthy having a dedicated addon for this. Including the snapshotter, CRDs + the webhook admission controller.

olemarkus:
Looking through https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md and the docs of the various plugins, I wonder if for kOps, it makes most sense to:
* deploy the vendor CSI driver itself as a standalone addon. The controller running as a daemonset on masters
* deploy the external-* plugins as a standalone addon _daemonset_. Need to use hostmounts to communicate with above CSI socket
* The snapshotter should also be a daemonset rather than a statefulset (the use of statefulset was somewhat inconsistent. The reason is to avoid running two of them at the same time. But snapshotter supports leader election, so for us a daemonset makes more sense. Especially given that it talks to the local CSI unix socket).

I don't think there is any particular point in running the snapshotter as a separate pod as long as we manage it outside of the EBS CSI driver itself.

olemarkus:
This is more or less done now.

Since we need a cert for the admission controller, this now blocked by #10321

mikesplain:
This is awesome, unfortunately I only found this PR after having set this up myself as well haha.  

Only comment, would we want to also modify the storage class for this? For this to be usable you need to modify the `provisioner: ebs.csi.aws.com` in https://github.com/kubernetes/kops/blob/22a9a13abf29aa24a7d2c85641a4a96827164be5/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml#L7

I could see that being a follow up PR or a separate flag but just thought I'd mention it.  Also an easy time to enable volume expansion by default.

olemarkus:
Updating the provisioner field is forbidden, so we need a new storageclass here. Or force-change the old one.

Do you know if changing this really is needed when `CSIMigrationAWS` feature gate is set?

olemarkus:
> **CSIMigrationAWS**: Enables shims and translation logic to route volume operations from the AWS-EBS in-tree plugin to EBS CSI plugin. Supports falling back to in-tree EBS plugin if a node does not have EBS CSI plugin installed and configured. Requires CSIMigration feature flag enabled.
**CSIMigrationAWSComplete**: Stops registering the EBS in-tree plugin in kubelet and volume controllers and enables shims and translation logic to route volume operations from the AWS-EBS in-tree plugin to EBS CSI plugin. Requires CSIMigration and CSIMigrationAWS feature flags enabled and EBS CSI plugin installed and configured on all nodes in the cluster.

Looks like we should not create a new storage class. And we probably should set the `CSIMigrationAWSComplete` flag to ensure our setup works as expected.

olemarkus:
I removed the snapshotter out of the PR now, so we shouldn't have any webhooks issues.
Also set the feature flag that will block the in-cluster EBS plugin so we 100% know the out-of-tree plugin is used.

k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -214,6 +214,14 @@ spec:
               cloudConfig:
                 description: CloudConfiguration defines the cloud provider configuration
                 properties:
+                  awsEBSCSIDriver:
```
rifelpet, line 217: There are `azure` and `openstack` fields within `cloudConfig`, i wonder if it makes sense to put this in an `aws` field? Given there are other aws-related fields in `cloudConfig` directly so it wouldn't be consistent until we can move them in a future apiVersion.

pkg/model/components/kubelet.go
```diff
@@ -213,5 +213,16 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
...
+	if clusterSpec.CloudConfig != nil && clusterSpec.CloudConfig.AWSEBSCSIDriver != nil && fi.BoolValue(clusterSpec.CloudConfig.AWSEBSCSIDriver.Enabled) {
+		if _, found := clusterSpec.Kubelet.FeatureGates["CSIMigrationAWSComplete"]; !found {
+			clusterSpec.Kubelet.FeatureGates["CSIMigrationAWSComplete"] = "true"
+		}
+	}
+	if _, found := clusterSpec.Kubelet.FeatureGates["ExperimentalCriticalPodAnnotation"]; !found {
+		if b.IsKubernetesLT("1.16") {
+			clusterSpec.Kubelet.FeatureGates["ExperimentalCriticalPodAnnotation"] = "true"
+		}
+	}
```
rifelpet, line 225: I think this is duplicated from Line 210

rifelpet (commented):
A few minor comments but looks good overall. We'll want docs for this as well as a release note. Being sure to mention the migration steps involving the feature gates, or just linking to wherever that process is documented.

k8s/crds/kops.k8s.io_clusters.yaml
olemarkus: If I remember correctly, we discussed this earlier and opted for keeping consistency. So AWS fields would remain in cloudconfig in this API version.

pkg/model/components/kubelet.go
olemarkus: Thanks for catching!

PR #10469, opened 2020-12-18
Boot nodes without state store access

justinsb:
kops-controller can now serve the instance group & cluster config to
nodes, as part of the bootstrap process.

This enables nodes to boot without access to the state
store (i.e. without S3 / GCS / etc permissions)

Feature-flagged behind the KopsControllerStateStore feature-flag.


justinsb:
This is a revamp of #8198

olemarkus:
This is really cool.

A challenge I can see is that we add another sizeable chunk to cloudinit. I wonder if it makes sense for kOps to have a new public readable VFS path (e.g a new S3 bucket) for public data where both cluster CA and JWKS can be stored. There are other benefits to that as well, such as being able to generate a kubectl config without having access to secret store (for those who don't use e.g OIDC).

justinsb:
@olemarkus it's a good point.  We do add a few small fields and then a (larger) CA certificate.  My thought was that we could continue to move the _other_ fields to be provided by kops-controller, so that it paves the way to reducing the size pretty dramatically.

I do agree that we should have that public bucket for IRSA, but I don't know if we want to require it (?)

olemarkus:
To some extent these values are needed to determine if a node needs rotating or not. But adding a hash of those values would do the trick as well. Not sure if that may affect the usefulness of `kops update cluster`

Agree that the public VFS should be optional.

olemarkus:
@justinsb do you mind if we set milestone to 1.21? I am hoping for a short cycle on 1.20 and this feature may need just a little bit more time. Just in case. We can merge this one in just after we cut the first 1.20 alpha.

justinsb:
That's a good idea.  I'd like to get the 1.21 branch open quickly in that case - it'll help us stabilize 1.20

/milestone 1.21

PR #10471, opened 2020-12-19
Add new-pod-scale-up-delay in Cluster Autoscaler spec

akshedu:
Allow specifying `newPodScaleUpDelay` in cluster autoscaler spec which cases autoscaler to ignore unscheduled pods for the given interval


akshedu:
Signed CLA

olemarkus:
You perhaps want to make use of this variable in the CAS template as well?
`upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template`

akshedu:
I signed it

hakman:
Thanks @akshedu. Looks good from my point of view.
@olemarkus leaving the final decision and cherrypick to you.
/approve

olemarkus:
/lgtm

Thanks for this!

PR #10472, opened 2020-12-19
Replace (some) deprecated ResourceHolder with Resource

justinsb:
This removes more of the deprecated type, but it also simplifies 
refactoring the GCE InstanceTemplate.


hakman:
Something is not working right here with the Prow tests. Even the CLA check is not reporting the status.

PR #7974, opened 2019-11-20
WIP: Install docker from static binary archive

austinmoore-:
Add the functionality to install docker from a static binary archive.

This solution was discussed in the November 8th office hours. This is a solution to a problem that I've run into where I cannot download docker binaries from download.docker.com and must use an internal mirror. The initial fix I had for this problem was in #6956, but that solution doesn't handle heterogenous clusters.

NOTE: This assumes that all instance groups in your cluster are running with the same CPU architecture.

PR #6956, opened 2019-05-13
Override Docker download URL

austinmoore-:
Allows users to specify where to download Docker with a new `spec.dockerInstall.overrideSources` spec. See the added [docs](https://github.com/austinmoore-/kops/blob/eea95c26eaf75af9b5a597f378f6ecd89d5c1dbe/docs/cluster_spec.md#source-overrides) for details on how to use it.

This is effectively a way to override the built in [dockerVersions](https://github.com/austinmoore-/kops/blob/eea95c26eaf75af9b5a597f378f6ecd89d5c1dbe/nodeup/pkg/model/docker.go#L89) slice and specify your own. This provides a nice escape hatch if you cannot download from external repos and must use internal mirrors (as in my company's situation). Ideally this would just be a url and hash. However, some packages have dependencies that need to be specified and other packages need to be installed simultaneously with other packages.

There are some considerations to keep in mind with this feature. This solution isn't the most intuitive for a user because the user will probably just end up looking at `dockerVersions` to find their OS and Docker version. If the user changes their OS or Docker version, this might cause problems if they forget to update this configuration. I'm open to ideas on how to handle these situations.

I think this might be a stop-gap measure for a more robust solution. For example, in the future it might be better to provide a way to override the base download URL like we do for some other resources. I didn't do this because I figured it was better to expose full control for something like this. I also wasn't sure if we need to consider mirrored repos that do not have the same path (not sure how common this would be).

Another option could be to allow a way to externalize the data in `dockerVersions` to a YAML file in S3 as part of the cluster that the user could then edit or specify.

PR #10473, opened 2020-12-19
Add config options for container runtime package URL and Hash

hakman:
A more recent implementation of #6956 and #7974.


hakman:
FYI @michaelajr.

Issue #6356, opened 2019-01-17
Support additional kubelet flags in componentconfig 


silveryfu:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

Some kubelet flags listed here: 

https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

..are seemingly not supported by kops: 

https://github.com/kubernetes/kops/blob/f5ce011fa036034e8b40d0dc94bc5bfc2419bc56/pkg/apis/kops/componentconfig.go

We found some are useful, e.g., --housekeeping-interval.

**2. Feel free to provide a design supporting your feature request.**

nareshku:
good explanation about performance improvements with `--housekeeping-interval` option enabled.
https://github.com/kubernetes/kubernetes/issues/18044

silveryfu:
@nareshku Are you suggesting that the decision of "having this flag enabled by default" should be made by kops and for all kops users? Excuse me but did you read this issue carefully?? :)

CarpathianUA:
It will be useful to have the ability to configure a kubelet's housekeeping.

integrii:
We have been having issues with Docker crashing due to [PLEG errors](https://www.ibm.com/support/knowledgecenter/en/SSBS6K_3.2.0/troubleshoot/docker_pods_overload.html).  The suggested mitigation is to increase the housekeeping interval described here.  Shame that this feature is not part of kops!

jsonmp-k8:
Can this be enabled for the kubelet?

rifelpet:
/reopen
This seems pretty straight forward and would be a good candidate for #9920

Issue #6150, opened 2018-12-04
Cluster Nodes & Masters Failing kube-bench Security Checks - Question


cheynewallace:
I just ran `kube-bench` (https://github.com/aquasecurity/kube-bench)  across my KOPS 1.10 provisioned clusters (Kubernetes `1.10.3` -  kope.io/k8s-1.9-debian-jessie-amd64-hvm-ebs-2018-03-11)  and I was quite shocked to see probably 80 of tests failed. 

I expected a few, but now quite this many.  Is there any reason why there seems to be so many security issues with both node and master machines provisioned with KOPS?

Some output
```
[INFO] 2 Worker Node Security Configuration
[INFO] 2.1 Kubelet
[FAIL] 2.1.1 Ensure that the --allow-privileged argument is set to false (Scored)
[FAIL] 2.1.2 Ensure that the --anonymous-auth argument is set to false (Scored)
[FAIL] 2.1.3 Ensure that the --authorization-mode argument is not set to AlwaysAllow (Scored)
[FAIL] 2.1.4 Ensure that the --client-ca-file argument is set as appropriate (Scored)
[FAIL] 2.1.5 Ensure that the --read-only-port argument is set to 0 (Scored)
[FAIL] 2.1.6 Ensure that the --streaming-connection-idle-timeout argument is not set to 0 (Scored)
[FAIL] 2.1.7 Ensure that the --protect-kernel-defaults argument is set to true (Scored)
[PASS] 2.1.8 Ensure that the --make-iptables-util-chains argument is set to true (Scored)
[FAIL] 2.1.9 Ensure that the --hostname-override argument is not set (Scored)
[FAIL] 2.1.10 Ensure that the --event-qps argument is set to 0 (Scored)
[FAIL] 2.1.11 Ensure that the --tls-cert-file and --tls-private-key-file arguments are set as appropriate (Scored)
[PASS] 2.1.12 Ensure that the --cadvisor-port argument is set to 0 (Scored)
[FAIL] 2.1.13 Ensure that the --rotate-certificates argument is not set to false (Scored)
[FAIL] 2.1.14 Ensure that the RotateKubeletServerCertificate argument is set to true (Scored)
[FAIL] 2.1.15 Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers (Not Scored)
[INFO] 2.2 Configuration Files
[FAIL] 2.2.1 Ensure that the kubelet.conf file permissions are set to 644 or more restrictive (Scored)
[...]

PR #10486, opened 2020-12-21
Added event-qps and event-burst flags to kubelet

DOboznyi:
#6356 Enable kubelet --event-qps flag.
#6150 Will provide the functionality to fix `[WARN] 4.2.9 Ensure that the --event-qps argument is set to 0 or a level which ensures appropriate event capture (Not Scored)` in kube-bench report (CIS benchmarks)

### DIFF
--- a/pkg/apis/kops/componentconfig.go
+++ b/pkg/apis/kops/componentconfig.go
@@ -207,6 +207,10 @@ type KubeletConfigSpec struct {
 	CgroupDriver string `json:"cgroupDriver,omitempty" flag:"cgroup-driver"`
 	// HousekeepingInterval allows to specify interval between container housekeepings.
 	HousekeepingInterval *metav1.Duration `json:"housekeepingInterval,omitempty" flag:"housekeeping-interval"`
+	// EventQPS if > 0, limit event creations per second to this value.  If 0, unlimited.
+	EventQPS *int32 `json:"eventQPS,omitempty" flag:"event-qps" flag-empty:"0"`
+	// EventBurst temporarily allows event records to burst to this number, while still not exceeding EventQPS. Only used if EventQPS > 0.
+	EventBurst *int32 `json:"eventBurst,omitempty" flag:"event-burst"`
 }
 
 // KubeProxyConfig defines the configuration for a proxy
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -591,6 +591,28 @@ spec:
     housekeepingInterval: 30s
 ```
 
+### Event QPS
+{{ kops_feature_table(kops_added_default='1.19') }}
+
+The limit event creations per second in kubelet. Default value is `0` which means unlimited event creations.
+
+```yaml
+spec:
+  kubelet:
+    eventQPS: 0
+```
+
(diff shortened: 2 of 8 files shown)


olemarkus:
Thanks for this.
Wondering if the best practice is to set this to 0, then shouldn't we default it to zero? Defaulting it to omitted, which will explicitly make the test fail, doesn't seem appropriate at least.

DOboznyi:
> Thanks for this.
> Wondering if the best practice is to set this to 0, then shouldn't we default it to zero? Defaulting it to omitted, which will explicitly make the test fail, doesn't seem appropriate at least.

Maybe better to set it to 5. Because 0 can provide a security issue from my perspective. It will make it unlimited.

olemarkus:
I think the CIS check is there because if you rate limit, important events can be missed. Thus the 
> [WARN] 4.2.9 Ensure that the --event-qps argument is set to 0 or a level which ensures appropriate event capture (Not Scored)

Setting it to 5 will retain current behaviour, which is also fine.

DOboznyi:
You're completely right. Where could i add default value as 0? I'm not familiar with kops code base.

DOboznyi:
@olemarkus I changed default value to 0. Could you review PR again?

pkg/apis/kops/componentconfig.go
```diff
@@ -207,6 +207,8 @@ type KubeletConfigSpec struct {
 	CgroupDriver string `json:"cgroupDriver,omitempty" flag:"cgroup-driver"`
 	// HousekeepingInterval allows to specify interval between container housekeepings.
 	HousekeepingInterval *metav1.Duration `json:"housekeepingInterval,omitempty" flag:"housekeeping-interval"`
+	// EventQPS if > 0, limit event creations per second to this value.  If 0, unlimited.
+	EventQPS *int32 `json:"eventQPS,omitempty" flag:"event-qps" flag-empty:"0"`
```
hakman, line 211: I think `event-burst` should also be added together with `event-qps`.
https://github.com/kubernetes/kubernetes/blob/87984d84d18c7816835e5efedebb17ed13aa2509/cmd/kubelet/app/options/options.go#L468

DOboznyi: Done

hakman:
Thanks @DOboznyi. Looks good from my point of view. @olemarkus any thoughts before merging?
/approve

pkg/apis/kops/componentconfig.go
hakman: Could you run `make apimachinery` once more, please?

hakman:
One more thing @DOboznyi, could you squash the commits?

DOboznyi:
@hakman done, all commits squahed.

DOboznyi:
@hakman do you have any ideas why test failed?

hakman:
> @hakman do you have any ideas why test failed?

Flake, no worries. :)

DOboznyi:
> > @hakman do you have any ideas why test failed?
> 
> Flake, no worries. :)

From the second attempt test passed. Is it ready to merge now?

PR #10500, opened 2020-12-27
[WIP] Run protokube as a static pod

bharath-123:
As per our discussions on slack channel #kops-dev, this is a POC to run protokube as a pod to make it independent of the container run time on which kOps runs on. The aim is to understand the edge cases involved in this effort and whether it is worth it or not. I did my best to replicate the current protokube configuration.

Although we would still have dependencies on the container run time for tasks like building the protokube image from a tar file which is a feature not supported by CRI runtimes.

I have tested this configuration by building basic clusters using `kops create cluster` and also by creating a HA cluster like the below. I have tested in both on docker and containerd runtimes. I have tested this with etcd-manager and not legacy etcd. 

[ha-kops.txt](https://github.com/kubernetes/kops/files/5744549/ha-kops.txt)

Seems like github does not support attaching yaml files ^^^

Looking forward for any feedback!

### DIFF
--- a/nodeup/pkg/model/convenience.go
+++ b/nodeup/pkg/model/convenience.go
@@ -85,6 +85,26 @@ func addHostPathMapping(pod *v1.Pod, container *v1.Container, name, path string)
 	return &container.VolumeMounts[len(container.VolumeMounts)-1]
 }
 
+// addHostPathMappingWithMount is shorthand for mapping a host path into a container at a given mount point
+func addHostPathMappingWithMount(pod *v1.Pod, container *v1.Container, name, path, mount string) *v1.VolumeMount {
+	pod.Spec.Volumes = append(pod.Spec.Volumes, v1.Volume{
+		Name: name,
+		VolumeSource: v1.VolumeSource{
+			HostPath: &v1.HostPathVolumeSource{
+				Path: path,
+			},
+		},
+	})
+
+	container.VolumeMounts = append(container.VolumeMounts, v1.VolumeMount{
+		Name:      name,
+		MountPath: mount,
+		ReadOnly:  true,
+	})
+
+	return &container.VolumeMounts[len(container.VolumeMounts)-1]
+}
+
 // addHostPathVolume is shorthand for mapping a host path into a container
 func addHostPathVolume(pod *v1.Pod, container *v1.Container, hostPath v1.HostPathVolumeSource, volumeMount v1.VolumeMount) {
 	vol := v1.Volume{
--- a/nodeup/pkg/model/protokube.go
+++ b/nodeup/pkg/model/protokube.go
@@ -23,6 +23,12 @@ import (
 	"path/filepath"
 	"strings"
 
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kops/pkg/k8scodecs"
+	"k8s.io/kops/pkg/kubemanifest"
+
 	kopsbase "k8s.io/kops"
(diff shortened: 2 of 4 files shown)


olemarkus:
Looks good to me so far. We want to especially test gossip clusters since protokube does play a special role there.

bharath-123:
@olemarkus have tested this by creating a gossip cluster and it does not seem to work on it. It gets stuck at this phase
```
VALIDATION ERRORS
KIND    NAME                                                    MESSAGE
Machine i-07d44250be6a66da9                                     machine "i-07d44250be6a66da9" has not yet joined cluster
Pod     kube-system/kube-dns-696cb84c7-7pxrg                    system-cluster-critical pod "kube-dns-696cb84c7-7pxrg" is pending
Pod     kube-system/kube-dns-autoscaler-55f8f75459-6wfcg        system-cluster-critical pod "kube-dns-autoscaler-55f8f75459-6wfcg" is pending
```
I am not sure as to why this is so as I don't have enough in-depth knowledge on protokube.... need to do some digging here.

bharath-123:
there doesn't seem to be an e2e test for gossip cluster as far as i can tell... is there a reason for that?

olemarkus:
Looks like worker nodes are not able to join the cluster. You'd have to log in to the node to see why.
My guess would be that gossip doesn't work so the node doesn't have the API server DNS.

bharath-123:
after a bit of investigation, the worker node does not have the dns entries of the master as you rightly said @olemarkus . The reason for that i think is because protokube which updates /etc/hosts in the case of gossip clusters does not run because the kubelet is not started. kubelet is necessary now as we are making it a pod or deploymeny. because protokube is not started, /etc/hosts is not updated which leads to calls to kops-controller to fail. 

In the case of protokube running as a standalone container, it runs successfully because the docker container runtime is already present. 

I believe the solution would be to run kubelet before starting protokube. But in my observations it seems like protokube starts before kubelet and since kubelet is necessary for running a pod, protokube does not run.

Please do correct me if i m wrong. I maybe missing something here.

olemarkus:
The challenge is that kubelet depends on nodeup creating the kubelet client certificate. Nodeup depends on talking to kops-controller to provision this cert, which in turn depends on protokube. And protokube won't start because kubelet won't start because it is missing that cert ...

Not sure if you can get kubelet to run pointing at a non-existing cert and then that it reloads the cert after being signed.

Alternative is, as discussed before, that gossip runs as a systemd service.

I would also not object to not allowing cri-o on gossip cluster until we find a better solution.

bharath-123:
I do think that protokube being a systemd service would be better. I am not sure why it is running as a container tbh. 
Even if we disable crio for gossip clusters, we would still need to figure out how to load an image from a tar file without using containerd... that seems to be difficult. Making protokube a systemd service would make things far easier since it just seems to be a binary... I am willing to put in the work to make protokube a systemd service and raise a POC for it.. would like to know if there is any objection or anything i m missing here before moving on with it.

hakman:
@bharath-123 As I said in slack, best to have a decision tomorrow during Office Hours.
If I remember correctly, at some point the idea was to have all kOps running as containers, including NodeUp. @justinsb would know more.
My opinion still stands, I would rather have a decision based on kOps goals, not on CRI-O limitations. The whole podman part is more than annoying. Worst case scenario, we can limit availability for some time to Ubuntu 20.10+, which has podman already installed.

olemarkus:
Another solution is that nodeup uses EC2 tags to discover the API rather than using DNS. This is what gossip does when seeding the cluster anyway.

bharath-123:
Thanks for you feedback @olemarkus and @hakman . Let's discuss this in office hours in more details.

bharath-123:
Closing this PR.

PR #10501, opened 2020-12-27
Remove support for Kubenet with containerd

hakman:
Kubenet is not actually supported for containerd and is unmaintained in general. Our current Kubenet setup for containerd is actually a "hack" that should be used only for development.

I had a longer discussion with some people involved in SIG-Networking and the recommendation is to remove this hack and also to stop using Kubenet in general (especially not as a default).

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -40,6 +40,7 @@ import (
 	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/pkg/kubeconfig"
 	"k8s.io/kops/pkg/kubemanifest"
+	"k8s.io/kops/pkg/model/components"
 	"k8s.io/kops/upup/pkg/fi"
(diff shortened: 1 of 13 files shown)


cmd/kops/create_cluster.go
```diff
@@ -488,6 +488,9 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 	if c.ContainerRuntime != "" {
 		cluster.Spec.ContainerRuntime = c.ContainerRuntime
 	}
+	if cluster.Spec.ContainerRuntime == "containerd" && cluster.Spec.Networking.Kubenet != nil {
```
olemarkus, line 491: There are a couple of other things to test for here too. `Networking.Kopeio` and `Networking.External`. Maybe use `components.UsesCNI(cluster.Spec.Networking) `?

hakman: Thanks. I think this should do it.

PR #10504, opened 2020-12-27
Run k/k's e2e suite via new kubetest2 make target

rifelpet:
ref: https://github.com/kubernetes/test-infra/pull/20349


rifelpet:
I think this is ready for review. The test is failing because kubetest2 doesn't yet download the same kubectl version as the e2e test version because it assumes we're building k/k from source. In our case it falls back to the kubectl baked into the prow job's image which i believe is 1.20.X and causes a minor test failure due to some text change.

Ideally kubetest2 would support downloading the kubernetes client tar along side the [test tar](https://github.com/kubernetes-sigs/kubetest2/blob/master/pkg/testers/ginkgo/package.go#L58) but we can probably build that into our implementation in the short term.

Regardless, I dont think that is a blocker for this PR. Next steps are:
* set KOPS_BASE_URL so that the cluster uses the same nodeup, kops-controller, etc. as the kops cli that is being built - we have a race condition in the current logic and I'm hesitant to add it to the new test code so I might explore alternatives (we publish the URL to a file in the gcs bucket that the tests then reference. two presubmit test runs could overwrite the file and end up using the wrong KOPS_BASE_URL)
* Download the kubernetes client tar and use its kubectl for the e2e tests
* Add manifest template support so that we can customize clusters beyond `create cluster` flags.

justinsb:
This looks great - excited to see this!  I would love to get a kops upgrade test going!

/approve
/lgtm

PR #10507, opened 2020-12-28
Add containerd option for registry mirrors

hakman:
This simplifies the way containerd options can be added, using `pelletier/go-toml`.
Next step for 1.20 would be to move config generation to NodeUp.

/cc @olemarkus @zetaab 
@bharath-123


bharath-12345:
looks good. I ll rework my PR to default systemd as a cgroup-driver to use this toml library. Thanks for doing this!

rifelpet:
can you use the field in an integration test so we can confirm the expected output?

hakman:
> can you use the field in an integration test so we can confirm the expected output?

Sounds good to me. Done.

hakman:
Seems like it's ready now. Thanks everyone :).
/hold cancel

hakman:
Thanks @olemarkus

PR #10516, opened 2021-01-02
Create command for rotating cluster CA

### LINKED ISSUE #1020: Provide full secret rotation
We should provide a way to rotate all our secrets: usernames, tokens, CA key, SSH key.

### LINKED ISSUE #8774: Certificate renewal
Add option to kops that renews certificates during an upgrade similar to kubeadm which can renew all the certificates during control plane upgrade.

Add option `--certificate-renewal` to perform renewal of certificates when upgrading the cluster (hopefully with no service downtime on workers).

```
kops upgrade cluster kubernetes-cluster.example.com --yes --state=s3://kops-state-1234 --certificate-renewal=true
```

### LINKED ISSUE #10498: Automate rotation of cluster secrets
This is a follow-up to https://github.com/kubernetes/kops/issues/8919

Kops clusters have an admin account - the credentials, stored in S3, can be exported with kops export kubecfg.
These credentials are not automatically rotated - the following procedure  describes how to do this manually https://kops.sigs.k8s.io/rotate-secrets/

This however requires running commands via SSH on every node, and is a "disruptive procedure".
Rotating secrets an employee has had access to when that employee leaves the company seems important, especially if those might give admin access to Kubernetes clusters. The current way to do so however, is quite inconvenient.

Could there be a kops command to automate this - ideally in a way that would not disrupt workloads running on the cluster? Or a CronJob that would rotate secrets regularly (every few weeks/months)?
I do appreciate this might be a pretty large undertaking.. is this automatable at all? Many thanks!

olemarkus:
Fixes #8774
Fixes #10498
Fixes #1020 

Pretty much following https://kubernetes.io/docs/tasks/tls/manual-rotation-of-ca-certificates/ but in an order that makes sense for kOps


johngmyers:
My main concern with this is how the admin is supposed to recover if the command fails halfway through, for example with a power loss on the machine running kops.

I designed the similar WIP PR #9556 with smaller, atomic steps. In that PR, to fully distrust the previous keys one has to run the command three times, waiting enough time for tokens to be renewed between each run.

Like that PR, it might be easier to simplify by always keeping a "next CA" (and possibly "previous CA") in the trust list.

Another idea is to put a hash of the relevant CA set in userdata, so worker and/or control plane nodes will be detected as needing update by the rolling update code.

olemarkus:
We rotate the cluster twice in this case, once with a CA bundle where both certs are trusted, and once with only one the new CA after everything trust the new CA.

I have not managed to end up in a state where the cluster is broken. Only manage to lock myself out by explicitly exporting admin credentials signed with the new CA too early, before the api servers trust it.

olemarkus:
@johngmyers I see your PR is about SA signing token keys, which is much more messy as all api servers and all pods should rotate at exactly the same time. There is https://github.com/kubernetes/kubernetes/issues/20165 but it is pretty much unresolved

johngmyers:
@olemarkus my PR is about graceful rotation of SA signing token keys. It is not necessary for all API servers and pods to rotate at the same time.

Since kubernetes/kubernetes#20165 was filed the `--service-account-key-file` flag started taking multiple keys.

olemarkus:
Ah wasn't aware of that. Thanks. I need to have a look on how that one works then.

johngmyers:
@olemarkus Would you be willing to hold this for a while to give #11204 a chance to land first? I believe that puts down some infrastructure that would improve this PR.

olemarkus:
Yeah. I like the ability to decide when to move to the next key rather than always using the highest number. I think that may require an additional roll of the control plane though, as on the first roll, the kube-controller-manager would not use the new keys. Unless we tell it to always use the latest cert even if it is not primary.

cmd/kops/rotate_ca.go
```diff
@@ -0,0 +1,310 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to rotate cluster: %v", err)
+	}
+
+	klog.Info("rotating the control plane")
+
+	//Update masters. This will issue new certs for k8s using the new CA.
+	//New nodes, service accounts etc will use new CA
+	ruo.InstanceGroupRoles = []string{"master", "apiserver"}
+
```
johngmyers, line 303: While the new certs/keys currently take effect without an `apply_cluster`, this is undesirable and there are plans to move more of this stuff to userdata and/or versioned files in the store. We should ensure that changing either the set of trusted CAs or the active CA causes the control plane nodes to be treated as "NeedsUpdate". We should also allow for a future need to do an `apply_cluster`.

cmd/kops/rotate_ca.go
```diff
@@ -0,0 +1,310 @@
...
+	ruo.InstanceGroupRoles = []string{"node"}
+
+	err = RunRollingUpdateCluster(ctx, f, out, ruo)
+	if err != nil {
+		return fmt.Errorf("failed to rotate cluster: %v", err)
+	}
+
+	klog.Info("rotating the control plane")
+
+	//Update masters. This will issue new certs for k8s using the new CA.
```
johngmyers, line 300:
```suggestion
	// Update control plane. This will issue new certs for k8s using the new CA.
```

cmd/kops/rotate_ca.go
```diff
@@ -0,0 +1,310 @@
...
+	ruo.ClusterName = cluster.ObjectMeta.Name
+	ruo.Force = true
+	ruo.InstanceGroupRoles = []string{"node"}
+
+	err = RunRollingUpdateCluster(ctx, f, out, ruo)
+	if err != nil {
+		return fmt.Errorf("failed to rotate cluster: %v", err)
+	}
+
+	klog.Info("rotating the control plane")
```
johngmyers, line 298: The new `certificate-authority-data` needs to be propagated to all the kubeconfig files that live outside the cluster before proceeding to change the signing CA. This is part of the reason I don't think this should be an "all in one go" type of command.

johngmyers (changes requested):

cmd/kops/rotate_ca.go
```diff
@@ -0,0 +1,310 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to delete secondary key: %w", err)
+	}
+
+	if !model.UseKopsControllerForNodeBootstrap(cluster) {
+		return fmt.Errorf("only clusters using kops-controller for boostrapping nodes are supported")
+	}
+
+	exportAdmin := clientConfig.Contexts[contextName].AuthInfo == contextName
```
johngmyers, line 102: This appears to be a bit too heuristic. It should check that the corresponding `AuthInfo` has nonempty `ClientCertificateData`

cmd/kops/rotate_ca.go
```diff
@@ -0,0 +1,310 @@
...
+		return err
+	}
+
+	pool, err := keyStore.FindCertificatePool(fi.CertificateIDCA)
+	if err != nil {
+		return fmt.Errorf("could not fetch the CA pool: %w", err)
+	}
+
+	if len(pool.Secondary) > 0 {
+		klog.Info("Secondary CA cert already in the pool. Not issuing a new CA")
```
johngmyers, line 147: What if the previous run died while removing trust from the old CA?

johngmyers:
Another reason I don't like an all-in-one command: We have a workload (rhymes with CrewBeeper) which frequently fails on a node rolling update. For this reason, we only do node rolling updates with a locally written controller. That controller implements local constraints, such as notifying the relevant workload's maintainers and delaying the update of their instance group until they can be on deck to fix the resulting failure of their workload. So we would not want the command that updates the keys to perform the rolling update.

olemarkus:
/close

Closing in favour of the great work done by John!

PR #10518, opened 2021-01-03
Use custom containerd socket with Flatcar

hakman:
Flatar uses `/run/docker/libcontainerd/docker-containerd.sock` for the containerd socket as mentioned in https://github.com/kinvolk/Flatcar/issues/284.

This should fix the failing periodic test:
https://testgrid.k8s.io/kops-distros#kops-aws-distro-flatcar


rifelpet:
I wonder if this was a recent Flatcar change or if Flatcar always used this socket location and our containerd+Flatcar support just never worked
/lgtm
/hold if you think we should key off of the Flatcar OS version as well, otherwise feel free to remove this.

hakman:
Thanks for the review @rifelpet. Flatcar is a rolling release, so we only care about latest.
/hold cancel

hakman:
This was good as a test run, but seem a more permanent fix will be required:
https://github.com/kinvolk/Flatcar/issues/284#issuecomment-754086168

hakman:
Closing in favor of #10540.
/close

PR #10522, opened 2021-01-04
e2e - dump cluster manifests into artifacts and add --kubernetes-version 

rifelpet:
the cluster manifest will be useful when we add e2e templating support to confirm it is rendered as intended.


rifelpet:
the manifest dump worked, YAML files are here: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kops/10522/pull-kops-e2e-kubernetes-aws-kubetest2/1345925653471432704/artifacts/

the test itself still fails due to the kubectl version mismatch issue, I'm hoping to tackle that next along with the manifest templating.

PR #10523, opened 2021-01-04
Simple upgrade test using kubetest2 framework

justinsb:
Starting very simple and hard-coded!

### DIFF
--- a/tests/e2e/scenarios/upgrade/run-test
+++ b/tests/e2e/scenarios/upgrade/run-test
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+echo "CLOUD_PROVIDER=${CLOUD_PROVIDER}"
+
+REPO_ROOT=$(git rev-parse --show-toplevel);
+
+KUBETEST2_COMMON_ARGS="-v=2 --cloud-provider=${CLOUD_PROVIDER} --cluster-name=${CLUSTER_NAME:-} --kops-binary-path=${REPO_ROOT}/bazel-bin/cmd/kops/linux-amd64/kops"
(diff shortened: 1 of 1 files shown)


justinsb:
cc @rifelpet I think this is the right way to do things?

/assign rifelpet

tests/e2e/scenarios/upgrade/run-test
```diff
@@ -0,0 +1,84 @@
...
+trap finish EXIT
+
+kubetest2 kops ${KUBETEST2_COMMON_ARGS} --up
+
+KUBECONFIG=${HOME}/.kube/config
+TEST_ARGS="--kubeconfig=${KUBECONFIG}"
+if [[ "${CLOUD_PROVIDER}" == "aws" ]]; then
+  ZONES=`kops get cluster ${CLUSTER_NAME} -ojson | jq -r .spec.subnets[].zone`
+  CLUSTER_TAG="${CLUSTER_NAME}"
+  TEST_ARGS="${TEST_ARGS} --provider=aws --gce-zone=${ZONES[0]} --cluster-tag=${CLUSTER_TAG}"
```
rifelpet, line 55: Might be worthy of a comment that --gce-zone is actually provider agnostic (this tripped me up for a minute)

https://github.com/kubernetes/test-infra/blob/1a5044709fb8b5350c99a2a8e5ac208827a04c07/kubetest/kops.go#L692-L697

tests/e2e/scenarios/upgrade/run-test
```diff
@@ -0,0 +1,84 @@
...
+kops set cluster ${CLUSTER_NAME} cluster.spec.kubernetesVersion=1.20.0
+kops update cluster
+kops update cluster --admin --yes
+
+kops rolling-update cluster
+kops rolling-update cluster --yes
+
+kops validate cluster
+
+kubetest2 kops ${KUBETEST2_COMMON_ARGS} --down
```
rifelpet, line 82: any reason not to just let the trap function run every time? You could remove these last 3 lines and simplify the two functions above.

justinsb: My thought was that we wanted to preserve the error, but actually I think that you're right and it's just over-complicating things - we want to know if down failed and there's still a lingering cluster!

tests/e2e/scenarios/upgrade/run-test
justinsb: Good call - added

tests/e2e/scenarios/upgrade/run-test
```diff
@@ -0,0 +1,82 @@
...
+# Always tear-down the cluster when we're done
+function kubetest2_down {
+  kubetest2 kops ${KUBETEST2_COMMON_ARGS} --down
+}
+function finish {
+  if [[ "${RUN_KUBETEST2_DOWN}" == "y" ]]; then
+    kubetest2_down || echo "kubetest2 down failed"
+  fi
+}
+trap finish EXIT
```
rifelpet, line 46:
```suggestion
function finish {
  kubetest2 kops ${KUBETEST2_COMMON_ARGS} --down || echo "kubetest2 down failed"
}
trap finish EXIT
```

justinsb:
Thanks @rifelpet - I incorporated your change!

rifelpet:
👍🏻 
/lgtm

PR #10529, opened 2021-01-05
IRSA - continue adding route53 permisions to masters

rifelpet:
These are needed by protokube to create the kops-controller DNS record to allow nodes to bootstrap. Ideally we could figure out a way for protokube to not rely on the ec2 instance's instance profile and role but until then we can revert to our original behavior here.

See these logs: https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-grid-scenario-public-jwks/1345956556562239488/artifacts/ip-172-20-48-1.sa-east-1.compute.internal/protokube.log

```
I0104 05:03:51.264472    6482 dnscache.go:74] querying all DNS zones (no cached results)
I0104 05:03:51.264570    6482 route53.go:53] AWS request: route53 ListHostedZones
W0104 05:03:51.389485    6482 dnscontroller.go:124] Unexpected error in DNS controller, will retry: error querying for zones: error querying for DNS zones: AccessDenied: User: arn:aws:sts::768319786644:assumed-role/masters.e2e-kops-scenario-public-jwks.test-cncf-aws.k8s.io/i-05b1db10d1a5b8637 is not authorized to perform: route53:ListHostedZones
```

and the nodeup logs on nodes that couldn't join the cluster:

```
Jan 04 04:55:53.500187 ip-172-20-38-84 nodeup[2070]: W0104 04:55:53.500117    2070 executor.go:131] error running task "BootstrapClient/BootstrapClient" (9m52s remaining to succeed): Post "https://kops-controller.internal.e2e-kops-scenario-public-jwks.test-cncf-aws.k8s.io:3988/bootstrap": dial tcp: lookup kops-controller.internal.e2e-kops-scenario-public-jwks.test-cncf-aws.k8s.io on 127.0.0.53:53: no such host
```

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -257,12 +257,10 @@ func (r *NodeRoleMaster) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 		addKMSIAMPolicies(p, stringorslice.Slice(b.KMSKeys), b.Cluster.Spec.IAM.Legacy)
 	}
 
-	if !b.UseServiceAccountIAM {
-		if b.Cluster.Spec.IAM.Legacy {
-			addLegacyDNSControllerPermissions(b, p)
-		}
-		AddDNSControllerPermissions(b, p)
+	if b.Cluster.Spec.IAM.Legacy {
+		addLegacyDNSControllerPermissions(b, p)
 	}
+	AddDNSControllerPermissions(b, p)
 
 	if b.Cluster.Spec.IAM.Legacy || b.Cluster.Spec.IAM.AllowContainerRegistry {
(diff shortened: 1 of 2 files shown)


olemarkus:
If we have to support CSR bootstrap mechanism for bottlerocket, this could be used here as well.

/lgtm

rifelpet:
@justinsb the route53 record is still failing to be created: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-grid-scenario-public-jwks/1346681344700190720

We dont collect the dns-controller logs in the job artifacts but my new theory is that kubetest only allows the prow job's public IP to access apiserver but it needs to be 0.0.0.0/0 so that IAM can reach the JWKS endpoints.

PR #10538, opened 2021-01-05
Refactor and centralize distribution logic

justinsb:
Use of a struct makes it more sustainable, centralizing into the 
distribution package makes it simpler to follow.

### DIFF
--- a/nodeup/pkg/model/kubectl.go
+++ b/nodeup/pkg/model/kubectl.go
@@ -22,7 +22,6 @@ import (
 	"k8s.io/kops/pkg/rbac"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
-	"k8s.io/kops/util/pkg/distributions"
 
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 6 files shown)


nodeup/pkg/model/kubectl.go
```diff
@@ -104,20 +103,9 @@ func (b *KubectlBuilder) Build(c *fi.ModelBuilderContext) error {
...
-		users = []string{"centos"}
-	case distributions.DistributionAmazonLinux2, distributions.DistributionRhel7, distributions.DistributionRhel8:
-		users = []string{"ec2-user"}
-	case distributions.DistributionFlatcar:
-		users = []string{"core"}
-	default:
-		klog.Warningf("Unknown distro; won't write kubeconfig to homedir %s", b.Distribution)
+	users, err := b.Distribution.DefaultUsers()
+	if err != nil {
+		klog.Warningf("won't write kubeconfig to homedir for distribution %s: %v", b.Distribution, err)
```
seh, line 108: Will we start wrapping inner errors via the `%w` verb?

justinsb: I think we should yes, but annoyingly klog doesn't yet recognize it (last I checked!), so we have to use %v.

So yes for fmt.Errorf, not yet for klog

PR #10540, opened 2021-01-06
Add containerd config file to Flatcar based instances

hakman:
This makes it possible to configure Flatcar from kOps and also fixes the issue of custom socket location (different approach to #10518).

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -51,16 +51,14 @@ func (b *ContainerdBuilder) Build(c *fi.ModelBuilderContext) error {
 	switch b.Distribution {
 	case distributions.DistributionFlatcar:
 		klog.Infof("Detected Flatcar; won't install containerd")
-		if err := b.buildContainerOSConfigurationDropIn(c); err != nil {
-			return err
+		if b.Cluster.Spec.ContainerRuntime == "containerd" {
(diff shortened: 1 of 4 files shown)


nodeup/pkg/model/containerd.go
```diff
@@ -216,16 +200,35 @@ func (b *ContainerdBuilder) buildContainerOSConfigurationDropIn(c *fi.ModelBuild
...
+		Path:     "/etc/systemd/system/containerd.service.d/10-kops.conf",
+		Contents: fi.NewStringResource(contents),
+		Type:     nodetasks.FileType_File,
+		OnChangeExecute: [][]string{
+			{"systemctl", "daemon-reload"},
+			{"systemctl", "restart", "containerd.service"},
+			// We need to restart kops-configuration service since nodeup needs to load images
+			// into containerd with the new config. Restart is on the background because
+			// kops-configuration is of type 'one-shot' so the restart command will wait for
+			// nodeup to finish executing
```
seh, line 224:
```suggestion
			// into containerd with the new config. We restart in the background because
			// kops-configuration is of type "one-shot," so the restart command will wait for
			// nodeup to finish executing.
```

hakman: Thanks. This is embarrassing. Just copy pasted this without reading. 😁

PR #10549, opened 2021-01-10
Kubetest2 - use our own tester that wraps kubetest2's ginkgo tester

rifelpet:
This allows us to share tester flags (package version and bucket) for downloading kubectl while passing them to the ginkgo tester. The kubectl.go file is essentially copied from kubetest2-tester-ginkgo's [package.go](https://github.com/kubernetes-sigs/kubetest2/blob/master/pkg/testers/ginkgo/package.go) that downloads the e2e.test binary

### DIFF
--- a/tests/e2e/pkg/tester/kubectl.go
+++ b/tests/e2e/pkg/tester/kubectl.go
@@ -1,0 +1,192 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tester
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/kubetest2/pkg/exec"
+)
+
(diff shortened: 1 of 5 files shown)


olemarkus:
Looks like it is still using 1.20.1 client against the 1.19.4 test cases.

tests/e2e/pkg/tester/kubectl.go
```diff
@@ -0,0 +1,192 @@
...
+		lines, err := exec.OutputLines(cmd)
+		if err != nil {
+			return "", fmt.Errorf("failed to get latest release name: %s", err)
+		}
+		if len(lines) == 0 {
+			return "", fmt.Errorf("getting latest release name had no output")
+		}
+		t.Ginkgo.TestPackageVersion = lines[0]
+
+		klog.V(1).Infof("Kubectl package version was not specified. Defaulting to latest: %s", t.Ginkgo.TestPackageVersion)
```
justinsb, line 55: I think we could log this without V(1) (i.e. at v=0), particularly as it seems like it's still giving us some trouble!

tests/e2e/pkg/tester/tester.go
```diff
@@ -0,0 +1,66 @@
...
+	return t.Ginkgo.Test()
+}
+
+func (t *Tester) pretestSetup() error {
+	kubectlPath, err := t.AcquireKubectl()
+	if err != nil {
+		return fmt.Errorf("failed to get kubectl package from published releases: %s", err)
+	}
+	existingPath := os.Getenv("PATH")
+	os.Setenv("PATH", fmt.Sprintf("%v:%v", kubectlPath, existingPath))
```
justinsb, line 47: I _think_ kubectlPath is the full path to kubectl here.  If that's right, we would probably want something like `filepath.Dir(kubectlPath)`

justinsb:
I like the approach - taking control of this code will let us better understand what's happening in the tests and add our own tests.

rifelpet:
Ok it looks like the upstream tester's flag parsing doesn't happen until after we can download kubectl, so the bucket and version strings aren't yet set.

https://github.com/kubernetes-sigs/kubetest2/blob/5a41a5dbafe58ff0dc59a85acc6cd62f08db5d8f/pkg/testers/ginkgo/ginkgo.go#L114-L117

I'll do some more experimenting locally and try and come up with a better way of wrapping the upstream ginkgo tester

olemarkus:
Looks good now!
/lgtm

PR #10551, opened 2021-01-10
Spotinst: Specify Spot percentage per Instance Group

liranp:
### Description

This PR allows users to specify Spot percentage per Instance Group (i.e. Ocean Launch Spec).

### Usage Example
```yaml
kind: InstanceGroup
metadata:
  labels:
    ...
    spotinst.io/spot-percentage: "70"
  name: ...
spec:
  ...
```


hakman:
Looks good. Thanks @liranp.
/lgtm
/approve

Issue #10506, opened 2020-12-27
Data Races reported by `go test -race`


rifelpet:
https://github.com/kubernetes/kops/pull/10443 exposed a few data races that are worth triaging:

<details>
<summary>awstasks.(*SecurityGroup).FindDeletions()</summary>

```
==================
WARNING: DATA RACE
Read at 0x00c00096b9b0 by goroutine 29:
  k8s.io/kops/upup/pkg/fi/cloudup/awstasks.(*SecurityGroup).FindDeletions()
      src/k8s.io/kops/upup/pkg/fi/cloudup/awstasks/securitygroup.go:417 +0x898
  k8s.io/kops/upup/pkg/fi.DefaultDeltaRunMethod()
      src/k8s.io/kops/upup/pkg/fi/default_methods.go:88 +0x30b
  k8s.io/kops/upup/pkg/fi/cloudup/awstasks.(*SecurityGroup).Run()
      src/k8s.io/kops/upup/pkg/fi/cloudup/awstasks/securitygroup.go:133 +0x4e
  k8s.io/kops/upup/pkg/fi.(*executor).forkJoin.func1()
      src/k8s.io/kops/upup/pkg/fi/executor.go:179 +0x30d

Previous write at 0x00c00096b9b0 by goroutine 39:
  k8s.io/kops/upup/pkg/fi/cloudup/awstasks.(*SecurityGroup).Find()
      src/k8s.io/kops/upup/pkg/fi/cloudup/awstasks/securitygroup.go:83 +0x507
  runtime.call64()
      /usr/local/Cellar/go/1.15.5/libexec/src/runtime/asm_amd64.s:541 +0x3d
  reflect.Value.Call()
      /usr/local/Cellar/go/1.15.5/libexec/src/reflect/value.go:337 +0xd8
  k8s.io/kops/util/pkg/reflectutils.InvokeMethod()
      src/k8s.io/kops/util/pkg/reflectutils/walk.go:77 +0x413
  k8s.io/kops/upup/pkg/fi.invokeFind()
      src/k8s.io/kops/upup/pkg/fi/default_methods.go:123 +0xc9
  k8s.io/kops/upup/pkg/fi.DefaultDeltaRunMethod()
      src/k8s.io/kops/upup/pkg/fi/default_methods.go:47 +0xa33
  k8s.io/kops/upup/pkg/fi/cloudup/awstasks.(*SecurityGroup).Run()
[...]

PR #10553, opened 2021-01-10
Openstack: Prevent data race in servergroup member list

justinsb:
We were adding to the ServerGroup without a mutex, so we introduce a mutex.

Also introduce some defense against the member list changing once we've
observed it, though this is already enforced by GetDependencies.


justinsb:
Issue #10506

Issue #7931, opened 2019-11-15
Calcio > 3.8.x and CoreOS fail


sstarcher:
**1. What `kops` version are you running? kops 1.15.0.beta1
**2. What Kubernetes version are you running? 1.14.8

**3. What cloud provider are you using?** aws and CoreOS

kops update cluster
kops rolling-update cluster

Calico fails with
```
 Warning  FailedMount  30s (x8 over 94s)  kubelet, ip  MountVolume.SetUp failed for volume "flexvol-driver-host" : mkdir /usr/libexec/kubernetes: read-only file system
```

Workaround 
https://docs.projectcalico.org/v3.10/reference/faq#are-the-calico-manifests-compatible-with-coreos

Issue on calco side
https://github.com/projectcalico/calico/issues/2712

z1oy:
kops version
Version 1.15.0 (git-9992b4055)
kubernetesVersion: 1.15.6

OS: CoreOS

I have the same problem, too

mazzy89:
This PR I've worked on should solve the thing https://github.com/kubernetes/kops/pull/7545 but unfortunately it never got the right love

r00twayne:
also seeing this issue on an upgrade from k8s v1.14.9 -> v1.15.5 .  (kops 1.15.0)

AlexLast:
Also seeing this `k8s` `1.14.9` => `1.15.6` running `kops` `1.15.0`

hakman:
I think this was fixed by #8062.

phspagiari:
For AWS users, we added these lines on our cluster deployment script until the release of `v1.15.1`:

```
# After kops update
aws s3 cp s3://<STATE-STORE-BUCKET>/<CLUSTER-NAME>/addons/networking.projectcalico.org/k8s-1.12.yaml
sed -i 's#/usr/libexec/kubernetes/kubelet-plugins/volume/exec/nodeagent~uds#/var/lib/kubelet/volumeplugins/nodeagent~uds#g' k8s-1.12.yaml
aws s3 cp k8s-1.12.yaml s3://<STATE-STORE-BUCKET>/<CLUSTER-NAME>/addons/networking.projectcalico.org/k8s-1.12.yaml
# Before kops rolling update
```

ifosch:
I found the same issue when creating a cluster from scratch using kops 1.15.2, k8s 1.15.10 on AWS with latest CoreOS AMI:
``` 
[...]

PR #10554, opened 2021-01-10
Updates GCE channels to use ubuntu over COS

geojaz:
I didn't realize we were defaulting to COS on GCE- this changes the default for >=1.17.0. I discovered this while researching #9832. When I tried to use calico on COS on GCE, I ran into the same  filesystem mount issues we saw with coreos on https://github.com/kubernetes/kops/issues/7931. 
Changing to this image fixes everything and calico comes up properly :)


hakman:
Very nice. Ubuntu FTW!
/lgtm

PR #10555, opened 2021-01-11
GCE/Refactoring the scopes- cloud-platform ftw

geojaz:
I think we should update our node authz model for GCE to leverage CLoud IAM instead of the old-school scopes. Ie: https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances

This will probably need some iteration before it looks good.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,9 @@ test: ${BINDATA_TARGETS}  # Run tests locally
 test-windows: ${BINDATA_TARGETS}  # Run tests locally
 	go test -v $(go list ./... | grep -v /nodeup/)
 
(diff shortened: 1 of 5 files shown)


hakman:
@geojaz do you want to add a manual job to easily test on GCE when needed?

geojaz:
@hakman Yeah... I wanna think about this a bit more tomorrow :) zzz

rifelpet:
I think our kubetest2 setup should work for gce presubmits, it should just be a matter of duplicating the `pull-kops-e2e-kubernetes-aws-kubetest2` job in test-infra and adding a new Make target

PR #10556, opened 2021-01-11
update gophercloud dependency

zetaab:


go.mod
```diff
@@ -47,7 +47,7 @@ replace k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.20.0
 // This should match hack/go.mod
 replace k8s.io/code-generator => k8s.io/code-generator v0.20.0
 
-replace github.com/gophercloud/gophercloud => github.com/gophercloud/gophercloud v0.11.0
+replace github.com/gophercloud/gophercloud => github.com/gophercloud/gophercloud v0.15.0
```
olemarkus, line 50: Do we actually need this line? We are just using gophercloud for our own benefits so we may as well use latest version similar to what we do with the aws client.

PR #10557, opened 2021-01-11
[OpenStack] Use new hash format in instance names

zetaab:
So this will modify the instance name format that we will see in OpenStack instance names AND in Kubernetes cluster. The current problem that we have is that if we delete instance from OpenStack and run `kops update cluster --yes` - the new instance will come back with same name. However, the OpenStack instanceid is changed and it leads to problems (there is mismatch between Kubernetes API node instanceid and real OpenStack instanceid). The current way to fix this is to delete node from Kubernetes API and restart OpenStack instance.

old format: `<ig name>-<index>-<clustername>`
new format: `<ig name>-<6 chars random hash>`

tested with following cases:

### with new code

- create new cluster `calc.k8s.local`

```
% kubectl get nodes
NAME                   STATUS   ROLES    AGE     VERSION
master-zone-1-kivf21   Ready    master   2m16s   v1.19.5
master-zone-2-8hxz8y   Ready    master   2m29s   v1.19.5
master-zone-3-yv9sdw   Ready    master   2m25s   v1.19.5
nodes-zone-1-nm2gcp    Ready    node     67s     v1.19.5
nodes-zone-2-ryxop5    Ready    node     64s     v1.19.5
```

- run kubectl update cluster without changing cluster config (no changes, `kops update cluster --name calc.k8s.local`)
- update k8s version in cluster config and run update cluster (no changes, `kops update cluster --name calc.k8s.local`)
- run rolling-update (`kops rolling-update cluster calc.k8s.local --yes`)

```
% kubectl get nodes
NAME                   STATUS   ROLES    AGE     VERSION
master-zone-1-chvtmx   Ready    master   28m     v1.19.6
master-zone-2-n7zs4y   Ready    master   21m     v1.19.6
master-zone-3-wmskzc   Ready    master   11m     v1.19.6
nodes-zone-1-miqpky    Ready    node     6m23s   v1.19.6
nodes-zone-2-bujc69    Ready    node     48s     v1.19.6
```

- scale down instancegroup nodes from 1 -> 0 (should delete one node, `kops edit ig nodes-zone-2` && `kops update cluster --name calc.k8s.local`)

```
% kubectl get nodes
NAME                   STATUS   ROLES    AGE     VERSION
master-zone-1-chvtmx   Ready    master   32m     v1.19.6
master-zone-2-n7zs4y   Ready    master   26m     v1.19.6
master-zone-3-wmskzc   Ready    master   16m     v1.19.6
nodes-zone-1-miqpky    Ready    node     11m     v1.19.6
```

- delete cluster

### with old code create and then updating with newer code

- create new cluster with old naming format (using old code)

```
% kubectl get nodes
NAME                             STATUS   ROLES    AGE     VERSION
master-zone-1-1-calc-k8s-local   Ready    master   2m38s   v1.19.5
master-zone-2-1-calc-k8s-local   Ready    master   2m35s   v1.19.5
master-zone-3-1-calc-k8s-local   Ready    master   2m2s    v1.19.5
nodes-zone-1-1-calc-k8s-local    Ready    node     94s     v1.19.5
nodes-zone-2-1-calc-k8s-local    Ready    node     90s     v1.19.5
```

- compile new code and use it
- run kubectl update cluster without changing cluster config (no changes, `kops update cluster --name calc.k8s.local`)
- update k8s version in cluster config and run update cluster (no changes, `kops update cluster --name calc.k8s.local`)
- run rolling-update (`kops rolling-update cluster calc.k8s.local --yes`)

```
% kubectl get nodes
NAME                   STATUS   ROLES    AGE     VERSION
master-zone-1-orb9gv   Ready    master   31m     v1.19.6
master-zone-2-diw28n   Ready    master   18m     v1.19.6
master-zone-3-srsk15   Ready    master   12m     v1.19.6
nodes-zone-1-lhgo7i    Ready    node     6m51s   v1.19.6
nodes-zone-2-gyziid    Ready    node     2m12s   v1.19.6
```

- scale down instancegroup nodes from 1 -> 0 (should delete one node, `kops edit ig nodes-zone-2` && `kops update cluster --name calc.k8s.local`)

```
% kubectl get nodes
NAME                   STATUS   ROLES    AGE     VERSION
master-zone-1-orb9gv   Ready    master   33m     v1.19.6
master-zone-2-diw28n   Ready    master   20m     v1.19.6
master-zone-3-srsk15   Ready    master   14m     v1.19.6
nodes-zone-1-lhgo7i    Ready    node     8m35s   v1.19.6
```

- delete cluster

### with old code create and then scaling down with new code

- create new cluster with old naming format (using old code)

```
% kubectl get nodes
NAME                             STATUS   ROLES    AGE     VERSION
master-zone-1-1-calc-k8s-local   Ready    master   2m20s   v1.19.6
master-zone-2-1-calc-k8s-local   Ready    master   2m32s   v1.19.6
master-zone-3-1-calc-k8s-local   Ready    master   2m34s   v1.19.6
nodes-zone-1-1-calc-k8s-local    Ready    node     82s     v1.19.6
nodes-zone-2-1-calc-k8s-local    Ready    node     85s     v1.19.6
```

- compile new code and use it
- scale down instancegroup nodes from 1 -> 0 (should delete one node, `kops edit ig nodes-zone-2` && `kops update cluster --name calc.k8s.local`)

```
% kubectl get nodes
NAME                             STATUS   ROLES    AGE     VERSION
master-zone-1-1-calc-k8s-local   Ready    master   4m17s   v1.19.6
master-zone-2-1-calc-k8s-local   Ready    master   4m29s   v1.19.6
master-zone-3-1-calc-k8s-local   Ready    master   4m31s   v1.19.6
nodes-zone-1-1-calc-k8s-local    Ready    node     3m19s   v1.19.6
```

- delete cluster

this code is backward compatible with old clusters with old format

### DIFF
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -100,6 +100,8 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 		securityGroups = append(securityGroups, b.LinkToSecurityGroup(b.Cluster.Spec.MasterPublicName))
 	}
 
+	r := strings.NewReplacer("_", "-", ".", "-")
+	groupName := r.Replace(strings.ToLower(ig.Name))
 	// In the future, OpenStack will use Machine API to manage groups,
 	// for now create d.InstanceGroups.Spec.MinSize amount of servers
 	for i := int32(0); i < *ig.Spec.MinSize; i++ {
@@ -141,8 +143,14 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 		}
 		c.AddTask(portTask)
 
+		metaWithName := make(map[string]string)
+		for k, v := range igMeta {
+			metaWithName[k] = v
+		}
+		metaWithName[openstack.TagKopsName] = fi.StringValue(instanceName)
 		instanceTask := &openstacktasks.Instance{
 			Name:             instanceName,
+			NamePrefix:       s(groupName),
 			Region:           fi.String(b.Cluster.Spec.Subnets[0].Region),
 			Flavor:           fi.String(ig.Spec.MachineType),
 			Image:            fi.String(ig.Spec.Image),
@@ -151,7 +159,7 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 			Role:             fi.String(string(ig.Spec.Role)),
 			Port:             portTask,
 			UserData:         startupScript,
-			Metadata:         igMeta,
+			Metadata:         metaWithName,
 			SecurityGroups:   ig.Spec.AdditionalSecurityGroups,
 			AvailabilityZone: az,
 		}
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -150,7 +150,7 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 		metaWithName[openstack.TagKopsName] = fi.StringValue(instanceName)
 		instanceTask := &openstacktasks.Instance{
 			Name:             instanceName,
-			NamePrefix:       s(groupName),
+			GroupName:        s(groupName),
 			Region:           fi.String(b.Cluster.Spec.Subnets[0].Region),
 			Flavor:           fi.String(ig.Spec.MachineType),
 			Image:            fi.String(ig.Spec.Image),
--- a/pkg/resources/openstack/floatingip.go
+++ b/pkg/resources/openstack/floatingip.go
@@ -17,6 +17,7 @@ limitations under the License.
 package openstack
 
 import (
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	l3floatingip "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	"k8s.io/kops/pkg/resources"
 	"k8s.io/kops/upup/pkg/fi"
@@ -55,14 +56,14 @@ func (os *clusterDiscoveryOS) listL3FloatingIPs(routerID string) ([]*resources.R
 	return resourceTrackers, nil
 }
 
-func (os *clusterDiscoveryOS) listFloatingIPs(instanceID string) ([]*resources.Resource, error) {
+func (os *clusterDiscoveryOS) listFloatingIPs(instance servers.Server) ([]*resources.Resource, error) {
 	var resourceTrackers []*resources.Resource
-	instance, err := os.osCloud.GetInstance(instanceID)
-	if err != nil {
-		return resourceTrackers, err
+	name := instance.Name
+	if val, ok := instance.Metadata[openstack.TagKopsName]; ok {
+		name = val
 	}
 	floatingIPs, err := os.osCloud.ListL3FloatingIPs(l3floatingip.ListOpts{
-		Description: "fip-" + instance.Name,
+		Description: "fip-" + name,
 	})
 	if err != nil {
 		return resourceTrackers, err
--- a/pkg/resources/openstack/floatingip.go
+++ b/pkg/resources/openstack/floatingip.go
@@ -59,6 +59,7 @@ func (os *clusterDiscoveryOS) listL3FloatingIPs(routerID string) ([]*resources.R
 func (os *clusterDiscoveryOS) listFloatingIPs(instance servers.Server) ([]*resources.Resource, error) {
 	var resourceTrackers []*resources.Resource
 	name := instance.Name
+	// this is needed for backwards compatbility issues with new and old instance name format
 	if val, ok := instance.Metadata[openstack.TagKopsName]; ok {
 		name = val
 	}
--- a/pkg/resources/openstack/floatingip.go
+++ b/pkg/resources/openstack/floatingip.go
@@ -58,8 +58,9 @@ func (os *clusterDiscoveryOS) listL3FloatingIPs(routerID string) ([]*resources.R
 
 func (os *clusterDiscoveryOS) listFloatingIPs(instance servers.Server) ([]*resources.Resource, error) {
 	var resourceTrackers []*resources.Resource
+	// we can find real instance name from instance name in old format
+	// however, in new format the real name can be found in metadata
 	name := instance.Name
-	// this is needed for backwards compatbility issues with new and old instance name format
 	if val, ok := instance.Metadata[openstack.TagKopsName]; ok {
 		name = val
 	}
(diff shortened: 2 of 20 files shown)


zetaab:
seems that tests will fail, I will fix tests tomorrow

zetaab:
@olemarkus @rifelpet could you guys help me here, why tests are failing? I cannot find the reason for that, some fields missing from dummy data?

zetaab:
I do not understand why its trying to create two instances with same name? If I add debug https://github.com/kubernetes/kops/blob/master/cloudmock/openstack/mockcompute/servers.go#L193 I can see that it will create two nodes always with same name!?

olemarkus:
Does that happen prior to this PR? I tried to duplicate that on master, but doesn't seem to happen there.

olemarkus:
Just wondering if `Find` actually can find anything using the mock. The new behaviour expects a tag to be set, but we don't set metadata in the mock.

zetaab:
hmm the metadata is set here https://github.com/kubernetes/kops/blob/master/cloudmock/openstack/mockcompute/servers.go#L199 ?

pkg/model/openstackmodel/servergroup.go
```diff
@@ -141,8 +143,11 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 		}
 		c.AddTask(portTask)
 
+		igMeta[openstack.TagKopsName] = fi.StringValue(instanceName)
```
olemarkus, line 146: The issue is probably this one. igMeta is defined outside of the loop, so we are setting the same metadata on all instances.

zetaab: yeah it seems that if we have more than 1 instance in same instancegroup, the instancename will be same in all of them

zetaab: but the weird thing is that it works when I execute this against openstack.. blah

zetaab: % kubectl get nodes
NAME                   STATUS   ROLES    AGE     VERSION
master-zone-1-prb0qx   Ready    master   7m43s   v1.19.6
nodes-zone-1-epkztj    Ready    node     6m29s   v1.19.6
nodes-zone-1-ldgep8    Ready    node     6m26s   v1.19.6

nodes-zone-1-epkztj 
metadata
KopsName nodes-zone-1-2-calc-k8s-local

nodes-zone-1-ldgep8 
metadata
KopsName nodes-zone-1-1-calc-k8s-local

olemarkus: Not sure. Probably some serialisation/deserialisation that does not happen outside of tests.

zetaab:
@olemarkus now it works! What is your opinion about this?

Ping @mitch000001

olemarkus:
I think this looks good.

/lgtm

pkg/model/openstackmodel/servergroup.go
```diff
@@ -141,8 +143,14 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
...
 		c.AddTask(portTask)
 
+		metaWithName := make(map[string]string)
+		for k, v := range igMeta {
+			metaWithName[k] = v
+		}
+		metaWithName[openstack.TagKopsName] = fi.StringValue(instanceName)
 		instanceTask := &openstacktasks.Instance{
 			Name:             instanceName,
+			NamePrefix:       s(groupName),
```
mitch000001, line 153: If the `NamePrefix` is always the group name, why not changing the field to be called `GroupName`?

pkg/resources/openstack/floatingip.go
```diff
@@ -55,14 +56,14 @@ func (os *clusterDiscoveryOS) listL3FloatingIPs(routerID string) ([]*resources.R
...
 }
 
-func (os *clusterDiscoveryOS) listFloatingIPs(instanceID string) ([]*resources.Resource, error) {
+func (os *clusterDiscoveryOS) listFloatingIPs(instance servers.Server) ([]*resources.Resource, error) {
 	var resourceTrackers []*resources.Resource
-	instance, err := os.osCloud.GetInstance(instanceID)
-	if err != nil {
-		return resourceTrackers, err
+	name := instance.Name
+	if val, ok := instance.Metadata[openstack.TagKopsName]; ok {
```
mitch000001, line 62: If this is also used for backward compatibility, can you add a comment reflecting this?

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -36,6 +37,7 @@ import (
 type Instance struct {
 	ID               *string
 	Name             *string
+	NamePrefix       *string
```
mitch000001, line 40: I'd also reflect here that the `NamePrefix` matches the InstanceGroup name by renaming the field to `GroupName`

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -235,10 +257,30 @@ func (_ *Instance) ShouldCreate(a, e, changes *Instance) (bool, error) {
 	return false, nil
 }
 
+// makeServerName generates name for the instance
+// the instance format is [namePrefix]-[6 character hash]
+func makeServerName(e *Instance) (string, error) {
```
mitch000001, line 262: I'd prefer `generateInstanceName`, but that's more a matter of taste.

mitch000001 (commented):
Overall looks good, just some minor naming issues :)

olemarkus:
/hold

In case you want to implement the requested changes

mitch000001:
looks good to me

zetaab:
@olemarkus can you recheck (and remove hold if ok)

mitch000001:
@zetaab is that a change we want to also cherry-pick for 1.19?

zetaab:
@mitch000001 no idea, do we want it already in 1.19? I am fine with that if someone wants it

olemarkus:
/hold cancel
/lgtm

I think this needs to bake a while, so a bit too late for 1.19.

kciredor:
Missed this one, but while upgrading to kops 1.20 from 1.19 on OpenStack I found out ;-) I understand this solves a problem, but it also means that the server name does not include the cluster name anymore. Meaning a quick `openstack server list` does not show which vm belongs to which cluster (It's not uncommon to have multiple clusters per OpenStack project). Ofcourse it's easy to display vm's in more detail, but still. What if introducing the hash would not imply dropping the cluster name?

mitch000001:
I think the introduced change does more than just adding a hash to the name. The prior problem was that the server name also contained the order in the instance group, e.g. if the instance group has 3 members (size 3) then the first node was called `ig-name-1-cluster-name`, the second `ig-name-2-cluster-name` and so on. Secondly the hostnames may only be 64 characters normally, so if your cluster name was too long you also ran into problems.
If you "just" want to have the cluster name in place to make it easier to find servers, I mean, I don't want to just point out `grep`ing for it (as we do) but I am not convinced that this outweighs the solved problems with this PR.
Having said that, this also is a preparation on enabling rolling update features later on like detaching nodes from the management of kOps (surging, +1/-1) which is only possible by stripping the instance group number because otherwise we can't easily create a replacement node because names could collide.

kciredor:
Thanks @mitch000001 this makes a lot of sense now!

PR #10559, opened 2021-01-12
kubetest2 - Add manifest template support

rifelpet:
This adds support for defining manifest templates used in e2e tests, allowing us to test more specific configurations not exposed through `create cluster` cli flags. I'm envision using this for testing ACM certificates on the API NLB, additional cilium configurations, mixed instances policies, etc.

I'm updating the test-e2e make target as I go to make the presubmit job use it but soon I'll work on how we'll actually configure and invoke kubetest2 from our variety of jobs.

Specifically how we'll access these kubetest2 binaries and other static files like the upgrade scenario script or these template files, including using the appropriate kubetest2-kops version for testing on release branches.

1. Have the periodic jobs include [extra_refs](https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md#writing-a-prowjob-that-uses-pod-utilities) to clone the kops repo and use its Makefile, kubetest2-kops, etc. The release branch jobs could clone the appropriate kops release branch. 
2. Build a new image that contains all the appropriate dependencies. It would be rebuilt on postsubmit and have separate tags per release branch.
3. Upload the dependencies as a part of `make gcs-publish-ci`. This would mean the logic to download a specific version of kops would need to live outside of kubetest2 in order to also download kubetest2. 

I'm leaning towards 1 because it would be simplest and easiest. There are a few concerns about version mismatch, particularly during the [updown pipeline](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kops/kops-pipeline.yaml) and the various version markers. For a new merge to master, the periodic jobs would immediately use the new commit's version of kubetest2-kops but wouldn't use the kops version from that commit until its gone through the pipeline. I don't think this is a big deal though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,6 +198,7 @@ test-e2e:
 		--cloud-provider=aws \
 		--kops-binary-path=/home/prow/go/src/k8s.io/kops/bazel-bin/cmd/kops/linux-amd64/kops \
 		--kubernetes-version=v1.19.4 \
+		--template-path=tests/e2e/templates/simple.yaml.tmpl \
 		--test=kops \
 		-- \
 		--test-package-version=v1.19.4 \
--- a/tests/e2e/go.mod
+++ b/tests/e2e/go.mod
@@ -5,6 +5,7 @@ go 1.15
 require (
 	github.com/octago/sflags v0.2.0
 	github.com/spf13/pflag v1.0.5
+	gopkg.in/yaml.v2 v2.3.0
 	k8s.io/klog/v2 v2.4.0
 	sigs.k8s.io/kubetest2 v0.0.0-20201130212850-d9dad7c8699c
 )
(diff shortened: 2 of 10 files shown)


olemarkus:
1 sounds good to me.

In this case, do you want to use `kops create -f` instead of `replace`? The former is a bit more convenient wrt keys and kubeconfig.

`replace` could be useful if we want to spin up a cluster, then apply a new spec and validate again. E.g if we want to test an addon upgrade/migration.

rifelpet:
good idea. I added the SSHCredential to the template rather than `kops create secret` hoping that will help.

/test pull-kops-e2e-kubernetes-aws-kubetest2

rifelpet:
This is (finally) ready for review 👍🏻

olemarkus:
This looks good!

/lgtm

PR #10564, opened 2021-01-12
Make cluster proportional autoscaler image configurable.

bjhaid:
Similar to how we can configure coreDNS image we will like to configure
cluster Proportional autoscaler so we can use our internal docker
registry rather than gcr.io.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -1867,6 +1867,10 @@ spec:
                     description: CoreDNSImage is used to override the default image
                       used for CoreDNS
                     type: string
+                  cpaImage:
+                    description: CPAImage is used to override the default image used
+                      for Cluster Proportional Autoscaler
+                    type: string
                   cpuRequest:
                     anyOf:
                     - type: integer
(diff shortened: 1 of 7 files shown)


k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -1867,6 +1867,10 @@ spec:
                     description: CoreDNSImage is used to override the default image
                       used for CoreDNS
                     type: string
+                  cpaImage:
```
joshbranham, line 1870: I'm not sure this image override field should live under the `kubeDNS` spec

bjhaid: the image is only used by coredns and kubedns... which is why I kept it there.

bjhaid: I am open to putting it elsewhere... under which object would you suggest I put it?

joshbranham: Yeah I guess I was thinking since its more a general purpose autoscaler than _just_ DNS but if its only used for DNS out of the box in a new cluster then seems fine for it be here!

bjhaid:
@olemarkus @joshbranham does this change look good enough to be approved? is either of you an approver? if not can you please help tag someone who can approve, thanks!

bjhaid:
@olemarkus thanks for approving! what do I need to do to get this backported?

olemarkus:
You need to create a backport issue and then we'll consider if it makes sense. See here https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/cherry-picks.md

PR #10567, opened 2021-01-13
Fix NLB listener -> target group association for TF & CF

### LINKED ISSUE #10562: Terraform bug with API load balancer with NLB and sslCertificate
 this information.**

Version 1.19.0-beta.3 (git-e43f1cc6e3c77d093935d1706042861095d75eb7)

 a `kops` flag.**

Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:32:58Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}

AWS

I configured the API load balancer with class: Network and sslCertificate. When I let kops generate Terraform from this manifest and apply it, Terraform eventually quits with:

Error: error creating ELBv2 Listener: IncompatibleProtocols: The listener and the following target groups have incompatible protocols: arn:aws:elasticloadbalancing:XXX:XXX:targetgroup/tls-XXX/XXX
        status code: 400, request id: XXX.

Terraform cannot be successfully applied. See error above.

The listeners and the target groups that are instantiated should be compatible and Terraform can be applied successfully.

```yaml
kind: Cluster
metadata:
  name: mycluster
spec:
  api:
    loadBalancer:
      type: Internal
      sslCertificate: arn:aws:acm:XXX:XXX:certificate/XXX
      class: Network
```

The connection between the two listeners and the two target groups seem to be mixed up. When I manually connect the TLS listener with port 443 to the "tls-" target group and the TCP listener with port 8443 to the "tcp-" target group it seems to work. The generated terraform code is the other way around:

```terraform
resource "aws_lb_listener" "api-XXXXX-443" {
  certificate_arn = "arn:aws:acm:XXX:XXX:certificate/XXX"
  default_action {
    target_group_arn = aws_lb_target_group.tcp-XXX.id
    type             = "forward"
  }
  load_balancer_arn = aws_lb.api-XXX.id
  port              = 443
  protocol          = "TLS"
}

resource "aws_lb_listener" "api-XXX-8443" {
  default_action {
    target_group_arn = aws_lb_target_group.tls-XXX.id
    type             = "forward"
  }
  load_balancer_arn = aws_lb.api-XXX.id
  port              = 8443
  protocol          = "TCP"
}```

rifelpet:
The old code made the incorrect assumption that the NLB's list of TargetGroup tasks is in the same order as the NLB's list of listeners for their associations.
Because the model adds them in opposite orders this resulted in the TLS listener being forwarded to the TCP TG and vice versa.

This updates the terraform and cloudformation generation code to search the NLB's list of target groups by name for the target group that should be associated with the listener.

This matches the logic used in the "direct" target:
https://github.com/kubernetes/kops/blob/a140168c70de521939399f19fb0bdec7dd51ffa4/upup/pkg/fi/cloudup/awstasks/network_load_balancer.go#L87-L95

fixes #10562

### DIFF
--- a/tests/integration/update_cluster/complex/cloudformation.json
+++ b/tests/integration/update_cluster/complex/cloudformation.json
@@ -1206,7 +1206,7 @@
           {
             "Type": "forward",
             "TargetGroupArn": {
-              "Ref": "AWSElasticLoadBalancingV2TargetGrouptcpcomplexexamplecomvpjolq"
+              "Ref": "AWSElasticLoadBalancingV2TargetGrouptlscomplexexamplecom5nursn"
             }
           }
         ],
@@ -1225,7 +1225,7 @@
           {
             "Type": "forward",
             "TargetGroupArn": {
-              "Ref": "AWSElasticLoadBalancingV2TargetGrouptlscomplexexamplecom5nursn"
+              "Ref": "AWSElasticLoadBalancingV2TargetGrouptcpcomplexexamplecomvpjolq"
             }
           }
         ],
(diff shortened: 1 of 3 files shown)


hakman:
Looks legit.
/lgtm

PR #10572, opened 2021-01-13
Allow setting gp3 for terraform volumes without setting throughput

hakman:
Until https://github.com/hashicorp/terraform-provider-aws/pull/16517 is merged, TF will not be able to use the `throughput` arg for `aws_ebs_volume`.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/ebsvolume.go
+++ b/upup/pkg/fi/cloudup/awstasks/ebsvolume.go
@@ -198,12 +198,18 @@ type terraformVolume struct {
 }
 
 func (_ *EBSVolume) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *EBSVolume) error {
+	// TODO: Remove when Terraform gets support for "throughput" with "aws_ebs_volume"
+	// https://github.com/hashicorp/terraform-provider-aws/pull/16517
+	throughput := e.VolumeThroughput
+	if fi.Int64Value(e.VolumeThroughput) <= 125 {
+		throughput = nil
+	}
 	tf := &terraformVolume{
 		AvailabilityZone: e.AvailabilityZone,
 		Size:             e.SizeGB,
 		Type:             e.VolumeType,
 		Iops:             e.VolumeIops,
-		Throughput:       e.VolumeThroughput,
+		Throughput:       throughput,
 		KmsKeyId:         e.KmsKeyId,
 		Encrypted:        e.Encrypted,
 		Tags:             e.Tags,


upup/pkg/fi/cloudup/awstasks/ebsvolume.go
```diff
@@ -198,12 +198,18 @@ type terraformVolume struct {
 }
 
 func (_ *EBSVolume) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *EBSVolume) error {
+	// TODO: Remove when Terraform gets support for "throughput" with "aws_ebs_volume"
+	// https://github.com/hashicorp/terraform-provider-aws/pull/16517
+	throughput := e.VolumeThroughput
+	if fi.Int64Value(e.VolumeThroughput) <= 125 {
```
rifelpet, line 204: why are we only setting it to nil with this condition? if the throughput field isn't yet supported at all in terraform then I'd think we should always set it to nil?

rifelpet: also we should have an integration test that generates this terraform output so that the verify-terraform job will catch it.

hakman: It is set to nil if you don't use gp3 volumes. We have integration test in master https://github.com/kubernetes/kops/pull/10569, but not sure how we can "catch" it.

rifelpet: looking at the [api docs](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#CreateVolumeInput)

```
    // The throughput to provision for a volume, with a maximum of 1,000 MiB/s.
    //
    // This parameter is valid only for gp3 volumes.
    //
    // Valid Range: Minimum value of 125. Maximum value of 1000.
    Throughput *int64 `type:"integer"`
```

I feel like we should fail api validation if throughput is < 125 rather than ignoring the value. We could also fail validation if RootVolumeThroughput is set for RootVolumeTypes other than gp3 (and if cloudprovider != aws )

This is the only throughput validation we have currently:

https://github.com/kubernetes/kops/blob/0410058bb0886cce28b2f45606b3dfd0830310ca/pkg/apis/kops/validation/instancegroup.go#L72-L74

hakman: We could, but it's not really this hack's purpose. I just want to not put the value in the TF template if the value is the minimum to not need a new kOps release when TF decides to merge the throughput arg. We set values lower than the minimum to minimum in model.

PR #10575, opened 2021-01-13
[DigitalOcean] add e2e tests

srikiz:
Initial work for adding e2e tests for DO.

Tested locally and it seems to work fine after adding the below command arguments.
` go run kubetest2-kops/main.go --up --cloud-provider digitalocean --cluster-name test.myclustername.com \
        --env KOPS_STATE_STORE=do://***-kops-space \
	--env DIGITALOCEAN_ACCESS_TOKEN=bd5f*****472dae3 \
	--env S3_ENDPOINT=sfo2.digitaloceanspaces.com \
	--env S3_ACCESS_KEY_ID=PPO******NSRW \
	--env S3_SECRET_ACCESS_KEY=rm******oIc \
	--kops-binary-path /home/kops \
	--ssh-public-key ~/.ssh/id_rsa.pub \
	--kubernetes-version 1.19.2`

I'll separately work with @timoreimann and @rifelpet to include the necessary environment variables.

### DIFF
--- a/tests/e2e/kubetest2-kops/do/zones.go
+++ b/tests/e2e/kubetest2-kops/do/zones.go
@@ -34,7 +34,7 @@ var allZones = []string{
 var ErrNoEligibleRegion = errors.New("No eligible DO region found with enough zones")
 
 // ErrMoreThanOneZone indicates the requested number of zones was more than one
-var ErrMoreThanOneZone = errors.New("More than 1 zone is choosen. DO only works with 1 zone")
+var ErrMoreThanOneZone = errors.New("More than 1 zone is chosen. DO only works with 1 zone")
 
 // RandomZones returns a random set of availability zones within a region
 func RandomZones(count int) ([]string, error) {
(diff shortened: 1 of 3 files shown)


tests/e2e/kubetest2-kops/do/zones.go
```diff
@@ -0,0 +1,53 @@
...
+	"sfo1",
+	"sfo2",
+	"tor1",
+}
+
+// ErrNoEligibleRegion indicates the requested number of zones is not available in any region
+var ErrNoEligibleRegion = errors.New("No eligible DO region found with enough zones")
+
+// ErrMoreThanOneZone indicates the requested number of zones was more than one
+var ErrMoreThanOneZone = errors.New("More than 1 zone is choosen. DO only works with 1 zone")
```
rifelpet, line 37: this should fix the verify job:
```suggestion
var ErrMoreThanOneZone = errors.New("More than 1 zone is chosen. DO only works with 1 zone")
```

rifelpet (commented):
nice work! glad to see it isn't much effort here to add support
/ok-to-test

timoreimann:
Do we have a DO test account set up with kops already?

I know there's one that Cluster API uses which is good for sharing with other CNCF projects, but wasn't sure if that has happened already (or may be happening implicitly).

PR #10550, opened 2021-01-10
Upgrade Go v1.15.6 / Bazel v3.4.1

bmelbourne:
**What this PR does / why we need it**:

- Upgrade Go v1.15.5 -> v1.15.6
- Upgrade Bazel v2.2.0 -> v3.4.1
- Upgrade Bazel Rules v0.24.7 -> v0.25.1
- Upgrade Bazel Gazelle v0.19.1/v0.22.2 -> v0.22.3

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

PR #10578, opened 2021-01-14
Use Bazel 3.4.1 for postsubmit jobs

hakman:
Presubmits jobs are failing because of missing Bazel 3.4.1 from #10550:
https://testgrid.k8s.io/kops-misc#kops-postsubmit-push-to-staging

There is no `bazelbuild` image for that, so experimenting with `kubekins-e2e` image that is used in presubmit jobs.

/cc @olemarkus @rifelpet

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -5,7 +5,7 @@ options:
   machineType: 'N1_HIGHCPU_8'
 steps:
 # Push the images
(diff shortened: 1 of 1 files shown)


hakman:
I think we have a winner, but we still should check with @justinsb if there's a more appropriate image.
https://prow.k8s.io/log?container=test&id=1349642210504609792&job=kops-postsubmit-push-to-staging

PR #10579, opened 2021-01-14
Various cleanups around apply_cluster and awsmodel

olemarkus:
A lot of the AWS specific builders exist in the generic `model` package. Most of them have now been moved to `awsmodel`.
Some are left behind because they are (ab)used elsewhere in kOps.

Also merged spotinst builder into `awsmodel` because spotinst isn't a true cloud provider and there are cyclic dependencies between the two.

Also merged the two large switch statements.

/kind cleanup


hakman:
As discussed during Office Hours, Justin will give the final approval on this.
/assign @justinsb

hakman (approved):

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -67,7 +66,7 @@ func (b *AutoscalingGroupModelBuilder) Build(c *fi.ModelBuilderContext) error {
 		name := b.AutoscalingGroupName(ig)
 
 		if featureflag.SpotinstHybrid.Enabled() {
-			if spotinstmodel.HybridInstanceGroup(ig) {
+			if HybridInstanceGroup(ig) {
```
justinsb, line 69: We might want to call this `SpotInstHybridInstanceGroup` if it's spotinst specific

pkg/model/awsmodel/bastion.go
```diff
@@ -270,11 +271,11 @@ func (b *BastionModelBuilder) Build(c *fi.ModelBuilderContext) error {
 		// Here we implement the bastion CNAME logic
 		// By default bastions will create a CNAME that follows the `bastion-$clustername` formula
 		t := &awstasks.DNSName{
-			Name:      s(bastionPublicName),
+			Name:      fi.String(bastionPublicName),
```
justinsb, line 274: I think I agree with this change (it's more verbose but clearer _not_ to use the helper functions), but it might have been easier to do this as a follow-on PR or separate commit, just from a change readability perspective.

justinsb:
One suggestion to rename a spotinst-specific method ("HybridInstanceGroup") but non-blocking and this is good

/approve
/lgtm

pkg/model/awsmodel/bastion.go
olemarkus: Yeah. This change was mainly done because `awstasks` does not have these helper functions, and I didn't see the point of copying them.

Issue #10122, opened 2020-10-27
Cluster cannot start with older versions of k8s


hakman:
After merging https://github.com/kubernetes/kops/pull/10048, the new way of installing Docker is conflicting with the pre-installed Docker in the `kope.io` images.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-aws-k8s-1-15/1320685447990480896

A similar problem existed in 1.17 and was fixed in https://github.com/kubernetes-sigs/image-builder/pull/273, by just copying the Docker deb files to `/var/cache/nodeup/packages` and letting kOps install them if the hash matches.

All `kope.io` images from 1.11 to 1.16, maybe also 1.17.

rifelpet:
This issue is impacting kubernetes 1.17 as well: https://testgrid.k8s.io/kops-versions#kops-aws-k8s-1.17

https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-1-17/1344395270774853632/artifacts/18.193.88.194/protokube.log

```
Starting Kubernetes Protokube Service...
Started Kubernetes Protokube Service.
ctr: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"/var/run/dbus\\\" to rootfs \\\"/run/containerd/io.containerd.runtime.v2.task/k8s.io/protokube/rootfs\\\" at \\\"/var/run/dbus\\\" caused \\\"stat /var/run/dbus: no such file or directory\\\"\"": unknown
protokube.service: Main process exited, code=exited, status=1/FAILURE
```

hakman:
Yes, dbus should be installed on those images.
Ref: https://github.com/kubernetes-sigs/image-builder/issues/130

atam003:
I am using kops1.19 +Kubernetes 1.16. I tried using image latest available kops image "kope.io/k8s-1.16-debian-stretch-amd64-hvm-ebs-2020-07-20". I see this issue.
Is there any latest published image of kops which can be used?

hakman:
@atam003 until 1.19 is released, you can use the official Debian 9 [image](https://kops.sigs.k8s.io/operations/images/#debian-9-stretch) or the Ubuntu 20.04 [image](https://kops.sigs.k8s.io/operations/images/#ubuntu-2004-focal).

sonowz:
Is there any workaround with this issue?
[...]

PR #10581, opened 2021-01-14
Allow nodeup (and others) to replace in-use files

justinsb:
By creating a tempfile and then moving the file into place, we both
write more atomically and we can overwrite in-use files.

Issue #10122


justinsb:
This does seem to workaround #10122 .  I'm still going to update the images, but ... we can tolerate images that have different docker versions!

PR #10582, opened 2021-01-14
Dial-down logging on flagbuilder

justinsb:
We still run nodeup at V=8, so the distinction between V=10 and V=8
matters.

This message really is just development logging, so downgrade to V=10.


zetaab (approved):
/lgtm

PR #10583, opened 2021-01-14
Install dbus if needed for protokube with containerd

justinsb:
Issue #10122

### DIFF
--- a/nodeup/pkg/model/protokube.go
+++ b/nodeup/pkg/model/protokube.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
 	"k8s.io/kops/util/pkg/architectures"
+	"k8s.io/kops/util/pkg/distributions"
 	"k8s.io/kops/util/pkg/proxy"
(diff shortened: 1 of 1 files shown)


justinsb:
Marking as WIP because I'm not entirely sure when we actually need the dbus mount; it might suffice only to use it when we're not using etcd-manager (and we should probably try so that we can remove the dbus dependency)

nodeup/pkg/model/protokube.go
```diff
@@ -101,6 +101,15 @@ func (t *ProtokubeBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 	c.AddTask(service)
 
+	if t.Cluster.Spec.ContainerRuntime == "containerd" {
```
hakman, line 104: This is not related just to containerd, but in general, because it is needed by Systemd to work properly (at least that is how I remember from testing a while back).

hakman: At most, the condition should be to install `dbus` if distro is Kopeio.

rifelpet:
note that this should fix the legacy etcd prow job as well: https://testgrid.k8s.io/kops-misc#kops-aws-legacy-etcd

PR #10584, opened 2021-01-14
Fix default make target

rifelpet:
Adding the e2e.mk file changed the default target for `make`

This reverts the behavior back to the original `kops-install`


zetaab (approved):
/lgtm

PR #10586, opened 2021-01-15
WIP: containerd: Support for kubenet-style networking modes

justinsb:
By configuration conf_template, we can continue to support these
networking modes.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -40,7 +40,6 @@ import (
 	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/pkg/kubeconfig"
 	"k8s.io/kops/pkg/kubemanifest"
-	"k8s.io/kops/pkg/model/components"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup"
 	"k8s.io/kops/upup/pkg/fi/utils"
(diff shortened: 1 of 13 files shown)


hakman:
@justinsb I removed something similar in https://github.com/kubernetes/kops/commit/c02e5a20ea248ccb250a564c42013bb07e4d9fff.
The problem with it isn't that it wouldn't work at all, but that is just unsupported and should not be used as default or in production.

justinsb:
@hakman that's surprising; this is how all the CNI plugins work :-)  Do you have a link to what's deprecated?

I can imagine that maybe conf_template is deprecated, in which case each provider (e.g. kopeio networking) might have to implement this behaviour.

We can also change the default to a different CNI provider, though then we'd have to agree what the new default should be, which will be tricky...

justinsb:
Ah - I bet the confusion is passing `--networking=kubenet` to kubelet, vs using CNI in a kubenet-style.  By kubenet-style I mean that nodes are configured externally with routes (either by the cloudprovider or by a CNI provider), and that we don't need to do anything per-pod.

Maybe we should come up with a better name for "kubenet-style" to avoid the confusion!

The reason I pursued this was because otherwise we have to figure out a migration strategy for clusters using kubenet/kopeio/gce-networking.

hakman:
Yes, I meant the template from containerd conf and what it does behind the scenes.
Let's make this the main topic of today's call

hakman:
@justinsb I added https://github.com/kubernetes/kops/pull/10651 back to the master branch and cherry-picked for 1.19.
I see kopeio networking tests working again:

justinsb:
Thanks for fixing @hakman - this is now unneeded

/close

PR #10591, opened 2021-01-15
Release 1.20.0-alpha.1

hakman:
First alpha release of not so many 😃


rifelpet:
I do think there is value in having integration tests for older k8s versions to ensure we don't change behavior on older clusters unexpectedly. The test case names should probably be named after their k8s version (`tests/integration/update_cluster/k8s-1-17` for example) to make that more clear. This would all be in a separate PR though 🙂 

/lgtm

hakman:
I agree @rifelpet. Just wanted to make it easier to migrate to containerd.

Issue #7976, opened 2019-11-21
similiar behavior for the kops set command but for instancegroups


cpanato:
Hello

I was looking for a way to do a similar thing that we use for `kops set cluster....` command but to set resources in the instance group.
the edit is ok when I want to edit manually, I was looking to add some automation and would like to change some items in the instancegroup files

is this something the kops team has in mind as well? I can help to develop this 

thanks!

PR #10593, opened 2021-01-15
Add `set instancegroup` command

gabrieljackson:
This change adds a new command and functionality for updating
instance group configuration via command line arguments. This
behavior mimics the `set cluster` command.

This is a follow-up to https://github.com/kubernetes/kops/pull/8531 and addresses https://github.com/kubernetes/kops/issues/7976 

This implements the original behavior using the new reflect-based
logic. I opted to open a new PR as the original was outdated and
figured this would help with reviews.


gabrieljackson:
Hey there, just checking in on this to see if anything else is needed from me to begin reviewing. Let me know if that is the case. Thanks!

cmd/kops/set.go
```diff
@@ -33,7 +33,8 @@ var (
 
 	setExample = templates.Examples(i18n.T(`
     # Set cluster to run kubernetes version 1.17.0
-    kops set cluster k8s-cluster.example.com spec.kubernetesVersion=1.17.0
+	kops set cluster k8s-cluster.example.com spec.kubernetesVersion=1.17.0
+	kops set instancegroup k8s-cluster.example.com spec.maxSize=4
```
rifelpet, line 37: this will require the cli docs to be regenerated:

```suggestion
	kops set instancegroup --name k8s-cluster.example.com nodes spec.maxSize=4
```

pkg/commands/helpers_readwrite.go
```diff
@@ -67,6 +67,38 @@ func UpdateCluster(ctx context.Context, clientset simple.Clientset, cluster *kop
...
+	err = cloudup.PerformAssignments(cluster, cloud)
+	if err != nil {
+		return fmt.Errorf("error populating configuration: %v", err)
+	}
+
+	assetBuilder := assets.NewAssetBuilder(cluster, "")
+	fullCluster, err := cloudup.PopulateClusterSpec(clientset, cluster, cloud, assetBuilder)
+	if err != nil {
+		return err
+	}
```
rifelpet, line 86: I _think_ these can be removed because they're only operating on the Cluster object rather than the InstanceGroup we're modifying.

gabrieljackson: I am pretty sure I need to keep this if I am going to do `DeepValidate` in the following lines. If this isn't the correct method of checking the resources though I can change it, but I didn't see another similar validation method for just instancegroups.

cmd/kops/set.go
gabrieljackson: Good catch, thanks. I had this changed locally, but missed including it in the branch. Should be fixed now.

pkg/commands/helpers_readwrite.go
rifelpet: Ah you're right, I missed that. thanks! I think in this case we can follow what `kops edit instancegroup` does which is largely the same:

https://github.com/kubernetes/kops/blob/97c40a3c5dd6edc952a69a307f10155391c8417d/cmd/kops/edit_instancegroup.go#L167-L183

but it only validates with `CrossValidateInstanceGroup` rather than `DeepValidate`

gabrieljackson: Thanks for the pointer. I modified the validation call to use the same one as you referenced.

rifelpet:
Looks good, thanks!

/lgtm
/approve

PR #10595, opened 2021-01-16
Removed default hostname-override

### LINKED ISSUE #10590: Possibility to disable hostname-override in kubelet
According to the CIS benchmark kubelet shouldn't have the hostname-override flag. So it will be great to have the possibility to disable hostname-override.

Ref #6150

DOboznyi:
There's no sense in hostname-override because it's not working when cloud-provider flag set.

Fix #10590

Signed-off-by: Dmytro Oboznyi <dmytro.oboznyi@syncier.com>


DOboznyi:
@olemarkus why such tests failing? How can I fix them?

olemarkus:
Run `./hack/update-expected`

DOboznyi:
./hack/update-expected

DOboznyi:
/hack/update-expected

olemarkus:
It's a script found in the git repos

DOboznyi:
@olemarkus seems as without hostname override resolving not working well...

DOboznyi:
@olemarkus the problem that from AWS metadata kubelet receive FQDN, but on node it gets hostname from `/proc/sys/kernel/hostname` ([Link](https://github.com/kubernetes/kubernetes/blob/cb8ad64243d48d9a3c26b11b2e0945c098457282/pkg/util/node/node.go#L64)). So can it be fixed on kops side? Or only in kubelet code? Also, I wonder how it's working on our env. Which OS used in the test environment with e2e tests?

pkg/model/components/kubelet.go
```diff
@@ -207,22 +184,11 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
 	if clusterSpec.Kubelet.FeatureGates == nil {
 		clusterSpec.Kubelet.FeatureGates = make(map[string]string)
 	}
-
-	if clusterSpec.CloudConfig != nil && clusterSpec.CloudConfig.AWSEBSCSIDriver != nil && fi.BoolValue(clusterSpec.CloudConfig.AWSEBSCSIDriver.Enabled) {
-		if _, found := clusterSpec.Kubelet.FeatureGates["CSIMigrationAWSComplete"]; !found {
-			clusterSpec.Kubelet.FeatureGates["CSIMigrationAWSComplete"] = "true"
-		}
-	}
```
hakman, line 215: I think this was removed by mistake.

pkg/model/components/kubelet.go
```diff
@@ -207,22 +184,11 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
...
 		if b.IsKubernetesLT("1.16") {
 			clusterSpec.Kubelet.FeatureGates["ExperimentalCriticalPodAnnotation"] = "true"
 		}
 	}
 
-	// Set systemd as the default cgroup driver for kubelet from k8s 1.20
-	if b.IsKubernetesGTE("1.20") && clusterSpec.Kubelet.CgroupDriver == "" {
-		clusterSpec.Kubelet.CgroupDriver = "systemd"
-	}
-
```
hakman, line 226: I think this was removed by mistake also.

olemarkus:
Most tests are running against Ubuntu 20.04.

In itself, it shouldn't be a problem that hostname is used instead of FQDN. The hostnames also resolve. But this may be unexpected to some of the kOps components. E.g anything that would try to get nodes would have to use `kubernetes.io/hostname` label for filtering instead of name. Could also be that hostname needs to be added to certs created by nodeup.

olemarkus:
The test failures are because of cgroup/systemd mismatch though. See https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/10595/pull-kops-e2e-kubernetes-aws/1350895281662595072/artifacts/99.79.195.34/kubelet.log

hakman:
> The test failures are because of cgroup/systemd mismatch though. See https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/10595/pull-kops-e2e-kubernetes-aws/1350895281662595072/artifacts/99.79.195.34/kubelet.log

That is only for the docker tests and caused by the incorrect removal of some code.
The containerd tests shows some different issues:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/10595/pull-kops-e2e-k8s-containerd/1350895281717121024/artifacts/13.238.118.41/kube-apiserver.log
```
I0117 20:21:09.349041       1 node_authorizer.go:291] NODE DENY: 'ip-172-20-40-160'
&authorizer.AttributesRecord{User:(*user.DefaultInfo)(0xc0060223c0), Verb:"get", Namespace:"",
APIGroup:"storage.k8s.io", APIVersion:"v1", Resource:"csinodes", Subresource:"",
Name:"ip-172-20-40-160.ap-southeast-2.compute.internal", ResourceRequest:true,
Path:"/apis/storage.k8s.io/v1/csinodes/ip-172-20-40-160.ap-southeast-2.compute.internal"}
```

pkg/model/components/kubelet.go
DOboznyi: Yes, sorry. I will remove this changes

pkg/model/components/kubelet.go
DOboznyi: Yes, sorry. I will remove this changes

DOboznyi: Fixed.

pkg/model/components/kubelet.go
DOboznyi: Fixed

pkg/model/components/kubelet.go
hakman: 👍

pkg/model/components/kubelet.go
hakman: 👍

DOboznyi:
Still same error with resolving on master

olemarkus:
Looking at the kubelet logs, the nodename is the same as before (i.e FQDN). I think the culprit may be that nodeup does something it should not based on the value of `node-override`.

hakman:
Most likely related to the `evaluateHostnameOverride` and `EvaluateHostnameOverride` functions. For sure it requires more digging.

DOboznyi:
On clusters with 1.18.2 with only changes in kubelet.go everything works. But also 2-3 min kubelet push error about resolving fqdn in logs

DOboznyi:
@hakman i think you are right. `EvaluateHostnameOverride` returns empty ([Link](https://github.com/kubernetes/kops/blob/789877984bbf746ffa2affba4feac0bd6ac17ea3/nodeup/pkg/model/context.go#L493) and it used as `nodeName` in bootstraping master labels ([Link](https://github.com/kubernetes/kops/blob/d5d08a43034dd4c7242cf1faa020cf9a8c3965e2/protokube/pkg/protokube/kube_boot.go#L146)) as i understand. So probably `nodeName` definition logic should be changed...

DOboznyi:
Also found that kube-api reject messages
```
I0118 10:10:37.271530       1 node_authorizer.go:291] NODE DENY: 'ip-172-20-54-197' &authorizer.AttributesRecord{User:(*user.DefaultInfo)(0xc0064a0fc0), Verb:"get", Namespace:"", APIGroup:"storage.k8s.io", APIVersion:"v1", Resource:"csinodes", Subresource:"", Name:"ip-172-20-54-197.eu-west-2.compute.internal", ResourceRequest:true, Path:"/apis/storage.k8s.io/v1/csinodes/ip-172-20-54-197.eu-west-2.compute.internal"}
```

olemarkus:
Those reject messages are a consequence of kOps passing in wrong value to the node flags. Should go away once you fix the functions mentioned above.

DOboznyi:
> Those reject messages are a consequence of kOps passing in wrong value to the node flags. Should go away once you fix the functions mentioned above.

You mean `evaluateHostnameOverride` and `EvaluateHostnameOverride`?

hakman:
I think the 2 functions should get a second arg for cloud provider and return same as before when either cloud provider or override match some provider. The purpose would be for kOps to behave same as before but just ignore the override when provider is set.

DOboznyi:
> I think the 2 functions should get a second arg for cloud provider and return same as before when either cloud provider or override match some provider. The purpose would be for kOps to behave same as before but just not need the override when provider is set.

But according to CIS hostname-override should be not used in kubelet. So i want to disable this flag completely by default.

olemarkus:
I would change those methods to test the CloudProvider instead.

DOboznyi:
`EvaluateHostnameOverride` works well, it returns empty string and after hostname got from OS ([Link](https://github.com/kubernetes/kops/blob/789877984bbf746ffa2affba4feac0bd6ac17ea3/nodeup/pkg/model/context.go#L499)).

olemarkus:
What it should return depends on the cloud provider. For AWS it is the internal FQDN, for DO, it is private ipv4 address. You need to ensure these functions keep returning the same values as before.

DOboznyi:
okay, I'll try to implement such changes

hakman:
> `EvaluateHostnameOverride` works well, it returns empty string and after hostname got from OS ([Link](https://github.com/kubernetes/kops/blob/789877984bbf746ffa2affba4feac0bd6ac17ea3/nodeup/pkg/model/context.go#L499)).

It does not work well. This is why this is not working. It should return same thing as before, but instead of checking for override should check for cloud provider.

DOboznyi:
@hakman what the purpose of [this code](https://github.com/kubernetes/kops/blob/ce30b3afbd39a4e1cf4186d39514fb2465885666/upup/pkg/fi/nodeup/command.go#L313)? How can i change it to such behaviour: If hostnameOverride not described in clusterSpec then it will not be substituted with an empty string.

nodeup/pkg/model/context.go
```diff
@@ -485,12 +485,13 @@ func (c *NodeupModelContext) BuildPrivateKeyTask(ctx *fi.ModelBuilderContext, na
 func (c *NodeupModelContext) NodeName() (string, error) {
 	// This mirrors nodeutil.GetHostName
 	hostnameOverride := c.Cluster.Spec.Kubelet.HostnameOverride
```
olemarkus, line 487: I don't think you need to recognize the user-set hostname. As far as I can tell, you cannot set this to anything meaningful. You can just add documentation to the API field (`pkg/apis/kops`) that setting this flag is deprecated and is ignored as of kOps 1.20.

Then you can test below only based on `cloudProvider`, which should make things a lot simpler for you.

nodeup/pkg/model/context.go
```diff
@@ -507,10 +508,15 @@ func (c *NodeupModelContext) NodeName() (string, error) {
...
 
 // EvaluateHostnameOverride returns the hostname after replacing some well-known placeholders
-func EvaluateHostnameOverride(hostnameOverride string) (string, error) {
-	if hostnameOverride == "" || hostnameOverride == "@hostname" {
+func EvaluateHostnameOverride(hostnameOverride string, cloudProvider string) (string, error) {
+	if hostnameOverride == "@hostname" {
 		return "", nil
 	}
+
+	if hostnameOverride == "" && cloudProvider != "external" {
```
olemarkus, line 516: I think also external cloudproviders generate the node names. But again, rather than generating hostnameOverride and testing on it, just test on cloudProvider.

nodeup/pkg/model/context.go
```diff
@@ -485,12 +485,13 @@ func (c *NodeupModelContext) BuildPrivateKeyTask(ctx *fi.ModelBuilderContext, na
 func (c *NodeupModelContext) NodeName() (string, error) {
 	// This mirrors nodeutil.GetHostName
 	hostnameOverride := c.Cluster.Spec.Kubelet.HostnameOverride
+	cloudProvider := c.Cluster.Spec.Kubelet.CloudProvider
```
olemarkus, line 488: You should use `c.Cluster.Spec.CloudProvider` here. This one is always set to the actual cloud provider and not to values such as `""` and `"external"`

olemarkus (changes requested):

DOboznyi:
@olemarkus it still failing. Do you have any ideas what i should change?

DOboznyi:
@hakman could you help me with debugging? I tried to remove hostname override almost completely (with deprecation mark as @olemarkus proposed), but it's still failing. Kubelet seems as working, but test failing.

olemarkus:
Am I missing something or is the change in `evaluateHostnameOverride` in `upup/pkg/fi/nodeup/command.go` missing?

DOboznyi:
> Am I missing something or is the change in `evaluateHostnameOverride` in `upup/pkg/fi/nodeup/command.go` missing?
Should I also remove hostnameOverride from kube-proxy configuration?

olemarkus:
I think so. It is subject to the same issue.

DOboznyi:
@olemarkus @hakman it seems as without hostname override kubelet set nodeName as private DNS FQDN from the cloud provider, but for calls to API it uses system node name which is the only hostname without zone. That's why it can't reach API and the cluster can't be stabilized. So can you give me some more info on how to add RBAC permissions? Because according to [docs](https://kubernetes.io/docs/reference/access-authn-authz/node/#kubelets-with-undifferentiated-usernames) it's not possible in such case reach API via Node Authorization.

nodeup/pkg/model/context.go
```diff
@@ -554,7 +544,8 @@ func EvaluateHostnameOverride(hostnameOverride string) (string, error) {
 	if len(result.Reservations[0].Instances) != 1 {
 		return "", fmt.Errorf("too many instances returned for the single instance-id")
 	}
-	return *(result.Reservations[0].Instances[0].PrivateDnsName), nil
+	hostname := strings.Split(*(result.Reservations[0].Instances[0].PrivateDnsName), ".")[0]
```
olemarkus, line 547: Have you verified that this function returns the internal FQDN and not just the hostname?

olemarkus (commented):
The "hostname" nodes are using comes from the certificate that nodeup builds.
Right now, it looks like nodes perhaps cannot correctly provision that certificate.

nodeup/pkg/model/context.go
DOboznyi: yes, i rolled out it locally. With completely removed hostnameOverride in kubelet `kubectl describe node` return node with label `kubernetes.io/hostname=ip-<xxx>-<xxx>-<xxx>-<xxx>`. So protokube should query node by `ip-<xxx>-<xxx>-<xxx>-<xxx>` which is element 0 if we split FQDN by `'.'`.

DOboznyi:
> The "hostname" nodes are using comes from the certificate that nodeup builds.
> Right now, it looks like nodes perhaps cannot correctly provision that certificate.

Where should I change the name used in the certificate?

olemarkus:
This most likely comes from inconsistent label use in protokube. I have filed a PR #10780  that will take care of that. Once that one is merged, you should be one step closer.

DOboznyi:
@olemarkus thank you so much.

HELLRAZOR6666 (approved):

olemarkus:
Looks like you are missing a `./hack/update-bazel.sh`

olemarkus:
You can verify tests locally by running `make bazel-test quick-ci`

DOboznyi:
> Looks like you are missing a `./hack/update-bazel.sh`

Still the error

nodeup/pkg/model/context.go
olemarkus: protokube now uses node name instead of `kubernetes.io/hostname` annotation. So this can be removed.

pkg/model/components/kubelet.go
```diff
@@ -209,13 +196,11 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
 	}
 
 	if clusterSpec.CloudConfig != nil && clusterSpec.CloudConfig.AWSEBSCSIDriver != nil && fi.BoolValue(clusterSpec.CloudConfig.AWSEBSCSIDriver.Enabled) {
-		if _, found := clusterSpec.Kubelet.FeatureGates["CSIMigrationAWS"]; !found {
```
olemarkus, line 212: Don't remove this bit, please

nodeup/pkg/model/context.go
```diff
@@ -507,16 +503,10 @@ func (c *NodeupModelContext) NodeName() (string, error) {
 }
 
 // EvaluateHostnameOverride returns the hostname after replacing some well-known placeholders
-func EvaluateHostnameOverride(hostnameOverride string) (string, error) {
-	if hostnameOverride == "" || hostnameOverride == "@hostname" {
+func EvaluateHostnameOverride(cloudProvider string) (string, error) {
+	if cloudProvider != "aws" {
 		return "", nil
```
olemarkus, line 508: You should move this block in here:
````
	if nodeName == "" {
		hostname, err := os.Hostname()
		if err != nil {
			klog.Fatalf("Couldn't determine hostname: %v", err)
		}
		nodeName = hostname
	}
````
That way this function will always return the appropriate hostname and you can remove the blank check above.

**Comment by olemarkus** [comment]
Those test failures you are seeing comes from them actually abusing the `hostnameOverride` param to create a consistent nodename for the test.

**Comment by DOboznyi** [comment]
> Those test failures you are seeing comes from them actually abusing the `hostnameOverride` param to create a consistent nodename for the test.

Thanks, I'll change all of them according to your suggestions.

File: nodeup/pkg/model/context.go (hunk 8 above)
Comment (DOboznyi) [reply nodeup/pkg/model/context.go:508]: Then do we need this function? Or it can be removed?

Comment (olemarkus) [reply nodeup/pkg/model/context.go:508]: The only thing this does is pass in the node name to protokube, which is in turn used to query its own node labels. 

Interestingly enough, if you remove that function entirely and always set the os name, I believe that going _back_ to using the hostname annotation will make this work.

I wonder if you can actually just go into `protokube/pkg/protokube/labeler.go` and just fetch the os hostname there.

Comment (olemarkus) [reply nodeup/pkg/model/context.go:508]: Scratch that. Forgot that it was also used for the certname. Then we actually need to know the actual nodename. (in `buildMasterKubeletKubeconfig`)

**Comment by DOboznyi** [comment]
@olemarkus maybe i missed something again, but it still not working...

**Comment by olemarkus** [comment]
Technically this is caused by missing `./hack/update-expected`.
But I am not sure how you can deterministically set the node-name. It will always change depending on where the test is running.

**Comment by DOboznyi** [comment]
> Technically this is caused by missing `./hack/update-expected`.
> 
> But I am not sure how you can deterministically set the node-name. It will always change depending on where the test is running.

Yes, I know, so should I change somehow tests?

**Comment by olemarkus** [comment]
If you cannot see a way of mocking that, I think we may just have to remove it for now

**Comment by DOboznyi** [comment]
@olemarkus but how can fix aws test? It still failing :(

**Comment by DOboznyi** [comment]
@olemarkus CI tests removed, could you help with e2e fails?

**Comment by olemarkus** [comment]
The cert is wrong. You see the error:
```
I0223 12:18:32.931137    4277 kubelet_node_status.go:71] Attempting to register node ip-172-20-59-153.eu-west-1.compute.internal
E0223 12:18:32.933453    4277 kubelet_node_status.go:93] Unable to register node "ip-172-20-59-153.eu-west-1.compute.internal" with API server: nodes "ip-172-20-59-153.eu-west-1.compute.internal" is forbidden: node "ip-172-20-59-153" is not allowed to modify node "ip-172-20-59-153.eu-west-1.compute.internal"
```
Further down there is
```
I0223 12:18:33.336972    4277 csi_plugin.go:1016] Failed to contact API server when waiting for CSINode publishing: csinodes.storage.k8s.io "ip-172-20-59-153.eu-west-1.compute.internal" is forbidden: User "system:node:ip-172-20-59-153" cannot get resource "csinodes" in API group "storage.k8s.io" at the cluster scope: can only access CSINode with the same name as the requesting node
```
Use r` system:node:ip-172-20-59-153` should be `system:node:ip-172-20-59-153..eu-west-1.compute.internal"

File: nodeup/pkg/model/context.go [hunk 9]
```diff
@@ -484,79 +481,16 @@ func (c *NodeupModelContext) BuildPrivateKeyTask(ctx *fi.ModelBuilderContext, na
 // NodeName returns the name of the local Node, as it will be created in k8s
 func (c *NodeupModelContext) NodeName() (string, error) {
```
olemarkus, line 482: This function need to return FQDN as it is used for the node cert name.See my previous comment.

DOboznyi: Okay, i'll return code which return FQDN

DOboznyi:
@olemarkus seems as previous problem gone. Are the other tests critical?

nodeup/pkg/model/context.go
```diff
@@ -484,11 +484,7 @@ func (c *NodeupModelContext) BuildPrivateKeyTask(ctx *fi.ModelBuilderContext, na
...
 	// This mirrors nodeutil.GetHostName
-	hostnameOverride := c.Cluster.Spec.Kubelet.HostnameOverride
-
-	if c.IsMaster && c.Cluster.Spec.MasterKubelet.HostnameOverride != "" {
-		hostnameOverride = c.Cluster.Spec.MasterKubelet.HostnameOverride
-	}
+	hostnameOverride := string('@') + c.Cluster.Spec.CloudProvider
```
olemarkus, line 487: I assume this is just temporary for testing?

pkg/model/components/kubelet.go
```diff
@@ -208,6 +195,9 @@ func (b *KubeletOptionsBuilder) BuildOptions(o interface{}) error {
 		if _, found := clusterSpec.Kubelet.FeatureGates["CSIMigrationAWS"]; !found {
 			clusterSpec.Kubelet.FeatureGates["CSIMigrationAWS"] = "true"
 		}
+		if _, found := clusterSpec.Kubelet.FeatureGates["CSIMigrationAWSComplete"]; !found {
+			clusterSpec.Kubelet.FeatureGates["CSIMigrationAWSComplete"] = "true"
+		}
```
olemarkus, line 200: Can you rebase this against master? If not you may get into conflicts.

DOboznyi: Fixed

nodeup/pkg/model/context.go
DOboznyi: How should i change this part?

olemarkus: Isn't the goal of completely getting rid of hostnameOverride?

nodeup/pkg/model/context.go
```diff
@@ -484,11 +484,7 @@ func (c *NodeupModelContext) BuildPrivateKeyTask(ctx *fi.ModelBuilderContext, na
...
 
 	nodeName, err := EvaluateHostnameOverride(hostnameOverride)
```
olemarkus, line 489: Is this now the only call in the code going to `EvaluateHostnameOverride`? if so, I think you can inline the function here. `NodeName` is a much more appropriate name than `EvaluateHostnameOverride` for what this function is used for now.

wdyt?

hakman:
Just wanted to leave a comment here that this may have gone against what upstream was thinking.
Seems that the `--cloud-provider` flag will be removed soon, as discussed in https://github.com/kubernetes/kubernetes/pull/90408.

PR #10609, opened 2021-01-19
feat: implement azure get api ingress status fn

ngalantowicz:
Fix:
- Updates azure getting started doc to instruct correct feature flag value

Feature:
- Creates azureup network interfaces client with List method

Update:
- Uses vmScaleSet client and new network interfaces client to get masters ip for api ingress status

### DIFF
--- a/upup/pkg/fi/cloudup/azure/azure_cloud.go
+++ b/upup/pkg/fi/cloudup/azure/azure_cloud.go
@@ -17,7 +17,9 @@ limitations under the License.
 package azure
 
 import (
+	"context"
 	"errors"
+	"fmt"
 
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"k8s.io/kops/dnsprovider/pkg/dnsprovider"
@@ -52,20 +54,22 @@ type AzureCloud interface {
 	VMScaleSetVM() VMScaleSetVMsClient
 	Disk() DisksClient
 	RoleAssignment() RoleAssignmentsClient
+	NetworkInterface() NetworkInterfacesClient
 }
 
 type azureCloudImplementation struct {
-	subscriptionID        string
-	location              string
-	tags                  map[string]string
-	resourceGroupsClient  ResourceGroupsClient
-	vnetsClient           VirtualNetworksClient
-	subnetsClient         SubnetsClient
-	routeTablesClient     RouteTablesClient
-	vmscaleSetsClient     VMScaleSetsClient
-	vmscaleSetVMsClient   VMScaleSetVMsClient
-	disksClient           DisksClient
-	roleAssignmentsClient RoleAssignmentsClient
+	subscriptionID          string
+	location                string
+	tags                    map[string]string
+	resourceGroupsClient    ResourceGroupsClient
+	vnetsClient             VirtualNetworksClient
+	subnetsClient           SubnetsClient
+	routeTablesClient       RouteTablesClient
+	vmscaleSetsClient       VMScaleSetsClient
+	vmscaleSetVMsClient     VMScaleSetVMsClient
+	disksClient             DisksClient
+	roleAssignmentsClient   RoleAssignmentsClient
+	networkInterfacesClient NetworkInterfacesClient
 }
 
 var _ fi.Cloud = &azureCloudImplementation{}
@@ -78,17 +82,18 @@ func NewAzureCloud(subscriptionID, location string, tags map[string]string) (Azu
 	}
 
 	return &azureCloudImplementation{
-		subscriptionID:        subscriptionID,
-		location:              location,
-		tags:                  tags,
-		resourceGroupsClient:  newResourceGroupsClientImpl(subscriptionID, authorizer),
-		vnetsClient:           newVirtualNetworksClientImpl(subscriptionID, authorizer),
-		subnetsClient:         newSubnetsClientImpl(subscriptionID, authorizer),
-		routeTablesClient:     newRouteTablesClientImpl(subscriptionID, authorizer),
-		vmscaleSetsClient:     newVMScaleSetsClientImpl(subscriptionID, authorizer),
-		vmscaleSetVMsClient:   newVMScaleSetVMsClientImpl(subscriptionID, authorizer),
-		disksClient:           newDisksClientImpl(subscriptionID, authorizer),
-		roleAssignmentsClient: newRoleAssignmentsClientImpl(subscriptionID, authorizer),
+		subscriptionID:          subscriptionID,
+		location:                location,
+		tags:                    tags,
+		resourceGroupsClient:    newResourceGroupsClientImpl(subscriptionID, authorizer),
+		vnetsClient:             newVirtualNetworksClientImpl(subscriptionID, authorizer),
+		subnetsClient:           newSubnetsClientImpl(subscriptionID, authorizer),
+		routeTablesClient:       newRouteTablesClientImpl(subscriptionID, authorizer),
+		vmscaleSetsClient:       newVMScaleSetsClientImpl(subscriptionID, authorizer),
+		vmscaleSetVMsClient:     newVMScaleSetVMsClientImpl(subscriptionID, authorizer),
+		disksClient:             newDisksClientImpl(subscriptionID, authorizer),
+		roleAssignmentsClient:   newRoleAssignmentsClientImpl(subscriptionID, authorizer),
+		networkInterfacesClient: newNetworkInterfacesClientImpl(subscriptionID, authorizer),
 	}, nil
 }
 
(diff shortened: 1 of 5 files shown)


ngalantowicz:
CLA has been signed 👍

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -128,9 +134,34 @@ func (c *azureCloudImplementation) AddClusterTags(tags map[string]*string) {
 }
 
 func (c *azureCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([]kops.ApiIngressStatus, error) {
-	// TODO(kenji): Implement this. Currently we return nil as we
-	// don't create any resources for ingress to the API server.
-	return nil, nil
+	var ingresses []kops.ApiIngressStatus
+
+	// Get scale sets in cluster resource group and find masters scale set
+	scaleSets, err := c.vmscaleSetsClient.List(context.TODO(), cluster.Spec.CloudConfig.Azure.ResourceGroupName)
```
kenji-cloudnatix, line 140: How about ` cluster.AzureResourceGroupName()`? The function handles a case where a resource group is unspecified in the cluster spec (and kops creates a new resource group with the same name as the cluster).

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -128,9 +134,34 @@ func (c *azureCloudImplementation) AddClusterTags(tags map[string]*string) {
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting Cluster Master Scale Set for API Ingress Status: %s", err)
+	}
+	var vmssName string
+	for i := range scaleSets {
+		if strings.Contains(*scaleSets[i].Name, ".masters.") {
```
kenji-cloudnatix, line 146: Wondering if there is a better way to do.. Maybe using tags is better? I believe other people can better chime-in here.

kenji-cloudnatix (commented):
Thanks for making the change, @ngalantowicz ! I just have one comment on the use of `cluster.AzureResourceGroupName()`.

rifelpet:
hm the CLA bot still isn't detecting that you signed it. Are you sure you signed with the email address(es) associated with these git commits?

upup/pkg/fi/cloudup/azure/azure_cloud.go
rifelpet: I agree tags would be better here. This is how openstack does it:

https://github.com/kubernetes/kops/blob/55e38c96d3b83c7dd7c89c205fb0a1f2914e7851/upup/pkg/fi/cloudup/openstack/cloud.go#L704-L709

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -128,9 +134,34 @@ func (c *azureCloudImplementation) AddClusterTags(tags map[string]*string) {
...
+	nis, err := c.NetworkInterface().ListScaleSetsNetworkInterfaces(context.TODO(), cluster.Spec.CloudConfig.Azure.ResourceGroupName, vmssName)
+	if err != nil {
+		return nil, fmt.Errorf("error getting Master Scale Set Network Interfaces for API Ingress Status: %s", err)
+	}
+	for _, ni := range nis {
+		for _, i := range *ni.IPConfigurations {
+			ingresses = append(ingresses, kops.ApiIngressStatus{
+				IP: *i.PrivateIPAddress,
+			})
+		}
```
rifelpet, line 162: I'm wondering how we should handle VMs that have multiple network interfaces or multiple IPConfigurations. I _think_ we would want to only have one ApiIngressStatus for each master VM. Would there be situations in which a master VM might have multiple? Perhaps a CNI provider that adds NIs or IPs to VMs, for example. Openstack has some additional metadata it uses to look for a specific "fixed" IP on the instance, does Azure support anything similar?

https://github.com/kubernetes/kops/blob/55e38c96d3b83c7dd7c89c205fb0a1f2914e7851/upup/pkg/fi/cloudup/openstack/utils.go#L111

ngalantowicz:
Thanks for the quick comments.  Based on your suggestions i made the following updates
- utilize `cluster.AzureResourceGroupName()`
- use tags for master scale set discovery
```
		val, ok := scaleSet.Tags["KubernetesCluster"]
		val2, ok2 := scaleSet.Tags["k8s.io_role_master"]
		if ok && *val == cluster.Name && ok2 && *val2 == "1" {
			vmssName = *scaleSet.Name
			break
		}
```
- identify single network interface ip based on "Primary" metadata
```
               if *ni.Primary == true {
			for _, i := range *ni.IPConfigurations {
				ingresses = append(ingresses, kops.ApiIngressStatus{
					IP: *i.PrivateIPAddress,
				})
			}
		}
```

Regarding CLA:
Screen Shot 2021-01-20 at 12 12 03 AM
Screen Shot 2021-01-20 at 12 08 53 AM
According to the CNCF CLA Signing FAQ
```
...the Github user that the commit is attributed to (based on a match between the commit authorship and Github profile) must be linked to a Linux Foundation ID user (on the "social network logins" tab), who is authorized by a signed ICLA and/or a CCLA.
```
Not sure what i'm missing

rifelpet:
@ngalantowicz i pulled the branch locally and noticed your git commit is authored by your gmail.com address rather than your t1cg.com address. [here is](https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit) some straight forward instructions for changing the email on an existing commit.

rifelpet:
there we go!
/ok-to-test

rifelpet:
@kenji-cloudnatix this looks good to me. If it looks good to you then we can get this merged

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -128,9 +133,39 @@ func (c *azureCloudImplementation) AddClusterTags(tags map[string]*string) {
...
+	var rg string = cluster.AzureResourceGroupName()
+
+	// Get scale sets in cluster resource group and find masters scale set
+	scaleSets, err := c.vmscaleSetsClient.List(context.TODO(), rg)
+	if err != nil {
+		return nil, fmt.Errorf("error getting Cluster Master Scale Set for API Ingress Status: %s", err)
+	}
+	var vmssName string
+	for _, scaleSet := range scaleSets {
+		val, ok := scaleSet.Tags["KubernetesCluster"]
```
kenji-cloudnatix, line 146: How about 

```go
val, ok := scaleSet.Tags[TagClusterName]
val2, ok2 := scaleSet.Tags[TagNameRolePrefix + TagRoleMaster]
```

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -128,9 +133,39 @@ func (c *azureCloudImplementation) AddClusterTags(tags map[string]*string) {
...
+		val2, ok2 := scaleSet.Tags["k8s.io_role_master"]
+		if ok && *val == cluster.Name && ok2 && *val2 == "1" {
+			vmssName = *scaleSet.Name
+			break
+		}
+	}
```
kenji-cloudnatix, line 152: How about returning an error if `vmssName` is an empty string?

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -128,9 +133,39 @@ func (c *azureCloudImplementation) AddClusterTags(tags map[string]*string) {
...
+
+	// Get masters scale set network interfaces and append to api ingress status
+	nis, err := c.NetworkInterface().ListScaleSetsNetworkInterfaces(context.TODO(), rg, vmssName)
+	if err != nil {
+		return nil, fmt.Errorf("error getting Master Scale Set Network Interfaces for API Ingress Status: %s", err)
+	}
+	for _, ni := range nis {
+		if *ni.Primary {
```
kenji-cloudnatix, line 160: Not sure the convention in kops, but might be nicer to do

```go
if !*ni.Primary {
  continue
}
for _, i := range *ni.IPConfigurations {
  ingresses = append(ingresses, kops.ApiIngressStatus{
    IP: *i.PrivateIPAddress,
  }
}
```

kenji-cloudnatix (approved):
Have minor comments, but looks good to me!

upup/pkg/fi/cloudup/azure/azure_cloud.go
ngalantowicz: I like this better too. When looking over recent comments in the remaining azure work issue I started having concern about this.  Thanks for the suggestion 👍

rifelpet:
Thanks!

/lgtm
/approve

ngalantowicz:
rebased

PR #10610, opened 2021-01-19
Increase CoreDNS default ttl

johanneswuerbach:
Increase the CoreDNS default ttl to be inline with kubernetes upstream https://github.com/kubernetes/kubernetes/pull/76238 and the previous kube-dns ttl.

We switching from kube-dns to CoreDNS, we noticed a visible bump in coredns-nodelocal request duration and an increase in cache misses.

image
image

This change should make the switch less impactful for others.


hakman:
Thanks @johanneswuerbach. It's a nice update.
/lgtm
/approve

PR #10616, opened 2021-01-20
Add network and router availability zone hints to OpenStack

ottosulin:
This pull request makes it possible to add availability zone hints to routers and networks if the hints are supported.


zetaab (approved):
/lgtm

cc @olemarkus

upup/pkg/fi/values.go
```diff
@@ -30,6 +30,20 @@ func StringValue(s *string) string {
 	return *s
 }
 
+//StringSliceValue takes a slice of string pointers and returns a slice of strings
+func StringSliceValue(stringSlice []*string) []string {
+	if stringSlice == nil {
```
zetaab, line 35: still checking is this correct or not...

zetaab (changes requested):

upup/pkg/fi/values.go
zetaab, line 35: this if is not needed, and should be removed

zetaab (changes requested):

upup/pkg/fi/values.go
```diff
@@ -43,6 +57,18 @@ func String(s string) *string {
 	return &s
 }
 
+// StringSlice is a helper that builds a []*string from a slice of strings
+func StringSlice(stringSlice []string) []*string {
+	if stringSlice == nil {
```
zetaab, line 62: delete this also

zetaab (approved):

olemarkus:
Thanks. This looks god to me.

Issue #10412, opened 2020-12-11
Remaining work for Azure support


kenji-cloudnatix:
This issues track the remaining TODOs from https://github.com/kubernetes/kops/issues/3957 (Azure support)

- [ ] Be able to configure some more parameters for VM Scale Set such as Windows
- [ ] Be able to configure some more parameters for Azure Disks such as zone
- [ ] Be able to configure some more parameters for subnets such as security groups
- [ ] Support DNS
- [ ] Support tag update of virtual networks
- [ ] Support ACL for Azure blob VFSImplement
- [x] Implement `azureCloudImplementation.GetApiIngressStatus`. Currently we don't create any resources for ingress to the API server.
- [x] ~Be able to mount etcd volumes from protokube. Currently we rely on etcd-manager for volume mount~
- [ ] Populate node labels from `nodeIdentifier.IdentifyNodeChange`.
- [ ] Implement `azureCloudImplementation.buildCloudInstanceGroup` so that kops can correclty tell whether a VM is up-to-date or not.

Aut0R3V:
anything I can do here?

olemarkus:
Any of these are up for grabs, I would think.

rifelpet:
I think we can remove `Be able to mount etcd volumes from protokube. Currently we rely on etcd-manager for volume mount` from the list. That is only needed for the "legacy" etcd option before etcd-manager. Given that it will eventually go away, I don't see a need to add support for it in Azure.

ngalantowicz:
Not sure how tasks are divided up for an issue like this or if work started needs to be reported, but folks I work with and myself are interested in starting

> Implement azureCloudImplementation.GetApiIngressStatus. Currently we don't create any resources for ingress to the API server.

Will be first time kops contributors, so any pertinent info to contributing that may be omitted from relevant docs is much appreciated

rifelpet:
[...]

PR #10619, opened 2021-01-20
Use correct tag when creating node labels from azure cloud tags

NickSchleicher:
Fixes first bug mentioned here https://github.com/kubernetes/kops/issues/10412#issuecomment-763880477
`nodeLabels` specified in the InstanceGroup spec weren't being labeled on nodes created in Azure

### DIFF
--- a/pkg/nodeidentity/azure/BUILD.bazel
+++ b/pkg/nodeidentity/azure/BUILD.bazel
@@ -11,6 +11,7 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//pkg/nodeidentity:go_default_library",
+        "//upup/pkg/fi:go_default_library",
         "//vendor/github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute:go_default_library",
         "//vendor/github.com/Azure/go-autorest/autorest/azure/auth:go_default_library",
         "//vendor/k8s.io/api/core/v1:go_default_library",
--- a/pkg/nodeidentity/azure/identify.go
+++ b/pkg/nodeidentity/azure/identify.go
@@ -27,13 +27,15 @@ import (
 	expirationcache "k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/nodeidentity"
+	"k8s.io/kops/upup/pkg/fi"
 )
 
 const (
 	// InstanceGroupNameTag is the key of the tag used to identify an
 	// instance group that VM ScaleSet belongs.
 	InstanceGroupNameTag = "kops.k8s.io_instancegroup"
-
+	// ClusterNodeTemplateLabel is the prefix used on node labels when copying to cloud tags.
+	ClusterNodeTemplateLabel = "k8s.io_cluster_node-template_label_"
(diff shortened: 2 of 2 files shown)


kenji-cloudnatix:
This looks good to me, but I believe we also need to convert "_" back to "/" in `v` as we convert  "/" to  "_" in https://github.com/kubernetes/kops/blob/309dc9fc26a5eb65614c3961239fd102f272ace5/pkg/model/azuremodel/context.go#L108 ?

If we do that, I guess we also need to convert "_" to "__" or something else in `CloudTagsForInstanceGroup` so that we can distinguish (a) original "_" and (b) "_" after conversion.

NickSchleicher:
hey @kenji-cloudnatix, i think the formatting got messed up on your message, i'm seeing empty quotes
but I remember one of the concerns i had about converting `/` back to `_` was if someone had this in their instanceGroup spec
which is part of what i think you're trying to describe
```
  nodeLabels:
    disk_type: ssd
```

I feel like we would almost have to convert user defined `/` into some pattern other than `_` in order to convert back to `/` as k8s labels without any side effects
that way they don't end up with `disk/type: ssd` as a label on their node, in this example

kenji-cloudnatix:
Ah, sorry. Yes, that's what I meant!

I believe something like below work?

- Convert `_` to `__`.
- Convert `/` to `_s`

Then we can apply the following rule to get the original string:

- First convert `_s` back to `/`
- Then convert `__` back to `_`

NickSchleicher:
thanks @kenji-cloudnatix! I don't think i see any issues going that route

also @kenji-cloudnatix and @rifelpet i can't get to the `_` `/` issue atm with a lot of work needed for my org, but I do intend to since my org will potentially have an azure based contract in the coming months. If possible I would like to get this fix in so we at least won't have to fork kops to get nodeLabels applied when that time comes. 

To recap: 
Right now node labels applied through `nodeLabels` in instance group spec are being created as tags in the cloud, but aren't being applied to nodes as k8s labels due to this error in the controller
```
I0120 18:14:43.199009       1 node_controller.go:144] sending patch for node "web000001": "{\"metadata\":{\"labels\":{\"\":\"web\"}}}"
W0120 18:14:43.204913       1 node_controller.go:110] failed to patch node labels on web000001: error applying patch to node: Node "web000001" is invalid: [metadata.labels: Invalid value: "": name part must be non-empty, metadata.labels: Invalid value: "": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')]
```

This pr fixes that issue. I can create another issue with the relevant notes, for the `/` conversion that needs to happen in the preset cloud tags applied by kops, along with the ones that can be created through `nodeLabels` now that it will work again. And then get to that issue as soon as I can, I already know we will eventually run into it. Let me know if any concerns!

kenji-cloudnatix:
Thanks for the update, @NickSchleicher ! Works for me.

kenji-cloudnatix:
FYI: Just created https://github.com/kubernetes/kops/pull/10935 for the remaining issue discussed in this PR.

PR #10622, opened 2021-01-21
[Digital Ocean] Add SFO3 region. Also update e2e tests to use full list of supported zones

srikiz:
Small addition to use SFO3 to the list of regions DO KOPS supports.


tests/e2e/go.mod
```diff
@@ -8,5 +8,5 @@ require (
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/yaml.v2 v2.3.0
 	k8s.io/klog/v2 v2.4.0
-	sigs.k8s.io/kubetest2 v0.0.0-20210115020551-4275dd0a0d63
```
rifelpet, line 11: I think this is downgrading the package. Any reason why that's necessary? Otherwise I think we can revert this go.mod and go.sum change

srikiz: Oh, I didn't realize that, thanks. I will revert the go.mod and go.sum change.

srikiz: @rifelpet - I think it should be okay now. Please have a look when you get a chance. Thanks !

tests/e2e/go.mod
```diff
@@ -9,4 +9,4 @@ require (
 	gopkg.in/yaml.v2 v2.3.0
 	k8s.io/klog/v2 v2.4.0
 	sigs.k8s.io/kubetest2 v0.0.0-20210115020551-4275dd0a0d63
-)
```
rifelpet, line 12: @srikiz I think these trailing newline changes in go.mod and go.sum need to be reverted as well.

srikiz: Sure, updated.

rifelpet (approved):
/lgtm
/approve

PR #10629, opened 2021-01-21
Remove coredns dnsprovider

olemarkus:
This code depends on a 2017 version of coredns. There is also no path in kOps that will configure coredns as flag to protokube/dns-controller. So I vote we punt this.

/kind cleanup


hakman:
Seems a good thing to me.

PR #10633, opened 2021-01-22
Add startup probe for calico-kube-controllers

hakman:
This should reduce considerably the issues during rolling update when https://github.com/projectcalico/calico/issues/3751 is hit, until next version of Calico is released.


upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
```diff
@@ -313,7 +313,7 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*chann
 	if kubeDNS.Provider == "CoreDNS" {
 		{
 			key := "coredns.addons.k8s.io"
-			version := "1.7.0-kops.2"
+			version := "1.7.0-kops.3"
```
hakman, line 316: This was forgot in a previous PR.

PR #10634, opened 2021-01-22
Replace gopkg yaml with k8s-sigs yaml

olemarkus:


olemarkus:
/hold

will switch to sigs yaml instead to make things more consistent.

PR #10651, opened 2021-01-24
Add back support for kubenet style networking with containerd

hakman:
Mostly reverting #10501.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -40,7 +40,6 @@ import (
 	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/pkg/kubeconfig"
 	"k8s.io/kops/pkg/kubemanifest"
-	"k8s.io/kops/pkg/model/components"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup"
 	"k8s.io/kops/upup/pkg/fi/utils"
@@ -488,9 +487,6 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 
 	if c.ContainerRuntime != "" {
 		cluster.Spec.ContainerRuntime = c.ContainerRuntime
-	}
-	if c.ContainerRuntime == "containerd" && components.UsesKubenet(cluster.Spec.Networking) {
-		return fmt.Errorf("--networking with CNI plugin is required for containerd")
 	}
 
 	if c.NetworkCIDR != "" {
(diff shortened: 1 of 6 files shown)


hakman:
Test results are good when forcing kubenet style networking with containerd:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/10651/pull-kops-e2e-k8s-containerd/1353390037454032896

cmd/kops/create_cluster.go
```diff
@@ -489,8 +488,8 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 	if c.ContainerRuntime != "" {
 		cluster.Spec.ContainerRuntime = c.ContainerRuntime
 	}
-	if c.ContainerRuntime == "containerd" && components.UsesKubenet(cluster.Spec.Networking) {
-		return fmt.Errorf("--networking with CNI plugin is required for containerd")
+	if c.ContainerRuntime == "containerd" && c.Networking == "" {
```
rifelpet, line 491: we default to networking = kubenet correct? if the default `--networking` value now works with containerd then we should just drop this "if" statement altogether rather than require a `--networking` flag.

hakman: True. I wanted to force people to choose. Removed.

hakman:
Just checked and the kope.io tests are running once again. The cherry-pick should be safe to merge to 1.19:

nodeup/pkg/model/containerd.go
```diff
@@ -294,3 +303,37 @@ runtime-endpoint: unix:///run/containerd/containerd.sock
 		Type:     nodetasks.FileType_File,
 	})
 }
+
+// buildCNIConfigTemplateFile is responsible for creating a special template for setups using Kubenet
+func (b *ContainerdBuilder) buildCNIConfigTemplateFile(c *fi.ModelBuilderContext) {
+	contents := `{
+    "cniVersion": "0.4.0",
```
justinsb, line 310: This is a different configuration from the one we reverted - where did you get it from?

(I'm trying to track down the "Kubernetes e2e suite.[sig-network] Services should preserve source pod IP for traffic thru service cluster IP [LinuxOnly]" flakes)

justinsb: I proposed taking the upstream configuration in https://github.com/kubernetes/kops/pulls

hakman: I tried this version at some point, but more tests were failing. We can merge #10759 as an experiment as see what happens. PR tests are done with Calico.

The current config is very similar from what was reverted and was taken from :
https://github.com/containerd/containerd/blob/774cb16a446bf7210584a455b47e58d488455aab/script/setup/install-cni#L37-L68

Issue #10206, opened 2020-11-10
nodeup error downloading assets from storage.googleapis.com when using official Ubuntu 20.04 AMI


CheyiLin:
**1. What `kops` version are you running?**
`Version 1.18.2 (git-84495481e4)`

**2. What Kubernetes version are you running?**
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.13", GitCommit:"30d651da517185653e34e7ab99a792be6a3d9495", GitTreeState:"clean", BuildDate:"2020-10-15T01:06:31Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.13", GitCommit:"30d651da517185653e34e7ab99a792be6a3d9495", GitTreeState:"clean", BuildDate:"2020-10-15T00:59:17Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
`kops rolling-update cluster --name my-cluster-name --yes`

**5. What happened after the commands executed?**
The new node hasn't joined cluster over 20 minutes.

<details>
<summary>kops logs</summary>

```
...
I1110 16:21:43.388467   23599 instancegroups.go:143] waiting for 15s after detaching instance
I1110 16:21:58.391501   23599 instancegroups.go:383] Validating the cluster.
I1110 16:22:01.241446   23599 instancegroups.go:440] Cluster did not pass validation, will retry in "30s": InstanceGroup "nodes-a1" did not have enough nodes 0 vs 1.
I1110 16:22:34.057496   23599 instancegroups.go:440] Cluster did not pass validation, will retry in "30s": machine "i-0de5271375ff81ce1" has not yet joined cluster.
I1110 16:23:07.378220   23599 instancegroups.go:440] Cluster did not pass validation, will retry in "30s": machine "i-0de5271375ff81ce1" has not yet joined cluster.
I1110 16:23:40.039431   23599 instancegroups.go:440] Cluster did not pass validation, will retry in "30s": machine "i-0de5271375ff81ce1" has not yet joined cluster.
I1110 16:24:12.913123   23599 instancegroups.go:440] Cluster did not pass validation, will retry in "30s": machine "i-0de5271375ff81ce1" has not yet joined cluster.
[...]

PR #10654, opened 2021-01-25
Set the tcp_rmem sysctl in bootstrap script

justinsb:
This ensures that we're using our settings for downloading nodeup
itself and any assets that nodeup downloads.  This is a workaround for
reported problems with the initial download on some kernels otherwise.

Issue #10206


olemarkus:
This looks good.

Do we know if there is any kernel issue worth tracking? I suspect it won't hit older distros like Debian 9 anyway though.

hakman:
Seems there was some discussion around the patch and workaround, but no conclusion:
https://lore.kernel.org/netdev/CANn89iL48fnjMQ=rt7oGRC_cejhuwfvcyy5JcUs=k5Cet+-Vwg@mail.gmail.com/

justinsb:
Removed the WIP, based on feedback (thanks @olemarkus  and @hakman ), and a similar workaround described here: https://github.com/kubernetes/kops/issues/10206#issuecomment-766852332

The sysctls might be setting up larger windows that we should by default, but we can tackle that separately.

PR #10655, opened 2021-01-25
Default IMDSv2 to "optional" for AWS

hakman:
IMDSv2 breaks any setup using Flatcar or Debian 9 as base image and seems to be the reason kube-router tests stoped working. The change, even though recommended by AWS seems to be highly disruptive in our case and for now should remain optional.

/cc @rifelpet @olemarkus


rifelpet:
Maybe we should default to required for new clusters only. I believe that would involve setting it in `setupMasters` and `setupNodes` [here](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster.go)

hakman:
@rifelpet This would mean adding the setting to each IG on creation and may look a bit odd:
```
  instanceMetadata:
    httpTokens: required
```

olemarkus:
It would be interesting to know why it fails. It may be one rather wants to increase allowed number of hops.

hakman:
> It would be interesting to know why it fails. It may be one rather wants to increase allowed number of hops.

I think kube-router is because of the number of hops.
Debian 9 fails because of very old cloud-init version.
Flatcar fails because of strange way of getting userdata from Ignition: https://github.com/kinvolk/Flatcar/issues/220.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -782,6 +792,16 @@ func setupNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap ma
 			g.Spec.Zones = []string{zone}
 		}
 
+		if api.CloudProviderID(cluster.Spec.CloudProvider) == api.CloudProviderAWS && (g.Spec.InstanceMetadata == nil || g.Spec.InstanceMetadata.HTTPTokens == nil) {
+			k8sVersion, err := version.ParseKubernetesVersion(cluster.Spec.KubernetesVersion)
+			if err == nil && version.IsKubernetesGTE("1.20", *k8sVersion) {
+				g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
```
rifelpet, line 798: this will overwrite InstanceMetadata if it already exists but HTTPTokens is nil. We should probably only create InstanceMetadata if it is nil, otherwise just set HTTPTokens and HTTPPutResponseHopLimit if they arent set.

hakman: Thanks. I think this last change should do it.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -674,6 +674,16 @@ func setupMasters(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap
 				g.Spec.Zones = []string{zone}
 			}
 
+			if api.CloudProviderID(cluster.Spec.CloudProvider) == api.CloudProviderAWS && (g.Spec.InstanceMetadata == nil || g.Spec.InstanceMetadata.HTTPTokens == nil) {
+				k8sVersion, err := version.ParseKubernetesVersion(cluster.Spec.KubernetesVersion)
+				if err == nil && version.IsKubernetesGTE("1.20", *k8sVersion) {
```
olemarkus, line 679: Looks like it will be enabled on new clusters with k8s 1.20+. Do we really need the k8s check here when it is only for new clusters?

hakman: I discussed a bit with @rifelpet and he is worried that this may break some existing clusters if operators don't read the release notes.
I would question a bit the requirement for 1.20+ and change to skip for debian 9 and flatcar images.

olemarkus: If the intention is to only do this for new clusters, it won't break any existing ones (I hope)

rifelpet: - [ ] I think historically we've made node-level cloud provider changes by only affecting new clusters and the newly supported k8s version. That way someone creating k8s 1.18 clusters wont have a change in behavior when upgrading from kops 1.18 to 1.19.

This might be especially important for changes that could break workloads like this one.

I don't nt have a particularly strong opinion about this though as long as the release notes make it abundantly clear. Given that this is for 1.20 perhaps we get more input at the mext office hours?

hakman: Good point, Friday it is then. 😄

bharath-123: I have observed that there are some components which break if IMDSv2 is enforced. eg: Kube2iam which does not support IMDSv2 properly. An upgrade could break workloads which depend on kube2iam for IAM access. Enforcing IMDSv2 on a cluster without the operator clearly knowing about it can make for a pretty bad day at the office....

rifelpet:
as discussed in office hours:
/lgtm

PR #10660, opened 2021-01-26
Fix cluster_spec.md indentation

trondhindenes:
Fixed wrong intendation


olemarkus:
Thanks for catching.

Also see cluster autoscaler did not make it to https://kops.sigs.k8s.io/addons/
And that the addon page also has some indentation issues.

Not good ...

/ok-to-test

Issue #10664, opened 2021-01-26
InstanceGroup externalLoadBalancer - found duplicate tasks


kxesd:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
```
Version 1.19.0-beta.3 (git-e43f1cc6e3c77d093935d1706042861095d75eb7)
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-15T11:35:50Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.12", GitCommit:"7cd5e9086de8ae25d6a1514d0c87bac67ca4a481", GitTreeState:"clean", BuildDate:"2020-11-12T09:11:15Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
a. Create 2 or more InstanceGroup's with the same `externalLoadBalancers` in the spec
b. `kops update cluster mycluster.tld`

**5. What happened after the commands executed?**
Received an error message with a bunch of goroutine information
```
[...]

PR #10666, opened 2021-01-27
Allow attaching same external load balancer to multiple instance groups

hakman:
Similar to https://github.com/kubernetes/kops/pull/10335
Fixes: https://github.com/kubernetes/kops/issues/10664

/cc @rifelpet @rdrgmnzs

### DIFF
--- a/tests/integration/update_cluster/externallb/kubernetes.tf
+++ b/tests/integration/update_cluster/externallb/kubernetes.tf
@@ -86,7 +86,7 @@ resource "aws_autoscaling_group" "master-us-test-1a-masters-externallb-example-c
     id      = aws_launch_template.master-us-test-1a-masters-externallb-example-com.id
     version = aws_launch_template.master-us-test-1a-masters-externallb-example-com.latest_version
   }
-  load_balancers      = ["my-external-elb-2", "my-external-elb-3"]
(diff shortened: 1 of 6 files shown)


tests/integration/update_cluster/externallb/kubernetes.tf
```diff
@@ -86,7 +86,7 @@ resource "aws_autoscaling_group" "master-us-test-1a-masters-externallb-example-c
     id      = aws_launch_template.master-us-test-1a-masters-externallb-example-com.id
     version = aws_launch_template.master-us-test-1a-masters-externallb-example-com.latest_version
   }
-  load_balancers      = ["my-external-elb-2", "my-external-elb-3"]
+  load_balancers      = ["master-us-test-1a.masters.externallb.example.com-my-external-elb-1", "master-us-test-1a.masters.externallb.example.com-my-external-elb-2", "master-us-test-1a.masters.externallb.example.com-my-external-elb-3"]
```
rifelpet, line 89: I wouldn't expect the terraform and cloudformation outputs to be changing. We don't want to change the names of the load balancers we're attaching to the ASGs, only the names of their fi tasks

hakman: True. These latest changes should do the trick.
I guess now this has to its way into 1.19 and 1.18?

Issue #10647, opened 2021-01-23
Don't allow invalid paths in `set cluster`


rifelpet:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

Currently `kops set cluster` accepts paths that aren't valid. 

`kops set cluster foo.k8s.local spec.foo=invalid`

This can lead to confusion when paths contain typos. We should return an error when a path doesn't match an actual field in the cluster spec.

**2. Feel free to provide a design supporting your feature request.**

bharath-123:
Have been a victim of this many a times, especially when setting kubernetes versions :( I think this would require messing around with stuff in reflect utils. Will take this up.

PR #10692, opened 2021-01-30
Throw error if path being set by kops set is not present in struct

bharath-123:
Currently, when using kops set cluster <cluster_name> <path>=<value>, if the path is not present it fails silently. eg: 
`kops set cluster test-cluster.k8s.local spec.abc.def.ghi=jkl ` will not throw any error despite the fields not being present in the kops.Cluster struct. This can lead to confusion and mistakes. 

This PR explicitly throws an error if a field being set using `kops set` does not exist in the associated struct. 

Fixes: https://github.com/kubernetes/kops/issues/10647

Depends on: https://github.com/kubernetes/kops/pull/10690

### DIFF
--- a/util/pkg/reflectutils/access.go
+++ b/util/pkg/reflectutils/access.go
@@ -31,6 +31,8 @@ func SetString(target interface{}, targetPath string, newValue string) error {
 		return fmt.Errorf("cannot parse field path %q: %w", targetPath, err)
 	}
 
+	var fieldSet = false
+
 	visitor := func(path *FieldPath, field *reflect.StructField, v reflect.Value) error {
 		if !targetFieldPath.HasPrefixMatch(path) {
 			return nil
@@ -45,6 +47,7 @@ func SetString(target interface{}, targetPath string, newValue string) error {
 				return fmt.Errorf("cannot set field %q: %v", path, err)
 			}
 
+			fieldSet = true
 			return nil
 		}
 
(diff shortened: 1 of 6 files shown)


util/pkg/reflectutils/access.go
```diff
@@ -45,6 +47,7 @@ func SetString(target interface{}, targetPath string, newValue string) error {
 				return fmt.Errorf("cannot set field %q: %v", path, err)
 			}
 
+			fieldSet = true
```
hakman, line 50: Seems that this is the only place where this is set to `true` and the function exists immediately after. Maybe I am missing something?

bharath-123: Yup. So the way this works as per my understanding is that, we recurse into the structure searching for the targetFieldPath. We set the targetFieldPath to the given value if we find an exact match in the structure. If during searching we see that the struct is `nil` i.e if we are setting `spec.networking.cilium.prometheusPort` but `spec.networking.cilium == nil` then it will autopopulate the structure value with `spec.networking.cilium` . The recursion begins again and we end up finding an exact match after which we set the field to the value. 

In the case of a field not being present, we ll never get a match and just recurse the entire struct and exit silently.

hakman: Ah, missed the recursive function part. 😄

rifelpet:
/lgtm
/approve

Thanks for improving this!

PR #10695, opened 2021-01-30
Add support for cilium 1.9

olemarkus:


upup/pkg/fi/cloudup/bootstrapchannelbuilder/cilium.go
```diff
@@ -46,12 +43,27 @@ func addCiliumAddon(b *BootstrapChannelBuilder, addons *api.Addons) {
...
+				addons.Spec.Addons = append(addons.Spec.Addons, &api.AddonSpec{
+					Name:               fi.String(key),
+					Version:            fi.String(version),
+					Selector:           networkingSelector(),
+					Manifest:           fi.String(location),
+					Id:                 id,
+					NeedsRollingUpdate: "all",
+				})
+			}
+		} else if ver.Minor == 9 {
```
rifelpet, line 61: Previously there was a lone `} else {`, was this acting as a default cilium version if the api field isn't set? If so, we should probably keep the `} else {` here so that 1.9 is installed by default.

olemarkus: The code expects that the code running prior does the right thing. Added some guards against that not being the case now.

hakman: Was just thinking about that error handling.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template
```diff
@@ -0,0 +1,977 @@
...
+  name: hubble-relay-config
+  namespace: kube-system
+data:
+  config.yaml: |
+    peer-service: unix:///var/run/cilium/hubble.sock
+    listen-address: :4245
+    dial-timeout: 
+    retry-timeout: 
+    sort-buffer-len-max: 
+    sort-buffer-drain-timeout:
```
hakman, line 225: Could you remove the trailing spaces?

olemarkus: Seems pointless to me to set empty keys like that. Removed them.

pkg/model/components/cilium.go
```diff
@@ -39,7 +39,7 @@ func (b *CiliumOptionsBuilder) BuildOptions(o interface{}) error {
 	}
 
 	if c.Version == "" {
-		c.Version = "v1.8.4"
+		c.Version = "v1.9.3"
```
hakman, line 42: Maybe now is a good time to use latest version?
```suggestion
		c.Version = "v1.9.4"
```

PR #10696, opened 2021-01-30
Update code reference links in docs

bharath-123:
Update out of date code reference links in the [how it works](https://kops.sigs.k8s.io/contributing/how_it_works/) doc. These out of date links had me confused a bit while reading the docs.

### DIFF
--- a/docs/contributing/how_it_works.md
+++ b/docs/contributing/how_it_works.md
@@ -36,7 +36,7 @@ List()
 
 The `kops` API is a definition of struct members in Go code found [here](https://github.com/kubernetes/kops/tree/master/pkg/apis/kops). The `kops` API does **NOT** match the command line interface (by design). We use the native Kubernetes API machinery to manage versioning of the `kops` API.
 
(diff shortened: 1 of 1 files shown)


hakman:
I don't think it's sustainable to change these line numbers all the time. I thin wither using the current master tag or keeping master and removing line would be better (once you get to the right file, a quick search is not that hard).
What do you think?

bharath-123:
Yeah that was in my mind while raising this PR. I think we should point to the right file and be more explicit in the docs about the struct name and method name so that the reader knows what to search for at least.

olemarkus:
It may be better to replace e.g https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/apply_cluster.go#L89 with a link to https://pkg.go.dev/k8s.io/kops/upup/pkg/fi/cloudup#ApplyClusterCmd

These docs has up-to-date line references.

bharath-123:
Thats a good idea @olemarkus! Thanks! Will fix up the links to point to the pkg.go.dev references!

PR #10697, opened 2021-01-30
Kubetest2 - terraform support

rifelpet:
This adds support for `--target terraform` and applying/destroying the cluster using terraform.

The specified terraform version is downloaded and unzipped each run, ideally we could cache this or use tfenv or something.

### DIFF
--- a/tests/e2e/pkg/util/zip.go
+++ b/tests/e2e/pkg/util/zip.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"archive/zip"
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// UnzipToTempDir will decompress the provided bytes into a temporary directory that is returned
+func UnzipToTempDir(data []byte) (string, error) {
+	reader, err := zip.NewReader(bytes.NewReader(data), int64(len(data)))
+	if err != nil {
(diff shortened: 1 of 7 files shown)


tests/e2e/pkg/util/zip.go
```diff
@@ -0,0 +1,76 @@
...
+
+		fileReader, err := r.Open()
+		if err != nil {
+			return "", err
+		}
+
+		_, err = io.Copy(output, fileReader)
+
+		output.Close()
+		fileReader.Close()
```
bharath-123, line 69: Should we put this in a defer stack? Seems to me if `r.open` fails then `output` will not be closed and can lead to good number of open files in the system.

rifelpet: If we defer these, they wont be closed until UnzipToTempDir returns which means a zip containing many files could lead to a stack overflow. Given that the kubetest2-kops process will exit if `r.open` fails, I think not deferring these is acceptable.

bharath-123: Ah I didn't have enough context on kubetest2-kops process. The stack overflow part makes sense too. Thanks for clearing it up :D

rifelpet:
I backed out the makefile update so this should be ready for review now.

I noticed an interesting behavior that I'm hoping is the source of some of our test job problems...

When I was porting over the kops commands from kubetest1 to kubetest2 I copied the `validate cluster` command as-is and didn't copy the node readiness loop that ensures nodes arent flapping their readiness:

https://github.com/kubernetes/test-infra/blob/dee9d4bff3ffb5130c88dd3c122ec3c3e4f95272/kubetest/kops.go#L497-L500

We built that logic into `validate cluster` but didnt update kubetest1 to use it instead of its own node readiness check.

The k8s e2e test framework has a [--num-nodes](https://github.com/kubernetes/kubernetes/blob/faecb196815e248d3ecfb03c680a4507229c2a56/test/e2e/framework/test_context.go#L348) flag that is used in some tests and defaults to the [number of Ready and Schedulable nodes](https://github.com/kubernetes/kubernetes/blob/faecb196815e248d3ecfb03c680a4507229c2a56/test/e2e/e2e.go#L261-L266) fetched at startup.

Kops not waiting for nodes to stop flapping their readiness caused the e2e framework to incorrectly detect the number of nodes, and some tests were being skipped as a result. See [Line 7335 here](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-aws-distro-imageubuntu2004/1361397364752912384#1:build-log.txt%3A7335). I'm hoping the `--count 10` on `kops validate cluster` will fix this and hopefully some other test related issues.

PR #10700, opened 2021-01-31
Fix rendering issue created by #10414

avdhoot:
This might break upgrade if cluster spec have `prometheusMetricsEnabled: true`. I have validated this by upgrading our cluster. This is caused by #10414

### DIFF
--- a/upup/models/bindata.go
+++ b/upup/models/bindata.go
@@ -37449,7 +37449,7 @@ spec:
           name: calico-typha
           protocol: TCP
         {{- if .Networking.Calico.TyphaPrometheusMetricsEnabled }}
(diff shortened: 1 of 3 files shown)


hakman (changes requested):
Thanks @avdhoot. Please also bump the manifest version here to `3.17.1-kops.3`:
https://github.com/kubernetes/kops/blob/40ae752e46182e60568eceff3d94dca39f79165a/upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go#L744

hakman:
Looks good. One more thing, if you don't mind, could you squash the 2 commits?

avdhoot:
@hakman Sorry took time to understand local `squash`. Thanks.

PR #10702, opened 2021-01-31
Increase IMDSv2 hop limit on control plane nodes

olemarkus:
Non-hostNetworking fails to talk to the instance metadata otherwise. Breaking e.g CSI controller

See https://github.com/kubernetes/cloud-provider-aws/issues/169

Note: This should work for kubenet, but other CNIs may need other hop limits. E.g with Cilium, I had to use 3 to get this to work. Will have to investigate a bit more first. Meanwhile this should fix the broken test.

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -679,7 +679,9 @@ func setupMasters(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap 
 				k8sVersion, err := version.ParseKubernetesVersion(cluster.Spec.KubernetesVersion)
 				if err == nil && version.IsKubernetesGTE("1.18", *k8sVersion) {
 					if g.Spec.InstanceMetadata == nil {
-						g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{}
+						g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+							HTTPPutResponseHopLimit: fi.Int64(2),
+						}
 					}
 					g.Spec.InstanceMetadata.HTTPTokens = fi.String(ec2.LaunchTemplateHttpTokensStateRequired)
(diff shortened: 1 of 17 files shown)


upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -679,7 +679,9 @@ func setupMasters(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap
 				k8sVersion, err := version.ParseKubernetesVersion(cluster.Spec.KubernetesVersion)
 				if err == nil && version.IsKubernetesGTE("1.18", *k8sVersion) {
 					if g.Spec.InstanceMetadata == nil {
-						g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{}
+						g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+							HTTPPutResponseHopLimit: fi.Int64(2),
+						}
```
hakman, line 684: Should check if HTTPPutResponseHopLimit is already set by user. Probably all this logic has to be updated a bit.

hakman: Also, should be IMDSv2, if I'm not mistaken.

olemarkus: I only set hop limit if InstanceMetadata is nil, so user should not have set anything. IMDSv2 is used when `httpTokens: required` and this is set ... somewhere else.

hakman: `IMDSv2` meant in title.
Here there is a case when HTTPTokens can be set, but HTTPPutResponseHopLimit won't be set. Would not happen in practice but code should be consistent.

olemarkus: The goal of this one was just to see if we can get the tests passing. We need to add more logic here soon to get CNI (well cilium anyway) working.

PR #10712, opened 2021-02-02
Docs: Fix ServiceAccountVolume proposed configuration for Istio

dntosas:
Seems that after merging https://github.com/kubernetes/kops/pull/9534 there is no need to specify paths for key-files for service Account Token Volume Projection. 

Instead, the current configuration proposed in docs is making Istio unable to verify JWT token with errors.

In this commit, we remove the paths, letting Kops handles the default ones that are working as expected.


olemarkus:
/lgtm
/approve

Thanks for catching!

PR #10722, opened 2021-02-04
Apiserver nodes

olemarkus:
This PR introduces an additional node role that only runs kube-apiserver.
The control plane nodes still run kube-apiserver for bootstrapping the cluster and get dns-controller running. Then the additional api-servers join the cluster at the same time as worker nodes.

In order to distinguish API vs etcd clusters, new etcd cluster-specific DNS entries are created and used by apiserver nodes.

Current implementation grants apiserver nodes access to S3 to self-provision certs. Technically kube-controller could provision these certs, but that would require changes to kube-controller so it only provisions kube-apiserver certs to apiserver role.

Apiservers also reuse the master SG, but should really have a tighter dedicated SG at some point.

Future improvements could be being able to attach ASG scaling policies to it so these ASG scales with load.


seh:
> Future improvements could be being able to attach ASG scaling policies to it so these ASG scales with load.

I've done this in the past based on CloudWatch metrics for CPU utilization. It was hard to get it right, given the many-minute delay in new machines starting up and being ready to serve. I found that often the CPU usage among the three or more API servers was inconsistent; one of them may be much more busy than the rest. I had been using the average utilization across the machines in the ASGs.

Also, though I may have been doing something incorrectly, I recall that setting the CPU usage threshold correctly in CloudWatch required me to know how many CPUs were available on these machines. It wasn't possible to express something like "60% of the available CPUs" so that the rules would work irrespective of the machine type.

If I recall correctly, the policy worked something like this:
- If the average CPU utilization was between 25% and 40%, increase the ASG size by 33%.
- If the average CPU utilization was 40% or higher, increase the ASG size by 50%.
- If the average CPU utilization was 10% or lower for 20 minutes, decrease the ASG size by one.

olemarkus:
> > Future improvements could be being able to attach ASG scaling policies to it so these ASG scales with load.
> 
> I've done this in the past based on CloudWatch metrics for CPU utilization. It was hard to get it right, given the many-minute delay in new machines starting up and being ready to serve. I found that often the CPU usage among the three or more API servers was inconsistent; one of them may be much more busy than the rest. I had been using the average utilization across the machines in the ASGs.
> 
> Also, though I may have been doing something incorrectly, I recall that setting the CPU usage threshold correctly in CloudWatch required me to know how many CPUs were available on these machines. It wasn't possible to express something like "60% of the available CPUs" so that the rules would work irrespective of the machine type.
> 

Yeah, so where I work we never used ASG policies at all, but wrote our own custom controller that watch various sources (in our case, is there an upcoming game in an important league etc) or prometheus. 

Since we have metrics-server as an addon, we could let kops-controller just watch the metrics API (or scrape prom metrics such as API  and set the desired flag.

Either way, that is for a follow-up :)

cmd/kops/rollingupdatecluster.go
```diff
@@ -189,7 +189,7 @@ func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().DurationVar(&options.PostDrainDelay, "post-drain-delay", options.PostDrainDelay, "Time to wait after draining each node")
 	cmd.Flags().BoolVarP(&options.Interactive, "interactive", "i", options.Interactive, "Prompt to continue after each instance is updated")
 	cmd.Flags().StringSliceVar(&options.InstanceGroups, "instance-group", options.InstanceGroups, "List of instance groups to update (defaults to all if not specified)")
-	cmd.Flags().StringSliceVar(&options.InstanceGroupRoles, "instance-group-roles", options.InstanceGroupRoles, "If specified, only instance groups of the specified role will be updated (e.g. Master,Node,Bastion)")
+	cmd.Flags().StringSliceVar(&options.InstanceGroupRoles, "instance-group-roles", options.InstanceGroupRoles, "If specified, only instance groups of the specified role will be updated (e.g. Master,Node,Bastion,Apiserver)")
```
seh, line 192: Consider capitalizing the new one differently:
```suggestion
	cmd.Flags().StringSliceVar(&options.InstanceGroupRoles, "instance-group-roles", options.InstanceGroupRoles, "If specified, only instance groups of the specified role will be updated (e.g. Master,Node,Bastion,APIServer)")
```
I suppose this is predicated on whether we think that the term is "apiserver" (as occurs sometimes in Kubernetes code, but not often in documentation) or "API Server." I favor the latter.

docs/cli/kops_create_instancegroup.md
```diff
@@ -36,7 +36,7 @@ kops create instancegroup [flags]
       --edit             If true, an editor will be opened to edit default values. (default true)
   -h, --help             help for instancegroup
   -o, --output string    Output format. One of json|yaml
-      --role string      Type of instance group to create (Node,Master,Bastion) (default "Node")
+      --role string      Type of instance group to create (Node,Master,Apiserver,Bastion) (default "Node")
```
seh, line 39: Consider capitalizing the new one differently:
```suggestion
      --role string      Type of instance group to create (Node,Master,APIServer,Bastion) (default "Node")
```

docs/cli/kops_rolling-update_cluster.md
```diff
@@ -75,7 +75,7 @@ kops rolling-update cluster [flags]
       --force                          Force rolling update, even if no changes
   -h, --help                           help for cluster
       --instance-group strings         List of instance groups to update (defaults to all if not specified)
-      --instance-group-roles strings   If specified, only instance groups of the specified role will be updated (e.g. Master,Node,Bastion)
+      --instance-group-roles strings   If specified, only instance groups of the specified role will be updated (e.g. Master,Node,Bastion,Apiserver)
```
seh, line 78: Consider capitalizing the new one differently:
```suggestion
      --instance-group-roles strings   If specified, only instance groups of the specified role will be updated (e.g. Master,Node,Bastion,APIServer)
```

cmd/kops/rollingupdatecluster.go
seh: Are these roles ordered from most to least likely to be pertinent here? I see that the order differs here, in _kops_create_instancegroup.md_, and in _kops_rolling-update_cluster.md_.

docs/cli/kops_create_instancegroup.md
seh: This one's role order is the odd one compared to those above and below.

nodeup/pkg/model/context.go
```diff
@@ -57,6 +57,9 @@ type NodeupModelContext struct {
 	// IsMaster is true if the InstanceGroup has a role of master (populated by Init)
 	IsMaster bool
 
+	// HasAPIServer is true if the InstanceGroup has a role of master or apiserver (pupulated by Init)
```
seh, line 60:
```suggestion
	// HasAPIServer is true if the InstanceGroup's machines run the Kubernetes API Server, having a role of either master or apiserver (populated by Init)
```

nodeup/pkg/model/kube_apiserver.go
```diff
@@ -316,19 +316,30 @@ func (b *KubeAPIServerBuilder) buildPod() (*v1.Pod, error) {
...
 
+	var mainCluster, eventsCluster string
+	if b.IsMaster {
+		mainCluster = "https://127.0.0.1:4001"
+		eventsCluster = "https://127.0.0.1:4002"
+	} else {
+		host := b.Cluster.ObjectMeta.Name
+		mainCluster = "https://main.etcd." + host + ":4001"
+		eventsCluster = "https://events.etcd." + host + ":4002"
+
```
seh, line 327: Remove blank line.

nodeup/pkg/model/kube_apiserver.go
```diff
@@ -316,19 +316,30 @@ func (b *KubeAPIServerBuilder) buildPod() (*v1.Pod, error) {
 		}
 	}
 
+	var mainCluster, eventsCluster string
```
seh, line 319: Consider renaming these to `mainEtcdCluster` and `eventsEtcdCluster` for clarifying which kind of cluster we mean.

nodeup/pkg/model/kubelet.go
```diff
@@ -490,6 +491,10 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 			// (Even though the value is empty, we still expect <Key>=<Value>:<Effect>)
 			c.Taints = append(c.Taints, nodelabels.RoleLabelMaster16+"=:"+string(v1.TaintEffectNoSchedule))
 		}
+		if len(c.Taints) == 0 && isAPIServer {
+			// (Even though the value is empty, we still expect <Key>=<Value>:<Effect>)
+			c.Taints = append(c.Taints, nodelabels.RoleLabelAPIServer16+"=:"+string(v1.TaintEffectNoSchedule))
```
seh, line 496: What does the "16" suffix mean here and above on line 492?

nodeup/pkg/model/secrets.go
```diff
@@ -70,7 +70,7 @@ func (b *SecretBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	// if we are not a master we can stop here
```
seh, line 72:
```suggestion
	// If we do not run the Kubernetes API server we can stop here.
```

pkg/apis/kops/instancegroup.go
```diff
@@ -59,12 +58,15 @@ const (
 	InstanceGroupRoleNode InstanceGroupRole = "Node"
 	// InstanceGroupRoleBastion is a bastion role
 	InstanceGroupRoleBastion InstanceGroupRole = "Bastion"
+	// InstanceGroupRoleAPIServer is an API server role
+	InstanceGroupRoleAPIServer InstanceGroupRole = "Apiserver"
```
seh, line 62: Consider capitalizing the new one differently:
```suggestion
	InstanceGroupRoleAPIServer InstanceGroupRole = "APIServer"
```

pkg/apis/kops/instancegroup.go
```diff
@@ -278,27 +280,32 @@ func (g *InstanceGroup) IsMaster() bool {
...
 		return true
-	case InstanceGroupRoleNode:
-		return false
-	case InstanceGroupRoleBastion:
+	default:
 		return false
+	}
+}
+
+// IsAPIServerOnly checks if instanceGroup only runs the API Server
```
seh, line 288:
```suggestion
// IsAPIServerOnly checks if instanceGroup runs only the API Server
```

pkg/apis/kops/util/labels.go
```diff
@@ -25,9 +25,15 @@ func GetNodeRole(node *v1.Node) string {
 	if _, ok := node.Labels["node-role.kubernetes.io/master"]; ok {
 		return "master"
 	}
+	if _, ok := node.Labels["node-role.kubernetes.io/control-plane"]; ok {
+		return "control-plane"
+	}
 	if _, ok := node.Labels["node-role.kubernetes.io/node"]; ok {
 		return "node"
 	}
+	if _, ok := node.Labels["node-role.kubernetes.io/api-server"]; ok {
```
seh, line 34: This one is hard to settle. The API Server is part of the control plane, but here we're projecting to only a single role, so even though a node could be labeled with both, we can only take the first one that matches.

pkg/instancegroups/rollingupdate.go
```diff
@@ -160,6 +163,21 @@ func (c *RollingUpdateCluster) RollingUpdate(groups map[string]*cloudinstances.C
 		}
 	}
 
+	// Upgrade API servers
+	{
+		for k := range apiServerGroups {
+			results[k] = fmt.Errorf("function panic nodes")
```
seh, line 169:
```suggestion
			results[k] = errors.New("function panic nodes")
```
That error message is confusing.

pkg/model/iam/iam_builder.go
```diff
@@ -235,6 +235,54 @@ func (b *PolicyBuilder) BuildAWSPolicy() (*Policy, error) {
 	return p, nil
 }
 
+// BuildAWSPolicy generates a custom policy for a Kubernetes master.
```
seh, line 238:
```suggestion
// BuildAWSPolicy generates a custom policy for a machine running the Kubernetes API Server.
```

pkg/model/iam/subject.go
```diff
@@ -68,13 +77,13 @@ func BuildNodeRoleSubject(igRole kops.InstanceGroupRole) (Subject, error) {
 	switch igRole {
 	case kops.InstanceGroupRoleMaster:
 		return &NodeRoleMaster{}, nil
-
+	case kops.InstanceGroupRoleAPIServer:
+		return &NodeRoleAPIServer{}, nil
 	case kops.InstanceGroupRoleNode:
 		return &NodeRoleNode{}, nil
```
seh, line 84: Consider removing this blank line.

pkg/model/names.go
```diff
@@ -36,7 +36,7 @@ func (b *KopsModelContext) SecurityGroupName(role kops.InstanceGroupRole) string
 		return "bastion." + b.ClusterName()
 	case kops.InstanceGroupRoleNode:
 		return "nodes." + b.ClusterName()
-	case kops.InstanceGroupRoleMaster:
+	case kops.InstanceGroupRoleMaster, kops.InstanceGroupRoleAPIServer:
 		return "masters." + b.ClusterName()
```
seh, line 40: This is an interesting move. We could introduce another security group for API servers, and attach it to both control plane machines and API server-only machines, though at the moment I can't think of firewall rules that the former require separately from the latter.

pkg/model/names.go
```diff
@@ -136,6 +138,8 @@ func (b *KopsModelContext) IAMName(role kops.InstanceGroupRole) string {
 	switch role {
 	case kops.InstanceGroupRoleMaster:
 		return "masters." + b.ClusterName()
+	case kops.InstanceGroupRoleAPIServer:
+		return "apiserver." + b.ClusterName()
```
seh, line 142: Should that be "apiservers" (pluralized) to align with the rest of the role names here?

pkg/validation/validate_cluster.go
```diff
@@ -364,6 +364,18 @@ func (v *ValidationCluster) validateNodes(cloudGroups map[string]*cloudinstances
 				}
 
 				v.Nodes = append(v.Nodes, n)
+			} else if n.Role == "apiserver" {
+				if !ready {
+					v.addError(&ValidationError{
+						Kind:          "Apiserver",
```
seh, line 370:
```suggestion
						Kind:          "APIServer",
```

pkg/validation/validate_cluster.go
```diff
@@ -364,6 +364,18 @@ func (v *ValidationCluster) validateNodes(cloudGroups map[string]*cloudinstances
...
+						Name:          node.Name,
+						Message:       fmt.Sprintf("apiserver %q is not ready", node.Name),
```
seh, line 372: Maybe:
```suggestion
						Message:       fmt.Sprintf("API Server %q is not ready", node.Name),
```

seh (commented):
I offered some copyediting suggestions, and asked a few clarifying questions.

It's easier to consider the things that you did change than to recognize something else missed here that also needs to change. What about validating that cluster includes at least one "master" or "control-plane" _InstanceGroup_, as opposed to having only "api-server" _InstanceGroups_?

pkg/model/names.go
rifelpet: The masters security group allows inbound etcd traffic which wouldn't be applicable on apiserver hosts. I believe it is exposed in a wider range than just the etcd ports themselves, so it could potentially impact other use cases. We also allow all traffic between control plane hosts which isn't applicable for dedicated apiserver hosts.

while it would add complexity to this PR (or a followup), given the pain of revoking security group rules down the road I think it would be best to create a dedicated apiserver security group now (or in a followup).

pkg/validation/validate_cluster.go
olemarkus: Unfortunately, `kops rolling-update cluster --instance-group-role <role>` title-case the input. so the consistent user-facing name need to be `Apiserver`

pkg/model/names.go
olemarkus: This was a short-cut, I must admit.

Right now, I am thinking
1.  merge this
2. create an e2e test that verifies this is working
3. tighten the rules

3 definitely need to happen before beta or we need to featureflag this. I should have plenty of time to fix this before beta though.

nodeup/pkg/model/kubelet.go
olemarkus: Added as of kops/k8s 1.16

pkg/apis/kops/util/labels.go
olemarkus: This is the crux I think.
Should we add this role also to the "main" CP nodes? They would then now have three roles. four if they are spots.

Curious what other k8s implementations do here.

pkg/instancegroups/rollingupdate.go
olemarkus: Error is copypasted from the other roles

pkg/apis/kops/util/labels.go
seh: I think we should, as this style of annotation deliberately accommodates any number of roles per node. It would be possible to have a machine run, say, the controller manager and scheduler, and _not_ run the API server, so these multiple roles occurring together make sense to clarify what's running there.

pkg/apis/kops/validation/instancegroup_test.go
```diff
@@ -300,3 +300,69 @@ func TestIGCloudLabelIsIGName(t *testing.T) {
...
 }
+
+func TestValidInstanceGroup(t *testing.T) {
+	grid := []struct {
+		IG             *kops.InstanceGroup
+		ExpectedErrors int
+		Description    string
+	}{
+		{
+
```
seh, line 311: Remove blank line.

seh (approved):
My remaining suggestions are minimal.

I think we need to figure out the security group thing before going too much further.

pkg/apis/kops/util/labels.go
olemarkus: Well. _This_ particular part of the code doesn't really deal with multiple roles. It is just used by some funny parts of `dns-controller`

olemarkus:
I don't think SG is too hard. It is basically node + the external access stuff + etcd client ports.
But preferably that comes later.

olemarkus:
This is as much as makes sense for a single PR I think.

I will have a look at SGs soon, but as we'll go from 2^2 to 3^2 SG relationships I want to take the opportunity to refactor those builders a bit.

nodeup/pkg/model/kube_apiserver.go
```diff
@@ -316,19 +316,29 @@ func (b *KubeAPIServerBuilder) buildPod() (*v1.Pod, error) {
...
 	}
 
+	var mainEtcdCluster, eventsEtcdCluster string
+	if b.IsMaster {
+		mainEtcdCluster = "https://127.0.0.1:4001"
+		eventsEtcdCluster = "https://127.0.0.1:4002"
+	} else {
+		host := b.Cluster.ObjectMeta.Name
+		mainEtcdCluster = "https://main.etcd." + host + ":4001"
+		eventsEtcdCluster = "https://events.etcd." + host + ":4002"
```
rifelpet, line 326: I'm wondering if these should use `b.Cluster.Spec.MasterPublicName` or `b.Cluster.Spec.MasterInternalName` instead of `b.Cluster.ObjectMeta.Name`

pkg/model/components/etcdmanager/model.go
```diff
@@ -258,7 +258,7 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec) (*v1.Pod
 
 	etcdInsecure := !b.UseEtcdTLS()
 
-	clientHost := "__name__"
+	clientHost := etcdCluster.Name + ".etcd." + b.ClusterName()
```
rifelpet, line 261: same as above, consider using MasterPublicName or MasterInternalName API field here?

nodeup/pkg/model/kube_apiserver.go
olemarkus: so `main.etcd.api.test.example.com`?
If they do not directly relate to api nodes (and in the future, may not even run api servers), does that make sense?

rifelpet: hm you're right, having `api` in there doesn't make sense. Maybe down the road we expose a third control plane name field in the clusterspec. If users are wanting to override the default master name fields, they'll likely want to override these etcd names as well.

pkg/model/iam/iam_builder.go
```diff
@@ -235,6 +235,54 @@ func (b *PolicyBuilder) BuildAWSPolicy() (*Policy, error) {
...
+func (r *NodeRoleAPIServer) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
+	resource := createResource(b)
```
justinsb, line 240: Can we extract a common "control plane" method which both NodeRoleMaster and NodeRoleAPIServer call into (likely passing their role)?  Otherwise there's a lot of duplication here.

pkg/nodelabels/builder.go
```diff
@@ -22,13 +22,15 @@ import (
...
+	RoleLabelName15           = "kubernetes.io/role"
+	RoleLabelName16           = "kubernetes.io/role"
+	RoleMasterLabelValue15    = "master"
+	RoleAPIServerLabelValue15 = "api-server"
+	RoleNodeLabelValue15      = "node"
 
-	RoleLabelMaster16 = "node-role.kubernetes.io/master"
-	RoleLabelNode16   = "node-role.kubernetes.io/node"
+	RoleLabelMaster16    = "node-role.kubernetes.io/master"
+	RoleLabelAPIServer16 = "node-role.kubernetes.io/api-server"
```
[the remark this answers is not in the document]
justinsb: If we're in doubt, we could just label with "control-plane" and add labels in future so we can easily distinguish etcd and non-etcd.  It's much easier to add later than it is to take away something once we've shipped it.

justinsb:
I really like this.  I think the biggest uncertainty I had was around node labels & taints.  I *suggest* that unless we are sure, we should continue to label these with the same labels / taints / AWS tags that we give to masters today (moving towards the preferred `control-plane` terminology).  We can always add additional labels in future, we can't ever take them away.

One thing I really wasn't sure about was whether we intended for these to run e.g. kube-scheduler and kube-controller-manager - and other controllers as they exist.  i.e. are these really "apiserver only" nodes, or are these "not etcd" nodes?

nodeup/pkg/model/kube_apiserver.go
olemarkus: I think there were someone wanting this field also for the newly-introduced kops-controller DNS. I can try to use the same approach as we did there (which was replace "api" with "kops-controller", I believe).

The introduction of the suffix field can be done in a follow-up.

pkg/model/components/etcdmanager/model.go
olemarkus: We didn't use `__name__` for the cilium etcd cluster. So this sort of just followed that logic.

nodeup/pkg/model/etcd_manager_tls.go
```diff
@@ -32,7 +32,7 @@ var _ fi.ModelBuilder = &EtcdManagerTLSBuilder{}
 
 // Build is responsible for TLS configuration for etcd-manager
 func (b *EtcdManagerTLSBuilder) Build(ctx *fi.ModelBuilderContext) error {
-	if !b.IsMaster || !b.UseEtcdManager() {
+	if !b.HasAPIServer || !b.UseEtcdManager() {
```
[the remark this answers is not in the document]
olemarkus: Partly right. The etcd CAs doesn't serve much purpose and should be skipped. But we need to get down to `buildKubeAPIServerKeypair` so we have certs we can use to authenticate to etcd.

pkg/nodelabels/builder.go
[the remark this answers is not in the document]
olemarkus: We are inventing these here. I couldn't find anyone else with a similar setup, and also discussed a bit with @rifelpet what they had been using.

I think having the role visible in `kubectl get nodes` is useful, but it is not a must.

More importantly is that APIserver nodes may not have all the IAM privileges that the "full" CP nodes have. So it could break things that use this label as node selector.

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -184,6 +185,10 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, cl
 		allErrs = append(allErrs, ValidateMasterInstanceGroup(g, cluster)...)
 	}
 
+	if g.Spec.Role == kops.InstanceGroupRoleAPIServer && kops.CloudProviderID(cluster.Spec.CloudProvider) != kops.CloudProviderAWS {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "role"), "Apiserver role only supported on AWS"))
```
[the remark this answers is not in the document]
olemarkus: With this PR, we only modify cloudup tasks for AWS (SGs, IAM etc)

olemarkus:
> One thing I really wasn't sure about was whether we intended for these to run e.g. kube-scheduler and kube-controller-manager - and other controllers as they exist. i.e. are these really "apiserver only" nodes, or are these "not etcd" nodes?

My intention was to only run kube-apiserver on this. I see the value of running etcd on dedicated nodes too as there is still a risk that overloaded apiserver may cause etcd to become slower. The main blocker for that is that etcd-manager relies on dns-controller to provision its DNS. Scheduler and Controller-Manager are leader-elected, so running those on a larger set of nodes doesn't help a lot.

nodeup/pkg/model/kube_apiserver.go
olemarkus: Seems to have remembered incorrectly. kops-controller does `dns.alpha.kubernetes.io/internal: kops-controller.internal.{{ ClusterName }}`

So maybe stick with current, but add support for configurable prefix later?

olemarkus:
All changes that would affect current clusters is now behind feature flag (as proven by passing the `origin/master` golden outputs).

Ready to merge from my side.

docs/cli/kops_rolling-update_cluster.md
```diff
@@ -75,7 +75,7 @@ kops rolling-update cluster [flags]
       --force                          Force rolling update, even if no changes
   -h, --help                           help for cluster
       --instance-group strings         List of instance groups to update (defaults to all if not specified)
-      --instance-group-roles strings   If specified, only instance groups of the specified role will be updated (e.g. Master,Node,Bastion)
+      --instance-group-roles strings   If specified, only instance groups of the specified role will be updated (e.g. Master,Apiserver,Node,Bastion)
```
seh, line 78: Why do we write "Apiserver" for role name for _kops rolling-update cluster_, but write "APIServer" for the role name for _kops create instancegroup_?

docs/cli/kops_create_instancegroup.md
```diff
@@ -36,7 +36,7 @@ kops create instancegroup [flags]
       --edit             If true, an editor will be opened to edit default values. (default true)
   -h, --help             help for instancegroup
   -o, --output string    Output format. One of json|yaml
-      --role string      Type of instance group to create (Node,Master,Bastion) (default "Node")
+      --role string      Type of instance group to create (Master,APIServer,Node,Bastion) (default "Node")
```
seh, line 39: Why do we write "APIServer" for the role name for _kops create instancegroup_, but write "Apiserver" for role name for _kops rolling-update cluster_?

cmd/kops/rollingupdatecluster.go
```diff
@@ -189,7 +188,7 @@ func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().DurationVar(&options.PostDrainDelay, "post-drain-delay", options.PostDrainDelay, "Time to wait after draining each node")
 	cmd.Flags().BoolVarP(&options.Interactive, "interactive", "i", options.Interactive, "Prompt to continue after each instance is updated")
 	cmd.Flags().StringSliceVar(&options.InstanceGroups, "instance-group", options.InstanceGroups, "List of instance groups to update (defaults to all if not specified)")
-	cmd.Flags().StringSliceVar(&options.InstanceGroupRoles, "instance-group-roles", options.InstanceGroupRoles, "If specified, only instance groups of the specified role will be updated (e.g. Master,Node,Bastion)")
+	cmd.Flags().StringSliceVar(&options.InstanceGroupRoles, "instance-group-roles", options.InstanceGroupRoles, "If specified, only instance groups of the specified role will be updated (e.g. Master,Apiserver,Node,Bastion)")
```
seh, line 191: Why do we write "Apiserver" for role name for _kops rolling-update cluster_, but write "APIServer" for the role name for _kops create instancegroup_?

nodeup/pkg/model/etcd_manager_tls.go
```diff
@@ -32,16 +32,25 @@ var _ fi.ModelBuilder = &EtcdManagerTLSBuilder{}
...
+	}
+
 	for _, k := range []string{"main", "events"} {
 		d := "/etc/kubernetes/pki/etcd-manager-" + k
 
 		keys := make(map[string]string)
-		keys["etcd-manager-ca"] = "etcd-manager-ca-" + k
-		keys["etcd-peers-ca"] = "etcd-peers-ca-" + k
+
+		// Only nodes running etcd needs the peers CA
```
seh, line 49:
```suggestion
		// Only nodes running etcd need the peers CA
```

nodeup/pkg/model/etcd_manager_tls.go
```diff
@@ -32,16 +32,25 @@ var _ fi.ModelBuilder = &EtcdManagerTLSBuilder{}
...
+		if b.IsMaster {
+			keys["etcd-manager-ca"] = "etcd-manager-ca-" + k
+			keys["etcd-peers-ca"] = "etcd-peers-ca-" + k
+		}
 		// Because API server can only have a single client-cert, we need to share a client CA
```
seh, line 54:
```suggestion
		// Because API server can only have a single client certificate for etcd, we need to share a client CA
```

pkg/nodelabels/builder_test.go
```diff
@@ -63,7 +63,8 @@ func TestBuildNodeLabels(t *testing.T) {
 			expected: map[string]string{
 				RoleLabelMaster16:       "",
 				RoleLabelControlPlane20: "",
-				RoleLabelName15:         RoleMasterLabelValue15,
+				//RoleLabelAPIServer16:    "",
```
seh, line 66: Did you mean to leave this commented?

seh (approved):
I made a few small suggestions and asked a few questions, but this looks like it's going to work, with some room left for improvement in the future.

cmd/kops/rollingupdatecluster.go
```diff
@@ -176,6 +177,11 @@ func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 		Example: rollingupdateExample,
 	}
 
+	var allRoles []string
```
seh, line 180: Consider:
```suggestion
	allRoles := make([]string, 0, len(kopsapi.AllInstanceGroupRoles))
```

cmd/kops/rollingupdatecluster.go
```diff
@@ -188,7 +194,7 @@ func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().DurationVar(&options.PostDrainDelay, "post-drain-delay", options.PostDrainDelay, "Time to wait after draining each node")
 	cmd.Flags().BoolVarP(&options.Interactive, "interactive", "i", options.Interactive, "Prompt to continue after each instance is updated")
 	cmd.Flags().StringSliceVar(&options.InstanceGroups, "instance-group", options.InstanceGroups, "List of instance groups to update (defaults to all if not specified)")
-	cmd.Flags().StringSliceVar(&options.InstanceGroupRoles, "instance-group-roles", options.InstanceGroupRoles, "If specified, only instance groups of the specified role will be updated (e.g. Master,Apiserver,Node,Bastion)")
+	cmd.Flags().StringSliceVar(&options.InstanceGroupRoles, "instance-group-roles", options.InstanceGroupRoles, "If specified, only instance groups of the specified role will be updated ("+strings.Join(allRoles, ",")+")")
```
seh, line 197: Clever!

seh (approved):

pkg/nodelabels/builder_test.go
olemarkus: Yes. The test runs without APIServerNodes feature flag. Will remove the comment when we remove/enable the ff by defult

pkg/nodelabels/builder.go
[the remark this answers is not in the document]
rifelpet: I recall from office hours this week that we were leaning towards using either `apiserver` or `kube-apiserver` for this label.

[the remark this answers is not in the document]
olemarkus: uh, we went a bit back and forth on that. I think `api-server` was preferred as it matched `control-plane`. And specifically we did not want `kube-apiserver`?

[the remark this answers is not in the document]
justinsb: My thought is that it should either be `kube-apiserver` or `apiserver`, because we should match `kube-apiserver`.  But then we also have `kube-controller-manager` so maybe there is no pattern and we should just go with what we think :-)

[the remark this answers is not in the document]
olemarkus: We are behind a feature flag now, so we can choose one and see how people react.

justinsb:
This LGTM - thanks for the tweaks and the feature flag (which lets us tweak some of the last unsettled points, like `apiserver` vs `api-server`!)

rifelpet:
/retest

we should definitely get an e2e job setup to test this

Issue #10719, opened 2021-02-03
"does not contain acceptable node role" error when starting a k8s 1.19 cluster specifying custom iam roles


alanbover:
**1. What `kops` version are you running? The command `kops version`, will display
1.19.0

**2. What Kubernetes version are you running? 
1.19.7

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
kops create -f ${BUILD_PATH}/cluster.yaml --state ${ASSETS_S3_PATH} -v9

**5. What happened after the commands executed?**
The cluster master instances boots correctly. But the nodes are not able to start. Seems that they are failing with the error 
```
Feb 03 15:45:37 ip-10-82-39-61.ec2.internal nodeup[4153]: W0203 15:45:37.535214    4153 executor.go:139] error running task "BootstrapClient/BootstrapClient" (33s remaining to succeed): bootstrap returned status code 403: failed to verify token: arn "arn:aws:sts::<hiddenAWSaccount>:assumed-role/eu-west-1-dev-k8sWorker/i-0a508b0db9c93ce33" does not contain acceptable node role
```

**6. What did you expect to happen?**
The nodes starts correctly, as they do for previous versions of kubernetes (< 1.19)

**7. Please provide your cluster manifest. Execute

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: null
  generation: 1
  name: loko2.kops.mydomain.com
spec:
  additionalSans:
  - api.loko2.us-east-1.admin.mydomain.com
  - api.loko2.us-east-1.mydomain.com
  api:
    loadBalancer:
      class: Classic
      type: Internal
  authorization:
    rbac: {}
  channel: stable
  cloudLabels:
    billed-service: schip-next
    billed-team: cpr
    monitor-with-datadog: enabled
    realm: dev
    support-team: cpr
  cloudProvider: aws
  configBase: s3://schip-kops-dev-us-east-1-<accountId>/loko2.kops.mydomain.com
  containerRuntime: docker
  etcdClusters:
  - cpuRequest: "1"
    etcdMembers:
    - instanceGroup: master-us-east-1a
      name: a
      volumeIops: 100
      volumeSize: 100
      volumeType: io1
    - instanceGroup: master-us-east-1b
      name: b
      volumeIops: 100
      volumeSize: 100
      volumeType: io1
    - instanceGroup: master-us-east-1c
      name: c
[...]

PR #10728, opened 2021-02-04
Fill Role names in kops-controller-config instead of instance profile names when it is specified

h3poteto:
refs: #10719 

The role names are checked in node bootstrap. If profile names are provided, bootstrap will fail. Because profile name and role name do not always mactch in AWS IAM. So I fixed `nodesRoles` in kops-controller-config to store IAM Role names associated with specified IAM Instance Profile.

I think that this should be cherry-picked to 1.19.

### DIFF
--- a/upup/pkg/fi/cloudup/awsup/aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/aws_cloud.go
@@ -1751,3 +1751,18 @@ func (c *awsCloudImplementation) AccountInfo() (string, string, error) {
 	}
 	return arn.AccountID, arn.Partition, nil
 }
+
+// GetRolesInInstanceProfile return role names which are associated with the instance profile specified by profileName.
+func GetRolesInInstanceProfile(c AWSCloud, profileName string) ([]string, error) {
+	output, err := c.IAM().GetInstanceProfile(&iam.GetInstanceProfileInput{
+		InstanceProfileName: aws.String(profileName),
+	})
+	if err != nil {
+		return nil, err
+	}
+	var roleNames []string
+	for _, role := range output.InstanceProfile.Roles {
+		roleNames = append(roleNames, *role.RoleName)
+	}
+	return roleNames, nil
+}
--- a/upup/pkg/fi/cloudup/template_functions.go
+++ b/upup/pkg/fi/cloudup/template_functions.go
@@ -59,6 +59,8 @@ import (
 // TemplateFunctions provides a collection of methods used throughout the templates
 type TemplateFunctions struct {
 	model.KopsModelContext
+
+	cloud fi.Cloud
 }
 
 // AddTo defines the available functions we can use in our YAML models.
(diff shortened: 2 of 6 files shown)


rifelpet:
/cc @johngmyers if you have time to review this.

I'm debating the pros and cons of having kops-controller fetch the list of roles for each instance profile at startup time vs having the kops CLI get that list and populate the configmap with the list of roles.

I'll admit I don't fully understand how having multiple roles associated with an instance profile and having that instance profile attached to an instance affects credentials provided via instance metadata, specifically when nodeup creates an sts.GetCallerIdentity request. How is the role chosen between all of the roles associated with the instance's instance profile?

h3poteto:
> I'm debating the pros and cons of having kops-controller fetch the list of roles for each instance profile at startup time vs having the kops CLI get that list and populate the configmap with the list of roles.

First I considered it, that is the kops CLI gets IAM Roles and fill it in kops-controller-config. But it is difficult because kops-controller-config is generated in template functions: https://github.com/kubernetes/kops/blob/0271f0e8cb8115a064ee8dc5645dbb950da91507/upup/pkg/fi/cloudup/template_functions.go#L454
So it is difficult to call AWS API in this function.

> How is the role chosen between all of the roles associated with the instance's instance profile?

OK, I'll check it.

johngmyers:
Per AWS documentation, an instance profile can contain at most one IAM role. (I admit to being baffled as to why instance profiles exist, as opposed to assigning roles directly.)

The role in an instance profile can change, but I don't think kops needs to support noticing this change without an apply_cluster. I would prefer early binding of the permitted role, during apply_cluster.

I think we need to add a field of type `fi.Cloud` to `TemplateFunctions`. Then it would be possible to call the AWS IAM API from within `KopsControllerConfig`.

h3poteto:
Indeed, if `TemplateFunctions` has `fi.Cloud`, it is possible.

Which do you think is better, to get IAM Roles in kops-controller or to get IAM Roles in template functions and write it in configmap?

h3poteto:
> Per AWS documentation, an instance profile can contain at most one IAM role.

Thank you.
I also tried, but can't assign multiple role to a instance profile.

rifelpet:
My mistake, I was remembering incorrectly and was thrown off by the [Roles field here](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#InstanceProfile) being a list. I agree in preferring the lookup happen in TemplateFunctions and providing the role list in the config file. We will need to provide a fi.Cloud to template functions. TemplateFunctions is setup in [apply_cluster.go](https://github.com/kubernetes/kops/blob/ae3579076d60d162687d58b2adc2f1d7ab4dc02d/upup/pkg/fi/cloudup/apply_cluster.go#L493) which already has a fi.Cloud so it should be straight forward to pass in.

h3poteto:
I see. I will rewrite this changes.

h3poteto:
@rifelpet I rewrote this pull request, please review this again.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -445,9 +447,26 @@ func (tf *TemplateFunctions) KopsControllerConfig() (string, error) {
...
-						return "", fmt.Errorf("getting role for ig %s: %v", ig.Name, err)
+						return "", fmt.Errorf("getting profile for ig %s: %v", ig.Name, err)
+					}
+					// The IAM Instance Profile has not been created at this point if it is not specified.
+					// Because the IAM Instance Profile and the IAM Role are created in IAMModelBuilder tasks.
+					// Therefore, the IAM Role associated with IAM Instance Profile is acquired only when it is not specified.
+					if ig.Spec.IAM != nil && ig.Spec.IAM.Profile != nil {
+						fmt.Printf("[h3poteto debug1] %+v\n", *profile.Name)
+						fmt.Printf("[h3poteto debug2] %+v\n", tf.cloud)
+						roles, err := tf.cloud.GetRolesInInstanceProfile(*profile.Name)
```
rifelpet, line 458: Rather than adding `GetRolesInInstanceProfile` to the cloud interface and all of the cloud provider implementations, it's probably easier to add `GetRolesInInstanceProfile` to the AWSCloud interface and use type assertions on `tf.cloud` to get an AWSCloud, like this:

https://github.com/kubernetes/kops/blob/8c5dc5f0c96132ae14a8fd128ec4aa8b0386927f/pkg/resources/aws/vpc.go#L32

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -445,9 +447,26 @@ func (tf *TemplateFunctions) KopsControllerConfig() (string, error) {
...
+						if err != nil {
+							return "", fmt.Errorf("getting role from profile %s: %v", *profile.Name, err)
+						}
+						fmt.Printf("[h3poteto debug4] %+v\n", roles)
```
rifelpet, line 462:
```suggestion
```

upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -1751,3 +1751,23 @@ func (c *awsCloudImplementation) AccountInfo() (string, string, error) {
...
 	return arn.AccountID, arn.Partition, nil
 }
+
+// GetRolesInInstanceProfile return role names which are associated with the instance profile specified by profileName.
+func (c *awsCloudImplementation) GetRolesInInstanceProfile(profileName string) ([]string, error) {
+	return getRolesInInstanceProfile(c, profileName)
+}
+
+func getRolesInInstanceProfile(c AWSCloud, profileName string) ([]string, error) {
+	fmt.Printf("[h3poteto debug3] called")
```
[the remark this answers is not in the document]
h3poteto: Oh, sorry...

upup/pkg/fi/cloudup/template_functions.go
h3poteto: OK, I see.

h3poteto: I fixed it.

rifelpet:
This looks good, thanks!

/lgtm
/approve

PR #10732, opened 2021-02-04
add user agent to openstack api requests

zetaab:


olemarkus:
/lgtm

This can be quite useful

zetaab:
@olemarkus can you check again, I just modified the names little bit. The problem that we have currently is that some kops component is calling openstack apis quite heavily. We have currently 30 kops clusters in openstack and those makes 1.2M requests in 24 hours against api. The problem is not csi-cinder or openstack cloud controller manager those do have own user agents.

PR #10741, opened 2021-02-05
Allow to control which subnets and IPs get used for the API loadbalancer

codablock:
This PR implements the new field `subnets` for the api loadbalancer in the cluster spec:
```
spec:
  api:
    loadBalancer:
      type: Internal
      subnets:
        - name: subnet-a
          privateIPv4Address: 172.16.1.10
```

It allows to control which subnets are used by the api load balancer. This is useful when VPCs and subnets are managed externally and a more granular separation between subnets is desired, e.g. when you have subnets that offer VPN access to on-prem networks, subnets which offer public IPs and subnets which are completely private.

In such case, the subnets must be carefully chosen to not expose the API in the wrong place.

Review can be done on a per-commit basis as I've tried to separate changes.


codablock:
FYI, I've tested this on the 1.19 branch and would be really happy if this could be backported/cherry-picked into 1.19 :)

docs/cluster_spec.md
```diff
@@ -101,6 +101,45 @@ spec:
...
+    loadBalancer:
+      type: Public
+      subnets:
+        - name: subnet-a
+        - name: subnet-b
+        - name: subnet-c
+````
+
+It is only allowed to add more subnets and forbidden to remove existing ones. This is due to limitations on AWS
+ELBs and NLBs.
```
timothyclarke, line 125: Is there the possibility to remove existing ones if you accept that the consequence is destroying the loadbalancer and creating a new one? eg `--force-recreate-loadbalancer`
I'd consider losing access to the loadbalancer acceptible during a maintenance window.

codablock: I'd assume that this would be possible. But this is not a new problem and thus should be part of another PR. Currently, when you change your set of shared subnets, you'd run into the same issue already as the automatically chosen subnets would differ then.

timothyclarke: My question was primarily around the validation side of things. If I wanted to change those values and accepted that I'd have to (manually) remove the loadbalancer,  would your changes to kops reject my edits?

codablock: It would accept the edits and only fail when running the update. The failure is a little bit unfortunate as it keeps retrying but then of course never succeeds.

rifelpet: A workaround for the above behavior is to manually delete the load balancer. A `kops update cluster` should then report that it will (re)create the load balancer which should succeed with `--yes`. You'll run into certificate issues with gossip clusters though since the load balancer's DNS name is used in various certs.

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -770,11 +804,18 @@ func (e *NetworkLoadBalancer) TerraformLink(params ...string) *terraform.Literal
...
 type cloudformationNetworkLoadBalancer struct {
-	Name    string                    `json:"Name"`
-	Scheme  string                    `json:"Scheme"`
-	Subnets []*cloudformation.Literal `json:"Subnets"`
-	Type    string                    `json:"Type"`
-	Tags    []cloudformationTag       `json:"Tags"`
+	Name           string                         `json:"Name"`
+	Scheme         string                         `json:"Scheme"`
+	Subnets        []*cloudformation.Literal      `json:"Subnets"`
+	SubnetMappings []*cloudformationSubnetMapping `json:"SubnetMappings"`
```
rifelpet, line 810:
```suggestion
	Subnets        []*cloudformation.Literal      `json:"Subnets,omitempty"`
	SubnetMappings []*cloudformationSubnetMapping `json:"SubnetMappings,omitempty"`
```

docs/cluster_spec.md
rifelpet: Ah I see you mentioned this a few lines below. I believe you'll also need to `kops rolling-update cluster --instance-group-roles=Master --cloudonly` after `kops update cluster --yes` recreates the NLB in order to pick up new certificates.

codablock: @rifelpet I did not observe the need for rolling-update while doing my testing, which I did on a semi-live (not prod, but a lot of stuff deployed) cluster.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -63,9 +63,39 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+				if subnet.Name == clusterSubnet.Name {
+					found = true
+					elbSubnet := b.LinkToSubnet(&clusterSubnet)
+					elbSubnets = append(elbSubnets, elbSubnet)
+					nlbSubnetMappings = append(nlbSubnetMappings, &awstasks.SubnetMapping{
+						Subnet: elbSubnet,
+					})
+					if subnet.PrivateIPv4Address != nil {
+						if subnet.PrivateIPv4Address == nil {
+							return fmt.Errorf("privateIPv4Address can't be nil")
```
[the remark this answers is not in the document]
codablock: True. I assume I meant to check for empty strings here...fixing it.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -63,9 +63,39 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+						}
+						if lbSpec.Class != kops.LoadBalancerClassNetwork || lbSpec.Type != kops.LoadBalancerTypeInternal {
+							return fmt.Errorf("privateIPv4Address only allowed for internal NLBs")
+						}
```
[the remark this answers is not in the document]
codablock: Moved it and also added more validation (e.g. subnet cidr matching)

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -63,9 +63,39 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+
+						nlbSubnetMappings[len(nlbSubnetMappings)-1].PrivateIPv4Address = subnet.PrivateIPv4Address
```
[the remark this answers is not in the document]
codablock: Agree and fixed.

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -354,7 +352,16 @@ func (e *NetworkLoadBalancer) Find(c *fi.Context) (*NetworkLoadBalancer, error)
 	}
 
 	for _, az := range lb.AvailabilityZones {
-		actual.Subnets = append(actual.Subnets, &Subnet{ID: az.SubnetId})
+		sm := &SubnetMapping{
+			Subnet: &Subnet{ID: az.SubnetId},
+		}
+		for _, a := range az.LoadBalancerAddresses {
```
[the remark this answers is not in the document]
codablock: My understanding is that this should never be the case, as the NLB only allows to assign one PrivateIPv4Address. But I can add a check to ensure that we never run into that case and abort just to make sure.

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -354,7 +352,16 @@ func (e *NetworkLoadBalancer) Find(c *fi.Context) (*NetworkLoadBalancer, error)
...
+			if a.PrivateIPv4Address != nil {
+				sm.PrivateIPv4Address = a.PrivateIPv4Address
```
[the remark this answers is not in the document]
codablock: In case someone does not specify `privateIPv4Address` when initially deploying the NLB, AWS will enable automatic assignment of IPs from the free IP pool of the subnet. These automatically assigned IPs do not appear in `LoadBalancerAddresses` (you can only find them when looking at the network interfaces of the NLB). Also, changing the IP assignment in any way is not possible afterwards.

So I assume the current behaviour is the correct one. The only thing an admin can modify after initial creation of the NLB is assigning new subnets to previously unassigned AZs (each AZ can have only one subnet). In that case, kops would detect that there is a new SubnetMapping for the NLB and then try to remove that one, which will of course fail. We can't really avoid this situation, because otherwise a removal of a subnet in the spec would be ignored as well as kops can't know the difference.

codablock:
@rifelpet Thanks for your review. I've gone through it and fixed things where necessary. You can review the changes per-commit.

docs/cluster_spec.md
rifelpet: It may only be gossip clusters that need the rolling update (cluster names ending in `.k8s.local`, was yours gossip or using route53 ? Now that I think about it more, I think the nodes themselves might need to be rolled too to pick up the new load balancer dns name which is pretty invasive. We could say "It is not recommended to make this change on gossip clusters due to the requirement of performing a --cloudonly rolling update on the entire cluster which results in temporary downtime of all workloads. It is recommended to create a new cluster instead."

pkg/apis/kops/validation/aws.go
```diff
@@ -196,3 +199,49 @@ func awsValidateSSLPolicy(fieldPath *field.Path, spec *kops.LoadBalancerAccessSp
...
+
+func awsValidateLoadBalancerSubnets(fieldPath *field.Path, spec kops.ClusterSpec) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	lbSpec := spec.API.LoadBalancer
+
+	for i, subnet := range lbSpec.Subnets {
+		var clusterSubnet *kops.ClusterSubnetSpec
+		if subnet.Name == "" {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Index(i).Child("name"), subnet, "subnet name can't be empty"))
```
rifelpet, line 211:
```suggestion
			allErrs = append(allErrs, field.Required(fieldPath.Index(i).Child("name"), subnet, "subnet name can't be empty"))
```

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#Required

pkg/apis/kops/validation/aws.go
```diff
@@ -196,3 +199,49 @@ func awsValidateSSLPolicy(fieldPath *field.Path, spec *kops.LoadBalancerAccessSp
...
+		} else {
+			for _, cs := range spec.Subnets {
+				if subnet.Name == cs.Name {
+					clusterSubnet = &cs
+					break
+				}
+			}
+			if clusterSubnet == nil {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Index(i).Child("name"), subnet, fmt.Sprintf("subnet %q not found in cluster subnets", subnet.Name)))
```
rifelpet, line 220:
```suggestion
				allErrs = append(allErrs, field.NotFound(fieldPath.Index(i).Child("name"), subnet, fmt.Sprintf("subnet %q not found in cluster subnets", subnet.Name)))
```

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#NotFound

pkg/apis/kops/validation/aws.go
```diff
@@ -196,3 +199,49 @@ func awsValidateSSLPolicy(fieldPath *field.Path, spec *kops.LoadBalancerAccessSp
...
+			}
+		}
+
+		if subnet.PrivateIPv4Address != nil {
+			if *subnet.PrivateIPv4Address == "" {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Index(i).Child("privateIPv4Address"), subnet, "privateIPv4Address can't be empty"))
```
rifelpet, line 226:
```suggestion
				allErrs = append(allErrs, field.Required(fieldPath.Index(i).Child("privateIPv4Address"), subnet, "privateIPv4Address can't be empty"))
```

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#Required

pkg/apis/kops/validation/aws.go
```diff
@@ -196,3 +199,49 @@ func awsValidateSSLPolicy(fieldPath *field.Path, spec *kops.LoadBalancerAccessSp
...
+				_, ipNet, err := net.ParseCIDR(clusterSubnet.CIDR)
+				if err == nil { // we assume that the cidr is actually valid
+					if !ipNet.Contains(ip) {
+						allErrs = append(allErrs, field.Invalid(fieldPath.Index(i).Child("privateIPv4Address"), subnet, "privateIPv4Address is not part of the subnet CIDR"))
+					}
+				}
+
+			}
+			if lbSpec.Class != kops.LoadBalancerClassNetwork || lbSpec.Type != kops.LoadBalancerTypeInternal {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Index(i).Child("privateIPv4Address"), subnet, "privateIPv4Address only allowed for internal NLBs"))
```
rifelpet, line 241:
```suggestion
				allErrs = append(allErrs, field.Forbidden(fieldPath.Index(i).Child("privateIPv4Address"), subnet, "privateIPv4Address only allowed for internal NLBs"))
```

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#Forbidden

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -680,15 +706,22 @@ func (_ *NetworkLoadBalancer) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Ne
...
 type terraformNetworkLoadBalancer struct {
-	Name                   string               `json:"name" cty:"name"`
-	Internal               bool                 `json:"internal" cty:"internal"`
-	Type                   string               `json:"load_balancer_type" cty:"load_balancer_type"`
-	Subnets                []*terraform.Literal `json:"subnets" cty:"subnets"`
-	CrossZoneLoadBalancing bool                 `json:"enable_cross_zone_load_balancing" cty:"enable_cross_zone_load_balancing"`
+	Name                   string                                      `json:"name" cty:"name"`
+	Internal               bool                                        `json:"internal" cty:"internal"`
+	Type                   string                                      `json:"load_balancer_type" cty:"load_balancer_type"`
+	Subnets                []*terraform.Literal                        `json:"subnets,omitempty" cty:"subnets"`
```
rifelpet, line 712: Ah I just realized we're removing the Subnets field from the `NetworkLoadBalancer` struct above. In that case we can just remove `Subnets` from here and Cloudformation, and remove `omitempty` from the SubnetMappings json field tags since we should always be setting it.

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -354,7 +352,18 @@ func (e *NetworkLoadBalancer) Find(c *fi.Context) (*NetworkLoadBalancer, error)
...
 
 	for _, az := range lb.AvailabilityZones {
-		actual.Subnets = append(actual.Subnets, &Subnet{ID: az.SubnetId})
+		sm := &SubnetMapping{
+			Subnet: &Subnet{ID: az.SubnetId},
+		}
+		for _, a := range az.LoadBalancerAddresses {
+			if a.PrivateIPv4Address != nil {
+				if sm.PrivateIPv4Address != nil {
+					return nil, fmt.Errorf("NLB has more then one PrivateIPv4Address, which is unexpected")
```
rifelpet, line 361:
```suggestion
					return nil, fmt.Errorf("NLB has more then one PrivateIPv4Address, which is unexpected. This is a bug in kOps, please open a GitHub issue.")
```

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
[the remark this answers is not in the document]
rifelpet: ah I didn't realize the automatically assigned IPs aren't in this field. In that case I agree the current behavior should work.

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
codablock: Ah true. I've removed the Subnets field now.

codablock:
@rifelpet I pushed a few more commits to handle your latest comments.

rifelpet:
this looks good to me, thanks for implementing this! I'll have someone else take a pass at it too

/ok-to-test

codablock:
@rifelpet Can you help me find the logs of the test failures that appeared now?

rifelpet:
Yea you can click the Details button in the list of checks at the bottom of the PR. here are the two failures:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/10741/pull-kops-verify-staticcheck/1359527983353696256
looks like a pretty trivial fix in the validation test file.


https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/10741/pull-kops-verify-cloudformation/1359527985023029248

is a bit strange, i think the line numbers are off but its essentially complaining that the allocation ID and private IP v4 address fields in the cloudformation JSON file are null. We should probably not include them at all (`omitempty` field tag) if that is the case.

codablock:
@rifelpet Thanks, that helped me :) Should be fixed now.

rifelpet:
/lgtm

I'll give others a chance for input otherwise we can get this merged soon. I'll bring it up at kOps office hours on friday too. Excited to get this in :)

codablock:
@rifelpet What do you think about the chances of this being cherry-picked into 1.19?

rifelpet:
1.19.X releases will contain bug fixes only. Since this is a new feature it likely wont get backported to 1.19 though we are aiming for an accelerated 1.20.0 release so the wait shouldn't be too long.

codablock:
ok, having this in 1.20.0 would already be very cool.

codablock:
Just seen that this got merged, thanks :) Any steps required from my side to get this cherry-picked into 1.20?

rifelpet:
@codablock I took care of it in https://github.com/kubernetes/kops/pull/10829

Issue #10718, opened 2021-02-03
[BUG] KOPS 1.19 Seg Fault when removing an in-use instance type from a mixed instance group


timothyclarke:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
`Version 1.19.0 (git-04d36d7d92c72601efd918877fc180c846129ffb)`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:23:52Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:15:20Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
`kops edit ig <instance group name>`, `kops update cluster`, `kops update cluster --yes`

1. Create a mixed instance group, apply and make sure there is atleast 1 instance in the group active
2. Edit the instance group and remove an instance type that is active
3. Attempt to update the cluster

An example would be
```
apiVersion: kops.k8s.io/v1alpha2
kind: InstanceGroup
metadata:
  creationTimestamp: "2021-02-03T09:39:25Z"
  generation: 3
  labels:
    kops.k8s.io/cluster: dev-euw2.example.com
  name: sites
spec:
  cloudLabels:
    BrandName: sites
    k8s.io/cluster-autoscaler/enabled: '""'
  image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20210119.1
  machineType: t3a.2xlarge
  maxSize: 11
  minSize: 1
  mixedInstancesPolicy:
    instances:
    - t3a.2xlarge
    - m5a.xlarge
    - m5a.2xlarge
    - m5a.4xlarge
    - m4.2xlarge
    - m5.2xlarge
    - m5.4xlarge
    - r5a.xlarge
[...]

PR #10742, opened 2021-02-05
Use expected LaunchTemplateId in updating ASG when MixedInstancePolicy is changed

h3poteto:
refs: #10718 

When user changes only mixedInstancePolicy, LaunchTemplateId is not changed. 
But current logic expect changes to LaunchTemplateId; therefore nil pointer dereference has occur.

In this case, it is not necessary to change LaunchTemplateId in AutoScalingGroup, so I fixed it.

---
I don't know the reason for this logic. If anyone knows why, please tell me.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
+++ b/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
@@ -454,7 +454,7 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 
 		if changes.LaunchTemplate != nil {
 			spec := &autoscaling.LaunchTemplateSpecification{
-				LaunchTemplateId: changes.LaunchTemplate.ID,
+				LaunchTemplateId: e.LaunchTemplate.ID,
 				Version:          aws.String("$Latest"),
 			}
 			if e.UseMixedInstancesPolicy() {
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -489,7 +489,7 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 			if setup(request).LaunchTemplate == nil {
 				setup(request).LaunchTemplate = &autoscaling.LaunchTemplate{
 					LaunchTemplateSpecification: &autoscaling.LaunchTemplateSpecification{
-						LaunchTemplateId: changes.LaunchTemplate.ID,
+						LaunchTemplateId: e.LaunchTemplate.ID,
```
hakman, line 492: > I don't know the reason for this logic. If anyone knows why, please tell me.

My mistake. Thanks for fixing.
Could you do same on line 457?

h3poteto: In line 457 `changes.LaunchTemplate` is not nil because it is checked on line 455. So I don't think it needs to be changed.

hakman: Just for consistency. I agree it won't improve anything.

h3poteto: OK, I see. I will add it.

h3poteto: I added: https://github.com/kubernetes/kops/pull/10742/commits/f22bafb43fb26388808c183fe6d062ba7510f4ca

hakman: 👍 
Can you also squash the 2 commits?

h3poteto: I squashed and force pushed.

hakman:
Thanks @h3poteto!
/lgtm
/approve

PR #10744, opened 2021-02-05
add azure support for internal loadbalancer to k8s api

collin-woodruff-t1cg:
Creates Basic type Loadbalancer that can be used to connect to the k8s api. 

Notes:
- Kops will try to delete the loadbalancer first and receive an error but will successfully get deleted on the retry since resources are no longer using it.
- After creating will need to use --insecure-skip-tls-verify to successfully connect to the api.

Any guidance to fix these issues would be appreciated. I'd be happy to add the fixes to my pr.


upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -136,39 +139,71 @@ func (c *azureCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([
...
-	// Get masters scale set network interfaces and append to api ingress status
-	nis, err := c.NetworkInterface().ListScaleSetsNetworkInterfaces(context.TODO(), rg, vmssName)
-	if err != nil {
-		return nil, fmt.Errorf("error getting Master Scale Set Network Interfaces for API Ingress Status: %s", err)
-	}
-	for _, ni := range nis {
-		if !*ni.Primary {
-			continue
+		for _, lb := range lbs {
+			if lb.Tags[TagClusterName] != nil {
```
kenji-cloudnatix, line 151: How about the following to reduce the number of nests?

```go
val := b.Tags[TagClusterName]
if val == nil ||  *val != cluster.Name
  continue
}
```

upup/pkg/fi/cloudup/azuretasks/loadbalancer.go
```diff
@@ -0,0 +1,233 @@
...
+				{
+					Name: to.StringPtr("LoadBalancerBackEnd"),
+				},
+			},
+			Probes: &[]network.Probe{
+				{
+					Name: to.StringPtr("Health-TCP-80"),
+					ProbePropertiesFormat: &network.ProbePropertiesFormat{
+						Protocol:          network.ProbeProtocolTCP,
+						Port:              to.Int32Ptr(80),
```
kenji-cloudnatix, line 167: Question: Do we need to have a probe for port 80? Or just 443 will be sufficient?

pkg/model/azuremodel/api_loadbalancer.go
```diff
@@ -38,5 +45,122 @@ func (b *APILoadBalancerModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+	return a[i].subnet.Name < a[j].subnet.Name
+}
+
+// subnetForLoadBalancer returns the subnet the loadbalancer will use.
+func (c *AzureModelContext) subnetForLoadBalancer() (*kops.ClusterSubnetSpec, error) {
+	var subnets []*kops.ClusterSubnetSpec
+	for i := range c.Cluster.Spec.Subnets {
+		subnet := &c.Cluster.Spec.Subnets[i]
+
+		if c.Cluster.Spec.API.LoadBalancer.Type == kops.LoadBalancerTypeInternal {
```
kenji-cloudnatix, line 122: Nitpick: we can combine these two `if`s into a single one with `&&`?

pkg/model/azuremodel/api_loadbalancer.go
```diff
@@ -38,5 +45,122 @@ func (b *APILoadBalancerModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+	}
+
+	if len(subnets) == 0 {
+		return nil, nil
+	}
+	if len(subnets) == 1 {
+		return subnets[0], nil
+	}
+
+	migSubnets := sets.NewString()
```
kenji-cloudnatix, line 138: Could you please add some comment and explain the logic? Couldn't follow..

pkg/model/azuremodel/api_loadbalancer.go
```diff
@@ -38,5 +45,122 @@ func (b *APILoadBalancerModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+	case kops.LoadBalancerTypePublic:
+		// TODO: Implement creating public ip and attach to public loadbalancer
+		return fmt.Errorf("only internal loadbalancer for API server is implemented in Azure")
+	default:
+		return fmt.Errorf("unhandled LoadBalancer type %q", lbSpec.Type)
+	}
+
+	// Create LoadBalancer for API ELB
+	var lb *azuretasks.LoadBalancer
+	{
```
kenji-cloudnatix, line 66: nitpick: Can we remove `{` and just have 

```go
lb :=  &azuretasks.LoadBalancer{ ...
}
```
?

kenji-cloudnatix (commented):
Thanks for making the change, @collin-woodruff-t1cg ! I'm not very familiar with the LB logic in kops. Just made several superficial comments..

pkg/model/azuremodel/api_loadbalancer.go
collin-woodruff-t1cg: i got this function from the aws load balancer file. but it basically searches through all the subnets for the master instance group subnets and if found, scored with a value of 1. then at the end it sorts the subnets from greatest to least and the first element in the array is returned. this way it should be in a master subnet. 

i'm fine with if you'd rather just have the first master subnet found be returned and the internal loadbalancer be assigned to that subnet. if you think thats a better way to do it?

kenji-cloudnatix: Ah, I see. Thanks for the explanation! I'll defer to other reviewers who know the code better.

pkg/model/azuremodel/api_loadbalancer.go
```diff
@@ -17,9 +17,16 @@ limitations under the License.
 package azuremodel
 
 import (
-	"errors"
+	"fmt"
+	"sort"
 
+	"github.com/Azure/go-autorest/autorest/to"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog"
```
kenji-cloudnatix, line 25: Ah, we need to change this to v2.. (`"k8s.io/klog/v2"`)..

kenji-cloudnatix (approved):
Looks good to me!

pkg/model/azuremodel/api_loadbalancer.go
collin-woodruff-t1cg: just updated that in that file and the loadbalancer.go file in azuretasks. let me know if you see any that i might have missed.

pkg/model/azuremodel/api_loadbalancer.go
```diff
@@ -38,5 +45,120 @@ func (b *APILoadBalancerModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+	switch lbSpec.Type {
+	case kops.LoadBalancerTypeInternal:
+		lb.RequirePublicIP = to.BoolPtr(false)
+		subnet, err := b.subnetForLoadBalancer()
+		if err != nil {
+			return err
+		}
+		lb.Subnet = b.LinkToAzureSubnet(subnet)
+	case kops.LoadBalancerTypePublic:
+		lb.RequirePublicIP = to.BoolPtr(true)
```
rifelpet, line 81: Above we return an error that public isnt supported, but here we're setting values on the LB task as if it is supported. If it is supported we should drop the error message above and if it isn't supported then we dont need this case statement. I see some tests later in this PR that have `RequirePublicIP` set to true too.

pkg/model/azuremodel/api_loadbalancer.go
rifelpet: I think the reason this complexity exists in AWS is because kops needs to choose a subnet per availability zone to use in its load balancer.

Do Azure load balancers support multiple subnets? if not then we could reduce this down to just returning the first master subnet as you mentioned.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
```diff
@@ -173,14 +177,15 @@ func (s *VMScaleSet) Find(c *fi.Context) (*VMScaleSet, error) {
...
-		PrincipalID:        found.Identity.PrincipalID,
+		RequirePublicIP:     to.BoolPtr(ipConfig.PublicIPAddressConfiguration != nil),
+		RequireLoadBalancer: to.BoolPtr(ipConfig.LoadBalancerBackendAddressPools != nil),
+		SKUName:             found.Sku.Name,
+		Capacity:            found.Sku.Capacity,
+		ComputerNamePrefix:  osProfile.ComputerNamePrefix,
+		AdminUser:           osProfile.AdminUsername,
+		SSHPublicKey:        sshKeys[0].KeyData,
+		Tags:                found.Tags,
+		PrincipalID:         found.Identity.PrincipalID,
```
rifelpet, line 188: We will need to set the LoadBalancerName field in this Find function or else `kops update cluster` dry runs will always report trying to add a load balancer to the VMSS. Can we determine the load balancer name based on the `LoadBalancerBackendAddressPools` ?

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -136,39 +139,70 @@ func (c *azureCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([
...
-	}
-	for _, ni := range nis {
-		if !*ni.Primary {
-			continue
+		for _, lb := range lbs {
+			val := lb.Tags[TagClusterName]
+			if val == nil || *val != cluster.Name {
+				continue
+			}
+			for _, i := range *lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations {
```
rifelpet, line 155: Do we need to do any nil checks on this nested reference? for example could `LoadBalancerPropertiesFormat` be nil?

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -136,39 +139,70 @@ func (c *azureCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([
...
+				switch lbSpec.Type {
+				case kops.LoadBalancerTypeInternal:
+					ingresses = append(ingresses, kops.ApiIngressStatus{
+						IP: *i.FrontendIPConfigurationPropertiesFormat.PrivateIPAddress,
```
rifelpet, line 159: same as above, could `FrontendIPConfigurationPropertiesFormat` be nil?

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -136,39 +139,70 @@ func (c *azureCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([
...
+					})
+				case kops.LoadBalancerTypePublic:
+					ingresses = append(ingresses, kops.ApiIngressStatus{
+						IP: *i.FrontendIPConfigurationPropertiesFormat.PublicIPAddress.PublicIPAddressPropertiesFormat.IPAddress,
```
rifelpet, line 163: could `FrontendIPConfigurationPropertiesFormat` or `PublicIPAddress` or `PublicIPAddressPropertiesFormat` be nil?

upup/pkg/fi/cloudup/azuretasks/loadbalancer.go
```diff
@@ -0,0 +1,204 @@
...
+
+// LoadBalancer is an Azure Cloud LoadBalancer
+type LoadBalancer struct {
+	Name          *string
+	Lifecycle     *fi.Lifecycle
+	ResourceGroup *ResourceGroup
+	Subnet        *Subnet
+
+	// RequirePublicIP is set to true when the Loadbalancer requires a public IP.
+	RequirePublicIP *bool
```
rifelpet, line 40: nit: `Require` is redundant in this field name. Skimming the azure docs, perhaps `External *bool` would be consistent with terminology?

upup/pkg/fi/cloudup/azuretasks/loadbalancer.go
```diff
@@ -0,0 +1,204 @@
...
+	// PublicIPName is the name of the ip created to be attached to the loadbalancer
+	PublicIPName *string
```
rifelpet, line 42: likely related to my first comment about whether we support `LoadBalancerTypePublic` but I don't see this field set by azuremodel nor do I see it being set by Find below. Both will need to set the field.

pkg/model/azuremodel/api_loadbalancer.go
collin-woodruff-t1cg: Public isn't supported because we would need to create a public ip address first and then attach it to the load balancer using its id. I can remove all the logic for creating a public load balancer if you think that would be better? i just thought it couldn't hurt having the logic there but stopping it from executing at the beginning.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
collin-woodruff-t1cg: I don't believe azure supports multiple subnets yet so returning the first master subnet should work. 

I don't think we'll be able to get the name of the load balancer from the backend address pools. But since the name is just `api-<clustername>` i could write a function in azure_cloud.go file. Would you suggest writing a get tags function and then getting the cluster name that way or a cluster name function similar to the subscription id function?

upup/pkg/fi/cloudup/azure/azure_cloud.go
collin-woodruff-t1cg: I'll add checks its always possible.

upup/pkg/fi/cloudup/azure/azure_cloud.go
collin-woodruff-t1cg: I'll add checks its always possible.

upup/pkg/fi/cloudup/azure/azure_cloud.go
collin-woodruff-t1cg: I'll add checks its always possible.

upup/pkg/fi/cloudup/azuretasks/loadbalancer.go
collin-woodruff-t1cg: I'll change it to External. I agree with you that is more fitting.

upup/pkg/fi/cloudup/azuretasks/loadbalancer.go
collin-woodruff-t1cg: I can remove this with the rest of the public ip logic. If we keep the logic we'll run into the same issue we have with LoadBalancerName in the VMSS so this may just be better to remove all together.

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -136,39 +139,82 @@ func (c *azureCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([
...
-			continue
+		for _, lb := range lbs {
+			val := lb.Tags[TagClusterName]
+			if val == nil || *val != cluster.Name {
+				continue
+			}
+			for _, i := range *lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations {
+				if lb.LoadBalancerPropertiesFormat == nil || lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations == nil {
+					continue
+				}
```
rifelpet, line 158: We should ensure `LoadBalancerPropertiesFormat` is not nil before the range loop. `FrontendIPConfigurations` is okay to be nil, the range loop will treat it as an empty collection.
 
```suggestion
			if lb.LoadBalancerPropertiesFormat == nil {
				continue
			}
			for _, i := range lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations {
```

upup/pkg/fi/cloudup/azure/azure_cloud.go
```diff
@@ -136,39 +139,82 @@ func (c *azureCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([
...
+				switch lbSpec.Type {
+				case kops.LoadBalancerTypeInternal:
+					if i.FrontendIPConfigurationPropertiesFormat == nil || i.FrontendIPConfigurationPropertiesFormat.PrivateIPAddress == nil {
```
rifelpet, line 161: nit: `FrontendIPConfigurationPropertiesFormat == nil` is being checked in both `case` blocks so we could probably move it to its own check above the switch statement (where the `if` statement used to be in my above suggestion)

pkg/model/azuremodel/api_loadbalancer.go
rifelpet: If you think we'll be able to add public support soon then I'm fine with keeping the underlying logic behind it in this PR rather than reverting and re-adding later. It sounds like we just need to create a "public ip address" azuretask and connect that to the load balancer task?

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
rifelpet: We need a way to determine the LoadBalancer's Name given the VMSS response object. I see below in `RenderAzure()` that the `ID` field in `LoadBalancerBackendAddressPools` resembles a URL path that includes the load balancer name. Does the VMSS object here contain that same path? If so we could parse the load balancer name out of the path.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
```diff
@@ -73,6 +73,10 @@ type VMScaleSet struct {
 	StorageProfile *VMScaleSetStorageProfile
 	// RequirePublicIP is set to true when VMs require public IPs.
 	RequirePublicIP *bool
+	// RequireLoadBalancer is set to true when VMs are using loadbalancer.
+	RequireLoadBalancer *bool
+	// LoadBalancer is the Load Balancer object the VMs will use.
+	LoadBalancerName *string
```
rifelpet, line 79: Because this name is a reference to another task that kops manages (an instance of the load balancer task) we should make this field's type the task itself, that way kops can build an accurate dependency tree when applying these changes rather than trying to add a LB to a VMSS before the LB exists.

This variable only needs the LB's `Name` field set rather than all of the properties set in `pkg/model/azuremodel/api_loadbalancer.go` Kops will use the Name field to compare against the list of tasks it knows about and will populate this variable's remaining fields automatically. Hopefully this makes sense :) 

```suggestion
	LoadBalancer *LoadBalancer
```



This means the Find function below needs to set this value as:

```go
LoadBalancer: &LoadBalancer {
  Name: "foo",
}
```

and RenderAzure() will reference `LoadBalancer.Name`

Lastly, `pkg/model/azuremodel/vmscaleset.go` will need to set the VMSS task as follows:

```go
# if IG role == master ...
t.LoadBalancer = &azuretasks.LoadBalancer {
    Name: b.NameForLoadBalancer(),
}
```

collin-woodruff-t1cg: made the changes.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
```diff
@@ -73,6 +73,10 @@ type VMScaleSet struct {
 	StorageProfile *VMScaleSetStorageProfile
 	// RequirePublicIP is set to true when VMs require public IPs.
 	RequirePublicIP *bool
+	// RequireLoadBalancer is set to true when VMs are using loadbalancer.
+	RequireLoadBalancer *bool
```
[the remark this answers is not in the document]
collin-woodruff-t1cg: made the change.

upup/pkg/fi/cloudup/azure/azure_cloud.go
collin-woodruff-t1cg: made the change.

upup/pkg/fi/cloudup/azure/azure_cloud.go
collin-woodruff-t1cg: made the change.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
collin-woodruff-t1cg: you were right in the first comment that we were able to get the load balancer name from the backend pools within the vmss profile. i setup something pretty much identical to the way kenji did the subnet id.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
```diff
@@ -150,11 +181,23 @@ func (s *VMScaleSet) Find(c *fi.Context) (*VMScaleSet, error) {
...
 
+	var loadBalancerID *LoadBalancerID
+	for _, i := range *ipConfig.LoadBalancerBackendAddressPools {
+		loadBalancerID, err = ParseLoadBalancerID(*i.ID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse loadbalancer ID %s", *ipConfig.Subnet.ID)
+		}
+		if !strings.Contains(loadBalancerID.LoadBalancerName, "api") {
+			continue
+		}
```
rifelpet: I'm not sure why this is here given that it is at the end of the for loop.  I realize VMSS can have multiple LBs and we're currently only supporting having the one API LB, so maybe this just needs to be moved up above `ParseLoadBalancerID` ?

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
```diff
@@ -173,6 +218,9 @@ func (s *VMScaleSet) Find(c *fi.Context) (*VMScaleSet, error) {
 		StorageProfile: &VMScaleSetStorageProfile{
 			VirtualMachineScaleSetStorageProfile: profile.StorageProfile,
 		},
+		LoadBalancer: &LoadBalancer{
+			Name: to.StringPtr(loadBalancerID.LoadBalancerName),
+		},
```
rifelpet, line 223: We should only be setting `LoadBalancer` if `loadBalancerID` was set above, otherwise we should keep `LoadBalancer` as its default value of nil.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
collin-woodruff-t1cg: made the change.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
collin-woodruff-t1cg: added an if statement to check if loadbalancerId is nil and if not then set LoadBalancer name.

pkg/model/azuremodel/api_loadbalancer.go
collin-woodruff-t1cg: yeah that's all that needs to be done for a public loadbalancer.

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
```diff
@@ -60,6 +60,35 @@ func ParseSubnetID(s string) (*SubnetID, error) {
 	}, nil
 }
 
+// LoadBalancerID contains the resource ID/names required to construct a loadbalancer ID.
+type LoadBalancerID struct {
```
rifelpet, line 64: We should probably not export this to make it clear that its not relevant outside of this package.

```suggestion
// loadBalancerID contains the resource ID/names required to construct a loadbalancer ID.
type loadBalancerID struct {
```

rifelpet (commented):
sorry one more minor fix and then we can get this merged

upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
collin-woodruff-t1cg: made the change. 

no worries! i appreciate all the help and feedback you've provided me with during the code reviewing process.

PR #10746, opened 2021-02-05
Deprecate aliyun

olemarkus:
We no longer have any contributors to this cloud provider.
Example of issues we may be facing in near future is that the implementation is currently largely based on the unofficial github.com/denverdino/aliyungo instead of github.com/aliyun/alibaba-cloud-sdk-go 

Hopefully this prompts someone to step up rather than that we have to punt it.

/milestone v1.20

### DIFF
--- a/docs/releases/1.20-NOTES.md
+++ b/docs/releases/1.20-NOTES.md
@@ -47,6 +47,8 @@
 
 * The experimental node-authorizer that could be enabled using `nodeAuthorization` has been removed. Setting this value is now forbidden.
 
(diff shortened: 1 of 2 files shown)


hakman:
👍 from me. Not sure if not better to add it to 1.19.1, as it is deprecated now too.

olemarkus:
It is unlikely the code will be a problem in 1.19, so 1.20 made more sense to me.

hakman:
OK
/lgtm

PR #10747, opened 2021-02-06
Add validation for instanceType and ami architecture 

bharath-123:
When an ami and instanceType of different architectures are specified,`kops update cluster` does not fail even though its an invalid combination.

This can cause side effects like creation of unwanted versions of a launch template, failed cluster creations and so on. 

This PR adds checks to validate the chosen instance type architecture against the ami architecture. This PR adds the validation for launch configuration, launch templates and mixed instance policies.

**Open question:** We are making a `DescribeInstanceType` call for each instance when validating for mixed instance policies. This can be expensive as more instances are added. Should we feature flag this validation?

**UPDATE:** I think the number of calls to DescribeInstanceType should be fine as this done only during spec validation and not during cluster creation.

Fixes: https://github.com/kubernetes/kops/issues/10726

### DIFF
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -65,7 +65,7 @@ func awsValidateInstanceGroup(ig *kops.InstanceGroup, cloud awsup.AWSCloud) fiel
 
 	allErrs = append(allErrs, awsValidateAdditionalSecurityGroups(field.NewPath("spec", "additionalSecurityGroups"), ig.Spec.AdditionalSecurityGroups)...)
 
-	allErrs = append(allErrs, awsValidateInstanceType(field.NewPath(ig.GetName(), "spec", "machineType"), ig.Spec.MachineType, cloud)...)
+	allErrs = append(allErrs, awsValidateInstanceTypeWithImageArch(field.NewPath(ig.GetName(), "spec", "machineType"), ig.Spec.MachineType, ig.Spec.Image, cloud)...)
 
 	allErrs = append(allErrs, awsValidateSpotDurationInMinute(field.NewPath(ig.GetName(), "spec", "spotDurationInMinutes"), ig)...)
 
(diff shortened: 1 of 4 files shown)


hakman:
Quick thought @bharath-123, when we debated validation of instance types, we decided to do it at config validation time:
https://github.com/kubernetes/kops/blob/a140168c70de521939399f19fb0bdec7dd51ffa4/pkg/apis/kops/validation/aws.go#L121-L132

Probably we should do the same for images and reduce the number of API calls later, when applying changes. I think the instance types are even cached to reduce the number of calls.

/cc @rifelpet

bharath-123:
I was not aware of this validation code. I ll work with this code. I think it makes sense to include during the deep validation phase. 

Plus I think the instanceType caching is feature flagged?

hakman:
> Plus I think the instanceType caching is feature flagged as per https://github.com/kubernetes/kops/blob/master/pkg/featureflag/featureflag.go#L52 ?

Yes, this is the PR https://github.com/kubernetes/kops/pull/9908. I think can be ignored for the initial implementation.

bharath-123:
@hakman I ll work on moving the validation logic to config validation time. I see the GetMachineTypeInfo  https://github.com/kubernetes/kops/blob/d5d08a43034dd4c7242cf1faa020cf9a8c3965e2/upup/pkg/fi/cloudup/awsup/machine_types.go#L69 which caches and returns instanceTypeInfo. If I can add a field to cache the architecture field, I think that should be useful over here, don't you think?

bharath-123:
@olemarkus @rifelpet @hakman Do have a look at this.

pkg/apis/kops/validation/aws.go
```diff
@@ -67,6 +67,8 @@ func awsValidateInstanceGroup(ig *kops.InstanceGroup, cloud awsup.AWSCloud) fiel
 
 	allErrs = append(allErrs, awsValidateInstanceType(field.NewPath(ig.GetName(), "spec", "machineType"), ig.Spec.MachineType, cloud)...)
```
hakman, line 68: Does this make sense anymore? No reason to check the instance type in 2 places.

bharath-123: Agreed we can push it to the arch instance type validation function. Will update this

rifelpet: We should also check against all instance types in a mixed instance policy in addition to spec.machineType

pkg/apis/kops/validation/aws.go
```diff
@@ -160,6 +186,7 @@ func awsValidateMixedInstancesPolicy(path *field.Path, spec *kops.MixedInstances
 	// @step: check the instance types are valid
 	for i, x := range spec.Instances {
 		errs = append(errs, awsValidateInstanceType(path.Child("instances").Index(i), x, cloud)...)
+		errs = append(errs, awsValidateInstanceTypeWithImageArch(path.Child("instances").Index(i), x, ig.Spec.Image, cloud)...)
```
bharath-123, line 189: @rifelpet here I am adding the check for the ImageArch for mixed instance policies. Will be refactoring the code to remove awsValidateInstanceType

rifelpet: ah my mistake, this lgtm 👍🏻

bharath-123:
I think this should be good for another round of review? @hakman @rifelpet

rifelpet:
I'm trying to find any validation that ensures `InstanceGroup.Spec.Image` is not empty but I don't see anything. I'm not sure if that is a valid case, but if the image were to be empty we would no longer be validating the instance type on its own.

PR #10748, opened 2021-02-06
alpha channel: Update older images

justinsb:
To pick up the sudo fixes, also addresses some problems with docker
preinstallation.

Issue #10122


justinsb:
Will be tested in the grid by https://github.com/kubernetes/test-infra/pull/20740

PR #10754, opened 2021-02-07
Use EnsureTask instead of prepending IG names to external ELB tasks

### LINKED ISSUE #10715: InstanceGroup name being prepended to externalLoadBalancer name
 this information.**
```
Version 1.19.0 (git-04d36d7d92c72601efd918877fc180c846129ffb)
```

 a `kops` flag.**
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-15T11:35:50Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.15", GitCommit:"73dd5c840662bb066a146d0871216333181f4b64", GitTreeState:"clean", BuildDate:"2021-01-13T13:14:05Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

AWS

a. Add an externalLoadBalancer to an InstanceGroup
b. `kops update cluster mycluster.tld`

```
Will modify resources:
  AutoscalingGroup/nodes-a.mycluster.tld
        LoadBalancers            [] -> [name:nodes-a.mycluster.tld-my-loadbalancer id:nodes-a.mycluster.tld-my-loadbalancer]

4739 executor.go:139] error running task "AutoscalingGroup/nodes-a.mycluster.tld" (9m59s remaining to succeed): error attaching LoadBalancers: ValidationError: Provided Load Balancers may not be valid. Please ensure they exist and try again.
        status code: 400, request id: 792b0944-adf0-4a28-a1cd-eea4c0f2c9e2
```

The externalLoadBalancer to be added to the InstanceGroup using the correct name (kOps v1.19.0-beta.3 has the correct name but suffers the bug in #10664)
```
Will modify resources:
  AutoscalingGroup/nodes-a.mycluster.tld
        LoadBalancers            [] -> [name:my-loadbalancer id:my-loadbalancer]
```

I'm assuming that the fix in #10666 has caused the InstanceGroup name to be prepended to the externalLoadBalancer name.

rifelpet:
Fixes #10715

Partially reverts https://github.com/kubernetes/kops/pull/10666

This way we end up with one CLB task per CLB regardless of how many ASGs it is attached to. Less brittle than creating many CLB tasks with concatenated names and exceptions for known CLBs.

The `Name` -> `LoadBalancerName` shouldn't make a functional difference, just improving consistency with how we attach CLBs to ASGs being created vs updated.

### DIFF
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -450,17 +450,13 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 
 	for _, extLB := range ig.Spec.ExternalLoadBalancers {
 		if extLB.LoadBalancerName != nil {
-			loadBalancerName := fi.StringValue(extLB.LoadBalancerName)
-			if loadBalancerName != awsup.GetResourceName32(b.Cluster.Name, "api") && loadBalancerName != awsup.GetResourceName32(b.Cluster.Name, "bastion") {
-				loadBalancerName = name + "-" + loadBalancerName
-			}
 			lb := &awstasks.ClassicLoadBalancer{
-				Name:             fi.String(loadBalancerName),
+				Name:             extLB.LoadBalancerName,
 				LoadBalancerName: extLB.LoadBalancerName,
 				Shared:           fi.Bool(true),
 			}
 			t.LoadBalancers = append(t.LoadBalancers, lb)
-			c.AddTask(lb)
+			c.EnsureTask(lb)
 		}
 
 		if extLB.TargetGroupARN != nil {
--- a/tests/integration/update_cluster/complex/cloudformation.json
+++ b/tests/integration/update_cluster/complex/cloudformation.json
@@ -89,6 +89,9 @@
             ]
           }
         ],
+        "LoadBalancerNames": [
+          "my-external-lb-1"
+        ],
         "TargetGroupARNs": [
           {
             "Ref": "AWSElasticLoadBalancingV2TargetGrouptcpcomplexexamplecomvpjolq"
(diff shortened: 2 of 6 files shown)


pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -450,17 +450,13 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
...
-			}
 			lb := &awstasks.ClassicLoadBalancer{
-				Name:             fi.String(loadBalancerName),
+				Name:             extLB.LoadBalancerName,
 				LoadBalancerName: extLB.LoadBalancerName,
 				Shared:           fi.Bool(true),
 			}
 			t.LoadBalancers = append(t.LoadBalancers, lb)
-			c.AddTask(lb)
+			c.EnsureTask(lb)
```
rifelpet, line 459: note for reviewers: EnsureTask returns an error if the second invocation provides a task that isn't DeepEqual:

https://github.com/kubernetes/kops/blob/0aa9cf19219b7cb53a1bb96ca7def6b9278188be/upup/pkg/fi/task.go#L70-L74

here, the CLB tasks should always be equal if a CLB is in multiple IGs' externaLoadBalancers because the only task fields that are set use the `LoadBalancerName` and `shared=true`, so this should never return an error.

hakman: Didn't know about this. May be a good idea to simplify the NLB side as well someday.

PR #10756, opened 2021-02-07
Instance roles for service accounts (IRSA) contd

olemarkus:
This is a continuation of #9271 

The workflow is changed a little bit. Users now only need to provide a `publicsDataStore` VFS (only S3 works now) and kOps will handle the rest.

This PR also works with #9352. dns-controller gets a dedicated IAM role and can assume it using the projected tokens.

Further work:
* Unfortunately, I could not figure out a way to make the uploaded files public so that is still a manual step.
* With the number of addons we'd like to provision roles for, the remap logic probably doesn't scale that well. But we can improve this using the IRSA webhook.


pkg/model/awsmodel/oidc_provider.go
```diff
@@ -43,34 +64,156 @@ const (
...
+	keyResponse := KeyResponse{Keys: keys}
+	jsonBytes, err := json.MarshalIndent(keyResponse, "", "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal json: %w", err)
+	}
+
+	return bytes.NewReader(jsonBytes), nil
+}
+
+func getFingerprint(issuer string) (string, error) {
```
olemarkus, line 198: This one will be interesting as S3 is changing certs these days. Luckily we should be good as this most likely won't make it in until 23rd March :)

olemarkus:
Added in using https://api.<clustername> as issuer as k8s 1.20 requires this.

The broken test had to be removed as PublicJWKS won't work with FS VFS. Since we try to call the https url to get the cert. If this gets merged, we can do a kubetest of this. We have the rights to write the jwks files to a public store, I believe (we are writing other assets somewhere).

The commits are a bit of a mess, so I should clean this up first.
But reviews would be welcomed.

pkg/model/awsmodel/oidc_provider.go
johngmyers: I think it might be better to either hardcode the CA cert fingerprints and/or allow them to be specified in the cluster spec. Getting them by observing what you get from a connection allows an MITM attacker to substitute out the last-listed cert.

olemarkus: My reasoning behind this was if someone wanted a different store than S3. But for well-known ones, I suppose hardcoding the fingerprints makes sense.

johngmyers: For stores other than S3 (or well-known ones) we could support a field in the cluster spec.

olemarkus: True. I think for the time being, only s3 will be supported anyway. So may as well add those two fingerprints.

pkg/model/awsmodel/oidc_provider.go
```diff
@@ -35,43 +43,162 @@ type OIDCProviderBuilder struct {
...
+	}
+
+	return bytes.NewReader(jsonBytes), nil
+}
+
+func getFingerprints() []string {
+
+	return []string{
+		"9e99a48a9960b14926bb7f3b02e22da2b0ab7280",
+		"a9d53002e97e00e043244f3d170d6f4c414104fd",
```
rifelpet, line 201: Probably worth a comment describing where these came from and how they're derived

olemarkus: Made a quick comment about that now.

pkg/model/awsmodel/oidc_provider.go
```diff
@@ -35,43 +43,163 @@ type OIDCProviderBuilder struct {
...
 	}
+	c.AddTask(discoveryFile)
 
 	c.AddTask(&awstasks.IAMOIDCProvider{
 		Name:        fi.String(b.ClusterName()),
 		Lifecycle:   b.Lifecycle,
-		URL:         fi.String(issuerURL),
+		URL:         fi.String(serviceAccountIssuer),
 		ClientIDs:   []*string{fi.String(defaultAudience)},
-		Tags:        b.CloudTags(b.ClusterName(), false),
```
justinsb, line 72: Why did we remove the Tags here?

pkg/model/awsmodel/oidc_provider.go
```diff
@@ -35,43 +43,163 @@ type OIDCProviderBuilder struct {
...
 
-	if issuerURL == "" {
-		return nil
+	skTask := signingKeyTaskObject.(*fitasks.Keypair)
+
+	keys := &OIDCKeys{
+		SigningKey: skTask,
+	}
+
+	discovery, err := buildDicoveryJSON(serviceAccountIssuer)
```
justinsb, line 92: Nit: typo "dicovery"

pkg/model/awsmodel/oidc_provider.go
```diff
@@ -35,43 +43,163 @@ type OIDCProviderBuilder struct {
...
+		Base:      fi.String(b.Cluster.Spec.PublicDataStore),
+		Public:    fi.Bool(true),
+	}
+	c.AddTask(keysFile)
+
+	discoveryFile := &fitasks.ManagedFile{
+		Contents:  fi.NewBytesResource(discovery),
+		Lifecycle: b.Lifecycle,
+		Location:  fi.String("oidc/.well-known/openid-configuration"),
+		Name:      fi.String("discovery.json"),
```
justinsb, line 110: Nit: maybe `public-oidc-configuration` (and similar prefix on keys.json).  It matters most for the terraform, I suspect (and primarily only optics there)

pkg/model/awsmodel/oidc_provider.go
```diff
@@ -35,43 +43,163 @@ type OIDCProviderBuilder struct {
...
+		ResponseTypes:         []string{"id_token"},
+		SubjectTypes:          []string{"public"},
+		SigningAlgs:           []string{"RS256"},
+		ClaimsSupported:       []string{"sub", "iss"},
+	}
+	return json.MarshalIndent(d, "", "")
+}
+
+type KeyResponse struct {
+	Keys []jose.JSONWebKey `json:"keys"`
```
justinsb, line 141: Nit: are taking the dependency on the jose library _only_ for the structs?  Should we just embed them?

upup/pkg/fi/cloudup/awstasks/iamoidcprovider.go
```diff
@@ -35,7 +35,7 @@ type IAMOIDCProvider struct {
 	Lifecycle *fi.Lifecycle
 
 	ClientIDs   []*string
-	Thumbprints []fi.Resource
+	Thumbprints []*string
```
justinsb, line 38: I thought we needed this for the initial preview... I guess if tests are passing then we don't (or we have a gap in our tests!)

upup/pkg/fi/cloudup/bootstrapchannelbuilder_test.go
```diff
@@ -53,20 +53,6 @@ func TestBootstrapChannelBuilder_BuildTasks(t *testing.T) {
 	runChannelBuilderTest(t, "awsiamauthenticator", []string{"authentication.aws-k8s-1.12"})
 }
 
-func TestBootstrapChannelBuilder_PublicJWKS(t *testing.T) {
```
justinsb, line 56: Is there are reason to drop this test?

justinsb:
Had some questions/nits but this LGTM.  As this is feature-flagged, I propose we merge and then I'll probably send a follow-on PR to try to e.g. add the test back if we do need it.

/approve
/lgtm

pkg/model/awsmodel/oidc_provider.go
rifelpet: @olemarkus this might be a merge conflict resolution issue. I added Tags in https://github.com/kubernetes/kops/pull/10832 so we probably can just add this line back.

upup/pkg/fi/cloudup/awstasks/iamoidcprovider.go
olemarkus: We used to because the thumbprint came from the cluster CA cert. Now they are just well-known strings.

I suspect this may be something we can change back in order to support anonymousauth+public apiserver as oidc provider.

upup/pkg/fi/cloudup/bootstrapchannelbuilder_test.go
olemarkus: Yeah. This PR only supports using s3 as store for jwks, so this fails because filesystem VFS cannot be used for https.

pkg/model/awsmodel/oidc_provider.go
olemarkus: Yep. This was a mistake.

rifelpet (approved):

Issue #10704, opened 2021-02-01
kops-controller enters crashloop on update to k8s 1.19.7


MichaelJDrK:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Version 1.19.0 (git-04d36d7d92c72601efd918877fc180c846129ffb)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-17T02:13:01Z", GoVersion:"go1.15.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.15", GitCommit:"73dd5c840662bb066a146d0871216333181f4b64", GitTreeState:"clean", BuildDate:"2021-01-13T13:14:05Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

1. Create a new cluster with kops 1.19.0 and Kubernetes 1.18.15. (1 master, 2 nodes)
2. Set "kubernetesVersion: 1.19.7" in the cluster.yaml
3. Update cluster.yaml in S3: "kops replace -f cluster.yaml"
4. Update cloud resources: "kops update cluster $NAME --yes"

**5. What happened after the commands executed?**

This will trigger an update to the kops-controller deployment. The kops-controller will now be in a crashloop because of the new config.yaml. 

E0201 09:34:10.278898       1 main.go:107] setup "msg"="unable to start server" "error"="reading \"ca\" certificate: open /etc/kubernetes/kops-controller/pki/ca.pem: no such file or directory"                                   
 stream closed  

kops-controller expects certificates on the node, which are not yet available, because there wasn't a rolling update yet.
You cannot do a rolling-update, because the cluster validation will not pass (because of the crashloop)

**6. What did you expect to happen?**

The cluster should remain stable. kops-controller should not enter a crashloop because of the missing certificates. There should be some smooth transition.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
[...]

PR #10758, opened 2021-02-07
Always generate kops-controller certs

hakman:
This is a quick simple fix, but can be easily be reverted and a better one added later.

Fixes: #10704

### DIFF
--- a/nodeup/pkg/model/kops_controller.go
+++ b/nodeup/pkg/model/kops_controller.go
@@ -46,10 +46,6 @@ func (b *KopsControllerBuilder) Build(c *fi.ModelBuilderContext) error {
 		Mode: s("0755"),
 	})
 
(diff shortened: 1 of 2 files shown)


justinsb:
Yes, I concur with this fix; it is consistent with our always labelling with the pki label!

Thanks @hakman!

hakman:
Thanks @justinsb. I guess now all blockers for 1.19.1 are fixed?

justinsb:
I think we also want to figure out the kubenet + containerd thing... I borrowed your trick and am trying to get that to pass...  They all pass on GCE somehow (https://k8s-testgrid.appspot.com/google-gce#gce-containerd)

I'm generally double-checking the list...  we should probably also check this fix against the two reported scenarios  (at least mentally run through the scenarios!)

PR #10759, opened 2021-02-07
kubenet containerd: match upstream

justinsb:
Configure kubenet in containerd/CNI mode to match upstream configuration.

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -103,6 +103,9 @@ func (b *ContainerdBuilder) Build(c *fi.ModelBuilderContext) error {
 		// https://github.com/containerd/containerd/blob/master/docs/cri/config.md#cni-config-template
 		if components.UsesKubenet(b.Cluster.Spec.Networking) {
 			b.buildCNIConfigTemplateFile(c)
+			if err := b.buildIPMasqueradeRules(c); err != nil {
+				return err
+			}
 		}
 
 	}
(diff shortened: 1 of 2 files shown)


hakman:
The current approach is from the containerd project instructions and and results in 1 test failing:
https://github.com/containerd/containerd/blob/master/script/setup/install-cni
```
[sig-network] Services should preserve source pod IP for traffic thru service cluster IP [LinuxOnly]
```

The `ptp` approach from this PR results in 5 tests failing:
https://github.com/kubernetes/kops/pull/10761
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/10761/pull-kops-e2e-k8s-containerd/1358693640242728960
```
[sig-network] Services [It] should have session affinity timeout work for NodePort service [LinuxOnly] [Conformance]
[sig-network] Services [It] should be able to update service type to NodePort listening on same port number but different protocols
[sig-network] Services [It] should be able to switch session affinity for NodePort service [LinuxOnly] [Conformance]
[sig-network] Services [It] should be able to change the type from ExternalName to NodePort [Conformance]
[sig-network] Services [It] should have session affinity work for NodePort service [LinuxOnly] [Conformance]
```

If you prefer to use the `ptp` approach then we should skip the `NodePort` tests.

justinsb:
Oh - thanks for confirming where it came from!  Losing the pod source IP is ... not good.  But I don't think we had this problem before, so I'm wondering if something else also changed.

But having the nodeport tests fail is also ... not good!

Thanks for pointing me in the right direction - I'll take a look and try to figure it out!

/hold

hakman:
> Oh - thanks for confirming where it came from! Losing the pod source IP is ... not good. But I don't think we had this problem before, so I'm wondering if something else also changed.

This was always a problem of the containerd kubenet integration.

justinsb:
So after one or two iterations here this is working!

The thing I was missing was that if we don't pass `networking=kubenet` to kube-proxy it doesn't set up the default destination NAT rules.  So all outbound traffic from the pods was broken; that was the NodePort failures AFAICT.  (It seems we don't have any tests that check that directly, which is kinda surprising...)

upstream added a more sophisticated masquerade rule than kube-proxy was previously installing.

New way:
https://github.com/kubernetes/kubernetes/blob/e0bba105cca88c897c60ad80d4aa0a54c984dbb5/cluster/gce/gci/configure-helper.sh#L153-L170

Old logic:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/network/kubenet/kubenet_linux.go#L197-L203

I'm inclined to say that we take the CNI/containerd opportunity to sync up with what upstream is doing.

Along those lines, upstream has also switched to ptp (and the code passed with ptp).  I'm pretty sure we can make it work with either.  I don't know whether there's an advantage to one or the other.  I'm inclined to switch to match upstream, open to arguments either way!

hakman:
Nicely done! Looks good to me either way, but upstream would probably get better support in the future.

justinsb:
OK so cleaned up and containerd is passing.  I went with ptp to match upstream, and I went with a NAT configuration that more closely matched what we did before, as I'm not convinced that the upstream configuration is suitable for AWS / anywhere where the routing might not be symmetric.  WDYT @hakman et al?

If we agree I can rebase to remove the force-kubenet trick!

nodeup/pkg/model/containerd.go
```diff
@@ -304,18 +307,74 @@ runtime-endpoint: unix:///run/containerd/containerd.sock
...
+
+	// This is based on rules from gce/cos/configure-helper.sh and the old logic in kubenet_linux.go
+
+	// We stick closer to the logic in kubenet_linux, both for compatability, and because the GCE logic
+	// skips masquerading for all private CIDR ranges, but this depends on an assumption that is likely GCE-specific.
+	// On GCE custom routes are at the network level, on AWS they are at the route-table / subnet level.
+	// We cannot generally assume that because something is in the private network space, that it can reach us.
+	// If we adopt "native" pod IPs (GCE ip-alias, AWS VPC CNI, etc) we can likely move to rules closer to the upstream ones.
+	script := `#!/bin/bash
+	# Built by kOps - do not edit
```
hakman, line 323:
```suggestion
# Built by kOps - do not edit
```

hakman:
Maybe we can discuss this more in detail during Office Hours, but I think it looks good as is now.
I vote to merge it as is and release 1.19.1, as it contains quite a lot of bug fixes.

@olemarkus Any thoughts?

justinsb:
Yes, I think we can probably reach consensus tomorrow in office hours, and I think it's time for 1.19.1.  The last issue I want to look at is 10719, though it looks like we have a fix available :-)

hakman:
Storage driver failure, retesting.
/retest

hakman:
Seems like it's time to merge :).
/lgtm

Issue #10738, opened 2021-02-05
Installing kops on GCP west4 fails with "unable to infer cloud provider from zones (is there a typo in --zones?)"


vhamburger:
**1. What `kops` version are you running? 
Version 1.19.0 (git-04d36d7d92c72601efd918877fc180c846129ffb)

**3. What cloud provider are you using?**
Google Cloud

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
kops create cluster server-cluster.k8s.local --zones us-west4-a --state "gs://$PROJECT-kops-clusters/"/ --project=$PROJECT --node-count=4

**5. What happened after the commands executed?**
ERROR: unable to infer cloud provider from zones (is there a typo in --zones?)
**6. What did you expect to happen?**
Setup the cluster

**9. Anything else do we need to know?**
The command works when applying the "--cloud gce" flag. I guess that should be documented since I couldn't find that in the readme or elsewhere - was a total trial and error game...

bharath-123:
Yup. kOps defaults to aws cloud. I think it's not very explicit that the default for the option `cloud` is aws. I think we should make it explicit in the docs that the option defaults to aws.

bharath-123:
@vhamburger you can see the docs here https://kops.sigs.k8s.io/cli/kops_create_cluster/ . The default for cloud option is not explicitly specified. Will add that.

justinsb:
@vhamburger  yes - sorry - it's a good idea to specify cloud, but you shouldn't have to (it's a bug that we don't know that us-west4-a is GCE).

@bharath-123 thanks for looking into this!  The default value for the cloud is in GuessCloudForZone.  If you don't specify cloud (which you don't normally have to), then we will use the zone to figure out which cloud you meant.  Unfortunately that list is out of date and doesn't have the latest GCE zones.

Two things I think we can do:

* Change the "typo in zones" message to suggest passing `--cloud` explicitly
* Add the latest zones/regions to the list of known zones

I'm happy to do both, but if you want to tackle them instead, just let me know!

bharath-123:
Hey @justinsb! What you say makes sense! I would love to put in the work to do this! I wrote a small PR explicitly stating the default for the `cloud` option to be aws. But I think your suggestions are better for the long term. Will raise a PR soon with you suggestions :)

PR #10771, opened 2021-02-08
Update GCE zones

bharath-123:
This PR essentially does 2 things. It updates the GCE zones and it refactors the error message thrown if the cloud provider is not inferred from the zones. 

I didn't bother making 2 seperate PRs as the changes are very small. Please do let me know if we would want to make 2 seperate PRs for this. Will do so.

Fixes: https://github.com/kubernetes/kops/issues/10738


geojaz (approved):
/lgtm

PR #10776, opened 2021-02-09
Reject update when architectures for each instance do not match

h3poteto:
closes: https://github.com/kubernetes/kops/issues/10726

I fixed to return an error when instances of different architectures are mixed.

---
I will add a test for this method, so this pull request is still WIP.

### DIFF
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -131,6 +131,35 @@ func awsValidateInstanceType(fieldPath *field.Path, instanceType string, cloud a
 	return allErrs
 }
 
+func awsValidateMixedInstanceTypes(fieldPath *field.Path, instanceTypes []string, cloud awsup.AWSCloud) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if instanceTypes == nil || cloud == nil {
+		return allErrs
+	}
+
+	var architectures []*string
+	for i, instanceType := range instanceTypes {
+		for _, typ := range strings.Split(instanceType, ",") {
+			info, err := cloud.DescribeInstanceType(typ)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


hakman:
Hi @h3poteto. I think the right solution here would be to check the arch against the image arch.
@bharath-123 stated something similar in https://github.com/kubernetes/kops/pull/10747.
Any thoughts?

bharath-123:
Hi @h3poteto I was actually working on this already in https://github.com/kubernetes/kops/pull/10747 ! The idea is to check the architecture supported by the image over the architecture supported by the instance types. The right solution would be to check whether the supported architectures for the AMI are supported for the instance. There is a catch too, we ideally want to call `DescribeInstanceType` as less as possible.
I implemented a solution but have not raised the PR yet. But here is my solution:
1. Check out this code https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/awsup/machine_types.go#L28. It caches the instance type info temporarily. We can add a field to extract the SupportedArchitectures
2. Use this new field to get the SupportedArchitectures for an instanceType and validate if the image architecture is in the SupportedArchitectures list in the awsValidateInstanceTypeFunction. 

If we nail the awsValidateInstanceType function, we should be good in the case of mixedPolicyInstances too.

Let me know if you are picking this up. I ll close my existing PR and would be glad to review your work! :)

bharath-123:
@hakman I think the right solution would be to validate the AMI arch to the archs supported by the instance type. launching instances should fail if there is a mismatch in the supported archs.

h3poteto:
@bharath-123 Oh, I'm sorry, I didn't realize your pull request.

> The right solution would be to check whether the supported architectures for the AMI are supported for the instance.

That's right. Please proceed with your work. I will close this pull request.

PR #10780, opened 2021-02-10
Make protokube CP label setting consistent with kops-controller

olemarkus:
Both kops-controller and protokube sets labels on CP nodes. Right now, protokube sets the first master label, which allows kops-controller to be scheduled. kops-controller then sets more labels. This PR consolidates the source of the labels to a common location so that when we change the set of labels CP nodes should have, we only need to do so in one place.

As an additional cleanup, I also changed protokube to select nodes by name rather than using the hostname label selector.


hakman:
@olemarkus could you add a small description? Thanks!

olemarkus:
Certainly. Done.

DOboznyi:
@olemarkus is it ready to merge?

olemarkus:
It is

DOboznyi:
> It is

So are you waiting for review?🙂

olemarkus:
Yep. Like all PRs, also this one requires a review from my peers

pkg/nodelabels/builder.go
```diff
@@ -76,3 +76,13 @@ func BuildNodeLabels(cluster *kops.Cluster, instanceGroup *kops.InstanceGroup) m
...
 	return nodeLabels
 }
+
+// BuildMandatoryControlPlaneLabels returns the list of labels all CP nodes must have
+func BuildMandatoryControlPlaneLabels() map[string]string {
+	nodeLabels := make(map[string]string)
+	nodeLabels[RoleLabelMaster16] = ""
+	nodeLabels[RoleLabelControlPlane20] = ""
+	nodeLabels[RoleLabelName15] = RoleMasterLabelValue15
+	nodeLabels["kops.k8s.io/kops-controller-pki"] = ""
```
hakman, line 86: I have some concern that this should be only for k8s 1.19+. Am I right?

olemarkus: That label is installed unconditionally on any _kOps_ 1.19+ cluster. See https://github.com/kubernetes/kops/pull/10780/files#diff-dacb133dab486c89cbabcce60635171df5295d7058d4b12565787bbfd7d46c0cL51

hakman: OK, let's see how this goes.

PR #10797, opened 2021-02-11
Allow managed images for Azure instance groups

NickSchleicher:
Allows us to use our own managed images for virtual machine scale sets and not just ones that are published to the marketplace.

Related to
`Be able to configure some more parameters for VM Scale Set such as Windows` in #10412

Not sure if it's necessary we should check for a 
`/subscriptions/<subscription id>/resourceGroups/<resource group>/providers/<provider>/images/<image>` format

### DIFF
--- a/pkg/model/azuremodel/vmscaleset.go
+++ b/pkg/model/azuremodel/vmscaleset.go
@@ -170,9 +170,7 @@ func getStorageProfile(spec *kops.InstanceGroupSpec) (*compute.VirtualMachineSca
 		storageAccountType = compute.StorageAccountTypesPremiumLRS
 	}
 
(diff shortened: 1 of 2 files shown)


rifelpet:
/ok-to-test

This looks good to me. For reference, cluster-api doesn't perform any input validation on the ID contents so I think its acceptable for kops not to either: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/29ba7f37f1f1e2274bee24bc66ef9561dd158c73/api/v1alpha3/azureimage_validation.go

NickSchleicher:
Thanks for that reference @rifelpet was curious on the protocol for things like that, will remember in the future

PR #10799, opened 2021-02-12
kubetest2 - Dump all pod logs in addition to host logs

rifelpet:
`kubectl cluster-info dump` will dump all etcd data into the provided directory organized by namespace. I'm hoping this will assist with troubleshooting our timeout issues once we add timeout support to kubetest2 so that we can invoke this shutdown process after the tests timeout.

### DIFF
--- a/tests/e2e/kubetest2-kops/deployer/dumplogs.go
+++ b/tests/e2e/kubetest2-kops/deployer/dumplogs.go
@@ -46,6 +46,10 @@ func (d *deployer) DumpClusterLogs() error {
 		return err
 	}
 
+	if err := d.dumpClusterInfo(); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


tests/e2e/kubetest2-kops/deployer/dumplogs.go
```diff
@@ -46,6 +46,10 @@ func (d *deployer) DumpClusterLogs() error {
 		return err
 	}
 
+	if err := d.dumpClusterInfo(); err != nil {
+		return err
```
justinsb, line 50: A thought  ... we may want to do these as best-effort (e.g. trying each dump step and returning a list of errors at the end)

justinsb:
Thanks @rifelpet ... assuming this passes lgtm!

What look like infra flakes, so:
/retest

/approve
/lgtm

rifelpet:
Thanks! rather than waiting for kubetest2 to support a --timeout flag I'm thinking we could add one ourselves in our tester that would interrupt [our Execute()](https://github.com/kubernetes/kops/blob/ec495195a7288d3b26550f62bb57eb77d0d6b0b5/tests/e2e/pkg/tester/tester.go#L68) call to kubetest2's ginkgo execution. Perhaps we could also issue an ABRT to the ginkgo process to get its stack traces.

PR #10805, opened 2021-02-12
WIP Kubetest2 - Add a timeout flag and issue SIGABRT to child processes 

rifelpet:
The goal here is to catch when the timeouts are happening in prow and ensure our shutdown + log dump processes happens cleanly.

I'm also hoping we can issue a SIGABRT to the e2e.test process which will dump its stack trace, also helping in the timeout troubleshooting.


rifelpet:
I discovered some flags built into the e2e tests and ginkgo that i'm going to experiment with first before proceeding with this strategy.

moving my efforts here: https://github.com/kubernetes/test-infra/pull/20850

PR #10812, opened 2021-02-13
iptables: Use the lock when checking for existing rules

justinsb:
Otherwise we sometimes get an "in-use" message, telling us to use the
lock, if another iptables process is running concurrently.


justinsb:
/retest

Unrelated flake from curl

PR #10828, opened 2021-02-15
Allow multi-CNI setups to set usesSecondaryIP

ravens:
This PR removes a previously introduced check on the CNI plugin presence. 

The original PR https://github.com/kubernetes/kops/pull/8617 seems to introduce some checks on the use of incompatible CNI plugins, but in fact the CNI plugin is used to set the `--node-ip` Kubelet option in case the `useSecondaryIP` flag is set to true. A Kops cluster operator should be able to declare that when using a CNI plugin like Calico for example.

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -548,6 +548,10 @@ func validateNetworking(cluster *kops.Cluster, v *kops.NetworkingSpec, fldPath *
 		optionTaken = true
 	}
 
+	if v.CNI != nil && optionTaken {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("cni"), "only one networking option permitted"))
+	}
+
 	if v.Weave != nil {
 		if optionTaken {
 			allErrs = append(allErrs, field.Forbidden(fldPath.Child("weave"), "only one networking option permitted"))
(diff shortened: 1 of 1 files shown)


olemarkus:
This option does prevent the install of incompatible networking options. The check you remove in this PR would allow specifying a user-managed CNI in addition to a kOps-managed CNI (or kubeet-like networking), which kOps cannot support.

You can however set `spec.networking.cni` and manage two CNIs independently yourself.

Could you elaborate more on your usecase for this?

ravens:
This case (cni + another cni) used to work with kops 1.17 - our use case on AWS is to use calico as main CNI and to use Multus to bring ENI to specific pods. Without the CNI option setup, Kubelet does not set up the --node-ip option and Kubelet starts to grab up all the IPs, creating duplicate internalIPs after a while. 

While I do see some CNI options being incompatible (i.e. Calico + Cilium), this particular one should not be marked as incompatible IMO.

olemarkus:
So it is actually `spec.networking.cni.usesSecondaryIP` that you want to configure. It is perhaps not placed in the in the best struct for this ...

I think maybe if you move this `if` block below `kopeio` and remove the `optionTaken = true` part from the block we achieve what we want. CNI cannot be combined with kubenet-like networking (a lot of the code in kOps would break over this), but it can be combined with any actual CNIs.

olemarkus:
/cc @hakman
you may have an opinion on this as well, perhaps

ravens:
@olemarkus thanks for the suggestion ! Indeed the focus is actually on the `spec.networking.cni.usesSecondaryIP` option. I amended the PR to reflect your proposal.

olemarkus:
Thanks. Lets see if anyone else has any comments. If not, I'll merge in not too long.

/assign

olemarkus (approved):

PR #10846, opened 2021-02-16
Actually enable systemd cgroup for containerd

codablock:
#10419 forgot to actually enable the systemd cgroup for containerd.

I assume this must be backported into kops 1.20.


bharath-123:
@codablock check this out https://github.com/kubernetes/kops/pull/10419#discussion_r547394545 .

bharath-123:
@codablock I think we should only set SystemdCgroup=true for k8s version >= 1.20. Plus there seems to be some merge conflicts.

hakman:
> @codablock I think we should only set SystemdCgroup=true for k8s version >= 1.20. Plus there seems to be some merge conflicts.

@bharath-123 I think it's ok as is, same as for runtime v2.
I would appreciate an explanation about why this is needed.

bharath-123:
Based on our previous discussions, I think we decided to enable this only for k8s cluster >= 1.20. I had added the if stmt here https://github.com/kubernetes/kops/blob/master/pkg/model/components/kubelet.go#L226 to enable this only on k8s versions >= 1.20. This would break for clusters created with versions < 1.20 and with containerd.

hakman:
How would it break older clusters?

bharath-123:
sorry i meant new clusters. I was not able to bring up a cluster with k8s 1.19.7 with this patch applied. because kubelet is running cgroupfs and containerd is using systemd.

hakman:
OK, agreed, but for older versions I would prefer to set it to 'false' explicitly.

bharath-123:
Sounds good @hakman . Thanks for this PR @codablock . Do resolve the merge conflicts and do set it to false for clusters with versions less than 1.20.

codablock:
@bharath-123  @hakman done. Not sure about the test failures...I'll check after it has re-run.

pkg/model/components/containerd.go
```diff
@@ -62,6 +62,9 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 				config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "registry", "mirrors", name, "endpoint"}, endpoints)
 			}
 			config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", "runc", "runtime_type"}, "io.containerd.runc.v2")
+			// only enable systemd cgroups for kubernetes >= 1.20
+			systemdCgroup := b.IsKubernetesGTE("1.20")
+			config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", "runc", "options", "SystemdCgroup"}, systemdCgroup)
```
hakman, line 67:
```suggestion
			config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", "runc", "options", "SystemdCgroup"}, b.IsKubernetesGTE("1.20"))
```

codablock: applied and squashed into initial commit

hakman:
I think the bot has gone 🍌

bharath-123:
not sure why the ci robot is going berserk here.

PR #10864, opened 2021-02-17
Update go to 1.16 for PR jobs

bharath-123:
https://blog.golang.org/go1.16 go 1.16 is released


hakman:
Thanks @bharath-123, but we plan to wait a bit more for this change, at least until Kubernetes does the move.
/close

PR #10870, opened 2021-02-18
Add more `kops toolbox template` functions

kesor:
Add additional functions to templates used by `kops toolbox template`:
* `toYaml`/`fromYaml`/`fromYamlArray`
* `toJson`/`fromJson`/`fromJsonArray`

### DIFF
--- a/pkg/util/templater/template_functions.go
+++ b/pkg/util/templater/template_functions.go
@@ -43,6 +43,10 @@ func (r *Templater) templateFuncsMap(tm *template.Template) template.FuncMap {
 		return content
 	}
 
+	// More template functions borrowed from Helm project, originally added in
+	// - https://github.com/helm/helm/pull/1686
+	// - https://github.com/helm/helm/pull/2000
+	// - https://github.com/helm/helm/pull/7712
 	funcs["toYaml"] = toYAML
 	funcs["fromYaml"] = fromYAML
 	funcs["fromYamlArray"] = fromYAMLArray
--- a/pkg/util/templater/template_functions.go
+++ b/pkg/util/templater/template_functions.go
@@ -43,6 +43,13 @@ func (r *Templater) templateFuncsMap(tm *template.Template) template.FuncMap {
 		return content
 	}
 
+	funcs["toYaml"] = toYAML
+	funcs["fromYaml"] = fromYAML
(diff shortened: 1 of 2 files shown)


olemarkus:
Hi @kesor,

Can you describe the use case for these functions?

kesor:
> Hi @kesor,
> 
> Can you describe the use case for these functions?

This is a common pattern used in Helm chart templates. In the context of Kops, here is one of the many examples we use in our templated clusters -

We want to have `etcd` receive common environment variables into both main and events clusters for each cluster. In the cluster template we use something like this for each availability zone :

```
etcdClusters:
  manager:
    env: {{ $etcd.common.env | toYaml | nindent 6 }}
```

And in the variables provided we include a definition like this:
```
etcd:
  common:
    env:     
      - name: ETCD_LOGGER
        value: zap
      - name: ETCD_METRICS
        value: extensive
      - name: ETCD_ENABLE_PPROF
        value: "true"
      - name: ETCD_QUOTA_BACKEND_BYTES
        value: "6442450944"
      - name: ETCD_SNAPSHOT_COUNT
        value: "5000"
```

In some cases where a "single line" needs to be added we could use `toJson` to convert a multi-line YAML object into a single-line string. These things have been done in Haml all over the place, for more examples search `fromYaml` in https://github.com/helm/charts

olemarkus:
Ah. I see these functions are more or less copied from helm3? We should comment where they are from.

We already vendor helm3 for handling values .. wonder if we can just use the helm3 engine for rendering the templates too. That would also make these functions available.

olemarkus:
Using the helm engine won't work due to various design decisions made by helm.

Question is if not kops solves this use case with snippets: https://kops.sigs.k8s.io/operations/cluster_template/#snippets

kesor:
> Using the helm engine won't work due to various design decisions made by helm.
> 
> Question is if not kops solves this use case with snippets: https://kops.sigs.k8s.io/operations/cluster_template/#snippets

We have been using a fork of `kops` with these additional functions, and it has been working just fine. Why wouldn't it work?

olemarkus:
The functions are not a problem, but need to be copied over and attributed as mentioned above.

kesor:
Added attribution to the original contributions, and did rebase on kops `master` branch.

PR #10871, opened 2021-02-18
Only update kops-controller pods on deletion

olemarkus:
This will in most cases mean kops-controller is only updated when the CP node is rolled and should increase the chances of compatible controller<->CP updates

### DIFF
--- a/upup/models/cloudup/resources/addons/kops-controller.addons.k8s.io/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/kops-controller.addons.k8s.io/k8s-1.16.yaml.template
@@ -25,9 +25,7 @@ spec:
     matchLabels:
       k8s-app: kops-controller
   updateStrategy:
-    type: RollingUpdate
-    rollingUpdate:
-      maxUnavailable: 1
+    type: OnDelete
   template:
     metadata:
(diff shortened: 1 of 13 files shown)


hakman:
`kops-controller` uses leader election to decide the leader. Just noticed that the mechanism changed there from ConfigMaps locking to Lease API. Update on delete would break upgrading to the new mechanism.
https://github.com/kubernetes-sigs/controller-runtime/pull/1144

Not sure what the plan is long term, but having new and old `kops-controller` at the same time in the cluster may generate unexpected results depending on how old the leader is and what are the tasks that it has to do.

olemarkus:
The only controller that kops-controller is running is node labeller. This one only does idempotent operations and can run perfectly fine in parallel. For CP nodes, it already fights with protokube.

Running rolling-update of kops-controller will result in similar conflict, only for a (much) shorter duration.

johngmyers:
The leader election method change would also be an issue for the status quo, just for the shorter amount of time that it takes for the DaemonSet controller to do its rolling update.

/lgtm
/approve
/hold for any last-minute objections

rifelpet (approved):

rifelpet:
@olemarkus can you rebase this?

Issue #10706, opened 2021-02-01
[Question / Feature Request] AWS NLB - EIP Allocations


timothyclarke:
Firstly thanks for the work people have put in for the the Network Loadbalancer in AWS clusters.

Is there a way (without using `-out ... --target=terraform` and editing the tf config) to use IP allocations so we can make use of static IP's?

I haven't thought through the cli elements to create the cluster, but I would expect that the spec would be something like
```
spec:
  api:
    loadBalancer:
      class: Network
      type: Public
      elasticIPs:
      - eipalloc-000abc123
      - eipalloc-111def456
      - eipalloc-222ghi789
```

bharath-123:
I don't think kOps currently supports specifying allocationIds of elasticIps to attach to a network load balancer. You probably would have to manually associate an elastic ip at this point of time. 

But I do see a way to do this in kOps. The `SubnetMapping` struct in `CreateLoadBalancerInput` accepts an allocationId for an elasticIp for a specified subnet.([link](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2@v1.1.0/types#SubnetMapping)) which can be used to associate elastic ips to the nlb.

bharath-123:
I think if we can specify a mapping of subnetId to allocationId of the elastic ip in the lb spec, we should be able to associate elastic ips to the nlb.

rifelpet:
Agreed, I think it would be reasonable to support this. We'll need the new API fields to be able to represent which subnet ID to use for each allocation ID. I _believe_ that the [name](https://github.com/kubernetes/kops/blob/9d0dd3c4c1957f829504376778d0ee67f8f4d28f/tests/integration/update_cluster/complex/in-v1alpha2.yaml#L68) field of subnets in the ClusterSpec are unique, so we might be able to get away with referencing them by name


Maybe something along the lines of:

```yaml
spec:
  api:
    loadBalancer:
      class: Network
      type: Public
      subnetMapping: # or a more generic term for this
      - subnet: us-east-1a-public
        allocationID: eipalloc-111def456
      - subnet: us-east-1b-public
        allocationID: eipalloc-222ghi789
```

Then we just find the subnet ID associated with each subnet name and pass all of that into the CreateLoadBalancerInput.

timothyclarke:
Shouldn't the subnet ID's come (indirectly) from the master instance groups ?
[...]

PR #10872, opened 2021-02-18
Adding Elastic IP Allocations to NLB API

timothyclarke:
This is an implementation of https://github.com/kubernetes/kops/issues/10706
This allows (static) Elastic IP's to be added to the external loadbalancer of admin interface.
The Loadbalancer must be Network and Public for this change

### DIFF
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -241,6 +241,16 @@ func awsValidateLoadBalancerSubnets(fieldPath *field.Path, spec kops.ClusterSpec
 				allErrs = append(allErrs, field.Forbidden(fieldPath.Index(i).Child("privateIPv4Address"), "privateIPv4Address only allowed for internal NLBs"))
 			}
 		}
+
+		if subnet.AllocationID != nil {
+			if *subnet.AllocationID == "" {
+				allErrs = append(allErrs, field.Required(fieldPath.Index(i).Child("allocationID"), "allocationID can't be empty"))
+			}
+
+			if lbSpec.Class != kops.LoadBalancerClassNetwork || lbSpec.Type == kops.LoadBalancerTypeInternal {
+				allErrs = append(allErrs, field.Forbidden(fieldPath.Index(i).Child("allocationID"), "allocationID only allowed for Public NLBs"))
+			}
+		}
 	}
 
 	return allErrs
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -242,14 +242,8 @@ func awsValidateLoadBalancerSubnets(fieldPath *field.Path, spec kops.ClusterSpec
 			}
 		}
 
-		if subnet.AllocationID != nil {
-			if *subnet.AllocationID == "" {
-				allErrs = append(allErrs, field.Required(fieldPath.Index(i).Child("allocationID"), "allocationID can't be empty"))
-			}
-
-			if lbSpec.Class != kops.LoadBalancerClassNetwork || lbSpec.Type == kops.LoadBalancerTypeInternal {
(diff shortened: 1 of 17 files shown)


timothyclarke:
Lots of thanks to @rifelpet for helping me get the changes to this point and @codablock for adding a similar change in https://github.com/kubernetes/kops/pull/10741 which gave me a basis for this.

pkg/apis/kops/validation/aws.go
```diff
@@ -241,6 +241,16 @@ func awsValidateLoadBalancerSubnets(fieldPath *field.Path, spec kops.ClusterSpec
 				allErrs = append(allErrs, field.Forbidden(fieldPath.Index(i).Child("privateIPv4Address"), "privateIPv4Address only allowed for internal NLBs"))
 			}
 		}
+
+		if subnet.AllocationID != nil {
+			if *subnet.AllocationID == "" {
```
bharath-123, line 246: Can we use fi.StringValue over here? Something like:
```
if fi.StringValue(subnet.AllocationID) == "" {

}
```
That takes care of the case when subnet.AllocationID is nil too?

timothyclarke: My GO isn't that great so I'm not sure how much of a difference there is between `!= nil` and ` ==""`

This conditional was mostly a copy of the above `PrivateIPv4Address` but without the `ParseIP` and switching the condition on the `lbSpec.Type` to require a Public rather than Internal NLB

I'm happy to switch to `if fi.StringValue(subnet.AllocationID) != "" {` and keep the `lbSpec` conditionals on 250.
Does `!= ""` cover where AllocationID is `nil` or undefined value

bharath-123: Yup I ll paste the implementation of fi.StringValue here:
```
func StringValue(s *string) string {
	if s == nil {
		return ""
	}
	return *s
}
```
subnet.AllocationId!=nil checks if the string pointer is nil but subnet.AllocationId=="" checks if the pointer points to an empty string. The string pointer can be non nil but the string which the pointer is pointing to can be "". fi.StringValue returns "" if the pointer nil. So it seems cleaner. You can avoid 2 if stmts and make it to one.....

bharath-123: I think fi.StringValue should take care of nil case too. I think its nicer since almost the entire kOps codebase uses it.... Do try it out once...

timothyclarke: @bharath-123 Please take a look. 
I've reduced the conditional. As `fi.StringValue(subnet.AllocationID) != ""` is also covering the set but empty string I only needed to handle the loadbalancer being of the wrong type (classic or internal)

bharath-123: Got it. I missed this use case of validating AllocationId to not be empty when it is present.

upup/pkg/fi/cloudup/awstasks/network_load_balancer.go
```diff
@@ -362,6 +362,12 @@ func (e *NetworkLoadBalancer) Find(c *fi.Context) (*NetworkLoadBalancer, error)
 				}
 				sm.PrivateIPv4Address = a.PrivateIPv4Address
 			}
+			if a.AllocationId != nil {
+				if sm.AllocationID != nil {
+					return nil, fmt.Errorf("NLB has more then one AllocationID per subnet, which is unexpected. This is a bug in kOps, please open a GitHub issue.")
```
hakman, line 367: Nit, but nice to fix. Same for above.
```suggestion
					return nil, fmt.Errorf("NLB has more than one AllocationID per subnet, which is unexpected. This is a bug in kOps, please open a GitHub issue.")
```

hakman (approved):

Issue #10777, opened 2021-02-09
Install CSI driver by default in preparation of CSI Migration


Jiawei0227:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

CSI Migration is a Kubernetes feature that when turn on, it will redirect in-tree plugin traffic to the corresponding CSI driver. It has been Beta in k8s since v1.17 without turning on by default.

Recently, we decide to push this feature forward and it will be turn on by default in v1.22 for a lot of plugins according to our [plan]((https://docs.google.com/document/d/1j6uF9sOoGCQ9nNFVrdvCWyn77o9k3-758qsEj2K3BAY/edit)).

It would be good if kOps can prepare for this upcoming change. Specifically, kOps should deploy the corresponding CSI drivers by default for the corresponding cloud. The driver is a requisite for CSI migration to work.

- [x] AWS - AWS EBS CSI Driver
- [ ] GCP - GCE PD CSI Driver
- [ ] Azure - Azuredisk/Azurefile CSI Driver

**2. Background**

- [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/625-csi-migration/README.md)
- [Tracking Issues](https://github.com/kubernetes/enhancements/issues?q=is%3Aissue+is%3Aopen+csi+migration)
- [Original design doc](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/csi-migration.md)
- [Public documentation](https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/)
- [CSI Migration meeting doc](https://docs.google.com/document/d/1j6uF9sOoGCQ9nNFVrdvCWyn77o9k3-758qsEj2K3BAY/edit)

/kind feature
/cc @msau42

olemarkus:
Can you clarify which feature you refer to? There are
* CSIMigrationAWS, which will use the CSI driver if installed, but fallback otherwise. Hence does not strictly need the driver. Thi sis the flag that is currently in beta, but disabled
* CSIMigrationAWSComplete, which strictly requires the driver and is currently in alpha.

We have support for the AWS CSI driver, where we will set the latter flag This is also used by the aws-cloud-provider e2e testing. I doubt the we will enabled the addon in 1.21 though. We recently discussed this and decided not to. We may extend our e2e grid to include testing with this enabled though. kOps 1.21 is many months away so there should be ample time reconsider later.
 
[...]

PR #10879, opened 2021-02-18
Fixes for 1.21 e2e tests

olemarkus:
* Set the correct jwks path. Should fix the failing OIDC test.
* Feature gate for CSI deregistration changed name. See #10777


pkg/model/components/discovery.go
```diff
@@ -55,20 +54,8 @@ func (b *DiscoveryOptionsBuilder) BuildOptions(o interface{}) error {
 		kubeAPIServer.ServiceAccountIssuer = &serviceAccountIssuer
 	}
 
-	// We set apiserver ServiceAccountKey and ServiceAccountSigningKeyFile in nodeup
-
-	if useJWKS {
-		if kubeAPIServer.FeatureGates == nil {
-			kubeAPIServer.FeatureGates = make(map[string]string)
-		}
-		kubeAPIServer.FeatureGates["ServiceAccountIssuerDiscovery"] = "true"
```
hakman, line 64: This part is removed because `ServiceAccountIssuerDiscovery` is enable by default in 1.20+, right?

olemarkus: Correct

hakman:
👍 
/lgtm

PR #10886, opened 2021-02-19
fix loadBalancerID null pointer

collin-woodruff-t1cg:
I missed one of the null pointers but I fixed it in this pr. 

*e.LoadBalancer.Name is null pointer but i've moved it inside a check for nil.


hakman:
Thanks @collin-woodruff-t1cg 
/ok-to-test
/lgtm
/approve

PR #10888, opened 2021-02-19
Use internal api url for jwks

olemarkus:
The public api url cannot be used by pods and nodes if access is restricted. So by default we need to use the internal one.
This should finally pass the OIDC e2e test


olemarkus:
As discussed in the office hours, this is needed for the conformance test to work (and also for token volume projections to work in many cases).

This can now be merged.

/approve

PR #10890, opened 2021-02-20
WIP - update presubmit job to match periodic job that times out

rifelpet:
this will make it easier to iterate on


rifelpet:
It seems adding an extra gig of memory to the prow jobs fixed the timeout issue. closing this since we no longer need this troubleshooting

PR #10897, opened 2021-02-21
Enforce 1.14 deprecation

olemarkus:
Also remove bits of code referencing 1.14 version


rifelpet:
There are 1.12 and 1.13 jobs in testgrid that we should remove

PR #10912, opened 2021-02-23
Instruct GH to collapse BUILD.bazel diffs by default

rifelpet:
Since these are generated, viewing their diff is not of much value when viewing a commit or PR.

I would include a trivial change to a BUILD.bazel file to demonstrate but it would fail the gazelle verification CI

See https://github.com/github/linguist#using-gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+**/zz_generated.*.go linguist-generated=true
+BUILD.bazel linguist-generated=true
+bindata.go linguist-generated=true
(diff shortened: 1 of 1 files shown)


hakman:
Maybe also add entries for docs and vendor and maybe bindata.go?

rifelpet:
good idea. updated.

rifelpet:
https://github.com/kubernetes/kops/pull/10929 shows what the diffs would look like

olemarkus (approved):
Big thumbs up from me!

PR #10913, opened 2021-02-23
Honor OS update policy at InstanceGroup level too

seh:
As with the _Cluster_ "spec.updatePolicy" field, add a similar field at the _InstanceGroup_ level, allowing overriding of the cluster-level choice in each _InstanceGroup_.

Introduce a new value for the field ("automatic") as equivalent to the default value applied when the field is absent. Honoring this new value allows disabling automatic updates at the cluster level, but then enabling them again for particular _InstanceGroups_. Without such a positive affirmation, it's not possible to override a cluster-level "external" policy at the _InstanceGroup_ level, as there's no way to specify positively that you want to recover the default value. Instead, expressing the explicit "automatic" value is clear and unambiguous.

See [this thread in the "kops-users" channel](https://kubernetes.slack.com/archives/C3QUFP0QM/p1614031129048800) of the "Kubernetes" Slack workspace for preceding discussion.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/kops/10913)

### DIFF
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -133,8 +133,8 @@ type ClusterSpec struct {
 	IsolateMasters *bool `json:"isolateMasters,omitempty"`
 	// UpdatePolicy determines the policy for applying upgrades automatically.
 	// Valid values:
-	//   'external' do not apply updates automatically - they are applied manually or by an external system
-	//   missing: default policy (currently OS security upgrades that do not require a reboot)
+	//   'automatic' (default): apply updates automatically (apply OS security upgrades, avoiding rebooting when possible)
+	//   'external': do not apply updates automatically; they are applied manually or by an external system
 	UpdatePolicy *string `json:"updatePolicy,omitempty"`
 	// ExternalPolicies allows the insertion of pre-existing managed policies on IG Roles
 	ExternalPolicies *map[string][]string `json:"externalPolicies,omitempty"`
--- a/pkg/apis/kops/instancegroup.go
+++ b/pkg/apis/kops/instancegroup.go
@@ -82,9 +82,9 @@ var (
 	SupportedFilesystems = []string{BtfsFilesystem, Ext4Filesystem, XFSFilesystem}
 )
 
-// InstanceGroupSpec is the specification for a instanceGroup
+// InstanceGroupSpec is the specification for an InstanceGroup
 type InstanceGroupSpec struct {
-	// Type determines the role of instances in this group: masters or nodes
+	// Type determines the role of instances in this instance group: masters or nodes
 	Role InstanceGroupRole `json:"role,omitempty"`
 	// Image is the instance (ami etc) we should use
 	Image string `json:"image,omitempty"`
@@ -92,7 +92,7 @@ type InstanceGroupSpec struct {
 	MinSize *int32 `json:"minSize,omitempty"`
 	// MaxSize is the maximum size of the pool
 	MaxSize *int32 `json:"maxSize,omitempty"`
-	// Autoscale determines if autoscaling will be enabled for the group if cluster autoscaler is enabled
+	// Autoscale determines if autoscaling will be enabled for this instance group if cluster autoscaler is enabled
 	Autoscale *bool `json:"autoscale,omitempty"`
 	// MachineType is the instance class
 	MachineType string `json:"machineType,omitempty"`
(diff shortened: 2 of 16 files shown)


pkg/apis/kops/cluster.go
```diff
@@ -133,8 +133,8 @@ type ClusterSpec struct {
 	IsolateMasters *bool `json:"isolateMasters,omitempty"`
 	// UpdatePolicy determines the policy for applying upgrades automatically.
 	// Valid values:
-	//   'external' do not apply updates automatically - they are applied manually or by an external system
-	//   missing: default policy (currently OS security upgrades that do not require a reboot)
```
seh, line 137: Note that this avoidance of reboots was and is not true for Flatcar Container Linux. It _only_ installs updates by way of rebooting. It downloads and verifies them ahead of time, but only pivots to using the new files upon reboot.

seh (commented):
I remain uncertain whether we need to change file _upup/pkg/fi/nodeup/nodetasks/package.go_. At present, it pays attention to the "external" policy, but its treatment might be inverted accidentally.

seh (commented):
I asked @hakman whether I should build a `loader.OptionsBuilder` for this field, or to add it to an existing one, in order to set the default value to "automatic." We could consider that for the `ClusterSpec` type, but should **not** do so for the `InstanceGroupSpec` type, because the latter should default to not overriding the former, by way of making no statement about its policy.

pkg/apis/kops/instancegroup.go
```diff
@@ -174,6 +174,11 @@ type InstanceGroupSpec struct {
 	CompressUserData *bool `json:"compressUserData,omitempty"`
 	// InstanceMetadata defines the EC2 instance metadata service options (AWS Only)
 	InstanceMetadata *InstanceMetadataOptions `json:"instanceMetadata,omitempty"`
+	// UpdatePolicy determines the policy for applying upgrades automatically.
```
seh, line 177: Add a note here that this overrides the "spec.updatePolicy" field in the _Cluster_ spec.

pkg/apis/kops/v1alpha2/instancegroup.go
```diff
@@ -172,6 +172,11 @@ type InstanceGroupSpec struct {
 	CompressUserData *bool `json:"compressUserData,omitempty"`
 	// InstanceMetadata defines the EC2 instance metadata service options (AWS Only)
 	InstanceMetadata *InstanceMetadataOptions `json:"instanceMetadata,omitempty"`
+	// UpdatePolicy determines the policy for applying upgrades automatically.
```
seh, line 175: Add a note here that this overrides the "spec.updatePolicy" field in the _Cluster_ spec.

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -140,6 +140,8 @@ func ValidateInstanceGroup(g *kops.InstanceGroup, cloud fi.Cloud) field.ErrorLis
 		allErrs = append(allErrs, validateExternalLoadBalancer(&lb, path)...)
 	}
 
+	allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "updatePolicy"), g.Spec.UpdatePolicy, []string{kops.UpdatePolicyAutomatic, kops.UpdatePolicyExternal})...)
```
hakman, line 143: Why not check if `g.Spec.UpdatePolicy != nil`?
Same in other places?

seh: That was me using the same `IsValidValue` function that we use [when validating the similar _Cluster_ field](https://github.com/kubernetes/kops/blob/4e8c6558993124af16c6a14efc88548c88daa805/pkg/apis/kops/validation/validation.go#L108), which I used as my example. [`IsValidValue` checks if the pointer argument is nil.](https://github.com/kubernetes/kops/blob/4e8c6558993124af16c6a14efc88548c88daa805/pkg/apis/kops/validation/helpers.go#L38)

Does that address your concern, or have I misunderstood you?

hakman: Yes. All good. Thanks.

seh:
I was able to rebase this cleanly, two documentation conflicts notwithstanding. PTAL.

seh:
@johngmyers, can you speak to the concern that I raised in [this Slack message](https://kubernetes.slack.com/archives/C3QUFP0QM/p1614094895070000?thread_ts=1614031129.048800&cid=C3QUFP0QM) regarding [`(*Package).findDpkg`](https://sourcegraph.com/github.com/kubernetes/kops/-/blob/upup/pkg/fi/nodeup/nodetasks/package.go#L201-L203) and `(*Package).FindYum`?

seh:
> can you speak to the concern that I raised in [this Slack message](https://kubernetes.slack.com/archives/C3QUFP0QM/p1614094895070000?thread_ts=1614031129.048800&cid=C3QUFP0QM) regarding [`(*Package).findDpkg`](https://sourcegraph.com/github.com/kubernetes/kops/-/blob/upup/pkg/fi/nodeup/nodetasks/package.go#L201-L203) and `(*Package).FindYum`?

@hakman set me straight. He's the original author of the blocks in question, and he explained that it's working as intended, but needs to be augmented to take my new _InstanceGroup_-level field into account.

seh:
Per private discussion with @hakman, in commit 2fc68564be758abd1714a9a3d32b2e93ad78ce0c I noted that for now we can't take this new _InstanceGroup_-level overriding of the _Cluster_-level update policy into account for the _dpkg_ and _YUM_ OS package installation tasks.

hakman (approved):

PR #10915, opened 2021-02-23
add support for azure public loadbalancer

collin-woodruff-t1cg:


rifelpet:
/ok-to-test

/cc @kenji-cloudnatix 

Looks reasonable to me. I noticed we give the public IP address the same name as the load balancer. Can a load balancer have more than one public IP address? If so we may want to suffix the IP address name.

kenji-cloudnatix (approved):
Looks good to me too!

collin-woodruff-t1cg:
I think it is possible for the load balancer to have multiple IP addresses. How would you want me to suffix the name?

rifelpet:
hm it would make sense to suffix the public iP name with something that would differentiate it in situations in which we'd have multiple public IPs. From my cursory search of the docs it seems one would use multiple public IPs for differentiating traffic that uses the same port and protocol. I can't imagine a need for that for our api load balancer, so perhaps keeping the name identical to the load balancer is sufficient.

https://docs.microsoft.com/en-us/azure/virtual-network/public-ip-addresses#internet-facing-load-balancers
https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-multivip-overview

rifelpet (approved):

PR #10920, opened 2021-02-25
Fix typos in docs/getting_started

roim:
Found this while going through docs. I then ran a quick check on everything under getting_started.

Admins: feel free to patch directly instead of accepting the PR.


roim:
some bad commit metadata, will send another pr

PR #10924, opened 2021-02-25
Fix nil pointer deference for image ID with spotinst

hakman:
/cc @liranp @olemarkus


liranp:
Thanks, @hakman.

/lgtm
/approve

PR #10931, opened 2021-02-25
Update SSH documentation for ubuntu

jpugliesi:
We've found that, with kops 1.19, the bastion and k8s nodes user is `ubuntu` when using the default kops configuration


hakman:
Thanks @jpugliesi.
/approve

PR #10932, opened 2021-02-25
copy protokube and channels binaries to upload dir in bazel-version-ci

bharath-123:
@hakman I think we would have to add this to the 1.21-alpha release?
/cc @hakman


hakman:
Let's just keep this in the protokube PR, there will have to be other similar changes there to remove stuff.

bharath-123:
Got it. Will do so.

PR #10934, opened 2021-02-26
Add support for CPU Credits on AWS t2 and t3 instance families

### LINKED ISSUE #4032: Add the ability to deploy T2 Unlimited instances on AWS
1. AWS recently announced [T2 Unlimited instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/t2-unlimited.html), which is essentially just a setting on a t2 class instance that will make it "purchase" extra credits instead of returning to the baseline performance level once credits are exhausted. This would be ideal for a master in a low-utilization cluster, as it would allow for a cheaper instance type, while also not running the risk of baseline performance causing problems with system components. Kops should allow the user to deploy a T2 unlimited by specifying it in an instance group configuration.

2. This might be annoying to implement, since it is an additional configuration option specific to a certain class of instances.

Essentially, you either call RunInstances with CreditSpecification=unlimited if they are new, or run ModifyInstanceCreditSpecification if it is being updated. See docs [here](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html).

As for configuration, two possibilities that come to mind would be either including specific machineTypes for unlimited instances (e.g. t2.medium.unlimited) or adding an additional configuration option such as machineCreditSpecification.

rifelpet:
closes #4032

### DIFF
--- a/k8s/crds/kops.k8s.io_instancegroups.yaml
+++ b/k8s/crds/kops.k8s.io_instancegroups.yaml
@@ -101,6 +101,10 @@ spec:
                 description: CompressUserData compresses parts of the user data to
                   save space
                 type: boolean
+              cpuCredits:
+                description: CPUCredits is the credit option for CPU Usage on burstable
+                  instance types (AWS only)
+                type: string
               detailedInstanceMonitoring:
                 description: DetailedInstanceMonitoring defines if detailed-monitoring
                   is enabled (AWS only)
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -79,6 +79,10 @@ func awsValidateInstanceGroup(ig *kops.InstanceGroup, cloud awsup.AWSCloud) fiel
 		allErrs = append(allErrs, awsValidateInstanceMetadata(field.NewPath("spec", "instanceMetadata"), ig.Spec.InstanceMetadata)...)
 	}
 
+	if ig.Spec.CPUCredits != nil {
+		allErrs = append(allErrs, awsValidateCPUCredits(field.NewPath("spec"), &ig.Spec, cloud)...)
+	}
+
 	return allErrs
 }
 
@@ -275,5 +279,12 @@ func awsValidateLoadBalancerSubnets(fieldPath *field.Path, spec kops.ClusterSpec
 		}
 	}
 
+	return allErrs
+}
+
+func awsValidateCPUCredits(fieldPath *field.Path, spec *kops.InstanceGroupSpec, cloud awsup.AWSCloud) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, IsValidValue(fieldPath.Child("cpuCredits"), spec.CPUCredits, []string{"standard", "unlimited"})...)
 	return allErrs
 }
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -177,6 +177,13 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 			lt.RootVolumeThroughput = fi.Int64(int64(fi.Int32Value(ig.Spec.RootVolumeThroughput)))
 		}
 	}
+
+	if ig.Spec.CPUCredits != nil {
+		lt.CPUCredits = ig.Spec.CPUCredits
+	} else {
+		lt.CPUCredits = fi.String("")
+	}
+
 	return lt, nil
 }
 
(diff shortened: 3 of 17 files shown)


pkg/apis/kops/validation/aws.go
```diff
@@ -277,3 +282,24 @@ func awsValidateLoadBalancerSubnets(fieldPath *field.Path, spec kops.ClusterSpec
...
+func awsValidateCPUCredits(fieldPath *field.Path, spec *kops.InstanceGroupSpec) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if !featureflag.EnableLaunchTemplates.Enabled() {
+		allErrs = append(allErrs, field.Forbidden(fieldPath.Child("cpuCredits"), "launch templates must be enabled with `export KOPS_FEATURE_FLAGS=EnableLaunchTemplates`"))
+	}
+
+	allErrs = append(allErrs, IsValidValue(fieldPath.Child("cpuCredits"), spec.CPUCredits, []string{"standard", "unlimited"})...)
+	if spec.MixedInstancesPolicy != nil {
+		for i, instanceType := range spec.MixedInstancesPolicy.Instances {
```
hakman, line 295: Also t4g

pkg/apis/kops/validation/aws.go
```diff
@@ -277,3 +282,24 @@ func awsValidateLoadBalancerSubnets(fieldPath *field.Path, spec kops.ClusterSpec
...
+			if !strings.HasPrefix(instanceType, "t2") && !strings.HasPrefix(instanceType, "t3") {
+				allErrs = append(allErrs, field.Forbidden(fieldPath.Child("mixedInstancesPolicy", "instanceType").Index(i), "cpuCredits requires the t2 or t3 instance families"))
+			}
+		}
+	}
+	if spec.MachineType != "" && !strings.HasPrefix(spec.MachineType, "t2") && !strings.HasPrefix(spec.MachineType, "t3") {
```
hakman, line 301: Also t4g

pkg/apis/kops/validation/aws.go
hakman: Not sure if it applies to those instances or just to the main instance type, or how would be handled by non-burstable instances.

rifelpet: ah you're right. In fact I can create a launch template using one non-burstable instance type and have credit specification set, and I can still create an ASG with no MIP. I guess we could drop the instance type checks here altogether.

hakman: Good idea.

k8s/crds/kops.k8s.io_instancegroups.yaml
```diff
@@ -101,6 +101,10 @@ spec:
                 description: CompressUserData compresses parts of the user data to
                   save space
                 type: boolean
+              cpuCredits:
+                description: CPUCredits is the credit option for CPU Usage on some
```
hakman, line 105: Would it be better?
```suggestion
                description: CPUCredits is the credit option for CPU Usage for burstable
```

pkg/apis/kops/validation/aws.go
```diff
@@ -277,3 +282,14 @@ func awsValidateLoadBalancerSubnets(fieldPath *field.Path, spec kops.ClusterSpec
 
 	return allErrs
 }
+
+func awsValidateCPUCredits(fieldPath *field.Path, spec *kops.InstanceGroupSpec, cloud awsup.AWSCloud) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if !featureflag.EnableLaunchTemplates.Enabled() {
+		allErrs = append(allErrs, field.Forbidden(fieldPath.Child("cpuCredits"), "launch templates must be enabled with `export KOPS_FEATURE_FLAGS=EnableLaunchTemplates`"))
+	}
```
hakman, line 291: Maybe would be best to remove this too, as we are removing LC support in 1.21.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -117,6 +117,11 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 		return nil, fmt.Errorf("error building cloud tags: %v", err)
 	}
 
+	cpuCredits := fi.StringValue(ig.Spec.CPUCredits)
+	if cpuCredits == "" {
+		cpuCredits = "unlimited"
+	}
+
```
hakman, line 124: I think this should be moved at the end of the function and keep similar style.

rifelpet: done, though some of the later assignments in that function are unnecessary. We're checking if a pointer is non-nil and conditionally assigning it to the task's pointer field. We could unconditionally assign it (where my CPUCredits assignment originally was) without any change in behavior. Maybe a cleanup opportunity after the launch configuration removal is merged.

hakman: 👍

hakman (approved):

PR #10935, opened 2021-02-26
Fix node label conversion in Azure

kenji-cloudnatix:
As Azure doesn't allow "/" in label keys, we used to convert "/" to
"_", but we didn't convert back "_" to "/" in kops-controller. This
commit fixes this.

Change the conversion semantics so that "/" is converted to "_s".  "_"
is converted to "__" so that we can distinguish the original "_" and
"_s".

### DIFF
--- a/pkg/nodeidentity/azure/identify.go
+++ b/pkg/nodeidentity/azure/identify.go
@@ -27,7 +27,6 @@ import (
 	expirationcache "k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/nodeidentity"
-	"k8s.io/kops/upup/pkg/fi"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/nodeidentity/azure/identify.go
```diff
@@ -109,8 +109,14 @@ func (i *nodeIdentifier) IdentifyNode(ctx context.Context, node *corev1.Node) (*
 	}
 
 	for k, v := range vmss.Tags {
+		// Obtain the original label key by reverting the change made by
+		// AzureModelContext.CloudTagsForInstanceGroup.
+		// Convert "_s" back to "/", and then convert "__" back to "_".
+		k = strings.ReplaceAll(k, "_s", "/")
```
rifelpet, line 115: I'm not sure this round trip conversion guarantees equivalence.  for example, a label of `foo_s` would get converted to an azure tag of `foo__s` but this conversion back to a kops label would result in `foo_/`. Even reversing the order of these two `ReplaceAll`s would result in `foo/` rather than `foo_s`.

kenji-cloudnatix: You're totally right! I have updated the code so that label keys are stored in tag values.  

Suppose that we have the following node labels:

- key: `foo`, value: `bar`
- key: `baz`, value: `qux`

We create the following tags:

- key: `k8s.io_cluster_node-template_label_v2_0`, value: `foo=bar`
- key: `k8s.io_cluster_node-template_label_v2_1`, value: `baz=qux`

Does this work? I'm not sure we should worry about the backward compatibility now, but kept the compatibility just in case.

hakman: We shouldn't worry about backwards compatibility. Azure is classified as Alpha and no official release was made. Let's keep it simple :).

kenji-cloudnatix: ack! updated the PR.

hakman (approved):

hakman:
Ah... one more thing, would you mind squashing the commits?

kenji-cloudnatix:
Sure! Done!

hakman:
Thanks! :)
/hold cancel
/lgtm

PR #10937, opened 2021-02-26
Remove support for launch configurations

bharath-123:


hakman:
Not sure if there is a need to have multiple PRs for this. Should be reasonably simple to remove support for launch configurations.

bharath-123:
This should be good for a review. Would love some feedback on the integration tests related changes. 

/cc @hakman

hakman:
> This should be good for a review. Would love some feedback on the integration tests related changes.

I think the LT integration test should be removed completely. As everything uses LT now, it should be no longer needed.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -196,11 +183,6 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchConfigurationTask(c *fi.ModelB
 		volumeType = DefaultLegacyVolumeType
 	}
 
-	rootVolumeDeleteOnTermination := DefaultVolumeDeleteOnTermination
-	if ig.Spec.RootVolumeDeleteOnTermination != nil {
-		rootVolumeDeleteOnTermination = fi.BoolValue(ig.Spec.RootVolumeDeleteOnTermination)
-	}
-
```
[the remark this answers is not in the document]
bharath-123: No it should not be. I think I have confused it for this https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go#L52 but thats for network interfaces. 

We would have to add RootVolumeDeleteOnTermination support for launch templates in kOps. we don't do that currently.

bharath-123:
> > This should be good for a review. Would love some feedback on the integration tests related changes.
> 
> I think the LT integration test should be removed completely. As everything uses LT now, it should be no longer needed.

agreed. will do that.

pkg/model/awsmodel/autoscalinggroup.go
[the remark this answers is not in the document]
bharath-123: Actually wait, we mark it as true by default in cloudFormation and terraform like this https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_cloudformation.go#L268 . Should we keep it as true or allow the user to specify it?

cmd/kops/integration_test.go
```diff
@@ -347,18 +341,6 @@ func TestDockerCustom(t *testing.T) {
...
-// TestLaunchConfigurationASG tests ASGs using launch configurations instead of launch templates
-func TestLaunchConfigurationASG(t *testing.T) {
-	featureflag.ParseFlags("-EnableLaunchTemplates")
-	unsetFeatureFlags := func() {
-		featureflag.ParseFlags("+EnableLaunchTemplates")
-	}
-	defer unsetFeatureFlags()
-
-	newIntegrationTest("launchtemplates.example.com", "launch_templates").withZones(3).withLaunchConfiguration().runTestTerraformAWS(t)
-	newIntegrationTest("launchtemplates.example.com", "launch_templates").withZones(3).withLaunchConfiguration().runTestCloudformation(t)
```
rifelpet, line 359: With this being removed, we can remove everything under `tests/integration/update_cluster/launch_templates`

pkg/model/awsmodel/autoscalinggroup.go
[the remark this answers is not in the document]
hakman: Not quite sure what the confusion is about. We allow people to set it via `ig.Spec.RootVolumeDeleteOnTermination`.
If not set by the user, the default `DefaultVolumeDeleteOnTermination ` is used.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -181,7 +168,7 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 }
 
 // buildLaunchConfigurationTask is responsible for building a launch configuration task into the model
```
hakman, line 170:
```suggestion
// buildLaunchTemplateHelper is responsible for building a launch configuration task into the model
```

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -181,7 +168,7 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
...
-func (b *AutoscalingGroupModelBuilder) buildLaunchConfigurationTask(c *fi.ModelBuilderContext, name string, ig *kops.InstanceGroup) (*awstasks.LaunchConfiguration, error) {
+func (b *AutoscalingGroupModelBuilder) buildLaunchTemplate(c *fi.ModelBuilderContext, name string, ig *kops.InstanceGroup) (*awstasks.LaunchTemplate, error) {
```
hakman, line 171:
```suggestion
func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateHelper(c *fi.ModelBuilderContext, name string, ig *kops.InstanceGroup) (*awstasks.LaunchTemplate, error) {
```

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -343,7 +324,7 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchConfigurationTask(c *fi.ModelB
 	// @step: set up instance spot pricing
 	if fi.StringValue(ig.Spec.MaxPrice) != "" {
 		spotPrice := fi.StringValue(ig.Spec.MaxPrice)
-		t.SpotPrice = spotPrice
+		t.SpotPrice = fi.String(spotPrice)
 	}
```
hakman, line 329: This is no longer needed, see the change in `buildLaunchTemplateTask`.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -147,7 +134,7 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 	//   You cannot use a launch template that is set to request Spot Instances (InstanceMarketOptions)
 	//   when you configure an Auto Scaling group with a mixed instances policy.
 	if ig.Spec.MixedInstancesPolicy == nil {
```
hakman, line 136:
```suggestion
	if ig.Spec.MixedInstancesPolicy == nil && ig.Spec.MaxPrice != nil {
```

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -147,7 +134,7 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
...
-		lt.SpotPrice = fi.String(lc.SpotPrice)
+		lt.SpotPrice = lc.SpotPrice
```
hakman, line 137:
```suggestion
		lt.SpotPrice = ig.Spec.MaxPrice
```

pkg/model/awsmodel/autoscalinggroup.go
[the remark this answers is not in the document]
bharath-123: I think where i got confused was at the part where we set it to true in cloudFormation, terraform and launchtemplate api. I assumed there was some reasoning behind it. I ll work on allowing those to be configured by the user.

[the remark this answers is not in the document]
hakman: 👍

hakman:
You were right, `RootVolumeDeleteOnTermination` had to be deprecated and ignored. I just removed that to speed things up as it was my mistake.
Thanks for the effort here. It will make things easier to maintain for AWS tasks.

hakman (approved):

rifelpet:
We should probably have a release note for this.

bharath-123:
@rifelpet I have already added this to the 1.21 release notes :)

rifelpet:
I don't see anything in here mentioning that launch configuration support has been removed: https://github.com/kubernetes/kops/blob/master/docs/releases/1.21-NOTES.md it should probably be in one of the first few sections in the docs rather than just a bullet point for the PR

PR #10945, opened 2021-02-27
Exclude CP nodes from load balancers

olemarkus:
The new service node exclusion feature went GA in 1.21. We should mark CP nodes so they don't become targets for LB services. 
See https://github.com/kubernetes/kubernetes/pull/97543


pkg/nodelabels/builder.go
```diff
@@ -84,5 +84,6 @@ func BuildMandatoryControlPlaneLabels() map[string]string {
 	nodeLabels[RoleLabelControlPlane20] = ""
 	nodeLabels[RoleLabelName15] = RoleMasterLabelValue15
 	nodeLabels["kops.k8s.io/kops-controller-pki"] = ""
+	nodeLabels["node.kubernetes.io/exclude-from-external-load-balancers"] = ""
```
rifelpet, line 87: Should we conditionally add this based on kubernetes minor version?

olemarkus: If people have enabled the new service logic, they need this label as well. If they haven't, this is no-op. So my preference is to do this unconditionally.

rifelpet (approved):

pkg/nodelabels/builder.go
kuzaxak: We have the logic when Masters hosts ingresses, can we somehow disable this behaviour?

PR #10948, opened 2021-02-28
Add AWS Transit Gateway support

### LINKED ISSUE #8837: Enable usage of AWS Transit Gateway

Cluster spec should allow the usage of AWS Transit Gateways, like it's possible to define `egress` to use specific NAT gateways (https://github.com/kubernetes/kops/blob/master/docs/cluster_spec.md#egress)

I tried setting it to the resource `tgw-aedf1230` but validation of cluster spec fails with error `error populating cluster spec: spec.Subnets[0].Egress: Invalid value: "tgw-aedf1230": egress must be of type NAT Gateway or NAT EC2 Instance or 'External'`

Only the route table entry has to be changed for `0.0.0.0/0` to use `tgw-aedf1230` aws resource

rifelpet:
closes #8837

I opted to just use a `*string` for the TGW ID in `awstasks.Route` rather than a new `awstasks.TransitGateway` because kops will never be creating TGWs itself, they will always be defined by ID, therefore we don't need any Find or Render logic.


hakman (approved):
I agree with the logic. 👍

PR #10957, opened 2021-03-01
Add support for enable-cadvisor-json-endpoints with Kubelet

adrianmoisey:
Kubernetes 1.18 disables this by default.
For backwards compatibility with monitoring tools, it would be nice to
be able to re-enable this.

https://github.com/kubernetes/kubernetes/issues/68522

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -517,6 +517,12 @@ func validateKubelet(k *kops.KubeletConfigSpec, c *kops.Cluster, kubeletPath *fi
 			}
 		}
 
+		if k.EnableCadvisorJsonEndpoints != nil {
+			if c.IsKubernetesLT("1.18") && c.IsKubernetesGTE("1.21") {
+				allErrs = append(allErrs, field.Forbidden(kubeletPath.Child("enableCadvisorJsonEndpoints"), "enableCadvisorJsonEndpoints requires Kubernetes 1.18-1.20"))
+			}
+		}
+
 	}
 	return allErrs
 }
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -2318,6 +2318,10 @@ spec:
                     description: DockerDisableSharedPID uses a shared PID namespace
                       for containers in a pod.
                     type: boolean
+                  enableCadvisorJsonEndpoints:
+                    description: EnableCadvisorJsonEndpoints enables cAdvisor json
+                      `/spec` and `/stats/*` endpoints. Defaults to False.
+                    type: boolean
                   enableCustomMetrics:
                     description: Enable gathering custom metrics.
                     type: boolean
@@ -2719,6 +2723,10 @@ spec:
                   dockerDisableSharedPID:
                     description: DockerDisableSharedPID uses a shared PID namespace
                       for containers in a pod.
+                    type: boolean
+                  enableCadvisorJsonEndpoints:
+                    description: EnableCadvisorJsonEndpoints enables cAdvisor json
+                      `/spec` and `/stats/*` endpoints. Defaults to False.
                     type: boolean
                   enableCustomMetrics:
                     description: Enable gathering custom metrics.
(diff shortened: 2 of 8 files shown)


adrianmoisey:
@hakman I can't tell if the failed tests are related to this change. Is there anything I need to do?

olemarkus:
This flag will be removed in 1.21. We need to consider if this is a good time to add support for that flag.
See https://github.com/kubernetes/kubernetes/issues/68522

adrianmoisey:
> This flag will be removed in 1.21. We need to consider if this is a good time to add support for that flag.

This is something I thought about. Our situation is that we're currently upgrading from Kubernetes 1.17 to Kubernetes 1.18. Our monitoring service (datadog) doesn't have a workaround for the disabled `enable-cadvisor-json-endpoints` yes, so we're forced to enable this to get some metrics.

I do realise that it's a bit strange adding a deprecated flag currently. I'm unsure what kOps' policy on this is.

hakman:
Maybe we could add it to 1.20 only?

olemarkus:
> Maybe we could add it to 1.20 only?

If we add this, we should validate the k8s version (i.e from 1.18 up to and including 1.20), but we can put this in any kOps version supported these k8s version.

adrianmoisey:
For my use-case I'd need it in 1.18.

Is there any other way of manipulating the command line arguments that get sent to kubelet?

olemarkus:
I think the approach in the PR is fine. Just need that validation 0

adrianmoisey:
Great! Anything I need to do?

Also, who is responsible for cherry-picking it back into 1.18 and 1.19? Could I do that?

hakman:
You would have to add something similar to:
https://github.com/kubernetes/kops/blob/9d8dec60d6250165d49f68752f0ef2d6810bdbe6/pkg/apis/kops/validation/validation.go#L513-L518

Regarding cherry-picking, we will take care of that at the end, but it will be to 1.20 and possibly 1.19, but for sure not 1.18.

adrianmoisey:
> You would have to add something similar to:

Done this.

I was unsure of the values at the end of the line in https://github.com/kubernetes/kops/pull/10957/files#diff-ae412ac68b83570fd50e9d5d63873f060eb8f503f5e44be78d710076294c3285R521

Is that correct?

hakman:
> Is that correct?

No need for that, just the version check(s).

adrianmoisey:
Is there anything else needed from me?

hakman:
Please also squash all commits.

adrianmoisey:
Done

pkg/apis/kops/validation/validation.go
```diff
@@ -517,6 +517,12 @@ func validateKubelet(k *kops.KubeletConfigSpec, c *kops.Cluster, kubeletPath *fi
 			}
 		}
 
+		if k.EnableCadvisorJsonEndpoints != nil {
+			if c.IsKubernetesGTE("1.18") && c.IsKubernetesLT("1.21") {
+				allErrs = append(allErrs, field.Forbidden(kubeletPath.Child("enableCadvisorJsonEndpoints"), "enableCadvisorJsonEndpoints requires at least Kubernetes 1.18"))
```
hakman, line 522: Sorry, my mistake. Please change and squash one more time.
```suggestion
			if c.IsKubernetesLT("1.18") && c.IsKubernetesGTE("1.21") {
				allErrs = append(allErrs, field.Forbidden(kubeletPath.Child("enableCadvisorJsonEndpoints"), "enableCadvisorJsonEndpoints requires at Kubernetes 1.18-1.20"))
```
```suggestion
			if c.IsKubernetesGTE("1.18") && c.IsKubernetesLT("1.21") {
				allErrs = append(allErrs, field.Forbidden(kubeletPath.Child("enableCadvisorJsonEndpoints"), "enableCadvisorJsonEndpoints requires at least Kubernetes 1.18"))
```

hakman (changes requested):

pkg/apis/kops/validation/validation.go
adrianmoisey: > Sorry, my mistake. Please change and squash one more time.

No worries, I should have caught that.

Fixed

hakman (approved):

hakman:
I think this was it. Thanks @adrianmoisey.

adrianmoisey:
Perfect! Thanks for the help! I really appreciate it.

PR #10963, opened 2021-03-01
[Digital Ocean] Add an e2e job for DO

srikiz:
Testing only.
Trying to set an e2e job for DO. I'll update the title once it's ready.


tests/e2e/e2e.mk
```diff
@@ -38,3 +38,25 @@ test-e2e-aws-simple-1-20: test-e2e-install
...
+.PHONY: test-e2e-do-simple-1-20
+test-e2e-do-simple-1-20: test-e2e-install
+	kubetest2 kops \
+		-v 6 \
+		--up --down \
+		--cloud-provider=digitalocean \
+		--cluster-name e2e-test-do.k8s.local \
+		--create-args networking=cilium \
+		--create-args api-loadbalancer-type=public \
+		--master-count=3 \
```
rifelpet, line 51: kubetest2-kops' --create-args is of type `string` rather than `[]string` so it can only be specified once. its value is space-separate flags though.

```suggestion
		--create-args "--networking=cilium --api-loadbalancer-type=public --master-count=3" \
```

srikiz: Oh I see, got it. Thanks @rifelpet

tests/e2e/e2e.mk
```diff
@@ -38,3 +38,24 @@ test-e2e-aws-simple-1-20: test-e2e-install
...
+		--up --down \
+		--cloud-provider=digitalocean \
+		--cluster-name e2e-test-do.k8s.local \
+    --create-args "--networking=cilium --api-loadbalancer-type=public --master-count=3" \
+		--env S3_ENDPOINT=sfo3.digitaloceanspaces.com \
+		--env KOPS_STATE_STORE=do://e2e-kops-space \
+		--env JOB_NAME=pull-kops-e2e-kubernetes-do-kubetest2 \
+		--kops-version-marker=https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt \
+		--kubernetes-version=https://storage.googleapis.com/kubernetes-release/release/stable-1.20.txt \
+		--template-path=tests/e2e/templates/simple.yaml.tmpl \
```
rifelpet, line 55:
```suggestion
```

lets leave this flag out since it affects the resulting cluster manifest and i'm not sure what might need to change for that template to work with DO. without the flag it will use `kops create cluster --cloud digitalocean` like normal.

Also I think we should set the state store here rather than an `--env` flag:

https://github.com/kubernetes/kops/blob/cb1edd047eaa683f1411a6dd600ee2c4cd7bf6a8/tests/e2e/kubetest2-kops/deployer/common.go#L186-L197

rifelpet:
I think we need to explicitly pass in the DIGITALOCEAN_ACCESS_TOKEN and S3_* environment variables here:

https://github.com/kubernetes/kops/blob/e1a17e94a5870a509e7af585699df623c615ec25/tests/e2e/kubetest2-kops/deployer/common.go#L120-L144

I do see the env vars being set in the job [here](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/10963/pull-kops-e2e-kubernetes-do-kubetest2/1366737504253775872/podinfo.json), so the preset seems to be working.

srikiz:
It looks like ```DIGITALOCEAN_ACCESS_TOKEN``` exists, but the other 2 variables we added ```S3_ACCESS_KEY_ID``` and ```S3_SECRET_ACCESS_KEY``` are missing.

rifelpet:
I'm seeing the env vars being defined [here](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/10963/pull-kops-e2e-kubernetes-do-kubetest2/1366770196542394368/podinfo.json) referencing the `spaces-digitalocean-s3` secret. Can we confirm that secret exists and is populated? We'll likely need help from someone with access to the infra.

srikiz:
Yeah. We initially got it added with ```DO_S3_ACCESS_KEY_ID``` and ```DO-S3_SECRET_ACCESS_KEY``` env vars. Later I changed the presets by removing "DO_", I am wondering if that's causing issues. I can also probably create a PR to infra, to make it with DO_... again, and verify if those env vars show up.

srikiz:
Thanks @fejta - Per discussion, it looks like the secrets does exist. So may be I am missing something, will continue to investigate.
https://kubernetes.slack.com/archives/C7J9RP96G/p1614712856037400?thread_ts=1614711416.033500&cid=C7J9RP96G

FYI - @timoreimann

rifelpet:
Oh 🤦🏻  I wonder if it is Makefile that is unsetting the env var:

https://github.com/kubernetes/kops/blob/e1a17e94a5870a509e7af585699df623c615ec25/Makefile#L51

the reason we unset it there was to keep it from impacting tests. Maybe we try temporarily removing it in this PR and then we figure out a longer term solution before merging?

rifelpet:
For the ssh key, we'll need to update this code to reference the new env vars introduced by the preset:

https://github.com/kubernetes/test-infra/blob/22b941d3af0b43af5447c1757ae1c9fc7510de6b/config/jobs/kubernetes/kops/kops-presets-do.yaml#L19-L20


https://github.com/kubernetes/kops/blob/8a8a4c8e40caa7af04c67c1044c374083034130d/tests/e2e/kubetest2-kops/deployer/common.go#L66-L71

We should add the private key to the prow cluster's secret as well, we use it with AWS to SSH into the instances to get cluster logs and other useful troubleshooting information.

rifelpet:
It seems the cluster validation is failing - apiserver isn't responding properly likely because of an issue with one of its dependencies or something earlier in the bootstrap process. It would help a lot if we had the SSH private key so that kops would SSH into the nodes and dump the logs files into the prow job's artifacts so we can view them. Would it be possible to get the SSH private key setup in the cluster secret along side the public key? This would match how we have it setup with AWS.

rifelpet:
Hm I'm not sure why its failing to SSH and dump logs from the instances. Can you confirm that `kops toolbox dump` works on your own DO clusters?

Also I'm curious if you can create your own cluster with these cluster and instancegroup manifests: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kops/10963/pull-kops-e2e-kubernetes-do-kubetest2/1372279591074271232/artifacts/

srikiz:
@rifelpet - Realized DO hasn't implemented the Dumpable interface yet. That's probably why we are not seeing any logs yet.
I tested locally and it looks like volume is not getting bound to the droplets. I am looking into it.

tests/e2e/e2e.mk
```diff
@@ -37,3 +37,22 @@ test-e2e-aws-simple-1-20: test-e2e-install
...
+test-e2e-do-simple-1-20: test-e2e-install
+	kubetest2 kops \
+		-v 6 \
+		--up --down \
+		--cloud-provider=digitalocean \
+		--cluster-name e2e-test-do.k8s.local \
+		--env S3_ENDPOINT=sfo3.digitaloceanspaces.com \
+		--env JOB_NAME=pull-kops-e2e-kubernetes-do-kubetest2 \
+		--create-args "--networking=cilium --api-loadbalancer-type=public --master-count=3" \
+		--kops-version-marker=https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt \
```
srikiz, line 51: @rifelpet - is there a way I can use an older kops binary here (say kops binary released for v1.18 instead ?)

rifelpet: give this a shot:

https://storage.googleapis.com/k8s-staging-kops/kops/releases/markers/release-1.18/latest-ci.txt

rifelpet:
I pushed a commit to this branch: https://github.com/rifelpet/kops/commit/eca86c6a51915f6d6d4962a62180c1947b183728 can you cherry-pick it onto this PR's branch? and feel free to iterate on it as well, I left a few comments on the GH commit itself. My intent is to add dump support for DO droplets so that we can get logs in the e2e job more easily.

It's a rough copy of the aws function:

https://github.com/kubernetes/kops/blob/9db39aa7d155a468d47bfe185ee83aa9542e32fa/pkg/resources/aws/aws.go#L540-L579

rifelpet:
Looking at the [testgrid history](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#e2e-do-kubetest2) of the past 4 runs it seems we have a handful of flakes and then 2 tests that are consistently failing:

* Kubernetes e2e suite.[sig-network] KubeProxy should set TCP CLOSE_WAIT timeout [Privileged]
* Kubernetes e2e suite.[sig-network] Services should be able to update service type to NodePort listening on same port number but different protocols

The first one always fails with --networking=cilium, we actually [skip it](https://github.com/kubernetes/test-infra/blob/c3460250be15b780dba7513506fc4b1fb706ed3f/config/jobs/kubernetes/kops/build_jobs.py#L286-L288) in our other jobs so we should probably do so here as well. I'm not sure we've seen failures of the second test before, so it might indicate an actual bug.

olemarkus:
The second one flakes on cilium. Supposedly to be fixed in their next release. We probably should be skipping that one too until they have closed that issue.

Makefile
```diff
@@ -46,10 +46,10 @@ GCFLAGS?=
 UPLOAD_CMD=$(KOPS_ROOT)/hack/upload ${UPLOAD_ARGS}
 
 # Unexport environment variables that can affect tests and are not used in builds
-unexport AWS_ACCESS_KEY_ID AWS_REGION AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN CNI_VERSION_URL DNS_IGNORE_NS_CHECK DNSCONTROLLER_IMAGE DO_ACCESS_TOKEN GOOGLE_APPLICATION_CREDENTIALS
-unexport KOPS_BASE_URL KOPS_CLUSTER_NAME KOPS_RUN_OBSOLETE_VERSION KOPS_STATE_STORE KOPS_STATE_S3_ACL KUBE_API_VERSIONS NODEUP_URL OPENSTACK_CREDENTIAL_FILE SKIP_PACKAGE_UPDATE
-unexport SKIP_REGION_CHECK S3_ACCESS_KEY_ID S3_ENDPOINT S3_REGION S3_SECRET_ACCESS_KEY
-
+unexport AWS_ACCESS_KEY_ID AWS_REGION AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN CNI_VERSION_URL DNS_IGNORE_NS_CHECK DNSCONTROLLER_IMAGE GOOGLE_APPLICATION_CREDENTIALS
```
srikiz, line 49: @rifelpet - thanks a lot for all the help. The tests are now passing.
Wanted your thoughts on this change - are we okay to remove the unexport option on some of these environment variables?

rifelpet: I'm debating whether we should keep the `kubetest2 kops` invocations in e2e.mk at all. I originally added it there so I could iterate on it as a presubmit quickly, but the real presubmit and periodic jobs dont rely on the kubetest2 command in the makefile at all, they invoke `kubetest2 kops` directly in the prow job yaml. I think the DO periodic and/or presubmit jobs should do the same.

I do think theres value in having the DO kubetest2 command already setup somewhere, perhaps we can replace the command in e2e.mk with a shell script like this [upgrade test scenario](https://github.com/kubernetes/kops/blob/master/tests/e2e/scenarios/upgrade/run-test)? Create a new `tests/e2e/scenarios/digital-ocean/run-test` script and revert the changes in Makefile and e2e.mk.

If we do that then the next step would be to add the new periodic job and update the presubmit job to invoke kubetest2 directly.

srikiz: This sounds good, thanks for the direction @rifelpet .

srikiz: @rifelpet - Sorry for the delay - I have made the changes. Please have a look when you get a chance. Another question I had was invoking these scripts? So the `tests/e2e/scenarios/digital-ocean/run-test` is only for reference and not invoked from the real presubmit job correct?

rifelpet: I agree that we could use this script as reference but have the presubmit job invoke similar kubetest2 commands directly. We'll need to update the presubmit job in test-infra to reflect these changes.

One more thing is that we just merged https://github.com/kubernetes/kops/pull/10847 which adds a --provider flag when calling the e2e.test binary. From searching the kubernetes/kubernetes/test/e2e directory, I don't see any digitalocean code including any [DO provider registered with the e2e tests](https://github.com/kubernetes/kubernetes/tree/master/test/e2e/framework/providers). I think we'll run into [this "unknown provider"](https://github.com/kubernetes/kubernetes/blob/3d5ebcc37f8eef9a3619a094ed12bffb1775eae1/test/e2e/framework/test_context.go#L466-L492) error for DO, so we might need to update [this function](https://github.com/kubernetes/kops/blob/f1fbde95e69c9fda23d8074458eab1a03050eca5/tests/e2e/pkg/tester/tester.go#L157-L162) to not set --provider for DO clusters. We may need to do the same with the other flags in the nearby functions (--gce-zone, --gce-multizone, --gce-region, all defined in e2e.test [here](https://github.com/kubernetes/kubernetes/blob/3d5ebcc37f8eef9a3619a094ed12bffb1775eae1/test/e2e/framework/test_context.go#L344-L347))

rifelpet: and feel free to open a test-infra PR now to update the job with these updated kubetest2 commands. We can merge it and test the job against this PR and then proceed with merging this.

rifelpet:
@srikiz okay the job seems to be configured correctly now. The current failure is due to some logic we added in kubetets2-kops that sets some cloud-specific flags to the e2e.test binary.

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/10963/pull-kops-e2e-kubernetes-do-kubetest2/1388903263197204480#1:build-log.txt%3A1072

since e2e.test doesn't have any DO-specific logic, we can just not pass these flags. I think it would be best to just have this logic default to not doing anything rather than returning an error. Feel free to update these in this PR:

https://github.com/kubernetes/kops/blob/fdce6638ddf1a44d29c1e09d97436b0fa3a8cd5e/tests/e2e/pkg/tester/tester.go#L157-L162

https://github.com/kubernetes/kops/blob/fdce6638ddf1a44d29c1e09d97436b0fa3a8cd5e/tests/e2e/pkg/tester/tester.go#L219-L227

https://github.com/kubernetes/kops/blob/fdce6638ddf1a44d29c1e09d97436b0fa3a8cd5e/tests/e2e/pkg/tester/tester.go#L297-L299

tests/e2e/pkg/tester/tester.go
```diff
@@ -155,10 +155,10 @@ func (t *Tester) addProviderFlag() error {
 
 	provider := ""
 	switch cluster.Spec.CloudProvider {
-	case "aws", "gce":
+	case "aws", "gce", "digitalocean":
 		provider = cluster.Spec.CloudProvider
```
rifelpet, line 159:
```suggestion
	case "aws", "gce":
		provider = cluster.Spec.CloudProvider
	case "digitalocean": 
```

srikiz:
@rifelpet - the aws & digital ocean jobs failed when it ran yesterday. It looks like the allocated time wasn't sufficient. I'll go ahead and increase the time duration if I see that happening again.

rifelpet (commented):
This LGTM. If you can squash all the commits then we can get this merged and set up a periodic job as well.

srikiz:
@rifelpet - this is good for your another review. Thanks a ton for all your help in getting to this point :)

PR #10978, opened 2021-03-04
Removing duplicate local and output values in terraform(#10786)

mmerrill3:
Signed-off-by: mmerrill3 <michael.merrill@vonage.com>
- fix for #10786 removing duplicated values in local and output variables for terraform


rifelpet (approved):
thanks for the fix!

Issue #10539, opened 2021-01-05
Impossible to deactivate PrometheusGoMetrics for Calico Networking


henninge:
We are using kops 1.18 with a 1.18 cluster on AWS but that information is mostly irrelevant to this report. I have identified the cause of the bug in the current master branch (0ca0e38518) but am unsure how to fix it.

Consider the following networking configuration.

      networking:
        calico:
          majorVersion: v3
          prometheusMetricsEnabled: true
          prometheusGoMetricsEnabled: false

After deployment the DeamonSet for calico-node has this in its pod template (redacted for readabilty):

    spec:
    template:
        spec:
        containers:
        - env:
            - name: FELIX_PROMETHEUSMETRICSENABLED
              value: "true"
            - name: FELIX_PROMETHEUSGOMETRICSENABLED
              value: "true"

What should happen instead, is that `FELIX_PROMETHEUSGOMETRICSENABLED` be set to "false". The same issue applies to `prometheusProcessMetricsEnabled` and its corresponding `FELIX_PROMETHEUSPROCESSMETRICSENABLED` as well as the same values in the "canal" configs.

The source of the problem is that in Calico these settings default to "true" but Go defaults Boolean values to "false". This becomes a problem in line 3945 of  `upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template` and other places.

        - name: FELIX_PROMETHEUSGOMETRICSENABLED
          value: "{{- or .Networking.Calico.PrometheusGoMetricsEnabled "true" }}"

The "or" function  always evulates to "true" because `.Networking.Calico.PrometheusGoMetricsEnabled` is a boolean and "or" will return the first non-empty (that includes non-false) value of its parameters.

I was going to directly submit a PR but the solution is not as straight forward as I would have expected. There is no way to pass the "use default value" information from the yaml because the Marshaller will initizalize the value to its Go-Builtin default "false" if it's missing. Later on it is impossible to determine if the "false" was set by the user or not.

I came up with three possible solutions:

  - Make the value a string and simply pass that through. Works but would be insconsistent with other Booleans in the config.
  - Let the default be "false" which would be different from how Calico specifies it.
  - Inverse the name of the config value and let the default be "false", i.e. `PrometheusGoMetricsDisabled`.

As I said, I am happy to implement either of them and create the PR. But which would be the preferred fix?

hakman:
Nice find @henninge. Thanks for reporting it.

[...]

PR #10982, opened 2021-03-05
Disable Calico Prometheus metrics by default

hakman:
Fixes the incorrect defaults implementation but also defaults to the official Calico values.

Fixes: #10539


olemarkus:
Should we have a release note for this one? Can perhaps break a lot of monitoring.

hakman:
The important stuff was already disabled, only the Go and Process metrics were enabled by default. Not sure if it really deserves a release note.

olemarkus (approved):

PR #10987, opened 2021-03-06
Remove manually added labels from addons

hakman:
This removes `role.kubernetes.io/networking: "1"` and `role.kubernetes.io/authentication: "1"` labels which are now added automatically since https://github.com/kubernetes/kops/pull/10796.

/cc @olemarkus

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
@@ -25,7 +25,6 @@ metadata:
   name: weave-net
   labels:
     name: weave-net
-    role.kubernetes.io/networking: "1"
   namespace: kube-system
 rules:
   - apiGroups:
@@ -68,7 +67,6 @@ metadata:
   name: weave-net
   labels:
     name: weave-net
-    role.kubernetes.io/networking: "1"
   namespace: kube-system
 roleRef:
   kind: ClusterRole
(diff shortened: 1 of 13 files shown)


olemarkus (approved):
Thanks for taking the time to do this one.

hakman:
Was a pleasure. Now I don't have to see them every time I update Calico.
😁

upup/models/cloudup/resources/addons/networking.weave/k8s-1.12.yaml.template
```diff
@@ -125,20 +123,17 @@ metadata:
...
   labels:
     name: weave-net
-    role.kubernetes.io/networking: "1"
   namespace: kube-system
 spec:
   # Wait 5 seconds to let pod connect before rolling next pod
   selector:
     matchLabels:
       name: weave-net
-      role.kubernetes.io/networking: "1"
```
gautam2187, line 135: @hakman  this breaks the weave addon update as selector labels are immutable, logs from protokube 

```
The DaemonSet "weave-net" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"name":"weave-net"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable

error updating "networking.weave": error applying update from "s3://kops...": error running kubectl: exit status 1
```

olemarkus: Support for weave has been removed now due to the inactivity of upstream. So I while we did an unintentional breaking change here, I think the path forward is to migrate to another CNI.

gautam2187: Thanks for the reply, migration from weave is planned but this added an extra twist.

PR #10988, opened 2021-03-06
Fix kops-controller rbac due to leader election change

olemarkus:
This one was merged to 1.20 branch, but not master. But since the controller-runtime also was updated now, we need this for a working master branch.

/cc @hakman 
/approve


hakman:
I didn't notice that the version changed in master also. Thanks.
/lgtm

PR #11002, opened 2021-03-09
Storage: Amend default choice for StorageClass management to honor a specified OpenStack-related value

seh:
When choosing a default value for the _Cluster_ spec's "cloudConfig.manageStorageClasses" field, first check whether a user specified a concrete value for the related OpenStack "blockStorage.createStorageClass" field. If so, use that value as the effective default value for the former field as well, so as to avoid an unnecessary conflict between the field values on the second validation pass.

See #10733, @zetaab's [findings there](https://github.com/kubernetes/kops/pull/10733/files#r590046073), and [the ensuing discussion in the "kops-dev" channel](https://kubernetes.slack.com/archives/C8MKE2G5P/p1615296250086900?thread_ts=1615291312.086800&cid=C8MKE2G5P) in the "Kubernetes" Slack workspace for context.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/kops/11002)

### DIFF
--- a/pkg/apis/kops/validation/validation_test.go
+++ b/pkg/apis/kops/validation/validation_test.go
@@ -1066,7 +1066,7 @@ func Test_Validate_CloudConfiguration(t *testing.T) {
 				}},
 		},
 		{
-			Description: "os false",
+			Description: "os true",
 			Input: kops.CloudConfiguration{
 				Openstack: &kops.OpenstackConfiguration{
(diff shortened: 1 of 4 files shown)


zetaab (approved):

zetaab:
this needs to be backported to release-1.20 also

seh:
> this needs to be backported to release-1.20 also

See #11003.

PR #11009, opened 2021-03-10
Spotinst: Add support for block device mappings in Ocean Launch Spec

liranp:
This PR adds support for block device mappings in Ocean Launch Spec, which would allow users to configure additional volume options, such as `rootVolumeType`, `rootVolumeIops`, and `rootVolumeThroughput`.


estahn:
@olemarkus Can we cherry-pick this into 1.19?

hakman (approved):

PR #11010, opened 2021-03-10
Remove extraneous field from integration test

rifelpet:
The api design for using existing instance profiles must have changed during its PR and I never removed the old field from the integration test.
grep shows that this field doesn't exist anywhere else in the codebase.


hakman (approved):

jcmartins:
But what's happen with this feature?
How can I change the kops default roles ?

jcmartins:
Need I change the kops default roles to use my role name pattern ( BURoleForKopsMaster and BURoleForKopsNode )
How can I do it?

PR #11015, opened 2021-03-11
azure: fix null pointer when updating in place cluster

collin-woodruff-t1cg:
Null pointer exception occurs when trying to update cluster in place on azure. This occurs because other virtual machine scale sets aside from the master do not have backend loadbalancer pools. 

This checks to make sure that the name of the scale set contains "masters" to differentiate between the master and other virtual machine scale sets.

### DIFF
--- a/upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
+++ b/upup/pkg/fi/cloudup/azuretasks/vmscaleset.go
@@ -182,7 +182,7 @@ func (s *VMScaleSet) Find(c *fi.Context) (*VMScaleSet, error) {
 	}
 
 	var loadBalancerID *loadBalancerID
-	if *ipConfig.LoadBalancerBackendAddressPools != nil {
+	if ipConfig.LoadBalancerBackendAddressPools != nil {
 		for _, i := range *ipConfig.LoadBalancerBackendAddressPools {
 			if !strings.Contains(*i.ID, "api") {
 				continue


hakman:
@kenji-cloudnatix maybe you have some thoughts on this.
/ok-to-test

kenji-cloudnatix:
Thanks for making the change, @collin-woodruff-t1cg ! Could you explain where the null pointer access is happening?

Is this coming from `*i.ID`? I'm wondering if we can add direct null-check instead of implicitly relying on the assumption that the VMSS name convention (and also on the assumption that the master is the only one that is backed by a load balancer).

collin-woodruff-t1cg:
yeah the null pointer is happening on line 185 with `*ipConfig.LoadBalancerBackendAddressPools`. On the other scale sets that field is null so I was getting this exception when trying to update a cluster already in place.

collin-woodruff-t1cg:
although thinking about it now if we don't use a loadbalancer this issue will still be prevelant on the master at that point...

kenji-cloudnatix:
Ah I see. Thanks! So I believe the right fix is to remove `*` from line 185?

```go
if ipConfig.LoadBalancerBackendAddressPools != nil {
```

The type of `LoadBalancerBackendAddressPools` is `*[]SubResource`.

collin-woodruff-t1cg:
I gotcha. i just tested that on my end and i no longer receive the null pointer error. Thanks @kenji-cloudnatix!

kenji-cloudnatix (approved):
Looks good to me! Thanks!

rifelpet (approved):

PR #11016, opened 2021-03-11
user-configurable IAM roles for ServiceAccounts

olemarkus:
This PR allows users to configure IRSA for their own workloads. It is the kOps equivalent of https://eksctl.io/usage/iamserviceaccounts/#usage-with-config-files

/cc @justinsb @rifelpet


pkg/apis/kops/validation/legacy.go
```diff
@@ -400,6 +400,34 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+			allErrs = append(allErrs, field.Duplicate(p, key))
+		}
+		sas[key] = ""
+
+		if sa.IAMPolicyARN == "" && sa.InlinePolicy == "" {
+			allErrs = append(allErrs, field.Required(p, "inlinePolicy or iamPolicyARN must be set"))
+		}
+		if sa.IAMPolicyARN != "" && sa.InlinePolicy != "" {
+			allErrs = append(allErrs, field.Forbidden(p, "cannot set both inlinePolicy and iamPolicyARN"))
+		}
```
rifelpet, line 429: nit, might be worth validating the policy arn:

https://github.com/kubernetes/kops/blob/b03b50fdb14d5c44dec0144c17e2187425ffb1cd/pkg/apis/kops/validation/instancegroup.go#L267-L271

pkg/model/iam.go
```diff
@@ -103,26 +104,63 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
 
+	irsa := b.Cluster.Spec.IAMRolesForServiceAccounts
+	if irsa != nil {
+		for _, sa := range irsa.ServiceAccounts {
+			var p *iam.Policy
+			if sa.InlinePolicy != "" {
+				bp, err := b.buildPolicy(sa.InlinePolicy)
+				p = bp
+				if err != nil {
+					return fmt.Errorf("error inline policy: %w", err)
```
rifelpet, line 115:
```suggestion
					return fmt.Errorf("error parsing inline policy: %w", err)
```

pkg/model/iam.go
```diff
@@ -103,26 +104,63 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+				Policy: p,
+			}
+			iamRole, err := b.BuildServiceAccountRoleTasks(serviceAccount, c)
+			if err != nil {
+				return fmt.Errorf("error building service account role tasks: %w", err)
+			}
+			if sa.IAMPolicyARN != "" {
+				name := "external-" + fi.StringValue(iamRole.Name)
+				externalPolicies := []string{sa.IAMPolicyARN}
+				c.AddTask(&awstasks.IAMRolePolicy{
```
rifelpet, line 132: Would we ever want to support attaching multiple policy ARNs to a single role? if so we should make that api field a list and include a reference to the policy in the `name` here.

rifelpet (commented):
I don't see the OIDC provider being created in the integration test's terraform output. any idea why that is?

olemarkus:
`PublicJWKS` feature flag isn't set on this particular test, so no OIDC config is created.

pkg/model/iam.go
olemarkus: That's a good point. I don't see a reason why we shouldn't allow that.

rifelpet:
The iam roles' assume role policies reference `arn:aws:iam::123456789012:oidc-provider/api.internal.minimal.example.com` if that's not being created in kubernetes.tf not specified anywhere in the cluster spec then where did it come from? Is this even a valid configuration?

olemarkus:
Not really. Those two flags would almost always be used together.

It depends a bit what we want here. My idea was to add an `enable` flag on this struct, and that would mean provision OIDC (i.e `PublicJWKS`). Having this enabled, would also use IRSA on all kops addons that supports it (i.e `UseServiceAccountIAM`). 

Question is if we want to force that? Or do we want users to enable IRSA, but not for the kops addons? Do we want users to enable IRSA individually on all addons?

rifelpet:
Hm for migration purposes I could see the benefit of being able to migrate individual addons to IRSA. It'd be safer to migrate one at a time rather than be forced to do them all at once. I do like the idea of an `enable` field here that would supplement `PublicJWKS`. Perhaps we update the integration test to use both feature flags, and (potentially in a followup PR) add api validation to ensure PublicJWKS is enabled if UseServiceAccountIAM is enabled? And if we want to relax that restriction down the road (eg. to allow specifying an existing OIDC provider ARN in the cluster spec) then we can do that.

pkg/apis/kops/cluster.go
```diff
@@ -205,6 +205,21 @@ type ClusterSpec struct {
...
 	// ClusterAutoscaler defines the cluster autoscaler configuration.
 	ClusterAutoscaler *ClusterAutoscalerConfig `json:"clusterAutoscaler,omitempty"`
+
+	// IAMRolesForServiceAccount defines the IRSA configuration.
+	IAMRolesForServiceAccounts *IAMRolesForServiceAccountsConfig `json:"iamRolesForServiceAccounts,omitempty"`
+}
+
+// InstanceRoleForServiceAccountConfig defines the IRSA configuration.
+type IAMRolesForServiceAccountsConfig struct {
+	ServiceAccounts []ServiceAccountMapping `json:"serviceAccounts,omitempty"`
```
johngmyers, line 215: I think this should go into `IAMSpec` instead of its own struct

pkg/apis/kops/validation/legacy.go
```diff
@@ -400,6 +400,34 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 		}
 	}
 
+	if c.Spec.IAMRolesForServiceAccounts != nil {
```
johngmyers, line 403: Could this go into `validateClusterSpec()` instead of legacy.go? Let's try not to add to tech debt.

pkg/apis/kops/validation/legacy.go
```diff
@@ -400,6 +400,34 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+
+func validateIRSA(irsa *kops.IAMRolesForServiceAccountsConfig, spec *field.Path) (allErrs field.ErrorList) {
+	if !featureflag.PublicJWKS.Enabled() {
+		allErrs = append(allErrs, field.Forbidden(spec, "IAM roles for ServiceAccounts requires PublicJWKS feature flag to be enabled"))
+	}
+
+	sas := make(map[string]string)
+	for _, sa := range irsa.ServiceAccounts {
+		p := spec.Child("serviceAccounts")
+		key := fmt.Sprintf("%s/%s", sa.Namespace, sa.Name)
```
johngmyers, line 417: Validate `Namespace` and `Name` are non-empty? Have `Namespace` default to `default`?

pkg/apis/kops/validation/legacy.go
```diff
@@ -400,6 +400,34 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+		_, duplicate := sas[key]
+		if duplicate {
+			allErrs = append(allErrs, field.Duplicate(p, key))
+		}
+		sas[key] = ""
+
+		if len(sa.IAMPolicyARNs) == 0 && sa.InlinePolicy == "" {
+			allErrs = append(allErrs, field.Required(p, "inlinePolicy or iamPolicyARN must be set"))
```
johngmyers, line 425:
```suggestion
			allErrs = append(allErrs, field.Required(p, "either inlinePolicy or iamPolicyARN must be set"))
```

pkg/model/iam.go
```diff
@@ -103,26 +104,63 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+					Name:      sa.Name,
+					Namespace: sa.Namespace,
+				},
+				Policy: p,
+			}
+			iamRole, err := b.BuildServiceAccountRoleTasks(serviceAccount, c)
+			if err != nil {
+				return fmt.Errorf("error building service account role tasks: %w", err)
+			}
+			if len(sa.IAMPolicyARNs) == 0 {
```
johngmyers, line 129:
```suggestion
			if len(sa.IAMPolicyARNs) > 0 {
```
?

pkg/apis/kops/cluster.go
```diff
@@ -205,6 +205,21 @@ type ClusterSpec struct {
...
+}
+
+type ServiceAccountMapping struct {
+	Name          string   `json:"name"`
+	Namespace     string   `json:"namespace"`
+	IAMPolicyARNs []string `json:"iamPolicyARN,omitempty"`
```
johngmyers, line 221: Should the json name be `iamPolicyARNs`?

johngmyers (changes requested):

pkg/apis/kops/cluster.go
olemarkus: We went a bit back and fort on this. I am not against moving it. But the intention is to merge this with the fields from #11194 so has to consider then if it should be still a substruct or not.

johngmyers: I commented in #11194 that the `OIDCProviderLocation` field is not specific to IRSA. It should not be merged into any IAM-specific substruct.

justinsb:
So I finally got around to this - sorry!

I agree with @johngmyers on the points he raised.  The big one for me is figuring out what the name of the field should be.  Nesting it under iam does feel attractive to me  (e.g. `iam`.`serviceAccountMappings`)

I do think we will end up separating this out of the main cluster object, into a separate top-level object, as that's then easier for addons to provide.  From our discussions though in office hours, we don't want to block on this.

/lgtm though we should reach consensus on the schema-visible points that john raised at least.

olemarkus:
> I do think we will end up separating this out of the main cluster object, into a separate top-level object, as that's then easier for addons to provide. From our discussions though in office hours, we don't want to block on this.
> 

Is it realistic to get a top level object for this in 1.21? 
There are some aspects of this that are attractive, especially if channels deploy these into the cluster as well. I belive I saw that happening with coredns. 

Anyways, I'll move this

justinsb:
That's a great question @olemarkus ... I could do an implementation of a top-level object and we can see what it looks like.  I think we actually have the mechanics for additional top-level objects already in place, but we won't really know until we try!

justinsb:
I did a prototype in #11324 - see in particular the 3rd commit and the diff to `tests/integration/update_cluster/irsa/in-v1alpha2.yaml`.  I started with a kubebuilder framework so that we could create a controller version of this in future, but I've commented out the controller main (and removed most of the boilerplate).

This doesn't isolate us from having to link the code, but it does allow us to specify the API / objects separately.

Personally, I quite like it, but feedback is welcome!

justinsb:
(Also, the logic is all your work @olemarkus  ... this really is just shunting the API into a separate object!)

olemarkus:
@justinsb how do you feel about the naming in this PR now?

johngmyers (commented):
The "mappings" part of the name seems a little odd to me. This is defining the IAM policies for service accounts. If we don't go with just `serviceAccounts` per the EKS example, perhaps `serviceAccountPolicies`? (The EKS example differs in that it also provisions and owns the Kubernetes `ServiceAccount` resource.)

pkg/featureflag/featureflag.go
```diff
@@ -91,9 +91,9 @@ var (
 	LegacyIAM = New("LegacyIAM", Bool(false))
 	// ClusterAddons activates experimental cluster-addons support
 	ClusterAddons = New("ClusterAddons", Bool(false))
-	// UseServiceAccountIAM controls whether we use pod-level IAM permissions for our system pods.
+	// UseServiceAccountIAM controls whether we use pod-level IAM permissions for our system pods and kOps addons.
 	UseServiceAccountIAM = New("UseServiceAccountIAM", Bool(false))
-	// PublicJWKS enables public jwks access.  This is generally not as secure as republishing.
+	// PublicJWKS enables public jwks access.
```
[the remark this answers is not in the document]
olemarkus: We don't need anonymousAuth enabled if discovery docs are published to public store though.

pkg/apis/kops/validation/validation.go
```diff
@@ -246,6 +246,44 @@ func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *fie
...
+	return allErrs
+}
+
+func validateSAMappings(mappings []kops.ServiceAccountMapping, path *field.Path) (allErrs field.ErrorList) {
+	if len(mappings) == 0 {
+		return allErrs
+	}
+	if !featureflag.PublicJWKS.Enabled() {
+		allErrs = append(allErrs, field.Forbidden(path, "IAM roles for ServiceAccounts requires PublicJWKS feature flag to be enabled"))
+	}
```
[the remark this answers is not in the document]
olemarkus: Right now it won't really work without this flag, as it determines of the oidc trust with AWS is configured.

But when the remaining work has been merged, all checks against feature flags has been replaced with configuration flags that are more specific. E.g that the OIDC issuer discovery location is not None

olemarkus:
It's a good suggestion as well. Hope we can conclude the next office hours

justinsb:
Yes - looking forward to discussing it today!  I think the field names are fine (FYI, on GCP there is a concept called a "ServiceAccount" which is a system/robot account, which is an unfortunate naming collision).

I think the decision we should try to make is whether we want to put this into the cluster object, vs putting it into additional objects.  It's a tricky one, because it feels like net new and additional functionality, so I don't think our usual expediency / proximity arguments apply as readily.

olemarkus:
> I think the decision we should try to make is whether we want to put this into the cluster object, vs putting it into additional objects. It's a tricky one, because it feels like net new and additional functionality, so I don't think our usual expediency / proximity arguments apply as readily.

I think additinal objects are a bit immature. It should be trivial to merge mapping/policies defined in the cluster spec and in the new additinal objects, so supporting both long-term sounds good. That shouldn't be more than kOps generating CRs and owning them/keeping up to date. We probably want the same pattern for other parts of the cluster spec we may split out as part of new API versions.

rifelpet:
Does Cluster API or any of the relevant Cluster API providers support similar functionality? If so we could consider modeling our design off of theirs to reduce friction when adopting Cluster API support.

rifelpet:
A summary of the discussion from office hours:

* Move `inlinePolicy` and `iamPolicyARNs` to an `aws` substruct
* Rename serviceAccountMappings to something indicating these are mappings to external permissions (cloud providers, etc.) `serviceAccountExternalMappings` was proposed.
* We need to investigate if multiple fields are supported as a merge key for patch operations since an element in the serviceAccountMappings list would be keyed by name + namespace

As an aside, I think if we move `iamPolicyARNs` to an `aws` substruct we could probably drop the `iam` prefix from the field name, to be more consistent with `inlinePolicy` and because the presence of the `aws` field provides sufficient context that these policy ARNs are for AWS IAM policies.

olemarkus:
Amended according to our discussions today.

Note that right now, one can configure this without enabling AWS OIDC provider. Once that is configurable, we may want to add validation for that. Easy thing that can be done in a follow-up.

johngmyers:
Why use the word "mappings" for these? That word appears to be overly uninformative as to what is being provisioned. What is being provisioned is something more like "permissions" or "roles". They're the out-of-cluster analogue of `Role` and/or `RoleBinding`.

johngmyers:
Okay, at least that word isn't exposed through the API.

pkg/apis/kops/cluster.go
```diff
@@ -207,6 +207,22 @@ type ClusterSpec struct {
...
+type IAMRolesForServiceAccountsConfig struct {
+	ServiceAccounts []ServiceAccountMapping `json:"serviceAccounts,omitempty"`
+}
+
+// ServiceAccountMapping defines the relationship between a Kubernetes ServiceAccount and an IAM Role.
+type ServiceAccountMapping struct {
+	Name          string   `json:"name"`
+	Namespace     string   `json:"namespace"`
+	IAMPolicyARNs []string `json:"iamPolicyARNs,omitempty"`
+	InlinePolicy  string   `json:"inlinePolicy,omitempty"`
```
johngmyers, line 225: I thought the consensus was to push these down into an "aws" substruct

pkg/apis/kops/cluster.go
```diff
@@ -207,6 +207,22 @@ type ClusterSpec struct {
 	ClusterAutoscaler *ClusterAutoscalerConfig `json:"clusterAutoscaler,omitempty"`
 	// WarmPool defines the default warm pool settings for instance groups (AWS only).
 	WarmPool *WarmPoolSpec `json:"warmPool,omitempty"`
+
+	// IAMRolesForServiceAccount defines the IRSA configuration.
```
johngmyers, line 211:
```suggestion
	// IAMRolesForServiceAccount defines external permission roles for service accounts.
```

The consensus was to name this to indicate it is provisioning external roles. I question the need to create a new substruct containing a single field. I think we should either create the array at top level or put it in `iam` with a clearer name. So something like `serviceAccountIAMRoles` or `iam.serviceAccountRoles`.

pkg/apis/kops/validation/validation_test.go
```diff
@@ -1177,3 +1178,89 @@ func Test_Validate_CloudConfiguration(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateIRSA(t *testing.T) {
+
+	featureflag.ParseFlags("+PublicJWKS")
+	unsetFeatureFlags := func() {
+		featureflag.ParseFlags("-PublicJWKS")
+	}
```
johngmyers, line 1187: Would be good to not require the feature flag. Not blocking.

johngmyers (changes requested):

pkg/apis/kops/cluster.go
olemarkus: Looks like I had a bad merge here. They should have been prefixed.

pkg/featureflag/featureflag.go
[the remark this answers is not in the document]
johngmyers: These changes are technically out of scope of this PR. Not blocking.

pkg/model/awsmodel/iam.go
```diff
@@ -124,26 +125,64 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	iamSpec := b.Cluster.Spec.IAM
+	if iamSpec != nil {
+		for _, sa := range iamSpec.ServiceAccountExternalPermissions {
```
johngmyers, line 130: Do we have code to delete the relevant AWS objects when items are removed from this list?

pkg/model/awsmodel/iam.go
```diff
@@ -124,26 +125,64 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+			var p *iam.Policy
+			aws := sa.AWS
```
johngmyers, line 132: We might consider handling `aws == nil` to save overlooking when someone adds a second external system.

johngmyers (changes requested):

pkg/model/awsmodel/iam.go
olemarkus: No, as usual, kOps is quite bad at this. I am wondering if we need a more generic solution that. Maybe we can tag all `RenderAWS`-created resources with `kops.k8s.io/task=taskname` and then at the end of cloudup, remove all resources that match tasks that wasn't created.

I think we need to look at that separately.

pkg/apis/kops/validation/validation.go
```diff
@@ -254,6 +254,52 @@ func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *fie
 		}
 	}
 
+	if spec.IAM != nil {
+		if len(spec.IAM.ServiceAccountExternalPermissions) > 0 {
+			if spec.ServiceAccountIssuerDiscovery == nil || !spec.ServiceAccountIssuerDiscovery.EnableAWSOIDCProvider {
```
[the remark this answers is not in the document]
olemarkus: The intention here was to ensure that AWS OIDC provider has been enabled if user tries to configure external permissions. Technically one should look for the `AWS` substruct of `ServiceAccountExternalPermissions` but since this we only support `AWS` now, we can defer that to later.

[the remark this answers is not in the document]
johngmyers: Sorry, I misread the condition as testing two separate things.

johngmyers (approved):

PR #11043, opened 2021-03-13
Improve instance type validation error message

bharath-123:
This should be reasonably safe as the machineInfo field should not be nil on a valid instance type. Would like some opinions on this.

### DIFF
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -153,7 +153,7 @@ func awsValidateInstanceTypeAndImage(instanceTypeFieldPath *field.Path, imageFie
 		}
 		if !found {
 			allErrs = append(allErrs, field.Invalid(instanceTypeFieldPath, instanceType,
-				fmt.Sprintf("machine type architecture does not match image architecture: %+v - %q", machineInfo.ProcessorInfo, imageArch)))
+				fmt.Sprintf("machine type architecture does not match image architecture: %s - %s", strings.Join(fi.StringSliceValue(machineInfo.ProcessorInfo.SupportedArchitectures), ","), imageArch)))
 		}
 	}
 
--- a/upup/pkg/fi/cloudup/awsup/mock_aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/mock_aws_cloud.go
@@ -308,7 +308,7 @@ func (c *MockAWSCloud) DescribeInstanceType(instanceType string) (*ec2.InstanceT
 	}
 
 	switch instanceType {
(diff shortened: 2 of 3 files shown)


upup/pkg/fi/cloudup/awsup/mock_aws_cloud.go
```diff
@@ -308,7 +308,7 @@ func (c *MockAWSCloud) DescribeInstanceType(instanceType string) (*ec2.InstanceT
 	}
 
 	switch instanceType {
-	case "c5.large", "m3.medium", "m4.large", "m5.large", "m5.xlarge", "t2.micro", "t2.medium", "t3.medium", "t3.large":
+	case "c5.large", "m3.medium", "m4.large", "m5.large", "m5.xlarge", "t2.micro", "t2.medium", "t3.medium", "t3.large", "c4.large":
```
hakman, line 311: Not sure what changes for "c4.large" to be added here.

pkg/apis/kops/validation/aws.go
```diff
@@ -153,7 +153,7 @@ func awsValidateInstanceTypeAndImage(instanceTypeFieldPath *field.Path, imageFie
 		}
 		if !found {
 			allErrs = append(allErrs, field.Invalid(instanceTypeFieldPath, instanceType,
-				fmt.Sprintf("machine type architecture does not match image architecture: %+v - %q", machineInfo.ProcessorInfo, imageArch)))
+				fmt.Sprintf("machine type architecture does not match image architecture: %+v - %s", strings.Join(fi.StringSliceValue(machineInfo.ProcessorInfo.SupportedArchitectures), ","), imageArch)))
```
hakman, line 156: I think now we don't need `%+v` anymore.
```suggestion
				fmt.Sprintf("machine type architecture does not match image architecture: %s - %s", strings.Join(fi.StringSliceValue(machineInfo.ProcessorInfo.SupportedArchitectures), ","), imageArch)))
```

bharath-123: yup.

upup/pkg/fi/cloudup/awsup/mock_aws_cloud.go
bharath-123: Just for tests to pass.. :)

bharath-123: I don't think anything there will be any sort of issues with this..

hakman: 👍

hakman (approved):

Issue #11023, opened 2021-03-12
[toolbox]: instance selector with arm64 arch, outputs wrong arch image


sachincool:
kops Version: 1.19.1 

cli-command:

`kops toolbox  instance-selector nodes --vcpus 4 --flexible --usage-class spot --cpu-architecture arm64 --dry-run `

output:
```
apiVersion: kops.k8s.io/v1alpha2
kind: InstanceGroup
metadata:
  creationTimestamp: null
  labels:
    kops.k8s.io/cluster: test.k8s.local
  name: nodes
spec:
  cloudLabels:
    k8s.io/cluster-autoscaler/enabled: "1"
    k8s.io/cluster-autoscaler/prod.k8s.local: "1"
    kops.k8s.io/instance-selector: "1"
  image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20210119.1
  machineType: a1.xlarge
  maxSize: 10
  minSize: 1
  mixedInstancesPolicy:
    instances:
    - a1.xlarge
    - c6g.xlarge
    onDemandAboveBase: 0
    onDemandBase: 0
    spotAllocationStrategy: capacity-optimized
  nodeLabels:
    kops.k8s.io/instancegroup: nodes
  role: Node
  subnets:
  - ap-south-1a
  - ap-south-1b
  - ap-south-1c
  ```
  
  The AMI  : `099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20210119.1` is x86_64
  
  Expected:  Output AMI should be of the same Arch

PR #11046, opened 2021-03-14
Add channels entries for image architecture

hakman:
This should fix most requests around having correct images for new clusters and upgrades.

Fixes: #11023

/cc @rifelpet @olemarkus @bharath-123

### DIFF
--- a/channels/stable
+++ b/channels/stable
@@ -3,72 +3,94 @@ spec:
     # We put the "legacy" version first, for kops versions that don't support versions ( < 1.5.0 )
     - name: kope.io/k8s-1.4-debian-jessie-amd64-hvm-ebs-2017-07-28
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.4.0 <1.5.0"
     - name: kope.io/k8s-1.5-debian-jessie-amd64-hvm-ebs-2018-08-17
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.5.0 <1.6.0"
     - name: kope.io/k8s-1.6-debian-jessie-amd64-hvm-ebs-2018-08-17
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.6.0 <1.7.0"
     - name: kope.io/k8s-1.7-debian-jessie-amd64-hvm-ebs-2018-08-17
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.7.0 <1.8.0"
     - name: kope.io/k8s-1.8-debian-stretch-amd64-hvm-ebs-2018-08-17
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.8.0 <1.9.0"
     - name: kope.io/k8s-1.9-debian-stretch-amd64-hvm-ebs-2018-08-17
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.9.0 <1.10.0"
     - name: kope.io/k8s-1.10-debian-stretch-amd64-hvm-ebs-2018-08-17
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.10.0 <1.11.0"
     # Stretch is the default for 1.11 (for nvme)
     - name: kope.io/k8s-1.11-debian-stretch-amd64-hvm-ebs-2021-02-05
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.11.0 <1.12.0"
     - name: kope.io/k8s-1.12-debian-stretch-amd64-hvm-ebs-2021-02-05
       providerID: aws
+      architectureID: amd64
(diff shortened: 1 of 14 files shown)


channels/stable
```diff
@@ -3,72 +3,94 @@ spec:
...
       kubernetesVersion: ">=1.17.0 <1.18.0"
     - name: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20210119.1
       providerID: aws
+      architectureID: amd64
       kubernetesVersion: ">=1.18.0"
-    - providerID: gce
+    - name: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64-server-20210119.1
+      providerID: aws
+      architectureID: arm64
+      kubernetesVersion: ">=1.18.0"
```
rifelpet, line 68: At the end of `pkg/apis/kops/channel.go`'s `FindImage` function we log a warning if multiple matches are found. 

This means that as soon as this PR is merged, all existing releases of kops will begin emitting a warning on k8s clusters >= k8s 1.18.0. Perhaps we only add an arm64 image starting with the k8s version associated with the kops version we anticipate this being released in?

hakman: Agreed. Should be fine now.

bharath-123:
Tried testing this out with the toolbox command, got this error
```
bharath12345@Bharaths-MacBook-Pro kops % kops toolbox instance-selector nodes --vcpus 4 --flexible --usage-class spot --cpu-architecture arm64 --cluster-autoscaler --dry-run
Using cluster from kubectl context: test-stuff.kops-cluster.com

I0314 18:48:18.623902   99607 populate_instancegroup_spec.go:260] Cannot set default Image for CloudProvider="aws"

unable to determine default image for InstanceGroup nodes
```

pkg/apis/kops/channel.go
```diff
@@ -259,13 +262,16 @@ const (
...
+func (c *Channel) FindImage(provider CloudProviderID, kubernetesVersion semver.Version, architecture architectures.Architecture) *ChannelImageSpec {
 	var matches []*ChannelImageSpec
 
 	for _, image := range c.Spec.Images {
 		if image.ProviderID != string(provider) {
 			continue
 		}
+		if image.ArchitectureID != string(architecture) {
+			continue
+		}
```
rifelpet, line 274: I'm wondering if we shouldn't `continue` if `image.ArchitectureID != ""`. I dont know the workflows that users have for hosting their own channels files but if they dont copy the associated architecture additions in this PR when upgrading to this version of kops then they'll no longer find any images. Perhaps a release note mentioning the change for offline installations would be sufficient.

rifelpet:
@bharath-123 your kops build is likely using the upstream master channels file which doesn't have the architecture additions in it. If you update your cluster spec's `channel` field to the file's URL from this PR then it should work.

https://raw.githubusercontent.com/kubernetes/kops/a5f5f984c99a9cf6b31ab87beee15a5248f50107/channels/stable

pkg/apis/kops/channel.go
hakman: I think for now we should go with the release note. We can see if later we need to improve.

bharath-123:
> @bharath-123 your kops build is likely using the upstream master channels file which doesn't have the architecture additions in it. If you update your cluster spec's `channel` field to the file's URL from this PR then it should work.
> 
> https://raw.githubusercontent.com/kubernetes/kops/a5f5f984c99a9cf6b31ab87beee15a5248f50107/channels/stable

Got it @rifelpet Thanks for clearing that out!

pkg/util/templater/template_functions.go
```diff
@@ -63,9 +64,9 @@ func (r *Templater) templateFuncsMap(tm *template.Template) template.FuncMap {
 
 	}
 
-	funcs["ChannelRecommendedImage"] = func(cloud, k8sVersion string) string {
+	funcs["ChannelRecommendedImage"] = func(cloud, k8sVersion string, architecture architectures.Architecture) string {
```
rifelpet, line 67: normally changing a template function's signature would be worthy of a release note but this hasnt been in a stable release yet so we should be fine. we do need to update the docs that use this function though https://github.com/kubernetes/kops/blob/master/docs/operations/cluster_template.md

hakman: Done.

rifelpet (approved):

yurrriq:
I'm running into issues using kOps 1.20.0: `wrong type for value; expected architectures.Architecture; got string`

using this `snippets/common/ami`:
```
{{ default (ChannelRecommendedImage $.cloud (include "kubernetesVersion" $) $.architecture) $.image }}
```
snippets/common/kubernetesVersion:
```
{{ default ChannelRecommendedKopsKubernetesVersion $.kubernetesVersion }}
```

with this template
```
{{ $image := include "ami" $) }}
```

and these values:
```yaml
architecture: amd
cloud: aws
```

PR #10910, opened 2021-02-23
Further improve cloudLabel validation

olemarkus:
We were too strict on some labels that blocked common use cases such as CAS. At the same time we allowed IG-level cloudLabels that could result in broken clusters.

Fixes #10862

PR #11048, opened 2021-03-14
Remove instance-selector label

bharath-123:
`kops toolbox instance-selector` fails due to https://github.com/kubernetes/kops/pull/10910 which improves the label validation. Making kops.k8s.io/instance-selector: "1" a node label rather than a cloud label should bypass this validation check. I also think it should make more sense to make it a node label rather than a cloud label.

### DIFF
--- a/cmd/kops/toolbox_instance_selector.go
+++ b/cmd/kops/toolbox_instance_selector.go
@@ -500,12 +500,6 @@ func decorateWithMixedInstancesPolicy(instanceGroup *kops.InstanceGroup, usageCl
 		return nil, fmt.Errorf("error node usage class not supported")
 	}
 
-	generatedWithLabelKey := "kops.k8s.io/instance-selector"
-	if ig.Spec.CloudLabels == nil {
-		ig.Spec.CloudLabels = make(map[string]string)
-	}
-	ig.Spec.CloudLabels[generatedWithLabelKey] = "1"
-
 	return ig, nil
 }
 


cmd/kops/toolbox_instance_selector.go
```diff
@@ -501,10 +501,10 @@ func decorateWithMixedInstancesPolicy(instanceGroup *kops.InstanceGroup, usageCl
 	}
 
 	generatedWithLabelKey := "kops.k8s.io/instance-selector"
```
hakman, line 503: @olemarkus what do you think, should we remove this completely?

bwagner5: My intention for this label was to show infrastructure operators within the AWS console that the group was generated with instance selector.  I thought it would be more helpful as a real tag rather than a node label for that reason.

bharath-123: i understand @bwagner5 . Curious to know why that info will be useful for operators?

bharath-123: @bwagner5 would it cause any issues if this label was removed?

bwagner5: I don't think so, I guess it's not too useful. I was thinking it would be nice to know which ones were generated with instance-selector in-case there were instance types that weren't actually fit for the compute nodes that were getting included. But I suppose it doesn't really matter since you can just hand modify the list.

hakman (approved):
Thanks everyone for the feedback!

PR #11054, opened 2021-03-15
Download kubectl to /opt/kops/bin on Flatcar OS

rifelpet:
Also add it to protokube's PATH.

[All of our flatcar jobs](https://testgrid.k8s.io/kops-distro-flatcar#Summary) are currently failing because channels arent being applied.
A newly added error log reports that kubectl isn't in protokube's PATH.

This adds the kubectl's location (/opt/bin) to protokube's PATH.

See https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-distro-imageflatcar/1371379886664454144/artifacts/54.206.100.130/protokube.log (search `error running kubectl` )

### DIFF
--- a/nodeup/pkg/model/context.go
+++ b/nodeup/pkg/model/context.go
@@ -398,7 +398,7 @@ func (c *NodeupModelContext) UseSecureKubelet() bool {
 func (c *NodeupModelContext) KubectlPath() string {
 	kubeletCommand := "/usr/local/bin"
 	if c.Distribution == distributions.DistributionFlatcar {
-		kubeletCommand = "/opt/bin"
+		kubeletCommand = "/opt/kops/bin"
 	}
 	if c.Distribution == distributions.DistributionContainerOS {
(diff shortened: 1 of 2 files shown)


bharath-123:
This makes sense i think. When protokube was containerized it had its own filesystem with kubectl. now that we are running it as a systemd process, we are picking up kubectl from the host filesystem and we need to give the right PATHs for it.

Should we add another case for containerOS too? because that seems to be adding kubectl to /home/kubernetes/bin

rifelpet:
@hakman was wondering if it would make more sense to add kubectl to /opt/kops/bin alongside the other binaries that kops downloads. I tend to agree but we're going to ask for other input at this thursday's office hours before proceeding.

bharath-123:
I am thinking if that would be a good idea to include other binaries like kubelet and so too? That would remove dependencies on the underlying distribution used and we can have a fixed location for the binaries added by kOps.

hakman (approved):

PR #11057, opened 2021-03-16
Update node local dns cache

zetaab:


upup/models/cloudup/resources/addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -192,3 +192,24 @@ spec:
           items:
             - key: Corefile
               path: Corefile.base
+---
```
olemarkus, line 195: We don't ship these headless services typically used by prom operator as we don't ship prom operator either.

Also, `PodMonitor` avoids the need for these services.

olemarkus (changes requested):
Looks good except for the headless service that I don't feel belongs as part of the addon.

zetaab:
@olemarkus I just updated the node local dns, maybe that is enough for now.

olemarkus (approved):
Thanks man

PR #11080, opened 2021-03-20
Logging cleanup

rifelpet:
After staring at prow job outputs and nodeup logs for a long time, i'm removing some of the unnecessary logging statements.

The first will remove these in the kops cli:

```
 I0318 08:50:47.113867    2772 executor.go:111] Tasks: 80 done / 82 total; 2 can run
I0318 08:50:48.918239    2772 executor.go:137] Task "AutoscalingGroup/master-ap-northeast-1a.masters.e2e-e6110d5e7d-a98da.test-cncf-aws.k8s.io" not ready: waiting for the IAM Instance Profile to be propagated
I0318 08:50:48.918286    2772 executor.go:137] Task "AutoscalingGroup/nodes-ap-northeast-1a.e2e-e6110d5e7d-a98da.test-cncf-aws.k8s.io" not ready: waiting for the IAM Instance Profile to be propagated
I0318 08:50:48.918297    2772 executor.go:155] No progress made, sleeping before retrying 2 task(s) 
```

The second will remove these from the [nodeup logs](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-distro-imagerhel8/1372953273908072448/artifacts/ip-172-20-37-77.ap-southeast-1.compute.internal/journal.log):

```
Mar 19 16:52:10.691883 ip-172-20-37-77.ap-southeast-1.compute.internal cloud-init[2044]: I0319 16:52:10.684448    4573 task.go:103] task *nodetasks.Service does not implement HasLifecycle
```

### DIFF
--- a/upup/pkg/fi/executor.go
+++ b/upup/pkg/fi/executor.go
@@ -134,7 +134,7 @@ func (e *executor) RunTasks(taskMap map[string]Task) error {
 
 				remaining := time.Second * time.Duration(int(time.Until(ts.deadline).Seconds()))
 				if _, ok := err.(*TryAgainLaterError); ok {
-					klog.Infof("Task %q not ready: %v", ts.key, err)
+					klog.V(2).Infof("Task %q not ready: %v", ts.key, err)
 				} else {
 					klog.Warningf("error running task %q (%v remaining to succeed): %v", ts.key, remaining, err)
 				}
--- a/upup/pkg/fi/task.go
+++ b/upup/pkg/fi/task.go
@@ -98,18 +98,18 @@ func (c *ModelBuilderContext) setLifecycleOverride(task Task) Task {
 	// TODO(@chrislovecnm) - wonder if we should update the nodeup tasks to have lifecycle
 	// TODO - so that we can return an error here, rather than just returning.
 	// certain tasks have not implemented HasLifecycle interface
-	hl, ok := task.(HasLifecycle)
-	if !ok {
-		klog.V(8).Infof("task %T does not implement HasLifecycle", task)
-		return task
-	}
-
 	typeName := TypeNameForTask(task)
 	klog.V(8).Infof("testing task %q", typeName)
 
 	// typeName can be values like "InternetGateway"
 	value, ok := c.LifecycleOverrides[typeName]
 	if ok {
+		hl, okHL := task.(HasLifecycle)
+		if !okHL {
+			klog.Warningf("task %T does not implement HasLifecycle", task)
+			return task
+		}
+
 		klog.Warningf("overriding task %s, lifecycle %s", task, value)
 		hl.SetLifecycle(value)
 	}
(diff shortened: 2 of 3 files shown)


upup/pkg/fi/executor.go
```diff
@@ -134,7 +134,7 @@ func (e *executor) RunTasks(taskMap map[string]Task) error {
 
 				remaining := time.Second * time.Duration(int(time.Until(ts.deadline).Seconds()))
 				if _, ok := err.(*TryAgainLaterError); ok {
-					klog.Infof("Task %q not ready: %v", ts.key, err)
+					klog.V(2).Infof("Task %q not ready: %v", ts.key, err)
```
hakman, line 137: This is a special case where things fail because something else is not ready due to how AWS works.
Usually you expect the tasks to succeed in the order they are and any retries imply some issue.

rifelpet: I think if we're dealing with eventual consistency (what TryAgainLaterError seems to imply) then it is within the expected behavior of the system, and therefor isn't worth alerting to the user at the default verbosity levels that it has occurred. If tasks consistently fail or fail for different reasons those situations will still be logged normally.

I just don't think this is a valuable enough piece of information to surface to the user.

upup/pkg/fi/task.go
```diff
@@ -98,18 +98,18 @@ func (c *ModelBuilderContext) setLifecycleOverride(task Task) Task {
...
 	// TODO - so that we can return an error here, rather than just returning.
 	// certain tasks have not implemented HasLifecycle interface
-	hl, ok := task.(HasLifecycle)
-	if !ok {
-		klog.V(8).Infof("task %T does not implement HasLifecycle", task)
-		return task
-	}
-
 	typeName := TypeNameForTask(task)
 	klog.V(8).Infof("testing task %q", typeName)
```
justinsb, line 102: This log line feels very V(8) :-)

upup/pkg/fi/executor.go
justinsb: I'd say you're probably both right ...

As a user, I don't much care about one or two failures here.  I probably do care if it gets "stuck" to know why it's stuck.  And I probably also want to know roughly what is going on ("waiting for 2 tasks" ?).

Maybe we should do something different on iteration 1 vs iteration 10?

rifelpet: if a task is stuck then kops will keep retrying it until 10 minutes have elapsed and log a more descriptive error:

https://github.com/kubernetes/kops/pull/11080/files#diff-d73723c1c245d32c779df66e821830949e518a324bd48cfcf8299ef00a2194dfR105

kops also logs if no progress has been made on any remaining tasks. I think both of these are sufficient enough to warrant not needing to log the specific task that isn't ready. Users can always `-v 2` if they need more detail.

hakman (approved):
Thought about this a little more and I think the fix is ok as is for now.

PR #11081, opened 2021-03-20
Ensure a publicdatastore exists for jwks and that it can only be s3

olemarkus:


pkg/apis/kops/validation/legacy.go
```diff
@@ -371,6 +371,20 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 			allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("keyStore"), "Vault keystore is only available on AWS"))
 		}
 	}
+
+	publicDataStore := c.Spec.PublicDataStore
+	if publicDataStore != "" {
+		if !strings.HasPrefix(publicDataStore, "s3://") {
+			allErrs = append(allErrs, field.Invalid(fieldSpec.Child("publicDataStore"), publicDataStore, "S3 is the only supported VFS for publicStore"))
```
justinsb, line 378: We might want to insert "currently" or something like that :-)

justinsb:
Looks good.  I'm not sure if we should consider specifying publicDataStore without the feature flag to be an error or a warning or just ignore it, but that's possibly a future PR.

In any case, this would have helped me, so thank you!

/approve
/lgtm

PR #11084, opened 2021-03-20
Update containerd to v1.3.10/v1.4.4

bmelbourne:
Update containerd to `v1.3.10` and `v1.4.4` which contains a fix for CVE-2021-21334.

[containerd CRI plugin: environment variables can leak between containers](https://github.com/containerd/containerd/security/advisories/GHSA-6g2q-w5j3-fwh4)

**Which issue(s) this PR fixes**:
N/A

None


upup/pkg/fi/cloudup/containerd.go
```diff
@@ -168,22 +168,26 @@ func findContainerdVersionHash(arch architectures.Architecture, version string)
...
 func findAllContainerdDockerMappings() map[string]string {
 	versions := map[string]string{
-		"1.3.7": "19.03.13",
-		"1.3.9": "19.03.14",
-		"1.4.3": "20.10.0",
+		"1.3.7":  "19.03.13",
+		"1.3.9":  "19.03.14",
+		"1.3.10": "19.03.14",
+		"1.4.3":  "20.10.0",
+		"1.4.4":  "20.10.0",
```
hakman, line 190: I don't think this is correct. There is no v1.4.4 released as part of the docker packages.

bmelbourne: After reviewing the code, my understanding of the `findAllContainerdDockerMappings` function to install the corresponding version of Docker where the version of containerd doesn't exist, such as for ARM64 architectures.

hakman: Yes, but only if the desired version exists as part of a Docker package.
If not, it should use the `containerdFallbackVersion`, which should be used in case the selected version is not available for ARM64.

upup/pkg/fi/cloudup/containerd.go
```diff
@@ -35,7 +35,7 @@ const (
 	// containerd legacy packages URLs for v1.2.x and v1.3.x
 	containerdLegacyUrlAmd64 = "https://storage.googleapis.com/cri-containerd-release/cri-containerd-%s.linux-amd64.tar.gz"
 	// containerd version that is available for both AMD64 and ARM64, used in case the selected version is not available for ARM64
-	containerdFallbackVersion = "1.4.3"
+	containerdFallbackVersion = "1.4.4"
```
hakman, line 38: Please revert here, at the moment there is not 1.4.4 version for ARM64, so the fallback remains 1.4.3.

bmelbourne: There are `v1.4.4` rpms available for CentOS 7/8 and Ubuntu 20.10 I believe, but it's not widely available, so happy to keep the fallback version as `v1.4.3`.

hakman: kOps uses static binaries, not the rpm/deb. Unfortunately we can't do much until the next release.

upup/pkg/fi/cloudup/containerd.go
bmelbourne: I've removed v1.4.4 from the `findAllContainerdDockerMappings` function and re-running the E2E tests now.

upup/pkg/fi/cloudup/containerd.go
```diff
@@ -168,22 +168,25 @@ func findContainerdVersionHash(arch architectures.Architecture, version string)
...
 }
 
 func findAllContainerdDockerMappings() map[string]string {
 	versions := map[string]string{
-		"1.3.7": "19.03.13",
-		"1.3.9": "19.03.14",
-		"1.4.3": "20.10.0",
+		"1.3.7":  "19.03.13",
+		"1.3.9":  "19.03.14",
+		"1.3.10": "19.03.14",
```
hakman, line 188: I think this has to be reverted also. Or maybe there is a newer Docker version.

bmelbourne: Yes, removed that one as well now.

hakman:
Thanks @bmelbourne. 👍

hakman (approved):

PR #11085, opened 2021-03-20
Improve error messages around PublicJWKS

justinsb:
I left off the publicDataStore (must pass --overwrite on create, I
believe), and the error message was a type-cast failure.


olemarkus (approved):
I also have #11081 which solves the thing. Good to guard this in multiple places though.

PR #11086, opened 2021-03-20
Don't add control-plane DNS permissions with UseServiceAccountIAM

justinsb:
Should not be needed; dns-controller should run on the control-plane
node so there should not be a bootstrapping problem with the nodes.

Reverts #10529

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -257,10 +257,12 @@ func (r *NodeRoleMaster) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 		addKMSIAMPolicies(p, stringorslice.Slice(b.KMSKeys), b.Cluster.Spec.IAM.Legacy)
 	}
 
-	if b.Cluster.Spec.IAM.Legacy {
-		addLegacyDNSControllerPermissions(b, p)
+	if !b.UseServiceAccountIAM {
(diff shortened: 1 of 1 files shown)


justinsb:
cc @rifelpet 

I'm not quite sure what happened before; dns-controller should have had permissions to set up the DNS names, and it should run on the control-plane.

justinsb:
/assign @rifelpet 

I think you have the best context here!

rifelpet:
I'm wondering if this would cause issues when enabling UseServiceAccountIAM on an existing cluster. `kops update cluster --yes` would immediately revoke the IAM permissions that dns-controller relies on, and then eventually dns-controller gets redeployed with the service account and the `AWS_WEB_IDENTITY_TOKEN_FILE` env var. During the time between those, DNS records won't be getting updated and this would likely be during cluster maintenance in which control plane instances are getting replaced and records need changing.

Do you think this is significant enough of an issue to warrant addressing?

olemarkus:
I made the same change as part of https://github.com/kubernetes/kops/pull/11088

If you run single-node CP and you roll the CP node, dns-controller is gone anyway.
If you run three-node CP and you roll a CP node, channels will update dns-controller long before the new node has booted. So I think this fairly safe.

rifelpet (approved):
Makes sense 👍🏻

PR #11089, opened 2021-03-20
Replace go-bindata with go:embed

rifelpet:
Now that we're on Go 1.16 we can replace our use of go-bindata with go:embed directives.

/hold for comment and input on timing


WORKSPACE
```diff
@@ -6,19 +6,19 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "7904dbecbaffd068651916dce77ff3437679f9d20e1a7956bff43826e7645fcc",
+    sha256 = "69de5c704a05ff37862f7e0f5534d4f479418afc21806c887db544a316f3cb6b",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.25.1/rules_go-v0.25.1.tar.gz",
-        "https://github.com/bazelbuild/rules_go/releases/download/v0.25.1/rules_go-v0.25.1.tar.gz",
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.27.0/rules_go-v0.27.0.tar.gz",
+        "https://github.com/bazelbuild/rules_go/releases/download/v0.27.0/rules_go-v0.27.0.tar.gz",
```
justinsb, line 12: If you wanted to split the rules_go / gazelle commit out, I think it's an easy merge...  It's not _that_ big though, so I'm fine with leaving it in

upup/models/vfs.go
```diff
@@ -17,17 +17,21 @@ limitations under the License.
 package models
 
 import (
+	"embed"
 	"errors"
 	"io"
+	"io/fs"
```
justinsb, line 23: Unrelated, but maybe worth looking at whether we can use more of `io/fs` in our vfs layer :-)

WORKSPACE
rifelpet: I ran into some issues with the rules_go upgrade that you'll see in the prow job failures shortly. Fixing it is a bit beyond my bazel knowledge though, so the upgrade will likely end up as a separate PR

upup/models/vfs.go
```diff
@@ -68,28 +72,24 @@ func (p *AssetPath) WriteTo(out io.Writer) (int64, error) {
...
+		return nil, os.ErrNotExist
 	}
 	return data, err
 }
 
 func (p *AssetPath) ReadDir() ([]vfs.Path, error) {
-	files, err := AssetDir(p.location)
+	files, err := content.ReadDir(p.location)
 	if err != nil {
-		// Yuk
```
justinsb, line 84: Hooray!

justinsb:
Personally I'm in favor.  I don't think it makes backporting much harder.  The only real risk would be if we found a critical problem with go 1.16 and had to go back to 1.15, but I don't think that's particularly likely.

WORKSPACE
rifelpet: the bazel prow jobs didn't fail... 🤔 they did on my mac. Heres the output:

```
ERROR: /path/to/src/k8s.io/kops/protokube/cmd/protokube/BUILD.bazel:29:10: Configurable attribute "out" doesn't match this configuration (would a default condition help?).
Conditions checked:
 @io_bazel_rules_go//go/platform:linux_amd64
 @io_bazel_rules_go//go/platform:linux_arm64
ERROR: Analysis of target '//protokube/cmd/protokube:protokube' failed; build aborted: /path/to/src/k8s.io/kops/protokube/cmd/protokube/BUILD.bazel:29:10: Configurable attribute "out" doesn't match this configuration (would a default condition help?).
Conditions checked:
 @io_bazel_rules_go//go/platform:linux_amd64
 @io_bazel_rules_go//go/platform:linux_arm64
INFO: Elapsed time: 3.473s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (18 packages loaded, 18 targets configured)
FAILED: Build did NOT complete successfully (18 packages loaded, 18 targets configured)
    Fetching @distroless_base_arm64; Restarting.
    Fetching @distroless_base_amd64; Restarting.
    Fetching @go_puller_darwin; fetching
    Fetching @local_config_cc; fetching
    Fetching @org_golang_x_tools; Restarting.
make: *** [bazel-test] Error 1
```

rifelpet: Seems to be that `bazel test` fails on a mac because we dont have darwin outputs for protokube and channels:

https://github.com/kubernetes/kops/blob/2b46042241f98d57c1d9523b470e7be4d264ed38/protokube/cmd/protokube/BUILD.bazel#L29-L37

https://github.com/kubernetes/kops/blob/2b46042241f98d57c1d9523b470e7be4d264ed38/channels/cmd/channels/BUILD.bazel#L16-L24

Adding outputs for darwin works fine but I'm not sure if thats the best solution here.

hakman: It is, don’t worry:)

hakman:
Same here, don’t see any issues regarding backporting and timing.

hakman (approved):

Issue #11050, opened 2021-03-14
Improve time to resolve API DNS record of new clusters


rifelpet:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

master

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

1.18 though the behavior happens on various minor versions

**3. What cloud provider are you using?**

aws

This is happening on various e2e prow jobs. I'm going to use [this](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-u1804-k18-containerd/1370875829445201920) one as an example. The cluster takes 5m16s to have its DNS record updated from the placeholder IP to the control plane hosts.

A timeline of events:
* [23:15:47](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-u1804-k18-containerd/1370875829445201920#1:build-log.txt%3A127) - the DNS placeholder records are created w/ a 10s TTL as a part of `kops create cluster`
* [23:15:48](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-u1804-k18-containerd/1370875829445201920#1:build-log.txt%3A140) - `validate cluster` is invoked with 15 minute timeout
* [23:15:59](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-u1804-k18-containerd/1370875829445201920#1:build-log.txt%3A161) - validation resolves the DNS record to the placeholder IP
* [23:21:03](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-u1804-k18-containerd/1370875829445201920#1:build-log.txt%3A630) - DNS record now points to apiserver, **this is the time i'd like to see if we can reduce**

On the control plane host we see this timeline:

* [23:16:20](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-grid-u1804-k18-containerd/1370875829445201920/artifacts/ip-172-20-48-213.eu-west-1.compute.internal/journal.log) - nodeup starts for the first time
* [23:17:20](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-grid-u1804-k18-containerd/1370875829445201920/artifacts/ip-172-20-48-213.eu-west-1.compute.internal/journal.log) - protokube starts for the first time
[...]

PR #11093, opened 2021-03-21
Ensure protokube can connect to kube-apiserver before starting the sync loop

olemarkus:
This commit will have protokube eagerly retry to connect to kube-apiserver instead of the 1 minute sync loop.
Another benefit is that we know that tasks are run in order and that labels are always applied before channels etc

With very few samples, it looks like a cluster gets the DNS entry about 1-2 minutes earlier, which I think makes sense.

Ref #11050

### DIFF
--- a/protokube/pkg/protokube/kube_boot.go
+++ b/protokube/pkg/protokube/kube_boot.go
@@ -18,10 +18,12 @@ package protokube
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"path/filepath"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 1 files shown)


protokube/pkg/protokube/kube_boot.go
```diff
@@ -101,8 +102,23 @@ func (k *KubeBoot) Init(volumesProvider Volumes) {
 
 // RunSyncLoop is responsible for provision the cluster
 func (k *KubeBoot) RunSyncLoop() {
+	ctx := context.Background()
+
+	for {
+		client, err := k.Kubernetes.KubernetesClient()
+		if err != nil {
+			klog.Warningf("could not create kubernetes client: %v", err)
```
justinsb, line 110: This one probably needs a sleep and continue?  Or klog.Fatalf?  I think we'll panic on the `client.CoreV1()` call anyway..

protokube/pkg/protokube/kube_boot.go
```diff
@@ -101,8 +102,23 @@ func (k *KubeBoot) Init(volumesProvider Volumes) {
...
+		}
+
+		_, err = client.CoreV1().Namespaces().Get(ctx, "kube-system", metav1.GetOptions{})
+		if err == nil {
+			klog.Info("connected to the apiserver")
+			break
+		}
+		klog.Infof("Could not connect to apiserver: %v", err)
```
justinsb, line 118: Nit: something to indicate that this is normal(ish) might be helpful here.  e.g. "polling for apiserver readiness; not yet ready: %v" or something like that?

justinsb:
This is a nice win.  LGTM; one error handling comment and one nit.

/approve
/lgtm

/hold

Feel free to remove the hold if you don't think the nits are worth it :-)

protokube/pkg/protokube/kube_boot.go
olemarkus: We may as well panic here. There isn't much we can do if this one fails.

rifelpet:
FWIW, Looking at the protokube logs from the first commit's [kubernetes-aws](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/11093/pull-kops-e2e-kubernetes-aws/1373576840320389120/artifacts/ip-172-20-60-121.ap-northeast-2.compute.internal/journal.log) and [k8s-containerd](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/11093/pull-kops-e2e-k8s-containerd/1373576840349749248/artifacts/ip-172-20-34-189.ap-northeast-2.compute.internal/journal.log) jobs it seems we retried 19 and 22 times respectively. We'll see how the most recent jobs perform as well.

rifelpet (approved):
most recent jobs both retried 15 times (75 seconds)

justinsb:
Looks good, going to remove the hold (I had meant it to mean "feel free to remove hold when you want it to merge")

/hold cancel

PR #10672, opened 2021-01-28
Remove BLM banner 

hakman:
Banner was already removed from the main Kubernetes site. I think it's time to do the same:
https://github.com/kubernetes/website/blob/87c6923445193d6a5d8b3d029c1f4bb4ee6562cc/data/announcements/scheduled.yaml#L55-L63

/kind office-hours

PR #11094, opened 2021-03-21
Add values page

justinsb:
As discussed in #10672, we are maintaining the BLM statement on a page
dedicated to our values.

Additional values mirror the kubernetes project values.  We can
document additional values by agreement in office hours.

### DIFF
--- a/docs/values.md
+++ b/docs/values.md
@@ -1,0 +1,69 @@
+# kOps values
+
+kOps is developed by mutual collaboration; we document our values so
+that we can collaborate more effectively and without conflict.
+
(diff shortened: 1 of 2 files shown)


rifelpet:
Should this block be removed? https://github.com/kubernetes/kops/blob/c65658f79ff32ed3217a1126e1a6807da8e8beda/docs/overrides/main.html#L30-L34

justinsb:
@rifelpet yes, but I think we can do that by merging #10672 once this one has merged.

rifelpet (approved):

PR #11069, opened 2021-03-18
Dns controller fixes

olemarkus:
My original intention was to try to correctly document how dns-controller worked. I had to test some code to check the actual behaviour, so may as well commit those too.

Ref #11055

PR #11095, opened 2021-03-21
Deeper validation in dns controller tests

justinsb:
Follow up to #11069; use cmp.Diff to compare got vs want.


olemarkus (approved):
Thanks for cleaning this up!

rifelpet (approved):

PR #11096, opened 2021-03-21
Update google SDK libraries

justinsb:
Updating to latest google SDK libraries, will be needed for
cluster-api support.


justinsb:
/hold

Introduces new protobuf dependency which I want to validate...

olemarkus:
Not really related, but what is our plan for cluster-api support these days?

justinsb:
Re cluster-api, I think the blocking pieces have merged.  I was going to start the PR train to enable them (optionally) on GCE.  Targeting nodes only initially.  The big piece we needed (and might need a few more things for) was for kops-controller to support booting without the state store - but I _think_ that's mostly in now.

justinsb:
OK so I was worried about that proto dependency, but it looks fine!

/hold cancel

olemarkus (approved):

PR #11098, opened 2021-03-21
Use etcd-manager built from etcdadm repo

justinsb:
We can now use etcd-manager as built from the kubernetes-sigs/etcdadm
repo.

### DIFF
--- a/pkg/model/components/etcd.go
+++ b/pkg/model/components/etcd.go
@@ -24,7 +24,7 @@ import (
 	"k8s.io/kops/upup/pkg/fi/loader"
 )
 
-const DefaultBackupImage = "kopeio/etcd-backup:3.0.20210122"
+const DefaultBackupImage = "k8s.gcr.io/etcdadm/etcd-backup:3.0.20210430"
 
 // EtcdOptionsBuilder adds options for etcd to the model
 type EtcdOptionsBuilder struct {
--- a/pkg/model/components/etcdmanager/model.go
+++ b/pkg/model/components/etcdmanager/model.go
@@ -172,7 +172,7 @@ metadata:
   namespace: kube-system
 spec:
   containers:
-  - image: kopeio/etcd-manager:3.0.20210228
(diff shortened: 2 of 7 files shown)


justinsb:
This image is built from https://github.com/kubernetes-sigs/etcdadm . The intent is that we can make that repo the canonical location.  Before merging, we want to get the image built using the k8s staging and image promotion systems.

This image is currently built manually, and is just intended to verify that the version built from the new repo works (i.e. that no problems have been introduced in the repo move)

justinsb:
cc @hakman 

So I finally got the build working and [proposed the latest image for promotion](https://github.com/kubernetes/k8s.io/pull/1991)

One thing I realized .... I didn't promote the -amd64 and -arm64 tags.  Are these needed?  If so I can add them to the process (which is a bit of a WIP) and propose another promotion.

Another snafu is (as you pointed out) that it's not as clear which version we are promoting ... we need to tag in the source repo and/or maybe plumb through the version in something like our versions.go.

hakman:
Thanks @justinsb. This is pretty cool.
There is no need for the -amd64 and the -arm64 tags, at least for us. I don't think we do it for the kOps images either.
Not quite sure what's the best way forward for versions. Tagging in the source repo would be good, but not sure how well this will go long term.

hakman:
Btw, this is ready from my point of view.
After we merge https://github.com/kubernetes-sigs/etcdadm/pull/205 we could switch to v3.1.x.
/lgtm

justinsb:
As discussed in office hours, the remaining things shouldn't be blockers for kops adoption, so removing WIP

hakman:
Thanks @justinsb. I tracked the commits that removed support for etcd v2 in kOps and we are good later with back-porting to 1.20:
https://github.com/kubernetes/kops/pull/10141
https://github.com/kubernetes/kops/pull/10245

hakman:
@justinsb should this also bump the backup image?

hakman (approved):

PR #11099, opened 2021-03-21
Don't grant S3 node IAM permissions when reading from kops-controller

justinsb:
If we're reading from kops-controller, we don't need the S3 IAM permissions.

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/apis/kops/model"
+	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/pkg/util/stringorslice"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awstasks"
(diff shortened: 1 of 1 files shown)


justinsb:
Part of the cluster-api stream.

With this PR & feature-flag, the nodes have these permissions:

```
    "PolicyDocument": {
        "Statement": [
            {
                "Action": [
                    "ec2:DescribeInstances",
                    "ec2:DescribeRegions"
                ],
                "Effect": "Allow",
                "Resource": [
                    "*"
                ]
            },
            {
                "Action": [
                    "ecr:GetAuthorizationToken",
                    "ecr:BatchCheckLayerAvailability",
                    "ecr:GetDownloadUrlForLayer",
                    "ecr:GetRepositoryPolicy",
                    "ecr:DescribeRepositories",
                    "ecr:ListImages",
                    "ecr:BatchGetImage"
                ],
                "Effect": "Allow",
                "Resource": [
                    "*"
                ]
            }
        ],
        "Version": "2012-10-17"
    }
```

justinsb:
/retest

Odd DNS problem:

```
error determining default DNS zone: error querying zones: Throttling: 
	status code: 400, request id: 6d486a47-686e-44c9-a502-adaca320e8de
```

rifelpet:
we've been experiencing route53 throttling a lot with our prow jobs. Route53 has a per-region [5 req / sec](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests-route-53-resolver) rate limit. We've seen this throttling error on both kops cli and dns-controller. the dns controller errors will retry on the next sync and succeed but lead to delayed cluster validation. the kops cli errors cause a failure as seen here.

I attempted a fix in https://github.com/kubernetes/kops/pull/11020 but its not clear whether it actually helped or not (too many jobs to search for throttling errors in their logs) I'd definitely like to have kops and dns-controller handle this throttling better.

I'm also hoping to increase the splay time of many of our jobs, currently jobs that are run a [certain # of times per day](https://github.com/kubernetes/test-infra/blob/9081550c34dc5c30a04276846874cb1b91316ad0/config/jobs/kubernetes/kops/build_jobs.py#L671) are only splayed across 1 hour rather than the entire interval between jobs. I'm hoping that helps avoid throttling. Since the quota is per-region we may also be able to look into adding any recently introduced regions.

rifelpet:
should this have an integration test that more clearly shows the different node IAM permissions?

PR #11101, opened 2021-03-21
Have nodeup retry kops-controller bootstrapping sooner if DNS isn't setup

rifelpet:
ref: https://github.com/kubernetes/kops/issues/11050#issuecomment-803620244

### DIFF
--- a/upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
+++ b/upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
@@ -27,13 +27,16 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"path"
+	"time"
 
 	"k8s.io/kops/pkg/apis/nodeup"
 	"k8s.io/kops/pkg/pki"
 	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/cloudup"
 )
 
 type BootstrapClientTask struct {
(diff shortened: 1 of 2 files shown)


olemarkus (approved):

rifelpet:
hm looks like these errors are being hit over 10 minutes after the dns record is created. Will need some additional troubleshooting

justinsb:
Yeah, not sure what's going on there but I see the same thing.  Maybe log the error and the host name we're resolving?

upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go
```diff
@@ -155,6 +159,19 @@ func (b *KopsBootstrapClient) QueryBootstrap(ctx context.Context, req *nodeup.Bo
...
 
+	if ips, err := net.LookupIP(b.BaseURL.Host); err != nil {
+		if dnsErr, ok := err.(*net.DNSError); ok && dnsErr.IsNotFound {
+			return nil, fi.NewTryAgainLaterError("kops-controller DNS not setup yet")
+		}
+		return nil, err
+	} else {
+		for _, ip := range ips {
+			if ip.String() == cloudup.PlaceholderIP {
+				return nil, fi.NewTryAgainLaterError("kops-controller DNS not setup yet")
```
justinsb, line 170: Probably also a good idea to differentiate the two messages e.g. ("kops-controller DNS not setup yet (placeholder IP found)" vs "kops-controller DNS not setup yet (error was %v)"

rifelpet: Good idea, it turns out `URL.Host` includes the port number which will never resolve. Updating it to `URL.Hostname()` seems to have shaved off a couple minutes on the `kops validate cluster` time in this small sample size of presubmit jobs.

olemarkus (approved):

PR #11102, opened 2021-03-21
[DigitalOcean] Fix DO Tag issue

srikiz:
We add tags to the droplets in order to match the correct volume that needs to be attached. The volumes for etcd-main and etcd-events are also tagged and the etcd-manager does this mapping.

Due to this [change](https://github.com/kubernetes/kops/commit/31f8cbd571964f19d3c31024ddba918998d29929), this was breaking DO droplets to volume matching logic and the new clusters wasn't coming up.

FYI - @rifelpet @timoreimann 

I can work on a better logic going further, but would need this change to go in so I can continue to test the e2e tests.

Please let me know your thoughts, thanks !

### DIFF
--- a/pkg/model/domodel/droplets.go
+++ b/pkg/model/domodel/droplets.go
@@ -71,7 +71,9 @@ func (d *DropletBuilder) Build(c *fi.ModelBuilderContext) error {
 
 		if ig.IsMaster() {
 			masterIndexCount++
-			clusterTagIndex := do.TagKubernetesClusterIndex + ":" + strconv.Itoa(masterIndexCount)
+			// create tag based on etcd name. etcd name is now prefixed with etcd-
+			// Ref: https://github.com/kubernetes/kops/commit/31f8cbd571964f19d3c31024ddba918998d29929
+			clusterTagIndex := do.TagKubernetesClusterIndex + ":" + "etcd-" + strconv.Itoa(masterIndexCount)
 			droplet.Tags = append(droplet.Tags, clusterTagIndex)
 			droplet.Tags = append(droplet.Tags, clusterMasterTag)
 			droplet.Tags = append(droplet.Tags, digitalocean.TagKubernetesInstanceGroup+":"+ig.Name)


rifelpet:
Ah, nice find! I believe we only prepend `etcd-` if the etcd member name starts with a digit. Should we add that same logic here as well?

cc @hakman

hakman:
@rifelpet @srikiz We should keep the logic somehow, otherwise it will match only some of the cases.

srikiz:
@hakman @rifelpet - In the case of DO, we don't actually have an availability zone. All the master droplets will be deployed in the same region. (example for torento region is **tor1**). So for a master count of 3, etcd-cluster names that gets into the trim function would be `master-1-tor1-1`, `master-1-tor1-2`, `master-1-tor1-3`. And the output would be "etcd-1", "etcd-2", "etcd-3". Previously the output was always "1" or "2" or "3" based on the number of masters.

For the test case examples as mentioned [here](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster_test.go#L53) it always results in the below output based on the above scenario. So I think it's very safe to assume the name as "etcd-1" or "etcd-2" etc. in DO case. 

```
		{
			Input:  []string{"us-test-1a-1", "us-test-1a-2", "us-test-1a-3"},
			Output: []string{"etcd-1", "etcd-2", "etcd-3"},
		},
```

Please let me know if this helps..
FYI - @timoreimann

hakman:
@srikiz Can it be a single master droplet? How does it look then?

srikiz:
@hakman - for a single master cluster it looks like below.
`
 etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-tor1-1
      name: etcd-1
    memoryRequest: 100Mi
    name: main
  - cpuRequest: 100m
    etcdMembers:
    - instanceGroup: master-tor1-1
      name: etcd-1
    memoryRequest: 100Mi
    name: events
`

For a multi-master cluster, it looks as below - 
`etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-tor1-1
      name: etcd-1
    - instanceGroup: master-tor1-2
      name: etcd-2
    - instanceGroup: master-tor1-3
      name: etcd-3
    memoryRequest: 100Mi
    name: main
  - cpuRequest: 100m
    etcdMembers:
    - instanceGroup: master-tor1-1
      name: etcd-1
    - instanceGroup: master-tor1-2
      name: etcd-2
    - instanceGroup: master-tor1-3
      name: etcd-3
    memoryRequest: 100Mi
    name: events`

hakman:
Then I think the change should be ok. What do you think @rifelpet ?

rifelpet (approved):
agreed. lets give it a shot 👍🏻

PR #11104, opened 2021-03-22
WIP: Use CertificateSigningRequests when requesting kops-controller certificates

justinsb:
This is slightly more secure, because we prove that we control the private key.

Also introduce a structured deprecation tracker.


johngmyers:
The CSR doesn't prove the private key is owned by the node because the signature doesn't cover  the node name that is going into the certificate. Any old CSR for that key would work.

PR #11042, opened 2021-03-13
Create an environment file for kops-configuration systemd process

bharath-123:

PR #11114, opened 2021-03-22
Load env vars from file for kops-configuration service

hakman:
Refs: #11042
Fixes: #11113


rifelpet:
Im curious why our e2e testing didn't catch this.

rifelpet:
And/or if we could add a unit test that verifies the behavior to make it easier to catch going forwards

rifelpet (approved):

PR #11118, opened 2021-03-23
future proof AWS endpoint and IAM Prefix generation to handle new AWS…

guydog28:
IAM generations fixes for non-standard aws partitions and future new ones. Offload to AWS SDK.

https://github.com/kubernetes/kops/issues/11074


guydog28:
starting diff branch without bad commits

PR #11119, opened 2021-03-23
replace hard coded aws region checks with aws sdk calls

guydog28:
future proof AWS endpoint and IAM Prefix generation to handle new AWS Partitions and future new ones. Offload to AWS SDK

### DIFF
--- a/pkg/model/BUILD.bazel
+++ b/pkg/model/BUILD.bazel
@@ -52,6 +52,7 @@ go_library(
         "//util/pkg/architectures:go_default_library",
         "//util/pkg/mirrors:go_default_library",
         "//util/pkg/vfs:go_default_library",
+        "//vendor/github.com/aws/aws-sdk-go/aws/endpoints:go_default_library",
         "//vendor/github.com/blang/semver/v4:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/runtime:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/util/net:go_default_library",
@@ -68,6 +69,7 @@ go_test(
     srcs = [
         "bootstrapscript_test.go",
         "firewall_test.go",
+        "iam_test.go",
     ],
     data = glob(["tests/**"]),  #keep
     embed = [":go_default_library"],
--- a/pkg/model/iam.go
+++ b/pkg/model/iam.go
@@ -21,6 +21,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/dns"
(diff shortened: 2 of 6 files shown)


guydog28:
I signed it

guydog28:
I signed it

guydog28:
I didn't want to muddy up this pull request, but additionally there are partitions whose regions are not represented in upup/pkg/fi/cloud.go (zonesToCloud).  Why not generate this list from the aws-sdk-go?  This way as new regions are reported, if the sdk continues to be updated, the new regions will be available to kops users.

hakman:
Thanks @guydog28.
/ok-to-test

hakman:
@guydog28 Please squash and make the commit message shorter. Will make it easier to track in the future.

rifelpet:
Regarding `zonesToCloud`, I agree that using the SDK to fetch an AZ list would be preferable, my concern would be that at the time we use that function, we don't necessarily know which cloud provider the `kops create cluster` is for, so we'd have to invoke each one and deal with credential failures since the user likely doesn't have credentials setup for cloud providers they're not using.

If we can figure out a lightweight way to do that for any providers that support providing this info via API, I would support that effort 👍🏻 

That mapping only exists because we support inferring the cloud provider from the provided --zones. I would strongly prefer that we just make --cloud required and drop this map entirely, but I'm not sure how others feel.

hakman:
> That mapping only exists because we support inferring the cloud provider from the provided --zones. I would strongly prefer that we just make --cloud required and drop this map entirely, but I'm not sure how others feel.

I agree with this.

guydog28:
@hakman do i need to create a new branch/pr to do that?  or is there an easy way to do that here - this is the first time I'm actally submitting a PR instead of just code snippets in a comment.

guydog28:
> Regarding `zonesToCloud`...
> 
> If we can figure out a lightweight way to do that for any providers that support providing this info via API, I would support that effort 👍🏻
> 
> That mapping only exists because we support inferring the cloud provider from the provided --zones. I would strongly prefer that we just make --cloud required and drop this map entirely, but I'm not sure how others feel.

@hakman @rifelpet If it just to make a guess, why not remove the map altogether and use a few regexs in the `GuessCloudForZone` func?

```
aws : = /^[a-z]{2}-[a-z\-]+?-[0-9]{1,2}[a-z]$/
gce := /^[a-z]+?-[a-z]+?[0-9]{1,2}-[a-z]$/
do := /^[a-z]{3}[0-9]+$/
ali := /^[a-z]{2}-[a-z]+-[a-z]+$/
```

Just test it against each regex to guess the provider instead of that big list?

guydog28:
I have added the changes to cloud.go to this PR as well, as they seem pertinent to the goal of it.

guydog28:
@hakman I'm not sure the release process, but if this is approved, and there is a 1.19.2, would love to have this cherry picked into that release.

guydog28:
> I have added the changes to cloud.go to this PR as well, as they seem pertinent to the goal of it.

Additionally, I took the old hardcoded zones and used them to create a unit test to test the regex I created to be sure this method still returns the same results.

hakman:
@guydog28 Let's keep the 2 things separate. Please create a separate PR for the zones thing and we can discuss there.
The partitions thing seems reasonable, the only thing missing would be some unit tests for the changed functions, if you don't mind. It also seems innocent enough to cherry-pick to 1.19, right @rifelpet?

rifelpet:
Yes I agree

guydog28:
@hakman ok I pulled out the cloud.go into its own PR, and added tests for the changes.

hakman:
Could you also update the commit message and PR title to match the content?

hakman (approved):

PR #11126, opened 2021-03-24
[kubelet] Remove deprecated flag allowPrivileged

dntosas:
Kubelet has removed the support of allowPrivileged flag since version
1.15 of Kubernetes. As a result, utilizing this flag will make both
master and regular nodes unable to join the cluster as they will stale
on Kubelet starting.

In this commit, we completely remove this flag to prevent users from
using it and introduce disturbance on their clusters.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -2233,10 +2233,6 @@ spec:
               kubelet:
                 description: KubeletConfigSpec defines the kubelet configuration
                 properties:
-                  allowPrivileged:
-                    description: AllowPrivileged enables containers to request privileged
-                      mode (defaults to false)
-                    type: boolean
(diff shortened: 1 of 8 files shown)


hakman:
Thanks @dntosas! :)

olemarkus:
I am not sure we can actually remove any fields from the API. But what we can do is to block users from setting this through API validation (`pkg/apis/kops/validation`). That is also nice as it alerts users they are doing something incorrectly with a "user friendly" error. 

Removing the code referencing the API is certainly fine. Support for k8s 1.14 has been removed from kops 1.21 so we don't need it.

johngmyers:
Apimachinery rules prohibit us from removing fields from an API version. So putting in an API validation check is the best we can do until we create a v1beta1 API.

johngmyers:
A PR to add an API validation preventing this field from being explicitly set to false would be welcomed.

PR #11127, opened 2021-03-24
Validate that kube-apiserver has the necessary authz modes set

### LINKED ISSUE #11014: kubeAPIServer authorizationMode still requires adding Node,RBAC on k8s 1.19.x
 this information.**
1.19.1
 a `kops` flag.**
1.19.8
AWS
rolling update from k8s 1.18.13 to 1.19.8
First master node didn't come up
First master node comes up
**7. Anything else do we need to know?**
Issue was
```
  kubeAPIServer:
     authorizationMode: RBAC
```
had to be changed to
```
  kubeAPIServer:
     authorizationMode: Node,RBAC
```
The rolling update then worked. This slack thread ( [https://kubernetes.slack.com/archives/C3QUFP0QM/p1614880910266900?thread_ts=1614872952.265000&cid=C3QUFP0QM](url) ) indicates that for k8s >= 1.19 Node should be auto-included per [ https://github.com/kubernetes/kops/blob/master/pkg/model/components/apiserver.go#L89](url) but that's clearly not happening.

olemarkus:
Fixes #11014

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -454,12 +454,33 @@ func validateKubeAPIServer(v *kops.KubeAPIServerConfig, c *kops.Cluster, fldPath
 		}
 	}
 
-	if v.AuthorizationMode != nil && strings.Contains(*v.AuthorizationMode, "Webhook") {
-		if v.AuthorizationWebhookConfigFile == nil {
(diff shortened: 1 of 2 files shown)


hakman (approved):
Nice. Please create the cherry-picks when you get the time.

PR #11135, opened 2021-03-24
use regex instead of hard-coded incomplete zones to guess cloud provider

guydog28:
@hakman @rifelpet created to continue the conversation from https://github.com/kubernetes/kops/pull/11119

### DIFF
--- a/upup/pkg/fi/cloud.go
+++ b/upup/pkg/fi/cloud.go
@@ -17,6 +17,8 @@ limitations under the License.
 package fi
 
 import (
+	"regexp"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/kops/dnsprovider/pkg/dnsprovider"
 	"k8s.io/kops/pkg/apis/kops"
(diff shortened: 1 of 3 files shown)


upup/pkg/fi/cloud.go
```diff
@@ -63,371 +65,43 @@ type SubnetInfo struct {
...
-
 // GuessCloudForZone tries to infer the cloudprovider from the zone name
 // Ali has the same zoneNames as AWS in the regions outside China, so if use AliCloud to install k8s in the regions outside China,
 // the users need to provide parameter "--cloud". But the regions inside China can be easily identified.
 func GuessCloudForZone(zone string) (kops.CloudProviderID, bool) {
-	c, found := zonesToCloud[zone]
-	return c, found
+
+	structuredProviders := map[kops.CloudProviderID]string{
+		kops.CloudProviderAWS: "^[a-z]{2}-[a-z\\-]+?-[0-9]{1,2}[a-z]$",
```
bharath-123, line 74: ap-south-1a and ab-south-1a will both match in this case. I am assuming there will be further checks validating the zone and region specified?

guydog28: My understanding is that this method tries to guess the cloud provider when --cloud has not been provided, but --zones has been provided.  Later, when it actually tries to communicated with aws it will error that ab-south-1a does not exist.  

I think the _main_ drawback of this approach is the regex for azure being so generic.  It will assume azure instead of asking the user to specify a cloud provider in the case that none of the other providers match and the zone is alphanumeric.  So it could guess Azure too often.  But again, if that is wrong, the user can specify --cloud with the correct provider.  

This may be a good temporary solution to deprecating allowing --cloud to be optional.  This goes away altogether if --cloud becomes required.

PR #11140, opened 2021-03-25
Add scaleDownDelayAfterAdd to clusterAutoscaler spec

### LINKED ISSUE #11117: cluster-autoscaler addon: allow specifying scaleDownDelayAfterAdd parameter
After the upgrade to kOps 1.19 we’ve switched over to the cluster-autoscaler addon instead of the Helm chart, however we currently have a need to adjust the `scale-down-delay-after-add` parameter, but the configuration options for the addon don’t allow that.

I'd like to see the possibility to change this (and possibly other relevant) parameter(s) of the cluster-autoscaler addon.

jurriaanpro:
Resolves #11117

Added some parameters to the docs that weren't specified before as well.

### DIFF
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -33,8 +33,14 @@ Cluster autoscaler can be enabled to automatically adjust the size of the kubern
 spec:
   clusterAutoscaler:
     enabled: true
+    expander: least-waste
+    balanceSimilarNodeGroups: false
+    scaleDownUtilizationThreshold: 0.5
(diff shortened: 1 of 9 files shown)


Jonathan1:
Awesome change! Hopefully this gets reviewed and added soon 🎉 😎

olemarkus (approved):
Thanks. Looks good to me.

Thanks for improving the docs as well.

PR #11150, opened 2021-03-26
Update Getting Started AWS guide

allir:
Fix some markdown linting and also fix headings.

The "Creating your first cluster" heading was top level heading so it breaks the right hand side table of contents as it will stop listing the headings after that point.


olemarkus (approved):
Thanks for cleaning this up!

PR #11155, opened 2021-03-27
Refactor kubeconfig generation

justinsb:
Tidying up the existing logic.

I think there's more that can be done here - I'm not sure the
innermost Config layer simplifies the code any more.  But one step at
a time!


justinsb:
Internal cleanup only (at the moment).

/milestone 1.22

PR #11160, opened 2021-03-30
Add an option to skip NTP installation

kenji-cloudnatix:
Add NTPConfig to ClusterSpec. NTPConfig has the SkipInstall option.

https://github.com/kubernetes/kops/issues/9661


pkg/apis/kops/ntpconfig.go
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kops
+
+// NTPConfig is the configuration for NTP.
+type NTPConfig struct {
+	// SkipInstall is set to true when we prevent kops from installing and modifying NTP.
+	SkipInstall bool `json:"skipInstall,omitempty"`
```
hakman, line 22: I think we were planning to move away from `SkipInstall` and add `Enabled` and/or `Managed`:
`Enabled` = false would mean the service is actually disabled by the task.
`Managed` = false would just skip the task.

/cc @justinsb @rifelpet @olemarkus

kenji-cloudnatix: Thanks for the review, @hakman ! I've added `Managed`. I wasn't sure we want to have `Enabled` for NTP since I believe either `systemd-timesyncd` or `chrony` is needed in every env. Does this make sense?

pkg/apis/kops/v1alpha2/ntpconfig.go
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+package v1alpha2
+
+// NTPConfig is the configuration for NTP.
+type NTPConfig struct {
+	// Managed is set to true if the NTP configuration is managed by kops.
+	// The NTP configuration task is skilled if this is set to false.
+	Managed bool `json:"managed,omitempty"`
```
hakman, line 23: In case `Managed` is null, should be ignored (treated as true). Only on false it should skip the task.
```suggestion
	Managed *bool `json:"managed,omitempty"`
```

pkg/apis/kops/v1alpha2/ntpconfig.go
```diff
@@ -0,0 +1,24 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+// NTPConfig is the configuration for NTP.
+type NTPConfig struct {
+	// Managed is set to true if the NTP configuration is managed by kops.
+	// The NTP configuration task is skilled if this is set to false.
```
hakman, line 22:
```suggestion
	// The NTP configuration task is skipped if this is set to false.
```

pkg/apis/kops/v1alpha2/ntpconfig.go
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+// NTPConfig is the configuration for NTP.
+type NTPConfig struct {
+	// Managed is set to true if the NTP configuration is managed by kops.
```
hakman, line 21:
```suggestion
	// Managed controls if the NTP configuration is managed by kOps.
```

hakman (approved):

pkg/apis/kops/ntpconfig.go
hakman: 👍

hakman (changes requested):
Please see my inline comments.
Approved this prematurely by mistake. Sorry.

kenji-cloudnatix:
Sure, @hakman ! Applied your comments.

hakman:
Thanks @kenji-cloudnatix. I appreciate the nudge on this.
@rifelpet Would you mind taking a final look?
/approve

rifelpet:
Agreed this looks good to me but could use a section in https://kops.sigs.k8s.io/cluster_spec/

I was debating whether it would make sense to have this be a property of the instance group rather than the cluster, given that it is used by nodeup. I can't think of legitimate use-cases where a cluster might need node-specific NTP configurations though. Maybe others can?

kenji-cloudnatix:
Sure, @rifelpet ! updated https://kops.sigs.k8s.io/cluster_spec/ 

Please let me know the NTP config should be somewhere level.. Thanks!

rifelpet (approved):

PR #11161, opened 2021-03-31
'sv' may have 'nil' or other unexpected value 

Akiros001:
'sv' may have 'nil' or other unexpected value


hakman (changes requested):

pkg/model/context.go
```diff
@@ -348,8 +348,10 @@ func (m *KopsModelContext) KubernetesVersion() semver.Version {
 	if err != nil {
 		klog.Fatalf("unable to determine kubernetes version from %q", kubernetesVersion)
 	}
-
-	return *sv
+	if sv != nil {
+		return *sv
+	}
+	return semver.Version{}
```
hakman, line 354:
```suggestion
	return *sv
```

hakman (approved):

PR #11169, opened 2021-04-02
Filter kOps NatGateways from route table

zetaab:
current solution is not filtering multiple natgateways from route table. We do have use-case where we have multiple natgateways. So when I add filtering to this function, it will only manage kops natgateways and not others.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/natgateway.go
+++ b/upup/pkg/fi/cloudup/awstasks/natgateway.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/klog/v2"
+	raws "k8s.io/kops/pkg/resources/aws"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
 	"k8s.io/kops/upup/pkg/fi/cloudup/cloudformation"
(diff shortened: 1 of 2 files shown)


upup/pkg/fi/cloudup/awstasks/natgateway.go
```diff
@@ -209,19 +209,36 @@ func findNatGatewayFromRouteTable(cloud awsup.AWSCloud, routeTable *RouteTable)
...
 
-			if len(natGatewayIDs) == 0 {
-				klog.V(2).Infof("no NatGateway found in route table %s", *rt.RouteTableId)
-			} else if len(natGatewayIDs) > 1 {
-				return nil, fmt.Errorf("found multiple NatGateways in route table %s", *rt.RouteTableId)
+				for _, tag := range gw.Tags {
+					if fi.StringValue(tag.Key) == awsup.TagClusterName && fi.StringValue(tag.Value) == clusterName {
+						filteredNatGateways = append(filteredNatGateways, gw)
+						break
+					}
```
hakman, line 233: I believe the new way to detect if a component belongs to the cluster is to check for `"kubernetes.io/cluster/" + clusterName`:
https://github.com/kubernetes/kops/blob/97869057129b30ea284c3ed1bdf1db36e752701d/pkg/resources/aws/tags.go#L27

upup/pkg/fi/cloudup/awstasks/natgateway.go
```diff
@@ -209,19 +209,36 @@ func findNatGatewayFromRouteTable(cloud awsup.AWSCloud, routeTable *RouteTable)
...
 				if route.NatGatewayId != nil {
 					natGatewayIDs = append(natGatewayIDs, route.NatGatewayId)
 				}
 			}
+			filteredNatGateways := []*ec2.NatGateway{}
+			for _, natGatewayID := range natGatewayIDs {
+				gw, err := findNatGatewayById(cloud, natGatewayID)
+				if err != nil {
+					return nil, err
+				}
```
hakman, line 227: Because in the past there was no requirement for NatGW to be tagged in any way, one could use a shared VPC without any tags at all. Maybe this filtering logic should be used only when `len(natGatewayIDs) > 1`?

hakman (changes requested):

hakman:
@zetaab Looks good to me, just not sure about the commit message.

zetaab:
@hakman better now?

hakman:
Better, thanks!
Please also cherry-pick to 1.20 if you need it.

hakman (approved):

PR #11170, opened 2021-04-02
Tag hostedzone with the shared tag

olemarkus:
We never own a hostedzone, so we just tag it as shared

### DIFF
--- a/cloudmock/aws/mockroute53/api.go
+++ b/cloudmock/aws/mockroute53/api.go
@@ -30,6 +30,7 @@ type zoneInfo struct {
 	hostedZone *route53.HostedZone
 	records    []*route53.ResourceRecordSet
 	vpcs       []*route53.VPC
+	tags       map[string]string
 }
 
 type MockRoute53 struct {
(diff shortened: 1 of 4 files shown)


rifelpet:
I had debated doing this a while ago. One concern I have is route53's limit of 50 tags per hosted zone and there not being a straight forward to opt-out of this behavior, so a HZ could be used by a max of 50 clusters (less if it has other tags).

Maybe this is a question for users, but I'm curious how tags on shared resources created with something like terraform are handled. Do users have their IAC solution ignore shared tags that kops adds for each cluster? Or do they explicitly add the shared tag themselves for each cluster that will use the resource?

olemarkus:
I wasn't aware of that constraint. That is a good point.

A motivation for doing this was that I wanted to use `aws-zone-tags` as filter for `external-dns`. Using the cluster name as filter have at least had some issues in the past, and hosted zone ID isn't that easy to get hold of automatically.

rifelpet:
Are we applying the external-dns manifest as an addon? could we create a template function that returns the hosted zone ID from the cluster spec and use that in the container's args?

olemarkus:
There is an external-dns addon, where the idea seems to be it to replace dns-controller. I have been looking into what it takes to finish that work, and it is not impossible. it requires some more additions to external-dns first.

But yes, deploying external-dns this way would be ideal.

olemarkus:
I'll close this one then. Seems like this could potentially break stuff

PR #11172, opened 2021-04-02
Adding a "quick fix" section

lboix:


olemarkus:
Thanks for the suggestion. However, the affected stable versions are fairly old. The only correct solution to this is to upgrade kOps.

hakman:
Thanks @lboix, but as @olemarkus said, it is not something that we want to add back to the docs. We had the workaround in the past as part of the advisory and cluster operators had various issues, depending on kOps version.  We prefer to recommend either upgrading kOps or etcd-manager.

PR #11173, opened 2021-04-03
correct a wrong word

zhangshj-inspur:
correct a wrong word


olemarkus:
This change is with a vendored file. Please file an issue with the actual project instead: https://github.com/Masterminds/goutils/

PR #11088, opened 2021-03-20
Enable use of irsa for aws load balancer controller

olemarkus:
This also improves how we can add irsa data to a manifest.

PR #11175, opened 2021-04-04
Add integration test for aws lb controller

olemarkus:
/cc @rifelpet


rifelpet:
I don't see a role + policy being created for lb controller, only dns-controller. Is that expected?

olemarkus:
Yep. This test is the "before" version of the PR. Should be easier to see the diff easier once this is merged and #11088 has been rebased.

rifelpet (approved):

PR #11178, opened 2021-04-06
Kubetest2 - Add flag to expose cluster validation wait time

rifelpet:
The flatcar jobs are failing because the OS performs package updates and a reboot after initial launch, even with the latest AMI.
This causes the cluster to timeout on its validation.

Exposing a flag will allow us to conditionally extend the validation for the flatcar tests

See https://testgrid.k8s.io/kops-distro-flatcar


rifelpet:
xref: https://github.com/kubernetes/test-infra/pull/21667

hakman (approved):

PR #11180, opened 2021-04-06
Ensure anonymousAuth is disabled if authorizationMode is set.

olemarkus:
If anonymousAuth is enabled on the kubelet on k8s 1.19, commands like `kubectl logs` stop working. This seems to be due to the change of defaults we do for authorizationMode.


olemarkus:
/hold
autonomeousAuth defaults to true so need to change some checks here.

Also, the problem is really that if autonomeousAuth is disabled, we do not use the node system account for authenticating. These are really two different things. Will do another PR that will always use secure node auth as of k8s 1.21.

olemarkus:
The original problem was solved by #11185 

/close

PR #6828, opened 2019-04-25
Add GCE NatGateway task

ttousai:
This PR adds a NatGateway task for GCE as part of a larger effort to improve private networking in GCP.
It also updates the Google API Go client to a version which has Google Cloud Router support.
cc @justinsb

Issue #6827, opened 2019-04-25
Add GCE Router task


ttousai:
To support upcoming improvements to GCP networking. The PR is #6828.

PR #11184, opened 2021-04-07
Add GCE Router task

kenji-cloudnatix:
This commit picks up the change from the previous attempt
(https://github.com/kubernetes/kops/pull/6828).

- Add Router to GCE tasks
- Add the `HasExternalIP` field to `InstanceTemplate`
- Create a `Router task and set `HasExternalIP` to `false` when a private topology is specified.

https://github.com/kubernetes/kops/issues/6827


rifelpet:
this looks good to me! Could we add an integration test for this? it may require some changes in the cloudmock directory, I can provide assistance for that if needed.

kenji-cloudnatix:
Thanks! Let me take a look. Will get back if I have any question!

kenji-cloudnatix:
Hi @rifelpet , took a while but added an integration test with [a separate PR](https://github.com/kubernetes/kops/pull/11291). Does the direction look good to you?

hakman:
@kenji-cloudnatix I think for this should be enough to add a TF output like:
https://github.com/kubernetes/kops/tree/master/tests/integration/update_cluster/minimal_gce

kenji-cloudnatix:
Sounds good! Just updated. @hakman ,  @rifelpet , could you take a look again?

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
```diff
@@ -133,6 +140,17 @@ func (e *InstanceTemplate) Find(c *fi.Context) (*InstanceTemplate, error) {
...
 			}
+
+			acs := ni.AccessConfigs
+			if len(acs) > 0 {
+				if len(acs) != 1 {
+					return nil, fmt.Errorf("unexpected number of access configs: %d", len(acs))
+				}
+				if acs[0].Type != accessConfigOneToOneNAT {
+					return nil, fmt.Errorf("unexpected access type: %s", acs[0].Type)
+				}
```
hakman, line 151: Maybe mention InstanceTemplate name in error messages.

hakman (changes requested):
Thanks @kenji-cloudnatix. Just a small nit, other than that looks good.
/lgtm

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
kenji-cloudnatix: done!

kenji-cloudnatix (commented):
Thanks for the review!

hakman (approved):

PR #11185, opened 2021-04-07
Always secure api -> kubelet communication

olemarkus:
Without secure node auth enabled, commands like `kubectl logs` may fail with certain configurations.

As of kubernetes 1.11, we explicitly disable anonymousauth on kubelet for new clusters. We also always disable AlwaysAllow auth mode for kubelet as of 1.19. This combination works for new clusters and those who have already disabled anonymousAuth for kubelet, which accounts for most users, it seems.

But old clusters upgrading from pre 1.11 face issues with commands like `kubectl logs` because for these clusters, apiserver is anonymous and anonymous users are not allowed to talk to kubelet per our RBAC rules.

The logic for this is 
```
// UseSecureKubelet checks if the kubelet api should be protected by a client certificate.
func (c *NodeupModelContext) UseSecureKubelet() bool {
	return c.NodeupConfig.KubeletConfig.AnonymousAuth != nil && !*c.NodeupConfig.KubeletConfig.AnonymousAuth
}
```
anonymousAuth and secure kubelet comms are two different things as far as I can tell. So not sure why we made the check this way. Checking for `AuthorizationMode="" or AlwaysAllow` would make more sense.

This PR enables secure comms unconditionally

### DIFF
--- a/nodeup/pkg/model/context.go
+++ b/nodeup/pkg/model/context.go
@@ -397,11 +397,6 @@ func (c *NodeupModelContext) UseBootstrapTokens() bool {
 	return c.Cluster.Spec.Kubelet != nil && c.Cluster.Spec.Kubelet.BootstrapKubeconfig != ""
 }
 
-// UseSecureKubelet checks if the kubelet api should be protected by a client certificate.
-func (c *NodeupModelContext) UseSecureKubelet() bool {
-	return c.NodeupConfig.KubeletConfig.AnonymousAuth != nil && !*c.NodeupConfig.KubeletConfig.AnonymousAuth
-}
-
 // KubectlPath returns distro based path for kubectl
 func (c *NodeupModelContext) KubectlPath() string {
 	kubeletCommand := "/usr/local/bin"
--- a/nodeup/pkg/model/kube_apiserver.go
+++ b/nodeup/pkg/model/kube_apiserver.go
@@ -107,19 +107,16 @@ func (b *KubeAPIServerBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
-	// @check if we are using secure client certificates for kubelet and grab the certificates
-	if b.UseSecureKubelet() {
-		issueCert := &nodetasks.IssueCert{
-			Name:    "kubelet-api",
-			Signer:  fi.CertificateIDCA,
-			Type:    "client",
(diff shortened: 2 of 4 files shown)


nodeup/pkg/model/context.go
```diff
@@ -399,7 +399,11 @@ func (c *NodeupModelContext) UseBootstrapTokens() bool {
 
 // UseSecureKubelet checks if the kubelet api should be protected by a client certificate.
 func (c *NodeupModelContext) UseSecureKubelet() bool {
-	return c.NodeupConfig.KubeletConfig.AnonymousAuth != nil && !*c.NodeupConfig.KubeletConfig.AnonymousAuth
+	if c.IsKubernetesGTE("1.21") {
+		return true
```
seh, line 403: I don't see much logging elsewhere in this file, but should we warn users if they had enabled anonymous authentication, and we're overriding their intention here?

nodeup/pkg/model/context.go
```diff
@@ -399,7 +399,11 @@ func (c *NodeupModelContext) UseBootstrapTokens() bool {
 
 // UseSecureKubelet checks if the kubelet api should be protected by a client certificate.
```
seh, line 400:
```suggestion
// UseSecureKubelet checks if the kubelet API should be protected by a client certificate.
```

nodeup/pkg/model/context.go
```diff
@@ -399,7 +399,11 @@ func (c *NodeupModelContext) UseBootstrapTokens() bool {
...
 func (c *NodeupModelContext) UseSecureKubelet() bool {
-	return c.NodeupConfig.KubeletConfig.AnonymousAuth != nil && !*c.NodeupConfig.KubeletConfig.AnonymousAuth
+	if c.IsKubernetesGTE("1.21") {
+		return true
+	} else {
+		return c.NodeupConfig.KubeletConfig.AnonymousAuth != nil && !*c.NodeupConfig.KubeletConfig.AnonymousAuth
+	}
 }
 
 // KubectlPath returns distro based path for kubectl
```
seh, line 409:
```suggestion
// KubectlPath returns a distribution-based path for kubectl.
```

nodeup/pkg/model/context.go
seh: Ah, I just saw #11180, which I think takes care of catching that discrepancy.

hakman (approved):

PR #11186, opened 2021-04-08
Update kops_create_secret_dockerconfig.md

integrii:
Today we were implementing an authenticated docker user, but it was unclear how exactly to do that.  We learned that simply making this secret within kops was all that was needed for the docker config to start appearing on newly built nodes.  It would be nice if the documentation here reflected that.  It would have saved us some time.

### DIFF
--- a/cmd/kops/create_secret_dockerconfig.go
+++ b/cmd/kops/create_secret_dockerconfig.go
@@ -35,7 +35,12 @@ var (
 	createSecretDockerconfigLong = templates.LongDesc(i18n.T(`
 	Create a new docker config, and store it in the state store.
 	Used to configure docker on each master or node (i.e. for auth)
-	Use update to modify it, this command will only create a new entry.`))
+	Use update to modify it, this command will only create a new entry.
+	
+	After creating a dockerconfig secret, a .docker/config.json file will 
+	be added to all newly created nodes without running kops update.
+	This file should also apply auth to containerd, which is the default
+	container runtime in kops 1.20 and above.`))
 
 	createSecretDockerconfigExample = templates.Examples(i18n.T(`
 	# Create a new docker config.
--- a/cmd/kops/create_secret_dockerconfig.go
+++ b/cmd/kops/create_secret_dockerconfig.go
@@ -37,10 +37,10 @@ var (
 	Used to configure docker on each master or node (i.e. for auth)
 	Use update to modify it, this command will only create a new entry.
 	
-	After creating a dockerconfig secret, a .docker/config.json file will 
-	be added to all newly created nodes without running kops update.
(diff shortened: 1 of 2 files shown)


cmd/kops/create_secret_dockerconfig.go
```diff
@@ -35,7 +35,12 @@ var (
 	createSecretDockerconfigLong = templates.LongDesc(i18n.T(`
 	Create a new docker config, and store it in the state store.
 	Used to configure docker on each master or node (i.e. for auth)
-	Use update to modify it, this command will only create a new entry.`))
+	Use update to modify it, this command will only create a new entry.
+	
+	After creating a dockerconfig secret, a .docker/config.json file will 
+	be added to all newly created nodes without running kops update.
+	This file should also apply auth to containerd, which is the default
+	container runtime in kops 1.20 and above.`))
```
hakman, line 43:
```suggestion
	After creating a dockerconfig secret, a /root/.docker/config.json file will 
	be added to newly created nodes (without running kops update).
	This should also work when using containerd, which is the default
	container runtime in kOps 1.20 and above.`))
```

hakman (changes requested):
Thanks @integrii. Just a small suggestion from me. Please also squash the commits.

cmd/kops/create_secret_dockerconfig.go
olemarkus: Can we say with confidence that it _will_ work? not only _should_ :)

hakman: Sure! 🤣

hakman (approved):

docs/cli/kops_create_secret_dockerconfig.md
```diff
@@ -9,6 +9,8 @@ Create a docker config.
 
 Create a new docker config, and store it in the state store. Used to configure docker on each master or node (i.e. for auth) Use update to modify it, this command will only create a new entry.
 
+ After creating a dockerconfig secret, a .docker/config.json file will be added to all newly created nodes without running kops update. This file should also apply auth to containerd, which is the default container runtime in kops 1.20 and above.
+
```
hakman, line 13:
```suggestion
```

docs/cli/kops_create_secret_dockerconfig.md
```diff
@@ -59,4 +59,3 @@ kops create secret dockerconfig [flags]
 ### SEE ALSO
 
 * [kops create secret](kops_create_secret.md)	 - Create a secret.
```
hakman, line 61:
```suggestion
* [kops create secret](kops_create_secret.md)	 - Create a secret.

```

hakman (approved):

PR #11189, opened 2021-04-09
Update cilium.md

recollir:
Link to Cilium doc about NodePort resulted in a 404.


olemarkus (approved):
Thanks for sending this!

PR #11194, opened 2021-04-09
Make IRSA more configurable

olemarkus:
* Introduces IRSA in the spec
 * Re-enables ability to use IRSA with the APIServer
 * Replaces various feature flag checks with spec checks


pkg/apis/kops/cluster.go
```diff
@@ -205,6 +205,17 @@ type ClusterSpec struct {
 	RollingUpdate *RollingUpdate `json:"rollingUpdate,omitempty"`
 	// ClusterAutoscaler defines the cluster autoscaler configuration.
 	ClusterAutoscaler *ClusterAutoscalerConfig `json:"clusterAutoscaler,omitempty"`
+	// IAMRolesForServiceAccount defines the IRSA configuration.
+	IAMRolesForServiceAccounts *IAMRolesForServiceAccountsConfig `json:"iamRolesForServiceAccounts,omitempty"`
```
rifelpet, line 209: I'm wondering if we should merge these fields into IAMSpec or if they're worth keeping separate.

rifelpet (commented):
Even if we use PublicDataStore, we'll still set the oidc apiserver flags correct? because they're required now that the feature went GA. I think this makes sense overall and appreciate the per-addon granularity. I'd like to have Justin take a look too

pkg/apis/kops/validation/legacy.go
```diff
@@ -394,10 +394,61 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+		switch irsa.OIDCLocation {
+		case "", "PublicDataStore":
+			if c.Spec.PublicDataStore == "" {
+				allErrs = append(allErrs, field.Required(fieldSpec.Child("publicDataStore"), "OIDC location set to PublicDataStore, but publicDataStore is not configured"))
+			}
+		case "APIServer":
+			if c.Spec.KubeAPIServer == nil || !fi.BoolValue(c.Spec.KubeAPIServer.AnonymousAuth) {
+				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubeAPIServer", "anonymousAuth"),
+					"Exposing OIDC metadata on the API Server requires anonmousAuth to be enabled"))
+			}
```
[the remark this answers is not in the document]
olemarkus: We considered, but landed on using s3 I believe. It became complicated with adding additional loadbalancers and such. Not really too complicated, but s3 was just easier. So I'd say this is on hold for now unless someone is really eager to implement this.

pkg/apis/kops/validation/legacy.go
```diff
@@ -394,10 +394,61 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+			found := false
+			for _, cidr := range c.Spec.KubernetesAPIAccess {
+				if cidr == "0.0.0.0/0" {
+					found = true
+				}
+			}
+			if !found {
+				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubernetesAPIAccess"),
+					"Exposing OIDC metadata on the API Server requires API server to be reachable from any IP"))
+			}
```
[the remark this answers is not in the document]
olemarkus: We only add tasks for AWS, so it makes sense to validate cloudprovider == aws.

For any other clouds, it should be possible to provision this themselves. We may just have to allow for OIDCLocation == SelfManaged or something so that they can enable IRSA on various addons.

pkg/apis/kops/cluster.go
olemarkus: I think it should be its own struct. If not, we'll have to prefix all the fields. The entire struct could be moved to IAMSpec, I suppose. IAMSpec also doesn't really have "a clear purpose".

olemarkus:
Fixed most of the comments now

pkg/apis/kops/cluster.go
```diff
@@ -205,6 +205,27 @@ type ClusterSpec struct {
...
+}
+
+// OIDCLocation determines where OIDC metadata will be stored
+type OIDCLocation string
```
johngmyers, line 213: `OIDCProviderLocation`?

pkg/apis/kops/validation/legacy.go
```diff
@@ -394,10 +394,65 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+
+		switch irsa.OIDCLocation {
+		case kops.OIDCLocationPublicDataStore:
+			if c.Spec.PublicDataStore == "" {
+				allErrs = append(allErrs, field.Required(fieldSpec.Child("publicDataStore"), "OIDC location set to PublicDataStore, but publicDataStore is not configured"))
+			}
+		case kops.OIDCLocationAPIServer:
+			if c.Spec.KubeAPIServer == nil || !fi.BoolValue(c.Spec.KubeAPIServer.AnonymousAuth) {
+				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubeAPIServer", "anonymousAuth"),
+					"Exposing OIDC metadata on the API Server requires anonmousAuth to be enabled"))
```
johngmyers, line 438:
```suggestion
					"Exposing OIDC metadata on the API Server requires anonymousAuth to be enabled"))
```

pkg/apis/kops/validation/legacy.go
```diff
@@ -394,10 +394,65 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
 		}
+
+		if !featureflag.PublicJWKS.Enabled() {
+			allErrs = append(allErrs, field.Forbidden(irsaSpec, "IAMRolesForServiceAccounts requires the PublicJWKS feature flag to be enabled."))
+		}
+
+		switch irsa.OIDCLocation {
+		case kops.OIDCLocationPublicDataStore:
+			if c.Spec.PublicDataStore == "" {
+				allErrs = append(allErrs, field.Required(fieldSpec.Child("publicDataStore"), "OIDC location set to PublicDataStore, but publicDataStore is not configured"))
```
johngmyers, line 433:
```suggestion
				allErrs = append(allErrs, field.Required(fieldSpec.Child("publicDataStore"), "must be configured when OIDC location is set to PublicDataStore"))
```

pkg/model/components/apiserver.go
```diff
@@ -155,8 +155,10 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
 		c.EnableAdmissionPlugins = append(c.EnableAdmissionPlugins, c.AppendAdmissionPlugins...)
 	}
 
-	// We make sure to disable AnonymousAuth
-	c.AnonymousAuth = fi.Bool(false)
+	// We default AnonymousAuth to off
```
johngmyers, line 158: Perhaps have ability to enable `AnonymousAuth` under the `PublicJWKS` feature flag?

pkg/apis/kops/validation/legacy.go
```diff
@@ -394,10 +394,65 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+
+	if irsa != nil && fi.BoolValue(irsa.Enabled) {
+		irsaSpec := fieldSpec.Child("iamRolesForServiceAccounts")
+
+		if kops.CloudProviderID(c.Spec.CloudProvider) != kops.CloudProviderAWS {
+			allErrs = append(allErrs, field.Forbidden(irsaSpec.Child("enabled"), "IAMRolesForServiceAccounts is only supported on AWS."))
 		}
+
+		if !featureflag.PublicJWKS.Enabled() {
+			allErrs = append(allErrs, field.Forbidden(irsaSpec, "IAMRolesForServiceAccounts requires the PublicJWKS feature flag to be enabled."))
```
johngmyers, line 427: I believe `PublicJWKS` should only cover the use of `OIDCLocationAPIServer`. I'm thinking we perhaps want to remove support for `OIDCLocationAPIServer` entirely.

johngmyers (changes requested):

pkg/model/components/discovery.go
```diff
@@ -59,11 +59,14 @@ func (b *DiscoveryOptionsBuilder) BuildOptions(o interface{}) error {
 
 	// We set apiserver ServiceAccountKey and ServiceAccountSigningKeyFile in nodeup
 
-	if useJWKS {
+	if b.IsKubernetesLT("1.20") {
```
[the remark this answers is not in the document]
johngmyers: Why are we unconditionally enabling this feature gate on old k8s releases?

johngmyers:
I'm concerned with defaults. If one sets up publishing to a s3 bucket, then all the various addons should default to using  IRSA instead of the node role. One should not have to change a setting for each subcomponent or addon that needs IAM privileges.

johngmyers:
I'm now wondering why one would legitimately want a per-subcomponent setting in the first place.

rifelpet:
I initially thought having per-subcomponent settings could reduce risk during enabling/disabling of this functionality on existing clusters. Perhaps it is unnecessary though.

johngmyers:
I don't see how it significantly reduces risk. The risks I can think of are:

* The OIDC provider is misconfigured such that AWS doesn't recognize it.
Nothing using IRSA is going to work. The fix is to revert to not using IRSA. We could perhaps allow deploying a custom servieceaccount role without using IRSA for addons so the admin could first deploy a test workload to verify AWS can use the OIDC provider.

* An addon's role is missing a needed permission, so depends on a permission in the instance role.
This would be a kops bug. The fix would be to revert to the previous state until kops is fixed.

* When turning use of IRSA on or off, in the time between apply_cluster removing permissions from the instance role (or deleting the serviceaccount role) and a rolling update updating the addons the addons could break due to not having necessary permissions. We might want an option to put the permissions in both serviceaccount and instance roles.

olemarkus:
For us the challenge would be extra permissions needed on the policies. Like allowing cert manager to do dns verification on other domains. We have a mechanism for that on normal instance roles, but not sure how to implement this in a good way for IRSA roles. 

So we as users would also like this to be enabled per addon.

pkg/apis/kops/validation/legacy.go
```diff
@@ -394,10 +394,65 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+			}
+			found := false
+			for _, cidr := range c.Spec.KubernetesAPIAccess {
+				if cidr == "0.0.0.0/0" {
+					found = true
+				}
+			}
+			if !found {
+				allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubernetesAPIAccess"),
+					"Exposing OIDC metadata on the API Server requires API server to be reachable from any IP"))
```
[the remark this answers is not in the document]
olemarkus: I don't believe these are well-known. I suspect they are too many to be configured on an SG as well.

pkg/model/components/discovery.go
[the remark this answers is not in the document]
olemarkus: Think this is the result of a rebase. It shouldn't be unconditionally.

pkg/apis/kops/cluster.go
```diff
@@ -471,6 +492,8 @@ type ExternalDNSConfig struct {
 	WatchIngress *bool `json:"watchIngress,omitempty"`
 	// WatchNamespace is namespace to watch, defaults to all (use to control whom can creates dns entries)
 	WatchNamespace string `json:"watchNamespace,omitempty"`
+	// UseIRSA determines IRSA is used instead of assuming the control plane role.
```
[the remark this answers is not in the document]
olemarkus: Something in that direction sounds good. `UseCloudIdentity` or `UseCloudIAM` could detach it even more from AWS (they are the only ones using roles I think). The latter would also be inline with the `UseServiceAccountIAM` feature flag. Think @justinsb also had some opinions here.

pkg/apis/kops/cluster.go
```diff
@@ -205,6 +205,27 @@ type ClusterSpec struct {
...
+	OIDCLocationPublicDataStore OIDCProviderLocation = "PublicDataStore"
+	// OIDCProviderLocationNone will not store the OIDC metadata and disable any cloud OIDC provider
+	OIDCProviderLocationNone OIDCProviderLocation = "None"
+)
+
+// InstanceRoleForServiceAccountConfig defines the IRSA configuration.
+type IAMRolesForServiceAccountsConfig struct {
+	// JWKSLocation determines the location of the OIDC metadata.
+	// Valid locations are "APIServer" and "PublicDataStore".
+	OIDCProviderLocation OIDCProviderLocation `json:"oidcLocation,omitempty"`
```
johngmyers, line 228: The OIDC Provider location is not specific to IRSA. The OIDC provider can be consumed by other clients. It should not be in a substruct about IAM.

johngmyers: Whether workloads should consume IAM permissions through serviceaccount or node roles does arguably belong under `IAMSpec`.

pkg/apis/kops/validation/legacy.go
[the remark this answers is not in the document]
johngmyers: One could be deliberately excluding IP ranges known not to be owned by AWS.

pkg/apis/kops/cluster.go
olemarkus: That is fair. But then we probably need to put `enabled` back into this struct as and additionally require `OIDCProviderLocation != None`. This struct is very much about enabling the IAM OIDC provider on AWS (and someone will maybe implement support for workload identity on GCE).

johngmyers:
Perhaps #11016 should be extended to allow adding permissions to the baked-in roles. If one specifies a pre-baked namespace/serviceaccount, the inline policies would be merged.

rifelpet:
Perhaps `spec.additionalPolicies` and/or `spec.externalPolicies` could be extended to support service account roles in addition to node roles?

johngmyers:
Or we could sprinkle `additionalPolicies` fields around the various addon structs like this PR currently sprinkles `useIRSA` fields.

johngmyers:
I think we should pull out a separate PR that only deals with publishing the OIDC provider to VFS. It should touch the consuming-by-IAM code as little as practicable. Publishing the provider to VFS should not be under either the `PublicJWKS` or `UseServiceAccountIAM` feature flags. I am ambivalent about adding a third feature flag.

After that we can move tests to use the new feature instead of PublicJWKS. We could either remove PublicJWKS entirely or leave it a feature flag long term.

olemarkus:
I don't think a third feature flag makes sense for just publishing the OIDC metadata. I don't think that behaviour is likely to change.
I can do that, we just need to agree where `OIDCProviderLocation` belongs. Right now it seems directly in the cluster spec.

johngmyers:
One possibility might be in `AuthenticationSpec`, but that seems to be more around authenticating *to* the API server.

I'm thinking there should be an `OIDCProviderSpec` starting out with a `Location` field.

johngmyers:
A third feature flag would be if we think we might need to make an incompatible change later.

johngmyers:
Possibly `PublicDataStore` should be moved under `OIDCProviderSpec`.

johngmyers:
A subtlety: it should be possible to publish the OIDC provider to a non-public VFS path; the requirement that the path be public applies to enabling provisioning/use of service account IAM roles.

olemarkus:
I have some other use for `PublicDataStore` in mind as well.

In OIDC, the provider tends to be the identity provider, which is technically the control plane. This is more about the OIDC discovery metadata for the service account issuer. K8s docs tend to call this Service Account Issuer Discovery. Something to consider when naming this.

Supporting arbitrary locations should be straightforward. Perhaps something some users want together with external service account issuers.

johngmyers:
I see this substruct as configuring aspects of the cluster's OIDC Provider. Part of that is configuration of the OpenID Provider Configuration document and its location.

olemarkus:
We could drop the OIDC bit entirely and use the k8s namings. If so it would be `ServiceAccountIssuerSpec` struct with a `DiscoveryLocation` field.

johngmyers:
I could go with `ServiceAccountIssuerSpec`.

I'm not quite sure about the contents of the substruct.

The kops behaviors that we need to control somewhere are:

1) The value of the service account issuer URL
2) Whether to publish the discovery document into VFS (which it should just do if (1) derives from VFS)
3) Whether to enable anonymous auth and provision a ClusterRoleBinding enabling anonymous discovery (Which is hat I think the PublicJWKS feature flag is supposed to control, since we only need this behavior to support a conformance test.)
4) Whether to configure AWS to consume the service account issuer URL. 
5) When (4) is yes, whether any particular addon/subcomponent should use either a serviceaccount or instance role.

There is a possible use case where we set the service account issuer URL to something other than the API server or something in VFS. In that case, we would rely on some external process publishing the discovery document to that URL. We don't have to support that use case now, but the spec should be reasonably extensible to cover that case.

What would be the other uses of `PublicDataStore`? Would we want to require all such uses share the same base path? Or could there be a use case for separate public VFS paths?

I'm thinking we have a field for the service account issuer URL. If that field value has a VFS scheme, then we publish the document to that VFS path and set the service account issuer URL to the corresponding HTTPS endpoint. If that field is not specified and the PublicJWKS feature flag is enabled we instead use the public APIServer endpoint.

Is there any downside to just doing (4) when there a service account issuer URL other than the default internal APIServer endpoint?

olemarkus:
> 
> I'm not quite sure about the contents of the substruct.
> 
> The kops behaviors that we need to control somewhere are:
> 
>     1. The value of the service account issuer URL

Today this is in two locations:
1. one can set this explicitly on API server
2. it is auto-configured based on OIDC issuer discovery location

1 overrides 2
 
>     2. Whether to publish the discovery document into VFS (which it should just do if (1) derives from VFS)

I prefer some automation here since there are two files involved (discovery docs and jwks docs). Also VFS location is on the format "s3://" while the format above is "https://". It is abit too automagical for kOps to try to derive https://-urls into a VFS.

>     3. Whether to enable anonymous auth and provision a ClusterRoleBinding enabling anonymous discovery (Which is hat I think the PublicJWKS feature flag is supposed to control, since we only need this behavior to support a conformance test.)

1. it is not needed for conformance. We already pass all conformance tests without these flags using the internal API url as SA issuer internally.
2. The flag does control using the public API domain as issuer.
3. PublicJWKS also controls deploying JWKS docs to VFS

Question is if it really should block 2 and 3. The main reason for the flag is to make API changes easier should we need to, so in that sense, it makes sense that it blocks both as these configurations are related.

>     4. Whether to configure AWS to consume the service account issuer URL.

I think this can be a dedicated flag in the struct. One may want to publish Issuer Discovery docs, but not necessarily set up the trust relationship with AWS.
This flag would then also control if we use IRSA for kOps addons or not?

This flag would also be required for what I called ServiceAccountMappings.

> 
>     5. When (4) is yes, whether any particular addon/subcomponent should use either a serviceaccount or instance role.
> 
> 
> There is a possible use case where we set the service account issuer URL to something other than the API server or something in VFS. In that case, we would rely on some external process publishing the discovery document to that URL. We don't have to support that use case now, but the spec should be reasonably extensible to cover that case.

This would be setting the config directly on the API Server. I believe we already support this today.

> What would be the other uses of `PublicDataStore`? Would we want to require all such uses share the same base path? Or could there be a use case for separate public VFS paths?
> 

Let's forget about this one. We already have fields that refers to various locations within the same VFS (secretstore, configstore etc). I am happy with a specific VFSPath for Issuer Discovery with a more apt name.

> I'm thinking we have a field for the service account issuer URL. If that field value has a VFS scheme, then we publish the document to that VFS path and set the service account issuer URL to the corresponding HTTPS endpoint. If that field is not specified and the PublicJWKS feature flag is enabled we instead use the public APIServer endpoint.
 
Only thing we need to agree on here is if PublicJWKS should control both options or just what you describe.

> Is there any downside to just doing (4) when there a service account issuer URL other than the default internal APIServer endpoint?

We need to do some sanity checking here. E.g we need to ensure audience is properly configured as well. Should we always append `amazonaws.com`?
It seems easier to let users explicitly say they want the AWS trust relationship configured, and then kOps can take care all the right information is included.

If kOps should publish Issuer Discovery docs to VFS for other purposes, we also need to let users specify the audience as it is now hardcoded to `amazonaws.com`.

johngmyers:
What I think the remaining issues might be:

* Whether we can remove the PublicJWKS feature flag (and all behaviors behind it)
  * It permits publishing OIDC discovery on k8s before 1.20
  * It loads a ClusterRoleBinding granting system:service-account-issuer-discovery to system:anonymous
* Whether we can default the UseServiceAccountIAM feature flag to true
  * I think we need to first implement merging the baked-in service account permissions with `iam.ServiceAccountExternalPermissions` for the same service account.
  * So probably not for 1.21

olemarkus:
> What I think the remaining issues might be:
> 
>     * Whether we can remove the PublicJWKS feature flag (and all behaviors behind it)
>       * It permits publishing OIDC discovery on k8s before 1.20

This we should allow. I am guessing it is just a lack of cleanup in the area that blocks this.

>       * It loads a ClusterRoleBinding granting system:service-account-issuer-discovery to system:anonymous

I think it still should block this. I don't want to use the logic we have for infering OIDC issuer discovery on public DNS (anonymous auth + 0.0.0.0/0 source IP) to also deploy these resources. And we may choose to just remove this part and let users deploy the RBAC themselves as there may be other reasons for public API/anonymousAuth (e.g some external monitoring tool that wants to list Deployments :shrug: )

> 
>     * Whether we can default the UseServiceAccountIAM feature flag to true
>       
>       * I think we need to first implement merging the baked-in service account permissions with `iam.ServiceAccountExternalPermissions` for the same service account.
>       * So probably not for 1.21

I think we want the feature gate here. The permissions we grant addons may not be the full set of permissions needed for all use cases and such. We have not done enough testing to see if this works. I'd like to get an e2e going first at least.

johngmyers:
> > What I think the remaining issues might be:
> >   * It permits publishing OIDC discovery on k8s before 1.20
> 
> This we should allow. I am guessing it is just a lack of cleanup in the area that blocks this.

Submitted #11395 for this.

> >   * It loads a ClusterRoleBinding granting system:service-account-issuer-discovery to system:anonymous
> 
> [...] And we may choose to just remove this part and let users deploy the RBAC themselves [...]

This is what I was thinking.

> > * Whether we can default the UseServiceAccountIAM feature flag to true
> >   
> >   * I think we need to first implement merging the baked-in service account permissions with `iam.ServiceAccountExternalPermissions` for the same service account.
> >   * So probably not for 1.21
> 
> I think we want the feature gate here. The permissions we grant addons may not be the full set of permissions needed for all use cases and such.

Which is why we need to implement that merging.

> We have not done enough testing to see if this works. I'd like to get an e2e going first at least.

So probably targeting kops 1.22.

johngmyers:
I think everything in this PR has been mined elsewhere. I have a list of remaining work in the office hours agenda.
/close

PR #11207, opened 2021-04-11
Rename the service account key

johngmyers:


hakman (approved):

johngmyers:
There's some nodeup builder that's still referencing the old name. It's probably something to do with IRSA, but I'm trying to find it.

johngmyers:
It looks to me like the e2e tests are running an old version of nodeup.

PR #11216, opened 2021-04-12
Don't start kubelet if instance is entering the warm pool

olemarkus:
AWS just launched a feature where one can create free cold standbys. These nodes run for an arbitrary amount of second as if they were normal instances, then stop. When an ASG needs to scale up, they boot up again as normal instances. 

This PR prevents kubelet from starting if the instance is entering the warm pool. That prevents the instance from joining the cluster. quickly ending up in NotReady nodes that don't go away because the instances still exists.

I have all the code already for supporting other aspects of this new feature (rolling cluster for example), but splitting across multiple PRs to make review easier.

To test this PR, you can create a new cluster, then manually click the "create warm pool" button in the console.

### DIFF
--- a/nodeup/pkg/model/context.go
+++ b/nodeup/pkg/model/context.go
@@ -42,8 +42,13 @@ import (
 	"github.com/blang/semver/v4"
 )
 
+const (
+	ConfigurationModeWarming string = "Warming"
+)
+
 // NodeupModelContext is the context supplied the nodeup tasks
 type NodeupModelContext struct {
+	Cloud         fi.Cloud
 	Architecture  architectures.Architecture
 	Assets        *fi.AssetStore
 	Cluster       *kops.Cluster
@@ -62,6 +67,10 @@ type NodeupModelContext struct {
 
 	kubernetesVersion semver.Version
 	bootstrapCerts    map[string]*nodetasks.BootstrapCert
+
+	// ConfigurationMode determines if we are prewarming an instance or running it live
+	ConfigurationMode string
+	InstanceID        string
 }
 
 // Init completes initialization of the object, for example pre-parsing the kubernetes version
--- a/nodeup/pkg/model/kubelet.go
+++ b/nodeup/pkg/model/kubelet.go
@@ -293,6 +293,11 @@ func (b *KubeletBuilder) buildSystemdService() *nodetasks.Service {
 
 	service.InitDefaults()
 
+	if b.ConfigurationMode == "Warming" {
+
+		service.Running = fi.Bool(false)
+	}
+
 	return service
 }
 
@@ -443,14 +448,7 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 			instanceTypeName = *b.NodeupConfig.DefaultMachineType
 		}
 
-		region, err := awsup.FindRegion(b.Cluster)
-		if err != nil {
-			return nil, err
-		}
-		awsCloud, err := awsup.NewAWSCloud(region, nil)
-		if err != nil {
-			return nil, err
-		}
+		awsCloud := b.Cloud.(awsup.AWSCloud)
 		// Get the instance type's detailed information.
 		instanceType, err := awsup.GetMachineTypeInfo(awsCloud, instanceTypeName)
 		if err != nil {
--- a/nodeup/pkg/model/kubelet_test.go
+++ b/nodeup/pkg/model/kubelet_test.go
@@ -157,8 +157,34 @@ func Test_RunKubeletBuilder(t *testing.T) {
 		t.Fatalf("error loading model %q: %v", basedir, err)
 		return
 	}
+	runKubeletBuilder(t, context, nodeUpModelContext)
 
-	builder := KubeletBuilder{NodeupModelContext: nodeUpModelContext}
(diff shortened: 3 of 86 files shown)


upup/pkg/fi/nodeup/command.go
```diff
@@ -637,3 +667,23 @@ func getNodeConfigFromServer(ctx context.Context, config *nodeup.ConfigServerOpt
...
+
+	svc := c.Cloud.(awsup.AWSCloud).Autoscaling()
+
+	result, err := svc.DescribeAutoScalingInstances(&autoscaling.DescribeAutoScalingInstancesInput{
+		InstanceIds: []*string{&c.InstanceID},
+	})
+	if err != nil {
+		return "", fmt.Errorf("error describing instances: %v", err)
+	}
+	lifecycle := fi.StringValue(result.AutoScalingInstances[0].LifecycleState)
```
rifelpet, line 681: I haven't checked, but I wonder if the lifecycle state is available via the instance metadata service? That would reduce the per-node API calls and avoid the IAM changes.

olemarkus: It isn't, unfortunately. I passed on a request to AWS to add that.

Same goes for completing lifecycle hooks. See AWS own example here https://github.com/aws-samples/amazon-ec2-auto-scaling-group-examples/blob/main/features/lifecycle-hooks/userdata-managed-linux/template.yaml#L282

olemarkus:
VPC limit exceeded ...
/retest

rifelpet:
> VPC limit exceeded ...

FYI: https://github.com/kubernetes/k8s.io/issues/1921

olemarkus:
Thanks for the info @rifelpet. hoping for a swift resolution :crossed_fingers: 

/retest

olemarkus:
Yay. @rifelpet think this one is good to go?

nodeup/pkg/model/context.go
```diff
@@ -62,6 +67,10 @@ type NodeupModelContext struct {
 
 	kubernetesVersion semver.Version
 	bootstrapCerts    map[string]*nodetasks.BootstrapCert
+
+	// ConfigurationMode determines if we are prewarming an instance or running it live
+	ConfigurationMode string
```
johngmyers, line 72: Perhaps this should be its own type?

nodeup/pkg/model/kubelet.go
```diff
@@ -293,6 +293,11 @@ func (b *KubeletBuilder) buildSystemdService() *nodetasks.Service {
 
 	service.InitDefaults()
 
+	if b.ConfigurationMode == "Warming" {
+
+		service.Running = fi.Bool(false)
```
johngmyers, line 298: Shouldn't we add a test to verify this behavior?

johngmyers (changes requested):

johngmyers:
What happens when a warmed instance is added to the pool? Does nodeup run a second time or does it just start up kubelet?

olemarkus:
There are two modes: one is where the instances are kept running and you pay for them being hot (this one doesn't make much sense to me). The mode we support here they are entering full stopped state and you don't pay for them. When they boot up, it would be the same as a reboot. I.e the `kops-configuration` service is started again. I like this behaviour as nodeup is fairly declarative and we should end up in a known state.

nodeup/pkg/model/context.go
olemarkus: I would have to think about where to put the type to avoid cyclic dependencies. That is why the const is not used close to your other comment.

johngmyers:
I presume you pay for the root volume in stopped state?

I think running nodeup again will result in it asking kops-controller for new certificates. We might want to work on ensuring that nodeup is truly idempotent if we need it to be.

olemarkus:
Yep.you pay for everyhting you would otherwise on any other stopped instance.

I think not reprovisioning certs is a good optimisation, but perhaps not needed for this PR?
I plan on reviewing other tasks too.

johngmyers:
Probably not needed for this PR. This PR just makes reboots more likely.

johngmyers:
One question is how much time re-running nodeup takes compared to just starting kubelet.

nodeup/pkg/model/context.go
johngmyers: Can be done later.

nodeup/pkg/model/kubelet_test.go
```diff
@@ -157,8 +157,34 @@ func Test_RunKubeletBuilder(t *testing.T) {
...
 		return
 	}
+	runKubeletBuilder(t, context, nodeUpModelContext)
 
-	builder := KubeletBuilder{NodeupModelContext: nodeUpModelContext}
+	testutils.ValidateTasks(t, filepath.Join(basedir, "tasks.yaml"), context)
+
+}
+
+func Test_RunKubeletBuilderConfigurationMode(t *testing.T) {
```
johngmyers, line 166:
```suggestion
func Test_RunKubeletBuilderWarmPool(t *testing.T) {
```

johngmyers (approved):
/hold to give opportunity to fix nit

olemarkus:
Anecdotally, it seems like kubelet joins the cluster 60-90 seconds faster (roughly 25-30 seconds after boot). Most of that time is the usual linux boot stuff like hw probing, dhcp etc.

johngmyers (approved):

PR #11227, opened 2021-04-13
Give kOps CLI knowledge about ASG warm pools

olemarkus:
* `kops get instances` mark warm pool nodes
* rolling update warm pool
   * Warm pool instances will terminate first
   * Warm pool instances will terminate without validation as they are not part of the cluster

Based on #11216 
/hold

### DIFF
--- a/pkg/instancegroups/instancegroups.go
+++ b/pkg/instancegroups/instancegroups.go
@@ -124,6 +124,21 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(group *cloudinstances.
 		}
 	}
 
+	nonWarmPool := []*cloudinstances.CloudInstance{}
+	// Run through the warm pool and delete all instances directly
+	for _, instance := range update {
+		if instance.State == cloudinstances.WarmPool {
+			klog.Infof("deleting warm pool instance %q", instance.ID)
+			err := c.Cloud.DeleteInstance(instance)
+			if err != nil {
+				return fmt.Errorf("failed to delete warm pool instance %q: %w", instance.ID, err)
+			}
+		} else {
+			nonWarmPool = append(nonWarmPool, instance)
+		}
+	}
+	update = nonWarmPool
+
 	if c.Interactive {
 		if maxSurge > 1 {
 			maxSurge = 1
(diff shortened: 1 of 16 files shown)


pkg/instancegroups/instancegroups.go
```diff
@@ -384,9 +392,13 @@ func (c *RollingUpdateCluster) drainTerminateAndWait(u *cloudinstances.CloudInst
...
 
-	// Wait for the minimum interval
-	klog.Infof("waiting for %v after terminating instance", sleepAfterTerminate)
-	time.Sleep(sleepAfterTerminate)
+	if u.State != cloudinstances.WarmPool {
+		// Wait for the minimum interval
+		klog.Infof("waiting for %v after terminating instance", sleepAfterTerminate)
+		time.Sleep(sleepAfterTerminate)
+	} else {
+		klog.Infof("%q is in the warm pool. Not sleeping", u.ID)
```
johngmyers, line 400: I don't think there's a need to log anything here.

johngmyers (changes requested):
This capping of `maxSurge`

https://github.com/kubernetes/kops/blob/ab1b85818d3240c5b9b04bade5c0d3465ce7944d/pkg/instancegroups/instancegroups.go#L111-L113

needs to ignore warm instances in `update`, otherwise rolling update could end up detaching warm instances.

olemarkus:
In `detachInstance`, I added a check for the state, so it should just return if an instance is in the warm pool. AWS will fail a detach instance call on a warm pool instance.

olemarkus:
@johngmyers you were right. there were dragons in my previous approach. Now I am a bit more brutish and just iterate over all the warmpool nodes and quickly terminate them immediately in the beginning.

johngmyers:
I like simple.

hakman (approved):

pkg/instancegroups/instancegroups.go
```diff
@@ -124,6 +124,21 @@ func (c *RollingUpdateCluster) rollingUpdateInstanceGroup(group *cloudinstances.
 		}
 	}
 
+	nonWarmPool := []*cloudinstances.CloudInstance{}
+	// Run through the warm pool and delete all instances directly
+	for _, instance := range update {
+		if instance.State == cloudinstances.WarmPool {
```
johngmyers, line 130: I don't think you need to delete them unless they're NeedsUpdate.

Issue #11062, opened 2021-03-16
KOPS wants to change encrypted volume on etcd on running cluster (but does nothing but complain everytime)


KlavsKlavsen:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
kops 1.18.3

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
1.18.15

**3. What cloud provider are you using?**
aws

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
set cluster with etcd encrypted volume NOT defined.

When I do kops update cluster (after changing minsize to 3) it says:
```
Will modify resources:
  AutoscalingGroup/nodes.k8s.ops.example.com
        MinSize                  2 -> 3

  EBSVolume/a.etcd-events.k8s.ops.example.com
        Encrypted                true -> false

  EBSVolume/a.etcd-main.k8s.ops.example.com
        Encrypted                true -> false
```
If I run kops update cluster - without having changed anything - it shows the above 2 changes - but it never changes them - and I tried setting encryptedVolume: true - in config (as it appearently IS encrypted) - but thats now allowed.

hakman:
Hi @KlavsKlavsen, could you check in the AWS console if those volumes are indeed encrypted or not?

KlavsKlavsen:
All volumes under the instance is marked as Encrypted in AWs console.

hakman:
In this case, I think you have enabled encryption by default or something similar:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html#encryption-by-default

At the moment kOps doesn't know this and it assumes that you had unencrypted volumes that later transformed into encrypted ones. The only way to fix this is to update the config to match reality.

KlavsKlavsen:
@hakman I tried adding that using kops edit cluster - but that was rejected by kops (as not being changeable later on)

rifelpet:
Maybe kops should change where it performs this immutable validation, rather than at the API layer it could ensure the encryption setting isnt changing closer to where it would otherwise call ec2.ModifyVolume.

This would mean the error would only happen if the setting would actually be changed on a volume, during `kops update cluster --yes` rather than during `kops edit cluster` or `kops replace`

h3poteto:
I can reproduce it in master (git-e465d57d81).
[...]

PR #11228, opened 2021-04-14
Remove validations for EBS from cluster validation

h3poteto:
Refs: https://github.com/kubernetes/kops/issues/11062

We can create clusters without specifying `encryptedVolume`, but default encryption is determined by [this setting in AWS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html#encryption-by-default). And default value of `encryptedVolume` in kops spec is `false`. As a result,  there is a difference between the kops spec and the AWS entity. 
To resolve this difference, we have to update `encryptedVolume` in our kops spec. But at the moment kops rejects changes to `encryptedVolume`.

Therefore, I removed a validation for `encryptedVolume` in `kops edit`.
Even if I remove the validation in `kops edit`, there is a validation for `encryptedVolume` in `kops update` so it is no problem.

### DIFF
--- a/pkg/apis/kops/validation/cluster.go
+++ b/pkg/apis/kops/validation/cluster.go
@@ -110,18 +110,6 @@ func validateEtcdMemberUpdate(fp *field.Path, obj kops.EtcdMemberSpec, status *k
 		allErrs = append(allErrs, field.Forbidden(fp.Child("instanceGroup"), "instanceGroup cannot be changed"))
 	}
 
-	if fi.Int32Value(obj.VolumeSize) != fi.Int32Value(old.VolumeSize) {
-		allErrs = append(allErrs, field.Forbidden(fp.Child("volumeSize"), "volumeSize cannot be changed"))
-	}
-
-	if fi.StringValue(obj.KmsKeyId) != fi.StringValue(old.KmsKeyId) {
-		allErrs = append(allErrs, field.Forbidden(fp.Child("kmsKeyId"), "kmsKeyId cannot be changed"))
-	}
-
-	if fi.BoolValue(obj.EncryptedVolume) != fi.BoolValue(old.EncryptedVolume) {
-		allErrs = append(allErrs, field.Forbidden(fp.Child("encryptedVolume"), "encryptedVolume cannot be changed"))
-	}
-
 	return allErrs
 }
 


pkg/apis/kops/validation/cluster.go
```diff
@@ -118,10 +118,6 @@ func validateEtcdMemberUpdate(fp *field.Path, obj kops.EtcdMemberSpec, status *k
 		allErrs = append(allErrs, field.Forbidden(fp.Child("kmsKeyId"), "kmsKeyId cannot be changed"))
 	}
 
-	if fi.BoolValue(obj.EncryptedVolume) != fi.BoolValue(old.EncryptedVolume) {
-		allErrs = append(allErrs, field.Forbidden(fp.Child("encryptedVolume"), "encryptedVolume cannot be changed"))
-	}
-
```
hakman, line 124: @h3poteto I think would be enough to remove the checks for encryption, key and, we are are here, also size.

upup/pkg/fi/cloudup/awstasks/ebsvolume.go
```diff
@@ -181,6 +178,29 @@ func (_ *EBSVolume) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *EBSVolume) e
...
+		if len(volumes.Volumes) == 0 {
+			return fmt.Errorf("could not find volume for: %s", *a.ID)
+		}
+		currentVolume := volumes.Volumes[0]
+		if changes.Encrypted != nil {
+			if *e.Encrypted != *currentVolume.Encrypted {
+				return fmt.Errorf("encryptedVolume cannot be changed")
+			}
+		}
+
```
hakman, line 203: This seems to be doing the same thing changes.Encrypted does, by comparing actual and expected.

hakman (changes requested):

pkg/apis/kops/validation/cluster.go
h3poteto: In this case, I think that it is enough to delete the validation for encryptedVolume. Do I need to change the key and size validation?

hakman: You should remove those too, as key is related to encryption and size can be changed now, so would be nice to not do another PR for it.

hakman: Btw, this will work with other commands like `create -f` or `replace -f`, this is why better to not mention the `edit` comand.

h3poteto: I see, I will remove those.

hakman: Thanks. Please also squash and should be good to merge.

hakman: 👍

hakman (approved):

PR #11235, opened 2021-04-14
Make it possible to enable/configure warm pool

olemarkus:
/hold


pkg/apis/kops/v1alpha2/instancegroup.go
```diff
@@ -178,6 +178,15 @@ type InstanceGroupSpec struct {
 	//   'automatic' (default): apply updates automatically (apply OS security upgrades, avoiding rebooting when possible)
 	//   'external': do not apply updates automatically; they are applied manually or by an external system
 	UpdatePolicy *string `json:"updatePolicy,omitempty"`
+	// WarmPool configures an ASG warm pool for the instance group
+	WarmPool *WarmPoolSpec `json:"warmPool,omitempty"`
+}
+type WarmPoolSpec struct {
```
johngmyers, line 184: These fields strike me as belonging in `InstanceGroupSpec` next to `MinSize` and `MaxSize`, not off in their own subspec.

upup/pkg/fi/cloudup/awstasks/warmpool.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (*WarmPool) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *WarmPool) error {
+
+	svc := t.Cloud.Autoscaling()
+	if changes != nil {
+		if fi.BoolValue(e.Enabled) {
+			request := &autoscaling.PutWarmPoolInput{
+				AutoScalingGroupName:     e.Name,
+				MaxGroupPreparedCapacity: e.MaxSize,
```
johngmyers, line 88: The AWS SDK documentation states

> To remove a value that you previously set, include the property but specify -1 for the value.

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -206,10 +215,15 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, cl
...
 
-	if g.Spec.RootVolumeType != nil && kops.CloudProviderID(cluster.Spec.CloudProvider) == kops.CloudProviderAWS {
-		allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "rootVolumeType"), g.Spec.RootVolumeType, []string{"standard", "gp3", "gp2", "io1", "io2"})...)
+	if kops.CloudProviderID(cluster.Spec.CloudProvider) == kops.CloudProviderAWS {
+		if g.Spec.RootVolumeType != nil {
+			allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "rootVolumeType"), g.Spec.RootVolumeType, []string{"standard", "gp3", "gp2", "io1", "io2"})...)
+		}
+	} else {
+		if g.Spec.WarmPool != nil {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool only allowed on AWS"))
```
johngmyers, line 224:
```suggestion
			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool only supported on AWS"))
```

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -143,6 +143,15 @@ func ValidateInstanceGroup(g *kops.InstanceGroup, cloud fi.Cloud) field.ErrorLis
 
 	allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "updatePolicy"), g.Spec.UpdatePolicy, []string{kops.UpdatePolicyAutomatic, kops.UpdatePolicyExternal})...)
 
+	warmPool := g.Spec.WarmPool
+	if warmPool != nil {
+		if g.Spec.Role != kops.InstanceGroupRoleNode && g.Spec.Role != kops.InstanceGroupRoleAPIServer {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool only allowed on IG with node or apiserver role"))
+		}
+		if g.Spec.MixedInstancesPolicy != nil {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool cannot be combined with a mixed instance policy"))
```
johngmyers, line 152:
```suggestion
			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool cannot be combined with a mixed instances policy"))
```

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -143,6 +143,15 @@ func ValidateInstanceGroup(g *kops.InstanceGroup, cloud fi.Cloud) field.ErrorLis
 
 	allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "updatePolicy"), g.Spec.UpdatePolicy, []string{kops.UpdatePolicyAutomatic, kops.UpdatePolicyExternal})...)
 
+	warmPool := g.Spec.WarmPool
+	if warmPool != nil {
+		if g.Spec.Role != kops.InstanceGroupRoleNode && g.Spec.Role != kops.InstanceGroupRoleAPIServer {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool only allowed on IG with node or apiserver role"))
```
johngmyers, line 149:
```suggestion
			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool only allowed on instance group with role Node or APIServer"))
```

pkg/apis/kops/v1alpha2/instancegroup.go
olemarkus: I think prefixed fields vs subspec is a matter of preference. I find subspecs to be much more user friendly, and finding docs for the fields become a little less needle/haystack in our godocs.

I also expect to expand on the warm pool configuration with more fields, such as configuration for an opt-in lifecycle hook that will guarantee nodeup has finished before stopping the instance.

upup/pkg/fi/cloudup/awstasks/warmpool.go
olemarkus: Oh. Thanks for that one. I was not aware.

pkg/apis/kops/v1alpha2/instancegroup.go
johngmyers: The question is where the subspec boundaries are.

These settings strike me as being about the scaling properties of the IG, so in the same category as `MinSize` and `MaxSize`. The object called a "warm pool" strikes me as an AWS-specific implementation detail. A separate struct would make more sense to me if the pool could be shared across IGs.

Would we go with a separate "warm pool" subspec if the AWS API had these fields directly on the ASG object?

olemarkus: The naming is definitely inherited from AWS. In a not too distant future Google probably comes along and launches "Cold Stand-by Instances" (they are stopped after all).

I would most likely always do substruct over common prefix regardless of what AWS does. AWS chose to do `MinSize` and `MaxGroupPreparedCapacity`, which is ... an interesting choice of naming. I went with `MinSize` and `MaxSize`.

My feelings are not too strong about this though? What are you proposing as alternative? `WarmPoolMinSize`?

johngmyers: The `MaxGroupPreparedCapacity` is a better name than `MaxSize`. It's not the max size of the warmed pool, the number of desired instances are subtracted from it to get the desired number of standby instances. Perhaps something like `MaxWarmOrActive`?

We should perhaps state that it defaults to the value of the IG `MaxSize`.

The AWS documentation doesn't state whether the max setting takes precedence over the min setting. It would be a bit silly if it didn't.

I can see the case for a substruct and we don't currently have the option of moving the IG `MinSize` and `MaxSize` into a new substruct. I just want to ensure due consideration is put into the terminology for things that aren't necessarily AWS-specific. "Warm" is probably the best term for this concept. "Pool" I'm not so sure of.

olemarkus: > The MaxGroupPreparedCapacity is a better name than MaxSize. It's not the max size of the warmed pool, the number of desired instances is subtracted from it to get the desired number of standby instances. Perhaps something like MaxWarmOrActive?

Are you sure about this? If I set max to 10 on the warm pool, I get 10 instances in the warm pool regardless of the amount of nodes I have in the ASG.

I can even set warm pool min and max higher than max of the ASG itself.

johngmyers: That's how I read the AWS documentation, modulo the information lacking on the interaction between min and max. If the documentation doesn't match the behavior, that's an issue for AWS.

olemarkus: I have a few more things that could go into a struct. I have a proof-of-concept code for:
* mentioned life cycle hook. That should be opt-in as if it is enabled without proper max-hop restrictions on metadata service, any pod can ABANDON life cycle hooks.
* pulling down known images for addons (kube-proxy, CNI, other daemonsets). E.g cilium image is fairly large, and pulling it during warming phase shaves of 10
* A list of user-provided images that typically run on the given instance group. It is not uncommon to prebake such images into an AMI. This simplifies that.

The behaviour of warm pools are inconsistent. often there are more instances there than it should be. But we can follow the documentation.

Had I designed this, I would probably call this (hot) standby instances. We could also do set MaxGroupPreparedCapacity = group.maxSize + warmPool.maxSize. From user perspective I find that more intuitive. And I can easily think many made the same mistake I did.

I think I will insist on the substruct. But I have no meaning on the field names. If you have, I am happy to just use them. Same with the name of the substruct itself.

johngmyers: I'm not quite sure why one would want the max prepared capacity to be larger than the max IG size. Would that perhaps be to be able to quickly replace terminated instances when the IG is at max size?

I think the default of using the max IG size for the max prepared capacity (when warming enabled) is reasonable; this means any instance one is willing to spin up should be kept warm. I can also see setting it lower; this means you're willing to scale up if needed but don't want to pay past a certain scale to keep things warm.

I do wonder what happens if min warm pool size is greater than the difference between desired and max IG size.

olemarkus: I think that the warm pool and the active asg nodes are seen as two different things configured independently. It may not make sense, but setting warmPool.minSize > asg.maxSize is possible, and there is no real reason to block this other than common sense. There may always be that one person with an edge case requiring this. with a asg min and max set to 1, I can still set warmPool.minSize 3. You'll never even be able to scale up, but you have a warmPool. My experience is when you terminate nodes rather than scaling up, the ASG is very slow with replacing that terminated node, so this doesn't seem super useful.

Anyways ... naming things ...

johngmyers: I'll need to experiment to see how these settings interact. Perhaps Monday.

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -143,6 +143,15 @@ func ValidateInstanceGroup(g *kops.InstanceGroup, cloud fi.Cloud) field.ErrorLis
 
 	allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "updatePolicy"), g.Spec.UpdatePolicy, []string{kops.UpdatePolicyAutomatic, kops.UpdatePolicyExternal})...)
 
+	warmPool := g.Spec.WarmPool
+	if warmPool != nil {
+		if g.Spec.Role != kops.InstanceGroupRoleNode && g.Spec.Role != kops.InstanceGroupRoleAPIServer {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool only allowed on instance group with role Node or APIServer"))
+		}
+		if g.Spec.MixedInstancesPolicy != nil {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool cannot be combined with a mixed instances policy"))
```
johngmyers, line 152: Apparently it also can't be combined with spot instances.

upup/pkg/fi/cloudup/awstasks/warmpool.go
```diff
@@ -0,0 +1,130 @@
...
+			if minSize == nil {
+				minSize = fi.Int64(-1)
+			}
+			maxSize := e.MaxSize
+			if maxSize == nil {
+				maxSize = fi.Int64(-1)
+			}
+			request := &autoscaling.PutWarmPoolInput{
+				AutoScalingGroupName:     e.Name,
+				MaxGroupPreparedCapacity: maxSize,
```
johngmyers, line 96: In AWS, the max prepared capacity can be no less than the minsize of the warm pool. We should either enforce that through API validation or silently increase the max prepared capacity to be at least the minsize. The fact that the max prepared capacity defaults to the IG maxsize is a complication.

upup/pkg/fi/cloudup/awstasks/warmpool.go
```diff
@@ -0,0 +1,130 @@
...
+	// Name is the name of the ASG
+	Name *string
+	// Lifecycle is the resource lifecycle
+	Lifecycle *fi.Lifecycle
+
+	Enabled *bool
+	// MaxSize is the max number of nodes in the warm pool
+	MaxSize *int64
+	// MinSize is the smallest number of nodes in the warm pool
+	MinSize *int64
```
johngmyers, line 42: Since this defaults to 0, should not be a pointer.

pkg/apis/kops/v1alpha2/instancegroup.go
johngmyers: Experimentation shows the algorithm for number of warm instances is max(maxPrepared - asgDesired, minWarm).

upup/pkg/fi/cloudup/awstasks/warmpool.go
johngmyers: Although AWS allows the maxPrepared to be smaller than the IG minSize, such a configuration is nonsensical. Perhaps we should add an API validation for that.

pkg/apis/kops/v1alpha2/instancegroup.go
johngmyers: Okay, I'm convinced that `MaxSize` is the right name for the max group prepared capacity. It is the number of warm instances you'll get if the desired size is set to 0, otherwise you'll get fewer.

pkg/apis/kops/instancegroup.go
```diff
@@ -182,6 +182,16 @@ type InstanceGroupSpec struct {
 	//   'automatic' (default): apply updates automatically (apply OS security upgrades, avoiding rebooting when possible)
 	//   'external': do not apply updates automatically; they are applied manually or by an external system
 	UpdatePolicy *string `json:"updatePolicy,omitempty"`
+	// WarmPool configures an ASG warm pool for the instance group
+	WarmPool *WarmPoolSpec `json:"warmPool,omitempty"`
+}
+
+type WarmPoolSpec struct {
+	// MinSize is the minimum size of the pool
```
johngmyers, line 190:
```suggestion
	// MinSize is the minimum size of the warm pool.
```

pkg/apis/kops/instancegroup.go
```diff
@@ -182,6 +182,16 @@ type InstanceGroupSpec struct {
...
+	MinSize *int64 `json:"minSize,omitempty"`
+	// MaxSize is the maximum size of the pool
+	// If omitted, it will use the max size of the IG
```
johngmyers, line 193:
```suggestion
	// MaxSize is the maximum size of the warm pool. The desired size of the instance group
	// is subtracted from this number to determine the desired size of the warm pool
	// (unless the resulting number is smaller than MinSize).
	// The default is the instance group's MaxSize.
```

pkg/apis/kops/instancegroup.go
```diff
@@ -182,6 +182,16 @@ type InstanceGroupSpec struct {
 	//   'automatic' (default): apply updates automatically (apply OS security upgrades, avoiding rebooting when possible)
 	//   'external': do not apply updates automatically; they are applied manually or by an external system
 	UpdatePolicy *string `json:"updatePolicy,omitempty"`
+	// WarmPool configures an ASG warm pool for the instance group
```
johngmyers, line 185:
```suggestion
	// WarmPool specifies a pool of pre-warmed instances for later use (AWS only).
```

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -143,6 +143,15 @@ func ValidateInstanceGroup(g *kops.InstanceGroup, cloud fi.Cloud) field.ErrorLis
 
 	allErrs = append(allErrs, IsValidValue(field.NewPath("spec", "updatePolicy"), g.Spec.UpdatePolicy, []string{kops.UpdatePolicyAutomatic, kops.UpdatePolicyExternal})...)
 
+	warmPool := g.Spec.WarmPool
+	if warmPool != nil {
+		if g.Spec.Role != kops.InstanceGroupRoleNode && g.Spec.Role != kops.InstanceGroupRoleAPIServer {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool only allowed on instance group with role Node or APIServer"))
```
johngmyers, line 149:
```suggestion
			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "warmPool"), "warm pool only allowed on instance groups with role Node or APIServer"))
```

upup/pkg/fi/cloudup/awstasks/warmpool.go
```diff
@@ -0,0 +1,130 @@
...
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
+	"k8s.io/kops/upup/pkg/fi/cloudup/cloudformation"
+	"k8s.io/kops/upup/pkg/fi/cloudup/terraform"
+)
+
+// AutoscalingGroup provdes the definition for a autoscaling group in aws
+// +kops:fitask
+type WarmPool struct {
+	// Name is the name of the ASG
```
johngmyers, line 33: All these sentences should end with a period.

johngmyers (changes requested):

upup/pkg/fi/cloudup/awstasks/warmpool.go
olemarkus: I am hesitant to restrict our API more than the AWS API. But if users could get burned by this, I think it makes sense.

johngmyers (approved):
A few nits that can be addressed in a followup PR.

PR #11248, opened 2021-04-17
[cilium] Add support for choosing resources

dntosas:
Cilium as a CNI is a critical component for the cluster so it would be safe
to have some guaranteed resources as well as allowing the users to
define them based on their needs.

In this commit, we init default requested resources and add the
capability of user-defined values.


dntosas:
it seems we hit AWS limits again in test jobs like https://github.com/kubernetes/k8s.io/issues/1921

docs/contributing/adding_a_feature.md
```diff
@@ -87,6 +87,8 @@ Then we conditionally move cilium-operator to masters:
       {{ '{{ end }}' }}
 ```
 
+After changing manifest files remember to run `bash hack/update-expected.sh` in order to get updated [manifestHash](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/cilium/manifest.yaml#L74) values.
```
olemarkus, line 90: Thanks for making this bit more clear. Note that not all manifests have tests that requires `update-expected`.

Perhaps split this to a dedicated commit?

docs/networking/cilium.md
```diff
@@ -138,6 +138,18 @@ Once the secret has been created, encryption can be enabled by setting `enableEn
...
+#### Resources in Cilium
+{{ kops_feature_table(kops_added_default='1.21', k8s_min='1.20') }}
+
+As of kOps 1.20, it is possible to choose your own values for Cilium Agents + Operator. Example:
+```yaml
+  networking:
+    cilium:
+      cpuRequest: "25m"
+      memoryRequest: "128Mi"
+      cpuLimit: "100m"
```
olemarkus, line 150: Setting limits on system pods is a bit risky. I'd rather not allow it. It is better that cilium consumes more memory than expected rather than that it OOMs.

olemarkus (changes requested):

docs/contributing/adding_a_feature.md
dntosas: done

docs/networking/cilium.md
dntosas: removed limits ^^

olemarkus (approved):

PR #11250, opened 2021-04-17
Kubetest2 - Cleanup leaked resources from previous clusters

rifelpet:
If --up is specified then we always delete any previous cluster with the same name since it would conflict otherwise and ensures leaked resources are cleaned up.

We don't do this if --terraform is specified because Down() will run `terraform destroy` but each kubetest2 invocation uses a random temp directory for its terraform state, so it wouldn't know of any resources to destroy

### DIFF
--- a/tests/e2e/kubetest2-kops/deployer/common.go
+++ b/tests/e2e/kubetest2-kops/deployer/common.go
@@ -197,9 +197,14 @@ func (d *deployer) featureFlags() string {
 // defaultClusterName returns a kops cluster name to use when ClusterName is not set
 func defaultClusterName(cloudProvider string) (string, error) {
 	jobName := os.Getenv("JOB_NAME")
+	jobType := os.Getenv("JOB_TYPE")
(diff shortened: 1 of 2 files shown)


rifelpet:
/hold

this was intended to fix the leaked resources identified in https://github.com/kubernetes/kops/pull/11249 but each job invocation in one PR will have different cluster names because the cluster name is dependent on the JOB_NAME (which remains constant) and BUILD_ID (which varies every invocation), so there will never be a cluster with the same name that is running at the start of an invocation.

One option is to have presubmit jobs replace BUILD_ID with the PR number in their cluster name.

hakman (approved):

rifelpet:
/retest

Its possible that all of the regions are "full" of leaked VPCs from other PRs. It seems that aws-janitor has been failing: https://testgrid.k8s.io/sig-testing-maintenance#ci-aws-janitor so we may need to get that working first :/

Issue #11253, opened 2021-04-18
"internal consistency error" when removing mixedInstancesPolicy


h3poteto:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
master (git-27f05ca939)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:28:09Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:25:06Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
At first, `kops edit ig`, and remove `mixedInstancesPolicy` from it.
Next, `kops update cluster`

**5. What happened after the commands executed?**
`internal consistency error` has occur.

```
I0418 15:41:26.003551  102999 executor.go:111] Tasks: 0 done / 100 total; 47 can run
I0418 15:41:26.873204  102999 executor.go:111] Tasks: 47 done / 100 total; 24 can run
I0418 15:41:27.687372  102999 executor.go:111] Tasks: 71 done / 100 total; 23 can run
I0418 15:41:29.501150  102999 executor.go:111] Tasks: 94 done / 100 total; 6 can run
I0418 15:41:29.694244  102999 executor.go:111] Tasks: 100 done / 100 total; 0 can run
Will modify resources:
  AutoscalingGroup/nodes-ap-northeast-1c.playground.k8s.h3poteto.dev
   internal consistency error!
[...]

PR #11255, opened 2021-04-18
Make it possible to detect field changes when mixedInstancePolicy is removed

h3poteto:
Refs: #11253 

When I remove mixedInstancesPolicy using `kops edit`, there are difference actual and expected. But kops don't detect that changes. So, it was causing `internal consistency error`.
I changed `buildChangeList`, so kops can detect field changes when I remove mixedInstancePolicy.


johngmyers:
This brings up the issue of all those other places where we check a subfield of `changes` for `nil`.

johngmyers (approved):

PR #11258, opened 2021-04-18
Pre-pull cilium and kube-proxy in warming mode

olemarkus:
This PR will optimise Node initialisation time by pulling kube-proxy and cilium images if we are in warming mode.

### DIFF
--- a/nodeup/pkg/model/kube_proxy.go
+++ b/nodeup/pkg/model/kube_proxy.go
@@ -61,6 +61,14 @@ func (b *KubeProxyBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	if b.ConfigurationMode == "Warming" {
+		pullTask := &nodetasks.PullImageTask{
+			Name:    kubeProxyImage(b.NodeupModelContext),
+			Runtime: b.Cluster.Spec.ContainerRuntime,
+		}
+		c.AddTask(pullTask)
+	}
+
 	{
 		pod, err := b.buildPod()
 		if err != nil {
@@ -185,11 +193,7 @@ func (b *KubeProxyBuilder) buildPod() (*v1.Pod, error) {
 		flags = append(flags, `--resource-container=""`)
 	}
 
-	image := c.Image
-	if b.Architecture != architectures.ArchitectureAmd64 {
-		image = strings.Replace(image, "-amd64", "-"+string(b.Architecture), 1)
-	}
-
+	image := kubeProxyImage(b.NodeupModelContext)
 	container := &v1.Container{
 		Name:  "kube-proxy",
 		Image: image,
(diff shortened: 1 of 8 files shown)


upup/pkg/fi/nodeup/nodetasks/pull_image.go
```diff
@@ -0,0 +1,100 @@
...
+	Runtime string
+}
+
+var _ fi.Task = &PullImageTask{}
+var _ fi.HasDependencies = &PullImageTask{}
+
+func (t *PullImageTask) GetDependencies(tasks map[string]fi.Task) []fi.Task {
+	// LoadImageTask depends on the docker service to ensure we
+	// sideload images after docker is completely updated and
+	// configured.
```
johngmyers, line 41:
```suggestion
	// LoadImageTask depends on the container runtime service to ensure we
	// sideload images after the container runtime is completely updated and
	// configured.
```

upup/pkg/fi/nodeup/nodetasks/pull_image.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		if svc, ok := v.(*Service); ok && svc.Name == dockerService {
+			deps = append(deps, v)
+		}
+	}
+	return deps
+}
+
+func (e *PullImageTask) Find(c *fi.Context) (*PullImageTask, error) {
+	klog.Warningf("LoadImageTask checking if image present not yet implemented")
```
johngmyers, line 55: This isn't information that would be important when diagnosing a failure, so warning level is too high.

upup/pkg/fi/nodeup/nodetasks/pull_image.go
```diff
@@ -0,0 +1,100 @@
...
+	return nil, nil
+}
+
+func (e *PullImageTask) Run(c *fi.Context) error {
+	return fi.DefaultDeltaRunMethod(e, c)
```
johngmyers, line 60: Since `Find()` can't do anything, it would be simpler to skip the `DefaultDeltaRunMethod` and just inline the contents of `RenderLocal` here.

nodeup/pkg/model/kube_proxy.go
```diff
@@ -61,6 +61,14 @@ func (b *KubeProxyBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	if b.ConfigurationMode == "Warming" {
+		pullTask := &nodetasks.PullImageTask{
+			Name:    kubeProxyImage(b.NodeupModelContext),
+			Runtime: b.Cluster.Spec.ContainerRuntime,
+		}
+		c.AddTask(pullTask)
+	}
```
johngmyers, line 70: It would be simpler to wrap this up in a receiver off of `NodeupModelContext`. It would be even simpler still to push the `ConfigurationMode` test down into the task (renaming it `WarmImageTask`.)

johngmyers (changes requested):
Addressing the warning log level is the only thing I would consider potentially blocking.

johngmyers:
@olemarkus what's your take on extracting out some of the boilerplate to common code? I suspect these aren't going to be the only two images we're going to want to warm-pull.

olemarkus:
Something like this, perhaps?

johngmyers:
Something like that, modulo the compile error.

johngmyers (approved):

PR #11259, opened 2021-04-18
Make nodeup able to complete the warming life cycle hook

olemarkus:
This PR makes nodeup complete the lifecycle action if it exists. It is a no op unless one manually provisions it.
The ability to configure this can be done as an addition to #11235 or as a followup

### DIFF
--- a/pkg/apis/kops/v1alpha2/instancegroup.go
+++ b/pkg/apis/kops/v1alpha2/instancegroup.go
@@ -159,6 +159,9 @@ type WarmPoolSpec struct {
 	// (unless the resulting number is smaller than MinSize).
 	// The default is the instance group's MaxSize.
 	MaxSize *int64 `json:"maxSize,omitempty"`
+	// EnableLifecyleHook determines if an ASG lifecycle hook will be added ensuring that nodeup runs to completion.
+	// Note that the metadata API must be protected from arbitrary Pods when this is enabled.
+	EnableLifecyleHook bool `json:"enableLifecycleHook,omitempty"`
 }
 
 // InstanceMetadataOptions defines the EC2 instance metadata service options (AWS Only)
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -21,6 +21,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
@@ -96,6 +97,26 @@ func (b *AutoscalingGroupModelBuilder) Build(c *fi.ModelBuilderContext) error {
 			if warmPool != nil {
 				warmPoolTask.MinSize = warmPool.MinSize
 				warmPoolTask.MaxSize = warmPool.MaxSize
+
+				if warmPool.EnableLifecyleHook {
+					name := "kops-warmpool"
+
+					lifecyleTask := &awstasks.AutoscalingLifecycleHook{
+						ID:               aws.String(name),
+						Name:             aws.String(name),
+						Lifecycle:        b.Lifecycle,
+						AutoscalingGroup: b.LinkToAutoscalingGroup(ig),
+						DefaultResult:    aws.String("ABANDON"),
+						// We let nodeup have 10 min to complete. Normally this should happen much faster,
+						// but CP nodes need 5 min or so to start on new clusters, and we need to wait for that.
+						HeartbeatTimeout:    aws.Int64(600),
+						LifecycleTransition: aws.String("autoscaling:EC2_INSTANCE_LAUNCHING"),
+					}
+
+					c.AddTask(lifecyleTask)
+
+				}
+
 			}
 			c.AddTask(warmPoolTask)
 		}
--- a/pkg/model/iam.go
+++ b/pkg/model/iam.go
@@ -75,7 +75,15 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
 
 	// Generate IAM tasks for each shared role
 	for profileARN, igRole := range sharedProfileARNsToIGRole {
-		role, err := iam.BuildNodeRoleSubject(igRole)
+		lchPermissions := false
+		for _, ig := range b.InstanceGroups {
+			if ig.Spec.Role == igRole && ig.Spec.WarmPool != nil && ig.Spec.WarmPool.EnableLifecyleHook {
+				lchPermissions = true
+				break
+
+			}
+		}
+		role, err := iam.BuildNodeRoleSubject(igRole, lchPermissions)
 		if err != nil {
 			return err
 		}
@@ -92,7 +100,15 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
 
 	// Generate IAM tasks for each managed role
 	for igRole := range managedRoles {
-		role, err := iam.BuildNodeRoleSubject(igRole)
+		warmPool := false
+		for _, ig := range b.InstanceGroups {
+			if ig.Spec.Role == igRole && ig.Spec.WarmPool != nil && ig.Spec.WarmPool.EnableLifecyleHook {
+				warmPool = true
+				break
+
+			}
+		}
+		role, err := iam.BuildNodeRoleSubject(igRole, warmPool)
 		if err != nil {
 			return err
 		}
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -245,7 +245,7 @@ func (r *NodeRoleAPIServer) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 	}
 
 	addMasterEC2Policies(p, resource, b.Cluster.Spec.IAM.Legacy, b.Cluster.GetName())
-	addASLifecyclePolicies(p, resource, b.Cluster.GetName())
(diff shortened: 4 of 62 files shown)


pkg/model/iam/iam_builder.go
```diff
@@ -1028,8 +1029,32 @@ func addMasterASPolicies(p *Policy, resource stringorslice.StringOrSlice, legacy
...
+				Effect: StatementEffectAllow,
+				Action: stringorslice.Of(
+					"autoscaling:DescribeLifecycleHooks",
+				),
+				Resource: resource,
+			},
+		)
+	}
 	p.Statement = append(p.Statement,
+
```
hakman, line 1057: This seems extra, maybe you want it before?

upup/pkg/fi/nodeup/command.go
```diff
@@ -363,6 +361,42 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
 		klog.Exitf("error closing target: %v", err)
 	}
 
+	if api.CloudProviderID(c.cluster.Spec.CloudProvider) == api.CloudProviderAWS {
```
hakman, line 364: Should this also check if warmPool is enabled?

upup/pkg/fi/nodeup/command.go
```diff
@@ -363,6 +361,42 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
...
+	svc := cloud.(awsup.AWSCloud).Autoscaling()
+	hooks, err := svc.DescribeLifecycleHooks(&autoscaling.DescribeLifecycleHooksInput{
+		AutoScalingGroupName: &asgName,
+		LifecycleHookNames:   []*string{&hookName},
+	})
+	if err != nil {
+		return err
+	}
+
+	klog.Infof("Checking if we should complete a lifecycle hook for %q", asgName)
```
hakman, line 385: Maybe remove from here and add some message on the else branch below.
As is here, it says "Checking" and then nothing.

upup/pkg/fi/nodeup/command.go
```diff
@@ -363,6 +361,42 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
...
+		}
+	}
+	return nil
+}
+
+func completeWarmingLifecycleAction(cloud awsup.AWSCloud, modelContext *model.NodeupModelContext) error {
+	asgName := modelContext.InstanceGroup.GetName() + "." + modelContext.Cluster.GetName()
+	hookName := "kops-warming"
+	svc := cloud.(awsup.AWSCloud).Autoscaling()
+	hooks, err := svc.DescribeLifecycleHooks(&autoscaling.DescribeLifecycleHooksInput{
```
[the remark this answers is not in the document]
olemarkus: Not really. Metadata about the scale-out event, name of the ASG, roles etc.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -96,6 +97,26 @@ func (b *AutoscalingGroupModelBuilder) Build(c *fi.ModelBuilderContext) error {
 			if warmPool != nil {
 				warmPoolTask.MinSize = warmPool.MinSize
 				warmPoolTask.MaxSize = warmPool.MaxSize
+
+				if warmPool.EnableLifecyleHook {
+					name := ig.Name + "-WarmPoolLifecycleHook"
```
hakman, line 102: I think the convention is to use lower case names and not sure "LifecycleHook" adds any value.
```suggestion
					name := ig.Name + "-warmpool"
```

upup/pkg/fi/nodeup/command.go
```diff
@@ -363,6 +361,44 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
...
+				return fmt.Errorf("failed to complete lifecylce action: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+func completeWarmingLifecycleAction(cloud awsup.AWSCloud, modelContext *model.NodeupModelContext) error {
+	asgName := modelContext.InstanceGroup.GetName() + "." + modelContext.Cluster.GetName()
+	hookName := modelContext.InstanceGroup.GetName() + "-WarmPoolLifecycleHook"
```
hakman, line 377: I think the convention is to use lower case names and not sure "LifecycleHook" adds any value.
```suggestion
	hookName := modelContext.InstanceGroup.GetName() + "-warmpool"
```

pkg/model/iam/iam_builder.go
```diff
@@ -1033,8 +1034,32 @@ func addMasterASPolicies(p *Policy, resource stringorslice.StringOrSlice, legacy
...
 		&Statement{
```
hakman, line 1063:
```suggestion

	p.Statement = append(p.Statement,
		&Statement{
```

hakman (changes requested):

pkg/model/awsmodel/autoscalinggroup.go
olemarkus: I don't disagree. I went with this because NTH hook had the same convention.

Personally I would prefer just "warmpool" or "kops-warmpool" for example. The name only need to be unique within the ASG, and adding the IG name to it feels excessive

hakman: I also agree, that we could use just `warmpool`

upup/pkg/fi/nodeup/command.go
hakman: Maybe also just move it to the call as `fi.String("warmpool")

hakman (approved):

pkg/apis/kops/v1alpha2/instancegroup.go
```diff
@@ -159,6 +159,9 @@ type WarmPoolSpec struct {
 	// (unless the resulting number is smaller than MinSize).
 	// The default is the instance group's MaxSize.
 	MaxSize *int64 `json:"maxSize,omitempty"`
+	// EnableLifecyleHook determines if an ASG lifecycle hook will be added ensuring that nodeup runs to completion.
+	// Note that the metadata API must be protected from arbitrary Pods when this is enabled.
+	EnableLifecyleHook bool `json:"enableLifecycleHook,omitempty"`
```
johngmyers, line 164: I would like to question the need for this setting, or at least its default.

What is the risk of enabling this when the metadata API is not protected from arbitrary pods, over and above the risk of having the metadata API be not protected even without enabling this?

olemarkus: Does https://kops.sigs.k8s.io/instance_groups/#lifecycle-hook answer the question?

johngmyers: The documentation states that the risk is that a workload pod could complete the lifecycle hook with the `ABANDONED` result, which would prevent subsequently created instances from joining the cluster (it is not clear to me how that would be). Would the pod be able to abandon the lifecycle hook even though the instance wouldn't be warming? Would the pod not be able to achieve the same result by abandoning some other lifecycle hook?

olemarkus: The lifecycle hook is used for any launching instance, both those going into the warm pool and those going directly into the ASG. The risk is introduced both because we define the launching hook and because we give the node IAM role the ability to complete them.

johngmyers: Can the lifecycle hook be abandoned after it has been completed? Could we not complete that lifecycle hook in nodeup even when the instance isn't warming?

johngmyers: I'm trying to figure out how this is particularly worse than giving pods access to the rest of the instance role's permissions.

We happen to mitigate this by having a post-nodeup hook which asks a lambda to change the instance role to one with only DescribeInstance.

pkg/model/iam.go
```diff
@@ -75,7 +75,15 @@ func (b *IAMModelBuilder) Build(c *fi.ModelBuilderContext) error {
 
 	// Generate IAM tasks for each shared role
 	for profileARN, igRole := range sharedProfileARNsToIGRole {
-		role, err := iam.BuildNodeRoleSubject(igRole)
+		lchPermissions := false
+		for _, ig := range b.InstanceGroups {
+			if ig.Spec.Role == igRole && ig.Spec.WarmPool != nil && ig.Spec.WarmPool.EnableLifecyleHook {
+				lchPermissions = true
```
johngmyers, line 81: So if any IG sets `EnableLifecycleHook` then the permissions will be added to the nodes policy, allowing pods in any nodes group to abandon lifecycle hooks. It seems to me this setting should be cluster-level, not per-IG.

pkg/apis/kops/v1alpha2/instancegroup.go
olemarkus: We complete the hook in all cases, if not the instance would be stuck `pending:wait` until it is terminated because it fails to complete. It can't be abandoned after being completed with `CONTINUE`, but that happens last in nodeup. So plenty of time for a pod to iterate across all instance and try to abandon the lifecycle hook action.

But I don't think there is much case for running with max-hop-limit > 1 on regular nodes. So we may just set that as default. 

We tried to set that as default on all nodes, I believe, but that caused CP nodes to fail because pods there do need to talk to the metadata API.

I.e if we set max-hop-limit = 1 by default, we can flip the default on for the lifecycle hooks as well.

johngmyers: We could start with max-hop-limit=1 for worker nodes to start with. Once we have IRSA in we could also make that the default for control plane nodes when IRSA is enabled.

We should file a ticket with AWS asking for the ability to restrict the IAM permission to same instance.

olemarkus: Yeah. That makes sense.

PR #11261, opened 2021-04-18
Automated cherry pick of #11255: Make it possible to detect field changes when

johngmyers:
Cherry pick of #11255 on release-1.19.

#11255: Make it possible to detect field changes when


hakman:
Please hold a little before cherry-pick. I think the second part of the fix is not exactly correct, even though it may work.
/hold

hakman:
Working on an update for this, but the fix is good enough as is for now.
/hold cancel

hakman (approved):

PR #11271, opened 2021-04-20
Add Azure image to alpha/stable channel

kenji-cloudnatix:
Remove the default image hard-coded in populate_instancegroup_spec.go.

### DIFF
--- a/channels/alpha
+++ b/channels/alpha
@@ -78,6 +78,10 @@ spec:
       providerID: gce
       architectureID: amd64
       kubernetesVersion: ">=1.18.0"
+    - name: "Canonical:UbuntuServer:20.04-LTS:latest"
+      providerID: azure
+      architectureID: amd64
+      kubernetesVersion: ">=1.20.0"
   cluster:
     kubernetesVersion: v1.5.8
     networking:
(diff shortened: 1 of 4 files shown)


channels/alpha
```diff
@@ -78,6 +78,10 @@ spec:
       providerID: gce
       architectureID: amd64
       kubernetesVersion: ">=1.18.0"
+    - name: "Canonical:UbuntuServer:20.04-LTS:latest"
+      providerID: azure
+      architectureID: amd64
+      kubernetesVersion: ">=1.18.0"
```
hakman, line 84: Any special reason for 1.18+? I would use the same version as when kOps started supporting Azure.
```suggestion
      kubernetesVersion: ">=1.20.0"
```

kenji-cloudnatix: I see. No strong reason. Let me update this and the other channel.

hakman: Great. Thanks. :)

hakman (approved):
Thanks @kenji-cloudnatix. Should make it easier to get started with Azure.

kenji-cloudnatix:
Great! Thanks for the quick review, @hakman !

PR #11276, opened 2021-04-21
If one tries to use eip with a public ip that doesn't exist, fail

olemarkus:


upup/pkg/fi/cloudup/awstasks/elastic_ip.go
```diff
@@ -135,6 +135,9 @@ func (e *ElasticIP) find(cloud awsup.AWSCloud) (*ElasticIP, error) {
 		}
 
 		if response == nil || len(response.Addresses) == 0 {
+			if publicIP != nil {
+				return nil, fmt.Errorf("cannot find eip with public ip %q", *e.PublicIP)
+			}
```
hakman, line 140:
```suggestion
			return nil, fmt.Errorf("found no ElasticIPs for: %v", e)
```

hakman (approved):

hakman (approved):

PR #11281, opened 2021-04-21
Add ability to set a default Issuer in certManager addon

javipolo:
This PR allows to set a default clusterIssuer in cert-manager

We're using this heavily to enable automated certs in many ingresses only with the acme annotation


zetaab (approved):
Looks good

PR #11296, opened 2021-04-23
fix: create.go doesnt add --name flag to the prompt: kops update cluster

ebarped:
When you execute `kops create`, it will prompt:
`To deploy these resources, run: kops update cluster %s --yes`, and if you copy/paste this line it wont work.
The correct line is:
`To deploy these resources, run: kops update cluster --name %s --yes`


hakman:
Thanks for fixing this @ebarped.

hakman (approved):

PR #11303, opened 2021-04-23
fix permissions required for NTH Queue Processor

haugenj:
There were more missing permissions for NTH Queue Processor for deleting/updating the resources. Perhaps at this point it would be reasonable to grant full access to sqs and eventbridge, similar to how the other kops user permissions are configured? 

Additionally I noticed a small error when deleting EventBridge rules that had already had their targets removed, so that check is included in here. I can split it into a different pr if that's the preference

### DIFF
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -143,20 +143,27 @@ The kOps CLI requires additional IAM permissions to manage the requisite EventBr
 
 ```json
 {
-    "Version": "2012-10-17",
-    "Statement": [
-        {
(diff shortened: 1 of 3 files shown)


rifelpet (approved):
I think what you have here is good for now 👍🏻
I was hoping to eventually do add a doc on "least permissive IAM policy required for a cluster" complete with conditionals for cluster name tags on applicable resources, perhaps we can do a bigger overhaul on documented IAM permissions then

PR #11304, opened 2021-04-24
[csi/aws] Bump templates + add support for warm pools

dntosas:
AWS EBS CSI Driver:
    - Update manifests
    - Bump components version
    - Add API capability of setting Version + VolumeLimit
    - Add support for warm pools
    - Remove snapshot-controller resources as it should be independent from
    any CSI driver, check here --> https://kubernetes.io/blog/2020/12/10/kubernetes-1.20-volume-snapshot-moves-to-ga/#how-to-deploy-volume-snapshots
 
### DIFF
--- a/pkg/apis/kops/componentconfig.go
+++ b/pkg/apis/kops/componentconfig.go
@@ -836,8 +836,19 @@ type CloudConfiguration struct {
 
 // AWSEBSCSIDriver is the config for the AWS EBS CSI driver
 type AWSEBSCSIDriver struct {
-	//Enabled enables the AWS EBS CSI driver
+	// Enabled enables the AWS EBS CSI driver
+	// Default: false
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// Version is the container image tag used.
+	// Default: The latest stable release which is compatible with your Kubernetes version
+	Version *string `json:"version,omitempty"`
+
+	// VolumeAttachLimit is the maximum number of volumes attachable per node.
+	// If specified, the limit applies to all nodes.
+	// If not specified, the value is approximated from the instance type.
+	// Default: -
+	VolumeAttachLimit *int `json:"volumeAttachLimit,omitempty"`
 }
 
 // NodeTerminationHandlerConfig determines the node termination handler configuration.
--- a/pkg/model/components/awsebscsidriver.go
+++ b/pkg/model/components/awsebscsidriver.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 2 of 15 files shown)


dntosas:
qq: where IAM resources that described in [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/example-iam-policy.json) are handled?

olemarkus:
they are handled in `pkg/model/iam/iam_builder.go`.

It would be nice to collect all the privs needed for CSI to one function though, to make this more maintainable. That would also allow us to use IRSA for this addon further down the road.

dntosas:
@olemarkus are not IAM permissions for attaching etc volumes should be present on regular nodes role also? without them how will the pods of node daemonSet can be allowed to do such operations?

olemarkus:
As I understand it, it is the ebs-csi-controller that does the AWS calls. The purpose of the DaemonSet is to extend kubelet with the driver specific functionality.

dntosas:
> As I understand it, it is the ebs-csi-controller that does the AWS calls. The purpose of the DaemonSet is to extend kubelet with the driver specific functionality.

y makes sense ^^

johngmyers:
I don't see the code for calculating the default value of `Version`

dntosas:
@johngmyers i thought something like [this](https://github.com/kubernetes/kops/blob/e8c8b92b95b49aba161f13a67f080b28c39abe97/upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template#L249) should be enough ^^ if there is a more proper way or place to implement this logic, can you give a hint on where to look?

johngmyers:
@dntosas Okay, that does work; I had missed that. Unfortunately it is error prone to have the version in so many places. If one gets missed on an update that is unlikely to be caught in code review.

The other way is to define a template function, like https://github.com/kubernetes/kops/blob/dbd23473efe25c7eca1576f462b60233cd715d4d/upup/pkg/fi/cloudup/template_functions.go#L588-L608

dntosas:
@olemarkus @johngmyers thanks for your help :)

pushed changes, i think we are better now ^^

pkg/apis/kops/componentconfig.go
```diff
@@ -838,6 +838,16 @@ type CloudConfiguration struct {
 type AWSEBSCSIDriver struct {
 	//Enabled enables the AWS EBS CSI driver
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// Version of the CSI controller container
```
olemarkus, line 842: Nit: go comments should have proper punctuation.

pkg/model/components/awsebscsidriver.go
```diff
@@ -0,0 +1,50 @@
...
+var _ loader.OptionsBuilder = &AWSEBSCSIDriverOptionsBuilder{}
+
+func (b *AWSEBSCSIDriverOptionsBuilder) BuildOptions(o interface{}) error {
+	clusterSpec := o.(*kops.ClusterSpec)
+	c := clusterSpec.CloudConfig.AWSEBSCSIDriver
+	if c == nil {
+		return nil
+	}
+
+	if c.Enabled == nil {
```
olemarkus, line 39: nit: this check is typically noops as we tend to use `if fi.BoolValue()` which defaults to false.
And you _have_ to use `{{ if WithDefaultBool <defaultVaule> }}` in templates as templates will only check if the pointer is nil or not.

olemarkus (changes requested):

olemarkus:
Just a few nits, otherwise looks good to go to me.

olemarkus (approved):

PR #11312, opened 2021-04-24
Use the full operator instead of the generic one

### LINKED ISSUE #11290: Cilium 1.9 uses incorrect cilium-operator
In https://github.com/cilium/cilium/issues/9920 and https://github.com/cilium/cilium/pull/10972, Cilium started shipping a "fat" operator (`cilium-operator`) with everything built in and "slim" operators (eg `cilium-operator-generic` and `cilium-operator-aws`) with only the specific cloud provider options enabled.

The 1.9 addon templates reference the `cilium-operator-generic` (one of the "slim" operators): https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template#L798  This is missing much of the cloud specific functionality, such as AWS ENI IPAM support, which leads to error messages like:

```
level=fatal msg="eni allocator is not supported by this version of cilium-operator-generic" subsys=cilium-operator-generic
```

The 1.8 addon template correctly references the "fat" operator `cilium-operator`: https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.8.yaml.template#L727   which properly includes all the generic functionality _plus_ the cloud specific functionality such as AWS ENI IPAM.

I believe the fix would be to update the `k8s-1.12-v1.9.yaml.template` file to use the `cilium-operator` instead.  As a temporary workaround with kops version 1.20.0 you can update your cluster config to a specific 1.8 based cilium like:

```yaml
spec:
  networking:                                                                                                                                                                                                                                                                                                                 
    cilium:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
      version: v1.8.9 
```

olemarkus:
Fixes #11290

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template
@@ -795,7 +795,7 @@ spec:
         - --config-dir=/tmp/cilium/config-map
         - --debug=$(CILIUM_DEBUG)
         command:
-        - cilium-operator-generic
+        - cilium-operator
         env:
         - name: K8S_NODE_NAME
           valueFrom:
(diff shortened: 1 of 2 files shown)


olemarkus:
This uses the full operator instead of the specific one. We can look to use cloud provider ones when we are adding 1.10 support

johngmyers (approved):
This is a minimal fix. It would be better to pick the appropriate operator for the cloud provider, but I suppose that's not required.

PR #11314, opened 2021-04-24
Expose hubble agent when hubble is enabled

olemarkus:
/kind bug

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template
@@ -193,6 +193,12 @@ data:
   enable-hubble: "true"
   # UNIX domain socket for Hubble server to listen to.
   hubble-socket-path:  "/var/run/cilium/hubble.sock"
+  # An additional address for Hubble server to listen to (e.g. ":4244").
+  hubble-listen-address: ":4244"
+  hubble-disable-tls: "false"
+  hubble-tls-cert-file: /var/lib/cilium/tls/hubble/tls.crt
+  hubble-tls-key-file: /var/lib/cilium/tls/hubble/tls.key
+  hubble-tls-client-ca-files: /var/lib/cilium/tls/hubble/ca.crt
   {{ if .Hubble.Metrics }}
   hubble-metrics-server: ":9091"
   hubble-metrics:
@@ -214,8 +220,13 @@ data:
   config.yaml: |
     peer-service: unix:///var/run/cilium/hubble.sock
     listen-address: :4245
-    disable-client-tls: true
+
     disable-server-tls: true
+
+    tls-client-cert-file: /var/lib/hubble-relay/tls/client.crt
+    tls-client-key-file: /var/lib/hubble-relay/tls/client.key
+    tls-hubble-server-ca-files: /var/lib/hubble-relay/tls/hubble-server-ca.crt
+
 {{ end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
@@ -653,6 +664,9 @@ spec:
           readOnly: true
         - mountPath: /run/xtables.lock
           name: xtables-lock
+        - mountPath: /var/lib/cilium/tls/hubble
+          name: hubble-tls
+          readOnly: true
 {{ if CiliumSecret }}
         - mountPath: /etc/ipsec
           name: cilium-ipsec-secrets
@@ -767,6 +781,10 @@ spec:
         secret:
           secretName: cilium-ipsec-keys
 {{ end }}
+      - name: hubble-tls
+        secret:
+          secretName: hubble-server-certs
+          optional: true
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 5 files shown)


upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template
```diff
@@ -190,6 +190,9 @@ data:
   enable-hubble: "true"
   # UNIX domain socket for Hubble server to listen to.
   hubble-socket-path:  "/var/run/cilium/hubble.sock"
+  # An additional address for Hubble server to listen to (e.g. ":4244").
+  hubble-listen-address: ":4244"
+  hubble-disable-tls: "true"
```
johngmyers, line 195: I'm not a big fan of exposing things without TLS.

olemarkus: Nor do I. But it is a choice of this staying broken in 1.20 or having the feature, but no tls. For 1.21 this will hopefully be fixed

johngmyers: I would prefer not having the feature. There is nothing as permanent as a temporary shortcut.

olemarkus: So you suggest blocking hubble from being enabled through validation? The flag is already there, so we cannot remove this altogether.

olemarkus: I'll implement serving TLS, but mTLS will be more tricky as kops-controller doesn't have any way of provisioning certs to Deployments.

johngmyers: You could have kops provision a `Keypair` and then render it to a `Secret` using a template function.

johngmyers: ...ah, that would make the addons manifest secret. We've had a conversation like this before. Perhaps an auxiliary manifest-like thing which provisions secrets from the keystore.

johngmyers: By the way, kops-controller certs have lifetimes that are tied to nodes. Deployments would need a different lifetime and renewal mechanism. This is why people tend to use JWT over mTLS--Kubernetes has the infrastructure for the former.

olemarkus: Turns out cilium starts normal operations, then wait for the availability of the certificate. So there really isn't any chicken/egg issue here wrt cert-manager.

I have deployed addon PKI + 3 certs (hubble server cert, relay client cert, relay server cert) and tried to get them to work, but something in the mTLS setup is failing. It works when I connect with hubble CLI using the same certs and flags that hubble-relay is supposed to have.

olemarkus: My problems were caused by our addon CAs only setting common name. This PR contains a workaround now, but we need to resolve the CA certificate problem later on.

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template
```diff
@@ -958,5 +983,72 @@ spec:
...
+          - secret:
+              name: hubble-relay-server-certs
+              items:
+                - key: tls.crt
+                  path: server.crt
+                - key: tls.key
+                  path: server.key
+        name: tls
+---
+apiVersion: cert-manager.io/v1alpha2
```
codablock, line 1006: Shouldn't this be v1 already?

olemarkus: ouch yes. not sure how v1alpha2 made it there.  thanks for cat

codablock:
fyi, I have this PR merged and running in one of our clusters and already succesfully used it to debug network failures.

olemarkus:
I am not sure why hubble-ui is failing. but this still solves the hubble issue, so I think we can just merge it.

codablock:
Looks like hubble-ui is failing for 2 different reasons:
1. See cilium Slack: https://cilium.slack.com/archives/CQRL1EPAA/p1619706586139400
    This is fixed in the hubble-ui image v0.7.7
2. hubble-ui seems to only work when hubble-relay is running without TLS and there is currently no way to enable TLS in hubble-relay. At the same time, you have fixed the hubble-relay service-port to 80, even though TLS is enabled, which seems to be misleading. Looking into the cilium Helm chart, I see that they use port 443 when TLS is enabled.

So, for this PR, I assume that at least 2. should be fixed first.

olemarkus:
There. Now it works for me.

We'll just have to live with relay TLS false when the UI doesn't support TLS.

hakman (approved):

PR #11316, opened 2021-04-24
Allow disabling warm pool by setting WarmPool.MaxSize to 0

johngmyers:


olemarkus:
What is the use case for this? Perhaps if one wants to keep the struct around, but disable the pool, the standard `enabled` flag would be better?

johngmyers:
This would allow us to add a `warmPool` struct at cluster level, providing defaults. Much like we do for `rollingUpdate`.

olemarkus (approved):
Sounds good

Issue #11251, opened 2021-04-17
1.21.0 addon bumps


olemarkus:
We should bump at least the following addons before 1.21

* [x] AWS EBS CSI driver
* [x] cluster autoscaler
* [x] cert-manager
* [x] AWS LB controller
* [x] metrics-server
* [x] cilium
* [x] node local dns cache
* [x] node termination handler #10995 
* [x] coredns

olemarkus:
All updated :tada: 

/close

PR #11319, opened 2021-04-25
[metrics-server] Bump manifest to latest stable

dntosas:
- Bump image to v0.4.3
- Set new default kubelet options based on upstream behavior
- Set default requested resources

As part of https://github.com/kubernetes/kops/issues/11251

### DIFF
--- a/upup/models/cloudup/resources/addons/metrics-server.addons.k8s.io/k8s-1.11.yaml.template
+++ b/upup/models/cloudup/resources/addons/metrics-server.addons.k8s.io/k8s-1.11.yaml.template
@@ -1,4 +1,4 @@
-# sourced from https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.7/components.yaml
+# sourced from https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.4.3/components.yaml
 ---
 apiVersion: v1
 kind: ServiceAccount
@@ -79,7 +79,6 @@ subjects:
   name: metrics-server
   namespace: kube-system
 ---
----
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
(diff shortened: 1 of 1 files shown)


upup/models/cloudup/resources/addons/metrics-server.addons.k8s.io/k8s-1.11.yaml.template
```diff
@@ -133,6 +130,8 @@ spec:
       containers:
       - args:
           - --secure-port=4443
+          - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
```
olemarkus, line 133: Did you test this one? I am not sure if either InternalIP or ExternalIP is part of the kublet cert. Iirc we only add hostname.

dntosas: you are right ^^ these can only work when going insecure.

should we put them into insecure block or remove them completely?

olemarkus: Maybe just explicitly set it to `Hostname` and add a comment why.

dntosas: i would suggest removing it completely and just let the `metrics-server` use its default behavior which is always Hostname --> DNS --> IPs as shown in [here](https://github.com/kubernetes-sigs/metrics-server/blob/master/pkg/utils/address_resolver.go#L24-L38) and avoid confusing maintainers. wdyt?

dntosas: also is this way, in edge cases where DNS is down, metrics-server would still operate fallback into IPs

olemarkus: As long as the defaults do not change, that is fine :) with secure: true, the fallbacks won't work anyway though.

olemarkus (approved):

PR #11320, opened 2021-04-25
Fix kops-controller's GCE ListManagedInstances filter

rifelpet:
We are seeing [logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/10847/pull-kops-e2e-k8s-gce/1385969563077709824/artifacts/cluster-info/kube-system/kops-controller-pbmbh/logs.txt) of "found instances with mismatched id" indicating the filter is not working correctly (see Line 194 of identify.go).

The [documentation](https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroupManagers/listManagedInstances#response-body) mentions that the ID field is actually a string, so this quotes the value of the id filter to match [other filter examples](https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroupManagers/listManagedInstances#query-parameters) using string fields.

### DIFF
--- a/pkg/nodeidentity/gce/identify.go
+++ b/pkg/nodeidentity/gce/identify.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strconv"
 	"strings"
 
 	"cloud.google.com/go/compute/metadata"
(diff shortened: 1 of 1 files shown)


rifelpet:
I anticipate this job failing but the kops-controller logs should be lacking that warning.

/test pull-kops-e2e-k8s-gce

rifelpet:
testing this more it appears that filtering is transparently ignored on this particular API. `gcloud --log-http` shows the filter isn't even included in the request and instead performs the filtering client-side. I suppose we could delete the warning log but i'll just close this for now.

/close

PR #11324, opened 2021-04-26
WIP: Well-known additional objects

justinsb:
Builds on #11016

Prototype how we can define objects out of the main cluster spec.

Paving the way for operators, though not currently implemented as an
operator.


olemarkus:
I don't think we want this particular functionality (anything that touches IAM) to run from within the cluster. even with IRSA, that sounds risky.

But I do like how this can be used to pull the cloud-provider specific configuration out of the cluster spec.

justinsb:
> I don't think we want this particular functionality (anything that touches IAM) to run from within the cluster. even with IRSA, that sounds risky.

I agree with that.  In future, we probably do want it to run in a management cluster if we're running kOps in a kubernetes cluster that creates other clusters (as opposed to the normal CLI way).

I'm thinking we can either have an annotation to say "this is bootstrap" or we can hard-code the list of things that should always run in bootstrap, because we can't do bootstrap without linking into the code.

We can always add an annotation later, if we start with hard-coding.

PR #11325, opened 2021-04-26
Add support for cri-o runtime

bharath-123:
Cri-o is a fairly young container runtime which aims to be a runtime made for Kubernetes. It implements the Kubelet CRI spec using OCI conformant runtimes. It is highly tied to Kubernetes releases. 

This change adds two new fields in the cluster spec:
1. A new value "crio" to the list of support containerRuntimes in kOps.
2. crio field to the cluster spec to specify the crio configuration.

All the crio configuration fields have defaults. A lot of the default settings have been delegated to the defaults suggested by cri-o `https://github.com/cri-o/cri-o/blob/master/docs/crio.conf.5.md.
```
 containerRuntime: crio // To select the crio runtime
 crio: 
     configOverride:
         [api]
         listen=/run/crio/crio.sock
     containerPolicyOverride:
        "{ "default": [{ "type": "insecureAcceptAnything" }] }"
     containerRegistriesOverride:
         unqualified-search-registries=["docker.io", "k8s.gcr.io", "quay.io"]
      logLevel:
          info
      version: 
          1.21.0
```

This change does not support cri-o for kubenet cni plugin as I think there is an intent to deprecate it. I have tested this with calico CNI plugin. 

Cri-o support is added only for kubernetes versions >= 1.21. cri-o was not supporting arm bundled binaries till their 1.21 release.

This change adds podman to build an e2e test for cri-o on ubuntu 20.10. podman is intended to handle image loading from a tar file.


nodeup/pkg/model/crio.go
```diff
@@ -0,0 +1,194 @@
...
+		t := &nodetasks.File{
+			Path:     "/etc/crio/crio.conf",
+			Contents: fi.NewStringResource(fi.StringValue(b.Cluster.Spec.Crio.ConfigOverride)),
+			Type:     nodetasks.FileType_File,
+		}
+		c.AddTask(t)
+	}
+
+	// Add binaries from assets
+	if b.Cluster.Spec.ContainerRuntime == "crio" {
```
johngmyers, line 89: Can't ever be false

nodeup/pkg/model/crio.go
```diff
@@ -0,0 +1,194 @@
...
+		f := b.Assets.FindMatches(regexp.MustCompile(`bin/(crio|conmon|pinns|crio-status|crun|runc)`))
+		//f := b.Assets.FindMatches(regexp.MustCompile(`(bin/crio|bin/conmon|bin/pinns|bin/crio-status)`))
+		if len(f) == 0 {
+			return fmt.Errorf("unable to find any crio binaries in assets")
+		}
+		for k, v := range f {
+			fileTask := &nodetasks.File{
+				Path:     filepath.Join("/usr/bin", k),
```
johngmyers, line 97: Can we even write to /usr/bin on distributions like Flatcar or ContainerOS?

pkg/model/components/crio.go
```diff
@@ -0,0 +1,125 @@
...
+		if fi.StringValue(crio.ContainerRegistriesOverride) == "" {
+			crio.ContainerRegistriesOverride = c.buildCrioDefaultContainerRegistriesPolicy()
+		}
+
+		if crio.LogLevel == nil {
+			crio.LogLevel = fi.String("info")
+		}
+
+	} else {
+		crio.SkipInstall = true
```
johngmyers, line 78: Would be easier to follow if you inverted the sense of the if block and then did an early return here.

nodeup/pkg/model/packages.go
```diff
@@ -52,6 +52,12 @@ func (b *PackagesBuilder) Build(c *fi.ModelBuilderContext) error {
 		c.AddTask(&nodetasks.Package{Name: "pigz"})
 		c.AddTask(&nodetasks.Package{Name: "socat"})
 		c.AddTask(&nodetasks.Package{Name: "util-linux"})
+		if b.Distribution == distributions.DistributionUbuntu2010 {
+			// install podman for crio testing
+			c.AddTask(&nodetasks.Package{Name: "podman"})
+		}
+		// For ARM testing
+		c.AddTask(&nodetasks.Package{Name: "runc"})
```
johngmyers, line 60: We probably shouldn't install these if the container runtime isn't crio.

pkg/apis/kops/validation/validation.go
```diff
@@ -1101,74 +1105,124 @@ func validateCalicoIPPoolEncapsulationMode(mode string, fldPath *field.Path) fie
...
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("version"), config.Version,
+				fmt.Sprintf("unable to parse version string: %s", err.Error())))
+		}
+		if sv.LT(semver.MustParse("1.3.4")) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("version"), config.Version,
+				"unsupported legacy version"))
+		}
+	}
+
+	//if config.Packages != nil {
```
johngmyers, line 1179: Please remove all commented-out code.

johngmyers (changes requested):

go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	github.com/aliyun/alibaba-cloud-sdk-go v1.61.1046
 	github.com/aws/amazon-ec2-instance-selector/v2 v2.0.2
 	github.com/aws/aws-sdk-go v1.38.25
+	github.com/blang/semver v3.5.1+incompatible
```
hakman, line 50: This should not be added back, probably you are importing semver somewhere without /v4.

nodeup/pkg/model/packages.go
hakman: Agreed and also should not install `runc`. This should be installed from the cri-o package.

hakman (changes requested):

pkg/model/components/crio.go
```diff
@@ -0,0 +1,125 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package components
+
+import (
+	"fmt"
+
+	"github.com/blang/semver"
```
rifelpet, line 22: This should negate the go.mod change

```suggestion
	"github.com/blang/semver/v4"
```

hakman:
At the moment there is no plan to continue work on CRI-O support.

PR #11326, opened 2021-04-26
Fix arguments to csi-provisioner after bump to v2.2.0

codablock:
Parameters for leader elections have changed in v2.2.0, which causes the
csi-provisioner to not start anymore.


johngmyers:
@codablock please run `hack/update-expected.sh`

olemarkus (approved):
Good catch. thanks

PR #11329, opened 2021-04-27
Set IMDSv2 on by default for nodes and apiservers

olemarkus:
This should mitigate security concerns around things like e.g enabling lifecycle hooks by default

### DIFF
--- a/docs/releases/1.22-NOTES.md
+++ b/docs/releases/1.22-NOTES.md
@@ -6,6 +6,20 @@ This is a document to gather the release notes prior to the release.
 
 # Significant changes
 
+## Instance metadata service version 2
+
+ On AWS, kOps will enable [Instance Metadata Service Version 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) by default with a max-hop-limit of 1 on new clusters that use Kubernetes 1.22. This means that any container running in the cluster will be unable to connect to the instance metadata _unless_ the container is running with `hostNetworking: true`. This will increase security by default, but may break some types of workloads. In order to revert to old behavior, add the following to the InstanceGroup:
+
+```
+spec:
+  instanceMetadata:
+    httpTokens: optional
+```
+
+This change only affects dedicated API server nodes and worker nodes. It does not affect control plane nodes.
+
+## Other significant changes
+
 * New clusters running Kubernetes 1.22 will have AWS EBS CSI driver enabled by default.
 
 # Breaking changes
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -693,6 +693,13 @@ func setupMasters(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap 
 				g.Spec.Zones = []string{zone}
 			}
 
+			if cluster.IsKubernetesGTE("1.22") {
+				g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+					HTTPPutResponseHopLimit: fi.Int64(3),
+					HTTPTokens:              fi.String("required"),
+				}
+			}
+
 			masters = append(masters, g)
 		}
 	}
@@ -807,6 +814,13 @@ func setupNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap ma
 			g.Spec.Zones = []string{zone}
 		}
 
+		if cluster.IsKubernetesGTE("1.22") {
+			g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+				HTTPPutResponseHopLimit: fi.Int64(1),
+				HTTPTokens:              fi.String("required"),
+			}
+		}
+
 		nodes = append(nodes, g)
 	}
 
@@ -846,6 +860,13 @@ func setupAPIServers(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetM
 		g.Spec.Subnets = []string{subnet.Name}
 		if cp := api.CloudProviderID(cluster.Spec.CloudProvider); cp == api.CloudProviderGCE || cp == api.CloudProviderAzure {
 			g.Spec.Zones = []string{zone}
+		}
+
+		if cluster.IsKubernetesGTE("1.22") {
+			g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+				HTTPPutResponseHopLimit: fi.Int64(1),
+				HTTPTokens:              fi.String("required"),
+			}
 		}
 
 		nodes = append(nodes, g)
@@ -998,6 +1019,14 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
 			if api.CloudProviderID(cluster.Spec.CloudProvider) == api.CloudProviderGCE {
 				bastionGroup.Spec.Zones = allZones.List()
 			}
+
+			if cluster.IsKubernetesGTE("1.22") {
+				bastionGroup.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+					HTTPPutResponseHopLimit: fi.Int64(1),
+					HTTPTokens:              fi.String("required"),
+				}
+			}
+
 		}
 
 	default:
--- a/tests/integration/create_cluster/minimal-1.22/expected-v1alpha2.yaml
+++ b/tests/integration/create_cluster/minimal-1.22/expected-v1alpha2.yaml
@@ -66,6 +66,9 @@ metadata:
   name: master-us-test-1a
 spec:
   image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20210415
+  instanceMetadata:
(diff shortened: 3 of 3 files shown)


rifelpet:
I recall we initially defaulted this to enabled but reverted it in https://github.com/kubernetes/kops/pull/10655. Can we address the issues that lead to the revert?

olemarkus:
The revert was because it caused any pod that piggybacked on the instance's IAM role to fail. This _should_ not happen on API server or worker nodes though. I hope it is rare that people add extra privileges to normal nodes and rather rely on KIAM (who should have their servers running on the control plane).

It also caused breakage on EBS CSI nodes because they rely on talking to the metadata service. This breakage was technically caused by me by removing `hostnetworking: true` from the DS manifest. We have added this back in now. Turns out it was exactly because of this they ran with `hostNetworking`.

I.e we should have addressed the issues now.

rifelpet:
Typically changes like this we only enable for new clusters (and sometimes only for new clusters with k8s minor versions >= the kops minor version)

Any reason not to follow that same pattern here?

olemarkus:
I want to play the security card here. But I am fine if we constrain this to a given k8s version too. But I think we really need this on by default on normal nodes. @johngmyers what do you think?

olemarkus:
What makes flatcar break? Does it handle routing differently than other distros?

hakman:
Flatcar breaks because of the outdated AWS SDK: https://github.com/kinvolk/ignition/pull/19.
I agree that we should have secure defaults, but seems that this will require too many exceptions. Maybe it's not yet time for it to be a default.

olemarkus:
Honestly, to me it looks like in this case we should use IMDSv2 by default and flatcar users should explicitly disable it. That this hasn't been fixed in flatcar sounds odd and I would prefer this not holding every other distro back from secure by default.

We can add an additinal callout here: https://kops.sigs.k8s.io/operations/images/#flatcar

hakman:
Same story with Debian 9. It will most likely break many obscure workflows or setups where people were lazy to update.
Security minded people have the option to enable IMDSv2 quite easily. They can also debug any issues that might appear, much more than the others that won't know what hit them.

rifelpet:
I agree that we should create secure clusters by default. 

Does debian 10 work? I'm fine with documenting that using certain distros (particularly older ones) on new clusters will need to have an instancegroup field overridden. We would need to do something similar for userdata generation for bottlerocket and windows anyways, so it seems like a reasonable pattern to follow.

docs/releases/1.21-NOTES.md
```diff
@@ -26,6 +26,19 @@ See the [warm pool](https://kops.sigs.k8s.io/instance_groups/#warmpool-aws-only)
...
+ On AWS, kOps will enable [Instance Metadata Service Version 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) by default with a max-hop-limit of 1. This means that any container running in the cluster will be unable to connect to the instance metadata _unless_ the container is running with `hostNetworking: true`. This will increase security by default, but may break some types of workloads. In order to revert to old behavior, add the following to the InstanceGroup:
+
+```
+spec:
+  instanceMetadata:
+    httpPutResponseHopLimit: 1
+    httpTokens: optional
+```
+
+This change does not affect control plane nodes.
```
hakman, line 40: Maybe best moved in main message.

hakman:
Debian 10 does work. I can't say that I really agree that these are corner cases, but I agree that the vote is to go forward like this.
A problem that I see is that there is no easy way to keep automated tests working for Flatcar, as there is no flag to override this.
Also, as Peter was saying earlier, it would apply for older k8s versions, that we still support and for which the default image is Debian 9.

rifelpet:
In office hours we decided that we can enable this for _new_ k8s 1.22 clusters. This address the Debian 9 issue because it is already EOL, leaving flatcar compatibility unknown.

rifelpet:
We have a [k8s-latest job](https://testgrid.k8s.io/kops-versions#kops-aws-k8s-latest&show-stale-tests=) but apparently its history is too large for testgrid and we need to reduce the retention on it. We also have an [arm64 job](https://testgrid.k8s.io/kops-misc#kops-aws-misc-arm64-ci&show-stale-tests=) that you could observe

olemarkus:
Both of those cilium tests use latest. Should be good enough to see if this breaks anything.

docs/releases/1.22-NOTES.md
```diff
@@ -6,6 +6,20 @@ This is a document to gather the release notes prior to the release.
...
 # Significant changes
 
+## Instance metadata service version 2
+
+ On AWS, kOps will enable [Instance Metadata Service Version 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) by default with a max-hop-limit of 1 on new clusters that use Kubernetes 1.22. This means that any container running in the cluster will be unable to connect to the instance metadata _unless_ the container is running with `hostNetworking: true`. This will increase security by default, but may break some types of workloads. In order to revert to old behavior, add the following to the InstanceGroup:
+
+```
+spec:
+  instanceMetadata:
+    httpPutResponseHopLimit: 1
```
johngmyers, line 16: Is this line necessary?

docs/releases/1.22-NOTES.md
```diff
@@ -6,6 +6,20 @@ This is a document to gather the release notes prior to the release.
...
+    httpTokens: optional
+```
+
+This change does not affect control plane nodes.
```
hakman, line 20: I think this change will affect control plane nodes also.

docs/releases/1.22-NOTES.md
olemarkus: No. Removed.

docs/releases/1.22-NOTES.md
olemarkus: Changed the wording to it is more clear.

hakman: The problem is not the wording but the intention.
As @rifelpet was saying above, conclusion from office hours was:
> **all node roles**, token requred, hop limit 1

olemarkus: We  cannot implement this on master roles though. Not until IRSA is used for all components and enabled by default.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -798,6 +798,13 @@ func setupNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap ma
 		g.Spec.MaxSize = fi.Int32(count)
 		g.ObjectMeta.Name = "nodes-" + zone
 
+		if cluster.IsKubernetesGTE("1.22") {
+			g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+				HTTPPutResponseHopLimit: fi.Int64(1),
+				HTTPTokens:              fi.String("required"),
+			}
+		}
+
```
hakman, line 807: Could you move at the end of the function, same you did for API servers?

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -999,6 +1020,14 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
 			if api.CloudProviderID(cluster.Spec.CloudProvider) == api.CloudProviderGCE {
 				bastionGroup.Spec.Zones = allZones.List()
 			}
+
+			if cluster.IsKubernetesGTE("1.22") {
+				bastionGroup.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+					HTTPPutResponseHopLimit: fi.Int64(3),
```
hakman, line 1026: Any special reason for using 3 as hop limit for Bastions?

olemarkus: No. Seems like a mistake.

hakman (approved):

docs/releases/1.22-NOTES.md
```diff
@@ -6,6 +6,20 @@ This is a document to gather the release notes prior to the release.
 
 # Significant changes
 
+## Instance metadata service version 2
+
+ On AWS, kOps will enable [Instance Metadata Service Version 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) by default with a max-hop-limit of 1 on new clusters that use Kubernetes 1.22. This means that any container running in the cluster will be unable to connect to the instance metadata _unless_ the container is running with `hostNetworking: true`. This will increase security by default, but may break some types of workloads. In order to revert to old behavior, add the following to the InstanceGroup:
```
rifelpet, line 11: should we clarify that control plane nodes get a limit of 3?

docs/releases/1.22-NOTES.md
```diff
@@ -6,6 +6,20 @@ This is a document to gather the release notes prior to the release.
...
+
+```
+spec:
+  instanceMetadata:
+    httpTokens: optional
+```
+
+This change only affects dedicated API server nodes and worker nodes. It does not affect control plane nodes.
```
rifelpet, line 19: I think this release note needs updating to match the final behavior in this PR. It _does_ affect control plane nodes but with a hop limit of 3 (along with apiserver nodes)

Issue #5004, opened 2018-04-14
Deprecate etcd event cluster


klausenbusk:
1. Describe IN DETAIL the feature/behavior/change you would like to see.

According to @chrislovecnm the etcd event cluster isn't needed anymore, but as it is legacy thing it is hard to remove.

2. Feel free to provide a design supporting your feature request.

Add logic so new clusters is created without a etcd event cluster, but still support old cluster with a etcd event cluster.

andrewsykim:
Thread: https://kubernetes.slack.com/archives/C8MKE2G5P/p1523721679000164

andrewsykim:
It seems like this is actually doable by removing `events` from [here](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/populate_cluster_spec.go#L43) and this would only be applied for new clusters since we only read those values when we write the initial cluster spec [here](https://github.com/kubernetes/kops/blob/master/cmd/kops/create_cluster.go#L630-L651). I'm not sure  if we need to account for iops/size for new volumes but that change seems like a decent start. @justinsb @chrislovecnm thoughts?

andrewsykim:
The original reason for having a separate events cluster was due to scalability issues with etcd (https://github.com/kubernetes/kubernetes/issues/32361) and at the time it was recommended to separate events to a separate cluster.  Having said that, it seems in the issue most of the scalability problems have been addressed in etcd3(?) so we may want to revisit this when we support etcd3. cc @justinsb

PR #11330, opened 2021-04-27
Make events etcd cluster optional

codablock:
Related to https://github.com/kubernetes/kops/issues/5004

The events cluster is not strictly needed anymore as etcd v3 is able to handle events and main in the same cluster. This PR only removes the strict requirement, but will still create the cluster via "kops create cluster". Future versions should consider not creating the cluster by default.


codablock:
Any chance to get this into 1.21? Would make things easier for us as I'm already actively using this

hakman:
Hi @codablock. We will discuss about this today at our weekly meeting and get back to you.

pkg/model/components/apiserver.go
```diff
@@ -133,8 +133,6 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
 
 	c.AllowPrivileged = fi.Bool(true)
 	c.ServiceClusterIPRange = clusterSpec.ServiceClusterIPRange
-	c.EtcdServers = []string{"http://127.0.0.1:4001"}
-	c.EtcdServersOverrides = []string{"/events#http://127.0.0.1:4002"}
```
hakman, line 137: I would prefer this logic to remain in model and in NodeUp to just use whatever was set earlier. What do you think @codablock?

hakman (approved):

pkg/model/components/apiserver.go
codablock: I'm not sure I understand. You want to leave these two lines as they were? If this is what you mean, then it becomes strange in `buildPod`, as I'd have to clear `EtcdServersOverrides` in case of `! hasEventsEtcdCluster`, which IMHO would be strange and not immediately clear why this is done (for example, it was a surprise for me that `EtcdServersOverrides` was still set, even though I skipped the assignment in `buildPod` when I started trying out things).

pkg/model/components/apiserver.go
```diff
@@ -133,8 +133,6 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
...
```
hakman, line 136: Trying to explain my previous comment, model section:
```suggestion

	for _, etcdCluster := range clusterSpec.EtcdClusters {
		switch etcdCluster.Name {
		case "main":
			if etcdCluster.EnableEtcdTLS {
				c.EtcdServers = append(c.EtcdServers, "https://127.0.0.1:4001")
			} else {
				c.EtcdServers = append(c.EtcdServers, "http://127.0.0.1:4001")
			}
		case "events":
			if etcdCluster.EnableEtcdTLS {
				c.EtcdServersOverrides = append(c.EtcdServersOverrides, "/events#https://127.0.0.1:4002")
			} else {
				c.EtcdServersOverrides = append(c.EtcdServersOverrides, "/events#http://127.0.0.1:4002")
			}
		}
	}

```

nodeup/pkg/model/kube_apiserver.go
```diff
@@ -312,6 +312,13 @@ func (b *KubeAPIServerBuilder) buildPod() (*v1.Pod, error) {
 		}
 	}
 
+	hasEventsEtcdCluster := false
+	for _, etcdCluster := range b.Cluster.Spec.EtcdClusters {
+		if etcdCluster.Name == "events" {
+			hasEventsEtcdCluster = true
+		}
+	}
```
hakman, line 320: Trying to explain my previous comment, nodeup section:
```suggestion
	if !b.IsMaster {
		clusterName := b.Cluster.ObjectMeta.Name
		mainEtcdDNSName := "main.etcd." + clusterName
		eventsEtcdDNSName := "events.etcd." + clusterName
		for i := range kubeAPIServer.EtcdServers {
			kubeAPIServer.EtcdServers[i] = strings.ReplaceAll(kubeAPIServer.EtcdServers[i], "127.0.0.1", mainEtcdDNSName)
		}
		for i := range kubeAPIServer.EtcdServersOverrides {
			if strings.HasPrefix(kubeAPIServer.EtcdServersOverrides[i], "/events") {
				kubeAPIServer.EtcdServersOverrides[i] = strings.ReplaceAll(kubeAPIServer.EtcdServersOverrides[i], "127.0.0.1", eventsEtcdDNSName)
			}
		}
	}
```

pkg/model/components/apiserver.go
hakman: I tried to explain with code what I meant. Easier than words :).

codablock:
@hakman It feels strange to solve it this way tbh. Replacing some IP (127.0.0.1) with a DNS name is not immediately clear why it is happening (it took me some time to understand it). Can't we do all this (127.0.0.1 vs DNS) in `pkg/model/components/apiserver.go` instead of later replacing it?

I'm also trying to understand the `b.IsMaster`, is this for apiservers running in dedicated instance groups?

hakman:
@codablock I think we only get the info about the node type in NodeUp, so we can decide if we use 127.0.0.1 or hostname based on Master or APIServer type. In model we have a warning about adding a dependency on InstanceGroups. 
https://github.com/kubernetes/kops/blob/1761076b38648018a4de7f93709a7654a4517326/pkg/model/components/apiserver.go#L186-L191

@rifelpet @johngmyers @olemarkus any thoughts on which approach would be better?

codablock:
@hakman I've update the PR to use your suggested solution (a few changes were required to make tests pass)

johngmyers:
According to the apiserver command line reference, this flag is only in use when `--endpoint-reconciler-type=master-count` is enabled. The default is `--endpoint-reconciler-type=lease` and kOps does not support changing it from the default.

I believe we can remove the code for setting the `--apiserver-count` flag.

pkg/model/components/apiserver.go
```diff
@@ -133,8 +133,21 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
...
+	c.EtcdServers = nil
+	c.EtcdServersOverrides = nil
```
hakman, line 137: Are these actually needed? Should be nil by default.

hakman: I understand now :). Builder runs until there are no changes between runs, but without clearing the object first.

codablock: Yeah exactly...showed up in tests :)

hakman: I'll pretend it never happened :D

Issue #11282, opened 2021-04-22
WarmPool task errors upon new cluster creation


johngmyers:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

Version 1.21.0-alpha.3 (git-efd34a80aa)

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

`kops create cluster` yada yada

**5. What happened after the commands executed?**

```
I0421 21:32:19.588735   63019 keypair.go:195] Issuing new certificate: "etcd-manager-ca-main"
I0421 21:32:19.611300   63019 keypair.go:195] Issuing new certificate: "etcd-manager-ca-events"
W0421 21:32:21.750329   63019 executor.go:139] error running task "WarmPool/nodes-us-east-1b.k8s.REDACTED" (9m56s remaining to succeed): ValidationError: AutoScalingGroup name not found - No such Auto Scaling group: nodes-us-east-1b.k8s.REDACTED
	status code: 400, request id: 89e245fc-f525-455d-869e-81e3d17f0e63
W0421 21:32:21.750415   63019 executor.go:139] error running task "WarmPool/nodes-us-east-1a.k8s.REDACTED" (9m56s remaining to succeed): ValidationError: AutoScalingGroup name not found - No such Auto Scaling group: nodes-us-east-1a.k8s.REDACTED
	status code: 400, request id: 9778cb2b-5244-4c44-a916-2e0758fdc84d
W0421 21:32:21.750432   63019 executor.go:139] error running task "WarmPool/master-us-east-1a.masters.k8s.REDACTED" (9m56s remaining to succeed): ValidationError: AutoScalingGroup name not found - No such Auto Scaling group: master-us-east-1a.masters.k8s.REDACTED
	status code: 400, request id: 9afc2c8c-b964-4548-b041-ef653a9ae2fe
I0421 21:32:21.750462   63019 executor.go:111] Tasks: 41 done / 90 total; 22 can run
```
and so on

**6. What did you expect to happen?**

No "error running task" log lines.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
```

[...]

PR #11332, opened 2021-04-27
Skip warmpool deletion when creating clusters

h3poteto:
Fixes: #11282 

At the moment, kOps calls `DeleteWarmPool` method when creating cluster. So it says error. 
I changed to skip this logic when create new clusters.


rifelpet:
duplicate of https://github.com/kubernetes/kops/pull/11331 ?

h3poteto:
Oh, I'm sorry, I missed it.
Exactly the same fix.

PR #11333, opened 2021-04-27
Add support for configuring Cilium enable-host-reachable-services.

bjhaid:
After upgrading Cilium to 1.8 via kops one of our clusters had a total
outage due to cilium reporting errors as below:

```
level=error msg="endpoint regeneration failed" containerID= datapathPolicyRevision=0 desiredPolicyRevision=1 endpointID=592 error="Failed to load tc filter: exit status 1" identity=40147 ipv4= ipv6= k8sPodName=/ subsys=endpoint
```

upon searching Cilium slack we found the below thread:

https://cilium.slack.com/archives/C1MATJ5U5/p1616400216167600

which recommended setting `enable-host-reachable-services` to true will
address the problems. We set the field and it fixed our issues too,
however we observed that kops does not have a means to configure this
hence this PR.

We will like to have this backported after it has been merged.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -3459,6 +3459,12 @@ spec:
                         description: 'EnableEncryption enables Cilium Encryption.
                           Default: false'
                         type: boolean
+                      enableHostReachableServices:
+                        description: 'EnableHostReachableServices configures Cilium
+                          to enable services to be reached from the host namespace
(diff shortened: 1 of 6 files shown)


olemarkus:
Thanks for the PR. This change will add the needed configuration flag, but it will not make any actual changes to the cilium manifests.

You can find the template you need to modify here: `upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.12-v1.9.yaml.template`

bjhaid:
thanks @olemarkus, addressed

olemarkus (approved):

PR #11335, opened 2021-04-27
Fix lifecycle hook naming

olemarkus:
Since tasks need to be unique, but we need to reuse the hookname across all ASGs, we distinguish between task and actual name of the hook


pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -100,7 +100,8 @@ func (b *AutoscalingGroupModelBuilder) Build(c *fi.ModelBuilderContext) error {
 				warmPoolTask.MaxSize = warmPool.MaxSize
 
 				if warmPool.EnableLifecycleHook {
-					name := "kops-warmpool"
+					hookName := "kops-warmpool"
+					name := fmt.Sprintf("%s-%s", hookName, ig.GetName())
 
 					lifecyleTask := &awstasks.AutoscalingLifecycleHook{
```
rifelpet, line 106: should HookName be passed in here?

olemarkus: yeah just caught this one. think we need some integration tests for this :)

rifelpet (approved):

PR #11349, opened 2021-04-28
Ignore detached nodes when doing validate cluster

rajatjindal:
Ran into this issue while doing rolling-update on our cluster today. The rolling updated was cancelled (for unrelated reasons) before it could complete.

### DIFF
--- a/pkg/validation/validate_cluster.go
+++ b/pkg/validation/validate_cluster.go
@@ -311,6 +311,10 @@ func (v *ValidationCluster) validateNodes(cloudGroups map[string]*cloudinstances
 					nodeExpectedToJoin = false
 				}
 
+				if member.Status == cloudinstances.CloudInstanceStatusDetached {
+					nodeExpectedToJoin = false
+				}
+
 				if nodeExpectedToJoin {
 					v.addError(&ValidationError{
 						Kind:          "Machine",
--- a/pkg/validation/validate_cluster_test.go
+++ b/pkg/validation/validate_cluster_test.go
@@ -840,3 +840,55 @@ func Test_ValidateBastionNodes(t *testing.T) {
 	})
 
 }
+
+func Test_ValidateDetachedNodesNotValidated(t *testing.T) {
+	groups := make(map[string]*cloudinstances.CloudInstanceGroup)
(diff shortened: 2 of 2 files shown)


rifelpet:
I'm curious why a detached node should be ignored from validation? the node and underlying instance should still be healthy so why not consider it for validation?

rajatjindal:
those nodes had no pods running on them and we ended up deleting them manually from AWS console.

i do not understand detached nodes logic 100%, may be @johngmyers can shed some light here. 

i think from AWS and kops perspective those nodes still existed, but from k8s perspective the nodes were gone?

johngmyers:
If a node is detached then it is on its way out of the cluster. A lack of being in the cluster is not a reason to hold up further disruption to the cluster.

/lgtm
Not approving due to employer proximity issues.

johngmyers:
To add more color, since detached nodes don't contribute to satisfying the IG target size check, the joined-to-cluster check is not not necessary to ensure the IG has enough capacity.

johngmyers:
Thinking about this further, we'll probably want to also exempt detached nodes from the node readiness and system-node-critical unready pod checks, at least for purposes of rolling update.

rajatjindal:
> Thinking about this further, we'll probably want to also exempt detached nodes from the node readiness and system-node-critical unready pod checks, at least for purposes of rolling update.

can that be a separate PR or should I make that change as part of this PR?

johngmyers:
It could be separate, but the intent here is "don't let extra sick nodes hold up a rolling update" and there is value in getting all of the "sick node" validation failures taken care of at once.

I have been thinking of a tighter integration between cluster validation and rolling update. There was previous work on ignoring "sick node" validation failures for nodes outside of the IG being updated. The next step would be to count sick nodes in the same IG against maxUnavailable, allowing `maxUnavailable - sick` drains to proceed in parallel.

rajatjindal:
Hi @rifelpet  do you think its reasonable to move ahead with the PR. if so, could you please help with the next steps?

thanks
Rajat Jindal

rifelpet (approved):
yes, I agree with the proposed logic now.

rajatjindal:
Thank you very much

PR #11352, opened 2021-04-29
Mount /run inside etcd-manager pods for systemd mounts

hakman:
xRef: https://github.com/kubernetes-sigs/etcdadm/pull/213

This will have no effect without the necessary binaries in the `etcd-manager` image.
`/run` is already available in `/rootfs/run` so there are no new security concerns.

/cc @olemarkus


olemarkus:
Good that you mentioned the bit about security :)

/lgtm

olemarkus (approved):

PR #11354, opened 2021-04-29
Allow cert-manager to be provisioned externally

codablock:
This PR allows provisioning of cert-manager externally. It actually leads to a no-op, but makes checks (`IsCertManagerEnabled`) succeed so that other dependent addons are deployable.

### DIFF
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -69,7 +69,23 @@ spec:
     defaultIssuer: yourDefaultIssuer
 ```
 
-**Warning: cert-manager only supports one installation per cluster. If you are already running cert-manager, you need to remove this installation prior to enabling this addon. As long as you are using v1 versions of the cert-manager resources, it is safe to remove existing installs and replace it with this addon**
+**Warning: cert-manager only supports one installation per cluster. If you are already running cert-manager, you need to
+either remove this installation prior to enabling this addon, or mark cert-manger as not being managed by kOps (see below).
+As long as you are using v1 versions of the cert-manager resources, it is safe to remove existing installs and replace it with this addon**
+
+##### Self-provisioned cert-manager
+{{ kops_feature_table(kops_added_default='1.21', k8s_min='1.16') }}
+
+The following cert-manager configuration allows provisioning cert-manager externally and allows all dependent plugins
+to be deployed. Please note that addons might run into errors until cert-manager is deployed.
+
+```yaml
+spec:
+  certManager:
+    enabled: true
+    managed: false
+```
+
 
 Read more about cert-manager in the [official documentation](https://cert-manager.io/docs/)
 
--- a/pkg/apis/kops/componentconfig.go
+++ b/pkg/apis/kops/componentconfig.go
@@ -937,6 +937,10 @@ type CertManagerConfig struct {
 	// Default: false
 	Enabled *bool `json:"enabled,omitempty"`
 
+	// Managed controls if cert-manager is manged and deployed by kOps.
+	// The deployment of cert-manager is skipped if this is set to false.
+	Managed *bool `json:"managed,omitempty"`
+
 	// Image is the docker container used.
 	// Default: the latest supported image for the specified kubernetes version.
 	Image *string `json:"image,omitempty"`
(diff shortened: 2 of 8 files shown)


pkg/apis/kops/validation/validation.go
```diff
@@ -1304,6 +1308,22 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
 	return allErrs
 }
 
+func validateCertManager(spec *kops.CertManagerConfig, fldPath *field.Path) (allErrs field.ErrorList) {
+	allErrs = field.ErrorList{}
+
+	if fi.BoolValue(spec.Enabled) && fi.BoolValue(spec.External) {
+		allErrs = append(allErrs, field.Forbidden(fldPath, "Cert manager must be either enabled or external but not both"))
+	}
+	if fi.BoolValue(spec.External) && fi.StringValue(spec.Image) != "" {
```
olemarkus, line 1317: This seems a bit overkill. I think we can just document that when `external` is set, any other flag is ignored.

pkg/apis/kops/componentconfig.go
```diff
@@ -930,6 +930,10 @@ type CertManagerConfig struct {
 	// Default: false
 	Enabled *bool `json:"enabled,omitempty"`
 
+	// External Tell kops that cert manager is deployed externally.
+	// Default: false
+	External *bool `json:"external,omitempty"`
```
olemarkus, line 935: Let me check if this naming is consistent with other structs where we allow this.

olemarkus (changes requested):

pkg/apis/kops/componentconfig.go
hakman: I think some other time we discussed about "Managed".

pkg/apis/kops/validation/validation.go
codablock: Removed the last 2 checks.

pkg/apis/kops/componentconfig.go
codablock: you mean setting "managed" to false?

hakman: Example: https://github.com/kubernetes/kops/blob/ba3914a30b41a986f687f25d51cff36951d03070/pkg/apis/kops/ntpconfig.go#L23

codablock: Changed the PR to do it the same way

docs/addons.md
```diff
@@ -69,7 +69,20 @@ spec:
     defaultIssuer: yourDefaultIssuer
 ```
 
-**Warning: cert-manager only supports one installation per cluster. If you are already running cert-manager, you need to remove this installation prior to enabling this addon. As long as you are using v1 versions of the cert-manager resources, it is safe to remove existing installs and replace it with this addon**
+**Warning: cert-manager only supports one installation per cluster. If you are already running cert-manager, you need to
+either remove this installation prior to enabling this addon, or mark cert-manger as not being managed by kOps (see below).
+As long as you are using v1 versions of the cert-manager resources, it is safe to remove existing installs and replace it with this addon**
+
+The following cert-manager configuration allows provisioning cert-manager externally and allows all dependent plugins
```
olemarkus, line 76: Can you add the version macro?
```{{ kops_feature_table(kops_added_default='1.21', k8s_min='1.16') }}```

codablock: I was not sure where to add this, I hope I got it right

olemarkus: Usually we just add a subheading and place the macro under.

codablock: Added another commit. I'm still unsure if this is how you expect it, if not, can you show me how you expect it?

olemarkus: Looks perfect

olemarkus (changes requested):
Can you squash? then this is ready to go.

codablock:
Squashed

olemarkus (approved):
Much obliged

kellycampbell:
Could this be cherry-picked back to 1.20? We're running into an error trying to upgrade from 1.19 to 1.20 due to metricsServer being enabled, but now certManager requirement is checked. We already have cert manager in the cluster and want to take more time to make sure all our certs are going to port over if we switch to the add-on. (I'm working around this for now by disabling metricsServer since this on a dev cluster.)

PR #11359, opened 2021-04-30
Split oidc_provider

olemarkus:
* one builder concerned with publishing issuer discovery metadata
* one builder concerned with creating aws oidc provider


hakman (approved):

hakman (approved):

pkg/model/BUILD.bazel
```diff
@@ -59,7 +61,6 @@ go_library(
 go_test(
     name = "go_default_test",
     srcs = ["bootstrapscript_test.go"],
-    data = glob(["tests/**"]),  #keep
```
hakman, line 62: This has to be added back :)

johngmyers: But verify-bazel passed?

hakman: This is manually added and cannot be verified by that, if I recall correctly.

PR #11361, opened 2021-04-30
Configure aws oidc provider

olemarkus:
Move VFS path and create flag for enabling AWS OIDC provider

/hold for #11359


pkg/apis/kops/cluster.go
```diff
@@ -207,6 +205,17 @@ type ClusterSpec struct {
 	ClusterAutoscaler *ClusterAutoscalerConfig `json:"clusterAutoscaler,omitempty"`
 	// WarmPool defines the default warm pool settings for instance groups (AWS only).
 	WarmPool *WarmPoolSpec `json:"warmPool,omitempty"`
+
+	// ServiceAccountIssuerDiscovery configures the OIDC Issuer for ServiceAccounts.
+	ServiceAccountIssuerDiscovery *OIDCIssuerConfig `json:"serviceAccountIssuerDiscovery,omitempty"`
```
johngmyers, line 210: I think the type should have the field name as a prefix.

johngmyers (changes requested):

pkg/apis/kops/cluster.go
olemarkus: I was wondering if we'd use this type for other properties too. But I guess not. I'll amend.

pkg/apis/kops/cluster.go
```diff
@@ -207,6 +205,17 @@ type ClusterSpec struct {
...
+}
+
+// OIDCIssuerConfig configures an OIDC Issuer.
+type OIDCIssuerConfig struct {
+	// DiscoveryStore is the VFS path to where OIDC Issuer Discovery metadata is stored.
+	DiscoveryStore string `json:"discoveryStore,omitempty"`
+	// EnableAWSOIDCProvider will provision an AWS OIDC provider that trusts the ServiceAccount Issuer
```
johngmyers, line 217:
```suggestion
	// EnableAWSOIDCProvider will provision an AWS OIDC provider that trusts the ServiceAccount Issuer.
```

pkg/model/iam/subject.go
```diff
@@ -101,12 +101,10 @@ func BuildNodeRoleSubject(igRole kops.InstanceGroupRole, enableLifecycleHookPerm
 // ServiceAccountIssuer determines the issuer in the ServiceAccount JWTs
 func ServiceAccountIssuer(clusterSpec *kops.ClusterSpec) (string, error) {
 	if featureflag.PublicJWKS.Enabled() {
```
johngmyers, line 103: Shouldn't this feature flag check be removed?

johngmyers (changes requested):

johngmyers:
Could one of the two integration tests with the PublicJWKS feature flag have it removed? Perhaps a followup PR.

pkg/model/iam/subject.go
olemarkus: This one got tricky to remove just like that. There is no flag that determines what the actual ServiceAccountIssuer is. It is coming in the next PR.

johngmyers (approved):

PR #11362, opened 2021-04-30
Use VFS as service account issuer if configured

olemarkus:
Think this is the closest path to both a test and getting rid of the PublicJWKS feature flag when using VFS.

@johngmyers don't remember if we concluded we wanted a flag to explicitly state if the user wants VFS or API server for OIDC, or if we just want to infer that like we do right now.


johngmyers:
My recollection was to infer.

Since we learned that newer k8s can have multiple service account issuers, we might just want to add them all. That requires figuring out a backwards-compatible API change, though–perhaps a second `serviceAccountIssuers` field which takes a slice?

cmd/kops/integration_test.go
```diff
@@ -291,21 +291,30 @@ func TestPrivateDns2(t *testing.T) {
...
 	// We have to use a fixed CA because the fingerprint is inserted into the AWS WebIdentity configuration.
-	newIntegrationTest("minimal.example.com", "public-jwks").
+	newIntegrationTest("minimal.example.com", "public-jwks-apiserver").
 		withCAKey().
 		withServiceAccountRole("dns-controller.kube-system", true).
 		runTestTerraformAWS(t)
 
 }
 
+func TestPublicJWKSVFS(t *testing.T) {
```
johngmyers, line 309: TestVFSServiceAccountIssuerDiscovery?

olemarkus:
> My recollection was to infer.
> 
> Since we learned that newer k8s can have multiple service account issuers, we might just want to add them all. That requires figuring out a backwards-compatible API change, though–perhaps a second `serviceAccountIssuers` field which takes a slice?

We still should only configure one AWS OIDC Provider. But we may say we only support using the VFS for that. Same with IRSA for our own addons.

johngmyers:
Ah, I was talking about what goes into componentconfig. But this takes more research. Later PR; infer VFS if configured for now.

johngmyers (approved):

PR #11364, opened 2021-05-01
skip detached nodes from readiness and system-*-critical checks

rajatjindal:
in continuation to PR: https://github.com/kubernetes/kops/pull/11349 and comment: https://github.com/kubernetes/kops/pull/11349#issuecomment-828658240, ignoring detached nodes from ready and system-*-critical checks as well.

### DIFF
--- a/pkg/validation/validate_cluster.go
+++ b/pkg/validation/validate_cluster.go
@@ -299,8 +299,11 @@ func (v *ValidationCluster) validateNodes(cloudGroups map[string]*cloudinstances
 		}
 
 		for _, member := range allMembers {
-			node := member.Node
+			if member.Status == cloudinstances.CloudInstanceStatusDetached {
+				continue
+			}
 
+			node := member.Node
 			if node == nil {
 				nodeExpectedToJoin := true
 				if cloudGroup.InstanceGroup.Spec.Role == kops.InstanceGroupRoleBastion {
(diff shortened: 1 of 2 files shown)


pkg/validation/validate_cluster.go
```diff
@@ -299,8 +299,11 @@ func (v *ValidationCluster) validateNodes(cloudGroups map[string]*cloudinstances
 		}
 
 		for _, member := range allMembers {
-			node := member.Node
+			if member.Status == cloudinstances.CloudInstanceStatusDetached {
+				continue
+			}
```
hakman, line 304: @olemarkus @johngmyers Should the bastion and warm pool checks be handled in the same way?

johngmyers: Bastions and warm instances will never join, so the node readiness and critical pod checks are never relevant for them.

johngmyers:
I'm going to walk back my previous comment. Nodes that aren't joined to the cluster either never previously contributed to capacity or are in the process of being terminated so will never contribute to capacity in the future. It is possible for these checks to fail intermittently, so such nodes might be needed to maintain the minimum capacity that rolling update is trying to maintain.

So I think we might want to think some more about what capacity invariant rolling update is trying to maintain and then come up with a design. In general we want to not reduce capacity by more than `maxUnavailable` than the ASG's target size, but we have exceptions.

Possible exceptions:

1. If we start off below target capacity due to unattached NeedsUpdate instances we can use the starting capacity instead of the target capacity.
2. If, assuming NeedsUpdate nodes that started off unhealthy stay so, we would otherwise deadlock, we can reduce capacity by one node.

If we can get rolling update to be able to drain unhealthy NeedsUpdate nodes, then we should prioritize the healthy NeedsUpdate nodes for detaching.

johngmyers:
Filed #11367 with an initial design.

PR #11369, opened 2021-05-02
Mark control-plane node for update when etcd manager config changes

hakman:
This is needed when using custom images or env vars with etcd-manager.


johngmyers (approved):
To save space in the userdata we should probably have a separate struct that gets hashed, with the hash put in the userdata.

hakman:
> To save space in the userdata we should probably have a separate struct that gets hashed, with the hash put in the userdata.

@johngmyers This is not that bad as it's not added to user data unless someone wants something different from defaults.
User data compression also is pretty good and should be more than enough for anyone that needs to reduce user data size.

PR #11371, opened 2021-05-03
Refactor terraform writing

johngmyers:
When working on #9621 I tried pushing Terraform generation into VFS but ran into an import loop. In order to break that import loop this breaks out a separate `terraformWriter.TerraformWriter` API in a separate package so that things that generate Terraform don't have to depend on the `fi` package.


rifelpet (approved):
This looks good to me, thanks for splitting the commits up

PR #11387, opened 2021-05-06
Enable reading shared config when possibly from CLI

johngmyers:
Remove the annoyance of having to set `AWS_SDK_LOAD_CONFIG=1` when using assumed roles.


justinsb:
This is a good idea, though I don't think we should backport it so we have time to vet it ourselves.

Can you also send a PR for the 1.22 release notes?  (It's an enhancement, but people should be made aware of it in case they are surprised)

/approve
/lgtm

Issue #11370, opened 2021-05-02
Failed Kops upgrade 1.18.3 => 1.19.2 - Issuing new certificate: "master"


jlsan92:
After running `kops update cluster ...` I notice `Issuing new certificate: "master"`. Then after a couple of seconds, `kubectl` and even `kops rolling-update cluster` fail with 

```
Unable to connect to the server: x509: certificate is valid for service-account, not api.my.cluster.net
```

To be able to connect again, I need to run `kops update cluster` again but with Kops 1.18.3 (The version I want to move from). Same behaviour, a couple of seconds after then I can connect back to my cluster's API. Seems impossible to upgrade to Kops 1.19. Not sure if `kops rolling-update cluster --cloudonly` would return a healthy cluster.

Not sure if I'm missing an important step but [Required Actions](https://kops.sigs.k8s.io/releases/1.19-notes/#required-actions) don't mention anything regarding certificates.

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

`Version 1.19.2 (git-e288df46e173ba8ce44ac52502283206c0d211ee)`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

```
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.18", GitCommit:"6f6ce59dc8fefde25a3ba0ef0047f4ec6662ef24", GitTreeState:"clean", BuildDate:"2021-04-15T03:23:41Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

```
kops update cluster --name $KOPS_CLUSTER_NAME --lifecycle-overrides IAMRole=ExistsAndWarnIfChanges,IAMRolePolicy=ExistsAndWarnIfChanges,IAMInstanceProfileRole=ExistsAndWarnIfChanges --out=config/terraform --target=terraform
```

**5. What happened after the commands executed?**

Normal `kops update cluster` output except for a new line => `57165 vfs_castore.go:590] Issuing new certificate: "master"`

**6. What did you expect to happen?**

Normal `kops update cluster` behaviour without rotating master's certs

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
[...]

PR #11388, opened 2021-05-06
Backport rename of service-account key to 1.20

johngmyers:
Addresses #11370


justinsb:
Like #11390, this looks good, but I propose we get an upgrade e2e periodic test in asap

/hold

hakman:
@justinsb should be covered by @rifelpet's new upgrade test:
https://testgrid.k8s.io/kops-misc#kops-aws-kops-upgrade

I see this in logs, which I am not sure if it's the expected change:
```
  Keypair/service-account
  	Signer              	 id:cn=service-account -> <nil>
  	LegacyFormat        	 true -> false
```

johngmyers:
@hakman The LegacyFormat change is saying that it will upgrade the format (by copying it from the "master" slot to "service-account". For the "Signer" field, `nil` means "self-signed" so it is not an actual change.

johngmyers:
For an upgrade from 1.18 or earlier, the signer would change from "ca" to being self-signed.

justinsb:
Reproducing #11390 is non-trivial (which is a good thing, as it means this isn't _as_ critical as it first appears)!

/hold cancel

/approve
/lgtm

PR #11393, opened 2021-05-06
Add elasticloadbalancing:ModifyTargetGroupAttributes to aws lb controller

### LINKED ISSUE #11297: AWS LoadbalancerController IAM is missing ModifyTargetGroupAttributes
 this information.**
```
kops version
Version 1.20.0 (git-8ea83c6d233a15dacfcc769d4d82bea3f530cf72)
```

 a `kops` flag.**

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:31:21Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.6", GitCommit:"8a62859e515889f07e3e3be6a1080413f17cf2c3", GitTreeState:"clean", BuildDate:"2021-04-15T03:19:55Z", GoVersion:"go1.15.10", Compiler:"gc", Platform:"linux/amd64"}
```

AWS with the LoadbalancerController Addon

try to create a service loadbalancer with targetGroup modifications

```yaml
..
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: "nlb-ip"
    service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: deregistration_delay.timeout_seconds=1
    service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: deregistration_delay.connection_termination.enabled=true
```

```json
{"level":"error","ts":1619172828.867188,"logger":"controller","msg":"Reconciler error","controller":"service","name":"service-nlb-ip","namespace":"test","error":"AccessDenied: User: arn:aws:sts::xxxxxxx:assumed-role/masters.meek/i-xxxxxxx is not authorized to perform: elasticloadbalancing:ModifyTargetGroupAttributes on resource: arn:aws:elasticloadbalancing:eu-west-1:xxxxxxx:targetgroup/k8s-test-servicet-ce1780c640/cb7ab39900cbf1d0\n\tstatus code: 403, request id: xxxxxx-a9fc-4fb2-894d-70144d367ab2"}
```

it to work

quickfix for everyone who runs into the same thing

```yaml
spec:
  additionalPolicies:
    master: |
      [
         {
            "Effect": "Allow",
            "Action": "elasticloadbalancing:ModifyTargetGroupAttributes",
            "Resource": "*"
         }
      ]
```

olemarkus:
Fixes #11297

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -759,6 +759,7 @@ func AddAWSLoadbalancerControllerPermissions(p *Policy, resource stringorslice.S
 				"elasticloadbalancing:DescribeRules",
 				"elasticloadbalancing:DescribeTargetHealth",
 				"elasticloadbalancing:DescribeListenerCertificates",
+				"elasticloadbalancing:ModifyTargetGroupAttributes",
 				"elasticloadbalancing:CreateRule",
 			),
 			Resource: resource,
--- a/tests/integration/update_cluster/aws-lb-controller/data/aws_iam_role_policy_aws-load-balancer-controller.kube-system.sa.minimal.example.com_policy
+++ b/tests/integration/update_cluster/aws-lb-controller/data/aws_iam_role_policy_aws-load-balancer-controller.kube-system.sa.minimal.example.com_policy
@@ -121,6 +121,7 @@
         "elasticloadbalancing:DescribeRules",
         "elasticloadbalancing:DescribeTargetHealth",
         "elasticloadbalancing:DescribeListenerCertificates",
+        "elasticloadbalancing:ModifyTargetGroupAttributes",
         "elasticloadbalancing:CreateRule"
       ],
       "Effect": "Allow",


peter-svensson:
When using the annotation 
`alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'`

The following permissions are needed in order to setup HTTPs certificate:

```json
{
    "Effect": "Allow",
    "Action": "acm:DescribeCertificate",
    "Resource": "arn:aws:acm:*:<accountid>:certificate/*"
},
{
    "Effect": "Allow",
    "Action": "acm:ListCertificates",
    "Resource": "*"
}
```

peter-svensson:
When changing an ingress I also need to set 

```
          "arn:aws:elasticloadbalancing:*:<account>:listener-rule/app/*/*/*/*",
          "arn:aws:elasticloadbalancing:*:<account>:listener-rule/net/*/*/*/*"
     
```

olemarkus:
I think the ACM part needs more thinking. Maybe require that the certs have certain tags for them to be used by a cluster.

peter-svensson:
> I think the ACM part needs more thinking. Maybe require that the certs have certain tags for them to be used by a cluster.

I'd say the same might be applicable to the ALBs created by the ingress controller? But that might be harder to achieve?

olemarkus:
> > I think the ACM part needs more thinking. Maybe require that the certs have certain tags for them to be used by a cluster.
> 
> I'd say the same might be applicable to the ALBs created by the ingress controller? But that might be harder to achieve?

ALBs created by the LB controller has the `elbv2.k8s.aws/cluster: <cluster>` tag. But those are created by the cluster.
ACM certs are never created by the cluster itself, so we need to use tags so that ALB controller suddenly doesn't get to use any cert for any purpose.

This is why we don't copy/paste the permissions from the LB controller docs.. they are just way to open for anyone security minded.

peter-svensson:
> > > I think the ACM part needs more thinking. Maybe require that the certs have certain tags for them to be used by a cluster.
> > 
> > 
> > I'd say the same might be applicable to the ALBs created by the ingress controller? But that might be harder to achieve?
> 
> ALBs created by the LB controller has the `elbv2.k8s.aws/cluster: <cluster>` tag. But those are created by the cluster.
> ACM certs are never created by the cluster itself, so we need to use tags so that ALB controller suddenly doesn't get to use any cert for any purpose.
> 
> This is why we don't copy/paste the permissions from the LB controller docs.. they are just way to open for anyone security minded.

I can image that. 
What  I meant was the the permission that we set with kops when enabling the AWS ALB addon should reflect that as well. So the permission to delete/update rules for example should be only for the ALBs with the correct tags as well.

olemarkus:
ah yes. more resources could have 
```
		Condition: Condition{
			"StringEquals": map[string]string{
				"ec2:ResourceTag/elbv2.k8s.aws/cluster": clusterName,
			},
		},
```

hakman (approved):

PR #11396, opened 2021-05-06
Remove the PublicJWKS feature flag

johngmyers:


johngmyers:
@rifelpet There's a kops-grid-scenario-public-jwks test that depends on this. Can it be changed to remove this feature flag?

rifelpet:
Yes, we can remove this: https://github.com/kubernetes/test-infra/blob/afb819233b06d912ae4bf111ee3e19e77ab4c72f/config/jobs/kubernetes/kops/build_jobs.py#L660

We could also add `--override=cluster.spec.serviceAccountIssuerDiscovery.discoveryStore=s3://k8s-kops-prow/...` to the job.

PR #11401, opened 2021-05-07
Add another update cluster dryrun to upgrade tests

rifelpet:
This way we can confirm that no changes are reported after update cluster --yes


justinsb:
Thanks - I'll add this line to mine as well :-)

/approve
/lgtm

PR #11403, opened 2021-05-07
Create upgrade_ab e2e scenario for complex upgrades

justinsb:
The intent is that we'll be able to perform arbitrary
upgrades/downgrades; creating a new scenario to avoid breaking
existing tests.

### DIFF
--- a/tests/e2e/scenarios/upgrade-ab/run-test.sh
+++ b/tests/e2e/scenarios/upgrade-ab/run-test.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+echo "CLOUD_PROVIDER=${CLOUD_PROVIDER}"
+
+if [ -z "$KOPS_VERSION_A" ] || [ -z "$K8S_VERSION_A" ] || [ -z "$KOPS_VERSION_B" ] || [ -z "$K8S_VERSION_B" ]; then
+  >&2 echo "must set all of KOPS_VERSION_A, K8S_VERSION_A, KOPS_VERSION_B, K8S_VERSION_B env vars"
+  exit 1
+fi
+
+WORKDIR=$(mktemp)
+
+KOPS_A=${WORKDIR}/kops-${KOPS_VERSION_A}
+wget -O "${KOPS_A}" "https://github.com/kubernetes/kops/releases/download/$KOPS_VERSION_A/kops-$(go env GOOS)-$(go env GOARCH)"
+chmod +x "${KOPS_A}"
+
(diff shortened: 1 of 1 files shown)


justinsb:
/assign @rifelpet 

I didn't realize you were actively working on these, so this may overlap with your work!

tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -0,0 +1,89 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+echo "CLOUD_PROVIDER=${CLOUD_PROVIDER}"
+
+if [ -z "$KOPS_VERSION_A" ] || [ -z "$K8S_VERSION_A" ] || [ -z "$KOPS_VERSION_B" ] || [ -z "$K8S_VERSION_B" ]; then
+  >&2 echo "must set all of KOPS_VERSION_A, K8S_VERSION_A, K8S_VERSION_B, K8S_VERSION_B env vars"
```
rifelpet, line 24:
```suggestion
  >&2 echo "must set all of KOPS_VERSION_A, K8S_VERSION_A, KOPS_VERSION_B, K8S_VERSION_B env vars"
```

tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -0,0 +1,89 @@
...
+KOPS_A=${WORKDIR}/kops-${KOPS_VERSION_A}
+wget -O "${KOPS_A}" "https://github.com/kubernetes/kops/releases/download/$KOPS_VERSION_A/kops-$(go env GOOS)-$(go env GOARCH)"
+chmod +x "${KOPS_A}"
+
+KOPS_B=${WORKDIR}/kops-${KOPS_VERSION_B}
+if [[ "${KOPS_VERSION_B}" == "source" ]]; then
+  export KOPS_FEATURE_FLAGS="SpecOverrideFlag,${KOPS_FEATURE_FLAGS:-}"
+  REPO_ROOT=$(git rev-parse --show-toplevel);
+
+  cp "${REPO_ROOT}/bazel-bin/cmd/kops/linux-amd64/kops" "${KOPS_B}"
```
rifelpet, line 39: side note: it would be great if our scenario scripts could run on more than linux/amd64 but the make task used by prow only builds linux-amd64 builds of the kops cli. Maybe we could extend it to always build the local GOOS / GOARCH as well.

tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -0,0 +1,89 @@
...
+		--kubernetes-version="${K8S_VERSION_A}" \
+		--create-args="--networking calico"
+
+
+"${KOPS_B}" set cluster "${CLUSTER_NAME}" "cluster.spec.kubernetesVersion=${K8S_VERSION_B}"
+
+"${KOPS_B}" update cluster
+"${KOPS_B}" update cluster --admin --yes
+# Verify no additional changes
+"${KOPS_B}" update cluster
```
rifelpet, line 75: other side note: it would be convenient if we had an `update cluster` flag that would set the exit code in a way that would fail the test if there were additional changes.

tests/e2e/scenarios/upgrade-ab/run-test.sh
justinsb: Good point.  Could we also just build it here (in the script)?

AFAIK all of our test runners are amd64 anyway (at the moment), so might not be a P0?

justinsb:
> this is more flexible than what I came up with so I'm happy to replace the one I just added with this!

It's basically a rename of your script with a few tweaks (primarily env vars).  I hadn't realized that your upgrade script was recent, so I didn't feel like I should replace it.  But if you want to we can, or we can get this one working and then move the existing upgrade test to it and then remove the existing ones...  whatever you think best!

hakman:
👍

hakman (approved):

PR #11404, opened 2021-05-07
Create kops-specific e2e test

justinsb:
Admittedly a lot of overhead to set us up with a nice abstraction.


justinsb:
/assign @rifelpet 

Do we have any kops e2e specific tests yet?  Is this the right place to put these?

I have some more tests locally which I want to push up after this, hence the overkill levels of abstraction!

rifelpet:
We do not have any kops-specific e2e tests yet, I think tests/e2e/tests is a fine place to put them. I'll admit i was a bit confused by the term `fluent`, i originally thought this was referring to fluentd. What is `fluent`'s intended usage here?

PR #11411, opened 2021-05-08
Set the output base for fitask

hakman:
```
> fitask --help
  -o, --output-base string               Output base; defaults to $GOPATH/src/ or ./ if $GOPATH is not set.
```

/cc @olemarkus @rifelpet

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ UPLOAD_DEST?=$(S3_BUCKET)
 GCS_LOCATION?=gs://must-override
 GCS_URL=$(GCS_LOCATION:gs://%=https://storage.googleapis.com/%)
 LATEST_FILE?=latest-ci.txt
-GOPATH_1ST:=$(shell go env | grep GOPATH | cut -f 2 -d \")
+GOPATH_1ST:=$(shell go env | grep GOPATH | cut -f 2 -d '"' | sed 's/ /\\ /g')
 UNIQUE:=$(shell date +%s)
 BUILD=$(KOPS_ROOT)/.build
 LOCAL=$(BUILD)/local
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -152,7 +152,8 @@ ${KOPS}:
 codegen:
 	go install k8s.io/kops/upup/tools/generators/...
 	${GOPATH_1ST}/bin/fitask --input-dirs k8s.io/kops/upup/pkg/fi/... \
-	    --go-header-file "hack/boilerplate/boilerplate.generatego.txt"
+	    --go-header-file "hack/boilerplate/boilerplate.generatego.txt" \
+	    --output-base ${GOPATH_1ST}/src/k8s.io/kops/
```
olemarkus, line 156: Maybe quote this one too?

hakman: Considering that the fitask path is unquoted too, a better solution was to just escape the spaces at source.
Also, the quoting on the header file path was not very useful.

Makefile
```diff
@@ -151,8 +151,19 @@ ${KOPS}:
 .PHONY: codegen
 codegen:
 	go install k8s.io/kops/upup/tools/generators/...
-	${GOPATH_1ST}/bin/fitask --input-dirs k8s.io/kops/upup/pkg/fi/... \
-	    --go-header-file "hack/boilerplate/boilerplate.generatego.txt"
+	${GOPATH_1ST}/bin/fitask \
+		--input-dirs k8s.io/kops/upup/pkg/fi/... \
+		--go-header-file hack/boilerplate/boilerplate.generatego.txt \
+		--output-base ${GOPATH_1ST}/src/k8s.io/kops/
```
rifelpet, line 157: We're trying to reduce our reliance on `GOPATH`, could we use `KOPS_ROOT` instead? It would probably be good to compile the fitask binary into `_output/bin` to match the `TOOLS_BIN` references in `hack/`.

hakman: Good point. Done.

rifelpet (approved):

Issue #8432, opened 2020-01-29
IPv6 support


aojea:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

With the graduation from IPv6 to beta in Kubernetes https://github.com/kubernetes/enhancements/issues/508#issuecomment-574126795
the project has a CI that tests conformance and e2e in a synthetic environment with KIND.

This guarantees a minimum of quality and reduces the risk of regression, however, it needs more user validation. In order to achieve this is important that the lifecycle projects support it allowing users to deploy IPv6 Kubernetes clusters.

**2. Feel free to provide a design supporting your feature request.**

As of today, only 2 cloud providers support IPv6, since AWS is the more mature one and Azure is working on supporting IPv6 with its cluster management tool it's reasonable to implement IPv6 first in AWS and iterate over it.

I have done a sketch of an API implementation for AWS for the cluster-API project
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1370

The setup of the infrastructure, VMs, VPCs, ... is working, the missing part is the configuration of the Kubernetes components.

In order to configure IPv6 in Kubernetes you just have to configure the components with IPv6 addresses, which is not an opt-in feature.

johngmyers:
It would be helpful to distinguish between dual stack versus v6-only. I presume you mean the former, as AWS doesn't support the latter.

aojea:
Right, sorry, this is a bit confusing.

Let me give some context, dual-stack is still alpha in k8s, but IPv6 has improved a lot and the goal is to graduate it to beta with 1.18. In order to move Ipv6 to GA it needs more user validation, and the best way is making it accessible to users, hence this request.

If the infra where the cluster is running is dual-stack, like in AWS,  the cluster can be IPv4 only, IPv6 only or dual-stack, it will just use the IP family he has configured. I mean, IPv6 only can run over a dual-stack infrastructure, but no the other way around.

olemarkus:
Would be interesting to support ipv6 with kops, but provisioning VPC/subnet with ipv6 is somewhat painful to do with the way kops works.

fvasco:
A first step might be to support IPv6 on already IPv6 configured subnet (my case).

hakman:
This seems to be a recurring request and may be worth some attention.
@olemarkus @rifelpet I thing you did some thinking on this topic, maybe you can share your thoughts.
/reopen
/remove-lifecycle rotten
/lifecycle frozen

[...]

PR #11442, opened 2021-05-09
Add initial support for configuring IPv6 with AWS

hakman:
Inspired by https://github.com/kubernetes/kops/pull/11163.

At the moment, the IPv6 feature configures the following AWS components:
* VPC:
  * always requests the IPv6 CIDR from AWS
  * cannot be disabled
* Routes:
  * always creates the default IPv6 route (for public topologies)
* Security Group:
  * always allows IPv6 egress
  * adds / removes entries based on `spec.kubernetesApiAccess` and `spec.sshAccess`
* Subnets:
  * users can assign / remove an IPV6 CIDR (but cannot change it)
* Launch Templates:
  * IPv6 addresses are assigned to instances based on subnet setting, which allows marking instances as needing update

TODO:
- [ ] docs, docs, docs, ... 😄 
- [ ] allow configuring Kubernetes IPv4/IPv6 dual-stack:
  * https://kubernetes.io/docs/concepts/services-networking/dual-stack
- [ ] set subnet CIDR based on VPC CIDR on cluster creation:
  * https://www.terraform.io/docs/language/functions/cidrsubnet.html
  * https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-cidr.html
- [ ] allow setting API NLB to dual-stack
- [ ] create Egress-only Internet Gateway for private topologies
- [x] handle path MTU discovery
  * https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/network_mtu.html#path_mtu_discovery
  * AWS doesn't allow setting type and code for ICMPv6, which may be an issue

Ref: #8432 

/cc @rifelpet @johngmyers @olemarkus


johngmyers (changes requested):
Things are changing out from under my review, so submitting this early. Will review more later.

upup/pkg/fi/cloudup/awstasks/vpc.go
```diff
@@ -253,6 +275,7 @@ type terraformVPC struct {
 	CIDR               *string           `json:"cidr_block,omitempty" cty:"cidr_block"`
 	EnableDNSHostnames *bool             `json:"enable_dns_hostnames,omitempty" cty:"enable_dns_hostnames"`
 	EnableDNSSupport   *bool             `json:"enable_dns_support,omitempty" cty:"enable_dns_support"`
+	EnableIPv6         *bool             `json:"assign_generated_ipv6_cidr_block,omitempty" cty:"assign_generated_ipv6_cidr_block"`
 	Tags               map[string]string `json:"tags,omitempty" cty:"tags"`
 }
```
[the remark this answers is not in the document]
johngmyers: Seems it's handled by the separate task. Perhaps a comment would be appropriate?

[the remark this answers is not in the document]
hakman: Agreed, will add more comments before removing the WIP. Only terraform is done in the VPC task.

pkg/apis/kops/cluster.go
```diff
@@ -621,8 +621,10 @@ const (
 type ClusterSubnetSpec struct {
 	// Name is the name of the subnet
 	Name string `json:"name,omitempty"`
-	// CIDR is the network cidr of the subnet
+	// CIDR is the network CIDR of the subnet.
 	CIDR string `json:"cidr,omitempty"`
+	// IPv6CIDR is the network CIDR of the subnet.
+	IPv6CIDR string `json:"ipv6cidr,omitempty"`
```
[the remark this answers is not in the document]
hakman: No idea, but `ipv6CIDR looks odd.

pkg/apis/kops/cluster.go
```diff
@@ -621,8 +621,10 @@ const (
 type ClusterSubnetSpec struct {
 	// Name is the name of the subnet
 	Name string `json:"name,omitempty"`
-	// CIDR is the network cidr of the subnet
+	// CIDR is the network CIDR of the subnet.
 	CIDR string `json:"cidr,omitempty"`
+	// IPv6CIDR is the network CIDR of the subnet.
```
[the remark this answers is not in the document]
hakman: Already done :)

pkg/apis/kops/cluster.go
```diff
@@ -621,8 +621,10 @@ const (
 type ClusterSubnetSpec struct {
 	// Name is the name of the subnet
 	Name string `json:"name,omitempty"`
-	// CIDR is the network cidr of the subnet
+	// CIDR is the network CIDR of the subnet.
```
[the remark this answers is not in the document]
hakman: Already done :)

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -155,10 +165,14 @@ func (s *Subnet) CheckChanges(a, e, changes *Subnet) error {
...
 		}
 		if changes.AvailabilityZone != nil {
-			errors = append(errors, fi.FieldIsImmutable(a.AvailabilityZone, e.AvailabilityZone, fieldPath.Child("AvailabilityZone")))
+			errors = append(errors, fi.FieldIsImmutable(e.AvailabilityZone, a.AvailabilityZone, fieldPath.Child("AvailabilityZone")))
 		}
 		if changes.CIDR != nil {
-			errors = append(errors, fi.FieldIsImmutable(a.CIDR, e.CIDR, fieldPath.Child("CIDR")))
+			errors = append(errors, fi.FieldIsImmutable(e.CIDR, a.CIDR, fieldPath.Child("CIDR")))
+		}
+		if changes.IPv6CIDR != nil && a.IPv6CIDR != nil {
```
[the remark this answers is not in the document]
hakman: Agreed, but I would like to keep this as small as possible and improve in future PRs.

tests/integration/update_cluster/minimal-ipv6/in-v1alpha2.yaml
```diff
@@ -35,6 +35,7 @@ spec:
     nodes: public
   subnets:
   - cidr: 172.20.32.0/19
+    ipv6cidr: 2a05:d014:000:100::/64
```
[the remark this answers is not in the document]
hakman: It can, if you already have the VPC created, or shared VPC.

pkg/model/awsmodel/network.go
```diff
@@ -178,12 +189,19 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
 				RouteTable:      publicRouteTable,
 				InternetGateway: igw,
 			})
+			c.AddTask(&awstasks.Route{
+				Name:            fi.String("ipv6-default"),
+				Lifecycle:       b.Lifecycle,
+				IPv6CIDR:        fi.String("::/0"),
+				RouteTable:      publicRouteTable,
+				InternetGateway: igw,
+			})
```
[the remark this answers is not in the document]
hakman: I don't think adding and removing routes or security groups is very tested.
I assume that if we allocate an IPv6 block by default, we don't mind adding a few routes and security groups to allow IPv6 too.

tests/integration/update_cluster/aws-lb-controller/kubernetes.tf
```diff
@@ -506,12 +506,18 @@ resource "aws_launch_template" "nodes-minimal-example-com" {
   user_data = filebase64("${path.module}/data/aws_launch_template_nodes.minimal.example.com_user_data")
 }
 
-resource "aws_route" "route-0-0-0-0--0" {
+resource "aws_route" "route-ipv4-default" {
```
rifelpet, line 509: This change means that existing clusters will have their route deleted and recreated which would result in a network outage. I've tried to make this exact change before and terraform ended up failing because it tried to create the new route before deleting the old route, causing the `apply` to fail and having to rerun `apply` a second time to create the replacement route. During that period of time the route was missing.

Alternatively we document yet another `terraform state mv` that is needed when upgrading kops

hakman: Ah... well.... I guess the IPv6 one will be named ----0 for Terraform.

pkg/apis/kops/validation/validation.go
```diff
@@ -410,6 +410,10 @@ func validateSubnet(subnet *kops.ClusterSubnetSpec, fieldPath *field.Path) field
 	if subnet.CIDR != "" {
 		allErrs = append(allErrs, validateCIDR(subnet.CIDR, fieldPath.Child("cidr"))...)
 	}
+	// IPv6CIDR
+	if subnet.IPv6CIDR != "" {
+		allErrs = append(allErrs, validateCIDR(subnet.IPv6CIDR, fieldPath.Child("ipv6cidr"))...)
```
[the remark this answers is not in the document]
hakman: Added extra validation for IPv6.
CIDR is mandatory for AWS, but not sure how other cloud providers work.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -155,10 +165,14 @@ func (s *Subnet) CheckChanges(a, e, changes *Subnet) error {
...
+			klog.Infof("Changes in IPv6CIDR: %q - %q", aws.StringValue(a.IPv6CIDR), aws.StringValue(e.IPv6CIDR))
```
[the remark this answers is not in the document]
hakman: Was trying to debug something. Not needed anymore.

pkg/apis/kops/validation/validation.go
```diff
@@ -410,6 +410,10 @@ func validateSubnet(subnet *kops.ClusterSubnetSpec, fieldPath *field.Path) field
 	if subnet.CIDR != "" {
 		allErrs = append(allErrs, validateCIDR(subnet.CIDR, fieldPath.Child("cidr"))...)
 	}
+	// IPv6CIDR
+	if subnet.IPv6CIDR != "" {
```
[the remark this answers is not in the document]
hakman: Done.

pkg/apis/kops/validation/validation.go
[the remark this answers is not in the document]
johngmyers: Presumably mandatory for GCE. Might not be mandatory for Azure.

upup/pkg/fi/cloudup/awstasks/subnet_test.go
```diff
@@ -137,6 +137,186 @@ func TestSubnetCreate(t *testing.T) {
...
+		}
+		cidr1 := &VPCAmazonIPv6CIDRBlock{
+			VPC:       vpc1,
+			CIDRBlock: s("2001:db8::/56"),
+		}
+		subnet1 := &Subnet{
+			Name:     s("subnet1"),
+			VPC:      vpc1,
+			CIDR:     s("172.20.1.0/24"),
+			IPv6CIDR: s("cidrsubnet-8-1"),
```
hakman, line 251: @johngmyers @justinsb This is my proposed way of describing a subnet CIDR, using the network CIDR as a base.

hakman: https://www.terraform.io/docs/language/functions/cidrsubnet.html

hakman: I also don't mind to split this in a separate PR and continue the discussion there.

johngmyers: I'd prefer bikeshedding a syntax that looks more like a function call. Perhaps something like `cidrsubnet(8,1)`.

I think a separate PR would help as the rest is a bit more baked.

hakman: OK, separate PR it is. :)

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -301,85 +302,118 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
 					Name:          fi.String("icmp-pmtu-api-elb-" + cidr),
 					Lifecycle:     b.SecurityLifecycle,
 					CIDR:          fi.String(cidr),
 					FromPort:      fi.Int64(3),
 					Protocol:      fi.String("icmp"),
-					SecurityGroup: masterGroup.Task,
+					SecurityGroup: lbSG,
 					ToPort:        fi.Int64(4),
 				})
+			}
```
johngmyers, line 358: Don't we need to open type 2 ICMPv6 traffic for IPv6 PMTU discovery?

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -301,85 +302,118 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+					c.AddTask(&awstasks.SecurityGroupRule{
+						Name:          fi.String("icmp-pmtu-api-elb-" + cidr),
+						Lifecycle:     b.SecurityLifecycle,
+						CIDR:          fi.String(cidr),
+						FromPort:      fi.Int64(3),
+						Protocol:      fi.String("icmp"),
+						SecurityGroup: masterGroup.Task,
+						ToPort:        fi.Int64(4),
+					})
+				}
```
johngmyers, line 399: Same here for IPv6 PMTU discovery.

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -211,6 +212,11 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.ModelBuilde
 		case kops.SubnetTypePrivate:
 			lt.AssociatePublicIP = fi.Bool(false)
 		}
+
+		// @step: add an IPv6 address
+		if subnets[0].IPv6CIDR != "" {
```
johngmyers, line 217: Shouldn't we do this if any of the subnets have an IPv6 CIDR? `GatherSubnets` enforces that there can't be multiple subnet types, but I think it is possible for some but not all subnets to have an IPv6 address.

pkg/model/awsmodel/bastion.go
```diff
@@ -77,14 +78,26 @@ func (b *BastionModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+				Lifecycle:     b.SecurityLifecycle,
+				SecurityGroup: src.Task,
+				Egress:        fi.Bool(true),
+				CIDR:          fi.String("0.0.0.0/0"),
+			}
+			AddDirectionalGroupRule(c, t)
+		}
+		{
+			t := &awstasks.SecurityGroupRule{
+				Name:          fi.String("ipv6-bastion-egress" + src.Suffix),
```
johngmyers, line 93: Consider making one `SecurityGroupRule` that allows egress for both IPv4 and IPv6 CIDRs.

johngmyers (changes requested):

pkg/model/awsmodel/autoscalinggroup.go
hakman: Good point, done.

pkg/model/awsmodel/api_loadbalancer.go
johngmyers: ...or even open all of ICMPv6?

hakman: I tried doing this manually and got into some trouble with the AWS web UI, which seems to not allow setting custom ICMPv6, but seems to work from API.

pkg/model/awsmodel/api_loadbalancer.go
hakman: Should be good now.

pkg/apis/kops/validation/validation.go
```diff
@@ -312,7 +313,11 @@ func validateCIDR(cidr string, fieldPath *field.Path) field.ErrorList {
 		if !strings.Contains(cidr, "/") {
 			ip := net.ParseIP(cidr)
 			if ip != nil {
-				detail += fmt.Sprintf(" (did you mean \"%s/32\")", cidr)
+				if ip.To4() != nil {
```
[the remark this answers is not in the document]
hakman: Done. Both `To4()` and `To6()` are not doing that much, so checking for ":" seems a good compromise.

upup/pkg/fi/utils/cidr.go
```diff
@@ -0,0 +1,53 @@
...
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"net"
+)
+
+func IsIPv4CIDR(cidr string) bool {
```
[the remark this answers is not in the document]
hakman: Done.

pkg/model/awsmodel/bastion.go
hakman: Seems a good idea for a future PR. Would have to check first if what is simplified is worth the added complexity.

hakman:
I will squash / re-stage some of the commits after addressing all review comments and getting the OK.
/hold

pkg/model/awsmodel/bastion.go
johngmyers: It would avoid changing the name of the rule, so an upgrade wouldn't delete and recreate.

hakman: The rule name doesn't change actually, because it uses `AddDirectionalGroupRule`, that overwrites the name for the task based on rule components. This is why we don't see any changes in TF or CF output either.

upup/pkg/fi/cloudup/awstasks/vpc.go
```diff
@@ -41,6 +41,11 @@ type VPC struct {
 	EnableDNSHostnames *bool
 	EnableDNSSupport   *bool
 
+	// Used only for Terraform rendering.
+	// Direct and CloudFormation rendering is handled via the VPCAmazonIPv6CIDRBlock task
```
johngmyers, line 45: nit: should start with an `// AmazonIPv6 is ...` comment.

upup/pkg/fi/cloudup/awstasks/vpc.go
```diff
@@ -41,6 +41,11 @@ type VPC struct {
...
+	AmazonIPv6 *bool
+	IPv6CIDR   *string
```
johngmyers, line 47: These two should probably go before `EnableDNSHostnames`.

upup/pkg/fi/cloudup/awstasks/vpcamazonipv6cidrblock.go
```diff
@@ -0,0 +1,149 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
johngmyers, line 2: nit: 2021

upup/pkg/fi/cloudup/awstasks/vpcamazonipv6cidrblock.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return nil
+}
+
+func (_ *VPCAmazonIPv6CIDRBlock) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *VPCAmazonIPv6CIDRBlock) error {
+	shared := aws.BoolValue(e.Shared)
+	if shared && a == nil {
+		// VPC not owned by kOps, no changes will be applied
+		// Verify that the Amazon IPv6 provided CIDR block was found.
+		return fmt.Errorf("IPv6 CIDR block provided by Amazon not found")
```
johngmyers, line 109: Does this require a "Required changes" release note?

tests/integration/update_cluster/complex/in-v1alpha2.yaml
```diff
@@ -19,7 +19,6 @@ spec:
           allocationId: eipalloc-012345a678b9cdefa
   kubernetesApiAccess:
   - 1.1.1.0/24
-  - 2001:0:8500::/40
```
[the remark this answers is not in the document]
hakman: The output for it was incorrect and should be handled in the new ipv6 test better.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -303,3 +338,74 @@ func (e *Subnet) CloudformationLink() *cloudformation.Literal {
...
+	var removals []fi.Deletion
+	for _, association := range subnet.Ipv6CidrBlockAssociationSet {
+		// Skip when without state
+		if association == nil || association.Ipv6CidrBlockState == nil {
+			continue
+		}
+
+		// Skip when already disassociated
+		state := aws.StringValue(association.Ipv6CidrBlockState.State)
+		if state == ec2.SubnetCidrBlockStateCodeDisassociated || state == ec2.SubnetCidrBlockStateCodeDisassociating {
```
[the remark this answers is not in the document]
hakman: I don't think there's any waiting, as deletions are done after applying new config. Also, disassociating a second time may make the deletion task fail and trigger errors.

upup/pkg/fi/cloudup/awstasks/vpcamazonipv6cidrblock.go
hakman: We allow, but via model. If it's shared we don't create this task.
Though inside the task it is correct to check if it matches the expected state, similar to the regular vpccidrblock.go task.

hakman:
Thanks for the review @johngmyers. I restaged the commits to include the review feedback.
/hold cancel

johngmyers (approved):

PR #11444, opened 2021-05-10
Sort --extra-tags of ebs-csi-driver

codablock:
Without sorting, the order is random and thus causes unnecessary cluster
updates.


codablock:
I assume that this is a 1.21 backport candidate

johngmyers (approved):
Thanks!

PR #11447, opened 2021-05-10
Always install the latest plugin versions for Terraform tests

hakman:
/cc @rifelpet

### DIFF
--- a/hack/verify-terraform.sh
+++ b/hack/verify-terraform.sh
@@ -30,7 +30,7 @@ while IFS= read -r -d '' -u 3 test_dir; do
   [ -f "${test_dir}/kubernetes.tf" ] || [ -f "${test_dir}/kubernetes.tf.json" ] || continue
   echo -e "${test_dir}\n"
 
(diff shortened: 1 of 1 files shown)


hakman:
Should still respect the constraints, upgrade the plugins cached on dev machines.
```
  -upgrade=false       If installing modules (-get) or plugins, ignore
                       previously-downloaded objects and install the
                       latest version allowed within configured constraints.
```

johngmyers:
Should we have the tests always specify the lowest provider versions?

johngmyers:
Test twice, once with oldest and once with newest?

hakman:
This is how I noticed that it doesn't work. It failed all tests for me this morning because the AWS provider was too old locally and it didn't what to do. Had to add this change to work.
Yes, I tested and respects the constraints for minimum provider version.

rifelpet:
ah, that makes (a bit) more sense. does `terraform init -upgrade` then download 3.34.0 or the latest 3.X ?

hakman:
Yup :)

rifelpet:
but using `>= 3.34.0` will download the latest 3.X ?

rifelpet (approved):
you're right that this preserves the existing functionality of the prow job so I'm inclined to merge as-is and if we can add more testing in a separate PR with `=` instead of `>=` that would be ideal.

PR #11454, opened 2021-05-11
Fix KCM livenessProbe to use secure port

rifelpet:
The hourly test that uses the HEAD of k/k [has been failing](https://testgrid.k8s.io/kops-latest#kops-aws-k8s-latest&show-stale-tests=) due to connection refused errors on the KCM livenessProbe. 

The failure is due to [this PR](https://github.com/kubernetes/kubernetes/pull/96216) which landed around the same time.

This updates the KCM livenessProbe to match upstream.

[This commit](https://github.com/kubernetes/kubernetes/commit/eb27b61cdbddd17a2990d61f3fb8250636220f74) shows the secure port has been present since k8s 1.12 so I think it should be safe to update unconditionally, but feedback is welcome.

/hold for feedback

### DIFF
--- a/nodeup/pkg/model/kube_controller_manager.go
+++ b/nodeup/pkg/model/kube_controller_manager.go
@@ -173,9 +173,10 @@ func (b *KubeControllerManagerBuilder) buildPod() (*v1.Pod, error) {
 		LivenessProbe: &v1.Probe{
 			Handler: v1.Handler{
 				HTTPGet: &v1.HTTPGetAction{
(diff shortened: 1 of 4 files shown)


hakman (approved):

rifelpet:
these tests will confirm that this doesn't break for older k8s versions: https://testgrid.k8s.io/kops-versions#kops-aws-k8s-1-17&show-stale-tests=

/hold cancel

PR #11457, opened 2021-05-11
upup: gcetasks: force send AutoCreateSubnetworks field when set to false

nicktrav:
When a struct contains a default value, it is not present in the
serialized output attached to request bodies sent to GCP. In the case
that a user wishes to explicitly set the `AutoCreateSubnetworks` field
to `false`, the request sent to GCP does not contain this directive, and
the network is provisioned with the default of `true`.

Explicitly specify that the `false` value should be present on the
request sent to the GCP API, making use of the `ForceSendFields` field
in the requset struct.

Signed-off-by: Nick Travers <n.e.travers@gmail.com>


hakman (approved):
Wow :). Thanks for the fix!

PR #11459, opened 2021-05-11
fix(coredns/rbac): add permission to list and watch endpointslices

nettoclaudio:
Following the changes on CoreDNS 1.8.3 (added in [commit](https://github.com/coredns/coredns/commit/9f72db12e73d6f1b7f69812baa4381fc7a1f04b3)), this PR adds permission to list and watch EndpointSlices resources.


olemarkus (approved):
Thanks for this!

PR #11460, opened 2021-05-11
upup: gcetasks: fix diffs in instance template and router

nicktrav:
Use the last component of the self-url for a router's region, rather
than the full URL.

Explicitly set the boolean pointer for `HasExternalIP` if the interface
attached to the instance does not have an access config (i.e. has no
public internet access).

Without this patch, both the router and instance template show pending
updates, neither of which can be applied as both are immutable
resources.

Signed-off-by: Nick Travers <n.e.travers@gmail.com>


kenji-cloudnatix:
Hi @hakman , @rifelpet , this PR will fix a bug that I've introduced in https://github.com/kubernetes/kops/pull/11184 . Thanks!

hakman (approved):

hakman:
Thanks you both @nicktrav & @kenji-cloudnatix.

PR #11470, opened 2021-05-12
Use kubernetes.default for OIDC discovery in gossip clusters

rifelpet:
It doesn't make sense to use a gossip hostname as the discovery url because it wont be resolveable.
For gossip clusters that dont provide a public VFS store, we can at least use kubernetes.default for internal oidc usage.

the goal of this is to fix the oidc tests for the e2e prow jobs that use gossip clusters:

https://testgrid.k8s.io/kops-gce#kops-gce-kubetest2&show-stale-tests=

https://testgrid.k8s.io/kops-misc#e2e-kops-do-calico&show-stale-tests=

slack thread: https://kubernetes.slack.com/archives/C8MKE2G5P/p1620837506354700

/hold for feedback


olemarkus (approved):

johngmyers:
Does the `MasterInternalName` make sense for non-gossip clusters without public discovery? I'm wondering if we should use `kubernetes.default` as the default only for gossip clusters, allowing things outside of the cluster but inside the local network to consume OIDC.

rifelpet:
Good point, I've updated this to only affect gossip clusters

johngmyers (approved):

johngmyers:
An integration test would not be amiss.

johngmyers (approved):

PR #11487, opened 2021-05-13
Set the test cluster-tag

olemarkus:
This should set the necessary tags for the iam roles to be able to use the test volumes

/cc @rifelpet


rifelpet:
the CLUSTER_NAME var isnt set by the job:

https://github.com/kubernetes/test-infra/blob/fe7ddc074339cc719f722b88beea80077aa6f7df/config/jobs/kubernetes/kops/kops-periodics-misc2.yaml#L177-L181

and so kubetest2 will generate a cluster name based on JOB_NAME and a few other inputs. It may make sense to copy what kubetest2 is currently using as a CLUSTER_NAME and explicitly set it in the prow job definition. Then we could enforce that it is set within this script and use it safely.

olemarkus:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kops/kops-periodics-misc.yaml#L3120

It should be set here

rifelpet (approved):
oops you're right. i was looking at the wrong test. too many tests :D

PR #11491, opened 2021-05-14
Update CAS manifest

olemarkus:
Upstream examples were missing a number of improvements.
This commit also adds template functions for making controllers such as CAS more HA on HA clusters

### DIFF
--- a/upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
+++ b/upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
@@ -1,5 +1,19 @@
 {{ with .ClusterAutoscaler }}
-# Sourced from https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-release-1.20/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-multi-asg.yaml
+# Sourced from https://github.com/kubernetes/autoscaler/
+---
+# Source: cluster-autoscaler/templates/pdb.yaml
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  labels:
+    k8s-addon: cluster-autoscaler.addons.k8s.io
+    k8s-app: cluster-autoscaler
+  name: cluster-autoscaler
+spec:
+  selector:
+    matchLabels:
+      k8s-app: "cluster-autoscaler"
+  maxUnavailable: 1
 ---
 apiVersion: v1
 kind: ServiceAccount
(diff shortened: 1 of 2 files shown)


upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -1,5 +1,21 @@
 {{ with .ClusterAutoscaler }}
-# Sourced from https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-release-1.20/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-multi-asg.yaml
+# Sourced from https://github.com/kubernetes/autoscaler/
+{{ if HasHighlyAvailableControlPlane }}
```
johngmyers, line 3: Why make this conditional?

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -272,6 +275,29 @@ func (tf *TemplateFunctions) GetInstanceGroup(name string) (*kops.InstanceGroup,
...
+	}
+	return 1
+}
+
+// HasHighlyAvailableControlPlane returns true of the cluster has more than one control plane node. False otherwise.
+func (tf *TemplateFunctions) HasHighlyAvailableControlPlane() bool {
+	cp := 0
+	for _, ig := range tf.InstanceGroups {
+		if ig.Spec.Role == kops.InstanceGroupRoleMaster {
+			cp++
```
johngmyers, line 292: Perhaps add `*ig.Spec.MinSize` instead if it's non-nil?

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -272,6 +275,29 @@ func (tf *TemplateFunctions) GetInstanceGroup(name string) (*kops.InstanceGroup,
 	return ig, nil
 }
 
+// ControlPlaneControllerReplicas returns the amount of replicas for a controllers that should run in the cluster
+// If the cluster has a highly available control plane, this function will return 2, if it has 1 control plane node, it will return 1
+func (tf *TemplateFunctions) ControlPlaneControllerReplicas() int {
+	if tf.HasHighlyAvailableControlPlane() {
+		return 2
```
johngmyers, line 282: Perhaps this should take the `2` as an input parameter? Return the max of input parameter and number of API servers?

johngmyers: I'm trying to figure out the benefit of having this sort of logic inside the template functions, versus having a template function for returning the number of apiservers and leaving the logic in go template.

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
olemarkus: Yep this one is perhaps unnecessary.

upup/pkg/fi/cloudup/template_functions.go
olemarkus: I think in almost all cases one wants this to be 2 for HA as these are leader elected.

This is a somewhat simple function, but I think it would be used a lot, so it will reduce some boilerplate.

johngmyers: Perhaps. A lot of these things are one-per-apiserver, but that's probably for other reasons. We might want to give weak zone anti-affinity for the 5-apiservers-in-3-zones case.

olemarkus: I don't think there are any good arguments for that. We have the same discussions around KCM, CCM etc .. these leader-elected controllers only run a single active instance and having two standbys is not very useful.

There are other components that are webhooks (or like metrics-server) and could be scaled using the autoscaler we use for the kube-dns service, but they often do not have to run on CP nodes at all.

But the leader-elected scenario is why I named the functions like I did as well.

olemarkus: Also worth mentioning that the PDB in this case guarantees that there will be an active leader at any point in time (given the PDB is respected)

johngmyers: Yes, I am convinced two are sufficient for leader-elected controllers, provided such controllers have some form of zone anti-affinity (when there are multiple zones).

As to naming, there's the ambiguity of what a "control plane node" is when etcd and apiserver are split. But that's a different issue—it would not make sense to have HA etcd with non-HA apiserver or vice versa.

upup/pkg/fi/cloudup/template_functions.go
olemarkus: If it is a CP IG, will that not always be 1?

upup/pkg/fi/cloudup/template_functions.go
olemarkus: Good point on topology key. For CP nodes that's the same, but with IRSA we have to remember to change this to zone.

"control plane node" are those with the role "control-plane", which we in an office hour informally concluded to be "what is running KCM, kube-scheduler, and similar". Doesn't necessarily have to run etcd. Unknown what this means in the long-term. Different discussion though.

upup/pkg/fi/cloudup/template_functions.go
johngmyers: That's how `kops create cluster` provisions them, but I don't think that's a constraint.

olemarkus: You are right. One can actually change a master IG to min/maxSize: 2. Which is ... interesting. We don't actually validate that we end up with odd number of masters then. The etcd-manager volume discovery also relies on single instance per IG. 

I think for now, it is safe to say it must be 1 and we lack some validation in this area.

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -138,6 +242,16 @@ spec:
...
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - cluster-autoscaler
+            topologyKey: kubernetes.io/hostname
```
johngmyers, line 254: As mentioned elsewhere, would like a soft anti-affinity to zone, to handle the 5 nodes in 3 zones case.

olemarkus: I think we could change `topologyKey` to `topology.kuberenetes.io/zone`. It wouldn't cover if someone puts all 3 CP nodes in the same zone. But I wonder `ControlPlaneControllerReplicas` should then return 1.

johngmyers: It still makes sense to have 2 replicas in a single-zone multi-control-plane-node cluster.

hakman (approved):

PR #11495, opened 2021-05-15
Set priorityClassName on critical addons

olemarkus:


upup/models/cloudup/resources/addons/certmanager.io/k8s-1.16.yaml.template
```diff
@@ -26688,6 +26688,7 @@ spec:
         imagePullPolicy: IfNotPresent
         name: cert-manager
         resources: {}
+      priorityClassName: system-cluster-critical
```
johngmyers, line 26691: Note this cannot be backported to kOps 1.21 as pods outside the `kube-system` namespace cannot have priorityClassName `system-cluster-critical` in k8s 1.16.

johngmyers (approved):

PR #11497, opened 2021-05-15
Cleanup orphaned IAM service account roles in direct render

johngmyers:


pkg/model/awsmodel/iam.go
```diff
@@ -441,3 +444,41 @@ func (b *IAMModelBuilder) buildAWSIAMRolePolicy(role iam.Subject) (fi.Resource,
 
 	return fi.NewStringResource(policy), nil
 }
+
+func (b *IAMModelBuilder) FindDeletions(context *fi.ModelBuilderContext, cloud fi.Cloud) error {
```
hakman, line 448: Is model the right place to implement this?

olemarkus: I think this is no less awkward than what I did with SG rules.
The challenge is that there is no relationship between tasks and models. Another model could also provision iam roles if that made sense.

An easy solution to this would be that the model add an additional tag to the resources it creates. And then replace ownershipTag below with that tag. E.g `kops.k8s.io/<cluster name>/owner: awsmodel/iammodelbuilder`

hakman: > I think this is no less awkward than what I did with SG rules.

I think https://github.com/kubernetes/kops/pull/10251 is not yet merged.

> The challenge is that there is no relationship between tasks and models.

It's always a challenge to track and handle orphaned resources and various hacks were made for that reason. I think the main reason this was not made easier initially was the lack of tags at that time. Another reason is that we 
Moving this logic to the model not such a good idea  and if we don't find a good alternative now will become a permanent thing.

johngmyers: This is absolutely the right place to implement this. The model knows what is supposed to be there, so after it has constructed what should be there it can compare that against what actually is there. Just like any well-behaved controller.

One could talk about consolidating the cluster and instancegroup destruction code into this mechanism in order to reduce the possibility of logic discrepancies, but that would be for future PRs.

johngmyers: > The challenge is that there is no relationship between tasks and models.

There is a relationship: models create tasks. This PR uses the tasks generated by `Loader.BuildTasks()` as a record of which objects the model says should exist. This is why `FindDeletions` runs as a second pass.

rifelpet:
IAM role tags were added in kops 1.20 (https://github.com/kubernetes/kops/commit/4ee5d7a5433a6b46343d4527fd9dc1cfced9f94e) which means if a cluster was last updated with kops <1.20 and then deleted with a version of kops that includes this PR, the untagged roles would not be cleaned up. Is that something we should be concerned with?

rifelpet:
`kops delete cluster` could return an error if its version doesn't match the cluster's kops-version.txt. I'm not sure that is a restriction worth imposing though.

johngmyers:
I don't think deleting a cluster with a newer version of kops than was used to create it is a use case we should spend a lot of time supporting. We could do the `kops-version.txt` check for known inflection points like this. We should probably do a `kops-version.txt` check for the case of deleting with an older kops version.

upup/pkg/fi/task.go
```diff
@@ -50,6 +50,14 @@ type ModelBuilder interface {
...
 }
 
+// HasDeletions is a ModelBuilder that creates tasks to delete cloud objects that no longer exist in the model.
+type HasDeletions interface {
+	ModelBuilder
+	// FindDeletions finds cloud objects that are owned by the cluster but no longer in the model and creates tasks to delete them.
+	// It is not called for the Terraform or Cloudformation targets.
+	FindDeletions(context *ModelBuilderContext, cloud Cloud) error
+}
+
```
rifelpet, line 60: We already have this:

https://github.com/kubernetes/kops/blob/da6eee03b0cfb96fb3b5622a9d7fc907ebe927be/upup/pkg/fi/deletions.go#L19-L28

would it make sense to consolidate these into one interface? I'm not sure I understand the difference between how they're both used.

johngmyers: `ProducesDeletions` finds everything of a type that is owned by the cluster and creates a `Deletion` that gets rid of it. It is only useful if you're deleting the cluster because it has no concept of objects owned by the cluster that need to be kept.

`HasDeletions` is a subclass of `ModelBuilder` that, during a second pass during apply_cluster, looks at the `Task`s representing the expected state. It finds objects owned by the cluster that have no corresponding expected `Task` and adds additional `Task`s to delete them.

johngmyers: It might make sense to later consolidate `ProducesDeletions` into `HasDeletions`, but there are a lot of deletions that only happen during cluster teardown and the `Task` interface has more requirements than the `Deletion` interface. I see such consolidation as potential future work.

rifelpet:
This looks good to me. Do the new deletions show up in the dryrun output? It would be inconvenient if someone created an IAM role externally that followed the naming and tagging convention and they inadvertently end up deleting the role after a kops upgrade.

johngmyers:
Deletions do show up in the dryrun output. This is a big reason why I used `Task`s as the output.

rifelpet (approved):

pkg/model/awsmodel/iam.go
hakman: I was not opposed to this, just not a fan of adding AWS SDK code in model.
Also, the FindDeletions name is something that I would see more as FindOrphaned and leave FindDeletions for tasks.
In general, making use of this may fix some of the awkwardness in some task deletions.

johngmyers: Well, this particular `ModelBuilder` is AWS-specific and in a directory for AWS-specific model code.

I made `HasDeletions` require `ModelBuilder` so that both could be added to `Loader.Builders`. I wanted to avoid errors where something that was both didn't get added to both lists.

But `HasDeletions` is closer to `Task` than `ModelBuilder`, despite needing to take a `ModelBuilderContext` in order to construct new `Task`s.

PR #11500, opened 2021-05-16
Bump CoreDNS manifests to latest stable version 1.8.3

dntosas:
In this commit, we update manifest files of CoreDNS. We also do a small
refactoring on yaml blocks to match the upstream source and to make
updating and reviewing procedures easier on the future.

### DIFF
--- a/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
@@ -1,3 +1,6 @@
+# CoreDNS
+# Source: https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed
+---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
@@ -74,6 +77,7 @@ data:
         health {
           lameduck 5s
         }
+        ready
         kubernetes {{ KubeDNS.Domain }}. in-addr.arpa ip6.arpa {
           pods insecure
           fallthrough in-addr.arpa ip6.arpa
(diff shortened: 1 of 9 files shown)


upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -219,8 +179,8 @@ spec:
           failureThreshold: 5
         readinessProbe:
           httpGet:
-            path: /health
-            port: 8080
+            path: /ready
+            port: 8181
```
dntosas, line 183: this seems not to be working

checking and if needed we will also push a PR on upstream

dntosas: oh ok, needed `ready` plugin on configuration

fixed

upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -219,8 +180,8 @@ spec:
...
```
dntosas, line 184: could that be a breaking change for some setup that use custom configuration? WDYT?

`ready` plugin is needed, that we add in our default config file ^^

olemarkus: I cannot imagine this will be a problem.

olemarkus (approved):

upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
dntosas: @olemarkus there were some users that use custom coreDNS config reporting this to be a breaking change for their setups. maybe good idea of adding a release note to guide of needed actions in such cases? WDYT?

PR #11502, opened 2021-05-16
bump aws lb controller to 2.2.0

olemarkus:
With these changes, we also pass the upstream e2e tests (with some amendments related to them expecting something more similar to EKS).

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -736,21 +736,35 @@ func addCalicoSrcDstCheckPermissions(p *Policy) {
 func AddAWSLoadbalancerControllerPermissions(p *Policy, resource stringorslice.StringOrSlice, clusterName string) {
 	addMasterEC2Policies(p, resource, false, clusterName)
 	addMasterELBPolicies(p, resource, false)
-	p.Statement = append(p.Statement, &Statement{
-		Effect: StatementEffectAllow,
-		Action: stringorslice.Of(
-			"ec2:AuthorizeSecurityGroupIngress", // aws.go
-			"ec2:DeleteSecurityGroup",           // aws.go
-			"ec2:RevokeSecurityGroupIngress",    // aws.go
-		),
-		Resource: resource,
-		Condition: Condition{
-			"StringEquals": map[string]string{
-				"ec2:ResourceTag/elbv2.k8s.aws/cluster": clusterName,
+	p.Statement = append(p.Statement,
+		&Statement{
+			Effect: StatementEffectAllow,
+			Action: stringorslice.Of(
+				"ec2:DescribeAvailabilityZones",
+			),
+			Resource: stringorslice.Slice([]string{"*"}),
+		},
+		&Statement{
+			Effect: StatementEffectAllow,
+			Action: stringorslice.Of(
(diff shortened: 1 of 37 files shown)


pkg/model/iam/iam_builder.go
```diff
@@ -736,21 +736,33 @@ func addCalicoSrcDstCheckPermissions(p *Policy) {
...
+			Effect: StatementEffectAllow,
+			Action: stringorslice.Of(
+				"ec2:AuthorizeSecurityGroupIngress", // aws.go
+				"ec2:DeleteSecurityGroup",           // aws.go
+				"ec2:RevokeSecurityGroupIngress",    // aws.go
 
+				"elasticloadbalancing:ModifyTargetGroupAttributes",
+				"elasticloadbalancing:ModifyRule",
+				"elasticloadbalancing:DeleteRule",
+				"elasticloadbalancing:RemoveTags",
```
rifelpet, line 757: `RemoveTags` is being added in both statements. Can we consolidate it to just one statement?

tests/e2e/scenarios/aws-lb-controller/run-test.sh
```diff
@@ -0,0 +1,70 @@
...
+cd "$(mktemp -dt kops.XXXXXXXXX)"
+go get github.com/onsi/ginkgo/ginkgo
+
+git clone --branch e2e-filter-non-eligible-targets https://github.com/olemarkus/aws-load-balancer-controller .
+
+ginkgo -v -r test/e2e -- \
+    -cluster-name="${CLUSTER_NAME}" \
+    -aws-region="${REGION}" \
+    -aws-vpc-id="$VPC" \
+	-ginkgo.reportFile="${REPORT_DIR}/junit.xml"
```
rifelpet, line 70:
```suggestion
    -ginkgo.reportFile="${REPORT_DIR}/junit.xml"

```

tests/e2e/scenarios/aws-lb-controller/run-test.sh
```diff
@@ -0,0 +1,70 @@
...
+VPC=$(${KOPS} toolbox dump -o json | jq -r .vpc.id)
+
+ZONE=$(${KOPS} get ig -o json | jq -r '[.[] | select(.spec.role=="Node") | .spec.subnets[0]][0]')
+
+REGION=${ZONE::-1}
+
+cd "$(mktemp -dt kops.XXXXXXXXX)"
+go get github.com/onsi/ginkgo/ginkgo
+
+git clone --branch e2e-filter-non-eligible-targets https://github.com/olemarkus/aws-load-balancer-controller .
```
rifelpet, line 64: Might be worth a comment linking to https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2012, suggesting we can switch back to upstream/main when it has been merged.

pkg/model/iam/iam_builder.go
olemarkus: Hm. Is the last statement redundant? Technically they should be for different types of resources (ELBs vs listeners)

olemarkus: Think this should work just as well.

rifelpet (approved):

PR #11510, opened 2021-05-18
Set OpenStack lb listener connection limit to 500k

zetaab:
We are currently facing issue https://bugzilla.redhat.com/show_bug.cgi?id=1724480 and because of that I recommend that we limit the Kubernetes API loadbalancer listener max conns to 500k. At least for me it is pretty difficult to imagine that someone could have more than 500k simultaneous connections to Kubernetes API


zetaab:
looks like this issue is only valid if we have more than one pool, and kops uses only one pool -> no need to do changes

PR #11517, opened 2021-05-18
Add default tags to LB controller and cilium eni resources

olemarkus:


upup/models/cloudup/resources/addons/aws-load-balancer-controller.addons.k8s.io/k8s-1.9.yaml.template
```diff
@@ -801,6 +801,7 @@ spec:
         - --enable-wafv2=false
         - --enable-shield=false
         - --ingress-class=alb
+        - "--default-tags={{ CsiExtraTags }}"
```
rifelpet, line 804: can we rename template function to something more generic?

olemarkus: hehe. yep. done

johngmyers (approved):

PR #11523, opened 2021-05-19
Calculate IPv6 subnet CIDR based on cluster CIDR

hakman:
Ref: https://github.com/kubernetes/kops/issues/8432#issuecomment-843777374

Inspired by https://github.com/kubernetes/kops/pull/11163.

This PR makes it possible to set subnet CIDR based on VPC CIDR on cluster creation similarly to:
  * https://www.terraform.io/docs/language/functions/cidrsubnet.html
  * https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-cidr.html

There is some feedback from @johngmyers in https://github.com/kubernetes/kops/pull/11442#discussion_r631712081, which suggests using a `cidrsubnet(8,1)` syntax instead of a `cidrsubnet-8-1` syntax for configuring a subnet.
The reason I chose the `cidrsubnet-8-1` is to maybe allow using this as a flag on create cluster and that it seems similar to other places, like https://kops.sigs.k8s.io/cluster_spec/#egress.

/cc @justinsb @johngmyers @rifelpet

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -333,7 +333,7 @@ func validateIPv6CIDR(cidr string, fieldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if strings.HasPrefix(cidr, "/") {
-		newSize, _, err := utils.ParseCidrSubnet(cidr)
+		newSize, _, err := utils.ParseCIDRNotation(cidr)
 		if err != nil {
 			allErrs = append(allErrs, field.Invalid(fieldPath, cidr, fmt.Sprintf("IPv6 CIDR subnet is not parsable: %v", err)))
 			return allErrs
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -330,10 +330,23 @@ func validateCIDR(cidr string, fieldPath *field.Path) field.ErrorList {
 }
 
 func validateIPv6CIDR(cidr string, fieldPath *field.Path) field.ErrorList {
-	allErrs := validateCIDR(cidr, fieldPath)
+	allErrs := field.ErrorList{}
 
-	if !utils.IsIPv6CIDR(cidr) {
-		allErrs = append(allErrs, field.Invalid(fieldPath, cidr, "Network is not an IPv6 CIDR"))
+	if strings.HasPrefix(cidr, "/") {
+		newSize, _, err := utils.ParseCidrSubnet(cidr)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fieldPath, cidr, fmt.Sprintf("IPv6 CIDR subnet is not parsable: %v", err)))
+			return allErrs
+		}
+		if newSize < 0 || newSize > 128 {
+			allErrs = append(allErrs, field.Invalid(fieldPath, cidr, "IPv6 CIDR subnet size must be a value between 0 and 128"))
+		}
+	} else {
+		allErrs = append(allErrs, validateCIDR(cidr, fieldPath)...)
+
+		if !utils.IsIPv6CIDR(cidr) {
+			allErrs = append(allErrs, field.Invalid(fieldPath, cidr, "Network is not an IPv6 CIDR"))
+		}
 	}
 
 	return allErrs
--- a/upup/pkg/fi/cloudup/awstasks/subnet.go
+++ b/upup/pkg/fi/cloudup/awstasks/subnet.go
@@ -465,12 +465,12 @@ func calculateSubnetCIDR(vpcCIDR, subnetCIDR *string) (*string, error) {
 		return nil, fmt.Errorf("expecting subnet cidr to start with %q: %q", "/", aws.StringValue(subnetCIDR))
 	}
 
-	newSize, netNum, err := utils.ParseCidrSubnet(aws.StringValue(subnetCIDR))
+	newSize, netNum, err := utils.ParseCIDRNotation(aws.StringValue(subnetCIDR))
 	if err != nil {
 		return nil, fmt.Errorf("error parsing CIDR subnet: %v", err)
 	}
 
-	newCIDR, err := utils.CidrSubnet(aws.StringValue(vpcCIDR), newSize, netNum)
+	newCIDR, err := utils.CIDRSubnet(aws.StringValue(vpcCIDR), newSize, netNum)
 	if err != nil {
 		return nil, fmt.Errorf("error calculating CIDR subnet: %v", err)
 	}
--- a/upup/pkg/fi/cloudup/awstasks/subnet.go
+++ b/upup/pkg/fi/cloudup/awstasks/subnet.go
@@ -100,18 +100,20 @@ func (e *Subnet) Find(c *fi.Context) (*Subnet, error) {
 		}
 
 		actual.IPv6CIDR = association.Ipv6CidrBlock
-		if e.VPC.IPv6CIDR != nil && strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "/") {
-			subnetIPv6CIDR, err := calculateSubnetCIDR(e.VPC.IPv6CIDR, e.IPv6CIDR)
-			if err != nil {
-				return nil, err
-			}
-			e.IPv6CIDR = subnetIPv6CIDR
-		}
 		break
 	}
 
 	klog.V(2).Infof("found matching subnet %q", *actual.ID)
 	e.ID = actual.ID
+
+	// Calculate expected IPv6 CIDR if possible and in the "/64#N" format
+	if e.VPC.IPv6CIDR != nil && strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "/") {
+		subnetIPv6CIDR, err := calculateSubnetCIDR(e.VPC.IPv6CIDR, e.IPv6CIDR)
+		if err != nil {
+			return nil, err
+		}
+		e.IPv6CIDR = subnetIPv6CIDR
+	}
 
 	// Prevent spurious changes
 	actual.Lifecycle = e.Lifecycle // Not materialized in AWS
--- a/upup/pkg/fi/cloudup/awstasks/subnet.go
+++ b/upup/pkg/fi/cloudup/awstasks/subnet.go
@@ -18,6 +18,7 @@ package awstasks
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -98,6 +99,14 @@ func (e *Subnet) Find(c *fi.Context) (*Subnet, error) {
 		}
 
 		actual.IPv6CIDR = association.Ipv6CidrBlock
+		if e.IPv6CIDR != nil && strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "/") {
+			subnetIPv6CIDR, err := calculateSubnetCIDR(e.VPC.IPv6CIDR, e.IPv6CIDR)
+			if err != nil {
+				return nil, err
+			}
+			e.IPv6CIDR = subnetIPv6CIDR
+		}
+		break
 	}
 
 	klog.V(2).Infof("found matching subnet %q", *actual.ID)
@@ -196,6 +205,22 @@ func (_ *Subnet) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Subnet) error {
 		}
 	}
 
+	if e.IPv6CIDR != nil && strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "/") {
+		vpcIPv6CIDR, err := findVPCIPv6CIDR(t.Cloud, e.VPC.ID)
+		if err != nil {
+			return err
+		}
+		if vpcIPv6CIDR == nil {
+			return fi.NewTryAgainLaterError("waiting for the VPC IPv6 CIDR to be assigned")
+		}
+
+		subnetIPv6CIDR, err := calculateSubnetCIDR(vpcIPv6CIDR, e.IPv6CIDR)
+		if err != nil {
+			return err
+		}
+		e.IPv6CIDR = subnetIPv6CIDR
+	}
+
 	if a == nil {
 		klog.V(2).Infof("Creating Subnet with CIDR: %q", *e.CIDR)
 
@@ -273,6 +298,12 @@ func (_ *Subnet) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *Su
 		return t.AddOutputVariableArray("subnet_ids", terraformWriter.LiteralFromStringValue(*e.ID))
 	}
 
+	if strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "/") {
+		// TODO: Implement using "cidrsubnet"
+		// https://www.terraform.io/docs/language/functions/cidrsubnet.html
+		return fmt.Errorf("<cidrsubnet> in not supported with Terraform target: %q", aws.StringValue(e.IPv6CIDR))
+	}
+
 	tf := &terraformSubnet{
 		VPCID:            e.VPC.TerraformLink(),
 		CIDR:             e.CIDR,
@@ -312,6 +343,12 @@ func (_ *Subnet) RenderCloudformation(t *cloudformation.CloudformationTarget, a,
 		// Not cloudformation owned / managed
 		// We won't apply changes, but our validation (kops update) will still warn
 		return nil
+	}
+
+	if strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "/") {
+		// TODO: Implement using "Fn::Cidr"
+		// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-cidr.html
+		return fmt.Errorf("<cidrsubnet> in not supported with CloudFormation target: %q", aws.StringValue(e.IPv6CIDR))
 	}
 
 	cf := &cloudformationSubnet{
@@ -408,4 +445,28 @@ func (d *deleteSubnetIPv6CIDRBlock) TaskName() string {
 
 func (d *deleteSubnetIPv6CIDRBlock) Item() string {
 	return fmt.Sprintf("%v: ipv6cidr=%v", *d.vpcID, *d.ipv6CidrBlock)
+}
+
+func calculateSubnetCIDR(vpcCIDR, subnetCIDR *string) (*string, error) {
+	if vpcCIDR == nil {
+		return nil, fmt.Errorf("expecting VPC CIDR to not be <nil>")
+	}
+	if subnetCIDR == nil {
+		return nil, fmt.Errorf("expecting subnet CIDR to not be <nil>")
+	}
+	if !strings.HasPrefix(aws.StringValue(subnetCIDR), "/") {
+		return nil, fmt.Errorf("expecting subnet cidr to start with %q: %q", "/", aws.StringValue(subnetCIDR))
+	}
+
+	newSize, netNum, err := utils.ParseCidrSubnet(aws.StringValue(subnetCIDR))
+	if err != nil {
+		return nil, fmt.Errorf("error parsing CIDR subnet: %v", err)
+	}
+
+	newCIDR, err := utils.CidrSubnet(aws.StringValue(vpcCIDR), newSize, netNum)
+	if err != nil {
+		return nil, fmt.Errorf("error calculating CIDR subnet: %v", err)
+	}
+
+	return aws.String(newCIDR), nil
 }
--- a/upup/pkg/fi/cloudup/awstasks/subnet.go
+++ b/upup/pkg/fi/cloudup/awstasks/subnet.go
@@ -100,7 +100,7 @@ func (e *Subnet) Find(c *fi.Context) (*Subnet, error) {
 		}
 
 		actual.IPv6CIDR = association.Ipv6CidrBlock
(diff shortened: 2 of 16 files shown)


upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -273,6 +309,12 @@ func (_ *Subnet) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *Su
 		return t.AddOutputVariableArray("subnet_ids", terraformWriter.LiteralFromStringValue(*e.ID))
 	}
 
+	if strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "cidrsubnet-") {
+		// TODO: Implement using "cidrsubnet"
+		// https://www.terraform.io/docs/language/functions/cidrsubnet.html
+		return fmt.Errorf("<cidrsubnet> in not upported with Terraform target: %q", aws.StringValue(e.IPv6CIDR))
```
johngmyers, line 315:
```suggestion
		return fmt.Errorf("<cidrsubnet> in not supported with Terraform target: %q", aws.StringValue(e.IPv6CIDR))
```

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -314,6 +356,12 @@ func (_ *Subnet) RenderCloudformation(t *cloudformation.CloudformationTarget, a,
 		return nil
 	}
 
+	if strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "cidrsubnet-") {
+		// TODO: Implement using "Fn::Cidr"
+		// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-cidr.html
+		return fmt.Errorf("<cidrsubnet> in not upported with CloudFormation target: %q", aws.StringValue(e.IPv6CIDR))
```
johngmyers, line 362:
```suggestion
		return fmt.Errorf("<cidrsubnet> in not supported with CloudFormation target: %q", aws.StringValue(e.IPv6CIDR))
```

upup/pkg/fi/utils/cidr.go
```diff
@@ -56,3 +62,42 @@ func IsIPv6CIDR(cidr string) bool {
 
 	return true
 }
+
+// ParseCidrSubnet parses a string in the format "cidrsubnet-<newBits>-<netNum>"
+// and returns the values of <newBits> and <netNum>
```
johngmyers, line 67:
```suggestion
// ParseCidrSubnet parses a string in the format "cidrsubnet-<newBits>-<netNum>"
// and returns the values of <newBits> and <netNum>.
```

johngmyers (changes requested):

johngmyers:
`egress` takes AWS resource IDs, whereas this is a function. I would like something that looks more like a function than an identifier.

hakman:
> `egress` takes AWS resource IDs, whereas this is a function. I would like something that looks more like a function than an identifier.

True. Still, exposing this as a function would be something totally new.
I am not opposed to this change, just think it would be best to agree which way to go during office hours.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -273,6 +289,12 @@ func (_ *Subnet) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *Su
 		return t.AddOutputVariableArray("subnet_ids", terraformWriter.LiteralFromStringValue(*e.ID))
 	}
 
+	if strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "cidrsubnet-") {
+		// TODO: Implement using "cidrsubnet"
+		// https://www.terraform.io/docs/language/functions/cidrsubnet.html
+		return fmt.Errorf("<cidrsubnet> in not supported with Terraform target: %q", aws.StringValue(e.IPv6CIDR))
```
rifelpet, line 295: Let me know if you need help adding support for `cidrsubnet()`.

Looking at the literal, writer, and hcl2 code right now, it is a bit more involved than I would like. I think we could refactor the `file()`/`filebase64()` functionality to more easily support arbitrary functions like `cidrsubnet()`

hakman:
One other idea coming from @rifelpet would be to use a struct field to describe this:
```go
type IPv6CIDRSubnet struct {
    NewBits int
    NetNum int
}
```

justinsb:
Could we use a syntax like `ipv6CIDR: /64` on the subnet - for when the user wants a /64 but doesn't much care which one.

I figure if we have use cases for more precise allocation in future we can add those.

I did a prototype https://github.com/kubernetes/kops/pull/11579 and the experience feels intuitive.  (My prototype uses "::/64" so it's a valid CIDR, but that's not a great experience because `kops edit` keeps removing the (required) quotes around that).

We'd have to figure out how we assign to terraform consistently.  We could do something like

`0:0:0:1::/64` through `0:0:0:ff::/64`; that indicates the bits we care about.

I think it's reasonable to require this specification for terraform.  It would be nice if `kops create cluster` auto assigned those to subnets in ipv6 mode.  We could then also require that for direct mode, though I'm not wild about the idea of making kOps less user-friendly because of terraform limitations.

justinsb:
Another possible syntax for absolute positioning: `/64#3` meaning "subnet of size 64, number 3".  That's close to what users care about - I think - and in keeping with the easier syntax `/64`.

(I don't know if `#` is ever meaningful in IPv6 and we should choose another character.  `%` is scope so I think we should avoid that.)

hakman:
I like the `/64#3` or `::/64#3` notations.
Regarding auto-assignment, I would be happy to do it in create cluster only, where this would be done based on simple subnet number. I would prefer not to add a mechanism for auto-assignment for existing clusters. On those, people can decide on these values themselves.

johngmyers:
I don't see any particular reason not to support auto-assignment for existing clusters. Why would would we require admins of existing clusters to manually do this themselves?

What's the problem with Terraform? Can't we query the existing state of the subnets in order to specify that state in the configuration when it exists?

hakman:
The reason I see not to support auto-assignment after cluster creation in the complications in logic for something that is not needed to make IPv6 work. It is nice to have, but would gladly live without it for now.
We already have differences between how things are done in Direct, CF and TF because of how settings are implemented of not there.

hakman:
Final implementations uses the `/64#3` notation for IPv6 subnet, where `64` is a decimal value and `3` is a hex value (as discussed during Office Hours.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -196,6 +205,22 @@ func (_ *Subnet) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Subnet) error {
 		}
 	}
 
+	if e.IPv6CIDR != nil && strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "/") {
```
johngmyers, line 208: A comment would be helpful to indicate this handles the case where the VPC's IPv6CIDR was not known at the time `e.Find()` is run. 

And actually, `e.Find()` should `NewTryAgainLaterError` until it can get the association. Otherwise the code path is timing-dependent.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -196,6 +205,22 @@ func (_ *Subnet) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Subnet) error {
...
+		vpcIPv6CIDR, err := findVPCIPv6CIDR(t.Cloud, e.VPC.ID)
```
johngmyers, line 209: Why limit this to Amazon-provided IPv6 associations? Why not just use e.VPC.IPv6CIDR?

johngmyers, line 209: There's no dependency on the `VPCAmazonIPv6CIDRBlock` task, so `RunTasks` might schedule the `VPCAmazonIPv6CIDRBlock` task in a later slot.

johngmyers (changes requested):

pkg/apis/kops/validation/validation.go
```diff
@@ -334,10 +334,23 @@ func validateCIDR(cidr string, fieldPath *field.Path) field.ErrorList {
 }
 
 func validateIPv6CIDR(cidr string, fieldPath *field.Path) field.ErrorList {
-	allErrs := validateCIDR(cidr, fieldPath)
+	allErrs := field.ErrorList{}
 
-	if !utils.IsIPv6CIDR(cidr) {
-		allErrs = append(allErrs, field.Invalid(fieldPath, cidr, "Network is not an IPv6 CIDR"))
+	if strings.HasPrefix(cidr, "/") {
```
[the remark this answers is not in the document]
johngmyers: Never mind, this works off of the VPC's `CIDR` or `IPv6CIDR` field, so isn't AWS-specific.

johngmyers:
We should support this syntax on the `Subnet.CIDR` field, but that can be a separate PR.

upup/pkg/fi/utils/cidr.go
```diff
@@ -56,3 +64,47 @@ func IsIPv6CIDR(cidr string) bool {
 
 	return true
 }
+
+// ParseCidrSubnet parses a string in the format "/<newSize>#<netNum>"
+// and returns the values of <newSize> and <netNum>.
+func ParseCidrSubnet(subnet string) (int, int64, error) {
```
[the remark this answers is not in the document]
hakman: Well, "generated" is not ideal either. It is just a notation.
When using networking as the context, "subnet" has a broader meaning.
I would see it more as a `ParseSubnetNotation()` name.

pkg/apis/kops/validation/validation.go
```diff
@@ -334,10 +334,23 @@ func validateCIDR(cidr string, fieldPath *field.Path) field.ErrorList {
...
+		newSize, _, err := utils.ParseCidrSubnet(cidr)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fieldPath, cidr, fmt.Sprintf("IPv6 CIDR subnet is not parsable: %v", err)))
+			return allErrs
+		}
+		if newSize < 0 || newSize > 128 {
```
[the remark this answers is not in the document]
hakman: I agree this validation is pretty trivial and can be improved in a future PR.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -196,6 +205,22 @@ func (_ *Subnet) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Subnet) error {
...
+		if err != nil {
+			return err
+		}
+		if vpcIPv6CIDR == nil {
+			return fi.NewTryAgainLaterError("waiting for the VPC IPv6 CIDR to be assigned")
```
johngmyers, line 214: It's also possible it's a shared VPC with no IPv6 allocation forthcoming.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -98,6 +99,14 @@ func (e *Subnet) Find(c *fi.Context) (*Subnet, error) {
 		}
 
 		actual.IPv6CIDR = association.Ipv6CidrBlock
+		if e.IPv6CIDR != nil && strings.HasPrefix(aws.StringValue(e.IPv6CIDR), "/") {
+			subnetIPv6CIDR, err := calculateSubnetCIDR(e.VPC.IPv6CIDR, e.IPv6CIDR)
```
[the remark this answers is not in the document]
hakman: `VPCAmazonIPv6CIDRBlock.RenderAWS()` cannot write the CIDR, because the request doesn't actually return which block it associates. Because of that, there is no dependency and `Subnet.RenderAWS()` waits for the association to be done, using `fi.NewTryAgainLaterError()`.

I agree that `Find()` writing to `e` is awkward, but it was that or writing the `e.IPv6CIDR` notation to `a.IPv6CIDR`.
Also makes it easier later in `FindDeletions()`. I could change that to set `a.IPv6CIDR = e.IPv6CIDR`, if it makes things less awkward.

upup/pkg/fi/cloudup/awstasks/subnet.go
hakman: This correct here, as it handles the case of `TargetDryRun`. We only care about the exact CIDR at render time, when we know that the VPC will eventually get and IPv6 CIDR.

upup/pkg/fi/cloudup/awstasks/subnet.go
hakman: I cannot test with customer owned CIDRs. The AWS docs regarding IPv6 are not exactly clear either.
So, for now, this is meant to only support AWS owned CIDRs, where I know that can be exactly 1 per subnet with a /64 size and from exactly 1 bloc of /56 size.

hakman: There is no value in the dependency, as `VPCAmazonIPv6CIDRBlock` will not return the associated block.
This is the reason for using `NewTryAgainLaterError` in `Subnet.RenderAWS`.

johngmyers: But unless `VPCAmazonIPv6CIDRBlock.Run()` is called, the IPv6 association won't ever arrive. If there's no dependency, `RunTasks` might schedule `VPCAmazonIPv6CIDRBlock` to a later tranche than `Subnet`.

johngmyers: If there's a customer-specified `VPC.IPv6CIDR`, the intended behavior is obvious: take the slice off of the specified VPC IPv6CIDR.

upup/pkg/fi/cloudup/awstasks/subnet.go
johngmyers: `TargetDryRun` will never call `RenderAWS()`, so that case will never happen.

upup/pkg/fi/cloudup/awstasks/subnet.go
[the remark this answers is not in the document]
johngmyers: That's still not correct. The code should not take a different path based on whether the `Subnet` has an IPv6 association. The actual of the `Subnet` should have no effect on the expected.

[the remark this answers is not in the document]
hakman: OK, if you insist, will change.

upup/pkg/fi/cloudup/awstasks/subnet.go
hakman: `Find()` has nothing to do with `VPC.IPv6CIDR`. It just needs to find the current state of the subnet.
If by the time it gets to `RenderAWS()` is still missing the `VPC.IPv6CIDR`, `NewTryAgainLaterError` waits for it.

upup/pkg/fi/cloudup/awstasks/subnet.go
hakman: > The IPv6 association won't ever arriveThere is a dependency in VPC, so it will arrive.

There is a dependency in VPC, so it will arrive, sooner or later.

> RunTasks might schedule VPCAmazonIPv6CIDRBlock to a later tranche than Subnet.

This is why `NewTryAgainLaterError` waits for it to appear.

upup/pkg/fi/cloudup/awstasks/subnet.go
hakman: In this case, there will be an error from `VPCAmazonIPv6CIDRBlock` that says: "IPv6 CIDR block provided by Amazon not found".

upup/pkg/fi/utils/cidr.go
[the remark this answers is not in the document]
johngmyers: Yeah, it's not necessarily generated. It's cut out from the VPC's singular CIDR.

(except VPCs, especially shared ones, can technically have multiple CIDRs.)

upup/pkg/fi/cloudup/awstasks/subnet.go
johngmyers: `Find()` is using `VPC.IPv6CIDR` to get the block to make allocations from.

`Find()` and `RenderAWS()` are using different rules for making the allocation and which one gets run depends on timing.

upup/pkg/fi/cloudup/awstasks/subnet.go
johngmyers: It won't appear due to how `RunTasks` schedules tasks. If tasks `b` and `c` depend on `a`, `d` depends on `b`, and `e` depends on `c`, the scheduler effectively puts false dependencies of `d` depending on `c` and `e` depending on `b`.

upup/pkg/fi/cloudup/awstasks/subnet.go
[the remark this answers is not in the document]
johngmyers: I think `Find()` writing to `e` is okay, I'm just having to figure out what the rules and model are. It seems `Find()` can refine the expected based on existing state, including state produced by tasks listed as dependencies. But it has to deal with those dependencies not running due to `TargetDryRun`.

There's an opportunity here to improve the developer documentation.

This could be used in the future to support a `/64` notation where the particular sub-block doesn't need to be specified.

But there is a dependency on `VPCAmazonIPv6CIDRBlock` as the association won't happen unless that task is run. And there shouldn't be two blocks of code when one would do, especially not if the two blocks of code implement different semantics.

upup/pkg/fi/cloudup/awstasks/subnet.go
johngmyers: You could test using a shared VPC with a site-local IPv6 CIDR association.

It make sense to support only AWS-provided CIDRs in managed VPCs, especially for egress. But we shouldn't design out the shared VPC case. There's already code that pulls from `VPC.IPv6CIDR`; we should be consistent about using that when it's available.

johngmyers:
I think we might need to make significant changes to `subnet_test.go` to exercise all of this.

johngmyers:
I believe there are two outstanding issues:

1) Missing dependency from `Subnet` to `VPCAmazonIPv6CIDRBlock`.
2) Looking up the VPC's IPv6 association in one place (`Find()`) or two (`Find()` and `RenderAWS()`).

For issue (1):

`RunTasks` works by scheduling `Task`s into tranches. The first tranche is tasks that have no dependencies. The second tranche is tasks that depend only on tasks in the first tranche. The third tranche is tasks that depend only on tasks in either the first or second tranche. And so on.

It then runs all the tasks in a given tranche in parallel. It waits for all tasks in a tranche to complete (without retries) before running any tasks in the next tranche.

`Subnet` depends only on `VPC`. So it will be in the tranche immediately following the one that `VPC` is in. `VPCAmazonIPv6CIDRBlock` currently depends only on `VPC`, so currently is scheduled in the same tranche as `Subnet`. So the two tasks race.

If there were a subsequent change that required `VPCAmazonIPv6CIDRBlock` to depend on an intermediate task that in turn depended on `VPC`, then it would then be scheduled in a tranche after `Subnet`. That means it would never run until all `Subnet` tasks completed. So if `Subnet` were to wait for an association that it is expecting `VPCAmazonIPv6CIDRBlock` to initiate, it would wait forever.

There also might be a race. If `Subnet.RenderAWS()` calls `findVPCIPv6CIDR()` before `VPCAmazonIPv6CIDRBlock.RenderAWS()` runs, it won't find an IPv6 association in progress, so might return a permanent error instead of a TryAgainLaterError.

I'll write up issue (2) later.

hakman:
@johngmyers I addressed (1).

hakman:
At the moment, task order due to dependencies is the following:
* VPC (no dependency)
* VPCAmazonIPv6CIDR (depends on VPC)
* Subnet (depends on VPC and VPCAmazonIPv6CIDR)

In the subnet `Find()` function, we have the following case with regards to IPv6:
* VPC doesn't exists -> No subnet will exist so no CIDR calculation is needed
* VPC exists but doesn't have IPv6 CIDR -> No subnet will have a IPv6 CIDR so no calculation is needed
* VPC exists and has IPv6 CIDR -> `VPC.IPv6CIDR` will be available for calculation.

In the subnet `RenderAWS()` function we have the following case with regards to IPv6:
* VPC doesn't exists -> It is a dependency, VPC will always exist when it runs
* VPC exists but doesn't have IPv6 CIDR ->  it will retry until the CIDR allocated by `VPCAmazonIPv6CIDR` is available.
* VPC exists and has IPv6 CIDR -> `VPC.IPv6CIDR` will be available for calculation and `findVPCIPv6CIDR() will not be needed.

@johngmyers Do you see any possible race condition anymore?

upup/pkg/fi/utils/cidr.go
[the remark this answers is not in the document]
hakman: Does `ParseCIDRNotation()` sound better?

johngmyers:
The initial look seems quite promising. It will likely be 24 hours before I can give it a proper review.

hakman:
> The initial look seems quite promising. It will likely be 24 hours before I can give it a proper review.

Thanks, no rush.

upup/pkg/fi/cloudup/awstasks/subnet_test.go
```diff
@@ -139,6 +139,192 @@ func TestSubnetCreate(t *testing.T) {
...
+			t.Fatalf("Unexpected Subnet: expected=%v actual=%v", expected, actual)
+		}
+	}
+
+	{
+		allTasks := buildTasks()
+		checkNoChanges(t, cloud, allTasks)
+	}
+}
+
```
johngmyers, line 327: Missing a test for the case where we have to wait for VPCAmazonIPv6CIDRBlock

johngmyers (approved):

PR #11524, opened 2021-05-19
Don't download nodeup if already in the AMI

johngmyers:


olemarkus:
This is great :)
I have been meaning to create a nodeup configuration mode similar to warmpool, but that can be run with packer or similar to build AMIs. This would fit nicely with such a feature.

pkg/model/resources/nodeup.go
```diff
@@ -131,7 +130,10 @@ function download-release() {
   esac
 
   cd ${INSTALL_DIR}/bin
-  try-download-release
+  local -r nodeup_urls=( $(split-commas "${NODEUP_URL}") )
+  download-or-bust nodeup "${NODEUP_HASH}" "${nodeup_urls[@]}"
+
+  chmod +x nodeup
```
hakman, line 136: Probably best to move url (not used anywhere else) and chmod to `download-or-bust()` and keep `download-release()` clean.
```suggestion
  download-or-bust nodeup "${NODEUP_HASH}" "${nodeup_urls[@]}"
```

johngmyers: `download-or-bust` has two return points, so it is simpler to have the chmod outside it. Also, it's not completely impossible that we might later need to download a second, non-executable file before running nodeup.

johngmyers:
There's something to be said for having the wrong `nodeup` binary cause a bootstrap failure. In the case of an out-of-date AMI, a noticeable failure might be better than slowness.

hakman:
If nodeup is re-downloaded, in most cases means a few seconds. Most people should not notice that part.
If we stop a cluster from scaling or replacing nodes just because the AMI is outdated, seems we would create quite a big issue. Maybe printing a warning would make it easier for people to notice this in logs.

johngmyers:
A few seconds might be enough to blow your SLA. It might not be noticed until peak traffic and even then the cause wouldn't be obvious.

Failure to replace nodes on the rolling update following a configuration change would be much more likely to be noticed quickly, before it becomes a big problem.

hakman:
If anyone is that serious about not starting if nodeup is out of date, could have a boot script that checks the sha256 and stops the process.
😄

hakman (approved):
I am ok with this as is. Thanks.
In case you would like more feedback...
/hold

johngmyers:
I'm okay with this for now. We can iterate as follow-up.
/hold cancel

PR #11531, opened 2021-05-19
feat(openstack): enable configuration of servergroup affinities

mitch000001:
This enables us to change the ServerGroup affinity policies using
annotations on instance groups.

The default affinity policy still is "anti-affinity".


mitch000001:
This was tested in our Openstack and the ServerGroup affinity policies are applied successfully.
Screenshot 2021-05-19 at 14 53 33

zetaab (approved):
looks valid, thank you!

PR #11540, opened 2021-05-19
Support terraform 0.12+'s filebase64() in json output

rifelpet:
Originally the JSON output was meant as a bridge between the 0.11 and 0.12 support.
Now that we've dropped support for 0.11, we can use filebase64() instead of encoding the userdata in the file ourselves.


hakman (approved):
Yeeeeey!

PR #11542, opened 2021-05-20
Add support for arbitrary terraform functions

rifelpet:
This allows `terraformWriter.LiteralFunctionExpression` to be used with any terraform function.

The arguments must be strings, which means that any string arguments must include quotes. I was debating between this and using a slice of interfaces for `Literal.FnArgs` and passing each to `json.Marshal` to get their string representations (which would quote the strings for us). This would require either returning the `json.Marshal` error up through all of the HCL2 rendering (which doesn't currently return errors) or adding a panic which I'm not excited about.

Feedback welcome.

/hold

### DIFF
--- a/upup/pkg/fi/cloudup/terraform/hcl2.go
+++ b/upup/pkg/fi/cloudup/terraform/hcl2.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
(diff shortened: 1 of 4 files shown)


johngmyers:
The thing that concerns me is lack of quoting `\`, `${`, or `%{` in the arguments. Probably unlikely to exist in pathnames, but would we want to provide a better quoting function than `fmt.Sprintf("%q"`?

johngmyers:
Other than that, looks good. I don't consider that blocking.
/lgtm

hakman:
Same here.
/lgtm

rifelpet:
We can improve escaping in a followup. The HCL library may be able to help us there

/hold cancel

PR #11558, opened 2021-05-22
Fix deletion of IAM roles and policies

johngmyers:


upup/pkg/fi/cloudup/awstasks/iamrole.go
```diff
@@ -143,6 +143,87 @@ func (s *IAMRole) CheckChanges(a, e, changes *IAMRole) error {
...
+			klog.V(2).Infof("Detaching IAM role policy %q", policy)
+			request := &iam.DetachRolePolicyInput{
+				RoleName:  a.Name,
+				PolicyArn: policy.PolicyArn,
+			}
+			_, err := t.Cloud.IAM().DetachRolePolicy(request)
+			if err != nil {
+				return fmt.Errorf("error detaching IAM role policy %q: %v", *policy.PolicyArn, err)
+			}
+		}
```
johngmyers, line 216: I wonder if it works to `UpdateAssumeRolePolicy` to a PolicyDocument that grants nothing, but I consider that future cleanup work.

upup/pkg/fi/cloudup/awstasks/iamrole.go
```diff
@@ -143,6 +143,87 @@ func (s *IAMRole) CheckChanges(a, e, changes *IAMRole) error {
 }
 
 func (_ *IAMRole) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *IAMRole) error {
+	if e.RolePolicyDocument == nil {
+		klog.V(2).Infof("Deleting IAM role %q", a.Name)
```
olemarkus, line 147: What was the reason for doing deletion in `RenderAWS` rather than in a `fi.Deletion`? I believe you can just do the same check as above in `findDeletion` and all the logic below in `Delete`.

johngmyers: We want the dryrun behavior of `fi.Task`. For other task types we'll probably want the cloud provider selection.

A followup work item is to write a `fi.Deletion` adapter to `fi.Task`, so the cluster deletion code could be subsumed into the task reconciliation code for object such as this that have to be deleted in both cluster delete and apply.

olemarkus: Sounds good. For some tasks, we added `Delete` flag that sort of gives both the dryrun behaviour and `fi.Deletion`, but it would be nice if we could allow deletion without that property.

olemarkus (approved):

upup/pkg/fi/cloudup/awstasks/iamrole.go
johngmyers: We do need some way to represent the deleted state so that dryrun can print something useful. I do need to look at how dryrun represents `IamRole` changes, as it doesn't currently show well what attached policies are changing.

PR #11566, opened 2021-05-22
Set flags on AWS CCM mimicking KCM

olemarkus:
With this PR, my local setup is down to 13 failures.


pkg/model/components/awscloucontrollermanager.go
```diff
@@ -0,0 +1,79 @@
+/*
```
rifelpet, line 1: filename is missing the `d` in `cloud`

hakman (approved):

Issue #8823, opened 2020-03-31
Kops delete cluster command should provide lifecycle-overrides option


tobernguyen:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
1.17.0-beta.1

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-13T18:08:14Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
I created a cluster using existing AWS Instance Profiles (https://github.com/kubernetes/kops/blob/master/docs/iam_roles.md#use-existing-aws-instance-profiles)

```
kops update cluster ${CLUSTER_NAME} --yes --lifecycle-overrides IAMRole=ExistsAndWarnIfChanges,IAMRolePolicy=ExistsAndWarnIfChanges,IAMInstanceProfileRole=ExistsAndWarnIfChanges
```

Then I tried to delete the cluster:
```
kops delete cluster --name $CLUSTER_NAME --yes
```

**5. What happened after the commands executed?**
The `kops delete cluster` command tries to delete the attached AWS Instance Profiles, IAM Roles, and IAM Policies which were managed outside kops (it were provisioned by Terraform in our project).

**6. What did you expect to happen?**
`kops delete cluster` should provide `lifecycle-overrides` option like `kops update` command so we tell kops to not delete AWS Instance Profiles, IAM Roles, and Policies.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**
I don't think this is needed

**8. Please run the commands with most verbose logging by adding the `-v 10` flag.
[...]

PR #11568, opened 2021-05-23
Cleanup InstanceProfile only that have ownership tags in delete cluster

h3poteto:
Refs #8823 

I change the logic for deleting IAM Instance Profiles from name to using ownership tags. This change is using the same logic as #11497 .
After this change, kops avoid deleting IAM Instance Profiles which were managed outside kops when delete cluster.


johngmyers (approved):
Thanks!

PR #11571, opened 2021-05-23
Only allow deletion of snapshots owned by the cluster

olemarkus:
/cc @rifelpet


rifelpet (commented):
Does snapshot controller include this tag by default? I don't see the cluster name being passed in to the controller's args. 

Or is the snapshot's tags copied from the volume?

olemarkus:
The way snapshoting works is a bit involved. It is the CSI driver that does the actual snapshoting. The controller just watches CRDs and calls the api. 

Also, when I ran e2e tests, they pass. So I assume everything is tagged appropriately.

rifelpet (approved):

PR #11575, opened 2021-05-23
Avoid error when first creating VPC with IPv6

justinsb:
When the VPC isn't yet created, the search for associations was failing
as we were trying to pass an empty ID.


hakman (approved):
Ah, thanks for fixing this.

PR #11577, opened 2021-05-23
VPC: allow IPv6 attachment in update

justinsb:
Add support for attaching after initial creation.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/vpc.go
+++ b/upup/pkg/fi/cloudup/awstasks/vpc.go
@@ -201,6 +201,18 @@ func (_ *VPC) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *VPC) error {
 		e.ID = response.Vpc.VpcId
 	}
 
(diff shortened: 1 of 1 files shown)


hakman:
@justinsb Isn't this same as the code in `vpcamazonipv6cidrblock.go`?
https://github.com/kubernetes/kops/blob/e7b63882469d4c7a915cfa7a989a9e6ac684699e/upup/pkg/fi/cloudup/awstasks/vpcamazonipv6cidrblock.go#L109-L128

upup/pkg/fi/cloudup/awstasks/vpc.go
```diff
@@ -201,6 +201,18 @@ func (_ *VPC) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *VPC) error {
 		e.ID = response.Vpc.VpcId
 	}
 
+	if a != nil && e.AmazonIPv6 != nil && changes.AmazonIPv6 != nil {
```
johngmyers, line 204: Shouldn't this be done by `VPCAmazonIPv6CIDRBlock`?

PR #11579, opened 2021-05-23
WIP: auto assign cidrs

justinsb:


hakman:
@justinsb Could we only do this for new clusters, on creation? Would make things simpler and consistent, making model code more complex.

johngmyers:
Subsumed by #11523

PR #11605, opened 2021-05-26
Enable AWS EBS CSI driver by default

olemarkus:


hakman (approved):

hakman:
Thinking about this more, I am wondering why enable it by default only for newly created clusters?
/cc @rifelpet

PR #11610, opened 2021-05-27
Cleanup AWSEBSCSIDriver setup

hakman:
What this does:
* Generate AWSEBSCSIDriver model only when needed
* Don't pre-pull AWSEBSCSIDriver images in NodeUp

/cc @olemarkus @rifelpet


upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -239,6 +239,15 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		dest["EnableSQSTerminationDraining"] = func() bool { return *cluster.Spec.NodeTerminationHandler.EnableSQSTerminationDraining }
 	}
 
+	if cluster.Spec.CloudConfig != nil && fi.BoolValue(cluster.Spec.CloudConfig.ManageStorageClasses) {
```
olemarkus, line 242: This sort of logic shouldn't be template functions. Template functions aren't scoped to any particular component or logic and makes things harder to maintain. That is precisely why we have option builders.

hakman: If you mean that `fi.BoolValue(cluster.Spec.CloudConfig.ManageStorageClasses)` part should be removed, I agree.
If you mean that we should always materialize `AWSEBSCSIDriver.Enabled`, I am not so sure.

olemarkus: We have to materialize it. Or make the templatest more complex.

The problem with the materialization now is that it triggers an unnecessary change to the bootstrap script. I think we can rather restrict what from `CloudConfig` is added to that.

pkg/model/components/awsebscsidriver.go
```diff
@@ -31,23 +31,16 @@ var _ loader.OptionsBuilder = &AWSEBSCSIDriverOptionsBuilder{}
 
 func (b *AWSEBSCSIDriverOptionsBuilder) BuildOptions(o interface{}) error {
 	clusterSpec := o.(*kops.ClusterSpec)
-	cc := clusterSpec.CloudConfig
-	if cc.AWSEBSCSIDriver == nil {
-		cc.AWSEBSCSIDriver = &kops.AWSEBSCSIDriver{
-			Enabled: fi.Bool(false),
```
olemarkus, line 37: When the driver is enabled by default, the code will stay like this, but flipping this bool.

hakman: Sounds fair. Just more complicated than I expected.

hakman:
@olemarkus Any other thoughts about this?

nodeup/pkg/model/awsebscsidriver.go
```diff
@@ -1,58 +0,0 @@
...
-// Build is responsible for configuring the EBS CSI Driver stuff
-func (b *AWSEBSCSIDriverBuilder) Build(c *fi.ModelBuilderContext) error {
-	csi := b.Cluster.Spec.CloudConfig.AWSEBSCSIDriver
-
-	if csi == nil || !fi.BoolValue(csi.Enabled) {
-		return nil
-	}
-
-	// Pulling CSI driver image
-	image := "k8s.gcr.io/provider-aws/aws-ebs-csi-driver:" + *csi.Version
```
johngmyers, line 39: There's value in doing the warmpull. We should keep this, but figure out why we're dereferencing nil here.

johngmyers (changes requested):

nodeup/pkg/model/awsebscsidriver.go
hakman: I disagree in this case. This should not have been done at all for anything other than CNI and proxy, which affect startup.
A generic way, with container images assets, would be the way to go forward.

upup/pkg/fi/cloudup/template_functions.go
olemarkus: Think this was answered in office hours. Short version is that I wanted e2e to pass before enabling by default for all clusters. That work is almost done now.

nodeup/pkg/model/awsebscsidriver.go
olemarkus: If you have pending workloads with volumes this will speed up pod startup. So while it does not contribute to node startup, it contributed to pod scheduling time.

upup/pkg/fi/cloudup/template_functions.go
hakman: Agreed, answered and the PR was was updated.

nodeup/pkg/model/awsebscsidriver.go
olemarkus: The nil dereferencing works because it assumes cloudup always set this value. That is the case now, but I wouldn't exactly call this very defensive.

hakman: This is a general argument for any pre-pulled image. Realistically, the pull time for these images is very small.
As I said, I don't disagree with the idea, if done correctly. I am not ok with using it in this case and having extra logic in NodeUp for a special case.

olemarkus: Note quite. There are three images pulled in succession, and one of them sometimes takes up to 20 seconds.
A random example:
```
FirstSeen              Count    From      Type     Reason      Message
<nil>                  <none>   <none>    Normal   Scheduled   Successfully assigned kube-system/ebs-csi-node-9sbqq to ip-172-20-98-90.eu-central-1.compute.internal
2021-06-03T17:34:17Z   1        kubelet   Normal   Pulling     Pulling image "k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v1.0.0"
2021-06-03T17:34:20Z   1        kubelet   Normal   Pulled      Successfully pulled image "k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v1.0.0" in 3.187918027s
2021-06-03T17:34:21Z   1        kubelet   Normal   Created     Created container ebs-plugin
2021-06-03T17:34:21Z   1        kubelet   Normal   Started     Started container ebs-plugin
2021-06-03T17:34:21Z   1        kubelet   Normal   Pulling     Pulling image "k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.1.0"
2021-06-03T17:34:34Z   1        kubelet   Normal   Pulled      Successfully pulled image "k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.1.0" in 12.506681986s
2021-06-03T17:34:34Z   1        kubelet   Normal   Created     Created container node-driver-registrar
2021-06-03T17:34:34Z   1        kubelet   Normal   Started     Started container node-driver-registrar
2021-06-03T17:34:34Z   1        kubelet   Normal   Pulling     Pulling image "k8s.gcr.io/sig-storage/livenessprobe:v2.2.0"
2021-06-03T17:34:35Z   1        kubelet   Normal   Pulled      Successfully pulled image "k8s.gcr.io/sig-storage/livenessprobe:v2.2.0" in 1.094773878s
2021-06-03T17:34:35Z   1        kubelet   Normal   Created     Created container liveness-probe
2021-06-03T17:34:36Z   1        kubelet   Normal   Started     Started container liveness-probe
```

hakman: Probably we can debate this forever. I doubt any of us will change its mind. Hopefully this will change in the near future to something more maintainable.

PR #11617, opened 2021-05-28
Add "kops get assets" command

johngmyers:
This is rough around the edges, but gives the idea.


hakman (changes requested):
Small suggestion, maybe skip the `mirror` field when not needed.
Also, would be nice to move this to V(1):
```
I0529 08:59:26.979910   39040 dns.go:97] Private DNS: skipping DNS validation
```

johngmyers:
I support suppressing chatter like that DNS log line, though I think that's perhaps a different PR.

On the mirror field, I'm a little wary of requiring consumers of that data to implement special cases for edge cases like this.
Perhaps I'm using the wrong terminology? Should I stick with "Download" and "Canonical"?

I could see a desire for less information when downloading from the canonical location, but then I don't see any use cases for this command in that situation.

hakman:
I understood perfectly the terminology, no worries. Though the name of the field "mirror" is misleading.
One can use this to inspect what assets will be downloaded in advance, for example to see what CNI or addon versions will be used by default. Another use case (more popular these days) is for building AMIs with these assets pre-loaded.

johngmyers:
@hakman the concern wasn't about understanding, but with which terminology was appropriate for the feature. I've changed it to be more consistent and accurate about what the values are.

hakman (approved):
Agreed. Thanks!

johngmyers:
It's a bit late to add features to 1.21.

PR #11628, opened 2021-05-30
Pre-install nvidia container runtime + drivers on GPU instances

olemarkus:
* [x] install nvidia container runtime
* [x] install nvidia drivers
* [x] make it opt-in to avoid clash with those who prebake/use gpu operator or similar
* [x] add device plugin as addon


olemarkus:
Worth mentioning that we override the containerd configuration override. Or at least modify it. kOps only knows about if the instance will have GPUs during nodeup, while cloudup will set the config override to the kOps default config by default (hence the containerd config override will always be set to the assumed "final" config).

hakman:
The idea is good in general, we should make this easy for people.

In particular, here are some of my thoughts:
* I think the vision for nodeup is to become dumb, as dump as possible. This is why we moved the containerd logic out of it. I think you can call it defensive programming, but probably it's a longer topic that would be good for tomorrow's office hours.
* Maybe add this as a new option of the container runtime, allowing to override of the runtime to something other than runc. Describing instance types just for this purpose seems wasteful for every node. It can also be done much earlier when generating the cloud model.
* The APT source task, maybe would be better just as APT key instead and a separate file task for the repo file.
* The APT list is focused on Ubuntu, there is a different list for debian10 for example.
* The APT key should be sent as a file asset url, same as the containerd package.
* containerd config is done at the moment outside nodeup. The reason it didn't change was because of logic moving to nodeup that we want to avoid. Again, topic for office hours. Though, it either should move completely to nodeup or manage it outside it as we do now.

olemarkus:
>     * I think the vision for nodeup is to become dumb, as dump as possible. This is why we moved the containerd logic out of it. I think you can call it defensive programming, but probably it's a longer topic that would be good for tomorrow's office hours.
> 
>     * Maybe add this as a new option of the container runtime, allowing to override of the runtime to something other than runc. Describing instance types just for this purpose seems wasteful for every node. It can also be done much earlier when generating the cloud model.
> 

The above has a challenge if someone creates a mixed instance group with both GPU and non-GPU instance types. I am not sure why anyone would do that, but it is possible, and if nvidia runtime is used on non-GPU, it will break.

That being said, we could do some hardware probing instead, I guess. Just seemed more reliable to describe the instance type.

>     * The APT source task, maybe would be better just as APT key instead and a separate file task for the repo file.
> 
>     * The APT list is focused on Ubuntu, there is a different list for debian10 for example.
> 

Yeah, I am aware. I plan on adding other distros as needed and in the interim solve this with some distro checks and documentation.

>     * The APT key should be sent as a file asset url, same as the containerd package.

Thanks. I didn't like the way I did this one. asset file url is a much better idea.

>     * containerd config is done at the moment outside nodeup. The reason it didn't change was because of logic moving to nodeup that we want to avoid. Again, topic for office hours. Though, it either should move completely to nodeup or manage it outside it as we do now.

If we find a way to do this in cloudup I'd also like to see if this can be put the final config aux. I don't like using the override flag as a carry mechanism cloudup -> nodeup.

hakman:
> > ```
> > * I think the vision for nodeup is to become dumb, as dump as possible. This is why we moved the containerd logic out of it. I think you can call it defensive programming, but probably it's a longer topic that would be good for tomorrow's office hours.
> > 
> > * Maybe add this as a new option of the container runtime, allowing to override of the runtime to something other than runc. Describing instance types just for this purpose seems wasteful for every node. It can also be done much earlier when generating the cloud model.
> > ```
> 
> 
> The above has a challenge if someone creates a mixed instance group with both GPU and non-GPU instance types. I am not sure why anyone would do that, but it is possible, and if nvidia runtime is used on non-GPU, it will break.
> 
> 
> That being said, we could do some hardware probing instead, I guess. Just seemed more reliable to describe the instance type.
> 
> 

No need for that, we have validation. If we can validate that instances should not have a mix of ARM and AMD, we can make sure GPU instances are validated in same way

> > ```
> > * The APT source task, maybe would be better just as APT key instead and a separate file task for the repo file.
> > 
> > * The APT list is focused on Ubuntu, there is a different list for debian10 for example.
> > ```
> 
> 
> Yeah, I am aware. I plan on adding other distros as needed and in the interim solve this with some distro checks and documentation.
> 
> 

Meant that the code checks for Debian family instead of Ubuntu specific

> > ```
> > * The APT key should be sent as a file asset url, same as the containerd package.
> > ```
> 
> 
> Thanks. I didn't like the way I did this one. asset file url is a much better idea.
> 
> 

👍 

> > ```
> > * containerd config is done at the moment outside nodeup. The reason it didn't change was because of logic moving to nodeup that we want to avoid. Again, topic for office hours. Though, it either should move completely to nodeup or manage it outside it as we do now.
> > ```
> 
> 
> If we find a way to do this in cloudup I'd also like to see if this can be put the final config aux. I don't like using the override flag as a carry mechanism cloudup -> nodeup.

I agree, but to keep this on track we can do it later, when the config aux is merged. I agree that aux would be a better place for it.

olemarkus:
So if we don't allow mixing GPU and non-GPU instance types, building the config cloudup should be fine.
I would need to send some setting to nodeup telling it to do those package installs. Again seems like something worth putting on aux.

hakman:
A prerequisite would be to allow configuring containerd for each IG, not sure if possible now. Once that is possible, the package installs can just be done in the `ContainerRuntime` builder. Package installs would still be hardcoded I guess for now based on OS type.

olemarkus:
Right. I think that is a challenge with how nodeup config is being built. Probably the containerd config should be built based on IG and cluster spec and then written to one of the configs rather than writing it back to the IG/cluster spec. But we are a long way away from that.

I think continuing the current path makes sense for now. There is nothing here that is exposed to the user, so things can be moved later on (maybe even before 1.22 GA)

olemarkus:
I didn't do many of the APT changes. Using assets was challenging as it is arch dependent, which doesn't make sense in this case. It could also run into run-time issues if we use hashes (upstream can change keys) and it is the TLS cert we trust here.

olemarkus:
I'll defer device addon to another PR.

johngmyers:
Determining the arch of an instance and whether or not it has a GPU appears to be firmly in the bailiwick of nodeup. I don't see why there's an objection to putting such logic in there. Making the admin configure this through the API is just causing them grief.

I do not understand or agree with this desire to make nodeup dumb. It should do the things it is well suited for and not do the things it is not well suited for.

hakman:
GPU stuff is something that very few operators need or use. Checking each node in every cluster if it has that capability just for that is not exactly ideal. More over, this has to work on the various supported platforms, not just AWS.

johngmyers:
Can't nodeup to the equivalent of `lspci` or somesuch?

pkg/apis/nodeup/config.go
```diff
@@ -76,6 +76,8 @@ type Config struct {
 
 	// APIServerConfig is additional configuration for nodes running an APIServer.
 	APIServerConfig *APIServerConfig `json:",omitempty"`
+	// Nvidia contains the configuration for nvidia
+	Nvidia *kops.NvidiaConfiguration
```
hakman, line 80: Should it be committed if empty?
```suggestion
	Nvidia *kops.NvidiaConfiguration `json:",omitempty"`
```

olemarkus:
Maybe a start is to agree on how this should behave from user perspective? I would prefer that we set driver version and (skip)install flag on cluster level similar to docker and containerd, and then just ignore that struct on non-GPU nodes. Ignoring that struct can be done in cloudup though.

Doing hardware-probing may be doable, but yeah, typically involves grepping lspci or lshw, and the latter is at least going to be slower, and is somewhat fragile (can accidentally match other hardware, can miss some hardware if AWS starts supporting other cards). I'd rather not go down that path when the API gives a very clear response.

So I'd rather use the AWS API either in cloudup or in nodeup.

hakman:
@olemarkus would something like this work for detection?
https://github.com/google/cadvisor/blob/master/accelerators/nvidia.go#L77

olemarkus:
Yeah. We could test for `0x10de`. But then I want to ggenerate ccontainerd config in nodeup so the same test is done there.

olemarkus:
Where do we go with this?
Move nvidia config to the cluster resource (next to the other runtimes) and use device detection in nodeup?

Containerd config in nodeup?

hakman:
> Where do we go with this?
> Move nvidia config to the cluster resource (next to the other runtimes) and use device detection in nodeup?
> Containerd config in nodeup?

I would put it in containerd config, as we may get other CRI support later and don't want to confuse users.

pkg/apis/kops/cluster.go
```diff
@@ -208,12 +208,20 @@ type ClusterSpec struct {
...
 	// ServiceAccountIssuerDiscovery configures the OIDC Issuer for ServiceAccounts.
 	ServiceAccountIssuerDiscovery *ServiceAccountIssuerDiscoveryConfig `json:"serviceAccountIssuerDiscovery,omitempty"`
-
 	// SnapshotController defines the CSI Snapshot Controller configuration.
 	SnapshotController *SnapshotControllerConfig `json:"snapshotController,omitempty"`
+	// NvidiaConfiguration configures the Nvidia GPU runtime.
+	Nvidia *NvidiaConfiguration `json:"nvidia,omitempty"`
+}
+
+type NvidiaConfiguration struct {
```
hakman, line 219: Everything else has the `Config` suffix. Also, there is a mix of Nvidia, GPU and Nvidia GPU terminology.
```suggestion
type NvidiaConfig struct {
```

pkg/apis/kops/cluster.go
```diff
@@ -208,12 +208,20 @@ type ClusterSpec struct {
...
+	// Package is the name of the nvidia driver package that will be installed.
+	// Default is nvidia-kernel-common-460-server.
+	DriverPackage string `json:"package,omitempty"`
+	// Install determines if kOps will install the Nvidia GPU runtime
+	Install *bool `json:"install,omitempty"`
```
hakman, line 224: Again, to be consistent.
```suggestion
	Enable *bool `json:"install,omitempty"`
```

nodeup/pkg/model/context.go
```diff
@@ -693,3 +693,7 @@ func (c *NodeupModelContext) CNIBinDir() string {
 func (c *NodeupModelContext) CNIConfDir() string {
 	return "/etc/cni/net.d/"
 }
+
+func (c *NodeupModelContext) InstallNvidiaRuntime() bool {
```
hakman, line 697: Maybe `UseNvidiaGPU` would be a better name:
```suggestion
func (c *NodeupModelContext) UseNvidiaGPU() bool {
```

nodeup/pkg/model/context.go
```diff
@@ -693,3 +693,7 @@ func (c *NodeupModelContext) CNIBinDir() string {
...
+	return c.NodeupConfig.Nvidia != nil && fi.BoolValue(c.NodeupConfig.Nvidia.Install)
```
hakman, line 698: I think in most cases we assume true if Enable / Install is nil.
```suggestion
	return c.NodeupConfig.Nvidia != nil && (c.NodeupConfig.Nvidia.Install == nil || fi.BoolValue(c.NodeupConfig.Nvidia.Install))
```

upup/pkg/fi/nodeup/nodetasks/aptsource.go
```diff
@@ -0,0 +1,82 @@
...
+	return fi.DefaultDeltaRunMethod(f, c)
+}
+
+func (*AptSource) CheckChanges(a, e, changes *AptSource) error {
+	return nil
+}
+
+func (f *AptSource) RenderLocal(t *local.LocalTarget, a, e, changes *AptSource) error {
+
+	filename := "/tmp/" + f.Name + ".gpg"
```
hakman, line 59:
```suggestion
      tmpDir, err := ioutil.TempDir("", "aptsource")
	if err != nil {
		return fmt.Errorf("error creating temp dir: %v", err)
	}
	defer func() {
		if err := os.RemoveAll(tmpDir); err != nil {
			klog.Warningf("error deleting temp dir %q: %v", tmpDir, err)
		}
	}()
	filename := path.Join(tmpDir, f.Name + ".gpg")
```

nodeup/pkg/model/containerd.go
```diff
@@ -456,5 +463,35 @@ func (b *ContainerdBuilder) buildContainerdConfig() string {
...
+
+	if b.InstallNvidiaRuntime() {
+		if err := appendGPURuntimeContainerdConfig(config); err != nil {
+			return "", err
+		}
+	}
+	return config.String(), nil
+}
+
+func appendGPURuntimeContainerdConfig(config *toml.Tree) error {
```
hakman, line 475: It is Nvidia specific and already in the containerd builder.
```suggestion
func appendNvidiaGPURuntimeConfig(config *toml.Tree) error {
```

hakman (changes requested):

hakman:
> /hold for missing device detection

Detection code can also be added separately in a later PR. Would be just a case of updating the `UseNvidiaGPU` function.

olemarkus:
> > Where do we go with this?
> > Move nvidia config to the cluster resource (next to the other runtimes) and use device detection in nodeup?
> > Containerd config in nodeup?
> 
> I would put it in containerd config, as we may get other CRI support later and don't want to confuse users.

Well. Any other CRI that can make use of runc can also use the nvidia runtime.
But if we enable nvidia by default in the cluster spec, this can become awkward as one would have to explicitly disable nvidia (or we silently ignore nvidia configs when that CRI is used). It may make sense to disable nvidia by default on cluster level. Or enable it by default only on new clusters if a supported CRI is used.

I think I would only enable nvidia by default on new clusters if containerd is used. Then validation can handle other cases.

hakman:
> > > Where do we go with this?
> > > Move nvidia config to the cluster resource (next to the other runtimes) and use device detection in nodeup?
> > > Containerd config in nodeup?
> > 
> > 
> > I would put it in containerd config, as we may get other CRI support later and don't want to confuse users.
> 
> Well. Any other CRI that can make use of runc can also use the nvidia runtime.
> But if we enable nvidia by default in the cluster spec, this can become awkward as one would have to explicitly disable nvidia (or we silently ignore nvidia configs when that CRI is used). It may make sense to disable nvidia by default on cluster level. Or enable it by default only on new clusters if a supported CRI is used.
> 
> I think I would only enable nvidia by default on new clusters if containerd is used. Then validation can handle other cases.

I think you are thinking too far in the future. For now, Nvidia GPU should be containerd only, nothing else. One can disable/enable at containerd cluster level and will be fine. Don't need to add another awkward setting at cluster level.

nodeup/pkg/model/context.go
olemarkus: What we do is:
* install it
* configure it
* set it to be used by default by containerd

But from kOps perspective, we don't really use it. Containerd will most likely use it, but technically a user can choose not to.

hakman: It doesn't install it either, it sets it up.
You can see it also as the desire to use the local CPU with k8s.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -304,6 +309,20 @@ func NewCluster(opt *NewClusterOptions, clientset simple.Clientset) (*NewCluster
 	return &result, nil
 }
 
+func setupRuntimes(opt *NewClusterOptions, cluster *api.Cluster) {
```
olemarkus, line 312: I wonder if not we should just set containerd here. It means we cannot change the implicit default later on, but I think changing the default would be very far into the future, and likely will involve changing default for new clusters first anyway.

hakman: We can't. We just remove something similar from here.
It is not a good idea to do this, and materialize it for all new clusters. If you really want it enabled by default maybe consider enabled when not set explicitly.

olemarkus: Well we can. It is just a matter of setting it here.

We cannot enable  nvidia on existing clusters though, so this would be the most helpful. Or decide it must always be enabled explicitly.

hakman: I don't mind enabling it explicitly. I don't mind enabling it on existing clusters either.
I think it is a really low risk move, considering you actually have to use GPU instances for this.

olemarkus: If a cluster is using GPU operator, it may break things. It could break things on prebaked images if they expect other drivers or other paths as well. I am not sure it is that safe to just enable it on existing clusters and I fear those who already install things their own way will not pick up this change from the release notes.

olemarkus: My idea with adding the nvidia conf here was to make it "just work" for those using `kops create cluster`. Those using specs would have to enable this explicitly in all cases.

It is not a must to keep this particular commit.

It could also be an option to add a flag, but ... seems a bit much.

hakman: OK, then enabling explicitly is the way, from my point of view.
If we get feedback that "it's hard to use/enable" we can discuss more.

nodeup/pkg/model/context.go
olemarkus: I am not sure I follow this one. This function determines if the driver and runtime packages are installed. As in getting `apt-get install`ed. It is used to configure containerd but that is a smallish part of it.

nodeup/pkg/model/context.go
olemarkus: Scanning `bootstrapchannelbuilder.go` we don't assume that for most addons.

nodeup/pkg/model/context.go
hakman: This function should check if Nvidia "stuff" is enabled or not, not if installed. Install is just a consequence of it being enabled and supported by the hardware.

nodeup/pkg/model/packages.go
```diff
@@ -33,11 +33,13 @@ var _ fi.ModelBuilder = &DockerBuilder{}
 
 // Build is responsible for installing packages
 func (b *PackagesBuilder) Build(c *fi.ModelBuilderContext) error {
+
```
hakman, line 36: This file should have no changes.

util/pkg/architectures/architectures.go
```diff
@@ -29,6 +29,12 @@ const (
 	ArchitectureArm64 Architecture = "arm64"
 )
 
+type GPUVendor string
+
+const (
+	GPUVendorNvidia GPUVendor = "nvidia"
+)
+
```
hakman, line 37: This has absolutely nothing to do with architecture. It is a sidenote at best that should be confined to a few files in nodeup.

upup/pkg/fi/nodeup/command.go
```diff
@@ -269,6 +270,24 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
...
+		// Get the instance type's detailed information.
+		instanceType, err := awsup.GetMachineTypeInfo(awsCloud, modelContext.MachineType)
+		if err != nil {
+			return err
+		}
+
+		if instanceType.GPU {
+			klog.Info("instance supports GPU acceleration")
+			modelContext.GPUVendor = architectures.GPUVendorNvidia
+		}
```
hakman, line 289: Why does this have to be done for every system? Isn't this only needed if Nvidia is enabled in conf?
Seems like a replay of the previous commit. In short, I don't understand why you need autodetection.

hakman:
It's been a long time and, sorry to say, I think we are not in sync on this.

Based on Office Hours chat about this:
1. This should NOT be enabled by default.
2. This should be OPT-IN, not opt-out at least for the first few releases, like we do with most new features.
3. Having `Nvidia` or `Containerd.Nvidia` should be thought a little more, probably some other people should think about it (maybe @johngmyers and @rifelpet).

Given the above, it removes the need for any hardware detection in this PR. I only see a need for it as a doublecheck function if Nvidia is enabled at cluster level and so it can skip nodes where there's no support.

olemarkus:
> It's been a long time and, sorry to say, I think we are not in sync on this.
> 
> Based on Office Hours chat about this:
> 
>     1. This should NOT be enabled by default.
> 
>     2. This should be OPT-IN, not opt-out at least for the first few releases, like we do with most new features.
> 
>     3. Having `Nvidia` or `Containerd.Nvidia` should be thought a little more, probably some other people should think about it (maybe @johngmyers and @rifelpet).
> 
> 
> Given the above, it removes the need for any hardware detection in this PR. I only see a need for it as a doublecheck function if Nvidia is enabled at cluster level and so it can skip nodes where there's no support.

1 and 2 should be fine. Users do have to explicitly enable this in the spec.
3. If nvidia is moved under containerd, how would this look if we at some point implement support for cri-o or any other CRI that can use runc/nvidia-container-runtime? cri-o support had some traction a whlie back so it is not that infeasible.

I thought we already covered the need for device detection. We want this config to be on the cluster level, and then we need to know which instances should use runc by default and which should use nvidia-container-runtime.

olemarkus:
What we can do, and most likely will be needed anyway in a PR is to do detection in cloudup. The reason for that is I plan on making it possible to install RuntimeClass objects, and these need a selector/toleration. So cloudup need to configure some labels that are applied to IGs with GPUs.

hakman:
> 3. If nvidia is moved under containerd, how would this look if we at some point implement support for cri-o or any other CRI that can use runc/nvidia-container-runtime? cri-o support had some traction a whlie back so it is not that infeasible.

Same way we deal with Docker now, it is completely separate. We have 2 options:
* add Nvidia settings to each CRI, based on how and if we want to support it
* add Nvidia settings at cluster level and keep adding exceptions in validation.
Seeing it as CRI specific feature and makes more sense to me.

> I thought we already covered the need for device detection. We want this config to be on the cluster level, and then we need to know which instances should use runc by default and which should use nvidia-container-runtime.

I thought we did. I understood that we would enable at cluster level and each node will decide if it can & will install and configure Nvidia. This only requires 1 function to detect Nvidia support, run only after checking that it has been enabled in cluster spec. Adding another constant for it, adding a new field for GPU vendor in nodeup and running the check even if the Nvidia feature is not enabled seems overkill to me.

hakman:
> What we can do, and most likely will be needed anyway in a PR is to do detection in cloudup. The reason for that is I plan on making it possible to install RuntimeClass objects, and these need a selector/toleration. So cloudup need to configure some labels that are applied to IGs with GPUs.

I would raise this in office hours. If we go ahead with it, it will require some thoughts on where and how to do it. Should take into account other cloud providers too, at least to be easy to add support for them.

olemarkus:
I originally through to do the device plugin addon in a separate PR, but it seemed easy enough to add it to this one.
Once this PR goes in, a cluster can accept GPU workloads that have the correct RuntimeClass set.

olemarkus:
What do others do:
* Rancher use [GPU operator or refer to nvidia device plugin](https://rancher.com/blog/2020/get-up-and-running-with-nvidia-gpus)
* [EKS uses nvidia runtime](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html#gpu-ami) 
* [AKS uses nvidia's runtime](https://docs.microsoft.com/en-us/azure/aks/gpu-cluster#update-your-cluster-to-use-the-aks-gpu-image-preview)
* [Kubespray use the Google approach](https://github.com/kubernetes-sigs/kubespray/tree/master/roles/kubernetes-apps/container_engine_accelerator/nvidia_gpu/templates) (but no change [since 2019](https://github.com/kubernetes-sigs/kubespray/blame/daed3e5b6a085ac99e076b51d314fcf76e4127b4/roles/kubernetes-apps/container_engine_accelerator/nvidia_gpu/defaults/main.yml))
* [Minikube uses a mix. Google for installing the driver, but the nvidia device plugin](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/assets/addons.go#L419)

My preference is still to go with the nvidia approach. It's use seems more widespread and personally I understand more of the dependencies. It decouples installing the driver from running the container runtime, which also makes it easier to install this on other distros as needed.

upup/pkg/fi/cloudup/populate_instancegroup_spec.go
```diff
@@ -161,6 +161,20 @@ func PopulateInstanceGroupSpec(cluster *kops.Cluster, input *kops.InstanceGroup,
...
+		mt, err := awsup.GetMachineTypeInfo(cloud.(awsup.AWSCloud), ig.Spec.MachineType)
+		if err != nil {
+			return ig, fmt.Errorf("error looking up machine type info: %v", err)
+		}
+		if mt.GPU {
+			if ig.Spec.NodeLabels == nil {
+				ig.Spec.NodeLabels = make(map[string]string)
+			}
+			ig.Spec.NodeLabels["kops.k8s.io/gpu"] = "1"
+			ig.Spec.Taints = append(ig.Spec.Taints, "nvidia.com/gpu:NoSchedule")
```
justinsb, line 175: I think this is a behavioral change?  i.e. workloads that previously ran on GPU instances will need to add this toleration to remain schedulable?

Or is there a webhook or similar that will add this automatically?

pkg/model/iam/iam_builder.go
```diff
@@ -765,6 +765,7 @@ func addNodeupPermissions(p *Policy, enableHookSupport bool) {
 	addASLifecyclePolicies(p, enableHookSupport)
 	p.unconditionalAction.Insert(
 		"ec2:DescribeInstances", // aws.go
+		"ec2:DescribeInstanceTypes",
```
justinsb, line 768: I was originally a little sad that we couldn't get this from the metadata or DescribeInstances.  But I came round to your point of view, because InstanceTypes are generic data that won't really be sensitive, and if there's one we should get rid of it's probably DescribeInstances :-)

justinsb:
I think this looks great, and as we agreed in office hours we can nest it under containerd.

My one remaining challenge is the automatic taint, because I'm worried that we'll break clusters that upgrade that are using GPU instances until users add that toleration.  Or is there some other mechanism at play here: e.g. everyone is already using this toleration, or there's a webhook that adds it?

upup/pkg/fi/cloudup/populate_instancegroup_spec.go
olemarkus: Good catch. I think this one should additionally be blocked by NvidiaGPU.enabled. That way one has explicitly subscribed to kOps way of doing things. I'll add docs to match as well.

olemarkus:
These concerns should be addressed now.

justinsb:
Thanks @olemarkus 

This lgtm now and we did agree on this at our last discussion; I'm going to approve but hold until after office hours.

/approve
/lgtm
/hold

PR #11629, opened 2021-05-30
Add --ipv6 experimental cli flag

hakman:
/hold for #11523

/cc @johngmyers @justinsb

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -260,6 +260,10 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	// TODO: Can we deprecate this flag - it is awkward?
 	cmd.Flags().BoolVar(&associatePublicIP, "associate-public-ip", false, "Specify --associate-public-ip=[true|false] to enable/disable association of public IP for master ASG and nodes. Default is 'true'.")
 
+	if featureflag.AWSIPv6.Enabled() {
+		cmd.Flags().BoolVar(&options.IPv6, "ipv6", false, "Allocate IPv6 CIDRs to sunets for clusters with public topology on AWS")
+	}
+
 	cmd.Flags().StringSliceVar(&options.NodeSecurityGroups, "node-security-groups", options.NodeSecurityGroups, "Add precreated additional security groups to nodes.")
 	cmd.Flags().StringSliceVar(&options.MasterSecurityGroups, "master-security-groups", options.MasterSecurityGroups, "Add precreated additional security groups to masters.")
 
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -91,6 +91,8 @@ type NewClusterOptions struct {
 	UtilitySubnetIDs []string
 	// Egress defines the method of traffic egress for subnets.
 	Egress string
+	// IPv6 adds IPv6 CIDRs to subnets
+	IPv6 bool
 
 	// OpenstackExternalNet is the name of the external network for the openstack router.
 	OpenstackExternalNet     string
@@ -940,6 +942,19 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
 
 		for i := range cluster.Spec.Subnets {
 			cluster.Spec.Subnets[i].Type = api.SubnetTypePublic
+		}
+
+		if opt.IPv6 {
+			if api.CloudProviderID(cluster.Spec.CloudProvider) == api.CloudProviderAWS {
+				klog.Warningf("IPv6 support is EXPERIMENTAL and can be changed or removed at any time in the future!!!")
+				for i := range cluster.Spec.Subnets {
+					// Start IPv6 CIDR numbering from "1" to reserve /64#0 for later use
+					// with NonMasqueradeCIDR, ClusterCIDR and ServiceClusterIPRange
+					cluster.Spec.Subnets[i].IPv6CIDR = fmt.Sprintf("/64#%x", i+1)
+				}
+			} else {
+				klog.Errorf("IPv6 support is available only on AWS")
+			}
 		}
 
 	case api.TopologyPrivate:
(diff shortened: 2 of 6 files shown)


upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -934,6 +936,13 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
 			cluster.Spec.Subnets[i].Type = api.SubnetTypePublic
 		}
 
+		if opt.IPv6 && api.CloudProviderID(cluster.Spec.CloudProvider) == api.CloudProviderAWS {
```
johngmyers, line 939: Perhaps error out if the option is set and the provider doesn't support it?

cmd/kops/create_cluster.go
```diff
@@ -260,6 +260,10 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	// TODO: Can we deprecate this flag - it is awkward?
 	cmd.Flags().BoolVar(&associatePublicIP, "associate-public-ip", false, "Specify --associate-public-ip=[true|false] to enable/disable association of public IP for master ASG and nodes. Default is 'true'.")
 
+	if featureflag.AWSIPv6.Enabled() {
+		cmd.Flags().BoolVar(&options.IPv6, "ipv6", false, "Add IPv6 CIDRs to AWS clusters with public topology")
```
johngmyers, line 264: I don't see any restriction to public topology.

```suggestion
		cmd.Flags().BoolVar(&options.IPv6, "ipv6", false, "Allocate IPv6 CIDRs to subnets")
```

johngmyers (changes requested):

cmd/kops/create_cluster.go
hakman: There is no restriction, but there is no allocation of subnets either.
At the moment, there is egress only internet gateway created so, private instances will have some trouble with internet access.

johngmyers: I'm failing to see why this shouldn't allocate IPv6 CIDRs to subnets in Private topology.

There is a separate task to provide IPv6 egress for the private route table.

hakman: The node could not be bootstrapped because of the missing egress gateway.
I am trying to get this working with public topologies and then enable other use cases.
Other than that, this is a helper flag, for convenience. Once an always experiment by modifying the cluster config, where there isn't any limitation.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -934,6 +936,17 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
 			cluster.Spec.Subnets[i].Type = api.SubnetTypePublic
 		}
 
+		if opt.IPv6 {
+			if api.CloudProviderID(cluster.Spec.CloudProvider) == api.CloudProviderAWS {
+				klog.Warningf("IPv6 support is EXPERIMENTAL and can be changed or removed at any time in the future!!!")
+				for i := range cluster.Spec.Subnets {
+					cluster.Spec.Subnets[i].IPv6CIDR = fmt.Sprintf("/64#%x", i)
```
johngmyers, line 943: We should reserve `/64#0` for the ClusterCIDR.

hakman: Sounds good to me, will change.

johngmyers: For future reference, read `NonMasqueradeCIDR` instead of `ClusterCIDR`. The latter is carved out of the former.

hakman: Yes, got it, though, we should discuss (again) about it. `NonMasqueradeCIDR` may go away soon as it's part of the dockershim implementation IIRC. Seems used just by the kubenet family.

johngmyers: The main use that is relevant here is that it's what `populateClusterSpec.assignSubnets()` carves the default `ClusterCIDR` and `ServiceClusterIPRange` out from.

johngmyers (approved):
/hold cancel

hakman:
@johngmyers This should be good to go now. Please take another look.

cmd/kops/create_cluster.go
```diff
@@ -500,6 +504,10 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 		cluster.Spec.NetworkCIDR = c.NetworkCIDR
 	}
 
+	if c.IPv6 {
+		cluster.Spec.NonMasqueradeCIDR = "fd00:10:96::/56"
```
johngmyers, line 508: Why this address? The `10:96` is not random as the spec requires.

johngmyers: If we are going to automatically assign the `NonMasquereadeCIDR` it should come from the VPC's IPv6 allocation. We should not automatically assign a ULA, especially if it isn't U.

hakman: For IPv4 we were using a fixed IP, but don't mind making it random.
What would be the benefit of using a block from the VPC IPv6 allocation?
Should I just remove this and discuss separately?
Is this a good case for making `NonMasquereadeCIDR` user configurable on `create cluster`?

johngmyers: Please remove this and discuss separately.

A benefit of using a block from the VPC IPv6 allocation is that one could then choose to make pods directly accessible from outside the cluster. Also, one wouldn't have to NAT the IPv6 outbound from pods, so the IPv6 destinations would be logging the IPv6 addresses of the pods.

I believe `NonMasqueradeCIDR` could be set using SpecOverride. Possible future work would be to default it to "/64#0" (and make that work).

johngmyers: FYI, making "/64#0" work for `NonMasqueradeCIDR` depends on work in my #9229 track. In that track I plan to refactor the writing of the completed cluster spec into a `ManagedFile` task. Once that is done, `PopulateClusterSpec` would need to be turned into a `Task` and made dependent on `VPCAmazonIPv6CIDRBlock` for managed VPCs.

johngmyers (approved):

PR #11639, opened 2021-05-30
WIP Stop feature-flagging SpecOverride

johngmyers:
We turn it into an unchecked default-on feature flag because the e2e upgrade tests use it.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -295,9 +295,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.Flags().BoolVar(&options.DryRun, "dry-run", options.DryRun, "If true, only print the object that would be sent, without sending it. This flag can be used to create a cluster YAML or JSON manifest.")
 	cmd.Flags().StringVarP(&options.Output, "output", "o", options.Output, "Output format. One of json|yaml. Used with the --dry-run flag.")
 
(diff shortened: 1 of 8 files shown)


olemarkus:
I don't know who need to approve this. But I agree with this change.

/lgtm

hakman:
@johngmyers https://github.com/kubernetes/kops/pull/9840#issuecomment-684869619 were some thoughts when we last discussed about this during office hours.

hakman:
I think the only thing remaining is documenting how to set multiple array elements, like:
```
--set=cluster.spec.kubeDNS.upstreamNameservers=2620:119:53::35
--set=cluster.spec.kubeDNS.upstreamNameservers=2620:119:53::53
```

Setting map elements should also be possible in the future but we should create an issue and document it as a limitation:
```
--set=cluster.spec.cloudLabels=app=test
...
cannot set field "spec.cloudLabels": unhandled type "map[string]string"
```

johngmyers:
I think we want to have a release before removing the feature flag.

johngmyers:
Does `--set=cluster.spec.cloudLabels.app=test` work?

hakman:
> Does `--set=cluster.spec.cloudLabels.app=test` work?

No, it triggers an error.
```
field spec.cloudLabels.app not found in *Cluster
```

johngmyers:
Office Hours result: call for review of current syntax.

PR #11645, opened 2021-05-31
Protokube needs dns-controller IAM permissions

johngmyers:


rifelpet (approved):
For reference, here is the exact failure: https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-grid-scenario-service-account-iam/1399255814341922816/artifacts/ip-172-20-44-209.us-west-1.compute.internal/protokube.log

PR #11647, opened 2021-05-31
Fix service-account-issuer url when specify s3 bucket as serviceAccountIssuerDiscovery

h3poteto:
This PR is a small change.

Service account issuer URL is wrong when I specify  serviceAccountIssuerDiscovery according to https://kops.sigs.k8s.io/cluster_spec/#service-account-issuer-discovery-and-aws-iam-roles-for-service-accounts-irsa .

For example, when I specify following spec 
```yaml
spec:
  serviceAccountIssuerDiscovery:
    discoveryStore: s3://h3poteto-playground-oidc
    enableAWSOIDCProvider: true
```

kube-apiserver's parameters are

```
    - --service-account-issuer=https://h3poteto-playground-oidc.s3.us-east-1.amazonaws.com//oidc
    - --service-account-jwks-uri=https://h3poteto-playground-oidc.s3.us-east-1.amazonaws.com//oidc/openid/v1/jwks
    - --service-account-key-file=/srv/kubernetes/service-account.key
    - --service-account-signing-key-file=/srv/kubernetes/service-account.key
    - --service-cluster-ip-range=100.64.0.0/13
```

There is no conent in `https://h3poteto-playground-oidc.s3.us-east-1.amazonaws.com//oidc`, it should be `https://h3poteto-playground-oidc.s3.us-east-1.amazonaws.com/oidc` . So I removed trailing slash from S3 bucket URL.

### DIFF
--- a/util/pkg/vfs/s3fs.go
+++ b/util/pkg/vfs/s3fs.go
@@ -447,7 +447,8 @@ func (p *S3Path) GetHTTPsUrl() (string, error) {
 		}
 		p.bucketDetails = bucketDetails
 	}
-	return fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", p.bucketDetails.name, p.bucketDetails.region, p.Key()), nil
+	url := fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", p.bucketDetails.name, p.bucketDetails.region, p.Key())
+	return strings.TrimSuffix(url, "/"), nil
 }
 
 // AWSErrorCode returns the aws error code, if it is an awserr.Error, otherwise ""


johngmyers:
I wonder if it would be better to use something like `URL.ResolveReference()` in the caller.

h3poteto:
@johngmyers It is no problem to use `ResolveReference()`. For example: https://play.golang.org/p/shrZENCWcqe

Which is better? I will rewrite this if necessary.

h3poteto:
Oh, I'm sorry, I overlooked.

Probably I can't use `ResolveReference()` in the caller. 

This function is called in https://github.com/kubernetes/kops/blob/ef43708cacb656ca65051530b2552be2f0415aec/pkg/model/iam/subject.go#L125

And if I change to following code
```go
			baseURLStr, err := base.GetHTTPsUrl()     // => https://hogehoge.s3.us-east-1.amazonaws.com/some-obj
			if err != nil {
				return "", err
			}
			baseURL, err := url.Parse(baseURLStr)
			if err != nil {
				return "", err
			}
			oidcPath, err := url.Parse("oidc")
			if err != nil {
				return "", err
			}
			return baseURL.ResolveReference(oidcPath).String(), nil  // => https://hogehoge.s3.us-east-1.amazonaws.com/oidc
```
path will be wrong. `ResolveReference` does not join the path, it will override the path.

johngmyers:
Looks like you might need to use `path.Join()`.

h3poteto:
The object path and URL are fixed in https://github.com/kubernetes/kops/pull/11649. So I close this PR.

PR #11649, opened 2021-05-31
Fix jwks object path in S3 for IRSA

h3poteto:
This is a small change.

jwks object path is wrong when I specify serviceAccountIssuerDiscovery according to https://kops.sigs.k8s.io/cluster_spec/#service-account-issuer-discovery-and-aws-iam-roles-for-service-accounts-irsa .

For example, when I specify following spec:
```yaml
spec:
  serviceAccountIssuerDiscovery:
    discoveryStore: s3://h3poteto-playground-oidc
    enableAWSOIDCProvider: true
```
kube-apiserver's parameters are
```
    - --service-account-issuer=https://h3poteto-playground-oidc.s3.us-east-1.amazonaws.com/oidc
    - --service-account-jwks-uri=https://h3poteto-playground-oidc.s3.us-east-1.amazonaws.com/oidc/openid/v1/jwks
    - --service-account-key-file=/srv/kubernetes/service-account.key
    - --service-account-signing-key-file=/srv/kubernetes/service-account.key
```

But, there is no object in `https://h3poteto-playground-oidc.s3.us-east-1.amazonaws.com/oidc/openid/v1/jwks`. At the momen the jwks object is located in `/openid/v1/jwks`. It should be `/oidc/openid/v1/jwks`, so I fixed the path.


johngmyers:
`oidc/openid` is a bit redundant. Perhaps the fix should be to take it out of `iam.ServiceAccountIssuer()`?

olemarkus:
If I remember my own reasoning correctly, it is to have a common prefix for both the `.well-known/openid-configuration` and `openid/v1/jwks` paths.

johngmyers:
Wouldn't the `DiscoveryStore` be the common prefix? The `oidc` prefix makes less sense now that `DiscoveryStore` is under `ServiceAccountIssuerDiscoveryConfig`.

olemarkus:
I agree. This is leftovers from when that VFS was intended to be shared for other things as well.

olemarkus:
@h3poteto i.e line 129 or so in `pkg/model/iam/subject.go` should just return `baseURL`

h3poteto:
OK, I will try it.

pkg/model/issuerdiscovery.go
```diff
@@ -80,7 +80,7 @@ func (b *IssuerDiscoveryModelBuilder) Build(c *fi.ModelBuilderContext) error {
 	keysFile := &fitasks.ManagedFile{
 		Contents:  keys,
 		Lifecycle: b.Lifecycle,
-		Location:  fi.String("/openid/v1/jwks"),
+		Location:  fi.String("oidc/openid/v1/jwks"),
```
johngmyers, line 83: @h3poteto please also change the `.well-known` path below on line 93. Also, please remove the appending of "oidc" around line 125 in `pkg/model/iam/subject.go`

h3poteto:
@olemarkus @johngmyers How about this?
https://github.com/kubernetes/kops/pull/11649/commits/a7a659d658097f67ee541d751bd041c32706ed47

util/pkg/vfs/s3fs.go
```diff
@@ -447,7 +447,8 @@ func (p *S3Path) GetHTTPsUrl() (string, error) {
 		}
 		p.bucketDetails = bucketDetails
 	}
-	return fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", p.bucketDetails.name, p.bucketDetails.region, p.Key()), nil
+	baseURL := fmt.Sprintf("https://%s.s3.%s.amazonaws.com", p.bucketDetails.name, p.bucketDetails.region)
+	return path.Join(baseURL, p.Key()), nil
```
johngmyers, line 451: `fmt.Println(path.Join("https://example.com", "/path"))` returns `https:/example.com/path` which has too few `/` after the `https:`.

johngmyers (changes requested):

util/pkg/vfs/s3fs.go
johngmyers: Perhaps use a `URL` struct literal.

h3poteto: Oh...sorry. I fixed it.
https://github.com/kubernetes/kops/pull/11649/commits/d52ec60c02473bfa886fbec36ecf60178d3fe3fc

johngmyers (approved):
Thank you.

PR #11654, opened 2021-05-31
Create document on asset repositories

johngmyers:
This covers the basics. There are a few things that could be done to polish the procedure.

My test environment doesn't have a non-S3 fileRepository, so I wasn't able to test that.

I'm wondering if asset phase should default to `LifecycleIgnore`.


docs/operations/asset-repository.md
```diff
@@ -0,0 +1,58 @@
...
+## Copying
+
+You can copy assets into their repositories either by running `kops update cluster` or through an external process.
+
+When doing a `kops update cluster`, kOps will copy assets into their respective repositories if
+they do not already exist there.
+
+For file assets, kOps only supports copying to a repository that is either an S3 or GCS bucket.
+An S3 bucket must be configured using the [regional naming conventions of S3](https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region).
+A GCS bucket must be configured with a prefix of `https://storage.googleapis.com/`.
```
johngmyers, line 53: I'm not quite sure why `fileRepository` doesn't take VFS paths. Perhaps a future PR.

hakman: I think it doesn't accept a VFS path because it is used for NodeUp download too, and that is done via wget.

johngmyers: It could convert from VFS form to https form for use by nodeup/userdata. That's a lot more natural than going the other way.

hakman: It could, but would have to be checked for access and maybe provide some override in case deduced http location is incorrect.

docs/operations/asset-repository.md
```diff
@@ -0,0 +1,60 @@
...
+
+```yaml
+spec:
+  assets:
+    fileRepository: https://example.com/files
+```
+
+## Copying assets into repositories
+
+You can copy assets into their repositories either by running `kops update cluster --phase assets --create-kube-config=false --yes` or through an external process.
```
hakman, line 46: Any idea when this was added? It is not what I would expect.
This is more of a `toolbox` thing than `update cluster` thing.

mkdocs.yml
```diff
@@ -76,6 +76,7 @@ nav:
     - Using Manifests and Customizing: "manifests_and_customizing_via_api.md"
     - High Availability: "operations/high_availability.md"
     - Scaling: "operations/scaling.md"
+    - Local asset repositories: "operatiosn/asset-repository.md"
```
rifelpet, line 79:
```suggestion
    - Local asset repositories: "operations/asset-repository.md"
```

rifelpet (approved):

docs/releases/1.22-NOTES.md
```diff
@@ -58,4 +60,8 @@ This change only affects dedicated API server nodes and worker nodes. It does no
 
 * It is no longer necessary to set `AWS_SDK_LOAD_CONFIG=1` in the environment when using AWS assumed roles with the `kops` CLI.
 
+* There is a new command `kops get assets` for listing image and file assets used by a cluster.
+  It also includes a `--copy` flag to copy the assets to local repositories.
+  See the documentaion on [Using local asset repositories](../operations/asset-repository.md) for more information.
```
rifelpet, line 65:
```suggestion
  See the documentation on [Using local asset repositories](../operations/asset-repository.md) for more information.
```

rifelpet (approved):

PR #11668, opened 2021-06-02
Move common stuff in e2e scenarios to common.sh

olemarkus:


tests/e2e/scenarios/lib/common.sh
```diff
@@ -0,0 +1,65 @@
...
+
+export KOPS_FEATURE_FLAGS="SpecOverrideFlag,${KOPS_FEATURE_FLAGS:-}"
+export GO111MODULE=on
+
+if [[ -z "${AWS_SSH_PRIVATE_KEY_FILE-}" ]]; then
+    export AWS_SSH_PRIVATE_KEY_FILE="~/.ssh/id_rsa"
+fi
+if [[ -z "${AWS_SSH_PUBLIC_KEY_FILE-}" ]]; then
+    export AWS_SSH_PUBLIC_KEY_FILE="~/.ssh/id_rsa.pub"
+fi
```
rifelpet, line 38: not a blocker:

kubetest2-kops recognizing those flags was intended for backwards compatibility with the prow job env vars

https://github.com/kubernetes/test-infra/blob/2fd611583393415eda044de79eb27d7fd816386c/config/prow/config.yaml#L666-L683

I wish they didn't have AWS_ in the name given that they're provider-agnostic. Maybe we could specify the flags instead? or have kubetest2-kops also look for a more generic set of env var names? 

```
      --ssh-private-key string           The path to the private key used for SSH access to instances
      --ssh-public-key string            The path to the public key passed to the cloud provider
```

tests/e2e/scenarios/lib/common.sh
```diff
@@ -0,0 +1,65 @@
...
+
+KUBETEST2="kubetest2 kops -v=2 --cloud-provider=${CLOUD_PROVIDER} --cluster-name=${CLUSTER_NAME:-}"
+KUBETEST2="${KUBETEST2} --admin-access=${ADMIN_ACCESS:-}"
+
+# Always tear-down the cluster when we're done
+function kops-finish {
+  ${KUBETEST2} --kops-binary-path="${KOPS}" --down || echo "kubetest2 down failed"
```
rifelpet, line 45: I had to test this myself but can confirm that `KOPS` not being defined yet is acceptable here.

tests/e2e/scenarios/lib/common.sh
olemarkus: When I test these scripts locally I wanted to emulate the prow environment. I was a bit torn between doing the above or demanding those vars being set (because I always forget to export all that kubetest2 expects).

rifelpet: yea that makes sense to me. we can revisit this later

olemarkus:
I think this is a step in the right direction. Probably one can wrap the kubetest2 commands in a common function as well. They now seem to mostly take the same arguments (overrides, CNI and zones differ). But that is more for a follow up.

rifelpet (approved):

PR #11534, opened 2021-05-19
[cni/cilium] Add support for additional config fields

dntosas:
In this commit, we enable users to define their setup with following
additional fields:

- DisableEndpointCRD
- EnableEndpointHealthChecking
- IdentityAllocationMode
- IdentityChangeGracePeriod
- BPFLBAlgorithm
- BPFLBMaglevTableSize
- BPFNATGlobalMax
- BPFNeighGlobalMax
- BPFPolicyMapMax

Added also validation tests to prevent conflicting value combinations to
reach actual cluster state.

- [x] Add additional config fields
- [x] Add default values
- [x] Include validation/sanity checks on new fields and combinations

Signed-off-by: dntosas <ntosas@gmail.com>

PR #11678, opened 2021-06-03
[cni/cilium] Add support for additional config options

dntosas:
In this commit, we enable users define their setup with following
additional fields:

- DisableEndpointCRD
- EnableEndpointHealthChecking
- IdentityAllocationMode
- IdentityChangeGracePeriod
- BPFLBAlgorithm
- BPFLBMaglevTableSize
- BPFNATGlobalMax
- BPFNeighGlobalMax
- BPFPolicyMapMax

Added also validation tests to prevent conflicting value combinations to
reach actual cluster state.


dntosas:
reflog is our friend :stuck_out_tongue: :bowing_man:

pkg/apis/kops/v1alpha2/networking.go
```diff
@@ -256,6 +256,8 @@ type AmazonVPCNetworkingSpec struct {
 	Env []EnvVar `json:"env,omitempty"`
 }
 
+const CiliumIpamEni = "eni"
```
dntosas, line 259: @olemarkus moving your comment from the broken PR https://github.com/kubernetes/kops/pull/11534#discussion_r642598335

dntosas: you suggest including the above to CiliumSpec if i understand correctly?

olemarkus: I was thinking of something like 


```
type CiliumIpam string

const CiliumIpamENI CiliumIpam = "eni" 
```

dntosas: how about moving this into here to make things simpler? --> https://github.com/kubernetes/kops/blob/master/pkg/model/components/cilium.go

wdyt?

dntosas: tbh, i believe we can completely remove it

dntosas: ok, i made the changes as you suggested ^^ let's solve to a different PR if we need to revisit where this should be placed

dntosas: reverted, as this one will need additional changes, out of scope of PR i guess:
```
# k8s.io/kops/pkg/model/iam
pkg/model/iam/iam_builder.go:272:122: invalid operation: b.Cluster.Spec.Networking.Cilium.Ipam == kops.CiliumIpamEni (mismatched types string and kops.CiliumIpam)
pkg/model/iam/iam_builder.go:327:122: invalid operation: b.Cluster.Spec.Networking.Cilium.Ipam == kops.CiliumIpamEni (mismatched types string and kops.CiliumIpam)
# k8s.io/kops/pkg/model/iam
vet: pkg/model/iam/iam_builder.go:272:84: cannot compare b.Cluster.Spec.Networking.Cilium.Ipam == kops.CiliumIpamEni (mismatched types string and kops.CiliumIpam)
```

olemarkus (approved):
Thanks for this and thanks for your patience :)

PR #11679, opened 2021-06-03
Use release markers instead of releases

olemarkus:
/cc @rifelpet

### DIFF
--- a/tests/e2e/scenarios/upgrade-ab/run-test.sh
+++ b/tests/e2e/scenarios/upgrade-ab/run-test.sh
@@ -22,7 +22,9 @@ if [ -z "$KOPS_VERSION_A" ] || [ -z "$K8S_VERSION_A" ] || [ -z "$KOPS_VERSION_B"
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -22,7 +22,9 @@ if [ -z "$KOPS_VERSION_A" ] || [ -z "$K8S_VERSION_A" ] || [ -z "$KOPS_VERSION_B"
   exit 1
 fi
 
-KOPS_A=$(kops-download-release "${KOPS_VERSION_A}")
+export KOPS_BASE_URL
+KOPS_BASE_URL=$(kops-base-from-marker "${KOPS_VERSION_B}")
```
rifelpet, line 26:
```suggestion
KOPS_BASE_URL=$(kops-base-from-marker "${KOPS_VERSION_A}")
```

olemarkus: Good catch. thanks.

rifelpet (approved):

PR #11682, opened 2021-06-03
Drop trailing slash from oidc issuer

olemarkus:
This is where the trailing slash came from.

/kind bug
/cc @rifelpet

### DIFF
--- a/pkg/model/issuerdiscovery.go
+++ b/pkg/model/issuerdiscovery.go
@@ -102,7 +102,7 @@ func (b *IssuerDiscoveryModelBuilder) Build(c *fi.ModelBuilderContext) error {
 
 func buildDiscoveryJSON(issuerURL string) ([]byte, error) {
 	d := oidcDiscovery{
-		Issuer:                fmt.Sprintf("%v", issuerURL),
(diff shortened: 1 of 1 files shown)


hakman (approved):

pkg/model/issuerdiscovery.go
```diff
@@ -102,7 +102,7 @@ func (b *IssuerDiscoveryModelBuilder) Build(c *fi.ModelBuilderContext) error {
 
 func buildDiscoveryJSON(issuerURL string) ([]byte, error) {
 	d := oidcDiscovery{
-		Issuer:                fmt.Sprintf("%v/", issuerURL),
+		Issuer:                fmt.Sprintf("%v", issuerURL),
```
rifelpet, line 105:
```suggestion
		Issuer:                issuerURL,
```

hakman: Let's leave it for a different PR, the cherry-pick just merged.

hakman: Too late... :)

rifelpet (approved):

PR #11688, opened 2021-06-04
Add options for configuring IPv4 and IPv6 support with Calico

hakman:
Sample config to get started with AWS, Calico and IPv6, using a kOps cluster with public topology:
```yaml
  kubeAPIServer:
    bindAddress: "::"
  nonMasqueradeCIDR: fd00:10:96::/64
  serviceClusterIPRange: fd00:10:96::/108
  kubeControllerManager:
    clusterCIDR: fd00:10:96::/80
    allocateNodeCIDRs: false
  kubeDNS:
    provider: CoreDNS
    upstreamNameservers:
    - 2620:119:35::35
    - 2620:119:53::53
  networking:
    calico:
      ipv4Support: false
      ipv6Support: true
  subnets:
   - type: Public
     ipv6CIDR: 2a05:d014:...::/64
```

Ref: #8432 

/cc @justinsb @johngmyers @aojea

### DIFF
--- a/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
@@ -84,7 +84,7 @@ data:
           ttl 30
         }
         prometheus :9153
-        forward . /etc/resolv.conf {
+        forward . {{ or (join KubeDNS.UpstreamNameservers " ") "/etc/resolv.conf" }} {
           max_concurrent 1000
         }
         loop
--- a/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
@@ -52,6 +52,8 @@ data:
           "nodename": "__KUBERNETES_NODE_NAME__",
           "mtu": __CNI_MTU__,
           "ipam": {
+              "assign_ipv4": "{{ WithDefaultBool .Networking.Calico.IPv4Support true }}",
+              "assign_ipv6": "{{ WithDefaultBool .Networking.Calico.IPv6Support false }}",
               "type": "calico-ipam"
           },
           "policy": {
@@ -3825,7 +3827,9 @@ spec:
               value: "kops,bgp"
             # Auto-detect the BGP IP address.
             - name: IP
-              value: "autodetect"
+              value: "{{- if (WithDefaultBool .Networking.Calico.IPv4Support true) -}}autodetect{{- else -}}none{{- end -}}"
+            - name: IP6
+              value: "{{- if (WithDefaultBool .Networking.Calico.IPv6Support false) -}}autodetect{{- else -}}none{{- end -}}"
             - name: IP_AUTODETECTION_METHOD
               value: "{{- or .Networking.Calico.IPv4AutoDetectionMethod "first-found" }}"
             - name: IP6_AUTODETECTION_METHOD
@@ -3857,8 +3861,17 @@ spec:
             # The default IPv4 pool to create on startup if none exists. Pod IPs will be
             # chosen from this range. Changing this value after installation will have
             # no effect. This should fall within `--cluster-cidr`.
+            {{- if (WithDefaultBool .Networking.Calico.IPv6Support false) }}
+            - name: CALICO_IPV6POOL_CIDR
+              value: "{{ .KubeControllerManager.ClusterCIDR }}"
+            - name: CALICO_IPV6POOL_NAT_OUTGOING
+              value: "true"
+            - name: CALICO_ROUTER_ID
+              value: "hash"
+            {{- else if (WithDefaultBool .Networking.Calico.IPv4Support true) }}
             - name: CALICO_IPV4POOL_CIDR
               value: "{{ .KubeControllerManager.ClusterCIDR }}"
+            {{- end }}
             # Disable file logging so `kubectl logs` works.
             - name: CALICO_DISABLE_FILE_LOGGING
               value: "true"
@@ -3867,7 +3880,7 @@ spec:
               value: "ACCEPT"
             # Disable IPv6 on Kubernetes.
             - name: FELIX_IPV6SUPPORT
-              value: "false"
+              value: "{{ WithDefaultBool .Networking.Calico.IPv6Support false }}"
             - name: FELIX_HEALTHENABLED
               value: "true"
 
@@ -3924,6 +3937,9 @@ spec:
               {{- if eq .Networking.Calico.EncapsulationMode "ipip" }}
               - -bird-live
               {{- end }}
+              {{- if (WithDefaultBool .Networking.Calico.IPv6Support false) }}
+              - -bird6-live
+              {{- end }}
             periodSeconds: 10
             initialDelaySeconds: 10
             failureThreshold: 6
@@ -3934,6 +3950,9 @@ spec:
               - -felix-ready
               {{- if eq .Networking.Calico.EncapsulationMode "ipip" }}
               - -bird-ready
+              {{- end }}
+              {{- if (WithDefaultBool .Networking.Calico.IPv6Support false) }}
+              - -bird6-ready
               {{- end }}
             periodSeconds: 10
           {{- if .Networking.Calico.PrometheusMetricsEnabled }}
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -3233,6 +3233,10 @@ spec:
                           (default), "can-reach=DESTINATION", "interface=INTERFACE-REGEX",
                           or "skip-interface=INTERFACE-REGEX"'
                         type: string
+                      ipv4Support:
+                        description: 'IPv4Support enables Calico networking and security
+                          for IPv4 traffic. Default: "true"'
+                        type: boolean
                       ipv6AutoDetectionMethod:
                         description: 'IPv6AutoDetectionMethod configures how Calico
                           chooses the IP address used to route between nodes.  This
@@ -3241,6 +3245,10 @@ spec:
                           (default), "can-reach=DESTINATION", "interface=INTERFACE-REGEX",
                           or "skip-interface=INTERFACE-REGEX"'
                         type: string
+                      ipv6Support:
+                        description: 'IPv6Support enables Calico networking and security
+                          for IPv6 traffic. Default: "false"'
+                        type: boolean
                       logSeverityScreen:
                         description: 'LogSeverityScreen lets us set the desired log
                           level. (Default: info)'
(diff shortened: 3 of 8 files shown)


aojea:
is there ipv6 CI job?

hakman:
> is there ipv6 CI job?

Not yet. There are a few things left on another PR and after that will try to get a CI job to run.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -3857,8 +3861,17 @@ spec:
             # The default IPv4 pool to create on startup if none exists. Pod IPs will be
             # chosen from this range. Changing this value after installation will have
             # no effect. This should fall within `--cluster-cidr`.
+            {{- if (WithDefaultBool .Networking.Calico.IPv6Support false) }}
+            - name: CALICO_IPV6POOL_CIDR
+              value: "{{ .KubeControllerManager.ClusterCIDR }}"
+            - name: CALICO_IPV6POOL_NAT_OUTGOING
```
johngmyers, line 3867: Why do we need to NAT outgoing IPv6? Could we not assign a routable CIDR to the pool? Or does that require mucking about with ENI mode?

upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -84,7 +84,7 @@ data:
           ttl 30
         }
         prometheus :9153
-        forward . /etc/resolv.conf {
+        forward . {{ or (join KubeDNS.UpstreamNameservers " ") "/etc/resolv.conf" }} {
```
johngmyers, line 87: Is this an unrelated change?

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -3857,8 +3861,17 @@ spec:
             # The default IPv4 pool to create on startup if none exists. Pod IPs will be
             # chosen from this range. Changing this value after installation will have
             # no effect. This should fall within `--cluster-cidr`.
+            {{- if (WithDefaultBool .Networking.Calico.IPv6Support false) }}
+            - name: CALICO_IPV6POOL_CIDR
+              value: "{{ .KubeControllerManager.ClusterCIDR }}"
```
johngmyers, line 3866: According to https://kubernetes.io/docs/concepts/services-networking/dual-stack/ the KCM `--cluster-cidr` is, in dual-stack, an IPv4 CIDR and an IPv6 CIDR, separated by a comma. It doesn't look like calico/node accepts that syntax in either CALICO_IPV6POOL_CIDR or CALICO_IPV4POOL_CIDR.

I think we'd need a template function to parse `ClusterCIDR` into its two components.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -3857,8 +3861,17 @@ spec:
...
+            - name: CALICO_IPV6POOL_NAT_OUTGOING
+              value: "true"
+            - name: CALICO_ROUTER_ID
+              value: "hash"
```
johngmyers, line 3870: Per Calico documentation, this should only be set for IPv6-only configurations. So it should be conditional on `.Networking.Calico.IPv6Support` being `false`.

johngmyers (changes requested):
Not blocking, but it would be good if we could automatically set the IPv*Support fields based on a cluster-level setting as to whether the cluster is v4-only, dual-stack, or v6-only.

upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
johngmyers: Okay, missing feature from Coredns that you need in order to not muck with the node-level DNS? This should perhaps be a separate PR.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
johngmyers: Actually, I'm okay with not supporting dual-stack.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
johngmyers: Okay, we're not supporting dual-stack.

johngmyers:
I think the only thing blocking is API validation to prohibit the combination of Calico and dual-stack ClusterCIDR.

johngmyers:
Okay, existing ClusterCIDR validation prohibits multiple CIDRs.

johngmyers (approved):
All of these comments can be followup work.

johngmyers:
Suggest API validation to prohibit enabling both IPv4 and IPv6 for Calico.

johngmyers:
Hrm, perhaps a single, tri-state setting? With values 'ipv4', 'ipv6', 'dual', default 'ipv4'.

johngmyers:
Or key off of how `ClusterCIDR` parses?

upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
hakman: I thought about making this a separate PR, but it was a one liner, related to current work.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
hakman: For now, single stack ClusterCIDR is ok from my point of view too.
Dual-stack is also blocked by https://github.com/kubernetes/cloud-provider-aws/pull/230, as the kubelet expects the cloud provider to return both IPv4 and IPv6 for the node.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
hakman: Exactly. Would have to experiment more with this setting. In theory it should not be needed because would use the host IPv4. In practice, it was needed to get things working. Will be much easier once the `--ipv6` will be available and can add periodic and optional tests.

upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
hakman: There are a few considerations here:
* default behavior for Calico is to do outgoing NAT
* this is a setting that will be applied for the default IPv6 Pool created on first start.
* I don't see how assigning a routable CIDR would be possible until AWS adds a routable IPv6 CIDR blocks to each node. 
With these in mind, seems a safe default for now. It can easily be changed after initial start and, once there is some way to use routable CIRDS, it can be split into a separate option.

hakman:
> Or key off of how ClusterCIDR parses?

I think this may be a good way to pull the defaults, but would sill allow overriding for experiments.

hakman:
In other news, I ran the Conformance tests manually on a fresh cluster and got the following failures:
```
[Fail] [sig-network] Services [It] should have session affinity work for NodePort service [LinuxOnly] [Conformance]
/workspace/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/service.go:203

[Fail] [sig-network] Services [It] should be able to switch session affinity for NodePort service [LinuxOnly] [Conformance]
/workspace/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/service.go:203

[Fail] [sig-network] EndpointSlice [It] should create Endpoints and EndpointSlices for Pods matching a Service [Conformance]
/workspace/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/endpointslice.go:320

[Fail] [sig-network] HostPort [It] validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]
/workspace/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/github.com/onsi/ginkgo/internal/leafnodes/runner.go:113

[Fail] [sig-network] Services [It] should have session affinity timeout work for NodePort service [LinuxOnly] [Conformance]
/workspace/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/service.go:203
```

These are happening because the tests try to connect to the IPv4 address of the service exposed via NodePort. If the tests would connect to the IPv6 address instead, they would pass (tested by setting `--node-ip=<IPv6-Address>` and using a custom cloud provider binary). For some reason, using `--node-ip=::` did not help.

The cloud provider also only provides the IPv4 address for the node, so this is an expected failure.

johngmyers:
In retrospect, there shouldn't be `IPv4Support` and `IPv6Support` fields in the `CalicoNetworkingSpec`. The templates should key off of something that derives from whether the cluster has an IPv4 or IPv6 (or both) `ClusterCIDR`.

hakman:
I agree, but this may allow some use cases like some fake pod dual-stack that works at node level.
Also, those are the default IP pools and what kOps does is just give a simple way to set the defaults correctly. Advanced users will do a lot more magic there.
So, the fields should be ok to exist, as overrides for advanced users, but the settings should default to the IP family of what's set in `ClusterCIDR`.

johngmyers:
The Calico templates don't support dual-stack. For example, `CALICO_ROUTER_ID` should not be set to `"hash"` if it's dual-stack.

Advanced configurations will undoubtedly require code to configure all the things needed for the particular use case.

The fields don't make sense to expose. They just clutter up the API.

johngmyers:
I think we might need code in `pkg/model/kubelet.go` to set `--node-ip` correctly in the IPv6 case. There's that `b.UsesSecondaryIP()` case there that currently assumes IPv4.

hakman:
> The Calico templates don't support dual-stack. For example, CALICO_ROUTER_ID should not be set to "hash" if it's dual-stack.

This is more of a quote from the docs which are lacking a lot of details when it comes to IPv6. "hash" means a  CRC32 of the hostname, which contains the IP. It has nothing to do with how dual-stack is configured in Calico.

Calico allows lots of config using calicoctl, outside of the daemonset template, including creating new IP pools. kOps only configures the default IP pool, which is based on ClusterCIDR.

> I think we might need code in pkg/model/kubelet.go to set --node-ip correctly in the IPv6 case. There's that b.UsesSecondaryIP() case there that currently assumes IPv4.

I am aware of that option and will take it into consideration, weather to support it or consider it mutually exclusive with IPv6.

PR #11700, opened 2021-06-06
Move asset copying out of apply_cluster

johngmyers:
Removes the `assets` phase, moving the functionality to the new `kops get assets --copy` flag.

### DIFF
--- a/cmd/kops/BUILD.bazel
+++ b/cmd/kops/BUILD.bazel
@@ -90,6 +90,7 @@ go_library(
         "//pkg/util/templater:go_default_library",
         "//pkg/validation:go_default_library",
         "//upup/pkg/fi:go_default_library",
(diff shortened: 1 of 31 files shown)


rifelpet:
I'm wondering if we should announce the deprecation of the assets phase in the 1.21 release notes.

This will also need mentioning in the 1.22 release notes, and ideally have documentation that walks through the process of how to use `get assets --copy` but that can be in a followup PR.

johngmyers:
There's nothing a user could do with such a deprecation notice. The feature isn't going away; the command to invoke it is just changing.

My documentation PR #11654 is pending review, so I can't update it in this PR.

rifelpet (approved):
True, but I think the 1.22 release notes should still mention the move of functionality. I'm fine with that being in a followup PR

PR #11702, opened 2021-06-06
Make relnotes match the new max hop limit IMDS behaviour

olemarkus:


johngmyers:
Should also update the second paragraph of `docks/iam_roles.md`

johngmyers (approved):
/hold
in case further revision desired

olemarkus:
Updated the docs in a few other places as well.

/hold cancel

PR #11708, opened 2021-06-07
Limit concurrency of asset copy tasks 

johngmyers:
Why: parallelism after a certain point doesn't make things any faster: there's only so much bandwidth. Less parallelism increases the amount of feedback on progress and allows retries of interrupted copies to better avoid recopying by having more completed copies.

How: It uses a buffered channel as a semaphore.


hakman:
This would deserve a description with why and how it's done. Other than that...
/lgtm

johngmyers:
Why: parallelism after a certain point doesn't make things any faster: there's only so much bandwidth. Less parallelism increases the amount of feedback on progress and allows retries of interrupted copies to better avoid recopying by having more completed copies.

How: It uses a buffered channel as a semaphore.

johngmyers:
@hakman Is there a particular place you wanted that description? How can I move this forward?

hakman (approved):
Sorry, missed this.
Description was not for me but for anyone looking at this 6 months from now.

PR #11709, opened 2021-06-07
Add proxy envs to calico to make possible usage of AWS source destination check

DOboznyi:
After 1.19 update calico was also updated and source-destination check was included in calico natively. So when you're using proxy and cluster created it can't be validated without manual patch:
```
kubectl set env daemonset/calico-node -n kube-system http_proxy=http://<proxy_url>:<proxy_port>
kubectl set env daemonset/calico-node -n kube-system https_proxy=http://<proxy_url>:<proxy_port>
kubectl set env daemonset/calico-node -n kube-system no_proxy=<no_proxy_list>
```

Signed-off-by: Dmytro Oboznyi <dmytro.oboznyi@syncier.com>

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
@@ -3924,6 +3924,11 @@ spec:
             # Enable WireGuard encryption for all on-the-wire pod-to-pod traffic
             - name: FELIX_WIREGUARDENABLED
               value: "{{ .Networking.Calico.WireguardEnabled }}"
+            # Enable support for HTTP forward proxy
+            {{ range $name, $value := ProxyEnv }}
+            - name: {{ $name }}
+              value: {{ $value }}
+            {{ end }}
           securityContext:
             privileged: true
           resources:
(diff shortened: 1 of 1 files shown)


upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -3924,6 +3924,12 @@ spec:
             # Enable WireGuard encryption for all on-the-wire pod-to-pod traffic
             - name: FELIX_WIREGUARDENABLED
               value: "{{ .Networking.Calico.WireguardEnabled }}"
+{{- if .EgressProxy }}
+{{ range $name, $value := ProxyEnv }}
+            - name: {{ $name }}
+              value: {{ $value }}
+{{ end }}
+{{- end }}
```
hakman, line 3932: There is no need to check for `EgressProxy`, `ProxyEnv` already does that.
```suggestion
            # Enable support for HTTP forward proxy
            {{ range $name, $value := ProxyEnv }}
            - name: {{ $name }}
              value: {{ $value }}
            {{ end }}
```

hakman (changes requested):

hakman:
Please also squash the commits. It should be good to merge otherwise.

hakman (approved):

PR #11712, opened 2021-06-07
Use helm's kubeclient to apply manifests

olemarkus:
Helm's client diffs two versions of a manifest and cleans up resources no longer used.
This particular PR will allow e.g moving from SQS to IMDS mode for NodeTerminationHandler.
Further down the line, it could be used to delete addons as well.


channels/pkg/channels/addon.go
```diff
@@ -161,17 +168,90 @@ func (a *Addon) EnsureUpdated(ctx context.Context, k8sClient kubernetes.Interfac
...
+				}
+			})
+
+			if err != nil {
+				return nil, fmt.Errorf("unable to get information about existing resources: %w ", err)
+			}
+		}
+
+		for _, r := range toBeUpdated {
+			current.Append(r)
```
justinsb, line 245: I'm not sure I follow what's going on here ... is current the version on the cluster or in the previously applied yaml?  Is helm merging the two by gvknn?

justinsb:
So I really like the idea here.  But I'm very wary of taking a dependency on something that isn't part of kubernetes (helm), particularly when we have a bunch of solutions in kube itself (e.g. server-side-apply, the Application CRD etc).

I think a good way forwards is probably to discuss the alternatives available to us and decide whether which library we want to debug :-)  A good topic for office hours?

channels/pkg/channels/addon.go
olemarkus: The previously applied yaml. This comes from the challenge I mentioned with being able to extract existing resources from k8s directly. So both I and Helm rely on the state in k8s being the same as the state from the previously applied yaml.

olemarkus:
> So I really like the idea here. But I'm very wary of taking a dependency on something that isn't part of kubernetes (helm), particularly when we have a bunch of solutions in kube itself (e.g. server-side-apply, the Application CRD etc).
> 
> I think a good way forwards is probably to discuss the alternatives available to us and decide whether which library we want to 

Can server-side-apply delete resources? I have seen SSA diff-and-patching done by apiserver instead of the client.
Application CRD is an interesting project, but I thought this was abandoned. I may be wrong. Kops benefits from not having very complex requirements, so if we can resurrect Application CRD and get some influence over the project, that is interesting indeed.

But the benefit of Helm is that it is heavily used and well-maintained.

justinsb:
So I've been thinking a lot about this one... (it's pretty related to my dayjob!)

Server-side-apply doesn't solve the ownership/pruning problem for objects; it addresses field ownership.  I'm going to do an experiment to see if it solves the services-with-tcp-and-udp update problem!

For ownership/pruning:

* If/when we move to operator-based addons, we'll likely use ownerrefs; they are self-documenting and almost clean up by themselves.

* Owner-refs aren't perfect; there's no built-in call (currently) to get all "child" objects - it requires listing all kinds (slower, need full RBAC permissions).  But cleanup does _not_ require special RBAC permissions, and you can tell for any given object who the owner is.  Much of the behaviour is server-side, so it doesn't need special tooling.

* The labels approach is built-in to kubectl but can be a little difficult (requires listing all kinds).  It requires client-side tooling to make it work, but this approach is pretty generic.

* The manifest-of-things-installed approach is reasonable, and overcomes some of the limitations of ownerrefs and labels.  But it means that users are more locked in to using the tool, and it's more complex that just labels.

Labels or manifests require us to have been using them when we installed, so we would likely need an adoption period IIUC.

I saw that in bca857326f3225e93f62d3d3d702043255d51519 you added standard labels, and that is in 1.21.  (Thanks!).  I'm wondering if pruning using labels is therefore simpler and more universal.

I think we could pass the pruning label selector in the addon manifest yaml, so we could enable it selectively.  If we wanted to adopt additional objects explicitly, we could do that as well...

olemarkus:
> Server-side-apply doesn't solve the ownership/pruning problem for objects; it addresses field ownership. ~I'm going to do an experiment to see if it solves the services-with-tcp-and-udp update problem!~ I checked, and server-side-apply does solve the services-with-tcp-and-udp update problem; it can update the targetPort.
> 

Under the hood, this PR does use server-side-apply and `kops` is set as owner of the all fields.

> For ownership/pruning:
> 
>     * If/when we move to operator-based addons, we'll likely use ownerrefs; they are self-documenting and almost clean up by themselves.
> 

Even with operator-based addons we would run into the same problem as we do deploy operators and if we wanted to switch from a hypothetical kubedns-operator to coredns-operator we'd have to remove entire "addons" with operators themselves being those addons.


>     * The manifest-of-things-installed approach is reasonable, and overcomes some of the limitations of ownerrefs and labels.  But it means that users are more locked in to using the tool, and it's more complex that just labels.

I don't really see the lock-in. If we used Helm's mechanisms for storing those manifests, we would certainly be locked in, but this is handled by channels itself. What helm provides is just the diffing mechanism.

If we used the labels approach, we'd end up with a set of things-installed and then I'd feed _that_ into the helm kube client to do diff-and-prune. It is actually not that different from going through every object in the to-be-applied manifest to see if they already exists, and if they do, add them to the list-of-things-installed.

I.e labels is just another way of getting list-of-things-installed. It does not solve diff-and-prune. That is, unless I am missing something.

PR #11716, opened 2021-06-08
Generate AWSEBSCSIDriver model only when using AWS

hakman:
/cc @olemarkus @johngmyers


johngmyers (approved):

johngmyers:
There's more refinement to do here. 

`validateSnapshotController()` requires AWSEBSCSIDriver be enabled even when not using the AWS cloud provider.

Either `AWSEBSCSIDriverBuilder.Build()` should do nothing for non-AWS cloud providers or API validation should validate the cloud provider is AWS when enabling AWSEBSCSIDriver. Same for KubeControllerManagerOptionsBuilder and KubeletOptionsBuilder

PR #11717, opened 2021-06-08
Pre-pull all container images used by components and addons

hakman:
This should be a more generic way of pre-pulling images when using Warm Pools.

/cc @olemarkus @johngmyers


johngmyers:
I totally agree with the first commit.

I don't think it's reasonable to require the admin to manually list the images in the instancegroup spec.

hakman:
It is not required to list the images in the instancegroup spec. It is just an option to allow additional images that may be used later for the workload. All kOps referenced images are added automatically to that list. Please see the integration test output changes.

johngmyers:
Could you put those in the `nodeup.Config` instead of `conf/ig_spec.yaml`? You're working at cross-purposes to my #9294.

johngmyers:
...and please don't add more references to the instancegroup spec from nodeup.

olemarkus:
I think it would be nice to have the list of images added to auxconfig that comes with #9294.

Images that are deployed as part of Deployments should not be pre-pulled, especially if they can only run on the control plane (etcd-manager for example, but also includes kops-controller etc).

Docs should mention that credentials managed by Kubernetes (such as ecr and dockerconfig) will not be used for pre-warming, so the typical way of working around dockerhub rate limits does not work here.

hakman:
First of all, thanks for the feedback. I think should be better now.

> I think it would be nice to have the list of images added to auxconfig that comes with #9294.

Added to `nodeup.Config` as requested.

> Images that are deployed as part of Deployments should not be pre-pulled, especially if they can only run on the control plane (etcd-manager for example, but also includes kops-controller etc).

We could add an exception list in `getWarmPoolImages ` and skip some of the known images.
In general and in particular for Deployments, I think that it is note very useful. While in pre-initialization, the nodes are not meant to have a quick start while pulling images, but prepare as much as possible for a fast start when needed.

> Docs should mention that credentials managed by Kubernetes (such as ecr and dockerconfig) will not be used for pre-warming, so the typical way of working around dockerhub rate limits does not work here.

Agreed. Though that is a general consideration for the WarmPool feature.

nodeup/pkg/model/tests/golden/warm-pool/cluster.yaml
```diff
@@ -0,0 +1,83 @@
...
+    - coredns/coredns:1.8.3
+    - k8s.gcr.io/cpa/cluster-proportional-autoscaler:1.8.3
+    - k8s.gcr.io/etcdadm/etcd-manager:3.0.20210430
+    - k8s.gcr.io/kops/dns-controller:1.22.0-alpha.1
+    - k8s.gcr.io/kops/kops-controller:1.22.0-alpha.1
+    - k8s.gcr.io/kops/kube-apiserver-healthcheck:1.22.0-alpha.1
+    - k8s.gcr.io/kube-apiserver:v1.21.0
+    - k8s.gcr.io/kube-controller-manager:v1.21.0
+    - k8s.gcr.io/kube-proxy:v1.21.0
+    - k8s.gcr.io/kube-scheduler:v1.21.0
```
rifelpet, line 83: You should be able to remove all these and the list of tasks in the next file would be preserved, correct? That would confirm that users don't need to specify them in their IG specs and nodeup will still prepull the relevant images from assets.

hakman: Not really. The test harness is not doing a full mock from the spec, like the integration test does.
Somewhere there's a missing piece, but did not have time to dig that deep.

rifelpet: ah ok. so if we had the integration tests also generate nodeup task lists, we could see the functionality there.

hakman: would be cool to generate the nodeup tasks in same place, or in a similar way here.

olemarkus:
I am not concerned about the time spent in pre-init, but I am concerned about how much data is being transferred. Some of these images are fairly large, and if they are pulled 100s of times per day, but will not ever be used, it seems wasteful.
I think it is easier to maintain a list of images worth pulling than a list of images not worth pulling.

hakman:
At the moment, the only huge images not worth pulling over and over are the etcd-manager ones. The k8s images are pretty small. The rest are nice to have on node startup, in case any of them are needed.
From point of view of exclude list vs include list, include list will be more work to maintain. Exclude list could be exposed as an option and let operators specify patterns to skip.

johngmyers:
We could add additional arguments to `AssetBuilder.RemapImage()` to list the node roles that should warm-pull the image. We could have `RemapManifest` parse a magic comment with the same info.

hakman:
> We could add additional arguments to `AssetBuilder.RemapImage()` to list the node roles that should warm-pull the image. We could have `RemapManifest` parse a magic comment with the same info.

Not sure how much this would help, compared to a simple exclude list. Of course, if this becomes a very popular kOps feature, we can iterate and improve.

johngmyers:
It would help by excluding irrelevant images without action by the administrator. Including that information in the output of `kops get assets` could enable baking the images into AMIs.

hakman:
I don't think many people would care about creating N AMIs to match IG container images. Usually one will create an image with everything included.
We have to keep a balance between how much stuff we add in our code vs brute force and cost of "networking". Warm Pool is by design wasteful. It creates N instances that may never be used. Bootstrapping consumes network resources, EBS will use resources and so on.
Particularly about this case, we have 3 cases of waste, so we have to see which ones we are debating:
1. kOps and k8s images that only run on masters (easy to exclude because they are well known and rarely change)
2. pods that only run on masters (would require parsing the addons manifests)
3. pods that run on 1 or few nodes (would again require parsing addons manifests)

olemarkus:
I also have implementing an "AMI builder" configuration mode that runs nodeup from e.g packer. That may be a way for administrators to pre-build images per instance group quite easily and without any particular configuration.

hakman:
I think AMI baking is completely different topic. Having that would make Warm Pool not needed. At least I would be curious to see how startup times compare between custom AMI and Warm Pool.
My desire is to remove any trace of image names from nodeup and making it generic. We discussed this at length before.
What would it take to do that?

johngmyers:
There is one node role for all worker nodes, not one per IG. And we could probably exclude the images for node role "Master".

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1404,10 +1404,30 @@ func (n *nodeUpConfigBuilder) getWarmPoolImages(ig *kops.InstanceGroup) []string
 		}
 	}
 
+	// Add component and addon images that impact startup time
+	// TODO: Exclude images that only run on control-plane nodes in a generic way
+	desiredImagePrefixes := []string{
```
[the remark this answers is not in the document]
hakman: That should probably be a separate PR. As it is now it's not worse than current implementation.

tests/integration/update_cluster/minimal-warmpool/data/aws_launch_template_nodes.minimal-warmpool.example.com_user_data
```diff
@@ -212,18 +219,18 @@ KubeletConfig:
...
-- k8s.gcr.io/cpa/cluster-proportional-autoscaler:1.8.3
-- k8s.gcr.io/etcdadm/etcd-manager:3.0.20210430
-- k8s.gcr.io/kops/dns-controller:1.22.0-alpha.1
-- k8s.gcr.io/kops/kops-controller:1.22.0-alpha.1
-- k8s.gcr.io/kops/kube-apiserver-healthcheck:1.22.0-alpha.1
-- k8s.gcr.io/kube-apiserver:v1.21.0
-- k8s.gcr.io/kube-controller-manager:v1.21.0
 - k8s.gcr.io/kube-proxy:v1.21.0
-- k8s.gcr.io/kube-scheduler:v1.21.0
+- k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v1.0.0
```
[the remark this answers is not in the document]
hakman: Is there anything that should be done here besides this? Maybe we can merge it and just modify desiredImagePrefixes.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1384,5 +1385,67 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+		for _, image := range cluster.Spec.WarmPool.Images {
+			images[image] = true
+		}
+	}
+
+	// Add component and addon images that impact startup time
+	// TODO: Exclude images that only run on control-plane nodes in a generic way
+	desiredImagePrefixes := []string{
+		"602401143452.dkr.ecr.us-west-2.amazonaws.com/",
+		"docker.io/calico/",
```
olemarkus, line 1420: We need to have a think about docker.io images. Unless people configure cache/rewrite assets or similar, these will hit ratelimts quickly.

It could be an idea to properly implement registry auth for containerd before merging this one so people also have the option of using the creds they may have deployed as docker secret.

hakman: I would like to merge this. kOps 1.21 is in alpha and there is enough time to get feedback on this.
As this desiredImagePrefixes will be a new config fields soon. I would expect will address some of the concerns here.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1384,5 +1385,67 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+	if ig.Spec.WarmPool != nil || cluster.Spec.WarmPool != nil {
+		config.WarmPoolImages = n.getWarmPoolImages(ig)
+	}
+
 	return config, auxConfig, nil
 }
+
+// getWarmPoolImages returns a list of container images that should be pre-pulled during instance pre-initialization
+func (n *nodeUpConfigBuilder) getWarmPoolImages(ig *kops.InstanceGroup) []string {
+	if ig != nil && ig.Spec.Role != kops.InstanceGroupRoleNode && ig.Spec.Role != kops.InstanceGroupRoleAPIServer {
```
johngmyers, line 1397: Pre-pulling images could make sense for APIServer nodes.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1384,5 +1385,67 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+	if cluster != nil && cluster.Spec.WarmPool != nil {
+		for _, image := range cluster.Spec.WarmPool.Images {
+			images[image] = true
+		}
+	}
+
+	// Add component and addon images that impact startup time
+	// TODO: Exclude images that only run on control-plane nodes in a generic way
+	desiredImagePrefixes := []string{
+		"602401143452.dkr.ecr.us-west-2.amazonaws.com/",
```
johngmyers, line 1419: Perhaps comment what this one is for

pkg/apis/kops/cluster.go
```diff
@@ -891,6 +891,8 @@ type WarmPoolSpec struct {
 	// EnableLifecyleHook determines if an ASG lifecycle hook will be added ensuring that nodeup runs to completion.
 	// Note that the metadata API must be protected from arbitrary Pods when this is enabled.
 	EnableLifecycleHook bool `json:"enableLifecycleHook,omitempty"`
+	// Images is the list of container images that should be puled during Warm Pools pre-initialization
```
johngmyers, line 894: nit:

```suggestion
	// Images are the images that should be puled during Warm Pool pre-initialization.
```

upup/pkg/fi/cloudup/apply_cluster.go
hakman: Done, I think it's more clear now.

hakman:
@olemarkus @johngmyers Anything else you want me to do here before merging this?

upup/pkg/fi/cloudup/apply_cluster.go
olemarkus: I think we should remove docker.io and prevent docker.io from being used here for now until a solution has been fixed. If not, this will break my test setup very quickly as all nodes are behind the same NAT IPs.

They can be added back once we have a good solution for this.

olemarkus (changes requested):
Except the one comment here, I think this is good enough for merge.

upup/pkg/fi/cloudup/apply_cluster.go
hakman: Latest commit should address this.

hakman:
@olemarkus Ping?

olemarkus:
I am fine with this one now. It is still a bit rough approach, but ...

/lgtm

hakman:
Implemented feedback from Office Hours.

johngmyers (approved):

hakman:
Rebased after latest changes.

johngmyers (approved):

PR #11721, opened 2021-06-08
Make AWS EBS CSI Driver default as of k8s 1.22

olemarkus:


rifelpet (approved):

michaelajr:
Hey @olemarkus - want to point out that the Cluster Spec comments still says this disabled by default. https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#AWSEBSCSIDriver

olemarkus:
Thanks. Any chance you would be able to do a PR to fix that? Would be much appreciated.

PR #11722, opened 2021-06-08
Use quay images for cilium

olemarkus:
a) again I got a request for this over docker rate limiting
b) Since with containerd, docker credentials are not valid for image pulls outside of kubelet, docker.io images will hit rate limiting
c) upstream use quay by default

cilium with hubble already has a dependency on cert-manager, which is hosted on quay so quay is to some extent a dependency anyway.

### DIFF
--- a/nodeup/pkg/model/networking/cilium.go
+++ b/nodeup/pkg/model/networking/cilium.go
@@ -53,7 +53,7 @@ func (b *CiliumBuilder) Build(c *fi.ModelBuilderContext) error {
 		return err
 	}
 
-	image := "docker.io/cilium/cilium:" + cilium.Version
+	image := "quay.io/cilium/cilium:" + cilium.Version
 
 	b.WarmPullImage(c, image)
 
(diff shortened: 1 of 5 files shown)


hakman:
@olemarkus Any thoughts about doing this the same way we do for Calico?
https://github.com/kubernetes/kops/commit/372fde094974bcb25fe17eb231e1e2a0fbca618c

olemarkus:
Was considered and decided against: https://github.com/kubernetes/kops/pull/10204

hakman:
> Was considered and decided against: #10204

I guess we can revisit that. The decision was in the context of us doing an API change soon.
Generic string replace in image urls could be a solution, but I don't see anyone interested in doing that.
Docker to Quay will generate the same amount of requests the other way around when Quay goes down.

johngmyers:
Upstream default is a powerful argument.

For kOps 1.22, local image registries will be a lot easier to manage.

hakman:
> Upstream default is a powerful argument.

I agree.

> For kOps 1.22, local image registries will be a lot easier to manage.

Still harder than setting the image version or registry.

/approve

PR #11725, opened 2021-06-09
Simplify Calico IPv6 configuration

johngmyers:


hakman:
If you want to simplify, please automatically set IPv4 and IPv6 support based on NonMasqueradeCIDR.
The 2 flags should stay as they are for now, to allow override.

johngmyers:
There is no need for override. Dual-stack is out of scope. The CNI needs to operate in the same protocol version as the pod and service networks.

hakman:
Dual-stack is out of scope for k8s. CNI still has some tricks that can be done via internal dual-stack.
For example these overrides make it possible to get away without using NAT64.

johngmyers:
Those "tricks" could be coded up as separate features with their own settings.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -223,6 +223,7 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		return strings.Join(labels, ",")
 	}
 
+	dest["IsIPv6Only"] = tf.IsIPv6Only
```
hakman, line 226: Would you be ok with something like this? It would make the template more clear.
At this stage, it is irrelevant if cluster is IPv4 or IPv6, but only if they should be enabled or not.
Later, if we decide to add DualStack, we can keep the template and just modify the logic in the function(s).
```suggestion
	dest["enableIPv4"] = !tf.IsIPv6Only
	dest["enableIPv6"] = tf.IsIPv6Only
```

hakman (changes requested):

hakman (approved):
We can iterate on this later. For now, it makes it possible to run IPv6 setups easier.
/hold cancel

PR #11729, opened 2021-06-09
Allow master to touch volumes tagged with kubernetes.io/cluster/<clusterName>:owned

wongma7:
This fixes a specific situation where CSIMigrationAWS gets enabled then disabled, step by step details are below.

Basically, kops KCM in-tree EBS plugin and CSI EBS plugin need to agree on what tags to add to provisioned volumes and what IAM policies to ship with such that those provisioned volumes can be managed by either KCM or CSI. Otherwise when migration gets toggled, KCM/CSI won't be allowed to Attach/Delete volumes that were provisioned by CSI/KCM before the toggle. For more details, see: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/927#issuecomment-858034205

1. Enable CSIMigrationAWS
2. Deploy CSI EBS plugin and provide the clusterName via the ` k8s-tag-cluster-id ` flag. This tells it to create volumes with the kubernetes.io/cluster/<clusterName>:owned tag, to partially mimic KCM volume creation behavior. Partially because KCM creates volumes with the legacy KubernetesCluster:<clusterName> tag too.
3. With the default StorageClass with provisioner kubernetes.io/aws-ebs installed, create a PVC.
4. Since migration is enabled, CSI EBS plugin acts as provisioner kubernetes.io/aws-ebs. It creates a volume with tag  kubernetes.io/cluster/<clusterName>:owned and corresponding PV. This volume can be attached, detached, and deleted by CSI.
5. Disable CSIMigrationAWS (let's say I decide I don't like CSI...I expect my existing volumes with spec awsElasticBlockStore to go back to being managed by the in-tree plugin instead of CSI)
6. The volume created in step 3 cannot be attached or deleted by KCM because it is only allowed to attach or delete volumes with the legacy KubernetesCluster:<clusterName> tag. (This goes against my expectation.)

So to fix step 6, this PR grants master instance on which KCM is running permission to touch volumes tagged with kubernetes.io/cluster/<clusterName>:owned in addition to the legacy KubernetesCluster:<clusterName> tag.

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -905,6 +905,25 @@ func addMasterEC2Policies(p *Policy, resource stringorslice.StringOrSlice, clust
 				},
 			},
 		},
+		&Statement{
+			Effect: StatementEffectAllow,
+			Action: stringorslice.Of(
+				"ec2:AttachVolume",                  // aws.go
+				"ec2:AuthorizeSecurityGroupIngress", // aws.go
+				"ec2:CreateRoute",                   // aws.go
+				"ec2:DeleteRoute",                   // aws.go
+				"ec2:DeleteSecurityGroup",           // aws.go
+				"ec2:DeleteVolume",                  // aws.go
+				"ec2:DetachVolume",                  // aws.go
+				"ec2:RevokeSecurityGroupIngress",    // aws.go
+			),
+			Resource: resource,
+			Condition: Condition{
+				"StringEquals": map[string]string{
+					"ec2:ResourceTag/kubernetes.io/cluster/" + clusterName: "owned",
+				},
+			},
+		},
 	)
 }
 
--- a/pkg/model/iam/tests/iam_builder_master_strict.json
+++ b/pkg/model/iam/tests/iam_builder_master_strict.json
@@ -52,6 +52,27 @@
       ]
     },
     {
+      "Action": [
+        "ec2:AttachVolume",
+        "ec2:AuthorizeSecurityGroupIngress",
+        "ec2:CreateRoute",
+        "ec2:DeleteRoute",
+        "ec2:DeleteSecurityGroup",
+        "ec2:DeleteVolume",
+        "ec2:DetachVolume",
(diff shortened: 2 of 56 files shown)


johngmyers (approved):

pkg/model/iam/iam_builder.go
```diff
@@ -905,6 +905,25 @@ func addMasterEC2Policies(p *Policy, resource stringorslice.StringOrSlice, clust
 				},
 			},
 		},
+		&Statement{
+			Effect: StatementEffectAllow,
+			Action: stringorslice.Of(
+				"ec2:AttachVolume",                  // aws.go
+				"ec2:AuthorizeSecurityGroupIngress", // aws.go
+				"ec2:CreateRoute",                   // aws.go
+				"ec2:DeleteRoute",                   // aws.go
```
rifelpet, line 914: These aren't related to volumes and the EBS CSI driver so I'm curious why they're being included here.

olemarkus: Looks like a mistake to me. But I am planning on moving the EBS specific stuff into a dedicated function to support IRSA. Can remove it as part of that PR.

olemarkus:
I believe we add legacy tags to volumes provisioned by the CSI driver as well.

johngmyers:
We should be moving away from depending on the legacy tags. In a release we should probably drop the old policy statement.

pkg/model/iam/iam_builder.go
wongma7: yes I just copied the existing KubernetesCluster policy without questioning it

wongma7:
@olemarkus , the CSI driver needs this patch to add legacy tags https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/932. The CSI driver installed by kops will need to have the --k8s-tag-cluster-id argument set on it if not already, I'll check

olemarkus:
We pass it in through `--extra-tags`. It is a generic list of tags that users can configure on all resources kOps/k8s provision. By default it contains the `KubernetesCluster` tag.

PR #11737, opened 2021-06-11
Set BindAddress appropriately when in IPv6-only mode

johngmyers:

### DIFF
--- a/pkg/model/components/apiserver.go
+++ b/pkg/model/components/apiserver.go
@@ -129,7 +129,11 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
 	c.LogLevel = 2
 	c.SecurePort = 443
 
-	c.BindAddress = "0.0.0.0"
+	if clusterSpec.IsIPv6Only() {
+		c.BindAddress = "::"
+	} else {
+		c.BindAddress = "0.0.0.0"
+	}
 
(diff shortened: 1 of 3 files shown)


pkg/model/components/apiserver.go
```diff
@@ -129,7 +129,11 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
 	c.LogLevel = 2
 	c.SecurePort = 443
 
-	c.BindAddress = "0.0.0.0"
+	if clusterSpec.IsIPv6Only() {
```
hakman, line 132: Any thoughts about using `IsIPv6CIDR`?
```suggestion
	if cidr.IsIPv6CIDR(clusterSpec.NonMasqueradeCIDR) {
```

hakman (changes requested):

pkg/model/components/apiserver.go
johngmyers: The fact that we are testing `NonMasqueradeCIDR` is not knowledge that should be spread all over the codebase. It should be centralized in one place.

hakman: I don't think there will be that many places where this will be used, but I don't mind.
Maybe use `cidr.IsIPv6CIDR` inside `IsIPv6Only()`, to use the same approach everywhere in code?

hakman (approved):

PR #11743, opened 2021-06-11
Make forwardToKubeDNS work in the NodeLocal DNSCache template

ederst:
This fixes https://github.com/kubernetes/kops/issues/11742 which discusses the broken rendering of the Corefile of the NodeLocal DNSCache template when setting forwardToKubeDNS to false (or not setting it).

Previously, due to not dereferencing the bool pointer, the Corefile was always rendered with the true clause, due to checking the address instead of the actual value of the variable.

With this fix, the templating mechanism will actually respect the value of forwardToKubeDNS and render it accordingly.


johngmyers (approved):
Thanks!

PR #11745, opened 2021-06-12
Allow unsetting fields from the command line

johngmyers:


cmd/kops/create_cluster.go
```diff
@@ -70,8 +70,10 @@ type CreateClusterOptions struct {
 	// SSHPublicKeys is a map of the SSH public keys we should configure; required on AWS, not required on GCE
 	SSHPublicKeys map[string][]byte
 
-	// Overrides allows settings values direct in the spec
+	// Overrides allows setting values directly in the spec.
 	Overrides []string
```
hakman, line 74: Maybe we can rename 'Overrides' to `Sets` to be consistent.
/hold in case you want to do that
/lgtm
/approve

johngmyers:
I will do the rename in a followup PR.
/hold cancel

PR #11747, opened 2021-06-12
Allow using IRSA for EBS CSI Driver

olemarkus:


johngmyers:
Can we not remove permissions from the instance role when both the AWS driver and IRSA are enabled?

I notice a bunch of permissions are using the legacy tags. Are these all for dealing with volumes previously created? I hope we're not relying on having legacy tags on new volumes.

olemarkus:
We are not relying on the legacy tags beyond these permissions and we can change these policies later when the in-tree plugin is no longer supported.

olemarkus:
We can also remove permissions from the master role, but I want to see some green e2es before I start removing things. But I do commit to removing as much of the master permissions as we can for the 1.22 release.

johngmyers (approved):

PR #11748, opened 2021-06-12
Enable ability to use IRSA for cluster autoscaler

olemarkus:


upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
```diff
@@ -531,6 +532,23 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*chann
 				})
 			}
 		}
+
+		if b.UseServiceAccountIAM() {
+			awsModelContext := &awsmodel.AWSModelContext{
+				KopsModelContext: b.KopsModelContext,
+			}
+
+			serviceAccountRoles := []iam.Subject{&clusterautoscaler.ServiceAccount{}}
```
johngmyers, line 541: We should refactor this to avoid copy-pasta. Pull `serviceAccountRoles` to the beginning of `buildAddons()`, push `awsModelContext` and the for loop to the end of `buildAddons()`, and have this block of code and the ones for dns-controller and aws-load-balancer-controller just append to `serviceAccountRoles`

pkg/model/iam/iam_builder.go
```diff
@@ -762,6 +762,34 @@ func AddAWSLoadbalancerControllerPermissions(p *Policy, resource stringorslice.S
...
+				"StringEquals": map[string]string{
+					"autoscaling:ResourceTag/KubernetesCluster": clusterName,
+				},
+			},
+		},
+		&Statement{
+			Effect: StatementEffectAllow,
+			Action: stringorslice.Of(
+				"autoscaling:DescribeAutoScalingGroups",
+				"autoscaling:DescribeAutoScalingInstances",
```
johngmyers, line 785: `addMasterASPolicies()` has this conditioned on `"autoscaling:ResourceTag/KubernetesCluster": clusterName`

Can any of the privileges in `addMasterASPolicies()` be trimmed when using service account IAM?

pkg/model/components/addonmanifests/remap.go
```diff
@@ -119,6 +120,8 @@ func getWellknownServiceAccount(name string) iam.Subject {
 	switch name {
 	case "aws-load-balancer-controller":
 		return &awsloadbalancercontroller.ServiceAccount{}
+	case "cluster-autoscaler":
+		return &clusterautoscaler.ServiceAccount{}
```
johngmyers, line 124: Any corresponding change needed to `RemapAddonManifest()`?

johngmyers (changes requested):

johngmyers:
Typo in PR title: "abilitty"

pkg/model/iam/iam_builder.go
olemarkus: The goal is to strip down the CP privileges to only what is needed by KCM, nodeup, and protokube. The first of those can even be removed when we use external CCM.

But I am thinking it is best to do this in two iterations. Users may rely on assuming these privileges and we need to give users a version or two to migrate to IRSA.

`Describe*` doesn't support conditions: https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2autoscaling.html

johngmyers: I would expect the `addMasterASPolicies()` path to call `AddClusterAutoscalerPermissions()` in the non-IRSA case.

I don't see the benefit of leaving those permissions at the instance level when IRSA is enabled. Without the permissions removed, admins won't learn of things that are relying on them.

olemarkus: I completely agree with that. I am just wondering if this PR is the right time to do so. We are working on improving e2e for addons, including IRSA versions of those tests. With those in place we have more confident we are not accidentally removing permissions that other addons may need too.

johngmyers: It would make sense for the IRSA e2e test to enable as many of these as possible. But we should try to catch this sort of breakage close to the IRSAing of each addon.

olemarkus: For CAS there is no e2e test that I am aware of unfortunately. And it is also a bit of a special case as the permissions to run CAS on the CP nodes were added long before the addon was actually supported.

With LB controller (the other PR), I can drop the addon-specific permissions from CP when IRSA is enabled though.

johngmyers: We should definitely drop the permissions on the CP nodes if both CAS and ServiceAccountIAM are enabled.

We could consider including the permissions when CAS is disabled, even when ServiceAccountIAM is enabled, but I think this is a poor choice. Enabling ServiceAccountIAM is expressing some sort of desire to have tighter access control.

johngmyers:
I consider the lack of dropping permissions on the instance role as still pending.

olemarkus:
Yep. Have a look now.

johngmyers (approved):

PR #11750, opened 2021-06-13
Set containerd config on nodeup.Config instead of clusterspec

olemarkus:
This allows us to set a default containerd config per IG (e.g add a different config for GPU IGs)

Can also be considered a cleanup as we no longer use containerd.overrideConfig as a mechanism for bringing the default containerd config from cloudup to nodeup.


johngmyers:
Are these so big they can't fit in userdata? I'm trying to keep non-empty AuxConfig to the uncommon case if possible.

olemarkus:
It's a good question. toml format is quite verbose. The default config is 500bytes, with GPU config is twice as big. Then people can add x number of registry mirror configurations, that is around 300 bytes each.

Since it to some extent can be arbitrary bytes long, I thought it was a good candidate.

hakman:
We still have compression of user data. Maybe wait a little more with this?

olemarkus:
If we believe the config to be small enough, changing to nodeup.Config (which the former version of this PR used) is trivial. I am happy with either.

johngmyers:
Let's go with nodeup.Config for now. I'd like to keep as few things in AuxConfig as possible as a likely future refactor would merge AuxConfig into Config and spill Config to VFS when it gets too big.

johngmyers:
FYI, I'd like to move the docker config into nodeup.Config as well. Doesn't have to be this PR.

pkg/apis/nodeup/config.go
```diff
@@ -73,6 +73,9 @@ type Config struct {
 	ConfigServer *ConfigServerOptions `json:"configServer,omitempty"`
 	// AuxConfigHash holds a secure hash of the nodeup.AuxConfig.
 	AuxConfigHash string
+
+	// Containerd config holds the configuration for containerd
+	Containerd *ContainerdConfig `json:"containerd,omitempty"`
```
hakman, line 78: Can be extended later if there's any need for it.
```suggestion
	ContainerdConfig string `json:",omitempty"`
```

hakman (changes requested):

pkg/apis/nodeup/config.go
olemarkus: This config is unversioned. I don't think we can extend it in the future as that would break nodeup/cloudup marshaling.

hakman: But it's just part of the launch template, which is updated together with new version of nodeup?

olemarkus: If I understand @johngmyers correctly, there is an idea about writing this to VFS at some point.

hakman: That is an alternative for large user data cases, if I understand correctly.
For containerd there are no flags to control various aspects, everything can be done through the config file, which will be generated in cloudup. So, I don't see any need for extending this short term.

johngmyers: The fact that it's unversioned isn't a problem because the reader (nodeup) always matches the version of the writer (kops). So we can make incompatible changes with impunity.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1384,5 +1385,35 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+		config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", "runc", "options", "SystemdCgroup"}, cluster.IsKubernetesGTE("1.20"))
+		if components.UsesKubenet(cluster.Spec.Networking) {
+			// Using containerd with Kubenet requires special configuration.
+			// This is a temporary backwards-compatible solution for kubenet users and will be deprecated when Kubenet is deprecated:
+			// https://github.com/containerd/containerd/blob/master/docs/cri/config.md#cni-config-template
+			config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "cni", "conf_template"}, "/etc/containerd/config-cni.template")
+		}
+		nodeupConfig.ContainerdConfig = config.String()
+	} else {
+		nodeupConfig.ContainerdConfig = *cluster.Spec.Containerd.ConfigOverride
```
hakman, line 1417: Could you move this first and return early?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1384,5 +1385,35 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 	config.Channels = n.channels
 	config.EtcdManifests = n.etcdManifests[role]
 
+	buildContainerdConfig(config, cluster)
```
hakman, line 1388: Should preserve the style.
```suggestion
	config.ContainerdConfig = buildContainerdConfig(cluster)
```

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1384,5 +1385,35 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+
 	return config, auxConfig, nil
+
+}
+
+func buildContainerdConfig(nodeupConfig *nodeup.Config, cluster *kops.Cluster) {
+	if cluster.Spec.ContainerRuntime != "containerd" {
```
hakman, line 1395: An extra check for `cluster.Spec.Containerd != nil` wouldn't hurt.

hakman (changes requested):

upup/pkg/fi/cloudup/apply_cluster.go
olemarkus: You mean moving this logic to the apis packages? I don't think this logic really has much to do with the API itself.
Given that we move stuff from the Cluster and IG resources to Nodeup.Config, we probably need more dedicated builders at some point similar to how we populate IG and Cluster.

hakman: Meant that you pass the `config` arg just to set `config.ContainerdConfig` at the end. So, instead of passing it to set it, just return a string from the builder.

olemarkus: yeah I realised I didn't read that right. sorry about that. Should be better now.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1384,5 +1385,40 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 	config.Channels = n.channels
 	config.EtcdManifests = n.etcdManifests[role]
 
+	if cluster.Spec.Containerd == nil {
+		cluster.Spec.Containerd = &kops.ContainerdConfig{}
+	}
+	config.ContainerdConfig = buildContainerdConfig(cluster)
```
hakman, line 1391: One last suggestion. This only makes sense to be done if container runtime is containerd.
/lgtm
```suggestion
	if cluster.ContainerRuntime == "containerd" {
		if cluster.Spec.Containerd == nil {
			cluster.Spec.Containerd = &kops.ContainerdConfig{}
		}
		config.ContainerdConfig = buildContainerdConfig(cluster)
	}
```

johngmyers (approved):

PR #11753, opened 2021-06-14
Cilium: disable masquerade by default when in ENI IPAM mode

johngmyers:
It doesn't make much sense to default to a non-working configuration.

/cc @olemarkus


olemarkus (approved):
One of those flags that probably shouldn't have been exposed to the user ...

Thanks for this

PR #11754, opened 2021-06-14
Enable IPv6 support for Cilium

johngmyers:
/cc @olemarkus @hakman 

Needs `--override=spec.kubeControllerManager.nodeCIDRMaskSize=80`. Will send that as a separate PR as it isn't specific to Cilium.

### DIFF
--- a/upup/pkg/fi/cloudup/template_functions.go
+++ b/upup/pkg/fi/cloudup/template_functions.go
@@ -223,6 +223,7 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		return strings.Join(labels, ",")
 	}
 
+	dest["IsIPv6Only"] = tf.IsIPv6Only
 	dest["UseServiceAccountIAM"] = tf.UseServiceAccountIAM
 
 	if cluster.Spec.NodeTerminationHandler != nil {
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -820,6 +820,10 @@ func validateNetworkingCilium(cluster *kops.Cluster, v *kops.CiliumNetworkingSpe
 			allErrs = append(allErrs, field.Invalid(versionFld, v.Version, "Only versions 1.8 through 1.10 are supported"))
 		}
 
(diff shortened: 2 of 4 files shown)


upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -223,6 +223,7 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		return strings.Join(labels, ",")
 	}
 
+	dest["IsIPv6Only"] = tf.IsIPv6Only
```
hakman, line 226: Same as for Calico, would you be ok with something like this? It would make the template more clear.
At this stage, it is irrelevant if cluster is IPv4 or IPv6, but only if they should be enabled or not.
Later, if we decide to add DualStack, we can keep the template and just modify the logic in the function(s).
```suggestion
	dest["EnableIPv4"] = !tf.IsIPv6Only
	dest["EnableIPv6"] = tf.IsIPv6Only
```

johngmyers: Dual stack is out of scope. If and when it is implemented it will require extensive changes, not just to this.

hakman: I never said it would not require changes or that we will ever do it, just that using `EnableIPv4` and `EnableIPv6` feels more natural in templates.

johngmyers: I disagree. Which would be more natural for selecting the health probe host?

There are only two states we support; that should be evident when reading the templates.

hakman: I am not quite sure the health probe host needs changing.

olemarkus: Cilium requires the health check to explicitly use localhost. If this isn't set, the health check will use the node IP, which will fail.

johngmyers: When I put Cilium into IPv6-only mode, it stopped listening to `127.0.0.1`.

hakman: Fine by me.

olemarkus (approved):

PR #11762, opened 2021-06-15
Default the NodeCIDRMaskSize appropriately for IPv6

johngmyers:


hakman:
What would be the reason to default to more than /120 -> 2^8=256 or maybe 118 -> 2^10=1024 pods per node?
k8s recommendation is 110 pods per node: https://kubernetes.io/docs/setup/best-practices/cluster-large/.

johngmyers:
Well, the NodeCIDRMaskSize can't be more than 16 bits larger than the ClusterCIDR prefix, so if we limit the NodeCIDRMaskSize to a minimum of 118 we'd have to limit the ClusterCIDR to a /102.

Aesthetically, it makes sense to split things on the colon boundaries. So that would mean a PodCIDR of /96 and a NodeCIDRMaskSize of 112. But that would be a different PR adjusting the default PodCIDR.

hakman (approved):

PR #11763, opened 2021-06-15
Make the AdminAccess default inclusive of IPv6

johngmyers:
The previous behavior did not match the documentation.


hakman:
Related Q, any thoughts about NLB/CLB defaulting to dual-stack?

hakman (approved):

johngmyers:
Kubernetes documents `loadBalancerSourceRanges` as defaulting to `0.0.0.0/0`. It would probably take a KEP to change that. And that might not fly due to the possibility of a workload choking on an IPv6 address in the PROXY protocol or whatever.

johngmyers:
The LB for the API server we should be dual-stack if APIAccess has both IPv4 and IPv6.

hakman:
Is there any reason not to default to dual-stack? IPv4 / IPv6 access could be controlled only by firewall rules.

johngmyers:
I can't think of any reason not to have this be controlled only by firewall rules.

Issue #11766, opened 2021-06-15
Cannot to rollout nodes on clusters that have Spot IGs


nettoclaudio:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
```
Version 1.19.1 (git-8589b4d157a9cb05c54e320c77b0724c4dd094b2)
```

I think the problem is happening since we merged the PR https://github.com/kubernetes/kops/pull/9118. So we're getting this problem around the v1.18.0 release.

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T21:15:16Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.18", GitCommit:"6f6ce59dc8fefde25a3ba0ef0047f4ec6662ef24", GitTreeState:"clean", BuildDate:"2021-04-15T03:23:41Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
AWS and Spot integration

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

```bash
KOPS_FEATURE_FLAGS="Spotinst,SpotinstOcean,SpotinstHybrid" \
kops rolling-update cluster --state s3://my-bucket --name alpha.k8s.example.com --yes
```

**5. What happened after the commands executed?**

Node validation fails when it reaches IGs managed as Spot instances. 

```
I0614 14:54:46.315524   17718 instancegroups.go:471] Cluster did not pass validation within deadline: InstanceGroup "my-spot-ig" is missing from the cloud provider.
```

**6. What did you expect to happen?**

Node validation should skip the Spot Instance Groups as they don't appear in the cloud provider.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  name: alpha.k8s.example.com
spec:
  api:
    loadBalancer:
[...]

PR #11767, opened 2021-06-15
fix: skipping node validation for spot IGs

nettoclaudio:
Fixes: #11766


johngmyers (changes requested):
This fix looks incorrect. It skips what seems a necessary check and will result in validation ignoring the state of the nodes in the IG.

Shouldn't the spontinst `GetCloudGroups()` be returning information on these groups?

nettoclaudio:
We don't need this fix if we activate the Spot integratation properly before run the `kops rolling-update` command.

PR #11771, opened 2021-06-15
fix enable default SC when EBS driver is not installed

olemarkus:
Looks like we are not installing a default StorageClass if AWS EBS CSI driver is disabled.

@justinsb @johngmyers another one that probably should go into 1.21.

/kind bug


yurrriq:
Thanks for this. It was a surprise on a fresh 1.21.0-beta.3 cluster to see no default StorageClass.

hakman (approved):

PR #11776, opened 2021-06-16
add replaceBeforeVersion field for channel

johngmyers:
This is resurrecting #8722. The original contributor gave up.

This is needed in order to change the `targetPort`s for the kube-dns `Service`, as `kubectl apply` does not work for changing the `targetPort`s of a Service that has two `ports` with the same `port` but different `protocol`s.


olemarkus:
/hold

sorry, but can this wait for #11712?

PR #11778, opened 2021-06-16
Also set haveUserInfo=true in case --user was provided in "kops export kubecfg"

codablock:
Without setting it to true, --user is completely ignored.


codablock:
I assume this must be cherry-picked into the next 1.21 beta and also into 1.20

olemarkus:
@yurrriq is this the same issue you mentioned?

yurrriq:
> @yurrriq is this the same issue you mentioned?

Seems related at least.

olemarkus:
> > @yurrriq is this the same issue you mentioned?
> 
> Seems related at least.

Are you able to test this patch and see if it works for you as well?

olemarkus (approved):

Issue #11783, opened 2021-06-16
Launch Openstack instances with config drive


ederst:
/kind feature

Currently, instances launched by the openstack cloud provider only use the metadata service of Openstack to gather their user data.

Due to various circumstances (no metadata service available, or an unstable metadata service, no dhcp enabled on the Openstack network) it might be necessary to use a config drive for letting the instance gather this data.

Luckily, this is just possible by passing `--config-drive true` [when creating a server](https://docs.openstack.org/nova/latest/user/metadata.html#config-drives) and it is also supported by [gophercloud](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L197).

To let kops users configure this behavior, I propose to add an additional configuration in the openstack cloud configuration in the cluster spec, like this:

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  name: my-cluster.k8s.local
spec:
  cloudConfig:
    openstack:
      metadata:
        configDrive: true|false
```

This option then can be passed to the appropriate openstack task(s) in kops and subsequently to the gophercloud create instance statement.

I will provide a PR to showcase this feature momentarily, in case someone wants to comment on it.

PR #11784, opened 2021-06-16
Launch Openstack instances with config drive

ederst:
Closes https://github.com/kubernetes/kops/issues/11783 by making it possible to specify to use a config drive instead of the metadata service as a source for the user data (cloudinit) of Openstack instances.


ederst:
@olemarkus @drekle @zetaab is there something I still need to do, or can this be merged?

zetaab (approved):
/approve
/lgtm

PR #11785, opened 2021-06-16
[cilium] Add validation rule for IPtables/Masquerade config combination

dntosas:
The `IPTablesRulesNoinstall` option will instruct Cilium to not create
any iptables rules to mess with pod traffic. But in order to carry
this setting, Cilium also needs eBPF masquerade enabled. 

Otherwise, Cilium will fallback to absent iptables rules/chains and eventually
network nodes will gradually be totally unreachable, without any sign of failure 
as Cilium pods will continue to operate normally.

In this commit, we add a validation rule to prevent users from applying configuration
that may break their network layer.

Ref: https://docs.cilium.io/en/v1.10/concepts/networking/masquerading/

Co-authored-by: hwoarang <markos@chandras.me>

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -867,6 +867,10 @@ func validateNetworkingCilium(cluster *kops.Cluster, v *kops.CiliumNetworkingSpe
 		allErrs = append(allErrs, field.Forbidden(fldPath.Child("enableL7Proxy"), "Cilium L7 Proxy requires IPTablesRules to be installed"))
 	}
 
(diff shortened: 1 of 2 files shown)


johngmyers:
Would it make sense to add to the defaulting logic?
/ok-to-test
/cc @olemarkus

olemarkus:
Yeah, I think it makes sense to default to not using iptables and enable bpf masq.

The way I would do this is to check for nil on both of these properties and if BPF NodePort is enabled, default to enabling bpf masq.

Can you add a commit for that to this PR?

I would also validate that BPF NodePort is enabled when BPF masq is enabled.

dntosas:
Cilium folks reported that this combination is legitimate so there is no point in adding this validation in here. Closing ^^

PR #11791, opened 2021-06-17
Delete all files in the provided discoveryStore on cluster deletion

olemarkus:


olemarkus:
/cc @rifelpet 
This one is also preventing `--down` from working. It fails to delete the discovery store when it is within the config store because `/discovery` is unknown.

johngmyers:
Should we delete everything or just `openid/v1/jwks/keys.json` and `.well-known/openid-configuration/discovery.json`?

johngmyers (approved):

PR #11795, opened 2021-06-17
Fix validating presence of AWS EBS CSI

olemarkus:
As of k8s 1.22, EBS driver is always installed. This PR makes validation pass when not explicitly enabling EBS CSI Driver.


johngmyers:
Shouldn't that validation do a k8s version check?

olemarkus:
Version check was missing before as well. But added now.

hakman (approved):

PR #11801, opened 2021-06-18
Run scenarios as presubmit tests

olemarkus:


olemarkus:
@johngmyers @rifelpet this should allow us to define presubmits of all the scenarios

olemarkus:
Node that this failing test is not related to this PR.

rifelpet (approved):

olemarkus:
I guess I cannot just stop that test. But #11811 should fix this, I think.

rifelpet (approved):

PR #11810, opened 2021-06-19
Enable cross-subnet mode with Calico by default

hakman:
Most kOps setup have at least 2 subnets so, the optimal setup is to:
1. disable AWS src/dest checks
2. set encapsulation mode to `CrossSubnet` (other possible options are `Always` and `None`)

/cc @johngmyers @olemarkus

### DIFF
--- a/pkg/apis/kops/networking.go
+++ b/pkg/apis/kops/networking.go
@@ -95,7 +95,7 @@ type WeaveNetworkingSpec struct {
 type FlannelNetworkingSpec struct {
 	// Backend is the backend overlay type we want to use (vxlan or udp)
 	Backend string `json:"backend,omitempty"`
-	// DisableTxChecksumOffloading is deprecated as of kops 1.19 and has no effect
+	// DisableTxChecksumOffloading is deprecated as of kOps 1.19 and has no effect.
 	DisableTxChecksumOffloading bool `json:"disableTxChecksumOffloading,omitempty"`
 	// IptablesResyncSeconds sets resync period for iptables rules, in seconds
 	IptablesResyncSeconds *int32 `json:"iptablesResyncSeconds,omitempty"`
@@ -109,7 +109,7 @@ type CalicoNetworkingSpec struct {
 	Version string `json:"version,omitempty"`
 
 	// AWSSrcDstCheck enables/disables ENI source/destination checks (AWS only)
-	// Options: DoNothing (default), Enable, or Disable
+	// Options: Disable (default), Enable, or DoNothing
 	AWSSrcDstCheck string `json:"awsSrcDstCheck,omitempty"`
 	// BPFEnabled enables the eBPF dataplane mode.
 	BPFEnabled bool `json:"bpfEnabled,omitempty"`
@@ -132,17 +132,19 @@ type CalicoNetworkingSpec struct {
 	ChainInsertMode string `json:"chainInsertMode,omitempty"`
 	// CPURequest CPU request of Calico container. Default: 100m
 	CPURequest *resource.Quantity `json:"cpuRequest,omitempty"`
-	// CrossSubnet enables Calico's cross-subnet mode when set to true
-	CrossSubnet bool `json:"crossSubnet,omitempty"`
+	// CrossSubnet is deprecated as of kOps 1.22 and has no effect
+	CrossSubnet *bool `json:"crossSubnet,omitempty"`
 	// EncapsulationMode specifies the network packet encapsulation protocol for Calico to use,
 	// employing such encapsulation at the necessary scope per the related CrossSubnet field. In
 	// "ipip" mode, Calico will use IP-in-IP encapsulation as needed. In "vxlan" mode, Calico will
 	// encapsulate packets as needed using the VXLAN scheme.
 	// Options: ipip (default) or vxlan
 	EncapsulationMode string `json:"encapsulationMode,omitempty"`
-	// IPIPMode is the encapsulation mode to use for the default Calico IPv4 pool created at start
-	// up, determining when to use IP-in-IP encapsulation, conveyed to the "calico-node" daemon
-	// container via the CALICO_IPV4POOL_IPIP environment variable.
+	// IPIPMode determines when to use IP-in-IP encapsulation for the default Calico IPv4 pool.
+	// It is conveyed to the "calico-node" daemon container via the CALICO_IPV4POOL_IPIP
+	// environment variable. EncapsulationMode must be set to "ipip".
+	// Options: "CrossSubnet", "Always", or "Never".
+	// Default: "CrossSubnet" if EncapsulationMode is "ipip", "Never" otherwise.
 	IPIPMode string `json:"ipipMode,omitempty"`
 	// IPv4AutoDetectionMethod configures how Calico chooses the IP address used to route
 	// between nodes.  This should be set when the host has multiple interfaces
@@ -183,6 +185,12 @@ type CalicoNetworkingSpec struct {
 	TyphaPrometheusMetricsPort int32 `json:"typhaPrometheusMetricsPort,omitempty"`
 	// TyphaReplicas is the number of replicas of Typha to deploy
 	TyphaReplicas int32 `json:"typhaReplicas,omitempty"`
+	// VXLANMode determines when to use VXLAN encapsulation for the default Calico IPv4 pool.
+	// It is conveyed to the "calico-node" daemon container via the CALICO_IPV4POOL_VXLAN
+	// environment variable. EncapsulationMode must be set to "vxlan".
+	// Options: "CrossSubnet", "Always", or "Never".
+	// Default: "CrossSubnet" if EncapsulationMode is "vxlan", "Never" otherwise.
+	VXLANMode string `json:"vxlanMode,omitempty"`
 	// WireguardEnabled enables WireGuard encryption for all on-the-wire pod-to-pod traffic
 	// (default: false)
 	WireguardEnabled bool `json:"wireguardEnabled,omitempty"`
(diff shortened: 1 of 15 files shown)


johngmyers:
I'm not terribly knowledgeable about Calico. @olemarkus do you have expertise in this area?

olemarkus:
No, I'm afraid not. Our e2e are not multi-az either. That would have been handy to verify this doesn't cause something unexpected to happen.

hakman:
We have this HA test for masters.
/test pull-kops-e2e-k8s-containerd-ha

hakman:
Seems all good. I can add more periodic tests later for vxlan and maybe bpf.

pkg/apis/kops/networking.go
```diff
@@ -183,6 +183,11 @@ type CalicoNetworkingSpec struct {
 	TyphaPrometheusMetricsPort int32 `json:"typhaPrometheusMetricsPort,omitempty"`
 	// TyphaReplicas is the number of replicas of Typha to deploy
 	TyphaReplicas int32 `json:"typhaReplicas,omitempty"`
+	// VXLANMode is the encapsulation mode to use for the default Calico IPv4 pool created at start
+	// up, determining when to use VXLAN encapsulation, conveyed to the "calico-node" daemon
+	// container via the CALICO_IPV4POOL_VXLAN environment variable.
```
johngmyers, line 188: This is a run-on sentence. It should be broken up into multiple sentences and/or simplified.

Document the possible values?

pkg/apis/kops/networking.go
```diff
@@ -203,7 +208,7 @@ type CanalNetworkingSpec struct {
 	// DisableFlannelForwardRules configures Flannel to NOT add the
 	// default ACCEPT traffic rules to the iptables FORWARD chain
 	DisableFlannelForwardRules bool `json:"disableFlannelForwardRules,omitempty"`
-	// DisableTxChecksumOffloading is deprecated as of kops 1.19 and has no effect
+	// DisableTxChecksumOffloading is deprecated as of kOps 1.19 and has no effect
```
johngmyers, line 211:
```suggestion
	// DisableTxChecksumOffloading is deprecated as of kOps 1.19 and has no effect.
```

pkg/apis/kops/validation/validation.go
```diff
@@ -1085,6 +1085,10 @@ func validateNetworkingCalico(v *kops.CalicoNetworkingSpec, e kops.EtcdClusterSp
 		allErrs = append(allErrs, IsValidValue(fldPath.Child("awsSrcDstCheck"), &v.AWSSrcDstCheck, valid)...)
 	}
 
+	if v.CrossSubnet != nil {
+		field.Invalid(fldPath.Child("crossSubnet"), v.CrossSubnet, "crossSubnet is deprecated, use awsSrcDstCheck instead")
```
johngmyers, line 1089: As coded here, It's not just deprecated.

I think we shouldn't validate this. At most, only fail validation if `CrossSubnet` is true and `AWSSrcDstCheck` is `DoNothing`.

johngmyers (changes requested):

pkg/apis/kops/networking.go
```diff
@@ -132,8 +132,8 @@ type CalicoNetworkingSpec struct {
 	ChainInsertMode string `json:"chainInsertMode,omitempty"`
 	// CPURequest CPU request of Calico container. Default: 100m
 	CPURequest *resource.Quantity `json:"cpuRequest,omitempty"`
-	// CrossSubnet enables Calico's cross-subnet mode when set to true
-	CrossSubnet bool `json:"crossSubnet,omitempty"`
+	// CrossSubnet is deprecated as of kOps 1.22 and has no effect
+	CrossSubnet *bool `json:"crossSubnet,omitempty"`
```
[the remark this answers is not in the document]
hakman: Seems a better check for it being set, in case someone forgot about it and really wants it disabled.

pkg/apis/kops/validation/validation.go
hakman: This was semi deprecated in docs a while ago. It shouldn't be used since kOps 1.19.
We should make it very clear that something i not being used anymore instead of just moving on.

hakman: If you have a strong preference here, I don't mind doing it though.

johngmyers: We're effectively making it default to `true`. So we shouldn't make it fail validation if it explicitly requests the now-default behavior.

It wasn't possible to explicitly turn off since it wasn't previously a pointer. So the only thing where it makes sense to fail validation is if it was explicitly set and something else contradictory is set, for example `AWSSrcDstCheck` set to `DoNothing`.

pkg/apis/kops/networking.go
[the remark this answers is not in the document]
johngmyers: It wasn't previously possible to explicitly set to disabled. The new way of explicitly disabling is to set `AWSSrcDstCheck`.

pkg/apis/kops/validation/validation.go
hakman: OK, updated.

pkg/apis/kops/networking.go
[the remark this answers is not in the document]
hakman: I want to check that is not set, for whatever reason.

pkg/apis/kops/networking.go
```diff
@@ -183,6 +184,11 @@ type CalicoNetworkingSpec struct {
 	TyphaPrometheusMetricsPort int32 `json:"typhaPrometheusMetricsPort,omitempty"`
 	// TyphaReplicas is the number of replicas of Typha to deploy
 	TyphaReplicas int32 `json:"typhaReplicas,omitempty"`
+	// VXLANMode is the encapsulation mode to use for the default Calico IPv4 pool created at start
+	// up, determining when to use VXLAN encapsulation, conveyed to the "calico-node" daemon
+	// container via the CALICO_IPV4POOL_VXLAN environment variable.
```
johngmyers, line 189:
```suggestion
	// VXLANMode determines when to use VXLAN encapsulation for the default Calico IPv4 pool.
	// It isconveyed to the "calico-node" daemon container via the CALICO_IPV4POOL_VXLAN
	// environment variable.
```

hakman: 👍

pkg/apis/kops/networking.go
```diff
@@ -132,17 +132,18 @@ type CalicoNetworkingSpec struct {
...
 	// encapsulate packets as needed using the VXLAN scheme.
 	// Options: ipip (default) or vxlan
 	EncapsulationMode string `json:"encapsulationMode,omitempty"`
-	// IPIPMode is the encapsulation mode to use for the default Calico IPv4 pool created at start
-	// up, determining when to use IP-in-IP encapsulation, conveyed to the "calico-node" daemon
-	// container via the CALICO_IPV4POOL_IPIP environment variable.
+	// IPIPMode determines when to use IP-in-IP encapsulation for the default Calico IPv4 pool.
+	// It is conveyed to the "calico-node" daemon container via the CALICO_IPV4POOL_IPIP
+	// environment variable.
+	// Default: CrossSubnet (other options: Always, Never)
```
johngmyers, line 146: Should we mention this is only used when `EncapsulationMode` is `ipip`?

pkg/apis/kops/networking.go
```diff
@@ -183,6 +184,11 @@ type CalicoNetworkingSpec struct {
 	TyphaPrometheusMetricsPort int32 `json:"typhaPrometheusMetricsPort,omitempty"`
 	// TyphaReplicas is the number of replicas of Typha to deploy
 	TyphaReplicas int32 `json:"typhaReplicas,omitempty"`
+	// VXLANMode determines when to use VXLAN encapsulation for the default Calico IPv4 pool.
+	// It is conveyed to the "calico-node" daemon container via the CALICO_IPV4POOL_VXLAN
+	// environment variable.
+	// Default: CrossSubnet (other options: Always, Never)
```
johngmyers, line 190: Should we mention this is only used when `EncapsulationMode` is `vxlan`?

pkg/apis/kops/networking.go
hakman: Yes, why not.

johngmyers: Ah, it defaults to "Never" if `EncapsulationMode` isn't `ipip`. So complicated.

hakman: Yes... Probably EncapsulationMode should have never existed.

pkg/apis/kops/networking.go
```diff
@@ -132,17 +132,18 @@ type CalicoNetworkingSpec struct {
...
 	// encapsulate packets as needed using the VXLAN scheme.
 	// Options: ipip (default) or vxlan
 	EncapsulationMode string `json:"encapsulationMode,omitempty"`
-	// IPIPMode is the encapsulation mode to use for the default Calico IPv4 pool created at start
-	// up, determining when to use IP-in-IP encapsulation, conveyed to the "calico-node" daemon
-	// container via the CALICO_IPV4POOL_IPIP environment variable.
+	// IPIPMode determines when to use IP-in-IP encapsulation for the default Calico IPv4 pool.
+	// It is conveyed to the "calico-node" daemon container via the CALICO_IPV4POOL_IPIP
+	// environment variable. EncapsulationMode must be set to "ipip".
+	// Default: CrossSubnet (other options: Always, Never)
```
johngmyers, line 146:
```suggestion
	// Options: "CrossSubnet", "Always", or "Never". Default: "CrossSubnet" if EncapsulationMode is "ipip", "Never" otherwise.
```

pkg/apis/kops/networking.go
```diff
@@ -183,6 +184,11 @@ type CalicoNetworkingSpec struct {
 	TyphaPrometheusMetricsPort int32 `json:"typhaPrometheusMetricsPort,omitempty"`
 	// TyphaReplicas is the number of replicas of Typha to deploy
 	TyphaReplicas int32 `json:"typhaReplicas,omitempty"`
+	// VXLANMode determines when to use VXLAN encapsulation for the default Calico IPv4 pool.
+	// It is conveyed to the "calico-node" daemon container via the CALICO_IPV4POOL_VXLAN
+	// environment variable. EncapsulationMode must be set to "vxlan".
+	// Default: CrossSubnet (other options: Always, Never)
```
johngmyers, line 190:
```suggestion
	// Options: "CrossSubnet", "Always", or "Never". Default: "CrossSubnet" if EncapsulationMode is "vxlan", "Never" otherwise.
```

johngmyers (changes requested):

pkg/apis/kops/networking.go
hakman: Done

pkg/apis/kops/networking.go
hakman: Done

PR #11814, opened 2021-06-19
Reduce policy size

olemarkus:
/kind bug

Some configurations end up with a policy that is too big. That is causing some e2es to fail.


hakman (approved):

rifelpet:
@olemarkus it looks like LBC job is still failing due to the IAM policy being too large: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-aws-load-balancer-controller/1407033641455849472

PR #11816, opened 2021-06-19
Make aws-cni config more flexible and generalized

### LINKED ISSUE #11144: AWS_VPC_K8S_CNI_RANDOMIZESNAT environment variable for `spec.networking.amazonvpc` resets to `prng` on every upgrade
 this information.**

1.19.1

 a `kops` flag.**

1.18.12

aws

In `cluster.yaml`

```
spec:
  networking:
    amazonvpc:
      env:
        - name: AWS_VPC_K8S_CNI_EXTERNALSNAT
          value: "false"
        - name: AWS_VPC_K8S_CNI_RANDOMIZESNAT
          value: none
        - name: MAX_ENI
          value: none
        - name: WARM_IP_TARGET
          value: "2"
```

`kops update cluster && kops rolling-update cluster`

From `kubectl get ds -n kube-system aws-node -oyaml`

```
spec:
  template:
    spec:
      containers:
      - env:
        - name: ADDITIONAL_ENI_TAGS
          value: '{}'
        - name: AWS_VPC_CNI_NODE_PORT_SUPPORT
          value: "true"
        - name: AWS_VPC_ENI_MTU
          value: "9001"
        - name: AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER
          value: "false"
        - name: AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG
          value: "false"
        - name: AWS_VPC_K8S_CNI_EXTERNALSNAT
          value: "prng"
        - name: AWS_VPC_K8S_CNI_LOGLEVEL
          value: DEBUG
        - name: AWS_VPC_K8S_CNI_LOG_FILE
          value: /host/var/log/aws-routed-eni/ipamd.log
        - name: AWS_VPC_K8S_CNI_RANDOMIZESNAT
          value: none
        - name: AWS_VPC_K8S_CNI_VETHPREFIX
          value: eni
        - name: AWS_VPC_K8S_PLUGIN_LOG_FILE
          value: /var/log/aws-routed-eni/plugin.log
        - name: AWS_VPC_K8S_PLUGIN_LOG_LEVEL
          value: DEBUG
        - name: DISABLE_INTROSPECTION
          value: "false"
        - name: DISABLE_METRICS
          value: "false"
        - name: ENABLE_POD_ENI
          value: "false"
        - name: MY_NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        - name: WARM_ENI_TARGET
          value: "1"
        - name: CLUSTER_NAME
          value: [REDACTED]
        - name: MAX_ENI
          value: "1"
        - name: WARM_IP_TARGET
          value: "1"
```

`AWS_VPC_K8S_CNI_EXTERNALSNAT` should be set to `none`

```
kind: Cluster
metadata:
  creationTimestamp: "2018-10-31T22:42:45Z"
  name: [REDACTED]
spec:
  additionalPolicies:
    node: |
      [
        [REDACTED]
      ]
  api:
    loadBalancer:
      class: Classic
      type: Private
  authorization:
    rbac: {}
  channel: stable
  cloudLabels:
    environment: [REDACTED]
  cloudProvider: aws
  configBase: s3://[REDACTED]
  dnsZone: [REDACTED]
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-us-east-1a
      name: a
      volumeType: gp3
    - instanceGroup: master-us-east-1b
      name: b
      volumeType: gp3
    - instanceGroup: master-us-east-1c
      name: c
      volumeType: gp3
    name: main
  - etcdMembers:
    - instanceGroup: master-us-east-1a
      name: a
      volumeType: gp3
    - instanceGroup: master-us-east-1b
      name: b
      volumeType: gp3
    - instanceGroup: master-us-east-1c
      name: c
      volumeType: gp3
    name: events
  iam:
    allowContainerRegistry: true
    legacy: false
  kubeControllerManager:
    horizontalPodAutoscalerDownscaleStabilization: 22m0s
  kubeDNS:
    provider: CoreDNS
  kubernetesApiAccess:
  - 0.0.0.0/0
  kubernetesVersion: 1.18.12
  masterInternalName: api.internal.[REDACTED]
  masterPublicName: api.[REDACTED]
  networkCIDR: 10.3.0.0/16
  networking:
    amazonvpc:
      env:
      - name: AWS_VPC_K8S_CNI_EXTERNALSNAT
        value: "false"
      - name: AWS_VPC_K8S_CNI_RANDOMIZESNAT
        value: none
      - name: MAX_ENI
        value: "1"
      - name: WARM_IP_TARGET
        value: "1"
      imageName: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.3
  sshAccess:
  - 0.0.0.0/0
  subnets: [REDACTED]
  topology:
    bastion:
      bastionPublicName: bastion.[REDACTED]
    dns:
      type: Public
    masters: private
    nodes: private
```

`AWS_VPC_K8S_CNI_RANDOMIZESNAT` continues to reset to `"prng"` on every upgrade

MoShitrit:
Fixes #11144 
/cc @hakman


pkg/apis/kops/networking.go
```diff
@@ -248,12 +248,19 @@ type RomanaNetworkingSpec struct {
 
 // AmazonVPCNetworkingSpec declares that we want Amazon VPC CNI networking
 type AmazonVPCNetworkingSpec struct {
+	// Specifies whether ipamd should configure rp filter for primary interface.
+	// Setting this to false will require rp filter to be configured through init container.
+	AwsVpcK8sCniConfigureRpFilter string   `json:"awsVpcK8sCniConfigureRpFilter,omitempty"`
+	Env                           []EnvVar `json:"env,omitempty"`
```
johngmyers, line 254: Should not remove the documentation comment

pkg/apis/kops/networking.go
```diff
@@ -248,12 +248,19 @@ type RomanaNetworkingSpec struct {
 
 // AmazonVPCNetworkingSpec declares that we want Amazon VPC CNI networking
 type AmazonVPCNetworkingSpec struct {
+	// Specifies whether ipamd should configure rp filter for primary interface.
```
johngmyers, line 251: Doc comment needs to start with the name of the field.

pkg/apis/kops/networking.go
```diff
@@ -248,12 +248,19 @@ type RomanaNetworkingSpec struct {
...
+	// Setting this to false will require rp filter to be configured through init container.
+	AwsVpcK8sCniConfigureRpFilter string   `json:"awsVpcK8sCniConfigureRpFilter,omitempty"`
+	Env                           []EnvVar `json:"env,omitempty"`
+	// If ENABLE_POD_ENI is set to true, in order for the kubelet to connect via TCP (for liveness or readiness probes) to pods that are using per pod security groups,
```
johngmyers, line 255: Doc comment needs to start with the name of the field.

upup/models/cloudup/resources/addons/networking.amazon-vpc-routed-eni/k8s-1.16.yaml.template
```diff
@@ -106,42 +106,35 @@
...
+        # - "name": "DISABLE_METRICS"
+        #   "value": "false"
+        # - "name": "ENABLE_POD_ENI"
+        #   "value": "false"
         - "name": "MY_NODE_NAME"
           "valueFrom":
             "fieldRef":
               "fieldPath": "spec.nodeName"
-        - "name": "WARM_ENI_TARGET"
-          "value": "1"
```
johngmyers, line 144: Did we want to comment this one out as well?

johngmyers (changes requested):

MoShitrit:
Thanks for the review, @johngmyers !
I've addressed all your comments, please have a look when you get a chance and let me know if you had any other concerns 😄 🙏🏻

pkg/apis/kops/networking.go
```diff
@@ -248,12 +248,20 @@ type RomanaNetworkingSpec struct {
 
 // AmazonVPCNetworkingSpec declares that we want Amazon VPC CNI networking
 type AmazonVPCNetworkingSpec struct {
-	// The container image name to use
-	ImageName string `json:"imageName,omitempty"`
-	// The init container image name to use
-	InitImageName string `json:"initImageName,omitempty"`
+	// AwsVpcK8sCniConfigureRpFilter specifies whether ipamd should configure rp filter for primary interface.
+	// Setting this to false will require rp filter to be configured through init container.
+	AwsVpcK8sCniConfigureRpFilter string `json:"awsVpcK8sCniConfigureRpFilter,omitempty"`
```
johngmyers, line 253: All of this refers to the AWS VPC CNI, so that part of the name is redundant.

"RP" is an acronym for "Reverse Path".

The possible values are "true" and "false", so it's a boolean.

```suggestion
	ConfigureRPFilter *bool `json:"configureRPFilter,omitempty"`
```

pkg/apis/kops/networking.go
```diff
@@ -248,12 +248,20 @@ type RomanaNetworkingSpec struct {
 
 // AmazonVPCNetworkingSpec declares that we want Amazon VPC CNI networking
 type AmazonVPCNetworkingSpec struct {
-	// The container image name to use
-	ImageName string `json:"imageName,omitempty"`
-	// The init container image name to use
-	InitImageName string `json:"initImageName,omitempty"`
+	// AwsVpcK8sCniConfigureRpFilter specifies whether ipamd should configure rp filter for primary interface.
+	// Setting this to false will require rp filter to be configured through init container.
```
johngmyers, line 252: Should mention the default is true.

pkg/apis/kops/networking.go
```diff
@@ -248,12 +248,20 @@ type RomanaNetworkingSpec struct {
...
+	AwsVpcK8sCniConfigureRpFilter string `json:"awsVpcK8sCniConfigureRpFilter,omitempty"`
 	// Env is a list of environment variables to set in the container.
 	Env []EnvVar `json:"env,omitempty"`
+	// DisableTCPEarlyDemux should be set to true for amazon-k8s-cni-init container under initcontainers if ENABLE_POD_ENI is set to true,
+	// in order for the kubelet to connect via TCP (for liveness or readiness probes) to pods that are using per pod security groups.
+	// This will increase the local TCP connection latency slightly.
+	// To use this setting, a Linux kernel version of at least 4.6 is needed on the worker node.
+	DisableTCPEarlyDemux string `json:"disableTCPEarlyDemux,omitempty"`
```
johngmyers, line 260: Also a boolean.

We try to avoid negative booleans.

```suggestion
	TCPEarlyDemux *bool `json:"enableTCPEarlyDemux,omitempty"`
```

pkg/apis/kops/networking.go
```diff
@@ -248,12 +248,20 @@ type RomanaNetworkingSpec struct {
...
+	// ImageName is the container image name to use.
+	ImageName string `json:"imageName,omitempty"`
```
johngmyers, line 262: Other subcomponents use `Image`

```suggestion
	Image string `json:"image,omitempty"`
```

pkg/apis/kops/networking.go
```diff
@@ -248,12 +248,20 @@ type RomanaNetworkingSpec struct {
...
+	// InitImageName is the init container image name to use.
+	InitImageName string `json:"initImageName,omitempty"`
```
johngmyers, line 264:
```suggestion
	InitImage string `json:"initImageName,omitempty"`
```

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -149,6 +149,22 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		}
 	}
 
+	if cluster.Spec.Networking != nil && cluster.Spec.Networking.AmazonVPC != nil {
```
johngmyers, line 152: Use `WithDefaultBool` in the template.

johngmyers (changes requested):

hakman (changes requested):
I still think that initializing the Env vars with the options kOps overrides by default in template functions is better and also allows users to change those too in a generic way, without adding extra options.

MoShitrit:
> I still think that initializing the Env vars with the options kOps overrides by default in template functions is better and also allows users to change those too in a generic way, without adding extra options.

So I did that with everything else, however it seems like these two are actually different from the default values according to the docs, so I needed to make an override on them. Is there a better way to handle that, than the one I used?

upup/models/cloudup/resources/addons/networking.amazon-vpc-routed-eni/k8s-1.16.yaml.template
MoShitrit: Done 👍🏻 
Good catch, thanks!

pkg/apis/kops/networking.go
MoShitrit: Yep, I didn't add that, simply moved it. 
Do we not order these parameters in an alphabetical fashion? 
And should I revert the rename back to `ImageName` and `InitImageName`?

upup/pkg/fi/cloudup/template_functions.go
MoShitrit: Removed the template functions and switched to use `WithDefaultBool` as suggested. Thanks!

pkg/apis/kops/networking.go
johngmyers: Yes, revert the rename. I hadn't realized it existed before.

hakman:
Hey @MoShitrit, sorry for the delay, but was traveling for the past few days. Will take another look at this today.

hakman:
I would add a template function like this:
```go
	if cluster.Spec.Networking != nil && cluster.Spec.Networking.AmazonVPC != nil {
		c := cluster.Spec.Networking.AmazonVPC
		dest["AmazonVpcEnvVars"] = func() map[string]string {
			envVars := map[string]string{
				"AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER": "false",
			}
			for _, e := range c.Env {
				envVars[e.Name] = e.Value
			}
			return envVars
		}
	}
```
and change the template to use it instead of directly accessing the env vars:
```go
        {{- range $name, $value := AmazonVpcEnvVars }}
        - "name": "{{ $name }}"
          "value": "{{ $value }}"
        {{- end }}
```
This way we can change the defaults, but still allow overwriting them.

upup/models/cloudup/resources/addons/networking.amazon-vpc-routed-eni/k8s-1.16.yaml.template
```diff
@@ -106,48 +106,44 @@
...
+        # - "name": "DISABLE_INTROSPECTION"
+        #   "value": "false"
+        # - "name": "DISABLE_METRICS"
+        #   "value": "false"
+        # - "name": "ENABLE_POD_ENI"
+        #   "value": "false"
         - "name": "MY_NODE_NAME"
           "valueFrom":
             "fieldRef":
               "fieldPath": "spec.nodeName"
```
hakman, line 142: Please comment this too to have a single block of comments and add it just before `CLUSTER_NAME`.

upup/models/cloudup/resources/addons/networking.amazon-vpc-routed-eni/k8s-1.16.yaml.template
```diff
@@ -106,48 +106,44 @@
                 - "fargate"
       "containers":
       - "env":
-        - "name": "ADDITIONAL_ENI_TAGS"
-          "value": "{}"
-        - "name": "AWS_VPC_CNI_NODE_PORT_SUPPORT"
-          "value": "true"
-        - "name": "AWS_VPC_ENI_MTU"
-          "value": "9001"
```
hakman, line 114: These should be added back as comments, seem to just be gone.

hakman (changes requested):
These are just some remaining nits, otherwise looks good from my point of view. Maybe also squash the commits.
@rifelpet Want to take a look too?

upup/models/cloudup/resources/addons/networking.amazon-vpc-routed-eni/k8s-1.16.yaml.template
hakman: Seems that this was not a good idea, looks worse. Would you mind reverting this change and squash?

MoShitrit: Done! 😄

hakman: Thanks!

rifelpet:
/lgtm

Will let @hakman do the final approval

hakman (approved):
Thanks @MoShitrit!

PR #11822, opened 2021-06-20
Support creating new service-account keypairs

johngmyers:


cmd/kops/create_keypair.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2019 The Kubernetes Authors.
```
johngmyers, line 2: The diff looks strange here. What I did is delete `create_keypair.go` and rename `create_keypair_ca.go` to `create_keypair.go` (and modify it).

olemarkus (approved):

PR #11825, opened 2021-06-21
Set EnableExternalCloudController to true by default

hakman:
This is not just about AWS, the OpenStack one for example is stable.
It means that someone should still enable the feature via `spec.cloudControllerManager.cloudProvider`, but no longer need to set the feature flag too.


olemarkus (approved):

johngmyers:
There's little point to having a feature flag then. There's no point to disabling it. The only reason to keep it would be automated tests that need to work across multiple kOps versions.

hakman:
I thought the procedure is to set it to true in case we want to flip it back.

johngmyers:
If the flag affected behavior beyond simply allowing options to be set, it would make sense to allow it to be turned back off. But this flag merely allows a field in the cluster spec to be non-nil.

PR #11836, opened 2021-06-22
Don't reconcile roles and policies if a profile is provided

olemarkus:
If we are using a shared IAM profile, kops shouldn't try to do anything with policies and profile role.

### DIFF
--- a/pkg/model/awsmodel/iam.go
+++ b/pkg/model/awsmodel/iam.go
@@ -279,15 +279,6 @@ func (b *IAMModelBuilder) roleKey(role iam.Subject) (string, bool) {
 func (b *IAMModelBuilder) buildIAMTasks(role iam.Subject, iamName string, c *fi.ModelBuilderContext, shared bool) error {
 	roleKey, _ := b.roleKey(role)
 
-	iamRole, err := b.buildIAMRole(role, iamName, c)
-	if err != nil {
-		return err
-	}
-
-	if err := b.buildIAMRolePolicy(role, iamName, iamRole, c); err != nil {
-		return err
-	}
-
 	{
 		// To minimize diff for easier code review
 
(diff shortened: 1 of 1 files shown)


pkg/model/awsmodel/iam.go
```diff
@@ -333,10 +333,7 @@ func (b *IAMModelBuilder) buildIAMTasks(role iam.Subject, iamName string, c *fi.
 			}
 
 			c.AddTask(t)
-		}
 
-		// Generate additional policies if needed, and attach to existing role
-		if !shared {
```
hakman, line 339: Does it hurt in any way to leave this as 2 separate blocks of code?

olemarkus: `iamRole` is defined in one and used in the other.

hakman: Maybe just create 2 blocks and keep the original comments for each block?

olemarkus: yeah removing the comments were unecessary.

pkg/model/awsmodel/iam.go
```diff
@@ -297,24 +288,34 @@ func (b *IAMModelBuilder) buildIAMTasks(role iam.Subject, iamName string, c *fi.
...
+				Tags:      b.CloudTags(iamName, shared),
 			}
 			c.AddTask(iamInstanceProfile)
 		}
 
-		{
-			iamInstanceProfileRole := &awstasks.IAMInstanceProfileRole{
-				Name:      fi.String(iamName),
-				Lifecycle: b.Lifecycle,
+		// Create External Policy tasks
```
hakman, line 296: This is duplicated now.
```suggestion
```

pkg/model/awsmodel/iam.go
```diff
@@ -333,10 +334,8 @@ func (b *IAMModelBuilder) buildIAMTasks(role iam.Subject, iamName string, c *fi.
 			}
 
 			c.AddTask(t)
-		}
 
-		// Generate additional policies if needed, and attach to existing role
-		if !shared {
+			// Generate additional policies if needed, and attach to existing role
```
hakman, line 338: Having this as a block would make it more readable, but your choice.
```suggestion
			// Generate additional policies if needed, and attach to existing role
			{
```

pkg/model/awsmodel/iam.go
```diff
@@ -297,24 +288,34 @@ func (b *IAMModelBuilder) buildIAMTasks(role iam.Subject, iamName string, c *fi.
...
+					Name:      fi.String(iamName),
+					Lifecycle: b.Lifecycle,
+
+					InstanceProfile: iamInstanceProfile,
+					Role:            iamRole,
+				}
+				c.AddTask(iamInstanceProfileRole)
+			}
+
+			// Create External Policy tasks
```
hakman, line 318: Having this as a block would make it more readable, but your choice.
```suggestion
			// Create External Policy tasks
			{
```

hakman (changes requested):
A few more nits, will finish the review once they are addressed.

hakman (approved):

PR #11838, opened 2021-06-22
Kubetest2 fix periodic end to end tests

olemarkus:


tests/e2e/kubetest2-kops/aws/zones.go
```diff
@@ -75,7 +75,6 @@ var allZones = []string{
 	"us-east-2a",
 	"us-east-2b",
 	"us-east-2c",
-	"us-west-1a",
 	"us-west-1b",
```
hakman, line 78: Please just comment these same as the one below. We can live without any `us-west-1`.

hakman (changes requested):

tests/e2e/kubetest2-kops/aws/zones.go
olemarkus: Is there a benefit to comment them out?

hakman: Yes. At some point someone will wonder why there's no `us-west-1` and he will add it, then remove it again.
This way maybe he reads the comment first :D.

olemarkus: For me 1c exists. It probably depends on the account which one gets cut.
I'll leave this change out for now.

hakman (approved):

olemarkus:
Shouldn't that test have been skipped?

rifelpet:
In the ginkgo output i see:

```
volume snapshot controller  [It] should check snapshot fields
```

Which doesn't exactly match the skip regex. Maybe drop the "volume.snapshot.controller" portion.

olemarkus:
I thought some of those bits were only used when pretty printing. not actually used for the regex. But apparently I was wrong there.

PR #11841, opened 2021-06-22
Kubetest2 - Add --skip-regex logic

rifelpet:
This will move the logic from [build_jobs.py](https://github.com/kubernetes/test-infra/blob/e4a29d493d9fa28a4ce506abd544c60375879a2b/config/jobs/kubernetes/kops/build_jobs.py#L284-L312) into kubetest2 itself.
This way we can rely on this logic from scenario scripts and other places that aren't directly set by build_jobs.py.

ref: https://kubernetes.slack.com/archives/C8MKE2G5P/p1624366418361000

### DIFF
--- a/tests/e2e/pkg/tester/skip_regex.go
+++ b/tests/e2e/pkg/tester/skip_regex.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tester
+
+import (
+	"regexp"
+	"strings"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


tests/e2e/pkg/tester/skip_regex.go
```diff
@@ -0,0 +1,93 @@
...
+		if strings.Contains(cluster.Spec.KubernetesVersion, "v1.21.") {
+			// TODO(rifelpet): Remove once k8s tags has been created that include
+			// https://github.com/kubernetes/kubernetes/pull/101443
+			skipRegex += "|Invalid.AWS.KMS.key"
+		}
+		if strings.Contains(cluster.Spec.KubernetesVersion, "v1.22.") {
+			// TODO(rifelpet): Remove once volume limits tests have been fixed
+			// https://github.com/kubernetes/kubernetes/issues/79660#issuecomment-854884112
+			skipRegex += "|Volume.limits.should.verify.that.all.nodes.have.volume.limits"
+		}
```
rifelpet, line 73: I'm not particularly excited about this:

In build_jobs.py we have version markers that we can reference (`stable`, `latest`, etc.). kubetest2-kops takes these marker URLs, reads their contents and passes that to `kops create cluster` so we end up with a specific k8s version URL in the cluster spec like [this](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-pipeline-updown-kopsmaster/1407306000108621824/artifacts/cluster.yaml).

kubetest2-tester-kops doesn't have access to the version marker being passed into kubetest2-kops, so we rely on parsing these known versions from the cluster spec instead. The potential downside is that if these k/k issues aren't fixed in 1.22, we need to remember to add v1.23 to this logic whereas in bulid_jobs.py the `stable` and `latest` markers wouldn't need any followup work. Maybe it is better to explicitly require this followup work though to ensure that these tests still need to be skipped.

olemarkus (approved):

PR #11843, opened 2021-06-22
Reduce policy size further

olemarkus:
This PR reduces policies by using a set for the actions that have well-known conditions (any or tagged by the cluster tag).


pkg/model/iam/iam_builder.go
```diff
@@ -49,12 +49,35 @@ const PolicyDefaultVersion = "2012-10-17"
 
 // Policy Struct is a collection of fields that form a valid AWS policy document
 type Policy struct {
-	Statement []*Statement
-	Version   string
+	clusterName         string
+	unconditionalAction sets.String
+	clusterTaggedAction sets.String
+	Statement           []*Statement
+	Version             string
```
johngmyers, line 56: Does this need to be public? We could have separate builder and built types. (for future PR)

pkg/model/iam/iam_builder.go
```diff
@@ -49,12 +49,35 @@ const PolicyDefaultVersion = "2012-10-17"
...
+		})
+	}
+	if len(p.clusterTaggedAction) > 0 {
+		p.Statement = append(p.Statement, &Statement{
+			Effect:   StatementEffectAllow,
+			Action:   stringorslice.Of(p.clusterTaggedAction.List()...),
+			Resource: stringorslice.String("*"),
+			Condition: Condition{
+				"StringEquals": map[string]string{
+					"aws:ResourceTag/KubernetesCluster": p.clusterName,
```
johngmyers, line 75: Do we have a plan to move off of the legacy tags? Perhaps we could add a condition that allows both? (for future PR)

pkg/model/iam/iam_builder.go
```diff
@@ -236,17 +259,24 @@ func (b *PolicyBuilder) BuildAWSPolicy() (*Policy, error) {
 	return p, nil
 }
 
+func NewPolicy(clusterName string) *Policy {
```
johngmyers, line 262: Would it make sense for this to be a receiver on `*PolicyBuilder`? It could then store the `PolicyBuilder` in the `Policy` and a lot of receivers on `*PolicyBuilder` could instead be on `*Policy`.

I think `PolicyBuilder` might be better named `PolicyFactory`.

pkg/model/iam/iam_builder.go
```diff
@@ -747,105 +781,61 @@ func AddAWSLoadbalancerControllerPermissions(p *Policy, resource stringorslice.S
...
-		&Statement{
-			Effect: StatementEffectAllow,
-			Action: stringorslice.Of(
-				"autoscaling:DescribeAutoScalingGroups",
-				"autoscaling:DescribeAutoScalingInstances",
-				"autoscaling:DescribeLaunchConfigurations",
-			),
-			Resource: resource,
-		},
+func AddClusterAutoscalerPermissions(p *Policy) {
```
johngmyers, line 794: Perhaps these would be better written as receivers on `*Policy`? They could then return `p` for fluent use.

johngmyers (commented):
The first commit is extremely big to review. It would be much easier if it were broken up into smaller commits.

I like to put nontrivial refactors in separate commits than semantic changes. When golden files change significantly I like to put them in an immediately following commit.

pkg/model/iam/iam_builder.go
olemarkus: Yeah, I am not happy about this, but this is due to consumers calling `toJSON` on the built type directly. Its a bit more work to separate this out so that they rather receive a string resource or something.

pkg/model/iam/iam_builder.go
olemarkus: I _think_ in-tree CCM still relies on the legacy tags. But I think we can eventually move away from this soon. Putting this into one location anyways make this a lot simpler.

pkg/model/iam/iam_builder.go
olemarkus: Since all these Add* are becoming component-specific, I am considering moving these to be with the subject. Today this is in `pkg/model/components/addonmanifests/<addon>/iam.go`

olemarkus:
I think that was a lot of valuable input and suggestions, but perhaps out of the scope of this PR. Would be nice to get this in as the policy limit is blocking/breaking stuff right now.

johngmyers:
The only thing blocking is the difficulty of reviewing such a large commit.

olemarkus:
The code changes aren't that big, are they? For the generated output files, the changes would be fairly big even if this was done addon-by-addon. https://github.com/kubernetes/kops/pull/11843/files#diff-dfcb900da852396ff87561d730a05b2c55fb43e394e4cbe7586a913fc84c9400 gives perhaps the best idea of how this change.

The new function in one go both deduplicates and sorts as a consequence of using sets.

hakman:
I managed to follow the commits and seems fine. Now that they are split, it is a little easier.
/lgtm

rifelpet:
Some of the condition keys are changing for certain actions:
* `autoscaling:SetDesiredCapacity` is changing from `autoscaling:ResourceTag/KubernetesCluster` to `aws:ResourceTag/KubernetesCluster`
* `ec2:DeleteTags` and `ec2:CreateRoute` are changing from `ec2:ResourceTag/KubernetesCluster` to `aws:ResourceTag/KubernetesCluster`

Given that some of these actions are not exercised during the e2e tests, have we confirmed these are compatible changes?

The [aws:ResourceTag](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-resourcetag) docs mention:
> This key is included in the request context when the requested resource already has attached tags.

which I would guess would not apply to a `Create*` request and instead should use the [aws:RequestTag](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-requesttag) condition instead. I also don't know if there is a functional difference between the aws, ec2, and autoscaling variants of these conditions. Does anyone know if they are interchangeable?

olemarkus:
`aws:ResourceTag` is interchangeable with the more specific ones. They were added to make it easier to build policies.

`Create*` should indeed use `aws:RequestTag` in most cases, but `CreateRoute` doesn't actually create an AWS resource. It adds a route to a RouteTable, and the Conditions are against the RouteTable.  

You can verify both claims on this page: https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html

pkg/model/iam/iam_builder.go
```diff
@@ -1158,27 +1121,11 @@ func addAmazonVPCCNIPermissions(p *Policy, resource stringorslice.StringOrSlice,
...
+		"autoscaling:CompleteLifecycleAction",
+		"autoscaling:DescribeAutoScalingInstances",
+		"sqs:DeleteMessage",
+		"sqs:ReceiveMessage",
 	)
 }
-
-func createResource(b *PolicyBuilder) stringorslice.StringOrSlice {
-	var resource stringorslice.StringOrSlice
-	if b.ResourceARN != nil {
```
johngmyers, line 1274: Should we get rid of the rest of `ResourceARN`?

johngmyers (approved):

PR #11849, opened 2021-06-23
Use host network for ebs-csi-controller in IPv6 only clusters

hakman:
This should get the IPv6 storage tests green again: https://testgrid.k8s.io/kops-misc#kops-grid-scenario-ipv6

IMDS is not available on IPv6 and this is an easy workaround while IPv6 is in experimental stage.
```
Jun 22 11:52:16.640: INFO: At 2021-06-22 11:47:14 +0000 UTC - event for awstt6j2: {persistentvolume-controller } WaitForFirstConsumer: waiting for first consumer to be created before binding
Jun 22 11:52:16.640: INFO: At 2021-06-22 11:47:14 +0000 UTC - event for awstt6j2: {persistentvolume-controller } ExternalProvisioning: waiting for a volume to be created, either by external provisioner "ebs.csi.aws.com" or manually created by system administrator
Jun 22 11:52:16.640: INFO: At 2021-06-22 11:47:14 +0000 UTC - event for awstt6j2: {ebs.csi.aws.com_ebs-csi-controller-5df8cc7bfb-hs4r4_fab53035-fcf2-404e-8d20-8e2f0bdb89dc } Provisioning: External provisioner is provisioning volume for claim "volume-4387/awstt6j2"
Jun 22 11:52:16.640: INFO: At 2021-06-22 11:47:14 +0000 UTC - event for awstt6j2: {ebs.csi.aws.com_ebs-csi-controller-5df8cc7bfb-hs4r4_fab53035-fcf2-404e-8d20-8e2f0bdb89dc } ProvisioningFailed: failed to provision volume with StorageClass "volume-4387fpvrk": rpc error: code = Internal desc = NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, .
EC2RoleRequestError: no EC2 instance role found
caused by: RequestError: send request failed
caused by: Get "http://169.254.169.254/latest/meta-data/iam/security-credentials/": dial tcp 169.254.169.254:80: connect: network is unreachable
```

### DIFF
--- a/upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
+++ b/upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
@@ -421,6 +421,9 @@ spec:
       serviceAccountName: ebs-csi-controller-sa
       priorityClassName: system-cluster-critical
       {{ if not UseServiceAccountIAM }}
+      {{ if IsIPv6Only }}
+      hostNetwork: true
+      {{ end }}
       tolerations:
         - operator: Exists
       {{ end }}
@@ -460,7 +463,7 @@ spec:
               mountPath: /var/lib/csi/sockets/pluginproxy/
           ports:
             - name: healthz
-              containerPort: 9808
+              containerPort: 9809
               protocol: TCP
           livenessProbe:
             httpGet:
--- a/upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
+++ b/upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
@@ -1005,7 +1005,7 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*chann
 		if b.Cluster.Spec.CloudConfig != nil && b.Cluster.Spec.CloudConfig.AWSEBSCSIDriver != nil && fi.BoolValue(b.Cluster.Spec.CloudConfig.AWSEBSCSIDriver.Enabled) {
 			key := "aws-ebs-csi-driver.addons.k8s.io"
 
-			version := "1.0.0-kops.1"
+			version := "1.1.0-kops.1"
 			{
 				id := "k8s-1.17"
 				location := key + "/" + id + ".yaml"
--- a/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/awscloudcontroller/manifest.yaml
+++ b/upup/pkg/fi/cloudup/tests/bootstrapchannelbuilder/awscloudcontroller/manifest.yaml
@@ -63,8 +63,8 @@ spec:
   - id: k8s-1.17
     kubernetesVersion: '>=1.17.0'
     manifest: aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml
-    manifestHash: 59c7723a8a9271558c6a87ccb0f0c61c2e36c6ed
+    manifestHash: 037b4ae065af38e8347bbec053c1503aa3c51a6c
     name: aws-ebs-csi-driver.addons.k8s.io
     selector:
       k8s-addon: aws-ebs-csi-driver.addons.k8s.io
-    version: 1.0.0-kops.1
+    version: 1.1.0-kops.1


upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
```diff
@@ -460,7 +463,7 @@ spec:
               mountPath: /var/lib/csi/sockets/pluginproxy/
           ports:
             - name: healthz
-              containerPort: 9808
+              containerPort: 9809
```
johngmyers, line 466: Should register the port in `pkg/wellknownports/wellknownports.go`. Perhaps take 3987?

olemarkus: I don't want to pollute the templates with this test code very long, so there shouldn't be any need to document this.

hakman: This can stay as if forever, it's just a port number. There are others also that are not documented there.
I will try not to "pollute" the templates with my test code after this.

johngmyers: What's the exit strategy? Sure, we'll need NAT64/DNS64 somewhere for this to be practical, but that means we'll need to get it into our test infra somehow.

olemarkus: What I mean is that hostnetworking for ebs driver will have to be removed fairly shortly as we need a solution that will work for other addons too. Putting them all on hostnetworking is not a solution going forward.

olemarkus: https://kubernetes.io/docs/tasks/administer-cluster/ip-masq-agent/ apparently. Cilium also moving in that direction with their own bpf version.

hakman: Technically, how would the ip-masq-agent work in this case to route traffic to 169.254.169.254 when there is no IPv4 address on the pod?

hakman: In this case, the solution may be IRSA, as it seems it only needs some credentials. I chose this route to get IPv6 tested faster and address remaining issues in networking while there's interest from upstream in fixing them.

rifelpet: IRSA and IPv6 are at similar levels of maturity right? Could we require IRSA if a cluster is using IPv6 and EBS-CSI-controller ?

hakman: We can, but would rather do it during beta. This way each feature can be worked independently.

olemarkus: So a couple of points here. The ip-masq-agent does not support NAT64, so it is not a solution. Most CNIs do (with cilium you can curl ipv4-only addresses that is outside of the cluster), but they all seem to explicitly consider link local non-NATable. So there is a needed functionality gap here. One can hope for that IMDS on ipv6 is a part of the extended ipv6 support that has been on the AWS container roadmap for some time.

I consider IRSA for EBS CSI GA, and don't have anything more planned in that area. What remains on IRSA side is to add support for the remaining addons, but that is definitely a parallel task.

What should be done on the IRSA side is to move from feature flag to an API flag.

hakman: As I said, I don't mind IPv6 to depend on IRSA, but would rather not do it at this stage. That's all.
I even agreed to do it later in the dev cycle, so I think this doesn't deserve so much debate, unless I am missing something.

Regarding what CNIs do, accessing IPv4 only addresses requires an IPv4 address allocated to the pod, so DualStack at least at CNI level to get a default route for IPv4.
Calico can do it too, but as it's configured now, it would not allocate any IPv4 address for pods.

Regarding NAT64, my understanding is that it works together with DNS64.
DNS64 would convert DNS IPv4 A records to an IPv4-embedded IPv6 address and NAT64 would handle the routing.
In this case, if something tries to access directly an IPv4 address (like the local link ones), it will still fail.

olemarkus: yeah nat64 typically doesn't work without dns64. That is true. So that is again a special case with IMDS. CNI agents are dual-stack (running on host networking after all), so this works just fine on anything but link local and where we only have ipv4 literals. IMDS being both of those is then problematic.

What is preventing you from depending on IRSA at this stage?

hakman: > What is preventing you from depending on IRSA at this stage?

Nothing, just don't want to focus on any issues that IRSA would bring in this combination.

hakman:
Replaced by https://github.com/kubernetes/test-infra/pull/22639

PR #11852, opened 2021-06-24
Handle containerExec hooks when using containerd

hakman:
Fixes: #11726


olemarkus (approved):
Is this one worth backporting? It looks like something that could (should already have) wreaked some havoc.

hakman:
> Is this one worth backporting? It looks like something that could (should already have) wreaked some havoc.

I don't think people migrate so fast to containerd and existing clusters have Docker hardcoded in cluster spec.
My feeling is that should at least be part of kOps 1.21 and that it is simple enough to be safe at this stage.

PR #11853, opened 2021-06-24
Allow overriding the ServiceAccountIssuer for IRSA

johngmyers:
Allows admins to override `Cluster.Spec.KubeAPIServer.ServiceAccountIssuer`, allowing them to use IRSA with an external process for publishing the discovery documents.

This increases the risk of a nil dereference if someone turns on IRSA or other service account issuer features on an old version of Kubernetes where they haven't enabled the feature gate. This fixes one such bug in bootstrapchannelbuilder and adjusts the test to specify a k8s version that does support the feature under test.


johngmyers:
@olemarkus the pull-kops-e2e-aws-irsa job appears to be busted. See #11895.

olemarkus (approved):

PR #11859, opened 2021-06-24
Use ExternalDNS as drop-in for dns-controller

olemarkus:
All the changes needed should now be in externalDNS. The remaining work is mainly being able to define DNS provider in the cluster spec.

This will probably never work for gossip clusters though.


johngmyers:
The external-dns docs/faq.md should be updated to fix "Compatibility mode with Kops DNS Controller is planned to be added in the future."

Currently dns-controller has to watch all the pods in the cluster in order to find the annotations. It's unfortunate we're making dns-controller do the same. Could we migrate to using labels instead?

I think we should deprecate dns-controller's functionality for services and ingresses, in favor of having people use external-dns's API directly.

olemarkus:
I think it would be good to have a discussion on how we can use external-dns without compatibility mode, and what the future role of dns-controller may be (e.g it could become a gossip-only thing).

But alone, this PR will make it easier for us to support other DNS providers, and for other cloud providers to not have to implement DNS provider logic in dns-controller.

johngmyers:
kOps relies on DNS to NodePort pods, which external-dns doesn't support without compatibility mode. I think it would be better for external-dns to implement a native mechanism for NodePort pods using labels than for it to implement a compatibility mode for the dns-controller API.

olemarkus:
I am not sure replacing annotations with labels is the right thing to do. But one could add a label selector to the pod source to limit what it watches. That way we could feed it the node role label (by default) for example.

PR #11865, opened 2021-06-25
Add a note about running update-expected when updating base AMI

MoShitrit:
When updating the AMI version on `stable` channel, there's an additional step required, which is running `hack/update-expected.sh` to prevent the tests from failing.
Figured it's worth adding a note about it in the docs.

image


hakman (approved):
Thanks @MoShitrit!

PR #11867, opened 2021-06-25
Remove version from addons

hakman:
... as discussed and agreed during 2021.06.24 Office Hours.


channels/pkg/api/channel.go
```diff
@@ -72,27 +65,3 @@ type AddonSpec struct {
...
-			if addon.Name != nil {
-				name = *addon.Name
-			}
-
-			_, err := semver.ParseTolerant(*addon.Version)
-			if err != nil {
-				return fmt.Errorf("addon %q has unparseable version %q: %v", name, *addon.Version, err)
-			}
-		}
-		if addon.KubernetesVersion != "" {
```
johngmyers, line 92: Please don't remove this check.

johngmyers (changes requested):

channels/pkg/api/channel.go
hakman: Done.

johngmyers:
Oh, I have a new issue: how are we going to deal with #11776 and the need to replace instead of update for some transitions, such as changing the ports for coredns?

hakman:
> Oh, I have a new issue: how are we going to deal with #11776 and the need to replace instead of update for some transitions, such as changing the ports for coredns?

Maybe based on kOps version?

johngmyers (approved):
Actually, we could do it off of the Id.

PR #11868, opened 2021-06-25
Decrease default values for net.ipv4.tcp_rmem and net.ipv4.tcp_wmem

hakman:
Ref: https://github.com/kubernetes/kops/issues/10206#issuecomment-766852332
We can now reproduce the issue with IPv6: https://testgrid.k8s.io/kops-ipv6#kops-grid-scenario-ipv6-conformance.

### DIFF
--- a/nodeup/pkg/model/sysctls.go
+++ b/nodeup/pkg/model/sysctls.go
@@ -63,6 +63,7 @@ func (b *SysctlBuilder) Build(c *fi.ModelBuilderContext) error {
 			"")
 
 		// See https://github.com/kubernetes/kube-deploy/issues/261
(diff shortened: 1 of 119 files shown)


johngmyers:
I think we should have a comment in  `nodeup/pkg/model/sysctls.go` referencing #10206.

Also, the reference in `pkg/model/bootstrapscript.go` should probably be moved down to `func setSysctls()`

hakman:
> I think we should have a comment in `nodeup/pkg/model/sysctls.go` referencing #10206.
> Also, the reference in `pkg/model/bootstrapscript.go` should probably be moved down to `func setSysctls()`

Added extra comments referencing #10206. Should be better now.

johngmyers (approved):

PR #11870, opened 2021-06-26
Use DualStack API NLB for IPv6

hakman:
/cc @johngmyers @rifelpet

### DIFF
--- a/pkg/model/awsmodel/api_loadbalancer.go
+++ b/pkg/model/awsmodel/api_loadbalancer.go
@@ -183,7 +183,9 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 			Type:          fi.String("network"),
 			IpAddressType: fi.String("ipv4"),
 		}
-		if b.UseIPv6ForAPI() {
+		// DualStack can only be used for public NLB
+		// https://aws.amazon.com/premiumsupport/knowledge-center/elb-configure-with-ipv6
+		if b.UseIPv6ForAPI() && lbSpec.Type == kops.LoadBalancerTypePublic {
 			nlb.IpAddressType = fi.String("dualstack")
 		}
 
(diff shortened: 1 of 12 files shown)


pkg/model/awsmodel/dns.go
```diff
@@ -132,15 +142,25 @@ func (b *DNSModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+			})
+			if b.UseIPv6ForAPI() {
+				c.EnsureTask(&awstasks.DNSName{
+					Name:               fi.String(b.Cluster.Spec.MasterInternalName + "-AAAA"),
+					ResourceName:       fi.String(b.Cluster.Spec.MasterInternalName),
+					Lifecycle:          b.Lifecycle,
+					Zone:               b.LinkToDNSZone(),
+					ResourceType:       fi.String("AAAA"),
+					TargetLoadBalancer: targetLoadBalancer,
+				})
```
johngmyers, line 162: We should be checking the return values of these `EnsureTask` calls.

upup/pkg/fi/cloudup/awstasks/dnsname.go
```diff
@@ -196,7 +198,7 @@ func (e *DNSName) Run(c *fi.Context) error {
 
 func (s *DNSName) CheckChanges(a, e, changes *DNSName) error {
 	if a == nil {
-		if fi.StringValue(e.Name) == "" {
+		if fi.StringValue(e.ResourceName) == "" {
 			return fi.RequiredField("Name")
```
johngmyers, line 202:
```suggestion
			return fi.RequiredField("ResourceName")
```

Do we not also want to make `Name` required?

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -178,9 +178,13 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
 
-			Tags: tags,
-			VPC:  b.LinkToVPC(),
-			Type: fi.String("network"),
+			Tags:          tags,
+			VPC:           b.LinkToVPC(),
+			Type:          fi.String("network"),
+			IpAddressType: fi.String("ipv4"),
+		}
+		if b.UseIPv6ForAPI() {
```
johngmyers, line 186: According to AWS documentation, internal load balancers can't be dualstack.

johngmyers (approved):

PR #11884, opened 2021-06-28
support large/slow downloads

aojea:
current download code was using http.client timeout, that will
interrupt reading of the response body, breaking slow downloads
or very large files.

This patch modified the http client to detect idle downloads, and
fail after 30 seconds of innactivity. It still keeps a global
timeout of 5 minutes.

### DIFF
--- a/upup/pkg/fi/http.go
+++ b/upup/pkg/fi/http.go
@@ -17,8 +17,10 @@ limitations under the License.
 package fi
 
 import (
+	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


hakman:
5 minutes is a little much. Usually when nodes are IPv4 only, it takes 1-2 seconds to download such files. Let's see how this goes.
/test pull-kops-e2e-ipv6-conformance

aojea:
> 5 minutes is a little much. Usually when nodes are IPv4 only, it takes 1-2 seconds to download such files. Let's see how this goes.
> /test pull-kops-e2e-ipv6-conformance

I made it configurable, just choose a random value, the point is that now we can differentiate between idle and slow connections

hakman:
Cool. Thanks. 😄

hakman:
Mostly ok, doing a few more runs.
/test pull-kops-e2e-ipv6-conformance

hakman:
Just perfect, thanks @aojea!
/lgtm
/approve

PR #11895, opened 2021-06-30
WIP trivial change for e2e testing

johngmyers:
/ok-to-test
/test pull-kops-e2e-aws-irsa

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 # kops source root directory (without trailing /)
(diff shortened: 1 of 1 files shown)


olemarkus:
Some really strange errors there. and It is supposed to be identical to 
https://testgrid.k8s.io/kops-misc#kops-grid-scenario-service-account-iam

See https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kops/kops-presubmits.yaml#L575
and https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kops/kops-periodics-misc2.yaml#L275

rifelpet:
looks like that fixed it.

PR #11936, opened 2021-07-06
[Digital Ocean] Remove PrivateNetworking option in droplet since it's deprecated

srikiz:
Per [DO Documetation here](https://docs.digitalocean.com/reference/api/api-reference/#operation/create_droplet) PrivateNetworking option is deprecated and will be eventually replaced by `vpc_uuid`. 
If `private_networking` is not specified, the droplets will be assigned to the account's default VPC, which is the default behavior.

PR to include support for `--vpc` flag will follow soon.

FYI - @timoreimann


timoreimann (approved):
LGTM from DO perspective.

hakman (approved):

PR #11940, opened 2021-07-06
Implement some completion for "kops create cluster"

johngmyers:


olemarkus (approved):

cmd/kops/create_cluster.go
```diff
@@ -741,3 +793,141 @@ func loadSSHPublicKeys(sshPublicKey string) (map[string][]byte, error) {
...
+				"flannel",
+				"canal",
+				"kube-router",
+			)
+
+			if options.CloudProvider == "aws" || options.CloudProvider == "" {
+				completions = append(completions, "amazonvpc", "lyftvpc")
+			}
+
+			if cloudup.AlphaAllowGCE.Enabled() && (options.CloudProvider == "gce" || options.CloudProvider == "") {
```
rifelpet, line 894: I noticed a few odd behaviors with checking the feature flag here.

1. The feature flag needs to be set prior to `kops completion` evaluation rather than at the time of tab completion. I suppose this isn't a big deal since most users would have this particular feature flag set in their default shell environment.
2. With the feature flag set, `kops completion` logs some output that should be silenced:
  ```
  export KOPS_FEATURE_FLAGS=+AlphaAllowGCE
  eval $(./.build/local/kops completion zsh)
  I0718 08:50:04.388774   17741 featureflag.go:167] FeatureFlag "AlphaAllowGCE"=true
  ```

johngmyers: I don't think the logging behavior of `kops completion` changed. The old implementation didn't suppress that output either.

PR #11941, opened 2021-07-06
Use etcd v3.5.0 for Kubernetes 1.22+

hakman:


olemarkus:
Maybe we should do a k8s upgrade test for this PR as well.

hakman:
> Maybe we should do a k8s upgrade test for this PR as well.

@olemarkus I agree, we should try it. Should be pretty much fine, as it's tested in the etcd-manager integration tests.

johngmyers (approved):

Issue #11951, opened 2021-07-08
Terraform target doesn't work with Azure provider (--target terraform --out terraform )


alexivanoff:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
I0708 11:55:36.732986   14963 featureflag.go:165] FeatureFlag "Azure"=true
Version 1.21.0 (git-a5bdc3359e544b314d5695b3ed596829313fc6e3)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Unable to connect to the server: dial tcp 10.7.32.4:443: connect: network is unreachable

**3. What cloud provider are you using?**
Azure

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
kops create cluster --cloud azure --name test-1.k8s.local --zones eastus-1 --networking calico --azure-subscription-id "${AZURE_SUBSCRIPTION_ID}" --azure-tenant-id "${AZURE_TENANT_ID}" --azure-resource-group-name test-rg --azure-route-table-name test-route-table --azure-admin-user admin  --network-cidr 10.0.0.0/16 --target terraform --out terraform/kubernetes

**5. What happened after the commands executed?**
W0708 12:00:02.510285   14992 executor.go:139] error running task "ResourceGroup/test-rg" (9m56s remaining to succeed): could not find Render method on type *azuretasks.ResourceGroup (target *terraform.TerraformTarget)
I0708 12:00:02.510327   14992 executor.go:111] Tasks: 38 done / 51 total; 1 can run
I0708 12:00:02.510371   14992 executor.go:186] Executing task "ResourceGroup/test-rg": *azuretasks.ResourceGroup {"Name":"test-rg","Lifecycle":"Sync","Tags":{"KubernetesCluster":"test-1.k8s.local"},"Shared":true}
W0708 12:00:02.510505   14992 executor.go:139] error running task "ResourceGroup/test-rg" (9m56s remaining to succeed): could not find Render method on type *azuretasks.ResourceGroup (target *terraform.TerraformTarget)
[...]

PR #11953, opened 2021-07-08
Return a clearer error when terraform is used on an unsupported provider

rifelpet:
ref: #11951

How I determined the list of supported providers:

```
$ grep -rl RenderTerraform upup/pkg/fi/cloudup  | xargs -L1 dirname | uniq -c
   3 upup/pkg/fi/cloudup/spotinsttasks
  13 upup/pkg/fi/cloudup/gcetasks
  36 upup/pkg/fi/cloudup/awstasks
  16 upup/pkg/fi/cloudup/alitasks
   1 upup/pkg/fi/cloudup/dotasks
```

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -83,6 +83,9 @@ const (
 	OldestSupportedKubernetesVersion = "1.17.0"
 	// OldestRecommendedKubernetesVersion is the oldest kubernetes version that is not deprecated in kOps.
 	OldestRecommendedKubernetesVersion = "1.19.0"
+)
+
+var (
 	// TerraformCloudProviders is the list of cloud providers with terraform target support
 	TerraformCloudProviders = []kops.CloudProviderID{kops.CloudProviderAWS, kops.CloudProviderGCE, kops.CloudProviderALI}
 )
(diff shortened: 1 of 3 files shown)


johngmyers:
It would be good to extend `main.completeTarget()` to only include `TargetTerraform` when the provider supports it. (and the same for `TargetCloudformation`)

rifelpet:
@johngmyers I added completion for `create cluster` but didn't touch `update cluster` because the cloud provider isn't in UpdateClusterOptions and I wasn't sure how to get access to the ClusterSpec from within the completion function.

johngmyers:
@rifelpet You use `GetClusterForCompletion()` to get access to the cluster spec.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -89,6 +89,8 @@ var (
 	OldestSupportedKubernetesVersion = "1.17.0"
 	// OldestRecommendedKubernetesVersion is the oldest kubernetes version that is not deprecated in Kops
 	OldestRecommendedKubernetesVersion = "1.19.0"
+	// TerraformCloudProviders is the list of cloud providers with terraform target support
+	TerraformCloudProviders = []kops.CloudProviderID{kops.CloudProviderAWS, kops.CloudProviderGCE, kops.CloudProviderALI}
```
johngmyers, line 93: Could use a `sets.String()`

johngmyers (approved):

johngmyers (approved):

Issue #11949, opened 2021-07-07
nodeup fails to start using spotinst


olivierpilotte:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
```
Version 1.21.0 (git-a5bdc3359e544b314d5695b3ed596829313fc6e3)
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
v1.21.2
```

**3. What cloud provider are you using?**
```
AWS / SpotInst
```

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

Using SpotInst to launch instances instead of the AWS auto scaling groups, nodeup runs 'describeAutoscalingInstances'. The instances dont' show up in the result. 

This code assumes there's at least 1 instance in the result: https://github.com/kubernetes/kops/blob/0f9388c4cd109ee71ac5ec01d3e0bfddd214808b/upup/pkg/fi/nodeup/command.go#L725

Simple resolution: check if the result has at least 1 instance.

nodeup logs:
```
nodeup[1619]: I0707 23:23:35.815148    1619 request_logger.go:45] AWS request: autoscaling/DescribeAutoScalingInstances
nodeup[1619]: panic: runtime error: index out of range [0] with length 0
nodeup[1619]: goroutine 1 [running]:
nodeup[1619]: k8s.io/kops/upup/pkg/fi/nodeup.getAWSConfigurationMode(0xc0007eefc0, 0xc000938048, 0x13, 0xc000938060, 0x13)
nodeup[1619]:         upup/pkg/fi/nodeup/command.go:725 +0x412
nodeup[1619]: k8s.io/kops/upup/pkg/fi/nodeup.(*NodeUpCommand).Run(0xc00003caf0, 0x43480c0, 0xc000130008, 0x0, 0x0)
nodeup[1619]:         upup/pkg/fi/nodeup/command.go:268 +0x4376
nodeup[1619]: main.main()
nodeup[1619]:         cmd/nodeup/main.go:117 +0x1012
systemd[1]: kops-configuration.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
systemd[1]: kops-configuration.service: Failed with result 'exit-code'.
systemd[1]: Failed to start Run kOps bootstrap (nodeup).
```

olivierpilotte:
This is the PR that introduced the bug: https://github.com/kubernetes/kops/pull/11216

[...]

PR #11958, opened 2021-07-08
check if the instance is under an asg

olivierpilotte:
Nodeup crash when an instance is not created by an AWS Auto-Scaling Group.

Check if the instance is under an ASG before trying to get the lifeCycle value from the ASG.

Related bug: https://github.com/kubernetes/kops/issues/11949

### DIFF
--- a/upup/pkg/fi/nodeup/command.go
+++ b/upup/pkg/fi/nodeup/command.go
@@ -762,6 +762,10 @@ func getAWSConfigurationMode(c *model.NodeupModelContext) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("error describing instances: %v", err)
 	}
+	// If the instance is not a part of an ASG, it won't be in a warm pool either.
+	if len(result.AutoScalingInstances) < 1 {
+		return "", nil
+	}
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/nodeup/command.go
```diff
@@ -762,6 +762,10 @@ func getAWSConfigurationMode(c *model.NodeupModelContext) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("error describing instances: %v", err)
 	}
+	if len(result.AutoScalingInstances) < 1 {
+		klog.Info("instance not under any ASG")
+		return "", nil
+	}
```
olemarkus, line 768:
```suggestion
	// If the instance is not a part of an ASG, it won't be in a warm pool either.
	if len(result.AutoScalingInstances) < 1 {
		return "", nil
	}
```

olemarkus (changes requested):
Thanks for this. Just a small nit.

/ok-to-test

hakman:
@olivierpilotte Please also squash the commits.

olemarkus (approved):

PR #11961, opened 2021-07-08
Cilium etcd fixes

### LINKED ISSUE #10663: Cilium-dedicated etcd is reaching space quota limit without AUTO_COMPACTION
 this information.**

```
1.18.2 (git-84495481e4)
```

 a `kops` flag.**

```json
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"archive", BuildDate:"2020-11-25T13:19:56Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.14", GitCommit:"89182bdd065fbcaffefec691908a739d161efc03", GitTreeState:"clean", BuildDate:"2020-12-18T12:02:35Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

```
AWS
```

Create a new cluster using cilium dedicated etcd created with etcd-manager and wait for a few days.

Without AUTO_COMPACTION options configured, after a few days, the etcd cluster reaches its space quota limit of 2Gb, and Cilium stops working with the following message:
```
# snippet of cilium pod log
level=fatal msg="Unable to connect to kvstore" error="etcdserver: mvcc: database space exceeded" module=etcd subsys=kvstore
```

The etcd cluster reports the following alert:
```
memberID:2157140721128943973 alarm:NOSPACE 
memberID:17459609605570688463 alarm:NOSPACE
```

If I edit my cluster and add the following environment variables to the manager configuration, it starts working as expected
```
# snippet cluster.yaml spec for etcdClusters item.
  - name: cilium
    version: 3.3.10
    manager:
      env:
      - name: ETCD_AUTO_COMPACTION_MODE
        value: revision
      - name: ETCD_AUTO_COMPACTION_RETENTION
        value: "1000"
```

Then I could see this cluster reporting:
```
2021-01-26 19:40:54.223524 I | pkg/flags: recognized and used environment variable ETCD_AUTO_COMPACTION_MODE=revision
2021-01-26 19:40:54.223530 I | pkg/flags: recognized and used environment variable ETCD_AUTO_COMPACTION_RETENTION=1000
```

It's likely to be a default option for all etcd clusters, or at least kops should have a section about this in the documentation, especially for Cilium configuration.

olemarkus:
First commit fixes an issue where the status object of the CNP becomes too large to write to etcd. Also enables "scalability"
Second fixes #10663

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -1173,6 +1173,19 @@ func createEtcdCluster(etcdCluster string, masters []*api.InstanceGroup, encrypt
 		m.InstanceGroup = fi.String(ig.ObjectMeta.Name)
 		etcd.Members = append(etcd.Members, m)
 	}
+
+	// Cilium etcd server is not compacted by the k8s API server.
+	if etcd.Name == "cilium" {
+		if etcd.Manager == nil {
+			etcd.Manager = &api.EtcdManagerSpec{
+				Env: []api.EnvVar{
+					{Name: "ETCD_AUTO_COMPACTION_MODE", Value: "revision"},
+					{Name: "ETCD_AUTO_COMPACTION_RETENTION", Value: "2500"},
+				},
+			}
+		}
+	}
+
 	return etcd
 
 }
--- a/docs/networking/cilium.md
+++ b/docs/networking/cilium.md
@@ -44,6 +44,7 @@ kops create cluster \
 
 For existing clusters, add the following to `spec.etcdClusters`:
 Make sure `instanceGroup` match the other etcd clusters.
+You should also enable auto compaction.
 
 ```yaml
   - etcdMembers:
@@ -53,6 +54,12 @@ Make sure `instanceGroup` match the other etcd clusters.
       name: b
     - instanceGroup: master-az-1c
       name: c
+    manager:
+      env:
+      - name: ETCD_AUTO_COMPACTION_MODE
+        value: revision
+      - name: ETCD_AUTO_COMPACTION_RETENTION
+        value: "2500"
     name: cilium
 ```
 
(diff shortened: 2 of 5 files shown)


upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1173,6 +1173,19 @@ func createEtcdCluster(etcdCluster string, masters []*api.InstanceGroup, encrypt
...
 		etcd.Members = append(etcd.Members, m)
 	}
+
+	// Cilium etcd server is not compacted by the k8s API server.
+	if etcd.Name == "cilium" {
+		if etcd.Manager == nil {
+			etcd.Manager = &api.EtcdManagerSpec{
+				Env: []api.EnvVar{
+					{Name: "ETCD_AUTO_COMPACTION_MODE", Value: "revision"},
+					{Name: "ETCD_AUTO_COMPACTION_RETENTION", Value: "2500"},
```
johngmyers, line 1183: Perhaps we should enable this for existing clusters?

olemarkus: The way this is managed with env vars makes it a bit tricky. There is no way of saying "I want to manage this myself".

We could maybe highlight this more in release notes.

johngmyers: We could take any value explicitly mentioned in the spec as overriding. That leaves the problem of "I want to omit this environment variable", but that has some possible solutions.

olemarkus: Perhaps merge this one and cherry-pick to 1.21 and then have a longer ponder on how to set these by default for existing clusters.

johngmyers: The disadvantage of that would be that clusters created in the interim would get these values set in the cluster spec, making it harder to change the defaults later. But perhaps it's an obscure enough feature.

olemarkus: These defaults should be fine (same as cilium-etcd-operator used before it got deprecated). For clusters that needs something else, their cluster operators should know what to do.

johngmyers (approved):

eddycharly:
Isn't periodic defragmentation necessary too ?

olemarkus:
Drefrag will just reclaim disk space. Since we use dedicated disks for etcd, it probably is not worth the cost of freezing the etcd node regularly.

PR #11964, opened 2021-07-09
Refactor keysets for etcd-manager

johngmyers:
With this I was able to gracefully rotate the etcd-manager CA certs.

Fixes a couple of bugs, most notably that CA rotation requires the new CA to have a different Subject than the old one. So I have `kops create keypair` put the serial number of the new CA into its Subject.


nodeup/pkg/model/context.go
```diff
@@ -408,6 +408,15 @@ func (c *NodeupModelContext) KubectlPath() string {
 
 // BuildCertificatePairTask creates the tasks to create the certificate and private key files.
 func (c *NodeupModelContext) BuildCertificatePairTask(ctx *fi.ModelBuilderContext, name, path, filename string, owner *string, beforeServices []string) error {
+	return c.buildCertificatePairTask(ctx, name, path, filename, owner, beforeServices, true)
+}
+
+// BuildPrivateKeyTask creates the tasks to create the certificate and private key files.
```
olemarkus, line 414: This one builds only the private key, doesn't it?
```suggestion
// BuildPrivateKeyTask builds a task to create the private key file.
```

olemarkus (changes requested):
Just a small nit in one of the comments

olemarkus (approved):

PR #11966, opened 2021-07-09
Unconditionally reenable KMS and Volume Limit tests

rifelpet:
Fixes for these tests have now been released in all applicable k8s version markers

* Invalid KMS Key: https://github.com/kubernetes/kubernetes/pull/101956
  GH doesn't mention it on the commit but v1.21.2 includes the commit [here](https://github.com/kubernetes/kubernetes/blob/v1.21.2/test/e2e/storage/volume_provisioning.go#L828). Theoretically this would fail if someone specified k8s v1.21.0 or v1.21.1 but typically we only test with the latest patch version.
* Volume Limits: https://github.com/kubernetes/kubernetes/pull/103074
  GH shows it as released in v1.22.0-beta.1: https://github.com/kubernetes/kubernetes/commit/164ce31e7fbe04a4a7509cf7aff62c10c5229583

### DIFF
--- a/tests/e2e/pkg/tester/skip_regex.go
+++ b/tests/e2e/pkg/tester/skip_regex.go
@@ -18,7 +18,6 @@ package tester
 
 import (
 	"regexp"
-	"strings"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


rifelpet:
The default [e2e presubmit job uses k8s 1.21](https://github.com/kubernetes/test-infra/blob/263f2a6be4292c319e4b3becc6e59a2277cac4b2/config/jobs/kubernetes/kops/build_jobs.py#L825-L827) so it will confirm the KMS Key fix.

This will test the volume limits fix:
/test pull-kops-e2e-k8s-ci

olemarkus (approved):
Thanks for this.

PR #11986, opened 2021-07-13
Move containerd config from cloudup to nodeup

olemarkus:
Prepare for appending nvidia runtime configuration to containerd + potential support for registry authentication as secrets.


pkg/apis/nodeup/config.go
```diff
@@ -72,7 +72,7 @@ type Config struct {
 	// Hooks are for custom actions, for example on first installation.
 	Hooks [][]kops.HookSpec
 	// ContainerdConfig config holds the configuration for containerd
-	ContainerdConfig string `json:"containerdConfig,omitempty"`
+	ContainerdOverrideConfig *string `json:"containerdConfig,omitempty"`
```
hakman, line 75: I want to add `ContainerdConfigBase` so that people can merge some custom config with what kOps builds.
```suggestion
	ContainerdConfigOverride *string `json:"containerdConfigOverride,omitempty"`
```

hakman (changes requested):
Other than this, looks good to me.

nodeup/pkg/model/containerd.go
```diff
@@ -419,3 +426,34 @@ func (b *ContainerdBuilder) buildCNIConfigTemplateFile(c *fi.ModelBuilderContext
...
+	cluster := b.Cluster
+
+	if cluster.Spec.ContainerRuntime != "containerd" {
+		return ""
+	}
+
+	containerd := cluster.Spec.Containerd
+	if fi.StringValue(containerd.ConfigOverride) != "" {
+		return *cluster.Spec.Containerd.ConfigOverride
+	}
```
johngmyers, line 440:
```suggestion
```

nodeup/pkg/model/containerd.go
```diff
@@ -419,3 +426,34 @@ func (b *ContainerdBuilder) buildCNIConfigTemplateFile(c *fi.ModelBuilderContext
...
 }
+
+func (b *ContainerdBuilder) buildContainerdConfig() string {
+	cluster := b.Cluster
+
+	if cluster.Spec.ContainerRuntime != "containerd" {
+		return ""
+	}
+
+	containerd := cluster.Spec.Containerd
```
johngmyers, line 437: It would be better to pass necessary values, such as `RegistryMirrors` in the `NodeupConfig` than refer to the cluster spec. I'm trying to remove references to the cluster spec from nodeup as changes to cluster spec don't mark nodes as NeedUpdate.

johngmyers (changes requested):

nodeup/pkg/model/containerd.go
olemarkus: I just passed in the entire struct. All of them should trigger NeedUpdate.

nodeup/pkg/model/tests/containerdbuilder/from_docker_19.03.11/tasks.yaml
```diff
@@ -2,7 +2,7 @@ contents: ""
 path: /etc/containerd/config-kops.toml
 type: file
 ---
-contents: CONTAINERD_OPTS=--log-level=info
+contents: CONTAINERD_OPTS=
```
olemarkus, line 5: This diff is unexpected to me.

hakman (approved):
This should unblock GPU and other CRI related changes.

johngmyers:
`b.Cluster` does not come from `NodeupConfig` or anything else deriving from userdata. Changes to it do not cause NeedsUpdate.

PR #11987, opened 2021-07-14
Change set and unset commands into flags on "kops edit cluster"

johngmyers:


hakman:
Good idea, but once this merges, many tests will have to be updated, right?

johngmyers:
Will they? I thought most tests use the `--set` flag on `kops create cluster`

olemarkus:
Some of the scenario tests use it. But you can fix those in this PR

hakman:
Ah, right. So...
/lgtm
/approve

tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -52,7 +52,7 @@ KOPS_B=$(kops-download-from-base)
 
 KOPS="${KOPS_B}"
 
-"${KOPS_B}" set cluster "${CLUSTER_NAME}" "cluster.spec.kubernetesVersion=${K8S_VERSION_B}"
+"${KOPS_B}" edit cluster "${CLUSTER_NAME}" "--set=cluster.spec.kubernetesVersion=${K8S_VERSION_B}"
```
rifelpet, line 55: We use the master branch's scenario scripts for testing all kops versions, so this change is causing failures in kops 1.21 jobs:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-k120-ko120-to-k121-ko121/1416173182599041024

johngmyers: As I see it, we have three alternatives:

* Add "kops set" back in as a hidden command until we drop 1.21 as a destination for upgrade tests
* Make the script change the command based on the version of the destination kops
* Use the release branch's scenario script

PR #11989, opened 2021-07-14
Clean up extra spaces

jayonlau:
Clean up extra spaces, although these errors are not important, they affect the code specification.


olemarkus:
/ok-to-test

Thanks for this. Bear in mind that these addons are deprecated and may stop working on newer k8s versions.

olemarkus (approved):

PR #11991, opened 2021-07-14
Dedicated function for ccm permissons

olemarkus:
The refactoring work from #11818

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -273,10 +273,7 @@ func NewPolicy(clusterName string) *Policy {
 func (r *NodeRoleAPIServer) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 	p := NewPolicy(b.Cluster.GetClusterName())
 
-	AddMasterEC2Policies(p)
-	addASLifecyclePolicies(p, r.warmPool)
-	addCertIAMPolicies(p)
-	addKMSGenerateRandomPolicies(p)
+	addNodeupPermissions(p, r.warmPool)
 
 	var err error
 	if p, err = b.AddS3Permissions(p); err != nil {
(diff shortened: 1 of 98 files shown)


rifelpet (commented):
A few comments, otherwise looks good to me though it does seem to increase the policy document size substantially so I worry users will hit size limits more frequently with `additionalPolicies`

pkg/model/iam/iam_builder.go
```diff
@@ -759,6 +752,152 @@ func addCalicoSrcDstCheckPermissions(p *Policy) {
...
+		"ec2:CreateTags",
+	)
+}
+
+func AddCCMPermissions(p *Policy, clusterName string, cloudRoutes bool) {
+	resource := stringorslice.Slice([]string{"*"})
+
+	p.unconditionalAction.Insert(
+		"autoscaling:DescribeAutoScalingGroups",
+		"autoscaling:DescribeLaunchConfigurations",
```
johngmyers, line 803: I wonder if we still need this.

olemarkus: I don't think so. I couldn't see CCM use this. It is mentioned here, but this is probably dated https://cloud-provider-aws.sigs.k8s.io/prerequisites.html

olemarkus:
> A few comments, otherwise looks good to me though it does seem to increase the policy document size substantially so I worry users will hit size limits more frequently with `additionalPolicies`

Looking at the strict json golden output, it will increase the size from roughly 5k to 6k. So substantially, but still a way to go before the limit is hit. The added bytes come from improving the cluster isolation, which I think is worth it.

pkg/model/iam/iam_builder.go
```diff
@@ -759,6 +752,149 @@ func addCalicoSrcDstCheckPermissions(p *Policy) {
...
+	p.unconditionalAction.Insert(
+		"ec2:DescribeInstances", // aws.go
+	)
+}
+
+func addEtcdManagerPermissions(p *Policy) {
+	resource := stringorslice.Slice([]string{"*"})
+	p.unconditionalAction.Insert(
+		"ec2:DescribeVolumes", // aws.go
+
```
rifelpet, line 768: nit:
```suggestion
```

otherwise LGTM

rifelpet (approved):

PR #11995, opened 2021-07-15
WIP Implement setting and unsetting map entries

johngmyers:
This uses struct field syntax, like Helm does, rather than array index syntax. A disadvantage would be the inability to use '.' characters in map keys.

I suppose we could support either syntax.


rifelpet:
FWIW helm supports `.` in map keys with a specific escape pattern

`--set nodeSelector."kubernetes\.io/role"=master`

https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set

PR #11998, opened 2021-07-15
Use keypair IDs for non-kops-controller-issued worker node certs

### LINKED ISSUE #8774: Certificate renewal

johngmyers:
And with this, the keypair rotation feature is complete.

I don't have any test infrastructure on cloud providers that don't use kops-controller for bootstrap, so this is untested.

Closes #8774

### DIFF
--- a/docs/operations/rotate-secrets.md
+++ b/docs/operations/rotate-secrets.md
@@ -1,3 +1,4 @@
+
 # How to rotate all secrets / credentials
 
 There are two types of credentials managed by kOps:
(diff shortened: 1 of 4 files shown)


hakman (approved):
🎉 🥳 🎉 🥳 🎉 🥳 🎉

rifelpet:
We can add e2e scenarios for testing rotation on AWS and GCE

PR #12001, opened 2021-07-16
Add keypair rotation test scenario

rifelpet:
I copied the commands out of https://kops.sigs.k8s.io/operations/rotate-secrets/

/cc @johngmyers

### DIFF
--- a/tests/e2e/scenarios/keypair-rotation/run-test.sh
+++ b/tests/e2e/scenarios/keypair-rotation/run-test.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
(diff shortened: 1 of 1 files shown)


johngmyers:
You might want to add something that verifies the exported kubeconfigs in some way. I would suggest that it test the first exported kubeconfig can be used to talk to the cluster both before and after the following rolling update. It should also verify that the second exported kubeconfig has only one cert in the certificate-authority-data.

johngmyers (approved):

rifelpet:
@johngmyers after some iteration, the test is now failing due to there being two CAs in the generated kubeconfig instead of one:

https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-keypair-rotation/1417137546726477824/build-log.txt

```
unexpected number of CA certificates in kubeconfig
```

PR #12019, opened 2021-07-18
Fix certificate bootstrap for non-kops-controller-bootstrap cloud providers

johngmyers:


olemarkus (approved):
This code is getting a bit convoluted ... and the cyclomatic complexity makes it a bit hard to reason about.
In particular, I think we need to have a think about `apply_cluster.go` soon.

johngmyers:
@olemarkus

There are the various model builders which have to decide whether or not to create tasks.

There is `BootstrapScriptBuilder.ResourceNodeUp()` which has to decide which of those tasks to record as dependencies of the BootstrapScript task for the instance group.

Then there is `apply_cluster.go`'s `nodeUpConfigBuilder.BuildConfig()` which has to put the data from those dependencies into the `nodeup.Config` and `nodeup.BootConfig` structs.

Finally, there are the nodeup tasks which consume the data from the `NodeupConfig` and `BootConfig`.

I've tried to push NodeupConfig data that doesn't come from dependent tasks down into `nodeup.NewConfig()`. I've also tried to have `nodeUpConfigBuilder.BuildConfig()` defer to `BootstrapScriptBuilder.ResourceNodeUp()` as to whether or not data should be included for a given instance group by testing whether or not the source task is listed as a dependency.

But yes, I think you've correctly identified `nodeUpConfigBuilder.BuildConfig()` as something we should try to figure out how to move logic out of.

PR #12029, opened 2021-07-20
Assert the correct number of kubeconfig CAs during keypair rotation

rifelpet:
The keypair rotation test is failing because there are two CAs where we expected 1: https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-keypair-rotation/1417258356002787328/build-log.txt 

After reviewing the steps, I think it _should_ have 2 CAs at that point, and then one CA after the distrust command. This updates the scenario to reflect that.


johngmyers:
Yes, after the create and after the promote there should be two CAs.

johngmyers:
The procedure has an `export kubeconfig` after the create step. There is no export after the promote as the kubeconfig isn't supposed to change at that point. The test should verify there are two CAs in the kubeconfig after the create.

johngmyers (approved):

PR #12032, opened 2021-07-21
keypair rotation - export new credentials between promote and distruts

rifelpet:
The test is now failing during the third `rolling-update` (after `distrust`) with unauthorized errors:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-keypair-rotation/1417620742954029056

`I0721 00:38:31.819943    4501 instancegroups.go:513] Cluster did not validate within deadline: error listing nodes: Unauthorized.`

Rereading [the docs](https://kops.sigs.k8s.io/operations/rotate-secrets/#export-and-distribute-new-kubeconfig-admin-credentials), it seems like we _do_ need to export new credentials after promoting the new CA and before distrusting the old CA, that way the credentials will be signed from the new CA.

followup to https://github.com/kubernetes/kops/pull/12029

### DIFF
--- a/tests/e2e/scenarios/keypair-rotation/run-test.sh
+++ b/tests/e2e/scenarios/keypair-rotation/run-test.sh
@@ -40,6 +40,11 @@ ${KOPS} promote keypair all
 ${KOPS} update cluster --yes
 ${KOPS} rolling-update cluster --yes --validate-count=10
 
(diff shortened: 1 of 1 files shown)


johngmyers (approved):
Ah, yes, if you're using admin credentials you need a new kubeconfig after promoting, before the rolling update following distrust.

PR #12035, opened 2021-07-21
Render addons and channel manifest during task execution

olemarkus:


upup/pkg/fi/cloudup/bootstrapchannelbuilder/resources.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		klog.Infof("invalid manifest: %s", string(manifestBytes))
+		return nil, fmt.Errorf("error remapping manifest %s: %v", manifestPath, err)
+	}
+	manifestBytes = remapped
+
+	// Trim whitespace
+	manifestBytes = []byte(strings.TrimSpace(string(manifestBytes)))
+
+	rawManifest := string(manifestBytes)
```
johngmyers, line 63: Could simplify the conversion to `[]byte` and then back to `string`.

upup/pkg/fi/fitasks/managedfile.go
```diff
@@ -35,10 +36,12 @@ type ManagedFile struct {
...
-	Base     *string
-	Location *string
-	Contents fi.Resource
-	Public   *bool
+	Base            *string
+	Location        *string
+	Contents        fi.Resource
+	Public          *bool
+	hash            *string
+	renderedContent *string
```
johngmyers, line 44: This appears to be a redundant copy of `Contents`, converted to `string`. Why is it necessary?

upup/pkg/fi/fitasks/managedfile.go
```diff
@@ -60,13 +63,23 @@ func (e *ManagedFile) Find(c *fi.Context) (*ManagedFile, error) {
 		return nil, err
 	}
 
+	existingDataString := string(existingData)
```
johngmyers, line 66: This appears to be adding the assumption that a `ManagedFile` is reasonably representable in a `string`. I don't think that's the case: it could be pure binary data.

One can take the SHA1 of a `[]byte`, so converting it to a `string` is not necessary.

If we're going to bake knowledge of hashes into `ManagedFile`, we shouldn't be doing SHA1. It should be SHA256.

johngmyers:
I'm very skeptical of implementing memoizing and hashing in `ManagedFile`. It seems to belong more in an implementation of `fi.Resource`.

upup/pkg/fi/fitasks/managedfile.go
olemarkus: As you maybe saw, it is because rendering the resource in this case can be worth caching. But as you also commented, this memoization can/should be done in the resource itself.

upup/pkg/fi/fitasks/managedfile.go
olemarkus: I honestly did not look at the `HashString` method. The point was to not break any tests. If we want to change the hash function to sha256, I am all for, but as another PR.

rifelpet:
I'm wondering if we should revert https://github.com/kubernetes/kops/pull/12023 before we cut another 1.22 pre-release if this PR doesn't land in time.

olemarkus:
My plan is to move the memoization to the resource, but keep the hashing with the managed file.
Then it is a matter of getting the VPC data into the template functions.

I don't know when we plan the next pre-release, but hopefully this can get in within a week.

olemarkus:
For plain AWS, this should now be working as expected.
The tests fail because terraform. I don't know enough about how terraform works in order to get the VPC ID that terraform would produce into the template. I assume it would be something around `VPC.TerraformLink`, but not sure how to make use of it.

olemarkus:
I think the cleanup comes with this PR has merit on its own even if we are far away from being able to use tasks in template functions. Target 1.23?

PR #12045, opened 2021-07-24
Pull sshpubkey subcommands out of secrets

johngmyers:


rifelpet:
This LGTM but I'm curious what the motivation is for this change

johngmyers:
SSH public keys are not secrets. It does not make sense to conflate them under the same subcommand. Doing so leads to a confusing CLI.

rifelpet:
Given that these commands are in stable Kops versions, do we need to give any sort of deprecation notice or support both methods for a period of time?

johngmyers:
The PR includes backwards compatibility interfaces. For example, the block around line 43 of `create_secret.go` maintains a hidden `create secret sshpublickey` command. Similarly, the block around line 96 of `delete_secret.go` maintains a hidden `delete secret sshpublickey` command.

rifelpet (approved):

PR #12049, opened 2021-07-27
Add Option to Specify OpenStack Octavia Provider

iGene:
In newer version of OpenStack, there are multiple octavia provider to
choose from instead of only "octavia" as provider. This commit added a
command line option "os-octavia-provider", enabling user to specify the
octavia provider that will be use to create load balancers.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -1057,8 +1057,17 @@ func completeOpenstackLBSubnet(cmd *cobra.Command, args []string, toComplete str
 }
 
 func completeOpenstackOctaviaProvider(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-	// TODO call into cloud provider to get list of Octavia providers
-	return nil, cobra.ShellCompDirectiveNoFileComp
+	providers := []string{
+		"a10",
+		"amphora",
+		"amphorav2",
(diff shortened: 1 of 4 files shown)


cmd/kops/create_cluster.go
```diff
@@ -1054,6 +1056,11 @@ func completeOpenstackLBSubnet(cmd *cobra.Command, args []string, toComplete str
 	return nil, cobra.ShellCompDirectiveNoFileComp
 }
 
+func completeOpenstackOctaviaProvider(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	// TODO call into cloud provider to get list of Octavia providers
+	return nil, cobra.ShellCompDirectiveNoFileComp
```
johngmyers, line 1061: Is there a static list of common Octavia providers we can offer for completion? It looks like `octavia` is a possible value.

iGene: Added completion based on the list provided in this document.
https://docs.openstack.org/octavia/latest/admin/providers/index.html

`octaiva` is not listed in the document but is the default provider for older version of Octavia, so I've also included it in the list.

johngmyers:
Looks like you need to do a `make gen-cli-docs`

zetaab (approved):
looks valid from my perspective.

PR #12064, opened 2021-07-29
Disallow the use of legacy addons

hakman:
These are no longer maintained and kept around to not break very old versions of kOps.

### DIFF
--- a/addons/README.md
+++ b/addons/README.md
@@ -1,3 +1,5 @@
 ## Addons
 
-Read on [addons](../docs/addons.md)
+**Legacy addons are deprecated and unmaintained!!!**
+
+Use [managed addons](../docs/addons.md) instead.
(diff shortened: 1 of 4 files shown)


johngmyers:
This is more than just deprecating. This is removing.

hakman:
> This is more than just deprecating. This is removing.

Not quite removing, but disallowing the use. They have not been useable for some time.

johngmyers:
1.22 is already deprecating the legacy metrics server and cluster server addons. I wouldn't mind extending that to some or all of the other legacy addons. But this should be announced in release notes and anything further should wait until 1.23 at the earliest.

justinsb:
I agree with both of you (I think): this functionality probably is better done via another mechanism now, but we should do so carefully to avoid breaking anyone that might be using it!

hakman:
I agree to with the feedback.
What do you think about disabling it based on k8s version, starting with k8s 1.23?

olemarkus (approved):

PR #12065, opened 2021-07-29
Update AWS CNI to v1.9.0

MoShitrit:
https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.9.0

To stay aligned with our documentation of ENVs that ship with the original manifest which we vendor from, I've added both newly-introduced ENVs with their original values, and commented-them out so that they will pick up the default values.


johngmyers (approved):

johngmyers:
Do we want to set `ENABLE_PREFIX_DELEGATION`?

MoShitrit:
> Do we want to set `ENABLE_PREFIX_DELEGATION`?

Mmm, do you mean to override the default value of `false`? 
Given that all of these can be set through cluster.spec by the user- do we want to leave the default values as they are and leave it up to the users to decide?

PR #12074, opened 2021-07-30
WIP: add support for waf, wafv2 and shield for AWS loadbalancer controller

zetaab:


olemarkus:
The reason I did not implement this is the IAM permissions required. Need to have a think about how to keep this restricted to the cluster's resources

zetaab:
yeah the problem looks like the IAM for this

seh:
See #13636 for another attempt.

PR #12078, opened 2021-07-30
Set SilenceUsage on root so we don't print usage

justinsb:
As we're removing exitWithError, we can use this option instead.


johngmyers:
The `kops` command currently silences usage here:

https://github.com/kubernetes/kops/blob/224eb66169e6cbf742f48ce08c1f3196d83ba5c5/cmd/kops/root.go#L86-L88

It's a `PersistenPreRun`, so you get usage if it fails argument parsing but suppresses if it fails inside `RunE`.

PR #12079, opened 2021-07-30
GCE: TargetPool should ignore Lifecycle field

justinsb:
It's an internal field, it shouldn't be detected as a change to apply.


johngmyers (approved):
/retest

PR #12081, opened 2021-07-31
GCE: auto-enable compute.googleapis.com service

justinsb:
This enables kOps to work in a new GCP project.


johngmyers:
Should this be a Task? As written, it looks like it makes a change to the cloud even on a dry-run invocation.

justinsb:
> Should this be a Task? As written, it looks like it makes a change to the cloud even on a dry-run invocation.

Until the API is activated we can't even query it, so most Find tasks will fail as well.  So it has to be pretty early, and the alternative is to add handling of this logic to every Find command.

OTOH, this only happens on newly created projects, and it's not unreasonable to require users to just activate the core services before using a new project...

johngmyers:
I think a doc update to state a requirement to enable core services in a new project would be better than making changes on a dryrun. Alternatevely, if you can detect the particular error returned by a disabled service, then the Find routines could handle that as "success, nothing exists yet".

PR #12096, opened 2021-08-04
Update Docker to v20.10.8

hakman:
https://github.com/moby/moby/releases/tag/v20.10.8


upup/pkg/fi/cloudup/containerd.go
```diff
@@ -207,6 +207,7 @@ func findAllContainerdDockerMappings() map[string]string {
 		"1.4.3": "20.10.0",
 		"1.4.4": "20.10.6",
 		"1.4.6": "20.10.7",
+		"1.4.9": "20.10.8",
```
johngmyers, line 210: Should we change the value of `containerdFallbackVersion`?

johngmyers (approved):

upup/pkg/fi/cloudup/containerd.go
hakman: I want to cherry-pick this to 1.21 without changing the fallback.

PR #12104, opened 2021-08-05
Update AWS CCM tags

rifelpet:
WIP pending any action from https://github.com/kubernetes/cloud-provider-aws/issues/244

### DIFF
--- a/upup/models/cloudup/resources/addons/aws-cloud-controller.addons.k8s.io/k8s-1.18.yaml.template
+++ b/upup/models/cloudup/resources/addons/aws-cloud-controller.addons.k8s.io/k8s-1.18.yaml.template
@@ -28,7 +28,7 @@ spec:
       serviceAccountName: aws-cloud-controller-manager
       containers:
       - name: aws-cloud-controller-manager
-        image: {{ if .ExternalCloudControllerManager.Image }}{{ .ExternalCloudControllerManager.Image }}{{ else }}gcr.io/k8s-staging-provider-aws/cloud-controller-manager:{{AWSCCMTag}}{{ end }}
+        image: {{ .ExternalCloudControllerManager.Image }}
         imagePullPolicy: IfNotPresent
         args:
 {{- range $arg := CloudControllerConfigArgv }}
--- a/pkg/model/components/awscloudcontrollermanager.go
+++ b/pkg/model/components/awscloudcontrollermanager.go
@@ -73,5 +73,24 @@ func (b *AWSCloudControllerManagerOptionsBuilder) BuildOptions(o interface{}) er
 		return fmt.Errorf("no networking mode set")
 	}
 
+	if eccm.Image == "" {
+		var tag string
+		switch b.KubernetesVersion.Minor {
+		case 18:
+			tag = "v1.18.0-alpha.1"
+		case 19:
+			tag = "v1.19.0-alpha.1"
+		case 20:
+			tag = "v1.20.0-alpha.0"
+		case 21:
+			tag = "v1.21.0-alpha.0"
+		default:
+			// There is no latest tag, so default to the latest that exists
+			// See https://console.cloud.google.com/gcr/images/k8s-artifacts-prod/us/provider-aws/cloud-controller-manager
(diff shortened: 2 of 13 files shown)


rifelpet:
https://github.com/kubernetes/cloud-provider-aws/pull/267 has the new location for images (including going back to 1.18)

/test ?

upup/models/cloudup/resources/addons/aws-cloud-controller.addons.k8s.io/k8s-1.18.yaml.template
```diff
@@ -28,7 +28,7 @@ spec:
       serviceAccountName: aws-cloud-controller-manager
       containers:
       - name: aws-cloud-controller-manager
-        image: {{ if .ExternalCloudControllerManager.Image }}{{ .ExternalCloudControllerManager.Image }}{{ else }}gcr.io/k8s-staging-provider-aws/cloud-controller-manager:{{AWSCCMTag}}{{ end }}
+        image: {{ if .ExternalCloudControllerManager.Image }}{{ .ExternalCloudControllerManager.Image }}{{ else }}us.gcr.io/k8s-artifacts-prod/provider-aws/cloud-controller-manager:{{AWSCCMTag}}{{ end }}
```
olemarkus, line 31: I think we should move this + the tag logic to `pkg/model/components`. Then we can more easily fall back to latest for newer k8s versions.

Also, should the repository for prod be  `k8s.gcr.io/provider-aws/cloud-controller-manager`?

rifelpet: I'm trying to move it to pkg/model/components but the cluster spec as seen by `AWSCloudControllerManagerOptionsBuilder`:

https://github.com/kubernetes/kops/blob/8dee92b2f32cbb4e147bbca0f763db01701fb1ae/pkg/model/components/awscloudcontrollermanager.go#L39

shows `cloudControllerManager` as empty, even if the actual cluster spec includes the `cloudProvider` subfield.

---

I think a related symptom is that the integration test input includes the CCM struct and fields:

https://github.com/kubernetes/kops/blob/8dee92b2f32cbb4e147bbca0f763db01701fb1ae/tests/integration/update_cluster/many-addons-ccm/in-v1alpha2.yaml#L19-L20

but the completed cluster spec as stored in the state store has an empty CCM struct:

https://github.com/kubernetes/kops/blob/8dee92b2f32cbb4e147bbca0f763db01701fb1ae/tests/integration/update_cluster/many-addons-ccm/data/aws_s3_bucket_object_cluster-completed.spec_content#L21

I don't know why that is happening but I'm guessing it is related. Any help would be appreciated.

olemarkus: If I run `kops get cluster -o yaml --full` I do see:
```
  cloudControllerManager:
    allocateNodeCIDRs: true
    cloudProvider: aws
    clusterCIDR: fd00:10:96::/64
    clusterName: test.kops-dev.srsandbox.io
    configureCloudRoutes: false
```

olemarkus: With some rough testing, this only seems to be happening with the terraform tests, but why ... it doesn't make any sense to me.

rifelpet:
I just pushed what ./hack/update-expected.sh is setting for the integration test files. I don't understand why one manifest has `image: null` and another manifest has`image:` removed altogether.

rifelpet:
The external cloud controller manager data seems to be lost during decoding:

https://github.com/kubernetes/kops/blob/d69106e239b8a71048a512b2c58792c416833458/cmd/kops/create.go#L123

The input yaml has the fields but the outputted object does not.

rifelpet:
@olemarkus it turns out the decoding issues were due to the manifests having `provider:` instead of `cloudProvider:`. The test outputs now look as expected and e2e is passing

olemarkus (approved):

PR #12108, opened 2021-08-05
Support Debian 11 Bullseye

ReillyBrogan:
Add support for Debian 11 ahead of the impending release on August 14th. 

Tested by creating a cluster with the `debian-11-amd64-daily-20210805-725` image and verifying that both the cluster came up and that node logs did not contain any errors. Tested kubenet and cilium networking and both containerd and docker runtimes.

### DIFF
--- a/docs/operations/images.md
+++ b/docs/operations/images.md
@@ -42,6 +42,7 @@ The following table provides the support status for various distros with regards
 | Debian 8 | - | 1.5 | 1.17 | 1.18 |
 | [Debian 9](#debian-9-stretch) | 1.8 | 1.10 | 1.21 | - |
 | [Debian 10](#debian-10-buster) | 1.13 | 1.17 | - | - |
+| [Debian 11](#debian-11-bullseye) | 1.22 | - | - | - |
 | [Flatcar](#flatcar) | 1.15.1 | 1.17 | - | - |
 | [Kope.io](#kopeio) | - | - | 1.18 | - |
 | [RHEL 7](#rhel-7) | - | 1.5 | 1.21 | - |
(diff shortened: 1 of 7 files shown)


ReillyBrogan:
I signed it

johngmyers:
I think this deserves a release note.

Perhaps we should bake in the switch to iptables?

hakman:
> I think this deserves a release note.
For sure :).

> Perhaps we should bake in the switch to iptables?
This is optional, works well without it too. I think most CNI plugins support it these days. It is the default in newer Ubuntu too. Probably the should be rephrased.

ReillyBrogan:
> > I think this deserves a release note.
> 
> For sure :).

I looked at the commit history of the releases directory, I assume PR authors are not usually expected to add those entries to their commits? (I can if that's helpful)



> > Perhaps we should bake in the switch to iptables?
> 
> This is optional, works well without it too. I think most CNI plugins support it these days. It is the default in newer Ubuntu too. Probably the should be rephrased.

Looks like kube-proxy supports it as of 1.17, so with the drop of 1.15 and 1.16 this release all supported versions of Kubernetes have the fix. I took a look at the various CNI plugins and the following all have auto-detection:
- Cilium (and especially when running in kube-proxy replacement mode)
- Weave-net
- Flannel
- kube-router

Calico allows configuration of the mode to be in but does not have auto-detection from what I can find.

I have no idea what the situation is with Canal with only Flannel supporting auto-detection, but I suspect you would need to manually set Calico to the correct value as well.

I was unable to find a PR or even an issue for aws-vpc or lyft-vpc, perhaps it was never an issue with those?

I think the entire section on iptables can probably be deleted from both Buster and Bullseye and perhaps moved to the Calico/Canal docs as it looks like that's the only place it's still relevant.

hakman:
I agree that the iptables section can be removed from Bullseye. Buster implementation has various bugs and performance issues that were never addressed.

Calico has support for NFT and autodetection:
https://github.com/kubernetes/kops/blob/1d42125c6c5e45d3b30d38fe9aff4f4e2bab7560/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template#L4110-L4111

Canal... needs a maintainer :).
Lyft... will be removed soon.

ReillyBrogan:
> I agree that the iptables section can be removed from Bullseye

This is done.

johngmyers (approved):

PR #12109, opened 2021-08-05
Remove UseServiceAccountIAM feature flag and rename feature to UseServiceAccountExternalPermissions

olemarkus:


pkg/featureflag/featureflag.go
```diff
@@ -88,7 +88,7 @@ var (
 	// ClusterAddons activates experimental cluster-addons support
 	ClusterAddons = new("ClusterAddons", Bool(false))
 	// UseServiceAccountIAM controls whether we use pod-level IAM permissions for our system pods and kOps addons.
-	UseServiceAccountIAM = new("UseServiceAccountIAM", Bool(false))
+	UseServiceAccountIAM = new("UseServiceAccountIAM", Bool(true))
```
rifelpet, line 91: grep shows that this isnt used anymore so I think we can remove it as well as from `./tests/e2e/scenarios/lib/common.sh`

pkg/apis/kops/cluster.go
```diff
@@ -304,6 +304,9 @@ type IAMSpec struct {
 	Legacy                 bool    `json:"legacy"`
 	AllowContainerRegistry bool    `json:"allowContainerRegistry,omitempty"`
 	PermissionsBoundary    *string `json:"permissionsBoundary,omitempty"`
+	// UseServiceAccountExternalPermissions determines if managed ServiceAccounts will use external permissions directly.
+	// If this is set to false, ServiceAccounts will assume external permissions from the instance it runs on.
```
johngmyers, line 308:
```suggestion
	// If this is set to false, ServiceAccounts will assume external permissions from the instances they run on.
```

johngmyers (changes requested):

pkg/featureflag/featureflag.go
johngmyers: Unknown feature flags are logged and otherwise ignored, so this should be removed.

olemarkus: Done.

johngmyers (approved):

PR #12112, opened 2021-08-06
Implement support for AWS ipv6 prefixes

olemarkus:
Using this PR you can provision a cluster using AWS prefixes.

Going forward, I expect it to be possible to enable ipv6 prefixes as part of launch templates so kOps doesn't have to do it. Meanwhile I do believe this is k8s installer territory.

I hope that in the future, the IPAM controller can be a part of CCM, but meanwhile it runs fine as part of kops-controller.


olemarkus:
This is the cluster settings I used.
I would expect this to use with any CNI though.
```
  kubeDNS:
    upstreamNameservers:
    - 2620:119:53::53
  kubeControllerManager:
    allocateNodeCIDR: false
    controllers:
      - "*"
      - "-nodeipam"
  kubeProxy:
    enabled: false
  networkCIDR: 172.20.0.0/16
  networking:
    cilium:
      version: v1.10.3
      disableMasquerade: true
      tunnel: disabled
  nonMasqueradeCIDR: fd00:10:96::/64
```

nodeup/pkg/model/sysctls.go
```diff
@@ -148,6 +148,9 @@ func (b *SysctlBuilder) Build(c *fi.ModelBuilderContext) error {
 	sysctls = append(sysctls,
 		"# Prevent docker from changing iptables: https://github.com/kubernetes/kubernetes/issues/40182",
 		"net.ipv4.ip_forward=1",
+		"net.ipv4.ip_forward=1",
```
hakman, line 151: I think these should be added conditionally when IPv6 is enabled. My guess is that they were disabled for security reasons on IPv4 only hosts.
Also, could we get this part as a separate PR, I would like to play with adding prefixes via launch templates?
```suggestion
		"net.ipv6.ip_forward=1",
```

olemarkus: Nothing in this PR is of mergeable quality :) It just shows what I hacked together to get stuff working.

I think a PR with the sysctls can be done though. We can probably change accept_ra=2 on all interfaces rather than just the one.

hakman: That would be great. Thanks.

olemarkus:
Since there were actually interest in getting this one merged, I made the code paths configurable.

The commits are still a bit messy, but feedback on the overall semantics would be appreciated.

/cc @justinsb

olemarkus:
This is the cluster config I use for this PR now.
I have not tested NodePort, but it should probably work.
```
...
  kubeProxy:
    enabled: true
  networking:
    cilium:
      enableNodePort: false
  nonMasqueradeCIDR: fd00:10:96::/64
  podCIDRFromCloud: true
...
```

justinsb:
So I do want to see this, but I'm worried about committing to things (without a feature flag) that we'll just have to deprecate as IPv6 support becomes more mature.

I see the changes in 3 groups:

* Changes to make the ClusterCIDR allocation optional etc ... we need this regardless.
* Changes to request an IPv6 address in nodeup ... can this be done in a central controller?
* Changes to map the IPv6 address in kops-controller .... this likely does belong in the cloud provider.  I think it's in kops-controller as a convenience, but we should also figure out whether it's easier to run e.g. with a fork of aws cloud-controller-manager.


I am going to try this out though and play with it :-)

olemarkus:
> So I do want to see this, but I'm worried about committing to things (without a feature flag) that we'll just have to deprecate as IPv6 support becomes more mature.

This should definitely go in with a feature flag. I can add one. Unless we want to reuse the IPv6 flag. I suppose it could go either way with how we end up supporting ipv6.

>     * Changes to request an IPv6 address in nodeup ... can this be done in a central controller?

On Ubuntu with Cilium, this works out of the box. However, on other distros and with other CNIs we may need to do more work on the instance. This is due to some buggy behavior in systemd and/or AWS depending on how you see it. Cilium does some tricks to work around this buggy behavior.

With some additional work, you will get kubenet working on Ubuntu for about 5 minutes though :) (DHCP will fail to renew after you add the ipv6 prefix due to the change in RA flags). I have a branch with those changes that I'll push as a follow-up.

> 
>     * Changes to map the IPv6 address in kops-controller .... this likely does belong in the cloud provider.  I think it's in kops-controller as a convenience, but we should also figure out whether it's easier to run e.g. with a fork of aws cloud-controller-manager.

I opened  this one some time ago: https://github.com/kubernetes/cloud-provider-aws/issues/246

olemarkus:
This is now behind a feature flag.

/retest

justinsb:
WDYT about having podCIDR be empty instead of introducing podCIDRFromCloud?  I think we can merge anyway (as I can explore in a follow-up PR.

Otherwise just a few nits, but lgtm.  Needs a rebase, so maybe you can look through my nits at the same time :-)

/approve
/lgtm

olemarkus:
@justinsb did you comment the nits? I only see the comment above. Since we need to enable a controller i kops-controller etc we need a flag that shows user's intent. Empty `podCIDR` implicitly meaning cloud IPAM sounds a bit magical to me.

justinsb:
Ooops - I queued them up and forgot to hit send.  I actually can't submit the comments now, so I guess I'll open a follow-up PR or two :-)  It was mostly nits apart from the `podCIDR == ""` question, so this way probably works better anyway!

Issue #11929, opened 2021-07-05
Implement rendering of GSFS managed files in Terraform


johngmyers:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

Implement rendering of managed files in GSFS in Terraform.

/kind feature
/area provider/gcp

**2. Feel free to provide a design supporting your feature request.**

`var _ TerraformPath = &GSPath{}`

PR #12121, opened 2021-08-09
Support GSFS Terraform Managed Files

rifelpet:
This adds support for google storage objects rendered in Terraform.

I also added unit tests for both GS and S3. Ideally we could generate their HCL representations and save them to their own .tf files that could be checked by ./hack/verify-terraform.sh but our terraform target would require significant refactoring to be usable in isolated tests like this. Instead we just marshal the resources to JSON.

closes https://github.com/kubernetes/kops/issues/11929

WIP until I get around to testing a `terraform apply` with this.

### DIFF
--- a/util/pkg/vfs/gsfs.go
+++ b/util/pkg/vfs/gsfs.go
@@ -116,6 +116,15 @@ func (p *GSPath) String() string {
 	return p.Path()
 }
 
+// TerraformProvider returns the provider name and necessary arguments
+func (p *GSPath) TerraformProvider() (*TerraformProvider, error) {
+	provider := &TerraformProvider{
+		Name:      "google",
+		Arguments: map[string]string{}, // GCS doesn't need the project and region specified
+	}
+	return provider, nil
+}
+
 func (p *GSPath) Remove() error {
 	done, err := RetryWithBackoff(gcsWriteBackoff, func() (bool, error) {
 		err := p.client.Objects.Delete(p.bucket, p.key).Do()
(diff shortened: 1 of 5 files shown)


rifelpet:
I'm not seeing a way we can easily ignore the lack of credentials when setting up a GCS client the way that we can with an S3 client.

https://github.com/kubernetes/kops/blob/b330f368fd67c8c133a2eb5637149c0324c4216f/util/pkg/vfs/context.go#L402-L405

johngmyers:
/lgtm
but don't know GSFS.

util/pkg/vfs/gsfs.go
```diff
@@ -33,6 +34,7 @@ import (
 	storage "google.golang.org/api/storage/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
+	"k8s.io/kops/upup/pkg/fi/cloudup/terraformWriter"
```
justinsb, line 37: Was the mixed-case deliberate here?  It feels like the sort of thing which might cause problems, so I would probably have used simply `terraformwriter`

util/pkg/vfs/gsfs.go
```diff
@@ -374,6 +386,59 @@ func (p *GSPath) Hash(a hashing.HashAlgorithm) (*hashing.Hash, error) {
...
+	Bucket     string                   `json:"bucket" cty:"bucket"`
+	Object     *terraformWriter.Literal `json:"object" cty:"object"`
+	RoleEntity []string                 `json:"role_entity" cty:"role_entity"`
+	Provider   *terraformWriter.Literal `json:"provider,omitempty" cty:"provider"`
+}
+
+func (p *GSPath) RenderTerraform(w *terraformWriter.TerraformWriter, name string, data io.Reader, acl ACL) error {
+	bytes, err := ioutil.ReadAll(data)
+	if err != nil {
+		return fmt.Errorf("reading data: %v", err)
```
justinsb, line 406: Nit (no need to fix): I've started using %w in new code.  I realize this might be copied-and-pasted from somewhere though!

Eventually I suspect we'll just bulk-replace, so I don't think it's worth fixing here.

util/pkg/vfs/tests/gsfs_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"k8s.io/kops/util/pkg/vfs"
+)
+
+var credsFile = "./mock_gcp_credentials.json"
+
+func TestGSRenderTerraform(t *testing.T) {
+	creds, err := filepath.Abs(credsFile)
+	if err != nil {
+		t.Fatalf("failed to prepare mock gcp credentials: %v", err)
+		t.FailNow()
```
justinsb, line 39: Nit: I don't think you need FailNow with Fatalf.  Fatalf == Logf + FailNow.

https://pkg.go.dev/testing#B.Fatalf

util/pkg/vfs/tests/gsfs_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	t.Setenv("GOOGLE_APPLICATION_CREDENTIALS", creds)
```
justinsb, line 41: Yes, it's not great, but I've seen it done in other tests and the alternative is that we wire in our own authentication token that bypasses FindDefaultCredentials.  So I think this is correct!

https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentialsWithParams

util/pkg/vfs/gsfs.go
rifelpet: The package itself is mixed case. we could certainly modify that in a followup PR.

https://github.com/kubernetes/kops/tree/master/upup/pkg/fi/cloudup/terraformWriter

justinsb:
Thanks for doing this, lgtm.  I agree the auth bypass is not ideal, but I think it's our best choice.

/approve
/lgtm

PR #12122, opened 2021-08-09
Openstack: add port metadata

mitch000001:
This adds some more tags to the Openstack ports. 
Specifically the previous code added the cluster name as an unkeyed tag. Now the cluster name gets added with a named tag, following the server metadata:
The cluster name is tagged by adding `KubernetesCluster=cluster-name` as a tag. Also, the instance group name and the stripped port name (without instance group and cluster name).

This is a preparation of enabling rolling update with surge (detaching instances) for Openstack. The goal is to be able to remove the port name which also embeds the order of the port within the instance group and to replace it with a hashed name, similar to how the servers are created.


zetaab:
@mitch000001 when we are thinking to recreate ports in rolling update. We need also check that internal certificates will work after that (in k8s). Also another problem is that bastion host in that case will be changing floatingip all the time. So we need take care that old one is deleted correctly.

I will check content of this pr tomorrow. Anyways I like the idea of detaching instances

mitch000001:
Well, that PR is only attaching useful information to the ports for later use. I haven't looked into the issues you mentioned but thanks for the heads-up 🙂

johngmyers:
Consider using `kubernetes.io/cluster/NAME: owned` instead of legacy tag names.

zetaab:
@johngmyers the problem with openstack is that all metadata does not allow / chars and also the maximum length is quite small. I do not know what is the case with ports

zetaab:
for me this looks valid. However, @mitch000001 could you check comment https://github.com/kubernetes/kops/pull/12122#issuecomment-895445496 and test can we use those tags? It might be that we cannot

mitch000001:
So, I guess it could be possible to add that label. The problem is more that those things are really tags, so we'd use the artificial `key=value`. 
I can try out if slashes are valid characters for tags and add them if possible.

mitch000001:
Having said my earlier comment the main problem with tags is that they have a length limit, I think it's 64 characters, so the proposed label could be slightly too long given that it contains both key and value.

zetaab (approved):
/approve
/lgtm

PR #12123, opened 2021-08-09
Fix disabling unattended upgrades 

olemarkus:
Current default AMIs pre-install and pre-configure unattended upgrades. We therefor need to explicitly disable it if the update policy requires it.


nodeup/pkg/model/update_service.go
```diff
@@ -85,23 +85,27 @@ func (b *UpdateServiceBuilder) buildFlatcarSystemdService(c *fi.ModelBuilderCont
 }
 
 func (b *UpdateServiceBuilder) buildDebianPackage(c *fi.ModelBuilderContext) {
-	if b.NodeupConfig.UpdatePolicy != kops.UpdatePolicyExternal {
+	contents := ""
+	if b.NodeupConfig.UpdatePolicy == kops.UpdatePolicyExternal {
```
hakman, line 89: External means kOps doesn't touch it not that is disabled. We need a new value for disabling explicitly.
Docs need correcting too.

olemarkus: See `pkg/apis/kops/labels.go`:
```
	// UpdatePolicyExternal is a value for ClusterSpec.UpdatePolicy and InstanceGroup.UpdatePolicy indicating that upgrades are done externally, and we should disable automatic upgrades
	UpdatePolicyExternal = "external"
```

It is fairly explicit about the intended behaviour.

johngmyers: `UpdateSericeBuilder.buildFlatcarSystemdService()` already uses `External` to disable explicitly.

hakman: OK, seems I was wrong on this.

hakman (commented):
/lgtm

hakman (approved):

PR #12124, opened 2021-08-09
Fix cases when the VPC doesn't exist yet

mikesplain:
Similarly to the code here, we should catch cases when the VPC doesn't exist yet.  

This is similar to: https://github.com/kubernetes/kops/blob/e7b52981abd9c510106ab4e43a5fa7ae19f80fec/upup/pkg/fi/cloudup/awstasks/vpcamazonipv6cidrblock.go#L44-L48


hakman (approved):
Good catch :)

PR #12126, opened 2021-08-10
Fix cases when the VPC doesn't exist yet for vpccidrblocks in 1.21

mikesplain:
Backported version of https://github.com/kubernetes/kops/pull/12124.


mikesplain:
cc @hakman

hakman (approved):

PR #12130, opened 2021-08-10
Debian 11: python-apt is not available

ReillyBrogan:


hakman (approved):
Thanks @ReillyBrogan, looks like now everything works as expected!

PR #12132, opened 2021-08-11
Automated cherry pick of #12108: Support Debian 11 Bullseye
#12130: Debian 11: python-apt is not available

hakman:
Cherry pick of #12108 #12130 on release-1.21.

#12108: Support Debian 11 Bullseye
#12130: Debian 11: python-apt is not available

### DIFF
--- a/nodeup/pkg/model/miscutils.go
+++ b/nodeup/pkg/model/miscutils.go
@@ -51,7 +51,7 @@ func (b *MiscUtilsBuilder) Build(c *fi.ModelBuilderContext) error {
 			packages = append(packages, "apt-transport-https")
 
 			// TODO: Do we really need python-apt?
-			if b.Distribution.IsUbuntu() && b.Distribution.Version() >= 20.10 {
+			if (b.Distribution.IsUbuntu() && b.Distribution.Version() >= 20.10) || (!b.Distribution.IsUbuntu() && b.Distribution.Version() >= 11) {
 				// python-apt not available (though python3-apt is)
 			} else {
 				packages = append(packages, "python-apt")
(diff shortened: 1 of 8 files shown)


hakman:
Seems that everything works well on Debian 11 without any extra changes. Would be a god idea to have it available sooner for those that want to experiment with it.
/test pull-kops-e2e-k8s-debian11
/hold

johngmyers:
This is a feature, so not necessarily appropriate for 1.21.

hakman:
This is not exactly a feature. It is more a maintenance thing. We regularly back-ported new OS releases support in the past.
Makes sense (for me) to let operators experiment with Debian 11 and not have to wait for another 1-2 months.
That being said, this is why this is waiting for Office Hours, to decide if we should or we should not.

nodeup/pkg/model/miscutils.go
```diff
@@ -51,7 +51,7 @@ func (b *MiscUtilsBuilder) Build(c *fi.ModelBuilderContext) error {
 			packages = append(packages, "apt-transport-https")
 
 			// TODO: Do we really need python-apt?
-			if b.Distribution.IsUbuntu() && b.Distribution.Version() >= 20.10 {
+			if (b.Distribution.IsUbuntu() && b.Distribution.Version() >= 20.10) || (!b.Distribution.IsUbuntu() && b.Distribution.Version() >= 11) {
```
justinsb, line 54: We probably should have `IsDebian` rather than `!IsUbuntu`, just because it's more self-documenting, but ... that's just a nit.

hakman: Yes, probably we should look to add this in a future PR.

justinsb:
Given that it's something that users will hit and that we do aim to support debian, I'm going to approve this one.  The deciding factor for me was really looking at the code and it is very additive & low-risk; it really shouldn't be triggered except on debian11, where the current behaviour is that we just fail.

@hakman and I chatted and we agreed that we should try to support distros earlier to avoid having to make these calls in future.

/approve
/lgtm

PR #12138, opened 2021-08-12
Always enable IPv4 forwarding

hakman:
... event if cluster is "IPv6 only", IPv6 only is just at k8s level.

/cc @olemarkus

### DIFF
--- a/nodeup/pkg/model/sysctls.go
+++ b/nodeup/pkg/model/sysctls.go
@@ -145,15 +145,14 @@ func (b *SysctlBuilder) Build(c *fi.ModelBuilderContext) error {
 			"")
 	}
 
+	sysctls = append(sysctls,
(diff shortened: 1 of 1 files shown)


hakman:
> But if the k8s networks are v6 only, why would it need to _forward_ IPv4?

 Because I don't want JUST IPv6. Most setups would include some sort of IPv4. Also, there is no harm in allowing IPv4.

olemarkus:
It would help to know what breaks if we leave ipv4 forwarding off. If the cluster is doing ipv6, what is being forwarded using ipv4?

johngmyers:
It would be worth verifying that host network pods can reach external IPv4 endpoints. That would be the most likely thing to break if Docker requires the host to be an IPv4 router.

But if that works we shouldn't be unnecessarily enabling IPv4 router behavior.

PR #12141, opened 2021-08-12
Bump cilium to 1.10.3

olemarkus:


tests/integration/update_cluster/minimal-warmpool/data/aws_s3_bucket_object_minimal-warmpool.example.com-addons-networking.cilium.io-k8s-1.16_content
```diff
@@ -528,6 +532,10 @@ spec:
           path: /opt/cni/bin
           type: DirectoryOrCreate
         name: cni-path
+      - hostPath:
+          path: /sys/fs/cgroup/unified
+          type: DirectoryOrCreate
+        name: cilium-cgroup
```
rifelpet, line 538: @olemarkus this change caused all cilium + rhel8 jobs to fail

https://testgrid.k8s.io/kops-grid#kops-grid-cilium-rhel8-k22-containerd

[kubelet logs](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-grid-cilium-rhel8-k22-containerd/1438294568381452288/artifacts/ip-172-20-39-12.us-east-2.compute.internal/kubelet.log) report:

`E0916 00:19:04.290207   18754 nestedpendingoperations.go:301] Operation for "{volumeName:kubernetes.io/host-path/23431797-5c17-4404-b61b-33d2e8ec3d54-cilium-cgroup podName:23431797-5c17-4404-b61b-33d2e8ec3d54 nodeName:}" failed. No retries permitted until 2021-09-16 00:19:36.290185341 +0000 UTC m=+151.386939478 (durationBeforeRetry 32s). Error: MountVolume.SetUp failed for volume "cilium-cgroup" (UniqueName: "kubernetes.io/host-path/23431797-5c17-4404-b61b-33d2e8ec3d54-cilium-cgroup") pod "cilium-8rdpr" (UID: "23431797-5c17-4404-b61b-33d2e8ec3d54") : mkdir /sys/fs/cgroup/unified: read-only file system`

I dont see any reference to /sys/fs/cgroup/unified in the cilium repo, do you know where this value came from? I'm debating if we want to try to fix this or if we just stop testing cilium + rhel8.

/cc @hakman

olemarkus: Cilium's approach to this is to mount cgroup2 on /sys/fs/cilium-cgroup or something like that, and use that.
But "all modern distros" mount cgroup2 /sys/fs/cgroup, so I simplified things by just assuming it exists.

I overestimated the modernness of redhat and AL2 though.
We could mount cgroup2 explicitly somewhere else as well, like we used to with bpffs. Or we could drop support for not-that-modern-distros.

PR #12144, opened 2021-08-13
Update Debian 11 first supported kOps version

hakman:
As https://github.com/kubernetes/kops/pull/12132 was approved, docs need updating.


mikesplain (approved):

ReillyBrogan:
@hakman when you cherry pick this back to 1.21 it would probably be best to re-add the section about iptables for Bullseye given that 1.21 supports versions of k8s that do not have the iptables/nft fix in kine-proxy.

Also, not sure if this is a concern or not but Bullseye uses unified croups v2 by default and I think this is only supported by kubernetes versions >= 1.19. I'm not near a computer right now and won't be for several days so I can't check the kops code to see if that's handled already.

hakman:
@ReillyBrogan This is why it's still experimental :).
iptables NFT is supported since long time ago, was just a problem with Buster that had some old iptables version without some required fixes.

PR #12148, opened 2021-08-13
Reconcile if managedFile is public or not

olemarkus:
This should take care of the `<nil> -> true` on the OIDC files.


johngmyers:
nit: spelled "reconcile".
/lgtm

johngmyers (approved):

Issue #12140, opened 2021-08-12
nodeup ignores system proxy settings


stanssh:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Version 1.21.0 (git-a5bdc3359e544b314d5695b3ed596829313fc6e3)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
v1.20.9

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
kops create -f ./combined.yml ; 
kops update cluster --name qa.domain.com --yes --state s3://app-qa-kops-state

I guess the simplest way to reproduce the problem is to try creating a cluster in a proxy-restricted environment, with kops version 1.21.0

**5. What happened after the commands executed?**
k8s cluster spins up as expected, but the installation process never completes, which is pretty clear from the logs of kops-configuration unit. It _looks_ like the system proxy settings are ignored for some reason, the asset download fails with a time-out,  as is it would either bypass or ignore proxy

**6. What did you expect to happen?**
K8s cluster installation completion; with assets downloaded through proxy-server

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**
<details>
  <summary>Manifest here</summary>

   ```
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2020-01-13T09:36:22Z"
  generation: 1
  name: qa.domain.net
spec:
  DisableSubnetTags: true
  additionalPolicies:
    node: |
      [
        {
          "Effect": "Allow",
          "Action": [
                "autoscaling:DescribeAutoScalingGroups",
                "autoscaling:DescribeAutoScalingInstances",
                "autoscaling:DescribeTags",
                "autoscaling:DescribeLaunchConfigurations",
                "autoscaling:SetDesiredCapacity",
                "autoscaling:TerminateInstanceInAutoScalingGroup",
                "ec2:DescribeLaunchTemplateVersions",
                "ssm:GetParameterHistory",
                "ssm:GetParameters",
                "ssm:GetParameter",
                "ssm:GetParametersByPath",
                "sqs:*"
            ],
          "Resource": ["*"]
        },
        {
          "Effect": "Allow",
          "Action": [
[...]

PR #12150, opened 2021-08-13
leverage proxy env variables

aojea:
The default http transport uses the proxy if the proxy environment variables exist.
However, since we are creating a new transport to define our own timeout, we should set that proxy option.

https://pkg.go.dev/net/http#DefaultTransport

Fixes:  https://github.com/kubernetes/kops/issues/12140


hakman (approved):
Thanks @aojea! :)

stanssh:
Thanks @aojea ! ^^ 👍🏻

PR #12156, opened 2021-08-15
WIP: prune via labels

justinsb:
Support pruning of existing objects as part of addons.

We extend the addon spec to allow explicit specification of the
objects to be pruned, and automatically generate that specification in
the builder.

This gives us a little more flexibility and should let us handle more
edge cases in future.


justinsb:
cc @olemarkus this is the "other" approach to pruning - using labels that you've set up.  Because I figured we might not have perfect labelling, I made channels more "dumb" in that it goes through and prunes based on instructions in the list of addons.

The addon builder can automatically populate the list of prune instructions based on the manifest; I'm thinking we could also support additional kinds that may have been dropped from the manifest, or just always do a core set of kinds (e.g. Deployments, Daemonsets, StatefulSets)

olemarkus:
Given that we use the following two labels I think our labels are sufficent.
```
"app.kubernetes.io/managed-by" = "kops"
"addon.kops.k8s.io/name" = *addon.Name
```

I would prefer channels to be aware of those labels and use them rather than sending labels in the prune spec.

The only reason I can think of for not just using them is if we wanted to prune some objects that were removed from the manifest before those labels existed and we wanted to use another selector. But there is no way for channels to know that kops actually applied these in the first place, so that doesn't seem like something we should do.

justinsb:
So I finally fixed this up as we agreed - sorry it took so long.

I _think_ what we decided was that we would keep the list of prune directives in the manifest (because otherwise it's very difficult to solve pruning) and that we would use the automatically added labels as the label selector, rather than requiring it to be manually specified.  In future we might allow for additional types to be pruned, or for additional object selectors to be pruned, as this will likely avoid some of the problems that have previously required manual intervention.  But we can add those when they are needed.

If my recollection is wrong, please do correct me :-)

/assign @olemarkus

upup/pkg/fi/cloudup/bootstrapchannelbuilder/pruning.go
```diff
@@ -0,0 +1,115 @@
...
+
+		if err := b.addPruneDirectivesForAddon(addon); err != nil {
+			return fmt.Errorf("failed to configure pruning for %s: %w", id, err)
+		}
+	}
+	return nil
+}
+
+func (b *BootstrapChannelBuilder) addPruneDirectivesForAddon(addon *Addon) error {
+	if len(addon.Spec.Selector) == 0 {
```
olemarkus, line 46: We don't actually use Selector below, so this constraint should probably be removed

upup/pkg/fi/cloudup/bootstrapchannelbuilder/pruning.go
```diff
@@ -0,0 +1,115 @@
...
+				namespaces.Insert(namespace)
+			}
+		}
+		if namespaces.Len() != 0 {
+			pruneSpec.Namespaces = namespaces.List()
+		}
+
+		pruneSpec.LabelSelector = selector.String()
+
+		addon.Spec.Prune.Kinds = append(addon.Spec.Prune.Kinds, pruneSpec)
```
olemarkus, line 111: As I understand this code, you are extracting Kinds in the _new_ manifest. So e.g for NTH, which change a DaemonSet to a Deployment (or vice versa), it actually won't prune the superfluous resource.

That's why I wanted a global list of prunable objects. I am also somewhat skeptical of ever pruning CRDs as that can have rather fatal consequences. Helm is goes even further by only installing, but never mutating CRDs.

olemarkus (changes requested):

justinsb:
/retest

Looks like infra flakes

justinsb:
I was looking at `KOPS_BASE_URL="$(curl -s https://storage.googleapis.com/k8s-staging-kops/kops/releases/markers/release-1.21/latest-ci.txt)"` and I assumed that meant it was not using this branch when first bringing up the cluster.  Upgrade is definitely something we need to think about, anyway :-)

In the last test run, the apiserver just failed to come up without really logging a reason (https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/12156/pull-e2e-kops-addon-resource-tracking/1448311356632600576/artifacts/35.180.124.113/kube-apiserver.log).  Going to repeat the test in the hope of figuring out whether that was a one-off or something likely more related to this PR...

/test pull-e2e-kops-addon-resource-tracking

justinsb:
Test looks to be passing but I think the bash test logic was not right.  Made it a little less terse...

/test pull-e2e-kops-addon-resource-tracking

justinsb:
All passing but  test pull-kops-e2e-cni-flannel and pull-kops-e2e-cni-canal which looks to be failures to upload.

/retest

justinsb:
I think this PR is (finally) in good shape.  I had to implement the generation concept to be sure we would reapply the update after an update of the `channels` code (even if the same manifest had been applied by the previous, non-prune-aware, version)

justinsb:
cc @olemarkus

PR #11786, opened 2021-06-16
[cilium] Add support for encryption via WireGuard

dntosas:
In this commit, we enable users to choose WireGuard as their prefered
encryption type, leveraging this new feature from Cilium.

Ref: https://cilium.io/blog/2021/05/20/cilium-110#wireguard

Signed-off-by: dntosas <ntosas@gmail.com>

PR #12158, opened 2021-08-15
Cilium wireguard support

olemarkus:
This is a continuation of #11786 which went inactive.

/cc @dntosas

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -879,8 +879,24 @@ func validateNetworkingCilium(cluster *kops.Cluster, v *kops.CiliumNetworkingSpe
 		allErrs = append(allErrs, IsValidValue(fldPath.Child("bpfLBAlgorithm"), &v.BPFLBAlgorithm, []string{"random", "maglev"})...)
 	}
 
+	// Cilium with Wireguard integration follow-up --> https://github.com/cilium/cilium/issues/15462.
+	// The following rule of validation should be deleted as this combination
+	// will be supported on future releases of Cilium (>= v1.11.0).
+	if v.EncryptionType == "wireguard" && v.EnableEncryption && fi.BoolValue(v.EnableL7Proxy) {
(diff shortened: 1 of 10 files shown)


pkg/apis/kops/validation/validation.go
```diff
@@ -865,8 +869,26 @@ func validateNetworkingCilium(cluster *kops.Cluster, v *kops.CiliumNetworkingSpe
...
+			if fi.BoolValue(v.EnableL7Proxy) {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("encryptionType"), "Cilium EncryptionType=WireGuard cannot work with L7 Proxy enabled."))
+			}
+
+			// Cilium with Wireguard integration follow-up --> https://github.com/cilium/cilium/issues/15462.
+			// The following rule of validation should be deleted as this combination
+			// will be supported on future releases of Cilium (>= v1.11.0).
+			if fi.BoolValue(v.EnableL7Proxy) {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Child("enableL7Proxy"), "L7 proxy cannot be enabled if wireguard is enabled."))
+			}
```
johngmyers, line 886: These two validations are redundant with each other. One should be removed.

johngmyers (changes requested):
The documentation and CLI refer to the "ciliumpassword" secret as the "Cilium IPsec configuration". It looks like the code requires and provisions this for Wireguard as well. The terminology in the documentation/CLI should be updated.

johngmyers:
Ah, if the ciliumpassword secret isn't needed for Wireguard, then kOps shouldn't require it to exist and it shouldn't provision it to the nodes.

johngmyers:
https://github.com/kubernetes/kops/blob/cd1aa1ab53c95b706a5eefdeeb0eb02223a1bddb/upup/pkg/fi/cloudup/apply_cluster.go#L363-L370

Needs to test the encryption type.

johngmyers (approved):

PR #12159, opened 2021-08-15
Incorrect url to the swift container

yadneshk:
Swift path prefix in ReadTree and ReadDir should not be empty
when adding "/" to it.


yadneshk:
closes #11526

zetaab (approved):

PR #12161, opened 2021-08-16
Debian 11: Release AMIs use same AWS Owner ID as Buster

ReillyBrogan:
Now that the release has happened it appears the Debian cloud team is using a different AWS account for release images compared to the daily images.


hakman (approved):
Thanks @ReillyBrogan!

PR #12170, opened 2021-08-17
Initial IPv6 support for GCE

justinsb:
Supporting IPv6 values where they can be set by the user, and ensuring
that IPv4 and IPv6 firewall rules are split because on GCP they cannot
be in the same rule.

### DIFF
--- a/pkg/model/gcemodel/firewall.go
+++ b/pkg/model/gcemodel/firewall.go
@@ -17,6 +17,9 @@ limitations under the License.
 package gcemodel
 
 import (
+	"net"
+	"strings"
+
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/upup/pkg/fi"
@@ -63,15 +66,13 @@ func (b *FirewallModelBuilder) Build(c *fi.ModelBuilderContext) error {
 		// The traffic is not recognized if it's on the overlay network?
 		klog.Warningf("Adding overlay network for X -> node rule - HACK")
 
-		t := &gcetasks.FirewallRule{
-			Name:         s(b.SafeObjectName("cidr-to-node")),
+		b.AddFirewallRulesTasks(c, "cidr-to-node", &gcetasks.FirewallRule{
 			Lifecycle:    b.Lifecycle,
 			Network:      b.LinkToNetwork(),
 			SourceRanges: []string{b.Cluster.Spec.NonMasqueradeCIDR},
 			TargetTags:   []string{b.GCETagForRole(kops.InstanceGroupRoleNode)},
 			Allowed:      []string{"tcp", "udp", "icmp", "esp", "ah", "sctp"},
-		}
-		c.AddTask(t)
+		})
 	}
 
 	// Allow full traffic from master -> master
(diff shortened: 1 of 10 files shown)


pkg/cidr/cidrset.go
```diff
@@ -0,0 +1,89 @@
...
+		ipNets = append(ipNets, *ipNet)
+	}
+	return Set{cidrs: ipNets}, nil
+}
+
+// WhereIPV4 returns a CIDRSet containing only the IPv4 CIDRs
+func (s Set) WhereIPV4() Set {
+	var matching []net.IPNet
+	for i := range s.cidrs {
+		if s.cidrs[i].IP.To4() != nil {
```
johngmyers, line 47: Probably needs to exclude IPv4-mapped IPv6 addresses. See `validation.validateCIDR()`.

pkg/model/gcemodel/firewall.go
```diff
@@ -116,15 +121,37 @@ func (b *FirewallModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+	if len(ipv4s) == 0 {
+		ipv4s = append(ipv4s, "0.0.0.0/32")
+	}
+	return ipv4s
+}
+
+func ipv6SourceRange(cidrs cidr.Set) []string {
+	ipv6s := cidrs.WhereIPV6().ToStrings()
+	if len(ipv6s) == 0 {
+		ipv6s = append(ipv6s, "::/128")
```
johngmyers, line 154: The semantics here seem odd. If you only list IPv4 CIDRs on your SSH access list, it will open SSH to the entire IPv6 Internet?

johngmyers (changes requested):

pkg/model/gcemodel/firewall.go
```diff
@@ -63,11 +64,15 @@ func (b *FirewallModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+		}
 		t := &gcetasks.FirewallRule{
 			Name:         s(b.SafeObjectName("cidr-to-node")),
 			Lifecycle:    b.Lifecycle,
 			Network:      b.LinkToNetwork(),
-			SourceRanges: []string{b.Cluster.Spec.NonMasqueradeCIDR},
+			SourceRanges: ipv4SourceRange(nonMasqueradeCIDR),
 			TargetTags:   []string{b.GCETagForRole(kops.InstanceGroupRoleNode)},
 			Allowed:      []string{"tcp", "udp", "icmp", "esp", "ah", "sctp"},
 		}
```
johngmyers, line 78: Why aren't we building IPv6 firewall rules here?

pkg/model/gcemodel/firewall.go
justinsb: Actually I think I found one!  We can set disabled=true on the rule - I think that's clearer than both the odd IPs and deleting the rules.  (I _think_!)

pkg/model/gcemodel/firewall.go
justinsb: It _was_ because they weren't set by default in `kops create`.  However, now we're using a helper function, we do now build the rules for ipv4 and ipv6.

pkg/cidr/cidrset.go
justinsb: I didn't understand the reference to validateCIDR - was it checking that ip == cidr.IP?

IIUC these IPv4-mapped IPv6 addresses are not publicly routable, and are probably going to be special cases for most networking operations on clouds.  We can add them if people start using them, but I think right now they will be an error, which I think is the correct thing to do until we/I understand better what we should do here.

I could change validateCIDR (and friends) to reject them in the validation phase though?

johngmyers: `validateCIDR` uses `ip.To4() != nil && !strings.Contains(cidr, ":")` to exclude IPv4-mapped IPv6 addresses from the IPv4 code path.

justinsb: Gotcha - makes sense - thanks!  Added!

johngmyers (approved):

PR #11592, opened 2021-05-24
[Digital Ocean] Code cleanup with no functional modifications

srikiz:
- Using interfaces for better unit testing (haven't added yet, but will create that separately)
- Removing cyclic dependency - moved the code from pkg/resources/digitalocean/cloud.go to upup/pkg/fi/cloudup/do/cloud.go.
- No changes to the functionality, but just moving code so it's easier to maintain going forward.

FYI - @timoreimann

PR #12173, opened 2021-08-18
[DigitalOcean] Code refactor to use the existing dnsprovider package

srikiz:
- Move files from `pkg/resources/digitalocean/dns` to `dnsprovider/pkg/dnsprovider/providers/do` so it's consistent with other cloud providers.
- Was seeing issues where protokube was not able to fetch dnsprovider for DO. (it continued to work for gossip cluster, but had issues only for dnsProviderID with DO) Tested with this change and it's working as expected. 
- It looks like it was caused due to my previous change - https://github.com/kubernetes/kops/pull/11592

FYI @timoreimann


hakman (approved):
Thanks @srikiz :).

PR #12175, opened 2021-08-18
Backport moving updatePolicy to nodeup config

ReillyBrogan:
!12127 was an incomplete backport for the issue of updatePolicy not working. If `updatePolicy: external` was set at the cluster level but not the instance group level it would still result in automatic updates being enabled when they should not be. 

FYI @olemarkus


johngmyers:
I believe you need to run `hack/update-expected.sh`

johngmyers:
And I coincidentally ran into this bug today.

/lgtm
/approve

PR #12177, opened 2021-08-19
Hardcode Flatcar containerd exec command

hakman:
As part of https://github.com/kinvolk/coreos-overlay/pull/931 and https://github.com/kinvolk/update_engine/pull/13, Flatcar decided to change the containerd exec command and not allow an easy override of the config file.
https://kinvolk.io/docs/flatcar-container-linux/latest/container-runtimes/customizing-docker/#use-a-custom-containerd-configuration

/cc @rifelpet

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -249,8 +249,10 @@ func (b *ContainerdBuilder) buildSystemdServiceOverrideContainerOS(c *fi.ModelBu
 func (b *ContainerdBuilder) buildSystemdServiceOverrideFlatcar(c *fi.ModelBuilderContext) {
 	lines := []string{
 		"[Service]",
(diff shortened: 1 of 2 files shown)


olemarkus (approved):

jepio:
Hi, just wanted add an observation: the two referenced Flatcar PRs didn't change the containerd command, the previous override method was not successful (`${CONTAINERD_CONFIG}"` was still being looked up in the torcx directory) but containerd-1.4 ignores missing config files and runs with default config, whereas containerd-1.5 errors out in that case.

hakman:
Thanks for the explanation @jepio. Was a bit harder these days to keep track with any Flatcar related issues as the periodic tests keep failing since kOps switched to IMDSv2 by default for AWS. Not sure if anyone is looking into that, but seems like a pretty simple ignition issue (see: https://github.com/flatcar-linux/ignition/pull/32).

PR #12183, opened 2021-08-20
Consolidate statements in control plane IAM role

rifelpet:
This reduces the policy size by ~9%

/cc @olemarkus

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -767,7 +767,6 @@ func addNodeupPermissions(p *Policy, enableHookSupport bool) {
 }
 
 func addEtcdManagerPermissions(p *Policy) {
-	resource := stringorslice.Slice([]string{"*"})
 	p.unconditionalAction.Insert(
 		"ec2:DescribeVolumes", // aws.go
 	)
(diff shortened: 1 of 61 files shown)


pkg/model/iam/iam_builder.go
```diff
@@ -878,25 +875,6 @@ func AddCCMPermissions(p *Policy, cloudRoutes bool) {
...
-				"ec2:CreateSecurityGroup",
-				"ec2:CreateVolume",
-				"elasticloadbalancing:CreateListener",
-				"elasticloadbalancing:CreateTargetGroup",
-			),
-			Resource: resource,
-
-			Condition: Condition{
-				"StringEquals": map[string]string{
-					"aws:RequestTag/KubernetesCluster": p.clusterName,
```
olemarkus, line 896: The plan was to put all the create-like statements into a separate set and apply this condition. Removing the tag-on-create condition is perhaps less than ideal.

olemarkus (changes requested):

olemarkus:
Thanks for picking this one up. I have been meaning to sort out the create stuff too, just that other things got in the way. As mentioned in the comments, I'd like to consolidate those into a separate set that then applies the `requestTag` condition.

olemarkus (approved):

PR #12188, opened 2021-08-21
Add ip addresses into kubelet certs and let kubelet default to ipv6 if applicable

olemarkus:


nodeup/pkg/model/kubelet.go
```diff
@@ -244,6 +246,12 @@ func (b *KubeletBuilder) buildSystemdEnvironmentFile(kubeletConfig *kops.Kubelet
 		flags += " --tls-private-key-file=" + b.PathSrvKubernetes() + "/kubelet-server.key"
 	}
 
+	if b.Cluster.Spec.IsIPv6Only() {
+		flags += " --address=::"
+		// This makes kubelet prefer ipv6. E.g Pods with host network will have ipv6 addresses.
+		flags += " --node-ip=::"
```
hakman, line 252: This doesn't work when `clould-provider=external` and the address I think is set somewhere else in model.
https://github.com/kubernetes/kubernetes/blob/v1.22.1/pkg/kubelet/nodestatus/setters.go#L95

olemarkus: It doesn't affect NodeIP assignment, but this does still affect Pods with HostNetwork. Since we are in a hybrid dual-stack/single-stack mode, these get ipv4 addresses without this, ipv6 addresses with.

hakman: How? There is no code for taking this into consideration with external CCM?

hakman: This only depends on the oder the CCM sends the IP addresses, unless something changed since I tested.

olemarkus: I have largely tested this on 1.21, so I doubt it changed. But there may also be different behavior for static Pods and the timing of when CCM sets the Node IPs. Node objects will exist before CCM modifies them.

I did find that I needed both this flag _and_ the ordering of internal addresses to get the addresses correctly.

Worst case, this one is noop though, so there shouldn't be any harm in keeping it.

hakman: It is a harm because of the confusion it creates and expectations. 1.21 at least with kOps still defaults to KCM.
I suggest to test with 1.22 and have some scenarios that we can also reproduce.

I think what you are observing is somehow correct. The IPv6 only behaviour for dual-stack nodes is not exactly great and some things that should be done by kubelet are not quite well implemented and documented.

If we want to go this way, which is not exactly ideal, we may as well set the NodeIP to the first IPv6 address from the EC2 metadata.

olemarkus: Makes sense.

I removed the commit. The important bit was getting IPs into the kubelet certs.

hakman:
@johngmyers Could you take a look at this PR, please?

johngmyers:
What is the reason for adding the IP addresses to the kubelet-server cert?

The concern I have is that when there are two services with the same IP in their respective SANs, a client using that IP to talk to one can be tricked into talking to the other. There are clients that use IP addresses to talk to kube-apiserver, so they could then be redirected to talk to kubelet instead.

olemarkus:
Sorry, for the lack of explanation. I put it in the commit, but it didn't make it into the PR.

The challenge is that there is no host DNS record for ipv6. So if you have an ipv6-only pod (say metrics-server) that wants to talk to kubelet, it cannot verify the TLS cert. So either such Pods would need host networking or we would need to put IPs in the certificate. I find the latter more safe than the former.

Upstream is also fond of using IPs before hostname. https://github.com/kubernetes-sigs/metrics-server/blob/master/manifests/base/deployment.yaml#L26

PR #12198, opened 2021-08-24
[addons/metrics-server] Bump minor version

dntosas:
Updating image and manifest to latest stable version.


johngmyers:
/ok-to-test

Changing the port is concerning.

olemarkus:
Changing the port of the backend service will make the service blip until the new Pods are ready to serve. I am a bit ambivalent between just following upstream and if this change is really necessary. Upstream doesn't give a particularly good reason for using 443 other than that they couldn't do so before due to missing caps.

olemarkus (approved):
While running `watch kubectl top nodes` during the upgrade, I could not see a single disruption. There was also not a single complaint from the apiserver about not being able to reach metrics-server. So I think we are good.

PR #12207, opened 2021-08-27
Do not set ClusterCIDR for KubeProxy when using CNI networking and kubeProxy.clusterCIDR is not set

dezmodue:


olemarkus:
Wouldn't this approach pretty much break any existing cluster that use `cni: {}`?

A better approach may be to make `clusterCIDR` a pointer and then test for nil vs empty string. Nil would mean current behavior, while you would want to set this to empty string.

dezmodue:
Thanks for your comment and sorry for the late reply.
I understand this would be a breaking change, the intention was to align the behaviour of kops for the specific use case of `cni: {}` where kops doesn't have a lot of information to make a good call on defaults and the user is expected to take full ownership of the chi installation.

I have now modified the PR as suggested, this changes the behaviour of `clusterCIDR: ""` from "get the defaults" to omit the `--cluster-cidr` argument for kube-proxy.

An alternative could be to have a specific value of `clusterCIDR` that has the same effect, something like `clusterCIDR: "omit"`, this wouldn't change in any way the current behaviour and just add an extra option.

pkg/model/components/kubeproxy.go
```diff
@@ -68,8 +68,12 @@ func (b *KubeProxyOptionsBuilder) BuildOptions(o interface{}) error {
 		// If we're not using the AmazonVPC networking, and the KubeControllerMananger has
 		// a ClusterCIDR, use that because most networking plug ins draw pod IPs from this range.
 		if clusterSpec.Networking.AmazonVPC == nil && clusterSpec.KubeControllerManager != nil {
-			config.ClusterCIDR = clusterSpec.KubeControllerManager.ClusterCIDR
+			config.ClusterCIDR = &clusterSpec.KubeControllerManager.ClusterCIDR
 		}
+		// The user can explicitly omit --cluster-cidr from the kube-proxy configuration
+		// by setting spec.kubeProxy.clusterCIDR: ""
+	} else if *config.ClusterCIDR == "" {
+		config.ClusterCIDR = nil
```
johngmyers, line 76: If something like this is necessary, it should be done just before turning into flags, in `KubeProxyBuilder`.

johngmyers:
Please rebase and squash.

johngmyers:
You need to run the `hack/update-expected.sh` script.

pkg/model/components/kubeproxy.go
dezmodue: Not necessary - removed

johngmyers (approved):

Issue #12041, opened 2021-07-23
Make static UIDs in wellknownusers.go configurable


wesselvdv:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

I'd like the capability to change the static UIDs defined in https://github.com/kubernetes/kops/blob/e014fce41de10cf285c083a4e107caa4b75fa8eb/pkg/wellknownusers/wellknownusers.go#L22-L47 to be configurable. I am working in a large organisation that's using `ldap-sssd` to provide ssh access, unfortunately it's not possible for us to change the uid range of sssd, so we're getting conflicts on these static uids.

johngmyers:
I think it's extremely unlikely we'd make this configurable. It would be a significant increase in complexity.

ArchangelX360:
+1, we are having the same issue for a similar reason.

The significant increase in complexity is probably required here.
`kops` shouldn't just "hope" that there will be no conflicts with the users of the nodes.

Is there any workaround you would suggest @johngmyers?

ArchangelX360:
> I don't believe that Debian, for example, lets you change the static UIDs of its system processes

I am not sure I understand. This issue is about making `kops` hardcoded UIDs configurable, especially because some environments where `kops`'s `nodeup.sh` would run will have those UIDs already taken.
Making `kops` hardcoded UID configurable would fix the problem.

---

In both @wesselvdv and my case, it seems that it is related to our company's LDAP having users of those UID.
A workaround I applied on the master nodes was to run the same `useradd` command with a `--non-unique` flag, creating a local user despite the LDAP existing one (I discourage this workaround as it enters the realm of unknown behaviour).

It worked but had to be applied manually, ssh-ing to the master nodes, and would have to be applied every time an EC2 instance is replaced. That is sub-optimal.

To automate this, I thought about using `kops` [`additionalUserData`](https://kops.sigs.k8s.io/instance_groups/#additionaluserdata), like that:
```
additionalUserData:
    # FIXME: this would only work if it was executed after `nodeup.sh`
    - name: useradd_workaround.sh
      type: text/x-shellscript
      content: |
        #!/bin/bash
        set -o errexit
        set -o nounset
        set -o pipefail

        # Context: https://booking.slack.com/archives/C022EK8N1L5/p1629710094173500

        # `kops` assumes those users are not existing and `kops-configuration.service` will fail with both:
        #  - "useradd: UID 10011 is not unique"
[...]

PR #7381, opened 2019-08-06
use iptables legacy mode when using buster

zetaab:
fixes #7379 

There are two ways to do this: nodeup OR insert in imagebuilder. However, at least I would like to have support in all buster images - that is why I implemented it to nodeup. Imagebuilder does not support platforms like openstack.

/assign @justinsb 
/kind bug

PR #12217, opened 2021-08-28
Add possibility to prepend additional user data scripts

ArchangelX360:
Before this patch, AdditionalUserData scripts will only be placed
after `nodeup.sh`.
We needed to have a script running before to workaround issues that are inherent to `nodeup.sh`, such as https://github.com/kubernetes/kops/issues/12041

### DIFF
--- a/k8s/crds/kops.k8s.io_instancegroups.yaml
+++ b/k8s/crds/kops.k8s.io_instancegroups.yaml
@@ -72,6 +72,10 @@ spec:
                 items:
                   description: UserData defines a user-data section
                   properties:
+                    before:
+                      description: Before indicated whether this user data script
+                        should be executed before or after the default user data scripts
+                      type: boolean
                     content:
                       description: Content is the user-data content
                       type: string
--- a/pkg/apis/kops/instancegroup.go
+++ b/pkg/apis/kops/instancegroup.go
@@ -239,6 +239,8 @@ type UserData struct {
 	Type string `json:"type,omitempty"`
 	// Content is the user-data content
 	Content string `json:"content,omitempty"`
+	// Before indicated whether this user data script should be executed before or after the default user data scripts
+	Before bool `json:"before,omitempty"`
 }
 
 // VolumeSpec defined the spec for an additional volume attached to the instance group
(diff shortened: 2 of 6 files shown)


hakman:
@ArchangelX360 Did you try naming your additional scripts with a digit in front? If I remember correctly, they run in alphabetical order.

ArchangelX360:
@hakman I did not try, but that seems very unlikely given the code of `AWSNodeUpTemplate`, `nodeup.sh` will always be always placed first see https://github.com/kubernetes/kops/blob/master/pkg/model/resources/nodeup.go#L192.

Then, AWS executes sequentially the scripts in userdata in the order provided by the operator (I tried moving manually the script before `nodeup.sh` and it worked).

hakman:
Can you try the naming? We had same issue in the past and different names for the scripts helped.
See this for example:
https://github.com/kubernetes/kops/pull/7381#issuecomment-521878286

ArchangelX360:
> Then, AWS executes sequentially the scripts in userdata in the order provided by the operator (I tried moving manually the script before nodeup.sh and it worked).

I was wrong, naming does work! AWS will run then in alphabetical order, e.g. it works if I configure:

```
apiVersion: kops/v1alpha2
kind: InstanceGroup
// ...
spec:
  // ...
  additionalUserData:
    # prefixed a_ to be sure to run before nodeup, script are executed in alphabetical order
    - name: a_useradd_workaround.sh
      type: text/x-shellscript
      content: |
        #!/bin/bash
        set -o errexit
        set -o nounset
        set -o pipefail

        # `kops` assumes those users are not existing and `kops-configuration.service` will fail with both:
        #  - "useradd: UID 10011 is not unique"
        #  - "useradd: UID 10012 is not unique"
        # Those two UIDs are already attached to some users in our setup.
        # So we are forcing a local user creation with `--non-unique`, so that `kops` does not try to create them
        echo "Applying 'kops' UID workaround"
        sudo /usr/sbin/useradd --non-unique -u 10011 -s /sbin/nologin -o kops-controller
        sudo /usr/sbin/useradd --non-unique -u 10012 -s /sbin/nologin -d /etc/kubernetes/kube-apiserver-healthcheck/secrets kube-apiserver-healthcheck
```

Thank you so much @hakman!

PR #12219, opened 2021-08-29
[addons/node-local-dns] Bump version and make image field configurable

dntosas:
As per https://github.com/kubernetes/dns/commit/3b17e06879a46b2ec5d97105c611a315897fdb48, node-local-dns addon is now built with the latest coreDNS base (v1.8) and that brings great consistency between cache and upstream servers in a manner of configuration syntax, metrics name convention, etc.

So in this commit, we bump node-local-dns image to latest v1.20.0 which is  build upon latest coreDNS and also add support for overriding this field.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -2104,12 +2104,16 @@ spec:
                         pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                         x-kubernetes-int-or-string: true
                       enabled:
-                        description: Enabled activates the node-local-dns addon
+                        description: Enabled activates the node-local-dns addon.
                         type: boolean
                       forwardToKubeDNS:
                         description: If enabled, nodelocal dns will use kubedns as
                           a default upstream
                         type: boolean
+                      image:
+                        description: Image overrides the default docker image used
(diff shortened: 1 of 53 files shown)


johngmyers:
Clusters that cannot fetch from public image registries should use the `assets.containerRegistry` or `assets.containerProxy` field of the cluster spec. See https://kops.sigs.k8s.io/operations/asset-repository/

docs/addons.md
```diff
@@ -129,7 +129,9 @@ NodeLocal DNSCache can be enabled if you are using CoreDNS. It is used to improv
 
 `memoryRequest` and `cpuRequest` for the `node-local-dns` pods can also be configured. If not set, they will be configured by default to `5Mi` and `25m` respectively.
 
-If `forwardToKubeDNS` is enabled, kubedns will be used as a default upstream
+If `forwardToKubeDNS` is enabled, kubedns will be used as a default upstream.
+
+There is also an option to override the default image used by Kops by defining `image`.
```
olemarkus, line 134: nit: kOps, not Kops

docs/addons.md
```diff
@@ -139,6 +141,7 @@ spec:
       enabled: true
       memoryRequest: 5Mi
       cpuRequest: 25m
+      image: "k8s.gcr.io/dns/k8s-dns-node-cache:1.20.0"
```
olemarkus, line 144: Even if we support setting image, we shouldn't encourage people to do so in the docs.

olemarkus (changes requested):

docs/addons.md
dntosas: true that ^ changing

olemarkus (approved):

PR #12220, opened 2021-08-29
[addons/node-termination-handler] Bump version and make image field configurable

dntosas:
Bump patch version of NTH and adding support for image field override by user.


dntosas:
version already bumped in https://github.com/kubernetes/kops/pull/12237, closing ^

PR #12234, opened 2021-08-31
Add ability to provide custom CoreDNS tolerations and affinity

hierynomus:


hierynomus:
CLA has been signed.

hierynomus:
@mikesplain @olemarkus Is there something more needed for this PR? Or you just haven't gotten around to it yet?

olemarkus:
Hi, sorry it is on my list. Have you tested the PR locally? I'd be surprised if this actually works.

Further, offering this kind of customization on a core addon is unlikely to be added. We'd at least have to understand the usecase. Kops will not really work on a cluster where there aren't worker nodes workloads cannot be scheduled on by default.

hierynomus:
Hi Ole, No problem!
Yes, we've tested this on our own environment using a source-built kops executable in our provisioning pipeline. The change itself does what it needs to do.

The usecase for us is that our cluster is configured in such a way that by default we want everything in our k8s clusters to run on AWS spot nodes. But CoreDNS needs to be scheduled on the master and on-demand nodes only.

With this change the default behavior is unchanged, but it is now allowed to specify custom tolerations and affinity such as this (taken from our kops cluster definition):

```
  kubeDNS:
    provider: CoreDNS
    tolerations:
      - effect: NoSchedule
        operator: Exists
    affinity:
      nodeAffinity:
          nodeSelectorTerms:
          - matchExpressions:
            - key: kops.k8s.io/instancegroup
              operator: In
              values:
              - master
              - ondemand-nodes
      podAntiAffinity:
        preferredDuringSchedulingIgnoredDuringExecution:
        - podAffinityTerm:
            labelSelector:
              matchExpressions:
              - key: k8s-app
                operator: In
                values:
                - kube-dns
            topologyKey: kubernetes.io/hostname
          weight: 100
```

hierynomus:
@olemarkus Is the usecase that I described clear?

olemarkus:
So this is essentially to avoid coredns from running on spot instances? CoreDNS should work fine on spot instances. Even more so if the NTH addon is enabled. I would specifically avoid running anything on master nodes that absolutely does not have to run there.

hierynomus:
I think it's more generic, that in the case of setting up a kubernetes cluster with multiple instance-groups, that you would want to be able to control where certain pods land. We see the same at our customers who want to target specific workloads to specific instance-groups.

olemarkus:
Brought this up in office hours today and we decided this feature looks good.

/ok-to-test
/assign

hierynomus:
@olemarkus What is still needed to proceed with this PR? I see that there are a few CI jobs which require an approval before they're run (`build-*-amd64`)

Other than that, I think it needs a code-review/approval?

pkg/apis/kops/cluster.go
```diff
@@ -470,6 +471,10 @@ type KubeDNSConfig struct {
 	CacheMaxSize int `json:"cacheMaxSize,omitempty"`
 	// CacheMaxConcurrent is the maximum number of concurrent queries for dnsmasq
 	CacheMaxConcurrent int `json:"cacheMaxConcurrent,omitempty"`
+	// Tolerations	are tolerations to apply to the kube-dns deployment
```
olemarkus, line 474: This must be added to `pkg/apis/kops/v1alpha2/cluster.go` as well. Then you need to run `make crds apimachinery`

hierynomus: I've fixed that, the GitHub workflows need a new approval before they're re-run

hierynomus:
> I took the liberty to create a test for the change this PR adds: [238670b](https://github.com/kubernetes/kops/commit/238670b171e95454fed83822c638f229373b359c)
> It can be a bit tricky for first-time contributors to create these, so I thought this may save some time.
> Can you cherry that commit into this PR?
> 
> It would also be nice if you could squash some of the commits.

Thanks a lot! That would've indeed cost me considerably more time. I've added your commit and squashed them all into one.

olemarkus:
You need to run `./hack/update-expected.sh` to fix those failures

hierynomus:
> You need to run `./hack/update-expected.sh` to fix those failures

That's indeed what I tried locally before doing the `retest`, but it didn't show any changes after that command. I'll re-run it

hierynomus:
@olemarkus If I just urn `./hack/update-expected`, it executes a bunch of tests, but at the end there is no difference to any of the files. Am I missing something in how to execute this?

olemarkus:
Not sure if you noticed, but `hack/update-expected.sh` fails because of the following error:
```
bootstrapchannelbuilder_test.go:174: error from BootstrapChannelBuilder Build: error reading manifest addons/coredns.addons.k8s.io/k8s-1.12.yaml: error opening resource: error executing resource template "addons/coredns.addons.k8s.io/k8s-1.12.yaml": error executing template "addons/coredns.addons.k8s.io/k8s-1.12.yaml": template: addons/coredns.addons.k8s.io/k8s-1.12.yaml:121:32: executing "addons/coredns.addons.k8s.io/k8s-1.12.yaml" at <8>: wrong type for value; expected string; got []v1.Toleration
```

olemarkus:
It should happen to you locally as well. Do you see this command exiting successfully?

hierynomus:
Ok, after merging in master, I had the same error locally. The template now uses a newly added `ToYAML` TemplateFunction to render the `v1.Tolerations` and `v1.Affinity` objects as yaml.

hierynomus:
@olemarkus Any idea about why that test fails? Is it something I changed, or is it flaky somehow?

olemarkus:
Looks like a flake

/retest

olemarkus (approved):

hierynomus:
Thanks @olemarkus! Two tests now failing on that cni-kuberouter test, looks like flakes indeed. /retest

hierynomus:
Thanks for the guidance @olemarkus!

PR #12238, opened 2021-08-31
Fix etcd3-migration docs link

RetWolf:
Came across this broken link when working on upgrading some older clusters. Pretty minor fix, validated as working locally. I've signed an individual CLA, and made this contribution from a personal machine.


rifelpet (approved):
thanks for the fix!

/retest

PR #12245, opened 2021-09-01
Avoid setting hostname-override

### LINKED ISSUE #10590: Possibility to disable hostname-override in kubelet

olemarkus:
This is an alternative to #10595

Fixes #10590


nodeup/pkg/model/context.go
```diff
@@ -566,42 +566,33 @@ func (c *NodeupModelContext) BuildLegacyPrivateKeyTask(ctx *fi.ModelBuilderConte
 
 // NodeName returns the name of the local Node, as it will be created in k8s
 func (c *NodeupModelContext) NodeName() (string, error) {
-	// This mirrors nodeutil.GetHostName
+
+	// If user explicitly sets hostnameOverride, we respect that even though the CCM will likely just ignore this.
```
hakman, line 570: Maybe we should add validation ad just remove the option?

olemarkus:
As we eventually will move to external CCMs we do need to maintain hostname overrides to match what CCM expects. This goes for both kube-proxy and kubelet.

/hold

PR #12246, opened 2021-09-01
Set kube-apiserver as default logs container

olemarkus:
With this, we don't have to add the kube-apiserver container name when calling `kubectl logs kube-apiserver-xxx`

### DIFF
--- a/nodeup/pkg/model/kube_apiserver.go
+++ b/nodeup/pkg/model/kube_apiserver.go
@@ -732,6 +732,7 @@ func (b *KubeAPIServerBuilder) buildPod(kubeAPIServer *kops.KubeAPIServerConfig)
 
 func (b *KubeAPIServerBuilder) buildAnnotations() map[string]string {
 	annotations := make(map[string]string)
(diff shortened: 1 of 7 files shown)


nodeup/pkg/model/kube_apiserver.go
```diff
@@ -732,6 +732,7 @@ func (b *KubeAPIServerBuilder) buildPod(kubeAPIServer *kops.KubeAPIServerConfig)
 
 func (b *KubeAPIServerBuilder) buildAnnotations() map[string]string {
 	annotations := make(map[string]string)
+	annotations["kubectl.kubernetes.io/default-logs-container"] = "kube-apiserver"
```
rifelpet, line 735: [This KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/2227-kubectl-default-container/README.md) is moving to default-container to cover exec and other kubectl commands.

```suggestion
	annotations["kubectl.kubernetes.io/default-container"] = "kube-apiserver"
```

rifelpet (approved):

PR #12288, opened 2021-09-07
External dns fixes

olemarkus:


docs/cluster_spec.md
```diff
@@ -585,9 +585,11 @@ You can use [external-dns](https://github.com/kubernetes-sigs/external-dns/) as
 ```yaml
 spec:
   externalDns:
-    provider: ExternalDNS
+    provider: external-dns
 ```
 
+Note that you if you have dns-controller installed, you need to remove this deployment before updating the cluster with the new configuration.
```
rifelpet, line 591: can we use our pruning support to delete it instead?

rifelpet (approved):

docs/cluster_spec.md
olemarkus: When/if it merges :)

rifelpet (approved):

PR #12290, opened 2021-09-09
Enable IMDS IPv6 endpoint

rifelpet:
~This is unconditionally enabled.~

~There is an IPv4-equivlanet ["HttpEndpoint" field](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateInstanceMetadataOptions.html#API_LaunchTemplateInstanceMetadataOptions_Contents) that defaults to enabled and we dont expose a way to disable it.~

~HttpProtocolIpv6 defaults to disabled, so to have it match the same behavior of HttpEndpoint we hardcode it to enabled.~

Since we dont allow the ipv4 endpoint to be disabled, I dont think we need to expose a new API field to allow this to be overridden either.~

~I chose not to expose it through the model since any kops upgrade to include this functionality will always have other LaunchTemplate changes that will prompt a new LT version to be written which will include this change, so we dont need to find and detect any necessary LT updates for this field.~

This now enables the IPv6 endpoint whenever IPv6AddressCount is > 0

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
+++ b/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
@@ -45,6 +45,7 @@ func (t *LaunchTemplate) RenderAWS(c *awsup.AWSAPITarget, a, e, changes *LaunchT
 		MetadataOptions: &ec2.LaunchTemplateInstanceMetadataOptionsRequest{
 			HttpPutResponseHopLimit: t.HTTPPutResponseHopLimit,
 			HttpTokens:              t.HTTPTokens,
+			HttpProtocolIpv6:        t.HTTPProtocolIPv6,
 		},
 		NetworkInterfaces: []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest{
 			{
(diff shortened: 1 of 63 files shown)


hakman:
Wonderful :)

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -45,6 +45,7 @@ func (t *LaunchTemplate) RenderAWS(c *awsup.AWSAPITarget, a, e, changes *LaunchT
 		MetadataOptions: &ec2.LaunchTemplateInstanceMetadataOptionsRequest{
 			HttpPutResponseHopLimit: t.HTTPPutResponseHopLimit,
 			HttpTokens:              t.HTTPTokens,
+			HttpProtocolIpv6:        aws.String(ec2.LaunchTemplateInstanceMetadataProtocolIpv6Enabled),
```
hakman, line 48: The state should also be read back in Find()

hakman (changes requested):
> /hold for comment on whether we should only enable this on ipv6 / dualstack clusters. I also havent done any extensive testing around this field and its behavior on ipv4 / dualstack / ipv6 instances

I think we should enable it for now only when `t.IPv6AddressCount > 0`.
Usually operators want to limit access to IMDS for security reasons.

olemarkus:
I agree that it makes sense to only do this only for ipv6. But I'd use the `IPv6Only` function as that is what we do everywhere else. dualstack isn't a thing yet, and if we do dualstack many of the things behind ipv6only should change condition anyway.

hakman:
> I agree that it makes sense to only do this only for ipv6. But I'd use the `IPv6Only` function as that is what we do everywhere else. dualstack isn't a thing yet, and if we do dualstack many of the things behind ipv6only should change condition anyway.

Let's say same way it's done for the IPv6 address count.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
rifelpet: I ended up adding it to the model so it can mimic IPv6AddressCount, so it is now in Find().

olemarkus (commented):
/lgtm

hakman (approved):

PR #12291, opened 2021-09-09
Default to latest staging image for AWS CCM

hakman:
/cc @olemarkus @rifelpet

### DIFF
--- a/pkg/model/components/awscloudcontrollermanager.go
+++ b/pkg/model/components/awscloudcontrollermanager.go
@@ -74,22 +74,21 @@ func (b *AWSCloudControllerManagerOptionsBuilder) BuildOptions(o interface{}) er
 	}
 
 	if eccm.Image == "" {
-		var tag string
+		// See https://us.gcr.io/k8s-artifacts-prod/provider-aws/cloud-controller-manager
 		switch b.KubernetesVersion.Minor {
 		case 18:
-			tag = "v1.18.0-alpha.1"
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -659,23 +659,30 @@ func (tf *TemplateFunctions) OpenStackCCMTag() string {
 	return tag
 }
 
-// AWSCCMTag returns the correct tag for the cloud controller manager based on
+// AWSCCMImage returns the correct image for the cloud controller manager based on
 // the Kubernetes Version
-func (tf *TemplateFunctions) AWSCCMTag() (string, error) {
+func (tf *TemplateFunctions) AWSCCMImage() (string, error) {
```
olemarkus, line 664: Maybe we should move this to pkg/model/components/awscloudcontrollermanager.go with the other arguments?
then we only need `{{ .ExternalCloudControllerManager.Image }}` in the template.

hakman: I think @rifelpet did that in https://github.com/kubernetes/kops/pull/12104?

olemarkus:
You're right. Should hold for that one to go in then?

hakman:
> You're right. Should hold for that one to go in then?

I don' mind, noticed later.

/hold

olemarkus (approved):

PR #12308, opened 2021-09-11
Do not precreate dns record for api lbs

olemarkus:
Precreating DNS records that kops cli will overwrite shortly after doesn't provide much benefit. It is also hard to say to external-dns it doesn't own those records.


olemarkus:
/cc @rifelpet 

This seems to be the last missing piece.

rifelpet (approved):

PR #12314, opened 2021-09-11
Use sg rule ids and tags where possible

olemarkus:


upup/pkg/fi/cloudup/awstasks/securitygrouprule.go
```diff
@@ -146,50 +161,16 @@ func (e *SecurityGroupRule) matches(rule *ec2.IpPermission) bool {
...
-				continue
-			}
-
-			if aws.StringValue(spec.GroupId) == *e.SourceGroup.ID {
-				match = true
-				break
-			}
-		}
-		if !match {
+		if fi.StringValue(e.SourceGroup.ID) != fi.StringValue(rule.ReferencedGroupInfo.GroupId) {
```
rifelpet, line 173: Can `rule.ReferencedGroupInfo` be nil here?

olemarkus: I suppose it can in theory. That would mean there is no CIDRs on the rule (as that would have failed the tests above) and no referencing groups though.
But could perhaps change line 172 to also test for the referenced group.

rifelpet: I think if `e.SourceGroup == nil && rule.ReferencedGroupInfo != nil` or vice versa we should return false

rifelpet:
I wonder if certain AWS partitions don't have support for SG rule IDs yet and this might cause problems for kOps' support in those partitions, similar to the IAM InstanceProfile tagging mentioned in https://github.com/kubernetes/kops/issues/12184

upup/pkg/fi/cloudup/awstasks/securitygrouprule.go
olemarkus: This should do it, I think.

rifelpet (approved):

PR #12322, opened 2021-09-14
Create a second Terraform provider for managed files

rifelpet:
This adds a second Terraform provider definition for managed files. All managed files now have an alias field that refers to the new provider's alias.

Some notes:

* We decided over slack to add a second API field for extra configuration of the new provider definition. I'm not satisfied with the new field's name of `filesProviderExtraConfig` so any suggestions would be appreciated.
* This functionality requires Terraform 0.15 so I added a release note about upgrade requirements
* ~As mentioned in TODO comments, we're use the cloud provider of the cluster rather than of the VFS path, so this may need to be revisited when we begin supporting cross-cloud cluster+VFS combinations. I didn't see a straight forward way in our VFS code to determine a cloud provider based on a path but we could certainly add that later.~
* ~I use a datasource to lookup the bucket's region and use that in the new provider's configuration. Alternatively if we had a way of using the VFS code to determine the bucket's region, that would be preferable. Also I wasn't sure whether to use the ClusterSpec's `ConfigStore` or `ConfigBase`.~
* ~I also added a release note for the deprecation of the TerraformJSON feature flag as discussed in #kops-dev on slack. Terraform's JSON doesn't support provider aliasing so we can't define two AWS providers. Trying to have the MinimalJSON integration test support `-TerraformManagedFiles` proved to be too complex given we'll be removing the support in 1.23 so I'm removing the integration test altogether.~

### DIFF
--- a/docs/releases/1.22-NOTES.md
+++ b/docs/releases/1.22-NOTES.md
@@ -92,6 +92,9 @@ For file assets, it means adding an explicit path as shown below:
 
 # Required actions
 
+* Terraform support now requires Terraform >=0.15.0.
+  Users on older versions must follow Terraform's recommended upgrade path of applying one minor version at a time prior to running `kops update cluster --target terraform`.
+
 * The kOps Terraform support now renders managed files through the Terraform configuration instead 
   of writing them to S3 directly. If, after upgrading kOps and applying a new Terraform plan,
   you subsequently downgrade to an earlier version of kOps, the generated plan will delete these
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -4410,11 +4410,18 @@ spec:
                     description: TerraformSpec allows us to specify terraform config
                       in an extensible way
                     properties:
+                      filesProviderExtraConfig:
+                        additionalProperties:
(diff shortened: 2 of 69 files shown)


rifelpet:
The fact that we use the same memfs paths in AWS and GCE integration tests is causing us to render the GCS files as aws_s3_bucket_objects with this alias, which is causing the verify-terraform job to fail. Maybe we can somehow use our s3fs and gcsfs code for memfs integration tests somehow.

rifelpet:
This is ready for review. I refactored this a bit, now adding the provider definition to the vfs.TerraformPath interface. This allows each VFS provider to define both their terraform resource definitions and terraform provider definitions. I still have a new API field for additional provider configuration that gets merged into what is specified in the VFS code.

This has a nice side effect for our integration tests - because they all use memfs regardless of the cluster's cloud provider, and the memfs terraform resource definitions use `aws_s3_bucket_object`, I defined memfs' provider to use use `aws`, so now the GCE integration tests' kubernetes.tf show `aws_s3_bucket_object`s using an `aws` provider as if we supported that configuration natively, and all of the verify scripts pass.

In the future I could see vaultfs adding terraform support too.

hakman (approved):

PR #12330, opened 2021-09-15
Limited namePrefix to be 64 chars for eventbridge query

allen-mun:
This PR just truncates after the 64th byte in the cluster name string.  Don't know if I should transform the cluster name into runes and then truncate to make sure UTF-8 bytes don't get cut off.

### DIFF
--- a/pkg/resources/aws/eventbridge.go
+++ b/pkg/resources/aws/eventbridge.go
@@ -79,11 +79,17 @@ func ListEventBridgeRules(cloud fi.Cloud, clusterName string) ([]*resources.Reso
 
 	klog.V(2).Infof("Listing EventBridge rules")
 
+	// limit the given NamePRefix to 64 chars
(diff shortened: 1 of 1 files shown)


allen-mun:
@rifelpet is there anything i need to do in this PR to get it merged?

michaelajr:
Our cluster names are longer due to hosted zone names being long. This is a MUCH needed PR for us. Thanks.

rifelpet:
I put up an alternative implementation that adds a random suffix and should fix EventBridgeRule creation as well: https://github.com/kubernetes/kops/pull/12439

would any of you be able to test this to confirm it fixes deletion of clusters with long names?

rifelpet:
We ended up fixing this in https://github.com/kubernetes/kops/pull/12439. Thanks for bringing it to our attention though.

/close

PR #12335, opened 2021-09-16
Allow kops-controller to be scheduled onto unready control plane nodes

rifelpet:
Some [upgrade tests](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-k121-ko121-to-klatest-kolatest/1438334330005884928) are failing and during troubleshooting I noticed [kops-controller never got scheduled](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-k121-ko121-to-klatest-kolatest/1438334330005884928/artifacts/15.223.115.49/kubelet.log) which meant [the other nodes couldn't join the cluster](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-k121-ko121-to-klatest-kolatest/1438334330005884928/artifacts/) because the control plane node had the `node.kubernetes.io/not-ready` taint (in the upgrade case, we lose DNS access to the API so we cant get k8s resources dumped but in [other tests](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-grid-cilium-rhel8-k22-containerd/1438294568381452288/artifacts/cluster-info/nodes.yaml) we can see the similar issue and the nodes reveal the not-ready taint is due to CNI issues).

This should improve the time for new clusters to validate by a bit.

### DIFF
--- a/tests/integration/update_cluster/aws-lb-controller/data/aws_s3_bucket_object_minimal.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/aws-lb-controller/data/aws_s3_bucket_object_minimal.example.com-addons-bootstrap_content
@@ -6,7 +6,7 @@ spec:
   addons:
   - id: k8s-1.16
     manifest: kops-controller.addons.k8s.io/k8s-1.16.yaml
-    manifestHash: 05eb1c527361a625ad1a90200d3166ccaea69b00a4661b2a0fc4f1c0f5d135f5
+    manifestHash: 9da1ffdf784b9de349bb51191f9467375f50bce6236b968d7be3cb782d857554
     name: kops-controller.addons.k8s.io
     needsRollingUpdate: control-plane
     selector:
--- a/tests/integration/update_cluster/aws-lb-controller/data/aws_s3_bucket_object_minimal.example.com-addons-kops-controller.addons.k8s.io-k8s-1.16_content
+++ b/tests/integration/update_cluster/aws-lb-controller/data/aws_s3_bucket_object_minimal.example.com-addons-kops-controller.addons.k8s.io-k8s-1.16_content
@@ -68,12 +68,7 @@ spec:
       priorityClassName: system-node-critical
       serviceAccount: kops-controller
       tolerations:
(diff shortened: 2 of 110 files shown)


olemarkus:
/lgtm

I had a similar problem when using external CCM. That will make nodes tainted with `node.cloudprovider.kubernetes.io/uninitialized`, which also prevents `kops-controller` from scheduling.
I faintly recall some discussions on whether we want kops-controller to tolerate all taints or not, so lets wait a bit for more reviews.

hakman:
We may be a little too quick to tolerate all taints. In general we will want CCM to run first and configure the node, followed by the CNI and then everything else. Otherwise we may see inconsistent behavior, like random IP family for pods.
That's why I suggested using LBs instead of the DNS to access API servers, and not rely on kops-controller to set the DNS records.

rifelpet:
@hakman what would be the impact of kops-controller being scheduled before CCM? AFAIK the only area of kops-controller that deals with changes made by CCM already handles that gracefully:

https://github.com/kubernetes/kops/blob/51afc01e08f6e118ceac465ebc2bf5ac63fcce36/cmd/kops-controller/controllers/awsipam.go#L114-L119

Allowing kops-controller to be scheduled before CCM has the added benefit of allowing the nodes to join sooner, shortening the time to pass cluster validation.

rifelpet:
Actually I believe this PR would cause the same behavior that we intended to avoid in dns-controller in https://github.com/kubernetes/kops/pull/12389. During single-control-plane host upgrades, kops-controller would get rescheduled onto the old control plane host after it was drained (even with OnDelete) so I'll close this out, but I still think we should keep the explicit tolerations introduced in https://github.com/kubernetes/kops/pull/12112/commits/1323ed90408ad403b195f6cc3c648027405576a8

rifelpet:
I think we shouldn't update this anymore. We moved dns-controller away from `Operator: Exists` because of the cordon issue during `rolling-update cluster`. I think the CCM and not-ready tolerations are sufficient enough for kops-controller now.

PR #12342, opened 2021-09-17
feat: add support for wildcard in roles generated for IRSA

eddycharly:
This PR adds support for wildcard in roles generated for IRSA.

Added unit and integration tests.

### DIFF
--- a/pkg/model/awsmodel/iam.go
+++ b/pkg/model/awsmodel/iam.go
@@ -256,7 +256,7 @@ func (b *IAMModelBuilder) buildIAMRolePolicy(role iam.Subject, iamName string, i
 func (b *IAMModelBuilder) roleKey(role iam.Subject) (string, bool) {
 	serviceAccount, ok := role.ServiceAccount()
 	if ok {
-		return strings.ToLower(serviceAccount.Namespace + "-" + serviceAccount.Name), true
+		return strings.ToLower(strings.ReplaceAll(serviceAccount.Namespace+"-"+serviceAccount.Name, "*", "wildcard")), true
 	}
 
 	// This isn't great, but we have to be backwards compatible with the old names.
@@ -405,6 +405,33 @@ func IAMServiceEC2(region string) string {
 	return "ec2.amazonaws.com"
 }
 
+func formatAWSIAMStatement(accountId, oidcProvider, namespace, name string) (*iam.Statement, error) {
+	// disallow wildcard in the service account name
+	if strings.Contains(name, "*") {
+		return nil, fmt.Errorf("service account name cannot contain a wildcard %s", name)
+	}
+
+	// if the namespace contains a wildcard, use StringLike condition instead of StringEquals
+	condition := "StringEquals"
+	if strings.Contains(namespace, "*") {
+		condition = "StringLike"
+	}
+
+	return &iam.Statement{
+			Effect: "Allow",
+			Principal: iam.Principal{
+				Federated: "arn:aws:iam::" + accountId + ":oidc-provider/" + oidcProvider,
+			},
+			Action: stringorslice.String("sts:AssumeRoleWithWebIdentity"),
+			Condition: map[string]interface{}{
+				condition: map[string]interface{}{
+					oidcProvider + ":sub": "system:serviceaccount:" + namespace + ":" + name,
+				},
+			},
+		},
+		nil
+}
+
 // buildAWSIAMRolePolicy produces the AWS IAM role policy for the given role.
 func (b *IAMModelBuilder) buildAWSIAMRolePolicy(role iam.Subject) (fi.Resource, error) {
 	var policy string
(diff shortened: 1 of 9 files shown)


eddycharly:
I would appreciate a little help to get the terraform target working 🙏.
What shall i change to get a valid resource name (of course, having a `*` in the name is not valid) ?

/cc @olemarkus

olemarkus:
Any suggestions, @rifelpet?

tests/integration/update_cluster/irsa/kubernetes.tf
```diff
@@ -1,25 +1,27 @@
...
+  myapp-myotherserviceaccount_role_name             = aws_iam_role.myotherserviceaccount-myapp-sa-minimal-example-com.name
+  node_autoscaling_group_ids                        = [aws_autoscaling_group.nodes-minimal-example-com.id]
+  node_security_group_ids                           = [aws_security_group.nodes-minimal-example-com.id]
+  node_subnet_ids                                   = [aws_subnet.us-test-1a-minimal-example-com.id]
+  nodes_role_arn                                    = aws_iam_role.nodes-minimal-example-com.arn
+  nodes_role_name                                   = aws_iam_role.nodes-minimal-example-com.name
+  region                                            = "us-test-1"
+  route_table_public_id                             = aws_route_table.minimal-example-com.id
+  subnet_us-test-1a_id                              = aws_subnet.us-test-1a-minimal-example-com.id
+  test- * -myserviceaccountwithwildcardns_role_arn  = aws_iam_role.myserviceaccountwithwildcardns-test-wildcard-sa-minimal-e-im2tl2.arn
```
rifelpet, line 21: There is an `ExportWithID` field of the IAMRole task that is only used in building these `local` and `output` names:

https://github.com/kubernetes/kops/blob/6612ba12a596f46c42bc80938ceac05e067a5574/upup/pkg/fi/cloudup/awstasks/iamrole.go#L352-L355

The model sets it in the task here:

https://github.com/kubernetes/kops/blob/6612ba12a596f46c42bc80938ceac05e067a5574/pkg/model/awsmodel/iam.go#L207-L213

We have two options:
* Don't set ExportWithID when the SA namespace contains `*`
* Do the same strings.ReplaceAll that we do with `name` on ExportWithID as well.

eddycharly: Thanks @rifelpet 👍

pkg/model/awsmodel/iam.go
```diff
@@ -256,7 +256,7 @@ func (b *IAMModelBuilder) buildIAMRolePolicy(role iam.Subject, iamName string, i
 func (b *IAMModelBuilder) roleKey(role iam.Subject) (string, bool) {
 	serviceAccount, ok := role.ServiceAccount()
 	if ok {
-		return strings.ToLower(serviceAccount.Namespace + "-" + serviceAccount.Name), true
+		return strings.ToLower(strings.ReplaceAll(serviceAccount.Namespace+"-"+serviceAccount.Name, "*", "-")), true
```
rifelpet, line 259: The `test---myserviceaccount_role_arn` terraform outputs below are a bit awkward. Maybe we can do this?

```suggestion
		return strings.ToLower(strings.ReplaceAll(serviceAccount.Namespace+"-"+serviceAccount.Name, "*", "wildcard")), true
```

That should result in `test-wildcard-myserviceaccount_role_arn` which is a bit easier to understand

eddycharly: What bothers me with the wildcard literal is the potential collision with a namespace called wildcard.
Unfortunately, all substitution schemes seem to have collision issues though.

I’ll change the code as you suggest. We can always change that later if we come up with a better naming scheme.

eddycharly:
I squashed the commits, should be ready for review.

tests/integration/update_cluster/irsa/kubernetes.tf
```diff
@@ -316,6 +326,16 @@ resource "aws_iam_role" "myserviceaccount-default-sa-minimal-example-com" {
   }
 }
 
+resource "aws_iam_role" "myserviceaccount-test---sa-minimal-example-com" {
```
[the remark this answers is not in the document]
eddycharly: Shall I just change the replacing pattern here instead ?

pkg/model/awsmodel/iam.go
```diff
@@ -405,29 +405,48 @@ func IAMServiceEC2(region string) string {
 	return "ec2.amazonaws.com"
 }
 
+func formatAWSIAMStatement(acountId, oidcProvider, namespace, name string) (*iam.Statement, error) {
```
rifelpet, line 408:
```suggestion
func formatAWSIAMStatement(accountId, oidcProvider, namespace, name string) (*iam.Statement, error) {
```

pkg/model/awsmodel/iam.go
```diff
@@ -405,29 +405,48 @@ func IAMServiceEC2(region string) string {
...
+	// if the namespace contains a wildcard, use StringLike condition instead of StringEquals
+	condition := "StringEquals"
+	if strings.Contains(namespace, "*") {
+		condition = "StringLike"
+	}
+
+	return &iam.Statement{
+			Effect: "Allow",
+			Principal: iam.Principal{
+				Federated: "arn:aws:iam::" + acountId + ":oidc-provider/" + oidcProvider,
```
rifelpet, line 423:
```suggestion
				Federated: "arn:aws:iam::" + accountId + ":oidc-provider/" + oidcProvider,
```

rifelpet (commented):
one minor typo, otherwise this LGTM

pkg/model/awsmodel/iam.go
eddycharly: Good catch !

rifelpet (approved):

eddycharly:
@olemarkus shall i ask for cherry picking this in 1.22 branch ?

PR #12349, opened 2021-09-18
Use EC2 and Metadata IPv6 endpoints in IPv6 mode for EBS CSI Driver

hakman:
/cc @olemarkus @rifelpet


olemarkus (approved):
I spent like half a day looking for these config options, but couldn't find them ... Thanks!

hakman:
> I spent like half a day looking for these config options, but couldn't find them ... Thanks!

Yes, they are fairly new, but seems to work pretty well with them.
The only catch is that dual-stack endpoints are only available for select regions:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/Using_Endpoints.html#ipv6

PR #12360, opened 2021-09-19
Update Calico to v3.20.1

hakman:
https://docs.projectcalico.org/archive/v3.20/release-notes/#v3201


upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -538,13 +538,6 @@ spec:
                 description: 'BPFEnabled, if enabled Felix will use the BPF dataplane.
                   [Default: false]'
                 type: boolean
-              bpfExtToServiceConnmark:
```
rifelpet, line 541: this removal is due to a duplicate field: https://github.com/projectcalico/calico/pull/4805

rifelpet (approved):

PR #12375, opened 2021-09-20
Update Go to v1.17.1

hakman:
https://golang.org/doc/go1.17
https://golang.org/doc/devel/release#go1.17

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Set up go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16.8
+          go-version: 1.17.1
 
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


rifelpet:
All of our prow jobs are still on go 1.16, so we'll either need to switch kubekins image tags or wait for the experimental tag to be updated:

https://github.com/kubernetes/test-infra/blob/51a9929c946b1152f8f6dbb55d3409b6f943d77b/images/kubekins-e2e/variants.yaml#L2-L4

hakman:
> All of our prow jobs are still on go 1.16, so we'll either need to switch kubekins image tags or wait for the experimental tag to be updated:

Only those that use go directly. Bazel ones use go 1.17 with this. I don't see any issue with the jobs, but we can change them to master after this is merged.

rifelpet (approved):

PR #12378, opened 2021-09-20
gossip: support resolution of k8s.local names from pods

justinsb:
We add the hosts plugin to CoreDNS, and we populate a ConfigMap from
kops-controller (when in gossip mode).

This enables resolution of the internal apiserver DNS name from Pods,
even when gossip mode (k8s.local) is in use.  This should fix the
failing e2e tests which are assuming that the name in the JWT token is
resolvable from inside the cluster.

This is also a possible step towards a simpler gossip mode, now that
we have a central controller.

### DIFF
--- a/pkg/ipaddr/family.go
+++ b/pkg/ipaddr/family.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ipaddr
+
+import (
+	"fmt"
+	"net"
+	"strings"
+)
+
+// Family is the IP address family (type) for an address; typically IPV4 or IPV6
+type Family string
+
+const (
+	FamilyIPV4 Family = "ipv4"
+	FamilyIPV6 Family = "ipv6"
+)
+
+// GetFamily returns the family for an IP.
+// IPv4-in-IPv6 addresses are treated as IPv6.
+func GetFamily(ip string) (Family, error) {
+	// Verify it is an IP (and not a host:port)
+	parsed := net.ParseIP(ip)
+	if parsed == nil {
+		return "", fmt.Errorf("cannot parse IP %q", ip)
+	}
+
+	// Check for ipv6 using colons; this is hacky but means we classify IPv4-in-IPv6 as v6
+	// Context: https://github.com/golang/go/issues/37921
+	family := FamilyIPV4
+	if strings.Contains(ip, ":") {
+		family = FamilyIPV6
+	}
+
+	return family, nil
+}
--- a/cmd/kops-controller/BUILD.bazel
+++ b/cmd/kops-controller/BUILD.bazel
@@ -21,6 +21,7 @@ go_library(
         "//vendor/k8s.io/api/coordination/v1:go_default_library",
         "//vendor/k8s.io/api/core/v1:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/runtime:go_default_library",
+        "//vendor/k8s.io/apimachinery/pkg/types:go_default_library",
         "//vendor/k8s.io/client-go/plugin/pkg/client/auth/gcp:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
         "//vendor/k8s.io/klog/v2/klogr:go_default_library",
--- a/cmd/kops-controller/controllers/BUILD.bazel
+++ b/cmd/kops-controller/controllers/BUILD.bazel
@@ -13,6 +13,8 @@ go_library(
     deps = [
         "//pkg/apis/kops:go_default_library",
         "//pkg/apis/kops/registry:go_default_library",
+        "//pkg/apis/kops/util:go_default_library",
+        "//pkg/ipaddr:go_default_library",
         "//pkg/kopscodecs:go_default_library",
         "//pkg/nodeidentity:go_default_library",
         "//pkg/nodelabels:go_default_library",
@@ -34,7 +36,9 @@ go_library(
         "//vendor/k8s.io/client-go/kubernetes/typed/core/v1:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
         "//vendor/sigs.k8s.io/controller-runtime:go_default_library",
+        "//vendor/sigs.k8s.io/controller-runtime/pkg/builder:go_default_library",
         "//vendor/sigs.k8s.io/controller-runtime/pkg/client:go_default_library",
+        "//vendor/sigs.k8s.io/controller-runtime/pkg/event:go_default_library",
         "//vendor/sigs.k8s.io/controller-runtime/pkg/manager:go_default_library",
     ],
 )
--- a/cmd/kops-controller/controllers/BUILD.bazel
+++ b/cmd/kops-controller/controllers/BUILD.bazel
@@ -4,6 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = [
         "awsipam.go",
+        "hosts_controller.go",
         "legacy_node_controller.go",
         "node_controller.go",
     ],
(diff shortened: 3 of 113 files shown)


justinsb:
/assign @rifelpet 

I think this might fix the issue we're seeing where we require the JWT service account issuer to be resolvable, as long as we are trying within the cluster....  maybe we should add a GCP job :-)

I guess this is worth a try:
/test pull-kops-e2e-k8s-gce

justinsb:
So the e2e failed overall, but it looks like ServiceAccountIssuer passed.  I guess I should try this on a different PR to figure out whether it is the pull test or something I broke!

However, I also noticed that the tests are now passing - it looks like @rifelpet you modified the tests to specify the serviceAccountIssuer.  Which do you think we should pursue?  (We can always pursue both!)

We can also talk about this in office hours - I don't think this is more than a test issue, but my motivation was to get the GCE e2e producing signal again, but also work in the direction of trying to simplify the gossip code...

rifelpet:
I'm in favor of moving forward with this and reverting https://github.com/kubernetes/test-infra/pull/23666 when we're ready

olemarkus:
Big thumbs up from me as well.

rifelpet:
We have ongoing efforts to support external-dns as a drop-in replacement for dns-controller, and I'm not sure it would be reasonable to expect external-dns to support updating this configmap.

johngmyers:
If it isn't reasonable to have external-dns write the configmap, then perhaps gossip mode should require dns-controller.

It doesn't seem logical to have some use cases have domains managed by dns-controller and others have them managed by kops-controller. It seems particularly unreasonable for those two controllers to have separate implementations of the address gathering logic.

olemarkus:
Gossip is out of the scope for dns-controller, I think. it needs fairly intrusive amount of logic to support the mesh we are using at least. IIRC, gossip already blocks the external-dns provider.

justinsb:
So my intent here is actually to replace the dns-controller logic with this logic over time.

The dns-controller logic is:
* pretty tied to gossip but it turned out that basically we didn't use gossip (we need to seed with a discovery step that might as well just find an instance of kops-controller now)
* it's also very generic, but if using gossip we can't resolve these records anyway externally, and we were using it for what turned out to be a tiny set of cases.

So I'm using this to explore whether we can build a ConfigMap that has the subset of data we need.  We don't currently publish the etcd records, because we don't need those.  I do agree we should make the code do the same thing, particularly with respect to address families, and I'll look into how we can reuse the code here!

rifelpet:
> We don't currently publish the etcd records, because we don't need those.

We do rely on etcd records for the APIServer node role which is still under feature flag. I suppose we could require it not be used on gossip clusters, but it would be preferable to create the etcd records if possible.

https://github.com/kubernetes/kops/pull/10722/files#diff-f81768b75ea2816be1479eac878f905b34b772f08c233164966f7fbe237d91b1R325

justinsb:
As we discussed in office hours, I tried implementing sharing code with gossip.  However, I don't think we should share code, I think we want a new implementation as is proposed here.  This is my thought process:

* We want to get to external-dns.  At that time, we can remove dns-controller.
* I think we *can* replace dns-controller with external-dns even for gossip mode (I hadn't realized this yesterday).  This PR is a step in that direction, in that kops-controller can serve up all the IP addresses to any clients that need it, and then the problem reduces to discovering kops-controller.  That problem is equivalent to finding the seeds which we need for gossip anyway (typically a call to the cloud APIs).  It's probably actually an easier problem, and we could even do things like publishing the IP records to a public bucket. 
* The existing logic is not small, and there are two components to that: (1) the gossip logic itself and (2) the pod/node mapping.  In particular, that involves watching all Pods (because it's an annotation, not a label), which is very expensive.
* This PR aims to demonstrate that we actually don't need the pod/node mapping; that clients are capable of recovering if there are some additional IP addresses.  Where they aren't (e.g. missing timeouts), I think we should fix that instead - it has resiliency benefits beyond just gossip.
* Sharing code pulls in most of the dns-controller logic: pod watching, node watching, joining with the dns.Context, dnsproviders; we'd probably have to implement a new dnsprovider.Interface for the hosts file.
* I might be wrong and then I'll change this to share the code :-)  This is a pretty safe area to start to introduce the logic - it's not really critical path.

cc @johngmyers 

Also @rifelpet thanks for pointing out etcd records; I added them in APIServer mode.

johngmyers:
Should we migrate to using a label?

pkg/ipaddr/family.go
```diff
@@ -0,0 +1,50 @@
...
+// Family is the IP address family (type) for an address; typically IPV4 or IPV6
+type Family string
+
+const (
+	FamilyIPV4 Family = "ipv4"
+	FamilyIPV6 Family = "ipv6"
+)
+
+// GetFamily returns the family for an IP.
+// IPv4-in-IPv6 addresses are treated as IPv6.
```
johngmyers, line 34:
```suggestion
// IPv4-mapped IPv6 addresses are treated as IPv6.
```

pkg/ipaddr/family.go
```diff
@@ -0,0 +1,50 @@
...
+func GetFamily(ip string) (Family, error) {
+	// Verify it is an IP (and not a host:port)
+	parsed := net.ParseIP(ip)
+	if parsed == nil {
+		return "", fmt.Errorf("cannot parse IP %q", ip)
+	}
+
+	// Check for ipv6 using colons; this is hacky but means we classify IPv4-in-IPv6 as v6
```
johngmyers, line 42:
```suggestion
	// Check for ipv6 using colons; this is hacky but means we classify IPv4-mapped IPv6 as v6
```

johngmyers (commented):
One possibility would be to add support to external-dns for patching a configmap. The only reasons I can see for putting it in kops-controller would be the different RBAC needed or the desire to not have the other semantics of external-dns when in gossip mode.

Moving from a pod/node mapping to a straight node mapping has the disadvantage of not being able to take information from the scheduling of pods. If we roll out a new domain for an OnDelete dameonset, we might want to include only those nodes that have the new (or old) version of the pod. Or we might want to use a Deployment, which couldn't be handled without the pod/node mapping information.

I think it makes sense to migrate from using annotations to using labels. We can even use labels that we define in external-dns.

I very much dislike having different discovery semantics for gossip and non-gossip.

justinsb:
I understand the objection - introducing yet-another-way of doing DNS is not ideal.  But this is how we get rid of all that code and move to external-dns.  I'd argue that we actually already have two styles of exposing these services - when we use DNS to the node we join pods and nodes; when we use an apiserver all control-plane nodes are registered with the load balancer and we point DNS to the load balancer.  This is arguably closer to the load-balancer configuration, which is probably more widely used in production than DNS-to-the-node.

As for whether we should use a label to make watching cheaper, yes we probably should!  I think we should do that in external-dns though, perhaps as part of bringing that label to GA.  If we could get onto external-dns using their annotations/labels, I think that would help them feel confident in promoting those labels to GA.  I don't see them introducing support for writing a hosts-file / configmap, but we could try I guess.

My next step here though is to allow nodeup to discover kops-controller before protokube/gossip is ready, making use of the same approach as we have here.  So I agree we should aim to rationalize our logic, I just think we should rationalize onto simpler code than is in dns-controller :-)

johngmyers:
If we want to discover from nodes directly, then we should label or annotate the nodes. The existing external-dns node support doesn't fit our use case; it can only register using a name of the node and there's no way to specify using only the internal IP of the node.

Are you perhaps thinking of using the ConfigMap for all internal domains, even in non-gossip clusters?

I think the way to get rid of dns-controller is to first change it to use the annotations/labels of external-dns or whatever we're moving to.

justinsb:
I was poking around external-dns / dns-controller & discovery, and I think we should be able to just use kubernetes Services/Endpoints as the data source here - and we may even be able to use Services directly e.g. in external-dns.  It was a long time ago, but if I recall the motivation was primarily around the _public_ services, and internal services which want the node IP work just fine.

Comparing a test cluster with APIServerNodes, /etc/hosts has:
```
# Begin host entries managed by kops - do not edit
10.0.16.2       api.internal.foo.k8s.local events.etcd.foo.k8s.local kops-controller.internal.foo.k8s.local main.etcd.foo.k8s.local
10.0.16.3       api.internal.foo.k8s.local
10.0.16.4       api.internal.foo.k8s.local
10.0.16.6       api.internal.foo.k8s.local
10.0.16.7       api.internal.foo.k8s.local events.etcd.foo.k8s.local kops-controller.internal.foo.k8s.local main.etcd.foo.k8s.local
10.0.16.8       api.internal.foo.k8s.local events.etcd.foo.k8s.local kops-controller.internal.foo.k8s.local main.etcd.foo.k8s.local
# End host entries managed by kops
```

Setting up some example services, the endpoints work and expose the/a node internal IP:
```
kubectl get endpoints 
NAME                                     ENDPOINTS                                               AGE
api-internal-foo-k8s-local               10.0.16.2:443,10.0.16.3:443,10.0.16.4:443 + 3 more...   5m55s
etcd-internal-foo-k8s-local              10.0.16.2:4001,10.0.16.7:4001,10.0.16.8:4001            5m55s
kops-controller-internal-foo-k8s-local   10.0.16.2:6443,10.0.16.7:6443,10.0.16.8:6443            5m55s
```

The node name is actually in the endpoints record, so if we wanted to select an alternative node IP, we could do that without too much trouble.

The nice thing here is that we can label our services, so we can easily filter on them.

Obviously we'd have to check IPv6 works now, but as we have the node name there's a path (and services do allow us to define ipFamilies, so this might "just work")

This doesn't address the records for exposing the apiserver outside the cluster, but I think it's a nice simplification.  What do people think?

olemarkus:
If we can use (headless) services, I think this is a good idea.

justinsb:
Closing in favor of #12792 - which is now merged :-)

PR #12381, opened 2021-09-21
GCE: Fix subnet deletion

justinsb:
Subnets are created & owned for IPAlias mode.  We weren't deleting
them because of a bug deleting when there is a hyphen in the name (and
by default they are named after the region, which has a hyphen).


justinsb:
Actually I think I sent this one twice - disregard!

PR #12382, opened 2021-09-21
GCE: clean up subnet support

justinsb:
We should use the subnet spec in the Cluster.


pkg/model/gcemodel/network.go
```diff
@@ -35,35 +35,58 @@ var _ fi.ModelBuilder = &NetworkModelBuilder{}
 
 func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
 	network := &gcetasks.Network{
-		Name:      s(b.NameForNetwork()),
+		Name:      b.LinkToNetwork().Name,
 		Lifecycle: b.Lifecycle,
 		Mode:      "auto", // Automatically create subnets, but stop using legacy mode
+		Shared:    fi.Bool(true),
```
rifelpet, line 41: This means new clusters will always run in an existing network. Should we have this mimic the behavior in AWS where we create new networks by default unless the ID of an existing one is provided?

pkg/model/gcemodel/autoscalinggroup.go
```diff
@@ -225,11 +225,22 @@ func (b *AutoscalingGroupModelBuilder) splitToZones(ig *kops.InstanceGroup) (map
 
 func (b *AutoscalingGroupModelBuilder) Build(c *fi.ModelBuilderContext) error {
 	for _, ig := range b.InstanceGroups {
-		instanceTemplate, err := b.buildInstanceTemplate(c, ig)
+		subnets, err := b.GatherSubnets(ig)
 		if err != nil {
 			return err
 		}
 
+		// Subnets are regional on GCE.  So it's reasonable to have one subnet per IG.
```
zetaab, line 233: I hope that this does not mean OWN subnet per IG? Subnet can be shared between many IGs?

pkg/model/gcemodel/context.go
```diff
@@ -30,16 +30,7 @@ type GCEModelContext struct {
...
-}
-
-// NameForNetwork returns the name for the GCE Network the cluster is located in
-func (c *GCEModelContext) NameForNetwork() string {
-	networkName := c.Cluster.Spec.NetworkID
-	if networkName == "" {
-		networkName = "default"
-	}
-	return networkName
+	return &gcetasks.Network{Name: s("main")}
```
zetaab, line 33: why this returns always "main"?

upup/pkg/fi/cloudup/gcetasks/network.go
```diff
@@ -31,22 +31,25 @@ import (
 // +kops:fitask
 type Network struct {
 	Name      *string
```
zetaab, line 33: what is difference between name and gcename?

zetaab:
@justinsb could you check comments in this PR

PR #12389, opened 2021-09-22
Add specific taints to dns-controller.

olemarkus:
Also set kops-controller as cluster critical, not node critical


rifelpet:
For reference, we believe this might be contributing to our upgrade test failures:

https://testgrid.k8s.io/kops-misc#kops-aws-upgrade-k121-ko121-to-klatest-kolatest

the `kops rolling update` cordons and drains the control plane node, but dns-controller immediately gets rescheduled to the same node because of its `operator: Exists` toleration. After the drain completes, kops deletes the Node object and terminates the EC2 instance. The replacement control plane node's KCM logs reveal Pod Garbage Collection errors regarding this orphaned dns-controller node that still existed after its Node was deleted. We think this may be what prevents it from being rescheduled on the replacement node which prevents the API DNS record from being updated which causes the cluster validation to timeout.

this looks good to me once hack/update-expected is ran

rifelpet (approved):

PR #12393, opened 2021-09-22
Remove circular dependency between Cilium and coredns

johngmyers:
kOps 1.21.1 with Kubernetes 1.18.20 and Cilium 1.9.9

Cilium-agent pods on some nodes are in crashloop, failing with:

```
level=fatal msg="Unable to initialize Kubernetes subsystem" error="unable to create k8s client: unable to create k8s client: Get \"https://api.internal.k8s.REDACTED:443/api/v1/namespaces/kube-system\": dial tcp: lookup api.internal.k8s.REDACTED on 100.64.0.10:53: write udp 10.93.164.247:47610->100.64.0.10:53: write: operation not permitted" subsys=daemon
```

All coredns pods are stuck in ContainerCreating, failing with:

```
  Warning  FailedCreatePodSandBox  5m40s (x91 over 142m)  kubelet  (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "c7d7a7189de309da013e9f2f8c96b0552edb5b5ec1f0955a2ae827547f4cded1" network for pod "coredns-5df8594b56-48qft": networkPlugin cni failed to set up pod "coredns-5df8594b56-48qft_kube-system" network: unable to connect to Cilium daemon: failed to create cilium agent client after 30.000000 seconds timeout: Get "http:///var/run/cilium/cilium.sock/v1/config": dial unix /var/run/cilium/cilium.sock: connect: no such file or directory
Is the agent running?
```

So this is a circular dependency.

[Kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) states:

> For Pods running with hostNetwork, you should explicitly set its DNS policy "`ClusterFirstWithHostNet`".

This appears to be an error in the upstream charts.

### DIFF
--- a/tests/integration/update_cluster/minimal-warmpool/data/aws_s3_bucket_object_minimal-warmpool.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/minimal-warmpool/data/aws_s3_bucket_object_minimal-warmpool.example.com-addons-bootstrap_content
@@ -47,7 +47,7 @@ spec:
       k8s-addon: storage-aws.addons.k8s.io
   - id: k8s-1.16
     manifest: networking.cilium.io/k8s-1.16-v1.10.yaml
-    manifestHash: 561acbb6fc8947695eb062473e9aa2297a9350fea1508d0939189940f48a5b32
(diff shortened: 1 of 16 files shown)


johngmyers:
Looks like my failure might be an artifact of something else.

PR #12394, opened 2021-09-23
Add bidirectional BPF mount for Cilium >= 1.9.10 or >= 1.10.4

ReillyBrogan:
- Cilium versions 1.10.4 and 1.9.10 now auto-mount the bpf file-system automatically. This logic depends on the specific version of Cilium in use so that users who are manually specifying a different version keep the previous behavior.
- Also remove redundant capabilities (these are already automatically granted by virtue of this being a privileged container)

Reference Cilium PRs/commits:
- https://github.com/cilium/cilium/pull/16656
- https://github.com/cilium/cilium/commit/1b20f7f90bace440f52195cefa4df8cbe6fe4cf9#diff-264b5e646aa5ad3db682a4a0a9cd4b4cbbae238d88b033d340e901060c89394a
- https://github.com/cilium/cilium/pull/17131

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.10.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.10.yaml.template
@@ -1,4 +1,5 @@
 {{ with .Networking.Cilium }}
+{{ $semver := (trimPrefix "v" .Version) }}
 {{- if CiliumSecret }}
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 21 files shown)


upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.10.yaml.template
```diff
@@ -188,8 +189,10 @@ data:
   # [0] http://docs.cilium.io/en/stable/policy/language/#dns-based
   # [1] http://docs.cilium.io/en/stable/install/upgrade/#changes-that-may-require-action
   tofqdns-enable-poller: "{{- if .ToFqdnsEnablePoller -}}true{{- else -}}false{{- end -}}"
+  {{- if not (semverCompare ">=1.11 || ~1.10.4 || ~1.9.10" $semver) }}
```
olemarkus, line 192: We'll create new templates for 1.11 so testing for 1.11 is not necessary. 1.11 will also require many other changes to the manifest.

olemarkus (changes requested):
Can you also bump cilium version on pkg/model/components/cilium.go?

olemarkus (approved):

olemarkus:
The weave failure should be fixed by https://github.com/kubernetes/kops/pull/12415

PR #12398, opened 2021-09-23
Remove critical-pod scheduler annotation.

rifelpet:
This is no longer recognized in all supported k8s versions (1.16+)

https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/CHANGELOG/CHANGELOG-1.16.md#deprecations-and-removals

> the scheduler.alpha.kubernetes.io/critical-pod annotation is removed. Pod priority (spec.priorityClassName) should be used instead to mark pods as critical.

There are still references to this in addon manifests but we can debate between removing them now vs when they're removed upstream.


olemarkus (approved):

hakman:
@rifelpet any chance switching to `TF_TAG=1.0.7` will fix it?

rifelpet:
I haven't had any luck on my side pinging someone that can take a look at the issue, so if you want to try something in a new PR go ahead

PR #12406, opened 2021-09-24
Create v1alpha3 apiVersion

johngmyers:
This is an alternative to #9178

Introduce a v1alpha3 API as a copy of v1alpha2. We can then take multiple kOps release cycles to refine the new API, until we are ready to publish it as v1beta1.

The v1alpha3 is not intended to be used in production systems.

PR #12414, opened 2021-09-24
Add Cilium agent pod annotations support to improve personalization

sterchelen:
Annotations is pretty useful when you need third-party tool to add additional behavior
for a k8s resource.
Lots of auto-discovery tools are based on this annotations.

### DIFF
--- a/pkg/apis/kops/networking.go
+++ b/pkg/apis/kops/networking.go
@@ -434,6 +434,9 @@ type CiliumNetworkingSpec struct {
 	// Nat6Range is not implemented and may be removed in the future.
 	// Setting this has no effect.
 	Nat46Range string `json:"nat46Range,omitempty"`
+	// AgentPodAnnotations makes possible to add additional annotations to cilium agent.
+	// Default: none
+	AgentPodAnnotations map[string]string `json:"agentPodAnnotations,omitempty"`
 	// Pprof is not implemented and may be removed in the future.
 	// Setting this has no effect.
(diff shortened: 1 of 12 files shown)


sterchelen:
I signed it

pkg/apis/kops/networking.go
```diff
@@ -434,6 +434,9 @@ type CiliumNetworkingSpec struct {
 	// Nat6Range is not implemented and may be removed in the future.
 	// Setting this has no effect.
 	Nat46Range string `json:"nat46Range,omitempty"`
+	// PodAnnotations makes possible to add additional annotations
+	// Default: none
+	PodAnnotations string `json:"podAnnotations,omitempty"`
```
olemarkus, line 439: I would expect annotations to be of type map[string][string].

This also doesn't distinguish between the many pods that this addon deploys: agent, operator, hubble relay

sterchelen: 💯 , changes applied accordingly.

sterchelen:
Ofc, I've missed out the `toYAML` function. @olemarkus any advice on this? I don't want to implement something without being sure it is the right way/file :)

olemarkus:
The function name is upper cased here. Try `ToYAML`

johngmyers:
Needs to update the v1alpha3 API as well, now that #12406 has merged.

olemarkus (approved):

PR #12415, opened 2021-09-25
Revert "Remove unneeded network related sysctls"

olemarkus:
This reverts commit ce08ec68dfb5bdf2c74d61dfd008c5fa671ef200.

Removing these sysctls broke a number of CNIs (weave, aws vpc, and canal), and possibly breaks configurations we do not test as well.

The accept_ra sysctls can be removed in a followup.

### DIFF
--- a/nodeup/pkg/model/sysctls.go
+++ b/nodeup/pkg/model/sysctls.go
@@ -150,6 +150,18 @@ func (b *SysctlBuilder) Build(c *fi.ModelBuilderContext) error {
 			"")
 	}
 
+	if b.Cluster.Spec.IsIPv6Only() {
+		sysctls = append(sysctls,
+			"net.ipv6.conf.all.forwarding=1",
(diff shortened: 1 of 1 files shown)


rifelpet (approved):

hakman:
Any idea why those things broke? Canal is unmaintained at the moment.

olemarkus:
Most likely the other CNIs set forwarding if not already set. Both Cilium and Calico at least depend on forwarding happening on kernel level when not running in BPF mode. Weave documentation explicitly says it [expects forwarding to be enabled outside of the CNI](https://www.weave.works/docs/net/latest/kubernetes/kube-addon/#-things-to-watch-out-for).

hakman:
In this case you could have just added IPv4 forwarding instead of reverting the whole thing and adding back the `accept_ra` mess.

dawgwe1:
Close all issues and accounts 
Then delete all accounts

PR #12416, opened 2021-09-25
Add fixed version to all addons

### LINKED ISSUE #12249: Unable to update cluster from 1.21 -> 1.22, kops controller crashlooping in AWS
/kind bug

 this information.**

1.22 beta 1

 a `kops` flag.**

1.22.1

AWS

I am updating cluster from 1.21.x k8s (it was installed by using kops 1.21) to 1.22.1 (using kops 1.22 beta 1)

I see kops-controller in new 1.22.1 master crashlooping

```
kops-controller-fzdhz                                                    0/1     CrashLoopBackOff   8 (4m33s ago)   20m   10.124.44.193     ip-10-124-44-193.eu-central-1.compute.internal    <none>           <none>
```

```
% kubectl logs kops-controller-fzdhz
E0902 06:26:05.813281       1 deleg.go:144] setup "msg"="unable to start server" "error"="reading \"kubernetes-ca\" certificate: open /etc/kubernetes/kops-controller/pki/kubernetes-ca.pem: no such file or directory"
```

-

```yaml
```

olemarkus:
This should ensure addon updates are triggered before rolling the cluster, as expected.
We also update addons before labeling masters to ensure that kops-controller is updated before it is scheduled.

Should hopefully fix #12249

/cc @zetaab

### DIFF
--- a/upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
+++ b/upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
@@ -105,6 +105,11 @@ func (b *BootstrapChannelBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	for _, a := range addons.Spec.Addons {
+		// Older versions of channels that may be running on the upgrading cluster requires Version to be set
+		// We hardcode version to a high version to ensure an update is triggered on first run, and from then on
+		// only a hash change will trigger an addon update.
+		a.Version = "9.99.0"
+
 		key := *a.Name
 		if a.Id != "" {
 			key = key + "-" + a.Id
--- a/channels/pkg/api/channel.go
+++ b/channels/pkg/api/channel.go
@@ -67,6 +67,8 @@ type AddonSpec struct {
 
 	// NeedsPKI determines if channels should provision a CA and a cert-manager issuer for the addon.
 	NeedsPKI bool `json:"needsPKI,omitempty"`
+
+	Version string `json:"version,omitempty"`
 }
 
 func (a *Addons) Verify() error {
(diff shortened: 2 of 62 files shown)


hakman (changes requested):
Can you set Version as KopsVersion? Should be good enough for what we want upgraded and won't be a strange fixed number.

upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
```diff
@@ -105,6 +105,11 @@ func (b *BootstrapChannelBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	for _, a := range addons.Spec.Addons {
+		// Older versions of channels that may be running on the upgrading cluster requires Version to be set
+		// We hardcode version to a high version to ensure an update is triggered on first run, and from then on
+		// only a hash change will trigger an addon update.
```
rifelpet, line 110: Consider mentioning "This can be removed when kops no longer supports k8s 1.22"

olemarkus: Unfortunately this follows kops versioning and we don't have any skew/upgrade jump policy. This will have to remain forever unless we change this

olemarkus:
> Can you set Version as KopsVersion? Should be good enough for what we want upgraded and won't be a strange fixed number.

We need to set a version that is at least as high as the old addon versions. That is 4.0.0, I think. If we don't none of these addons will be updated.

hakman:
> > Can you set Version as KopsVersion? Should be good enough for what we want upgraded and won't be a strange fixed number.
> 
> We need to set a version that is at least as high as the old addon versions. That is 4.0.0, I think. If we don't none of these addons will be updated.

My understanding is that they will be updated once new control-plane nodes are started, we just need the kOps ones (like kops-controller) to be updated before that.

rifelpet (approved):

PR #12418, opened 2021-09-25
Revert "Enable ipv6 forwarding and router announcements"

hakman:
`net.ipv6.conf.all.forwarding=1` is not needed with Calico and Cilium (the only supported IPv6 CNIs)
`net.ipv6.conf.all.accept_ra=2` is a big mess that is conflicting with systemd-networkd on Ubuntu and in general, not working.

This reverts commit ab596a49

/cc @olemarkus @rifelpet

### DIFF
--- a/nodeup/pkg/model/sysctls.go
+++ b/nodeup/pkg/model/sysctls.go
@@ -150,17 +150,10 @@ func (b *SysctlBuilder) Build(c *fi.ModelBuilderContext) error {
 			"")
 	}
 
-	if b.Cluster.Spec.IsIPv6Only() {
-		sysctls = append(sysctls,
-			"net.ipv6.conf.all.forwarding=1",
-			"net.ipv6.conf.all.accept_ra=2",
-			"")
-	} else {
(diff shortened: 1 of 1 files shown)


olemarkus:
I am trying to get plain kubenet working with ipv6, and that seems to require ipv6 forwarding from kops. Technically speaking, weave should also support ipv6 so we do not need to intentionally break something that accidentally may work.

accept_ra has no effect and can be removed though. It does no harm leaving it as-is, but it doesn't serve any purpose.

hakman:
> accept_ra has no effect and can be removed though. It does no harm leaving it as-is, but it doesn't serve any purpose.

`accept_ra` can interfere with systemd-networkd. It's not supposed to be enabled at same time.

> I am trying to get plain kubenet working with ipv6

Then why not make it part of that PR, when it's ready? At the moment it does nothing useful.

hakman:
BTW, this is something that has to be removed from 1.22 too.

hakman:
Replaced by #12461

PR #12420, opened 2021-09-26
WIP: Support GCE TPM verification

justinsb:
Support for instance authentication using vTPM on GCE.


upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	if node.Zone == "" {
+		return nil, fmt.Errorf("zone is required")
+	}
+	if node.Instance == "" {
+		return nil, fmt.Errorf("instance is required")
+	}
+
+	if node.GCPProjectID != v.opt.ProjectID {
+		return nil, fmt.Errorf("projectID does not match expected: got %q, want %q", node.GCPProjectID, v.opt.ProjectID)
```
johngmyers, line 90: If the client's project ID has to match the server's project ID, why bother having the client pass a project ID in the first place?

upup/pkg/fi/cloudup/gce/tpm/gcetpmsigner/tpmauthenticator.go
```diff
@@ -0,0 +1,150 @@
...
+			Instance:     a.instance,
+		}
+
+		payload, err := json.Marshal(node)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal node claim: %w", err)
+		}
+		signature, err := tpmSign(key, payload)
+		if err != nil {
+			return nil, fmt.Errorf("failed to sign node claim: %w", err)
```
johngmyers, line 102: Why sign the node claim? What would an attacker gain by altering it?

upup/pkg/fi/cloudup/gce/tpm/gcetpmsigner/tpmauthenticator.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// double-check the signature.
+	keyRSA, ok := tpmKey.PublicKey().(*rsa.PublicKey)
+	if !ok {
+		return nil, fmt.Errorf("unexpected type for attestation key, got %T, want rsa.PublicKey", tpmKey.PublicKey())
+	}
+
+	attestHash := sha256.Sum256(payload)
+	if err := rsa.VerifyPKCS1v15(keyRSA, crypto.SHA256, attestHash[:], signature); err != nil {
+		return nil, fmt.Errorf("signature double-check failed: %w", err)
```
johngmyers, line 133: What's the purpose of the double-check?

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
```diff
@@ -0,0 +1,241 @@
...
+			sans = append(sans, iface.Ipv6Address)
+		}
+		break
+	}
+
+	return sans, nil
+}
+
+// ParseCSR decodes a PEM encoded CSR
+func ParseCSR(pemBytes []byte) (*x509.CertificateRequest, error) {
```
johngmyers, line 206: This doesn't appear to be referenced?

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	if clusterName != v.opt.ClusterName {
+		return nil, fmt.Errorf("clusterName does not match expected: got %q, want %q", clusterName, v.opt.ClusterName)
+	}
+	if instanceGroupName == "" {
+		return nil, fmt.Errorf("could not determine instance group for instance %s", instance.SelfLink)
+	}
+
+	// TODO can we include the attestation certificate (signed by GCE) in the claim, rather than getting it from the API?
```
johngmyers, line 128: If the attestation certificate also contains the zone, instance metadata and network interfaces you might be able to avoid using the API altogether. Otherwise, I don't see the value in merely reducing the number of calls.

Getting it from the API validates that the instance still exists.

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
```diff
@@ -0,0 +1,241 @@
...
+		if !verifySignature(tpmSigningKey, nodeClaim, claim.NodeSignature) {
+			return nil, fmt.Errorf("failed to verify claim signature for node: %w", err)
+		}
+	}
+
+	sans, err := GetInstanceCertificateAlternateNames(instance)
+	if err != nil {
+		return nil, err
+	}
+
```
johngmyers, line 149: This completely fails to verify the request body!!! An attacker obtaining data signed by the TPM that deserializes to a nodeClaim that identifies the TPM's instance could use it to obtain certs for keys of the attacker's choice.

johngmyers (changes requested):

pkg/apis/nodeup/bootstrap.go
```diff
@@ -30,6 +30,30 @@ type BootstrapRequest struct {
 	// IncludeNodeConfig controls whether the cluster & instance group configuration should be returned.
 	// This allows for nodes without access to the kops state store.
 	IncludeNodeConfig bool `json:"includeNodeConfig"`
+
+	// Claim is for including rich evidence for the node identity, such as a TPM attestation
+	Claim *Claim `json:"claim,omitempty"`
```
johngmyers, line 35: You can't put the signature inside the thing being signed.

upup/pkg/fi/cloudup/gce/tpm/gcetpmsigner/tpmauthenticator.go
justinsb: Removed, it was in the gce provider but I agree it wasn't doing anything.

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
justinsb: Downgraded from a TODO to a Note.  It's a cute trick, at least, and might be the default way for other TPM implementations.

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
justinsb: Well spotted, and fixed now!

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
justinsb: Thanks - removed!

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
justinsb: It's something we can relax in future.

upup/pkg/fi/cloudup/gce/tpm/gcetpmsigner/tpmauthenticator.go
justinsb: Changed the way the signature matches to match the way it works on the AWS implementation.

Previously, we were demonstrating that we had control of the signing key, but it's a lot more prone to subtle mistakes ... as you spotted :-)

justinsb:
I've refactored this to use the header, just as AWS does, so we don't immediately need #12423

As such, I think this is probably no longer WIP - reviews greatly appreciated (thanks for taking an early look @johngmyers !)

upup/pkg/fi/cloudup/gce/tpm/gcetpmsigner/tpmauthenticator.go
```diff
@@ -0,0 +1,136 @@
...
+
+	klog.Infof("TPM initialization took %v", time.Since(tpmStart))
+
+	token := &gcetpm.AuthToken{}
+
+	token.Data.GCPProjectID = a.projectID
+	token.Data.Zone = a.zone
+	token.Data.Instance = a.instance
+	token.Data.Timestamp = time.Now().Unix()
+	token.Data.Audience = gcetpm.AudienceNodeAuthentication
```
johngmyers, line 93: Including timestamp and audience in the signature is definitely good.

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
```diff
@@ -0,0 +1,247 @@
...
+	clusterName := ""
+	instanceGroupName := ""
+	for _, item := range instance.Metadata.Items {
+		switch item.Key {
+		case gce.MetadataKeyInstanceGroupName:
+			instanceGroupName = fi.StringValue(item.Value)
+		case "cluster-name":
+			clusterName = fi.StringValue(item.Value)
```
johngmyers, line 133: One thing to consider, perhaps not for this PR, is how to allow an instance to revoke its privilege to make this query of kops-controller. That way, any malicious code later obtaining access to the TPM wouldn't be able to escalate to having kubelet access. For AWS, my employer has a hook run after nodeup, sending a request to a lambda that changes the instance role to one with minimal privileges.

I think for GCE you could have it add or remove an instance metadata item, which this code would check.

upup/pkg/fi/cloudup/gce/tpm/gcetpmsigner/tpmauthenticator.go
```diff
@@ -0,0 +1,136 @@
...
+
+	payload, err := json.Marshal(token.Data)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal token data: %w", err)
+	}
+	signature, err := tpmSign(key, payload)
+	if err != nil {
+		return "", fmt.Errorf("failed to sign token data: %w", err)
+	}
+	token.Signature = signature
```
johngmyers, line 104: This is a bit fragile in that it relies on the client and server both serializing JSON the same way. When using a serialization format, such as JSON, which allows multiple possible encodings of the same thing it is better to transmit the precise byte sequence that got signed. Then again, for this particular case it is extremely likely for the client and server to be both using the same version of the same encoding library.

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
```diff
@@ -0,0 +1,247 @@
...
+	if token.Data.Instance == "" {
+		return nil, fmt.Errorf("instance is required")
+	}
+
+	// Verify node is in our cluster
+	if token.Data.GCPProjectID != v.opt.ProjectID {
+		return nil, fmt.Errorf("projectID does not match expected: got %q, want %q", token.Data.GCPProjectID, v.opt.ProjectID)
+	}
+
+	instance, err := v.computeClient.Instances.Get(token.Data.GCPProjectID, token.Data.Zone, token.Data.Instance).Do()
```
johngmyers, line 114: Consider the possibility of the client including '/' or other problematic character in the zone or instance field. Could that cause the computClient implementation to misbehave?

upup/pkg/fi/cloudup/gce/tpm/token.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Audience is the audience for this request (to help prevent replay attacks)
+	Audience string `json:"audience,omitempty"`
+}
+
+// GCETPMAuthenticationTokenPrefix is the prefix used for authentication using the GCE TPM
+const GCETPMAuthenticationTokenPrefix = "x-gce-tpm "
+
+// AudienceNodeAuthentication is used in case we have multiple audiences using the TPM in future
+const AudienceNodeAuthentication = "kops.k8s.io/node"
```
johngmyers, line 53: Perhaps "node-bootstrap"?

johngmyers (changes requested):
The NPE nit is the only blocker.

cmd/kops-controller/main.go
```diff
@@ -94,6 +95,12 @@ func main() {
 				setupLog.Error(err, "unable to create verifier")
 				os.Exit(1)
 			}
+		} else if opt.Server.Provider.GCE != nil {
```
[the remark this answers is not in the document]
justinsb: This is actually nested in an `opt.Server != nil` check.  That said, I do think we could think about having something like `GetServer()` that is nil safe (i.e. returns an empty `ServerOptions` if `Server` - or `opt` - was nil.  But probably that's a deeper change for another time!

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
justinsb: That's a really good idea - though I agree that as it isn't something we can do in the TPM itself (I assume!) it probably doesn't belong in this PR.  We could also consider having kops-controller remember that it issued the cert (maybe mark the node?) / refuse to issue to a machine that had been up for more than N minutes.

upup/pkg/fi/cloudup/gce/tpm/gcetpmsigner/tpmauthenticator.go
justinsb: Good point!  I guess it's not too hard to change it _now_ to be `payload []byte` and `signature []byte` (where it would be a pain later) and just do an additional JSON decode ... I'll make the change.

upup/pkg/fi/cloudup/gce/tpm/gcetpmverifier/tpmverifier.go
justinsb: Such a good point.  We shouldn't be trusting this data until we've checked the signature.  I had a look and the google SDK does do the right thing in terms of escaping (I'd feel happier if it was using proto, but the escaping logic looks pretty sound).  Adding a comment though to remind ourselves to be careful!

upup/pkg/fi/cloudup/gce/tpm/token.go
justinsb: Good call - done!

PR #12423, opened 2021-09-26
Refactor bootstrap logic to allow for alternative signatures

justinsb:
No real code changes (different rejection method on malformed JSON body, probably); but this allows for different signature methods - for example using vTPM on GCE.

Builds on #12422


johngmyers:
I could use a bit more information about why this refactoring is necessary. Why is this pushing the serialization/deserialization of `nodeup.BootstrapRequest` into the cloud-provider-specific code? That should be the same for all cloud providers, should it not?

Is it really the case that the GCE-specific signature can't be easily separated from the signed content? If so, then the methods could take/return the  `[]byte` serialized form of the request under signature.

justinsb:
Sorry @johngmyers  - should have put the link... https://github.com/kubernetes/kops/pull/12420  is the "big picture" PR that I'm carving up.

The reasoning is that the "proof" information doesn't really fit into the header any more.  I think this also helps AWS, in that the signature on AWS requires that both sides use json.Marshal/Unmarshal identically, and this makes it clearer (aka stops someone refactoring it).

johngmyers:
I don't understand what you mean by " requires that both sides use json.Marshal/Unmarshal identically". All cloud providers should be marshaling or unmarshaling the `BootstrapRequest` the same way: we should not be affording the cloud providers the ability to marshal `BootstrapRequest` a different way or to omit signing any part of it.

I'm a bit skeptical that an RSA signature, zone, and instance name don't fit in an HTTP header, but I'll have to try it out. If this can't be encoded in the header, then we'd have to have some outer serialization of the signed part and the signature in the body. Such a thing should be outside the cloud-provider-specific code.

PR #12426, opened 2021-09-26
Allow aws-iam-authenticator to be scheduled onto dedicated apiserver nodes

### LINKED ISSUE #12390: IAM Authenticator addon not updating nodeSelector/toleration for APIServer flag
/kind bug

 this information.** 1.21.1

 a `kops` flag.** 1.20.9

**3. What cloud provider are you using?** AWS

```
export KOPS_FEATURE_FLAGS="+APIServerNodes"
kops update cluster --yes
```

New dedicated apiserver nodes were launched, but didn't contain aws-iam-authenticator pods, as the addon defines tolerations and selectors as:
```
  nodeSelector:
    node-role.kubernetes.io/master: ""
  tolerations:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
```

This pod is required to run on all master nodes, so expected behavior would be that kops would modify the addon when the `+APIServerNodes` flag is set:

```patch
  nodeSelector:
-    node-role.kubernetes.io/master: ""
+    node-role.kubernetes.io/api-server: ""
  tolerations:
+  - effect: NoSchedule
+    key: node-role.kubernetes.io/api-server
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
```

rifelpet:
Also adding the CCM toleration to match dns-controller in https://github.com/kubernetes/kops/pull/12389

Fixes #12390

### DIFF
--- a/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
@@ -129,11 +129,15 @@ spec:
 
       # run on each master node
       nodeSelector:
-        node-role.kubernetes.io/master: ""
+        {{ APIServerNodeRole }}: ""
       priorityClassName: system-node-critical
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/api-server
+      - key: node.cloudprovider.kubernetes.io/uninitialized
+        operator: Exists
       - key: CriticalAddonsOnly
         operator: Exists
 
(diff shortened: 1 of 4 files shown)


justinsb:
Makes sense to me to tolerate the apiserver taint; apiserver talks to the authentication webhook over a localhost URL, so it has to run on every apiserver pod.

One question would be whether we should be running it with a universal toleration, but we can discuss this at office hours sometime.  Would be good to get some rules here!

/approve
/lgtm

I think we should @olemarkus approve though, as IIRC they were the main author of the dedicated apiserver nodes.

/hold for olemarkus to take a look

/assign @olemarkus

olemarkus:
Looks good

/hold cancel

PR #12431, opened 2021-09-27
Mount cgroupv2 for cilium at a custom location

olemarkus:
This should fix cilium for AL2 5.10 kernel.

### DIFF
--- a/nodeup/pkg/model/networking/cilium.go
+++ b/nodeup/pkg/model/networking/cilium.go
@@ -17,7 +17,9 @@ limitations under the License.
 package networking
 
 import (
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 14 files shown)


rifelpet:
The cilium's DaemonSet is OnDelete but there's still a chance that between `update cluster --yes` and all nodes being replaced that a cilium-agent pod being deleted results in it using the new mount path, even though the systemd unit was never ran on that node. Is that a problem we need to address here?

olemarkus:
In this case, cilium will mount cgroupv2 inside the pod, which works fine. The reason for mounting cgroup2 on the host is to preserve some socket loadbalancing state, which is not _that_ critical.

avdhoot:
This might fixes cilium in flatcar Linux stable ([2983.2.0](https://www.flatcar-linux.org/releases/#release-2983.2.0)). by default, flatcar mount cgroup2 on `/sys/fs/cgroup` . Currently, cilium does work with flatcar with kops `1.21.1`.  I have changed the mount path `cilium-cgroup` in cilium DaemonSet to `/sys/fs/cgroup` to make it work. 
```
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime,seclabel)
```
Is there any chance this can be cherry-picked in kops `1.21.2`  @olemarkus

olemarkus:
The change is unfortunately a bit too big for 1.21 branch, I think. I already had to fix one bug introduced by this change.

PR #12437, opened 2021-09-28
Make it possible to set CAS max-node-provision-time

olemarkus:


pkg/apis/kops/componentconfig.go
```diff
@@ -992,6 +992,8 @@ type ClusterAutoscalerConfig struct {
 	// CPURequest of cluster autoscaler container.
 	// Default: 100m
 	CPURequest *resource.Quantity `json:"cpuRequest,omitempty"`
+	// MaxNodeProvisionTime determines how long CAS will wait for a node to join the cluster.
+	MaxNodeProvisionTime *string `json:"maxNodeProvisionTime,omitempty"`
```
johngmyers, line 996: Why is this a pointer? What is the utility of setting this to the empty string?

olemarkus: I suppose that is the case for many of the above. Empty string and nil would mean the same.

Do you want to change this to non-pointer if we prefer that?

Issue #12329, opened 2021-09-15
Unable to delete cluster with a long cluster name, listing Eventbridge rules fails as the namePrefix given is too long


allen-mun:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Version 1.21.1 (git-ffabc3bf682bfc25ffdce99e54005345df94b467)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:52:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/amd64"}

server version from kops configuration yaml:
kubernetesVersion: 1.21.4

**3. What cloud provider are you using?**
aws

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
`kops delete cluster`

You can reproduce this issue by creating a cluster `kops create cluster` with a name greater than 64 characters, then attempting to delete the cluster.

**5. What happened after the commands executed?**
```
I0915 16:16:00.039164   33406 request_logger.go:45] AWS request: EventBridge/ListRules

error listing Eventbridge rules: ValidationException: 1 validation error detected: Value '<long_host_name>.<long_aws_domain>' at 'namePrefix' failed to satisfy constraint: Member must have length less than or equal to 64
	status code: 400, request id: <request_id>
```
AWS errors out when listing eventbridge as the namePrefix given (the cluster name) is over 64 characters long
**6. What did you expect to happen?**
The cluster should be deleted properly.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2021-09-15T20:14:59Z"
  name: <long_cluster_name>
spec:
  api:
    loadBalancer:
      class: Classic
      crossZoneLoadBalancing: true
      idleTimeoutSeconds: 3600
      securityGroupOverride: <sg_arn>
      type: Internal
  assets:
    containerRegistry: <internal_hosting_service>
    fileRepository: <internal_hosting_service>
[...]

PR #12439, opened 2021-09-28
Truncate cluster name in NTH EventBridgeRules

rifelpet:
ref: https://github.com/kubernetes/kops/issues/12329

alternative to https://github.com/kubernetes/kops/pull/12330

This truncates them with a random suffix and uses that shortened name in both the creation and deletion of event bridge rules. This should fix EventBridgeRule creation on clusters with long names.


tests/integration/update_cluster/nth_sqs_resources/kubernetes.tf
```diff
@@ -201,193 +201,193 @@ resource "aws_autoscaling_group" "nodes-nthsqsresources-example-com" {
...
   lifecycle_transition   = "autoscaling:EC2_INSTANCE_TERMINATING"
   name                   = "nodes-NTHLifecycleHook"
 }
 
-resource "aws_cloudwatch_event_rule" "nthsqsresources-example-com-ASGLifecycle" {
-  event_pattern = file("${path.module}/data/aws_cloudwatch_event_rule_nthsqsresources.example.com-ASGLifecycle_event_pattern")
-  name          = "nthsqsresources.example.com-ASGLifecycle"
+resource "aws_cloudwatch_event_rule" "nthsqsresources-longclustername-e-fkbaoh-ASGLifecycle" {
+  event_pattern = file("${path.module}/data/aws_cloudwatch_event_rule_nthsqsresources.longclustername.e-fkbaoh-ASGLifecycle_event_pattern")
+  name          = "nthsqsresources.longclustername.e-fkbaoh-ASGLifecycle"
```
rifelpet, line 229: here's the relevant change in the integration test

upup/pkg/fi/cloudup/awsup/aws_utils.go
```diff
@@ -221,6 +221,15 @@ func ELBv2Tags(tags map[string]string) []*elbv2.Tag {
 	return elbv2Tags
 }
 
+// EventBridgeRulePrefix returns a suitable prefix for event bridge rules given their length restrictions
+func EventBridgeRulePrefix(cluster string) string {
```
hakman, line 225: I would call this more like ClusterName40. But it's just a suggestion. Seems more consistent with GetResourceName32 that we use in other places.

hakman (approved):

BenWolstencroft:
@rifelpet - Thanks for this fix, any idea when there will be a re-release of kops 1.21 including this fix? - we've ended up in the unenviable situation where some of our cluster estate has upgrade and some has hit this cluster hostname length issue.

PR #12485, opened 2021-10-05
[DigitalOcean] Incorporate existing vpc support for kops

srikiz:
This PR introduces minor changes to create KOPS DO cluster in an existing VPC.
It uses the `--vpc` flag that already exists to pass the vpc uuid where the droplets must be created.

Example:
./kops create cluster --cloud=digitalocean --name=dev1.example.co.in --vpc=af287488-862e-46c7-a783-5e5fa89cb256 --networking=cilium --zones=tor1 --ssh-public-key=~/.ssh/id_rsa_github.pub

Tested to make sure the droplets are created in the vpc specified.

Next steps, may include additional support for creating new vpc as part of the kops create cluster option.

### DIFF
--- a/pkg/model/domodel/droplets.go
+++ b/pkg/model/domodel/droplets.go
@@ -17,7 +17,6 @@ limitations under the License.
 package domodel
 
 import (
-	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


hakman (approved):
👍

PR #12498, opened 2021-10-07
Use .bazel-bin to help gopls & VSCode

justinsb:
This means that gopls will ignore the bazel-* symlinks, in particular
when using the go LSP plugin (gopls) with VSCode.

Idea from https://bitworking.org/news/2021/06/bazel-and-gopls/

### DIFF
--- a/bazel-bin
+++ b/bazel-bin
@@ -1,0 +1,1 @@
+.bazel-bin


olemarkus:
Very interesting. Tried to make use of this PR, but gopls still doesn't seem entirely happy and VSCode needs a restart after bazel-build

PR #12521, opened 2021-10-11
Spotinst: Permission to create `bigdata.spot.io/bigdataenvironments`

liranp:
The PR updates the controller's cluster role to allow the creation of `bigdata.spot.io/bigdataenvironments` resources.


johngmyers:
@liranp Could you reference the appropriate controller documentation? Does it not need to update or delete said resources?

liranp:
Both Elastigroup and Ocean use the controller to sync cluster state with Spot. More information is available [here](https://docs.spot.io/ocean/tutorials/spot-kubernetes-controller/).

PR #12531, opened 2021-10-13
Stop skipping snapshot fields tests in EBS CSI e2e

rifelpet:
the linked issue's PR was merged into 1.22 so we'll see if we can stop skipping these tests


rifelpet:
I expect these to fail but we'll see if the unskipped tests pass or not:

/test pull-kops-e2e-aws-ebs-csi-driver
/test pull-kops-e2e-aws-ebs-csi-driver-irsa

olemarkus:
I am wondering if we should switch to running the tests in `e2e` rather than `e2e-kubernetes`. The former are tests specific for the EBS driver, while the latter just looks like a fork of k/k's tests, which we cover by all the other runs anyway.

olemarkus (approved):

PR #12544, opened 2021-10-16
Show additional ("addon") objects in kops get

justinsb:
If there are additional objects in the kOps state store, we should output them from `kops get` (which shows all objects).


johngmyers:
It is inconsistent that some things use "additional" while others use "addons". I think it should consistently use "addons".

I see a good case for also adding a `kops get addons` command.

justinsb:
Thanks @johngmyers - renamed to addons - much more consistent :-)

johngmyers (approved):

PR #12554, opened 2021-10-19
gossip: Seed /etc/hosts in nodeup

justinsb:
In some scenarios (e.g. cilium), we rely on the internal DNS name
being available, but this isn't the case with gossip clusters.

nodeup can seed /etc/hosts for the control-plane nodes, breaking the
deadlock.


justinsb:
This was causing gossip + cilium to fail, at least, e.g. https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-grid-gce-cilium-u2004-k22-docker/1450342567488000000/artifacts/cluster-info/kube-system/cilium-operator-558c4b6599-dc62t/logs.txt

cc @rifelpet

PR #3087, opened 2017-07-29
Add `kops create secret dockerconfig` feature

blakebarnett:
This adds a well-known secret name `dockerconfig` which will automatically
be used if present to create `/root/.docker/config.json` on all nodes. This will
allow private registries to be used for kops hooks as well as any k8s images
without the need to define `imagePullSecrets` in every namespace.

closes https://github.com/kubernetes/kops/issues/2505

PR #12562, opened 2021-10-20
Containerd repo mirror auth

haad:
Add support for registry mirror authentication with containerD. This was managed byt overiding cotnainerd config as whole.


haad:
Ping

hakman:
Hi @haad. Sorry for the delay in answering. We had some discussions about this topic at our weekly meeting and some concerns were raised about adding passwords directly into kOps config. There was no consensus on how to approach this last week, but it's still on our list of priorities.

One question came up about the use of Docker secrets managed by kOps which are directly used by k8s even with containerd. Did you try to see if those work somehow for the mirrors too?.

haad:
I understand, but We are not forcing our users to include those passwords there. Plus you have an option to use a token, too. Right no every Kops users has to override this config as a whole which isn't a good thing. It complicates workflow for them as they have to check if we have not changed default configuration(containerd/config.toml) before each upgrade.


We should be able to add some cloud based password insertion scheme (Use aws system parameters as passwords), but I would add/implment that functionality as a separate PR.

blakebarnett:
Shouldn't this be implemented similarly to the host-level docker auth (as a secret)? We could maybe abstract it to be a generic "registry auth" secret, and have it apply to both...

https://github.com/kubernetes/kops/pull/3087

olemarkus:
That would be my preferred way of implementing this

PR #12564, opened 2021-10-20
Disable CNP status updates by default

olemarkus:
This change was suggested by upstream to avoid ever-growing objects. See https://cilium.slack.com/archives/C53TG4J4R/p1634655498336400


yurrriq:
Can we backport to 1.22.2?

olemarkus:
It's tecnically a BC break, so I do not think we will backport this one. 1.23 is also not that far away now.

olemarkus:
Ended up making hte setting configurable at least, but defaults to disabled #12564

PR #12605, opened 2021-10-25
Make dns-controller delete placeholder addresses for IPv6 cluster

johngmyers:
On an IPv6 cluster, the internal domain records for api and kops-controller had two records, an A record with the placeholder address and the correct AAAA record for the single control plane node.

Fix dns-controller to remove A records for domains that only have AAAA records (and vice versa).


johngmyers:
I see a lot of back-and-forth in the dns-controller reconciliation loop upon cluster bootstrap. In an IPv6 cluster it deletes the placeholder A record then later puts in the correct A record. I'll be looking into this.

PR #12609, opened 2021-10-26
[WIP] Use prefixes for IPv6 with Calico

hakman:


hakman:
Replaced with a clean version as #12643.

PR #12621, opened 2021-10-28
GCE: Open BGP port for calico

justinsb:
Calico uses BGP for route propagation.


zetaab:
@justinsb perhaps override that test or how we should fix it?

justinsb:
I _think_ this is still needed, going to reopen while I figure it out.  It shouldn't be closed without a reason anyway!

PR #12626, opened 2021-10-28
Use InternalIP as preferred kubelet address only in ivp6 mode

olemarkus:
As metrics-server rolls before the worker nodes, and worker nodes do not yet have IP SANs, upgrade breaks if InternalIP is used.
IPv6 never worked with hostnames, so there is no BC break there.


yurrriq:
Confirmed this fixed the broken metrics-server on a kOps 1.22.1 / k8s 1.21.5 cluster of ours.

PR #12635, opened 2021-10-29
Check for hardcoded partitions in integration tests

rifelpet:
This should catch issues like #12633 earlier. Note that this will fail until the bug in question has been fixed.

Our integration tests use an `aws-test` partition in any ARNs, so there should never be an `arn:aws:` occurrence in the outputs.


rifelpet:
This is passing now

/cc @hakman

hakman (approved):

PR #12637, opened 2021-10-29
Integration test cleanup

rifelpet:
Some integration tests were missing their cloudformation invocations, so their cloudformation.json wasn't being updated during ./hack/update-expected.sh.

We also had some orphaned files in the integration test source directories that shouldn't be there but we didnt assert that the number of actual source data files matches the expected number of data files. This adds an assertion for that and removes the orphaned files.


hakman (approved):
👍

johngmyers:
@rifelpet `hack/update-expected.sh` is no longer creating missing golden files.

rifelpet:
Ok I'll take a look

johngmyers:
Never mind, it was something else

PR #12638, opened 2021-10-29
Fix hardcoded ARN partitions

### LINKED ISSUE #12633: Partition "aws" is not valid for resource "arn:aws:ec2:*:*:volume/*".

rifelpet:
IAMBuilder was hardcoding the partition in ARNs to `aws` in IAM Policy documents for OIDC Providers, EBS Volumes, and EBS Snapshots. This fixes them to use the current partition (`aws-test` in integration tests). This also fixes a few ARNs generated in cloudmock and updates any ARNs referenced in cluster manifests as well.

The end goal is that there are no ARNs with `aws` partitions in any of the integration test inputs or outputs so they can be validated for support for alternative partitions via #12635.

Fixes #12633

### DIFF
--- a/pkg/model/components/addonmanifests/awscloudcontrollermanager/iam.go
+++ b/pkg/model/components/addonmanifests/awscloudcontrollermanager/iam.go
@@ -32,7 +32,7 @@ var _ iam.Subject = &ServiceAccount{}
 func (r *ServiceAccount) BuildAWSPolicy(b *iam.PolicyBuilder) (*iam.Policy, error) {
 	clusterName := b.Cluster.ObjectMeta.Name
 	p := iam.NewPolicy(clusterName)
-	iam.AddCCMPermissions(p, b.Cluster.Spec.Networking.Kubenet != nil)
+	iam.AddCCMPermissions(p, b.Partition, b.Cluster.Spec.Networking.Kubenet != nil)
 	iam.AddLegacyCCMPermissions(p)
 	return p, nil
 }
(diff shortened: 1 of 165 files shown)


pkg/model/components/addonmanifests/awscloudcontrollermanager/iam.go
```diff
@@ -32,7 +32,7 @@ var _ iam.Subject = &ServiceAccount{}
 func (r *ServiceAccount) BuildAWSPolicy(b *iam.PolicyBuilder) (*iam.Policy, error) {
 	clusterName := b.Cluster.ObjectMeta.Name
 	p := iam.NewPolicy(clusterName)
-	iam.AddCCMPermissions(p, b.Cluster.Spec.Networking.Kubenet != nil)
+	iam.AddCCMPermissions(p, b.Partition, b.Cluster.Spec.Networking.Kubenet != nil)
```
johngmyers, line 35: Would it make sense to pass the partition into `NewPolicy()` instead of having to pass it to every `Add*Permissions()` call?

johngmyers (approved):

PR #12639, opened 2021-10-29
Log `kops toolbox dump` output to artifacts dir rather than stdout

rifelpet:
This should reduce some of the verbose clutter in the job logs during cluster teardown

/cc @hakman


hakman (approved):
Much better. Thanks @rifelpet :).

PR #12640, opened 2021-10-30
dns-controller: use aliases for internal host-network pods

johngmyers:
/kind bug

In IPv6-only clusters the api.internal DNS record is sometimes, upon cluster bootstrap, incorrectly set to the control plane node's IPv4 address.

This is reported to be a race between the kube-apiserver pod recording its podIPs and the CCM assigning an IPv6 address to the node. Sometimes the kube-apiserver host-network pod has its podIP set to the node's IPv4 address instead of its IPv6 address. dns-controller then takes this incorrect podIP and publishes it in the api.internal DNS record.

This changes dns-controller to instead use its `_alias` mechanism to take the addresses from the `InternalIP`s of the node resource of the pod's node. It additionally filters the `InternalIP`s to the address families of the cluster.

This should fix the `ServiceAccountIssuerDiscovery` e2e test failures when an IPv6 cluster is not using an internal load balancer.

### DIFF
--- a/dns-controller/cmd/dns-controller/main.go
+++ b/dns-controller/cmd/dns-controller/main.go
@@ -54,6 +54,7 @@ func main() {
 	fmt.Printf("dns-controller version %s\n", BuildVersion)
 	var dnsServer, dnsProviderID, gossipListen, gossipSecret, watchNamespace, metricsListen, gossipProtocol, gossipSecretSecondary, gossipListenSecondary, gossipProtocolSecondary string
 	var gossipSeeds, gossipSeedsSecondary, zones []string
+	var internalIpv4, internalIpv6 bool
 	var watchIngress bool
 	var updateInterval int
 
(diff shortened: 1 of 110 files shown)


dns-controller/cmd/dns-controller/main.go
```diff
@@ -73,6 +74,8 @@ func main() {
 	flag.StringVar(&gossipListenSecondary, "gossip-listen-secondary", fmt.Sprintf("0.0.0.0:%d", wellknownports.DNSControllerGossipMemberlist), "address:port on which to bind for gossip")
 	flags.StringVar(&gossipSecretSecondary, "gossip-secret-secondary", gossipSecret, "Secret to use to secure gossip")
 	flags.StringSliceVar(&gossipSeedsSecondary, "gossip-seed-secondary", gossipSeedsSecondary, "If set, will enable gossip zones and seed using the provided addresses")
+	flags.BoolVar(&ipv4, "ipv4", ipv4, "Internal network has IPv4")
+	flags.BoolVar(&ipv6, "ipv6", ipv4, "Internal network has IPv6")
```
hakman, line 78: Maybe good to be more specific?
```suggestion
	flags.BoolVar(&ipv4, "internal-ipv4", internal-ipv4, "Internal network has IPv4")
	flags.BoolVar(&ipv6, "internal-ipv6", internal-ipv6, "Internal network has IPv6")
```

dns-controller/cmd/dns-controller/main.go
```diff
@@ -85,6 +88,18 @@ func main() {
 	flags.AddGoFlagSet(flag.CommandLine)
 	flags.Parse(os.Args)
 
+	var internalRecordTypes []dns.RecordType
+	if ipv4 {
```
hakman, line 92: We could just default to ipv4.
```suggestion
	if ipv4 || !ipv6 {
```

hakman (changes requested):

dns-controller/cmd/dns-controller/main.go
johngmyers: I think it'd be more worthwhile to catch cases where this information has been omitted.

dns-controller/cmd/dns-controller/main.go
```diff
@@ -54,6 +54,7 @@ func main() {
 	fmt.Printf("dns-controller version %s\n", BuildVersion)
 	var dnsServer, dnsProviderID, gossipListen, gossipSecret, watchNamespace, metricsListen, gossipProtocol, gossipSecretSecondary, gossipListenSecondary, gossipProtocolSecondary string
 	var gossipSeeds, gossipSeedsSecondary, zones []string
+	var internal_ipv4, internal_ipv6 bool
```
hakman, line 57: Shouldn't this be camelCase?

dns-controller/cmd/dns-controller/main.go
hakman: 👍

hakman:
Looks good. Do you want to squash a some stuff, before merging?

PR #12641, opened 2021-10-30
Allow kops-controller to describe network interfaces

hakman:
Still needed by kops-controller for cloud IPAM:
https://testgrid.k8s.io/kops-ipv6#kops-grid-scenario-ipv6-cilium-cloudipam
https://github.com/kubernetes/kops/blob/fa29c9a6b29f21becb9453f204759f6a07a644f0/cmd/kops-controller/controllers/awsipam.go#L127

/cc @johngmyers @rifelpet @olemarkus


johngmyers (approved):

johngmyers:
At some point we should start carving out the permissions for the various control plane service roles. But this particular permission is pretty low-value.

PR #12642, opened 2021-10-30
Enable Router Advertisements for Debian 11

hakman:
Test run using Calico and Debian 11 here:
https://testgrid.k8s.io/kops-ipv6#pull-kops-e2e-ipv6-conformance

### DIFF
--- a/nodeup/pkg/model/sysctls.go
+++ b/nodeup/pkg/model/sysctls.go
@@ -151,6 +151,11 @@ func (b *SysctlBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	if b.Cluster.Spec.IsIPv6Only() {
+		if b.Distribution == distributions.DistributionDebian11 {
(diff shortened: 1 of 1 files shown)


johngmyers (approved):

nodeup/pkg/model/sysctls.go
```diff
@@ -151,6 +151,11 @@ func (b *SysctlBuilder) Build(c *fi.ModelBuilderContext) error {
 	}
 
 	if b.Cluster.Spec.IsIPv6Only() {
+		if b.Distribution == distributions.DistributionDebian11 {
+			sysctls = append(sysctls,
+				"# Enable Router Advertisements to get the default IPv6 route",
+				"net.ipv6.conf.ens5.accept_ra=2")
```
olemarkus, line 157: Just for the record, the interface name depends the instance type. On e.g t2, this will be ens3. On upcoming instance types it can again be something else.

PR #12649, opened 2021-10-30
Add initial IPv6 documentation

johngmyers:


docs/networking/ipv6.md
```diff
@@ -0,0 +1,35 @@
...
+# CNI
+
+kOps currently supports IPv6 on Calico, Cilium, and bring-your-own CNI only.
+
+CNIs must not masquerade IPv6 addresses.
+
+## Calico
+
+Running IPv6 with Calico requires a Debian 11-based AMI. As of the writing of this document, Ubuntu does not work due to an 
+[issue with systemd's handling of AWS's incorrect DHCP responses](https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1945888).
```
johngmyers, line 31: Do we have a public ticket reference for the AWS DHCP bug?  Do we want to describe it in the "Cloud providers" section?

hakman: Not that I know of.

docs/networking/ipv6.md
```diff
@@ -0,0 +1,35 @@
...
+## CNI
+
+kOps currently supports IPv6 on Calico, Cilium, and bring-your-own CNI only.
+
+CNIs must not masquerade IPv6 addresses.
+
+### Calico
+
+Running IPv6 with Calico requires a Debian 11-based AMI. As of the writing of this document, Ubuntu does not work due to an 
+[issue with systemd's handling of AWS's incorrect DHCP responses](https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1945888).
```
hakman, line 31: https://github.com/systemd/systemd/issues/20803 may be a better reference.

johngmyers: I was trying to go for something Ubuntu-specific, but I suppose AL2 would probably have the same issue.

hakman: I don't know if AL2 uses systemd-networkd. Each distro uses something different. I Just mention Debian 11 if it works with Cilium, and mention that Ubuntu 22.04 should work too.

johngmyers: Perhaps you could refine the language after the PR lands? I don't have complete enough knowledge of all the possibilities.

docs/networking/ipv6.md
```diff
@@ -0,0 +1,37 @@
+# IPv6
+
+{{ kops_feature_table(kops_added_ff='1.22') }}
```
hakman, line 3: Best not to confuse people:
```suggestion
{{ kops_feature_table(kops_added_ff='1.23') }}
```

docs/networking/ipv6.md
```diff
@@ -0,0 +1,37 @@
...
+
+kOps has experimental support for configuring clusters with IPv6-only pods and dual-stack nodes.
```
hakman, line 5: Dual-stack should probably not get mentioned.
```suggestion
kOps has experimental support for configuring clusters with IPv6-only pods on AWS instances with IPv6 enabled.
```

hakman (changes requested):

docs/networking/ipv6.md
hakman: 👍

docs/networking/ipv6.md
johngmyers: We did feature flag it in 1.22

docs/networking/ipv6.md
johngmyers: The nodes are dual-stack.

docs/networking/ipv6.md
hakman: Yes, but it's a totally different approach that we don't want to document or support. Mentioning 1.22 would confuse some users in thinking that it's same as what we will release in 1.23. Am I wrong?

johngmyers: I suppose it depends on what the table is supposed to mean and how people interpret "experimental". I suspect there's a good chance we'll remove the feature flag for 1.23.

hakman: I agree that the feature flag will probably be removed in 1.23. Though, kOps docs don't have a way to say, experimental in version X, beta in version Y, stable in version Z. I am afraid people will try this now with 1.22 and get something different than described here.

hakman (approved):

PR #12658, opened 2021-11-01
Start IPv6 CIDR numbering from 0

hakman:
/cc @johngmyers

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -973,8 +973,7 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
 			if api.CloudProviderID(cluster.Spec.CloudProvider) == api.CloudProviderAWS {
 				klog.Warningf("IPv6 support is EXPERIMENTAL and can be changed or removed at any time in the future!!!")
 				for i := range cluster.Spec.Subnets {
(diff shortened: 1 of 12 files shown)


upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -969,12 +969,13 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
 
 		if opt.IPv6 {
 			cluster.Spec.NonMasqueradeCIDR = "::/0"
-			cluster.Spec.ExternalCloudControllerManager = &api.CloudControllerManagerConfig{}
+			cluster.Spec.ExternalCloudControllerManager = &api.CloudControllerManagerConfig{
+				CloudProvider: "aws",
```
johngmyers, line 973: Can we do this through defaulting the value downstream?

hakman: We could just replace `eccm.CloudProvider` with `cluster.Spec.CloudProvider` here:
https://github.com/kubernetes/kops/blob/3fd7b446c06a04d9c133bb37c63c12b3ae560415/pkg/model/components/awscloudcontrollermanager.go#L41-L43

hakman: Does it look better?

johngmyers (approved):

PR #12659, opened 2021-11-01
Use new DescribeAutoScalingGroups tag filter support

rifelpet:
https://github.com/aws/aws-sdk-go/releases/tag/v1.41.3

This eliminates the `autoscaling.DescribeTags` call entirely and may reduce the `DescribeAutoScalingGroups` call volume due to previously insufficient tag filtering support in `DescribeTags` whose output was passed to `DescribeAutoScalingGroups`.

I confirmed this support has been rolled out to GovCloud. I have not confirmed whether it is in the iso or cn partitions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/aliyun/alibaba-cloud-sdk-go v1.61.1059
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/amazon-ec2-instance-selector/v2 v2.0.2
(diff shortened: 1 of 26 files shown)


olemarkus (approved):
/hold

Not sure if we want to hold this one or not waiting for confirmation. I am in favour of merging though.

hakman:
Until https://github.com/kubernetes/kops/issues/12184 is addressed, to not confuse people.
/lgtm cancel

Issue #12630, opened 2021-10-29
Kops 1.22.1 not adding spec.image field to instance groups


jim-barber-he:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

```
Version 1.22.1 (git-da710c931fbfe0fc8fe7bb3e2d41ba6e09c899f9)
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
 
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.3", GitCommit:"c92036820499fedefec0f847e2054d824aea6cd1", GitTreeState:"clean", BuildDate:"2021-10-27T18:41:28Z", GoVersion:"go1.16.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.3", GitCommit:"c92036820499fedefec0f847e2054d824aea6cd1", GitTreeState:"clean", BuildDate:"2021-10-27T18:35:25Z", GoVersion:"go1.16.9", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**

Amazon AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

We create our instance groups using the `kops create -f` command using manifests without the `spec.image` field set.
For example the manifest for one of our masters on a test cluster looks like so:
```yaml
---
apiVersion: kops.k8s.io/v1alpha2
kind: InstanceGroup
metadata:
  labels:
    kops.k8s.io/cluster: test3.he0.io
  name: master-ap-southeast-2a
spec:
  additionalSecurityGroups:
  - sg-0a63fd81e986e756f
  additionalUserData:
  - content: |
      #!/bin/sh
      /usr/bin/apt-get update -qq
      /usr/bin/unattended-upgrade
    name: apply-security-updates
    type: text/x-shellscript
  associatePublicIp: false
  cloudLabels:
    he:Application: kubernetes
    he:Component: kube-master
    he:EnvironmentName: test3
    he:EnvironmentType: test
  detailedInstanceMonitoring: true
  instanceMetadata:
    httpPutResponseHopLimit: 3
    httpTokens: required
  machineType: t3.large
  maxSize: 1
  minSize: 1
  nodeLabels:
    kops.k8s.io/instancegroup: master-ap-southeast-2a
[...]

PR #12660, opened 2021-11-01
Validate IGs more strictly after defaults have applied

olemarkus:
This commit will ensure IGs are validated in clientset prior to write similar to clusters. Also introduces strict flag similar to cluster, which only validate values where we have defaults after defaults have been applied.

This is a step towards fixing https://github.com/kubernetes/kops/issues/12630. I suspect somewhere we removed a call to `PopulateInstanceGroupSpec`, which allows IGs with bad values to be written.

I am not sure it is okay to do validation within the clientset. Seems a bit wrong, and we do not know the `Cloud` implementation there. But we already did this for `Cluster`, so there was precedence.

Also, we do validation in _many_ locations.


johngmyers (approved):

pkg/apis/kops/validation/aws_test.go
```diff
@@ -21,12 +21,14 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+
```
hakman, line 24: These changes are quite messy. At least it should be arranged somehow.

cmd/kops/create.go
```diff
@@ -26,14 +26,15 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
+	"k8s.io/kubectl/pkg/util/i18n"
+	"k8s.io/kubectl/pkg/util/templates"
+
```
hakman, line 31: No actual changes to this file.

hakman (changes requested):
Could you clean the imports a little before merging?

hakman (approved):

PR #12661, opened 2021-11-01
[WIP] Test IPv6 Cilium setup with Debian 11

hakman:


hakman:
Ref. https://github.com/kubernetes/test-infra/pull/24218

PR #12657, opened 2021-11-01
WIP Support IPv6 in AWS VPC CNI

johngmyers:

PR #12665, opened 2021-11-02
Don't create placeholder A record on IPv6 clusters without API LB

rifelpet:
dns-controller updates this record to the node IPs, but in this case nodes wont have any IPv4 addresses so dns-controller never updates the record from the placeholder value, causing `validate cluster` to fail (even if the AAAA placeholder is updated)

I noticed this in [this job](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/12657/pull-kops-e2e-ipv6-amazonvpc/1455030868530696192) (ref: https://github.com/kubernetes/kops/pull/12657#issuecomment-956195781)

`kops validate cluster` fails with the error message containing the IPv4 placeholder IP which shouldn't be relevant to an IPv6 cluster:

`The dns-controller Kubernetes deployment has not updated the Kubernetes cluster's API DNS entry to the correct IP address.  The API DNS IP address is the placeholder address that kops creates: 203.0.113.123.  Please wait about 5-10 minutes for a master to start, dns-controller to launch, and DNS to propagate.  The protokube container and dns-controller deployment logs may contain more diagnostic information.  Etcd and the API DNS entries must be updated for a kops Kubernetes cluster to start.`

`kops delete cluster` reports cleaning up these records, notice the A record for the API name:

```
route53-record		api.e2e-09c66f4ca2-5e791.test-cncf-aws.k8s.io.								ZEMLNXIIWQ0RV/A/api.e2e-09c66f4ca2-5e791.test-cncf-aws.k8s.io.
route53-record		api.e2e-09c66f4ca2-5e791.test-cncf-aws.k8s.io.								ZEMLNXIIWQ0RV/AAAA/api.e2e-09c66f4ca2-5e791.test-cncf-aws.k8s.io.
route53-record		api.internal.e2e-09c66f4ca2-5e791.test-cncf-aws.k8s.io.							ZEMLNXIIWQ0RV/AAAA/api.internal.e2e-09c66f4ca2-5e791.test-cncf-aws.k8s.io.
route53-record		kops-controller.internal.e2e-09c66f4ca2-5e791.test-cncf-aws.k8s.io.					ZEMLNXIIWQ0RV/AAAA/kops-controller.internal.e2e-09c66f4ca2-5e791.test-cncf-aws.k8s.io. 
```

[dns-controller logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/12657/pull-kops-e2e-ipv6-amazonvpc/1455030868530696192/artifacts/cluster-info/kube-system/dns-controller-58dc6bc8b9-vrb2c/logs.txt) report adding AAAA record but not any A record.

Neither the [apiserver pod](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/12657/pull-kops-e2e-ipv6-amazonvpc/1455030868530696192/artifacts/cluster-info/kube-system/pods.yaml) nor the [nodes](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/12657/pull-kops-e2e-ipv6-amazonvpc/1455030868530696192/artifacts/cluster-info/nodes.yaml) have any IPv4 addresses.

### DIFF
--- a/upup/pkg/fi/cloudup/dns.go
+++ b/upup/pkg/fi/cloudup/dns.go
@@ -262,14 +262,8 @@ func buildPrecreateDNSHostnames(cluster *kops.Cluster) []recordKey {
 	if cluster.Spec.MasterPublicName != "" && !hasAPILoadbalancer {
 		recordKeys = append(recordKeys, recordKey{
 			hostname: cluster.Spec.MasterPublicName,
-			rrsType:  rrstype.A,
+			rrsType:  internalType,
 		})
-		if internalType != rrstype.A {
-			recordKeys = append(recordKeys, recordKey{
-				hostname: cluster.Spec.MasterPublicName,
-				rrsType:  internalType,
-			})
-		}
 	}
 
 	if cluster.Spec.MasterInternalName != "" && !useLBForInternalAPI {
--- a/upup/pkg/fi/cloudup/dns_test.go
+++ b/upup/pkg/fi/cloudup/dns_test.go
@@ -45,7 +45,6 @@ func TestPrecreateDNSNames(t *testing.T) {
 				},
 			},
 			expected: []recordKey{
-				{"api.cluster1.example.com", rrstype.A},
 				{"api.cluster1.example.com", rrstype.AAAA},
 				{"api.internal.cluster1.example.com", rrstype.AAAA},
 			},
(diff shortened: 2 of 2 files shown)


johngmyers:
In this case nodes are dual-stack, so have an IPv4 address. This will cause delays until the SOA TTL expires when the kops client only has IPv4 connectivity to the control plane.

rifelpet:
ah, so the problem is that the IPv4 addresses aren't being added to the node objects? I see the AWS CCM isnt running and the [job's cluster spec](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/12657/pull-kops-e2e-ipv6-amazonvpc/1455030868530696192/artifacts/cluster.yaml) doesn't have external CCM enabled. Perhaps we should default to enabling external CCM on IPv6 AWS clusters, given that it is required for node ipam.

johngmyers:
The problem, as I noted in the PR, is the CNI daemonset pod refused to start.

A problem with enabling the external CCM was recently fixed in #12658.

johngmyers:
If we ever support a cloud provider with IPv6-only instances, then the placeholder creation code will need to know whether or not the control plane nodes are going to be dual-stack.

rifelpet:
That makes sense to me, thanks for the clarification

/close

PR #12667, opened 2021-11-02
Fix error handling in kubetest2 dumplogs

rifelpet:
Make directories prior to creating files in them. Also cleanup some error handling so that the remaining teardown steps can run even if dumplogs fails.

Add --request-timeout to some kubectl commands to shorten wait times when the cluster isnt up


justinsb:
Thanks @rifelpet - all looks good!

/approve
/lgtm

PR #12676, opened 2021-11-03
Migrate to AWS CCM in k8s 1.24

johngmyers:
This is a proposal for how to migrate existing AWS clusters to external CCM

* In k8s 1.23 we default to enabling [leader migration](https://cloud-provider-aws.sigs.k8s.io/getting_started/#using-leader-migration) if using KCM on AWS
* In k8s 1.24 and higher we only support CCM for AWS.
* In k8s 1.24 we default to enabling leader migration on AWS.
* In k8s 1.25 and higher we default to disabling leader migration.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -522,6 +522,9 @@ spec:
                     description: ConfigureCloudRoutes enables CIDRs allocated with
                       to be configured on the cloud provider.
                     type: boolean
+                  enableLeaderMigration:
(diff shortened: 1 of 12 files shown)


olemarkus:
I am all for this. We need to see what is holding AWS CCM away from being stable. As I understand it, it is generally testing, but we have fairly good test coverage for kOps purpose on our side. Cc @nckturner 

/lgtm

johngmyers:
I don't think there's any particular reason to release note this for 1.23. We can change our mind about the CCM default in the 1.24 cycle if need be.

hakman (approved):

PR #12677, opened 2021-11-03
Fix that states AWS IAM Instance Profile blocks IAM Role

angeloskaltsikis:
According to [aws-cli docs](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/iam/delete-role.html#examples),
it is needed to delete any Instance profile that uses a role before deleting
the actual role. This fix adds a "blocks" statement to the IAM Instance Profile,
to declare that it should block the IAM Role deletion.


rifelpet (approved):
thanks for this!

PR #12674, opened 2021-11-03
Remove tag conditions on certain AWS IAM actions

rifelpet:
I checked these against the IAM docs for each API and moved the actions that dont support tag conditions:

* https://docs.aws.amazon.com/service-authorization/latest/reference/list_elasticloadbalancing.html#elasticloadbalancing-actions-as-permissions
* https://docs.aws.amazon.com/service-authorization/latest/reference/list_elasticloadbalancingv2.html#elasticloadbalancingv2-actions-as-permissions
* https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html#amazonec2-actions-as-permissions
* https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2autoscaling.html#amazonec2autoscaling-actions-as-permissions

If someone could double check all these changes that would be appreciated.

Fixes #12671

Issue #12671, opened 2021-11-02
Unable to create new loadbalancers in AWS


zetaab:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Version 1.23.0-alpha.2 (git-3a97dbaa8d754d6f3318890c3e9159bd5ea1282c)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

```
% kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:45:37Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.2", GitCommit:"8b5a19147530eaac9476b0ab82980b4088bbc1b2", GitTreeState:"clean", BuildDate:"2021-09-15T21:32:41Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

```
kops create cluster --name teestiing.k8s.local --zones eu-north-1a,eu-north-1b,eu-north-1c --master-count=3 --node-count=1 --node-size t3.small --master-size t3.small
kops update cluster teestiing.k8s.local --yes --admin=87600h
```

Then created 2 loadbalancers with proxy protocol and without:

```
apiVersion: v1
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "60"
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
  labels:
    app.kubernetes.io/name: nginx
  name: nginx-proxy
spec:
  externalTrafficPolicy: Local
  ports:
  - name: http
    port: 80
    targetPort: http
  selector:
    app.kubernetes.io/name: nginx
  type: LoadBalancer

---
apiVersion: v1
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "60"
  labels:
    app.kubernetes.io/name: nginx
  name: nginx
spec:
[...]

PR #12680, opened 2021-11-04
Fix ELB IAM conditions (part 2)

rifelpet:
This mostly reverts https://github.com/kubernetes/kops/pull/12674

I did some manual testing of all of these API actions and confirmed they all work even with `aws:ResourceTag` conditions even though the AWS documentation linked in the previous PR isn't particularly clear about the condition support for these actions.

The only changes that are needed are the classic ELB `CreateLoadBalancerListeners` and `CreateLoadBalancerPolicies` actions which were using `aws:RequestTag` but need to use `aws:ResourceTag`. This fixes one of the two problems reported in https://github.com/kubernetes/kops/issues/12671.

I wasn't able to reproduce the `ModifyLoadBalancerAttributes` error, I was successfully able to make those calls. @zetaab I see in the log you included that the PermissionDenied event was only 18 seconds old. I'm wondering if this call eventually succeeded for you, perhaps due to eventual consistency with the newly created ELB and its tags.

#12674 wasn't yet cherry-picked to release-1.22, so we can cherry-pick them both together once this lands and we confirm no changes are needed to `ModifyLoadBalancerAttributes`


olemarkus:
The ELB permissions were fixed in https://github.com/kubernetes/kops/pull/12611

It is true that there is a temporary permission denied that is logged that comes from trying to modify the ELB before it has been tagged (KCM doesn't do tag on create, which is what is causing some of the permissions failures), but once the tags are there the calls will succeed.

olemarkus (approved):

zetaab:
@rifelpet I compiled latest master yesterday and I see following again in our production:

```
 Warning event found: Error syncing load balancer: failed to ensure load balancer: error creating proxy protocol policy on load balancer: "AccessDenied: User: arn:aws:sts::xxxx:assumed-role/masters.updateawspr-12b8c8.k8s.local/i-0a0878c4875c6ab4b is not authorized to perform: elasticloadbalancing:CreateLoadBalancerPolicy on resource: arn:aws:elasticloadbalancing:eu-central-1:xxxx:loadbalancer/a1d884520d9e349b7bc635c7297654b0 because no identity-based policy allows the elasticloadbalancing:CreateLoadBalancerPolicy action\n\tstatus code: 403, request id: d6f87029-ee59-4fd8-bf62-ad6572a1a920"
 ```

So the problem is that when creating service type: loadbalancer in kubernetes. It might be that those loadbalancers do NOT have the cluster tag added. That is why there is no permissions to create those.

Basically it means that people cannot use proxy protocol loadbalancers currently with master branch

zetaab:
@rifelpet I opened #12671 because its not possible to create normal elbs currently

Issue #12678, opened 2021-11-03
Support allow_ip_forwarding flag in Calico


kuldazbraslav:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
I'd like kops to support [container_settings.allow_ip_forwarding](https://docs.projectcalico.org/reference/cni-plugin/configuration#container-settings) in [CalicoNetworkingSpec](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#CalicoNetworkingSpec).

The flag is set to false by default and sets `net.ipv4.ip_forward` / `net.ipv6.conf.all.forwarding` no matter what `securityContext.sysctl` in PodSpec says.

The workaround for would be using a privileged initContainer that sets the sysctl from command line - but that definitely isn't preferred option. The other option is overwriting calico-config ConfigMap which requires custom templating.

**2. Feel free to provide a design supporting your feature request.**
This should be easy to implement in https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template

zhengtianbao:
It's looks like appropriate for new comming, can I take this issue? :smile:

PR #12682, opened 2021-11-04
[calico] Add support for allow_ip_forwarding field

zhengtianbao:
This is a needed configuration option for calico users that want to set **allow_ip_forwarding** within the container namespace. 

Ref issue: #12678


hakman:
Thank you for your contribution @kuldazbraslav. This looks good to me.
@kuldazbraslav Is this what you were looking for?

hakman (approved):

kuldazbraslav:
@hakman Yes, this seems quite like what I was looking for. However after changing this on an existing cluster, the DaemonSet should be rolled over - is it the case?

Sorry for reacting so late, I totally missed this PR.

PR #12687, opened 2021-11-05
Add option use chrony for synchronizing time in Ubuntu 20.04

hakman:
Ubuntu is defaulting to `chrony` in latest versions and this was requested in the past. Will make it easier to make time synchronization more configurable in the future.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -5156,6 +5156,10 @@ spec:
               ntp:
                 description: NTPConfig is the configuration for NTP.
                 properties:
+                  chrony:
+                    description: Chrony replaces timesyncd for Ubuntu 20.04 and older
+                      when set to true.
+                    type: boolean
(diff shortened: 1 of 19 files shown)


k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -5156,6 +5156,10 @@ spec:
               ntp:
                 description: NTPConfig is the configuration for NTP.
                 properties:
+                  chrony:
```
justinsb, line 5159: Maybe a string `provider: chrony` would be more extensible?  (And it could serve as a "discriminated union" field https://github.com/kubernetes/community/issues/229 )

hakman: I was hoping to remove it later. It is only useful for Ubuntu 20.04 that will be removed in the next few years. I doubt anything exciting will happen with NTP by then.

hakman:
Dropping this per discussion during kOps office hours.

PR #12692, opened 2021-11-07
Watch Ingress by default when using the external-dns provider

johngmyers:
When using dns-controller, it's likely that there would be an externally installed external-dns creating DNS records for ingresses, so in order to avoid two controllers fighting over the same domains the default of `false` for `watchIngress` makes sense.

With a kops-managed external-dns, it makes more sense to enable `watchIngress` by default.


olemarkus (approved):
I think this is what most users would expect as well. I am fine with this change.

PR #12710, opened 2021-11-11
gce: don't over-warn on ManagedInstanceGroups filtering

justinsb:
It does seem that filtering is not yet implemented on this call.
Don't over-warn.

Also introduce pagination, because without a filter we might paginate.


rifelpet (approved):
for reference, this removes the warning noise in the kops-controller logs here: https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-gce-latest/1458748400299151360/artifacts/cluster-info/kube-system/kops-controller-6j5tk/logs.txt

PR #12713, opened 2021-11-11
set calico-node readiness/liveness timeout to 10s

estahn:
Provide calico-node with more time to come up.


JoshVee (approved):

hakman (approved):
Thanks @estahn! 👍

PR #12722, opened 2021-11-12
refactor: move from io/ioutil to io and os packages

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since kops has been upgraded to Go 1.17 in #12505, this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.


hakman (approved):
Thanks @Juneezee! :)

PR #12724, opened 2021-11-13
Upgrade external-dns to 0.10.1 for Kubernetes >= 1.19

johngmyers:
/cc @rifelpet

### DIFF
--- a/upup/models/cloudup/resources/addons/external-dns.addons.k8s.io/k8s-1.19.yaml.template
+++ b/upup/models/cloudup/resources/addons/external-dns.addons.k8s.io/k8s-1.19.yaml.template
@@ -1,0 +1,141 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: external-dns
+  namespace: kube-system
+  labels:
+    k8s-addon: external-dns.addons.k8s.io
+    k8s-app: external-dns
+    version: v0.10.1
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      k8s-app: external-dns
+  template:
+    metadata:
+      labels:
+        k8s-addon: external-dns.addons.k8s.io
+        k8s-app: external-dns
+        version: v0.10.1
+    spec:
+      priorityClassName: system-cluster-critical
+      serviceAccountName: external-dns
+      securityContext:
+        fsGroup: 65534
+      tolerations:
+      - key: "node-role.kubernetes.io/master"
+        effect: NoSchedule
+      - key: "node.kubernetes.io/not-ready"
+        effect: NoSchedule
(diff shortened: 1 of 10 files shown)


upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
```diff
@@ -490,8 +490,13 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*Addon
 			key := "external-dns.addons.k8s.io"
 
 			{
-				location := key + "/k8s-1.12.yaml"
-				id := "k8s-1.12"
+				location := key + "/k8s-1.19.yaml"
+				id := "k8s-1.19"
+
+				if !b.IsKubernetesGTE("1.19") {
```
hakman, line 496:
```suggestion
				if b.IsKubernetesLT("1.19") {
```

johngmyers:
The IRSA e2e test will hopefully be fixed by kubernetes/test-infra#24341

upup/models/cloudup/resources/addons/external-dns.addons.k8s.io/k8s-1.19.yaml.template
```diff
@@ -0,0 +1,141 @@
...
+          initialDelaySeconds: 5
+          periodSeconds: 10
+          timeoutSeconds: 5
+          failureThreshold: 6
+          successThreshold: 1
+        resources:
+          requests:
+            cpu: 50m
+            memory: 50Mi
+---
```
olemarkus, line 79: What is this service used for?

johngmyers: I believe it is for pulling metrics

olemarkus: Do we want to expose this? Usually we don't add services just for metrics as it is not given this is the way users want to scrape metrics.

hakman (approved):

upup/models/cloudup/resources/addons/external-dns.addons.k8s.io/k8s-1.19.yaml.template
johngmyers: The upstream chart unconditionally creates this service.

olemarkus: Upstream does not always do the right thing though. but well .. it's in now anyway.

PR #12727, opened 2021-11-13
addons: support for kopeio-networking addon

justinsb:
Adding support for the kopeio-networking addon.
    
We load the operator manifest (which changes rarely) from the channels directory for now.  We follow the same structure as operators themselves use so that we can support other backends in future.
    
The channels file includes the current versions of the operators.
    
During cluster creation, we create these additional objects.

All changes are feature-flagged behind the UseAddonOperators feature flag


justinsb:
Note that this builds on #12544

pkg/apis/kops/channel.go
```diff
@@ -352,3 +364,30 @@ func (c *Channel) HasUpstreamImagePrefix(image string) bool {
...
+
+		matches = append(matches, pkg)
+	}
+
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("found no packages in channel for name=%q", name)
+	}
+
+	if len(matches) != 1 {
+		return nil, fmt.Errorf("found multiple packages in channel for name=%q", name)
```
rifelpet, line 386: Is the intent that we'll only ever have one package per addon? what if we end up needing to support per-k8s-version packages, similar to bootstrapchannelbuilder now?

pkg/wellknownoperators/operators.go
```diff
@@ -38,65 +40,183 @@ type Builder struct {
...
-	var addons []*WellKnownAddon
-	var crds kubemanifest.ObjectList
+	var packages []*Package
+	var keepObjects kubemanifest.ObjectList
+
+	for _, object := range objects {
+		keep := true
+
+		// We may in future move ClusterPackage to the server-side,
+		// however remapping it client-side allow us to use our existing manifest logic
```
rifelpet, line 55: Does this support rendering manifest templates using values from the cluster spec? (or a different API type?) how do we plan to offer the equivalent level of customizability that we have now with addons and their api types?

pkg/apis/kops/channel.go
justinsb: Good point.  So the channel has two versions per "bundled" addon ... the operator version, and the version we should inject into the CRD.   The operator version, yes, I expect that we will resolve that to a single version.  We don't currently have it, but we could have kubernetesVersion selectors and/or kopsVersion selectors in the channels file.  There's probably a good case for adding per-k8s-version now; ideally these operators wouldn't be specific to kOps versions though so I'm not sure of the case for a per-kops version.

pkg/wellknownoperators/operators.go
justinsb: So generally my hope/expectation is that we don't customize the operator itself; by design it's supposed to be very generic.

The CRD-instance _would_ offer the customization.  Because that CRD is an operator-specific CRD (e.g. CoreDNS or KopeioNetworking) it has a specific schema for that customization.  Eventually I hope we get to a world where the user configures those addon CRD instances without kops even being aware of what's going on.

Where today we have configuration in the Cluster, we have to map that to a CRD.  Eventually we can have a migration tool or similar to get the values out of the Cluster, but today we can have the values specified in the Cluster generate the CRDs.  We don't have that yet / in this PR.

Where the CRD / operator has dependencies on other objects, the hope is that it can get those from the cluster itself.  For example, today the CoreDNS operator needs the DNS service IP address, and so the operator can look up the kube-dns service to default the value.

pkg/apis/kops/channel.go
justinsb: I added filtering support for kOps & k8s version - it's much hard to add later.  Thanks for the suggestion @rifelpet

justinsb:
Rebased and added filtering for k8s / kops versions, as suggested by @rifelpet , because it's much hard to add later!

This is all behind a feature flag .... how do people feel about it?

rifelpet:
/lgtm
Perhaps someone else can take a look for final approval

hakman:
👍 
/approve

PR #12731, opened 2021-11-14
[DigitalOcean] Show load balancer id in logs

srikiz:
The [e2e tests](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#e2e-kops-do-calico) for DO are failing to create a load balancer. I haven't been able to reproduce it locally, so adding the load balancer ID in logs for further troubleshooting.

FYI - @timoreimann

### DIFF
--- a/upup/pkg/fi/cloudup/dotasks/loadbalancer.go
+++ b/upup/pkg/fi/cloudup/dotasks/loadbalancer.go
@@ -129,6 +129,8 @@ func (_ *LoadBalancer) RenderDO(t *do.DOAPITarget, a, e, changes *LoadBalancer) 
 		HealthyThreshold:       5,
 	}
 
+	klog.Infof("In RenderDO: loadbalancer")
+
 	// check if load balancer exist.
 	loadBalancers, err := t.Cloud.GetAllLoadBalancers()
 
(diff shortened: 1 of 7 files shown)


upup/pkg/fi/cloudup/dotasks/loadbalancer.go
```diff
@@ -163,6 +163,8 @@ func (_ *LoadBalancer) RenderDO(t *do.DOAPITarget, a, e, changes *LoadBalancer)
 		return err
 	}
 
+	klog.Infof("Load balancer created with ID=%s", loadbalancer.ID)
```
olemarkus, line 166:
```suggestion
	klog.V(2).Infof("Load balancer created with ID=%s", loadbalancer.ID)
```

olemarkus (changes requested):
Maybe consider adding this to a higher verbosity.

upup/pkg/fi/cloudup/dotasks/loadbalancer.go
srikiz: @olemarkus - Strangely, I don't see the logs emitted if I use `klog.V(2)` when testing locally.
However, I do see if with `klog.Infof` as seen below

`I1114 10:36:38.306548    5486 loadbalancer.go:166] Load balancer created with ID=59a14963-08b5-4cfb-94cd-2ffd00d9b6e1`

Am I missing something here?

srikiz:
Figured out the issue and created a separate PR - https://github.com/kubernetes/kops/pull/12758 to fix it. 
Hence, closing this.

PR #12736, opened 2021-11-14
Enable CSI volumes in GCE

zetaab:
The remaining problem is that we should somehow delete old storageclass (if exists) which uses old provisioner. The old name for storageclass is `standard` and new one `standard-csi`

@justinsb do you have tip where we should do that? nodeup maybe? Can we just do that if there are old volumes still using that? At least for me it was working pretty well (created first volume using normal provisioner `kubernetes.io/gce-pd` after deploying CSI stuff). https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/ If I understand correctly it should work out of the box for old `kubernetes.io/gce-pd` provisioner.


zetaab:
@rifelpet any ideas what I am doing wrong?

    compare.go:45: error reading file "/bazel-scratch/.cache/bazel/_bazel_root/3fe4bf09df1f8c97d1bcca7cfce3cda6/sandbox/processwrapper-sandbox/2180/execroot/__main__/bazel-out/k8-fastbuild/bin/cmd/kops/go_default_test_/go_default_test.runfiles/__main__/tests/integration/update_cluster/ha_gce/data/aws_s3_bucket_object_ha-gce.example.com-addons-storage-gce.addons.k8s.io-k8s-1.22_content": open /bazel-scratch/.cache/bazel/_bazel_root/3fe4bf09df1f8c97d1bcca7cfce3cda6/sandbox/processwrapper-sandbox/2180/execroot/__main__/bazel-out/k8-fastbuild/bin/cmd/kops/go_default_test_/go_default_test.runfiles/__main__/tests/integration/update_cluster/ha_gce/data/aws_s3_bucket_object_ha-gce.example.com-addons-storage-gce.addons.k8s.io-k8s-1.22_content: no such file or directory

upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
```diff
@@ -712,7 +712,7 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.ModelBuilderContext) (*Addon
 		key := "storage-gce.addons.k8s.io"
 
 		{
-			id := "v1.7.0"
+			id := "k8s-1.22"
```
johngmyers, line 715: Per our naming convention, this suggests the manifest applies for k8s 1.22 and above. But the code is using it unconditionally, meaning it applies for all supported k8s versions, meaning 1.18 and above. If you intend to cherry-pick this to kOps 1.22, then it would apply to k8s 1.17 and above.

```suggestion
			id := "k8s-1.17"
```

johngmyers (changes requested):
The way we handle such changes in AWS is to create both old and new StorageClasses and make the new one the default. That is necessary for upgrades where there are existing PVs on the old StorageClass.

zetaab:
closing see https://github.com/kubernetes/kops/pull/12812

PR #12737, opened 2021-11-14
GCE: Fix race around route deletion

justinsb:
Because the control-plane can recreate routes, there's a race between
deleting instances and deleting routes.  Add a dependency so we don't
try to delete routes until after we've deleted all the instances.

### DIFF
--- a/pkg/resources/gce/BUILD.bazel
+++ b/pkg/resources/gce/BUILD.bazel
@@ -15,7 +15,6 @@ go_library(
         "//upup/pkg/fi/cloudup/gce:go_default_library",
         "//vendor/google.golang.org/api/compute/v1:go_default_library",
         "//vendor/google.golang.org/api/dns/v1:go_default_library",
-        "//vendor/k8s.io/apimachinery/pkg/util/sets:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
     ],
 )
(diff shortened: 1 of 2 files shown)


justinsb:
Race was observed in https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-gce-latest/1459699645369290752/build-log.txt

zetaab (approved):
from my perspective looks good, but run `./hack/update-bazel.sh`

justinsb:
I think the additional failures beyond the 2 we know about are due to skip-regex list not being in sync: https://github.com/kubernetes/test-infra/pull/24362

justinsb:
Rebased, OIDC test should be fixed.  Metadata concealment is the last one (other than a few flakes!)

/test pull-kops-e2e-k8s-gce

justinsb:
Just down to metadata concealment now.  I propose we merge and deal with that one separately (not least because this PR itself addresses a different test failure!)

rifelpet:
Agreed. You'll need to override the job's context or force push though.

rifelpet:
The metadata concealment test passes in the periodic job:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-gce-latest/1462598870054735872

The main differences are:
* presubmit uses cilium
* periodic sets `serviceAccountIssuer`
* presubmit uses k8s 1.22 but periodic uses k8s 1.23 - I didn't see any significant changes to the test or its pod image between 1.22 and 1.23

justinsb:
Thanks for lgtm @rifelpet and @johngmyers; and for the pointers on the metadata job @rifelpet !

PR #12738, opened 2021-11-14
Support apiserver-only nodes on GCE

justinsb:
Extending the existing AWS support to support GCE also.

### DIFF
--- a/pkg/apis/kops/validation/instancegroup.go
+++ b/pkg/apis/kops/validation/instancegroup.go
@@ -187,8 +187,14 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, cl
 		allErrs = append(allErrs, ValidateMasterInstanceGroup(g, cluster)...)
 	}
 
-	if g.Spec.Role == kops.InstanceGroupRoleAPIServer && kops.CloudProviderID(cluster.Spec.CloudProvider) != kops.CloudProviderAWS {
-		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "role"), "Apiserver role only supported on AWS"))
+	if g.Spec.Role == kops.InstanceGroupRoleAPIServer {
+		switch kops.CloudProviderID(cluster.Spec.CloudProvider) {
+		case kops.CloudProviderAWS, kops.CloudProviderGCE:
+			// Allowed
+
+		default:
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "role"), "Apiserver role only supported on AWS and GCE"))
+		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/apis/kops/validation/instancegroup.go
```diff
@@ -187,8 +187,14 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, cl
...
 
-	if g.Spec.Role == kops.InstanceGroupRoleAPIServer && kops.CloudProviderID(cluster.Spec.CloudProvider) != kops.CloudProviderAWS {
-		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "role"), "Apiserver role only supported on AWS"))
+	if g.Spec.Role == kops.InstanceGroupRoleAPIServer {
+		switch kops.CloudProviderID(cluster.Spec.CloudProvider) {
+		case kops.CloudProviderAWS, kops.CloudProviderGCE:
+			// Allowed
+
+		default:
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "role"), "Apiserver role only supported on AWS"))
```
johngmyers, line 196:
```suggestion
			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "role"), "Apiserver role only supported on AWS and GCE"))
```

pkg/model/gcemodel/autoscalinggroup.go
```diff
@@ -128,7 +128,7 @@ func (b *AutoscalingGroupModelBuilder) buildInstanceTemplate(c *fi.ModelBuilderC
 			}
 
 			switch ig.Spec.Role {
-			case kops.InstanceGroupRoleMaster:
+			case kops.InstanceGroupRoleMaster, kops.InstanceGroupRoleAPIServer:
```
johngmyers, line 131: Do APIServer nodes really need DNS permissions?

Should't we tag them with InstanceGroupRoleAPIServer?

johngmyers (changes requested):

pkg/model/gcemodel/autoscalinggroup.go
justinsb: Great points both - will fix!

justinsb: Actually, the tag one is trickier... we share security groups on AWS as well, and this tag is a network tag, controlling the GCE equivalent of AWS security groups.

johngmyers: Could use a comment then.

justinsb:
/reopen

Not resolved

PR #12741, opened 2021-11-14
Add create cluster flag for enabling IRSA

johngmyers:


olemarkus (approved):

hakman:
@johngmyers @olemarkus Should we cherry-pick this to 1.22 to make things easier for users?

johngmyers:
I think that would be something to bring up in Office Hours. This is a release-noteable feature, but it would make things much easier.

PR #12742, opened 2021-11-14
GCE: Support kops-controller, including in gossip mode

justinsb:
We discover the kops-controller in gossip mode using seeding code that
calls into the GCE API, just like gossip itself does.

We refactor the gossip code into a shared gcediscovery library with
minimal dependencies.


johngmyers:
Bootstrap changes look good. I don't know about the other GCE-specific changes.

justinsb:
/test pull-kops-e2e-cni-cilium-ipv6

Not quite sure how these got triggered, but they should be unrelated / should pass!

johngmyers:
@justinsb I added the IPv6 jobs to trigger for PRs that affect files for the corresponding CNIs. They might require the PR to be rebased to take in recent IPv6 changes.

upup/pkg/fi/cloudup/gce/gcediscovery/resolver.go
```diff
@@ -0,0 +1,266 @@
...
+
+// Resolve implements resolver.Resolve, providing name -> address resolution using GCE discovery.
+func (r *Discovery) Resolve(ctx context.Context, name string) ([]string, error) {
+	var records []string
+	klog.Infof("trying to resolve %q using GCEResolver", name)
+
+	var requiredTags []string
+
+	// We assume we are trying to resolve a component that runs on the control plane
+	requiredTags = append(requiredTags, gce.TagForRole(r.clusterName, kops.InstanceGroupRoleMaster))
```
rifelpet, line 233: Should we also find instances tagged with InstanceGroupRoleAPIServer?

justinsb: I don't know; but here's why I think we don't need to do it yet..

1) We don't need _all_ the candidates, we just want to find one we can connect to
2) Do we run kops-controller on apiserver only nodes?  (And will we continue to do?)

rifelpet (approved):

PR #12743, opened 2021-11-15
Use dualstack endpoint for s3

olemarkus:
If we are launching kops in an environment with only ipv6, we need to use dualstack endpoints for fetching things from S3.

There is a challenge in that for existing clusters, this will change the issuer urls, and we have run into problems with this in the past.

### DIFF
--- a/util/pkg/vfs/s3context.go
+++ b/util/pkg/vfs/s3context.go
@@ -36,13 +36,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
-var (
-	// matches all regional naming conventions of S3:
-	// https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region
-	// TODO: perhaps make region regex more specific, i.e. (us|eu|ap|cn|ca|sa), to prevent matching bucket names that match region format?
-	//       but that will mean updating this list when AWS introduces new regions
-	s3UrlRegexp = regexp.MustCompile(`(s3([-.](?P<region>\w{2}-\w+-\d{1})|[-.](?P<bucket>[\w.\-\_]+)|)?|(?P<bucket>[\w.\-\_]+)[.]s3([.](?P<region>\w{2}-\w+-\d{1}))?)[.]amazonaws[.]com([.]cn)?(?P<path>.*)?`)
-)
+// matches all regional naming conventions of S3:
+// https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region
+// TODO: perhaps make region regex more specific, i.e. (us|eu|ap|cn|ca|sa), to prevent matching bucket names that match region format?
+//       but that will mean updating this list when AWS introduces new regions
+var s3UrlRegexp = regexp.MustCompile(`(s3([-.](?P<region>\w{2}-\w+-\d{1})|[-.](?P<bucket>[\w.\-\_]+)|)?|(?P<bucket>[\w.\-\_]+)[.]s3([.](?P<region>\w{2}-\w+-\d{1}))?)[.]amazonaws[.]com([.]cn)?(?P<path>.*)?`)
 
 type S3BucketDetails struct {
 	// context is the S3Context we are associated with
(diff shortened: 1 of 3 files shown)


johngmyers:
Hrm, the `iss` claim can only be single-valued. So there will be authentication failures between apply and when the last token issued by an old control plane node is replaced.

olemarkus:
API server can be configured to accept multiple issuers though.

rifelpet:
This could be done outside of this particular PR, but do you think we should conditionally enable dualstack endpoints for all aws-sdk-go clients that run within the cluster?

https://github.com/aws/aws-sdk-go/releases/tag/v1.42.0

we could even include `AWS_USE_DUALSTACK_ENDPOINT` in any addon manifests that use the aws sdk.

johngmyers:
I was thinking of IRSA authentication failures. The tokens issued by controller-managers not updated yet wouldn't work with the new configuration of AWS.

johngmyers:
I see no reason to not always use dualstack endpoints. Why are there separate dualstack enpoints anyway?

olemarkus:
I think we should enable dualstack endpoints everywhere in a follow up PR. I did S3 specifically because we need to handle the migration somehow.

johngmyers:
I think we might need a two-upgrade migration. Accept from both endpoints in one version and switch the primary endpoint in the next.

johngmyers:
We might also need to publish two IDPs for the transition.

hakman:
@olemarkus We discussed this topic yesterday during office hours and agreed it would be best to limit this to IPv6 clusters.
AWS is not so good with deploying IPv6 features consistently and we may find situations when duals-stack endpoints are not enabled.

olemarkus:
That makes sense. It's only the discovery logic where this change is problematic, so it makes sense to do the conditional there, I think. To avoid having to do some "configuration" of VFS.

util/pkg/vfs/s3context.go
```diff
@@ -84,7 +82,7 @@ func (s *S3Context) getClient(region string) (*s3.S3, error) {
 		var err error
 		endpoint := os.Getenv("S3_ENDPOINT")
 		if endpoint == "" {
-			config = aws.NewConfig().WithRegion(region)
+			config = aws.NewConfig().WithRegion(region).WithUseDualStack(true)
```
johngmyers, line 85: Do all regions support the dual stack endpoint for S3?

johngmyers: The dual-stack S3 endpoints are missing in `cn-north-1` and `cn-northwest-1`.

They do exist in the GovCloud regions. I don't know the names of the isolated regions.

olemarkus: As far as I can tell, yes. At least I haven't found a region where this doesn't work.

I can feed in `IPv6Only` here as well to be safe. That would at least put this behind a feature flag.

johngmyers: We could require users in said regions to set the `S3_ENDPOINT` environment variable...

olemarkus: Yeah. Come to think of it, making this code be aware of `IPv6Only` will be very tricky.

johngmyers: Do we need a release note for folks in the China regions?

hakman: Seems that we don't need to do anything for China regions:
https://aws.amazon.com/about-aws/whats-new/2020/05/amazon-s3-adds-support-ipv6-protocol-beijing-sinnet-ningxia-nwcd/

johngmyers: Ah, different TLD.

hakman:
`[sig-auth] ServiceAccounts ServiceAccountIssuerDiscovery should support OIDC discovery of service account ...` is passing now:
https://testgrid.k8s.io/kops-ipv6#kops-grid-scenario-ipv6-calico

PR #12756, opened 2021-11-16
Enable DNS64 on ipv6 subnets

olemarkus:
There is nothing NATing these synthetic IPs so for now, this is more just to get kOps to provision instances where one can test the behavior of DNS64 itself.

### DIFF
--- a/pkg/model/awsmodel/network.go
+++ b/pkg/model/awsmodel/network.go
@@ -264,6 +264,7 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
 			subnet.IPv6CIDR = fi.String(subnetSpec.IPv6CIDR)
 			// TODO: set this to true once NAT64 is in place
 			subnet.DNS64 = fi.Bool(false)
+			subnet.ResourceBasedNaming = fi.Bool(true)
 		}
 		if subnetSpec.ProviderID != "" {
 			subnet.ID = fi.String(subnetSpec.ProviderID)
--- a/pkg/model/awsmodel/network.go
+++ b/pkg/model/awsmodel/network.go
@@ -21,11 +21,12 @@ import (
 	"strings"
 
 	"k8s.io/klog/v2"
+	"k8s.io/legacy-cloud-providers/aws"
+
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awstasks"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
-	"k8s.io/legacy-cloud-providers/aws"
 )
 
 // NetworkModelBuilder configures network objects
@@ -253,6 +254,7 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
 			CIDR:             fi.String(subnetSpec.CIDR),
 			Shared:           fi.Bool(sharedSubnet),
 			Tags:             tags,
+			DNS64:            fi.Bool(false),
 		}
 
 		if subnetSpec.IPv6CIDR != "" {
@@ -260,6 +262,8 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
 				subnet.AmazonIPv6CIDR = b.LinkToAmazonVPCIPv6CIDR()
 			}
 			subnet.IPv6CIDR = fi.String(subnetSpec.IPv6CIDR)
+			// TODO: set this to true once NAT64 is in place
+			subnet.DNS64 = fi.Bool(false)
 		}
 		if subnetSpec.ProviderID != "" {
 			subnet.ID = fi.String(subnetSpec.ProviderID)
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -46,7 +46,7 @@ require (
 	github.com/aliyun/alibaba-cloud-sdk-go v1.61.1059
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/amazon-ec2-instance-selector/v2 v2.0.2
-	github.com/aws/aws-sdk-go v1.40.38
+	github.com/aws/aws-sdk-go v1.42.5
```
hakman, line 49: Maybe it needs a rebase?
https://github.com/kubernetes/kops/commit/0c08f047ba38639837a7b40e62a01b3cd51edf6c

olemarkus: Probably not for the last time before this can go in...

We could merge this one and always set DNS64 to false though.

hakman: Probably a good idea.

rifelpet:
I opened https://github.com/hashicorp/terraform-provider-aws/issues/21791 for the terraform equivalent

johngmyers:
Why not just enable DNS64 in coredns? Why use AWS's DNS servers?

olemarkus:
I would think using AWS servers would be much simpler than configuring coredns/dns cache

johngmyers:
I believe coredns has DNS64 support.

I would think it better if the nodes, which are dual stack, were not behind DNS64. Especially if they're on a public subnet. That would cause them to unnecessarily use NAT64.

johngmyers:
It turns out that configuring CoreDNS in #12766 was much simpler than this PR.

johngmyers:
I believe we should only enable DNS64 on IPv6-only subnets. Dual-stack subnets should get DNS64 from coredns.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -261,6 +266,36 @@ func (_ *Subnet) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Subnet) error {
...
+		}
+		_, err := t.Cloud.EC2().ModifySubnetAttribute(request)
+		if err != nil {
+			return fmt.Errorf("error enabling resource based dns on subnet: %v", err)
+		}
+		request = &ec2.ModifySubnetAttributeInput{
+			SubnetId:                             e.ID,
+			EnableResourceNameDnsARecordOnLaunch: &ec2.AttributeBooleanValue{Value: e.ResourceBasedNaming},
+		}
+		_, err = t.Cloud.EC2().ModifySubnetAttribute(request)
```
johngmyers, line 293: Don't you need to set `PrivateDnsHostnameTypeOnLaunch`? 
Why two separate `ModifySubnetAttribute` calls? You can't set multiple attributes in one call?

pkg/model/awsmodel/network.go
```diff
@@ -253,13 +254,17 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
 		}
 
 		if subnetSpec.IPv6CIDR != "" {
 			if !sharedVPC {
 				subnet.AmazonIPv6CIDR = b.LinkToAmazonVPCIPv6CIDR()
 			}
 			subnet.IPv6CIDR = fi.String(subnetSpec.IPv6CIDR)
+			// TODO: set this to true once NAT64 is in place
+			subnet.DNS64 = fi.Bool(false)
+			subnet.ResourceBasedNaming = fi.Bool(true)
```
johngmyers, line 267: Because of CCM, we'll need to put an `IsKubernetesGTE("1.23")` condition on this.

upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -261,6 +266,36 @@ func (_ *Subnet) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Subnet) error {
...
+		_, err := t.Cloud.EC2().ModifySubnetAttribute(request)
+		if err != nil {
+			return fmt.Errorf("error enabling DNS64 on subnet: %v", err)
+		}
+	}
+
+	if changes.ResourceBasedNaming != nil {
+		request := &ec2.ModifySubnetAttributeInput{
+			SubnetId:                                e.ID,
+			EnableResourceNameDnsAAAARecordOnLaunch: &ec2.AttributeBooleanValue{Value: e.ResourceBasedNaming},
```
johngmyers, line 283: I think this should be keyed off of whether there are any IPv6 CIDRs assigned, not the value of `ResourceBasedNaming`.

pkg/model/awsmodel/network.go
hakman: Is this a reasonable thing to happen when someone upgrades its cluster from k8s 1.22 to 1.23?

olemarkus: This is only for ipv6 clusters, so there shouldn't be a problem here. We shouldn't support ipv6 without RBN,I think.
We can also use CCM 1.23 on k8s 1.22 clusters really. There is no other (significant) change to CCM at the moment.

hakman: But, what we really need to do is set a minimum k8s version for IPv6.

olemarkus: Agreed.

olemarkus:
/kind office-hours

Given that dns64 on coredns doesn't work as expected if nodelocaldns is enabled, and not at all on ipv6-only networks, I think we should reconsider this PR. It does NAT some extra traffic for public subnets not behind a NAT GW, but this only happens on instance-initiated traffic when the instances run in public subnets.

PR #12758, opened 2021-11-16
[Digital Ocean] Fix load balancer retry logic while retrieving ip

srikiz:
The [e2e tests for Digital Ocean cloud provider](https://testgrid.k8s.io/sig-cluster-lifecycle-kops#e2e-kops-do-calico) was failing after this PR got merged - https://github.com/kubernetes/kops/pull/12624/files

With the above changes, kubetest2 was modified to do a create and then followed by an update with `--yes`. But the DO's load balancer logic didn't handle the create logic previously. It assumed that creation happens with `--yes`. So when the args wasn't passed, it kept waiting until the IP address is created, and eventually timesout.

The fix now only checks if loadbalancer IP is created during ingress status check which includes an exponential backoff retry logic. This gets triggered only when cluster creation is triggered.

FYI - @rifelpet @timoreimann

### DIFF
--- a/upup/pkg/fi/cloudup/do/BUILD.bazel
+++ b/upup/pkg/fi/cloudup/do/BUILD.bazel
@@ -17,9 +17,11 @@ go_library(
         "//pkg/cloudinstances:go_default_library",
         "//protokube/pkg/etcd:go_default_library",
         "//upup/pkg/fi:go_default_library",
+        "//util/pkg/vfs:go_default_library",
         "//vendor/github.com/digitalocean/godo:go_default_library",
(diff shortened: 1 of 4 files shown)


srikiz:
FYI - @timoreimann @rifelpet - this is now good for review.

olemarkus (approved):

rifelpet:
this lgtm also, pending the squash request

srikiz:
Thanks @olemarkus @hakman @rifelpet - I have squashed the commits !

rifelpet (approved):
/test pull-kops-e2e-cni-cilium-ipv6

PR #12760, opened 2021-11-17
Partial backport of #12564

olemarkus:
Partial backport of #12564 
Keep CNP status updates enabled to avoid BC break.

### DIFF
--- a/pkg/model/components/cilium.go
+++ b/pkg/model/components/cilium.go
@@ -140,7 +140,7 @@ func (b *CiliumOptionsBuilder) BuildOptions(o interface{}) error {
 	}
 
 	if c.DisableCNPStatusUpdates == nil {
-		c.DisableCNPStatusUpdates = fi.Bool(true)
+		c.DisableCNPStatusUpdates = fi.Bool(false)
 	}
 
 	if c.CPURequest == nil {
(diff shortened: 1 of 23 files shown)


pkg/model/components/cilium.go
```diff
@@ -139,6 +139,10 @@ func (b *CiliumOptionsBuilder) BuildOptions(o interface{}) error {
 		c.EnableL7Proxy = fi.Bool(true)
 	}
 
+	if c.DisableCNPStatusUpdates == nil {
+		c.DisableCNPStatusUpdates = fi.Bool(false)
+	}
+
```
hakman, line 145: This should be the only change to this file, right?

hakman (changes requested):

pkg/model/components/cilium.go
olemarkus: VS Code started autoformatting some new stuff on save. I just let it. Seems to be improvements (I am really inconsistent when it comes to whitespace usage, so I like this).

hakman: Was just thinking that this would make cherry-picks harder later.

olemarkus: Probably not. But anyways, removed those changes now

hakman (approved):

yurrriq:
Will this land in 1.22.2 then? :crossed_fingers:

hakman:
> Will this land in 1.22.2 then? 🤞

yes

PR #12763, opened 2021-11-17
Bump etcd manager to 20211117

justinsb:
WIP because we haven't tagged the build yet; I want to put it through testing.

### DIFF
--- a/pkg/model/components/etcdmanager/model.go
+++ b/pkg/model/components/etcdmanager/model.go
@@ -179,7 +179,7 @@ metadata:
   namespace: kube-system
 spec:
   containers:
-  - image: k8s.gcr.io/etcdadm/etcd-manager:3.0.20211007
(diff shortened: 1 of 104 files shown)


rifelpet:
Unfortunately we dont have any presubmit upgrade tests which is what this etcd-manager version should fix, maybe we could add some? it may be straight forward using the `upgrade-ab` scenario.

justinsb:
I did run the upgrade test and it looks healthy (i.e. unrelated to etcd: one OIDC failure, one OwnerReferences failure - flake?)

I'm going to send the promotion PR and then update this PR to point to the promoted version.

justinsb:
Promotion PR is https://github.com/kubernetes/k8s.io/pull/3088

(I'm also working on a release process doc :-) )

hakman (approved):

justinsb:
Removed WIP, but grappling with a flake.

PR #12766, opened 2021-11-18
Enable DNS64 in CoreDNS if IPv6 enabled

johngmyers:
This is most useful when nodes are in dual-stack subnets.

In IPv6-only subnets we will want to enable AWS's DNS64 on the subnet, but having it also enabled in CoreDNS will not hurt.


johngmyers:
This doesn't work for NodeLocalDNS, but then NodeLocalDNS is broken on IPv6 clusters—a nameserver of `169.254.20.10` doesn't work too well from an IPv6-only pod.

I could put the nat64 support into the nodelocaldns manifest template regardless.

johngmyers:
I suppose one eventual use case might be an IPv6 cluster with no need for NAT64. But for now we need NAT64 to deal with critical v4-only AWS services.

olemarkus:
I agree that the best solution is to use dns64 in coredns for dualstack nodes. Let's merge this one, and I will amend my PR to distinguish between ipv6only pod networking and ipv6only node networking

olemarkus (approved):

PR #12779, opened 2021-11-18
WIP: e2e tests: create (additional) logs per-command

justinsb:
This makes it much easier to see the per-command logs that show what
is actually happening.


justinsb:
/wip until I figure out where to put these so they get captured

PR #12780, opened 2021-11-18
Stable-sort subnets by Name

justinsb:
This ensures a stable order, even if/when the IDs aren't set (e.g. in terraform)


justinsb:
I observed a test-failure here, where it looks like we are not stable-sorting subnet mappings: https://github.com/kubernetes/kops/runs/4253856935?check_suite_focus=true

I think the root cause is that only the Name is set in terraform.

hakman (approved):

justinsb:
OK, as this issue is blocking other PRs, removing WIP

PR #12784, opened 2021-11-18
Bump nvidia device plugin to 0.10.0

olemarkus:


johngmyers:
We should probably have an e2e test for this.
/lgtm
/approve

PR #12788, opened 2021-11-19
Remove featureflag for creating IPv6 clusters

hakman:
Makes it easier for operators to create IPv6 clusters, which and should be pretty good already in 1.23.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -314,9 +314,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	// TODO: Can we deprecate this flag - it is awkward?
 	cmd.Flags().BoolVar(&associatePublicIP, "associate-public-ip", false, "Specify --associate-public-ip=[true|false] to enable/disable association of public IP for master ASG and nodes. Default is 'true'.")
 
-	if featureflag.AWSIPv6.Enabled() {
-		cmd.Flags().BoolVar(&options.IPv6, "ipv6", false, "Allocate IPv6 CIDRs to subnets for clusters with public topology (AWS only)")
-	}
+	cmd.Flags().BoolVar(&options.IPv6, "ipv6", false, "Use IPv6 for the pod network (AWS only)")
 
 	cmd.Flags().StringSliceVar(&options.NodeSecurityGroups, "node-security-groups", options.NodeSecurityGroups, "Additional precreated security groups to add to worker nodes.")
 	cmd.RegisterFlagCompletionFunc("node-security-groups", completeSecurityGroup)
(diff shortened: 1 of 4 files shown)


johngmyers:
I think we should get a NAT64/DNS64 solution before doing this.

hakman:
> I think we should get a NAT64/DNS64 solution before doing this.

Depends on use case. It NAT64 is just a limitation that may not be a concern for some. We can document it and explain how someone could use an external NAT64 setup.
Having the `--ipv6` flag hidden is not that nice though.

olemarkus:
I think AWS and ipv6 is a bit too much of a moving target to bring this to 1.23. I think there may still be changes to kops behavior.

cmd/kops/create_cluster.go
```diff
@@ -314,9 +314,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	// TODO: Can we deprecate this flag - it is awkward?
 	cmd.Flags().BoolVar(&associatePublicIP, "associate-public-ip", false, "Specify --associate-public-ip=[true|false] to enable/disable association of public IP for master ASG and nodes. Default is 'true'.")
 
-	if featureflag.AWSIPv6.Enabled() {
-		cmd.Flags().BoolVar(&options.IPv6, "ipv6", false, "Allocate IPv6 CIDRs to subnets for clusters with public topology (AWS only)")
-	}
+	cmd.Flags().BoolVar(&options.IPv6, "ipv6", false, "Allocate IPv6 CIDRs to subnets for clusters with public topology (AWS only)")
```
johngmyers, line 317: This description should be improved. It's more around the pods being IPv6-only. And the "public topology" part is out of date.

Perhaps something like:

```suggestion
	cmd.Flags().BoolVar(&options.IPv6, "ipv6", false, "Make the pod network family be IPv6 (AWS only)")
```

But perhaps we should have a `--network-family` flag with values `ipv4` (default) and `ipv6`?

I've seen references from AWS to IPv6-only subnets, so we should think about how we'll handle those.

hakman: I agree on the message.
About the flag, we can discuss it during the next office hours. 1.23 is still alpha so, not really an issue.

hakman:
> I think AWS and ipv6 is a bit too much of a moving target to bring this to 1.23. I think there may still be changes to kops behavior.

I agree on this, this is why docs will still say that it is experimental. We can probably do a better job there.

olemarkus:
I think things landed quite well now. I don't mind removing the feature flag for dualstack nodes. Also for 1.23.

johngmyers:
The private topology PR should land first.

olemarkus:
Guessing we don't remove for 1.23 then.
I guess we should have a working service controller first too...

hakman:
Private topology can be a 1.24 thing, same for the service controller for that.
We can limit `--ipv6` in 1.24 to only allow public  topology.

johngmyers (approved):

PR #12791, opened 2021-11-19
Fix volume ratio comparisons

### LINKED ISSUE #12038: volumeIops to volumeSize ratio must be lower than 50. For 1.etcd-main ratio is 50.00
/kind bug

 this information.**
`Version 1.21.0 (git-a5bdc3359e544b314d5695b3ed596829313fc6e3)`

 a `kops` flag.**
` kubectl v1.21.0`
Last working versions are 1.20.0/1.21.0-alpha.2, first affected versions are 1.20.1 and 1.21.0-alpha.3

`aws`

`kops replace --force -f cluster.yaml --name $CLUSTER_NAME`

`error building tasks: volumeIops to volumeSize ratio must be lower than 50. For 1.etcd-main ratio is 50.00`

Cluster successfully created as was with kops 1.19.2

Relevant part is
```
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-eu-north-1a-1
      name: "1"
      volumeIops: 1000
      volumeSize: 20
      volumeType: io1
```

Nothing relevant there, unfortunately, I just see the values in spec, and then a failure
```
I0722 14:09:55.972668   31923 spec_builder.go:46] options: {
...
  "etcdClusters": [
    {
      "name": "main",
      "provider": "Manager",
      "etcdMembers": [
        {
          "name": "1",
          "instanceGroup": "master-eu-north-1a-1",
          "volumeType": "io1",
          "volumeIops": 1000,
          "volumeSize": 20
        },
...
I0722 14:10:10.839268   31923 task.go:108] testing task "ManagedFile"
I0722 14:10:10.839337   31923 task.go:108] testing task "ManagedFile"
I0722 14:10:10.839346   31923 task.go:108] testing task "Keypair"
I0722 14:10:10.839372   31923 task.go:108] testing task "Keypair"
I0722 14:10:10.839398   31923 task.go:108] testing task "Keypair"
I0722 14:10:10.839650   31923 build_flags.go:49] ignoring non-field:
I0722 14:10:10.839682   31923 build_flags.go:49] ignoring non-field:
I0722 14:10:10.840158   31923 task.go:108] testing task "ManagedFile"
I0722 14:10:10.840201   31923 task.go:108] testing task "ManagedFile"
I0722 14:10:10.840207   31923 task.go:108] testing task "Keypair"
I0722 14:10:10.840211   31923 task.go:108] testing task "Keypair"
I0722 14:10:10.840214   31923 task.go:108] testing task "Keypair"
I0722 14:10:10.840231   31923 task.go:81] EnsureTask ignoring identical

error building tasks: volumeIops to volumeSize ratio must be lower than 50. For 1.etcd-main.208.k8s.wixprod.net ratio is 50.00
```

Reviewing changelog for 1.21.0-alpha.3 it seems this pull is the source of the issue
https://github.com/kubernetes/kops/pull/11225/files
Looks like rounding is needed somewhere starting from this block
https://github.com/kubernetes/kops/blob/master/pkg/model/master_volumes.go#L146
But upon quick research it looks like Go doesn't support proper rounding for float64. Looking on the old code, I have a guess that previous calls to int64() maybe were helping there, but not sure. Also I've checked the smallest possible value for that relation, which should overflow, it is 50,0002, so a dirty hack here would be to multiply the relation by 10000, converting to int32, and compare it to 50000 )).

olemarkus:
Our ratio checks check for e.g == 50, which is actually allowed by AWS.
Fixes #12038

### DIFF
--- a/pkg/model/BUILD.bazel
+++ b/pkg/model/BUILD.bazel
@@ -65,7 +65,10 @@ go_library(
 
 go_test(
     name = "go_default_test",
-    srcs = ["bootstrapscript_test.go"],
+    srcs = [
+        "bootstrapscript_test.go",
+        "master_volumes_test.go",
+    ],
     data = glob(["tests/**"]),  #keep
     embed = [":go_default_library"],
     deps = [
--- a/pkg/model/master_volumes.go
+++ b/pkg/model/master_volumes.go
@@ -197,18 +197,18 @@ func validateAWSVolume(name, volumeType string, volumeSize, volumeIops, volumeTh
 	volumeThroughputIopsRatio := float64(volumeThroughput) / float64(volumeIops)
 	switch volumeType {
 	case ec2.VolumeTypeIo1:
-		if volumeIopsSizeRatio >= 50.0 {
+		if volumeIopsSizeRatio > 50.0 {
 			return fmt.Errorf("volumeIops to volumeSize ratio must be lower than 50. For %s ratio is %.02f", name, volumeIopsSizeRatio)
 		}
 	case ec2.VolumeTypeIo2:
-		if volumeIopsSizeRatio >= 500.0 {
+		if volumeIopsSizeRatio > 500.0 {
 			return fmt.Errorf("volumeIops to volumeSize ratio must be lower than 500. For %s ratio is %.02f", name, volumeIopsSizeRatio)
 		}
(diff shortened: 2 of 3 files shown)


hakman (approved):
Maybe do some cherry-picking if you don't mind.

PR #12792, opened 2021-11-19
gossip: support resolution of k8s.local names from pods (via services)

justinsb:
We add the hosts plugin to CoreDNS, and we populate a ConfigMap from
kops-controller (when in gossip mode).

This enables resolution of the internal apiserver DNS name from Pods,
even when gossip mode (k8s.local) is in use. This should fix the
failing e2e tests which are assuming that the name in the JWT token is
resolvable from inside the cluster.

This is also a possible step towards a simpler gossip mode, now that
we have a central controller.


justinsb:
This is the revamp of #12378 using headless services.  It is much simpler and more efficient.

We could have CoreDNS do the service lookup by adding additional kubernetes block(s), but if we do that we are very restricted in the service naming (I think it has to be `api` and `main` and `events`), and likely opens up a larger surface than we are opening here.  Additionally, we likely want this information in the kops-controller anyway, if we're going to assist gossip.

justinsb:
This should let the GCE OIDC test pass:
/test pull-kops-e2e-k8s-gce

justinsb:
Force pushed a few self-review tweaks:
/test pull-kops-e2e-k8s-gce

justinsb:
/test pull-kops-e2e-cni-cilium

Failure was related to cilium, which I _think_ is unrelated.

johngmyers (approved):

PR #12804, opened 2021-11-20
Check that there are extra fields and not fields explicitly false

olemarkus:
When we check for extra fields, we check only check for byte equality. But because of omitEmpty, fields set explicitly to false are removed, which cause a diff of removed line. This change introduces an actual strict unmarshal to ensure all the fields are known. Then it continues with the existing logic to preserve the output.

### DIFF
--- a/pkg/edit/edit.go
+++ b/pkg/edit/edit.go
@@ -20,6 +20,9 @@ import (
 	"bytes"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
+
+	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/diff"
 	"k8s.io/kops/pkg/kopscodecs"
 	"k8s.io/kops/upup/pkg/fi/utils"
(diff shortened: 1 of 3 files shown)


pkg/edit/edit.go
```diff
@@ -29,7 +32,18 @@ import (
...
 // If there are no extra fields it returns an empty string
-func HasExtraFields(yaml string, object runtime.Object) (string, error) {
+func HasExtraFields(yamlString string, object runtime.Object) (string, error) {
+	switch object.(type) {
+	case *kops.Cluster:
+		editedClusterObj := kops.Cluster{}
+		err := yaml.UnmarshalStrict([]byte(yamlString), &editedClusterObj)
+		if err == nil {
+			return "", nil
+		}
```
johngmyers, line 42: Don't we also need to handle `kops.InstanceGroup`?

olemarkus: We do.
There is very little testing around this. I think we should consolidate some of the logic in the edit commands with the replace command. Typos in yaml passed to replace is also a fairly common source of bugs.

johngmyers (approved):

PR #12805, opened 2021-11-20
Add action for automatically tagging releases

johngmyers:
Tested in the johngmyers/testactions repository.

Will tag an alpha to verify before removing the restriction to alpha. Then will backport script to release branches before enabling there.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,0 +1,23 @@
+name: tag-release
+
+on:
+  push:
+    branches:
+      - master
+      #- 'release-*'
+    paths:
+      - version.go
+
+jobs:
+  tag-release:
+    if: ${{ github.repository == 'kubernetes/kops' }}
+    runs-on: ubuntu-20.04
+
+    permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,23 @@
...
+  tag-release:
+    if: ${{ github.repository == 'kubernetes/kops' }}
+    runs-on: ubuntu-20.04
+
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/checkout@v2
+      - run: /usr/bin/git config --global user.email actions@github.com
```
rifelpet, line 21: I wonder if there are any CLA concerns with having a commit in the repo authored from this email address. Are there other kubernetes repos that author git commits in GitHub Actions?

johngmyers: Since it's not authored by a human there are unlikely to be any legal concerns. The commit isn't going through a PR, so it won't trip up the CLA gates.

hakman (approved):

PR #12806, opened 2021-11-21
Update channels

hakman:
/cc @johngmyers @rifelpet

### DIFF
--- a/channels/stable
+++ b/channels/stable
@@ -58,11 +58,11 @@ spec:
       providerID: aws
       architectureID: amd64
       kubernetesVersion: ">=1.17.0 <1.18.0"
-    - name: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20211015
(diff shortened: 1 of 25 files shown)


channels/stable
```diff
@@ -58,11 +58,11 @@ spec:
       providerID: aws
       architectureID: amd64
       kubernetesVersion: ">=1.17.0 <1.18.0"
-    - name: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20211015
+    - name: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20211118
```
johngmyers, line 61: I wonder if we should go to 10/21 first. But maybe that's not important for AMIs.

hakman: Not very important. Already updated in test and don't see any issues.

johngmyers (approved):

PR #12807, opened 2021-11-21
Promote alpha channel to stable

hakman:
This way we don't forget about it.

/cc @johngmyers @rifelpet


hakman:
I think it's time to promote alpha to stable.
/cc @olemarkus

olemarkus (approved):

PR #12811, opened 2021-11-21
Rename fields to fit acronym conventions

johngmyers:


johngmyers:
I left the "Amd" fields alone, though they are technically acronyms.

olemarkus (approved):

PR #12819, opened 2021-11-23
Add NodeProblemDetector clusterRoleBinding

### LINKED ISSUE #12817: Addon NodeProblemDetector missing privileges
/kind bug

**1. What `kops` version are you running?* Kops 1.22.2

**2. What Kubernetes version are you running?* 1.18.20

**3. What cloud provider are you using?** AWS

Install addon using the documentation steps -> https://kops.sigs.k8s.io/addons/#node-problem-detector

Pods log show the follower error:
```
E1122 21:46:22.131359       1 manager.go:162] failed to update node conditions: nodes "ip-10-200-20-18.eu-west-1.compute.internal" is forbidden: User "system:serviceaccount:kube-system:default" cannot patch resource "nodes/status" in API group "" at the cluster scope
```
Node Problem Detector pod would be able to update node status.

**7. Please provide your cluster manifest.** N/A

**8. Please run the commands with most verbose logging by adding the `-v 10` flag.** N/A

According to some digging, the current [configuration](https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/node-problem-detector.addons.k8s.io/k8s-1.17.yaml.template) is missing the `ServiceAccount ` and `ClusterRoleBinding` resources.

An example is available here -> https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/node-problem-detector/npd.yaml.

Related issue: https://github.com/kubernetes/node-problem-detector/issues/149

Happy to send a PR. Please let me know.

Regards

zhengtianbao:
**What this PR does / why we need it:**

When deploy node-problem-detector in kubernetes cluster, needs rolebinding to patch the node.

**Which issue(s) this PR fixes:**

Fixes #12817

### DIFF
--- a/upup/models/cloudup/resources/addons/node-problem-detector.addons.k8s.io/k8s-1.17.yaml.template
+++ b/upup/models/cloudup/resources/addons/node-problem-detector.addons.k8s.io/k8s-1.17.yaml.template
@@ -1,6 +1,31 @@
 {{ with .NodeProblemDetector }}
 # Sourced from https://github.com/kubernetes/node-problem-detector/tree/v0.8.8
 ---
+# We need service account and role binding for node-problem-detector
+# see: https://github.com/kubernetes/node-problem-detector/issues/149 for details
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: node-problem-detector
+  namespace: kube-system
+  labels:
+    app: node-problem-detector
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: node-problem-detector
+  labels:
+    app: node-problem-detector
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:node-problem-detector
+subjects:
+- kind: ServiceAccount
+  name: node-problem-detector
+  namespace: kube-system
+---
 # Source: node-problem-detector/deployment/node-problem-detector.yaml
 apiVersion: apps/v1
 kind: DaemonSet


upup/models/cloudup/resources/addons/node-problem-detector.addons.k8s.io/k8s-1.17.yaml.template
```diff
@@ -1,6 +1,59 @@
...
+  - events
+  verbs:
+  - create
+  - patch
+  - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: node-problem-detector-binding
```
johngmyers, line 45:
```suggestion
  name: node-problem-detector
```

upup/models/cloudup/resources/addons/node-problem-detector.addons.k8s.io/k8s-1.17.yaml.template
```diff
@@ -1,6 +1,59 @@
...
+  labels:
+    app: node-problem-detector
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: node-problem-detector
```
johngmyers, line 51: Why not use the existing `system:node-problem-detector` ClusterRole instead of defining our own here?

zhengtianbao: I didn't aware the system default clusterRole, Thanks for your advice.

johngmyers (approved):
Thanks!

PR #12824, opened 2021-11-24
WIP Set Resource Based Naming in dual-stack Subnets

johngmyers:


upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -55,6 +55,10 @@ func (t *LaunchTemplate) RenderAWS(c *awsup.AWSAPITarget, a, e, changes *LaunchT
 				Ipv6AddressCount:         t.IPv6AddressCount,
 			},
 		},
+		PrivateDnsNameOptions: &ec2.LaunchTemplatePrivateDnsNameOptionsRequest{
+			EnableResourceNameDnsAAAARecord: fi.Bool(*t.IPv6AddressCount > 0),
```
olemarkus, line 59: Should we not enable A records too?

olemarkus (commented):
There is also an attribute on the subnet to enable RBN. Not sure we need to set that one as well.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
johngmyers: They aren't default?

olemarkus: They are for ipv6 native, but not sure about dual stack

johngmyers:
Perhaps it can be set either on subnet or launch template? I suspect we'd be better off setting it on subnet, so users could customize how they like on shared subnets.

johngmyers:
There's an attribute on the `Subnet` but I don't see any API for setting it.

hakman:
> Perhaps it can be set either on subnet or launch template? I suspect we'd be better off setting it on subnet, so users could customize how they like on shared subnets.

LaunchTemplate is easier exactly because of shared subnets. We can add whatever settings we need and keep them consistent, without modifying or failing if shared subnets are misconfigured.

johngmyers:
I wouldn't say we "need" to set the `HostnameType`, it's just that RBN is a better default. It would be better for us to say "If you want IPBN, create your own shared subnets" than either add an option or prohibit them.

hakman:
Creating new subnets might be inconvenient. I agree that IPBN is a better default in general.

johngmyers:
Looks like CCM can't deal with RBN.

olemarkus:
yeah. I expected that.
Probably an easy fix. Maybe this time just use hostname as node name so that we don't have to pass in hostname override ...

johngmyers:
Here's the culprit:
https://github.com/kubernetes/cloud-provider-aws/blob/a1590733fac851b3a27d351c9c80e9b2bf8d6f7e/pkg/providers/v1/aws.go#L5174-L5182

johngmyers:
Filed kubernetes/cloud-provider-aws#285

PR #12837, opened 2021-11-26
Fix enabling of AWS CCM

johngmyers:
/kind bug

The code disabled KCM in 1.24, but didn't enable CCM. This broke the kops-aws-k8s-latest periodic test, as latest now identifies as 1.24.

### DIFF
--- a/pkg/model/components/kubecontrollermanager.go
+++ b/pkg/model/components/kubecontrollermanager.go
@@ -79,11 +79,7 @@ func (b *KubeControllerManagerOptionsBuilder) BuildOptions(o interface{}) error 
 	kcm.ClusterName = b.ClusterName
 	switch kops.CloudProviderID(clusterSpec.CloudProvider) {
 	case kops.CloudProviderAWS:
-		if b.IsKubernetesGTE("1.24") {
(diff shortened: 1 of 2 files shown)


pkg/model/components/kubecontrollermanager.go
```diff
@@ -79,11 +79,7 @@ func (b *KubeControllerManagerOptionsBuilder) BuildOptions(o interface{}) error
 	kcm.ClusterName = b.ClusterName
 	switch kops.CloudProviderID(clusterSpec.CloudProvider) {
 	case kops.CloudProviderAWS:
-		if b.IsKubernetesGTE("1.24") {
-			kcm.CloudProvider = "external"
-		} else {
-			kcm.CloudProvider = "aws"
-		}
+		kcm.CloudProvider = "aws"
```
hakman, line 82: Isn't external the only accepted value starting 1.24?

johngmyers: Lower down it sets it to `external` if `ExternalCloudControllerManager` is non-`nil`

hakman (approved):

PR #12842, opened 2021-11-27
Route NAT64 to NAT Gateway in IPv6 private topology

johngmyers:
Routes the well-known NAT64 prefix to the NAT gateway (or Transit Gateway if that's what's specified in `egress`) in IPv6 clusters with private topology.

A future PR will need to provide this option for public topology, but we will might want an option in the cluster spec in case the user doesn't need NAT64 and thus doesn't want to pay for a NAT Gateway.

We might want to later provide option to route NAT64 differently than IPv4 egress.


johngmyers:
I'm working on NAT64 for IPv6 public topology for a followup PR.

hakman (approved):

PR #12844, opened 2021-11-27
Use AWS metadata to retrieve local-hostname in nodeup

bwagner5:
Cloud Provider: AWS

This PR changes nodeups method of looking up an instance private DNS name from a DescribeInstances call to an instance metadata lookup. DescribeInstances is prone to throttling with large clusters and can cause significant delays when a node is trying to join a cluster.

### DIFF
--- a/nodeup/pkg/model/BUILD.bazel
+++ b/nodeup/pkg/model/BUILD.bazel
@@ -67,7 +67,6 @@ go_library(
         "//util/pkg/distributions:go_default_library",
         "//util/pkg/proxy:go_default_library",
         "//util/pkg/vfs:go_default_library",
-        "//vendor/github.com/aws/aws-sdk-go/aws:go_default_library",
         "//vendor/github.com/aws/aws-sdk-go/aws/ec2metadata:go_default_library",
         "//vendor/github.com/aws/aws-sdk-go/aws/session:go_default_library",
         "//vendor/github.com/aws/aws-sdk-go/service/ec2:go_default_library",
--- a/nodeup/pkg/model/context.go
+++ b/nodeup/pkg/model/context.go
@@ -36,9 +36,6 @@ import (
 	"k8s.io/kops/util/pkg/vfs"
 	"k8s.io/mount-utils"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/blang/semver/v4"
 )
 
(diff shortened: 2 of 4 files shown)


rifelpet:
This would break clusters with IMDS disabled. I think it would be reasonable to try IMDS first and fallback to DescribeInstances.

hakman:
> This would break clusters with IMDS disabled. I think it would be reasonable to try IMDS first and fallback to DescribeInstances.

Can / should a kOps cluster work without IMDS?

hakman:
Thanks @bwagner5. Seems like a good improvement from my point of view.
@rifelpet Leaving this to you for the final review.
/lgmt

bwagner5:
>> This would break clusters with IMDS disabled. I think it would be reasonable to try IMDS first and fallback to DescribeInstances.

> Can / should a kOps cluster work without IMDS?
Previous implementation also used IMDS to get metadata://aws/meta-data/instance-id.

There's some other places that use IMDS in nodeup as well so it might be worth going through and figuring out if it's reasonable to perform a fallback if IMDS is disabled everywhere. But it's almost impossible to do really large scale-ups if you're not using IMDS on node bootstrap so I'd lean towards getting rid of the API calls and requiring IMDS. 

I have a version of nodeup that does zero API calls and relies only on IMDS and OS calls for node info rather than DescribeInstanceTypes.  I may try to PR some of those after they're cleaned up.

johngmyers:
Don't you have to use IMDS to get the instance role credentials to authenticate to kops-controller? I don't think it's going to be possible to bootstrap without IMDS.

hakman:
> There's some other places that use IMDS in nodeup as well so it might be worth going through and figuring out if it's reasonable to perform a fallback if IMDS is disabled everywhere. But it's almost impossible to do really large scale-ups if you're not using IMDS on node bootstrap so I'd lean towards getting rid of the API calls and requiring IMDS.

In any case one would need instance ID, which is pretty hard to get without IMDS, right?

> I have a version of nodeup that does zero API calls and relies only on IMDS and OS calls for node info rather than DescribeInstanceTypes. I may try to PR some of those after they're cleaned up.

That would be very nice.
I think there are also some places where metadata api is called directly instead of using vfs to read it. It may be worth a PR to change all to use vfs.

olemarkus:
> In any case one would need instance ID, which is pretty hard to get without IMDS, right?

Instances with RBN enabled can just get this from `os.hostname` :)

rifelpet (approved):
It sounds like IMDS is required at this point then. This lgtm and I agree with proposed follow up PRs.

rifelpet:
Should we remove eco:DescribeInstances from the node IAM policy? Users may be depending on it for other workloads though, and having them add it back makes the policy size larger since it would become its own statement.

johngmyers:
Doesn't kubelet require `DescribeInstances`?

olemarkus:
We can remove it from the nodeup function (if it is there). Kubelet still needs it if the in-tree CCM is used, but it should be a part of the CCM function too.

PR #12848, opened 2021-11-27
Open nodeport to IPv6 in e2e tests

johngmyers:
/cc @rifelpet


rifelpet (approved):

johngmyers:
This broke e2e tests of kOps versions 1.20 and 1.21, as those don't accept IPv6 addresses in `nodePortAccess`.

I don't think the deployer has a good idea of which minor version of kOps it is deploying.

Any suggestions @rifelpet ?

PR #12851, opened 2021-11-28
Support NodeLocalDNS on IPv6 clusters

johngmyers:
The `dns64` plugin is not supported by node-cache, so I had to temporarily remove that from our template. I filed kubernetes/dns#489 to add the plugin.

IPv6 doesn't have unscoped link-local addresses, so I had to invent an ULA.

### DIFF
--- a/upup/models/cloudup/resources/addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml.template
@@ -54,7 +54,7 @@ data:
           force_tcp
         }
         prometheus :9253
-        health {{ KubeDNS.NodeLocalDNS.LocalIP }}:{{ NodeLocalDNSHealthCheck }}
+        health {{ joinHostPort KubeDNS.NodeLocalDNS.LocalIP NodeLocalDNSHealthCheck }}
     }
     {{- if WithDefaultBool KubeDNS.NodeLocalDNS.ForwardToKubeDNS false }}
     .:53 {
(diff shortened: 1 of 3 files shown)


upup/models/cloudup/resources/addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -54,7 +54,7 @@ data:
           force_tcp
         }
         prometheus :9253
-        health {{ KubeDNS.NodeLocalDNS.LocalIP }}:{{ NodeLocalDNSHealthCheck }}
+        health {{ if contains ":" KubeDNS.NodeLocalDNS.LocalIP }}[{{ end }}{{ KubeDNS.NodeLocalDNS.LocalIP }}{{ if contains ":" KubeDNS.NodeLocalDNS.LocalIP }}]{{ end }}:{{ NodeLocalDNSHealthCheck }}
```
olemarkus, line 57: Maybe we should expose `net.JoinHostPort` to templates. Bound to run into this many times.

upup/models/cloudup/resources/addons/nodelocaldns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -99,9 +99,6 @@ data:
         bind {{ KubeDNS.NodeLocalDNS.LocalIP }}
         forward . __PILLAR__UPSTREAM__SERVERS__
         prometheus :9253
-        {{- if IsIPv6Only }}
```
olemarkus, line 102: Should this be removed? Non-local lookups will be directly passed to AWS DNS, which will bypass dns64 if that is handled by coredns.

johngmyers: As mentioned in the initial comment, the coredns in node-cache does not have the dns64 plugin. So leaving this in will cause the container to go into crashloop.

olemarkus (approved):

PR #12860, opened 2021-11-30
If we use resource based naming, use the hostname as node name

olemarkus:


johngmyers:
I sent a competing PR #12862

PR #12862, opened 2021-12-01
Use instance ID as node name when AWS CCM supports it

johngmyers:
/cc @olemarkus

### DIFF
--- a/cmd/kops-controller/pkg/config/options.go
+++ b/cmd/kops-controller/pkg/config/options.go
@@ -55,6 +55,9 @@ type ServerOptions struct {
 	SigningCAs []string `json:"signingCAs"`
 	// CertNames is the list of active certificate names.
 	CertNames []string `json:"certNames"`
+
+	// UseInstanceIDForNodeName uses the instance ID instead of the hostname for the node name.
+	UseInstanceIDForNodeName bool `json:"useInstanceIDForNodeName,omitempty"`
 }
 
 type ServerProviderOptions struct {
--- a/upup/pkg/fi/nodeup/command.go
+++ b/upup/pkg/fi/nodeup/command.go
@@ -447,7 +447,7 @@ func completeWarmingLifecycleAction(cloud awsup.AWSCloud, modelContext *model.No
 }
 
 func evaluateSpec(c *NodeUpCommand, nodeupConfig *nodeup.Config) error {
-	hostnameOverride, err := evaluateHostnameOverride(api.CloudProviderID(c.cluster.Spec.CloudProvider))
+	hostnameOverride, err := evaluateHostnameOverride(api.CloudProviderID(c.cluster.Spec.CloudProvider), nodeupConfig.UseInstanceIDForNodeName)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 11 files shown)


cmd/kops-controller/pkg/config/options.go
```diff
@@ -55,6 +55,9 @@ type ServerOptions struct {
 	SigningCAs []string `json:"signingCAs"`
 	// CertNames is the list of active certificate names.
 	CertNames []string `json:"certNames"`
+
+	// UseInstanceIDForNodeName uses the instance ID instead of the hostname for the node name.
```
olemarkus, line 59: Currently only AWS supports verified nodes, but should we still say that this is AWS specific?

olemarkus (commented):
I don't mind merging this at all, but this perhaps should be discussed in office hours first?

upup/pkg/fi/nodeup/command.go
```diff
@@ -477,14 +477,19 @@ func evaluateSpec(c *NodeUpCommand, nodeupConfig *nodeup.Config) error {
...
+		source := "local-hostname"
+		if useInstanceIDForNodeName {
+			source = "instance-id"
+		}
+		nodeNameBytes, err := vfs.Context.ReadFile("metadata://aws/meta-data/" + source)
 		if err != nil {
-			return "", fmt.Errorf("error reading local-hostname from AWS metadata: %v", err)
+			return "", fmt.Errorf("error reading %s from AWS metadata: %v", source, err)
 		}
-		return string(hostnameBytes), nil
```
olemarkus, line 487: If we use the instance ID we should not return it here. That would cause hostname override to be set on k8s components, which we want to avoid when we can.

johngmyers: As far as my IDE can see, this goes into kubelet and kube-proxy. Both use it as the node name. So I don't see how we can avoid either.

olemarkus: `NodeName` has a fallback to `os.hostname` which is also what kubelet and kube-proxy use as default.

cmd/kops-controller/pkg/config/options.go
johngmyers: GCE also has a verifier, but it doesn't have this issue of its CCM supporting two node naming conventions.

upup/pkg/fi/nodeup/command.go
olemarkus: Right. I suggest an additional test for RBN and if so return empty string.

johngmyers: I'm failing to see the benefit of adding such logic.

olemarkus: The whole motivation for using just the hostname was to not set hostname override. It's one of the failing CIS checks we have from https://github.com/kubernetes/kops/issues/6150.

johngmyers: Ah, another silly, inappropriate CIS benchmark thing.

But can we really remove the override? The hostname would be a FQDN, whereas the CCM only takes an unqualifed instance ID.

olemarkus: Yes. The CIS benchmark is quite silly. But you know how it is with compliance, I am sure.

The fallback to `os.hostname()` would produce only the instance ID, not FQDN.

johngmyers: I think a good follow-up PR would be to return nothing if `os.hostname()` matches what this function would otherwise return.

rifelpet:
This is worthy of a release note since it may be surprising to users

upup/pkg/fi/nodeup/command.go
olemarkus: I can do that.

rifelpet (approved):

PR #12863, opened 2021-12-01
Don't assign CIDRs to shared subnets

johngmyers:


upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -201,12 +201,20 @@ func (s *Subnet) CheckChanges(a, e, changes *Subnet) error {
 	return nil
 }
 
+func (_ *Subnet) ShouldCreate(a, e, changes *Subnet) (bool, error) {
+	if fi.BoolValue(e.Shared) {
+		changes.IPv6CIDR = nil
```
hakman, line 206: Why not just check in `CheckChanges()`?

johngmyers: I suppose the question is if we want to error out or silently ignore.

hakman: I would say error out so that an operator doesn't have to investigate why it's not applied.

hakman (approved):

PR #12864, opened 2021-12-01
Set Resource Based Naming on managed subnets

johngmyers:


upup/pkg/fi/cloudup/awstasks/subnet.go
```diff
@@ -267,6 +286,41 @@ func (_ *Subnet) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Subnet) error {
...
+		request := &ec2.ModifySubnetAttributeInput{
+			SubnetId:                       e.ID,
+			PrivateDnsHostnameTypeOnLaunch: &hostnameType,
+		}
+		_, err := t.Cloud.EC2().ModifySubnetAttribute(request)
+		if err != nil {
+			return fmt.Errorf("error modifying hostname type: %w", err)
+		}
+
+		request = &ec2.ModifySubnetAttributeInput{
```
olemarkus, line 303: I discovered that in ipv6-native subnets, A records cannot be enabled. So here we need to know if the subnet has that enabled or not.

johngmyers: kOps doesn't support ipv6-native subnets yet. When we add that support, we will have to update this logic appropriately.

johngmyers:
During Office Hours, I heard two concerns:

* One, that some existing subsystem will depend on the IPv4-based names.
* Two, what I interpret as general conservatism around changing existing clusters.

I don't know what subsystems these would be that depend on the IPv4-based domain names. We've taken care of CCM and that depended on the node name, not the domain name.

This change only takes effect upon upgrading to k8s 1.23 or switching to CCM when on k8s 1.23 or later. For most people, this will only take effect upon upgrading to k8s 1.24, when we currently plan to enable CCM by default. At that time, existing nodes will continue as-is, whereas new nodes will get the resource-based node names. With this patch, they will get the resource-based domain names to match.

If admins really don't want the resource-based domain names, they can create their own shared subnets. I believe we can be a little opinionated about managed subnets as they generally only handle the simpler cases. kOps will not, for example, manage endpoint associations or peering connections.

hakman (approved):

PR #12865, opened 2021-12-01
Replace gofmt with gofumpt

hakman:
Ref: https://github.com/mvdan/gofumpt

/cc @olemarkus @rifelpet @johngmyers 

/hold for feedback

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -363,7 +363,7 @@ gomod: bazelisk gomod-prereqs
 
 .PHONY: gofmt
 gofmt:
-	find $(KOPS_ROOT) -name "*.go" | grep -v vendor | xargs ${BAZEL} run @go_sdk//:bin/gofmt -- -w -s
+	hack/update-gofmt.sh
 
 .PHONY: goimports
 goimports:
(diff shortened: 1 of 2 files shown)


johngmyers:
IntelliJ IDEA supports gofmt before commit. It does not support gofmpt before commit.  This change would make development harder.

olemarkus:
As in goland? Can't you configure it to run whatever? As in https://github.com/mvdan/gofumpt#goland
or run the hack script as a precommit?

johngmyers:
I also work with Java so use IDEA instead of Goland. Those Goland instructions don't match what's in IDEA and I can't find any corresponding settings.

hakman:
@johngmyers @olemarkus Should we commit the script and the one-time run as a first step and keep gofmt around until we find a more permanent solution?

olemarkus:
I am for that. I will probably keep using gofumpt if you don't mind, and doing that one-time pass will make my commits easier for you :)

johngmyers:
I have no objections to script and one-time pass.

PR #12875, opened 2021-12-02
tests gce: fix project arg on gsutil rm

justinsb:
Because we don't actually need to pass the project when deleting
objects (as it can be inferred from the bucket), the syntax for
passing it is a little different.


rifelpet (approved):
good catch, I see this in the logs:

```
I1202 13:38:44.638933    4791 gcs.go:89] gsutil rm -r gs://k8s-jkns-e2e-gce-state-44 -p k8s-jkns-e2e-gce
CommandException: "rm" command does not support "file://" URLs. Did you mean to use a gs:// URL?
```

PR #12877, opened 2021-12-03
Remove support for CentOS 8

johngmyers:


hakman:
@johngmyers This also needs a release note, right?

johngmyers:
Ah, put it in the copy that was under `.build`

hakman (approved):
We should go ahead and remove RHEL & CentOS 7 support in a future PR.
Seems like a good time for it.

PR #12878, opened 2021-12-03
If RBN, use IPv6 address instead of IPv4

olemarkus:
This code seems to only be used for gossip, but it is always run. This fixes IPv6 only networking

### DIFF
--- a/protokube/pkg/protokube/aws_volume.go
+++ b/protokube/pkg/protokube/aws_volume.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"k8s.io/klog/v2"
+
 	"k8s.io/kops/protokube/pkg/etcd"
 	"k8s.io/kops/protokube/pkg/gossip"
 	gossipaws "k8s.io/kops/protokube/pkg/gossip/aws"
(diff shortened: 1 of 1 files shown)


protokube/pkg/protokube/aws_volume.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"k8s.io/klog/v2"
+
```
hakman, line 34: I thought that running the gogumpt script would fix these issues...

olemarkus: Not sure what caused this one

hakman (approved):

protokube/pkg/protokube/aws_volume.go
```diff
@@ -125,7 +126,11 @@ func (a *AWSVolumes) discoverTags() error {
 
 	a.clusterTag = clusterID
 
-	a.internalIP = net.ParseIP(aws.StringValue(instance.PrivateIpAddress))
+	if *instance.PrivateDnsNameOptions.HostnameType == ec2.HostnameTypeResourceName {
```
johngmyers, line 129: RBN does not necessarily mean the  instance has IPv6.

PR #12879, opened 2021-12-03
Add support for IPv6 only clusters

olemarkus:
This PR is an umbrella PR that contains all the changes for supporting IPv6 only clusters, meaning clusters that run on AWS IPv6 native subnets. 

Some notes:
* kOps will not provision the subnets. You need to create these outside of kOps for now and then reference them in the cluster spec.
* You should not set networkCIDR or any IPv4 CIDRs on subnets.
* Ingress is tricky. AWS do not support instance targets in elbv2 for IPv6 addresses, so there is effectively no way of creating an ELB for the API. DNS works, but you'll need IPv6 on client side.
* Many addons may break.


olemarkus:
No worries. Most of this has been folded into the work John has been doing, so this is no longer necessary.

PR #12880, opened 2021-12-03
Revert "If RBN, use IPv6 address instead of IPv4"

johngmyers:
This reverts commit 878bedf2326d79b00ff988bf3674dd3a3d1d5c6a.

Having a Resource Based Name does not mean an instance has an IPv6 address.


protokube/pkg/protokube/aws_volume.go
```diff
@@ -126,11 +125,7 @@ func (a *AWSVolumes) discoverTags() error {
 
 	a.clusterTag = clusterID
 
-	if *instance.PrivateDnsNameOptions.HostnameType == ec2.HostnameTypeResourceName {
```
olemarkus, line 129: Could you fix this condition to test for the availability of an IPv6 address instead (as the line below) instead of reverting?

PR #12881, opened 2021-12-03
enable connection-draining for aws classic lb

heybronson:
This PR enables connection draining for aws classic LBs by default. 

Currently, kops doesn't enable connection draining and dropped connections can happen while cycling instanceGroups (master instanceGroups in particular).


heybronson:
I've signed the CNCF agreement.

johngmyers (approved):

PR #12890, opened 2021-12-04
WIP: GCE: We should now accept CIDRs

justinsb:
WIP because the network doesn't actually get a CIDR on GCE


zetaab:
@justinsb any change to get this merged?

justinsb:
/reopen

I need to revisit this, the robot is really unhelpful

PR #12891, opened 2021-12-04
e2e: fix test flakes where we specify a non-schedulable zone

justinsb:
If we specify a zone to the e2e tests, it will try to pre-provision
volumes into it for the inline tests.  We then will be unable to
attach a volume to test pods if we don't have (normal) nodes in that
zone.

Pick a zone that is schedulable for test pods.

### DIFF
--- a/tests/e2e/pkg/tester/tester.go
+++ b/tests/e2e/pkg/tester/tester.go
@@ -27,10 +27,10 @@ import (
 	"github.com/octago/sflags/gen/gpflag"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
-	"sigs.k8s.io/kubetest2/pkg/testers/ginkgo"
-
+	unversioned "k8s.io/kops/pkg/apis/kops"
 	api "k8s.io/kops/pkg/apis/kops/v1alpha2"
 	"k8s.io/kops/tests/e2e/pkg/kops"
+	"sigs.k8s.io/kubetest2/pkg/testers/ginkgo"
 )
 
 // Tester wraps kubetest2's ginkgo tester with additional functionality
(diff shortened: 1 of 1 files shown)


pkg/apis/kops/v1alpha2/instancegroup.go
```diff
@@ -49,6 +49,17 @@ type InstanceGroupList struct {
...
+	// InstanceGroupRoleMaster is a master role
+	InstanceGroupRoleMaster InstanceGroupRole = "Master"
+	// InstanceGroupRoleNode is a node role
+	InstanceGroupRoleNode InstanceGroupRole = "Node"
+	// InstanceGroupRoleBastion is a bastion role
+	InstanceGroupRoleBastion InstanceGroupRole = "Bastion"
+	// InstanceGroupRoleAPIServer is an API server role
+	InstanceGroupRoleAPIServer InstanceGroupRole = "APIServer"
+)
+
```
johngmyers, line 62: What are these used for?

justinsb: BTW I don't _really_ care too much here, if we don't want to duplicate the constants I can just import from the unversioned package.  I do think this is more correct though (even though it is more tedious).

johngmyers: There doesn't appear to be a need to have versioned constants. The versioned structs are only ever accessed through serialization/deserialization and the apimachinery conversion code.

justinsb: OK, I'll remove.

The side-question it opens up is whether we should be using the unversioned types.  For example, here we're parsing the output from `kops get cluster`, which is naturally versioned.  If we do need to convert it, should we convert to a versioned type?  (i.e. is the unversioned type actually just apimachinery internals "leaking").  But ... a topic for another day!

hakman (approved):

PR #12899, opened 2021-12-06
CCM now supports RBN

johngmyers:


hakman:
Shouldn't we wait for them to do a release?

johngmyers:
Since we got a new release of the NodeLocalDNS image, I'll fold this into that PR.

johngmyers:
Per normal, we're using CCM's "latest" tag for 1.23+. Our normal 1.23 release procedure would be to fix onto their 1.23 (probably alpha) tag after they cut it.

PR #12900, opened 2021-12-06
Simplify Flatcar containerd exec command

pothos:
The containerd command used in
https://github.com/kubernetes/kops/pull/12177 is a modification from
the torcx containerd unit. However, how torcx starts containerd is a
implementation detail and it's better to not hardcode torcx in case it
isn't used anymore.
Change the ExecStard command to use /usr/bin/containerd directly,
making it simpler and more future-proof.


hakman:
Thanks @pothos. :)
/ok-to-test

pothos:
Now with `gofmt`, sorry ;)

hakman (approved):
Thanks @pothos. Looking forward to see the periodic tests go green once the IMDSv2 issue is also addressed. Will take care of the cherry-picks soon.

PR #12904, opened 2021-12-07
Upgrade node-cache to 1.21.3

johngmyers:
Adds support for DNS64 when using NodeLocalDNS on an IPv6 cluster.


johngmyers:
Unfortunately I don't see a presubmit for NodeLocalDNS.

hakman (approved):

PR #12906, opened 2021-12-07
WIP Support Karpenter

olemarkus:


justinsb:
This is exciting ... I would really love to see us do this as an external addon (#12727 gets us really close I think), but I don't want to block on that.

pkg/apis/kops/instancegroup.go
```diff
@@ -82,8 +82,17 @@ const (
...
+
+const (
+	InstanceManagerCloudGroup InstanceManager = "CloudGroup"
+	InstanceManagerKarpenter  InstanceManager = "Karpenter"
+)
+
 // InstanceGroupSpec is the specification for an InstanceGroup
 type InstanceGroupSpec struct {
+	// InstanceManager determines what is managing the node lifecycle
+	InstanceManager InstanceManager `json:"instanceManager,omitempty"`
```
johngmyers, line 95: I think `provisioner` is a better name. If not that, the "instance" part is redundant, suggesting `manager`.

pkg/model/components/addonmanifests/karpenter/iam.go
```diff
@@ -0,0 +1,66 @@
...
+	}, true
+}
+
+func addKarpenterPermissions(p *iam.Policy) {
+	p.UnconditionalAction.Insert(
+		//		"ec2:CreateLaunchTemplate",
+		"ec2:CreateFleet",
+		"ec2:RunInstances",
+		"ec2:CreateTags",
+		"iam:PassRole",
```
johngmyers, line 55: Constrain to the nodes instance role?

pkg/model/components/addonmanifests/karpenter/iam.go
```diff
@@ -0,0 +1,66 @@
...
+		"ec2:TerminateInstances",
```
johngmyers, line 56: Constrain to tagged for cluster?

pkg/model/components/addonmanifests/karpenter/iam.go
```diff
@@ -0,0 +1,66 @@
...
+		Name:      "karpenter",
+	}, true
+}
+
+func addKarpenterPermissions(p *iam.Policy) {
+	p.UnconditionalAction.Insert(
+		//		"ec2:CreateLaunchTemplate",
+		"ec2:CreateFleet",
+		"ec2:RunInstances",
+		"ec2:CreateTags",
```
johngmyers, line 54: Is it possible to constrain to tag-on-launch?

pkg/model/iam/iam_builder.go
```diff
@@ -49,26 +50,26 @@ const PolicyDefaultVersion = "2012-10-17"
 // Policy Struct is a collection of fields that form a valid AWS policy document
 type Policy struct {
 	clusterName               string
-	unconditionalAction       sets.String
-	clusterTaggedAction       sets.String
-	clusterTaggedCreateAction sets.String
+	UnconditionalAction       sets.String `json:"-"`
+	ClusterTaggedAction       sets.String `json:"-"`
+	ClusterTaggedCreateAction sets.String `json:"-"`
```
johngmyers, line 55: Perhaps only expose receivers that insert into these? I don't think we want external modules to be able to remove permissions from a policy.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -0,0 +1,677 @@
...
+# Source: karpenter/templates/controller/rbac.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: karpenter-controller
+  namespace: kube-system
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
johngmyers, line 341: File an issue upstream to constrain this to the configmaps they need to modify?

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -0,0 +1,677 @@
...
+---
+# Source: karpenter/templates/webhook/rbac.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: karpenter-webhook
+rules:
+- apiGroups: ["admissionregistration.k8s.io"]
+  resources: ["validatingwebhookconfigurations", "mutatingwebhookconfigurations"]
+  verbs: ["get", "watch", "list", "update"]
```
johngmyers, line 302: File an issue to constrain to the configurations they need to update?

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -0,0 +1,677 @@
...
+metadata:
+  name: karpenter-webhook
+  namespace: kube-system
+rules:
+- apiGroups: [""]
+  resources: ["configmaps", "namespaces"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "list", "watch", "update"]
```
johngmyers, line 361: Constrain to the secrets they need?

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -0,0 +1,677 @@
...
+- apiGroups: ["coordination.k8s.io"]
+  resources: ["leases"]
+  verbs: ["get", "watch", "create", "update"]
```
johngmyers, line 364: Constrain to the leases they use?

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -0,0 +1,677 @@
...
+  labels:
+  {{ range $key, $value := . }}
+    {{ $key }}: {{ $value }}
+  {{ end }}
+{{ end }}
+  limits:
+    resources:
+      cpu: 1000
+  provider:
+    instanceProfile: nodes.{{ ClusterName }}
```
johngmyers, line 669: Should support using the iam.profile field.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -0,0 +1,677 @@
...
+    launchTemplate: {{ $name }}
+    securityGroupSelector:
+      Name: nodes.{{ ClusterName }}
```
johngmyers, line 672: There's a todo to support additionalSecurityGroups?

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -0,0 +1,677 @@
...
+    subnetSelector:
+      kops.k8s.io/karpenter/{{ $name }}: "*"
```
johngmyers, line 674: There's a todo to support subnets?

pkg/apis/kops/instancegroup.go
olemarkus: Sounds good. I think I will go with `manager`

pkg/model/components/addonmanifests/karpenter/iam.go
olemarkus: The IAM permissions are way too wide right now. But I wanted to get something working into the tree and get an e2e running before trying to fix these.
But It is definitely on my agenda to tighten these up.

pkg/model/components/addonmanifests/karpenter/iam.go
olemarkus: Yeah. That should be possible. Will look at that in a follow-up.

pkg/model/iam/iam_builder.go
olemarkus: That makes sense. Will do that.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
olemarkus: Same as with AWS IAM, I also found these too wide. Will tighten these in a follow-up.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
olemarkus: I wonder if SGs actually goes into the LT. It may be we do not need to set this and that Karpenter only use them for their own LTs (which we won't support).

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
olemarkus: We do. The code iterates over the mentioned subnets and add this tag to it. It may be that this doesn't work with shared subnets. But that's the intention.
Ideally, Karpenter would allow for multiple selectors, but currently they don't.

olemarkus:
Thanks for the review. I'll fix `InstanceManager` -> `Manager` and the pointers for adding policies.
The rest can be done in follow-ups, I think.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
johngmyers: We turn off `cluster.spec.tagSubnets` and manage the subnets with Terraform. It would be a real pain to have to tag them for each IG, especially since the Terraform code/config doesn't otherwise need to be changed much.

Possibly this should be feedback for upstream.

In our use cases, we have some IGs that spread across all 3 AZs and some that only go into one AZ, "a". So allowing a configurable subnetSelector would work well for us.

olemarkus: Should be easy enough to make subnetSelector configurable, but usability-wise having multiple selectors would be better.

I also don't think one can directly specify subnet IDs. Our `Find` do not require subnet name in spec to match the Name tag if ID is specified, so if there is a mismatch there, using a tag-based selector would fail.

I suggest at minimum having an upstream issue for multiple selectors.

olemarkus:
Added some TODOs in the original post.

johngmyers:
Filed aws/karpenter#973 on the too-permissive RBAC.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -280,6 +281,10 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		dest["EnableSQSTerminationDraining"] = func() bool { return *cluster.Spec.NodeTerminationHandler.EnableSQSTerminationDraining }
 	}
 
+	dest["SupportedArchitecture"] = tf.SupportedArchitecture
+
+	dest["ParseTaint"] = parseTaint
```
johngmyers, line 286: Maybe `ExpandTaint` ?

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -280,6 +281,10 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 		dest["EnableSQSTerminationDraining"] = func() bool { return *cluster.Spec.NodeTerminationHandler.EnableSQSTerminationDraining }
 	}
 
+	dest["SupportedArchitecture"] = tf.SupportedArchitecture
```
johngmyers, line 284: `ArchitectureOfAMI` ?

pkg/model/iam/iam_builder.go
```diff
@@ -50,26 +50,30 @@ const PolicyDefaultVersion = "2012-10-17"
 // Policy Struct is a collection of fields that form a valid AWS policy document
 type Policy struct {
 	clusterName               string
-	UnconditionalAction       sets.String `json:"-"`
-	ClusterTaggedAction       sets.String `json:"-"`
-	ClusterTaggedCreateAction sets.String `json:"-"`
+	unconditionalAction       sets.String `json:"-"`
+	clusterTaggedAction       sets.String `json:"-"`
+	clusterTaggedCreateAction sets.String `json:"-"`
```
johngmyers, line 55: Don't need the json tags.

johngmyers (changes requested):

johngmyers:
I went to file a ticket about `subnetSelector` and had challenges figuring out what to ask for.

Multiple `subnetSelector` doesn't really seem to help. To implement our existing API we'd want to be able to specify multiple subnet IDs. Tagging each subnet with its own ID and then using multiple selectors would work, but is indirect.

I'm wondering if we should allow specifying the `subnetSelector` in the IG spec and default it to `kubernetes.io/role/internal-elb`. This would even be a good feature to provide for ASG-provisioned IGs.

upup/pkg/fi/cloudup/template_functions.go
olemarkus: The function name came from copying the k/k function. I suggest we keep that link even though I may agree ExpandTaint is more descriptive.

olemarkus:
If we want to use subnet selectors, we can change the Tag I introduced above to `kops.k8s.io/instancegroup/{{ $name }}` and keep the logic. We'll also let people set their own selector.

I think there is a huge advantage in that kOps doesn't have to know anything about the concrete IDs to use for various cloud resources.

olemarkus:
I am wondering if we can merge in this PR now though. What we are changing/discussing should come as follow ups to ensure the amount of stuff in one PR doesn't blow up.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
olemarkus: I don't think this is actually used. See https://github.com/aws/karpenter/issues/966

johngmyers:
I think tagging the subnets with instance group names is a poor approach and I would prefer we not merge that part of the code.

johngmyers:
If you have shared subnets and `tagSubnets` is enabled, then this code will add the IG tags to the subnets but never remove them. So the tags can get out of sync with the IG spec.

olemarkus:
I don't think there any other way for kops to add subnets otherwise now. We use a feature flag so we can remove this behavior later

johngmyers:
I think for now we should use `kubernetes.io/role/internal-elb` and the cluster tag, so we don't support restricting the AZs.

olemarkus:
Fair enough. Will fix.

olemarkus:
It's a bit of a tag abuse here. internal-elb means something else, and is not necessarily where one wants Nodes to land. But for now this works and we can fix that later.

johngmyers (approved):

PR #12911, opened 2021-12-08
Tag on create for remaining CCM privileges

olemarkus:
The custom image tests https://github.com/kubernetes/cloud-provider-aws/pull/293

I can't get the actual `RequestTag` constraint working with security groups for some reason though. But otherwise, these constraints should be fine.

This PR also removes CCM EBS privileges since CSI driver is needed for CCM anyway.


olemarkus:
This is 1.24 material anyway, so I will wait for the LBC release rather than trying to hack around the failing CCM+addons test.

johngmyers (approved):

PR #12915, opened 2021-12-09
Set DNS PDB to a maxUnavailable percentage

heybronson:
Both kubeDNS and coreDNS have cluster proportional autoscalers which scale the corresponding DNS deployments as the cluster grows in size. However, the corresponding PDB for this deployment only ensures that 1 corresponding DNS pod is running. Ensuring a minimum of 1 pod is not enough as clusters to scale.

This PR updates the DNS PDBs for kubeDNS and coreDNS to `maxUnavailable: 50%`.

**Tests**
The min number of DNS pods is 2 due to the corresponding autoscalers having `preventSinglePointFailure: true`. PDB in this scenario:
```
NAME       MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS  
kube-dns   N/A             50%               1                              
```

Scenario where DNS autoscaler increased the DNS replica count to 22:
```
NAME       MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS  
kube-dns   N/A             50%               11                   
```

### DIFF
--- a/tests/integration/update_cluster/apiservernodes/data/aws_s3_bucket_object_minimal.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/apiservernodes/data/aws_s3_bucket_object_minimal.example.com-addons-bootstrap_content
@@ -20,7 +20,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.12
     manifest: coredns.addons.k8s.io/k8s-1.12.yaml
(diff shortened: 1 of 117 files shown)


olemarkus:
Hey. Thanks for this. Sounds like a good idea.

/ok-to-test

olemarkus (approved):

PR #12922, opened 2021-12-10
WIP: Bare metal work

justinsb:
This is an integration branch to let others see / contribute to my experiments in bare-metal support.


zetaab:
@justinsb how the provisioning architecture is going to look like? in clouds kops uses cloud apis quite heavily, how you are planning to do that in metal?

PR #12923, opened 2021-12-10
nodeup: store the CloudProvider in the context

justinsb:
This is a bit simpler than fetching it from the cluster every time, and also can allow things like mixed-cloud clusters (in future).


hakman (approved):
/retest

PR #12932, opened 2021-12-11
Set the default LT version to the new LT version

olemarkus:
Needed for Karpenter, which will always use the default version of a Launch Template.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
+++ b/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
@@ -20,10 +20,12 @@ import (
 	"encoding/base64"
 	"fmt"
 	"sort"
+	"strconv"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"k8s.io/klog/v2"
+
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
 )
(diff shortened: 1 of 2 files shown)


upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -164,8 +166,19 @@ func (t *LaunchTemplate) RenderAWS(c *awsup.AWSAPITarget, a, e, changes *LaunchT
 			LaunchTemplateName: t.Name,
 			LaunchTemplateData: data,
 		}
-		if _, err = c.Cloud.EC2().CreateLaunchTemplateVersion(input); err != nil {
+		if version, err := c.Cloud.EC2().CreateLaunchTemplateVersion(input); err != nil {
 			return fmt.Errorf("error creating LaunchTemplateVersion: %v", err)
+		} else {
```
hakman, line 171: Please keep the same style.
```suggestion
		version, err := c.Cloud.EC2().CreateLaunchTemplateVersion(input)
		if err != nil {
			return fmt.Errorf("error creating LaunchTemplateVersion: %v", err)
		}
		...
```

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
```diff
@@ -164,8 +166,19 @@ func (t *LaunchTemplate) RenderAWS(c *awsup.AWSAPITarget, a, e, changes *LaunchT
...
+			newDefault := strconv.FormatInt(*version.LaunchTemplateVersion.VersionNumber, 10)
+			_, err := c.Cloud.EC2().ModifyLaunchTemplate(
+				&ec2.ModifyLaunchTemplateInput{
+					DefaultVersion:   &newDefault,
+					LaunchTemplateId: version.LaunchTemplateVersion.LaunchTemplateId,
+				},
+			)
```
hakman, line 178: Could we do this just for Karpenter? Users already complain about the amount of API calls.

hakman (changes requested):

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
olemarkus: Users should not hit rate limits of this call at least. It's mostly non-mutating call that is a challenge, but these are separate buckets.

LTs do not know about karpenter, and I can't think of a reason why they would. I also can't think of how to inject such a value in a good way.

hakman: OK, we can go with it as is and optimize later.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
hakman: I really dislike the `if err {} else {}` pattern, it's hard to follow, but can be changed later.

hakman (approved):

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_api.go
johngmyers: Concur. There should not be an `else` after an if block that always returns.

PR #12937, opened 2021-12-11
Recognize debian bullseye as having "broken" resolv.conf

justinsb:
Debian bullseye also has the loopback resolv.conf that needs special
handling.

### DIFF
--- a/util/pkg/distributions/distributions.go
+++ b/util/pkg/distributions/distributions.go
@@ -95,6 +95,11 @@ func (d *Distribution) HasLoopbackEtcResolvConf() bool {
 		// Ubuntu > 16.04 has it
 		return d.version > 16.04
 	}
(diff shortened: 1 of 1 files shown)


justinsb:
TBH I'm not sure whether we should try to recognize the "loopback" resolv.conf - I think we could check for `nameserver 127.0.0.53`

hakman:
> TBH I'm not sure whether we should try to recognize the "loopback" resolv.conf - I think we could check for `nameserver 127.0.0.53`

We could try this or try to replace/own resolv.conf. It may not be that complicated.

hakman (approved):

olemarkus:
This PR seems to have broken the tests that runs on Debian Bullseye.

```
# cat /etc/resolv.conf 
domain us-west-2.compute.internal
search us-west-2.compute.internal
nameserver 172.20.0.2
```

PR #12949, opened 2021-12-13
WIP: Use GCE dedicated service accounts

justinsb:
Integration branch that shows the "big picture" of using per-InstanceGroup service accounts on GCE.

Will whittle this down to more reasonable PRs, but wanted to make my end-goal available.


pkg/model/gcemodel/service_accounts.go
```diff
@@ -0,0 +1,142 @@
...
+
+	Lifecycle fi.Lifecycle
+}
+
+var _ fi.ModelBuilder = &ServiceAccountsBuilder{}
+
+func (b *ServiceAccountsBuilder) Build(c *fi.ModelBuilderContext) error {
+	if b.Cluster.Spec.CloudConfig.GCEServiceAccount != "" {
+		serviceAccount := &gcetasks.ServiceAccount{
+			Name:      s("shared"),
```
zetaab, line 38: what is purpose of this shared serviceaccount?

justinsb: So this is the current behaviour.  If a GCEServiceAccount is specified at the cluster level, we currently use it for all instancegroups.  The current default is that all instancegroups use that cluster-level service account, or the GCE "default" service account.

I think we should change this so that each instancegroup role gets its own ServiceAccount.  I think we'll have to continue to honor the cluster-level GCEServiceAccount if it's set, but I think we should warn if it is set.

zetaab: sounds good

tests/integration/update_cluster/ha_gce/kubernetes.tf
```diff
@@ -564,7 +564,7 @@ resource "google_compute_instance_template" "master-us-test1-a-ha-gce-example-co
     preemptible         = false
   }
   service_account {
-    email  = "default"
+    email  = "control-plane-ha-gce-example-com@testproject.iam.gserviceaccount.com"
```
rifelpet, line 567: when we get to the specific PR for this change, I believe this should be a terraform link rather than a literal

```suggestion
    email  = google_service_account.serviceaccount-control-plane.email
```

The TerraformLink code looks correct:

https://github.com/kubernetes/kops/blob/d48e7ab52e00c90d0a3564a81d500dc061f6fa2f/upup/pkg/fi/cloudup/gcetasks/serviceaccount.go#L213-L226

which makes me wonder if we're setting the SA task's `Shared` field incorrectly... 🤔

justinsb: Ah, you're quite right - I see you did this in #12995 - thank you!

justinsb:
/reopen

We need to reach a conclusion here.

PR #12960, opened 2021-12-14
Do not skip NodePort tests for Calico

hakman:


hakman:
@rifelpet The regular e2e test uses Calico. Should be the same as ^^^.

rifelpet (approved):
ah you're right

rifelpet:
Do we know if this will pass on previous kops and k8s versions? I guess we'll find out

PR #12961, opened 2021-12-14
Use spread constraints rather than affinity to spread pods

olemarkus:
Spread constraints are supported on all supported k8s versions. And it is better at doing our intention than what anti affinity is.

### DIFF
--- a/upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
+++ b/upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
@@ -396,26 +396,23 @@ spec:
         app.kubernetes.io/instance: aws-ebs-csi-driver
         app.kubernetes.io/version: {{ .Version }}
     spec:
-      affinity:
-        podAntiAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-          - labelSelector:
-              matchExpressions:
-              - key: app
-                operator: In
-                values:
-                - ebs-csi-controller
-            topologyKey: kubernetes.com/hostname
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 100
-            podAffinityTerm:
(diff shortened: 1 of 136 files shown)


johngmyers:
Could you please put the golden outputs changes in a separate commit when they're not small? That would make review easier.

upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
```diff
@@ -396,26 +396,25 @@ spec:
...
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: "topology.kubernetes.io/zone"
+        whenUnsatisfiable: ScheduleAnyway
+        labelSelector:
+          matchLabels:
+            app: ebs-csi-controller
+            app.kubernetes.io/name: aws-ebs-csi-driver
+            app.kubernetes.io/instance: aws-ebs-csi-driver
+            app.kubernetes.io/version: {{ .Version }}
```
johngmyers, line 408: Why these three labels?

olemarkus: A general best practice is to put the same label selectors on topologySpreadConstraints as the Deployment has. So I am just using the same seletors. Same goes for PDB. The reason we ended up with this in the first place is because this is how it looked upstream.

johngmyers: The deployment selector doesn't have `app.kubernetes.io/version`

olemarkus: Ah. Yeah, version should not be a part of the selector. That would just lead to unnecessary during upgrades.

johngmyers (approved):

PR #12962, opened 2021-12-14
Do not explicitly skip Dashboard tests

hakman:
Should be covered by the `[Feature:.+` pattern now.
xRef: https://github.com/kubernetes/kubernetes/pull/80044

/cc @rifelpet


rifelpet (approved):
for reference, that upstream PR was in k8s 1.16 so we should be safe to remove this unconditionally

PR #12966, opened 2021-12-14
add instance connection draining for NLBs

heybronson:
This builds on https://github.com/kubernetes/kops/pull/12902 to add graceful connection draining for associated targetGroups.

Today, one can define `targetGroupArns` in an instanceGroup spec to associate instances in a corresponding instanceGroup with. When instances in the instanceGroup are cycled via rolling-updates, those instances aren't first properly deregistered with the corresponding `targetGroupArns`, causing dropped connections. 

This PR fixes this issue by ensuring that we're properly draining instances from their corresponding classic load balancers and NLB targetGroups before terminating the instance. This is a big improvement to the overall client experience when running rolling-updates on production clusters.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -73,6 +73,7 @@ require (
 	golang.org/x/crypto v0.0.0-20211215153901-e495a2d5b3d3
 	golang.org/x/net v0.0.0-20211112202133-69e39bad7dc2
 	golang.org/x/oauth2 v0.0.0-20210819190943-2bc19b11175f
+	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/sys v0.0.0-20211216021012-1d35b9e2eb4e
 	google.golang.org/api v0.57.0
 	gopkg.in/gcfg.v1 v1.2.3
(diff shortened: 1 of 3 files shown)


heybronson:
cc @olemarkus @johngmyers

rifelpet:
For reference, this deregistration occurs _after_ the node has been drained of its non-daemonset pods but kube-proxy is still running, so this specifically helps services exposed via kube-proxy.

This doesn't help target groups targeting pod IPs (which can be gracefully deregistered via finalizers and the aws-load-balancer-controller during the drain process) nor does it help targeting pods listening on the hosts' nodePorts directly (since they will have been drained already and their connections disconnected)

heybronson:
> For reference, this deregistration occurs after the node has been drained of its non-daemonset pods 

@rifelpet That's correct, this deregistration happens after node draining. However, this helps in scenarios where traffic is routed to daemonset pods running on a node, since those aren't drained. For example, Nginx-ingress in daemonset mode benefits from this change.

heybronson:
friendly bump

olemarkus (commented):
/lgtm

hakman (approved):

heybronson:
@rifelpet can we potentially keep this in the 1.23 release?

PR #12973, opened 2021-12-15
Various nill pointer fixes for karpenter

olemarkus:


pkg/instancegroups/instancegroups.go
```diff
@@ -680,7 +681,7 @@ func (c *RollingUpdateCluster) UpdateSingleInstance(cloudMember *cloudinstances.
 	if detach {
 		if cloudMember.CloudInstanceGroup.InstanceGroup.IsMaster() {
 			klog.Warning("cannot detach master instances. Assuming --surge=false")
-		} else {
+		} else if cloudMember.CloudInstanceGroup.InstanceGroup.Spec.Manager != api.InstanceManagerKarpenter {
```
johngmyers, line 684: Would a klog.Warning be appropriate if it's managed by Karpenter?

olemarkus: I thought about maybe doing a V(2) info. Compared to control plane, I think not detaching is more obvious for Karpenter since there is nothing to detach from.

johngmyers: `detach` would ordinarily only be true if the IG spec explicitly specified surging, so a visible warning should be appropriate. Though that assumes that the surge isn't inherited from the cluster spec, just like control plane IGs don't inherit surge from the cluster spec.

We should probably fix the inheritance logic, lest a large default surge get interpreted as a large default maxUnavailable.

olemarkus: I actually think you only hit this when using kops delete instance. That one has a flag that assumes detaching that one instance.

johngmyers: Perhaps we should define a `CanSurge()` receiver on `InstanceGroupSpec` and use that in `UpdateSingleInstance()`.

johngmyers (approved):

PR #12983, opened 2021-12-16
Drain OpenStack loadbalancers

zetaab:
Why OpenStack cannot use current structure at all if we want drain connections? Delete kubernetes node is executed before delete instance https://github.com/kubernetes/kops/blob/master/pkg/instancegroups/instancegroups.go#L418 in OpenStack cloudprovider this means that when node is deleted, it will automatically remove all loadbalancer members -> we cannot drain anything anymore.

Now when we add `DeregisterInstance` function to clouds, we have way to execute things BEFORE node is deleted from Kubernetes. By using this we should be able to do real draining in all clouds including for instance google cloud. AWS works in little bit different way than others (it keeps terminated instances in API lists for sometime).

I also was thinking could we use https://github.com/kubernetes/kops/blob/9ed4ec13f0dbe29e2e9066c554ce968f99ffdc1d/pkg/instancegroups/instancegroups.go#L641 but in fact, we cannot use that for drain Kubernetes API (created by kOps) loadbalancer. That is why I decided to make new function. 

as you are currently making similar thing in https://github.com/kubernetes/kops/pull/12966
cc @heybronson @rifelpet

### DIFF
--- a/upup/pkg/fi/cloudup/awsup/aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/aws_cloud.go
@@ -543,7 +543,12 @@ func (c *awsCloudImplementation) DeleteInstance(i *cloudinstances.CloudInstance)
 
 // DeregisterInstance drains a cloud instance and loadbalancers.
 func (c *awsCloudImplementation) DeregisterInstance(i *cloudinstances.CloudInstance) error {
-	klog.V(8).Info("AWS DeregisterInstance not implemented")
+	if i.CloudInstanceGroup.InstanceGroup.Spec.Manager != kops.InstanceManagerKarpenter {
+		err := deregisterInstance(c, i)
+		if err != nil {
+			return fmt.Errorf("failed to deregister instance from loadBalancer before terminating: %v", err)
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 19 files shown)


upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -541,6 +541,12 @@ func (c *awsCloudImplementation) DeleteInstance(i *cloudinstances.CloudInstance)
 	return deleteInstance(c, i)
 }
 
+// DeregisterInstance drains a cloud instance and loadbalancers.
+func (c *awsCloudImplementation) DeregisterInstance(i *cloudinstances.CloudInstance) error {
```
olemarkus, line 545: AWS actually has a `deregisterInstance` that does the same time. It is not exposed thought this interface though, but is called from `deleteInstance`. Think we should consolidate here.

zetaab: if I delete `deregisterInstance` from `deleteInstance` function, it means that in case of warmpool https://github.com/kubernetes/kops/blob/e0d22ce2f7700ed9be7af619ff5f029bc9d103dd/pkg/instancegroups/instancegroups.go#L133 we are not detaching lbs. Second option is that I leave `deregisterInstance` to `deleteInstance` function. In that case other than warmpool instances will get deregisterInstance twice. Which one we should use?

zetaab: actually third option is that we do not implement new `DeregisterInstance` to AWS at all, we leave it as-is which will work fine for AWS

olemarkus: Can the DeregisterInstance call also be called from `DeleteSingleInstance`?
I am not sure why warmpool matters here. Those instances have not joined the cluster or really the ASGs and should be deregistered at all. Or am I missing something?

zetaab: ah right, I did not know what means warmpool

zetaab:
@olemarkus afaik if warmpool instances are not part of the cluster then this current code could work fine? 

> Can the DeregisterInstance call also be called from DeleteSingleInstance?
In that case we make double calls to the apis, at least I do not see point of doing that?

olemarkus (approved):
can probably clean this up later if we need to.

PR #12987, opened 2021-12-17
gce: clean up networking objects by reference

justinsb:
We try to avoid cleaning up by name, and prefer checking references to
(e.g. targeting) a known resource, like an instancegroup.

### DIFF
--- a/pkg/resources/gce/gce.go
+++ b/pkg/resources/gce/gce.go
@@ -498,89 +498,153 @@ func (d *clusterDiscoveryGCE) listFirewallRules() ([]*resources.Resource, error)
 
 	ctx := context.Background()
 
-	frs, err := c.Compute().Firewalls().List(ctx, c.Project())
+	firewallRules, err := c.Compute().Firewalls().List(ctx, c.Project())
 	if err != nil {
 		return nil, fmt.Errorf("error listing FirewallRules: %v", err)
 	}
 
-	for _, fr := range frs {
-		if !d.matchesClusterNameMultipart(fr.Name, maxPrefixTokens) && !strings.HasPrefix(fr.Name, "k8s-") {
+nextFirewallRule:
+	for _, firewallRule := range firewallRules {
+		if !d.matchesClusterNameMultipart(firewallRule.Name, maxPrefixTokens) && !strings.HasPrefix(firewallRule.Name, "k8s-") {
 			continue
 		}
 
-		foundMatchingTarget := false
+		// TODO: Check network?  (or other fields?)  No label support currently.
+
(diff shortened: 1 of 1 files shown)


justinsb:
/assign @zetaab 

Tightening up the checks here to make sure it's "our" cluster.  Without this, I get errors like the below:

```
Error: error get HTTPHealthCheck: googleapi: Error 400: Invalid value for field 'httpHealthCheck': '2d9bcd4e53ff36ee-node'. Must be a match of regex '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?|[1-9][0-9]{0,19}', invalid
```

pkg/resources/gce/gce.go
```diff
@@ -498,89 +498,153 @@ func (d *clusterDiscoveryGCE) listFirewallRules() ([]*resources.Resource, error)
...
+			if forwardingRule.Target != "" {
+				forwardingRuleResource.Blocks = append(forwardingRuleResource.Blocks, typeTargetPool+":"+gce.LastComponent(forwardingRule.Target))
 			}
-			resourceTrackers = append(resourceTrackers, frResourceTracker)
+			k8sResources = append(k8sResources, forwardingRuleResource)
 
-			tp, err := c.Compute().TargetPools().Get(c.Project(), c.Region(), name)
+			// TODO: Can we get k8s to set labels on the ForwardingRule?
+
+			// TODO: Check description?  It looks like e.g. description: '{"kubernetes.io/service-name":"kube-system/guestbook"}'
```
zetaab, line 580: yep but what if we have multiple clusters in same project? GCP cloudprovider should have something similar like CSI has "custom labels" https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver#createvolume-parameters

However, these things are not going to happen before we have 1) gce external ccm in use 2) after that custom labels should be implemented

pkg/resources/gce/gce.go
```diff
@@ -498,89 +498,153 @@ func (d *clusterDiscoveryGCE) listFirewallRules() ([]*resources.Resource, error)
...
+				ID:      targetPool.Name,
 				Type:    typeTargetPool,
 				Deleter: deleteTargetPool,
-				Obj:     tp,
+				Obj:     targetPool,
 			}
-			resourceTrackers = append(resourceTrackers, tpResourceTracker)
+			k8sResources = append(k8sResources, targetPoolResource)
+
+			// TODO: Check description? (looks like description: '{"kubernetes.io/service-name":"k8s-dbb09d49d9780e7e-node"}' )
```
zetaab, line 602: multi cluster problem as well

zetaab (approved):

PR #12992, opened 2021-12-17
Use fi.Keyset instead of passing tasks around

justinsb:
Using a task leads to layering complexity.  We could introduce a new type, but fi.Keyset is the type we seem to want.
    
(We could move Keyset out of fi, but we don't need to yet)

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -67,7 +67,6 @@ import (
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstack"
 	"k8s.io/kops/upup/pkg/fi/cloudup/terraform"
 	"k8s.io/kops/upup/pkg/fi/cloudup/terraformWriter"
-	"k8s.io/kops/upup/pkg/fi/fitasks"
 	"k8s.io/kops/util/pkg/architectures"
 	"k8s.io/kops/util/pkg/hashing"
 	"k8s.io/kops/util/pkg/mirrors"
@@ -1237,7 +1236,7 @@ func newNodeUpConfigBuilder(cluster *kops.Cluster, assetBuilder *assets.AssetBui
 }
 
 // BuildConfig returns the NodeUp config and auxiliary config.
-func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAdditionalIPs []string, caTasks map[string]*fitasks.Keypair) (*nodeup.Config, *nodeup.BootConfig, error) {
+func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAdditionalIPs []string, keysets map[string]*fi.Keyset) (*nodeup.Config, *nodeup.BootConfig, error) {
 	cluster := n.cluster
 
 	if ig == nil {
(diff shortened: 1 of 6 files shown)


justinsb:
/assign @johngmyers 

I wasn't sure whether we were trying to move away from fi.Keyset, but I don't think we should be passing tasks around.

olemarkus (commented):
/lgtm

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1236,7 +1235,7 @@ func newNodeUpConfigBuilder(cluster *kops.Cluster, assetBuilder *assets.AssetBui
 }
 
 // BuildConfig returns the NodeUp config and auxiliary config.
-func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAdditionalIPs []string, caTasks map[string]*fitasks.Keypair) (*nodeup.Config, *nodeup.BootConfig, error) {
+func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAdditionalIPs []string, keys map[string]*fi.Keyset) (*nodeup.Config, *nodeup.BootConfig, error) {
```
johngmyers, line 1238:
```suggestion
func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAdditionalIPs []string, keysets map[string]*fi.Keyset) (*nodeup.Config, *nodeup.BootConfig, error) {
```

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1399,15 +1398,21 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 	return config, bootConfig, nil
 }
 
-func getTasksCertificate(caTasks map[string]*fitasks.Keypair, name string, config *nodeup.Config, includeKeypairID bool) error {
-	cas, err := fi.ResourceAsString(caTasks[name].Certificates())
+func loadCertificate(keysets map[string]*fi.Keyset, name string, config *nodeup.Config, includeKeypairID bool) error {
```
johngmyers, line 1401:
```suggestion
func loadCertificates(keysets map[string]*fi.Keyset, name string, config *nodeup.Config, includeKeypairID bool) error {
```

Since there can be many, especially during key rotation.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1399,15 +1398,21 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+	keyset := keysets[name]
+	if keyset == nil {
+		return fmt.Errorf("key %q not found", name)
+	}
+	certificates, err := keyset.ToCertificateBytes()
 	if err != nil {
-		// CA task may not have run yet; we'll retry
-		return fmt.Errorf("failed to read %s certificates: %w", name, err)
+		return fmt.Errorf("error getting certificate from keyset %q: %w", name, err)
```
johngmyers, line 1408: My understanding is that the word "error" is usually used before printing an `error`'s string, so I try to avoid using the word in an error's string.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1399,15 +1398,21 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
 	}
-	config.CAs[name] = cas
+	config.CAs[name] = string(certificates)
 	if includeKeypairID {
-		config.KeypairIDs[name] = caTasks[name].Keyset().Primary.Id
+		if keyset.Primary == nil {
+			return fmt.Errorf("key %q did not have primary set", name)
```
johngmyers, line 1413: `Keyset` maintains the invariant that there must be a `Primary` if there are any `Keypair`s.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1263,63 +1262,63 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
-				config.KeypairIDs["service-account"] = caTasks["service-account"].Keyset().Primary.Id
+				config.KeypairIDs["service-account"] = keys["service-account"].Primary.Id
 			}
 
 			config.APIServerConfig.EncryptionConfigSecretHash = n.encryptionConfigSecretHash
-			var err error
-			config.APIServerConfig.ServiceAccountPublicKeys, err = caTasks["service-account"].Keyset().ToPublicKeys()
+			serviceAccountPublicKeys, err := keys["service-account"].ToPublicKeys()
 			if err != nil {
-				return nil, nil, fmt.Errorf("encoding service-account keys: %w", err)
```
johngmyers, line 1317: Why remove this context information?

upup/pkg/fi/cloudup/apply_cluster.go
justinsb: Ah, I like that rule.  This wasn't a deliberate rephrase, I think I came over a few iterations, so I'll just revert this

upup/pkg/fi/cloudup/apply_cluster.go
justinsb: Ah - good to know.  I'll revert this and add a comment to Primary (if one is needed!)

justinsb: Actually, as this method returns an error anyway, I think we should keep it - as it's not obvious why the keys in keysets have any keypairs.

upup/pkg/fi/cloudup/apply_cluster.go
justinsb: Good point - reverted!

upup/pkg/fi/cloudup/apply_cluster.go
johngmyers: Keeping it is okay. The keys in keysets have at least one keypair because they come from a `fitasks.Keypair`.

johngmyers (approved):

upup/pkg/fi/cloudup/apply_cluster.go
johngmyers: Another invariant maintained is that the `Primary` has a non-nil `PrivateKey`.

Issue #9624, opened 2020-07-24
Delete all dependencies when VPC is "owned"


hakman:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

There are management tools that are used to manage access or resources at organisation level.
In some cases, these tools add extra security groups to the VPC and makes it impossible for `kops delete cluster` to finish the job. It is the case of AWS Firewall Manager, as described in https://github.com/kubernetes/kops/issues/9578#issuecomment-661264659.

**2. Feel free to provide a design supporting your feature request.**

Seems reasonable to assume that all security groups should be deleted in case the VPC is not shared (has the "owned" tag).
This should be done for all VPC dependencies that prevent its deletion.

rifelpet:
the amazonvpc CNI will occasionally run into this issue as well:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-aws-cni-amazon-vpc/1288078416746123264

my guess is that there are orphaned ENIs using a security group, causing the security group deletion to fail. We should extend the logic proposed in this issue to also apply to extra ENIs in the VPC when the VPC is being deleted.

PR #12994, opened 2021-12-18
Add support for graceful node shutdown

### LINKED ISSUE #11292: Feature: Support Graceful Node Shutdown

Add support for Graceful Node Shutdown.

Add `ShutdownGracePeriod` and `ShutdownGracePeriodCriticalPods` fields to `KubeletConfigSpec`.

As Kubernetes appears to not have added kubelet flags for these settings, this will mean we need to refactor kubelet configuration to use a `KubeletConfiguration`.

olemarkus:
Fixes #11292 

```
# systemd-inhibit --list
WHO                          UID USER PID  COMM            WHAT     WHY                                                       MODE 
Unattended Upgrades Shutdown 0   root 614  unattended-upgr shutdown Stop ongoing upgrades or perform upgrades before shutdown delay
kubelet                      0   root 5366 kubelet         shutdown Kubelet needs time to handle node shutdown                delay
```

Since this feature requires componentconfig, I had to bring in that. Flags override config, so I left everything else as-is. We can gradually remove the flags tag from our component config as we copy the values over to kubelet component config.


olemarkus:
AWS CNI is failing to clean up some resources. Do not think this is related to this PR. @rifelpet I believe we have seen this before?

johngmyers:
What could the resources be? I'm wondering if an ASG scaled up a new instance which got created after the initial search for resources to delete.

rifelpet:
The `kops delete cluster` logs show the ASGs being deleted so I dont think it would be orphaned instances. With the resources that are failing to be deleted due to dependencies:

```
	security-group:sg-04067b092e2d442c8
	vpc:vpc-0b07287562bffb052
	dhcp-options:dopt-092e405d759a19eb8
	subnet:subnet-0f808ee5038c7cec0
	route-table:rtb-0555cceb23bfcecf7
```

Given that we're only seeing this on the amazon VPC CNI, My guess is that an ENI is not being deleted which is preventing the security group and subnet from being deleted. The CNI sets [DeleteOnTermination](https://github.com/aws/amazon-vpc-cni-k8s/blob/dda97f26eee12ef7477d033822765c129d2b24cc/pkg/awsutils/awsutils.go#L749) in a separate API call after attaching the ENI. I wonder if that isn't being called for one interface. The CNI logs that we capture from every node may reveal such a scenario.

https://github.com/kubernetes/kops/issues/9624 was opened to expand the resource coverage during `kops delete cluster` but we never got around to it.

pkg/apis/kops/validation/validation.go
```diff
@@ -675,7 +675,19 @@ func validateKubelet(k *kops.KubeletConfigSpec, c *kops.Cluster, kubeletPath *fi
 				allErrs = append(allErrs, field.Forbidden(kubeletPath.Child("cpuCFSQuotaPeriod"), "cpuCFSQuotaPeriod has been removed on Kubernetes >=1.20"))
 			}
 		}
-
+		if k.ShutdownGracePeriod != nil {
+			if k.ShutdownGracePeriod.Seconds() > 30 {
+				allErrs = append(allErrs, field.Invalid(kubeletPath.Child("shutdownGracePeriod"), k.ShutdownGracePeriod.String(), "shutdownGracePeriod cannot exceed 30 seconds"))
```
johngmyers, line 680: Where is the reference for this limitation?

pkg/apis/kops/v1alpha3/componentconfig.go
```diff
@@ -223,6 +223,11 @@ type KubeletConfigSpec struct {
 	EnableCadvisorJsonEndpoints *bool `json:"enableCadvisorJsonEndpoints,omitempty" flag:"enable-cadvisor-json-endpoints"`
 	// PodPidsLimit is the maximum number of pids in any pod.
 	PodPidsLimit *int64 `json:"podPidsLimit,omitempty" flag:"pod-max-pids"`
+
+	// ShutdownGracePeriod specifies the total duration that the node should delay the shutdown by.
+	ShutdownGracePeriod *metav1.Duration `json:"shutdownGracePeriod,omitempty"`
+	// ShutdownGracePeriodCriticalPods specifies the duration used to terminate critical pods during a node shutdown.
+	ShutdownGracePeriodCriticalPods *metav1.Duration `json:"shutdownGracePeriodCriticalPods,omitempty"`
```
johngmyers, line 230: We should mention the defaults that we are providing.

johngmyers:
It looks like the CNI can't prevent leaking ENIs like this. Perhaps we need `DeleteSubnet()` to search for and delete orphaned ENIs?

pkg/apis/kops/validation/validation.go
olemarkus: This is the max amount of seconds a shutdown may be inhibited. I don't have any reference for this,but setting a higher value will fail.

nodeup/pkg/model/kubelet.go
```diff
@@ -154,6 +171,39 @@ func (b *KubeletBuilder) Build(c *fi.ModelBuilderContext) error {
 	return nil
 }
 
+func buildKubeletComponentConfig(kubeletConfig *kops.KubeletConfigSpec) (*nodetasks.File, error) {
+	componentConfig := kubelet.KubeletConfiguration{
+		ShutdownGracePeriod:             *kubeletConfig.ShutdownGracePeriod,
+		ShutdownGracePeriodCriticalPods: *kubeletConfig.ShutdownGracePeriodCriticalPods,
```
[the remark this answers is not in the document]
olemarkus: If we can. For kubelet, it is not that bad. But some of the component configs are a bit complex.
We also do a lot of defaulting and other types of checks. So I wonder if we should just set KubeletConfiguration as part of that process in the future. But right now, it is not that urgent.

Many users also want more direct control over KubeletConfiguration and others. That will be a more interesting problem to solve.

pkg/apis/kops/validation/validation.go
johngmyers: Where/how does it fail? This is really unfortunate; stateful applications should be able to inhibit this for hours while they increase the replication of the node-local data they're about to lose.

olemarkus: Kubelet errorred out with a message that it was too high and refused to set the inhibitor. Not sure where the error came from kubelet or systemd. I will reproduce and have a closer look.

Unattended upgrades holds an inhibitor lock for much longer than 30s when configured to upgrade on shutdown.

olemarkus:
```
Dec 22 07:46:49 ip-172-20-59-237 kubelet[5072]: E1222 07:46:49.647969    5072 kubelet.go:1443] "Failed to start node shutdown manager" err="node shutdown manager was unable to update logind InhibitDelayMaxSec to 1m0s (ShutdownGracePeriod), current value of InhibitDelayMaxSec (30s) is less than requested ShutdownGracePeriod"
```

So it is a logind config. I saw this one before, but didn't think of it much because the default is 5s, not 30s.
But turns out ubuntu has this:
```
# cat /usr/lib/systemd/logind.conf.d/unattended-upgrades-logind-maxdelay.conf 
[Login]
# delay
InhibitDelayMaxSec=30
```

We could drop in a file as well. But I suggest doing that in a followup.

olemarkus: For reference regarding filenames:
>  Files in the *.conf.d/ configuration subdirectories are sorted by their filename in lexicographic order, regardless of in which of the subdirectories they reside. When multiple files specify the same option, for options which accept just a single value, the entry in the file sorted last takes precedence (..)

So yay to unattended-upgrades for starting their file name at "u"

hakman (changes requested):

hakman:
Maybe we should relax the conditions for running the `pull-kops-e2e-cni-amazonvpc` test. I would remove anything `nodeup` related.
https://github.com/kubernetes/test-infra/blob/e9f391ba1562ceef64356d7d8c79e3290701843a/config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml#L10

nodeup/pkg/model/tests/kubelet/featuregates/tasks.yaml
```diff
@@ -3,11 +3,48 @@ path: /etc/kubernetes/manifests
...
+    x509: {}
+  authorization:
+    webhook:
+      cacheAuthorizedTTL: 0s
+      cacheUnauthorizedTTL: 0s
+  cpuManagerReconcilePeriod: 0s
+  evictionPressureTransitionPeriod: 0s
+  fileCheckFrequency: 0s
+  httpCheckFrequency: 0s
+  imageMinimumGCAge: 0s
```
hakman, line 28: Is this expected? Looks like a lot of unrelated settings with 0 value.

nodeup/pkg/model/kubelet.go
```diff
@@ -32,6 +33,10 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
+	kubelet "k8s.io/kubelet/config/v1beta1"
```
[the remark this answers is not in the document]
olemarkus: What we have to avoid is using the types from componentconfigs into our own API structs. But here we are just using the structs for generating the yaml file. The alternative would be to copy the structs into our tree, which I do not think is necessary unless upstream decides to remove `v1beta1`. But right now, they even have `v1alpha1` in their tree.

nodeup/pkg/model/tests/kubelet/featuregates/tasks.yaml
olemarkus: Yeah. Non-pointer Durations will serialise as 0s, which is not counted as empty value. When it is unserialised, these values are treated as empty and the default is used though.

hakman: Not ideal... We also have `infoBufferSize: "0"` below.

olemarkus: These do not have omitempty.
E.g https://github.com/kubernetes/component-base/blob/master/config/v1alpha1/types.go#L95

go.mod
```diff
@@ -89,6 +89,7 @@ require (
 	k8s.io/gengo v0.0.0-20210813121822-485abfe95c7c
 	k8s.io/klog/v2 v2.30.0
 	k8s.io/kubectl v0.23.0
+	k8s.io/kubelet v0.0.0-00010101000000-000000000000
```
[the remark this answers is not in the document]
hakman: I think this has to be changed again.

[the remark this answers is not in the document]
olemarkus: how do I change this? make gomod just adds it back in no matter what I do.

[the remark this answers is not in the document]
hakman: no idea, for me works without any problem.

nodeup/pkg/model/kubelet.go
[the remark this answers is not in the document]
hakman: 👍

go.mod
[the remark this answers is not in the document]
hakman: https://github.com/hakman/kops/commit/87c941ff83b591245ace01a2330867e1d9b4e42b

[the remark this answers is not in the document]
olemarkus: strange. works now.

pkg/apis/kops/validation/validation.go
hakman: Does Debian 11 have the same limitation? If not, we should remove this check and just document it.

olemarkus: I would assume unattened-upgrades install the same limit increase. If it doesn't, then the limit is 5s and the inhibitor lock won't be installed.

hakman: I still think such a validation is not very reliable for multiple OSes and documenting the limitation until we override would be preferable. Operators could drop an override by other means until then.

olemarkus: Sounds reasonable.

hakman (changes requested):

docs/cluster_spec.md
```diff
@@ -742,6 +742,23 @@ spec:
...
+
+By default, kOps will try to add a 30 second timeout for 30 seconds where the first 2o seconds is reserved for normal Pods and the last 10 seconds for critical Pods.
+
+This configuration can be changed as follows:
+
+```yaml
+spec:
+  kubelet:
+    shutdownGracePeriod: 60s
+    shutdownGracePeriodCriticalPods: 20s
```
hakman, line 757: To match defaults:
```suggestion
    shutdownGracePeriod: 30s
    shutdownGracePeriodCriticalPods: 10s
```

docs/cluster_spec.md
```diff
@@ -742,6 +742,23 @@ spec:
...
+```
+
+Note that Kubelet will fail to install the shutdown inhibtor on systems where logind is configured with an `InhibitDelayMaxSeconds` lower than `shutdownGracePeriod`. On Ubuntu, this setting is 30 seconds.
```
hakman, line 760:
```suggestion
Note that Kubelet will fail to install the shutdown inhibtor on systems where `systemd-logind` is configured with an `InhibitDelayMaxSeconds` lower than `shutdownGracePeriod`. On Ubuntu, this setting is 30 seconds.
```

hakman (changes requested):

olemarkus:
What I think happens here is:
* Cluster starts deletion, creates a list of resources to be deleted and starts going through the list.
* triggers termination of all instances
* aws-node detects rescheduling etc and starts spawning ENIs. This could also be because it detects its own ENIs being removed as part of termination.
* The new ENIs are not on the list kops creates and therefore becomes dangling. aws-node also does not manage to attach or set terminateOnDelete flag on the ENI.
* kops cannot delete SG/subnet because of the dangling ENI.

It could also be we are hitting https://github.com/aws/amazon-vpc-cni-k8s/issues/1223

A couple of ways to handle this:
* Delete all IAM roles first. aws-node would then not be able to do any mischief
* Make aws-node clean up its own ENIs on termination if also the node is terminating (issue above, I think)
* Build the list of resources that needs deletion say once per 10 deletion cycles (we do 42 of them).

For this PR, I think I will just not set any default for amazonvpc. Then we can see where to fix this properly. I haven't tested yet, but I suspect it will leak quite a lot of ENIs on plain cluster rolls too (but these will be cleaned up by remaining aws-node instances)

docs/cluster_spec.md
```diff
@@ -742,6 +742,23 @@ spec:
     logFormat: json
 ```
 
+### Graceful Node Shutdown
+
+Graceful node shutdown allows kubelet to prevent instance shutdown until Pods have been safely terminated or a timeout has been reached.
+
+For all CNIs except `amazonaws`, kOpw will try to add a 30 second timeout for 30 seconds where the first 2o seconds is reserved for normal Pods and the last 10 seconds for critical Pods. When using `amazonaws` this feature is disabled, as it leads to [leaking ENIs](https://github.com/aws/amazon-vpc-cni-k8s/issues/1223).
```
hakman, line 749:
```suggestion
For all CNIs except `amazonaws`, kOps will try to add a 30 second timeout for 30 seconds where the first 20 seconds is reserved for normal Pods and the last 10 seconds for critical Pods. When using `amazonaws` this feature is disabled, as it leads to [leaking ENIs](https://github.com/aws/amazon-vpc-cni-k8s/issues/1223).
```

docs/cluster_spec.md
olemarkus: The point was to show how to change the defaults. so feel like this can/should stay as-is.

hakman:
@justinsb Any objection to merging this as is in 1.23 and maybe change the approach later to reflection?

hakman:
@olemarkus Could you split the `hack/update-expected.sh` part into a separate commit?
/lgtn

hakman (approved):

PR #12998, opened 2021-12-19
componentconfig: expose advertise-address flag for kube-apiserver

justinsb:
This flag determines which IPs are advertised inside the cluster, in
the "kubernetes" service in the "default" namespace.


olemarkus:
How can this flag be used? If it is set by a user, wouldn't all servers get the same ip?

hakman:
@olemarkus Yes, this is correct. Its purpose is for "metal" setup testing and should be set to the LB IP.
I don't think it hurts in any way to add this field.
/lgtm

olemarkus (approved):
Ah. Sure if it is useful. It's not the only setting we have that doesn't make sense for most users.

PR #13007, opened 2021-12-20
Use kubelet --non-masquerade-cidr only for Docker with kubenet

hakman:
Kubelet `--non-masquerade-cidr` is/was only used in the dockershim kubenet implementation. The flag is also deprecated since Kubernetes 1.7.
https://github.com/kubernetes/kubernetes/blob/8f91d09865b7adb641a5e890d8b9b6053078b81a/pkg/kubelet/dockershim/network/kubenet/kubenet_linux.go#L187-L205

/cc @olemarkus @rifelpet


olemarkus (commented):
/lgtm

rifelpet (approved):

PR #13010, opened 2021-12-20
Do not install ClusterRole and binding used by in-tree volume provider if CSI is used

### LINKED ISSUE #11283: Understand (and rename?) RBAC and service account for AWS storage addon
Currently `storage-aws.addons.k8s.io` includes [a ClusterRole and ClusterRoleBinding](https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml.template#L46-L85) with the name `system:aws-cloud-provider`.  These were added to the storage addon in https://github.com/kubernetes/kops/pull/6899, because it was removed from upstream in https://github.com/kubernetes/kubernetes/pull/66635.  The new external cloud controller manager addon for aws (`aws-cloud-controller.addons.k8s.io`) uses the name: `system:cloud-controller-manager` for its service account and RBAC, but one difference is that the service account `aws-cloud-controller.addons.k8s.io` is managed in the addon yaml, whereas the service account for `storage-aws.addons.k8s.io` is created elsewhere (I haven't found it yet).  

To better understand the use of the `aws-cloud-provider` service account referenced in the RBAC, I'd like to answer the following questions:
1. When and how is the aws-cloud-provider service account created?
   According to a comment in the PR, perhaps by the client builder code.
2. What is it used for?  
   According to the comment by @gnufied in the linked PR, its used for "tainting the nodes when volumes are stuck in Attaching state for too long."
3. What cluster configurations require this RBAC and service account?
   Here I'm assuming its clusters with kube-controller-manager, with `--cloud-provider=aws` (no external cloud provider).

Perhaps we should do one or more of the following: 
1. rename the `aws-cloud-provider` service account and RBAC to something more specific, 
2. separate it from the storage classes, (which are probably shared with CSI and still useful) and 
3. deprecate it (on the same timeline as the in-tree provider).

olemarkus:
Fixes #11283

### DIFF
--- a/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml.template
+++ b/upup/models/cloudup/resources/addons/storage-aws.addons.k8s.io/v1.15.0.yaml.template
@@ -55,7 +55,7 @@ provisioner: ebs.csi.aws.com
 allowVolumeExpansion: true
 volumeBindingMode: WaitForFirstConsumer
 {{ end }}
-{{ end }}
+
 ---
 
 apiVersion: rbac.authorization.k8s.io/v1
@@ -98,3 +98,5 @@ subjects:
 - kind: ServiceAccount
   name: aws-cloud-provider
   namespace: kube-system
+
+{{ end }}


johngmyers:
This PR removes the ClusterRole and doesn't affect the ServiceAccount. Could you fix the commit message and PR title?

johngmyers (approved):

PR #13012, opened 2021-12-20
nodeup bash script: use explicit return code

justinsb:
In bash, `return` returns the exit code of the last statement.  Being
explicit here is safer, and I believe in the first case (the cached
download) the value was actually incorrect.


johngmyers:
Is this something shellcheck can check for?

johngmyers (approved):

PR #13013, opened 2021-12-20
staticcheck cleanup: fixup nodeup/pkg/model

justinsb:
These pop up in VSCode and are pretty simple to fix:

```
nodeup/pkg/model/cloudconfig_test.go:86:17: possible nil pointer dereference (SA5011)
        nodeup/pkg/model/cloudconfig_test.go:83:5: this check suggests that the pointer can be nil
nodeup/pkg/model/cloudconfig_test.go:155:17: possible nil pointer dereference (SA5011)
        nodeup/pkg/model/cloudconfig_test.go:152:5: this check suggests that the pointer can be nil
nodeup/pkg/model/sysctls.go:172:12: error strings should not be capitalized (ST1005)
nodeup/pkg/model/sysctls.go:184:12: error strings should not be capitalized (ST1005)
nodeup/pkg/model/volumes.go:59:11: error strings should not be capitalized (ST1005)
```

### DIFF
--- a/nodeup/pkg/model/cloudconfig_test.go
+++ b/nodeup/pkg/model/cloudconfig_test.go
@@ -81,7 +81,7 @@ func TestBuildAzure(t *testing.T) {
 		}
 	}
 	if task == nil {
-		t.Errorf("no File task found")
+		t.Fatalf("no File task found")
 	}
 	r, err := task.Contents.Open()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


johngmyers:
We should consider trimming our staticcheck exclude list. Perhaps have a second exclude list just for rules we want to start enabling.

johngmyers (approved):

PR #13021, opened 2021-12-22
Fix dangling ENIs from AWS VPC CNI 

olemarkus:
This should fix future dangling ENIs from AWS VPC CNI should it occur in the future.
Cilium already tags its ENIs so should the same problem happen there, this should fix that as well.

### DIFF
--- a/pkg/resources/aws/eni.go
+++ b/pkg/resources/aws/eni.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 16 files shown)


pkg/resources/aws/eni.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		if awsup.AWSErrorCode(err) == "InvalidNetworkInterfaceID.NotFound" {
+			// Concurrently deleted
+			return nil
+		}
+
+		if IsDependencyViolation(err) {
+			return err
+		}
+		return fmt.Errorf("error deleting ENI %q: %v", id, err)
```
johngmyers, line 50: Since errors are usually printed prefixed with the word "error" I tend not to include the word "error" in the error string.

johngmyers (approved):

PR #13029, opened 2021-12-24
Remove TerraformJSON feature flag

rifelpet:
TerraformJSON was [deprecated in 1.22](https://github.com/rifelpet/kops/blob/master/docs/releases/1.22-NOTES.md#deprecations). This removes the feature flag and the json field tags from the Terraform structs.

I propose we cherrypick this to 1.23 because the removal was initially declared as 1.23 and because it would reduce the conflicts in other cherrypicks.


johngmyers:
Should update terraform.md and experimental.md

hakman (approved):

PR #13030, opened 2021-12-24
Use IPv6-only subnets for worker nodes in private IPv6 topology

### LINKED ISSUE #12886: Determine supported topologies for IPv6
Per a discussion of IPv6 and subnets, I have created a taxonomy of topologies that appear to make sense.

From here we should decide which of these we want to support with managed subnets. The more specialized or obscure ones can be declared as only supportable through using shared subnets.

# IPv4 topologies (NonMasqueradeCIDR is IPv4)

## 1. "public"

IPv4-only subnets, type "public"
Instances get public IPv4 addresses
Instances get private IPv4 addresses
Load balancers placed in subnet
Egress to Internet Gateway

Use cases:
- Test clusters
- Other low-security environments
- Don't want to pay for NAT Gateway

## 2. "private"

IPv4-only subnets, two types:

- "utility"
  Contain load balancers and NAT Gateway
  Do not contain nodes
  Egress to Internet Gateway
- "private"
  Contain nodes
  Instances get private IPv4 addresses
  Egress to NAT Gateway in utility subnet for AZ

Advantages over (1):
- No direct ingress to nodes from public internet

Disadvantages over (1):
- More complicated
- Pay for NAT Gateway

Use cases:
- Don't want nodes to be directly exposed to public internet

## 3.

Same as (2) except no utility subnets
No NAT Gateway

Advantages over (2):
- Simpler
- No need to pay for NAT Gateway

Use cases:
- No need for ingress from or egress to public internet

## 4. What EKS supports

IPv4-only subnets, types "public" and "private"
Like (2), but nodes can be created in the "public" subnet
NAT Gateway and load balancers are placed in the "public" subnet

Use cases:
- Some workloads need direct ingress from public internet
- Only some workloads need public internet egress, in which case NAT
gateway can be omitted

# IPv6 topologies (NonMasqueradeCIDR is IPv6)

## 10. IPv6 analogue to "public"

Subnets are dual-stack, type "public"
Instances get IPv6 addresses
Instances get public IPv4 addresses
Instances get private IPv4 addresses
Load balancers and NAT Gateway placed in subnet
Egress to Internet Gateway
Route NAT64 prefix to NAT Gateway

Use cases:
- Test clusters
- Other low-security environments

## 11.

Same as (10) but without NAT Gateway

Use cases:
- No need for pods to egress to IPv4

## 12.

Two types of subnets:

- "utility" subnets
  Dual-stack
  Contain load balancers and NAT Gateway
  Do not contain nodes
  Egress to Internet Gateway
- "private"? subnets
  IPv6-only
  Contain nodes
  Instances get IPv6 addresses only
  NAT64 egress to NAT Gateway in utility subnet for AZ
  Other IPv6 egress to Egress-only Internet Gateway

Advantages over (10):
- No direct ingress to nodes from public internet
- No need to assign IPv4 addresses to nodes

Advantages over (2):
- No need to assign IPv4 addresses to pods or nodes
- Allows IPv6 ingress and egress

Use cases:
- Private IPv4 space is exceedingly scarce

## 13.

Same as (12) except private subnets are dual-stack

Private subnets egress IPv4 and NAT64 to NAT Gateway in utility subnet for AZ

Advantages over (12):
- Traffic from nodes or host-network pods to private IPv4 addresses
doesn't have to go through NAT

Advantages over (2):
- Don't need IPv4 addresses for pods, which greatly outnumber nodes

Disadvantages over (12):
- Need to assign an IPv4 addresss per node

Use cases:
- Private-IPv4-only container registry, file repository, or other services accessed from nodes

## 14.

Same as (12) except no NAT Gateway

Use cases:
- No need for IPv4 egress

## 15.

Same as (12) except utility subnet is IPv6-only

No NAT Gateway

Use cases:
- No need for IPv4 ingress or egress

## 16.

Dual-stack "public" subnet and IPv6-only "private" subnet
Like (4) but IPv6
Like (12) but nodes can be created in the "public" subnet

Use cases:
- Some workloads need direct ingress from public internet

## 17.

IPv6-only "public" and "private" subnets

No NAT Gateway

Use cases:
- Some workloads need direct ingress from public internet, no need for IPv4 ingress or egress

## 18.

(12) but with an additional type of subnet:

- "private-dualstack"? subnets
  Dual-stack
  Contain nodes
  IPv4 and NAT64 egress to NAT Gateway in utility subnet for AZ
  Other IPv6 egress to Egress-only Internet Gateway

Use cases:
- NLB doesn't support instance targets with ipv6 addresses, so we cannot attach a target group to ASGs. So nodes, such as control plane nodes, needing NLBs in front go in the dual-stack subnet.

johngmyers:
One alternative would be to omit creating dualstack subnets in zones that don't have control plane nodes when there are no apiserver nodes.

/cc @olemarkus @hakman 
Closes #12886

### DIFF
--- a/docs/networking/ipv6.md
+++ b/docs/networking/ipv6.md
@@ -13,13 +13,20 @@ kOps currently supports IPv6 on AWS only.
 
 IPv6 requires the external Cloud Controller Manager.
 
-## VPC and subnets
+## VPC, subnets, and topology
 
 The VPC can be either shared or managed by kOps. If shared, it must have an IPv6 pool associated.
 
 Subnet IPv6 CIDR allocations may be specified in the cluster spec using the special syntax `/LEN#N`,
 where "LEN" is the prefix length and "N" is the hexadecimal sequence number of the CIDR within the VPC's IPv6 CIDR.
 For example, if the VPC's CIDR is `2001:db8::/56` then the syntax `/64#a` would mean `2001:db8:0:a/64`.
+
+Public and utility subnets are expected to be dual-stack. Subnets of type `Private` are expected to be IPv6-only.
+There is a new type of subnet `DualStack` which is like `Private` but is dual-stack.
+The `DualStack` subnets are used by default for the control plane and APIServer nodes.
+
+IPv6-only subnets require Kubernetes 1.23 or later. For this reason, private topology on an IPv6 cluster also
+requires Kubernetes 1.23 or later.
 
 ## Routing and NAT64
 
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -633,11 +633,17 @@ type EtcdMemberSpec struct {
 type SubnetType string
 
 const (
-	// SubnetTypePublic means the subnet is public
+	// SubnetTypePublic means the subnet has external addresses.
+	// In IPv6 clusters it is typically dual-stack.
 	SubnetTypePublic SubnetType = "Public"
-	// SubnetTypePrivate means the subnet has no public address or is natted
+	// SubnetTypePrivate means the subnet has no public addresses.
+	// In IPv6 clusters it is typically IPv6-only.
 	SubnetTypePrivate SubnetType = "Private"
-	// SubnetTypeUtility mean the subnet is used for utility services, such as the bastion
+	// SubnetTypeDualStack means the subnet has no public addresses but is dual-stack.
+	SubnetTypeDualStack SubnetType = "DualStack"
+	// SubnetTypeUtility mean the subnet has external addresses but is not used for nodes.
+	// It is used for utility services, such as the bastion or load balancers.
+	// In IPv6 clusters it is typically dual-stack.
 	SubnetTypeUtility SubnetType = "Utility"
 )
 
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -471,7 +471,7 @@ func validateSubnet(subnet *kops.ClusterSubnetSpec, fieldPath *field.Path) field
 			allErrs = append(allErrs, field.Invalid(fieldPath.Child("egress"), subnet.Egress,
 				"egress must be of type NAT Gateway, NAT Gateway with existing ElasticIP, NAT EC2 Instance, Transit Gateway, or External"))
 		}
-		if subnet.Egress != kops.EgressExternal && subnet.Type != "Private" && (subnet.IPv6CIDR == "" || subnet.Type != "Public") {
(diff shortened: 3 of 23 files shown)


johngmyers:
/test pull-kops-ee-cni-calico-ipv6

We don't appear to have any e2e tests with private topology.

johngmyers:
Looks like this depends on #13208

olemarkus:
I think this looks good.

In the IPv6 docs we should probably also mention that classic ELBs will not work since kubelet reasonably only listens to IPv6 (kube-proxy would need some very interesting rules to bounce ipv6 to ipv4 traffic around).

When LBC finally gets 1.22+ support, we could consider enabling it by default and perhaps disable the service controller in CCM.

hakman:
> When LBC finally gets 1.22+ support, we could consider enabling it by default and perhaps disable the service controller in CCM.

Disabling the service controller sounds good, but enabling LBC by default seems like a pain. Can be enabled manually if anyone needs it.

rifelpet:
We haven't attempted to setup a private topology test yet because we would lose SSH access to nodes in the private subnets, so artifact collection wouldn't work. I recall exploring both [Run Command](https://aws.amazon.com/blogs/aws/new-ec2-run-command-remote-instance-management-at-scale/) and [Session Manager](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager.html).

Run Command requires the command output be stored in an S3 bucket which I suppose could work in the prow environment but isn't ideal.

Session Manager requires a [local plugin binary](https://github.com/aws/session-manager-plugin) passed to ssh's ProxyCommand.

Its probably worth revisiting using Session Manager and the plugin in kubetest2-kops to dump artifacts.

johngmyers:
Filed kubernetes-sigs/aws-load-balancer-controller#2431 to make progress on the issue of `Services` of type `LoadBalancer` in IPv6-only subnets.

hakman:
I missed yesterday's office hours and the discussion about this. Can delay the final review and the release until next week?
As it is described here, it makes IPv6 setup pretty unintuitive with the "DualStack" name that is not really about topology.
Maybe we can signal IPv6-only subnets instead.

olemarkus:
Is the plan to bring this into 1.23? sounds a bit late?

johngmyers:
I wasn't planning on pulling this into 1.23.

The DualStack type networks are the special ones, only needed for nodes that have to be instance targets.

johngmyers:
Todo: Put the DualStack type behind the IPv6 feature flag

johngmyers:
API validation now only permits `DualStack` type subnets when the cluster `IsIPv6Only()`. This satisfies the desire to have them behind the feature flag.

olemarkus (commented):
The private vs dualstack throws me off a bit. I can see private ipv6 being about egress-only vs full igw. But that it implicitly means ipv6-only while public means dualstack? Not sure.

Anyways, this is behind a feature flag, so ...
/lgtm

PR #13031, opened 2021-12-25
Use instance requirements with Karpenter

olemarkus:
This PR uses the getInstanceTypesFromInstanceRequirements API to add appropriate instance types to the Karpenter providers. This adds the flexibility one expects from using Karpenter without having to maintain ones own lists of instance types, and also allows us to ensure we have valid combinations of instance types (e.g not combining arm and amd).

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -876,24 +876,6 @@ func setupKarpenterNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSub
 		HTTPTokens:              fi.String("required"),
 	}
 
-	// Karpenter thinks all clusters run VPC CNI and schedules thinking Node Capacity is constrainted by number of ENIs.
-
-	// cpuMin is the reasonable lower limit for a Kubernetes Node
-	// Generally, it also avoids instances Karpenter thinks it can only schedule 4 Pods on.
-	cpuMin := resource.MustParse("2")
-	memoryMin := resource.MustParse(("2G"))
-
-	g.Spec.MixedInstancesPolicy = &api.MixedInstancesPolicySpec{
-		InstanceRequirements: &api.InstanceRequirementsSpec{
-			CPU: &api.MinMaxSpec{
-				Min: &cpuMin,
-			},
-			Memory: &api.MinMaxSpec{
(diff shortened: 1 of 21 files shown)


olemarkus:
/cc @johngmyers
Mind having a look at this one?

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -876,6 +876,24 @@ func setupKarpenterNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSub
...
 	}
 
+	// Karpenter thinks all clusters run VPC CNI and schedules thinking Node Capacity is constrainted by number of ENIs.
+
+	// cpuMin is the reasonable lower limit for a Kubernetes Node
+	// Generally, it also avoids instances Karpenter thinks it can only schedule 4 Pods on.
+	cpuMin := resource.MustParse("2")
+	memoryMin := resource.MustParse(("2G"))
+
+	g.Spec.MixedInstancesPolicy = &api.MixedInstancesPolicySpec{
```
johngmyers, line 886: Do we really want to bake this into the IG spec during cluster creation? Could we instead treat it as a default?

olemarkus: I think that is somewhat tricky. We wouldn't know the difference of just having machineType set and having machineType and instanceRequirements. I think we still need machineType to validate things like gpu-only and single arch.

johngmyers: This PR appears to remove support for specifying both machineType and mixedInstancesPolicy. If the latter is set, karpenterInstanceTypes ignores the value of the former.

It seems karpenterInstanceTypes could implement the defaulting when both mixedInstancesPolicy and machineType are empty. They could also default the minimum cpu and memory fields.

olemarkus: > This PR appears to remove support for specifying both machineType and mixedInstancesPolicy. If the latter is set, karpenterInstanceTypes ignores the value of the former.

That both were used was just a laziness on my side. ASGs also ignore machineType if mixedInstancesPolicy is set. So now the behavior is consistent.

I moved the defaulting to templates now. And removed/amended validation that required machineType to be set.

johngmyers (approved):

Issue #12980, opened 2021-12-16
Unable to create new kOps cluster in OpenStack


zetaab:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

master

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

1.22.3

**3. What cloud provider are you using?**

openstack

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

```
kops create cluster \
  --cloud openstack \
  --name draintest.k8s.local \
  --state ${KOPS_STATE_STORE} \
  --zones x,y,z \
  --network-cidr 10.2.0.0/16 \
  --image debian-11-221121-devops \
  --master-count=3 \
  --node-count=2 \
  --node-size m1.medium \
  --master-size m1.medium \
  --etcd-storage-type solidfire \
  --topology private \
  --bastion \
  --networking calico \
  --api-loadbalancer-type public \
  --os-octavia=true \
  --os-ext-net x-nap \
  --os-ext-subnet ext-ha-v4 \
  --os-lb-floating-subnet ext-ha-v4

kops update cluster --name draintest.k8s.local --yes --admin
```

**5. What happened after the commands executed?**

I can see following in logs

```
W1216 11:12:59.589334   56693 executor.go:139] error running task "SecurityGroupRule/IPv4-ingress-tcp-from-bastion.draintest.k8s.local-to-::/0-22-22" (9m42s remaining to succeed): error creating SecurityGroupRule in SG bastion.draintest.k8s.local: error creating security group rule {ingress  IPv4 4c7c0afd-7131-483f-bb9e-d5b505dd7db2 22 22 tcp  ::/0 }: Bad request with: [POST https://sdc.elisa.fi:13696/v2.0/security-group-rules], error message: {"NeutronError": {"type": "SecurityGroupRuleParameterConflict", "message": "Conflicting value ethertype IPv4 for CIDR ::/0", "detail": ""}}
I1216 11:12:59.589382   56693 executor.go:111] Tasks: 111 done / 150 total; 14 can run
[...]

PR #13032, opened 2021-12-25
Fix OpenStack SecurityGroupRule/LB When CIDR is IPv6

iGene:
When creating security group rule, the default EtherType is IPv4.
Currently, AdminAccess default includes IPv6 CIDR, in this case
EtherType should be changed to IPv6.

Similar issue also happens with load balancer creation as the VIP is
IPv4.

This commit fixed this issue by checking if the CIDR is IPv6 before
creating the rule, if it is, EtherType will be changed to IPv6. When
creating load balancer, if allowed CIDR doesn't match VIP type, remove
it from the list.

Fixes: #12980

### DIFF
--- a/upup/pkg/fi/cloudup/openstacktasks/BUILD.bazel
+++ b/upup/pkg/fi/cloudup/openstacktasks/BUILD.bazel
@@ -61,6 +61,7 @@ go_library(
         "//vendor/github.com/gophercloud/gophercloud/openstack/networking/v2/subnets:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/util/wait:go_default_library",
         "//vendor/k8s.io/klog/v2:go_default_library",
+        "//vendor/k8s.io/utils/net:go_default_library",
     ],
 )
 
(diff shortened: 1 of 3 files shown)


iGene:
Another approach will be to add the check at when creating every `SecurityGroupRule` when `RemoteIPPrefix` is set.

For example at these lines:
https://github.com/kubernetes/kops/blob/master/pkg/model/openstackmodel/firewall.go#L101
https://github.com/kubernetes/kops/blob/master/pkg/model/openstackmodel/firewall.go#L117
https://github.com/kubernetes/kops/blob/master/pkg/model/openstackmodel/firewall.go#L243
https://github.com/kubernetes/kops/blob/master/pkg/model/openstackmodel/firewall.go#L273

zetaab (approved):
looks good, thank you!

zetaab:
@iGene if you want cherry pick this to release 1.23 feel free to do it. Instructions https://kops.sigs.k8s.io/contributing/proposing-a-cherry-pick/

iGene:
@zetaab Thanks for the instructions, cherry-picked to 1.23 release.

PR #13033, opened 2021-12-25
LBC has to run on the control plane, so set replicas accordingly

olemarkus:


johngmyers:
Any particular reason we shouldn't deploy on worker nodes if using IRSA?

(And I noticed that `deployOnWorkersIfExternalPermissons` misspells "permissions".)

olemarkus:
The problem is that LBC without IMDS need `--vpc-id` specified. And it is really tricky to get the actual VPC ID into the templates. I have asked LBC if it is possible to use tag-based discovery of VPC as well, like they do for subnets.

PR #13038, opened 2021-12-27
Kube components log to stdout

olemarkus:
When we moved to go-runner, we stopped getting logs on stdout. This adds stdout logging back in again so we can do `kubectl logs`


rifelpet (approved):
confirmed this works locally with:

```
docker run -it --rm k8s.gcr.io/kube-scheduler:v1.23.0 --log-file=/var/log/kube-scheduler.log /usr/local/bin/kube-scheduler

docker run -it --rm k8s.gcr.io/kube-scheduler:v1.23.0 --log-file=/var/log/kube-scheduler.log --also-stdout /usr/local/bin/kube-scheduler
```

PR #13017, opened 2021-12-21
external CCM for GCE

jiahuif:
This PR adds `ExternalCloudControllerManager` support for gcp.

PR #13044, opened 2021-12-28
Automated cherry pick of #13017: external CCM for GCE

johngmyers:
Cherry pick of #13017 on release-1.23.

#13017: external CCM for GCE

### DIFF
--- a/pkg/model/components/gcpcloudcontrollermanager.go
+++ b/pkg/model/components/gcpcloudcontrollermanager.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package components
+
+import (
+	"k8s.io/kops/pkg/apis/kops"
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/cloudup/gce"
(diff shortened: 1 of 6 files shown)


pkg/model/components/gcpcloudcontrollermanager.go
```diff
@@ -0,0 +1,55 @@
...
+	// CCM interacts directly with the GCP API, use the name safe for GCP
+	ccmConfig.ClusterName = gce.SafeClusterName(b.ClusterName)
+	ccmConfig.AllocateNodeCIDRs = fi.Bool(true)
+	ccmConfig.CIDRAllocatorType = fi.String("CloudAllocator")
+	if ccmConfig.ClusterCIDR == "" {
+		ccmConfig.ClusterCIDR = clusterSpec.PodCIDR
+	}
+	if ccmConfig.Image == "" {
+		// TODO: Implement CCM image publishing
+		ccmConfig.Image = "k8scloudprovidergcp/cloud-controller-manager:v1.23.0"
```
zetaab, line 52: imo in future we should use something similar than openstack has. Which means that if we have for instance kubernetes 1.23.x it uses 1.23.0 image (which can be overrided of course with Image spec)

zetaab (approved):

pkg/model/components/gcpcloudcontrollermanager.go
zetaab: @jiahuif is CCM image publishing something that you could help? Like make CI for gcp ccm?

johngmyers: The OpenStack method would make it more challenging to run e2e tests on the latest Kubernetes, before the corresponding CCM has been tagged.

zetaab: @johngmyers correct, but can we for instance use 1.26 ccm image for 1.23 kubernetes? that is going to be the challenge. However, kops do have some tools to use different images also between versions

johngmyers: Overriding the image in tests would be more trouble than the current approach of tracking it in the CCM's OptionsBuilder. The tests tracking latest Kubernetes would have to be updated twice: once when the Kubernetes minor version bumps and again when the corresponding CCM is tagged.

PR #13049, opened 2021-12-29
Preload channel versions from namespaces

olemarkus:
Channels is a bit spammy towards the k8s API. This will reduce some of the spam by not loading all namespaces multiple times per installed addon.

### DIFF
--- a/channels/pkg/channels/addon.go
+++ b/channels/pkg/channels/addon.go
@@ -90,35 +90,34 @@ func (a *Addon) ChannelVersion() *ChannelVersion {
 }
 
 func (a *Addon) buildChannel() *Channel {
-	namespace := "kube-system"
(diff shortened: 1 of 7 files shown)


johngmyers (changes requested):
Previously, channels would only look at the version annotation in the namespace listed in the channel for the addon. This code takes the version from the last-listed namespace that has an annotation for this channel.

The code should ignore annotations in namespaces other than the one listed in the channel for the addon.

olemarkus:
The PR contains a bit of refactoring now in order to run a test case proving that we are looking only at the correct namespace.
Sorry for cramming a bit too much into the same commit. If you really want to, I can probably do the refactoring bit on the master version, and apply this change after.

johngmyers (approved):

PR #13050, opened 2021-12-29
Bump Cluster Autoscaler and update manifest

olemarkus:


upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -239,39 +273,16 @@ spec:
     metadata:
       labels:
         app: cluster-autoscaler
-      annotations:
-        prometheus.io/port: "8085"
-        prometheus.io/scrape: "true"
```
rifelpet, line 244: Is there any reason not to keep these?

olemarkus: Nope. I missed that in the diff. Thanks for catching

rifelpet (approved):

PR #13052, opened 2021-12-29
Tag IAM Roles with service account info

rifelpet:


johngmyers:
Is there a reference defining these tag names?

rifelpet:
There is no reference, I made them up and am open to changing them. We could use something under `kops.k8s.io`.

For reference, eksctl uses `alpha.eksctl.io/iamserviceaccount-name = <namespace>/<name>`.

https://github.com/weaveworks/eksctl/blob/debd9f9cef49adae2e44384101a6bd8b697af852/pkg/apis/eksctl.io/v1alpha5/types.go#L232-L233

 I can't find any other occurrences of service account metadata being tagged on cloud resources.

pkg/model/context.go
```diff
@@ -185,6 +186,13 @@ func (b *KopsModelContext) CloudTagsForInstanceGroup(ig *kops.InstanceGroup) (ma
 	return labels, nil
 }
 
+func (b *KopsModelContext) CloudTagsForServiceAccount(name string, sa types.NamespacedName) map[string]string {
+	tags := b.CloudTags(name, false)
+	tags[awstasks.CloudTagServiceAccountName] = sa.Name
+	tags[awstasks.CloudTagServiceAccountNamespace] = sa.Name
```
johngmyers, line 192:
```suggestion
	tags[awstasks.CloudTagServiceAccountNamespace] = sa.Namespace
```

johngmyers:
What is the reason for adding these tags?

olemarkus:
Since we truncate names, it would be convenient to have these tags in some occasions. Would perhaps also be useful for auditing purposes.

Suggest putting it under the kops namespace though.

johngmyers:
Fortunately we put the most important information first and the least important where it will be truncated. So unless the SA or namespace is super-long, the information will be in the name.

pkg/model/awsmodel/iam.go
```diff
@@ -202,15 +202,19 @@ func (b *IAMModelBuilder) buildIAMRole(role iam.Subject, iamName string, c *fi.M
...
-		Tags:               b.CloudTags(iamName, false),
 	}
 
 	if isServiceAccount {
 		// e.g. kube-system-dns-controller
 		iamRole.ExportWithID = fi.String(roleKey)
+		sa, ok := role.ServiceAccount()
+		if ok {
+			iamRole.Tags = b.CloudTagsForServiceAccount(iamName, sa)
+		}
```
justinsb, line 213: In the !ok case, should we set it to CloudTags?  Although it looks like it should be unreachable, unless roleKey changes its behaviour.

olemarkus:
@rifelpet I think this is worth revisiting. I have a use case where I need to programmatically fetch the role for a given SA, and since we truncate the name, I don't have anything to go by. Would you be able to get this one mergeable again?

/reopen

rifelpet:
IIRC there was a concern about the iso AWS partition not supporting role tags and there not being a way to opt out of them with kops. I'm not sure if that concern is still valid.

olemarkus:
We already rely on tags for deleting roles, so think that this is no longer valid.

PR #13059, opened 2021-12-30
Make cloudProvider a struct in v1alpha3 API

johngmyers:


olemarkus:
I favour having the flag for setting cloud provider, and then a struct for the details. We do this with e.g container runtime.
We also see this frequently in other API objects. e.g
```
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
```

I see having structs per cloudProvider is a must though. It is too much awkwardness right now with the `cloudConfig` struct, which went from being about the cloud.config file to aa lot more.

johngmyers:
We use the pattern of struct for selecting with CNIs, so it makes sense to use the pattern for cloudProvider as well.

Container runtime is a bit special since one might need to set containerd config when the runtime is Docker.

cmd/kops/update_cluster.go
```diff
@@ -488,11 +488,11 @@ func completeUpdateClusterTarget(f commandutils.Factory, options *UpdateClusterO
 			cloudup.TargetDryRun,
 		}
 		for _, cp := range cloudup.TerraformCloudProviders {
-			if cluster.Spec.CloudProvider == string(cp) {
+			if cluster.Spec.GetCloudProvider() == cp {
 				completions = append(completions, cloudup.TargetTerraform)
 			}
 		}
-		if cluster.Spec.CloudProvider == string(kops.CloudProviderAWS) {
+		if cluster.Spec.GetCloudProvider() == kops.CloudProviderAWS {
```
justinsb, line 495: Big +1 to using "getters" over raw fields (also helps with nil pointer checking and mutability)

pkg/apis/kops/v1alpha2/cluster.go
```diff
@@ -53,9 +54,11 @@ type ClusterSpec struct {
 	// This might be different that the location when the cluster spec itself is stored,
 	// both because this must be accessible to the cluster,
 	// and because it might be on a different cloud or storage system (etcd vs S3)
-	ConfigBase string `json:"configBase,omitempty"`
+	ConfigBase    string                 `json:"configBase,omitempty"`
+	CloudProvider kops.CloudProviderSpec `json:"-"`
```
justinsb, line 58: I think we will need to give it a json representation (once there is new state in here), as otherwise it may not round-trip.  Thinking about this, this is pretty important, because we need to figure out what the representation will be, and then I guess we need to consider whether that representation is "bad enough" to warrant a breaking change.

pkg/apis/kops/validation/legacy.go
```diff
@@ -55,41 +55,54 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+			allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("openstack"), "only one cloudProvider option permitted"))
+		}
+		optionTaken = true
+		requiresNetworkCIDR = false
 		requiresSubnetCIDR = false
-
-	case kops.CloudProviderDO:
+	}
+	if !optionTaken {
+		allErrs = append(allErrs, field.Required(fieldSpec.Child("cloudProvider"), ""))
```
justinsb, line 102: Is there a reason to set requiresSubnets /  requiresSubnetCIDR / requiresNetworkCIDR  here?  Changing the order makes the change a harder to review (I'll check the individual PRs)

pkg/apis/kops/validation/validation.go
```diff
@@ -1552,11 +1552,11 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
 func validateClusterAutoscaler(cluster *kops.Cluster, spec *kops.ClusterAutoscalerConfig, fldPath *field.Path) (allErrs field.ErrorList) {
 	allErrs = append(allErrs, IsValidValue(fldPath.Child("expander"), spec.Expander, []string{"least-waste", "random", "most-pods", "price", "priority"})...)
 
-	if fi.StringValue(spec.Expander) == "price" && kops.CloudProviderID(cluster.Spec.CloudProvider) != kops.CloudProviderGCE {
+	if fi.StringValue(spec.Expander) == "price" && cluster.Spec.CloudProvider.GCE == nil {
```
justinsb, line 1555: Should we just use `GetCloudProvider() != kops.CloudProviderGCE` ?  (Any reason for the different pattern?)

pkg/model/components/containerd_test.go
```diff
@@ -27,7 +27,9 @@ import (
 func buildContainerdCluster(version string) *kopsapi.Cluster {
 	return &kopsapi.Cluster{
 		Spec: kopsapi.ClusterSpec{
-			CloudProvider:     "aws",
+			CloudProvider: kopsapi.CloudProviderSpec{
```
justinsb, line 30: I'm wondering if we should have a SetCloudProvider helper.

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -289,14 +289,14 @@ spec:
           command:
             - ./cluster-autoscaler
             - --balance-similar-node-groups={{ .BalanceSimilarNodeGroups }}
-            - --cloud-provider={{ $.CloudProvider }}
-            {{ if (eq $.CloudProvider "aws") }}
+            - --cloud-provider={{ GetCloudProvider }}
```
justinsb, line 292: Could we name this CloudProvider (for the templates)?

justinsb:
Thanks for doing this!

Very much favor introducing GetCloudProvider; that is a big win.

Code looks good; a few suggestions/quibbles but nothing blocking.

I was thinking about the JSON representation and round-tripping.  In the light of our ingress experience, do we want to introduce a breaking API change?  Do we get a pay off if we don't delete v1alpha2?

Let's discuss in kOps office hours / bonus hours.

/approve
/lgtm

/hold for discussion

pkg/apis/kops/v1alpha2/cluster.go
johngmyers: When we add a new field in here we will either have to add manual conversion code or find a place to put a v1alpha2 `CloudProviderSpec` with the existing fields suppressed. The tricky thing about the latter approach is suppressing serialization of the new struct when there are no non-empty, non-suppressed members.

pkg/apis/kops/validation/legacy.go
johngmyers: I figured alphabetizing would make it easier to maintain going forward.

pkg/apis/kops/validation/validation.go
johngmyers: I've been considering the getter to be largely transitional. I've been targeting the patterns we use with CNIs.

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
johngmyers: I patterned this off of GetNodeInstanceGroups and HasHighlyAvailableControlPlane. Should template functions that call getters be named as getters or attributes?

johngmyers:
When we get to v1beta1 I was planning on having one kops release before changing the storage format to v1beta1. I don't see a big payoff in deleting v1alpha1 early and it would be helpful to users to allow them to upgrade from pre-v1beta1 versions of kops as long as there's overlapping k8s version support.

johngmyers:
Since we don't have controllers which do read-modify-write on these objects, once we change the storage format to the new version we could slack off on making sure new fields round-trip through the old API version.

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
justinsb: A good question - I was more just thinking it was odd to go from `$.CloudProvider` to `GetCloudProvider`.  If there's precedent, what you've done makes sense

pkg/apis/kops/validation/validation.go
justinsb: Oh OK - I quite like the getter and I think it would be nice if we had a similar helper method for CNIs.  I don't think we need to surface it in the API, just that it seems to improve the code quality - less casting etc.  But just something to consider!

justinsb:
/approve
/lgtm

The test looks like a CNI failure, I'm not sure it should hold up this PR from merging - maybe in our meeting today someone will know the "magic command".

PR #13063, opened 2021-12-31
Map kube-apiserver egress-selector-config-file flag

justinsb:


johngmyers:
Is the intention that this file be laid down by a hook? If it is instead to be managed by kOps I would prefer not exposing it though the API.

pkg/apis/kops/componentconfig.go
```diff
@@ -314,6 +314,8 @@ type KubeAPIServerConfig struct {
 	ServiceClusterIPRange string `json:"serviceClusterIPRange,omitempty" flag:"service-cluster-ip-range"`
 	// Passed as --service-node-port-range to kube-apiserver. Expects 'startPort-endPort' format e.g. 30000-33000
 	ServiceNodePortRange string `json:"serviceNodePortRange,omitempty" flag:"service-node-port-range"`
+	// EgressSelectorConfigFile is the file path with api-server egress selector configuration.
+	EgressSelectorConfigFile string `json:"egressSelectorConfigFile,omitempty" flag:"egress-selector-config-file"`
```
johngmyers, line 318:
```suggestion
	EgressSelectorConfigFile string `json:"-" flag:"egress-selector-config-file"`
```

I see in the baremetal PR that there's a plan for kOps to manage this file. So is there a need to have the admin configure it?

pkg/apis/kops/v1alpha2/componentconfig.go
```diff
@@ -314,6 +314,8 @@ type KubeAPIServerConfig struct {
 	ServiceClusterIPRange string `json:"serviceClusterIPRange,omitempty" flag:"service-cluster-ip-range"`
 	// Passed as --service-node-port-range to kube-apiserver. Expects 'startPort-endPort' format e.g. 30000-33000
 	ServiceNodePortRange string `json:"serviceNodePortRange,omitempty" flag:"service-node-port-range"`
+	// EgressSelectorConfigFile is the file path with api-server egress selector configuration.
+	EgressSelectorConfigFile string `json:"egressSelectorConfigFile,omitempty" flag:"egress-selector-config-file"`
```
johngmyers, line 318:
```suggestion
	EgressSelectorConfigFile string `json:"-" flag:"egress-selector-config-file"`
```

I see in the baremetal PR that there's a plan for kOps to manage this file. So is there a need to have the admin configure it?

justinsb: I don't believe we've historically made this distinction.  The configuration objects that are componentconfig express all the configuration options.  The user can specify them, but if they do we can't guarantee compatibility.  If the values aren't set, kOps can populate them as appropriate.

This is another reason why it would be good to split out e.g. the kubeapiserver config from the cluster config, because of the different philosophy / guarantee in the same object, only separate by a nesting level.

That said, I am actively thinking about how we want to manage this file.  It's an odd one, and I'm thinking we may want to treat apiserver.k8s.io/EgressConfiguration as another configuration object, that - like [KubeletConfiguration](https://github.com/kubernetes/kubelet/blob/71a1ccd1b3e715e00084484796348d6668eb86bb/config/v1beta1/types.go#L88) gets written to a file instead of to the API.  So I don't mind if we'd rather hold off on this PR and wait for my exploration to get a little further along.

johngmyers: In the v1alpha3 API, I've been removing the kops-managed settings from the API.

PR #13064, opened 2021-12-31
WIP: Introduce config file for kubelet

justinsb:
Trying out a simpler way of doing this, where we map directly to the yaml structure.  This should avoid needing to create another struct.


johngmyers:
My only concern here is if we end up needing to support two different apiversions of a config file. We might want to make the `configfile` tag a parameter to `MapToYAML`.

johngmyers:
Or an ordered list of tags. I suppose we could code that when we need it.

PR #13065, opened 2021-12-31
Move Azure settings to cloudProvider.azure

johngmyers:


pkg/apis/kops/componentconfig.go
```diff
@@ -823,10 +823,10 @@ type OpenstackConfiguration struct {
 	Metadata           *OpenstackMetadata           `json:"metadata,omitempty"`
 }
 
-// AzureConfiguration defines Azure specific cluster configuration.
-type AzureConfiguration struct {
+// AzureSpec defines Azure specific cluster configuration.
+type AzureSpec struct {
 	// SubscriptionID specifies the subscription used for the cluster installation.
-	SubscriptionID string `json:"subscriptionId,omitempty"`
+	SubscriptionID string `json:"subscriptionID,omitempty"`
```
justinsb, line 829: Aside: Are we actually using this JSON encoding anywhere?  I know there are some places e.g. in nodeup where we aren't we precise with the version, but this is in the unversioned form...

justinsb:
This LGTM.  I don't know if I'm reviewing them in priority order, so will add a hold, but feel free to cancel the hold to determine the order you want these to merge in!

/approve
/lgtm
/hold

pkg/apis/kops/componentconfig.go
johngmyers: I'm not sure. I think we keep it in sync with the latest versioned API just to make updating easier.

PR #13081, opened 2022-01-05
Support price and priority cluster-autoscaler expanders

danports:
This PR adds support for both of the remaining cluster-autoscaler expanders kops doesn't currently support. The `price` expander requires no additional configuration. The `priority` expander requires configuration in a separate ConfigMap, so I've added notes to the docs to that effect.

### DIFF
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -46,7 +46,7 @@ spec:
     memoryRequest: "300Mi"
 ```
 
-Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
+Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler). The autoscaler supports five different expander strategies; the `priority` expander requires additional configuration in a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md).
 
 ##### Disabling cluster autoscaler for a given instance group
 {{ kops_feature_table(kops_added_default='1.20') }}
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -49,7 +49,9 @@ spec:
 Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
 
 ##### Expander strategies
-The cluster autoscaler supports five different [expander strategies](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders). The `price` expander is only supported on GCE. The `priority` expander requires additional configuration in a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md) - please ensure that you have deployed this ConfigMap before enabling the `priority` expander.
+Cluster autoscaler supports several different [expander strategies](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders).
+
+Note that the `priority` expander requires additional configuration through a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md) - you will need to create this ConfigMap in your cluster before selecting this expander.
 
 ##### Disabling cluster autoscaler for a given instance group
 {{ kops_feature_table(kops_added_default='1.20') }}
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -46,7 +46,10 @@ spec:
     memoryRequest: "300Mi"
 ```
 
-Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler). The autoscaler supports five different expander strategies; the `priority` expander requires additional configuration in a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md) - please ensure that you have deployed this ConfigMap before enabling the `priority` expander.
+Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
+
+##### Expander strategies
+The cluster autoscaler supports five different [expander strategies](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders). The `price` expander is only supported on GCE. The `priority` expander requires additional configuration in a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md) - please ensure that you have deployed this ConfigMap before enabling the `priority` expander.
 
 ##### Disabling cluster autoscaler for a given instance group
 {{ kops_feature_table(kops_added_default='1.20') }}
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -46,7 +46,7 @@ spec:
     memoryRequest: "300Mi"
 ```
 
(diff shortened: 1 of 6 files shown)


pkg/apis/kops/validation/validation.go
```diff
@@ -1517,7 +1517,7 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
 }
 
 func validateClusterAutoscaler(cluster *kops.Cluster, spec *kops.ClusterAutoscalerConfig, fldPath *field.Path) (allErrs field.ErrorList) {
-	allErrs = append(allErrs, IsValidValue(fldPath.Child("expander"), spec.Expander, []string{"least-waste", "random", "most-pods"})...)
+	allErrs = append(allErrs, IsValidValue(fldPath.Child("expander"), spec.Expander, []string{"least-waste", "random", "most-pods", "price", "priority"})...)
```
olemarkus, line 1520: Is price actually supported on AWS? If it is only for certain cloud vendors, we should validate that.

danports: Good point, I could add validation and docs for that.

docs/addons.md
```diff
@@ -46,7 +46,7 @@ spec:
     memoryRequest: "300Mi"
 ```
 
-Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
+Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler). The autoscaler supports five different expander strategies; the `priority` expander requires additional configuration in a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md).
```
olemarkus, line 49: I think it is worth adding something like "please make sure that you have deployed this ConfigMap before enabling the priority expander"

danports: Sure, I can reword this. FWIW, the cluster-autoscaler docs make it sound like it will fall back to selecting a random node group if the expander selected can't pick one, so it may not actually break things too much if you try to use the `price` expander on AWS or the `priority` expander without a ConfigMap, but I haven't tested that behavior.

danports: Added this language to the docs.

pkg/apis/kops/validation/validation.go
danports: Added validation and docs. My knowledge of Go is basically zero so please let me know if I've done anything dumb.

pkg/apis/kops/validation/validation.go
```diff
@@ -1517,7 +1517,11 @@ func validateNodeLocalDNS(spec *kops.ClusterSpec, fldpath *field.Path) field.Err
...
+
+	if spec.Expander != nil && *spec.Expander == "price" && kops.CloudProviderID(cluster.Spec.CloudProvider) != kops.CloudProviderGCE {
```
olemarkus, line 1522:
```suggestion
	if fi.StringValue(spec.Expander) == "price" && kops.CloudProviderID(cluster.Spec.CloudProvider) != kops.CloudProviderGCE {
```

docs/addons.md
```diff
@@ -48,6 +48,9 @@ spec:
 
 Read more about cluster autoscaler in the [official documentation](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
 
+##### Expander strategies
+The cluster autoscaler supports five different [expander strategies](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders). The `price` expander is only supported on GCE. The `priority` expander requires additional configuration in a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md) - please ensure that you have deployed this ConfigMap before enabling the `priority` expander.
```
olemarkus, line 52:
```suggestion
Cluster autoscaler supports different [expander strategies](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders).

Please note that the "priority" expander requires additional configuration through a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md). The expander requires that the the ConfigMap exists before being enabled.
```

olemarkus (commented):
Just some nits, then it's good to go.

Thanks!

eddycharly:
Would it make sense to specify the priorities in the addon spec ?

docs/addons.md
danports: I'm puzzled about this change - it looks like you removed the note about the `price` expander only being supported on GCE. Wouldn't we want to have a note about this in the docs if we're going to validate it in the code? 🤔

danports:
@eddycharly That did occur to me, and I can definitely see the value in that, but this approach was simpler and I wasn't sure how far we wanted to go down the route of essentially building Helm charts in kops. What do you think @olemarkus? It looks like it wouldn't be difficult to add to the YAML template, though I am unsure of some of the semantics there - e.g. would the ConfigMap get removed automatically if the user switched to a different expander the way it would with a Helm chart?

eddycharly:
@danports it was just a thought and can be added later if necessary. It would be nice if the addon could be self contained and not require a manual operation on the end user side.

About deleting the config map, i'm not sure kops supports this, but keeping the config map even if the expander does not need it shouldn't hurt anyway.

docs/addons.md
olemarkus: We don't need to duplicate CAS docs where things are graceful. Users can't "harm" themselves if they tried to use price on AWS.
Starting CAS with priority expander without the configmap would be "harmful" though.

olemarkus:
kOps does support pruning of manifests. E.g nodeterminationhandler uses this as it can move from a DaemonSet to a Deployment. This is done by setting `addon.BuildPrune` in `bootstrapchannelbuilder.go`.

But as you say, building that configmap is a bit comlex and will expand the API a bit. I am not against that change, but I suppose there is a reason we didn't add support for this expander before.

Earlier, I have also been thinking that this addon may not be the right thing for those who want to configure CAS to this level of detail.

danports:
How do we define the right level of detail for an addon? I understand not wanting to overcomplicate or reinvent the wheel with our addons, but on the other hand, arbitrarily setting a line beyond which the addons should not go can lead to a frustrating UX, where you start with an addon out of the box with your shiny new cluster, and then configure more options over time, until the point at which you want to add just one more option, but the addon doesn't support it, so you have to disable it and reconfigure the whole thing from scratch with a separate manifest or Helm chart. (I ran into a similar situation recently with the external-dns addon, which is admittedly fairly new and so will probably accumulate more options over time.)

I think the wisest approach would be to release addons with a fairly basic set of options and allow folks to vote with their PRs so to speak and contribute the addon options that would be most useful to them.

All that to say I think in this case we should merge this as it stands now (pending that open doc nit) and consider adding an option for the priority ConfigMap later. 🙂 Thoughts?

docs/addons.md
danports: OK, I understand the distinction you are drawing, and I agree that it is better not to duplicate the autoscaler's docs unnecessarily. I reworded the docs in my latest commit.

olemarkus:
> How do we define the right level of detail for an addon? I understand not wanting to overcomplicate or reinvent the wheel with our addons, but on the other hand, arbitrarily setting a line beyond which the addons should not go can lead to a frustrating UX, where you start with an addon out of the box with your shiny new cluster, and then configure more options over time, until the point at which you want to add just one more option, but the addon doesn't support it, so you have to disable it and reconfigure the whole thing from scratch with a separate manifest or Helm chart. (I ran into a similar situation recently with the external-dns addon, which is admittedly fairly new and so will probably accumulate more options over time.)
> 

Sorry, I was only explaining my thinking. When someone does PRs to implement additional support we tend to be permissive with what functionality gets added. We only need to be careful about the design and ensure it is maintainable. So e.g with the configmap it may be that we'd just inline the whole yaml similar to `spec.kubeDNS.ExternalCoreFile`. There would be more discussion on that particular part.


> I think the wisest approach would be to release addons with a fairly basic set of options and allow folks to vote with their PRs so to speak and contribute the addon options that would be most useful to them.

Absolutely. 

> All that to say I think in this case we should merge this as it stands now (pending that open doc nit) and consider adding an option for the priority ConfigMap later. slightly_smiling_face Thoughts?

100% agree.

PR #13088, opened 2022-01-08
Add ipv6 to relnotes

olemarkus:
We should hilight the state of ipv6 in the relnotes.


johngmyers:
I'm not sure this quite qualifies as "significant". It's still work in progress, not all that useful yet.

olemarkus:
Fair enough. Moved this down to "other changes of notes".

johngmyers (approved):
/retest

PR #13090, opened 2022-01-11
Containerd Config file path is now tweakable

bbourbie:
Some containerd shims can be installed via daemonSets.
(eg. katacontainers)

Usually, these daemonsets  add the shims and modify the containerd config file on the fly
by adding the necessary parameters for their shims.

Example of such code
https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/scripts/kata-deploy.sh

The config file path is usually at /etc/containerd/config.toml but
kops has set the path to /etc/containerd/config-kops.toml

I felt it was necessary to add that option because changing the
config file path to the regular path could have been seen as
breaking change.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -691,6 +691,9 @@ spec:
                     description: ConfigOverride is the complete containerd config
                       file provided by the user.
                     type: string
+                  ConfigFilePath:
+                    description: ConfigFilePath Overwrites the containerd config file path by default
+                    type: string
                   logLevel:
                     description: LogLevel controls the logging details [trace, debug,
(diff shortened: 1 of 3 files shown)


hakman:
@bbourbie The config file name is a relic from the times when kOps installed containerd from deb/rpm.
Renaming the config file to `config.toml` was something I was planning to do sooner or later so should be fine to do that instead.
The config file path should still be added in case someone wants the backwards compatibility.
The new option should be added also to the `v1alpha2` and `v1alpha3` API versions and then run `make crds` and `make apimachinery`.

@olemarkus Any thoughts?

olemarkus:
All of that makes sense to me.

czomo:
Hey guys! Any update on this? Without it we can't use dragonfly p2p image system.

PR #13093, opened 2022-01-12
fix ipv4+ipv6 sec groups/listeners in OpenStack

zetaab:
seems that https://github.com/kubernetes/kops/pull/13032 is not working correctly when updating the existing cluster. It do solves the problem that cluster can be created from the scratch, but updating existing clusters does not work at all.

This PR is modifying security groups working correctly. However, there is no easy way to make api loadbalancer to support ipv6 addresses like this. It needs more investigation how that can be done. I am pretty sure that none is using ipv6 api loadbalancer because it has not been supported by kOps

cc @iGene


zetaab:
@olemarkus could you help reviewing this? Currently master / release-1.23 does not work for openstack

olemarkus (approved):

PR #13100, opened 2022-01-13
WIP use AWS LBC to create the API server load balancer

olemarkus:
Just testing if this is feasible at all.
LBC can do IP targeting and may work on IPv6-only networks. This makes a few addons non-optional though.


johngmyers:
I'm not sure we necessarily want to pull in LBC for this. We could have kops-controller update the IP-based targets.

olemarkus:
Yeah not sure either.
At first I was thinking of using protokube actually, but the complexity of handling and cleaning up targets is not insignificant. Hence LBC.

johngmyers:
I don't think it has to be in protokube. The node would presumably be able to come up enough to run kops-controller before it needs to be made a target of the load balancer. But getting all the dependencies of LBC up before the control plane is healthy could be a challenge.

PR #13103, opened 2022-01-13
add node-drain-timeout flag to rolling-update

heybronson:
sets default eviction timeout to 15minutes. Previously, rolling-updates would retry infinitely when failing to drain a given node.

We can now properly rely on `--fail-on-drain-error=false/true` in the event that it takes longer than 15 minutes to drain a node. 

Additionally, expose a new `--drain-timeout` flag to configure the max drain timeout if needed.

I decided to name the flag `drain-timeout` to be consistent with the naming conventions of other flags. Specifically, `drain-timeout` aligns with flags:
- `validation-timeout`
- `fail-on-drain-error`
- `post-drain-delay`

### DIFF
--- a/docs/releases/1.23-NOTES.md
+++ b/docs/releases/1.23-NOTES.md
@@ -62,6 +62,7 @@ It is recommended to keep using the `v1alpha2` API version.
 
 * IPv6 pod subnets is in a working state using public IPv6 addresses for the Pod network. This works with both Cilium and Calico. IPv6 is still behind a feature flag until service controllers and addons implement support for IPv6. See [the IPv6 documentation](https://kops.sigs.k8s.io/networking/ipv6/).
 
+* The `kops rolling-update cluster` command has a new `--drain-timeout` flag for specifying the maximum amount of time to wait when attempting to drain a node. Previously, rolling-updates would attempt to drain a node for an indefinite amount of time. If `--drain-timeout` is not specified, a default of 15 minutes is applied.
 
 # Full change list since 1.22.0 release
 
--- a/cmd/kops/rolling-update_cluster.go
+++ b/cmd/kops/rolling-update_cluster.go
@@ -107,6 +107,9 @@ type RollingUpdateOptions struct {
 	// does not validate, after a validation period.
 	FailOnValidate bool
 
+	// DrainTimeout is the maximum time to wait while draining a node.
+	DrainTimeout time.Duration
+
 	// PostDrainDelay is the duration of a pause after a drain operation
 	PostDrainDelay time.Duration
 
@@ -154,6 +157,8 @@ func (o *RollingUpdateOptions) InitDefaults() {
 	o.PostDrainDelay = 5 * time.Second
 	o.ValidationTimeout = 15 * time.Minute
 	o.ValidateCount = 2
+
+	o.DrainTimeout = 15 * time.Minute
 }
 
 func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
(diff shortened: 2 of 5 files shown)


olemarkus:
I do think we need to make this timeout configurable before it goes in.

heybronson:
>I do think we need to make this timeout configurable before it goes in.

@olemarkus I've added a new flag `--node-drain-timeout` to make this configurable.

olemarkus:
Looks good. Can you add a line to the release notes under "other changes of note"?

docs/releases/1.23-NOTES.md
```diff
@@ -62,6 +62,7 @@ It is recommended to keep using the `v1alpha2` API version.
 
 * IPv6 pod subnets is in a working state using public IPv6 addresses for the Pod network. This works with both Cilium and Calico. IPv6 is still behind a feature flag until service controllers and addons implement support for IPv6. See [the IPv6 documentation](https://kops.sigs.k8s.io/networking/ipv6/).
 
+* The `kops rolling-update cluster` command has a new `--node-drain-timeout` flag for specifying the maximum amount of time to wait when attempting to drain a node.
```
olemarkus, line 65: I think the change from hanging perpetually to 15 min by default is worth mentioning.

olemarkus:
When you are done, can you also squash?

olemarkus (commented):
/lgtm

johngmyers:
I'm not sure about that default. It seems awfully short.

johngmyers:
It seems to me this should be set through the IG spec rather than be a flag. The desired timeout derives more from the type of workload than the circumstances of a particular rolling update invocation. I think it's a `RollingUpdate` strategy, so belongs in that struct.

olemarkus:
I think the timeout is reasonable given that it only causes distruptions if one has also set ` --fail-on-drain-error=false`. In other cases, upgrade will bail, which I think is good behavior as in almost all cases, this would be caused by bad PDBs that need further investigation. Those who know they need a long time to rotate their nodes because of graceful terminations/finalizers etc, can easily adjust their settings.

I do agree that this fit nicely in the rollingUpdate struct. One especially need to be careful with drain errors on the control plane.

heybronson:
The reason why I added `--node-drain-timeout` as a flag instead of a `RollingUpdate` field is because `--fail-on-drain-error` is also a flag. Since both have have an impact on the overall rolling-update behavior, I figured it made sense for them to both be flags.

If you folks feel that it would be better as a new `RollingUpdate` field, I'm happy to update. @johngmyers @olemarkus

olemarkus:
We have similar master vs node flags. I'll accept either way, I think.

heybronson:
My opinion is that the current flag configuration in this makes sense; I'd want finer control over both flags depending on what cluster environment I'm operating on, or if a CICD pipeline is doing the operation.

olemarkus:
The flags prefixed with `node` is expected to work only on worked nodes. While the ones on master is prefixed with `master`. I think that convention should at least be maintained. Sorry for not catching that on the first run.

heybronson:
> The flags prefixed with node is expected to work only on worked nodes. While the ones on master is prefixed with master.

That makes sense to me. In that case, I think I'll add `DrainTimeout` to the `RollingUpdate` struct instead, like @johngmyers mentioned. It'll be more flexible in allowing folks to define different timeouts depending on the instanceGroup.

heybronson:
I've decided to change the name from `node-drain-timeout` to `drain-timeout`. This keeps the naming convention consistent with other flag names like `post-drain-delay`, `fail-on-drain-error`, and `validation-timeout`.

heybronson:
@johngmyers do you want to take another look now that the new flag `drain-timeout` corresponds to existing flags better?

PR #13104, opened 2022-01-14
Create helper function for ec2 create/tag-on-create IAM permissions

olemarkus:
It's quite difficult to get tag-on-create correctly for EC2 because of the createAction rules etc. The policies coming from AWS are also quite different in strictness, while trying to achieve the same thing. This helper function should help us be consistent with how we create these rules.

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -61,6 +61,50 @@ func (p *Policy) AddUnconditionalActions(actions ...string) {
 	p.unconditionalAction.Insert(actions...)
 }
 
+func (p *Policy) AddEC2CreateAction(actions, resources []string, partition string) {
+	actualActions := []string{}
+	for _, action := range actions {
+		actualActions = append(actualActions, "ec2:"+action)
+	}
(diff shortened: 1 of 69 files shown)


pkg/model/iam/iam_builder.go
```diff
@@ -962,48 +1002,15 @@ func AddAWSEBSCSIDriverPermissions(p *Policy, partition string, appendSnapshotPe
...
-				},
-			),
-			Condition: Condition{
-				"StringEquals": map[string]string{
-					"aws:ResourceTag/KubernetesCluster": p.clusterName,
-				},
-			},
+		[]string{
+			fmt.Sprintf("arn:%v:ec2:*:*:volume/*", partition),
+			fmt.Sprintf("arn:%v:ec2:*:*:snapshot/*", partition),
```
johngmyers, line 1013: Perhaps some of this partition logic could be pushed down?

olemarkus: partition should probably be on the policy object itself. but that seems like something for a follow-up.

johngmyers (approved):

PR #13109, opened 2022-01-16
Log OIDC providers we don't delete

olemarkus:


olemarkus:
/hold
Just to see why we have ~100 OIDC providers in our account.

justinsb:
We should probably mark this one as "WIP" or "housekeeping", because with the `foundK8sTag` check it's pretty dangerous if we accidentally merged it :-)

justinsb:
FYI I am proposing that the janitor also clean these up: https://github.com/kubernetes-sigs/boskos/pull/121

There will always be some instances where the e2e run is interrupted and `kops delete cluster` is not run.

PR #13113, opened 2022-01-16
Allow PrefixList for sshAccess and kubernetesApiAccess

hierynomus:
Signed-off-by: Jeroen van Erp <jeroen@hierynomus.com>


rifelpet:
kops' cloudmock packages will need to be updated to recognize the new prefix list field. All of the changes needed should be in [here](https://github.com/kubernetes/kops/blob/master/cloudmock/aws/mockec2/securitygroups.go)

hierynomus:
Thanks for the pointer @rifelpet. Awaiting new build results now ;)

pkg/apis/kops/validation/validation.go
```diff
@@ -85,17 +85,23 @@ func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *fie
 
 	// SSHAccess
 	for i, cidr := range spec.SSHAccess {
-		allErrs = append(allErrs, validateCIDR(cidr, fieldPath.Child("sshAccess").Index(i))...)
+		if !strings.HasPrefix(cidr, "pl-") {
```
hierynomus, line 88: Should we (can we) add validation that the PrefixListID is actually valid by interrogating AWS, or don't we do that during validation?

hierynomus:
@rifelpet Another question: Should we also add the PrefixList for the "ICMP for PMTU discovery" block? See:

https://github.com/kubernetes/kops/blob/7e645187bca6716f11adaef474fd976b14f56f40/pkg/model/awsmodel/api_loadbalancer.go#L379-L381

If so, what should be the `FromPort` and `ToPort`?

pkg/apis/kops/validation/validation.go
johngmyers: Shouldn't this only be allowed if the cloudprovider is AWS?

pkg/model/awsmodel/utils.go
```diff
@@ -0,0 +1,42 @@
...
+	"k8s.io/kops/upup/pkg/fi/cloudup/awstasks"
+	"k8s.io/kops/upup/pkg/fi/utils"
+)
+
+// IsPrefixListID checks whether the string resembles a prefix list ID
+func IsPrefixListID(id string) bool {
+	return strings.HasPrefix(id, "pl-")
+}
+
+func LimitIngress(t *awstasks.SecurityGroupRule, cidr string) *awstasks.SecurityGroupRule {
```
johngmyers, line 32: Perhaps make this a receiver on `awstasks.SecurityGroupRule`? Perhaps call it `SetPrefix()`?

docs/cluster_spec.md
```diff
@@ -257,6 +257,8 @@ spec:
     - 12.34.56.78/32
 ```
 
+Instead of listing all CIDRs, it is possible to specify a pre-existing [AWS Prefix List](https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html) ID
```
johngmyers, line 260:
```suggestion
Instead of listing all CIDRs, it is possible to specify a pre-existing [AWS Prefix List](https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html) ID.
```

docs/cluster_spec.md
```diff
@@ -269,6 +271,8 @@ spec:
...
```
johngmyers, line 274:
```suggestion
Instead of listing all CIDRs, it is possible to specify a pre-existing [AWS Prefix List](https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html) ID.
```

johngmyers (changes requested):
Perhaps modify one of the IPv6 e2e tests to use this?

johngmyers:
As to ICMP/ICMPv6, you would need to query whether the prefix list in question contains IPv4 or IPv6 addresses, then add the appropriate rule. Per AWS documentation, a prefix list may contain only one address family.

pkg/model/awsmodel/utils.go
hierynomus: I've added that as `SetCidrOrPrefix`, as I thought that `SetPrefix` wasn't fully covering the functionality.

docs/cluster_spec.md
hierynomus: Updated that to include the `.`

docs/cluster_spec.md
hierynomus: Updated that to include the `.`

hierynomus:
> As to ICMP/ICMPv6, you would need to query whether the prefix list in question contains IPv4 or IPv6 addresses, then add the appropriate rule. Per AWS documentation, a prefix list may contain only one address family.

I've added this functionality, by passing the `fi.Cloud.(awsup.AWSCloud)` into the builder struct. Feel free to suggest a better way to obtain the AWS API. 

I've kept the commit separate for now to ease the review process, I will squash it once you've OK'ed the implementation 😄

pkg/model/awsmodel/utils.go
johngmyers: My thinking was that a CIDR specifies a prefix. A "pl-" string is a prefix *list*. Both CIDR and PrefixList specify the subset of the address space that the SGR applies to.

I suppose "Prefix" is incorrectly singular when applied to prefix lists. I'm trying to find a good term to express the concept of the subset of address space.

hierynomus: Yes agreed, basically either limit the ingress to certain address(blocks).

So that's why I initially chose LimitIngress

johngmyers: Except a security group rule can be for egress instead.

hierynomus:
Any ideas already @johngmyers on the naming of the method? Furthermore besides that, is there anything else needed? I've updated the `complex`e2e test to include 2 prefix lists (1 ipv4, 1 ipv6), which other e2e test needs something extra?

johngmyers:
Sorry, I haven't been able to get back to this due to the large review backlog.

upup/pkg/fi/cloudup/awstasks/securitygrouprule.go
```diff
@@ -191,6 +210,9 @@ func (_ *SecurityGroupRule) CheckChanges(a, e, changes *SecurityGroupRule) error
 		if e.CIDR != nil && e.IPv6CIDR != nil {
 			return field.Forbidden(field.NewPath("CIDR/IPv6CIDR"), "Cannot set more than 1 CIDR or IPv6CIDR")
 		}
+		if e.PrefixList != nil && (e.CIDR != nil || e.IPv6CIDR != nil) {
+			return field.Forbidden(field.NewPath("CIDR/IPv6CIDR"), "Cannot set CIDR or IPv6CIDR when PrefixList is set")
```
hakman, line 214:
```suggestion
			return field.Forbidden(field.NewPath("PrefixList"), "Cannot set PrefixList when CIDR or IPv6CIDR is set")
```

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -418,35 +426,44 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+				icmpIpv6 := false
+				if strings.HasPrefix(cidr, "pl-") {
+					b, err := b.IsIPv6PrefixList(cidr)
+					if err != nil {
+						return err
+					}
+					icmpIpv6 = b
+				} else {
+					icmpIpv6 = utils.IsIPv6CIDR(cidr)
+				}
```
hakman, line 442: I don't think model should check the kind of prefix with AWS. I would rather skip the ICMP rules for prefixes, for now.

As a curiosity, any idea what happens if you set IPv4 PrefixList to ICMPv6 protocol or IPV6 PrefixList to ICMPv4 protocol?

docs/cluster_spec.md
```diff
@@ -257,6 +257,8 @@ spec:
     - 12.34.56.78/32
 ```
 
+Instead of listing all CIDRs, it is possible to specify a pre-existing [AWS Prefix List](https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html) ID.
```
hakman, line 260:
```suggestion
{{ kops_feature_table(kops_added_default='1.23') }}

In AWS, instead of listing all CIDRs, it is possible to specify a pre-existing [AWS Prefix List](https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html) ID.
```

docs/cluster_spec.md
```diff
@@ -269,6 +271,8 @@ spec:
...
```
hakman, line 274:
```suggestion
{{ kops_feature_table(kops_added_default='1.23') }}

In AWS, instead of listing all CIDRs, it is possible to specify a pre-existing [AWS Prefix List](https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html) ID.
```

hakman (changes requested):
Hi @hierynomus. I did a first pass of the PR.
There are mostly small changes from my point of view, but would like to not do cloud calls in model.

pkg/apis/kops/validation/validation.go
```diff
@@ -85,17 +85,35 @@ func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *fie
 
 	// SSHAccess
 	for i, cidr := range spec.SSHAccess {
-		allErrs = append(allErrs, validateCIDR(cidr, fieldPath.Child("sshAccess").Index(i))...)
+		if strings.HasPrefix(cidr, "pl-") {
+			if kops.CloudProviderID(spec.CloudProvider) != kops.CloudProviderAWS {
```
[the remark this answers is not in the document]
hierynomus: Actually your suggestion exhibits different behaviour... If we'd take your suggestion, it will need another `else if` because the `pl-` prefix is allowed when it is on AWS. So that's why I made a nested if.

pkg/model/awsmodel/api_loadbalancer.go
hierynomus: Just tried that out, AWS does allow to configure it like that, but not sure what the net effect is. I added a ICMPv4 inbound rule with an IPv6 type PrefixList, and the console accepted and created that.

pkg/apis/kops/validation/validation.go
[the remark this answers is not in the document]
hakman: Somehow was expecting the `strings.HasPrefix(cidr, "pl-")` check to also be part of `validateCIDR()`.

pkg/model/awsmodel/api_loadbalancer.go
hakman: So, let's skip this for now, to get it into 1.23 before the release. We can try later to add the ICMP rules and see what happens.

pkg/apis/kops/validation/validation.go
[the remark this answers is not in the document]
hierynomus: No, I actually didn't add that, as that method is used to check subnet.CIDR also where I don't think a prefix list would be allowed.

pkg/model/awsmodel/api_loadbalancer.go
hierynomus: I'm not sure what the effect would be of not rendering ICMP rules when a prefix list is in use? I took the approach of querying the AWS API from the IAM model, where that was already used, so I figured that was allowed ;).

Are you sure we can skip them?

pkg/apis/kops/validation/validation.go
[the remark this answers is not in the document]
hakman: 👍

pkg/model/awsmodel/api_loadbalancer.go
hakman: It is required for [Path MTU Discovery](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/network_mtu.html#path_mtu_discovery).
I am sure I don't want to call the cloud for this. 😄

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -418,35 +415,36 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
 						Protocol:      fi.String("icmpv6"),
 						SecurityGroup: masterGroup.Task,
 						ToPort:        fi.Int64(-1),
-					})
-				} else {
-					c.AddTask(&awstasks.SecurityGroupRule{
+					}
+					t.SetCidrOrPrefix(cidr)
+					c.AddTask(t)
+				} else if utils.IsIPv4CIDR(cidr) {
```
hakman, line 437:
```suggestion
				} else {
```

cloudmock/aws/mockec2/prefixlists.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"k8s.io/klog/v2"
+)
+
+func (m *MockEC2) DescribeManagedPrefixLists(request *ec2.DescribeManagedPrefixListsInput) (*ec2.DescribeManagedPrefixListsOutput, error) {
```
hakman, line 28: Are these used anywhere? Could you do some cleanup in mock?

hakman (changes requested):

cloudmock/aws/mockec2/prefixlists.go
hierynomus: Probably not anymore now that we've removed the PMTU ICMP security group rules.

hierynomus: Cleaned up the mock

hakman:
Great. Would you mind also squashing the commits before merging the PR?

hierynomus:
> Great. Would you mind also squashing the commits before merging the PR?

Sure, will do that now!

hakman (approved):
Looks great. Thanks for taking the time to add this feature!

PR #13114, opened 2022-01-17
Add DescribeRegions to nodeup privs

### LINKED ISSUE #13079: nodeup fails with missing ec2:DescribeRegions in edge case
/kind bug

 this information.**
v1.22.2

 a `kops` flag.**
v1.21.7

AWS

`kops replace`, `kops update cluster`, `kops rolling-update cluster`

nodeup fails on new worker nodes.

```
Jan  4 14:52:38 ip-10-225-8-165 nodeup[2859]: W0104 14:52:38.373744    2859 main.go:133] got error running nodeup (will retry in 30s): error loading Cluster "s3://my-state-store/my-cluster.example.com/cluster-completed.spec": Unable to list AWS regions: UnauthorizedOperation: You are not authorized to perform this operation.
```

nodeup completes successfully.

Not relevant.

kOps commands complete successfully.

With kOps 1.21.2 the nodes IAM role had the ec2:DescribeRegions grant, so everything worked fine.

The issue seems to only occur when the state store lives in a different account and region from the cluster. For example, we have a shared state store in Account A in eu-central-1, and kOps 1.22.2 nodeup failed for a cluster in Account B in eu-west-1 and another in Account C in us-east-1. nodeup works fine for other clusters colocated by account with their state store, irrespective of region.

We were able to work around it with the following addition to the cluster spec:

```yaml
additionalPolicies:
  node: |
    [
      {
        "Effect": "Allow",
        "Action": ["ec2:DescribeRegions"],
        "Resource": "*"
      }
    ]
```

olemarkus:
In some cases, buildCloud will eventually end up in a DescribeRegion call, but this has been removed from nodeup. This PR adds it back in again.

Fixes #13079

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -763,6 +763,7 @@ func (b *PolicyBuilder) addNodeupPermissions(p *Policy, enableHookSupport bool) 
 	addKMSGenerateRandomPolicies(p)
 	addASLifecyclePolicies(p, enableHookSupport)
 	p.unconditionalAction.Insert(
+		"ec2:DescribeRegions",
 		"ec2:DescribeInstances", // aws.go
 		"ec2:DescribeInstanceTypes",
 	)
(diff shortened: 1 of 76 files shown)


johngmyers (approved):

johngmyers:
amazon-vpc test failed because the node's limit of 35 pods prevented an additional pod from scheduling. Perhaps the amazonvpc e2e test needs to use a larger instance type?

olemarkus:
Or enable prefix mode by default

johngmyers:
Not a bad idea. Perhaps by default for k8s >= 1.23?

PR #13121, opened 2022-01-19
Use prefix mode for VPC CNI on new clusters

olemarkus:
Some of the e2e tests require more pods than we have room for. 
I also cannot think of a reason why one wouldn't want this by default ...

This PR will only use prefix mode on new clusters. We may want to "upgrade" old clusters as well, but that can be done in a follow-up.

### DIFF
--- a/docs/releases/1.23-NOTES.md
+++ b/docs/releases/1.23-NOTES.md
@@ -12,6 +12,8 @@ This is a document to gather the release notes prior to the release.
 being used, then Kubernetes Node resources will be named after their AWS instance ID instead of their domain name and
 managed subnets will be configured to launch instances with Resource Based Names.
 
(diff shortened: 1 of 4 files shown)


johngmyers:
Any reason not to key this off of k8s version? "For new clusters" is more appropriate things that don't upgrade well. Keying off of k8s version lets people upgrade kops without taking the semantic change until they also upgrade k8s.

olemarkus:
Setting this on `kops create cluster` has a known state (i.e no state). Trying to make this change on an upgrade need to consider what people have set (e.g WARM_IP_TARGET instead of WARM_PREFIX_TARGET), so it becomes a lot more complicated. We would need to do a lot more validation of those arbitrary ENV values.

Or were you thinking of doing this on new clusters >= 1.23?

PR #13156, opened 2022-01-27
Add back container-runtime flag on kubelet as docker is the default

olemarkus:


hakman:
Where is Docker the default?

olemarkus:
It's the kubelet default value. See the flag values e.g here:

https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-k123-kolatest-to-klatest-kolatest/1486659161042194432/artifacts/34.214.123.179/kubelet.log

hakman:
1.24.0-alpha.2 will be released next week and should fix this. `remote` is already the default for `ci/latest`:
https://github.com/kubernetes/kubernetes/blob/5c9c181698c0a8944057c2d4c15ad78de8a96319/cmd/kubelet/app/options/container_runtime.go#L39

olemarkus:
Thanks for the update. I'll close then. We can live with another week of latest failing.

/close

PR #13161, opened 2022-01-30
dev: create scripts to make it easier to run e2e tests

justinsb:
Assumes an existing cluster, but then example usage:

```
./hack/run-e2e-from-source.sh aws.*Pre-provisioned.*ext4
```


hakman (approved):
/retest

PR #13164, opened 2022-01-30
Continue passing container-runtime flag in 1.24

justinsb:
Because the change is still working its way through the releases & CI,
we can continue passing the flag to avoid it defaulting to dockershim
(which is being removed).

Context: recent failures in https://testgrid.k8s.io/kops-latest#kops-grid-scenario-terraform


zetaab (approved):

hakman:
Let's wait until next week when the new alpha is released. The tests should fix themselves at that time.
/hold

hakman:
No longer needed. Tests are passing now.

Issue #12436, opened 2021-09-28
Install eks-pod-identity-webhook automatically for IRSA


h3poteto:
/kind feature

Refs: https://github.com/kubernetes/kops/issues/8264

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
At the moment, we can set up IRSA (IAM Roles for Service Accounts) using [serviceAccountIssuerDiscovery](https://github.com/kubernetes/kops/blob/master/docs/cluster_spec.md#service-account-issuer-discovery-and-aws-iam-roles-for-service-accounts-irsa) and `serviceAccountExternalPermissions`. But we still must install [eks-pod-identity-webhook](https://github.com/aws/amazon-eks-pod-identity-webhook) to inject volume and env to pods.
I hope that webhook will be installed automatically when I specify `serviceAccountIssuerDiscovery` in the cluster spec.

**2. Feel free to provide a design supporting your feature request.**
In Makefile,
https://github.com/aws/amazon-eks-pod-identity-webhook/blob/35a57cc479ae760760bfa9b5a628a488a46adad2/Makefile#L72-L88 
it set CA_BUNDLE and approve CSR when I deploy eks-pod-identity-webhook.

But probably Kubernetes custom controller can execute these processes, so I tried to implement it  on a trial basis.
https://github.com/h3poteto/eks-pod-identity-webhook-installer

I think that kops can install it automatically. There are two methods:

1. Execute the same logic in kops-controller
2. Setup CRD of [my repository](https://github.com/h3poteto/eks-pod-identity-webhook-installer) and create custom resource

What do you think about this?

olemarkus:
If you are willing to support the webhook as an addon, I support this.

Certificates can be handled by cert manager. We already use this for metrics-server and aws lbc.

The addon should not be installed when IRSA is enabled, but should require explicit config (e.g `spec.podIdentityWebhook.enabled: true`)

h3poteto:
Thank you.

I also though about using cert manager for webhook certificates, but maybe it is difficult. Because eks-pod-identity-webhook doesn't accept custom cert and key when I use in-cluster config.
https://github.com/aws/amazon-eks-pod-identity-webhook/blob/35a57cc479ae760760bfa9b5a628a488a46adad2/main.go#L153-L193

> spec.podIdentityWebhook.enabled: true

That's nice.

olemarkus:
From the looks of it, you just add the key/cert through cli flags...

h3poteto:
[...]

PR #13176, opened 2022-02-01
Add support to install EKS Pod Identity Webhook

h3poteto:
Refs #12436 

I add an option to install EKS Pod Identity Webhook. If you specify it in the cluster spec, the webhook will be installed. And it requires cert-manager to manage certificates of the webhook.
For example:
```yaml
kind: Cluster
metadata:
  name: playground
spec:
  certManager:
    enabled: true
    managed: true
  podIdentityWebhook:
    enabled: true
# ...
```


upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
```diff
@@ -0,0 +1,184 @@
...
+  - "pod-identity-webhook.default.svc.local"
+  isCA: true
+  duration: 2160h # 90d
+  renewBefore: 360h # 15d
+  issuerRef:
+    name: selfsigned
+    kind: ClusterIssuer
+---
+
+apiVersion: admissionregistration.k8s.io/v1
```
olemarkus, line 147: I am not sure if we should add an objectSelector to this one. Getting requests for every single pod in the cluster sounds like a potential issue.

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
```diff
@@ -0,0 +1,184 @@
...
+kind: MutatingWebhookConfiguration
+metadata:
+  name: pod-identity-webhook
+  namespace: default
+  annotations:
+    cert-manager.io/inject-ca-from: default/pod-identity-webhook
+webhooks:
+- name: pod-identity-webhook.amazonaws.com
+  failurePolicy: Ignore
```
olemarkus, line 156: Is this actually safe? won't this just end up in a running pod that will start misbehaving?

olemarkus (changes requested):

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
```diff
@@ -0,0 +1,184 @@
...
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - get
+  - watch
+  - list
+- apiGroups:
```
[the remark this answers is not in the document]
h3poteto: Oh, we don't need this permission.

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
h3poteto: I just copied it from the official repository: https://github.com/aws/amazon-eks-pod-identity-webhook/blob/b19c295a269ad9de50b10e3a9cdc9ec1f7d48a19/deploy/mutatingwebhook.yaml#L10
But I agree that it is not good.

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
h3poteto: Hmm... `objectSelector` is nice, but what value should we set?
The [official repository](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/deploy/mutatingwebhook.yaml) does not have any objectSelector, so it checks all pods in the cluster.
I guess should we allow users to customize the value of `objectSelector`?

olemarkus: It's a matter of documentation, I think. We can say that only pods with the label `kops.k8s.io/inject-aws-pod-identity` will be evaluated. I am not sure performance is a big concern though. Don't see much mention of it elsewhere.

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
```diff
@@ -0,0 +1,184 @@
...
+  name: pod-identity-webhook
+  namespace: default
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: pod-identity-webhook
+  namespace: default
+spec:
+  replicas: 1
```
olemarkus, line 88: We maybe want more than 1 replica?

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
h3poteto: OK, I see.

h3poteto: I update the document.
https://github.com/kubernetes/kops/pull/13176/commits/0144c1f66141c9de197ffd5d021946c6914e14f2

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
```diff
@@ -0,0 +1,173 @@
...
+  name: pod-identity-webhook
+  namespace: kube-system
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: pod-identity-webhook
+  namespace: kube-system
+spec:
+  replicas: {{ ControlPlaneControllerReplicas false }}
```
olemarkus, line 79: The webhook never runs on the control plane, so you can just set this one to 2.

olemarkus (commented):
/ok-to-test

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
```diff
@@ -0,0 +1,173 @@
+# sourced from https://github.com/aws/amazon-eks-pod-identity-webhook/tree/b19c295a269ad9de50b10e3a9cdc9ec1f7d48a19/deploy
```
olemarkus, line 1: this file should be named similar to the other addons

h3poteto: Is it a name like `k8s-1.23.yaml.template`, for example?

olemarkus: Yeah. Where the version is the minimum version supporting the manifests

upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/v0.3.0.yaml.template
```diff
@@ -0,0 +1,173 @@
...
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pod-identity-webhook
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
h3poteto: Removed: https://github.com/kubernetes/kops/pull/13176/commits/3cd20cb82a7bfbe3bf1847c3b158c27ce4e86590

olemarkus:
FYI: I am working on a feature where kops will build a configmap that the EKS hook can read SA to role mapping from. I'd like that feature to go in before we merge this.

h3poteto:
OK, I found your pull request in the amazon-eks-pod-identity-webhook repository.

olemarkus:
@h3poteto sorry inadvertently forced pushed to your branch.

https://github.com/aws/amazon-eks-pod-identity-webhook/pull/142 is taking a while, but I think we can still finish this PR. The implementation may change in that PR, but I expect the functionality/interface will be the same.

I also removed the object selector. The EKS team had not experienced any performance issues, so we do not need to introduce that burden, I think.

h3poteto:
OK, thank you.
I also removed unnecessary descriptions regarding the object selector in the doc.

olemarkus:
Are you able to test the new webhook functionality?

h3poteto:
Hmm...it does not work fine.
Because we specify `failurePolicy: Fail` and remove the object selector. So some required pods can not launch.
```
Feb 20 04:36:47 ip-172-20-37-9 kubelet[5300]: E0220 04:36:47.392591    5300 kubelet.go:1711] "Failed creating a mirror pod for" err="Internal error occurred: failed calling webhook \"pod-identity-webhook.amazonaws.com\": failed to call webhook: Post \"https://pod-identity-webhook.kube-system.svc:443/mutate?timeout=10s\": dial tcp 100.64.59.214:443: connect: connection refused" pod="kube-system/kube-controller-manager-ip-172-20-37-9.ap-northeast-1.compute.internal"
```

olemarkus:
Isn't that error transient? KCM is actually running. The wevhook cannot affect the mirror pod anyway.

h3poteto:
And the Issuer is not created even though I set `NeedsPKI: true`. :confused:

olemarkus:
There is a fun chicken/egg thing happening here:
* pod identity webhook pods won't start because it fails the webhook
* webhook fails because no pod identity webhook pods are running

We need to label pods managed by kops so that it doesn't match the webhook, I think.

h3poteto:
@olemarkus I fixed the object selector to ignore myself and the cert-manager pod. It works fine.
https://github.com/kubernetes/kops/pull/13176/commits/6e56fcc89081d7c2e74d39d2e1a69fb5ffb42a20

olemarkus:
Was that enough? I think perhaps there will be a race here ... say should the webhook be created before dns-controller pod or kops-controller, none of the worker nodes will be able to join.

olemarkus:
A test for the above-mentioned issue would be to do a `kops rolling-update cluster --yes --force --cloudonly`.

I pushed a commit now that will add a label to all pods in all kops addons, then ignore that in the webhook. Also ignoring kube-apiserver as the mirror pod is needed for dns-controller to create the API dns entries.

h3poteto:
Thank you. I agree that the `kops.k8s.io/managed-by` label is better than my change.

olemarkus:
With this branch, I could now do https://aws-controllers-k8s.github.io/community/docs/user-docs/install/
but instead of the IAM bit there, I just added this to the cluster spec:
```
    serviceAccountExternalPermissions:
    - aws:
        policyARNs:
        - arn:aws:iam::aws:policy/AmazonS3FullAccess
      name: ack-s3-controller
      namespace: ack-system
```

olemarkus:
Can you clean up the commits in this PR? And move the template file?

Then I'll get a second opinion on the changes.

olemarkus:
@h3poteto while we wait for upstream, perhaps you can add a PodDisruptionBudget and topologySpreadConstraint similar to most other addons?

h3poteto:
@olemarkus OK, I will add it.

olemarkus:
I added a commit with a few small changes. most notably not using a latest tag for the image.

I think I'd like to get this merged now even though it is not 100% ready upstream. that way I can work on some e2e tests and such to ensure this works as intended.

I can also work on extending the IRSA documentation a bit more as follow-up PRs

olemarkus (approved):

PR #13187, opened 2022-02-02
Always enable Leader Election for cloud-controller-manager

jiahuif:
... for aws & gcp cloud-controller-manager.
/kind feature

There is no significant downside. Plus, it is required for replicating the controller plane

Note that we already force leader election to be enabled for KCM.


jiahuif:
/assign @justinsb 
Could you take a look? Should be trivial.

olemarkus (approved):

PR #13188, opened 2022-02-03
[Issue-12293] Fix json output to keep it consistent for single or multiple objects

srikiz:
A minor fix, tested and works as expected.
For a single cluster you would still return an array as seen below.

```
./kops get clusters -o json | jq .
[
  {
    "kind": "Cluster",
    "apiVersion": "kops.k8s.io/v1alpha2",
    "metadata": {
      "name": "dev***ds.co.in",
      "creationTimestamp": "2022-02-01T14:17:41Z"
    },
    "spec": {
      "channel": "stable",
      "configBase": "do://sri-k***e/de****ds.co.in",
      "cloudProvider": "digitalocean",
      "kubernetesVersion": "1.22.5",
      "subnets": [
        {
          "name": "tor1",
          "zone": "tor1",
          "region": "tor1",
          "type": "Public"
        }
      ],
      "masterPublicName": "ap*****.co.in",
      "topology": {
        "masters": "public",
        "nodes": "public",
        "dns": {
          "type": "Public"
        }
      },
      "nonMasqueradeCIDR": "100.64.0.0/10",
      "sshAccess": [
        "0.0.0.0/0",
        "::/0"
      ],
      "kubernetesApiAccess": [
        "0.0.0.0/0",
        "::/0"
      ],
      "etcdClusters": [
        {
          "name": "main",
          "etcdMembers": [
            {
              "name": "etcd-1",
              "instanceGroup": "master-tor1-1"
            }
          ],
          "memoryRequest": "100Mi",
          "cpuRequest": "200m"
        },
        {
          "name": "events",
          "etcdMembers": [
            {
              "name": "etcd-1",
              "instanceGroup": "master-tor1-1"
            }
          ],
          "memoryRequest": "100Mi",
          "cpuRequest": "100m"
        }
      ],
      "kubeProxy": {
        "enabled": false
      },
      "kubelet": {
        "anonymousAuth": false
      },
      "networking": {
        "cilium": {
          "enableNodePort": true
        }
      },
      "api": {
        "dns": {}
      },
      "authorization": {
        "rbac": {}
      },
      "iam": {
        "legacy": false,
        "allowContainerRegistry": true
      }
    }
  }
]
```

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -720,7 +720,7 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 			}
 			return nil
 		case OutputJSON:
-			if err := fullOutputJSON(out, obj...); err != nil {
+			if err := fullOutputJSON(out, true, obj...); err != nil {
 				return fmt.Errorf("error writing cluster json to stdout: %v", err)
 			}
 			return nil
--- a/cmd/kops/create_instancegroup.go
+++ b/cmd/kops/create_instancegroup.go
@@ -221,7 +221,7 @@ func RunCreateInstanceGroup(ctx context.Context, f *util.Factory, out io.Writer,
 			}
 			return nil
 		case OutputJSON:
-			if err := fullOutputJSON(out, ig); err != nil {
+			if err := fullOutputJSON(out, true, ig); err != nil {
 				return fmt.Errorf("error writing cluster json to stdout: %v", err)
 			}
 			return nil
--- a/cmd/kops/get.go
+++ b/cmd/kops/get.go
@@ -158,7 +158,7 @@ func RunGet(ctx context.Context, f commandutils.Factory, out io.Writer, options 
 		return nil
 
 	case OutputJSON:
-		if err := fullOutputJSON(out, allObjects...); err != nil {
+		if err := fullOutputJSON(out, false, allObjects...); err != nil {
 			return fmt.Errorf("error writing json to stdout: %v", err)
 		}
 		return nil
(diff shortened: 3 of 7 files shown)


srikiz:
@olemarkus - please suggest if you see any issues with this. This should keep the `get clusters -o json` consistent to return an array for a single/multiple clusters. However, it also returns an array when using `kops get cluster test.example.com -ojson` as well. I fixed the test for the same, but not sure if we should expect it to return a cluster object instead of an array if a single cluster is specified. 
I still like the existing behavior, so not sure if we really need to modify this pattern here.

olemarkus:
Can you add a small mention in the 1.23 release notes (assume we can cherry-pick this there)

srikiz:
> Can you add a small mention in the 1.23 release notes (assume we can cherry-pick this there)

Okay. Should I do that after this gets merged into master, and when we cherry pick this into 1.23 release branch?

olemarkus:
The docs are served from master only, so add it to this PR please.

olemarkus:
I think you can squash the commits and then this is good.

srikiz:
@olemarkus - Fixed docs as per your suggestion. Also, realized I missed one use-case where we still need to return a single object if `kops get cluster dev.example.com -ojson` is specified. I have fixed that as well. The changes would ensure:

> `kops get clusters -o json` -- will always return a list (irrespective of whether the list contains a single or multiple clusters)
> `kops get cluster dev.example.com -ojson` -- will return a single object.
> `kops get ig -o json` -- will return based on the size of the ig.

srikiz:
> I think you can squash the commits and then this is good.

@olemarkus - Done. Please have a look when you get a chance. Thanks !

olemarkus (approved):

Issue #13171, opened 2022-01-31
kops with Spotinst feature flag: Cannot upgrade cluster when machineType value is a comma separated list


MeirP-3:
/kind bug

**kops version:** 1.21.2 (git-f86388fb1ec8872b0ca1819cf98f84d18f7263a4)
**kubernetes server version:** 1.21.5 
**Cloud provider:** AWS
**Feature Flags:** "Spotinst"=true, "SpotinstHybrid"=true

Kops allows to configure spotinst Elastic Groups when using the feature flag `Spotinst=true`.

The field `spec.machineType` can accept a comma separated list of types, for example `m5.large,m5a.large,m5.xlarge,m5a.xlarge`. Spot instances for this Elastic Group can only be instances of the listed types.

**The issue:** `kops upgrade cluster` emits errors when `machineType` is a list:
> upgrade_cluster.go:179] Error finding architecture for machine type "m5.large,m5a.large,m5.xlarge,m5a.xlarge": error finding instance info for instance type "m5.large,m5a.large,m5.xlarge,m5a.xlarge": InvalidInstanceType: The following supplied instance types do not exist: [m5.large,m5a.large,m5.xlarge,m5a.xlarge]

This problem is quite similar to [this issue](https://github.com/kubernetes/kops/issues/11513), and the solution can be similar to [it's solution](https://github.com/kubernetes/kops/pull/11560).

**A possible solution:**
In the below code, Call `cloudup.MachineArchitecture` with `strings.Split(ig.Spec.machineType, ",")[0]`, instead of `ig.Spec.machineType`. This way, even when the field is a comma separated list, only the first item will be taken.
https://github.com/kubernetes/kops/blob/f86388fb1ec8872b0ca1819cf98f84d18f7263a4/cmd/kops/upgrade_cluster.go#L176-L181

This solution relies on the current behavior of kops, which won't accept a list of mixed architecture anyway.

olemarkus:
This feature should probably be using the values from mixed instance policy instead of a comma separated list of machine types.

MeirP-3:
@olemarkus Thanks for your response.
I understand that the spot instance types feature for spotinst should use the values from mixed instance policy.
However, is it ok, while this feature uses the comma separated list and is supported with spotinst flag, to support upgrade cluster for this case as well?

olemarkus:
It's a bug that this broke, so yes. but it may be a part of the upgrade procedure to move to specifying this as mixedinstancepolicy. 

[...]

PR #13211, opened 2022-02-06
upgrade cluster: support comma separated list for machineType [1.22]

MeirP-3:
Solves #13171 for v1.22

Not needed in v1.23, since instance types for spotinst are moved to `mixedInstancesPolicy`: #12549


hakman (approved):
Thanks for this fix @MeirP-3!

PR #13212, opened 2022-02-07
Fix irsa for k8s < 1.20

olemarkus:
IRSA on 1.19 segfaults because we are not setting the feature gate that the aws model expects.


hakman:
@olemarkus Could you split this into 2 commits so that at least the main one would be easy to track in various branches?

pkg/model/components/apiserver.go
```diff
@@ -198,6 +198,12 @@ func (b *KubeAPIServerOptionsBuilder) BuildOptions(o interface{}) error {
 		}
 	}
 
+	if b.IsKubernetesLT("1.20") && clusterSpec.ServiceAccountIssuerDiscovery != nil && fi.BoolValue(&clusterSpec.ServiceAccountIssuerDiscovery.EnableAWSOIDCProvider) {
```
hakman, line 201: Do we need a minimum version here too, in case we want to cherry-pick this to other release branches?

olemarkus: I added some validation to cover this case.

pkg/apis/kops/validation/legacy.go
```diff
@@ -375,28 +375,39 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
...
+					}
+				default:
 					allErrs = append(allErrs, field.Invalid(saidStoreField, saidStore, "S3 is the only supported VFS for discoveryStore"))
 				}
-			default:
-				allErrs = append(allErrs, field.Invalid(saidStoreField, saidStore, "S3 is the only supported VFS for discoveryStore"))
 			}
 		}
-	}
+		if said.EnableAWSOIDCProvider {
```
hakman, line 400: Let's just add this a as separate block, if you don't mind.
```suggestion
		if said != nil && said.EnableAWSOIDCProvider {
```

olemarkus: The said stuff was getting too big. Moved to a dedicated function.

hakman (approved):

PR #13217, opened 2022-02-07
Use KUBETEST2_RUN_DIR for binaries if set

nckturner:
* Allow the environment variable `KUBETEST2_RUN_DIR` to be used for expected binary locations.
* If set, it is set as the `runDir` in the `execute()` method.  This is similar to what the ginkgo tester does in its `Execute()` method, which we don't use ([ref](https://github.com/kubernetes-sigs/kubetest2/blob/67b31b5cf8c6c559f6d5e6221164210912a6b65d/pkg/testers/ginkgo/ginkgo.go#L179-L182)). 
* This makes it easier to use `--use-built-binaries` in the ginkgo kubetest2 tester.
* This requires a change to kubetest2 to allow us to set `runDir` via a new `SetRunDir()` method (see https://github.com/kubernetes-sigs/kubetest2/pull/182)

### DIFF
--- a/tests/e2e/kubetest2-kops/deployer/common.go
+++ b/tests/e2e/kubetest2-kops/deployer/common.go
@@ -97,7 +97,7 @@ func (d *deployer) initialize() error {
 				d.boskos,
 				"gce-project",
 				5*time.Minute,
-				5*time.Minute,
+				10*time.Second,
 				d.boskosHeartbeatClose,
 			)
 			if err != nil {
--- a/tests/e2e/pkg/tester/kubectl.go
+++ b/tests/e2e/pkg/tester/kubectl.go
@@ -83,9 +83,14 @@ func (t *Tester) extractBinaries(downloadPath string) (string, error) {
 	}
 	tarReader := tar.NewReader(gzf)
 
-	kubectlDir, err := os.MkdirTemp("", "kubectl")
-	if err != nil {
-		return "", err
+	var kubectlDir string
+	if dir, ok := os.LookupEnv("KUBETEST2_RUN_DIR"); ok {
+		kubectlDir = dir
+	} else {
+		kubectlDir, err = os.MkdirTemp("", "kubectl")
+		if err != nil {
+			return "", err
+		}
 	}
 
 	// this is the expected path of the package inside the tar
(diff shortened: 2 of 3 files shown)


tests/e2e/kubetest2-kops/deployer/common.go
```diff
@@ -97,6 +97,7 @@ func (d *deployer) initialize() error {
 				d.boskos,
 				"gce-project",
 				5*time.Minute,
+				10*time.Second,
```
rifelpet, line 100: Can you update the kubetest2 dependency in tests/e2e/go.mod and then run `make gomod` to pick up the update that matches this change?

nckturner:
Strangely I'm getting a "security error" when running make gomod due to a checksum mismatch:
```
verifying github.com/bazelbuild/bazelisk@v1.10.1: checksum mismatch
	downloaded: h1:NOx8p+vDJ24UiQ1kG8zMqRLdi/sOBgqld3mXMKAp5HY=
	go.sum:     h1:2EWA2lRrt/k8B5ASt0mlTDQ+7mnzvKdF6ShNWLbk0o0=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
Makefile:351: recipe for target 'gomod' failed
make: *** [gomod] Error 1

```

Haven't looked into it yet.

rifelpet:
I'm not sure what the hash mismatch is about but I forgot youll also need to run `go mod clean` to keep the `verify-gomod` prow job happy

olemarkus:
I bumped kubetest2 in kops earlier today. If you rebase, you may find the tests passing.

tests/e2e/kubetest2-kops/deployer/common.go
olemarkus: Should this change be here?

tests/e2e/pkg/tester/kubectl.go
```diff
@@ -83,9 +83,14 @@ func (t *Tester) extractBinaries(downloadPath string) (string, error) {
 	}
 	tarReader := tar.NewReader(gzf)
 
-	kubectlDir, err := os.MkdirTemp("", "kubectl")
-	if err != nil {
-		return "", err
+	var kubectlDir string
+	if dir, ok := os.LookupEnv("KUBETEST2_RUN_DIR"); ok {
```
olemarkus, line 87: Annoying that rundir is entirely private to the ginkgo tester (which we compose ours of). With this, there will be different behavior if `--rundir` is set vs this variable. Maybe we should shadow `rundir` and do a similar init.

PR #13218, opened 2022-02-07
fix KCM LogLevel setting not honored

jiahuif:
The LogLevel was forced to be 2, which can be painful for troubleshooting. Changed to make it default to 2 but the value will be honored if set.

Downside is that there is no way to disable logging by setting `-v=0`. However muting logging should be rarely used.


justinsb:
Thanks!  This does mean it's not possible to set LogLevel to 0, but in order to fix that we'd have to switch LogLevel to a pointer (which we can do, but it's much more involved).  Being able to set it to "anything other than 0" is a big improvement :-)

/approve
/lgtm

PR #13219, opened 2022-02-07
CCM: use flagbuilder instead of manually building argv

jiahuif:
Because CCM did not use flagbuilder, flags defined in the configuration were not applied. Rewrite the argv building to use standard approach.  Also reorder flags in tests, and add tests for other flags.

/kind feature

PR #13220, opened 2022-02-07
always enable Leader Election for openstack CCM

jiahuif:
Leader Election is a must-have for replicated/HA control plane.

Kind of scary to think of current HA clusters running without leader election.

Needs https://github.com/kubernetes/kops/pull/13219 to be merged first.


hakman (approved):
Thanks @jiahuif!

PR #13228, opened 2022-02-09
Bump AWS CNI to 1.10.2

MoShitrit:
Version 1.10.2 was released 4 days ago - https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.10.2

### DIFF
--- a/tests/integration/update_cluster/many-addons-ccm-irsa/data/aws_s3_bucket_object_minimal.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/many-addons-ccm-irsa/data/aws_s3_bucket_object_minimal.example.com-addons-bootstrap_content
@@ -118,7 +118,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.16
     manifest: networking.amazon-vpc-routed-eni/k8s-1.16.yaml
-    manifestHash: 302d76bd06cc968ba03481ada5a894aa120c7687d7d41a02dd93f83c7a6cbb9f
+    manifestHash: 6f4d332e277fc7902448a11cca2691075541f871b6d3f5a78c3ee1c58d71911e
     name: networking.amazon-vpc-routed-eni
     needsRollingUpdate: all
     selector:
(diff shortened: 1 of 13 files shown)


hakman:
@MoShitrit There were some changes to the manifest since 1.9. Could you also update the template?

MoShitrit:
> @MoShitrit There were some changes to the manifest since 1.9. Could you also update the template?

Hey @hakman !
So the only thing I could see in their [upgrade PR](https://github.com/aws/amazon-vpc-cni-k8s/pull/1850) is the addition of one ENV:
image

However, since we now don't have these explicitly set, I wasn't sure if I should add that or not. I can probably add that and comment it out like the other ENVs there. Does that work?
Also- is there anything else I might've missed besides that?
Thanks!

hakman:
Hey @MoShitrit! :)
I think the changes in these commit are missing:
https://github.com/aws/amazon-vpc-cni-k8s/commit/01654383404fe88f18a89508d9406ad52c6eec1f
https://github.com/aws/amazon-vpc-cni-k8s/commit/cf017e525abcc038341929431c91c8767f32c672
https://github.com/aws/amazon-vpc-cni-k8s/commit/fe817b84a7d00a0037d7feefbea20f90a6804eae

Probably the easiest way to check would be to compare the template with https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml.

MoShitrit:
> Probably the easiest way to check would be to compare the template with https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml.

Good point! 
Let me do that and make the necessary update.
Thanks 😄

MoShitrit:
@hakman I took your suggestion and compared our template with the one from `aws-cni`'s `master` branch. There were a couple of cosmetic changes the were causing noise, so I moved things around a bit to reduce that noise.
Then I also noticed the additional changes you've mentioned. At this point, I think we should be good, but feel free to verify my sanity.
Thanks again! 😄

hakman:
Thanks @MoShitrit. Looks pretty good. Will do the final review during the weekend.

hakman (changes requested):
Looks great, just a few nits left.

hakman (approved):
Excellent! Thanks @MoShitrit. :)

PR #13234, opened 2022-02-10
Spotinst: Ocean as a template for VNGs

liranp:
This PR allows users to configure Ocean as a template by using the `SpotinstOceanTemplate` feature flag. In this mode, the launch specification defined on the Ocean object acts as a template only for Ocean Virtual Node Groups.


olemarkus:
Curious if we actually need all the spotinst feature flags. At some point these should be removed.

hakman (approved):

PR #13251, opened 2022-02-14
E2E HA Upgrade/Rollback for Leader Migration

jiahuif:
This PR adds a new scenario that
- create a 1.23 cluster with 3 control plane nodes
- upgrade and rolling-upgrade the control plane to 1.24, performing Leader Migration between 1.23 KCM to 1.24 CCM
- rollback and rolling-upgrade the control plane back to 1.23, performing Leader Migration between 1.23 KCM to 1.24 CCM

This PR implements a controller-to-manager recorder with `client-go` that runs in the cluster, because connections from the 
test host to the cluster may be interrupted during the rolling-upgrade. Also, this is a good way to ensure the data plane is not disturbed during the control plane upgrade.

This test also uses (and thus covers) per-controller health-check implemented as https://github.com/kubernetes/kubernetes/pull/104667

This PR is designed to run as a periodic job in GCE. Other providers support TODO.

Current status: passing, using my own GCP project.

### DIFF
--- a/tests/e2e/go.mod
+++ b/tests/e2e/go.mod
@@ -8,7 +8,9 @@ require (
 	github.com/octago/sflags v0.2.0
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/yaml.v2 v2.4.0
+	k8s.io/api v0.23.1
 	k8s.io/apimachinery v0.23.1
+	k8s.io/client-go v9.0.0+incompatible
 	k8s.io/klog/v2 v2.30.0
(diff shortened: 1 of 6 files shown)


jiahuif:
/assign @justinsb @hakman 
Could you take a look when available?

justinsb (approved):

justinsb:
This is an awesome contribution; these techniques are a big leap on our testing and I imagine we'll use this as the basis for a lot more tests like this - thank you!

/approve
/lgtm

PR #13252, opened 2022-02-14
CCM: allow setting Controllers for cloudControllerManagerConfig

jiahuif:
This PR enables the `Controllers` field for CCM, just like KCM.

Currently, it is possible to set the `--controllers` flag via the config for kube-controller-manager, but not `cloud-controller-manager`. However, being able to enable/disable some controllers for CCM is as important as that for KCM, if not more so. This is because for the same cloud provider, the cloud environment may have significant variation, for example, hosted Kubernetes, self-deployed Kubernetes, third-party hosted Kubernetes for the same cloud. The added flexibility of CCM config would help such use cases.

### DIFF
--- a/pkg/apis/kops/componentconfig.go
+++ b/pkg/apis/kops/componentconfig.go
@@ -681,6 +681,8 @@ type CloudControllerManagerConfig struct {
 	AllocateNodeCIDRs *bool `json:"allocateNodeCIDRs,omitempty" flag:"allocate-node-cidrs"`
 	// ConfigureCloudRoutes enables CIDRs allocated with to be configured on the cloud provider.
 	ConfigureCloudRoutes *bool `json:"configureCloudRoutes,omitempty" flag:"configure-cloud-routes"`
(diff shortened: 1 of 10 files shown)


pkg/apis/kops/componentconfig.go
```diff
@@ -681,6 +681,8 @@ type CloudControllerManagerConfig struct {
 	AllocateNodeCIDRs *bool `json:"allocateNodeCIDRs,omitempty" flag:"allocate-node-cidrs"`
 	// ConfigureCloudRoutes enables CIDRs allocated with to be configured on the cloud provider.
 	ConfigureCloudRoutes *bool `json:"configureCloudRoutes,omitempty" flag:"configure-cloud-routes"`
+	// Controllers is a list of controllers to enable on the controller-manager
+	Controllers []string `json:"controllers,omitempty" flag:"controllers"`
```
jiahuif, line 685: Note: copied from that of KubeControllerManagerConfig

hakman (approved):

PR #13253, opened 2022-02-14
CCM: add livenessProbe for GCP CCM

jiahuif:
Our CCM addons are missing `livenessProbe`. This PR adds the probe to GCP CCM. Other cloud providers should do the same.


upup/models/cloudup/resources/addons/gcp-cloud-controller.addons.k8s.io/k8s-1.23.yaml.template
```diff
@@ -41,6 +41,17 @@ spec:
         env:
         - name: KUBERNETES_SERVICE_HOST
           value: "127.0.0.1"
+        livenessProbe:
```
jiahuif, line 44: Config copied from our KCM manifest except port. Should be okay to use the same setting because the health check mechanism is shared among controller managers.

PR #13279, opened 2022-02-17
Release 1.22.4

heybronson:


justinsb:
Thanks for doing this.  We do want a release, but my understanding was that we needed to pick up the new etcd-manager with 3.5.1.  I that cherry-pick in https://github.com/kubernetes/kops/pull/13291

If I merge this, it should go on top of the branch (good).  But I think the SHAs in the manifests will no longer be correct.  It's worth verifying though, and I think I can also maybe send a commit to the PR to update the SHAs.

/retest

justinsb:
The tests are passing, and I verified it locally also after rebasing.   Cool!

/approve
/lgtm

PR #13280, opened 2022-02-17
Add missing permissions to aws lbc for irsa

### LINKED ISSUE #13020: AWS LBC cannot delete Security Groups it creates
/kind bug

 this information.**

master branch. Broken code appears to be in release-1.23 branch as well

 a `kops` flag.**

1.21.7

AWS

Examine the kops-aws-aws-load-balancer-controller e2e test.

Alternatively:
1. Create a cluster with the AWS Load Balancer controller.
2. Create an Ingress backed by an ALB.
3. Delete the ingress

The ALB controller is unable to delete the "k8s-traffic-" ingress for the ALB due to lack of IAM permissions. This is because the security group is missing the `KubernetesCluster` tag.

Security group is tagged with the tags that kOps passed AWS LBC through the `default-tags` flag. LBC is able to delete the security groups it creates.

```yaml
```

Introduced by #12920.

Filed upstream as kubernetes-sigs/aws-load-balancer-controller#2424

olemarkus:
Looks like we are missing a lot of permissions for LBC. 
fixes #13020

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -953,28 +953,61 @@ func AddCCMPermissions(p *Policy, cloudRoutes bool) {
 // AddAWSLoadbalancerControllerPermissions adds the permissions needed for the aws load balancer controller to the givnen policy
 func AddAWSLoadbalancerControllerPermissions(p *Policy) {
 	p.unconditionalAction.Insert(
-		"ec2:DescribeAvailabilityZones",
-		"ec2:DescribeNetworkInterfaces",
-		"elasticloadbalancing:DescribeTags",
-		"elasticloadbalancing:DescribeTargetGroupAttributes",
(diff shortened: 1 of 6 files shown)


hakman (approved):

olemarkus:
Not sure how we managed to get this test passing before, but the test tries to use create an LB with the cluster name in it. This name contains `.`, which is not allowed by AWS.

hakman:
> Not sure how we managed to get this test passing before, but the test tries to use create an LB with the cluster name in it. This name contains `.`, which is not allowed by AWS.

Seems like a test bug and should be fixed in k/k. Should we skip it for now?

olemarkus:
The tests are from lbc repo, but yeah. we I'll do rebase and push, and then we skip trying to run that test. I'll see if I can fix it in a follow-up.

PR #13283, opened 2022-02-18
test: use `T.TempDir` to create temporary test directory

Juneezee:
A small testing enhancement here.

We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete.

Reference: https://pkg.go.dev/testing#T.TempDir


hakman (approved):
Thanks @Juneezee.

PR #13289, opened 2022-02-18
Only delete node object on GCE

### LINKED ISSUE #13259: Kubelet recreating node lease after node object has been deleted
During rolling upgrades the following happens:
* Drain node
* Terminate instance
* Delete node object
* Node lease object deleted as part of the cascade since the lease has the node object as owner

After deleting the node object, and the lease with it, there is a high chance that kubelet will recreate the lease object. Especially after enabling graceful shutdowns. Since the node object is gone, it will not set the ownerRef on the lease, which is causing our E2E to fail. This behavior is somewhat questionable as kubelet is aware it is being shut down, and the lease doesn't make that much sense without the node object in the first place.

Consequence is anyways that lease objects are being created and never cleaned up.

There are a couple of ways this could be solved:
* Change behavior in kubelet
* cloudup or kops-controller cleans up lease objects of terminated instances
* We stop deleting node objects, and just let CCM clean them up. In general this is not a problem as CCM will clean up nodes faster than new nodes are being provisioned. There is a comment on GCP will recycle node names, but I am not sure if we then either make an exception for GCP or if we just leave it be.

I want to do the latter. Let CCM discover instances are terminated and delete the nodes.

/kind bug
/kind office-hours

olemarkus:
I don't see why the GCE CCM shouldn't have deleted the node object before the new kublet registers given that the NodeLifecycle controller runs every 40 seconds ... But until we can test, we keep this for GCE.

/cc @justinsb @hakman @johngmyers 

fixes #13259

### DIFF
--- a/pkg/instancegroups/instancegroups.go
+++ b/pkg/instancegroups/instancegroups.go
@@ -409,9 +409,8 @@ func (c *RollingUpdateCluster) drainTerminateAndWait(u *cloudinstances.CloudInst
 		}
 	}
 
-	// We unregister the node before deleting it; if the replacement comes up with the same name it would otherwise still be cordoned
-	// (It often seems like GCE tries to re-use names)
-	if !isBastion && !c.CloudOnly {
+	// GCE often re-uses names, so we delete the node object to prevent the new instance from using the cordoned Node object
+	if c.Cluster.Spec.GetCloudProvider() == api.CloudProviderGCE && !isBastion && !c.CloudOnly {
 		if u.Node == nil {
 			klog.Warningf("no kubernetes Node associated with %s, skipping node deletion", instanceID)
 		} else {
--- a/pkg/instancegroups/rollingupdate_test.go
+++ b/pkg/instancegroups/rollingupdate_test.go
@@ -222,17 +222,14 @@ func TestRollingUpdateAllNeedUpdate(t *testing.T) {
 		case testingclient.PatchAction:
 			if string(a.GetPatch()) == cordonPatch {
 				assertCordon(t, a)
-				assert.Equal(t, "", cordoned, "at most one node cordoned at a time")
 				assert.True(t, tainted[a.GetName()], "node", a.GetName(), "tainted")
 				cordoned = a.GetName()
 			} else if string(a.GetPatch()) == excludeLBPatch {
 				assertExclude(t, a)
-				assert.Equal(t, "", excluded, "at most one node excluded at a time")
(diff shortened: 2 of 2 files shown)


pkg/instancegroups/rollingupdate_test.go
```diff
@@ -222,17 +222,14 @@ func TestRollingUpdateAllNeedUpdate(t *testing.T) {
 		case testingclient.PatchAction:
 			if string(a.GetPatch()) == cordonPatch {
 				assertCordon(t, a)
-				assert.Equal(t, "", cordoned, "at most one node cordoned at a time")
```
johngmyers, line 225: I'm wondering if it would be worthwhile to hook into the `TerminateInstances()` mock in order to keep all of the `testingclient.DeleteAction` assertions.

olemarkus: I thought about it, but it _shouldn't_ actually delete any nodes, so seems strange to test for it.
We are also checking that the ASGs are empty at the end, so we know the terminations take place.

pkg/instancegroups/instancegroups.go
```diff
@@ -409,9 +410,8 @@ func (c *RollingUpdateCluster) drainTerminateAndWait(u *cloudinstances.CloudInst
 		}
 	}
 
-	// We unregister the node before deleting it; if the replacement comes up with the same name it would otherwise still be cordoned
-	// (It often seems like GCE tries to re-use names)
-	if !isBastion && !c.CloudOnly {
+	// GCE often re-uses names, so we delete the node object to prevent the new instance from using the cordoned Node object
+	if kops.CloudProviderID(c.Cluster.Spec.CloudProvider) == kops.CloudProviderGCE && !isBastion && !c.CloudOnly {
```
hakman, line 414:
```suggestion
	if api.CloudProviderID(c.Cluster.Spec.CloudProvider) == kops.CloudProviderGCE && !isBastion && !c.CloudOnly {
```

pkg/instancegroups/instancegroups.go
```diff
@@ -36,6 +36,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/drain"
 
+	"k8s.io/kops/pkg/apis/kops"
```
hakman, line 39: Already declared below.
```suggestion
```

hakman (changes requested):

hakman (approved):

PR #13308, opened 2022-03-01
Append policy config map arguments only if UsePolicyConfigmap is true

vivekjainx86:
Existing condition appends the policy config map arguments even if the UsePolicyConfigMap is set to false.

### DIFF
--- a/nodeup/pkg/model/kube_scheduler.go
+++ b/nodeup/pkg/model/kube_scheduler.go
@@ -192,7 +192,7 @@ func (b *KubeSchedulerBuilder) buildPod(kubeScheduler *kops.KubeSchedulerConfig)
 		flags = append(flags, "--"+flag+"kubeconfig="+defaultKubeConfig)
 	}
 
-	if kubeScheduler.UsePolicyConfigMap != nil {
+	if kubeScheduler.UsePolicyConfigMap != nil && *kubeScheduler.UsePolicyConfigMap {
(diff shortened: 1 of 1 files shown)


nodeup/pkg/model/kube_scheduler.go
```diff
@@ -192,7 +192,7 @@ func (b *KubeSchedulerBuilder) buildPod(kubeScheduler *kops.KubeSchedulerConfig)
 		flags = append(flags, "--"+flag+"kubeconfig="+defaultKubeConfig)
 	}
 
-	if kubeScheduler.UsePolicyConfigMap != nil {
+	if kubeScheduler.UsePolicyConfigMap != nil && *kubeScheduler.UsePolicyConfigMap {
```
hakman, line 195: fi.BoolValue would be better in this case:
https://github.com/kubernetes/kops/blob/97869057129b30ea284c3ed1bdf1db36e752701d/pkg/values/values.go#L48-L53
```suggestion
	if fi.BoolValue(kubeScheduler.UsePolicyConfigMap) {
```

hakman (changes requested):

nodeup/pkg/model/kube_scheduler.go
vivekjainx86: ohh.. thats great! This is better.

hakman:
Thanks @vivekjainx86. Looks good to me. Could you also squash the commits?

hakman (approved):

PR #13311, opened 2022-03-01
use own function to define CSI image version

zetaab:
this PR fixes 2 bugs:

1) there is typo in CCM tag, it should be `v1.23.1` not `1.23.1`
2) currently kOps assumes that CSI uses same version than CCM uses in OpenStack. However, that is not true currently.


zetaab:
this needs cherry-pick to 1.23

olemarkus (approved):

Issue #2214, opened 2017-03-28
Request to allow specifying extra routes to add to kops subnets to point to pre-existing VPN gateway


keiths-osc:
We have a pre-existing VPC in AWS with a VPN connection between our company's network and the VPC.  I have been able to successfully create a completely private Kubernetes cluster with kops by ensuring that I:

- set the api load balancer type to `Internal` in the kops cluster configuration
- set the dns, masters, and nodes topology to `private` in the kops cluster configuration (via flags to `kops create cluster`)
- do not pass the `--bastion` flag to `kops create cluster` so a bastion is not created

 I found in the [run_in_existing_vpc document](https://github.com/kubernetes/kops/blob/master/docs/run_in_existing_vpc.md) that I can use pre-existing subnets in the cluster configuration.  I gave this a try.  The route tables for the subnets I made have routes that point to our VPN gateway.  The kops utility subnets share a route table that points to an Internet gateway for the default route, and the private subnets each have a route table that points to a NAT gateway for the default route.  This all appears to have worked well.  I have been able to create `LoadBalancer` type Kubernetes services configured so that they create private load balancers, and I can access the services via the VPN.  (Note I did make sure to add the subnet tags mentioned by [this issue](https://github.com/kubernetes/kops/issues/2011) to ensure the load balancer ends up in one of the private subnets).

Rather than creating the Kubernetes resources in our pre-existing subnets, we would prefer to keep the kops cluster separated in its own dedicated subnets that are ideally managed by kops.  It would be nice if kops could create the subnets and their corresponding route tables with the VPN gateway routes added.  The [run_in_existing_vpc document](https://github.com/kubernetes/kops/blob/master/docs/run_in_existing_vpc.md) shows how to have kops create the subnets for you but use pre-existing NAT gateways by specifying an `egress` setting on the private subnets.  I gave this a try, and kops created all the subnets for me in our existing VPC and pointed the route tables to the NAT gateways I specified.  I then manually added the VPN gateway routes to both the utility and private subnet route tables that kops created (and the tags mentioned in [this issue](https://github.com/kubernetes/kops/issues/2011)) and found that all seems to work fine.  Without the VPN gateway routes, I can't communicate with the masters or nodes over the VPN because the networking on the AWS side doesn't know how to route the response back to me.

[...]

PR #13318, opened 2022-03-03
Possibility to add additional routes in route tables of subnets

guillomep:
This PR fixes the issue https://github.com/kubernetes/kops/issues/2214.


k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -5369,6 +5369,19 @@ spec:
...
                   properties:
+                    additionalRoutes:
+                      description: AdditionalRoutes to attach to the subnet's route
+                        table
+                      items:
+                        properties:
+                          cidr:
+                            description: CIDR destination of the route
+                            type: string
+                          target:
```
rifelpet, line 5380: Nothing stands out in this PR, but we will want to keep in mind future support for prefix lists instead of CIDRs for routes. We already support them in security group rules (https://github.com/kubernetes/kops/pull/13113). Essentially it would make the `cidr` field optional if a prefix list field is set.

pkg/model/awsmodel/network.go
```diff
@@ -564,6 +568,22 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+
+			subnets, err := b.LinkToPrivateSubnetInZone(zone)
+			if err != nil {
+				return err
+			}
+
+			for _, subnetSpec := range b.Cluster.Spec.Subnets {
+				for _, subnet := range subnets {
+					if strings.HasPrefix(*subnet.Name, subnetSpec.Name) {
+						err := addAdditionalRoutes(subnetSpec.AdditionalRoutes, subnetSpec.Name, rt, b.Lifecycle, c)
```
rifelpet, line 580: we're adding these routes to private subnets but `addAdditionalRoutes` includes `public` in the task names, perhaps we should leave out `public` if there are otherwise no naming collisions?

pkg/apis/kops/validation/aws.go
```diff
@@ -48,6 +48,13 @@ func awsValidateCluster(c *kops.Cluster) field.ErrorList {
 		allErrs = append(allErrs, awsValidateIAMAuthenticator(field.NewPath("spec", "authentication", "aws"), c.Spec.Authentication.AWS)...)
 	}
 
+	for i, subnet := range c.Spec.Subnets {
```
rifelpet, line 51: I think we should disallow AdditionalRoutes if the subnet is shared (if the ProviderID field is set)

https://github.com/kubernetes/kops/blob/b2bb322a577ec901de77dbd36e5d6ff12f48d628/pkg/apis/kops/cluster.go#L717-L718

rifelpet (commented):
Can we also add some docs for this? 

https://kops.sigs.k8s.io/cluster_spec/#clusterspec-subnet-keys

k8s/crds/kops.k8s.io_clusters.yaml
guillomep: If you think support prefix list is important for the future I can add it to this PR.
Maybe we can replace `CIDR` with `Destination` instead of having two fields:

```yaml
additionalRoutes:
  description: AdditionalRoutes to attach to the subnet's route table
  items:
    properties:
      destination:
        description: CIDR destination or prefix list of the route
        type: string
      target:
        description: Target of the route
        type: string
      type: object
    type: array
```
Then the validation check if it is a valid CIDR or prefix list

pkg/apis/kops/validation/aws.go
guillomep: I add the check and fail the validation if the `ProviderId` is not empty

guillomep:
Added documentation

docs/cluster_spec.md
```diff
@@ -351,6 +351,25 @@ spec:
     zone: us-east-1a
 ```
 
+### additionalRoutes
+
+{{ kops_feature_table(kops_added_default='1.24') }}
```
guillomep, line 356: Need to update that if kops 1.24 is released before this PR is merged

pkg/model/awsmodel/network.go
```diff
@@ -307,6 +307,10 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
 					infoByZone[subnetSpec.Zone].NATSubnets = append(infoByZone[subnetSpec.Zone].NATSubnets, subnetSpec)
 					infoByZone[subnetSpec.Zone].HaveIPv6PublicSubnet = true
 				} else {
+					err := addAdditionalRoutes(subnetSpec.AdditionalRoutes, subnetName, publicRouteTable, b.Lifecycle, c)
+					if err != nil {
+						return err
+					}
```
rifelpet, line 313: Since we only create one RouteTable for all public and utility subnets, attempting to add routes to the public route table from each public/utility SubnetSpec's AdditionalRoutes may result in collisions. For now I think we should consider adding API validation to ensure AdditionalRoutes is not used on public or utility subnets until we come up with a better solution for them (in a separate PR)


@olemarkus any additional concerns here for ipv6/dualstack?

pkg/model/awsmodel/network.go
rifelpet: In every case but as i alluded to above, i think we shouldn't support public subnets in this initial PR and try to add support for them in a followup PR due to the added complexity.

pkg/model/awsmodel/network.go
```diff
@@ -615,3 +635,104 @@ func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+	for _, r := range routes {
+		if strings.HasPrefix(r.Target, "pcx-") {
+			t := &awstasks.Route{
+				Name:                 fi.String("public-" + sbName + "." + r.CIDR),
+				Lifecycle:            lf,
+				CIDR:                 fi.String(r.CIDR),
+				RouteTable:           rt,
+				VPCPeeringConnection: fi.String(r.Target),
+			}
+			c.AddTask(t)
```
rifelpet, line 649: nit: we could avoid some duplication without sacrificing readability if we define one route task and then conditionally set the appropriate field based on the target
```suggestion
	for _, r := range routes {
		t := &awstasks.Route{
			Name:       fi.String("public-" + sbName + "." + r.CIDR),
			Lifecycle:  lf,
			CIDR:       fi.String(r.CIDR),
			RouteTable: rt,
		}

		if strings.HasPrefix(r.Target, "pcx-") {
			t.VPCPeeringConnection = fi.String(r.Target)
			c.AddTask(t)
```

upup/pkg/fi/cloudup/awstasks/route.go
```diff
@@ -45,6 +45,7 @@ type Route struct {
 	InternetGateway           *InternetGateway
 	NatGateway                *NatGateway
 	TransitGatewayID          *string
+	VPCPeeringConnection      *string
```
rifelpet, line 48:
```suggestion
	VPCPeeringConnectionID      *string
```

and all of the other references to this field

pkg/model/awsmodel/network.go
guillomep: I have removed support for Public and utility subnets.

There is still a question about the DualStack

pkg/model/awsmodel/network.go
olemarkus: I think we can deal with those later should any appear.

olemarkus (approved):

PR #13330, opened 2022-03-04
Support GPU in OpenStack

### LINKED ISSUE #12750: Container runtime configuration should be per instanceGroup
/kind feature

There are a number of use cases that require a different configuration for the container runtime per instance group. Currently it is only possible configure containerd or docker in the cluster_spec. This can cause issues especially for masters which have extremely different requirements for the runtime. As an example, GPU instances currently are supported but can only be configured/enabled cluster-wide.

I'd like to propose moving the container runtime configuration into the InstanceGroup spec.

zetaab:
fixes #12750

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -70,6 +70,7 @@ import (
 	"k8s.io/kops/util/pkg/architectures"
 	"k8s.io/kops/util/pkg/hashing"
 	"k8s.io/kops/util/pkg/mirrors"
+	"k8s.io/kops/util/pkg/reflectutils"
 	"k8s.io/kops/util/pkg/vfs"
 )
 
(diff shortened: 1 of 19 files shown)


zetaab:
@olemarkus as you did that AWS support, could you check this? With this PR I can get GPUs visible in OpenStack and everything looks good.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1397,14 +1398,26 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+	if ig.Spec.Containerd != nil && ig.Spec.Containerd.NvidiaGPU != nil {
+		if config.NvidiaGPU == nil {
+			config.NvidiaGPU = ig.Spec.Containerd.NvidiaGPU
+		} else {
+			reflectutils.JSONMergeStruct(&config.NvidiaGPU, ig.Spec.Containerd.NvidiaGPU)
+		}
+	}
+
+	if config.NvidiaGPU != nil && config.NvidiaGPU.DriverPackage == "" {
+		config.NvidiaGPU.DriverPackage = "nvidia-headless-460-server"
```
zetaab, line 1418: we should perhaps store this const to same common place. This is now defined in two files as hardcoded, which would be the correct place to store this information? there are wellknownports etc should we add something new?

olemarkus: It could certainly be moved to a const. Perhaps in the API package.

Here we are doing the merging in two locations as well. Didn't it work to move this to a function similar to how we merge workpool?

olemarkus (approved):

PR #13337, opened 2022-03-05
Fix introduced version for pod-identity-webhook in docs

h3poteto:
Recently, I added EKS pod identity webhook addon: https://github.com/kubernetes/kops/pull/13176 . In that PR, I updated addons docs.
And that PR was backported to 1.23: https://github.com/kubernetes/kops/pull/13315
That means we can use the addon, since kOps 1.23. So I fix the docs.

### DIFF
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -249,7 +249,7 @@ spec:
 
 #### EKS Pod Identity Webhook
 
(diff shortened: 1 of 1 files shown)


olemarkus:
Also a part of #13338

h3poteto:
Wow, it's the same timing :open_mouth:

PR #13354, opened 2022-03-10
Don't hash parts of the cluster name for e2e

olemarkus:


justinsb:
Looks good, but can we add a comment (either to the commit or the code itself) about why we're not hashing.  It does feel like we were trying to fit into a size limit before, but maybe we don't have that limit any more?

olemarkus:
Added to commit

justinsb:
Thanks @olemarkus 

/approve
/lgtm

PR #13368, opened 2022-03-16
Parallelize the way we read files from VFS

toninis:
In our deployments we have more than 150 instance groups. 
Reading all those states takes enormous amount of time . 
This is a short implementation to parallelize the work. 

**PS**
I am working on a way to also parallelize the [getCloudGroups](https://github.com/kubernetes/kops/blob/dd46a82065a7d9c341b71a19c5d374f6c6cfb526/upup/pkg/fi/cloudup/awsup/aws_cloud.go#L911) function without hitting the AWS API limits.

### DIFF
--- a/pkg/client/simple/vfsclientset/commonvfs.go
+++ b/pkg/client/simple/vfsclientset/commonvfs.go
@@ -23,8 +23,10 @@ import (
 	"os"
 	"reflect"
 	"sort"
+	"sync"
 	"time"
 
+	"golang.org/x/sync/errgroup"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


toninis:
@olemarkus  Test fail on `//cmd/kops:go_default_test` . I am not sure how my change affects those tests . There is something else going on that I am not able to figure out .

toninis:
> it looks like the order of resources returned from VFS is now nondeterministic because it happens in parallel. Fixing this may be as simple as sorting the results which may have a one-time impact on the ordering of the test outputs and you'll need to run `./hack/update-expected.sh` to update them all.

I Tried that and it did not work. But you are right. The expected manifests are incosistent . I will dig up a bit and try to mitigate this . 
Thanks 😄

PR #13373, opened 2022-03-16
If kubetest2 fails cluster validation, we run down before exiting

olemarkus:


rifelpet (approved):
We probably want it if `d.updateCluster(true)` returns an error as well, a few lines up. We can start with this though.

PR #13378, opened 2022-03-17
Do not return a '-1' exit if no keys found and json/yaml output

hierynomus:
In our CI scripts we wanted to adapt to the new `sshpublickey` command, however for a new cluster this will exit non-successfully if there are no keys yet. It is possible to work around this with some ugly hacks, however for scripts I think we should return successfully if there are no public keys. It's not like something bad happened.

This PR ensures that for json/yaml output, which is for machine consumption, that even for an empty list it will exit with a `0` exit code if no errors happen.


hakman (approved):
Seems like a good thing to me. Thanks @hierynomus.

PR #13382, opened 2022-03-17
[Digital Ocean] Remove sfo2 region from the list of supported DO regions

srikiz:
Saw e2e tests failing for SFO2 region - https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-do-calico/1504078673064497152

Per discussion with @timoreimann this is expected and customers should use the new `SFO3` region going forward. We currently have `SFO3` already in the list of supported regions.


srikiz:
FYI - @timoreimann

hakman (approved):
/retest

timoreimann (commented):
A few more regions are restricted to existing DO customers only. We should exclude those as well.

Full list of regions we should not use:

- nyc2
- sfo1
- sfo2
- ams2

(You can also see this list by logging into the DO cloud control panel UI and see which regions are created out on the droplet create page.)

PR #13390, opened 2022-03-18
Clean up kops build targets

olemarkus:


Makefile
```diff
@@ -184,34 +177,40 @@ test:
 test-windows:
 	go test -v $(go list ./... | grep -v /nodeup/)
 
+
+.PHONY: kops
+kops: crossbuild-$(shell go env GOOS)-$(shell go env GOARCH)
```
hakman, line 182: Maybe we should be more explicit and use `crossbuild-kops-...`, or just drop `crossbuild-` completely?

olemarkus: I thought about it, but didn't want to change the naming of targets in this PR. But it is something we should look at.

hakman: I think the only one that already exists is `crossbuild`, all the rest are new.

hakman (approved):

PR #13396, opened 2022-03-21
Add possibility to set PodAnnotations into NodeLocalDNS

anthonyhaussman:
Annotations is pretty useful when you need third-party tool to add additional behavior for a k8s resource.
Lots of auto-discovery tools are based on this annotations.

This PR add the possibility to set tune pod annotations in the node-local-dns daemonsets.app.


justinsb:
/ok-to-test

Thanks for doing this @anthonyhaussman 

This seems reasonable to me.  I do want to get addon-operators going, but this seems like a good feature and one we would likely want to support in the addon operators also.

We have some "prior art" in that cilium allows for `agentPodAnnotations` (also `map[string]string`).  I think this feature does make sense therefore.

I'm going to LGTM but not approve yet, to give others time to chime-in.  But if no objections I'm inclined to approve also :-)

/lgtm

PR #13400, opened 2022-03-22
Fix some flags in NTH Queue Process mode

hippolin:
This PR is to fix some flags in NTH Queue Process mode.


olemarkus:
Are these flags applicable to queue processing mode?
Looking at https://github.com/aws/aws-node-termination-handler/blob/main/config/helm/aws-node-termination-handler/templates/deployment.yaml they are not configurable there either.

cc @bwagner5

hippolin:
I trace the source code of the AWS node termination handler, too. Queue processing mode doesn't need these flags. So I close this PR.  Thank you. @olemarkus

Issue #13365, opened 2022-03-16
migrate to registry.k8s.io


ameukam:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

Migrate future releases of kOps to `registry.k8s.io` which is a replacement of `k8s.gcr.io` (and currently for the moment redirect to `k8s.gcr.io`)

**2. Feel free to provide a design supporting your feature request.**

/kind feature
/sig k8s-infra

ameukam:
cc @justinsb @olemarkus @hakman

ameukam:
cross-ref: https://github.com/kubernetes/k8s.io/issues/3411

PR #13402, opened 2022-03-23
Use k8s.gcr.io for k8s side-loaded images

hakman:
Refs: #13365


olemarkus (approved):

ameukam:
@hakman curious about why the switch to `k8s.gcr.io` ? Anything we need to improve for `registry.k8s.io` ?

hakman:
@ameukam We sideload k8s images from .tgz files. Those are built with `k8s.gcr.io` as registry. Once `registry.k8s.io` is GA, k/k should switch too.

PR #13410, opened 2022-03-24
Correctly detect GovCloud regions

mixja:
This PR correctly detects S3 HTTP URLs in GovCloud regions for `spec.assets.fileRepository`.

e.g.  this is the S3 object URL format for GovCloud
```
spec:
  assets:
    fileRepository: https://bucket-name.s3-us-gov-west-1.amazonaws.com/bucket-path
```

Prior to this PR, the bucket name is captured incorrectly in the example above as `s3-us-gov-west-1`.

### DIFF
--- a/util/pkg/vfs/s3context.go
+++ b/util/pkg/vfs/s3context.go
@@ -36,11 +36,12 @@ import (
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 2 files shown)


hakman (approved):
Thanks @mixja. 👍

PR #13426, opened 2022-03-28
Run CSI controller in masters

### LINKED ISSUE #13425: CSI provisioning does not work for GCE
/kind bug

 this information.**

master

 a `kops` flag.**

1.23.5

GCE

I created new cluster and after that I deployed application that uses PVC. 

```
% kubectl get pvc -n monitoring
NAME                                 STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
prometheus-k8s-db-prometheus-k8s-0   Pending                                      standard-csi   4m29s
prometheus-k8s-db-prometheus-k8s-1   Pending                                      standard-csi   3m50s
```

I see following in csi provisioner logs

```
% kubectl logs csi-gce-pd-controller-7cc567ddc9-9cqfn -c csi-provisioner
...
Reason: forbidden, Message: Required 'compute.disks.create' permission for 'projects/x/zones/europe-north1-a/disks/pvc-91238f39-a327-4004-8fc8-613809d13ecd'
Reason: forbidden, Message: Required 'compute.disks.setLabels' permission for 'projects/x/zones/europe-north1-a/disks/pvc-91238f39-a327-4004-8fc8-613809d13ecd'
```

So I think we are missing some permissions from GCE service account. Seems that CSI controller is running in nodes, so is it using node service account? There can be also other permissions that are missing, I think we could allow compute.disks.*?

zetaab:
fixes #13425

I tested creating new pvcs, resizing them and deleting. Everything works smoothly now.

### DIFF
--- a/tests/integration/update_cluster/minimal_gce/data/aws_s3_bucket_object_minimal-gce.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/minimal_gce/data/aws_s3_bucket_object_minimal-gce.example.com-addons-bootstrap_content
@@ -61,7 +61,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.23
     manifest: gcp-pd-csi-driver.addons.k8s.io/k8s-1.23.yaml
-    manifestHash: 1d39fdc2fcd7473ff3a0d9bd4141036be0933a94f4862f34e37be8b3163dcdb5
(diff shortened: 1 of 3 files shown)


olemarkus (approved):
Until we support pod identities on GCP

zetaab:
@olemarkus can you readd lgtm, I updated the expected test files

PR #13433, opened 2022-03-29
Use kubekins image for postsubmit, not imagebuilder

olemarkus:


rifelpet:
Out of curiosity, what's the reasoning for this? Does imagebuilder not have the dependencies we need?

Issue #13441, opened 2022-03-30
DNS controller crashes when running in Digital Ocean with multiple masters


zak905:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Version 1.23.0 (git-a067cd7742a497a5c512762b9880664d865289f1)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-17T03:51:43Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:52:18Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**

Digital Ocean

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

A simple way to reproduce this issue is to create a cluster in Digital ocean with multiple masters

for example:
``` 
kops create cluster --cloud=digitalocean --name=app.test.com --networking=flannel --master-count=3 --api-loadbalancer-type=public --node-size=s-2vcpu-4gb --image=debian-11-x64 --node-count=2 --zones=fra1 --ssh-public-key=~/.ssh/kops_rsa.pub
```

**5. What happened after the commands executed?**

First thing that I noticed was that:
* DNS records in Digital Ocean are not updated properly, there is only one record `api.app.test.com` and it does not have the right load balancer address but rather the placeholder address here https://github.com/kubernetes/kops/blob/master/dnsprovider/pkg/dnsprovider/providers/do/dns.go#L39
* This says that something went wrong or the domain update process was interupted
* After checking the dns controller pod logs I discovered that a panic was happening

**6. What did you expect to happen?**

To have two DNS record in Digital Ocean with correct IPs as values
one for api.test.app.com 
and one for api.internal.app.com

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
[...]

PR #13443, opened 2022-03-31
fix dns controller crashing for DO

zak905:
More details can be found in the issue #13441

### DIFF
--- a/dnsprovider/pkg/dnsprovider/providers/do/dns.go
+++ b/dnsprovider/pkg/dnsprovider/providers/do/dns.go
@@ -185,20 +185,27 @@ func (r *resourceRecordSets) List() ([]dnsprovider.ResourceRecordSet, error) {
 		return nil, err
 	}
 
-	var rrset *resourceRecordSet
 	var rrsets []dnsprovider.ResourceRecordSet
+	rrsetsWithoutDups := make(map[string]*resourceRecordSet)
+
 	for _, record := range records {
 		// digitalocean API returns the record without the zone
 		// but the consumers of this interface expect the zone to be included
 		recordName := dns.EnsureDotSuffix(record.Name) + r.Zone().Name()
-		rrset = &resourceRecordSet{
-			name:       recordName,
-			data:       record.Data,
-			ttl:        record.TTL,
-			recordType: rrstype.RrsType(record.Type),
+		if set, ok := rrsetsWithoutDups[recordName]; !ok {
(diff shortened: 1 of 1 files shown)


justinsb:
Thanks for tracking down and fixing.  Is making the DO provider accept multiple records an option?  That might avoid encoding the package name as a string (without introducing a circular package ref)

zak905:
you are welcome. I have already made an attempt here https://github.com/zak905/kops/commit/40136a0c8eceb088c58379c5dd941f876eeb6781 to make the DO DNS provider accomodate multiple values.

I have not tested it yet in my cluster, but if you confirm this is a better alternative, I can test it and update the PR

zak905:
I have just tested now the changes in https://github.com/zak905/kops/commit/40136a0c8eceb088c58379c5dd941f876eeb6781 
They seem to work well, no crashes, also the records seem to have the correct values. 
I am updating this PR in this case, if this is the prefeffered approach.

dnsprovider/pkg/dnsprovider/providers/do/dns.go
```diff
@@ -185,20 +185,27 @@ func (r *resourceRecordSets) List() ([]dnsprovider.ResourceRecordSet, error) {
 		return nil, err
 	}
 
-	var rrset *resourceRecordSet
 	var rrsets []dnsprovider.ResourceRecordSet
+	rrsetWithoutDups := make(map[string]*resourceRecordSet)
```
hakman, line 189: Should this be plural to keep similar with `rrsets` above?
```suggestion
	rrsetsWithoutDups := make(map[string]*resourceRecordSet)
```

hakman (changes requested):

dnsprovider/pkg/dnsprovider/providers/do/dns.go
zak905: sure

hakman: Thanks. Please also squash the commits and change the commit message to something explaining the change, like "[DigitalOcean]: accept multiple records in dns-controller".
/lgtm

zak905: Done.

hakman: 👍

hakman (approved):

PR #13445, opened 2022-03-31
Enable internal load balancers when running on GCP

nat-henderson:
* Add ILBs, broadly following the AWS model.  The following new
capabilities are added for clusters in GCP:
  * Cluster's spec.api.loadBalancer can be set to 'type: internal' on
    GCP.
    * Therefore, GCP can now create:
        * regional backend services
        * regional (non-legacy) healthchecks
        * firewall rules with "internal" load-balancing scheme
        * firewall rules with dot-notation-specified IP addresses
  * Cluster's spec.api.loadBalancer's 'subnets' field functions
    as in the AWS model.

A few incidental changes are included, either because this change
touched the relevant code or because my use case happened to trigger the
issues that are fixed here.

* Resource names which are too long to be allowed will
  now be shortened.  The cluster name will be preserved to allow
  tools which sort & filter by cluster to continue to work.

* The existing HealthCheck is moved to be named HTTPHealthCheck.
  * GCP supports several kinds of health check, and this change matches their
     naming scheme for the two types in use here.

* Cluster's spec.networkID field can be prefixed by project to use
  GCP's common cross-project networking model.
    * The presumption is that all specified subnets belong to this
      network and therefore this project.

* Add missing operation wait on forwarding rule creation.

* Some Terraform output improvements:
    * Permit no-ACL files in GCS buckets in Terraform output.
    * Enable marginally better cross-resource reference in Terraform outputs
    * Add project to network + subnetwork literals in Terraform output.
    * Add terraform output to backend services and health checks.

Co-authored-by: Travis Reid <travis_reid@apple.com>


hakman:
Thanks for improving the CGP implementation @nat-henderson. One more step towards getting it out of beta.
/cc @justinsb 
/assign @justinsb

pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -101,13 +84,118 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+	lbSpec := b.Cluster.Spec.API.LoadBalancer
+	hc := &gcetasks.HealthCheck{
+		Name:      s(b.NameForHealthCheck("api")),
+		Port:      443,
+		Lifecycle: b.Lifecycle,
+	}
+	c.AddTask(hc)
+	var igms []*gcetasks.InstanceGroupManager
+	for _, ig := range b.InstanceGroups {
+		if ig.Spec.Role != "Master" {
```
justinsb, line 116: Nit: It's a good idea to use the constants (InstanceGroupRoleMaster), particularly where at some stage we'll want to address the legacy naming here.

pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -101,13 +84,118 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
...
+		t := true
+		subnet := &gcetasks.Subnet{
+			Name:    s(sn.Name),
+			Network: network,
+			Shared:  &t,
+			// Override lifecycle because these subnets are specified
+			// to already exist.
+			Lifecycle: fi.LifecycleExistsAndWarnIfChanges,
+		}
+		if subnetNotSpecified(sn, b.Cluster.Spec.Subnets) {
```
justinsb, line 151: This is clever.  For a future PR, could we also automatically associate to the subnet(s) in use if no Subnets are specified?

pkg/resources/gce/gce.go
```diff
@@ -1037,6 +1041,105 @@ func deleteServiceAccount(cloud fi.Cloud, r *resources.Resource) error {
...
+func (d *clusterDiscoveryGCE) listBackendServices() ([]*resources.Resource, error) {
+	c := d.gceCloud
+
+	svcs, err := c.Compute().RegionBackendServices().List(context.Background(), c.Project(), c.Region())
+	if err != nil {
+		if gce.IsNotFound(err) {
+			klog.Infof("backend services not found, assuming none exist in project: %q region: %q", c.Project(), c.Region())
+			return nil, nil
+		}
+		return nil, fmt.Errorf("Failed to list backend services: %s", err)
```
justinsb, line 1072: Nit: %w

pkg/resources/gce/gce.go
```diff
@@ -1037,6 +1041,105 @@ func deleteServiceAccount(cloud fi.Cloud, r *resources.Resource) error {
 	return err
 }
 
+// containsOnlyListedIGMs returns true if all the given backend service's backends
+// are contained in the provided list of IGM resources.
+func containsOnlyListedIGMs(svc *compute.BackendService, igms []*resources.Resource) bool {
+	for _, be := range svc.Backends {
+		listed := false
+		for _, igm := range igms {
+			if strings.HasSuffix(be.Group, "/"+igm.Name) {
```
justinsb, line 1050: Nit: is this matching strict enough?  Maybe add a TODO to double-check / revisit if we can make it stricter.

upup/pkg/fi/cloudup/gce/utils.go
```diff
@@ -81,7 +84,22 @@ func SafeClusterName(clusterName string) string {
 // SafeObjectName returns the object name and cluster name escaped for GCE
 func SafeObjectName(name string, clusterName string) string {
 	gceName := name + "-" + clusterName
-
+	if len(gceName) > 64 {
```
justinsb, line 87: Can I suggest we pull this function into a separate PR?  The rest of the PR has just minor quibbles and I think is basically ready to go...

Here I guess there are some conflicting goals, in that we want to preserve the suffix as it's the primary mechanism we have for recognizing the cluster, but we want to preserve the prefix because it has more signal to the user than the common sufix.  On AWS when we're limited in length we focus on the prefix (check out GetResourceName32 and GetClusterName40).

We can also tweak the deletion logic to look for the hash suffix, which would let us prioritize the name over the suffix, but would require more code changes in deletion.

justinsb:
Thanks for an amazing contribution!  I had a few nits / suggestions, but the only one that I think is blocking is how we should truncate overly long names - I propose pulling that into a separate PR and if desired I can help out / experiment with a few approaches there.

The other thing we could do on this stream is to create a new test that generates the expected terraform output, like e.g. https://github.com/kubernetes/kops/tree/master/tests/integration/update_cluster/minimal_gce_private (referenced from https://github.com/kubernetes/kops/blob/d751d0f0569642dac97d05973a7491f6b88e3551/cmd/kops/integration_test.go#L255-L260).  It's easier to maintain than it looks - you copy the directory, edit in-v1alpha2.yaml, create the Test method, and then run `./hack/update-expected.sh` to update the "golden output".  We check in the golden files, and the big advantage of this is that then we can see the generated configuration (in terraform format) and in particular we can see if the terraform changes.  We try to avoid terraform changes if possible because it's user-visible and some changes can be disruptive, so it's good to have a test case that covers your use-case to avoid breakage :-) .  I don't consider this a blocker for merging this PR, however, as it looks like the changes only affect the newly-supported configurations.

util/pkg/vfs/gsfs.go
```diff
@@ -428,6 +428,10 @@ func (p *GSPath) RenderTerraform(w *terraformWriter.TerraformWriter, name string
 		RoleEntity: make([]string, 0),
 		Provider:   terraformWriter.LiteralTokens("google", "files"),
 	}
+	if acl == nil {
```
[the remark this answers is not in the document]
nat-henderson: Thanks, you're right!  I can make that change, another 10 lines of whitespace diff doesn't cost anyone anything.  (:

upup/pkg/fi/cloudup/gcetasks/subnet.go
```diff
@@ -50,8 +50,11 @@ func (e *Subnet) CompareWithID() *string {
 
 func (e *Subnet) Find(c *fi.Context) (*Subnet, error) {
 	cloud := c.Cloud.(gce.GCECloud)
-
-	s, err := cloud.Compute().Subnetworks().Get(cloud.Project(), cloud.Region(), *e.Name)
+	_, project, err := gce.ParseNameAndProjectFromNetworkID(c.Cluster.Spec.NetworkID)
+	if err != nil || project == "" {
```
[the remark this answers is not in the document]
nat-henderson: Ah, sure - I'll just have to make it so that `""` doesn't error, which is easy (and probably more correct anyway).

upup/pkg/fi/cloudup/gcetasks/subnet.go
```diff
@@ -277,8 +280,12 @@ func (e *Subnet) TerraformLink() *terraformWriter.Literal {
 			klog.Fatalf("GCEName must be set, if subnet is shared: %#v", e)
 		}
 
-		klog.V(4).Infof("reusing existing subnet with name %q", *e.Name)
-		return terraformWriter.LiteralFromStringValue(*e.Name)
+		name := *e.Name
+		if e.Network != nil && e.Network.Project != nil {
+			name = *e.Network.Project + "/" + name
```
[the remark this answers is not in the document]
nat-henderson: That's to set the value to a literal string - the string `project/network` - which is how terraform parses this value already.  :)

upup/pkg/fi/cloudup/gcetasks/network.go
```diff
@@ -47,8 +48,12 @@ func (e *Network) CompareWithID() *string {
 
 func (e *Network) Find(c *fi.Context) (*Network, error) {
 	cloud := c.Cloud.(gce.GCECloud)
+	project := cloud.Project()
+	if e.Project != nil {
```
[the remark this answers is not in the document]
nat-henderson: Oh, yeah, I was going on the Terraform theory - leave it unset to use the default - but I see that's not a universal standard and will swap to this in the future!

pkg/model/gcemodel/api_loadbalancer.go
nat-henderson: Done.

pkg/model/gcemodel/api_loadbalancer.go
nat-henderson: Added todo.

pkg/resources/gce/gce.go
nat-henderson: ah!  thanks for the catch.

pkg/resources/gce/gce.go
nat-henderson: Added note!

nat-henderson:
Thanks for the review! Tests on the CR changes are still running but I assume they will pass - or fail due to typos or something easily resolvable.

> the only one that I think is blocking is how we should truncate overly long names - I propose pulling that into a separate PR and if desired I can help out / experiment with a few approaches there.

I have pulled that out - we can talk about how best to do that in another pr, although we are going to need a solution especially now that there are forwarding rules including subnet names, which means _very_ long names.

> The other thing we could do on this stream is to create a new test that generates the expected terraform output, like e.g. https://github.com/kubernetes/kops/tree/master/tests/integration/update_cluster/minimal_gce_private ...

thanks for the guidance on this! I'm working on that - if I finish before you get back around to the review, I'll post it here, otherwise, it'll be a new PR.

nat-henderson:
Okay - this passes for me, so let's see what happens on prow...

hakman:
@nat-henderson A rebase would be needed to pass all the tests.

nat-henderson:
Got it - rebase worked fine, running local tests now...

nat-henderson:
Okay, _that_ should do it...

nat-henderson:
Hmm… I’m going to just have to try to run that again. /retest

nat-henderson:
Cool - thought so. In that case this is ready to go in, @justinsb, unless you wanna get anything else done here beforehand!

justinsb (approved):

justinsb:
Thanks for addressing my comments.  Looks like there's a small conflict so this needs a rebase, but this LGTM.  (I had a few nits about how many fields we need to find, but I'm not certain about them and they're easily addressed later).

/approve
/lgtm

(But it won't do anything until you rebase, I'm afraid)

upup/pkg/fi/cloudup/gcetasks/backend_service.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		if gce.IsNotFound(err) {
+			return nil, nil
+		}
+
+		return nil, fmt.Errorf("error listing Backend Services: %v", err)
+	}
+
+	actual := &BackendService{}
+	actual.Name = &r.Name
```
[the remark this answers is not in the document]
nat-henderson: I think I've fixed that - I'll let the tests run.  :)

upup/pkg/fi/cloudup/gcetasks/healthcheck.go
```diff
@@ -23,73 +23,125 @@ import (
...
+
+		return nil, fmt.Errorf("error listing Health Checks: %v", err)
 	}
-	// System fields
-	actual.Lifecycle = e.Lifecycle
-	e.SelfLink = r.SelfLink
+
+	actual := &HealthCheck{}
+	actual.Name = &r.Name
+	actual.Port = r.TcpHealthCheck.Port
```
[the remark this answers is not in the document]
nat-henderson: Good call - fixed.

hakman:
@nat-henderson If you don't mind, could you also squash some of the commits? Thanks!

nat-henderson:
Sure thing, let me fix that compile error there and squash it down...

nat-henderson:
Ah - the rebase - gotta update the goldens again.

PR #13449, opened 2022-04-01
Add support for fetching files from private GCS bucket.

nat-henderson:
* Service account authentication is available when running on GCE.
    * If we are accessing a google cloud storage bucket, we now fetch
      an auth token and apply it to any requests made.
    * This includes the initial request to download nodeup,
      subsequent requests made by nodeup to download additional files.
    * It also rewrites requests made as part of preflight checking
      (e.g. checking the existence of the nodeup binary) if those
      requests are GCS requests.

Room for improvement:
    * We could support rewrites in the other direction as well - allow
      the user to specify the gs:// scheme url and rewrite it
      to storage.googleapis.com when fetching nodeup.
    * This would be useful generically, and could be extended to include
      authenticated requests to S3 buckets.

### DIFF
--- a/pkg/model/resources/nodeup.go
+++ b/pkg/model/resources/nodeup.go
@@ -60,6 +60,22 @@ function ensure-install-dir() {
   cd ${INSTALL_DIR}
 }
 
+try-download-file() {
+  local -r url="$1"
+  local -r file="$2"
+  local -r auth="$3"
+  if [[ "$auth" == "" ]]; then
+        if curl -f --compressed -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
+        if wget --compression=auto -O "${file}" --connect-timeout=20 --tries=6 --wait=10 "${url}"; then return 0; fi
+        if curl -f -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
+        if wget -O "${file}" --connect-timeout=20 --tries=6 --wait=10 "${url}"; then return 0; fi
+  else
+        if curl -f --compressed -Lo "${file}" -v -H "Authorization: Bearer ${auth}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
+        if curl -f -Lo "${file}" -v -H "Authorization: Bearer ${auth}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
+  fi
+  return 1
+}
+
 # Retry a download until we get it. args: name, sha, urls
 download-or-bust() {
   local -r file="$1"
@@ -76,29 +92,28 @@ download-or-bust() {
 
   while true; do
     for url in "${urls[@]}"; do
-      commands=(
-        "curl -f --compressed -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10"
-        "wget --compression=auto -O "${file}" --connect-timeout=20 --tries=6 --wait=10"
-        "curl -f -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10"
-        "wget -O "${file}" --connect-timeout=20 --tries=6 --wait=10"
-      )
-      for cmd in "${commands[@]}"; do
-        echo "Attempting download with: ${cmd} {url}"
-        if ! (${cmd} "${url}"); then
-          echo "== Download failed with ${cmd} =="
-          continue
-        fi
-        if ! validate-hash "${file}" "${hash}"; then
-          echo "== Hash validation of ${url} failed. Retrying. =="
-          rm -f "${file}"
-        else
-          echo "== Downloaded ${url} (SHA256 = ${hash}) =="
-          return 0
-        fi
-      done
+      local access_token=""
+      if [[ "${GCE_ACCESS_TOKEN_URL:-}" != "" ]]; then
+        # The token always begins with 'ya29' and continues until the next double-quote.
+        # We would normally use jq to be more tolerant of format changes, 
+        # but it's not available in flatcar during startup (depends on docker, which isn't up yet.)
+        access_token=$(curl "$GCE_ACCESS_TOKEN_URL" -H "Metadata-Flavor: Google" | grep -Eo 'ya29\.[^\"]*')
+      fi
+
+      echo "Attempting download of ${url}"
+      if ! try-download-file "${url}" "${file}" "${access_token}" ; then
+        continue
+      fi
+      if ! validate-hash "${file}" "${hash}"; then
+        echo "== Hash validation of ${url} failed. Retrying. =="
+        rm -f "${file}"
+      else
+        echo "== Downloaded ${url} (SHA256 = ${hash}) =="
+        return
+      fi
     done
 
-    echo "All downloads failed; sleeping before retrying"
+    echo "Download failed with all plausible commands; sleeping before retrying"
     sleep 60
   done
 }
--- a/upup/pkg/fi/http.go
+++ b/upup/pkg/fi/http.go
@@ -18,11 +18,14 @@ package fi
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
+	"os/exec"
 	"path"
 	"time"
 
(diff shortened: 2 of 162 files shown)


nat-henderson:
Okay ... sorry for what may be a ton of PR noise here, but the same commands are passing locally, so there must be something about the execution environment I need to sort out, which means I gotta upload and re-run tests at least a few times...

nat-henderson:
Okay, that looks like a timeout, so I'm just going to try it again.

nat-henderson:
Hm, same problem:
> The dns-controller Kubernetes deployment has not updated the Kubernetes cluster's API DNS entry to the correct IP address.  The API DNS IP address is the placeholder address that kops creates: 203.0.113.123.  Please wait about 5-10 minutes for a master to start, dns-controller to launch, and DNS to propagate.  The protokube container and dns-controller deployment logs may contain more diagnostic information.  Etcd and the API DNS entries must be updated for a kops Kubernetes cluster to start.
[2414](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/13449/pull-kops-e2e-kubernetes-aws/1511472037783146496#1:build-log.txt%3A2414)

Suppose I'll leave that there, then - @justinsb Let me know if that looks like it reflects a real error caused by my PR during your review?

pkg/model/resources/nodeup.go
```diff
@@ -76,29 +92,28 @@ download-or-bust() {
...
-        if ! validate-hash "${file}" "${hash}"; then
-          echo "== Hash validation of ${url} failed. Retrying. =="
-          rm -f "${file}"
-        else
-          echo "== Downloaded ${url} (SHA256 = ${hash}) =="
-          return 0
-        fi
-      done
+      local access_token=""
+      if [[ "$GCE_ACCESS_TOKEN_URL" != "" ]]; then
```
hakman, line 96: Quick hint:
```
== nodeup node config starting ==
/var/lib/cloud/instance/scripts/part-001: line 66: GCE_ACCESS_TOKEN_URL: unbound variable
```

pkg/model/resources/nodeup.go
```diff
@@ -60,6 +60,22 @@ function ensure-install-dir() {
...
+  local -r url="$1"
+  local -r file="$2"
+  local -r auth="$3"
+  if [[ "$auth" == "" ]]; then
+		if curl -f --compressed -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
+		if wget --compression=auto -O "${file}" --connect-timeout=20 --tries=6 --wait=10 "${url}"; then return 0; fi
+		if curl -f -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
+		if wget -O "${file}" --connect-timeout=20 --tries=6 --wait=10 "${url}"; then return 0; fi
+  else
+		if curl -f --compressed -Lo "${file}" -v -H "Authorization: Bearer ${auth}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
```
justinsb: One of the things we have to be careful about here is logging the auth token in the output, in particularly for the e2e tests.  However the good news is that this script _doesn't_ have the `set -x` / `set -o xtrace` so we should be OK!

pkg/model/resources/nodeup.go
```diff
@@ -76,29 +92,28 @@ download-or-bust() {
...
-        else
-          echo "== Downloaded ${url} (SHA256 = ${hash}) =="
-          return 0
-        fi
-      done
-    done
-
-    echo "All downloads failed; sleeping before retrying"
+	  local access_token=""
+	  if [[ "$GCE_ACCESS_TOKEN_URL" != "" ]]; then
```
justinsb: As @hakman pointed out, we need to `if [[ "$GCE_ACCESS_TOKEN_URL" != "" ]]; then` -> `if [[ "${GCE_ACCESS_TOKEN_URL:-}" != "" ]]; then` because of `set -o nounset`.

It's a little hard to see, but it was logged in https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/13449/pull-kops-e2e-kubernetes-aws/1511472037783146496/artifacts/13.233.128.192/cloud-init-output.log  You can get there by clicking on the failing test, then the artifacts tab, then artifacts / pick an IP address / cloud--init-output.log.

pkg/model/resources/nodeup.go
```diff
@@ -76,29 +92,28 @@ download-or-bust() {
...
+		# The token always begins with 'ya29' and continues until the next double-quote.
+		# We would normally use jq to be more tolerant of format changes, 
+		# but it's not available in flatcar during startup (depends on docker, which isn't up yet.)
```
justinsb: TIL that the GCP metadata service supports an alt=text parameter, and we've even documented it :-)   https://cloud.google.com/compute/docs/metadata/querying-metadata#formatting_query_output

That might be easier to parse, though I don't see this as a blocker.

upup/pkg/fi/http.go
```diff
@@ -105,6 +159,9 @@ func downloadURLAlways(url string, destPath string, dirMode os.FileMode) error {
 	if err != nil {
 		return fmt.Errorf("Cannot create request: %v", err)
 	}
+	if err := addAuthHeadersFromEnvironment(httpClient, req); err != nil {
+		return fmt.Errorf("Cannot determine authentication headers: %v", err)
```
justinsb: Super-nit: we're trying to use %w for all new error messages (at some stage we'll go back and fix all the existing %v)

upup/pkg/fi/http.go
```diff
@@ -67,6 +70,57 @@ func DownloadURL(url string, dest string, hash *hashing.Hash) (*hashing.Hash, er
 	return hash, nil
 }
 
+func addAuthHeadersFromEnvironment(client *http.Client, req *http.Request) error {
```
justinsb, line 73: It _might_ be easier to pass the URL to nodeup etc as a `gs://` URL, rather than rewriting to storage.googleapis.com.  I _think_ that nodeup will then know to get a GCP token etc

pkg/model/resources/nodeup.go
```diff
@@ -65,13 +65,13 @@ try-download-file() {
   local -r file="$2"
   local -r auth="$3"
   if [[ "$auth" == "" ]]; then
-		if curl -f --compressed -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
-		if wget --compression=auto -O "${file}" --connect-timeout=20 --tries=6 --wait=10 "${url}"; then return 0; fi
-		if curl -f -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
-		if wget -O "${file}" --connect-timeout=20 --tries=6 --wait=10 "${url}"; then return 0; fi
+        if curl -f --compressed -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10 "${url}"; then return 0; fi
```
justinsb, line 68: One reason to prefer tabs _here_ (for once, a reason) is that we're a little space constrained on AWS - the limit is 16KB, though we can gzip. OTOH if this actually becomes an issue I think feeding this through a filter that just trimmed all leading whitespace on a line would be the right thing to do, so I agree this makes sense.

justinsb:
I really like the idea of tweaking that script to support metadata-based authentication (not least because it isn't that hard, though I'm a little scared about the AWS v2 exchange!)

Looking at the nodeup logic that maps https://storage.googleapis.com back to gs://, I'm wondering whether we can avoid the `gs://` ->  `https://storage.googleapis.com` translation in the first place and whether that might be easier.   I might try an experiment :-)

justinsb:
I did a little experiment, and it looks (as you suggested) we need some small changes e.g. to the download from https logic, but that it might be easier to accept a KOPS_BASE_URL that is `gs://`, and only rewrite NodeUpSourceAmd64 / NodeUpSourceArm64.  The advantage of this is that we keep the vfs layer a little "purer" - it doesn't "magically" switch from https://storage.googleapis.com to gs:// .  My scratch-pad (based on your changes) is here: https://github.com/justinsb/kops/tree/authenticated_gcs_download

upup/pkg/fi/http.go
nat-henderson: Cool - did that for this file!

util/pkg/vfs/context.go
```diff
@@ -122,7 +122,6 @@ func (c *VFSContext) ReadFile(location string, options ...VFSOption) ([]byte, er
 				return nil, fmt.Errorf("unknown metadata type: %q in %q", u.Host, location)
 			}
 
-		case "http", "https":
```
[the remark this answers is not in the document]
nat-henderson: That code is reached under the case that the scheme is not "metadata" - so, "http" or "gs" or "s3".  :)

upup/pkg/fi/http.go
nat-henderson: Unforunately, on flatcar, it doesn't seem to work - gsutil requires `docker` on flatcar, which isn't up yet at this point.  I can try to work around that if this is important.

pkg/model/resources/nodeup.go
nat-henderson: Yeah, I didn't modify it at first, but it causes the linter to complain - I can add a lint ignore tag if you like!

upup/pkg/fi/http.go
```diff
@@ -159,20 +164,32 @@ func downloadURLAlways(url string, destPath string, dirMode os.FileMode) error {
...
-		return fmt.Errorf("Cannot determine authentication headers: %v", err)
-	}
-
 	response, err := httpClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("error doing HTTP fetch of %q: %v", url, err)
 	}
-	defer response.Body.Close()
+
+	if response.StatusCode == 401 || response.StatusCode == 403 {
```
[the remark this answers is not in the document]
nat-henderson: Oh, cool - I was worried it was sloppy or otherwise undesirable, ha, but glad it looks good to you!

upup/pkg/fi/http.go
```diff
@@ -159,20 +164,32 @@ func downloadURLAlways(url string, destPath string, dirMode os.FileMode) error {
...
 	}
-	if err := addAuthHeadersFromEnvironment(httpClient, req); err != nil {
-		return fmt.Errorf("Cannot determine authentication headers: %v", err)
-	}
-
 	response, err := httpClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("error doing HTTP fetch of %q: %v", url, err)
 	}
-	defer response.Body.Close()
```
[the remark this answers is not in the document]
nat-henderson: Oh, nice!  Thank you, done.

util/pkg/vfs/context.go
```diff
@@ -121,8 +121,6 @@ func (c *VFSContext) ReadFile(location string, options ...VFSOption) ([]byte, er
 			default:
 				return nil, fmt.Errorf("unknown metadata type: %q in %q", u.Host, location)
 			}
-
-			return c.readHTTPLocation(location, nil, opts)
```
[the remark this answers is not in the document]
nat-henderson: Hmm... no, I think we should fall through for any HTTP / HTTPS URL, isn't that right?

upup/pkg/fi/http.go
nat-henderson: Ah, I see that it works on your config - maybe I can just make it work with mine somehow!  I'll take a look...

pkg/model/resources/nodeup.go
```diff
@@ -76,29 +92,28 @@ download-or-bust() {
...
+        echo "== Hash validation of ${url} failed. Retrying. =="
+        rm -f "${file}"
+      else
+        echo "== Downloaded ${url} (SHA256 = ${hash}) =="
+        return
+      fi
     done
 
-    echo "All downloads failed; sleeping before retrying"
+    echo "Download failed with all plausible commands; sleeping before retrying"
```
hakman, line 116: By any chance you meant `possible` instead of `plausible`?

nat-henderson: "possible" reads a lot cleaner, yeah.

nat-henderson:
/remove-lifecycle stale

Nope, I'm still here!  If you can come back to this PR I do still want to get it in.  :)

PR #13452, opened 2022-04-02
Adopt control-plane taint and remove master role labels

olemarkus:
Following kubeadm, we can move to the next stage of master label/taints removal.
The master labels are removed.
While kubeadm only adds the control plane taint, removing the master taint in 1.25, we swap the master taint with the control-plane taint in one go.

### DIFF
--- a/docs/releases/1.24-NOTES.md
+++ b/docs/releases/1.24-NOTES.md
@@ -25,6 +25,38 @@ kOps will directly manage the Karpenter Provisioner resources. Read more about h
 
 # Breaking changes
 
+## Control plane taints and labels
+
+As of Kubernetes version 1.24, the control plane (formerly master) nodes no longer have the deprecated `node-role.kubernetes.io/master` label.
+The deprecated `node-role.kubernetes.io/master` taint has been replaced by `node-role.kubernetes.io/control-plane`. If you run your own workload on the control plane, you have to adjust your Pod spec to accommodate for this change.
+
+The following shows a node affinity/node label selector and tolerations that works with both new and old control plane nodes:
+
+```yaml
+spec:
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: node-role.kubernetes.io/control-plane
+            operator: Exists
+        - matchExpressions:
+          - key: node-role.kubernetes.io/master
+            operator: Exists
+  tolerations:
+  - key: node-role.kubernetes.io/control-plane
+    operator: Exists
+  - key: node-role.kubernetes.io/master
+    operator: Exists
+```
+
+## Removing the `kubernetes.io/role` label
+
+The deprecated `kubernetes.io/role` label has been removed for all roles as of Kubernetes version 1.24. Use `node-role.kubernetes.io/<role>` label instead.
+
+## Other breaking changes
+
 * Support for Kubernetes version 1.18 has been removed.
 
 * Support for Aliyun/Alibaba Cloud has been removed.
@@ -42,8 +74,6 @@ kOps will directly manage the Karpenter Provisioner resources. Read more about h
 * Support for Kubernetes version 1.20 is deprecated and will be removed in kOps 1.26.
 
 * All legacy addons are deprecated in favor of managed addons, including the [metrics server addon](https://github.com/kubernetes/kops/tree/master/addons/metrics-server) and the [autoscaler addon](https://github.com/kubernetes/kops/tree/master/addons/cluster-autoscaler).
-
-* The `node-role.kubernetes.io/master` and `kubernetes.io/role` labels are deprecated and might be removed from control plane nodes in future versions of kOps.
 
 * Due to lack of maintainers, the CloudFormation support has been deprecated. The current implementation will be left as-is until the implementation needs updates or otherwise becomes incompatible. At that point, it will be removed. We very much welcome anyone willing to contribute to this target.
 
--- a/nodeup/pkg/model/kubelet.go
+++ b/nodeup/pkg/model/kubelet.go
@@ -588,7 +588,11 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 	{
 		if len(c.Taints) == 0 && isMaster {
 			// (Even though the value is empty, we still expect <Key>=<Value>:<Effect>)
-			c.Taints = append(c.Taints, nodelabels.RoleLabelMaster16+"=:"+string(v1.TaintEffectNoSchedule))
+			if b.IsKubernetesLT("1.24") {
+				c.Taints = append(c.Taints, nodelabels.RoleLabelMaster16+"=:"+string(v1.TaintEffectNoSchedule))
+			} else {
+				c.Taints = append(c.Taints, nodelabels.RoleLabelControlPlane20+"=:"+string(v1.TaintEffectNoSchedule))
+			}
 		}
 		if len(c.Taints) == 0 && isAPIServer {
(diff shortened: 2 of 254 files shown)


upup/models/cloudup/resources/addons/dns-controller.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -24,15 +24,29 @@ spec:
...
     spec:
       priorityClassName: system-cluster-critical
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
+                operator: Exists
+              - key: kops.k8s.io/kops-controller-pki
```
justinsb, line 34: I don't see this in the nodeSelector - copy/paste error?

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
```diff
@@ -241,6 +241,21 @@ spec:
         - key: node-role.kubernetes.io/master
           operator: Exists
       {{ end }}
+---
+apiVersion: policy/v1beta1
```
justinsb, line 245: I'm missing something - I don't see how a PDB is related....

upup/models/cloudup/resources/addons/aws-load-balancer-controller.addons.k8s.io/k8s-1.19.yaml.template
```diff
@@ -949,3 +959,18 @@ spec:
     apiGroup: elbv2.k8s.aws
     kind: IngressClassParams
     name: alb
+---
+apiVersion: policy/v1beta1
```
justinsb, line 963: As before - why are we adding PDBs?

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.10.yaml.template
```diff
@@ -1161,3 +1161,18 @@ spec:
   secretName: hubble-relay-client-certs
 {{ end }}
 {{ end }}
+---
+apiVersion: policy/v1beta1
```
justinsb, line 1165: PDB question again :-)

nodeup/pkg/model/kubelet.go
```diff
@@ -589,6 +589,9 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 		if len(c.Taints) == 0 && isMaster {
 			// (Even though the value is empty, we still expect <Key>=<Value>:<Effect>)
 			c.Taints = append(c.Taints, nodelabels.RoleLabelMaster16+"=:"+string(v1.TaintEffectNoSchedule))
+			if b.IsKubernetesGTE("1.24") {
+				c.Taints = append(c.Taints, nodelabels.RoleLabelControlPlane20+"=:"+string(v1.TaintEffectNoSchedule))
```
justinsb, line 593: It's probably a good thing to get this into testing, but we could also split this PR up to _not_ add the new taints / drop the old labels _yet_.  Ideally we would get this into 1.24, but if users have problems with it we might expose a feature flag..  That said, I think you've shown that it isn't too hard to write a compatible manifest.

justinsb:
Awesome stuff.  A few questions about why we need PDBs etc.

BTW if you don't know about it ... https://github.com/torbiak/git-autofixup/ - I find it really useful when I have these multi-commit PRs.

nodeup/pkg/model/kubelet.go
olemarkus: I don't think the migration will be easier than it is right now. And we've both warned about this for some time and kubeadm is now making the move.

olemarkus:
> Awesome stuff. A few questions about why we need PDBs etc.
> 

Got this split out and merged as part of a separate PR now. 
I think this one is good to go.

> BTW if you don't know about it ... https://github.com/torbiak/git-autofixup/ - I find it really useful when I have these multi-commit PRs.

Thanks, I did not know about this one. Very useful indeed.

docs/releases/1.24-NOTES.md
```diff
@@ -25,6 +25,34 @@ kOps will directly manage the Karpenter Provisioner resources. Read more about h
...
+          - key: node-role.kubernetes.io/control-plane
+            operator: Exists
+        - matchExpressions:
+          - key: node-role.kubernetes.io/master
+            operator: Exists
+  tolerations:
+  - key: node-role.kubernetes.io/control-plane
+    operator: Exists
+  - key: node-role.kubernetes.io/master
+    operator: Exist
```
rifelpet, line 51:
```suggestion
    operator: Exists
```

pkg/nodelabels/builder.go
```diff
@@ -67,13 +67,17 @@ func BuildNodeLabels(cluster *kops.Cluster, instanceGroup *kops.InstanceGroup) m
...
+		}
 	} else {
 		if nodeLabels == nil {
 			nodeLabels = make(map[string]string)
 		}
 		nodeLabels[RoleLabelNode16] = ""
-		nodeLabels[RoleLabelName15] = RoleNodeLabelValue15
+		if cluster.IsKubernetesLT("1.24") {
+			nodeLabels[RoleLabelName15] = RoleNodeLabelValue15
+		}
```
rifelpet, line 80: Is kubeadm also removing the `kubernetes.io/role: node` node label?

pkg/model/components/clusterautoscaler.go
```diff
@@ -43,6 +43,8 @@ func (b *ClusterAutoscalerOptionsBuilder) BuildOptions(o interface{}) error {
 		v, err := util.ParseKubernetesVersion(clusterSpec.KubernetesVersion)
 		if err == nil {
 			switch v.Minor {
+			case 24:
+				image = "registry.k8s.io/autoscaling/cluster-autoscaler:v1.23.0"
```
hakman, line 47: This should probably go separately.

pkg/nodelabels/builder.go
olemarkus: I don't think kubeadm used that label. https://github.com/kubernetes/kubernetes/pull/107533 does not mention it at all.

pkg/model/components/clusterautoscaler.go
olemarkus: Agree. https://github.com/kubernetes/kops/pull/13510

rifelpet:
this needs updating:

https://github.com/kubernetes/kops/blob/340ff24468565644c3ba5cf2859cdbb55d97a95d/pkg/validation/validate_cluster.go#L184-L201

pkg/nodelabels/builder.go
rifelpet: yea the only reference I can find upstream is this function which handles the `node-role.kubernetes.io/` case as well:

https://github.com/kubernetes/kubernetes/blob/4cdeab4696e86c9738d99e2650d2cdfdfb0f8d32/pkg/printers/internalversion/printers.go#L1653-L1671

olemarkus: Note that we have deprecated those labels explicitly so removing these is expected.

docs/releases/1.24-NOTES.md
olemarkus: Thanks for catching.

olemarkus:
> this needs updating:
> 
> https://github.com/kubernetes/kops/blob/340ff24468565644c3ba5cf2859cdbb55d97a95d/pkg/validation/validate_cluster.go#L184-L201

Good catch. Fixed. I guess we'll find more things like this. Also need to purge the code itself for `master` once this goes in.

PR #13454, opened 2022-04-03
Enable etcd corruption check as mitigatio of 3.5 corruption issue

olemarkus:


pkg/model/components/etcdmanager/options.go
```diff
@@ -49,27 +49,50 @@ func (b *EtcdManagerOptionsBuilder) BuildOptions(o interface{}) error {
...
+		if !etcdVersionIsSupported(version) {
 			if featureflag.SkipEtcdVersionCheck.Enabled() {
 				klog.Warningf("etcd version %q is not known to be supported, but ignoring because of SkipEtcdVersionCheck feature flag", etcdCluster.Version)
 			} else {
 				klog.Warningf("unsupported etcd version %q detected; please update etcd version.  Use export KOPS_FEATURE_FLAGS=SkipEtcdVersionCheck to override this check", etcdCluster.Version)
 				return fmt.Errorf("etcd version %q is not supported with etcd-manager, please specify a supported version or remove the value to use the default version.  Supported versions: %s", etcdCluster.Version, strings.Join(supportedEtcdVersions, ", "))
 			}
 		}
-	}
+		for _, s := range []string{"3.5.0", "3.5.1"} {
```
justinsb, line 62: I think it's 3.5.2 also... https://groups.google.com/a/kubernetes.io/g/dev/c/B7gJs88XtQc/m/rSgNOzV2BwAJ?utm_medium=email&utm_source=footer

justinsb:
Thanks for doing this - I think we need to add 3.5.2 also (unless you've heard otherwise?), but I agree this is a good thing to add.

Do you want to add 3.5.2 and remove hold?

Otherwise

/approve
/lgtm
/hold

olemarkus:
This list of etcd versions supported on the following line did not include 3.5.2, which is why I did not add it to this list either.
https://github.com/kubernetes/kops/pull/13454/files#diff-8d69c7aefecf6d6f5ac91444d56393c60d3b5937bef954f02555b68b288b7275L65

PR #13457, opened 2022-04-04
allow ecr pull-through-cache

### LINKED ISSUE #13456: Support ECR Pull through cache policy
/kind feature

- aws ecr now supporting [pull through cache repositories](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-ecr-cache-repositories/)
- For using upstream images via nodes we need to add additional IAM policy.
-- At least `ecr:BatchImportUpstreamImage` to retrieve the external image & add it to the private registry.

karancode:
- Adds `ecr:BatchImportUpstreamImage` to the nodes IAM. fixes #13456

### DIFF
--- a/docs/iam_roles.md
+++ b/docs/iam_roles.md
@@ -28,6 +28,7 @@ The additional permissions are:
   "Action": [
     "ecr:BatchCheckLayerAvailability",
     "ecr:BatchGetImage",
+    "ecr:BatchImportUpstreamImage",
     "ecr:DescribeRepositories",
     "ecr:GetAuthorizationToken",
     "ecr:GetDownloadUrlForLayer",
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -806,6 +806,7 @@ func addECRPermissions(p *Policy) {
 		"ecr:DescribeRepositories",
 		"ecr:ListImages",
 		"ecr:BatchGetImage",
+		"ecr:BatchImportUpstreamImage",
 	)
 }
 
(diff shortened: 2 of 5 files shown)


olemarkus:
Can you add some documentation on how to make use of this feature? I assume one also need to configure mirrors for containerd.

karancode:
Hi @olemarkus thanks for the review.
> Can you add some documentation on how to make use of this feature?
If anyone is using the aws ecr pull through cache then they need to create cache rules and for k8s(nodes) to be able to use those rules, they just need this additional iam policy. There is no other special config required. Using pull through cache, quay/amzn-ecr-public repos can be pulled directly just by using the ecr endpoint `aws_account_id.dkr.ecr.region.amazonaws.com/quay/repository_name/image_name:tag`

Let me know if I should add above to the readme?

> I assume one also need to configure mirrors for containerd.
Sorry I am not sure about this. Let me check.

olemarkus:
> Hi @olemarkus thanks for the review.
> 
> > Can you add some documentation on how to make use of this feature?
> 
> If anyone is using the aws ecr pull through cache then they need to create cache rules and for k8s(nodes) to be able to use those rules, they just need this additional iam policy. There is no other special config required. Using pull through cache, quay/amzn-ecr-public repos can be pulled directly just by using the ecr endpoint `aws_account_id.dkr.ecr.region.amazonaws.com/quay/repository_name/image_name:tag`
> 
> Let me know if I should add above to the readme?
> 

So if I understand this correctly, this is only useful for your own images if you explicitly change the image name? You cannot use it as a normal containerd image? Where the Pod would have the original image specified, but containerd would try to pull it from the given endpoint?

karancode:
@olemarkus sorry for late. So basically, it is just for pulling an external image and caching it our private ECR repo. I think it will be the same for docker/containerd. 🤔

ScOut3R:
@olemarkus I suspect this should work with specifying `aws_account_id.dkr.ecr.region.amazonaws.com/quay` as a registry mirror for containerd.

I tried it for Docker Hub hosted image (`aws_account_id.dkr.ecr.region.amazonaws.com/docker`) and kubelet tried to pull images from the correct URL but I was getting a 401 error. I suspect the said IAM permission is needed on the nodes.

Issue #13465, opened 2022-04-07
Bump AWS Node termination handler to 1.16 to support excludeLoadBalancers


DingGGu:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
https://github.com/aws/aws-node-termination-handler/pull/582/files
AWS Node termination handler 1.16 support `node.kubernetes.io/exclude-from-external-load-balancers` label during termination flows.

This will increase stability during ec2 terminate from load balacners.

PR #13467, opened 2022-04-07
Bump NTH 1.16 and add excludeFromLoadBalancers option

DingGGu:
- Bump NTH 1.16 and add excludeFromLoadBalancers option #13465

### DIFF
--- a/pkg/model/components/nodeterminationhandler.go
+++ b/pkg/model/components/nodeterminationhandler.go
@@ -51,9 +51,6 @@ func (b *NodeTerminationHandlerOptionsBuilder) BuildOptions(o interface{}) error
 	if nth.EnableRebalanceDraining == nil {
 		nth.EnableRebalanceDraining = fi.Bool(false)
 	}
-	if nth.ExcludeFromLoadBalancers == nil {
-		nth.ExcludeFromLoadBalancers = fi.Bool(false)
-	}
 
 	if nth.EnablePrometheusMetrics == nil {
 		nth.EnablePrometheusMetrics = fi.Bool(false)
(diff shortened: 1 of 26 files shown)


upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
```diff
@@ -183,6 +197,8 @@ spec:
               value: "false"
             - name: ENABLE_SQS_TERMINATION_DRAINING
               value: "true"
+            - name: EXCLUDE_FROM_LOAD_BALANCERS
+              value: "{{ .ExcludeFromLoadBalancers }}"
```
olemarkus, line 201: You need to use `WithDefaultBool` here since `ExcludeFromLoadBalancers` is a pointer and go templates will evaluate pointers as non-nil or not, also for boolean pointers.

pkg/model/components/nodeterminationhandler.go
```diff
@@ -51,6 +51,9 @@ func (b *NodeTerminationHandlerOptionsBuilder) BuildOptions(o interface{}) error
 	if nth.EnableRebalanceDraining == nil {
 		nth.EnableRebalanceDraining = fi.Bool(false)
 	}
+	if nth.ExcludeFromLoadBalancers == nil {
+		nth.ExcludeFromLoadBalancers = fi.Bool(false)
```
olemarkus, line 55: Any downsides to enabling this by default?

olemarkus (changes requested):

pkg/model/components/nodeterminationhandler.go
DingGGu: NTH defaults to false, but in kOps, we can set the default value to true because there is nothing bad effect with it.

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
DingGGu: Okay, I will try to use WithDefaultBool for all nth manifest.

hakman:
@DingGGu You may have to run `make apimachinery` and `hack/update-expected.sh` and commit the changes to get the tests passing.

DingGGu:
maybe need /ok-to-test for this PR

olemarkus (approved):

PR #13487, opened 2022-04-12
Bump Ubuntu AMIs in alpha and stable

olemarkus:


hakman:
@olemarkus I think the AMIs are safe to bump in both alpha and stable, as long as they are tested first in periodic tests.

PR #13491, opened 2022-04-13
Use aws_s3_object instead of deprecated aws_s3_bucket_object

hakman:
```
│ Warning: Argument is deprecated
│ 
│   with aws_s3_bucket_object.cluster-completed-spec,
│   on kubernetes.tf line 563, in resource "aws_s3_bucket_object" "cluster-completed-spec":
│  563:   bucket                 = "testingBucket"
│ 
│ Use the aws_s3_object resource instead
│ 
│ (and 37 more similar warnings elsewhere)
╵
Success! The configuration is valid, but there were some
validation warnings as shown above.
```
/cc @rifelpet


hakman:
For minimum version we are ok, as the new resource name was added in `3.52.0` and we require `>= 3.71.0`.
https://github.com/hashicorp/terraform-provider-aws/issues/9903#event-5081798638

rifelpet:
The changelog suggests it was added in 4.0.0

https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#400-february-10-2022

The 3.52.0 changelog doesn't mention aws_s3_object

hakman:
@rifelpet I think should be ok for the final check.

PR #13515, opened 2022-04-19
Update aws-sdk-go to v1.43.41

hakman:


rifelpet (approved):
This could be done with dependabot now, right?

PR #13530, opened 2022-04-20
Revert "docs for release process shouldn't assume remotes"

hakman:
Reverts kubernetes/kops#13295

/cc @olemarkus @justinsb


hakman:
Hmm... I thought this would fix creating release PRs with the correct base branch, not master always.
Seems it's not the case, so it can be closed.

Issue #13540, opened 2022-04-21
Issue with creating a new cluster


abssirohi90:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Version 1.23.0

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.5", GitCommit:"5c99e2ac2ff9a3c549d9ca665e7bc05a3e18f07e", GitTreeState:"clean", BuildDate:"2021-12-16T08:38:33Z", GoVersion:"go1.16.12", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:52:18Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

kops create -f cluster.yaml
kops create -f ig.yaml
kops update cluster
kops validate cluster

**5. What happened after the commands executed?**

The cluster was created, new EC2 instances came up. 
Upon running validate cluster, we are getting:
	machine "i-xxx" has not yet joined cluster
Machine	i-xxx					machine "i-xxx" has not yet joined cluster
Machine	i-xxx					machine "i-xxx" has not yet joined cluster
Machine	i-xxx					machine "i-xxx" has not yet joined cluster
Machine	i-xxx					machine "i-xxx" has not yet joined cluster

**6. What did you expect to happen?**

Cluster to be created properly, kops validate to validate the cluster as up and running.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  generation: 22
  name: xxx
spec:
  additionalPolicies:
    master: |
      [
        {
          "Effect": "Allow",
          "Action": [
            "elasticloadbalancing:*"
          ],
          "Resource": [
            "*"
          ]
        },
        {
          "Effect": "Allow",
          "Action": [
[...]

Issue #13492, opened 2022-04-13
Master nodes on AWS not joining the cluster after upgrade to kops 1.23.0


aharonh:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

1.23.0

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

1.21.8

**3. What cloud provider are you using?**

aws

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

```
kops upgrade cluster test.com --yes
kops update cluster test.com --yes
kops rolling-update cluster test.com --instance-group master-us-east-1b --yes
```

**5. What happened after the commands executed?**

the cluster did not pass the validation, the replaced upgraded node did not join back the cluster

**6. What did you expect to happen?**

the upgraded node joins the cluster

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2022-02-08T13:30:13Z"
  generation: 78
  name: test.com
spec:
  DisableSubnetTags: true
  additionalPolicies:
    master: |
      [
        {
          "Effect": "Allow",
          "Action": ["sts:AssumeRole"],
          "Resource": "*"
        },
        {
          "Effect": "Allow",
          "Action": [
              "route53:ChangeResourceRecordSets",
              "route53:ListResourceRecordSets",
              "route53:GetHostedZone"
          ],
          "Resource": [
              "arn:aws:route53:::hostedzone/Z03047592P43JSA8N9YP2"
          ]
        }
      ]
    node: |
      [
        {
          "Effect": "Allow",
          "Action": ["sts:AssumeRole"],
          "Resource": "*"
        }
      ]
  api:
    loadBalancer:
      class: Classic
      type: Internal
      useForInternalApi: true
  authentication:
    aws: {}
  authorization:
    rbac: {}
  channel: stable
  cloudConfig:
    awsEBSCSIDriver:
      enabled: false
  cloudLabels:
    env: dev
  cloudProvider: aws
  clusterAutoscaler:
    awsUseStaticInstanceList: false
    balanceSimilarNodeGroups: true
    enabled: true
    expander: random
[...]

PR #13542, opened 2022-04-21
Add relnote about breaking change with custom DHCP opts

olemarkus:
Been quite a few reports on this: #13540 #13492 and #13358

Alternative would be to revert, but I think we are too far out into the 1.23 lifecycle to do this. It will also fix itself on kops/k8s 1.24.

### DIFF
--- a/docs/releases/1.23-NOTES.md
+++ b/docs/releases/1.23-NOTES.md
@@ -12,6 +12,32 @@ managed subnets will be configured to launch instances with Resource Based Names
 
 # Breaking changes
 
+## Custom domain names on AWS
+
+If your EC2 instances runs in a VPC that has custom domain name set in its DHCP options or custom resolve configuarion, instances will not be able to join the cluster. The reason for this is that the AWS
+cloud provider requires that instances use the default regional domain names. Before kOps 1.23, kOps would fetch the expected node name from the AWS EC2 API, but this
+caused scaling issues on large clusters and kOps now rely on the DNS configuration instead.
+
+As of Kubernetes 1.22, one can enable the [external cloud provider](https://cloud-provider-aws.sigs.k8s.io/), which supports Resource Based Naming.
+When the external cloud provider is used, kOps will automatically use instance IDs as node names instead of IP address based naming. These node
+names do not contain the domain, and therefore will work also with custom DHCP domain names.
+
+In order to enable the external cloud provider, add this to the cluster spec:
+
+```
+spec:
+  cloudControllerManager:
+    cloudProvider: aws
+  cloudConfig:
+    awsEBSCSIDriver:
+    enabled: true
+  kubernetesVersion: v1.23.1 #Must be 1.22.0 or higher
+```
+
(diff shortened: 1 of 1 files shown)


hakman:
I would vote for adding back the EC2/Metadata API if possible, with option to use DNS instead?

olemarkus:
I am ambivalent. An approach would be to use the current approach when CCM is used, and the old approach if not.

hakman:
CCM is pretty much always used and soon will be the only option.
Complicating things for operators with small clusters is not exactly ideal. The ones with bigger clusters are more used to reading release notes and custom configs.

olemarkus:
I agree

docs/releases/1.23-NOTES.md
```diff
@@ -12,6 +12,32 @@ managed subnets will be configured to launch instances with Resource Based Names
 
 # Breaking changes
 
+## Custom domain names on AWS
+
+If your EC2 instances runs in a VPC that has custom domain name set in its DHCP options or custom resolve configuarion, instances will not be able to join the cluster. The reason for this is that the AWS
+cloud provider requires that instances use the default regional domain names. Before kOps 1.23, kOps would fetch the expected node name from the AWS EC2 API, but this
+caused scaling issues on large clusters and kOps now rely on the DNS configuration instead.
```
seh, line 19:
```suggestion
caused scaling issues on large clusters and kOps now relies on the DNS configuration instead.
```

docs/releases/1.23-NOTES.md
```diff
@@ -12,6 +12,32 @@ managed subnets will be configured to launch instances with Resource Based Names
...
+
+As of Kubernetes 1.22, one can enable the [external cloud provider](https://cloud-provider-aws.sigs.k8s.io/), which supports Resource Based Naming.
```
seh, line 21:
```suggestion
As of Kubernetes 1.22, one can enable the [external cloud provider](https://cloud-provider-aws.sigs.k8s.io/), which supports Resource-Based Naming.
```

docs/releases/1.23-NOTES.md
```diff
@@ -12,6 +12,32 @@ managed subnets will be configured to launch instances with Resource Based Names
...
+When the external cloud provider is used, kOps will automatically use instance IDs as node names instead of IP address based naming. These node
```
seh, line 22:
```suggestion
When the external cloud provider is used, kOps will automatically use instance IDs as node names instead of IP address-based naming. These node
```

docs/releases/1.23-NOTES.md
```diff
@@ -12,6 +12,32 @@ managed subnets will be configured to launch instances with Resource Based Names
 
 # Breaking changes
 
+## Custom domain names on AWS
+
+If your EC2 instances runs in a VPC that has custom domain name set in its DHCP options or custom resolve configuarion, instances will not be able to join the cluster. The reason for this is that the AWS
```
robinverduijn, line 17: `configuarion` --> `configuration`

PR #13547, opened 2022-04-22
Upgrade Karpenter to 0.10.0

rifelpet:
this job timed out, seeing if an upgrade will help: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-misc-karpenter/1517300168158875648

Fixes a few bugs in the manifest and updates it to more closely match the rendered helm chart

### DIFF
--- a/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
+++ b/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
@@ -10,6 +10,8 @@ metadata:
 spec:
   group: karpenter.sh
   names:
+    categories:
+    - karpenter
     kind: Provisioner
     listKind: ProvisionerList
     plural: provisioners
@@ -40,6 +42,17 @@ spec:
               Node properties are determined from a combination of provisioner and
               pod scheduling constraints.
             properties:
+              kubeletConfiguration:
+                description: KubeletConfiguration are options passed to the kubelet
+                  when provisioning nodes
+                properties:
+                  clusterDNS:
+                    description: clusterDNS is a list of IP addresses for the cluster
+                      DNS server. Note that not all providers may use all addresses.
+                    items:
+                      type: string
+                    type: array
+                type: object
               labels:
                 additionalProperties:
                   type: string
@@ -246,7 +259,7 @@ data:
         "thereafter": 100
       },
       "outputPaths": ["stdout"],
-    "errorOutputPaths": ["stderr"],
+      "errorOutputPaths": ["stderr"],
       "encoding": "console",
       "encoderConfig": {
         "timeKey": "time",
@@ -260,8 +273,8 @@ data:
       }
     }
   # Log level overrides
-  # loglevel.controller: info # debug
-  # loglevel.webhook: info # debug
+  loglevel.controller: debug
+  loglevel.webhook: debug
 ---
 # Source: karpenter/templates/controller/rbac.yaml
 apiVersion: rbac.authorization.k8s.io/v1
@@ -270,17 +283,23 @@ metadata:
   name: karpenter-controller
 rules:
 - apiGroups: ["karpenter.sh"]
-  resources: ["provisioners", "provisioners/status"]
+  resources: ["provisioners"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["karpenter.sh"]
+  resources: ["provisioners/status"]
   verbs: ["create", "delete", "patch", "get", "list", "watch"]
-- apiGroups: ["coordination.k8s.io"]
-  resources: ["leases"]
-  verbs: ["create", "get", "patch", "update", "watch"]
+- apiGroups: [""]
+  resources: ["persistentvolumes", "persistentvolumeclaims"]
+  verbs: ["get", "list", "watch", "update"]
+- apiGroups: ["storage.k8s.io"]
+  resources: ["storageclasses"]
+  verbs: ["get", "list", "watch"]
 - apiGroups: [""]
   resources: ["nodes", "pods"]
   verbs: ["get", "list", "watch", "patch", "delete"]
 - apiGroups: [""]
-  resources: ["configmaps", "persistentvolumes", "persistentvolumeclaims"]
-  verbs: ["get", "list", "watch", "update"]
+  resources: ["configmaps"]
+  verbs: ["get", "list", "watch"]
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["create"]
@@ -354,7 +373,7 @@ metadata:
   namespace: kube-system
 rules:
 - apiGroups: [""]
-  resources: ["configmaps", "namespaces"]
+  resources: ["namespaces"]
   verbs: ["get", "list", "watch"]
 - apiGroups: [""]
   resources: ["secrets"]
@@ -402,7 +421,7 @@ metadata:
 spec:
   ports:
     - port: 8080
-      targetPort: metrics
+      targetPort: http-metrics
   selector:
     karpenter: controller
 ---
@@ -415,7 +434,7 @@ metadata:
 spec:
   ports:
     - port: 443
-      targetPort: webhook
+      targetPort: https-webhook
   selector:
     karpenter: webhook
 ---
@@ -436,6 +455,7 @@ spec:
     metadata:
       labels:
         karpenter: controller
+        app.kubernetes.io/name: karpenter
     spec:
       affinity:
         nodeAffinity:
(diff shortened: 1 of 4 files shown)


hakman:
```
==== START logs for container webhook of pod kube-system/karpenter-webhook-7584d59bc7-hn7km ====
panic: Unable to hydrate the AWS launch template cache, RequestCanceled: request context canceled
caused by: context canceled

goroutine 1 [running]:
github.com/aws/karpenter/pkg/cloudprovider/aws.(*LaunchTemplateProvider).hydrateCache(0xc00007e820, {0x27171f8, 0xc000745230})
	github.com/aws/karpenter/pkg/cloudprovider/aws/launchtemplate.go:229 +0x4f4
github.com/aws/karpenter/pkg/cloudprovider/aws.NewLaunchTemplateProvider({0x27171f8, 0xc000745230}, {0x27e9a60, 0xc0002069c8}, 0xc0000c2160, 0xc000206a08, 0xc000756560, 0x0)
	github.com/aws/karpenter/pkg/cloudprovider/aws/launchtemplate.go:70 +0x225
github.com/aws/karpenter/pkg/cloudprovider/aws.NewCloudProvider({0x27171f8, 0xc0006c3650}, {0xc0006d7200})
	github.com/aws/karpenter/pkg/cloudprovider/aws/cloudprovider.go:90 +0x5d3
github.com/aws/karpenter/pkg/cloudprovider/registry.newCloudProvider(...)
	github.com/aws/karpenter/pkg/cloudprovider/registry/aws.go:27
github.com/aws/karpenter/pkg/cloudprovider/registry.NewCloudProvider({0x27171f8, 0xc0006c3650}, {0x1f0b6a0})
	github.com/aws/karpenter/pkg/cloudprovider/registry/register.go:25 +0x28
main.main()
	github.com/aws/karpenter/cmd/webhook/main.go:55 +0x225
==== END logs for container webhook of pod kube-system/karpenter-webhook-7584d59bc7-hn7km ====
```

rifelpet:
looks like the livenessProbe is failing:

`I0422 04:13:51.762824    4557 prober.go:121] "Probe failed" probeType="Liveness" pod="kube-system/karpenter-webhook-7584d59bc7-hn7km" podUID=67b19fb4-8bc5-4fa9-8bc7-5ed3ae3eee26 containerName="webhook" probeResult=failure output="Get \"https://100.96.0.123:8443/\": dial tcp 100.96.0.123:8443: connect: connection refused"`

olemarkus:
Not sure the e2e will actually pass. Karpenter scheduling is not conformant, so test may fail because of that.

rifelpet:
Right now I can't even get the webhook pod to startup properly. It does some [cloud provider initialization before opening the listening port](https://github.com/aws/karpenter/blob/dd34701f0df2b44390c314fce82d6a8b5d526b53/cmd/webhook/main.go#L54-L63). From tracing the code it appears that the logging framework isn't initialized by that point either, so we don't have any of the expected [logging statements](https://github.com/aws/karpenter/blob/dd34701f0df2b44390c314fce82d6a8b5d526b53/pkg/cloudprovider/aws/cloudprovider.go#L83) in the cloud provider initialization either even though the stack trace points to a few lines later in that same function.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -728,6 +777,10 @@ spec:
   {{ range $key, $value := . }}
     {{ $key }}: "{{ $value }}"
   {{ end }}
+{{ end }}
+{{ if and $spec.Kubelet $spec.Kubelet.ClusterDNS }}
```
olemarkus, line 781: This is used by karpenter to configure kubelet on EKS through user data. It shouldn't apply to kOps.

tests/integration/update_cluster/karpenter/in-v1alpha2.yaml
```diff
@@ -116,4 +116,5 @@ spec:
   role: Node
   subnets:
   - us-test-1a
-
+  kubelet:
```
olemarkus, line 119: This is just for testing this propagates to the template?

olemarkus (changes requested):

olemarkus:
> Right now I can't even get the webhook pod to start up properly. It does some [cloud provider initialization before opening the listening port](https://github.com/aws/karpenter/blob/dd34701f0df2b44390c314fce82d6a8b5d526b53/cmd/webhook/main.go#L54-L63). From tracing the code it appears that the logging framework isn't initialized by that point either, so we don't have any of the expected [logging statements](https://github.com/aws/karpenter/blob/dd34701f0df2b44390c314fce82d6a8b5d526b53/pkg/cloudprovider/aws/cloudprovider.go#L83) in the cloud provider initialization either even though the stack trace points to a few lines later in that same function.

The webhook tries to fetch launch templates from AWS. I never gave karpenter that permission because we don't want Karpenter doing anything with Launch Templates. And now it panics over it.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
rifelpet: 👍🏻 removed

tests/integration/update_cluster/karpenter/in-v1alpha2.yaml
rifelpet: correct, I've removed it since we'll no longer recognize these fields

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -451,27 +471,30 @@ spec:
       serviceAccountName: karpenter
       dnsPolicy: Default
       containers:
-        - name: manager
-          image: public.ecr.aws/karpenter/controller:v0.6.0
-          resources: 
+        - name: controller
```
bwagner5, line 474: as of v0.6.2 of Karpenter, we're running the controller and webhook container in the same pod. Do you intentionally not want to do that in kOps?

bwagner5 (commented):
I'm not sure why things aren't working, I'm guessing there's some config mismatch since this is deviating a little from what we're publishing in the current manifests. Is there a `helm template` generation script you're using?

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
olemarkus: Hm. Usually webhooks are horisontally scalable, while controllers are leader elected. So typically they should be split. Any benefits to merging them?

olemarkus: We may as well merge since both use the same serviceaccount as well. Webhook shouldn't really have access to the AWS API or anything like that. I doubt that is what is causing this issue though.

dewjam:
Seems like the issue is with retrieving the logging ConfigMap.  In my testing, the webhook can handle cases where the ConfigMap doesn't exist (`IsNotFound`) but panics for other errors.  So perhaps there is a connectivity issue with APIServer during the test?

https://github.com/knative/pkg/blob/d016976f25676949e77732ba4e8fe2af5090db2f/injection/sharedmain/main.go#L66-L81

https://github.com/knative/pkg/blob/d016976f25676949e77732ba4e8fe2af5090db2f/injection/sharedmain/main.go#L294

olemarkus:
There isn't any apiserver issues. And the controller can fetch config just fine. And since the two deployments share SA, there are no RBAC issues either.

dewjam:
>There isn't any apiserver issues. And the controller can fetch config just fine. And since the two deployments share SA, there are no RBAC issues either.

Hmm.  Am I looking at an old job then?  ([karpenter-webhook-5b8b7c9d8c-gw9hx](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/13547/pull-kops-e2e-aws-karpenter/1523818273144049664/artifacts/cluster-info/kube-system/karpenter-webhook-5b8b7c9d8c-gw9hx/logs.txt))

dewjam:
I noticed there is a CoreDNS pod in "Pending" state, which never becomes ready.  Is that expected?

olemarkus:
CoreDNS runs on worker nodes, and since karpenter hasn't created any, they are pending.
But that's a good call. I remember that for Karpenter controller, I added `dnsPolicy: Default` to avoid this, while the webhook did not need it. It may be that it does now.

olemarkus:
@rifelpet took the liberty to push that change to your branch.

🤞

rifelpet:
Looks like we're in better shape now. the webhook pod becomes Ready, the tests are now just failing because test pods aren't able to run

olemarkus:
It looks like these are errors due to not all nodes in the cluster being ready. It's expected due to karpenter IGs instance count being somewhat volatile. That's at least not a regression.

I think we can squash and merge.

olemarkus (approved):

PR #13548, opened 2022-04-22
Bump k8s deps to 1.24.0-rc.0

olemarkus:


hakman:
I think this should also be bumped:
https://github.com/kubernetes/kops/blob/39defc4d9179bce2197706fa72e17932eda033bc/Makefile#L38

rifelpet (approved):

PR #13554, opened 2022-04-24
WIP: Use Ubuntu 22.04 as the default image for Kubernetes 1.24+

hakman:
As agreed during the 22.04.2022 kOps office hours.

### DIFF
--- a/channels/alpha
+++ b/channels/alpha
@@ -61,11 +61,19 @@ spec:
     - name: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20220404
       providerID: aws
       architectureID: amd64
-      kubernetesVersion: ">=1.18.0"
+      kubernetesVersion: ">=1.18.0 <1.24.0"
     - name: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64-server-20220404
       providerID: aws
       architectureID: arm64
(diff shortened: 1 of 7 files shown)


olemarkus:
I have started using 22.04 and it is not without teething trouble ... I suggest we at least add it to the test grid and run it for a while before making it the default. It is after all using nftables by default etc.

zetaab:
ubuntu 22 uses also cgroupv2 by default. We have seen issues with debian-11 earlier with oldest java projects. Old java versions do not support cgroup v2

hakman:
`nftables by default` - It's been ages since that appeared and it's supported since long time ago by k8s and most CNIs.
`cgroup v2` - Is probably a good point. 
`old java` - Not sure what we can do about workloads using older Java versions and other programming languages. For sure clusters could have a separate IG for such apps.

olemarkus:
We could mount v1 like we mount cgroups for cilium. But not sure this is on us. Can also be solved by daemonset or initcontainers. But they would need to be privileged.

olemarkus:
Seems like ec2-instance-connect is broken on jammy.
```
May 01 07:27:45 ip-172-20-56-12 sshd[7914]: AuthorizedKeysCommand /usr/share/ec2-instance-connect/eic_run_authorized_keys ubuntu SHA256:**** failed, status 2
```

PR #13555, opened 2022-04-24
Always set cluster-id flag for Protokube

hakman:
Getting the cluster ID from the cloud provider is just not needed.
NodeUp knows the cluster ID and can set it when configuring Protokube.


justinsb:
Nice observation & simplification!

/approve
/lgtm

PR #13559, opened 2022-04-26
Add support for Rocky Linux 8

hakman:
... as an alternative to CentOS 8, after its deprecation and removal.

/cc @olemarkus @rifelpet


rifelpet (approved):
should we add a section to the images docs?

PR #13566, opened 2022-04-28
Add NodeTerminationHandler CPU request settings

glebiller:
The NodeTerminationHandler addon currently has hard-coded CPU requests.
This PR adds a `CPURequest` property to the `CloudControllerManagerConfig` that controls the configured CPU requests for the cloud-controller-manager deployment. The default value has been set to the original `200m` value.

The settings works for AWS, GCP and OpenStack.


pkg/apis/kops/componentconfig.go
```diff
@@ -691,6 +691,10 @@ type CloudControllerManagerConfig struct {
 	UseServiceAccountCredentials *bool `json:"useServiceAccountCredentials,omitempty" flag:"use-service-account-credentials"`
 	// EnableLeaderMigration enables controller leader migration.
 	EnableLeaderMigration *bool `json:"enableLeaderMigration,omitempty" flag:"enable-leader-migration"`
+	// CPURequest of NodeTerminationHandler container.
+	// Default: 200m
+	// +k8s:conversion-gen=false
```
hakman, line 696: I am not familiar with this and don't see it used elsewhere. Would you mind explaining?

glebiller: I am not completely familiar with it too, to be honest.
I had trouble with the conversion from different API versions while adding the field.

The conversion-gen=false is used for other fields and prevents the conversion generator to try to convert from the old versions to the new API version. Since the value has defaulted in the template, having empty values for the old version is picking the default value of 200m.

If someone could point out a better solution for it, I would be happy to make the changes!

glebiller: Let me try that!

hakman (approved):
Thanks @glebiller. Looks good now!

PR #13567, opened 2022-04-28
Add a nameservers parameter for cert-manager.

jim-barber-he:
Provide a way to override the pod's list of DNS nameservers to use so
that split-view DNS zones still work for things like DNS01 challenges.
Without this the DNS TXT records are searched for in the private DNS
zone instead of the public one and the challenge will never succeed.

### DIFF
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -92,6 +92,22 @@ spec:
     managed: false
 ```
 
+##### DNS nameserver configuration for cert-manager pod
+{{ kops_feature_table(kops_added_default='1.23.3', k8s_min='1.16') }}
+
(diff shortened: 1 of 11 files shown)


docs/addons.md
```diff
@@ -92,6 +92,22 @@ spec:
     managed: false
 ```
 
+##### DNS nameserver configuration for cert-manager pod
+{{ kops_feature_table(kops_added_default='1.23.2', k8s_min='1.16') }}
```
hakman, line 96: Unfortunately, it missed the `1.23.2` release.
```suggestion
{{ kops_feature_table(kops_added_default='1.23.3', k8s_min='1.16') }}
```

olemarkus (approved):

Issue #8983, opened 2020-04-25
Hetzner Cloud Support


nekinie:
**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

Utilize kops to deploy and manage Kubernetes clusters on [Hetzner Cloud](https://www.hetzner.com/cloud), also known has hcloud.

**2. Feel free to provide a design supporting your feature request.**

Hetzner provide an official Go library: https://github.com/hetznercloud/hcloud-go

robertwe:
I would love to see it moving forward +1

nekinie:
I have some work ongoing for it in the background, will start pushing my work into a branch a little later today

JWDobken:
hee @nekinie what is the status of your work?

tomekit:
I would love to try out kops on Hetzner Cloud. Is this something in progress or ditched?

olemarkus:
I don't think there is much point in trying to reopen this. If one wants support, one would have to provide PRs.

bokysan:
Seems so, yeah. :-/

olemarkus:
What you can do is to ask Hetzner to sponsor this. That is what many of the other cloud providers has done. DO especially, as of late

hakman:
This is mostly done and available for testing and feedback. That can be done as separate issues if needed.

PR #13572, opened 2022-04-29
Add initial support for Hetzner Cloud

hakman:
Refs: #8983

/cc @justinsb


cmd/kops-controller/main.go
```diff
@@ -206,6 +207,12 @@ func addNodeController(mgr manager.Manager, opt *config.Options) error {
 			return fmt.Errorf("error building identifier: %v", err)
 		}
 
+	case "hetzner":
+		identifier, err = nodeidentityhetzner.New(opt.CacheNodeidentityInfo)
+		if err != nil {
+			return fmt.Errorf("error building identifier: %v", err)
```
justinsb, line 213: Nit: we should try to use %w for new code :-)

pkg/model/hetznermodel/firewall.go
```diff
@@ -0,0 +1,129 @@
...
+// ExternalAccessModelBuilder configures Firewall objects
+type ExternalAccessModelBuilder struct {
+	*HetznerModelContext
+	Lifecycle fi.Lifecycle
+}
+
+var _ fi.ModelBuilder = &ExternalAccessModelBuilder{}
+
+func (b *ExternalAccessModelBuilder) Build(c *fi.ModelBuilderContext) error {
+	//var sshAccess []net.IPNet
```
justinsb, line 39: Nit: Not sure why this code is commented out.   Is it because hetzner accepts strings?  Maybe add a todo or comment or remove?

pkg/model/hetznermodel/firewall.go
```diff
@@ -0,0 +1,129 @@
...
+				Protocol:  string(hcloud.FirewallRuleProtocolTCP),
+				Port:      fi.String("22"),
+			},
+		},
+		Labels: map[string]string{
+			hetzner.TagKubernetesClusterName:  b.ClusterName(),
+			hetzner.TagKubernetesFirewallRole: "control-plane",
+		},
+	}
+	if !b.UseLoadBalancerForAPI() {
```
justinsb, line 77: It would be cool to put this logic into `pkg/model` and have all the cloud providers share it!  But not a blocker

pkg/model/hetznermodel/network.go
```diff
@@ -0,0 +1,48 @@
...
+
+var _ fi.ModelBuilder = &NetworkModelBuilder{}
+
+func (b *NetworkModelBuilder) Build(c *fi.ModelBuilderContext) error {
+	network := &hetznertasks.Network{
+		Name:      fi.String(b.ClusterName()),
+		Lifecycle: b.Lifecycle,
+		IPRange:   fi.String(b.Cluster.Spec.NetworkCIDR),
+		Subnets: []string{
+			b.Cluster.Spec.NetworkCIDR,
```
justinsb, line 39: In future, maybe this should be subnets.  But I agree no need today, and that the sensible default is that the subnet CIDR should == the network CIDR.  If we support e.g. vswitch in future, users might have to specify subnet CIDRs to use that.

pkg/model/hetznermodel/servers.go
```diff
@@ -0,0 +1,89 @@
...
+		labels[hetzner.TagKubernetesInstanceRole] = string(ig.Spec.Role)
+
+		userData, err := b.BootstrapScriptBuilder.ResourceNodeUp(c, ig)
+		if err != nil {
+			return err
+		}
+
+		for i := 1; i <= int(igSize); i++ {
+			// hcloud-cloud-controller-manager requires hostname to be same as server name.
+			// This means server names should not contain the cluster name (which contains "." chars"
```
justinsb, line 65: You could maybe also replace the dots with dashes.  However, I like what you've done here, and it looks like we have the cluster name in the labels.

I guess one maybe-problem is if users were running two clusters in the same account, these names could collide?  You could add a random suffix if that's a concern (like k8s does for deployment pods)

pkg/model/hetznermodel/servers.go
```diff
@@ -0,0 +1,89 @@
...
+			// https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/f7d624e83c2c3475c5606306214814250922cb8a/hcloud/util.go#L39
+			name := ig.Name + "-" + strconv.Itoa(i)
+
+			server := hetznertasks.Server{
+				Name:      fi.String(name),
+				Lifecycle: b.Lifecycle,
+				SSHKey:    &hetznertasks.SSHKey{Name: fi.String(b.ClusterName())},
```
justinsb, line 72: Nit: Adding a helper method like LinkToSSHKey / LinkToNetwork here can avoid spreading the logic around too much about how sshkey / network is named.

pkg/model/hetznermodel/sshkey.go
```diff
@@ -0,0 +1,45 @@
...
+type SSHKeyModelBuilder struct {
+	*HetznerModelContext
+	Lifecycle fi.Lifecycle
+}
+
+var _ fi.ModelBuilder = &SSHKeyModelBuilder{}
+
+func (b *SSHKeyModelBuilder) Build(c *fi.ModelBuilderContext) error {
+	sshkey := &hetznertasks.SSHKey{
+		Name:      fi.String(b.ClusterName()),
```
justinsb, line 35: As a trick, you could also use `LinktoSSHKey().Name` here also.  That's a trick I've just started using, but I think I like it :-)

pkg/model/hetznermodel/sshkey.go
```diff
@@ -0,0 +1,45 @@
...
+		Lifecycle: b.Lifecycle,
+		PublicKey: string(b.SSHPublicKeys[0]),
```
justinsb, line 37: Looks like hetzner might support more than one SSH key?  Absolutely fine to add a TODO at this stage though (assuming it does support multiple).

pkg/nodeidentity/hetzner/identify.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	if !strings.HasPrefix(providerID, "hcloud://") {
+		return nil, fmt.Errorf("providerID %q not recognized for node %s", providerID, node.Name)
+	}
+
+	serverID := strings.TrimPrefix(providerID, "hcloud://")
+
+	// If caching is enabled try pulling nodeidentity.Info from cache before doing a Hetzner Cloud API call.
+	if i.cacheEnabled {
+		obj, exists, err := i.cache.GetByKey(serverID)
```
justinsb, line 80: Is this a unique identifier - it looks like it is a server number, rather than e.g. the name?  If so, great!

pkg/nodeidentity/hetzner/identify.go
```diff
@@ -0,0 +1,148 @@
...
+	return key, nil
+}
+
+// getServer queries Hetzner Cloud for the server with the specified ID, returning an error if not found
+func (i *nodeIdentifier) getServer(id string) (*hcloud.Server, error) {
+	serverID, err := strconv.Atoi(id)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert server ID %q to int: %w", id, err)
+	}
+	server, _, err := i.client.Server.GetByID(context.TODO(), serverID)
```
justinsb, line 142: Nit: probably a good idea to pass in the context, because it's relatively hard to retrofit.  Even if you the next function up you still need a `context.TODO()` :-)

pkg/resources/hetzner/resources.go
```diff
@@ -0,0 +1,293 @@
...
+	return resourceTrackers, nil
+}
+
+func listSSHKeys(cloud fi.Cloud, clusterName string) ([]*resources.Resource, error) {
+	c := cloud.(hetzner.HetznerCloud)
+	var resourceTrackers []*resources.Resource
+
+	sshKeys, err := c.GetSSHKeys(clusterName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list ssh keys: %v", err)
```
justinsb, line 73: Nit: %w

pkg/resources/hetzner/resources.go
```diff
@@ -0,0 +1,293 @@
...
+	return nil
+}
+
+func deleteNetwork(cloud fi.Cloud, r *resources.Resource) error {
+	klog.Infof("Deleting Network: %s(%s)", r.Name, r.ID)
+
+	c := cloud.(hetzner.HetznerCloud)
+	client := c.NetworkClient()
+	network := r.Obj.(*hcloud.Network)
+	_, err := client.Delete(context.TODO(), network)
```
justinsb, line 231: This makes me want to go and add ctx to all our delete methods :-)  I'm not going to force the merge-conflict on you though!

protokube/pkg/protokube/hetzner_volume.go
```diff
@@ -0,0 +1,119 @@
...
+
+// HetznerVolumes defines the Hetzner Cloud volume implementation.
+type HetznerVolumes struct {
+	hcloudClient *hcloud.Client
+	server       *hcloud.Server
+}
+
+func (h HetznerVolumes) AttachVolume(volume *Volume) error {
+	//TODO implement me
+	panic("implement me")
```
justinsb, line 42: I take it we don't need this because protokube doesn't attach volumes any more, because we're using etcd-manager.  If so, we could tweak the comment to say that.

protokube/pkg/protokube/hetzner_volume.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, fmt.Errorf("failed to get info for the running server: %s", err)
+	}
+
+	klog.V(4).Infof("Found name of the running server: %q", server.Name)
+	if server.Datacenter != nil && server.Datacenter.Location != nil {
+		klog.V(4).Infof("Found location of the running server: %q", server.Datacenter.Location.Name)
+	} else {
+		return nil, fmt.Errorf("failed to find location of the running server")
+	}
+	if len(server.PrivateNet) > 0 {
```
justinsb, line 85: Not a blocker, but looked like you might be able to get this from private-networks in metadata?  https://docs.hetzner.cloud/#server-metadata

Although I suppose you're getting the server anyway, though if that's only to get the cluster label I thought you had a PR to pass that in as a parameter...

upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -0,0 +1,523 @@
...
+}
+
+func (c *hetznerCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([]fi.ApiIngressStatus, error) {
+	var ingresses []fi.ApiIngressStatus
+	done, err := vfs.RetryWithBackoff(readBackoff, func() (bool, error) {
+		if cluster.Spec.MasterPublicName != "" {
+			// Note that this must match Digital Ocean's lb name
+			klog.V(2).Infof("Querying Hetzner to find Loadbalancers for API (%q)", cluster.Name)
+
+			loadBalancers, err := c.GetAllLoadBalancers()
```
justinsb, line 335: Nit: looks like we can filter by name https://docs.hetzner.cloud/#load-balancers-get-all-load-balancers

Maybe add a TODO?

upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -0,0 +1,523 @@
...
+func (c *hetznerCloudImplementation) FindClusterStatus(cluster *kops.Cluster) (*kops.ClusterStatus, error) {
+	////etcdStatus, err := findEtcdStatus(c, cluster)
+	////if err != nil {
+	////	return nil, err
+	////}
+	//status := &kops.ClusterStatus{
+	//	EtcdClusters: etcdStatus,
+	//}
+	//klog.V(2).Infof("Cluster status (from cloud): %v", fi.DebugAsJsonString(status))
+	status := &kops.ClusterStatus{}
```
justinsb, line 377: Nit: add a TODO?  Although I'm not sure how much we can do here.

upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -0,0 +1,523 @@
...
+func getCloudGroups(c *hetznerCloudImplementation, cluster *kops.Cluster, instancegroups []*kops.InstanceGroup, warnUnmatched bool, nodes []v1.Node) (map[string]*cloudinstances.CloudInstanceGroup, error) {
+	nodeMap := cloudinstances.GetNodeMap(nodes, cluster)
+
+	groups := make(map[string]*cloudinstances.CloudInstanceGroup)
+	instanceGroups, err := findInstanceGroups(c, cluster.ObjectMeta.Name)
+	if err != nil {
+		return nil, fmt.Errorf("unable to find autoscale groups: %v", err)
+	}
+
+	for _, doGroup := range instanceGroups {
```
justinsb, line 390: Nit: `doGroup` -> `ig`?

upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -0,0 +1,523 @@
...
+		name := doGroup.InstanceGroupName
+
+		instancegroup, err := matchInstanceGroup(name, cluster.ObjectMeta.Name, instancegroups)
+		if err != nil {
+			return nil, fmt.Errorf("error getting instance group for doGroup %q", name)
```
justinsb, line 395: Nit: doGroup -> HetznerInstanceGroup

upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -0,0 +1,523 @@
...
+		}
+	}
+
+	klog.V(8).Infof("Cloud Instance Group Info = %v", groups)
+	return groups, nil
+}
+
+// findInstanceGroups finds instance groups matching the specified tags
+func findInstanceGroups(c *hetznerCloudImplementation, clusterName string) ([]HetznerInstanceGroup, error) {
+	var result []HetznerInstanceGroup
```
justinsb, line 416: Nit: Add a TODO.  I think we can implement something like this by grouping instances by their instance-group label.  Maybe?

upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -0,0 +1,523 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return lbs, nil
+}
+
+func (c *hetznerCloudImplementation) GetAllVPCs() ([]*hcloud.Network, error) {
+	var allVPCs []*hcloud.Network
+
```
justinsb, line 508: Nit: TODO?

upup/pkg/fi/cloudup/hetznertasks/firewall.go
```diff
@@ -0,0 +1,165 @@
...
+	for _, firewall := range firewalls {
+		if firewall.Name == fi.StringValue(v.Name) {
+			matches := &Firewall{
+				Lifecycle: v.Lifecycle,
+				Name:      fi.String(firewall.Name),
+				ID:        firewall.ID,
+				Selector:  firewall.AppliedTo[0].LabelSelector.Selector,
+				Labels:    firewall.Labels,
+			}
+			for _, rule := range firewall.Rules {
```
justinsb, line 80: Ah, I see .. the API assumes a strict ordering.  This approach makes sense then, we might need to improve the tasks logic to cope with complex structures!

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -541,6 +543,29 @@ func setupZones(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.Stri
 		zoneToSubnetMap[region] = subnet
 		return zoneToSubnetMap, nil
 
+	//case api.CloudProviderHetzner:
```
justinsb, line 546: Nit: Add a TODO here?

justinsb (approved):

justinsb:
A few nits / suggestions, but I think this looks good and it does appear to be additive and feature-flagged, so I'm in favor of merging and iterating!

hakman:
> A few nits / suggestions, but I think this looks good and it does appear to be additive and feature-flagged, so I'm in favor of merging and iterating!

Thanks @justinsb. Will try to clean things up a little until tomorrow so that it can be merged.

pkg/model/hetznermodel/firewall.go
```diff
@@ -0,0 +1,129 @@
...
+
+	controlPlaneLabelSelector := []string{
+		fmt.Sprintf("%s=%s", hetzner.TagKubernetesClusterName, b.ClusterName()),
+		fmt.Sprintf("%s=%s", hetzner.TagKubernetesInstanceRole, string(kops.InstanceGroupRoleMaster)),
+	}
+	controlPlaneFirewall := &hetznertasks.Firewall{
+		Name:      fi.String("control-plane." + b.ClusterName()),
+		Lifecycle: b.Lifecycle,
+		Selector:  strings.Join(controlPlaneLabelSelector, ","),
+		Rules: []*hetznertasks.FirewallRule{
```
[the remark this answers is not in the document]
hakman: I just need a struct to describe the rules, but executor sees it as unknown. This is the only way I could find that works:
```go
var _ fi.HasDependencies = &FirewallRule{}
func (e *FirewallRule) GetDependencies(tasks map[string]fi.Task) []fi.Task {
 	return nil
}
```

pkg/model/hetznermodel/servers.go
hakman: Actually was thinking of adding the nodeup config hash.
Servers names must be unique per project, so one could use the same account for multiple clusters in different projects.

protokube/pkg/protokube/hetzner_volume.go
hakman: Server metadata is returned mostly unparsed by the lib so would have meant a little more work. We need to list all peers anyway later, so using the cloud API is still needed. Added a TODO to do it later.

hakman:
I think this is ready to merge now.
/cc @olemarkus @rifelpet

hakman:
The `pull-kops-e2e-cni-amazonvpc` is now optional and will continue failing until CCM is fixed.

nodeup/pkg/model/kube_apiserver.go
```diff
@@ -64,6 +64,16 @@ func (b *KubeAPIServerBuilder) Build(c *fi.ModelBuilderContext) error {
 		kubeAPIServer = *b.NodeupConfig.APIServerConfig.KubeAPIServer
 	}
 
+	if b.UsesSecondaryIP() {
```
rifelpet, line 67: This means we'll start setting --advertise-address for AWS clusters as well, correct? Do we need that set? If so could we do that in a separate PR?

hakman: It is needed, otherwise the advertised address remains the external IP, which is firewalled.
Changed the condition to `b.CloudProvider == kops.CloudProviderHetzner` so it only affects Hetzner for now.

rifelpet (approved):

PR #13580, opened 2022-04-30
Bump nvidia device plugin to 0.11 and nvidia driver to 510

olemarkus:
This will also make it possible to use gpu + arm. That combination requires jammy AMIs.


rifelpet:
should we add a release note and update the docs regarding the AMI requirement?

olemarkus:
Good idea. Added.

rifelpet (approved):

ddelange:
fyi nvidia has [stopped](https://hub.docker.com/r/nvidia/k8s-device-plugin) CD to docker hub and moved here:
`nvcr.io/nvidia/k8s-device-plugin:v0.12.0`

olemarkus:
@ddelange thanks. would really appreciate if you could do a PR updating this.

PR #13582, opened 2022-04-30
Revert containerd v1.6.3 upgrade

rifelpet:
https://github.com/containerd/containerd/issues/6876

will revisit once 1.6.4 is out


hakman:
Let's just wait for 1.6.4. It only affects some periodic tests for now.

hakman (approved):
I don't see any work towards a new release at the moment and we need to unblock testing with etcd 3.5.4.

PR #13586, opened 2022-05-01
Remove remaining bazel references

rifelpet:


olemarkus (approved):

docs/contributing/testing.md
```diff
@@ -49,7 +49,7 @@ kubetest2 kops \
   --test \
   --cloud-provider=aws \
   --cluster-name=my.testcluster.com \
-  --kops-binary-path=${KOPS_ROOT}/.bazel-bin/cmd/kops/linux-amd64/kops \
+  --kops-binary-path=${KOPS_ROOT}/.build/dist/$(go env GOOS)/$(go env GOARCH)/kops \
```
hakman, line 52: Not quite sure GOOS and GOARCH should be used here. Most likely the kOps binary should be `linux/amd64`.

hakman: Yes, but this assumes the test cluster GOOS & GOARCH would be the same as local one.

PR #13577, opened 2022-04-29
Don't install nfs-common/nfs-utils by default

hakman:
This adds too much overhead by default and also is a security risk (automatically start `rpcbind` on port 111).
Random ref: https://book.hacktricks.xyz/pentesting/pentesting-rpcbind

/cc @justinsb @rifelpet

PR #13590, opened 2022-05-03
Add support for installing additional packages

hakman:
Easy way to install additional packages. Should allow kOps to not install optional packages, like https://github.com/kubernetes/kops/pull/13577.

/cc @rifelpet @justinsb


justinsb:
This is OS specific, but the instancegroup is already OS specific because of the image.

/approve
/lgtm

hakman:
> This is OS specific, but the instancegroup is already OS specific because of the image.

Yup, this is why only per instancegroup and not at cluster level.

rifelpet:
Could we add a section in the docs that includes the kops_feature_table that mentions the version in which it's introduced

PR #13595, opened 2022-05-03
fix immutable retry logic when updating member

zetaab:


upup/pkg/fi/cloudup/openstack/loadbalancer.go
```diff
@@ -345,11 +345,10 @@ func updateMemberInPool(c OpenstackCloud, poolID string, memberID string, opts v
 				return true, nil
 			}
 			// pool is currently in immutable state, try to retry
-			if errCode, ok := err.(gophercloud.ErrUnexpectedResponseCode); ok {
-				if errCode.Actual == http.StatusConflict {
-					klog.Infof("got error %v retrying...", err)
-					return false, nil
-				}
+			errCode, ok := err.(gophercloud.ErrDefault409)
```
zetaab, line 348: for some reason the old logic did not work. We are seeing failed cluster updates because of this.

justinsb:
Thanks @zetaab ... it might be a good idea to add lots of logging here if this is problematic, but you're the expert on this code!

/approve
/lgtm

Issue #13541, opened 2022-04-21
pod annotation for AWS csi driver when using kube2iam


ashish1099:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
Currently a new cluster with 1.22 works fine, when requesting for a PVC from the `ebs.csi.aws.com` provisioner.
The problem comes when we deploy the kube2iam and it runs with a `namespace-restrictor` flag, which create two problem for us.

a. We cannot add the required annotation on the aws-ebs-csi pod 
b. we have to add the annotation in the kube-system ns (which I would like to avoid, although not IMP, but still)

**2. Feel free to provide a design supporting your feature request.**
Can we allow to add the annotation in the clusterspec `cloudConfig.awsEBSCSIDriver.controller.annotation` or something, I don't whatever key we use, just able to add it.

Also I have seen many requesting the same feature not exactly for the CSI, but for external-dns or some other app.
and the change is quite simple too, if you guys are fine, I can spend sometime to raise a PR

olemarkus:
Hey. There is precedence for adding `podAnnotations` to our addons, so go right ahead. Thanks.

ashish1099:
hmm,

I can see this `aws-ebs-csi-driver.addons.k8s.io` as an addon name, buts whats the spec and I see no example for any other addon where we can add it.

I dont see anywhere in here as well https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template

any pointer would be helpful :)

Thanks

rifelpet:
You can use these two PRs as reference:

https://github.com/kubernetes/kops/pull/13511
https://github.com/kubernetes/kops/pull/13536

ashish1099:
great, I did saw these two PRs, but again these PR has added an ability to add the annotation for nodelocaldns and [cluster-autoscaler](https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/v1alpha2/componentconfig.go#L995) and I guess I have to raise a PR to add support for aws-csi-driver here https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/v1alpha2/componentconfig.go#L901 and relevant files as well.

Let me know, If I'm on right track, I can find some time to raise the PR

rifelpet:
[...]

PR #13536, opened 2022-04-20
fix pod annotations in addon yamls

heybronson:
PodAnnotations in both cluster autoscaler and node-local-dns are failing.

Changing the yaml definitions to correctly unmarshal `map[string][string]` to valid annotations.

Testing:
```
❯ ~/go/bin/kops update cluster
...
  ManagedFile/bronson.cluster.doordash.cloud-addons-cluster-autoscaler.addons.k8s.io-k8s-1.15
        Contents
                                ...
                                          prometheus.io/port: "8085"
                                          prometheus.io/scrape: "true"
                                +         test1: test
                                +         test2: "123"
                                        creationTimestamp: null
                                        labels:
                                ...


  ManagedFile/bronson.cluster.doordash.cloud-addons-nodelocaldns.addons.k8s.io-k8s-1.12
        Contents
                                ...
                                          prometheus.io/port: "9253"
                                          prometheus.io/scrape: "true"
                                +         test1: test
                                +         test2: "123"
                                        creationTimestamp: null
                                        labels:
                                ...

❯ kd po node-local-dns-2qvz5
Name:                 node-local-dns-2qvz5
Namespace:            kube-system
Priority:             2000001000
Priority Class Name:  system-node-critical
Node:                 ip-172-20-38-152.us-west-2.compute.internal/172.20.38.152
Start Time:           Wed, 20 Apr 2022 13:53:15 -0700
Labels:               controller-revision-hash=7f58bd4969
                      k8s-app=node-local-dns
                      kops.k8s.io/managed-by=kops
                      pod-template-generation=4
Annotations:          prometheus.io/port: 9253
                      prometheus.io/scrape: true
                      test1: test
                      test2: 123

❯ kd po cluster-autoscaler-8495fb4f76-k25s2
Name:                 cluster-autoscaler-8495fb4f76-k25s2
Namespace:            kube-system
Priority:             2000000000
Priority Class Name:  system-cluster-critical
Node:                 ip-172-20-63-214.us-west-2.compute.internal/172.20.63.214
Start Time:           Wed, 20 Apr 2022 13:53:05 -0700
Labels:               app=cluster-autoscaler
                      app.kubernetes.io/name=cluster-autoscaler
                      k8s-addon=cluster-autoscaler.addons.k8s.io
                      k8s-app=cluster-autoscaler
                      kops.k8s.io/managed-by=kops
                      pod-template-hash=8495fb4f76
Annotations:          prometheus.io/port: 8085
                      prometheus.io/scrape: true
[...]

PR #13600, opened 2022-05-04
add annotation support to aws-ebs-csi pods

avish42:


ashish1099:
This will fix #13600

upup/models/cloudup/resources/addons/aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml.template
```diff
@@ -395,6 +395,10 @@ spec:
         app.kubernetes.io/name: aws-ebs-csi-driver
         app.kubernetes.io/instance: aws-ebs-csi-driver
         app.kubernetes.io/version: {{ .Version }}
+      annotations:
```
ashish1099, line 398: this should be `range` https://github.com/kubernetes/kops/pull/13536/files#diff-387221074177e2de294fd64375019fcfcae1519aebd3b97d2c9dc7ae9426a53fR131

avish42: ok

avish42:
@k8s-ci-robot /retest

KlavsKlavsen:
@olemarkus Can you approve above checks to run? (it says first time contributors needs approval to run checks - we need this support in kops - to solve #13541 :)

olemarkus:
This only add annotations to the controller, but not the daemonset. Maybe you want to distinguish between the two?

KlavsKlavsen:
@olemarkus Good catch. @avish42 will add it, so one can also set same annotation for daemonset. the AWS Helm chart for this, does exactly that - sets same annotation for both controllers and nodes - so IMHO the same is needed here - they both need to use the same role in AWS.

avish42:
@olemarkus can you please approve it now - if it's all okay

pkg/apis/kops/v1alpha3/componentconfig.go
```diff
@@ -886,6 +886,10 @@ type AWSEBSCSIDriver struct {
 	// If not specified, the value is approximated from the instance type.
 	// Default: -
 	VolumeAttachLimit *int `json:"volumeAttachLimit,omitempty"`
+
+	// PodAnnotations are the annotations added to aws-ebs-csi pods when they are created.
```
olemarkus, line 890:
```suggestion
	// PodAnnotations are the annotations added to AWS EBS CSI node and controller Pods.
```

olemarkus (changes requested):
Just want to make it more explicit that the annotation is added to _all_ pods. Then we are good, I think.

olemarkus (approved):

KlavsKlavsen:
@olemarkus those tests that fail - fail with:
```
 ok  	k8s.io/kops/util/pkg/vfs/tests	0.037s
FAIL
make: *** [Makefile:170: test] Error 1
Error: Process completed with exit code 2.
```

Not much detail about whats actually failing in this test? anyone who can enlighten us here :)

hakman:
@avish42 Would you mind squashing and rebasing the PR? Thanks!

hakman:
Until tests pass, to avoid noise, canceling /lgtm.
/lgtm cancel

avish42:
@hakman looks like unit tests are passing

hakman:
Thanks @avish42. Please also squash the commits.

PR #13603, opened 2022-05-04
update OpenStack CSI & cloud-controller manifest

zetaab:


zetaab:
@hakman / @olemarkus do you know do we have any way to delete old resources? Basically I should execute following to migrate to deployment

```
kubectl delete role external-resizer-cfg -n kube-system
kubectl delete rolebinding csi-resizer-role-cfg -n kube-system
kubectl delete sts csi-cinder-controllerplugin -n kube-system
```

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml.template
```diff
@@ -440,7 +425,7 @@ spec:
       hostNetwork: true
       containers:
         - name: node-driver-registrar
-          image: registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.1.0
```
olemarkus, line 443: registry.k8s.io supersedes k8s.gcr.io, so we probably should not change this.

olemarkus (changes requested):

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml.template
zetaab: ok I was just copying values from openstack manifests.. I can change these back and verify that images exists

zetaab:
@olemarkus changed all back to registry.k8s.io and verified that all images do exist

olemarkus (approved):

PR #13609, opened 2022-05-05
Add IRSA for kube-router

rifelpet:
[kube-router jobs are failing](https://testgrid.k8s.io/kops-network-plugins#kops-aws-cni-kuberouter) now that k8s 1.24 is released. the jobs use the release/stable k8s version marker, so they're now on 1.24 which kops defaults to enabling IRSA.

[kube-router pod logs](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-cni-kuberouter/1521995867320487936/artifacts/cluster-info/kube-system/kube-router-c69fn/logs.txt) show failed attempts to disable the src/dest check on ec2 instances:

`E0504 23:37:12.829945       1 aws.go:60] Node does not have necessary IAM creds to modify instance attribute. So skipping disabling src-dst check.`

Before 1.24, [pod logs](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-cni-kuberouter/1521392063394877440/artifacts/cluster-info/kube-system/kube-router-h6csr/logs.txt) confirm the src-dst check is able to be disabled:

`I0503 07:37:01.551179       1 aws.go:65] Disabled source destination check for the instance: i-08feeddb530837cd8`

This sets up IRSA for the kube-router service account to have the necessary permissions. Its likely we'll run into this issue with multiple CNIs so I kept the iambuilder function name generic.

### DIFF
--- a/pkg/model/components/addonmanifests/kuberouter/iam.go
+++ b/pkg/model/components/addonmanifests/kuberouter/iam.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kuberouter
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+
+	"k8s.io/kops/pkg/model/iam"
+)
+
+// ServiceAccount represents the service-account used by kube-router when we're using IRSA
+// It implements iam.Subject to get AWS IAM permissions.
+type ServiceAccount struct{}
+
+var _ iam.Subject = &ServiceAccount{}
+
+// BuildAWSPolicy generates a custom policy for a ServiceAccount IAM role.
+func (r *ServiceAccount) BuildAWSPolicy(b *iam.PolicyBuilder) (*iam.Policy, error) {
+	clusterName := b.Cluster.ObjectMeta.Name
+	p := iam.NewPolicy(clusterName, b.Partition)
+	iam.AddKubeRouterPermissions(b, p)
+	return p, nil
+}
+
+// ServiceAccount returns the kubernetes service account used.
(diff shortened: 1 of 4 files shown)


rifelpet:
kube-router success:

```
I0505 03:32:24.840609       1 aws.go:65] Disabled source destination check for the instance: i-08670cbcfe86d2403
I0505 03:32:26.247560       1 aws.go:65] Disabled source destination check for the instance: i-0fd241f223510cc44
I0505 03:32:27.933689       1 aws.go:65] Disabled source destination check for the instance: i-0f53063c97f382b83
I0505 03:32:29.405674       1 aws.go:65] Disabled source destination check for the instance: i-0f4e634395f117045
I0505 03:32:30.790558       1 aws.go:65] Disabled source destination check for the instance: i-058b97ddd45e896f2
```

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/13609/pull-kops-e2e-cni-kuberouter/1522053559615492096/artifacts/cluster-info/kube-system/kube-router-ztm4n/logs.txt

We'll have to figure out what to do about these VPC CNI tests. Those still use the host's instance profile's permissions - we havent migrated it to IRSA yet. Though it shouldn't need the source/destination check disabled since VPC CNI (I can't find a ModifyInstanceAttribute call in its codebase).

I'm also a bit surprised kube-router is the only CNI that is experiencing this issue. Maybe it occurs with other CNIs but only with certain settings?

hakman:
@rifelpet I am trying to fix AWS VPC CNI in https://github.com/kubernetes/kops/pull/13610.

rifelpet:
I see we actually already have a permissions builder for calico that does this:

https://github.com/kubernetes/kops/blob/2b58a87a5c1b3fe101c428e7ce0f231b78d5ec48/pkg/model/iam/iam_builder.go#L817-L822

so if we don't find any other CNIs that need this setup I may rename the function to be kuberouter specific.

pkg/model/components/addonmanifests/kuberouter/iam.go
```diff
@@ -0,0 +1,45 @@
...
+
+package kuberouter
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+
+	"k8s.io/kops/pkg/model/iam"
+)
+
+// ServiceAccount represents the service-account used by kube-router
```
justinsb, line 25: Nit: we might want to modify the comment to specify "when we're using IRSA" or something like that to indicate that we don't always use this.  OTOH we're going to always be using it going forwards, so this is just a nit...

justinsb:
One nit suggestion but LGTM

hakman (approved):

PR #13617, opened 2022-05-07
Avoid "/etc/resolv.conf" file loopback for Flatcar Container Linux distribution

seh:
In [Flatcar Container Linux versions at 2748.0](https://www.flatcar-linux.org/releases#release-2748.0.0) and later, configure the kubelet to use an alternate DNS resolver configuration. Rather than using the _/etc/resolv.conf_ file, use _systemd-resolved_'s _/run/systemd/resolve/resolv.conf_ file.

Without this configuration, pods like CoreDNS that use [the "Default" DNS policy](https://github.com/kubernetes/api/blob/5e2f5add2db54c6b8eec87c2e467b53d2409dbf3/core/v1/types.go#L2716-L2718) wind up with [a DNS forwarding loop](https://coredns.io/plugins/loop/#troubleshooting).

See flatcar-linux/Flatcar#285 for pertinent discussion, as well as [a thread in the "kops-users" channel](https://kubernetes.slack.com/archives/C3QUFP0QM/p1650573952888199) of the "Kubernetes" Slack workspace that arose from diagnosing failures in the CoreDNS containers.

### DIFF
--- a/util/pkg/distributions/distributions.go
+++ b/util/pkg/distributions/distributions.go
@@ -99,6 +99,9 @@ func (d *Distribution) HasLoopbackEtcResolvConf() bool {
 		// Ubuntu > 16.04 has it
 		return d.version > 16.04
 	}
+	if d.project == "flatcar" {
+		return true
+	}
 	return false
 }
 


util/pkg/distributions/distributions.go
```diff
@@ -99,6 +99,13 @@ func (d *Distribution) HasLoopbackEtcResolvConf() bool {
 		// Ubuntu > 16.04 has it
 		return d.version > 16.04
 	}
+	if d.project == "flatcar" {
+		// Before version 2748.0.0. the "/etc/resolv.conf" file did not point to the stub resolver
+		// configuration. Instead, it shared content (by way of a symbolic link) with systemd's
+		// "/run/systemd/resolve/resolv.conf" file, as opposed to the
+		// "/run/systemd/resolve/stub-resolv.conf" file. The latter includes the loopback address.
+		return d.version >= 2748.0
+	}
```
hakman, line 108: kOps does not set the flatcar version it finds, so comparing with `2748.0` will not help much.
Being a rolling release and `2748.0` is no longer listed in https://www.flatcar.org/releases, I think it's safe to consider this true for Flatcar in general.
Also, not sure if the long comment is needed considering the function comment.
```suggestion
	if d.IsFlatcar() == "flatcar" {
		// Before version 2748.0.0. the "/etc/resolv.conf" file did not point to the stub resolver
		// configuration. Instead, it shared content (by way of a symbolic link) with systemd's
		// "/run/systemd/resolve/resolv.conf" file, as opposed to the
		// "/run/systemd/resolve/stub-resolv.conf" file. The latter includes the loopback address.
		return true
	}
```

hakman (changes requested):

util/pkg/distributions/distributions.go
seh: I had originally written it as you suggest, then stopped to go figure out why it hadn't been necessary to have this change in place in much older versions of Flatcar. Thinking about it again, though, while it wasn't _necessary_ to use the alternate _resolv.conf_ file in those older versions, it would have worked to do so. In other words, using either of the candidate files worked equally well, as far as I can tell.

hakman (approved):

PR #13624, opened 2022-05-09
Allow returning multiple addresses for API

hakman:
Hetzner load balancers have internal IPv4, external IPv4 and external IPv6.
This PR makes it possible for tasks with `HasAddress` implemented to return multiple addresses via `FindAddresses()`.


justinsb:
Looks good - and looking at the code returning `[]string` actually feels more fluent than returning `*string`

/approve
/lgtm

PR #13636, opened 2022-05-11
Allow the AWS Load Balancer Controller to use WAFs

seh:
By introducing a few new fields within the _Cluster_ spec's "awsLoadBalancerController" field, allow users to enable the AWS Load Balancer Controller to associate WAFs with EC2 Application Load Balancers (ALBs). It's possible to enable separately use of two kinds of WAF: [_WAF Classic_](https://docs.aws.amazon.com/waf/latest/developerguide/classic-waf-chapter.html) and [the never version 2-era _WAF_](https://docs.aws.amazon.com/waf/latest/developerguide/waf-chapter.html), the latter of which bears no distinguishing name.

Retain our default configuration of the AWS Load Balancer Controller in which this capability remains disabled via command-line flags, overriding the controller program's enabling of this capability by default.

See @zetaab's #12074 for an earlier pass, and [a thread in the "kops-users" channel](https://kubernetes.slack.com/archives/C3QUFP0QM/p1652112294790899) of the "Kubernetes" Slack workspace for discussion of the current gaps in AWS's IAM system for restricting which ALBs a given principal can associate with or—more importantly—dissociate from which WAFs.

For now, we continue to disable this controller's integration with [AWS Shield](https://aws.amazon.com/shield).


pkg/apis/kops/componentconfig.go
```diff
@@ -1055,6 +1055,12 @@ type AWSLoadBalancerControllerConfig struct {
 	Enabled *bool `json:"enabled,omitempty"`
 	// Version is the container image tag used.
 	Version *string `json:"version,omitempty"`
+	// EnableUseOfWAFClassic specifies whether the controller can use WAFs (Classic Regional).
+	// Default: false
+	EnableUseOfWAFClassic *bool `json:"enableWAFClassic,omitempty"`
+	// EnableUseOfWAF specifies whether the controller can use WAFs (V2).
+	// Default: false
+	EnableUseOfWAF *bool `json:"enableWAF,omitempty"`
```
hakman, line 1063: I think we can drop the `UseOf` part.
Any special reason to use pointers?
```suggestion
	// EnableWAFClassic specifies whether the controller can use WAFs (Classic Regional).
	// Default: false
	EnableWAFClassic bool `json:"enableWAFClassic,omitempty"`
	// EnableWAF specifies whether the controller can use WAFs (V2).
	// Default: false
	EnableWAF bool `json:"enableWAF,omitempty"`
```

hakman (changes requested):

pkg/apis/kops/componentconfig.go
seh: I was following the example from #12074. I can use a `bool` directly. That would cut out the need for `fi.BoolValue` elsewhere.

hakman: `Enabled` is special use case. `bool` should be enough when default is `false`.

pkg/model/components/addonmanifests/awsloadbalancercontroller/iam.go
```diff
@@ -32,7 +33,17 @@ func (r *ServiceAccount) BuildAWSPolicy(b *iam.PolicyBuilder) (*iam.Policy, erro
 	clusterName := b.Cluster.ObjectMeta.Name
 	p := iam.NewPolicy(clusterName, b.Partition)
 
-	iam.AddAWSLoadbalancerControllerPermissions(p)
+	var enableUseOfWAFClassic bool
+	var enableUseOfWAF bool
+	if c := b.Cluster.Spec.AWSLoadBalancerController; c != nil {
+		if fi.BoolValue(c.EnableUseOfWAFClassic) {
+			enableUseOfWAFClassic = true
```
olemarkus, line 40: This is probably easier written as `enableUseOfWAFClasic := fi.BoolValue(c.EnableUseOfWAFClassic)`

upup/models/cloudup/resources/addons/aws-load-balancer-controller.addons.k8s.io/k8s-1.19.yaml.template
```diff
@@ -749,8 +749,8 @@ spec:
       containers:
       - args:
         - --cluster-name={{ ClusterName }}
-        - --enable-waf=false
-        - --enable-wafv2=false
+        - --enable-waf={{ WithDefaultBool .AWSLoadBalancerController.EnableUseOfWAFClassic false }}
```
olemarkus, line 752: I prefer that the spec field and flag match. Any reason they shouldn't?

olemarkus (changes requested):

upup/models/cloudup/resources/addons/aws-load-balancer-controller.addons.k8s.io/k8s-1.19.yaml.template
seh: I found that the flag names didn't match AWS' language about these services or resources (e.g. the `--enable-waf` flag controls use of WAF Classic), so I tried to recast the field names in terms that users would confront in the AWS documentation.

However, if you see the _Cluster_ spec as being a summary of configuration that the users would have written themselves in a manifest like this one, I can accept your preference as being the appropriate choice.

pkg/model/components/addonmanifests/awsloadbalancercontroller/iam.go
seh: Is that still the case even if these fields are now of type `bool` rather than `*bool`, per @hakman's earlier suggestion (https://github.com/kubernetes/kops/pull/13636#discussion_r870521607)?

upup/models/cloudup/resources/addons/aws-load-balancer-controller.addons.k8s.io/k8s-1.19.yaml.template
```diff
@@ -749,8 +749,8 @@ spec:
       containers:
       - args:
         - --cluster-name={{ ClusterName }}
-        - --enable-waf=false
-        - --enable-wafv2=false
+        - --enable-waf={{ WithDefaultBool .AWSLoadBalancerController.EnableWAF false }}
+        - --enable-wafv2={{ WithDefaultBool .AWSLoadBalancerController.EnableWAFv2 false }}
```
hakman, line 753: Is there a need for `WithDefaultBool` ?
```suggestion
        - --enable-waf={{ .AWSLoadBalancerController.EnableWAF }}
        - --enable-wafv2={{ .AWSLoadBalancerController.EnableWAFv2 }}
```

pkg/model/iam/iam_builder.go
```diff
@@ -955,8 +955,8 @@ func AddCCMPermissions(p *Policy, cloudRoutes bool) {
 	}
 }
 
-// AddAWSLoadbalancerControllerPermissions adds the permissions needed for the aws load balancer controller to the givnen policy
-func AddAWSLoadbalancerControllerPermissions(p *Policy) {
+// AddAWSLoadbalancerControllerPermissions adds the permissions needed for the AWS Load Balancer Controller to the givnen policy
+func AddAWSLoadbalancerControllerPermissions(p *Policy, enableUseOfWAFClassic bool, enableUseOfWAF bool) {
```
hakman, line 959:
```suggestion
func AddAWSLoadbalancerControllerPermissions(p *Policy, enableWAF bool, enableWAFv2 bool) {
```

upup/models/cloudup/resources/addons/aws-load-balancer-controller.addons.k8s.io/k8s-1.19.yaml.template
olemarkus: By convention, we try to match flags. It can get awkward when upstream is not consistent, but people tend to look for "how do I enable this flag" rather than "how do I enable this feature". In my humble experience.

upup/models/cloudup/resources/addons/aws-load-balancer-controller.addons.k8s.io/k8s-1.9.yaml.template
```diff
@@ -749,8 +749,8 @@ spec:
       containers:
       - args:
         - --cluster-name={{ ClusterName }}
-        - --enable-waf=false
-        - --enable-wafv2=false
+        - --enable-waf={{ .AWSLoadBalancerController.EnableWAF }}
+        - --enable-wafv2={{ .AWSLoadBalancerController.EnableWAFv2 }}
```
hakman, line 753: Not quite sure if these should apply to pre 1.19 clusters, but no strong opinions.

docs/addons.md
```diff
@@ -22,6 +22,33 @@ spec:
     enabled: true
 ```
 
+Though the AWS Load Balancer Controller can integrate the AWS WAF and
+Shield services with your Application Load Balancers (ALBs), kOps
+disables those capabilities by default.
+
+{{ kops_feature_table(kops_added_default='1.24) }}
```
hakman, line 29:
```suggestion
{{ kops_feature_table(kops_added_default='1.24') }}
```

olemarkus (approved):

PR #13641, opened 2022-05-12
Update documentation for manageStorageClasses

minkimipt:
`spec.cloudConfig.manageStorageClasses` is undocumented despite being
cloud-wide setting. Adding description for that option.


olemarkus (approved):
That's a good update. Thanks!

PR #13643, opened 2022-05-12
[Digital Ocean] Add CSI driver for DO block storage

srikiz:
Minor change to add CSI drivers for digital ocean. Templates added from https://github.com/digitalocean/csi-digitalocean

Tested the csi controller pods are created on cluster creation and a csi volume was provisioned as expected.
```
kube-system   csi-do-controller-0                           5/5     Running   0               3m27s   1***7    10.1**.0.7   <none>           <none>
kube-system   csi-do-node-kcj57                             2/2     Running   0               61s     10.12**.0.7       10.1**2.0.7   <none>           <none>
```

PVC mounted:
```
kubectl exec -it my-csi-app -- df -h | grep pvc-
/dev/disk/by-id/scsi-0DO_Volume_pvc-14bd9094-0bd0-48cf-bb59-aaf1a82f51a2
```

### DIFF
--- a/upup/models/cloudup/resources/addons/digitalocean-csi-driver.addons.k8s.io/k8s-1.22.yaml.template
+++ b/upup/models/cloudup/resources/addons/digitalocean-csi-driver.addons.k8s.io/k8s-1.22.yaml.template
@@ -1,0 +1,1245 @@
+# Copyright 2022 DigitalOcean
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


srikiz:
FYI - @timoreimann

hakman:
This is pretty exciting and looks good to me. Just wondering if it should have some toleration blocks like the cloud controller addon:
https://github.com/kubernetes/kops/blob/53af129f41548ffb6f2725f17e4aeb308bd098ec/upup/models/cloudup/resources/addons/digitalocean-cloud-controller.addons.k8s.io/k8s-1.8.yaml.template#L42-L59

timoreimann:
LGTM

Re: tolerations: those are critical in CCM since one of its responsibilities is completing the Node provisioning process (specifically around discovering internal / external IP addresses to be associated with a Node) and subsequent removal of taints, until which certain regular workloads won't be able to come up.

CSI is a bit different in the sense that it is not an essential component for bootstrapping. I'd be in favor of not adding more tolerations for now, also because the current upstream driver e2e tests don't rely on it either.

hakman:
Thanks @srikiz & @timoreimann.
/lgtm

zak905:
this is great! thanks @srikiz

PR #13645, opened 2022-05-13
Fix code block in manageStorageClasses docs

minkimipt:
When posting the initial PR https://github.com/kubernetes/kops/pull/13641 I didn't build the docs and didn't realise
that code block was not closed. Fixing it now.

Here's the broken doc:

image

Here's the fixed one after this PR:

image


justinsb:
Thanks for tracking it down and fixing it @minkimipt

/ok-to-test
/approve
/lgtm

PR #13649, opened 2022-05-14
Use Calico v3.21 with older versions of k8s

hakman:
/cc @olemarkus @rifelpet


upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template
```diff
@@ -4692,7 +4897,7 @@ metadata:
 
 # This manifest creates a Pod Disruption Budget for Controller to allow K8s Cluster Autoscaler to evict
 
-apiVersion: policy/v1beta1
+apiVersion: policy/v1
```
olemarkus: policy/v1 is 1.21+ so this change shouldn't go into the 1.16 manifest.

hakman: This is actually reverted in the end, or it should be.

olemarkus (approved):

Issue #13614, opened 2022-05-06
Use kops upgrade for patch level and ig image updates


TwoStone:
/kind feature

**1. Problem**
It would be great to have the possibility to use kops upgrade cluster to upgrade a cluster to the latest patch level of the currently used kubernetes version and also the instance group images.
As you recommend to always use the latest kops version for working with the cluster, a kops upgrade will always try to upgrade to kubernetes version to the latest from the channel.

Our use case is, that we need some time before we can upgrade to newer minor versions of kubernetes as we have many resources to check and adjust for deprecated apis and so on before we can do a minor version update of k8s so we normally are lagging 1 or 2 minor versions behind the latest. But we have to do patch level and ami image updates on a more regular basis.
At the moment we have two options to do these:

1. Do not use the latest kops version for `kops upgrade cluster`, instead use the latest version corresponding to the k8s version
2. Lookup the latest patch level version and ami image in the stable channel yaml and edit the cluster spec manually

**2. Proposed solution**
To solve this problem, you could introduce a flag for the `upgrade cluster` command to only upgrade to the latest k8s version within the current minor version from the cluster spec. i.E. `--patch-level-only=true`
Another option would be to add a flag to the upgrade cluster command to specify the k8s minor version used to lookup the latest version. i.E. `--kubernetes-version=1.21`

Please tell me what do you think about this. 
If you find one of this a useful addition to kops, I would like to create a pr for this.

hakman:
@TwoStone Any thoughts about the #13652 approach? That should satisfy the minimum requirements. If you think minor version should be accepted also, sounds good to me and I would be happy to review such PR.

TwoStone:
Thank you @hakman for the implementation. I think providing the full k8s version is fine for our use case.

hakman:
👍

PR #13652, opened 2022-05-15
Allow overriding the kubernetes version when upgrading the cluster

hakman:
This will allow setting the exact (patch) k8s version when upgrading a cluster:
```sh
$ kops upgrade cluster --kubernetes-version=1.23.6
```

Fixes: #13614

### DIFF
--- a/cmd/kops/upgrade_cluster.go
+++ b/cmd/kops/upgrade_cluster.go
@@ -58,6 +58,8 @@ type UpgradeClusterOptions struct {
 	ClusterName string
 	Yes         bool
 	Channel     string
+	// KubernetesVersion is the k8s version to use for upgrade.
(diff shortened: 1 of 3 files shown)


olemarkus:
I interpreted the feature request to be about specifying minor version and then use the latest patch version from channels.

hakman:
> I interpreted the feature request to be about specifying minor version and then use the latest patch version from channels.

This is not even possible with `create cluster`.
Also, this is why the PR only refs the issue.

olemarkus:
/lgtm

Should we add a release note?

hakman:
> Should we add a release note?

Sounds good to me. Done.

olemarkus (approved):

PR #13663, opened 2022-05-18
Cilium: Add node rbac permissions for cilium-operator

ReillyBrogan:
Just updated one of our dev clusters to Cilium 1.11.5 and noticed the following error in the cilium-operator logs:
```
level=warning msg="github.com/cilium/cilium/operator/watchers/node.go:78: failed to list *v1.Node: nodes is forbidden: User \"system:serviceaccount:kube-system:cilium-operator\" cannot list resource \"nodes\" in API group \"\" at the cluster scope" subsys=klog
level=error msg=k8sError error="github.com/cilium/cilium/operator/watchers/node.go:78: Failed to watch *v1.Node: failed to list *v1.Node: nodes is forbidden: User \"system:serviceaccount:kube-system:cilium-operator\" cannot list resource \"nodes\" in API group \"\" at the cluster scope" subsys=k8s
```
This adds the required rbac permissions as shown in https://github.com/cilium/cilium/blob/0e9c6c6c6bb29e86c88a95bc01c2f527a0bb2faf/install/kubernetes/cilium/templates/cilium-operator/clusterrole.yaml

PR #13666, opened 2022-05-18
Bump Cilium to 1.11.5

olemarkus:
Since this introduced some backwards breaking RBAC changes, the manifest got forked

### DIFF
--- a/pkg/model/components/cilium.go
+++ b/pkg/model/components/cilium.go
@@ -40,7 +40,7 @@ func (b *CiliumOptionsBuilder) BuildOptions(o interface{}) error {
 	}
 
 	if c.Version == "" {
-		c.Version = "v1.11.4"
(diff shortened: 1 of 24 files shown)


ReillyBrogan:
This, or at least the RBAC changes, should be cherry-picked back to supported branches that otherwise support Cilium 1.11.

olemarkus:
This change is a bit too big for the 1.23 branch. There is also no new 1.23 version planned in the near future, so 1.24 may very will arrive first.

ReillyBrogan:
> This change is a bit too big for the 1.23 branch. There is also no new 1.23 version planned in the near future, so 1.24 may very will arrive first.

Yes, the full changeset is a bit too big, however you could cherry-pick just the additive RBAC changes (IE #13663) as those would be enough to fix Cilium errors but wouldn't break previous versions.

PR #13678, opened 2022-05-22
Add Support for OVN Load Balancer

ching-kuo:
OVN load balancer in OpenStack only supports SOURCE_IP_PORT as load
balancer method. This commits add support for OVN Octavia provider by

- Use SOURCE_IP_PORT as load balancer method when LB provider is OVN
- Disable allowed CIDR when LB provider is OVN


ching-kuo:
Current issues:

- LB method is hard coded string "SOURCE_IP_PORT" as newer version of gophercloud haven't release yet which includes the PR [1]
- OVN LB doesn't do source NAT so that user will have to manually open port 443 to 0.0.0.0/0 (or others allowed CIDR) to master security group in order to access Kubernetes API.

upup/pkg/fi/cloudup/openstacktasks/lbpool.go
```diff
@@ -132,9 +132,13 @@ func (_ *LBPool) RenderOpenstack(t *openstack.OpenstackAPITarget, a, e, changes
 			return fmt.Errorf("failed to loadbalancer ACTIVE provisioning status %v: %v", provisioningStatus, err)
 		}
 
+		LbMethod := v2pools.LBMethodRoundRobin
+		if *e.Loadbalancer.Provider == "ovn" {
```
zetaab, line 136: use fi.StringValue(e.Loadbalancer.Provider)

upup/pkg/fi/cloudup/openstacktasks/lblistener.go
```diff
@@ -154,7 +154,7 @@ func (_ *LBListener) RenderOpenstack(t *openstack.OpenstackAPITarget, a, e, chan
 		e.ID = fi.String(listener.ID)
 		return nil
 	} else if len(changes.AllowedCIDRs) > 0 {
-		if useVIPACL {
+		if useVIPACL && (*e.Pool.Loadbalancer.Provider != "ovn") {
```
zetaab, line 157: fi.StringValue(e.Loadbalancer.Provider)

upup/pkg/fi/cloudup/openstacktasks/lblistener.go
```diff
@@ -143,7 +143,7 @@ func (_ *LBListener) RenderOpenstack(t *openstack.OpenstackAPITarget, a, e, chan
 			ProtocolPort:   443,
 		}
 
-		if useVIPACL {
+		if useVIPACL && (*e.Pool.Loadbalancer.Provider != "ovn") {
```
zetaab, line 146: fi.StringValue(e.Loadbalancer.Provider)

zetaab (changes requested):

zetaab:
@ching-kuo does OVN support health monitors nowadays? 

https://docs.openstack.org/networking-ovn/latest/admin/loadbalancer.html says

> There are no Health Monitors currently in OVN’s Driver for Load Balancer. Therefore Health Checking is not possible with OVN.

https://docs.openstack.org/ovn-octavia-provider/latest/admin/driver.html does not mention this anymore

and commit like https://opendev.org/openstack/ovn-octavia-provider/commit/8bbd8f1b62d7811433079bff8ce940fbdd041a0c exists

ching-kuo:
@zetaab I can confirm that OVN provider support health monitor in Yoga release. I've tested the setup in my environment.

The only manual operation I'm doing now to make it works is to open port 443 to 0.0.0.0/0 (or others allowed CIDR) to master security group in order to access Kubernetes API as the LB doesn't do source NAT.

zetaab:
@ching-kuo can you fix that API firewall logic in https://github.com/kubernetes/kops/blob/master/pkg/model/openstackmodel/firewall.go#L248-L279

so if we use OVN we should open b.Cluster.Spec.KubernetesAPIAccess array instead of b.Cluster.Spec.NetworkCIDR

ching-kuo:
@zetaab Added the logic to open apiAccess if octavia provider is OVN. 
Thanks for pointing out the location.

zetaab:
@ching-kuo please execute `./hack/update-expected.sh` to update test data

ching-kuo:
@zetaab I've ran the script and there is no changes to any of the files.

zetaab (approved):

PR #13682, opened 2022-05-25
Fix kops update for OpenStack with LB

ching-kuo:
In the last PR to support OVN provider for LB, listener will refer to
load balancer provider for ACL settings. While currently get listener
API returns empty Pools, which will cause nil pointer dereference when
referring Pool.Loadbalancer.Provider.

This commit fix this issue by getting pool information with
DefaultPoolID when Pools is empty. As I added GetPool function, the
origin GetPool function is renamed to GetPoolMember.


zetaab (approved):
looks good from my perspective.

PR #13703, opened 2022-05-30
Copy weaveworks/mesh to avoid doing in-place changes

hakman:
/cc @olemarkus @justinsb


hakman:
Failures `pull-kops-e2e-kubernetes-do-gossip` should be addressed by https://github.com/kubernetes/kubernetes/pull/110277.

rifelpet:
Why can't we use the clean vendored version now that we're off bazel and can support build tags?

hakman:
> Why can't we use the clean vendored version now that we're off bazel and can support build tags?

Do we want to add those build tags to both test and protokube build?
I think the plan to find some other solution in the near future.

hakman:
https://github.com/kubernetes/kops/pull/13706 looks like a better and simpler alternative. Will close this PR.

hakman:
Doing some more testing with gossip enabled.
/hold

PR #13705, opened 2022-05-30
Fix links to go docs

diversario:
/kind documentation

Go doc link path is case-sensitive, so use all-lowercase `kops` in the path.


hakman (approved):
Thanks @diversario! 😄

PR #13706, opened 2022-05-30
Use build tags for Protokube

hakman:
Alternative to #13703 

/cc @rifelpet @olemarkus @justinsb


hakman:
peer_name_hash

hakman:
Not sure why this is failing at the moment.
/hold

hakman:
We have a winner! Cluster with gossip based dns is finally starting.
/hold cancel

hakman:
Those importing kOps as a lib should not care much about protokube or dns-controller. All other components are unaffected.

Issue #6517, opened 2019-02-22
Adding GuestAccelerators to InstanceTemplate


harishsearce:
As per steps mentioned to add GuestAccelerators #4277 

@chrislovecnm @roberthbailey i tried to add GuestAccelerators as suggested i.e:

* we need a new field in the instance group spec. Documentation https://github.com/kubernetes/kops/blob/master/docs/development/api_updates.md for how to add a field.
* follow @roberthbailey's instructions and add the `GuestAccelerators` I think in here https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/gcetasks/instancetemplate.go

As robert mentioned "So this is a field that kops would need to set when creating the instance template via some config exposed to the user in the kops api" where does this changes needs to be done in kops api?

If you could help on this, it would be great, i can raise PR once i am through.

FYI: i had added fields to the api by following steps mentioned in 1. https://github.com/kubernetes/kops/blob/master/docs/development/api_updates.md
GuestAccelerators []*AcceleratorConfig
I could see the updates in IG spec getting reflected with the changes but nothing happening:

Steps i followed:

* Add new field in the the api
* Following changes done in kops/upup/pkg/fi/cloudup/gcetasks/instancetemplate.go:
  ** Added GuestAccelerators []*AcceleratorConfig in type InstanceTemplate struct
  ** And added
  type AcceleratorConfig struct {
  // AcceleratorCount: The number of the guest accelerator cards exposed
  // to this instance.
  AcceleratorCount int64 `json:"acceleratorCount,omitempty"`
  // AcceleratorType: Full or partial URL of the accelerator type resource
  // to attach to this instance. If you are creating an instance template,
  // specify only the accelerator name.
  AcceleratorType string `json:"acceleratorType,omitempty"`
  }
* make apimachinery && make

harishsearce:
Hi team,

Any help would be appreciated.

Awaiting for your positive response!

harishsearce:
@chrislovecnm i was able to provision GPU accelerators on GCE by doing some modification in the source code, I would like to raise a PR against it.

I have gone through [contributing](https://github.com/kubernetes/kops/blob/master/CONTRIBUTING.md) section

Please do let me know how can i go ahead on this.

lime008:
Any news on this? Also if you @harishsearce have something working could you provide a link to a fork where this is implemented?

olemarkus:
[...]

PR #13704, opened 2022-05-30
Adding GuestAccelerators to InstanceTemplateFeature/accelerators

jonasasx:
Resolves https://github.com/kubernetes/kops/issues/6517

PR #13707, opened 2022-05-30
Adding GuestAccelerators to InstanceTemplate

jonasasx:
Resolves https://github.com/kubernetes/kops/issues/6517
Fix non-master PR https://github.com/kubernetes/kops/pull/13704


pkg/apis/kops/instancegroup.go
```diff
@@ -363,3 +365,9 @@ type LoadBalancer struct {
 	// TargetGroupARN to associate with this instance group (AWS ALB/NLB)
 	TargetGroupARN *string `json:"targetGroupARN,omitempty"`
 }
+
+// AcceleratorConfig defines an accelerator config
+type AcceleratorConfig struct {
+	AcceleratorCount int64  `json:"acceleratorCount,omitempty"`
+	AcceleratorType  string `json:"acceleratorType,omitempty"`
```
hakman, line 372: Why not just `count` and `type`? Should be pretty obvious they refer to accelerators.
```suggestion
	Count int64  `json:"count,omitempty"`
	Type  string `json:"type,omitempty"`
```

pkg/model/gcemodel/autoscalinggroup.go
```diff
@@ -90,7 +90,7 @@ func (b *AutoscalingGroupModelBuilder) buildInstanceTemplate(c *fi.ModelBuilderC
 				BootDiskImage:  s(ig.Spec.Image),
 
 				// TODO: Support preemptible nodes?
-				Preemptible: fi.Bool(false),
+				Preemptible: fi.Bool(ig.Spec.GuestAccelerators != nil && len(ig.Spec.GuestAccelerators) > 0),
```
hakman, line 93: Why is this related to preemption?

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
```diff
@@ -211,6 +213,16 @@ func (e *InstanceTemplate) Find(c *fi.Context) (*InstanceTemplate, error) {
 		// System fields
 		actual.Lifecycle = e.Lifecycle
 
+		if p.GuestAccelerators != nil {
```
hakman, line 216:
```suggestion
```

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
```diff
@@ -335,13 +347,26 @@ func (e *InstanceTemplate) mapToGCE(project string, region string) (*compute.Ins
 		})
 	}
 
+	var accelerators []*compute.AcceleratorConfig
+	if e.GuestAccelerators != nil {
```
hakman, line 351:
```suggestion
```

pkg/model/gcemodel/autoscalinggroup.go
```diff
@@ -168,6 +168,16 @@ func (b *AutoscalingGroupModelBuilder) buildInstanceTemplate(c *fi.ModelBuilderC
 			//}
 			//t.Labels = labels
 
+			if ig.Spec.GuestAccelerators != nil && len(ig.Spec.GuestAccelerators) > 0 {
```
hakman, line 171: This `if` can be removed. Setting this to empty slice allows unsetting it later if not wanted anymore. Otherwise, a `nil` value will be ignored.
```suggestion
```

hakman (changes requested):
Thanks @jonasasx. This looks mostly good, just some nits and Qs from my side.
Besides this there are a few things missing:
1. validation that both `count` and `type` are set.
2. add the fields to the terraform output also.

pkg/apis/kops/instancegroup.go
jonasasx: gcp fields of InstanceTemplate are named with the prefix `accelerator`. I think we need to keep uniformity.

pkg/model/gcemodel/autoscalinggroup.go
jonasasx: Google Cloud forbids to create instances with GPUs in `MIGRATE` maintenance mode. Only `TERMINATE` is allowed.

pkg/model/gcemodel/autoscalinggroup.go
jonasasx: I have been coded in Java for many years with endless NullPointerExceptions :)

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
jonasasx: I think, we must have this check. Otherwise `actual.GuestAccelerators` will be an empty slice instead of nil. It currupt back copatibility.

pkg/model/gcemodel/autoscalinggroup.go
jonasasx: Why setting and unsetting slice is better than a preventive check?

pkg/model/gcemodel/autoscalinggroup.go
hakman: In that case, the nil check is not needed.

pkg/apis/kops/instancegroup.go
hakman: Not a big fan of this, but can live with it.

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
hakman: Don't think I understand how it can corrupt backward compatibility.

pkg/model/gcemodel/autoscalinggroup.go
hakman: tasks check actual vs model, when model is not `nil`.

pkg/model/gcemodel/autoscalinggroup.go
jonasasx: fixed

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
jonasasx: All InstanceTemplates created before this implementation will have a difference on applying this implementation:
```
- 
+ guestAccelerators: [],
```

hakman: This is the kOps representation, not the internal one of GCP. `Find()` will return `[]AcceleratorConfig{}` wether `p.GuestAccelerators` is `nil` or empty slice.

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
hakman, line 351: I think this is a little better.
```suggestion
	if len(e.GuestAccelerators) > 0 {
```

pkg/model/gcemodel/autoscalinggroup.go
hakman: After reading a little, I think the correct way would be to set [Scheduling.OnHostMaintenance](https://pkg.go.dev/google.golang.org/api@v0.82.0/compute/v1#Scheduling.OnHostMaintenance).
`Preemptible ` seems to be the equivalent of Spot instances on AWS.

jonasasx: Fixed. Moved to `InstanceTemplate.mapToGCE`

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
jonasasx: fixed

pkg/model/gcemodel/autoscalinggroup.go
jonasasx: fixed

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
```diff
@@ -235,7 +245,7 @@ func (e *InstanceTemplate) mapToGCE(project string, region string) (*compute.Ins
 	// TODO: This is similar to Instance...
 	var scheduling *compute.Scheduling
 
-	if fi.BoolValue(e.Preemptible) {
+	if fi.BoolValue(e.Preemptible) || len(e.GuestAccelerators) > 0 {
```
hakman, line 248: This has nothing to do with preemption. Instead, add this after the current block:
```go
if len(e.GuestAccelerators) > 0 {
    // Instances with accelerators cannot be migrated.
    scheduling.OnHostMaintenance = "TERMINATE"
}

File: upup/pkg/fi/cloudup/gcetasks/instancetemplate.go [hunk 7]
```diff
@@ -335,13 +345,26 @@ func (e *InstanceTemplate) mapToGCE(project string, region string) (*compute.Ins
 		})
 	}
 
+	var accelerators []*compute.AcceleratorConfig
+	if len(e.GuestAccelerators) > 0 {
```
hakman, line 349: What happens when you remove accelerator from config? With this here it won't apply any change.

hakman (changes requested):

upup/pkg/fi/cloudup/gcetasks/accelerator_config.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (a *AcceleratorConfig) GetDependencies(tasks map[string]fi.Task) []fi.Task {
+	return nil
+}
+
+func (_ *AcceleratorConfig) ShouldCreate(a, e, changes *AcceleratorConfig) (bool, error) {
+	if e.AcceleratorCount < 0 {
+		return false, fmt.Errorf("acceleratorCount must be positive or 0")
+	}
+	if e.AcceleratorCount < 0 {
```
hakman, line 43:
```suggestion
	if e.AcceleratorType == "" {
```

jonasasx: fixed

hakman:
Please ignore the `pull-kops-e2e-k8s-gce` test failure, it is related to #13855.

hakman (approved):

PR #13717, opened 2022-06-02
[DigitalOcean] Restart journald service on node startup

srikiz:
Digital Ocean cloud provider image has `/etc/machine-id` baked into the image. And as a result when nodeup executes `system-id-machine-setup` it re-creates a new machine-id and as a result we lose track of journal entries. The only fix is to restart journald to view the system logs again. 

The fix here will ensure that journald is restarted on node startup so we don't need an explicit restart manually.

This should also help us to view the logs in our test-grid artifacts. Currently it shows empty [here](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-do-calico/1532344823695020032/artifacts/10.124.0.7/journal.log) 

FYI - @timoreimann 
Reference - https://serverfault.com/questions/1058259/journalctl-stops-working-randomly-after-boot-on-digitalocean-droplet

### DIFF
--- a/pkg/model/resources/nodeup.go
+++ b/pkg/model/resources/nodeup.go
@@ -149,6 +149,12 @@ function download-release() {
 
 /bin/systemd-machine-id-setup || echo "failed to set up ensure machine-id configured"
 
+{{- if eq GetCloudProvider "digitalocean" }}
+  # DO has machine-id baked into the image and journald should be flushed
+  # to use the new machine-id
+  systemctl restart systemd-journald
+{{- end }}
+
 echo "== nodeup node config starting =="
 ensure-install-dir
 
@@ -178,6 +184,7 @@ type NodeUpScript struct {
 	KubeEnv              string
 	CompressUserData     bool
 	SetSysctls           string
+	CloudProvider        string
 	ProxyEnv             func() (string, error)
 	EnvironmentVariables func() (string, error)
 	ClusterSpec          func() (string, error)
@@ -234,6 +241,10 @@ func (b *NodeUpScript) Build() (fi.Resource, error) {
 
 		"CompressUserData": func() bool {
 			return b.CompressUserData
+		},
+
+		"GetCloudProvider": func() string {
+			return b.CloudProvider
 		},
 
 		"SetSysctls": func() string {
--- a/pkg/model/bootstrapscript.go
+++ b/pkg/model/bootstrapscript.go
@@ -397,6 +397,8 @@ func (b *BootstrapScript) Run(c *fi.Context) error {
 	// See https://github.com/kubernetes/kops/issues/10206 for details.
 	nodeupScript.SetSysctls = setSysctls()
 
+	nodeupScript.CloudProvider = string(c.Cluster.Spec.GetCloudProvider())
+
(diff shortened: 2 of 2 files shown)


pkg/model/resources/nodeup.go
```diff
@@ -177,6 +184,7 @@ type NodeUpScript struct {
 	NodeUpAssets         map[architectures.Architecture]*mirrors.MirroredAsset
 	KubeEnv              string
 	CompressUserData     bool
+	RestartJournald      bool
```
hakman, line 187: Maybe just pass the CloudProvider instead? I am sure there will be other use cases for it.

pkg/model/resources/nodeup.go
```diff
@@ -149,6 +149,13 @@ function download-release() {
 
 /bin/systemd-machine-id-setup || echo "failed to set up ensure machine-id configured"
 
+{{ if RestartJournald -}}
+# DO cloud provider has /etc/machine-id baked in the image and 
+# systemd-machine-id-setup creates a new id and journalctld should be flushed
+# to use the new machine-id
+systemctl restart systemd-journald
+{{- end }}
+
```
hakman, line 158: This adds extra spaces so it will require to run `hack/update-expected` to fix the golden output.

hakman (changes requested):

pkg/model/resources/nodeup.go
srikiz: sure, updated.

pkg/model/resources/nodeup.go
```diff
@@ -149,12 +150,13 @@ function download-release() {
 
 /bin/systemd-machine-id-setup || echo "failed to set up ensure machine-id configured"
 
-{{ if RestartJournald -}}
-# DO cloud provider has /etc/machine-id baked in the image and 
-# systemd-machine-id-setup creates a new id and journalctld should be flushed
-# to use the new machine-id
-systemctl restart systemd-journald
-{{- end }}
+if [[ "${CLOUD_PROVIDER}" = "digitalocean" ]]; then
```
hakman, line 153: Maybe still do this at template level like before? No need to have it for all clouds, in bash.

hakman: Or maybe always restart `systemd-journald`, without any check?

srikiz: I will probably keep this check since this issue was only pertaining to Digital Ocean. Though, restarting journald should not cause any issues as such..

pkg/model/resources/nodeup.go
srikiz: `hack/update-expected` doesn't seem to work for me.. Trying few options here.

pkg/model/resources/nodeup.go
hakman: Also, please remove the comment or make it very short, without echo.
The size for user-data is quite small. We don't want to add extra unneeded chars.

srikiz: Sure, done.

hakman (approved):
Thanks for your patience @srikiz!

PR #13740, opened 2022-06-07
[WIP] Generate machine-id only when needed

hakman:
Alternative to #13717


olemarkus:
I wonder if we even need this. systemd should always create machine-id at some point.

hakman:
> I wonder if we even need this. systemd should always create machine-id at some point.

Yeah, and also it should be reproducible.
For some reason DO still broken so, I guess it's initialised with incorrect ID.

olemarkus:
A common AMI building issue is that when building the AMI, machine-id is generated and written, but not deleted upon cleanup. So keeps reusing the same ID everywhere. But that shouldn't really lead to an error.

PR #13753, opened 2022-06-08
Use dynamic client for applying channels manifest rather than calling kubectl

olemarkus:
The aim of this PR is to remove the dependency of channels on the kubectl binary. This makes k8s version a bit more consistent, and makes it easier to run clientside.

The approach is more or less a copy/paste of the pruner. Refactoring opportunities there.

@justinsb appreciate your eyes on this one. There are probably lots of things to think about when using the dynamic client like this.


hakman:
Current approach (since #13731) uses `kubectl replace` and causes some issues also:
https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-conformance-1-24/1534818290919018496/artifacts/i-0b53ec288383f99fa/protokube.log
```
I0609 11:07:32.775549   96689 s3fs.go:329] Reading file "s3://k8s-kops-prow/e2e-e2e-kops-aws-conformance-1-24.test-cncf-aws.k8s.io/addons/core.addons.k8s.io/v1.4.0.yaml"
I0609 11:07:32.959358   96689 apply.go:60] Running command: kubectl replace -f /tmp/channel2405827354/manifest.yaml --force --field-manager=kops
I0609 11:07:33.102919   96689 apply.go:63] error running kubectl replace -f /tmp/channel2405827354/manifest.yaml --force --field-manager=kops
I0609 11:07:33.102938   96689 apply.go:64] Error from server (Forbidden): error when deleting "/tmp/channel2405827354/manifest.yaml": namespaces "kube-system" is forbidden: this namespace may not be deleted
error updating "core.addons.k8s.io": error applying update from "s3://k8s-kops-prow/e2e-e2e-kops-aws-conformance-1-24.test-cncf-aws.k8s.io/addons/core.addons.k8s.io/v1.4.0.yaml": error running kubectl: exit status 1
```

channels/pkg/channels/apply.go
```diff
@@ -17,53 +17,114 @@ limitations under the License.
...
-			klog.Warningf("error deleting temp dir %q: %v", tmpDir, err)
+	objectsByKind := make(map[schema.GroupKind][]*kubemanifest.Object)
+	for _, object := range objects {
+		gv, err := schema.ParseGroupVersion(object.APIVersion())
+		if err != nil || gv.Version == "" {
+			return fmt.Errorf("failed to parse apiVersion %q", object.APIVersion())
+		}
+		kind := object.Kind()
+		if kind == "" {
+			return fmt.Errorf("failed to find kind in object")
```
justinsb, line 52: Nit: we might want to log more object info here (maybe %v of object), although I suspect this "won't happen" (tm)

channels/pkg/channels/apply.go
```diff
@@ -17,53 +17,114 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/restmapper"
 	"k8s.io/klog/v2"
+	"k8s.io/kops/pkg/kubemanifest"
 )
 
-// Apply calls kubectl apply to apply the manifest.
-// We will likely in future change this to create things directly (or more likely embed this logic into kubectl itself)
-func Apply(data []byte) error {
```
justinsb, line 31: I wonder if we should try to make this pluggable during the transition?  cf what we did here https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/pkg/applier/direct.go ...

I do like this applier better than reusing kubectl, as that is giving us trouble e.g. https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/225/files# ...

channels/pkg/channels/apply.go
```diff
@@ -17,53 +17,114 @@ limitations under the License.
...
-		return fmt.Errorf("error writing temp file: %v", err)
+		gvk := gv.WithKind(kind)
+		gk := gvk.GroupKind()
+		objectsByKind[gk] = append(objectsByKind[gk], object)
 	}
 
-	_, err = execKubectl("replace", "-f", localManifestFile, "--force", "--field-manager=kops")
-	return err
+	for gk := range objectsByKind {
+		if err := p.applyObjectsOfKind(ctx, gk, objectsByKind[gk]); err != nil {
```
justinsb, line 61: There might be interdependencies here (the classic is namespace before objects in that namespace); there are two possible strategies ... we could pre-sort, we could iterate a few times.  In practice because we're going to retry this anyway, I don't think it matters, just one to note for the future!

channels/pkg/channels/apply.go
```diff
@@ -17,53 +17,114 @@ limitations under the License.
...
+			return fmt.Errorf("failed to apply objects of kind %s: %w", gk, err)
```
justinsb, line 62: We may want to keep on going (i.e. accumulate errors) in case of dependencies

channels/pkg/channels/apply.go
```diff
@@ -17,53 +17,114 @@ limitations under the License.
...
 
-func execKubectl(args ...string) (string, error) {
-	kubectlPath := "kubectl" // Assume in PATH
-	cmd := exec.Command(kubectlPath, args...)
-	env := os.Environ()
-	cmd.Env = env
+func (p *Applier) applyObjectsOfKind(ctx context.Context, gk schema.GroupKind, expectedObjects []*kubemanifest.Object) error {
+	klog.V(2).Infof("applying objects of kind: %v", gk)
+
+	restMapping, err := p.RESTMapper.RESTMapping(gk)
```
justinsb, line 71: Nit: I think we should apply with the correct version, i.e. group by GVK.

channels/pkg/channels/apply.go
```diff
@@ -17,53 +17,114 @@ limitations under the License.
...
+	}
+
+	gvr := restMapping.Resource
 
-	human := strings.Join(cmd.Args, " ")
-	klog.V(2).Infof("Running command: %s", human)
-	output, err := cmd.CombinedOutput()
+	baseResource := p.Client.Resource(gvr)
+
+	actualObjects, err := baseResource.List(ctx, v1.ListOptions{})
```
justinsb, line 80: If we have to read the existing objects, we might want to scope by namespace, just for memory etc in big clusters.  We may not have to do this at all with server-side-apply, TBD maybe

channels/pkg/channels/apply.go
```diff
@@ -17,53 +17,114 @@ limitations under the License.
...
+		actualMap[key] = actualObject
+	}
+
+	for _, expectedObjects := range expectedObjects {
+		name := expectedObjects.GetName()
+		namespace := expectedObjects.GetNamespace()
+		key := namespace + "/" + name
+
+		var resource dynamic.ResourceInterface
+		if namespace != "" {
```
justinsb, line 104: Nit: If we pass in the restMapping, we can know if this object is namespaced or not (and raise an appropriate error/warning or default the namespace, whatever we want to do)

channels/pkg/channels/apply.go
```diff
@@ -17,53 +17,114 @@ limitations under the License.
...
+			resource = p.Client.Resource(gvr).Namespace(namespace)
+		} else {
+			resource = p.Client.Resource(gvr)
+		}
+
+		obj := expectedObjects.ToUnstructured()
+
+		if actual, found := actualMap[key]; found {
```
justinsb, line 112: I suspect we should be using server-side-apply here.  It's a little easier in some ways (it's a Patch, no need to worry about whether or not the object exists), although we've seen some of the field manager challenges during the transition.

justinsb (commented):
This is awesome and I would love to see us pursue this approach!

justinsb:
So I think this is great, and we should get this into 1.25.  I'm happy to basically merge this as-is and we can iterate on it or alternatively if you want to incorporate some of the suggestions first (if you agree!) we can do it that way.

PR #13794, opened 2022-06-14
Fix GCE firewall rule listing logic for teardown

rifelpet:
In https://github.com/kubernetes/kops/pull/13513 we updated the firewall name logic to support clusters with names longer than GCE naming limits.
This updates the listing logic for teardown to match.

Should fix at least some of the GCE periodic jobs:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-gce-latest/1536707209835057152#1:build-log.txt%3A84439

`Network:e2e-e2e-kops-gce-latest-k8s-local	error deleting resources, will retry: googleapi: Error 400: The network resource 'projects/k8s-jkns-e2e-gce-gci-serial/global/networks/e2e-e2e-kops-gce-latest-k8s-local' is already being used by 'projects/k8s-jkns-e2e-gce-gci-serial/global/firewalls/nodeport-external-to-node-ipv6-e2e-e2e-kops-gce-latest-k-vqb8db'`

### DIFF
--- a/pkg/resources/gce/gce.go
+++ b/pkg/resources/gce/gce.go
@@ -516,7 +516,7 @@ func deleteForwardingRule(cloud fi.Cloud, r *resources.Resource) error {
 	return c.WaitForOp(op)
 }
 
-// listFirewallRules discovers Firewall objects for the cluster
+// listFirewallRules discovers firewalls, forwarding rules, target pools, and health checks for the cluster
 func (d *clusterDiscoveryGCE) listFirewallRules() ([]*resources.Resource, error) {
 	c := d.gceCloud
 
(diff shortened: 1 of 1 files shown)


rifelpet:
not sure how stable this job is:

/test pull-kops-e2e-k8s-gce

rifelpet:
Hm this is going to be tricky because in our model, the "id" of the firewall rule is the prefix and the cluster name can be arbitrarily truncated, so in the listing logic where we only know the cluster name we'll have to search for some arbitrarily truncated cluster name which seems brittle. The best option may be to determine the longest "id" in the model today, determine how much of the original cluster name wont ever be truncated, and search for that. that still seems fairly brittle but may be our only option.

PR #13800, opened 2022-06-15
add kops binary to gitignore

remyleone:


olemarkus:
What's the reason behind this?

remyleone:
As a contributor when I'm building a go project I expect that my git status won't change. It is especially true if I got visualization in my terminal of the current state of git. As of today when I'm building kops I can see that the kops binary is considered untracked. As it is very unlikely that the kops binary will ever be committed to the git repository, I think it is useful to ignore it at the git level. By doing so, the git status is less cluttered.

olemarkus:
`make`, `make kops-install`, or `make kops` shouldn't put the binary in that location though.

remyleone:
Maybe but I wanted to build it using the native way in go :) Is that line that much of a hassle? What negative side effects could it have?

olemarkus:
More of a principle. We cannot add any custom development environment setup to our git configs.
Why don't you just add that path to your own `.git/info/exclude`?

hakman:
Also, keep in mind that without a `/` in front, it would ignore all things named `kops` from all the subdirs.

remyleone:
@hakman fixed :)

remyleone:
@olemarkus Should I close this PR? Will, you not merge it?

hakman:
TBH, most likely this PR will not be accepted as a patch. Buiding binaries in the root dir is not a good practice.
I recommend setting the output dir to `.build/`, which is already excluded.
Even more, there are some components like `protokube` and `dns-controller` that require special flags like `GOFLAGS="-tags=peer_name_alternative,peer_name_hash"`. This is why, using `make` for the build is usually a better idea.

PR #13801, opened 2022-06-15
add support for varcheck linter

remyleone:


olemarkus:
Isn't varcheck in the list of deprecated linters?

remyleone:
I don't see any deprecation notice here: https://golangci-lint.run/usage/linters/#varcheck besides it seems it caught some unused variables :)

olemarkus:
There is a comment with a link a few lines above the one you added there.

remyleone:
Would you still like to have the fix without enabling the linter? The fixes could be nice regardless of the status of the linter. The deprecation is still in the process and there does not seem to be an official replacement notice yet.

remyleone:
@hakman rebased :)

hakman (approved):
/test all

PR #13803, opened 2022-06-15
Remove vendor directory

rifelpet:


olemarkus:
Should we try to get this one in for 1.25?

hakman:
> Should we try to get this one in for 1.25?

Last discussion we had about this with @justinsb, the answer was negative.
Kubernetes still has the vendor dir also.

PR #13805, opened 2022-06-16
Use exported interface to detect SSH key type

AaronFriel:
Don't rely on the `golang.org/x/crypto` library to keep consistent,
non-exported type names.


hakman (approved):
Thanks @AaronFriel! :)

PR #13811, opened 2022-06-16
add metric port to nth deployment

raffis:
I am missing the port declaration for metrics in the nth deployment.
Also it seems its not possible to enable it at all.


olemarkus:
The PR looks good.
Could you run `./hack/update-expected.sh` to update the tests?

olemarkus (approved):

PR #13819, opened 2022-06-16
Replace manifests after apply

olemarkus:
* `kubectl replace --force=true` always deletes and creates objects, causing havoc and breakage everywhere.
* `kubectl apply` can update and create objects, but does a strategic merge, which means in many cases arrays are merged. This leaves behind containers, ports and other things that causes apply to fail.
* `kubectl replace --force=false` cannot create objects, but does a "normal" merge of objects, meaning lists are _not_ merged, but replaced entirely. This is the behavior we want.

Combining `kubectl apply` with `kubectl replace` seems to be what we want right now until we know the kubectl client side apply doesnt' manage any fields.

/kind office-hours

### DIFF
--- a/channels/pkg/channels/apply.go
+++ b/channels/pkg/channels/apply.go
@@ -45,9 +45,28 @@ func Apply(data []byte) error {
 	if err := os.WriteFile(localManifestFile, data, 0o600); err != nil {
 		return fmt.Errorf("error writing temp file: %v", err)
 	}
+	{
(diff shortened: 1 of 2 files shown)


justinsb:
I looked at what `kubectl replace` actually does, and it does a PUT, so I think that is very much in line with what we want to do.

Long-term I think we will need to clean up any duplicate field managers, and I think @apelisse has some suggestions for us there!

Thanks for this @olemarkus!

/approve
/lgtm

PR #13823, opened 2022-06-17
Recommend the latest kOps version in alpha & stable channels and add 1.24 to alpha

MoShitrit:
/cc @olemarkus @hakman

### DIFF
--- a/channels/alpha
+++ b/channels/alpha
@@ -87,6 +87,9 @@ spec:
     networking:
       kubenet: {}
   kubernetesVersions:
+  - range: ">=1.24.0"
+    recommendedVersion: 1.24.2
+    requiredVersion: 1.24.0
   - range: ">=1.23.0"
     recommendedVersion: 1.23.8
     requiredVersion: 1.23.0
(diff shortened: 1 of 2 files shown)


channels/alpha
```diff
@@ -130,28 +130,32 @@ spec:
     recommendedVersion: 1.11.10
     requiredVersion: 1.11.10
   kopsVersions:
+  - range: ">=1.24.0-alpha.1"
+    recommendedVersion: "1.23.2"
+    #requiredVersion: 1.23.0
+    kubernetesVersion: 1.23.8
```
hakman, line 136: Also, a section for 1.24 should be added above, in the `kubernetesVersions` section.
```suggestion
    kubernetesVersion: 1.24.2
```

hakman (changes requested):

channels/alpha
MoShitrit: Thanks @hakman!
Fixing.
As for the `1.24` section in `kubernetesVersions` - does the below look right? I mostly wasn't sure about the `requiredVersion` part, since technically `1.24.0` hasn't been released yet.
Do we just point to the most recent 1.24 release ([v1.24.0-beta.2](https://github.com/kubernetes/kops/releases/tag/v1.24.0-beta.2))? Or should it point to `1.23.0` for now?

hakman: Yup, that looks good. Required version refers to k8s version I think.

MoShitrit: Of course it does, lol 🤦🏻 
Thanks for clarifying!

hakman:
Looks good to me. Could you also squash?

hakman (approved):
👍

PR #13827, opened 2022-06-18
Add manual job for updating dependencies

hakman:
This is an alternative to Dependabot, which is really annoying with the many PRs.
Once we consider it stable, it can be run periodically.

/cc @justinsb @olemarkus @rifelpet


hakman:
Rebased and rerun `make depup`.
/hold cancel

olemarkus (approved):

PR #13845, opened 2022-06-21
Add config drive as a source for OpenStack instance metadata

ederst:
This adds the config drive as an additional source for instance metadata when using OpenStack.

The reasoning behind this is that the Metadata Service may not always be reachable. For example when not using DHCP on a tenant network, or using a provider network (and at least according to our OpenStack support, the config drive is far more stable to use than the Metadata Service).

Currently, the search order is hard coded to `configDrive, metadataService`. So the code tries to look up the metadata in the config drive first and if not successful it will query the Metadata Service Endpoint. I did it this way because I think/assume the config drive lookup will fail faster in case it is not available.

Basically, the code is lifted/borrowed and adapted from the external OpenStack provider:

* https://github.com/kubernetes/cloud-provider-openstack/blob/27b6fc483451b6df2112a6a4a40a34ffc9093635/pkg/util/metadata/metadata.go

I did this for the etcd-manager too: https://github.com/kubernetes-sigs/etcdadm/pull/318

### DIFF
--- a/protokube/pkg/protokube/openstack_volume.go
+++ b/protokube/pkg/protokube/openstack_volume.go
@@ -84,7 +84,7 @@ func getLocalMetadata() (*InstanceMetadata, error) {
 	return nil, err
 }
 
-// NewOpenStackCloudProvider builds a OpenstackVolume
+// NewOpenStackCloudProvider builds a OpenStackCloudProvider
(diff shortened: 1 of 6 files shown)


zetaab (approved):
looks good, and at least old behaviour should continue working

PR #13852, opened 2022-06-22
Fix unsetting ASG max price

olemarkus:
Fixes https://github.com/kubernetes/kops/issues/13849

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
+++ b/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
@@ -218,6 +218,10 @@ func (e *AutoscalingGroup) Find(c *fi.Context) (*AutoscalingGroup, error) {
 			actual.MixedSpotAllocationStrategy = mpd.SpotAllocationStrategy
 			actual.MixedSpotInstancePools = mpd.SpotInstancePools
 			actual.MixedSpotMaxPrice = mpd.SpotMaxPrice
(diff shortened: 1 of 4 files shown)


upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -218,6 +218,10 @@ func (e *AutoscalingGroup) Find(c *fi.Context) (*AutoscalingGroup, error) {
 			actual.MixedSpotAllocationStrategy = mpd.SpotAllocationStrategy
 			actual.MixedSpotInstancePools = mpd.SpotInstancePools
 			actual.MixedSpotMaxPrice = mpd.SpotMaxPrice
+			// MixedSpotMaxPrice must be set to "" in order to unset.
+			if mpd.SpotMaxPrice != nil && e.MixedSpotMaxPrice == nil {
+				e.MixedSpotMaxPrice = fi.String("")
```
hakman, line 223: Why do this in `Find()`? Seems more like a thing to do in model.

olemarkus: Then I think I'd have to convert the nil from AWS to "" here instead. This way seemed easier.

hakman: It is easier, but makes things harder to track later. Find should only set expected for ignored fields.

hakman (approved):

PR #13853, opened 2022-06-22
Fix openstack tag limitation

akkina2107:
Openstack is not allowing object tag size of more than 60 characters, as
we can not rename a cluster we have to truncate and limit length to 42
for the tag value.

### DIFF
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/dns"
 	"k8s.io/kops/pkg/model"
+	"k8s.io/kops/pkg/truncate"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstack"
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstacktasks"
@@ -43,6 +44,15 @@ var _ fi.ModelBuilder = &ServerGroupModelBuilder{}
 
 // See https://specs.openstack.org/openstack/nova-specs/specs/newton/approved/lowercase-metadata-keys.html for details
 var instanceMetadataNotAllowedCharacters = regexp.MustCompile("[^a-zA-Z0-9-_:. ]")
+
+// Constants for truncating Tags
+const MAX_TAG_LENGTH_OPENSTACK = 60
+
+var TRUNCATE_OPT = truncate.TruncateStringOptions{
+	MaxLength:     MAX_TAG_LENGTH_OPENSTACK,
+	AlwaysAddHash: false,
+	HashLength:    6,
+}
 
 func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *openstacktasks.ServerGroup, ig *kops.InstanceGroup) error {
 	sshKeyNameFull, err := b.SSHKeyName()
@@ -146,9 +156,9 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 			InstanceGroupName: &groupName,
 			Network:           b.LinkToNetwork(),
 			Tags: []string{
-				fmt.Sprintf("%s=%s", openstack.TagKopsInstanceGroup, groupName),
-				fmt.Sprintf("%s=%s", openstack.TagKopsName, portTagKopsName),
-				fmt.Sprintf("%s=%s", openstack.TagClusterName, b.ClusterName()),
+				truncate.TruncateString(fmt.Sprintf("%s=%s", openstack.TagKopsInstanceGroup, groupName), TRUNCATE_OPT),
+				truncate.TruncateString(fmt.Sprintf("%s=%s", openstack.TagKopsName, portTagKopsName), TRUNCATE_OPT),
+				truncate.TruncateString(fmt.Sprintf("%s=%s", openstack.TagClusterName, b.ClusterName()), TRUNCATE_OPT),
 			},
 			SecurityGroups:           securityGroups,
 			AdditionalSecurityGroups: ig.Spec.AdditionalSecurityGroups,
--- a/upup/pkg/fi/cloudup/openstacktasks/instance.go
+++ b/upup/pkg/fi/cloudup/openstacktasks/instance.go
@@ -29,6 +29,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/schedulerhints"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"k8s.io/klog/v2"
+	"k8s.io/kops/pkg/truncate"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstack"
 )
@@ -61,6 +62,15 @@ var (
 	_ fi.HasAddress      = &Instance{}
 	_ fi.HasDependencies = &Instance{}
 )
+
+// Constants for truncating Tags
+const MAX_TAG_LENGTH_OPENSTACK = 60
+
+var TRUNCATE_OPT = truncate.TruncateStringOptions{
+	MaxLength:     MAX_TAG_LENGTH_OPENSTACK,
+	AlwaysAddHash: false,
+	HashLength:    6,
+}
 
 // GetDependencies returns the dependencies of the Instance task
 func (e *Instance) GetDependencies(tasks map[string]fi.Task) []fi.Task {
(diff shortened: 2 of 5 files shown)


pkg/model/openstackmodel/servergroup.go
```diff
@@ -148,7 +164,7 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 			Tags: []string{
 				fmt.Sprintf("%s=%s", openstack.TagKopsInstanceGroup, groupName),
 				fmt.Sprintf("%s=%s", openstack.TagKopsName, portTagKopsName),
-				fmt.Sprintf("%s=%s", openstack.TagClusterName, b.ClusterName()),
+				fmt.Sprintf("%s=%s", openstack.TagClusterName, TruncateString(b.ClusterName(), 42)),
```
drekle, line 167: This would be a magic integer IMO.  I think what you mean is something like `MAX_TAG_LENGTH_OPENSTACK - len(openstack.TagClusterName) + 1`  +1 for the `=` here, which just happens to be 42 however would impact anyone much more than necessary if they changed the value of the constant `openstack.TagClusterName`.

pkg/model/openstackmodel/servergroup.go
```diff
@@ -43,7 +43,23 @@ var _ fi.ModelBuilder = &ServerGroupModelBuilder{}
 
 // See https://specs.openstack.org/openstack/nova-specs/specs/newton/approved/lowercase-metadata-keys.html for details
 var instanceMetadataNotAllowedCharacters = regexp.MustCompile("[^a-zA-Z0-9-_:. ]")
-
+// Truncate string.
+func TruncateString(str string, length int) string {
```
drekle, line 47: I believe this could be something more like:

```
const MAX_TAG_LENGTH_OPENSTACK = 60
```
```
func formatTag(tag string) string {
    if len(tag) <= MAX_TAG_LENGTH_OPENSTACK {
        return tag
    }
    return tag[:MAX_TAG_LENGTH_OPENSTACK ]
}
```
```
			Tags: []string{
				formatTag(fmt.Sprintf("%s=%s", openstack.TagKopsInstanceGroup, groupName)),
				formatTag(fmt.Sprintf("%s=%s", openstack.TagKopsName, portTagKopsName)),
				formatTag(fmt.Sprintf("%s=%s", openstack.TagClusterName, b.ClusterName())),
				
...
```

drekle (commented):
@akkina2107 This is a good find.  I appreciate you raising this.  It might be better served as an issue to have a discussion on.

I am not sure of the ramifications at the top of my mind where two clusters of different names have the same truncated string.  I think this may impact certain configurations for Openstack, specifically when it comes to gossip DNS for those without designate, however I believe the ClusterName for Gossip DNS will be derived from metadata.  It may be simpler to restrict the length of the clusterName, or enforce that the truncated string does not match any existing cluster.

rifelpet:
Kops' GCE code has logic to truncate a cluster name such that it still contains a hash suffix, which would avoid similarly-named clusters from being truncated down to identical names. We could probably reuse that here

https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/gce/utils.go#L52-L71

akkina2107:
I used the proposed function, please verify

mitch000001:
This PR is very much related to this one: https://github.com/kubernetes/kops/pull/13861

akkina2107:
please update the current status of this PR

pkg/model/openstackmodel/servergroup.go
```diff
@@ -148,7 +159,7 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.ModelBuilderContext, sg *
 			Tags: []string{
 				fmt.Sprintf("%s=%s", openstack.TagKopsInstanceGroup, groupName),
 				fmt.Sprintf("%s=%s", openstack.TagKopsName, portTagKopsName),
-				fmt.Sprintf("%s=%s", openstack.TagClusterName, b.ClusterName()),
+				fmt.Sprintf("%s=%s", openstack.TagClusterName, TruncateString(b.ClusterName(), 42)),
```
zetaab, line 162: perhaps we should add comment here why we are doing this

akkina2107: error message: {“NeutronError”: {“type”: “InvalidInput”, “message”: “Invalid input for operation: ‘KubernetesCluster=cluster_name_of_more_than_42_characters’ exceeds maximum length of 60.“, “detail”: “”}}

akkina2107: Added comment as suggested, sorry for the confusion

zetaab:
@akkina2107 could you fix the tests. Otherwise it looks correct

akkina2107:
I would appreciate if someone can let me know, on which version this issue will be fixed? we need a solution for 1.22 version is the current solution back ported? if not how can I achieve this? kindly let me know.

akkina2107:
@zetaab and @ederst @drekle
Please inspect the new changes and leave your comments

pkg/model/openstackmodel/servergroup.go
```diff
@@ -44,6 +45,15 @@ var _ fi.ModelBuilder = &ServerGroupModelBuilder{}
...
 
+// Constants for truncating Tags
+const MAX_TAG_LENGH_OPENSTACK = 60
+
+var TRUNCATE_OPT = truncate.TruncateStringOptions{
+	MaxLength:     MAX_TAG_LENGH_OPENSTACK,
+	AlwaysAddHash: false,
+	HashLength:    6,
+}
+
```
ederst, line 56: @akkina2107 What I'd probably do is, consolidate the truncating definitions from [here][1] and [there][2] into the `k8s.io/kops/upup/pkg/fi/cloudup/openstack` package. Either to [`cloudup/openstack/port.go`][3] or [`cloudup/openstack/utils.go`][4].

For example there are already some consts for instances handled like this in [`cloudup/openstack/instance.go`][5], which are used "all over the place".

Also some typo fix:
```suggestion
// Constants for truncating Tags
const MAX_PORT_TAG_LENGTH_OPENSTACK = 60

var TRUNCATE_OPT = truncate.TruncateStringOptions{
	MaxLength:     MAX_PORT_TAG_LENGTH_OPENSTACK,
	AlwaysAddHash: false,
	HashLength:    6,
}
```

Anyhow, I'm basically just suggesting, and have no authority to approve anything etc. Just trying to give some (hopefully) constructive feedback, since I ran into this issue too and wanted to fix it myself. In the end it's up to the official maintainers.

And speaking of them, AFAIK some of them are currently enjoying their vacation or are generally on some other sort of leave.

As for bringing this into 1.22. I am not sure if this will be done, but probably @hakman can clarify this. Or you simply can build your own forked version with it (with some rudimentary CI) like we do with some changes which are not backported.

[1]: https://github.com/kubernetes/kops/blob/eb7c9ab5495839455038c251c80140cc68301eb2/pkg/model/openstackmodel/servergroup.go#L48-L55
[2]: https://github.com/kubernetes/kops/blob/eb7c9ab5495839455038c251c80140cc68301eb2/upup/pkg/fi/cloudup/openstacktasks/instance.go#L66-L73
[3]: https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/openstack/port.go
[4]: https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/openstack/utils.go
[5]: https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/openstack/instance.go#L40-L47

DavidSie:
@zetaab  Please take a look this PR. Can it be merged ?

pkg/model/openstackmodel/servergroup.go
```diff
@@ -44,6 +45,15 @@ var _ fi.ModelBuilder = &ServerGroupModelBuilder{}
 // See https://specs.openstack.org/openstack/nova-specs/specs/newton/approved/lowercase-metadata-keys.html for details
 var instanceMetadataNotAllowedCharacters = regexp.MustCompile("[^a-zA-Z0-9-_:. ]")
 
+// Constants for truncating Tags
+const MAX_TAG_LENGH_OPENSTACK = 60
```
zetaab, line 49: MAX_TAG_LENGHT_OPENSTACK

pkg/model/openstackmodel/servergroup.go
```diff
@@ -44,6 +45,15 @@ var _ fi.ModelBuilder = &ServerGroupModelBuilder{}
...
+
+var TRUNCATE_OPT = truncate.TruncateStringOptions{
+	MaxLength:     MAX_TAG_LENGH_OPENSTACK,
```
zetaab, line 52: MAX_TAG_LENGHT_OPENSTACK

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -62,6 +63,15 @@ var (
 	_ fi.HasDependencies = &Instance{}
 )
 
+// Constants for truncating Tags
+const MAX_TAG_LENGH_OPENSTACK = 60
```
zetaab, line 67: typo

upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -62,6 +63,15 @@ var (
...
+
+var TRUNCATE_OPT = truncate.TruncateStringOptions{
+	MaxLength:     MAX_TAG_LENGH_OPENSTACK,
+	AlwaysAddHash: false,
+	HashLength:    6,
+}
+
```
zetaab, line 74: typo

zetaab (changes requested):

akkina2107:
@zetaab please verify

akkina2107:
@zetaab kindly review

olemarkus:
Probably want to cherry pick this one to release-1.25

PR #13854, opened 2022-06-22
increase backoff time when updating loadbalancer pool member

zetaab:
we are currently seeing that current backoff is not enough always. This will double the Steps


drekle (approved):
/lgtm

We might consider making these configurable

hakman (approved):

PR #13861, opened 2022-06-23
Be more specific when filtering OS instance ports

ederst:
This adds an additional filter for ports attached to an instance, so that only ports tagged with the cluster name will be considered.

Basically, this enables adding ports to the server independent from the provisioning process.

Without the change updating the cluster after adding custom ports it will throw the following error:

```
W0623 22:18:29.426867   27313 executor.go:139] error running task "Instance/instance-name" (9m59s remaining to succeed): found more than one port for instance edb5ab73-1234-1234-1234-01b3b5565ba4
```

### DIFF
--- a/upup/pkg/fi/cloudup/openstacktasks/instance.go
+++ b/upup/pkg/fi/cloudup/openstacktasks/instance.go
@@ -116,6 +116,29 @@ func (e *Instance) FindAddresses(context *fi.Context) ([]string, error) {
 	return nil, nil
 }
 
+// filterInstancePorts tries to get all ports of an instance tagged with the cluster name.
+// If no tagged ports are found it will return all ports of the instance, to not change the legacy behavior when there weren't tagged ports
+func filterInstancePorts(allPorts []ports.Port, clusterName string) []ports.Port {
+	clusterNameTag := fmt.Sprintf("%s=%s", openstack.TagClusterName, clusterName)
+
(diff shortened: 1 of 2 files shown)


upup/pkg/fi/cloudup/openstacktasks/instance.go
```diff
@@ -173,6 +173,7 @@ func (e *Instance) Find(c *fi.Context) (*Instance, error) {
 
 	ports, err := cloud.ListPorts(ports.ListOpts{
 		DeviceID: server.ID,
+		Tags:     fmt.Sprintf("%s=%s", openstack.TagClusterName, fi.StringValue(e.ServerGroup.ClusterName)),
```
zetaab, line 176: the problem that I see is that tags were added one year ago in kOps. If we have now clusters which are created before that, this is not going to work (because ports are missing tags).

Should we have way to rollback to old behaviour if we cannot find any port?

ederst: > Should we have way to rollback to old behaviour if we cannot find any port?

Either that, or what about using the port name instead of the tag then?

ederst: I thought about it and think it is a good idea to just fall back to the old behavior in case no port could be identified by tag.

Will code that later today.

PR #13892, opened 2022-06-27
aws: introduce maximum instance lifetime in cluster

### LINKED ISSUE #13820: maximum instance lifetime setting for AWS ASG
👋🏼 
I didn't find any settings related to [maximum instance lifetime](https://aws.amazon.com/about-aws/whats-new/2019/11/amazon-ec2-auto-scaling-supports-max-instance-lifetime/) in kops code. That is why I propose to implement this feature in kops.

Maximum instance lifetime can help to introduce ephemerality in a Kubernetes environment. 
From the AWS documentation:
> it helps you ensure that instances are recycled before reaching the specified lifetime, giving you an automated way to adhere to your security, compliance, and performance requirements.

I would be happy to implement this new feature.

sterchelen:
This patches set enables to set [maximum instance lifetime](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-max-instance-lifetime.html) for an AWS autoscaling group

Closes #13820

### DIFF
--- a/docs/instance_groups.md
+++ b/docs/instance_groups.md
@@ -320,3 +320,19 @@ spec:
     httpPutResponseHopLimit: 1
     httpTokens: required
 ```
+
+## maxInstanceLifetime (AWS Only)
+
+{{ kops_feature_table(kops_added_default='1.24') }}
+
+The maximum instance lifetime specifies the maximum amount of time (in go duration [format](https://pkg.go.dev/time#ParseDuration)) that an instance can be in service before it is terminated and replaced.
+A common use case might be a requirement to replace your instances on a schedule because of internal security policies or external compliance controls. 
+In other words, this feature helps you to put a bit of ephemerality in your cluster.
+You must specify a value of at least 24h (86,400 seconds). To clear a previously set value, specify a new value of 0.
+
+The following configuration enables a maximum instance lifetime to two days.
+
+```yaml
+spec:
+  maxInstanceLifetime: "48h"
+```
--- a/pkg/apis/kops/instancegroup.go
+++ b/pkg/apis/kops/instancegroup.go
@@ -196,6 +196,9 @@ type InstanceGroupSpec struct {
 	Packages []string `json:"packages,omitempty"`
 	// GuestAccelerators configures additional accelerators
 	GuestAccelerators []AcceleratorConfig `json:"guestAccelerators,omitempty"`
+	// MaxInstanceLifetime to the maximum amount of time, in seconds, that an instance can be in service.
+	// Value expected must be in form of duration ("ms", "s", "m", "h")
+	MaxInstanceLifetime *metav1.Duration `json:"maxInstanceLifetime,omitempty"`
 }
 
 const (
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/kops/pkg/apis/kops"
@@ -98,6 +99,21 @@ func awsValidateInstanceGroup(ig *kops.InstanceGroup, cloud awsup.AWSCloud) fiel
 
 	if ig.Spec.CPUCredits != nil {
 		allErrs = append(allErrs, awsValidateCPUCredits(field.NewPath("spec"), &ig.Spec, cloud)...)
+	}
+
+	if ig.Spec.MaxInstanceLifetime != nil {
+		allErrs = append(allErrs, awsValidateMaximumInstanceLifetime(field.NewPath(ig.GetName(), "spec"), ig.Spec.MaxInstanceLifetime)...)
+	}
+
+	return allErrs
+}
+
+func awsValidateMaximumInstanceLifetime(fieldPath *field.Path, maxInstanceLifetime *metav1.Duration) field.ErrorList {
+	allErrs := field.ErrorList{}
+	const minMaxInstanceLifetime = 86400
+	lifetimeSec := int64(maxInstanceLifetime.Seconds())
+	if lifetimeSec != 0 && lifetimeSec < minMaxInstanceLifetime {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("maxInstanceLifetime"), maxInstanceLifetime, fmt.Sprintf("max instance lifetime must be greater than %d or equal to 0", int64(minMaxInstanceLifetime))))
 	}
 
 	return allErrs
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -547,5 +547,11 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 		}
 	}
 
+	if ig.Spec.MaxInstanceLifetime != nil {
+		lifetimeSec := int64(ig.Spec.MaxInstanceLifetime.Seconds())
+		t.MaxInstanceLifetime = fi.Int64(lifetimeSec)
+	} else {
+		t.MaxInstanceLifetime = fi.Int64(0)
+	}
 	return t, nil
 }
--- a/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
+++ b/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
@@ -37,7 +37,7 @@ import (
 // CloudTagInstanceGroupRolePrefix is a cloud tag that defines the instance role
 const CloudTagInstanceGroupRolePrefix = "k8s.io/role/"
 
-// AutoscalingGroup provdes the definition for a autoscaling group in aws
+// AutoscalingGroup provides the definition for a autoscaling group in aws
 // +kops:fitask
 type AutoscalingGroup struct {
 	// Name is the name of the ASG
@@ -53,6 +53,8 @@ type AutoscalingGroup struct {
 	LaunchTemplate *LaunchTemplate
 	// LoadBalancers is a list of elastic load balancer names to add to the autoscaling group
 	LoadBalancers []*ClassicLoadBalancer
+	// MaxInstanceLifetime is the maximum amount of time, in seconds, that an instance can be in service.
+	MaxInstanceLifetime *int64
 	// MaxSize is the max number of nodes in asg
 	MaxSize *int64
 	// Metrics is a collection of metrics to monitor
@@ -110,9 +112,17 @@ func (e *AutoscalingGroup) Find(c *fi.Context) (*AutoscalingGroup, error) {
 	}
 
 	actual := &AutoscalingGroup{
-		Name:    g.AutoScalingGroupName,
-		MaxSize: g.MaxSize,
-		MinSize: g.MinSize,
+		Name:                g.AutoScalingGroupName,
+		MaxSize:             g.MaxSize,
+		MinSize:             g.MinSize,
+		MaxInstanceLifetime: g.MaxInstanceLifetime,
+	}
+
+	// Use 0 as default value when api returns nil (same as model)
+	if g.MaxInstanceLifetime == nil {
+		actual.MaxInstanceLifetime = fi.Int64(0)
+	} else {
+		actual.MaxInstanceLifetime = g.MaxInstanceLifetime
 	}
 
 	actual.LoadBalancers = []*ClassicLoadBalancer{}
@@ -344,6 +354,13 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 			VPCZoneIdentifier:                fi.String(strings.Join(e.AutoscalingGroupSubnets(), ",")),
 		}
 
+		//On ASG creation 0 value is forbidden
+		if fi.Int64Value(e.MaxInstanceLifetime) == 0 {
+			request.MaxInstanceLifetime = nil
+		} else {
+			request.MaxInstanceLifetime = e.MaxInstanceLifetime
+		}
+
 		for _, k := range e.LoadBalancers {
 			if k.LoadBalancerName == nil {
 				lbDesc, err := t.Cloud.FindELBByNameTag(fi.StringValue(k.GetName()))
@@ -521,6 +538,13 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 		if changes.Subnets != nil {
 			request.VPCZoneIdentifier = aws.String(strings.Join(e.AutoscalingGroupSubnets(), ","))
 			changes.Subnets = nil
+		}
+
+		if changes.MaxInstanceLifetime != nil {
+			request.MaxInstanceLifetime = e.MaxInstanceLifetime
+			changes.MaxInstanceLifetime = nil
+		} else {
+			request.MaxInstanceLifetime = fi.Int64(0)
 		}
 
 		var updateTagsRequest *autoscaling.CreateOrUpdateTagsInput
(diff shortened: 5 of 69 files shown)


Makefile
```diff
@@ -167,7 +167,11 @@ hooks: # Install Git hooks
 
 .PHONY: test
 test:
+ifndef RUN
```
olemarkus, line 170: `RUN` is perhaps a bit too generic. Maybe `TEST_RUN` is a better name?

docs/instance_groups.md
```diff
@@ -320,3 +320,19 @@ spec:
     httpPutResponseHopLimit: 1
     httpTokens: required
 ```
+
+## maxInstanceLifetime (AWS Only)
+
+{{ kops_feature_table(kops_added_default='1.24') }}
```
olemarkus, line 326: Given that we are (hopefully) days away from releasing 1.24, this probably will be 1.25.

olemarkus (changes requested):
Thanks for this contribution. I think it looks generally good, but I want to do some smoke testing as well.

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -234,6 +236,10 @@ func (e *AutoscalingGroup) Find(c *fi.Context) (*AutoscalingGroup, error) {
 		}
 	}
 
+	if g.MaxInstanceLifetime != nil {
+		actual.MaxInstanceLifetime = g.MaxInstanceLifetime
```
hakman, line 240: Move above to the `actual := &AutoscalingGroup{}` section.

Makefile
hakman: Not quite sure we need this at all.

pkg/apis/kops/validation/aws.go
```diff
@@ -100,6 +100,20 @@ func awsValidateInstanceGroup(ig *kops.InstanceGroup, cloud awsup.AWSCloud) fiel
...
+	}
+
+	return allErrs
+}
+
+func awsValidateMaximumInstanceLifetime(fieldPath *field.Path, maxInstanceLifetime *int64) field.ErrorList {
+	allErrs := field.ErrorList{}
+	const minMaxInstanceLifetime = 86400
+	if *maxInstanceLifetime < minMaxInstanceLifetime {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("maxInstanceLifetime"), maxInstanceLifetime, "max instance lifetime must be greater than "+strconv.Itoa(minMaxInstanceLifetime)))
```
hakman, line 114:
```suggestion
		allErrs = append(allErrs, field.Invalid(fieldPath.Child("maxInstanceLifetime"), maxInstanceLifetime, fmt.Sprintf("max instance lifetime must be greater than %d", int64(minMaxInstanceLifetime))))
```

pkg/apis/kops/validation/aws.go
```diff
@@ -100,6 +100,20 @@ func awsValidateInstanceGroup(ig *kops.InstanceGroup, cloud awsup.AWSCloud) fiel
...
+		allErrs = append(allErrs, awsValidateMaximumInstanceLifetime(field.NewPath(ig.GetName(), "spec"), ig.Spec.MaxInstanceLifetime)...)
+	}
+
+	return allErrs
+}
+
+func awsValidateMaximumInstanceLifetime(fieldPath *field.Path, maxInstanceLifetime *int64) field.ErrorList {
+	allErrs := field.ErrorList{}
+	const minMaxInstanceLifetime = 86400
+	if *maxInstanceLifetime < minMaxInstanceLifetime {
```
hakman, line 113:
```suggestion
	if fi.Int64Value(maxInstanceLifetime) < minMaxInstanceLifetime {
```

hakman (commented):
It looks mostly good from my end too. Thanks for taking the time to add this feature.

Makefile
sterchelen: Indeed this not a game changer but it helped to avoid running the whole test suite.

hakman: It seems a good idea, but maybe for separate PR. :)

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -542,5 +542,9 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 		t.MixedSpotMaxPrice = ig.Spec.MaxPrice
 	}
 
+	if ig.Spec.MaxInstanceLifetime != nil {
```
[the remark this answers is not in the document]
hakman: Should this block be removed also now?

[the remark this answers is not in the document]
sterchelen: Indeed. Thx.

hakman:
@sterchelen Please also squash the commits. Other than this...
/lgtm

hakman:
Yes, please squash all. Thanks!

pkg/apis/kops/instancegroup.go
```diff
@@ -196,6 +196,8 @@ type InstanceGroupSpec struct {
 	Packages []string `json:"packages,omitempty"`
 	// GuestAccelerators configures additional accelerators
 	GuestAccelerators []AcceleratorConfig `json:"guestAccelerators,omitempty"`
+	// MaxInstanceLifetime is the maximum amount of time, in seconds, that an instance can be in service.
+	MaxInstanceLifetime *int64 `json:"maxInstanceLifetime,omitempty"`
```
olemarkus, line 200: maybe `*metav1.Duration` is a better type for this. so one can specify "1d30h" as value.

pkg/apis/kops/validation/aws.go
```diff
@@ -100,6 +100,20 @@ func awsValidateInstanceGroup(ig *kops.InstanceGroup, cloud awsup.AWSCloud) fiel
...
 	}
 
+	if ig.Spec.MaxInstanceLifetime != nil {
+		allErrs = append(allErrs, awsValidateMaximumInstanceLifetime(field.NewPath(ig.GetName(), "spec"), ig.Spec.MaxInstanceLifetime)...)
+	}
+
+	return allErrs
+}
+
+func awsValidateMaximumInstanceLifetime(fieldPath *field.Path, maxInstanceLifetime *int64) field.ErrorList {
```
olemarkus, line 110: Could consider 0 as a legal value as it means "instance lives forever".

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -397,7 +397,8 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 			"GroupTotalInstances",
 		},
 
-		InstanceProtection: fi.Bool(false),
+		InstanceProtection:  fi.Bool(false),
+		MaxInstanceLifetime: ig.Spec.MaxInstanceLifetime,
```
olemarkus, line 401: Here you may want to set `fi.Int64(0)` if `ig.Spec.MaxInstanceLifetime` is `nil`. In order to unset the value with AWS, you have to pass `0`. Not setting the value will mean the value should be unchanged.

This is a similar issue to https://github.com/kubernetes/kops/issues/13849 (fixed by https://github.com/kubernetes/kops/pull/13852)

olemarkus (changes requested):

sterchelen:
@olemarkus, I did all the changes you've asked for.

pkg/apis/kops/instancegroup.go
sterchelen: I did the change b/c it really makes sense, though the authorized values are "ms", "s", "m", "h" not "d"

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -542,5 +542,11 @@ func (b *AutoscalingGroupModelBuilder) buildAutoScalingGroupTask(c *fi.ModelBuil
 		t.MixedSpotMaxPrice = ig.Spec.MaxPrice
 	}
 
+	if ig.Spec.MaxInstanceLifetime != nil {
+		lifetimeSec := int64(ig.Spec.MaxInstanceLifetime.Seconds())
+		t.MaxInstanceLifetime = fi.Int64(lifetimeSec)
+	} else {
+		t.MaxInstanceLifetime = fi.Int64(0)
```
rifelpet, line 549: the e2e test is failing with:

`Error: error running tasks: deadline exceeded executing task AutoscalingGroup/master-eu-west-1a.masters.e2e-pr13892.pull-kops-e2e-kubernetes-aws.test-cncf-aws.k8s.io. Example error: error creating AutoScalingGroup: maxInstanceLifetime must be between 86400 and 31536000 seconds (inclusive).`

```suggestion
```

and then `./hack/update-expected.sh` should remove all of the `max_instance_lifetime = 0` below

sterchelen: Thx, fixed.

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -110,9 +112,10 @@ func (e *AutoscalingGroup) Find(c *fi.Context) (*AutoscalingGroup, error) {
 	}
 
 	actual := &AutoscalingGroup{
-		Name:    g.AutoScalingGroupName,
-		MaxSize: g.MaxSize,
-		MinSize: g.MinSize,
+		Name:                g.AutoScalingGroupName,
+		MaxSize:             g.MaxSize,
+		MinSize:             g.MinSize,
+		MaxInstanceLifetime: g.MaxInstanceLifetime,
```
olemarkus, line 118: So if you try to remove `maxInstanceLifetime` from the IG, the effect won't be applied.

The reason for this may be that this one is not set to `0` if AWS returns `nil`.

olemarkus:
The 1.24 was pushed for a week, so if we can get this done soon, we can merge it into 1.24

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
sterchelen: I did the change on this commit 58b15841e8bb468f1b7d38870013f31322a8c361.
Is there any e2e tests that could help me validate the intended logic?

olemarkus: Unfortunately you need to manually test this by creating a cluster and adding/removing the setting.

olemarkus: The change in the mentioned commit won't work on its own. You need to feed `fi.Int64(0)` into the task from the model.

sterchelen: > You need to feed fi.Int64(0) into the task from the model.

So IIURC @olemarkus, I need to feed `fi.Int64(0)` by default into the model? here --> https://github.com/kubernetes/kops/pull/13892/files#diff-ae2fb14fcb27f448d20d1943a0c3164d3692f812bec7b0d1e3349e93de2bbd2cR545-R547

I thought that setting `fi.Int64(0)` if and only if there is an existing ASG & the `MaxInstanceLifetime`'s spec is set to nil. This is what the aforementioned commit was expected to do.

olemarkus: Correct

sterchelen: Ok, but it means that everybody using kops in an AWS environment would have all their ASG changed in terraform :/.

Change in fc0c1274031a9abe9c1487dafb60e7597e870975

sterchelen: This change breaks e2e tests 
```
Example error: error creating AutoScalingGroup: maxInstanceLifetime must be between 86400 and 31536000 seconds (inclusive).
```
It seems that 0 is only allowed when updating an ASG.

olemarkus: Pass nil to create if the value is 0.

Terraform resources will change because of this, yes. But ASGs change on kops updates anyway

sterchelen: @olemarkus After introducing the change to pass nil in asg creation, tests are broken.
We pass `MaxInstanceLifetime:0` from spec to model and after creating the ASG we check if there is a change... and there is indeed a change between what we have in aws api and the model.
```
model: MaxInstanceLifetime: 0
api: MaxInstanceLifetime: nil
```

 https://github.com/kubernetes/kops/blob/34490d07cf8bf396a21703f02d8ccf1973c36074/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go#L101

 wdyt if in ☝🏼 we also use 0 as default value to avoid the difference between model & api?
```go
	if g.MaxInstanceLifetime == nil {
		actual.MaxInstanceLifetime = fi.Int64(0)
	} else {
		actual.MaxInstanceLifetime = g.MaxInstanceLifetime
	}
```

Extract from tests output:
```
I0704 16:09:43.619841 1826014 executor.go:111] Tasks: 85 done / 85 total; 0 can run
Will modify resources:
  AutoscalingGroup/master-us-test-1a.masters.minimal.example.com
        MaxInstanceLifetime      <nil> -> 0

I0704 16:09:43.619960 1826014 featureflag.go:164] FeatureFlag "SpecOverrideFlag"=false
--- FAIL: TestLifecycleMinimalAWS (2.84s)
    lifecycle_integration_test.go:194: running lifecycle test for cluster minimal.example.com
    lifecycle_integration_test.go:205: Target had changes after executing: Will modify resources:
          AutoscalingGroup/master-us-test-1a.masters.minimal.example.com
                MaxInstanceLifetime      <nil> -> 0
```

olemarkus: Yeah. So if API returns `nil`, you need to set `actual.MaxInstanceLifetime` to `0`

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -519,6 +529,17 @@ func (v *AutoscalingGroup) RenderAWS(t *awsup.AWSAPITarget, a, e, changes *Autos
 			changes.Subnets = nil
 		}
 
+		if changes.MaxInstanceLifetime != nil {
+			request.MaxInstanceLifetime = e.MaxInstanceLifetime
+			changes.MaxInstanceLifetime = nil
+		} else {
+			//If max instance lifetime has been removed from spec and we have a max instance lifetime
+			//on the "physical" ASG, we need to reset the value to 0
+			if a.MaxInstanceLifetime != nil {
```
olemarkus, line 538: I don't think `a.MaxInstanceLifetime` can ever be nil. If we unset `MaxInstanceLifetime` in the spec, this value would be `0`, otherwise it would be `> 0`. In other cases, `changes.MaxInstanceLifetime` would be `nil` and we'd never enter this block.

sterchelen: Fixed in 5073d5b

upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
sterchelen: Fixed in 0d7b8d9453c46c4c78f771627efee26aa30620ca

olemarkus:
Can you rebase and squash?

docs/instance_groups.md
```diff
@@ -320,3 +320,19 @@ spec:
     httpPutResponseHopLimit: 1
     httpTokens: required
 ```
+
+## maxInstanceLifetime (AWS Only)
+
+{{ kops_feature_table(kops_added_default='1.25') }}
```
olemarkus, line 326: We can let this one go into 1.24 now

docs/instance_groups.md
```diff
@@ -320,3 +320,19 @@ spec:
...
+The maximum instance lifetime specifies the maximum amount of time (in seconds) that an instance can be in service before it is terminated and replaced.
+A common use case might be a requirement to replace your instances on a schedule because of internal security policies or external compliance controls. 
+In other words, this feature helps you to put a bit of ephemerality in your cluster.
+You must specify a value of at least 86,400 seconds (one day).
+
+The following configuration enables a maximum instance lifetime to 172,800 seconds (two days).
+
+```yaml
+spec:
+  maxInstanceLifetime: 87000
```
olemarkus, line 337: Maybe use duration as example? I also believe that this needs quoting otherwise.

olemarkus (changes requested):
Just a couple of minor nits left.

docs/instance_groups.md
sterchelen: Fixed in 43f0e2f

docs/instance_groups.md
sterchelen: Fixed in 43f0e2f

sterchelen: Fixed in 43f0e2f

olemarkus (approved):
Thanks for the solid work

PR #13897, opened 2022-06-28
Apply PKI even if addon fails

olemarkus:
It looks like `kubectl replace` fails on older k8s versions, while it does not on newer. It could be differences in the OpenAPI spec that causes this. This PR makes channels install PKI even if replace fails. The addon will still be listed as `not updated`, but in most cases, addons will still work. This doens't make the problem go away entirely, but there are fewer cases of a cluster not validating.

Channels will keep trying to apply changes and fail, but this is not a regression from 1.24.


channels/pkg/channels/apply.go
```diff
@@ -50,6 +50,11 @@ func Apply(data []byte) error {
 		if err != nil {
 			klog.Errorf("failed to apply the manifest: %v", err)
 		}
+		/*
+			if err == nil {
+				return nil
+			}
+		*/
```
hakman, line 57: Any reason for this comment to be added?
/lgtm

olemarkus: Yeah ... I was contemplating if we should do this or not.

Plus side is that more addons will be considered updated for lower k8s versions.
Downside is that in some cases, e.g containers we have removed from the manifest won't be won't be removed from actual.

Maybe apply, replace, apply and only check the error of the last apply is the way to go.

hakman: > Maybe apply, replace, apply and only check the error of the last apply is the way to go.

Sounds reasonable to me.

olemarkus:
In upgrade tests, A=latest means latest CI, while B=latest means current branch. So since CI is already failing tests, the test failure is expected. That pull-kops-e2e-aws-upgrade-123-ko123-to-klatest-kolatest-many-addons passes is a good sign.

I think this one can be merged now.

PR #13898, opened 2022-06-28
Document OpenStack CSI breaking changes

ederst:
This documents the changes done in https://github.com/kubernetes/kops/pull/13603 as breaking changes since it is necessary to manually delete the old `StatefulSet` when upgrading from an old kOps version.


zetaab (approved):

zetaab:
@ederst BuildPrune should do the trick. It should delete the old manifests afaik

zetaab:
@ederst did you test this with newest kops? If I remember correctly BuildPrune was doing the cleanup of old resources and that was why it was not documented

PR #13901, opened 2022-06-28
Use Calico v3.23 for Kubernetes 1.22+

hakman:
Ref: https://projectcalico.docs.tigera.io/getting-started/kubernetes/requirements#supported-versions
> We test Calico v3.23 against the following Kubernetes versions.
> * v1.21
> * v1.22
> * v1.23


hakman:
Looks like the Calico upgrade is more complicated.
/hold cancel

olemarkus (approved):

PR #13903, opened 2022-06-28
gce: Move out of beta, drop feature flag

justinsb:
GCE support seems stable now, and we have good clarity at the API
level and how that translates to GCP resources, which was our blocker
previously.

Drop the need for the feature-flag.


justinsb:
cc @hakman as you've suggested this in the past!

I'm inclined to do this for our upcoming 1.24, as I don't _think_ there's anything in the 1.25 (other than #13902, which I just sent) and that suggsts that GCE is in good shape.

hakman (approved):

hakman:
@justinsb Cherry-pick to 1.24?

PR #13917, opened 2022-06-29
Update Cilium to 1.11.6

ReillyBrogan:
- Upstream changed the default health port to 9879 so add support for that in a backwards compatible way ([ref](https://github.com/cilium/cilium/pull/19830))
- Upstream also removed the `scheduler.alpha.kubernetes.io/critical-pod` annotation from their manifests so do the same to ours as the last version of Kubernetes to support that annotation was 1.16 which is not supported by Kops ([ref](https://github.com/cilium/cilium/pull/18667))

### DIFF
--- a/pkg/apis/kops/validation/validation_test.go
+++ b/pkg/apis/kops/validation/validation_test.go
@@ -921,7 +921,7 @@ func Test_Validate_Cilium(t *testing.T) {
 		},
 		{
 			Cilium: kops.CiliumNetworkingSpec{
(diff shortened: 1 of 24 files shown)


ReillyBrogan:
This should be cherry-picked to 1.24 as well (or at least the manifest changes if we tend to avoid updating Cilium mid-cycle).

olemarkus:
1.24 will most likely be released the coming days, so won't cherry pick this one right now. Will consider it for 1.24.1

PR #13922, opened 2022-06-30
Mount /etc/hosts from host for CoreDNS

hakman:
This should fix the failing `ServiceAccounts ServiceAccountIssuerDiscovery should support OIDC discovery of service account issuer` test for GCE:
https://testgrid.k8s.io/kops-gce#kops-grid-gce-calico-u2004-k24-containerd


hakman:
`pull-kops-e2e-kubernetes-do-gossip` fails the StatefulSet tests, same as in periodic tests, due to a DO CSI driver bug:
https://testgrid.k8s.io/kops-misc#e2e-kops-do-calico

olemarkus (approved):

Issue #13945, opened 2022-07-05
Adding GCE SPOT support


jonasasx:
/kind feature

There is no GCE SPOT support in kops.

I suggest adding `provisioningModel: [STANDART|SPOT]` in InstanceGroup model.

Spot instance must be also `preemptible=true` and `instance-termination-action=STOP`

PR #13946, opened 2022-07-05
Adding GCE SPOT support

jonasasx:
Resolves https://github.com/kubernetes/kops/issues/13945

### DIFF
--- a/pkg/apis/kops/v1alpha2/instancegroup.go
+++ b/pkg/apis/kops/v1alpha2/instancegroup.go
@@ -165,6 +165,11 @@ type InstanceGroupSpec struct {
 	// MaxInstanceLifetime to the maximum amount of time, in seconds, that an instance can be in service.
 	// Value expected must be in form of duration ("ms", "s", "m", "h")
 	MaxInstanceLifetime *metav1.Duration `json:"maxInstanceLifetime,omitempty"`
+	// GCPProvisioningModel: Specifies the provisioning model of the GCP instance.
+	// Valid values:
+	//   'STANDARD': (default) standard provisioning with user controlled run time, no discounts
+	//   'SPOT': heavily discounted, no guaranteed run time.
+	GCPProvisioningModel *string `json:"gcpProvisioningModel,omitempty"`
 }
 
 // InstanceMetadataOptions defines the EC2 instance metadata service options (AWS Only)
--- a/pkg/model/gcemodel/autoscalinggroup.go
+++ b/pkg/model/gcemodel/autoscalinggroup.go
@@ -89,8 +89,8 @@ func (b *AutoscalingGroupModelBuilder) buildInstanceTemplate(c *fi.ModelBuilderC
 				BootDiskSizeGB: i64(int64(volumeSize)),
 				BootDiskImage:  s(ig.Spec.Image),
 
-				// TODO: Support preemptible nodes?
-				Preemptible: fi.Bool(false),
+				Preemptible:          fi.Bool(fi.StringValue(ig.Spec.GCPProvisioningModel) == "SPOT"),
+				GCPProvisioningModel: ig.Spec.GCPProvisioningModel,
 
 				HasExternalIP: fi.Bool(b.Cluster.Spec.Topology.Masters == kops.TopologyPublic),
 
--- a/upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
+++ b/upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
@@ -51,10 +51,11 @@ type InstanceTemplate struct {
 
 	Lifecycle fi.Lifecycle
 
-	Network     *Network
-	Tags        []string
-	Labels      map[string]string
-	Preemptible *bool
+	Network              *Network
+	Tags                 []string
+	Labels               map[string]string
+	Preemptible          *bool
+	GCPProvisioningModel *string
 
 	BootDiskImage  *string
 	BootDiskSizeGB *int64
@@ -132,6 +133,7 @@ func (e *InstanceTemplate) Find(c *fi.Context) (*InstanceTemplate, error) {
 
 		if p.Scheduling != nil {
 			actual.Preemptible = &p.Scheduling.Preemptible
+			actual.GCPProvisioningModel = &p.Scheduling.ProvisioningModel
 		}
 		if len(p.NetworkInterfaces) != 0 {
 			ni := p.NetworkInterfaces[0]
@@ -249,7 +251,7 @@ func (e *InstanceTemplate) mapToGCE(project string, region string) (*compute.Ins
 		scheduling = &compute.Scheduling{
 			AutomaticRestart:  fi.Bool(false),
 			OnHostMaintenance: "TERMINATE",
-			ProvisioningModel: "STANDARD", // TODO: Support Spot?
+			ProvisioningModel: fi.StringValue(e.GCPProvisioningModel),
 			Preemptible:       true,
 		}
 	} else {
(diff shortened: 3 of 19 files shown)


pkg/apis/kops/v1alpha2/instancegroup.go
```diff
@@ -165,6 +165,15 @@ type InstanceGroupSpec struct {
...
 	// Value expected must be in form of duration ("ms", "s", "m", "h")
 	MaxInstanceLifetime *metav1.Duration `json:"maxInstanceLifetime,omitempty"`
+	// Preemptible: Defines whether the instance is preemptible.
+	Preemptible *bool `json:"preemptible,omitempty"`
+	// ProvisioningModel: Specifies the provisioning model of the instance.
+	//
+	// Possible values:
+	//   "SPOT" - Heavily discounted, no guaranteed runtime.
+	//   "STANDARD" - Standard provisioning with user controlled runtime, no
+	// discounts.
```
hakman, line 175: Let's use the same format as with `InstanceMetadata`. We can allow both uppercase and lowercase spelling for values.
```suggestion
	// ProvisioningModel: Specifies the provisioning model of the instance.
	// Valid values:
	//   'standard: (default) standard provisioning with user controlled run time, no discounts
	//   'spot': heavily discounted, no guaranteed run time.
```

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
```diff
@@ -238,6 +240,14 @@ func (_ *InstanceTemplate) CheckChanges(a, e, changes *InstanceTemplate) error {
 	if fi.StringValue(e.MachineType) == "" {
 		return fi.RequiredField("MachineType")
 	}
+	if e.ProvisioningModel != nil {
+		if fi.StringValue(e.ProvisioningModel) != "SPOT" && fi.StringValue(e.ProvisioningModel) != "STANDARD" {
+			return fmt.Errorf("provisioningModel must be SPOT or STANDARD")
+		}
```
hakman, line 246: This should be checked during validation.

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
```diff
@@ -238,6 +240,14 @@ func (_ *InstanceTemplate) CheckChanges(a, e, changes *InstanceTemplate) error {
...
+		if e.Preemptible == nil || fi.BoolValue(e.Preemptible) == false {
+			return fmt.Errorf("provisioningModel must be true when provisioningModel is SPOT")
```
hakman, line 248: Probably something not quite right here.

pkg/apis/kops/v1alpha2/instancegroup.go
```diff
@@ -165,6 +165,15 @@ type InstanceGroupSpec struct {
 	// MaxInstanceLifetime to the maximum amount of time, in seconds, that an instance can be in service.
 	// Value expected must be in form of duration ("ms", "s", "m", "h")
 	MaxInstanceLifetime *metav1.Duration `json:"maxInstanceLifetime,omitempty"`
+	// Preemptible: Defines whether the instance is preemptible.
+	Preemptible *bool `json:"preemptible,omitempty"`
```
hakman, line 169: Do we need to expose `Preemptible`. If I understand correctly, `SPOT` has to have `Preemptible: true` and `STANDARD` has to have ``Preemptible: false`. This can be set later in model automatically.

hakman (changes requested):

pkg/apis/kops/v1alpha2/instancegroup.go
olemarkus: Is there any way of shoehorning this into the AWS model? It would be nice to avoid having concepts that are fairly similar across clouds different in our models.

jonasasx: There is `maxPrice` field with decimal type to enable spot mode in aws. I think it is very uncompatible with gce approach with enum type

jonasasx: fixed

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
jonasasx: fixed

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
jonasasx: fixed

pkg/apis/kops/v1alpha2/instancegroup.go
jonasasx: fixed

pkg/apis/kops/v1alpha2/instancegroup.go
olemarkus: There is also these settings.
https://pkg.go.dev/k8s.io/kops/pkg/apis/kops?utm_source=godoc#MixedInstancesPolicySpec

Would be awkward, but you could say that GCP only allows 0 or 100 for `OnDemandAboveBase`

hakman: After some discussions, the new name should be `GCPProvisioningModel`. Plase update the PR.

upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
```diff
@@ -247,10 +249,11 @@ func (e *InstanceTemplate) mapToGCE(project string, region string) (*compute.Ins
...
 	if fi.BoolValue(e.Preemptible) {
 		scheduling = &compute.Scheduling{
-			AutomaticRestart:  fi.Bool(false),
-			OnHostMaintenance: "TERMINATE",
-			ProvisioningModel: "STANDARD", // TODO: Support Spot?
-			Preemptible:       true,
+			AutomaticRestart:          fi.Bool(false),
+			OnHostMaintenance:         "TERMINATE",
+			ProvisioningModel:         fi.StringValue(e.GCPProvisioningModel),
+			InstanceTerminationAction: "STOP",
```
hakman, line 255: I am trying to understand why we need stop here. Would you mind providing some context?

jonasasx: unfortunately I can not reproduce it now. So, I can remove InstanceTerminationAction here

pkg/apis/kops/v1alpha2/instancegroup.go
jonasasx: Fixed

pkg/model/gcemodel/autoscalinggroup.go
```diff
@@ -89,8 +89,8 @@ func (b *AutoscalingGroupModelBuilder) buildInstanceTemplate(c *fi.ModelBuilderC
 				BootDiskSizeGB: i64(int64(volumeSize)),
 				BootDiskImage:  s(ig.Spec.Image),
 
-				// TODO: Support preemptible nodes?
-				Preemptible: fi.Bool(false),
+				Preemptible:          fi.Bool(strings.EqualFold(fi.StringValue(ig.Spec.GCPProvisioningModel), "spot")),
+				GCPProvisioningModel: ig.Spec.GCPProvisioningModel,
```
hakman, line 93: This can be with any case because or my proposal to allow that. Probably would be best to step back on that idea and only allow `SPOT` and `STANDARD` as options. Would make things much simpler I think.

jonasasx: Rollback to uppcased variants only?

hakman: Yes please. Thanks!

jonasasx: fixed

justinsb:
This is wonderful - thank you for the contribution @jonasasx 

/approve
/lgtm

PR #13962, opened 2022-07-09
Make it possible to remove addons

olemarkus:


olemarkus:
/kind office-hours
/cc @justinsb 

@justinsb appreciate your review on this one

olemarkus:
/hold
definitely needs some release notes.

channels/pkg/channels/channel_version.go
```diff
@@ -209,3 +209,25 @@ func (c *Channel) SetInstalledVersion(ctx context.Context, k8sClient kubernetes.
 	}
 	return nil
 }
+
+// Remove removes the channel from the namespace.
```
justinsb, line 213: Nit: "Removes the channel tracking-annotation from the namespace" might be less scary-sounding!

channels/pkg/cmd/apply_channel.go
```diff
@@ -159,6 +165,10 @@ func applyMenu(ctx context.Context, menu *channels.AddonMenu, k8sClient kubernet
 		}
 	}
 
+	for key := range deletions {
+		fmt.Printf("Will deleting addon %q\n", key)
```
justinsb, line 169: Nit: Will delete addon

channels/pkg/cmd/apply_channel.go
```diff
@@ -185,6 +195,13 @@ func applyMenu(ctx context.Context, menu *channels.AddonMenu, k8sClient kubernet
 		}
 	}
 
+	for key := range deletions {
+		err := deleteAddon(ctx, k8sClient, pruner, key)
+		if err != nil {
+			merr = multierr.Append(merr, fmt.Errorf("failed to prune %q: %w", key, err))
```
justinsb, line 201: Nit: maybe "failed to delete addon %q", just for consistency.

channels/pkg/cmd/apply_channel.go
```diff
@@ -204,6 +221,80 @@ func getUpdates(ctx context.Context, menu *channels.AddonMenu, k8sClient kuberne
...
+
+	// We always include a set of well-known group kinds,
+	// so that we prune even if we end up removing something from the manifest.
+	alwaysPruneGroupKinds := []schema.GroupKind{
+		{Group: "", Kind: "ConfigMap"},
+		{Group: "", Kind: "Service"},
+		{Group: "", Kind: "ServiceAccount"},
+		{Group: "apps", Kind: "Deployment"},
+		{Group: "apps", Kind: "DaemonSet"},
+		{Group: "apps", Kind: "StatefulSet"},
```
justinsb, line 262: Aside: it is a pity this one is on here, because it could involve data loss (because it deletes the disks).  That said, I don't think any addons use StatefulSets with data.

channels/pkg/cmd/apply_channel.go
```diff
@@ -273,3 +364,20 @@ func buildMenu(kubernetesVersion semver.Version, args []string, localFiles bool)
 	}
 	return menu, nil
 }
+
+func deleteAddon(ctx context.Context, k8sClient kubernetes.Interface, pruner *channels.Pruner, key string) error {
+	parts := strings.Split(key, ":")
```
justinsb, line 369: TBH this scares me a little.  We should have a parsing function, and that function should be strict.  Not a problem for this PR, but I might add such a function.

channels/pkg/cmd/apply_channel.go
```diff
@@ -273,3 +364,20 @@ func buildMenu(kubernetesVersion semver.Version, args []string, localFiles bool)
...
+	name := parts[1]
+	pruneSpec, _ := buildDeletionPruneSpec(name)
```
justinsb, line 371: We should check `err` here.

justinsb:
So this LGTM.  I had some thoughts, but nothing blocking - I am happy to send a follow-on PR.

As removing addons is sort-of-dangerous, I'm going to hold for you to actively approve @olemarkus .  There might be a problem with karpenter also.

/approve
/lgtm
/hold for active approval

channels/pkg/cmd/apply_channel.go
olemarkus: Doesn't statefullsets preserve disks by default?

olemarkus:
I'll fix some of the nits at least.

It is a bit scary to delete addons, but I think our prune spec preserves the important bits so they can always be re-enabled.

PR #13963, opened 2022-07-11
Applier should be more tolerant of errors

justinsb:
We also aim to be a little stricter about invalid input
(e.g. namespace specified for cluster-scoped objects).


justinsb:
Follow up to #13753

PR #13965, opened 2022-07-12
Fix AWS IAM Authenticator nodeSelector in k8s 1.24 

rifelpet:
1.24 stopped including the master node-role label:

https://github.com/kubernetes/kops/blob/d0583189fdbe8627edbe371c947cff5a9b7bbea1/pkg/nodelabels/builder.go#L87-L93


tests/integration/update_cluster/complex/data/aws_s3_object_complex.example.com-addons-authentication.aws-k8s-1.12_content
```diff
@@ -197,7 +197,7 @@ spec:
           name: output
       hostNetwork: true
       nodeSelector:
-        node-role.kubernetes.io/master: ""
+        node-role.kubernetes.io/control-plane: ""
```
rifelpet, line 200: here's the bugfix being exercised

hakman (approved):

PR #13982, opened 2022-07-15
Use only IPv4 for Hetzner servers

hakman:
Recently, Hetzner started allowing to configure the Primary IPs for their servers. This makes it possible to disable IPv6.

/cc @justinsb @olemarkus @rifelpet


justinsb:
Thanks @hakman 

/approve
/lgtm

Issue #13931, opened 2022-06-30
pod-identity-webhook forces replica 2


richard-scott:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
 
` Version 1.23.2 (git-4125bbbe975ca104c609ceaa625a9a5ff3ac19f4)`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.4", GitCommit:"e6c093d87ea4cbb530a7b2ae91e54c0842d8308a", GitTreeState:"clean", BuildDate:"2022-02-16T12:38:05Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.8", GitCommit:"a12b886b1da059e0190c54d09c5eab5219dd7acf", GitTreeState:"clean", BuildDate:"2022-06-16T05:51:36Z", GoVersion:"go1.17.11", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

Try and build a 1/master 1/node cluster with this enabled:

```
spec:
  certManager:
    enabled: true
  podIdentityWebhook:
    enabled: true

```
**5. What happened after the commands executed?**

The cluster starts to build but never becomes healthy.

**6. What did you expect to happen?**

I would expect the cluster to become healthy.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2022-06-30T19:52:59Z"
  generation: 1
  name: eu.co.uk
spec:
  api:
    dns: {}
  authorization:
    rbac: {}
  awsLoadBalancerController:
    enabled: true
  certManager:
    enabled: true
  channel: stable
  cloudProvider: aws
  configBase: s3://eu-co-uk-state-store/eu.co.uk
  dnsZone: eu.co.uk
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - encryptedVolume: true
[...]

PR #13986, opened 2022-07-16
Add option to set number of replicas for pod-identity-webhook

hakman:
Refs: #13931

### DIFF
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -223,7 +223,8 @@ type ClusterSpec struct {
 
 // PodIdentityWebhookConfig configures an EKS Pod Identity Webhook.
 type PodIdentityWebhookConfig struct {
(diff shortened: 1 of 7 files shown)


pkg/apis/kops/cluster.go
```diff
@@ -223,7 +223,8 @@ type ClusterSpec struct {
 
 // PodIdentityWebhookConfig configures an EKS Pod Identity Webhook.
 type PodIdentityWebhookConfig struct {
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled  bool `json:"enabled,omitempty"`
+	Replicas int  `json:"replicas,omitempty"`
```
justinsb, line 227: I was wondering if it would make sense to set replicas to 0, but I guess someone that wanted to do that would set enabled = false instead.

justinsb:
Thanks @hakman 

/approve
/lgtm

PR #14016, opened 2022-07-21
fix hyperlinks in calico docs

mostafahussein:
This PR fixes hyperlinks in Calico docs in order to be displayed properly


hakman (approved):
Thanks for fixing the links @mostafahussein.

PR #14017, opened 2022-07-21
Allow configuring OpenStack CCM networking options

ederst:
This will allow to configure the [OpenStack CCM networking options](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#networking).

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -479,6 +479,16 @@ spec:
                             items:
                               type: string
                             type: array
+                          internalNetworkNames:
+                            items:
+                              type: string
+                            type: array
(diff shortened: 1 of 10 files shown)


zetaab (approved):
looks ok, but just wondering can this be used with kops? kops does not add multiple interfaces to single instance?

ederst:
@zetaab no kops does not allow multiple openstack interfaces, but this can be beneficial when:
* interfaces get attached afterwards, via a controller or other provisioning logic
* when using subnets where a ipv6 is automatically attached, and no ipv6 should be reported to the k8s node resource
* using subnets where occm deems it is public and the net should be handles as private (or vice versa)

PR #14018, opened 2022-07-21
Add server group management for Hetzner

hakman:
/cc @justinsb

### DIFF
--- a/upup/pkg/fi/cloudup/hetznertasks/server.go
+++ b/upup/pkg/fi/cloudup/hetznertasks/server.go
@@ -18,8 +18,11 @@ package hetznertasks
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/base64"
 	"fmt"
-	"strconv"
+	"math/rand"
+	"strings"
 
 	"github.com/hetznercloud/hcloud-go/hcloud"
 	"k8s.io/kops/upup/pkg/fi"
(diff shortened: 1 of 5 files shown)


upup/pkg/fi/cloudup/hetznertasks/server.go
```diff
@@ -46,123 +51,131 @@ type Server struct {
...
+	labelSelector := []string{
+		fmt.Sprintf("%s=%s", hetzner.TagKubernetesClusterName, c.Cluster.Name),
+		fmt.Sprintf("%s=%s", hetzner.TagKubernetesInstanceGroup, v.Labels[hetzner.TagKubernetesInstanceGroup]),
+	}
+	listOptions := hcloud.ListOpts{
+		PerPage:       50,
+		LabelSelector: strings.Join(labelSelector, ","),
+	}
+	serverListOptions := hcloud.ServerListOpts{ListOpts: listOptions}
+	servers, err := client.AllWithOpts(context.TODO(), serverListOptions)
```
justinsb, line 67: Unrelated: We should probably just put the context.Context into *fi.Context; there's some guidance that we shouldn't do so but I think it's fine, particularly if the wrapper object doesn't outlast the context.Context.

justinsb:
I like this approach, and as it is specific to hetzner (which isn't yet fully GA)...

/approve
/lgtm

hakman:
Thanks @justinsb! :)

PR #14024, opened 2022-07-24
Revert to using instance private DNS name to lookup hostname

hakman:
https://github.com/kubernetes/kops/pull/13542 explains the reasons behind this change.
Also the various reports on the issue: https://github.com/kubernetes/kops/issues/13540 https://github.com/kubernetes/kops/issues/13492 and https://github.com/kubernetes/kops/issues/13358.

/cc @justinsb @olemarkus

### DIFF
--- a/upup/pkg/fi/nodeup/command.go
+++ b/upup/pkg/fi/nodeup/command.go
@@ -31,8 +31,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/autoscaling"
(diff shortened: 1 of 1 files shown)


olemarkus:
This should only target 1.23, no? In 1.24, this is not a problem since we use external CCM.

/hold

olemarkus:
Or at least conditional based on if CCM is used and k8s is new enough.

hakman:
> This should only target 1.23, no? In 1.24, this is not a problem since we use external CCM.

This should target master also, as it needs to apply to older k8s versions too, right?
Isn't the condition from resource based names good enough?
https://github.com/kubernetes/kops/blob/89faa0a865798f52336fa54a813d48ebe67233a2/pkg/apis/nodeup/config.go#L221

olemarkus:
Yeah you are right. I completely missed that condition.

/hold cancel
/approve
/lgtm

PR #14034, opened 2022-07-27
Enable rolling updates for Hetzner

hakman:
Follow-up to #14018.

/cc @justinsb @olemarkus

### DIFF
--- a/upup/pkg/fi/cloudup/hetzner/cloud.go
+++ b/upup/pkg/fi/cloudup/hetzner/cloud.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
+	"time"
 
 	"github.com/hetznercloud/hcloud-go/hcloud"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -257,21 +301,118 @@ func (c *hetznerCloudImplementation) Region() string {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to build cloud instance group for instance group %q: %w", instanceGroup.Name, err)
+		}
+	}
+
+	return cloudInstanceGroups, nil
+}
+
+// findServerGroups finds all server groups belonging to the cluster
+func findServerGroups(c *hetznerCloudImplementation, clusterName string) (map[string][]*hcloud.Server, error) {
```
justinsb, line 339: Nit: if you're only calling this from GetCloudGroups, I think you could pass the instancegroups in and directly match the instancegroups, instead of constructing the magic string (?)

upup/pkg/fi/cloudup/hetznertasks/servergroup.go
```diff
@@ -143,6 +148,28 @@ func (_ *ServerGroup) CheckChanges(a, e, changes *ServerGroup) error {
...
+				continue
+			}
+
+			server.Labels[hetzner.TagKubernetesInstanceNeedsUpdate] = ""
+			_, _, err = client.Update(context.TODO(), server, hcloud.ServerUpdateOpts{
+				Name:   server.Name,
+				Labels: server.Labels,
+			})
+			if err != nil {
+				return err
```
justinsb, line 168: Nit: when we get an error from calling the hetzner API (or any API), I don't know how good those errors are.  It might be nice to `fmt.Errorf("error setting needs-update label on %q: %w", serverName, err)` for example.

justinsb:
This is nice!

/approve
/lgtm

upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -237,14 +243,52 @@ func (c *hetznerCloudImplementation) DNS() (dnsprovider.Interface, error) {
...
+	client := c.ServerClient()
+	ctx := context.TODO()
+
+	server := &hcloud.Server{ID: id}
+	_, _, err := client.Shutdown(ctx, server)
+	if err != nil {
+		return fmt.Errorf("failed to stop server %q: %w", strconv.Itoa(id), err)
+	}
+
+	for i := 1; i <= 30; i++ {
```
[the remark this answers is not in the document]
hakman: Good point. Added on my TODO list.

PR #14057, opened 2022-07-29
Wait for load balancer to be ready for Hetzner

hakman:
`FindAddresses()` may run before the public addresses are allocated and the Kubernetes API certificate will not work.

### DIFF
--- a/upup/pkg/fi/cloudup/hetznertasks/loadbalancer.go
+++ b/upup/pkg/fi/cloudup/hetznertasks/loadbalancer.go
@@ -54,11 +54,12 @@ func (e *LoadBalancer) IsForAPIServer() bool {
 }
 
 func (v *LoadBalancer) FindAddresses(c *fi.Context) ([]string, error) {
+	ctx := context.TODO()
 	cloud := c.Cloud.(hetzner.HetznerCloud)
 	client := cloud.LoadBalancerClient()
 
 	// TODO(hakman): Find using label selector
-	loadbalancers, err := client.All(context.TODO())
+	loadbalancers, err := client.All(ctx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


upup/pkg/fi/cloudup/hetznertasks/loadbalancer.go
```diff
@@ -242,20 +263,26 @@ func (_ *LoadBalancer) RenderHetzner(t *hetzner.HetznerAPITarget, a, e, changes
...
-		})
-		if err != nil {
+		// Update the targets
+		if a.Target == "" {
+			action, _, err := client.AddLabelSelectorTarget(ctx, loadbalancer, hcloud.LoadBalancerAddLabelSelectorTargetOpts{
+				Selector:     e.Target,
+				UsePrivateIP: fi.Bool(true),
+			})
 			if err != nil {
+				if err != nil {
```
justinsb, line 280: Looks look an extra check of `err != nil`

upup/pkg/fi/cloudup/hetznertasks/loadbalancer.go
```diff
@@ -242,20 +263,26 @@ func (_ *LoadBalancer) RenderHetzner(t *hetzner.HetznerAPITarget, a, e, changes
...
+					return err
+				}
+			}
+			_, errCh := actionClient.WatchProgress(ctx, action)
+			if err := <-errCh; err != nil {
```
justinsb, line 285: Nit: we might want a helper here that calls WatchProgress, but also has a timer, so that we log e.g. every minute that we're waiting for something, rather than potentially silently waiting forever.

justinsb:
Two thoughts, but lgtm.

/approve
/lgtm

upup/pkg/fi/cloudup/hetznertasks/loadbalancer.go
hakman: Good idea. Thanks!

upup/pkg/fi/cloudup/hetznertasks/loadbalancer.go
hakman: Fixed this now, to make it easy to cherry-pick.

PR #14058, opened 2022-07-29
Add multiple SSH keys support for Hetzner

hakman:


pkg/model/hetznermodel/servers.go
```diff
@@ -33,6 +34,24 @@ type ServerGroupModelBuilder struct {
 var _ fi.ModelBuilder = &ServerGroupModelBuilder{}
 
 func (b *ServerGroupModelBuilder) Build(c *fi.ModelBuilderContext) error {
+	var sshkeyTasks []*hetznertasks.SSHKey
+	for _, sshkey := range b.SSHPublicKeys {
```
justinsb, line 38: Nit: If you have two ServerGroupModelBuilders, with overlapping SSHPublicKeys, this may result in duplicate tasks.  So you _might_ want to consider keeping the dedicated SSHPublicKeyBuilder task, or using EnsureTask (which allows for duplicate creation if they are identical, but is a bit more hacky).  But not a blocker!

upup/pkg/fi/cloudup/hetznertasks/server.go
```diff
@@ -33,7 +33,7 @@ import (
 type ServerGroup struct {
 	Name      *string
 	Lifecycle fi.Lifecycle
-	SSHKey    *SSHKey
+	SSHKeys   []*SSHKey
```
justinsb, line 36: Do we set this anywhere?  I didn't see if in the diff, which is surprising (?)

justinsb:
/approve
/lgtm

Looks good to me, but you might want to keep the SSH-specific builder to avoid duplicate tasks.  So I'll add a hold but feel free to cancel the hold if the potentially duplicate tasks aren't a problem in practice!

/hold

upup/pkg/fi/cloudup/hetznertasks/server.go
hakman: We don't set this anywhere, except for model. They are used only when creating a new server and otherwise ignored.
Hetzner Cloud API doesn't return the keys for a running server.

pkg/model/hetznermodel/servers.go
hakman: The ServerGroupModelBuilder creates servers for all instance groups, so there is no need for using EnsureTask at the moment.

hakman:
> Looks good to me, but you might want to keep the SSH-specific builder to avoid duplicate tasks. So I'll add a hold but feel free to cancel the hold if the potentially duplicate tasks aren't a problem in practice!

I need an easy way to access the SSHKey tasks. As they are only used for creating servers, seems the best approach for now.

/hold cancel

hakman:
Rebasing to get past merge conflict with #14034.

PR #14072, opened 2022-07-29
WIP: Expose channels through kops cli

olemarkus:
Mount the channels API on `kops addons`. I put everything under `addons` rather than reusing `kops get` etc since especially `apply` have very different semantics.

Still some work left. In particular getting default cluster name, state store etc. Should try to merge the two CLI factories in some way.

/kind office-hours


justinsb:
Meant to comment on this during office hours; I think having `get` is good here.  Apply we might not want to encourage users to use, so I proposed we put that command under `toolbox`, so we make fewer guarantees about it etc

justinsb:
This came up in office hours, are we agreed that it should go under toolbox?  And is this still important?

olemarkus:
I think this is still important. It's always annoying to tell people they have to log into a control plane node and look at protokube logs in order to determine if channels did something wrong.

johngmyers:
Per office hours, we want this to go into toolbox.

docs/releases/1.26-NOTES.md
```diff
@@ -11,6 +11,8 @@ This is a document to gather the release notes prior to the release.
 * Control-plane instance groups in new clusters are now created with names starting
 with "control-plane-". The names of groups for existing clusters are unchanged.
 
+* The channels CLI that kOps use to manage addons is now bundled with the kOps binary. The interface is considered unstable and experimental, but useful for addon diagnostics and troubleshooting.
```
hakman, line 14:
```suggestion
* The channels CLI that kOps use to manage addons is now bundled with the kOps binary. This is useful for addons diagnostics and troubleshooting.
```

cmd/kops/toolbox_addons.go
```diff
@@ -0,0 +1,43 @@
...
+	channelscmd "k8s.io/kops/channels/pkg/cmd"
+	"k8s.io/kops/cmd/kops/util"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmdToolboxAddons(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "addons",
+		Short:         "Addons manage the kOps addons",
```
hakman, line 31:
```suggestion
		Short:         "Manage addons",
```

cmd/kops/toolbox_addons.go
```diff
@@ -0,0 +1,43 @@
...
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	f := util.NewFactory(nil)
+
+	// create subcommands
+	cmd.AddCommand(channelscmd.NewCmdApply(f, out))
+	cmd.AddCommand(channelscmd.NewCmdGet(f, out))
```
hakman, line 40:
```suggestion
	cmd.AddCommand(&cobra.Command{
		Use:     "apply CHANNEL",
		Short:   "Applies updates from the given channel",
		Example: "kops toolbox addons apply s3://<state_store>/<cluster_name>/addons/bootstrap-channel.yaml",
		RunE: func(cmd *cobra.Command, args []string) error {
			ctx := context.TODO()
			return channelscmd.RunApplyChannel(ctx, f, out, &channelscmd.ApplyChannelOptions{}, args)
		},
	})
	cmd.AddCommand(&cobra.Command{
		Use:   "list",
		Short: "Lists or gets addons",
		RunE: func(cmd *cobra.Command, args []string) error {
			ctx := context.TODO()
			return channelscmd.RunGetAddons(ctx, f, out, &channelscmd.GetAddonsOptions{})
		},
	})
```

hakman (changes requested):
Works nicely, except for a few nits:
1. subcommands are awkward, so may be nicer to just redefine them as suggested.
2. `apply` requires the channel path, but would be nice to default to the cluster channel.
```
Error: unexpected number of arguments. Only one channel may be processed at the same time.
```

cmd/kops/toolbox_addons.go
olemarkus: Hm ... clever.

hakman: Thanks. I suggest to also clean the generated `.md` files

channels/pkg/cmd/apply_channel.go
```diff
@@ -46,8 +46,10 @@ func NewCmdApplyChannel(f Factory, out io.Writer) *cobra.Command {
 	var options ApplyChannelOptions
 
 	cmd := &cobra.Command{
-		Use:   "channel",
-		Short: "Apply channel",
+		Use:     "channel CHANNEL",
```
hakman, line 49: Changes to channels should not be needed anymore.

olemarkus: Yeah, but it's still more correct now :p I think we should keep this change.

hakman: Maybe just remove the example, or update the command?

hakman (approved):

PR #14074, opened 2022-07-30
Use SSA for updating addon channel objects

olemarkus:
Current implementation is effectively a `kubectl replace` and will remove fields not in the objects (i.e those set by other controllers). This PR should make channels use server-side apply instead with a non-strategic merge.

### DIFF
--- a/channels/pkg/channels/apply.go
+++ b/channels/pkg/channels/apply.go
@@ -18,17 +18,19 @@ package channels
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


olemarkus:
/cc @rifelpet 
Seems to work when testing locally.

rifelpet (approved):

rifelpet:
confirmed this fixed the job failure: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-k120-kolatest-to-k121-kolatest-many-addons/1553443348121915392

PR #14076, opened 2022-07-31
 Remove passing cluster name as positional argument 

olemarkus:
Mixing sub commands and positional args (e.g `kops get cluster` vs `kops get foo.k8s.local`) is confusing. It also complicates the arg/flag logic. This PR deprecates the use of cluster name as positional argument and nudges people over to always using `--name`.

### DIFF
--- a/cmd/kops/root.go
+++ b/cmd/kops/root.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/kops/pkg/client/simple"
 	"k8s.io/kops/pkg/commands"
 	"k8s.io/kops/pkg/commands/commandutils"
+	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kubectl/pkg/util/i18n"
 	"k8s.io/kubectl/pkg/util/templates"
 )
@@ -247,6 +248,15 @@ func (c *RootCmd) clusterNameArgsAllowNoCluster(clusterName *string) func(cmd *c
 //  * <clustername> (and --name not specified)
 // Everything else is an error.
 func (c *RootCmd) ProcessArgs(args []string) error {
+	if len(args) > 0 {
+		fmt.Printf("\n")
+		fmt.Printf("\nClusterName as positional argument is deprecated and will be removed\n")
+		fmt.Printf("Use `KOPS_FEATURE_FLAGS=PositionalClusterArg` to revert to the old behavior.")
+		fmt.Printf("\n")
+	}
+	if !featureflag.PositionalClusterArg.Enabled() {
+		return nil
+	}
 	if len(args) == 0 {
 		return nil
 	}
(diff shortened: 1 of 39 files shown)


cmd/kops/root.go
```diff
@@ -247,6 +248,9 @@ func (c *RootCmd) clusterNameArgsAllowNoCluster(clusterName *string) func(cmd *c
 //  * <clustername> (and --name not specified)
 // Everything else is an error.
 func (c *RootCmd) ProcessArgs(args []string) error {
+	if !featureflag.PositionalClusterArg.Enabled() {
+		return nil
+	}
```
hakman, line 253:
```suggestion
	if len(args) > 0 {
		fmt.Printf("\n")
		fmt.Printf("\nClusterName as positional argument is deprecated and will be removed\n")
		fmt.Printf("Use `KOPS_FEATURE_FLAGS=PositionalClusterArg` to revert to the old behavior.")
		fmt.Printf("\n")
	}
	if !featureflag.PositionalClusterArg.Enabled() {
		return nil
	}
```

docs/releases/1.25-NOTES.md
```diff
@@ -20,6 +20,12 @@ This changes the default behavior where the CSI snaphotter container was always
 
 So in case of manually deployed CRDs to make the snapshotter work it is now necessary to [enable the snapshot controller](https://kops.sigs.k8s.io/addons/#snapshot-controller).
 
+## kOps CLI no longer accepts cluster name as positional argument
+
+Prior to kOps 1.25, one could pass cluser name as a CLI argument to most functions. E.g `kOps get <clustername>`. As of kOps 1.25, kOps commands only accepts the cluster name through the `--name` flag or the `KOPS_CLUSTER_NAME` environment variable. The one exception to this change is the `kops get cluster` command, which still accepts one or more cluster names as positional arguments.
+
+This change is behind a feature flag. One can revert to the old behavior by setting `KOPS_FEATURE_FLAGS=EnablePositionalClusterArg`.
```
hakman, line 27:
```suggestion
This change is behind a feature flag. For now, one can revert to the old behavior by setting `KOPS_FEATURE_FLAGS=PositionalClusterArg`.
```

hakman (changes requested):

hakman (approved):

justinsb:
This does break user workflows, and does deviate from what kubectl does.

I do agree that `kops get clustername` is an abberation ... we could replace that with `kops get all <clustername>` or similar, as that might be clearer all round.

We could also make this a warning and remove the functionality in a future release; right now we immediately break scripts etc I think.

olemarkus:
I think we already deviate from kubectl in many ways. My original idea was to only print a deprecation warning, but I don't have too strong opinion so I went with @hakman's suggestion. We could also drop the `--name` flag, I suppose, but I want to get rid from the ambigous parsing that comes from `ProcessArgs`.

PR #14077, opened 2022-08-01
Fix Karpenter IAM permissions and make karpenter respect IG subnets

olemarkus:


pkg/model/components/addonmanifests/karpenter/iam.go
```diff
@@ -52,17 +52,19 @@ func addKarpenterPermissions(p *iam.Policy) {
...
+		"ec2:DescribeInstanceTypes",
 		"ec2:DescribeSecurityGroups",
+		"ec2:DescribeSpotPriceHistory",
 		"ec2:DescribeSubnets",
-		"ec2:DescribeInstanceTypes",
-		"ec2:DescribeInstanceTypeOfferings",
-		"ec2:DescribeAvailabilityZones",
+		"iam:PassRole",
+		"pricing:GetProducts",
 		"ssm:GetParameter",
```
hakman, line 68: Maybe sort them as in the generated policy below:
```
"ec2:CreateFleet",
"ec2:CreateTags",
"ec2:DescribeAvailabilityZones",
"ec2:DescribeInstanceTypeOfferings",
"ec2:DescribeInstanceTypes",
"ec2:DescribeInstances",
"ec2:DescribeLaunchTemplates",
"ec2:DescribeSecurityGroups",
"ec2:DescribeSpotPriceHistory",
"ec2:DescribeSubnets",
"ec2:RunInstances",
"ec2:TerminateInstances",
"iam:PassRole",
"pricing:GetProducts",
"ssm:GetParameter"
```

olemarkus: I started doing that but apparently stopped half-way through :)

olemarkus:
Using tags-based approach that I had in my original PR.

Note that this will tag all subnets regardless of if karpenter is used or not. Figured it could be useful for other things as well.

rifelpet:
I'm wondering if we should be concerned with clusters that have a high churn of instance groups (operators frequently creating/deleting IGs) resulting in unnecessary churn on subnet tags.

I suppose if they're using terraform, they'll be running a `terraform apply` for any change in IGs anyways, so updating the subnet tags should be fine.

This is also guarded by `Cluster.Spec.TagSubnets` bool which is all or nothing but I suppose thats acceptable here. It may be worth a comment in the kops+karpenter docs around preventing that field from being set to false (or even failing API validation if karpenter is enabled)

olemarkus:
> I'm wondering if we should be concerned with clusters that have a high churn of instance groups (operators frequently creating/deleting IGs) resulting in unnecessary churn on subnet tags.
 
I think these tags would be quite minor compared to the churn on EC2 instances.
 
> This is also guarded by `Cluster.Spec.TagSubnets` bool which is all or nothing but I suppose thats acceptable here. It may be worth a comment in the kops+karpenter docs around preventing that field from being set to false (or even failing API validation if karpenter is enabled)

I don't think validation is necessary here. It's not much different than LBC or CCM failing if one doesn't take care of tags properly.

Added some docs on the tag though.

docs/operations/karpenter.md
```diff
@@ -45,16 +45,16 @@ A Karpenter-managed InstanceGroup controls a corresponding Karpenter Provisioner
 
 Note that not all features of InstanceGroups are supported.
 
+## Subnets
+
+By default, kOps will tag subnets with `kops.k8s.io/instance-group/<intancegroup>: "true"` for each InstanceGroup the subnet is assigned to. If you enable manual tagging of subnets, you have to ensure you add these tags are added, if not Karpenter will fail to provision any instances.
```
rifelpet, line 50:
```suggestion
By default, kOps will tag subnets with `kops.k8s.io/instance-group/<intancegroup>: "true"` for each InstanceGroup the subnet is assigned to. If you enable manual tagging of subnets, you have to ensure these tags are added, if not Karpenter will fail to provision any instances.
```


maybe add `{{ kops_feature_table(kops_added_default='1.25') }}` too, or whatever version this is intended to land in

olemarkus: It's feature flagged so thinking this is not necessary.

rifelpet (approved):

PR #14080, opened 2022-08-02
Create etcd-manager config for each instance group

hakman:
kOps should give a way for `etcd-manager` to identify a specific volume for a specific control-plane node. This means having access to the instance group, as defined in the cluster config:
https://github.com/kubernetes/kops/blob/70c85280917f67c9c057ec41e31ee148bd5333c1/pkg/apis/kops/cluster.go#L664-L665

This can be done either directly in `etcd-manager`, by having some hardcoded tags, or by passing a specific config for each instance group, which includes the instance group tag. My preference is for the latter.


olemarkus:
Looks good. You are only using IG-specific tags for hetzner now, right? nothing changes for AWS?

hakman:
> Looks good. You are only using IG-specific tags for hetzner now, right? nothing changes for AWS?

Yes, only Hetzner and flag is still up to debate.

pkg/model/components/etcdmanager/model.go
```diff
@@ -68,25 +65,29 @@ func (b *EtcdManagerBuilder) Build(c *fi.ModelBuilderContext) error {
 			return fmt.Errorf("backupStore must be set for use with etcd-manager")
 		}
 
-		manifest, err := b.buildManifest(etcdCluster)
-		if err != nil {
-			return err
-		}
+		for _, member := range etcdCluster.Members {
+			instanceGroupName := fi.StringValue(member.InstanceGroup)
```
justinsb, line 69: It _might_ be that we should use the member name here.  On the one hand, it's the primary key and therefore maybe "more correct", and maybe avoids the problem of then wanting to lock instance groups to members.  On the other hand, it means users must be more aware of etcd member names, when they are probably more familiar with instance group names.

justinsb: Edit: but now I see that it the mapping in nodeup below becomes much more straightforward if we stick to instance groups...

justinsb (approved):

justinsb:
Thanks @hakman this lgtm, and will help us be more precise with rolling updates etc.  I now also understand better why you want to try to keep instance groups to a specific member - that makes a lot more sense now that I see it!

/approve
/lgtm

PR #14087, opened 2022-08-03
Add back conversion struct to cert-manager CRDs

olemarkus:
A combination of apiserver map handling when using server-side apply, CA injection and resource validation made it impossible for kops to remove this struct. However, kops should claim ownership of all fields so we can remove them all in a future version


rifelpet (approved):
looks like this was removed in https://github.com/kubernetes/kops/commit/43bfb934e7b0de89e0fd3072ec39761a0bcfda22#diff-b40eb2fd32f439fb653b6875c851d9f517e39809f7e3fe177e28e9da56e5f082

PR #14088, opened 2022-08-03
Hide klog flags from --help output

justinsb:
The klog flags just clutter the output and often push the most
important output off of smaller screens.

The flags are still supported, just hidden.


hakman (approved):
👍

PR #14098, opened 2022-08-06
Test the aws ebs csi driver in e2e if installed

olemarkus:


olemarkus:
I thought it would be clever to just reuse the storage class we install by default. But the tests are failing because the test framework is copying, not reusing, the storage class. And when multiple tests run in parallel, those depending on a default class fail.

We probably need to specify a class from file instead.

tests/e2e/pkg/tester/tester.go
```diff
@@ -373,12 +373,35 @@ func (t *Tester) addNonBlockingTaintsFlag() {
 		return
 	}
 	nbt := "node-role.kubernetes.io/master,"
-	nbt += "node-role.kubernetes.io/api-server,"
+	nbt += "node-role.kubCloudProviderAWSernetes.io/api-server,"
```
hakman, line 376: This looks out of place.

rifelpet (approved):
should we update the skip regex to skip the in-tree AWS driver tests when we set these new test args?

https://github.com/kubernetes/kops/blob/master/tests/e2e/pkg/tester/skip_regex.go

PR #14105, opened 2022-08-09
cilium: fix wrong pod annotations templating #1.23

sterchelen:
This patch fixes cilium pod annotations templating for 1.9 & 1.10.

very similar to https://github.com/kubernetes/kops/pull/13536

PR #14111, opened 2022-08-10
cilium: fix wrong pod annotations templating

sterchelen:
Same as https://github.com/kubernetes/kops/pull/14105 but for latest kops version


olemarkus (approved):
/ok-to-tesdt

PR #14112, opened 2022-08-10
Use semver for skipregex ifs instead of strings.Contains

olemarkus:


tests/e2e/pkg/tester/skip_regex.go
```diff
@@ -84,27 +89,26 @@ func (t *Tester) setSkipRegexFlag() error {
 		skipRegex += "|should.be.mountable.when.non-attachable"
 		// The in-tree driver and its E2E tests use `topology.kubernetes.io/zone` but the CSI driver uses `topology.gke.io/zone`
 		skipRegex += "|In-tree.Volumes.\\[Driver:.gcepd\\].*topology.should.provision.a.volume.and.schedule.a.pod.with.AllowedTopologies"
-	} else if strings.Contains(cluster.Spec.KubernetesVersion, "v1.1") ||
-		strings.Contains(cluster.Spec.KubernetesVersion, "v1.20.") ||
-		strings.Contains(cluster.Spec.KubernetesVersion, "v1.21.") ||
-		strings.Contains(cluster.Spec.KubernetesVersion, "v1.22.") {
+	}
+
+	if k8sVersion.Minor <= 22 {
```
hakman, line 94: I think this can be removed also from the above GCE block.

hakman, line 94: I think we can use `IsKubernetesGTE` for comparing versions instead of minor.

olemarkus: The version of that function that util provide is not that great. I prefer this check.

hakman (approved):

PR #14114, opened 2022-08-11
Merge kubeletConfigs earlier

olemarkus:
There are three places where one can configure `kubeletConfig` in our API: `kubelet` and `masterKubelet` in the cluster config, and `kubelet` in the `instancegroup`. If you want to do any sort of testing or validation of the kubeletconfig you have to validate all three places now as the configs are not merged until one tries to build nodeup config.

This PR does the merge in `populateInstanceGroup` instead so that one can always test against the IGs kubelet config.

Also some minor cleanups along the way. The golden output changes that comes a part of this should be fine.

### DIFF
--- a/nodeup/pkg/model/kubelet.go
+++ b/nodeup/pkg/model/kubelet.go
@@ -261,6 +261,9 @@ func (b *KubeletBuilder) kubeletPath() string {
 
 // buildManifestDirectory creates the directory where kubelet expects static manifests to reside
 func (b *KubeletBuilder) buildManifestDirectory(kubeletConfig *kops.KubeletConfigSpec) (*nodetasks.File, error) {
+	if kubeletConfig.PodManifestPath == "" {
+		return nil, fmt.Errorf("failed to build manifest path. Path was empty")
+	}
 	directory := &nodetasks.File{
 		Path: kubeletConfig.PodManifestPath,
 		Type: nodetasks.FileType_Directory,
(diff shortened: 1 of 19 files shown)


olemarkus:
Worth mentioning that we do a lot of generic defaults also in nodeup that should be moved to cloudup.

tests/integration/create_cluster/complex/expected-v1alpha2.yaml
```diff
@@ -64,6 +64,29 @@ metadata:
   name: master-us-test-1a
 spec:
   image: 099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20220615
+  kubelet:
```
justinsb, line 67: I _think_ this is indicating that we populate these on `kops create cluster`?

justinsb:
I like the idea a lot here - nodeup should have as little logic in it as possible!

However, does `kops create cluster` now set kubeletConfig in the InstanceGroup?  We've generally tried to avoid setting default values there unless we think the user intends to "lock" them (which is different from how k8s does e.g. Deployments, but I think that's a big part of why users complain that Deployments are complicated.... because of all those fields that magically appear!)

If we can keep this in the `--full` configuration, then I think this is great... I haven't yet tried just running a `kops create cluster` on this branch though - am I misreading the test output?

olemarkus:
Hm. It's a good call. I'll do some more testing. I thought `PopulateInstanceGroupSpec` would work similarly to `PopulateClusterSpec` in that it takes the user-created spec and converts it into the full spec. So the test outputs should only have diffs in the full spec.

olemarkus:
Okay. So instancegroups works quite differently than cluster here. I wonder if this is legacy due to nodeup using the whole IG spec earlier. While now, we use nodeup config. I think we can replace writing the populated spec spec with writing the input spec like we do for cluster.

hakman (approved):

PR #13264, opened 2022-02-16
Disable some flags in kube-apiserver when logging-format is not text

h3poteto:
Fixes #13245 

Disable these flags because these are not accepted when I use `json` as `logging-format`.

* `--logtostderr`
* `--alsologtostderr`
* `--log-file`

According to Kubernetes document: 
> Non-default formats don't honor these flags: --add-dir-header, --alsologtostderr, --log-backtrace-at, --log-dir, --log-file, --log-file-max-size, --logtostderr, --one-output, --skip-headers, --skip-log-headers, --stderrthreshold, --vmodule.

https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

And when I specify these flags, kube-apiserver says an error.

PR #14115, opened 2022-08-11
Disable some flags in kube-controller-manager and kube-scheduler when logging-format is not text

### LINKED ISSUE #14100: kube-controller-manager and kube-scheduler crash after setting "logFormat: json" (need the same patch as in #13245)
/kind bug

It's basically the same issue as in bug #13245.
It was fixed in https://github.com/kubernetes/kops/pull/13264 by @h3poteto by checking if logformat is set to json.
But the fix was provided for the api-server only, kube-controller-manager and kube-proxy and kube-scheduler were not fixed.
For some reason json logging was not enabled in my cluster for kube-proxy, so it's running. But kube-scheduler and kube-controller-manager stuck in CrashLoopBackOff state after upgrade to 1.22.12 using kops 1.22.6

 this information.**
Client/Server Version 1.22.6

 a `kops` flag.**
Client/Server Version 1.21.13

AWS

Edit cluster
Add parameter
```
spec:
  kubeAPIServer:
    logFormat: json
```
kops update cluster
kops rolling-update cluster --instance-group <one of the masters>

kube-controller-manager and [kube-proxy] and kube-scheduler crash with error:
```
Error: [non-default logging format doesn't honor flag: alsologtostderr, non-default logging format doesn't honor flag: log_file, non-default logging format doesn't honor flag: logtostderr]
```

Cluster work properly with all pods alive

too much sensitive information, no need to provide for a clone of already fixed issue

h3poteto:
Fixes #14100 

`kube-controller-manager` and `kube-scheduler` version of https://github.com/kubernetes/kops/pull/13264 .

Disable these flags because these are not accepted when logging-format is `json`.
* --logtostderr
* --alsologtostderr
* --log-file

https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/

And I confirmed kube-proxy, but there is no option regarding logging-format in [kube-proxy](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/). Of course, kops does not have such an option.
https://github.com/kubernetes/kops/blob/6f4bb0e4daf52084407491fb875f8621c1927eca/pkg/apis/kops/componentconfig.go#L238-L278

### DIFF
--- a/nodeup/pkg/model/kube_controller_manager.go
+++ b/nodeup/pkg/model/kube_controller_manager.go
@@ -252,11 +252,17 @@ func (b *KubeControllerManagerBuilder) buildPod(kcm *kops.KubeControllerManagerC
 		container.Args = append(container.Args, sortedStrings(flags)...)
 	} else {
 		container.Command = []string{"/usr/local/bin/kube-controller-manager"}
-		container.Args = append(
-			sortedStrings(flags),
-			"--logtostderr=false", // https://github.com/kubernetes/klog/issues/60
-			"--alsologtostderr",
-			"--log-file=/var/log/kube-controller-manager.log")
+		if kcm.LogFormat != "" && kcm.LogFormat != "text" {
+			// When logging-format is not text, some flags are not accepted.
+			// https://github.com/kubernetes/kops/issues/14100
+			container.Args = sortedStrings(flags)
+		} else {
+			container.Args = append(
+				sortedStrings(flags),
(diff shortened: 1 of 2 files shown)


hakman (approved):

hakman:
Thanks @h3poteto!

h3poteto:
Should we cherry-pick this change for 1.22, 1.23, and 1.24?

PR #14125, opened 2022-08-14
Revert back to using kubectl in channels

olemarkus:
I think it will take a while to get the custom client production grade. Reverting to 1.24 behavior so that we get working tests and that this does not block any release.

Introduces an Applier interface and keeps the custom client based implementation.

### DIFF
--- a/channels/pkg/channels/addon.go
+++ b/channels/pkg/channels/addon.go
@@ -41,6 +41,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type Applier interface {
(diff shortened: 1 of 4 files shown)


justinsb:
I'd really like to try to get this working - I am going to see about OSSing the code we have or reproducing it today!

That said, I do think an interface makes sense here.

olemarkus:
What do you think about merging this and then do another PR for switching back. Or using tags for switching implementation if we want to (feature flag could make sense, but it would take a lot more to implement and propagate through nodeup)

justinsb:
Yes, I think that's a good plan @olemarkus  - thanks for suggesting it!

PR #14127, opened 2022-08-16
Write the user provided IG spec to state store instead of the full spec

olemarkus:
With clusters, we write the input spec to the cluster store. The full spec is also uploaded, which is eventually used by nodeup.

With IGs, however, we write the full spec (after expanding and setting all defaults) to the cluster store. Further defaults are then done in places less appropriate such as during nodeup or during the building of nodeup config. The reason for this is probably that nodeup used the IG spec earlier, while now this isn't needed since nodeup only uses the nodeup config.

This PR writes the input spec to the cluster store, but it does validate the expanded cluster spec (_before_ the not so appropriate expansion though!). This will allow us to move more expansions to populateInstanceGroupSpec without mutating the user input before storing it.

Some of the setup logic done in populateInstanceGroupSpec such as setting image and machine type has been moved to new_cluster.

Some of the mutation has been skipped as well (some node labels and the provider type), which is the cause of the create cluster golden output diffs. There should be no diffs to update cluster or nodeup configs though.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -40,6 +40,7 @@ import (
 
 	kopsbase "k8s.io/kops"
 	"k8s.io/kops/cmd/kops/util"
+	"k8s.io/kops/pkg/apis/kops"
 	api "k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/apis/kops/registry"
 	kopsutil "k8s.io/kops/pkg/apis/kops/util"
@@ -68,9 +69,6 @@ type CreateClusterOptions struct {
 	NodeVolumeSize       int32
 	ContainerRuntime     string
 	OutDir               string
-	Image                string
-	NodeImage            string
-	MasterImage          string
 	DisableSubnetTags    bool
 	NetworkCIDR          string
 	DNSZone              string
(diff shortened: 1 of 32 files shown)


cmd/kops/create_cluster.go
```diff
@@ -549,22 +547,6 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
...
-		for _, group := range masters {
-			group.Spec.Image = c.MasterImage
-		}
-	}
-	if c.NodeImage != "" {
-		for _, group := range nodes {
-			group.Spec.Image = c.NodeImage
-		}
-	}
-
```
hakman, line 567: Where was this logic moved?

olemarkus: E.g https://github.com/kubernetes/kops/pull/14127/files#diff-cf0e850db1e4461eab5d4dac860d20f2176b940f236e50825ae538ee23688a1eR861

So in `new_cluster.go`

hakman: I don't see any reference to `NodeImage` and I don't see `Image` being set first and `MasterImage` and `NodeImage` overriding it.

olemarkus: https://github.com/kubernetes/kops/pull/14127/files#diff-cf0e850db1e4461eab5d4dac860d20f2176b940f236e50825ae538ee23688a1eR396

It happens the other way around now. So if the image doesn't get set during IG creation, it gets set by the default logic.

But I think you are right about NodeImage not being respected. We may be missing a test case here. Let me try to add one.

hakman (approved):

PR #14137, opened 2022-08-17
Bump cilium to 1.11.8

olemarkus:
Also fixes an issue with rp_filters on some clouds/distros combinations. See https://github.com/cilium/cilium/issues/10645


hakman (approved):

olemarkus:
We have enough tests failing for us to see the if this has any effect.

/hold cancel

PR #14142, opened 2022-08-18
Bump Karpenter to 0.15 and enable consolidation

olemarkus:


pkg/apis/kops/validation/validation.go
```diff
@@ -301,6 +301,9 @@ func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *fie
 		if !featureflag.Karpenter.Enabled() {
 			allErrs = append(allErrs, field.Forbidden(fieldPath.Child("karpenter", "enabled"), "karpenter requires the Karpenter feature flag"))
 		}
+		if c.IsKubernetesLT("1.21") {
+			allErrs = append(allErrs, field.Forbidden(fieldPath.Child("karpenter", "enabled"), "karpenter requires Kubernetes 1.21 or greater"))
```
rifelpet, line 305: Did older versions of karpenter work with 1.20? wouldn't this break clusters when upgrading to this kops version?

olemarkus: It won't break clusters, but it'll block upgrades. But I figured ... featureflag let us do this rather than having to maintain multiple manifests.

rifelpet: Ah that's right, I forgot this is feature flagged

rifelpet (approved):

PR #14151, opened 2022-08-20
Plug the IAM role leak

olemarkus:
Looks like we were indeed leaking IAM roles. The code has assumed that all roles are prefixed with the cluster name, which is true for the instance profile roles, but rarely true for the SA roles. The SA roles have almost always truncated names.


olemarkus:
We were leaking instance profiles too, it seems.

hakman (approved):

PR #14154, opened 2022-08-21
Update v1.10.1.yaml

lvnilesh:
apiVersion: apps/v1


hakman:
Hi @lvnilesh. As mentioned in the Legacy Addons readme, they are deprecated, unmaintained and will be removed sometime in the future.
https://github.com/kubernetes/kops/blob/8bbc88eb2e0fdb2cc903892b7b16671b85b3b3df/addons/README.md?plain=1#L3

Issue #14174, opened 2022-08-24
"kops update cluster" panics while creating JWKS for OIDC


seh:
**1. What `kops` version are you running?**

_Client version: 1.24.1 (git-v1.24.1)_

**2. What Kubernetes version are you running?**

Starting with version **1.19.9**, upgrading to version **1.21.14**.

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

```shell
kops replace --filename=cluster.yaml
kops update cluster --yes
```

**5. What happened after the commands executed?**

It appears that _kops update cluster_ fails when it panics preparing for publishing OIDC Discovery documents to an S3 bucket:
```
W0824 08:36:09.215431   12986 external_access.go:39] KubernetesAPIAccess is empty
I0824 08:36:10.848893   12986 executor.go:111] Tasks: 0 done / 393 total; 110 can run
I0824 08:36:11.805410   12986 executor.go:111] Tasks: 110 done / 393 total; 83 can run
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x108 pc=0x3ec3a7c]

goroutine 1531 [running]:
k8s.io/kops/pkg/model.(*OIDCKeys).Open(0xc0017a32a0?)
        k8s.io/kops/pkg/model/issuerdiscovery.go:134 +0x21c
k8s.io/kops/upup/pkg/fi.CopyResource({0x5c5c140, 0xc00104de60}, {0x5c61d40?, 0xc00108c7c8?})
        k8s.io/kops/upup/pkg/fi/resources.go:85 +0x72
k8s.io/kops/upup/pkg/fi.ResourceAsBytes({0x5c61d40, 0xc00108c7c8})
        k8s.io/kops/upup/pkg/fi/resources.go:112 +0x4c
k8s.io/kops/upup/pkg/fi/fitasks.(*ManagedFile).Render(0x5?, 0x0?, 0xc00168b740?, 0xc000ee9240, 0x2?)
        k8s.io/kops/upup/pkg/fi/fitasks/managedfile.go:154 +0x70
reflect.Value.call({0x4f99d60?, 0xc000ee9240?, 0x4?}, {0x537191a, 0x4}, {0xc000c28c60, 0x4, 0x5c91ab0?})
        reflect/value.go:556 +0x845
[...]

PR #14175, opened 2022-08-24
OIDC: Tolerate extra service-account key set items

seh:
When reading the _kOps_ "service-account" key set in preparation for publishing the OIDC JWKS file (such as to S3 alongside the discovery document), in some cases the set contains items that either lack an X.509 certificate or contain such a certificate issued for a subject with common name other than "service-account." Ignore these extra key set items and instead only project JWKS keys for those with an X.509
certificate with the expected subject common name.

Refs: #14174

### DIFF
--- a/pkg/model/issuerdiscovery.go
+++ b/pkg/model/issuerdiscovery.go
@@ -130,6 +130,9 @@ func (o *OIDCKeys) Open() (io.Reader, error) {
 		if item.DistrustTimestamp != nil {
 			continue
 		}
+		if item.Certificate == nil || item.Certificate.Subject.CommonName != "service-account" {
+			continue
+		}
 
 		publicKey := item.Certificate.PublicKey
 		publicKeyDERBytes, err := x509.MarshalPKIXPublicKey(publicKey)


olemarkus (commented):
This test looks sane, but I wonder if you can still end up in issues elsewhere. Especially if certificate rotation works.

I would assume that `kops get keypairs` still fails as well, so this wouldn't be a full fix of #14174.

hakman:
This would at least unblock deployment and it can be iterated later.
Also, one could always distrust older keys.

olemarkus:
Yeah. I just wanted to unlink the issue so it doesn't get closed.

/approve

seh:
> I would assume that `kops get keypairs` still fails as well, so this wouldn't be a full fix of #14174.

Yes, that's true. I didn't realize that _kops get keypairs_ fails until after I had proposed this fix for the failure in _kops update cluster_, and I neglected to change the "Fixes" comment.

I expect that there are several more places that read key pairs that will need similar treatment.

seh:
Please see #14370 for making _kops get keypairs_ more tolerant as well.

PR #14185, opened 2022-08-25
Bump the nvidia addon

olemarkus:
driver: 515
device addon: 0.12.2

Also use ubuntu release specific nvidia container runtime package.

### DIFF
--- a/nodeup/pkg/model/nvidia.go
+++ b/nodeup/pkg/model/nvidia.go
@@ -17,6 +17,8 @@ limitations under the License.
 package model
 
 import (
+	"fmt"
+
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/nodeup/nodetasks"
 )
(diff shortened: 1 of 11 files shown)


hakman (approved):

ddelange:
Hi @olemarkus 👋 

Could you do another bump when you have a spare minute?

CUDA v12 is out, and it [requires](https://docs.nvidia.com/deploy/cuda-compatibility/index.html) `>=525.60.13`

Many thanks in advance!

ddelange:
on a side-note: we moved from `nvidia-headless-5xx-server` to the `nvidia-headless-5xx` package, because this will install the `nvidia-utils-5xx` subpackage, which contains the `nvidia-smi` binary.

we were surprised to see it was missing, because it is generally used to monitor gpu usage, similar to `top`, and it's also used by downstream packages in machine learning applications to get information about your gpus.

olemarkus:
Right now I have a few other things I need to prioritise, I'm afraid. But you can take the relevant commits from this PR and easily do a new one.

Regarding smi, it's more common to run this in a container rather than expecting it exists host-side.

ddelange:
> Right now I have a few other things I need to prioritise, I'm afraid. But you can take the relevant commits from this PR and easily do a new one.

ok no worries

> Regarding smi, it's more common to run this in a container rather than expecting it exists host-side.

I was in opposite understanding, e.g. `nvidia/cuda` runtime images do not contain `nvidia-smi` 🤔

olemarkus:
Been a while while since I tested, but the cuda base flavour at least did have nvidia-smi.

olemarkus:
By the way, recommend filing a new issue for bumping nvidia.

PR #14194, opened 2022-08-26
Exit rolling updates when encountering specific errors

### LINKED ISSUE #14176: All instancegroups must fail before kops exits on rolling update
/kind bug

 this information.**

1.20.3

 a `kops` flag.**

1.20.15

AWS

```
$ kops rolling-update cluster my-cluster.io --state s3://plaid-kops --master-interval 5m0s --node-interval 5m0s --post-drain-delay 5s --validation-timeout 15m0s --yes
```

The rolling update went appropriately! But, we use an AWS IAM role in order to control permissions to run kops commands. Some of the clusters are rather large, so due to https://github.com/kubernetes/kops/issues/5130 we end up waiting for kops to fail with a specific error, refresh the role and then re-run the command.

When this happens we see the following logs:

<details>

```
[2022-08-24T16:07:58.971Z] E0824 16:07:53.703940     321 instancegroups.go:457] Cluster did not validate within 15m0s

[2022-08-24T16:07:58.971Z] I0824 16:07:53.703951     321 instancegroups.go:452] Validating the cluster.

[2022-08-24T16:13:36.722Z] I0824 16:13:22.259215     321 instancegroups.go:498] Cluster did not validate, will retry in "30s": unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T16:13:36.722Z] 	status code: 403, request id: 72e2c489-ec7c-4e80-98dc-cda58e3804ba.

[2022-08-24T16:19:00.195Z] I0824 16:18:57.924309     321 instancegroups.go:498] Cluster did not validate, will retry in "30s": unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T16:19:00.195Z] 	status code: 403, request id: 62df166c-b6d2-47d0-9967-db488cbe5b76.

[2022-08-24T16:24:54.070Z] I0824 16:24:53.177895     321 instancegroups.go:495] Cluster did not validate within deadline: unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T16:24:54.070Z] 	status code: 403, request id: abef1974-1752-42e8-82c0-c5cfe6a8b1d2.

[2022-08-24T16:24:54.070Z] E0824 16:24:53.177928     321 instancegroups.go:457] Cluster did not validate within 15m0s

[2022-08-24T16:24:54.070Z] I0824 16:24:53.177945     321 instancegroups.go:452] Validating the cluster.

[2022-08-24T16:32:17.885Z] I0824 16:32:03.178143     321 instancegroups.go:498] Cluster did not validate, will retry in "30s": unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T16:32:17.885Z] 	status code: 403, request id: ab436289-abec-4c19-8a56-1f68daf6ffe3.

[2022-08-24T16:37:41.710Z] I0824 16:37:40.916761     321 instancegroups.go:498] Cluster did not validate, will retry in "30s": unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T16:37:41.710Z] 	status code: 403, request id: 4068577b-0255-4bf8-a3fc-d6289d2db07e.

[2022-08-24T16:44:04.955Z] I0824 16:43:57.587224     321 instancegroups.go:495] Cluster did not validate within deadline: unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T16:44:04.955Z] 	status code: 403, request id: 6d00c28e-f560-4923-985e-c89e0ee45cbc.

[2022-08-24T16:44:04.955Z] E0824 16:43:57.587257     321 instancegroups.go:457] Cluster did not validate within 15m0s

[2022-08-24T16:44:04.955Z] I0824 16:43:57.587273     321 instancegroups.go:452] Validating the cluster.

[2022-08-24T16:49:30.646Z] I0824 16:49:17.122051     321 instancegroups.go:498] Cluster did not validate, will retry in "30s": unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T16:49:30.646Z] 	status code: 403, request id: 71f5d700-f207-4a77-a16d-5d66d4639967.

[2022-08-24T16:55:39.198Z] I0824 16:55:28.310586     321 instancegroups.go:498] Cluster did not validate, will retry in "30s": unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T16:55:39.198Z] 	status code: 403, request id: d47edcf4-60a9-4058-b003-93e75d567763.

[2022-08-24T17:02:48.991Z] I0824 17:02:40.994726     321 instancegroups.go:495] Cluster did not validate within deadline: unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T17:02:48.991Z] 	status code: 403, request id: 66c94c79-b5df-4fa6-9487-b36c722b3e64.

[2022-08-24T17:02:48.991Z] E0824 17:02:40.994764     321 instancegroups.go:457] Cluster did not validate within 15m0s

[2022-08-24T17:02:48.991Z] I0824 17:02:40.994780     321 instancegroups.go:452] Validating the cluster.

[2022-08-24T17:08:42.381Z] I0824 17:08:29.814311     321 instancegroups.go:498] Cluster did not validate, will retry in "30s": unable to find autoscale groups: error listing autoscaling cluster tags: ExpiredToken: The security token included in the request is expired

[2022-08-24T17:08:42.381Z] 	status code: 403, request id: 53a4ac96-10fc-4a45-969d-df6e77149d39.
```
</details>

What is happening is that each instance group (on some clusters we have quite a few of those) needs to fail before the validation command can completely exit: https://github.com/kubernetes/kops/blob/v1.20.3/pkg/instancegroups/rollingupdate.go#L175-L181

So (assuming we roll an entire cluster at once), if our validation timeout is 15 minutes, and we have X instancegroups, we can expect to take 15 minutes * X amount of time to fail the command. At a minimum, with 3 masters and 1 instancegroup for nodes this ends up being an hour. Some of our clusters have 5 masters and 6 instancegroups, so we end up waiting close to 3 hours for a failure.

I expected that kops would fail out early when an instancegroup could not be rolled. That way, the rolling update can only fail for as long as the validation timeout (a maximum of 15 minutes instead of hours).

Happy to provide, but I don't think this is actually relevant.

Please see the logs above!

Happy to PR this and have the change hidden behind a flag so that existing behavior is not changed.

jandersen-plaid:
Fixes #14176 

Add a new error to `instancegroups` which is returned when cluster validation times out. Then, check this error during cluster node and apiserver instancegroup rolls. If the cluster fails to be validated within the timeout then it is unlikely to succeed on subsequent node rolls, so the preference is to exit the cluster roll (to allow operators to fix the cluster validation issue).

### DIFF
--- a/pkg/instancegroups/instancegroups.go
+++ b/pkg/instancegroups/instancegroups.go
@@ -46,12 +46,16 @@ const rollingUpdateTaintKey = "kops.k8s.io/scheduled-for-update"
 // ValidationTimeoutError represents an error that occurs when
 // the cluster fails to validate within the designated timeout.
 type ValidationTimeoutError struct {
-	Operation string
-	Err       error
+	operation string
+	err       error
 }
 
 func (v *ValidationTimeoutError) Error() string {
-	return fmt.Sprintf("error validating cluster%s: %v", v.Operation, v.Err)
+	return fmt.Sprintf("error validating cluster%s: %s", v.operation, v.err.Error())
+}
+
+func (v *ValidationTimeoutError) Unwrap() error {
+	return v.err
 }
 
 // Is checks that a given error is a ValidationTimeoutError.
@@ -501,8 +505,8 @@ func (c *RollingUpdateCluster) maybeValidate(operation string, validateCount int
 			if c.FailOnValidate {
 				klog.Errorf("Cluster did not validate within %s", c.ValidationTimeout)
 				return &ValidationTimeoutError{
-					Operation: operation,
-					Err:       err,
+					operation: operation,
+					err:       err,
 				}
 			}
 
--- a/pkg/instancegroups/instancegroups.go
+++ b/pkg/instancegroups/instancegroups.go
@@ -54,6 +54,15 @@ func (v *ValidationTimeoutError) Error() string {
 	return fmt.Sprintf("error validating cluster%s: %v", v.Operation, v.Err)
 }
 
+// Is checks that a given error is a ValidationTimeoutError.
+func (v *ValidationTimeoutError) Is(err error) bool {
+	// Currently all validation timeout errors are equivalent
+	// If you wish to differentiate, please update the instances of `errors.Is` that check
+	// this error to take that into account
+	_, ok := err.(*ValidationTimeoutError)
+	return ok
+}
+
 // promptInteractive asks the user to continue, mostly copied from vendor/google.golang.org/api/examples/gmail.go.
 func promptInteractive(upgradedHostID, upgradedHostName string) (stopPrompting bool, err error) {
 	stopPrompting = false
--- a/pkg/instancegroups/instancegroups.go
+++ b/pkg/instancegroups/instancegroups.go
@@ -43,6 +43,17 @@ import (
 
 const rollingUpdateTaintKey = "kops.k8s.io/scheduled-for-update"
 
+// ValidationTimeoutError represents an error that occurs when
+// the cluster fails to validate within the designated timeout.
+type ValidationTimeoutError struct {
+	Operation string
+	Err       error
+}
+
+func (v *ValidationTimeoutError) Error() string {
+	return fmt.Sprintf("error validating cluster%s: %v", v.Operation, v.Err)
+}
+
 // promptInteractive asks the user to continue, mostly copied from vendor/google.golang.org/api/examples/gmail.go.
 func promptInteractive(upgradedHostID, upgradedHostName string) (stopPrompting bool, err error) {
 	stopPrompting = false
@@ -480,7 +491,10 @@ func (c *RollingUpdateCluster) maybeValidate(operation string, validateCount int
 
 			if c.FailOnValidate {
 				klog.Errorf("Cluster did not validate within %s", c.ValidationTimeout)
-				return fmt.Errorf("error validating cluster%s: %v", operation, err)
+				return &ValidationTimeoutError{
+					Operation: operation,
+					Err:       err,
+				}
 			}
 
 			klog.Warningf("Cluster validation failed%s, proceeding since fail-on-validate is set to false: %v", operation, err)
--- a/pkg/instancegroups/rollingupdate.go
+++ b/pkg/instancegroups/rollingupdate.go
@@ -18,6 +18,7 @@ package instancegroups
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"sort"
 	"sync"
@@ -187,7 +188,7 @@ func (c *RollingUpdateCluster) RollingUpdate(groups map[string]*cloudinstances.C
 
 		for _, k := range sortGroups(apiServerGroups) {
 			err := c.rollingUpdateInstanceGroup(apiServerGroups[k], c.NodeInterval)
-			if err != nil && exitableError(err) {
+			if err != nil && isExitableError(err) {
 				return err
 			}
 
@@ -209,7 +210,7 @@ func (c *RollingUpdateCluster) RollingUpdate(groups map[string]*cloudinstances.C
 
 		for _, k := range sortGroups(nodeGroups) {
 			err := c.rollingUpdateInstanceGroup(nodeGroups[k], c.NodeInterval)
-			if err != nil && exitableError(err) {
+			if err != nil && isExitableError(err) {
 				return err
 			}
 
@@ -237,13 +238,12 @@ func sortGroups(groupMap map[string]*cloudinstances.CloudInstanceGroup) []string
 	return groups
 }
 
-// exitableError inspects an error to determine if the error is
+// isExitableError inspects an error to determine if the error is
 // fatal enough that the rolling update cannot continue.
 //
 // For example, if a cluster is unable to be validated by the deadline, then it
 // is unlikely that it will validate on the next instance roll, so an early exit as a
 // warning to the user is more appropriate.
 func isExitableError(err error) bool {
-	_, ok := err.(*ValidationTimeoutError)
-	return ok
+	return stderrors.Is(err, &ValidationTimeoutError{})
 }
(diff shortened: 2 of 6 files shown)


olemarkus:
I'd rather not add a flag for this. I think it is enough to inspect the returned error and return directly if it does not make sense to continue.

johngmyers:
I think I'd prefer this be the default or only option.

/hold
/kind office-hours

johngmyers:
I believe the history was that the update on the IG would previously wait forever, not fail.

pkg/instancegroups/rollingupdate.go
```diff
@@ -236,3 +237,13 @@ func sortGroups(groupMap map[string]*cloudinstances.CloudInstanceGroup) []string
...
 }
+
+// exitableError inspects an error to determine if the error is
+// fatal enough that the rolling update cannot continue.
+//
+// For example, if a cluster is unable to be validated by the deadline, then it
+// is unlikely that it will validate on the next instance roll, so an early exit as a
+// warning to the user is more appropriate.
+func exitableError(err error) bool {
+	return strings.HasPrefix(err.Error(), "error validating cluster")
```
johngmyers, line 248: This should be some sort of type assertion, not a check of the error string.

olemarkus:
If a control plane IG fails, we already directly return an error. That is by far the most important behavior. If an IG fails and kOps keeps going to the next, and keeps going to the next and continues to gracefully drain and terminate nodes makes sense.

But in the case of a validation error, it doesn't make sense to keep going as kOps won't succeed with the next IG either.

pkg/instancegroups/rollingupdate.go
jandersen-plaid: Sounds good! I changed the returned error that this is checking in https://github.com/kubernetes/kops/pull/14194/commits/b8e503136a88582118559c36ce4023a9de0ce1e9 to be a specific error type instead of `fmt.Errorf` so I could do the error type checking here. Additionally, I figured it was better to return this from `maybeValidate` rather than `validateWithClusterTimeout` since this is the highest level of the chain validation call stack. I can also see an argument for the error being more conceptually in sync with `validateWithClusterTimeout` as well, so happy to change it with whatever fits.

pkg/instancegroups/rollingupdate.go
```diff
@@ -187,11 +187,11 @@ func (c *RollingUpdateCluster) RollingUpdate(groups map[string]*cloudinstances.C
 
 		for _, k := range sortGroups(apiServerGroups) {
 			err := c.rollingUpdateInstanceGroup(apiServerGroups[k], c.NodeInterval)
-			results[k] = err
-			if err != nil {
-				klog.Errorf("failed to roll InstanceGroup %q: %v", k, err)
```
olemarkus, line 192: If this isn't an exitable error, this should still log and continue.

pkg/instancegroups/rollingupdate.go
```diff
@@ -236,3 +236,14 @@ func sortGroups(groupMap map[string]*cloudinstances.CloudInstanceGroup) []string
...
 	return groups
 }
+
+// exitableError inspects an error to determine if the error is
+// fatal enough that the rolling update cannot continue.
+//
+// For example, if a cluster is unable to be validated by the deadline, then it
+// is unlikely that it will validate on the next instance roll, so an early exit as a
+// warning to the user is more appropriate.
+func exitableError(err error) bool {
```
olemarkus, line 246:
```suggestion
func isExitableError(err error) bool {
```

pkg/instancegroups/rollingupdate.go
```diff
@@ -236,3 +236,14 @@ func sortGroups(groupMap map[string]*cloudinstances.CloudInstanceGroup) []string
...
+	_, ok := err.(*ValidationTimeoutError)
```
olemarkus, line 247: Use `errors.Is(err, ValidationTimeoutError)` in case we wrap this error at some point.

olemarkus (changes requested):

pkg/instancegroups/rollingupdate.go
jandersen-plaid: See ac86445 for the implementation of this.

Since the struct takes in arbitrary strings, comparing out of the box via `errors.Is` would always return `false`. I implemented the `Is` interface with this typecheck, so -- if the `ValidationTimeoutError`s need to be differentiated from each other in the future -- it can be done there.

pkg/instancegroups/instancegroups.go
```diff
@@ -43,6 +43,26 @@ import (
 
 const rollingUpdateTaintKey = "kops.k8s.io/scheduled-for-update"
 
+// ValidationTimeoutError represents an error that occurs when
+// the cluster fails to validate within the designated timeout.
+type ValidationTimeoutError struct {
+	Operation string
+	Err       error
```
olemarkus, line 50: There should be an `Unwrap` method on this struct as well that returns this value. I also think neither `err` nor `operation` should be exported.

```
func (v *ValidationTimeoutError) Unwrap() error { 
  return v.Err 
}
```

pkg/instancegroups/instancegroups.go
```diff
@@ -43,6 +43,26 @@ import (
...
+}
+
+func (v *ValidationTimeoutError) Error() string {
+	return fmt.Sprintf("error validating cluster%s: %v", v.Operation, v.Err)
```
olemarkus, line 54:
```suggestion
	return fmt.Sprintf("error validating cluster%s: %s", v.Operation, v.Err.Error())
```

olemarkus (changes requested):
This looks good.

I think it is worth adding a brief explanation to the notes about the change in behavior. Something like:
```
* As of 1.24, kOps no longer hung indefinitely on eviction errors, but timed out after 15 minutes by default. After the timeout kOps will carry on to the next InstanceGroup. As of kOps 1.26, kOps will no longer carry on if a cluster validation error is encountered.
```

olemarkus (approved):

olemarkus:
@johngmyers you still want to hold this one?

PR #14213, opened 2022-08-31
Accommodate S3 VFS paths designating bucket owner

seh:
In VFS paths using the "s3" scheme designating paths within S3 buckets, accept the "x-amz-expected-bucket-owner" query parameter nominating [an AWS account ID that we expect will own the S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-owner-condition.html). When this query parameter is present in a parsed VFS path, include its value in AWS Go SDK calls that accept an "ExpectedBucketOwner" parameter. (If there are multiple occurrences of this parameter in the query string, use the first occurrence's value.) If the target S3 bucket turns out to be owned by an AWS account with a different ID, these API calls will fail.

Propagate this expected owning account ID when constructing new VFS paths from a path with such an ID already bound.

### DIFF
--- a/util/pkg/vfs/context.go
+++ b/util/pkg/vfs/context.go
@@ -299,7 +299,13 @@ func (c *VFSContext) buildS3Path(p string) (*S3Path, error) {
 		return nil, fmt.Errorf("invalid s3 path: %q", p)
 	}
 
-	s3path := newS3Path(c.s3Context, u.Scheme, bucket, u.Path, true)
+	var expectedBucketOwner string
+	const ownerParam = "x-amz-expected-bucket-owner"
+	if values := u.Query(); values.Has(ownerParam) {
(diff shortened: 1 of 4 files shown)


hakman:
@seh I understand that @olemarkus provided a different approach to this change. Do you think it's still needed?

seh:
> Do you think this is still needed?

Testing showed that it was never really _necessary_, but my investigation led me to discover this feature of the S3 API that I had not known about, and I thought it might be useful to some of our more security-conscious users. If you don't think it's worth adding to _kOps_, or would prefer a different approach, I understand. I didn't have high confidence that you'd accept this patch. I did think that it was sharing the idea, though.

seh:
This was an interesting idea, but since we don't really need it now, I'll withdraw this proposal.

PR #14233, opened 2022-09-05
GCE: change default control-plane instance type to e2-medium

justinsb:
This better matches the AWS machine, which is also a burstable 2 core
machine.  Without this pods sometimes fail to schedule on single core
machines.


olemarkus:
Can you run `hack/update-expected.sh`?

hakman (approved):

Issue #14171, opened 2022-08-24
ClusterAutoscaler addon: Add support to cordonNodeBeforeTerminating flag


dcfranca:
`Cluster Autoscaler` since at least version 1.19.3 has added support to a flag to indicate whether it should cordon a node before scaling down or not.

This can be seen on the list of parameters [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca)

> cordon-node-before-terminating | Should CA cordon nodes before terminating during downscale process | false

This flag is not supported by kops, is that possible for it to be added?

 I guess it needed to be added at some parts of the code
 
 https://github.com/kubernetes/kops/blob/master/k8s/crds/kops.k8s.io_clusters.yaml#L716
 
 https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/componentconfig.go#L1000 and `v1apha2/3` as well?
 
https://github.com/kubernetes/kops/blob/master/pkg/model/components/clusterautoscaler.go#L89

And here add a conditional to include the flag in case it was set: https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template#L332

In case it is planned to add it I can submit a PR, just probably need some guidance on the codebase and tests

olemarkus:
We tend to accept features such as these. Happy to review a PR.

All the API versions you mention above must be updated.

PR #14236, opened 2022-09-05
Add suport to --cordon-node-before-terminating autoscaler flag

dcfranca:
This feature was added on k8s autoscaler on Oct 2020: https://github.com/kubernetes/autoscaler/pull/3649
but kops didn't provide support to add it via the autoscaler addon
This PR adds it

Implements issue: https://github.com/kubernetes/kops/issues/14171

### DIFF
--- a/upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
+++ b/upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
@@ -329,6 +329,10 @@ spec:
             - --scale-down-delay-after-add={{ .ScaleDownDelayAfterAdd }}
             - --new-pod-scale-up-delay={{ .NewPodScaleUpDelay }}
             - --max-node-provision-time={{ .MaxNodeProvisionTime }}
+            # This flag does not exist before CAS 1.21
+            {{ if IsKubernetesGTE "1.21" }}
+            - --cordon-node-before-terminating="{{ WithDefaultBool .CordonNodeBeforeTerminating true }}"
+            {{ end }}
             - --logtostderr=true
             - --stderrthreshold=info
             - --v=4
--- a/docs/releases/1.25-NOTES.md
+++ b/docs/releases/1.25-NOTES.md
@@ -11,6 +11,7 @@ This is a document to gather the release notes prior to the release.
 * Karpenter support can be considered stable on Kubernetes versions 1.22 up until 1.24. Karpenter does not yet support Kubernetes above 1.25.
 
 * Cert Manager may now solve dns-01 challenges. See [the cert manager documentation](/addons/#enabling-dns-01-challenges).
+* Add support to --cordon-node-before-terminating on the cluster autoscaler addon (CordonNodeBeforeTerminating)
 
 * EBS CSI driver can now be self-managed. See the [addon docs](/addons/#self-managed-aws-ebs-csi-driver).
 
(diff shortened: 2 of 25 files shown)


upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -329,6 +329,9 @@ spec:
             - --scale-down-delay-after-add={{ .ScaleDownDelayAfterAdd }}
             - --new-pod-scale-up-delay={{ .NewPodScaleUpDelay }}
             - --max-node-provision-time={{ .MaxNodeProvisionTime }}
+            {{ if .CordonNodeBeforeTerminating }}
+            - --cordon-node-before-terminating=true
+            {{ end }}
```
olemarkus, line 334:
```suggestion
            # This flag does not exist before CAS 1.21
            {{ IsKubernetesGTE 1.21 }}
             - --cordon-node-before-terminating="{{ WithDefaultBool .CordonNodeBeforeTerminating false }}"
            {{ end }}
```

Although I am wondering if not we could default to true. Seems like the right thing to do.

olemarkus (changes requested):

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
dcfranca: This was introduced on some version of the auto scaler (I don't know exactly, but it was 1.19.x), wouldn't it break compatibility?

olemarkus: 1.21 is the earliest version. Hence the GTE 1.21 check.

olemarkus:
Also, please add a line to release notes.

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -329,6 +329,10 @@ spec:
             - --scale-down-delay-after-add={{ .ScaleDownDelayAfterAdd }}
             - --new-pod-scale-up-delay={{ .NewPodScaleUpDelay }}
             - --max-node-provision-time={{ .MaxNodeProvisionTime }}
+            # This flag does not exist before CAS 1.21
+            {{ if IsKubernetesGTE "1.21" }}
+            - --cordon-node-before-terminating="{{ WithDefaultBool .CordonNodeBeforeTerminating false }}"
```
olemarkus, line 334: @hakman @johngmyers  what do you think about having this enabled by default? When kOps does rolling updates, we also cordon before terminating, so enabling it seems reasonable to me.

hakman: I was wondering the same thing, so 👍

dcfranca: Done!

olemarkus (approved):

olemarkus:
This one probably need a quick rebase and update expected.

olemarkus:
Might actually be `make crds apimachinery` that's missing.

johngmyers:
/lgtm

Though I suggest it might not be necessary to allow turning this off.

rifelpet:
This is causing test failures

https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-kstable-kolatest-to-klatest-kolatest-many-addons/1568601968639741952/artifacts/cluster-info/kube-system/cluster-autoscaler-87b7c5845-9qqqs/logs.txt

```
invalid argument "\"true\"" for "--cordon-node-before-terminating" flag: parse error
```

PR #14286, opened 2022-09-17
 networkID support for Hetzner 

JJotah:


hakman:
@JJotah thanks for trying to add networkID support. It was a little more complicated than this and I implemented it as part of https://github.com/kubernetes/kops/pull/14294. You could give it a try using [kOps 1.25.0](https://github.com/kubernetes/kops/releases/tag/v1.25.0).

PR #14295, opened 2022-09-18
Replace --vpc flag with --network-id

hakman:
/cc @olemarkus @justinsb


justinsb:
Thanks @hakman  - we discussed in office hours and because it is non-breaking we think it's a good change!

/approve
/lgtm

PR #14299, opened 2022-09-19
hetzner: Move out of alpha and drop feature flag

hakman:
/cc @olemarkus @justinsb


README.md
```diff
@@ -15,7 +15,7 @@ We like to think of it as `kubectl` for clusters.
 `kops` will not only help you create, destroy, upgrade and maintain production-grade, highly
 available, Kubernetes cluster, but it will also provision the necessary cloud infrastructure.
 
-AWS (Amazon Web Services) is currently officially supported, with DigitalOcean, GCE, and OpenStack in beta support, and Azure and AliCloud in alpha.
+AWS (Amazon Web Services) and GCE (Google Compute Platform) are currently officially supported, with DigitalOcean, Hetzner and OpenStack in beta support, and Azure in alpha.
```
olemarkus, line 18: `README.md` and `docs/README.md` unfortunately is not the same file. So you probably want to update that file as well with similar content.

hakman: Right, thanks for noticing. Done.

olemarkus (approved):

PR #14306, opened 2022-09-19
Add release 1.25.0 to channels

hakman:
/cc @olemarkus

### DIFF
--- a/channels/stable
+++ b/channels/stable
@@ -87,6 +87,9 @@ spec:
     networking:
       kubenet: {}
   kubernetesVersions:
+  - range: ">=1.25.0"
+    recommendedVersion: 1.25.1
+    requiredVersion: 1.25.0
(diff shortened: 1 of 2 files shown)


channels/stable
```diff
@@ -133,6 +136,10 @@ spec:
     recommendedVersion: 1.11.10
     requiredVersion: 1.11.10
   kopsVersions:
+  - range: ">=1.25.0-alpha.1"
+    recommendedVersion: "1.25.0"
```
olemarkus, line 140: Should this be changed all the way back to K8s 1.20?

hakman: No, we do that once .1 is released.

olemarkus: Ah. TIL ...

olemarkus (approved):

Issue #14308, opened 2022-09-20
Hetzner instances are not joining the cluster


casperakos:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Client version: 1.25.0

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

v1.25.1

**3. What cloud provider are you using?**

Hetzner

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

kops create cluster --name=my-cluster.example.k8s.local \
  --ssh-public-key=~/.ssh/id_rsa.pub --cloud=hetzner --zones=nbg1 \
  --image=ubuntu-20.04 --networking=cilium --network-cidr=10.10.0.0/16

**5. What happened after the commands executed?**

Cluster has been created but 2 out five nodes are not joining the cluster. Cloud controller manager has this error : 
`error running controllers: failed to parse cidr value:"" with error:invalid CIDR address:
`
**6. What did you expect to happen?**

All nodes join the cluster

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2022-09-20T17:05:09Z"
  name: my-cluster.example.k8s.local
spec:
  api:
    loadBalancer:
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudProvider: hetzner
  configBase: s3://xxx/my-cluster.example.k8s.local
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-nbg1
      name: etcd-1
    memoryRequest: 100Mi
    name: main
  - cpuRequest: 100m
    etcdMembers:
    - instanceGroup: master-nbg1
      name: etcd-1
    memoryRequest: 100Mi
    name: events
  iam:
    allowContainerRegistry: true
    legacy: false
  kubeProxy:
    enabled: false
  kubelet:
    anonymousAuth: false
  kubernetesApiAccess:
  - 0.0.0.0/0
  - ::/0
  kubernetesVersion: 1.25.1
  masterPublicName: api.my-cluster.example.k8s.local
  networkCIDR: 10.10.0.0/16
  networking:
    cilium:
      enableNodePort: true
[...]

PR #14309, opened 2022-09-21
Hetzner: Generate CCM args from external CCM config

hakman:
Fixes: #14308


upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -450,7 +450,9 @@ func (tf *TemplateFunctions) CloudControllerConfigArgv() ([]string, error) {
 		argv = append(argv, fmt.Sprintf("--use-service-account-credentials=%t", true))
 	}
 
-	argv = append(argv, "--cloud-config=/etc/kubernetes/cloud.config")
+	if cluster.Spec.GetCloudProvider() != kops.CloudProviderHetzner {
+		argv = append(argv, "--cloud-config=/etc/kubernetes/cloud.config")
```
olemarkus, line 454: In most cases, this file is entirely empty. Hetzner CCM does not support this?

hakman: It supports it, just not mounted as part of the official manifests and didn't want to diverge too much.

olemarkus (approved):

upup/pkg/fi/cloudup/template_functions.go
olemarkus: Maybe consider a PR to upstream

Issue #13320, opened 2022-03-03
AWS LaunchTemplates with no userdata rebuild on each kops update cluster


kuldazbraslav:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Version 1.21.4 (git-53e6bf3e5b0a77c78df2a7c60baca6e926fd5105)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

v1.19.16

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

This spec is in S3:
```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: InstanceGroup
metadata:
  creationTimestamp: "2022-02-21T15:24:49Z"
  labels:
    kops.k8s.io/cluster: <REDACTED>
  name: bastions
spec:
  associatePublicIp: false
  cloudLabels:
    Owner: Core
    Service: Bastion
    env: dev
  image: <REDACTED>
  machineType: t3.micro
  maxSize: 2
  minSize: 1
  nodeLabels:
    ami/id: <REDACTED>
    ami/name: <REDACTED>
    ami/version: <REDACTED>
    environment: development
  role: Bastion
  rootVolumeOptimization: false
  rootVolumeSize: 10
  rootVolumeType: gp3
  subnets:
  - eu-west-1a
  - eu-west-1b
  - eu-west-1c
```

Then I run `kops --state s3://<REDACTED> update cluster <REDACTED>`

**5. What happened after the commands executed?**

Bastion instance group is reported with changes, even though in both cases there are no userdata either before or after.
```
Will modify resources:
  LaunchTemplate/bastions.<REDACTED>
  	UserData            	 <nil> -> <resource>
```

If I run `kops update cluster --yes`, new launch template is generated and bastion IG requires rolling update.

**6. What did you expect to happen?**

No changes reported, no rolling update required.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: null
  generation: 26
  name: <REDACTED>
spec:
  api:
    loadBalancer:
      class: Classic
      idleTimeoutSeconds: 300
      type: Public
  authorization:
[...]

Issue #14195, opened 2022-08-26
Instance groups NeedsUpdate after running rolling-update


yuzhouliu9:
/kind bug

**1. `kops` version.**
v1.24.0

**2. Kubernetes version.**
v1.21.14

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
After running `kops rolling-update`, my bastion and the first master that was rolled are still reporting "NeedsUpdate".

This has happened 3 times now, the bastion and the first master to roll are always marked as "NeedsUpdate" immediately after rollout. If I do the rolling-update command again, the master becomes Ready, but bastion stays at "NeedsUpdate".

```
NAME                STATUS      NEEDUPDATE  READY  MIN  TARGET  MAX  NODES
bastion             NeedsUpdate	1           0      1    1       1    0
master-us-east-1b   NeedsUpdate	1           0      1    1       1    1
master-us-east-1c   Ready       0           1      1    1       1    1
master-us-east-1f   Ready       0           1      1    1       1    1
```

I noticed the master-us-east-1b node had the annotation `kops.k8s.io/needs-update` which is used by kops as a [criteria](https://kops.sigs.k8s.io/operations/rolling-update/#instance-selection) for updating.

Why did kops leave this annotation even after rolling the node the first time?
Is there a way to find out why the bastion is marked as NeedsUpdate?

**5. What happened after the commands executed?**
Instance groups still marked as `NeedsUpdate`.

**6. What did you expect to happen?**
All Instance groups should report `Ready` since rolling-update just ran.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

[kops_cluster_spec.txt](https://github.com/kubernetes/kops/files/9434319/kops_cluster_spec.txt)

**8. Please run the commands with most verbose logging by adding the `-v 10` flag.
  Paste the logs into this report, or in a gist and provide the gist link here.**
I don't have this.

**9. Anything else do we need to know?**

olemarkus:
There only way bastion can need update is if the lASG aunch template version is higher than the launch template used to launch the bastion instance.

`kops.k8s.io/needs-update` are applied when updating certain addons. If you roll before addons are applied, you may need to roll one extra time.

seh:
[...]

PR #14318, opened 2022-09-23
Avoid spurious changes with bastion hosts due to user data

hakman:
Fixes: #13320 #14195

### DIFF
--- a/pkg/model/bootstrapscript.go
+++ b/pkg/model/bootstrapscript.go
@@ -245,7 +245,7 @@ func (b *BootstrapScriptBuilder) ResourceNodeUp(c *fi.ModelBuilderContext, ig *k
 
 		// Bastions can have AdditionalUserData, but if there isn't any skip this part
 		if len(ig.Spec.AdditionalUserData) == 0 {
-			return fi.NewStringResource(""), nil
+			return nil, nil
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/model/bootstrapscript.go
```diff
@@ -245,7 +245,7 @@ func (b *BootstrapScriptBuilder) ResourceNodeUp(c *fi.ModelBuilderContext, ig *k
 
 		// Bastions can have AdditionalUserData, but if there isn't any skip this part
 		if len(ig.Spec.AdditionalUserData) == 0 {
-			return fi.NewStringResource(""), nil
+			return nil, nil
```
seh, line 248: I see that this avoids returning a resource, and that we wind up setting the "UserData" field of the `awstasks.LaunchTemplate` to nil. How does that change avoid creation of a new launch template. Is it that two different `*fi.StringResource` values don't compare as equal, so _kOps_ thinks that there's a material change required in the launch template?

hakman: AWS seems to return `nil` when `UserData == ""`. This means that kOps tries to reconcile it.

seh:
Testing this patch atop the "v1.24.3" tag, I found that it resolves the problem, such that _kops update cluster_ no longer tries to create a new launch template version for the bastion ASG on every invocation.

olemarkus (approved):

PR #14332, opened 2022-09-25
Fix typo in doc for enabling cert-manager dns-01 challenges

ilyasotkov:
Following the doc resulted in an error from kOps CLI

```
Error: error creating cluster: spec.certManager: Forbidden: Cert Manager requires that service accounts use external permissions in order to do dns-01 validation
```

because there is a typo in the doc

### DIFF
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -158,7 +158,7 @@ spec:
     hostedZoneIDs:
     - ZONEID
   iam:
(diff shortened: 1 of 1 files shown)


hakman (approved):
Thanks for taking the time to fix this @ilyasotkov.
/ok-to-test

Issue #14334, opened 2022-09-25
Terraform Support for Version >=1.3 


ecda909:
/kind feature

1. I would like to know if Terraform versions 1.3 or higher are supported using —terraform. If it’s not supported. Then I would like to add this as a feature request.

hakman:
@ecda909 We are running regularly tests that validate kOps Terraform output against newer versions of Terraform:
https://testgrid.k8s.io/kops-presubmits#verify-terraform

I updated the test this morning to run against Terraform 1.3.0 in #14335 and I don't see any error. I suggest you give it a try and, if you find specific issues, please report them.

ecda909:
Okay, I will. Thank you very much!

ecda909:
Okay, thank you very much!

PR #14335, opened 2022-09-26
Bump Terraform tag to v1.3.0

hakman:
Fixes: #14334


olemarkus:
Terraform is not my thing.
/cc @rifelpet 
Just in case there are any caveats we are not aware of.

hakman:
@rifelpet Any thoughts about this?

rifelpet (approved):

PR #14322, opened 2022-09-23
Scaleway init and nodeup

Mia-Cross:
The goal of this PR is to initialize Scaleway support in kOps :

- by allowing the feature flag "Scaleway"
- by adding the right environment variables to the bootstrap script (therefore they will also be passed to nodeup)

PR #14345, opened 2022-09-27
Scaleway support in protokube

Mia-Cross:
The goal of this PR is to add support for Scaleway in protokube. This step will complete the bootstrapping of new nodes. For the time being, only gossip DNS is supported but a custom DNS will soon come in a next PR.


Mia-Cross:
@hakman I tried adding ScalewaySpec to the ClusterSpec in v1alpha2 as you suggested in my previous PR (#14322), but I was not sure where since I didn't see any cloud provider spec. It caused errors in the CI so I eventually removed it. It seems like the [CloudProviderSpec](https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/v1alpha2/cluster.go#L58) included in v1alpha2 references [this one](https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/cluster.go#L231) where ScalewaySpec is already present.

remyleone (approved):

hakman (changes requested):
Thanks @Mia-Cross. This looks really nice. 👍 
Please squash the commits and see a few nits and comments below:

> I tried adding ScalewaySpec to the ClusterSpec in v1alpha2...

No worries, please ignore that.

> For the time being, only gossip DNS is supported but a custom DNS will soon come in a next PR.

DNS part may be need more time and attention. With this in mind, you may want to switch to the cloud resources provisioning instead and come back to DNS once you have a working cluster.

hakman (approved):
Thanks @Mia-Cross! 😄

PR #14359, opened 2022-09-30
[WIP] hetzner: Cluster without DNS or Gossip

hakman:
/cc @olemarkus @justinsb

### DIFF
--- a/pkg/apis/nodeup/config.go
+++ b/pkg/apis/nodeup/config.go
@@ -90,6 +90,8 @@ type BootConfig struct {
 	ConfigBase *string `json:",omitempty"`
 	// ConfigServer holds the configuration for the configuration server.
 	ConfigServer *ConfigServerOptions `json:",omitempty"`
+	// MasterInternalName is the API server name/address.
+	MasterInternalName string
 	// InstanceGroupName is the name of the instance group.
 	InstanceGroupName string `json:",omitempty"`
 	// InstanceGroupRole is the instance group role.
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -302,9 +302,6 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
 	if cluster.Spec.KubernetesVersion == "" {
 		return fmt.Errorf("KubernetesVersion not set")
 	}
-	if cluster.Spec.DNSZone == "" && !dns.IsGossipHostname(cluster.ObjectMeta.Name) {
-		return fmt.Errorf("DNSZone not set")
-	}
 
 	l := &Loader{}
 	l.Init()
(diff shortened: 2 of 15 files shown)


olemarkus:
If I understand this PR correctly, it tries to make empty master internal name allowed. That will break a few addons that must run before the `kuberntes.default` service exists. Since the manifests are rendered before LB creation, I am not sure how that will work.

hakman:
> If I understand this PR correctly, it tries to make empty master internal name allowed. That will break a few addons that must run before the `kuberntes.default` service exists. Since the manifests are rendered before LB creation, I am not sure how that will work.

Any example of such manifests?

olemarkus:
Cilium always requires it. Calico when using BPF mode. Karpenter also references it, but not sure its strictly required anymore.

hakman:
Easiest way to go about this would be to add a new entry in `/etc/hosts` on each node.
Though, this won't work for LBs with stable DNS name only, like in AWS. At least I'm not sure that the internal IP is stable.

pkg/apis/nodeup/config.go
```diff
@@ -90,6 +90,8 @@ type BootConfig struct {
 	ConfigBase *string `json:",omitempty"`
 	// ConfigServer holds the configuration for the configuration server.
 	ConfigServer *ConfigServerOptions `json:",omitempty"`
+	// MasterInternalName is the API server name/address.
+	MasterInternalName string
```
justinsb, line 94: I like this... if we're adding a field, should we (1) use the non-deprecated naming and (2) be clearer about whether it is apiserver or kops-controller?

How about `APIServerInternalName` ?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1418,6 +1417,24 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 		config.ApiserverAdditionalIPs = apiserverAdditionalIPs
 	}
 
+	bootConfig.MasterInternalName = cluster.Spec.MasterInternalName
+	if bootConfig.MasterInternalName == "" {
+		for _, addr := range apiserverAdditionalIPs {
+			if net.ParseIP(addr).IsPrivate() {
+				bootConfig.MasterInternalName = addr
```
justinsb, line 1424: Now I'm wondering if we should make `MasterInternalName` a []string, or even a list of objects, like the `/api` endpoint:

```
>  kubectl get --raw /api
{"kind":"APIVersions","versions":["v1"],"serverAddressByClientCIDRs":[{"clientCIDR":"0.0.0.0/0","serverAddress":"1.2.3.4:443"}]}
```

I guess because the version of nodeup is tightly coupled it doesn't much matter, this is easy to change going forward if we need to.

justinsb:
Looking good - one naming nit :-)

pkg/apis/nodeup/config.go
hakman: Initially the intention was to keep it as close as possible to the original name. Though, I think `APISever` should be good enough. Thanks for the suggestion.

PR #14362, opened 2022-10-01
Let kOps take ownership of any field currently owned by kubectl

olemarkus:


pkg/applylib/applyset/applyset.go
```diff
@@ -141,7 +144,24 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
 			return nil, fmt.Errorf("unknown scope for gvk %s: %q", gvk, restMapping.Scope.Name())
 		}
 
-		j, err := json.Marshal(obj)
+		currentObj, err := dynamicResource.Get(ctx, name, v1.GetOptions{})
```
justinsb, line 147: I do wonder whether we should put this into a separate method, because (I would hope) this isn't going to be the common path.

One possible approach .... when we apply, we get the "after" object.  We could inspect that to see if field ownership changes are needed.  It's slightly less efficient in this case, because I'd imagine we'd just reprocess every object, but I would expect this case also to be the rare case.

pkg/applylib/applyset/applyset.go
```diff
@@ -103,11 +106,11 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
 
 	for i := range trackers.items {
 		tracker := &trackers.items[i]
-		obj := tracker.desired
+		expectedObject := tracker.desired
```
justinsb, line 109: Nit: should probably match the name of the tracker field.  So I would argue either to call this desiredObj or to rename track.desired to tracker.expected.

(Personally I think desired better captures the idea that this is what the user wants, but open either way ... we used `e` for expected in all our kOps tasks)

pkg/applylib/applyset/applyset_test.go
```diff
@@ -1,128 +0,0 @@
-/*
```
justinsb, line 1: Do we want to delete the test?  If the test doesn't pass I'd rather t.Skip() it with a reason, and then we can try to make our mocks better?  I'm trying to incorporate mockkubeapiserver (and applyset) into kubebuilder-declarative-pattern here https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/248 .  I can keep copying and pasting back and forth, but I'm also trying to establish a go.mod there as that might be a good place to maintain it (eventually!)

pkg/applylib/applyset/applyset.go
```diff
@@ -161,3 +181,31 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+	meta := &v1.ObjectMeta{}
+	meta.SetManagedFields(fixedManagedFields)
+	patchObject := map[string]interface{}{
+		"metadata": meta,
+	}
+
+	jsonData, err := json.Marshal(patchObject)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marsal %q into json: %w", currentObject.GetName(), err)
+	}
```
justinsb, line 209: Typo: marshal

justinsb (approved):

justinsb:
I think this looks great - one open question on whether we want to remove the test, and I think we should think about whether we want this in the common path (I would prefer it wasn't, for performance).

But I think the logic is sound, and I'd be in favor of merging and then I can see if the (slightly more complicated) code we have internally to do this handles any cases we are missing.

olemarkus:
This one started flaking after this PR: https://testgrid.k8s.io/kops-misc#kops-aws-addon-resource-tracking

It seems to be hitting the validation timeout. Or that there is some race condition that makes things fail. It's at least not entirely fit for production.

PR #14371, opened 2022-10-03
GCE: Support nested virtualization

justinsb:
Add a field spec.gcpNestedVirtualization on the InstanceGroup, which
sets the corresponding field on the instanceTemplate.


upup/pkg/fi/cloudup/gcetasks/instancetemplate.go
```diff
@@ -365,10 +373,20 @@ func (e *InstanceTemplate) mapToGCE(project string, region string) (*compute.Ins
...
 	}
 
+	var advancedMachineFeatures *compute.AdvancedMachineFeatures
+
+	if e.GCPNestedVirtualization != nil {
+		if advancedMachineFeatures == nil {
+			advancedMachineFeatures = &compute.AdvancedMachineFeatures{}
+		}
+		advancedMachineFeatures.EnableNestedVirtualization = fi.BoolValue(e.GCPNestedVirtualization)
+	}
```
hakman, line 383: Would be nice if we always set this to true or false, to also disable if option is removed from config.
```suggestion
	advancedMachineFeatures &compute.AdvancedMachineFeatures{
		EnableNestedVirtualization: fi.BoolValue(e.GCPNestedVirtualization)
	}
```

hakman (changes requested):

burnyd:
Looking for this feature for a kops cluster where hardware virtualization is needed within Kubevirt on GCE.  To pickup this PR do you want me to pickup this same PR or create a new one?  Did it not get merged due to the unit tests that it failed?

PR #14378, opened 2022-10-04
Karpenter : fallback on ondemand instance by default

noony:
Hi,

Having ondemand and spot in capacity type can secure the usage of karpenter by always having a solution to spawn a new instance if there is no capacity in spot. 

The karpenter [documentation](https://karpenter.sh/v0.16.3/faq/#what-if-there-is-no-spot-capacity-will-karpenter-use-on-demand) : `If your Karpenter Provisioner specifies flexibility to both Spot and On-Demand capacity, Karpenter will attempt to provision On-Demand capacity if there is no Spot capacity available`

This will not change the basic behavior of karpenter spawning a spot instance if possible first, but it adds a fallback to start an ondemand instance if there is no spot available.

Thanks for your time !

### DIFF
--- a/docs/operations/karpenter.md
+++ b/docs/operations/karpenter.md
@@ -66,6 +66,6 @@ As mentioned above, kOps will manage a Provisioner resource per InstanceGroup. I
 ### Other minor limitations
 
 * Control plane nodes must be provisioned with an ASG, not Karpenter.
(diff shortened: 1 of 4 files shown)


olemarkus (approved):
Thanks. Absolutely makes sense.

PR #14382, opened 2022-10-04
update node-problem-detector to version 0.8.12

jjinno:
node-problem-detector has added multi-arch support in 0.8.12
https://github.com/kubernetes/node-problem-detector/releases/tag/v0.8.12

This just updates to that version to avoid CrashLoopBackOff on ARM (etc)


hakman (approved):
Really nice. Thanks @jjinno.

PR #14389, opened 2022-10-05
Fix default launch spec detection for spotinst

2solt:
Currently cluster creation is failing if `SpotInstanceGroupLabelOceanDefaultLaunchSpec` label is set for an instance group.
This should fix it.

### DIFF
--- a/pkg/model/awsmodel/spotinst.go
+++ b/pkg/model/awsmodel/spotinst.go
@@ -341,30 +341,34 @@ func (b *SpotInstanceGroupModelBuilder) buildOcean(c *fi.ModelBuilderContext, ig
 		ocean.UseAsTemplateOnly = fi.Bool(true)
 	}
 
-	ig := igs[0].DeepCopy()
-	if len(igs) > 1 {
-		for _, g := range igs {
-			for k, v := range g.ObjectMeta.Labels {
-				if k == SpotInstanceGroupLabelOceanDefaultLaunchSpec {
-					defaultLaunchSpec, err := parseBool(v)
-					if err != nil {
-						continue
+	if len(igs) == 0 {
+		return nil
+	}
+	var ig *kops.InstanceGroup
+	for _, g := range igs {
+		for k, v := range g.ObjectMeta.Labels {
+			if k == SpotInstanceGroupLabelOceanDefaultLaunchSpec {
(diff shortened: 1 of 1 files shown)


pkg/model/awsmodel/spotinst.go
```diff
@@ -341,30 +341,31 @@ func (b *SpotInstanceGroupModelBuilder) buildOcean(c *fi.ModelBuilderContext, ig
...
+					ig = g.DeepCopy()
+					break
 				}
 			}
 		}
-
-		klog.V(4).Infof("Detected default launch spec: %q", b.AutoscalingGroupName(ig))
 	}
+	if ig == nil {
+		ig = igs[0].DeepCopy()
```
hakman, line 365: I would expect to return or error in this case,
```suggestion
		return nil
```

hakman (changes requested):

pkg/model/awsmodel/spotinst.go
2solt: I ok with that, but the current and pass behaviour was if you not set the label it will use the first instance group for the default launch spec

2solt: Im ok with that, but the current and pass behaviour was if you not set the label it will use the first instance group for the default launch spec

hakman: I looked again through the code and seems fine like this.

pkg/model/awsmodel/spotinst.go
```diff
@@ -341,30 +341,31 @@ func (b *SpotInstanceGroupModelBuilder) buildOcean(c *fi.ModelBuilderContext, ig
...
-					}
-					if fi.BoolValue(defaultLaunchSpec) {
-						if ig != nil {
-							return fmt.Errorf("unable to detect default launch spec: "+
-								"multiple instance groups labeled with `%s: \"true\"`",
-								SpotInstanceGroupLabelOceanDefaultLaunchSpec)
-						}
-						ig = g.DeepCopy()
-						break
+	var ig *kops.InstanceGroup
```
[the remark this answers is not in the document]
hakman: @2solt Please add this additional check and squash the commits. Once this is done, I can merge and cherry-pick the fix. Thanks!

hakman (approved):
👍

PR #14397, opened 2022-10-08
nodeup script: accept strongly typed nodeup.Config

justinsb:
More self-documenting and flexible than accepting a string.


justinsb:
cc @hakman ... I'm revisiting the split-cloud/non-cloud work and thought this one was one we should have regardless.

hakman (approved):
Nice catch!

PR #14413, opened 2022-10-14
Allow snapshot controller to create volumesnapshotcontent

TwoStone:
For using dynamic provisioning of volumesnapshot the snapshot-controller has to be able to create volumesnapshotcontent objects at the cluster level. When deploying the snapshot-controller as a kops addon the controller is not able to create volumesnapshotcontent object because it lacks the create permission. This PR adds the create permission for the content objects, which should be the default according to https://github.com/kubernetes-csi/external-snapshotter/blob/v6.0.1/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml


olemarkus:
Thanks for this.
You probably have to run `./hack/update-expected.sh` to make the tests pass.

PR #14420, opened 2022-10-16
Move setting role taints to cloudup

olemarkus:
Role taints doesn't require any specific knowledge that exists only nodeup and therefore should be done in cloudup.


nodeup/pkg/model/kubelet_test.go
```diff
@@ -49,7 +49,7 @@ func TestTaintsApplied(t *testing.T) {
 		{
 			version:           "1.9.0",
 			taints:            []string{"foo", "bar", "baz"},
-			expectTaints:      []string{"foo", "bar", "baz"},
+			expectTaints:      []string{"foo", "bar", "baz", "node-role.kubernetes.io/master=:NoSchedule"},
 			expectSchedulable: true,
```
hakman, line 53: This is no longer used. Same goes from the struct above.
```suggestion
```

hakman (changes requested):

hakman (approved):

PR #14431, opened 2022-10-18
Scaleway cloud interface, model and tasks

Mia-Cross:
In this PR, we will be adding the ScwCloud interface, along with the model and tasks for server instances and ssh keys.


quantumsheep (changes requested):

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -16,7 +16,393 @@ limitations under the License.
...
+	return ""
+}
+
+func (s *scwCloudImplementation) ProviderID() kops.CloudProviderID {
+	return kops.CloudProviderScaleway
+}
+
+func (s *scwCloudImplementation) DNS() (dnsprovider.Interface, error) {
+	//TODO(Mia-Cross) implement me
+	panic("Scaleway doesn't have a DNS yet")
```
remyleone, line 153: Remove the TODO. What if we don't implement this function? it does not seem to be required by `fi.CloudInterface`

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -16,7 +16,393 @@ limitations under the License.
...
+	if err != nil {
+		return fmt.Errorf("error deleting cloud instance %s of group %s: %w", i.ID, i.CloudInstanceGroup.HumanName, err)
+	}
+
+	return nil
+}
+
+func (s *scwCloudImplementation) DeregisterInstance(i *cloudinstances.CloudInstance) error {
+	//TODO(Mia-Cross) implement me
+	panic("implement me")
```
remyleone, line 193: Needs to be implemented before getting out of draft

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -16,7 +16,393 @@ limitations under the License.
...
+		if err != nil {
+			return fmt.Errorf("error deleting group %q: %w", group.HumanName, err)
+		}
+	}
+	return nil
+}
+
+func (s *scwCloudImplementation) DetachInstance(i *cloudinstances.CloudInstance) error {
+	//TODO(Mia-Cross) implement me
+	panic("implement me")
```
remyleone, line 209: Need to be implemented before getting out of draft

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -16,7 +16,393 @@ limitations under the License.
...
+	return cloudInstanceGroup, nil
+}
+
+func (s *scwCloudImplementation) FindClusterStatus(cluster *kops.Cluster) (*kops.ClusterStatus, error) {
+	klog.V(8).Info("FindClusterStatus is not implemented yet for Scaleway")
+	return nil, nil
+}
+
+func (s *scwCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([]fi.ApiIngressStatus, error) {
+	//TODO(Mia-Cross) implement me
```
remyleone, line 304: Why do you panic for `GetApiIngressStatus` and just emit a log message for `FindClusterStatus`?

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -16,7 +16,393 @@ limitations under the License.
...
+	}
+	for {
+		_, err := s.instanceAPI.GetServer(&instance.GetServerRequest{
+			Zone:     s.zone,
+			ServerID: server.ID,
+		})
+		if is404Error(err) {
+			break
+		}
+	}
```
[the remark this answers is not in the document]
remyleone: Or we just do a WaitServer which is what this function is for.

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -16,7 +16,393 @@ limitations under the License.
...
+		})
+		if err != nil {
+			return fmt.Errorf("delete server %s: error deleting volume %s: %w", server.ID, server.Volumes[i].Name, err)
+		}
+	}
+
+	return nil
+}
+
+func (s *scwCloudImplementation) DeleteVolume(volume *instance.Volume) error {
```
remyleone, line 399: Should you add a wait for Volume at the end of this function to be sure to return when you are in a safe state?

upup/pkg/fi/cloudup/scaleway/cloud.go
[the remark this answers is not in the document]
quantumsheep: Does WaitServer works when deleting a server ?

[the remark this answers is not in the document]
hakman: Not sure we care that much to wait for the server to be deleted. I guess that, if there's no error on `DeleteServer()`, it will eventually disappear. Also, is there any constraint to stop the server before deleting it?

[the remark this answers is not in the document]
quantumsheep: We later delete the volumes attached to the server, there might be a constraint that disallow this action while the server is on. I don't know if the constraint is also present if the server still exists. @Mia-Cross

upup/pkg/fi/cloudup/scaleway/cloud.go
Mia-Cross: The DNS function is indeed required by the fi.Cloud interface. I removed the TODO but it will not stay unimplemented for long since the DNS part is already finished on an other branch.

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -16,7 +16,404 @@ limitations under the License.
...
+
+	serverGroups, err := findServerGroups(s, cluster.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find server groups: %w", err)
+	}
+
+	for _, ig := range instancegroups {
+		serverGroup, ok := serverGroups[ig.Name]
+		if !ok {
+			//for igName, serverGroup := range serverGroups {
```
remyleone, line 236: To be removed

remyleone (approved):

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -16,7 +16,395 @@ limitations under the License.
...
+	zone   scw.Zone
+	tags   map[string]string
+
+	accountAPI  *account.API
+	instanceAPI *instance.API
+}
+
+// NewScwCloud returns a Cloud with a Scaleway Client using the env vars SCW_ACCESS_KEY, SCW_SECRET_KEY,
+// SCW_DEFAULT_PROJECT_ID, SCW_DEFAULT_REGION and SCW_DEFAULT_ZONE
+func NewScwCloud(tags map[string]string) (ScwCloud, error) {
```
olemarkus, line 88: Maybe use `errors.NewAggregate(errs)` from k8s.io/apimachinery/pkg/util/errors to go through each of these checks in one go and return an aggregate error. Tends to be more user friendly than having the user error on one env, fix, then error on the next and repeat.

olemarkus (changes requested):
Lot of code to go through here. One suggestion for now. I need a bit more time just to review a few other potential improvements. Apologies for taking some time.

hakman (changes requested):
I also have more more code to go through. Sorry for the delay in review, I plan to go through the rest by end of week.

upup/pkg/fi/cloudup/scalewaytasks/instance.go
```diff
@@ -0,0 +1,199 @@
...
+			ServerID: srv.Server.ID,
+			Zone:     zone,
+		})
+		if err != nil {
+			return fmt.Errorf("error waiting for instance %s of group %q: %w", srv.Server.ID, fi.StringValue(expected.Name), err)
+		}
+	}
+
+	// If newInstanceCount < 0, we need to delete instances of this group
+	for i := 0; i > expected.Count; i-- {
```
hakman, line 152: Should this loop start from `newInstanceCount`, otherwise it wouldn't do much?
```suggestion
	for i := newInstanceCount; i > expected.Count; i-- {
```

upup/pkg/fi/cloudup/scalewaytasks/instance.go
```diff
@@ -0,0 +1,199 @@
...
+
+		igInstances, err := cloud.GetClusterServers(cloud.ClusterName(actual.Tags), actual.Name)
+		if err != nil {
+			return fmt.Errorf("error deleting instance: %w", err)
```
hakman, line 156:
```suggestion
			return fmt.Errorf("error finding instances: %w", err)
```

upup/pkg/fi/cloudup/scalewaytasks/instance.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+
+		for _, igInstance := range igInstances {
+			err = cloud.DeleteServer(igInstance)
+			if err != nil {
+				return fmt.Errorf("error deleting instance of group %s: %w", igInstance.Name, err)
+			}
+		}
```
hakman, line 164: This pretty much deletes all the instances in the group, instead of one. Maybe the intention was different?

upup/pkg/fi/cloudup/scalewaytasks/sshkey.go
```diff
@@ -0,0 +1,148 @@
...
+		KeyPairFingerPrint: fi.String(k.Fingerprint),
+	}
+
+	// Avoid spurious changes
+	if strings.Contains(fi.StringValue(sshKey.KeyPairFingerPrint), fi.StringValue(s.KeyPairFingerPrint)) {
+		klog.V(2).Infof("SSH key fingerprints match; assuming public keys match")
+		sshKey.PublicKey = s.PublicKey
+		sshKey.KeyPairFingerPrint = s.KeyPairFingerPrint
+	} else {
+		klog.V(2).Infof("Computed SSH key fingerprint mismatch: %q %q", fi.StringValue(s.KeyPairFingerPrint), fi.StringValue(sshKey.KeyPairFingerPrint))
```
hakman, line 77: Should this be an error?

upup/pkg/fi/cloudup/scalewaytasks/sshkey.go
```diff
@@ -0,0 +1,148 @@
...
+			return fmt.Errorf("error creating SSH keypair: %w", err)
+		}
+		expected.KeyPairFingerPrint = fi.String(key.Fingerprint)
+		klog.V(2).Infof("Created a new SSH keypair, id=%q fingerprint=%q", key.ID, key.Fingerprint)
+
+		return nil
+	}
+
+	expected.KeyPairFingerPrint = actual.KeyPairFingerPrint
+	klog.V(2).Infof("Using an existing SSH keypair, fingerprint=%q", fi.StringValue(expected.KeyPairFingerPrint))
```
hakman, line 145: Render is called when actual is different from expected. In this case, you must either update the key, or return error, right?

hakman (changes requested):

upup/pkg/fi/cloudup/scalewaytasks/instance.go
Mia-Cross: there was indeed an error, but it's that the loop should stop at newInstanceCount like in the one where there are instances to add. If expected.Count is lower than actual.Count, newInstanceCount will have a negative value that indicated how many instances we have to remove.

hakman: Should there be a change in the loop here to reflect ^^^?

hakman: no worries. thanks.

hakman:
@Mia-Cross anything you want to squash before merging the PR?
/lgtm

PR #14433, opened 2022-10-19
Bump alpha and stable channel with the latest kops releases

olemarkus:


channels/alpha
```diff
@@ -137,11 +137,11 @@ spec:
     requiredVersion: 1.11.10
   kopsVersions:
   - range: ">=1.25.0-alpha.1"
-    recommendedVersion: "1.25.0"
+    recommendedVersion: "1.25.2"
     #requiredVersion: 1.25.0
     kubernetesVersion: 1.25.3
   - range: ">=1.24.0-alpha.1"
-    recommendedVersion: "1.24.3"
+    recommendedVersion: "1.24.4"
```
hakman, line 144: Time to default to 1.25.2 for older k8s versions too?
```suggestion
    recommendedVersion: "1.25.2"
```

hakman (changes requested):

channels/alpha
olemarkus: Yeah. Let's do both.
Not sure about the AMI given that another one was released today.

hakman: Let's update AMI to latest.

olemarkus: Updated AMIs for alpha.

olemarkus: good catch. done.

hakman (approved):

PR #14434, opened 2022-10-19
add option to query AMI IDs from SSM

heybronson:
Adds the option to query an SSM parameter for an AMI ID. This helps when one has continuous AMI builds and wants kops to dynamically pull an AMI ID without needing to change the `image` in the kops cluster config.

Working Example:
```
❯ /Users/bronson/go/bin/kops get ig nodes-us-west-2a -oyaml
kind: InstanceGroup
metadata:
  creationTimestamp: "2022-05-11T22:35:23Z"
  generation: 1
  labels:
    kops.k8s.io/cluster: <REDACTED>
  name: nodes-us-west-2a
spec:
  image: ssm:/aws/service/canonical/ubuntu/server/18.04/stable/current/amd64/hvm/ebs-gp2/ami-id
  instanceMetadata:
    httpPutResponseHopLimit: 1
    httpTokens: required
  machineType: t3.medium
  maxSize: 1
  minSize: 1
  nodeLabels:
    kops.k8s.io/instancegroup: nodes-us-west-2a
  role: Node
  subnets:
  - us-west-2a
 ```


olemarkus:
This looks like a good feature to me. I am wondering a bit if this should be a template function rather than being done inline though just so one has some form of change control of what AMI is being used.

Also, this definitely requires both documentation and release notes. Would very much appreciate if that could be added to this PR.

heybronson:
@olemarkus Thanks for your comment.

I've updated the 1.26 release nodes and made a new section to `working-with-instancegroups`. Please take a look when you get the chance.

heybronson:
Thanks, updated that page as well.

PR #14440, opened 2022-10-21
hetzner: Create cluster without DNS or Gossip

hakman:
This is an updated approach to #14359.


hakman:
@justinsb @olemarkus I implemented all the feedback from #14359, so I think this is ready for review.

pkg/apis/kops/validation/validation.go
```diff
@@ -1613,8 +1613,8 @@ func validateExternalDNS(cluster *kops.Cluster, spec *kops.ExternalDNSConfig, fl
 	}
 
 	if spec.Provider == kops.ExternalDNSProviderExternalDNS {
-		if cluster.IsGossip() {
-			allErrs = append(allErrs, field.Forbidden(fldPath.Child("provider"), "external-dns does not support gossip clusters"))
+		if cluster.IsGossip() || cluster.UsesNoneDNS() {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("provider"), "external-dns does not support clusters without real DNS"))
```
johngmyers, line 1617: "without public DNS"?

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -86,7 +91,7 @@ func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.Se
 
 		sort.Strings(targets)
 		if len(targets) == 0 {
-			klog.Warningf("Did not find API endpoint for gossip hostname; may not be able to reach cluster")
+			klog.Warningf("Did not find API endpoint, may not be able to reach cluster")
```
johngmyers, line 94:
```suggestion
			klog.Warningf("Did not find API endpoint; may not be able to reach cluster")
```

nodeup/pkg/model/etc_hosts.go
```diff
@@ -30,32 +30,37 @@ var _ fi.ModelBuilder = &EtcHostsBuilder{}
...
+		task.Records = append(task.Records, nodetasks.HostRecord{
+			Hostname:  b.Cluster.Spec.MasterInternalName,
+			Addresses: []string{b.BootConfig.APIServer},
+		})
+		if b.UseKopsControllerForNodeBootstrap() {
 			task.Records = append(task.Records, nodetasks.HostRecord{
-				Hostname:  b.Cluster.Spec.MasterPublicName,
-				Addresses: []string{"127.0.0.1"},
+				Hostname:  "kops-controller.internal." + b.Cluster.Name,
+				Addresses: []string{b.BootConfig.APIServer},
```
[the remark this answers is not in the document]
johngmyers: Perhaps apiserver nodes don't make sense when using none DNS? If that's the case then there should be API validation against the combination.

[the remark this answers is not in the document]
hakman: None DNS implies at this time a load balancer or something similar and I assume that it would handle both API servers and kops-controller connections. Probably would be best to keep this part out of this PR.

pkg/apis/kops/validation/validation.go
hakman: Doesn't it work also with private DNS, available only inside the VPC?

hakman:
I also added a preview for the AWS implementation in https://github.com/kubernetes/kops/pull/14452/commits/95dcf31c64e6a06955034aca230ed43579f0e849.

nodeup/pkg/model/etc_hosts.go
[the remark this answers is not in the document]
johngmyers: As I understand apiserver nodes, the load balancer for the apiserver would go to the apiserver nodes, not the control plane nodes. You'd need a separate load balancer for kops-controller. @olemarkus ?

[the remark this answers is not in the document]
hakman: @johngmyers For kops-controller, we need a new listener + target group. See https://github.com/kubernetes/kops/pull/14452 for details.

pkg/apis/kops/validation/validation.go
johngmyers: I think "real dns" is a bit ambiguous. Perhaps separate messages for Gossip and NoneDNS?

hakman: Sounds good to me.

nodeup/pkg/model/etc_hosts.go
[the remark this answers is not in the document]
hakman: I also added an integration test for it, so this should make it easier to follow the actual changes from Public DNS to None DNS: https://github.com/kubernetes/kops/pull/14452/commits/062adb078ccb7d9652066f1f70a53a7c8cd220cd.

hakman:
Thanks @johngmyers. I squashed the commits that addressed the review comments.

johngmyers:
I'd prefer someone else also take a look at this.

hakman:
> I'd prefer someone else also take a look at this.

I agree. Just wanted to have clean code before the next round of reviews.

pkg/apis/kops/cluster.go
```diff
@@ -807,10 +807,6 @@ func (c *Cluster) FillDefaults() error {
 		c.Spec.MasterInternalName = "api.internal." + c.ObjectMeta.Name
 	}
 
-	if c.Spec.MasterPublicName == "" {
```
olemarkus, line 810: I am not a big fan of this method. E.g if the API name fields should be set in new_cluster or in populate cluster spec or something when dns type is none. Some care needs to be taken here wrt all the paths that may create a cluster object that expects this being set here.

I could not see that we have validation ensuring this fiels is always set.

pkg/apis/nodeup/config.go
```diff
@@ -90,6 +90,8 @@ type BootConfig struct {
 	ConfigBase *string `json:",omitempty"`
 	// ConfigServer holds the configuration for the configuration server.
 	ConfigServer *ConfigServerOptions `json:",omitempty"`
+	// APIServer is the API server IP address.
+	APIServer string `json:",omitempty"`
```
olemarkus, line 94: I know this one is just carrying on the convention, but I do have a hard time understanding when/how this is being used just from looking at the field.

olemarkus (commented):
Two concerns:
* that the intent of `BootConfig.APIServer` is clear and how/when it is used
* that we have covered all the bases with the removal of default value for `MasterInternalName`.

For the latter, I think we only need a truly minimal integration test that sets no more than the absolutely required fields.

nodeup/pkg/model/etc_hosts.go
```diff
@@ -30,32 +30,31 @@ var _ fi.ModelBuilder = &EtcHostsBuilder{}
...
-	useGossip := b.Cluster.IsGossip()
-	if !useGossip {
-		return nil
-	}
 
-	if b.IsMaster {
-		task := &nodetasks.UpdateEtcHostsTask{
-			Name: "control-plane-bootstrap",
-		}
+	task := &nodetasks.UpdateEtcHostsTask{
```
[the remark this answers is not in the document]
hakman: Task is added later only if any records are set.

cmd/kops/integration_test.go
```diff
@@ -276,7 +276,7 @@ func TestMinimal_v1_26(t *testing.T) {
 // TestHetzner runs the test on a minimum configuration
 func TestHetzner(t *testing.T) {
 	t.Setenv("HCLOUD_TOKEN", "REDACTED")
-	newIntegrationTest("minimal.k8s.local", "minimal_hetzner").
```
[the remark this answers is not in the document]
hakman: Yes, wanted to show how a cluster changes from Gossip to None DNS.

pkg/apis/kops/cluster.go
hakman: Not quite sure why someone would expect this to be there. It should not be populated to private clusters or Gossip ones either.

pkg/apis/nodeup/config.go
hakman: True, the comment should be more detailed.

olemarkus (approved):

PR #14442, opened 2022-10-21
Fix pdb for identity webhook

olemarkus:
Also removing namespace from non-namespaced resource


tests/e2e/templates/many-addons.yaml.tmpl
```diff
@@ -50,6 +50,8 @@ spec:
     enabled: true
     enableSQSTerminationDraining: true
   nonMasqueradeCIDR: 100.64.0.0/10
+  podIdentityWebhook:
+    enabled: true
```
hakman, line 54: Does this trigger any change?

hakman (approved):

PR #14450, opened 2022-10-23
Drop support for k8s 1.23 in kOps 1.28

johngmyers:
With Kubernetes releasing three times a year, it makes sense to narrow our supported range a bit.

/hold for wide consensus
/kind office-hours

### DIFF
--- a/docs/releases/1.26-NOTES.md
+++ b/docs/releases/1.26-NOTES.md
@@ -19,7 +19,7 @@ This is a document to gather the release notes prior to the release.
 
 * Support for Kubernetes version 1.21 is deprecated and will be removed in kOps 1.27.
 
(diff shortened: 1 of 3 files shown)


hakman:
Maybe we can keep it the same it is now. Kubernetes seems to become more stable so there's not much extra work supporting older versions.

olemarkus:
My largest concern would be configuration file format drift. When K8s X requires v1beta1 and K8s X+4 has dropped support for v1beta1, things tend to get annoying in nodeup. So strictly from a maintenance perspective, I'd prefer 2 stable + 1 deprecated to match the K8s rolling window.

But right now I also see the maintenance burden somewhat low, and getting people on latest kOps is more important than getting people on latest k8s.

PR #14452, opened 2022-10-25
aws: Create cluster without DNS or Gossip

hakman:
Follow-up PR to https://github.com/kubernetes/kops/pull/14440.

/hold for #14440 to merge first


hakman:
Looks like e2e tests pass without issues using `--dns=none`.

johngmyers:
We should make presubmits for dnsnone and apiserver-nodes-dnsnone. (Unless the combination of apiserver-nodes and dnsnone doesn't make sense, in which case we should catch that in API validation.)

hakman:
> We should make a presubmit for apiserver-nodes-dns-none. (Unless the combination of apiserver-nodes and dnsnone doesn't make sense, in which case we should catch that in API validation.)

I added an extra check to keep the `kops-controller` target group limited to master asgs. The apiserver related target groups are also added to apiserver asgs.

I also added an e2e template for APIServer nodes with LB and without DNS.

hakman:
In the end, seems that APIServer nodes with LB and without DNS will not work for now, due to the lack of access to etcd, which would require additional target groups (if at all possible).

olemarkus:
You'd need an ELB per cluster, probably (since different etcd clusters can have different sizes/IGs)

hakman:
> You'd need an ELB per cluster, probably (since different etcd clusters can have different sizes/IGs)

Why not just different target groups added to the apiserver load balancer, each with his own listening port?

olemarkus:
Yeah that could be possible. I don't know what would be the least complex there.

hakman:
> Yeah that could be possible. I don't know what would be the least complex there.

It's also about the cost. One more LB would mean at least an extra USD 30.

johngmyers:
Does the combination make sense? Would a cluster that is big enough to need apiserver nodes want to run without DNS?

hakman:
> Does the combination make sense? Would a cluster that is big enough to need apiserver nodes want to run without DNS?

Not really. I think as is it makes sense to get started. Would make that part a little easier by removing any need for a DNS zone.

johngmyers:
Someone getting started wouldn't need apiserver node. I think we should just add API validation to prohibit this combination and move on.

hakman:
> Someone getting started wouldn't need apiserver node. I think we should just add API validation to prohibit this combination and move on.

That's what I meant. Already added.

olemarkus (commented):
/lgtm

olemarkus:
Add release notes maybe?

hakman:
> Add release notes maybe?

Will add as a separate PR.

pkg/model/awsmodel/api_loadbalancer.go
```diff
@@ -521,6 +557,33 @@ func (b *APILoadBalancerBuilder) Build(c *fi.ModelBuilderContext) error {
 		}
 	}
 
+	// Allow kops-controller to the master instances from the ELB
+	if b.Cluster.UsesNoneDNS() && b.APILoadBalancerClass() == kops.LoadBalancerClassNetwork {
```
johngmyers, line 561: Should we prohibit NoneDNS with Classic API loadbalancers?

johngmyers (approved):
/approve
/hold for question in review

pkg/model/awsmodel/api_loadbalancer.go
hakman: Sure, done.

PR #14457, opened 2022-10-27
WIP: Split task without generics

justinsb:


justinsb:
I think the interesting stuff here is all in the `fi` package, the rest is boilerplate.

upup/pkg/fi/executor.go
```diff
@@ -184,7 +184,12 @@ func (e *executor) forkJoin(tasks []*taskState) []error {
 			results[index] = fmt.Errorf("function panic")
 			defer wg.Done()
 			klog.V(2).Infof("Executing task %q: %v\n", ts.key, ts.task)
-			results[index] = ts.task.Run(e.context)
+			if nodeTask, ok := ts.task.(NodeTask); ok {
+				results[index] = nodeTask.Run(e.context.(*NodeContext))
+			} else {
+				cloudTask := ts.task.(CloudTask)
+				results[index] = cloudTask.Run(e.context.(*CloudContext))
+			}
```
johngmyers, line 192: More runtime errors which would be compile time with generics.

upup/pkg/fi/task.go
```diff
@@ -25,13 +25,23 @@ import (
 )
 
 type Task interface {
-	Run(*Context) error
+	HasName
```
johngmyers, line 28: I'd like to get it so node tasks don't have to implement unnecessary boilerplate methods such as `GetName`.

upup/pkg/fi/has_address.go
```diff
@@ -19,9 +19,9 @@ package fi
 // HasAddress is implemented by elastic/floating IP addresses in order to include
 // relevant dynamically allocated addresses in the api-server's server TLS certificate.
 type HasAddress interface {
-	Task
+	//Task
```
johngmyers, line 22: `CloudTask`? I don't want anything not a `Task` to implement this, as such things won't have `FindAddresses()` called.

justinsb:
Addressed by alternate approach

/close

PR #14460, opened 2022-10-27
hetzner: Use kops-controller for node bootstrap

hakman:
/cc @justinsb @johngmyers


upup/pkg/fi/cloudup/hetzner/verifier.go
```diff
@@ -0,0 +1,94 @@
...
+	if !strings.HasPrefix(token, HetznerAuthenticationTokenPrefix) {
+		return nil, fmt.Errorf("incorrect authorization type")
+	}
+	token = strings.TrimPrefix(token, HetznerAuthenticationTokenPrefix)
+
+	serverID, err := strconv.Atoi(token)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert server ID %q to int: %w", token, err)
+	}
+	server, _, err := h.client.Server.GetByID(context.TODO(), serverID)
```
olemarkus, line 67: Is there any good reason for creating a new `context.TODO` here instead of passing in `ctx` or building on top of that one?

olemarkus (changes requested):

upup/pkg/fi/cloudup/hetzner/verifier.go
hakman: Nop, copy/paste in action :).

olemarkus:
Looks good :+1:

upup/pkg/fi/cloudup/hetzner/authenticator.go
```diff
@@ -0,0 +1,45 @@
...
+	return &hetznerAuthenticator{}, nil
+}
+
+func (h hetznerAuthenticator) CreateToken(body []byte) (string, error) {
+	serverID, err := metadata.NewClient().InstanceID()
+	if err != nil {
+		return "", fmt.Errorf("failed to retrieve server ID: %w", err)
+	}
+
+	return HetznerAuthenticationTokenPrefix + strconv.Itoa(serverID), nil
```
hakman, line 44: `serverID` is not exactly the most secure token, but options are limited with Hetzner.
This could be improved in the future by also checking:
1. the source IP of the request
2. the age of the server (and allow only new servers to authenticate)
3. ...

olemarkus (approved):

PR #14461, opened 2022-10-27
ipv6: Tolerate multiple routes to the same NAT Gateway

johngmyers:


upup/pkg/fi/cloudup/awstasks/natgateway.go
```diff
@@ -225,14 +225,18 @@ func findNatGatewayFromRouteTable(cloud awsup.AWSCloud, routeTable *RouteTable)
 					return nil, fmt.Errorf("Could not find '%s' tag from route table", awsup.TagClusterName)
 				}
 				filteredNatGateways := []*ec2.NatGateway{}
+				natGatewayIDsSeen := map[string]bool{}
 				for _, natGatewayID := range natGatewayIDs {
 					gw, err := findNatGatewayById(cloud, natGatewayID)
 					if err != nil {
 						return nil, err
 					}
```
hakman, line 234: Nit: Would it be easier to read if we skip early?
```suggestion
					if natGatewayIDsSeen[*natGatewayID] {
						continue
					}
```

johngmyers: Moved it even earlier.

hakman (approved):

PR #14463, opened 2022-10-28
Fix Prometheus scraping for pod-identity-webhook

jim-barber-he:
The Service annotation for the pod-identity-webhook is asking Prometheus to scrape its metrics from `https://` but this results in a HTTP 404 Not Found response.
The pod-identity-webhook pods expose their metrics via HTTP on TCP port 9999 (the same port where the healthz checks go).
Therefore update the annotations to use `http://:9999` instead.


johngmyers:
/ok-to-test

The verify-goimports failed. The PR shouldn't reindent the comments in the clientset register.go

jim-barber-he:
> /ok-to-test
> 
> The verify-goimports failed. The PR shouldn't reindent the comments in the clientset register.go

Yeah I was using an older version of golang (1.17.1).
Updating to the current version and rerunning the formatting tools stops it from messing with the indent in those golang src files.

PR #14468, opened 2022-10-28
Create NLB instead of CLB for bastion

johngmyers:
Bastions didn't work in IPv6 clusters because they were placed in the IPv6-only private networks, which the CLB couldn't talk to. IPv6 bastions are now placed in dualstack networks, where they can be instance targets.

While I was at it, I switched the bastion LB from CLB to NLB. If there are any subnets with IPv6 CIDRs, it makes the NLB dualstack and publishes an AAAA DNS record for it.

Doesn't delete the old CLB (yet).


docs/releases/1.26-NOTES.md
```diff
@@ -6,6 +6,8 @@ This is a document to gather the release notes prior to the release.
 
 # Significant changes
 
+* On AWS, bastions are now fronted by a Network Load Balancer.
+
```
hakman, line 10: Maybe we should make an AWS section for specific changes?

johngmyers: Done

hakman: Thanks. Will try to look at the code during weekend.

hakman (approved):

PR #14473, opened 2022-10-29
aws: Set the target group health check interval to 10s

hakman:
The default health check interval is 30s, which seems too long to me:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html
> **HealthCheckIntervalSeconds**
> The approximate amount of time, in seconds, between health checks of an individual target. The range is 5–300 seconds. The default is 30 seconds if the target type is instance or ip and 35 seconds if the target type is lambda.


johngmyers:
I'm wary of changing AWS defaults, but LBC appears to default to 10s for Services.

Do we need an `Interval` field in `awstasks.TargetGroup`? It doesn't appear to be configurable and we can't change the value for existing TGs.

hakman:
> Do we need an Interval field in awstasks.TargetGroup? It doesn't appear to be configurable and we can't change the value for existing TGs.

We don't need an `Interval` field, but wanted to keep it similar to `HealthyThreshold` and `UnhealthyThreshold`. It may be nice to display as part of the task.

johngmyers (approved):

PR #14475, opened 2022-10-29
Refactor to avoid looking up SSH keypairs twice

johngmyers:


upup/pkg/fi/cloudup/awstasks/sshkey.go
```diff
@@ -66,14 +66,17 @@ func (e *SSHKey) find(cloud awsup.AWSCloud) (*SSHKey, error) {
...
-			return nil, nil
+			err = nil
 		}
 	}
 	if err != nil {
 		return nil, fmt.Errorf("error listing SSHKeys: %v", err)
 	}
 
 	if response == nil || len(response.KeyPairs) == 0 {
+		if e.IsExistingKey() && *e.Name != "" {
```
hakman, line 77: At this point `*e.Name` should already be set. In general, we should `e.Name != ""` for the task to be accepted, right?
Also, I think `find()` can be now inlined in `Find()`, as it's the only place where it's called.

johngmyers: `e.Name != ""` would fail to compile due to the type mismatch. This test comes directly from the previous code.

`find()` is also called from `RenderCloudformation()`.

hakman (approved):

PR #14485, opened 2022-11-02
Upgrade Amazon VPC CNI to v1.12.0

MoShitrit:
### [Release Page](https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.12.0)

/cc @rifelpet @hakman @olemarkus @johngmyers 
Hey y'all!

Wanted to point out the below updates from the release page:

> This new version removes dependency on CRI socket(e.g. dockershim.sock).
> 
> 🚨 🚨 🚨 Action Required For Upgrading
> 
> To upgrade to VPC CNI >=v1.12.0, you must upgrade to VPC CNI >=v1.7.x first. We recommend that you update one minor version at a time.
> aws-vpc-cni Helm chart v1.2.0 is released with VPC CNI v1.12.0, thus no longer supports the cri.hostPath.path. If you need to install a VPC CNI <v1.12.0 with helm chart, a aws-vpc-cni Helm chart with version <v1.2.0 should be used.

Regarding the `dockershim.sock` removal, I suppose we should be good since it was already remove everywhere else already, correct?

And about the requirement to have a minimum version of v1.7 first (of the CNI plugin) - I suppose we should be good there as well, since the general `kOps` best practice is to _always_ upgrade one major version at a time- is that a correct assumption? 

Let me know if something seems off, but I went by the general practice of vendorizing our template from the original chart (updated the link at the top of the template to point to the new version now).

Thanks!

### DIFF
--- a/tests/integration/update_cluster/many-addons-ccm-irsa/data/aws_s3_object_minimal.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/many-addons-ccm-irsa/data/aws_s3_object_minimal.example.com-addons-bootstrap_content
@@ -112,7 +112,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.16
     manifest: networking.amazon-vpc-routed-eni/k8s-1.16.yaml
-    manifestHash: da1a13abe387e36b409914ba1158040eaaa780f5dc1d925ee83f9d6cb73c9292
+    manifestHash: 2dbc5844330df7de478e4d739f390c1525ce00426302a9bcd5cd7f7ffa9af437
     name: networking.amazon-vpc-routed-eni
     needsRollingUpdate: all
     selector:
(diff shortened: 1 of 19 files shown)


johngmyers:
The kOps recommendation is to use the latest version of kOps that supports your version of Kubernetes. It is not to upgrade one version of kOps at a time.

kOps 1.26 supports down to k8s 1.21. kOps 1.21, the oldest version supporting k8s 1.21, uses amazon-vpc-cni version 1.7. So we're good.

MoShitrit:
> The kOps recommendation is to use the latest version of kOps that supports your version of Kubernetes. It is not to upgrade one version of kOps at a time.
> 
> kOps 1.26 supports down to k8s 1.21. kOps 1.21, the oldest version supporting k8s 1.21, uses amazon-vpc-cni version 1.7. So we're good.

Gotcha, thanks for the clarification!

johngmyers (approved):

PR #14488, opened 2022-11-02
Deprecate Ubuntu 18.04

johngmyers:


hakman:
Is there any technical reason to deprecate Ubuntu 18.04?
It should be still supported for another 5 years by Canonical and there are not not much maintenance burden for it.

johngmyers:
It's leaving "Standard Support" soon, so our test coverage for it will decrease.

hakman (approved):

PR #14492, opened 2022-11-03
hetzner: Add listener for kops-controller when using it for node bootstrap

hakman:
A few more nits...

/cc @olemarkus


olemarkus (approved):
To

PR #14497, opened 2022-11-05
WIP Change "GCE" to "GCP" in v1alpha3 API

johngmyers:


johngmyers:
All the Kubernetes components call it "gce", so this change is going against the grain.

hakman:
> All the Kubernetes components call it "gce", so this change is going against the grain.

Seems that there's a change there too. See https://github.com/kubernetes/cloud-provider-gcp.

johngmyers:
This is on hold until the gce-gce periodics pass

PR #14498, opened 2022-11-05
Update TopologySpec for v1alpha3 API

johngmyers:
I don't like that the values for `dns` are capitalized, whereas the others are not. In general, I'm not a fan of the capitalization, but is there a wider Kubernetes convention to follow?


pkg/apis/kops/v1alpha3/topology.go
```diff
@@ -22,30 +22,27 @@ const (
 )
 
 type TopologySpec struct {
-	// The environment to launch the Kubernetes masters in public|private
-	Masters string `json:"masters,omitempty"`
+	// ControlPlane specifies the environment for launching the control plane nodes. (public, private)
+	ControlPlane string `json:"control-plane,omitempty"`
```
hakman, line 26: Not quite sure we have any other place using `-`.
```suggestion
	ControlPlane string `json:"controlPlane,omitempty"`
```

johngmyers: Fixed

hakman (approved):

olemarkus:
> I don't like that the values for `dns` are capitalized, whereas the others are not. In general, I'm not a fan of the capitalization, but is there a wider Kubernetes convention to follow?

To answer this one: enum-type values are alaways capitalised (upper camel-case) in K8s APIs. I don't have a reference for that convention, but you see it more or less everywhere. So it's the other fields that are wrong. In v1alpha3, I think we should ensure the enums are typede typed and do conversions of the values so they get the correct case convention.

PR #14501, opened 2022-11-06
Boot nodes without state store access

hakman:
It's been a year since https://github.com/kubernetes/kops/pull/10469. I think it would be a good idea to always enable the feature when `kops-controller` is used for node bootstrap.

/cc @justinsb @johngmyers @olemarkus @rifelpet


nodeup/pkg/model/kops_controller.go
```diff
@@ -62,6 +62,9 @@ func (b *KopsControllerBuilder) Build(c *fi.ModelBuilderContext) error {
 		Subject:        nodetasks.PKIXName{CommonName: "kops-controller"},
 		AlternateNames: []string{"kops-controller.internal." + b.Cluster.ObjectMeta.Name},
 	}
+	if b.BootConfig.APIServer != "" {
+		issueCert.AlternateNames = append(issueCert.AlternateNames, b.BootConfig.APIServer)
```
johngmyers, line 66: kops-controller isn't the apiserver, so shouldn't have the apiserver domain in its SANs.

hakman: This is only set when using `--dns=none` and represents the IP address of the API server.

johngmyers: Perhaps rename it to `APIServerIP`? And what about HA control planes? Or IPv6 clusters, where APIServer nodes are dual-stack?

johngmyers: Or is it `APIServerLoadBalancerIP`?

johngmyers:
This should also remove the code that gives worker nodes IAM permissions to the bucket.

johngmyers:
Also remove the feature flag.

nodeup/pkg/model/kops_controller.go
hakman: `APIServerIP` it is :).

pkg/model/bootstrapscript.go
```diff
@@ -131,7 +131,7 @@ func (b *BootstrapScript) buildEnvironmentVariables(cluster *kops.Cluster) (map[
 		env["GOSSIP_DNS_CONN_LIMIT"] = os.Getenv("GOSSIP_DNS_CONN_LIMIT")
 	}
 
-	if os.Getenv("S3_ENDPOINT") != "" {
+	if os.Getenv("S3_ENDPOINT") != "" && (!model.UseKopsControllerForNodeBootstrap(cluster) || b.ig.IsMaster()) {
```
johngmyers, line 134: Given there's still IAM for `/addons/*` and `/secrets/dockerconfig`, is this not still needed?

pkg/model/iam/tests/iam_builder_node_strict_ecr.json
```diff
@@ -7,8 +7,6 @@
       "Effect": "Allow",
       "Resource": [
         "arn:aws-test:s3:::kops-tests/iam-builder-test.k8s.local/addons/*",
-        "arn:aws-test:s3:::kops-tests/iam-builder-test.k8s.local/cluster-completed.spec",
-        "arn:aws-test:s3:::kops-tests/iam-builder-test.k8s.local/igconfig/node/*",
         "arn:aws-test:s3:::kops-tests/iam-builder-test.k8s.local/pki/private/kube-proxy/*",
         "arn:aws-test:s3:::kops-tests/iam-builder-test.k8s.local/pki/private/kubelet/*",
```
johngmyers, line 11: Looks like a bug in the test. It probably needs to set AWS as the cloud provider.

pkg/model/bootstrapscript.go
hakman: `/addons/*` could also be removed on worker nodes, as Protokube only applies them from the control plane.
The only remaining nit is `/secrets/dockerconfig`. Maybe we could keep this around if at least a `dockerconfig` secret exists?

pkg/model/iam/tests/iam_builder_node_strict_ecr.json
hakman: https://github.com/kubernetes/kops/pull/14517

pkg/model/bootstrapscript.go
olemarkus: I guess a followup would be to also serve the secrets from kops-controller.

hakman: Yes, that would be a good followup.

johngmyers: Do we need to set these env vars in order for nodeup to be able to read dockerconfig?

hakman: Only for non-AWS clouds that don't have object storage, like Hetzner.

johngmyers: It looks like this would affect Hetzner and GCE using a custom S3 state stores and also using a Docker secret.

I think this change should be deferred until after the secrets are served from kops-controller.

hakman: From security point of view, giving access to S3 fully is much worse than not having a Docker secret.

johngmyers: Perhaps add a condition testing whether or not a dockerconfig secret exists?

hakman: Using config for bootstrap switches to using `configserverSecretStore` automatically, which is not implemented.
https://github.com/kubernetes/kops/blob/0be79b25b74e5bec3ddd0907918605b39600e1dd/pkg/configserver/secretstore.go#L38-L41
We get away with it due to error not being checked:
https://github.com/kubernetes/kops/blob/0be79b25b74e5bec3ddd0907918605b39600e1dd/nodeup/pkg/model/secrets.go#L52
I would say that, unless I am missing something, this doesn't work in general.

hakman: Please see https://github.com/kubernetes/kops/pull/14501/commits/8ba1fa46c3f36a5a9334ee52bbec085354e4446b for a quick implementation of serving secrets from kops-controller.

pkg/configserver/secretstore.go
```diff
@@ -26,18 +26,25 @@ import (
 
 // configserverSecretStore is a SecretStore backed by the config server.
 type configserverSecretStore struct {
-	nodeConfig *nodeup.NodeConfig
+	nodeSecrets map[string]string
 }
 
 func NewSecretStore(nodeConfig *nodeup.NodeConfig) fi.SecretStore {
```
justinsb, line 32: Might be clearer to just take the `NodeSecrets` field, but that's just my 2c

cmd/kops-controller/pkg/server/node_config.go
```diff
@@ -61,5 +61,21 @@ func (s *Server) getNodeConfig(ctx context.Context, req *nodeup.BootstrapRequest
...
+			"dockerconfig",
+		}
+		nodeConfig.NodeSecrets = make(map[string]string)
+		for _, id := range secretIDs {
+			secret, err := s.secretStore.FindSecret(id)
+			if err != nil {
+				return nil, fmt.Errorf("error loading secret %q: %w", id, err)
+			}
+			if secret != nil && secret.Data != nil {
+				nodeConfig.NodeSecrets[id] = string(secret.Data)
```
justinsb, line 75: Maybe we should make NodeSecrets `map[string][]byte` ?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1424,16 +1405,39 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
 				break
 			}
 		}
 	}
 
+	useConfigServer := apiModel.UseKopsControllerForNodeBootstrap(cluster) && (role != kops.InstanceGroupRoleMaster)
+	if useConfigServer {
+		host := "kops-controller.internal." + cluster.ObjectMeta.Name
+		if cluster.UsesNoneDNS() {
+			host = bootConfig.APIServerIP
```
justinsb, line 1422: This is where I think it would be neat to have a custom dialer.  I'll see if I can find the code I was thinking of!

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1392,25 +1392,6 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 		}
 	}
 
-	useConfigServer := featureflag.KopsControllerStateStore.Enabled() && (role != kops.InstanceGroupRoleMaster)
```
justinsb, line 1395: Ah, this is nice, I like the idea of using the helper function instead of the flag :-)

justinsb:
Overall this LGTM.  The only one that I think is a blocker is the json change on BootstrapResponse Certs -> certs (if I'm reading it right).  We should also think about whether secrets should be `map[string][]byte` because that's also potentially a breaking change.

Otherwise, things like using a custom dialer we can change later.

Will approve after we figure out the BootstrapResponse JSON format, as that's potentially breaking changes.

justinsb:
We do have a custom dialer / resolver BTW: https://github.com/kubernetes/kops/blob/ed6bbc74172998e7bb4183fa667b064f93756bef/upup/pkg/fi/nodeup/nodetasks/bootstrap_client.go#L150-L152

So you might not need to write to /etc/hosts (at least for this particular hostname, I suspect you'll still have to write the apiserver!).

pkg/apis/nodeup/config.go
```diff
@@ -90,9 +90,9 @@ type BootConfig struct {
 	ConfigBase *string `json:",omitempty"`
 	// ConfigServer holds the configuration for the configuration server.
 	ConfigServer *ConfigServerOptions `json:",omitempty"`
-	// APIServer is the API server IP address.
+	// APIServerIP is the API server IP address.
 	// This field is used for adding an alias for api.internal. in /etc/hosts, when Topology.DNS.Type == DNSTypeNone.
-	APIServer string `json:",omitempty"`
+	APIServerIP string `json:",omitempty"`
```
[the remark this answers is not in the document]
justinsb: More of a map of well-known hosts, like `Hosts map[string]string`

pkg/apis/nodeup/bootstrap.go
```diff
@@ -35,7 +35,7 @@ type BootstrapRequest struct {
 // BootstrapResponse is a response to a BootstrapRequest.
 type BootstrapResponse struct {
 	// Certs are the issued certificates.
-	Certs map[string]string
+	Certs map[string]string `json:"certs,omitempty"`
```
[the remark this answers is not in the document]
johngmyers: In general, nodes can only bootstrap off of kops-controllers of the same version because otherwise the hashes won't match. So there's no real problem.

This change isn't really necessary. The protocol is internal; it doesn't have to look pretty.

pkg/featureflag/featureflag.go
```diff
@@ -78,8 +78,6 @@ var (
 	ClusterAddons = new("ClusterAddons", Bool(false))
 	// Azure toggles the Azure support.
 	Azure = new("Azure", Bool(false))
-	// KopsControllerStateStore enables fetching the kops state from kops-controller, instead of requiring access to S3/GCS/etc.
-	KopsControllerStateStore = new("KopsControllerStateStore", Bool(false))
```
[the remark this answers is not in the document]
johngmyers: We'll get a fair amount of test coverage before we release. The only area of concern would be the lack of coverage for dockerconfig. But I think leaving in the old code would be riskier.

johngmyers:
We should put a hash of the dockerconfig in the nodeconfig, so that changes to it result in rolling update. But that is a preexisting issue.

johngmyers:
I have no major issues other than the ones that Justin identified.

pkg/apis/nodeup/bootstrap.go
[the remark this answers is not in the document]
hakman: `json:"Certs,omitempty"` seems reasonable for now. We can have a cleaner version later.

cmd/kops-controller/pkg/server/server.go
```diff
@@ -75,6 +77,12 @@ func NewServer(opt *config.Options, verifier bootstrap.Verifier) (*Server, error
 	}
 	s.configBase = configBase
 
+	p, err := vfs.Context.BuildVfsPath(opt.SecretStore)
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse SecretStore %q: %v", opt.SecretStore, err)
```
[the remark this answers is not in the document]
hakman: Somehow I missed this one. Thanks! :)

pkg/apis/nodeup/config.go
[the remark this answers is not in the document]
hakman: Seems like a good idea.

pkg/featureflag/featureflag.go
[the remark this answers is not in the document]
hakman: I think this is more clear later. I am using `UseKopsControllerForNodeBootstrap(cluster)` instead of the flag.

pkg/configserver/secretstore.go
hakman: I agree. Thanks for the suggestion.

cmd/kops-controller/pkg/server/node_config.go
hakman: That would be nice.

pkg/apis/nodeup/bootstrap.go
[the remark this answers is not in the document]
johngmyers: The `omitempty` seems superfluous, but whatever.

hakman:
Thanks for the reviews. I think I addressed most of the feedback, with some left for separate PRs.

justinsb:
Thanks - this lgtm!

/approve

PR #14502, opened 2022-11-06
[WIP] Stop generating unused cluster_spec.yaml config file for nodes

hakman:
/cc @johngmyers @justinsb @olemarkus


rifelpet:
Maybe worthy of a release note if folks are utilizing this file themselves?

hakman:
```
jgmyers: I think it's there to make sure that cluster spec changes cause rolling update.
jgmyers: What we want to axe is the cluster completed.
```

rifelpet:
Would reducing it to a hash in the userdata be sufficient?

hakman:
We already have it as a hash in the userdata. But if we want to axe cluster completed instead...

johngmyers:
It should at least show up somewhere in the integration test output

PR #14504, opened 2022-11-07
v1alpha3: Move API-related settings under API

johngmyers:


pkg/apis/kops/cluster.go
```diff
@@ -452,12 +444,20 @@ type RBACAuthorizationSpec struct{}
...
+type APISpec struct {
+	// DNS will be used to provide configuration for the Kubernetes API's DNS server.
 	DNS *DNSAccessSpec `json:"dns,omitempty"`
-	// LoadBalancer is the configuration for the kube-apiserver ELB
+	// LoadBalancer is the configuration for the Kubernetes API load balancer.
 	LoadBalancer *LoadBalancerAccessSpec `json:"loadBalancer,omitempty"`
+	// PublicName is the external DNS name for the Kubernetes API.
+	PublicName string `json:"publicName,omitempty"`
+	// InternalName is the internal DNS name for the Kubernetes API.
+	InternalName string `json:"internalName,omitempty"`
```
hakman, line 456: I was always curious why we have `InternalName` as a field. It implies it can be changed.
Can `InternalName` be replaced by a function that always returns `"api.internal." + cluster.Name`?

olemarkus (approved):

Issue #14518, opened 2022-11-09
Cluster-autoscaler autodiscovery and priorityexpander configMap


chubchubsancho:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
1.1. Autodiscovery:
Actually when deploying `cluster-autoscaler` addons, each instance group allowed to scale is added to `cluster-autoscaler` command line through:
```
- --nodes={{ $spec.MinSize }}:{{ $spec.MaxSize }}:{{ $name }}{{- if not (eq GetCloudProvider "gce") }}.{{ ClusterName }}{{ end -}}
```
`cluster-autoscaler` is able to auto discover instance group in different way (based on cloud provider).
eg for `AWS`: it search `ASG` with those tags:
```
k8s.io/cluster-autoscaler/enabled
k8s.io/cluster-autoscaler/{{ cluster_name }}
```
instead of writing one adding one argument per instance group to `cluster-autoscaler` command we could only add one and let `cluster-autoscaler` retrieve ASG.
```
- --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/{{ ClusterName }}
```

1.2. priority-expander confimMap
Currently we need to create a configMap for priority expander before switching to this.
It could be nice to be able to use it without out of the box on kOps deployement.

**2. Feel free to provide a design supporting your feature request.**

PR #14519, opened 2022-11-09
Feat(ca): priority expander config

chubchubsancho:
Fixes: #14518


upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -258,6 +258,22 @@ spec:
     app.kubernetes.io/name: "cluster-autoscaler"
   type: "ClusterIP"
 ---
+{{- if (eq GetClusterAutoscalerExpander "priority") }}
```
olemarkus, line 261: I am guessing this will break things for everyone already using priority if they are not also setting the priority values and letting kOps override the CM correctly. There need to be a migration path of some sort for this.

One way of doing it would be to require the value being set on all IGs with autoscaling enabled. That would however break things for those using regex, which means you may need an option for "bring your own CM".

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -308,6 +308,22 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
...
+			priorities := make(map[string][]string)
+			for name, spec := range tf.GetNodeInstanceGroups() {
+
+				if spec.Autoscale != nil {
+					priorities[fmt.Sprint(spec.AutoscalePriority)] = append(priorities[fmt.Sprint(spec.AutoscalePriority)], fmt.Sprintf("%s.%s", name, tf.ClusterName()))
+				}
+			}
+			return priorities
+		}
+		dest["GetClusterAutoscalerExpander"] = func() string {
```
olemarkus, line 322: Why is this function required? Can't you use the value directly?

pkg/model/context.go
```diff
@@ -153,6 +153,16 @@ func (b *KopsModelContext) CloudTagsForInstanceGroup(ig *kops.InstanceGroup) (ma
 		labels[fi.StringValue(nth.ManagedASGTag)] = ""
 	}
 
+	switch b.Cluster.Spec.GetCloudProvider() {
+	case kops.CloudProviderAWS:
+		// Apply labels for cluster autoscaler autodiscovery
+		cas := b.Cluster.Spec.ClusterAutoscaler
+		if cas != nil && fi.BoolValue(cas.Enabled) && fi.BoolValue(cas.AutoDiscovery) && ig.Spec.Autoscale != nil {
```
olemarkus, line 160: This belongs in a separate PR I think. Debatable if this feature is needed at all and what value it would bring.

olemarkus (changes requested):
Thanks for this. Most importantly, the two features should be split into two PRs.
Secondly, there should be an integration test covering this so we know the manifests render as expected.

pkg/model/context.go
chubchubsancho: Hi @olemarkus cluster-autoscaler autodiscovery have been moved https://github.com/kubernetes/kops/pull/14521

upup/pkg/fi/cloudup/template_functions.go
chubchubsancho: Without this function `eq` failed on template rendering with error `error calling eq: invalid type for comparison`

It seems that dereferencing pointers does not work with `eq`

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
chubchubsancho: I'll take a look on way to not break things for users that already use `priority` expander with their own configMap and add the ability to use a custom configMap.
I'm thinking about:
1. adding a new field on cluster autoscaler which will take precedence on `autoscalePriority` field of instance group if defined. 
Like:
```yaml
clusterAutoscaler:
  priorityExpanderConfig: |-
    100:
    - foo.*
    200:
    - bar.foo.*
```
2. Adding a field on cluster autoscaler in order to allow self manage configMap for cluster autoscaler. 
Like:
```yaml
clusterAutoscaler:
  selfManagedPriorityExpanderConfig: true
```

chubchubsancho:
Hi @olemarkus integration tests have been added for this feature

upup/pkg/fi/cloudup/template_functions.go
olemarkus: Hm. Expander shouldn't even be a pointer as `""` (empty string)  isn't a valid value and we always set this one.

Let me do a quick PR for that

docs/addons.md
```diff
@@ -89,7 +89,15 @@ Read more about cluster autoscaler in the [official documentation](https://githu
 ##### Expander strategies
 Cluster autoscaler supports several different [expander strategies](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders).
 
-Note that the `priority` expander requires additional configuration through a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md) - you will need to create this ConfigMap in your cluster before selecting this expander.
+Note that the `priority` expander requires additional configuration through a ConfigMap as described in [its documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md) - this ConfigMap will be generated based on instance group you can change priority of each instance group by adding the followig to instance group spec.
```
olemarkus, line 92: This is missing the configmap creation flag, I think.
Also please add `{{ kops_feature_table(kops_added_default='1.26') }}`

pkg/model/components/clusterautoscaler.go
```diff
@@ -93,6 +93,9 @@ func (b *ClusterAutoscalerOptionsBuilder) BuildOptions(o interface{}) error {
 	if cas.MaxNodeProvisionTime == "" {
 		cas.MaxNodeProvisionTime = "15m0s"
 	}
+	if cas.CreatePriorityExpenderConfig == nil {
```
olemarkus, line 96: I recommend changing this one to `true` by default iff the priority expander is set. If not, always leave it unset. That way this won't cause an unnecessary diff for everyone not using the priority expander.

olemarkus (changes requested):

pkg/apis/kops/componentconfig.go
```diff
@@ -993,7 +993,9 @@ type ClusterAutoscalerConfig struct {
 	// Expander determines the strategy for which instance group gets expanded.
 	// Supported values: least-waste, most-pods, random, price, priority.
 	// The price expander is only supported on GCE.
-	// The priority expander requires additional configuration via a ConfigMap.
+	// The priority expander ConfigMap will be autogenerated based on instance groups.
+	// Instance group will be selected if their autoscale field is set to true.
+	// Priority for each instance group is defined by AutoscalePriority field.
```
olemarkus, line 998:
```suggestion
	// By default, kOps will generate the priority expander ConfigMap based on the `autoscale` and `autoscalePriority` fields in the InstanceGroup specs.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -1040,6 +1042,12 @@ type ClusterAutoscalerConfig struct {
 	// PodAnnotations are the annotations added to cluster autoscaler pods when they are created.
 	// Default: none
 	PodAnnotations map[string]string `json:"podAnnotations,omitempty"`
+	// CreatePriorityExpenderConfig should kOps create priority-expander configmap
```
olemarkus, line 1045:
```suggestion
	// CreatePriorityExpenderConfig makes kOps create the priority-expander ConfigMap.
```

pkg/apis/kops/componentconfig.go
```diff
@@ -1040,6 +1042,12 @@ type ClusterAutoscalerConfig struct {
...
+	// Default: true
+	CreatePriorityExpenderConfig *bool `json:"createPriorityExpanderConfig,omitempty"`
+	// CustomPriorityExpanderConfig provide priority-expander configMap regardless of instance group autoscalePriority
```
olemarkus, line 1048:
```suggestion
	// CustomPriorityExpanderConfig overrides the priority-expander ConfigMap with the provided configuration. Any InstanceGroup configuration will be ignored if this is set.
```

pkg/apis/kops/instancegroup.go
```diff
@@ -104,6 +104,8 @@ type InstanceGroupSpec struct {
 	MaxSize *int32 `json:"maxSize,omitempty"`
 	// Autoscale determines if autoscaling will be enabled for this instance group if cluster autoscaler is enabled
 	Autoscale *bool `json:"autoscale,omitempty"`
+	// AutoscalePriority determines instance group priority for scaling when cluster autoscaler use priority expander
```
olemarkus, line 107:
```suggestion
	// AutoscalePriority determines the InstanceGroup priority for scaling when cluster autoscaler uses the priority expander.
```

olemarkus (changes requested):

olemarkus (commented):
Sorry for taking some time to revisit this one. It looks good now, but it needs rebasing.

chubchubsancho:
> Sorry for taking some time to revisit this one. It looks good now, but it needs rebasing.
Just rebased ;-)

olemarkus (approved):

johngmyers:
@chubchubsancho please run the `hack/update-expected.sh` script and commit the results (as a separate commit).

danports:
Nice work @chubchubsancho, thanks!

What will happen when existing clusters with `expander: priority` upgrade to kOps 1.26 if `createPriorityExpanderConfig: false` hasn't been set on the cluster spec? Will the autoscaler addon overwrite the `cluster-autoscaler-priority-expander` ConfigMap? If so, it might be nice to add a note to the 1.26 release notes about that.

olemarkus:
Added https://github.com/kubernetes/kops/pull/14998

danports:
👍 Thanks @olemarkus!

johngmyers:
Perhaps this should be called `managePriorityExpanderConfig` for consistency with similar options?

PR #14521, opened 2022-11-09
feat(cluster-autoscaler): Add AutoDiscovery in ClusterAutoscalerConfig

chubchubsancho:
Fixes: https://github.com/kubernetes/kops/issues/14518 1.1


olemarkus:
I am missing the motivation for using autodiscovery. What do we achieve that we can't do with the current implementation?

chubchubsancho:
> I am missing the motivation for using autodiscovery. What do we achieve that we can't do with the current implementation?

Beyond reducing the number of arguments passed to the cluster-autoscaler command, this allows when adding a new instance group not to redeploy cluster-autoscaler

olemarkus:
> > I am missing the motivation for using autodiscovery. What do we achieve that we can't do with the current implementation?
> 
> Beyond reducing the number of arguments passed to the cluster-autoscaler command, this allows when adding a new instance group not to redeploy cluster-autoscaler

I don't see either as particularly beneficial. Number of arguments has zero cost. Restarting CAS also has very little cost.

PR #14540, opened 2022-11-13
Bump AWS Load Balancer Controller to 2.4.5

olemarkus:
Due to https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2824 LBC can now run on regular nodes with max hop = 1 if we use the new node name convention. I simplified the checks a bit so we only do this if IRSA is enabled and K8s > 1.24. Technically older K8s versions would work to if external CCM was explicitly enabled, but this would complicate the checks a bit too much.


docs/releases/1.26-NOTES.md
```diff
@@ -8,6 +8,8 @@ This is a document to gather the release notes prior to the release.
 
 * The default networking provider for new clusters is now Cilium.
 
+* AWS Load Balancer Controller will now run on worker nodes if IRSA is enabled and the Kubernetes version is greater than 1.24.
```
johngmyers, line 11: This should be in the AWS Only subsection

johngmyers: Greater than or equal to 1.24

johngmyers: This can be misconstrued as stating that LBC will be automatically enabled when k8s is new enough and IRSA is enabled. I suggest some sort of "if it is enabled" wording.

johngmyers (changes requested):

Issue #14481, opened 2022-10-31
Add Support for AWS G5 Instances


sekasara:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
AWS recently announced the G5 instance type (https://aws.amazon.com/about-aws/whats-new/2021/11/announcing-general-availability-amazon-ec2-g5-instances). 
This is currently not a supported instance type, and it would be useful to be able to use these in our cluster, particularly for GPU-accelerated workloads.

olemarkus:
kOps should support this as long as the given region supports it. Are you seeing any errors when trying to use it?

sekasara:
@olemarkus Can you please help us with which version of kops is supporting G5 instance type ?  
getting below error.  
`spec.machineType: Invalid value: "g5.4xlarge": machine type specified is invalid`

olemarkus:
In which region are you trying to use these?

sekasara:
@olemarkus ap-south-1 region

olemarkus:
Right. I don't think the instance type exists there. Only exists in Oregon, North Virginia, and Ireland.

sekasara:
its available @olemarkus 
<img width="742" alt="image" src="https://user-images.githubusercontent.com/63733959/200287382-bf0e03d0-df15-4d65-b9c9-1e9b17e1b81e.png">
https://aws.amazon.com/ec2/pricing/on-demand/

olemarkus:
The error specifically comes from the ec2 `DescribeInstanceType` API endpoint not returning any values.

MinliangLin:
Hi @olemarkus. Not sure if this could help your debugging, I just tired the `DescribeInstanceType` api. It seems working well on my EC2 machine. The aws cli version I am using is `aws-cli/1.18.107 Python/2.7.18 Linux/4.14.192-147.314.amzn2.x86_64 botocore/1.17.31`

```
➜  ~ aws ec2 describe-instance-types --region ap-south-1 --instance-types g5.xlarge
{
    "InstanceTypes": [
        {
            "FreeTierEligible": false,
            "InstanceStorageSupported": true,
            "Hypervisor": "nitro",
            "PlacementGroupInfo": {
                "SupportedStrategies": [
                    "cluster",
                    "partition",
                    "spread"
                ]
            },
            "CurrentGeneration": true,
            "MemoryInfo": {
                "SizeInMiB": 16384
            },
            "SupportedVirtualizationTypes": [
                "hvm"
[...]

PR #14544, opened 2022-11-14
Clearer error messages around instance types not found

justinsb:
Issue #14481


upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -2387,10 +2387,10 @@ func describeInstanceType(c AWSCloud, instanceType string) (*ec2.InstanceTypeInf
 	}
 	resp, err := c.EC2().DescribeInstanceTypes(req)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error describing instance type %q in region %q: %w", instanceType, c.Region(), err)
```
johngmyers, line 2390: error strings should not start with the word "error" as they are printed with "Error: " prepended.

```suggestion
		return nil, fmt.Errorf("describing instance type %q in region %q: %w", instanceType, c.Region(), err)
```

johngmyers (changes requested):

PR #14545, opened 2022-11-14
Change the control-plane IG role to "ControlPlane" in v1alpha3 API

johngmyers:


tests/integration/update_cluster/additionalobjects/kubernetes.tf
```diff
@@ -636,7 +646,7 @@ resource "aws_s3_object" "manifests-static-kube-apiserver-healthcheck" {
 resource "aws_s3_object" "nodeupconfig-master-us-test-1a" {
   bucket                 = "testingBucket"
   content                = file("${path.module}/data/aws_s3_object_nodeupconfig-master-us-test-1a_content")
-  key                    = "tests/additionalobjects.example.com/igconfig/master/master-us-test-1a/nodeupconfig.yaml"
+  key                    = "tests/additionalobjects.example.com/igconfig/control-plane/master-us-test-1a/nodeupconfig.yaml"
```
rifelpet, line 649: will using `--target direct` cleanup the object at the old location or will it be orphaned? and if orphaned, will `kops delete cluster` delete the orphaned object or fail because it doesn't "recognize" the object key?

pkg/model/components/etcdmanager/model.go
```diff
@@ -383,6 +383,7 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
 			config.VolumeTag = []string{
 				fmt.Sprintf("kubernetes.io/cluster/%s=owned", b.Cluster.Name),
 				awsup.TagNameEtcdClusterPrefix + etcdCluster.Name,
+				awsup.TagNameRolePrefix + "control-plane=1",
```
hakman, line 386: Changes to this file feel incomplete:
1. new `control-plane` role is only added to some cloud providers
2. probably we want to remove the old `master` tag in this case, as `etcd-manager` should only look for the new one.
Any thoughts?

rifelpet: Wouldn't removing the old master tag break kops downgrades?

hakman: This is not about adding / removing the tag. This just generates a list of tags that etcd-manager should match for a volume. By removing "master", it should just ignore the existence of that tag.

tests/integration/update_cluster/additionalobjects/kubernetes.tf
johngmyers: I ran a test. They are orphaned. `kops delete cluster` deletes the orphaned objects anyway.

pkg/model/components/etcdmanager/model.go
johngmyers: I believe additional cloud providers is a separate PR. I only did the ones that were obvious.

If it's the query, I can remove the old tag out of `EtcdManagerBuilder`.

johngmyers: I switched Azure over. For GCE I don't know enough about the tagging restrictions

johngmyers: That would be an excellent topic for a follow-up PR. There are a lot of opportunities in the code base to update the terminology. I was trying to limit this PR to things that derive directly from the IG role, though I unfortunately got some scope creep into etc volumes on providers that don't directly derive from the IG role.

hakman: 👍

hakman (approved):

PR #14547, opened 2022-11-14
option to add extra volume tags with ebs csi driver

heybronson:
Adds `cloudConfig.awsEBSCSIDriver.extraVolumeTags` to add extra AWS tags to volumes provisioned by the EBS CSI Driver.

Changes the `join` template function to use the Sprig-defined function instead of `strings.Join` to get better type handling with the Sprig function. Updates other templates to fit the Sprig `join` expectations.

Testing:
`kops update cluster`:
```
Will modify resources:
  LaunchTemplate/master-us-west-2a.masters.bronson.cluster.doordash.cloud
        UserData
                                ...
                                      extraVolumeTags:
                                        testKey: testValue
                                +       testKey2: testValue2
                                      managed: true
                                      podAnnotations:
                                ...


  LaunchTemplate/nodes-us-west-2a.bronson.cluster.doordash.cloud
        UserData
                                ...
                                      extraVolumeTags:
                                        testKey: testValue
                                +       testKey2: testValue2
                                      managed: true
                                      podAnnotations:
                                ...


  LaunchTemplate/nodes-us-west-2b.bronson.cluster.doordash.cloud
        UserData
                                ...
                                      extraVolumeTags:
                                        testKey: testValue
                                +       testKey2: testValue2
                                      managed: true
                                      podAnnotations:
                                ...


  LaunchTemplate/nodes-us-west-2c.bronson.cluster.doordash.cloud
        UserData
                                ...
                                      extraVolumeTags:
                                        testKey: testValue
                                +       testKey2: testValue2
                                      managed: true
                                      podAnnotations:
                                ...


  ManagedFile/bronson.cluster.doordash.cloud-addons-aws-ebs-csi-driver.addons.k8s.io-k8s-1.17
        Contents
                                ...
                                          - --http-endpoint=0.0.0.0:3301
                                          - --v=5
                                +         - --extra-tags=testKey=testValue,testKey2=testValue2
                                -         - --extra-tags=testKey=testValue
                                          env:
                                          - name: CSI_ENDPOINT
                                ...


  ManagedFile/bronson.cluster.doordash.cloud-addons-bootstrap
        Contents
                                ...
                                    - id: k8s-1.17
                                      manifest: aws-ebs-csi-driver.addons.k8s.io/k8s-1.17.yaml
                                +     manifestHash: f84f76f9db04b6d51d42922182b570343fccf1d83f7a0cf1a257b68626f6776e
[...]

PR #14564, opened 2022-11-15
use sprig join for template functions

heybronson:
Switching template `join` function to Sprig's `join`, which can handle `[]interface{}` inputs instead of `[]string` only. This makes `join` easier to use in templates that define variables within the template itself.

Updates existing uses of `join` to fix the new function signature.

related to: https://github.com/kubernetes/kops/pull/14547


johngmyers (approved):
Thanks!

PR #14572, opened 2022-11-16
Set IMDS max hop limit to 1 for control plane nodes if IRSA is enabled

olemarkus:


docs/releases/1.26-NOTES.md
```diff
@@ -6,6 +6,18 @@ This is a document to gather the release notes prior to the release.
 
 # Significant changes
 
+## AWS meta data service hop limit
+
+As of Kubernetes version 1.26 and with IRSA enabled, control plane nodes will now run with a max hop limit of 1 for the metadata service. This will prevent Pods without host networking from accessing the instance metadata service.
```
johngmyers, line 11: Should be in the "AWS only" section

docs/releases/1.26-NOTES.md
```diff
@@ -6,6 +6,18 @@ This is a document to gather the release notes prior to the release.
...
+
+```
+spec:
+  instanceMetadata:
+    httpTokens: optional
+```
```
johngmyers, line 17: What is this for?

tests/integration/create_cluster/minimal-1.26/expected-v1alpha2.yaml
```diff
@@ -29,6 +29,7 @@ spec:
   iam:
     allowContainerRegistry: true
     legacy: false
+    useServiceAccountExternalPermissions: true
```
johngmyers: This change is unnecessary (and reverted in the subsequent commit)

johngmyers (changes requested):

docs/releases/1.26-NOTES.md
olemarkus: AWS only section is under "other significant changes", but this requires more attention than that. Doesn't the title indicate that it is AWS only?

docs/releases/1.26-NOTES.md
olemarkus: Forgot to fix the content. Done now.

docs/releases/1.26-NOTES.md
```diff
@@ -6,6 +6,20 @@ This is a document to gather the release notes prior to the release.
 
 # Significant changes
 
+## AWS meta data service hop limit
+
+As of Kubernetes version 1.26 and with IRSA enabled, control plane nodes will now run with a max hop limit of 1 for the metadata service. This will prevent Pods without host networking from accessing the instance metadata service.
+
+If you need to be able to allow Pods to assume the control plane instance role, you need to add the following to the InstanceGroup spec:
```
johngmyers, line 13:
```suggestion
If you need to be able to allow non-host-network Pods to assume the control plane instance role, you need to add the following to the InstanceGroup spec:
```

docs/releases/1.26-NOTES.md
johngmyers: Why does it require more attention than that?

olemarkus: It's borderline breaking change if you e.g run `kops create` as part of e2e testing.

johngmyers: To have an adverse impact, one would have to be running one's own workload on the control plane, with non-host-network, not using IRSA, requiring a permission granted to the control plane instance role, not running on worker nodes (because they wouldn't have access to the permission), and the cluster would have to have `useServiceAccountExternalPermissions` enabled.

Otherwise, it's just that the control plane nodes run a little more securely. I don't see this as meeting the bar for a top-line announcement.

olemarkus: Fair enough. Slimmed and moved under AWS

johngmyers (approved):

PR #14577, opened 2022-11-17
Add option for setting Kubernetes feature gates

hakman:
/cc @olemarkus @rifelpet @jsafrane

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kops"
 	api "k8s.io/kops/pkg/apis/kops"
-	kopsapi "k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/apis/kops/model"
 	"k8s.io/kops/pkg/apis/kops/util"
 	"k8s.io/kops/pkg/client/simple"
@@ -416,7 +415,7 @@ func NewCluster(opt *NewClusterOptions, clientset simple.Clientset) (*NewCluster
 				}
 
 			}
-		} else if g.Spec.Role == kopsapi.InstanceGroupRoleBastion {
+		} else if g.Spec.Role == api.InstanceGroupRoleBastion {
 			if g.Spec.MachineType == "" {
 				g.Spec.MachineType, err = defaultMachineType(cloud, &cluster, g)
 				if err != nil {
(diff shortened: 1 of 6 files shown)


hakman:
Managed to test this with multiple k8s feature gates. No issues noticed.
/cc @johngmyers

johngmyers:
@k8s-ci-robot needs-rebase

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -64,6 +63,8 @@ type NewClusterOptions struct {
 	DiscoveryStore string
 	// KubernetesVersion is the version of Kubernetes to deploy. It defaults to the version recommended by the channel.
 	KubernetesVersion string
+	// KubernetesVersion is the version of Kubernetes to deploy. It defaults to the version recommended by the channel.
```
johngmyers, line 66: Incorrect comment

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -233,9 +285,11 @@ func NewCluster(opt *NewClusterOptions, clientset simple.Clientset) (*NewCluster
 	cluster.Spec.IAM = &api.IAMSpec{
 		AllowContainerRegistry: true,
 	}
-	cluster.Spec.Kubelet = &api.KubeletConfigSpec{
-		AnonymousAuth: fi.PtrTo(false),
+
+	if cluster.Spec.Kubelet == nil {
+		cluster.Spec.Kubelet = &api.KubeletConfigSpec{}
```
johngmyers, line 290: It would be simpler to initialize `cluster.Spec.Kubelet` before the iteration over KubernetesFeatureGates

johngmyers (changes requested):
Please fix the comment.

johngmyers:
I wonder if we should put `KubernetesFeatureGates` in the cluster spec. It would be a lot easier to manage changes that way.

upup/pkg/fi/cloudup/new_cluster.go
hakman: Should be simpler now.

upup/pkg/fi/cloudup/new_cluster.go
hakman: Fixed.

hakman:
> I wonder if we should put `KubernetesFeatureGates` in the cluster spec. It would be a lot easier to manage changes that way.

For sure would be easier. Was thinking that it would be nice in v1alpha3 API as it would mean removing individual flags from each component.

johngmyers:
I don't think we can just remove the individual flags in v1alpha3. I was thinking of having the individual flags be merged on top of the cluster-level ones. Or we could have an API validation that says you can't set both.

hakman:
Not sure if that will help that much. I think I am ok with how it is for now.

johngmyers (approved):

PR #14591, opened 2022-11-18
Always run Cluster Autoscaler on the control plane

### LINKED ISSUE #14411: Node affinity for cluster autoscaler
While I fully agree with the changes in https://github.com/kubernetes/kops/pull/13846 there are certain use-cases where being able to configure the `nodeSelector` and `tolerations` for the cluster autoscaler would be ideal.

In my case I manage multiple clusters where, with the exception of the control plane, every other node group is made up from spot instances. On these clusters I wouldn't mind running the autoscaler on those node groups or even on the control plane.

Since the referenced feature every time I update the clusters via `kops` I have to patch the autoscaler's `Deployment` to remove the `affinity` option.

Unfortunately I'm not a Go engineer so I can't make any contributions to the code.

I appreciate your consideration.

olemarkus:
This matches what we do with karpenter and makes safer to run spot-only clusters. Should also avoid situations where small nodes are used and CAS end up not being initially scheduled.

Fixes #14411

### DIFF
--- a/tests/integration/update_cluster/many-addons-ccm-irsa/data/aws_s3_object_minimal.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/many-addons-ccm-irsa/data/aws_s3_object_minimal.example.com-addons-bootstrap_content
@@ -41,7 +41,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.15
     manifest: cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml
-    manifestHash: 47e4e76613b490ee644bbba65371780a4920a76756e02fc5feea4610644c2552
+    manifestHash: 246f758a9ace2054627a38a260dd480439a23bab669895187ca815ddfdaa7f15
     name: cluster-autoscaler.addons.k8s.io
     selector:
       k8s-addon: cluster-autoscaler.addons.k8s.io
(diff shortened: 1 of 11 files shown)


olemarkus:
Was this one discussed in the office hours?

johngmyers:
No, and I need to put together my comments.

johngmyers:
Pro:
* CAS doesn't get much attacker-controlled input, so is less likely to be broken into.
* CP nodes tend to be overprovisioned, so there's room.

Con:
* CP nodes are a high-value target due to the private keys they keep on-disk.
* CAS doesn't have high privileges, so isolating it to CP nodes isn't worth much.
* CAS is unlike Karpenter in that it doesn't have to provision itself. Even if all the nodes it can run on get terminated, the ASG will spin up new ones.
* If an all-spot cluster loses all its spot instances, having CAS running on the CP isn't going to help. When spot capacity is once again available, ASG will spin up new instances and CAS can then run due to its high scheduling priority.

olemarkus:
I think having a proper security context (which, admittedly, we don't right now) would mitigate the risks of CAS on CP. It would not have the privileges to reach a file even if it somehow would get access to the host filesystem. Of course, if the manifest is compromised, and someone would add a host volume, you'd be in trouble, but then, you could just add the selector/toleration to run on the CP as well.

On the last point, if you have multiple IGs (say one per AZ) and you lose capacity and CAS in one IG, CAS on the CP would be able to scale up the other IGs.

johngmyers:
Pushing upstreams to configure better security contexts would be a good project. But the control plane is simpler to reason about with fewer things running on it. There is a risk that a container escape could also bypass the security context.

If you lose capacity and CAS in one AZ, it will be able to reschedule into another AZ.

PR #14592, opened 2022-11-18
Use generics to refactor fi.Bool/Float*/Int*/String*()

hakman:


johngmyers:
I'm not sure about `NumberValue()`. Go can't have a `Value()`?

One would think something like `PtrTo()` would make its way into the standard libraries.

upup/pkg/fi/values.go
```diff
@@ -21,8 +21,16 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+
+	"golang.org/x/exp/constraints"
 )
 
+// PtrTo returns a pointer to any value
```
johngmyers, line 28:
```suggestion
// PtrTo returns a pointer to a copy of any value
```

johngmyers (changes requested):

hakman:
> I'm not sure about `NumberValue()`. Go can't have a `Value()`?

No idea, was just something fun to do. Seemed more "number" like, as we still have `StringValue()` and `BoolValue()`.

> One would think something like `PtrTo()` would make its way into the standard libraries.

That would be nice.

johngmyers (approved):

PR #14593, opened 2022-11-18
WIP Allow cluster-autoscaler to run on spot if nothing else is available

### LINKED ISSUE #14411: Node affinity for cluster autoscaler

johngmyers:
An alternative to #14591 

Fixes #14411

### DIFF
--- a/tests/integration/update_cluster/many-addons-ccm-irsa/data/aws_s3_object_minimal.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/many-addons-ccm-irsa/data/aws_s3_object_minimal.example.com-addons-bootstrap_content
@@ -41,7 +41,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.15
     manifest: cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml
-    manifestHash: 47e4e76613b490ee644bbba65371780a4920a76756e02fc5feea4610644c2552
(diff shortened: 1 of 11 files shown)


olemarkus:
This is not a solution for spot-only clusters, so I don't think it safely fixes https://github.com/kubernetes/kops/issues/14411

johngmyers:
@olemarkus could you expand on why this isn't a solution for spot-only clusters?

olemarkus (approved):
For spot-only, the only really safe place to put CAS would be on the control plane. However, using NTH in addition would probably mitigate this to the extent it is a very unlikely scenario.

If you have many IGs with very narrow mix of insistence types, you can still run into this though, but CAS trying to scale up an IG that has no spot capacity, and then get evicted before it has the chance to jump to the next one.

But yeah, this all may be theoretical since we also run two CAS instances, and the likelyhood of both getting reaped is in itself fairly low.

johngmyers:
Cluster Autoscaler is `system-cluster-critical`, so there's a good chance of it rescheduling if it gets evicted.

PR #14594, opened 2022-11-18
skip Loadbalancer UDP tests

aojea:
All the loadbalancer tests in the kubernetes e2e suite fail on IPv6, however, they were skipped until now because they were tagged as [Slow]. Antonio messed it up adding a new loadbalancer test without  the Slow tag and now it fails trying to run on IPv6.
Skip these tests temporary since they fail always on IPv6

### DIFF
--- a/tests/e2e/pkg/tester/skip_regex.go
+++ b/tests/e2e/pkg/tester/skip_regex.go
@@ -45,6 +45,13 @@ func (t *Tester) setSkipRegexFlag() error {
 
 	skipRegex := skipRegexBase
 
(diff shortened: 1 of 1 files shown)


aojea:
/assign @justinsb 

Sorry for the inconvenience
https://github.com/kubernetes/kubernetes/issues/113964

johngmyers:
The skip won't be temporary since we'll have to keep it for as long as kOps supports k8s 1.26.

But I suppose if we do start running Slow tests we could add a k8s minor-version check at that time.

johngmyers (approved):

PR #14601, opened 2022-11-19
Bump karpenter to 0.19.x

olemarkus:
Significant changes to the templates related to webhook and controller being compiled into a single binary.

Note that this is currently broken due to https://github.com/aws/karpenter/issues/2892


noony:
@olemarkus it's maybe fixed by this pr : https://github.com/aws/karpenter/pull/3146 ?

noony:
We can think moving on with last version + aws sqs queue for native support of interruptions.

olemarkus:
Yeah sounds good. Not sure it makes sense for us to support the "native" SQS mode since it is just bundling the NTH controller and we support easily installing that as a separate addon.

PR #14611, opened 2022-11-21
WIP toolbox-dump: add support for --bastion-ssh-user flag

johngmyers:
Debian 11 doesn't forward authentication by default, so those clusters will need to use a bastion on Ubuntu or some other distro. But then the bastion will need to use a different SSH user than the cluster nodes.


johngmyers:
Ubuntu bastion doesn't work either,

PR #14621, opened 2022-11-22
Implement subnet "/64#N" notation in Terraform

### LINKED ISSUE #8432: IPv6 support

johngmyers:
Closes #8432


johngmyers:
@rifelpet I'm having to really abuse the `terraformWriter` interface because all the `Literal` constructors take arguments of type `string` instead of `Literal`.

johngmyers:
@rifelpet I'm not quite sure how to address this PR's abuse of the `Literal` interface. It seems `Literal` is going through a lot of complexity in order to use HCL's constructs to generate the output. I'm not sure why this is necessary, especially since it also generates the output itself in order to serialize to JSON for unit tests.

upup/pkg/fi/cloudup/awstasks/vpc.go
```diff
@@ -296,6 +296,10 @@ func (_ *VPC) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *VPC)
 		return err
 	}
 
+	if err := t.AddOutputVariable("vpc_ipv6_cidr_length", terraformWriter.LiteralFunctionExpression("local.vpc_ipv6_cidr_block == null ? null : tonumber", "regex(\".*/(\\\\d+)\", local.vpc_ipv6_cidr_block)[0]")); err != nil {
```
rifelpet, line 299:
```suggestion
	if err := t.AddOutputVariable("vpc_ipv6_cidr_length", terraformWriter.LiteralFunctionExpression("local.vpc_ipv6_cidr_length")); err != nil {
```

rifelpet (commented):
Looking at [how we interact with hclwrite](https://github.com/kubernetes/kops/blob/55b5e998557e6ba230dc65a24b64db9027416d15/upup/pkg/fi/cloudup/terraform/hcl2.go#L111), we might be able to have the Literal struct just contain a `hclwrite.Tokens` field and have all of the constructors build the tokens themselves. That may allow for more flexibility in those constructors without all the complexities of their string arguments.

upup/pkg/fi/cloudup/awstasks/vpc.go
johngmyers: This `AddOutputVariable` call is what defines `local.vpc_ipv6_cidr_length`. You can't define a local variable as being itself.

johngmyers:
I would prefer we refactor the `Literal` constructors as a followup PR.

rifelpet (approved):

Issue #14575, opened 2022-11-17
[cilium - eni IPAM MODE]  problem with connection from POD to external resource in additional subnet after kops upgrade from 1.23 -> 1.25 


glomikuz:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Version 1.23.2 (git-4125bbbe975ca104c609ceaa625a9a5ff3ac19f4) → Client version: 1.25.2 (git-v1.25.2)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

1.23.8

Subnets:
  - cidr: 10.210.140.0/26
  - cidr: 10.210.140.64/26
  - cidr: 10.210.0.0/22

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

kops1.25 replace -f value.yaml
kops1.25 update cluster --name ${CLUSTER_NAME} --yes

**5. What happened after the commands executed?**

https://gist.github.com/glomikuz/5786db5989fa714cf80dd3642b068ee8
Cluster changes have been applied to the cloud.

**6. What did you expect to happen?**

Pods with IPs from my new subnet(  - cidr: 10.210.0.0/22 ) will able connect to external resources

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  name: k8s.tmp-test.test
spec:
  api:
    loadBalancer:
      type: Internal
      useForInternalApi: true
  authorization:
    rbac: {}
  cloudConfig:
    awsEBSCSIDriver:
      enabled: true
  cloudProvider: aws
  clusterAutoscaler:
    enabled: true
  configBase: s3://states/k8s.tmp-test.test
  containerRuntime: containerd
  docker:
    skipInstall: true
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-1a
      name: a
    manager:
      env:
      - name: ETCD_LISTEN_METRICS_URLS
        value: http://0.0.0.0:2379
      - name: ETCD_METRICS
        value: basic
      - name: ETCD_MANAGER_HOURLY_BACKUPS_RETENTION
        value: 1d
[...]

PR #14622, opened 2022-11-22
Add networking flag option for cilium in eni mode

olemarkus:
Ref #14575


cmd/kops/integration_test.go
```diff
@@ -516,6 +516,13 @@ func TestPrivateCiliumAdvanced(t *testing.T) {
 		runTestTerraformAWS(t)
 }
 
+func TestPrivateCiliumEni(t *testing.T) {
+	newIntegrationTest("privatecilium.example.com", "privateciliumeni").
```
johngmyers, line 520: perhaps a `-` before `eni` in the direectory name?

johngmyers:
This would be easier to review if the update-expected.sh changes were in a separate commit.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1141,6 +1141,8 @@ func setupNetworking(opt *NewClusterOptions, cluster *api.Cluster) error {
 	case "cilium-etcd":
 		addCiliumNetwork(cluster)
 		cluster.Spec.Networking.Cilium.EtcdManaged = true
+	case "cilium-eni":
+		addCiliumNetwork(cluster)
```
johngmyers, line 1145: It doesn't look like this is actually enabling ENI mode.

johngmyers:
I think we should make sure that `--networking=cilium-eni --ipv6` works and behaves the same as `--networking=cilium --ipv6`.

upup/pkg/fi/cloudup/new_cluster.go
olemarkus: Heh. I missed a commit as I wanted the test diff to be more visible. Added now.

olemarkus:
> I think we should make sure that `--networking=cilium-eni --ipv6` works and behaves the same as `--networking=cilium --ipv6`.

We don't support eni mode for ipv6. That's already blocked by validation. I assume there would be a difference in behavior between the two: that cilium-operator does the prefix delegation and node IPAM rather than kops-controller. This support wasn't added until 1.12 though, I believe.

cmd/kops/integration_test.go
```diff
@@ -510,6 +510,13 @@ func TestPrivateCiliumAdvanced(t *testing.T) {
 		runTestTerraformAWS(t)
 }
 
+func TestPrivateCiliumEni(t *testing.T) {
```
johngmyers, line 513:
```suggestion
func TestPrivateCiliumENI(t *testing.T) {
```

hakman (approved):

Issue #14520, opened 2022-11-09
client-go applier fails to upgrade resources


olemarkus:
The addon upgrade test has been failing since we migrated to the client-go-based applier in #14362: https://testgrid.k8s.io/kops-misc#kops-aws-addon-resource-tracking

This needs to be looked at or we need to prepare for flipping back to kubectl for another release.

/cc @justinsb

justinsb:
Thanks for calling it out - I'll take a look!

justinsb:
Looks like two separate problems:

1) Older kOps versions can't delete newer clusters, so once we leak one we "get stuck", because we clean up before we create a cluster:

`error removing cluster from state store: refusing to delete: unknown file found: s3://k8s-kops-prow/e2e-c0d41e2af2-13250.test-cncf-aws.k8s.io/cluster-completed.spec`

2) There might be a problem with a PodDisruptionBudget for kube-dns:

W1102 21:08:44.774906    6381 results.go:63] error from apply on policy/v1beta1, Kind=PodDisruptionBudget kube-system/kube-dns: error from apply: PodDisruptionBudget.policy "kube-dns" is invalid: spec: Invalid value: policy.PodDisruptionBudgetSpec{MinAvailable:(*intstr.IntOrString)(0xc0077a6e40), Selector:(*v1.LabelSelector)(0xc0077a6e60), MaxUnavailable:(*intstr.IntOrString)(0xc0077a6e80)}: minAvailable and maxUnavailable cannot be both set

justinsb:
For 2, I think the problem is that we're mixing old and new objects (we switched from minAvailable to maxUnavailable in https://github.com/kubernetes/kops/pull/12915 . I'm thinking what we need to do is to switch the field manager from kubectl-applier to our SSA owner _before_ we apply an addon object.  I know you had a WIP PR here @olemarkus , but I can't find it...?   However, I do think doing it _before_ might actually be (1) easier and (2) necessary because of this scenario - even if it means we probably always have to do a GET on each object before applying.

justinsb:
Actually I've found the logic to update the field manager, nicely merged into applyset.go.  It may be that we just need to cherry-pick that logic; I'm writing a test to be sure!

PR #14624, opened 2022-11-22
kubetest: dump poddisruptionbudgets and fieldManagers

justinsb:
Should help diagnose what's going on in some of our complex upgrade
tests when we detect conflicts.

Issue #14520


justinsb:
The test which should be doing the same thing as the failing upgrade tests is not showing a problem (#14623).

Adding more information to our dump so we can better understand what's going on here.

olemarkus (approved):

justinsb:
Thanks @olemarkus - I think I found the fix as well - I added the original `kubectl apply -f` applier to the tests and was able to get a failure, fix should be #14626

justinsb:
(Although we still want this PR, because the [test](https://testgrid.k8s.io/kops-misc#kops-aws-addon-resource-tracking) just passed but I'm not sure why...)

PR #14629, opened 2022-11-22
Scaleway integration

Mia-Cross:
This is the latest version of my work for the integration of Scaleway in kOps.
Commits are squashed in this version for better readability but the original history is still available in the branch scw_init that I left as it were before the big rebase I did after the merge of #14431 in upstream.

EDIT: I meant to open this PR on my own fork and not here as it's way too big, I'll keep making smaller PRs instead. Next up will be the nodeidentity part and the listing of resources for deletion, and then the DNS after that.


cmd/kops/create_cluster.go
```diff
@@ -727,6 +727,17 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
...
+					sshPublicKeyPath = "~/.ssh/id_ed25519.pub"
+					c.SSHPublicKeys, err = loadSSHPublicKeys(sshPublicKeyPath)
+					if err != nil {
+						// Don't wrap file-not-found
+						if os.IsNotExist(err) {
+							klog.V(2).Infof("ssh key not found at %s", sshPublicKeyPath)
+						} else {
+							return fmt.Errorf("error reading SSH key file %q: %v", sshPublicKeyPath, err)
+						}
+					}
```
hakman, line 740: This should be a separate PR.
Please put both keys in an array and loop over it. Stop at the first found.

Mia-Cross: I opened a new PR #14638 for theses lines alone.

PR #14631, opened 2022-11-22
Write Terraform vpc_cidr_block output for shared VPCs

johngmyers:
Will need this for IPv6 `/64#N` support.


johngmyers:
Fixing mock implementation bug in #14634 
/retest

rifelpet:
should the terraform output name suggest that it is the ipv4 cidr block? `vpc_ipv4_cidr_block` or something along those lines

johngmyers:
@rifelpet I was sticking with Terraform's terminology.

johngmyers:
Plus, that was the name we were using for owned VPCs.

rifelpet (approved):

PR #14635, opened 2022-11-23
Set AssignIPv6AddressOnCreation on subnets

johngmyers:


olemarkus:
Can we add a description to this PR? What's the motivation for this change? Don't we already have something for instance ENIs through the launch templates?

johngmyers:
I added a description.

olemarkus (approved):
:ok_hand: 
Thanks a lot

PR #14638, opened 2022-11-23
Add id_ed25519 to auto-loaded ssh keys

Mia-Cross:
In the cmd/kops/create_cluster.go:RunCreateCluster function, the auto-loading of ssh keys only looks for keys at `~/.ssh/id_rsa.pub` so we could add another path to also take ed25519 keys into account.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -31,6 +31,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"go.uber.org/multierr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -734,16 +735,26 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 		}
 
 		if autoloadSSHPublicKeys {
-			// Load from default location, if found
-			sshPublicKeyPath := "~/.ssh/id_rsa.pub"
-			c.SSHPublicKeys, err = loadSSHPublicKeys(sshPublicKeyPath)
-			if err != nil {
+			// Load from default locations, if found
+			sshPublicKeyPaths := []string{
+				"~/.ssh/id_rsa.pub",
+				"~/.ssh/id_ed25519.pub",
+			}
+			var merr error
+			for _, sshPublicKeyPath := range sshPublicKeyPaths {
+				c.SSHPublicKeys, err = loadSSHPublicKeys(sshPublicKeyPath)
+				if err == nil {
+					break
+				}
 				// Don't wrap file-not-found
 				if os.IsNotExist(err) {
 					klog.V(2).Infof("ssh key not found at %s", sshPublicKeyPath)
 				} else {
-					return fmt.Errorf("error reading SSH key file %q: %v", sshPublicKeyPath, err)
+					merr = multierr.Append(merr, err)
 				}
+			}
+			if merr != nil && len(c.SSHPublicKeys) == 0 {
+				return fmt.Errorf("error reading SSH public key files %q: %v", sshPublicKeyPaths, merr)
 			}
 		}
 	}


cmd/kops/create_cluster.go
```diff
@@ -734,10 +734,16 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
...
+				"~/.ssh/id_ed25519.pub",
+			}
+			for _, sshPublicKeyPath := range sshPublicKeyPaths {
+				c.SSHPublicKeys, err = loadSSHPublicKeys(sshPublicKeyPath)
+				if err == nil {
+					break
+				}
 				// Don't wrap file-not-found
 				if os.IsNotExist(err) {
 					klog.V(2).Infof("ssh key not found at %s", sshPublicKeyPath)
```
hakman, line 749: What do you think about using `multierr` in this case?
```suggestion
			var merr error
			for _, sshPublicKeyPath := range sshPublicKeyPaths {
				c.SSHPublicKeys, err = loadSSHPublicKeys(sshPublicKeyPath)
				if err == nil {
					break
				}
				// Don't wrap file-not-found
				if os.IsNotExist(err) {
					klog.V(2).Infof("ssh key not found at %s", sshPublicKeyPath)
				} else {
					merr = multierr.Append(merr, err)
				}
			}
			if err != nil && len(c.SSHPublicKeys) == 0 {
				return nil, fmt.Errorf("error reading SSH key file %q: %v", sshPublicKeyPath, merr))
			}
```

Mia-Cross: I didn't know `multierr`, I followed your suggestion but modified it a little because sshPublicKeyPath was out of scope at the end of the loop and the function returns only an error.

hakman: It will be easier with Go v1.20:
https://github.com/golang/go/issues/53435

cmd/kops/create_cluster.go
```diff
@@ -734,17 +735,27 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
...
 				} else {
-					return fmt.Errorf("error reading SSH key file %q: %v", sshPublicKeyPath, err)
+					merr = multierr.Append(merr, err)
 				}
 			}
+			if err != nil && len(c.SSHPublicKeys) == 0 {
```
hakman, line 756: My bad... :)
```suggestion
			if merr != nil && len(c.SSHPublicKeys) == 0 {
```

cmd/kops/create_cluster.go
```diff
@@ -734,17 +735,27 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
...
+				return fmt.Errorf("error reading SSH key file %q: %v", sshPublicKeyPaths, merr)
```
hakman, line 757:
```suggestion
				return fmt.Errorf("error reading SSH public key files %q: %v", sshPublicKeyPaths, merr)
```

hakman:
I think this is good to merge once commits are squashed.
/lgtm

cmd/kops/create_cluster.go
```diff
@@ -734,17 +735,27 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
...
 				if os.IsNotExist(err) {
 					klog.V(2).Infof("ssh key not found at %s", sshPublicKeyPath)
 				} else {
-					return fmt.Errorf("error reading SSH key file %q: %v", sshPublicKeyPath, err)
+					merr = multierr.Append(merr, err)
 				}
 			}
+			if merr != nil && len(c.SSHPublicKeys) == 0 {
+				return fmt.Errorf("error reading SSH public key files %q: %v", sshPublicKeyPaths, merr)
+			}
```
johngmyers, line 758: The new code will incorrectly fall through if `os.IsNotExist(err)` on both files.

johngmyers (changes requested):

cmd/kops/create_cluster.go
```diff
@@ -734,17 +735,27 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 		}
 
 		if autoloadSSHPublicKeys {
-			// Load from default location, if found
-			sshPublicKeyPath := "~/.ssh/id_rsa.pub"
-			c.SSHPublicKeys, err = loadSSHPublicKeys(sshPublicKeyPath)
-			if err != nil {
+			// Load from default locations, if found
+			sshPublicKeyPaths := []string{
+				"~/.ssh/id_rsa.pub",
```
olemarkus, line 740: Should we flip the order of these two?

hakman: Let's do that separately and add a release note.

hakman (approved):

Issue #14639, opened 2022-11-23
Retests not always running latest code


justinsb:
I think there is an error where we don't always run the latest code in our e2e tests.  I think the issue is this:

```
gsutil -h "Cache-Control:private, max-age=0, no-transform" -m cp -n -r /home/prow/go/src/k8s.io/kops/.build/upload/kops/* gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160
```

I think that behaves differently when the target directory exists vs does not exist:

```
> gsutil ls  gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160/
gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160/latest-ci.txt
gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160/1.26.0-alpha.2+v1.26.0-alpha.1-464-gcf3b5b5c60/
gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160/images/
gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160/linux/
```

```
> gsutil ls gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160/1.26.0-alpha.2+v1.26.0-alpha.1-464-gcf3b5b5c60/
gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160/1.26.0-alpha.2+v1.26.0-alpha.1-464-gcf3b5b5c60/images/
gs://kops-ci/pulls/pull-kops-e2e-k8s-aws-calico/pull-7a4070bda39ac25cbd484aeb4218001a36d08160/1.26.0-alpha.2+v1.26.0-alpha.1-464-gcf3b5b5c60/linux/
```

I think we always want the versioned copy (and need to figure out how to get the correct KOPS_BASE_URL)

PR #14640, opened 2022-11-23
e2e: always upload to a versioned directory

justinsb:
Otherwise on our first upload we were truncating one level of the
upload path.

Issue #14639


justinsb:
/wip while we make sure the e2e does what I think it should!

justinsb:

It's passing, but the whole point is that it should pass again :-)

Issue #14140, opened 2022-08-18
Ubuntu 2204 tracker


olemarkus:
Various things we need to cross of before defaulting to 2204

* [ ] Cilium tests are failing
* [ ] https://github.com/kubernetes/kops/issues/14139
* [ ] Are things ready for cgroupv2?

zetaab:
cgroupv2 is worth of mentioning in release notes at least

hakman:
With k8s 1.25, support for cgroups v2 [graduates to stable](https://kubernetes.io/blog/2022/08/23/kubernetes-v1-25-release/#support-for-cgroups-v2-graduates-to-stable).

olemarkus:
Was lucky enough to catch one of the weekly cilium u2204 tests passing this night:
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-grid-cilium-etcd-u2204-k22-docker

danports:
Last open issue in the checklist here was closed (well, rotted away) - perhaps I'll give 22.04 a try and open issues if I run into any problems with it. Are there any other known issues not already noted here?

danports:
FWIW, I haven't had any issues running on 22.04 in the last two weeks, though I haven't tested the specific scenarios that were raised in issues here, like dind.

hakman:
@danports Ubuntu 22.04 will be the default in kOps 1.27.

danports:
I've been running Ubuntu 22.04 with every version of kOps since 1.26.3. kOps 1.23 is over a year old now, so I'd recommend upgrading to the latest version if possible and seeing if the issue still persists.

PR #14641, opened 2022-11-23
Scaleway create cluster

Mia-Cross:
This PR adds almost everything we need for the command kops create cluster to make it until the end and launch instances. To do that we added Scaleway to all cloud provider switches and added the model and task for the etcd volumes.

The load-balancer part is not implemented yet so at the very end the GetApiIngressStatus function just returns nil, it will be implemented in a later PR.


upup/pkg/fi/cloudup/populate_instancegroup_spec.go
```diff
@@ -49,8 +50,10 @@ const (
 	defaultMasterMachineTypeDO      = "s-2vcpu-4gb"
 	defaultMasterMachineTypeAzure   = "Standard_B2s"
 	defaultMasterMachineTypeHetzner = "cx21"
+	defaultMasterMachineTypeSCW     = "DEV1-M"
 
-	defaultDONodeImage = "ubuntu-20-04-x64"
+	defaultDONodeImage  = "ubuntu-20-04-x64"
+	defaultSCWNodeImage = "debian_buster"
```
remyleone, line 56: Why not the latest ubuntu? or the latest debian?

upup/pkg/fi/cloudup/scalewaytasks/volume.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	for _, volume := range volumes.Volumes {
+		if volume.Name == fi.ValueOf(v.Name) {
+			return &Volume{
+				Name:      fi.PtrTo(volume.Name),
+				ID:        fi.PtrTo(volume.ID),
+				Lifecycle: v.Lifecycle,
+				Size:      fi.PtrTo(int64(volume.Size)),
+				Zone:      fi.PtrTo(string(volume.Zone)),
```
remyleone, line 63: Could you also return the type of volume (local or block)?

upup/pkg/fi/cloudup/scalewaytasks/volume.go
```diff
@@ -0,0 +1,116 @@
...
+	// Scaleway does not support changes to volumes for the moment
+	if a != nil {
+		return nil
+	}
+
+	instanceService := t.Cloud.InstanceService()
+	_, err := instanceService.CreateVolume(&instance.CreateVolumeRequest{
+		Zone:       scw.Zone(fi.ValueOf(e.Zone)),
+		Name:       fi.ValueOf(e.Name),
+		VolumeType: "b_ssd",
```
remyleone, line 110: Could you use `VolumeVolumeTypeBSSD`?

upup/pkg/fi/cloudup/populate_instancegroup_spec.go
hakman: We prefer Ubuntu 20.04 for now:
https://github.com/kubernetes/kops/issues/14140

Mia-Cross:
Note : we need to wait for the next release of the etcd-manager with Scaleway support for it to work properly.

pkg/model/components/etcdmanager/model.go
```diff
@@ -440,6 +441,16 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
...
 
+		case kops.CloudProviderScaleway:
+			config.VolumeProvider = "scaleway"
+
+			config.VolumeTag = []string{
+				fmt.Sprintf("%s=%s", scaleway.TagClusterName, b.Cluster.Name),
+				fmt.Sprintf("%s=%s", scaleway.TagNameEtcdClusterPrefix, etcdCluster.Name),
+				fmt.Sprintf("%s=%s", scaleway.TagNameRolePrefix, scaleway.TagRoleMaster),
+			}
+			config.VolumeNameTag = fmt.Sprintf("%s=%s-%s", scaleway.TagInstanceGroup, instanceGroupName, etcdCluster.Name)
```
hakman, line 452: Why does it have to be IG with Cluster prefix (%s=%s-%s"). There should be no need for the cluster part, as there is already a cluster tag.

Mia-Cross: If the `FindVolumes()` function finds more than 1 volume, the first one of the list gets mounted (either `main` or `events`) but not the other one, therefore the etcd-manager pod which volume is not mounted can't work properly. The only way I managed to prevent this is to make sure that the name tag is unique so that the list returned by `FindVolumes` only contains one element for each pod running the function.

hakman: Could you check https://github.com/kubernetes-sigs/etcdadm/pull/334? It is an example on how to deal with this issue. This approach combines creates 2 sets of tags:
* matchPeerTags = VolumeTags (used to find all peers of same type)
* matchNameTags = VolumeTags + VolumeNameTag (used to find the exact volume that needs mounting)

Cluster an EtcdClusterPrefix should be enough to match all peers. Adding the InstanceGroup tag should make it unique.

Mia-Cross: I reverted my latest changes on the name tag, now it's only the instance group name so we should be good to go for this one too !

hakman: Perfect. Thanks!

pkg/model/master_volumes.go
```diff
@@ -392,3 +396,25 @@ func (b *MasterVolumeBuilder) addAzureVolume(
 
 	return nil
 }
+
+func (b *MasterVolumeBuilder) addScalewayVolume(c *fi.ModelBuilderContext, name string, volumeSize int32, zone string, etcd kops.EtcdClusterSpec, m kops.EtcdMemberSpec, allMembers []string) {
+	tags := []string{
+		fmt.Sprintf("%s=%s", scaleway.TagClusterName, b.Cluster.ObjectMeta.Name),
+		fmt.Sprintf("%s=%s", scaleway.TagNameEtcdClusterPrefix, etcd.Name),
+		fmt.Sprintf("%s=%s", scaleway.TagNameRolePrefix, scaleway.TagRoleMaster),
+		fmt.Sprintf("%s=%s", scaleway.TagRoleVolume, etcd.Name),
```
hakman, line 405: Not used.
```suggestion
```

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -729,6 +731,11 @@ func setupZones(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.Stri
 				return nil, err
 			}
 		}
+
+	case api.CloudProviderScaleway:
+		if len(opt.Zones) > 1 {
+			return nil, fmt.Errorf("scaleway cloud provider currently supports only one zone (location)")
+		}
```
hakman, line 738: Not sure the "location" terminology is correct here or if you want clusters to only allow 1 AZ (zone).
Should this work for Scaleway?
```
kops create cluster --zones=fr-par-1,fr-par-2,fr-par-3
```

upup/pkg/fi/cloudup/scaleway/cloud.go
```diff
@@ -224,7 +227,7 @@ func (s *scwCloudImplementation) FindVPCInfo(id string) (*fi.VPCInfo, error) {
 
 func (s *scwCloudImplementation) GetApiIngressStatus(cluster *kops.Cluster) ([]fi.ApiIngressStatus, error) {
 	klog.V(8).Info("Scaleway doesn't have load-balancers yet so GetApiIngressStatus is not implemented")
```
hakman, line 229: Does this mean that Scaleway doesn't have LBs at all, or that this was not yet implemented for kOps?

upup/pkg/fi/cloudup/scalewaytasks/instance.go
```diff
@@ -75,6 +75,37 @@ func (s *Instance) Run(c *fi.Context) error {
 	return fi.DefaultDeltaRunMethod(s, c)
```
hakman, line 75: Why is the file called "instance"? I thought the Scaleway terminology is "server"?

upup/pkg/fi/cloudup/scalewaytasks/volume.go
```diff
@@ -0,0 +1,118 @@
...
+			return fi.RequiredField("Zone")
+		}
+	}
+	return nil
+}
+
+func (_ *Volume) RenderScw(t *scaleway.ScwAPITarget, a, e, changes *Volume) error {
+	// Scaleway does not support changes to volumes for the moment
+	if a != nil {
+		return nil
```
hakman, line 105: Maybe add a warning before exiting that volumes cannot be modified?

upup/pkg/fi/cloudup/scalewaytasks/sshkey.go
```diff
@@ -108,36 +108,35 @@ func (s *SSHKey) CheckChanges(actual, expected, changes *SSHKey) error {
 }
 
 func (*SSHKey) RenderScw(c *fi.Context, actual, expected, changes *SSHKey) error {
-	cloud := c.Cloud.(scaleway.ScwCloud)
-
-	if actual == nil {
+	// Scaleway does not support changes to ssh keys for the moment
+	if actual != nil {
+		return nil
```
hakman, line 113: Maybe add a warning before exiting that ssh keys cannot be modified?

hakman (changes requested):
Looks mostly good, few nits to address.

upup/pkg/fi/cloudup/scalewaytasks/instance.go
Mia-Cross: Technically you're right, "instance" is the name of our virtual machine product so it is related to servers, volumes, images and snapshots but we use interchangeably the terms server and instance. Would you prefer the term server ?

upup/pkg/fi/cloudup/scaleway/cloud.go
Mia-Cross: Load-balancers will be implemented in an other PR but they are ready. I reworded this message too.

upup/pkg/fi/cloudup/scalewaytasks/instance.go
hakman: I prefer which one you think it's more appropriate.

upup/pkg/fi/cloudup/scaleway/cloud.go
hakman: Perfect.

hakman (approved):
Cool. Thanks @Mia-Cross!

upup/pkg/fi/cloudup/scalewaytasks/instance.go
johngmyers: ...but please consistently use the same term for the same concept.

PR #14642, opened 2022-11-23
fix: use viper to get configuration strings

remyleone:
I would like to use a configuration file stored in `~/.kops.yaml` to store my credentials and many configuration preferences that will remain constant across several cluster creations. Therefore, I think it would be useful to use the helper `viper.GetString` to get information from the configuration file and in case they are not defined there, it will be go to the environment variables.

I guess we would need to do that across all Kops. At the moment many configuration options are not working for me when they are defined in the configuration file and not with environment variables without the use of `viper.GetString()`.

### DIFF
--- a/pkg/featureflag/featureflag.go
+++ b/pkg/featureflag/featureflag.go
@@ -26,10 +26,10 @@ package featureflag
 
 import (
 	"fmt"
-	"os"
 	"strings"
 	"sync"
 
+	"github.com/spf13/viper"
 	"k8s.io/klog/v2"
 )
 
@@ -39,7 +39,8 @@ const (
 )
 
 func init() {
-	ParseFlags(os.Getenv(Name))
+	_ = viper.BindEnv(Name)
+	ParseFlags(viper.GetString(Name))
 }
 
 var (
--- a/util/pkg/vfs/s3context.go
+++ b/util/pkg/vfs/s3context.go
@@ -33,6 +33,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/spf13/viper"
 	"k8s.io/klog/v2"
 )
 
@@ -82,7 +83,8 @@ func (s *S3Context) getClient(region string) (*s3.S3, error) {
 	if s3Client == nil {
 		var config *aws.Config
 		var err error
-		endpoint := os.Getenv("S3_ENDPOINT")
+		_ = viper.BindEnv("S3_ENDPOINT")
+		endpoint := viper.GetString("S3_ENDPOINT")
 		if endpoint == "" {
 			config = aws.NewConfig().WithRegion(region).WithUseDualStack(true)
 			config = config.WithCredentialsChainVerboseErrors(true)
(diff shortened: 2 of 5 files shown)


hakman:
Running `make gomod` should fix the missing deps.

remyleone:
@hakman I'm not sure about how to mock credentials in the context of this test

remyleone:
I've tried to do a search and replace across the whole code base and it seems to work except for this test

johngmyers:
Viper isn't seeing the change to the environment variable that is done in the test setup. The test setup probably has to be changed to push the value into viper (and clear it aterwards).

util/pkg/vfs/tests/s3fs_test.go
```diff
@@ -51,9 +53,8 @@ func TestS3RenderTerraform(t *testing.T) {
 			`,
 		},
 	}
-	t.Setenv("S3_ENDPOINT", "foo.s3.amazonaws.com")
-
-	t.Setenv("KOPS_STATE_S3_ACL", "bucket-owner-full-control")
+	viper.Set("S3_ENDPOINT", "foo.s3.amazonaws.com")
+	viper.Set("KOPS_STATE_S3_ACL", "bucket-owner-full-control")
```
hakman, line 57: This is not really required. The problem is mainly that `viper.BindEnv("ENV_VAR")` has to be called before using `viper.GetString("ENV_VAR")`.

util/pkg/vfs/s3context.go
```diff
@@ -146,10 +148,10 @@ func (s *S3Context) getDetailsForBucket(bucket string) (*S3BucketDetails, error)
 	}
 
 	// Probe to find correct region for bucket
-	endpoint := os.Getenv("S3_ENDPOINT")
+	endpoint := viper.GetString("S3_ENDPOINT")
```
hakman, line 151: Calling `viper.BindEnv()` should make env reading possible. A little annoying, but not sure how to go about it differently at the moment.
```suggestion
	_ = viper.BindEnv("S3_ENDPOINT")
	endpoint := viper.GetString("S3_ENDPOINT")
```

hakman (changes requested):

remyleone:
@hakman could you review again? :)

pkg/featureflag/featureflag.go
```diff
@@ -39,7 +40,7 @@ const (
 )
 
 func init() {
-	ParseFlags(os.Getenv(Name))
+	ParseFlags(viper.GetString(Name))
```
hakman, line 43: I think `viper.BindEnv()` would be required here also.
```suggestion
	_ = viper.BindEnv(Name)
	ParseFlags(viper.GetString(Name))
```

hakman (changes requested):

hakman:
@johngmyers Could you take a look too?

johngmyers:
I think we need a coherent strategy for using viper, probably that of https://github.com/carolynvs/stingoftheviper. I would prefer the viper keys be lowercase hyphen-separated, using the flag names, not all-caps underscore-separated.

hakman:
@remyleone Would you mind rebasing to address https://github.com/kubernetes/kops/pull/14642#issuecomment-1462082897?

hakman:
@remyleone Could you remove the go.mod & go.sum changes and squash the commits?

johngmyers:
I think my comments of Dec 18, 2022 are still relevant. We definitely shouldn't be defining new all-caps viper keys.

hakman:
> I think my comments of Dec 18, 2022 are still relevant. We definitely shouldn't be defining new all-caps viper keys.

To answer your comment from Dec 18, https://github.com/carolynvs/stingoftheviper may work in the example, but could not get it to work in practice.

Also, I only want to redo the way we do things for the existing env vars for cloud provider, cluster, state store and some other things used for development.

PR #14664, opened 2022-11-25
v1alpha3: move AWS-specific fields to AWSSpec

johngmyers:


olemarkus:
I am not entirely against moving AWS addons under the AWS spec. But for discoverability/usability, I am wondering if it makes more sense to have them under an `AddonsSpec` or something similar.

If people are using template includes (which I know some do), it is also more logical to group addon configs together.

johngmyers:
That can beg the question of what is an "addon" versus base functionality. Would AWS-CCM and EBS-CSI also go under this `AddonSpec`?

I could see NTH being enabled by default; it is incorrectly documented as such. There's a good argument for enabling Pod Identity Webhook by default. I could also see LBC enabled by default for new clusters.

olemarkus:
CCM, CSI and CNI are separate I think. But I see your point. I would love to get some more user input on this topic.

pkg/apis/kops/componentconfig.go
```diff
@@ -1088,8 +1088,8 @@ type CertManagerConfig struct {
 	HostedZoneIDs []string `json:"hostedZoneIDs,omitempty"`
 }
 
-// AWSLoadBalancerControllerConfig determines the AWS LB controller configuration.
-type AWSLoadBalancerControllerConfig struct {
+// LoadBalancerControllerSpec determines the AWS LB controller configuration.
+type LoadBalancerControllerSpec struct {
```
justinsb, line 1092: I agree with the Config -> Spec bit, not sure we want to drop the AWS bit.  If/when we add the GCE LoadBalancerController, won't we readd the AWS prefix?

I think we tend to end up with the path to the type in the type name, not the field name (e.g. PodSpec and NodeSpec).

pkg/apis/kops/validation/instancegroup.go
```diff
@@ -286,6 +276,10 @@ func CrossValidateInstanceGroup(g *kops.InstanceGroup, cluster *kops.Cluster, cl
 		}
 	}
 
+	if g.Spec.Containerd != nil {
```
justinsb, line 279: Nice catch here!

justinsb:
One nit on dropping the AWS prefix from AWSLoadBalancerControllerConfig, but I don't think that is surfaced in the end-user API.

/approve
/lgtm

/hold for contemplation of the AWS prefix :-)

pkg/apis/kops/componentconfig.go
johngmyers: We could always change it back at that point. It doesn't affect the API.

johngmyers:
Had to fix a conflict.

We could always change the name of the type later if a second cloud provider gets something called a "load balancer controller". It's similarly possible that another cloud could get a node termination handler.
/hold cancel

PR #14666, opened 2022-11-25
Release notes for 1.26.0-alpha.2

johngmyers:


hakman:
We don't add commits to the release notes anymore.

johngmyers:
We should update `release-process.md` then.

hakman:
Ok, will do that in the following days.

PR #14668, opened 2022-11-26
WIP Disallow running NTH and Karpenter together

johngmyers:
/cc @olemarkus


hakman:
Are you referring to https://github.com/aws/karpenter/issues/702?

johngmyers:
I was referring to https://karpenter.sh/v0.19.2/tasks/deprovisioning/#interruption

Seems that's an 0.19 feature and we'd have to turn it on.

olemarkus:
I think it's better to run NTH and Karpenter separately. Especially since we can mix karpenter and ASG-managed modes nodes.

PR #14669, opened 2022-11-26
v1alpha3: use new terminology for kubelet config

johngmyers:


hakman (approved):

pkg/apis/kops/validation/legacy.go
```diff
@@ -215,9 +215,9 @@ func ValidateCluster(c *kops.Cluster, strict bool) field.ErrorList {
 							allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("kubelet", "nonMasqueradeCIDR"), "kubelet nonMasqueradeCIDR did not match cluster nonMasqueradeCIDR"))
 						}
 					}
-					if fi.ValueOf(c.Spec.MasterKubelet.NonMasqueradeCIDR) != nonMasqueradeCIDRString {
-						if strict || c.Spec.MasterKubelet.NonMasqueradeCIDR != nil {
-							allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("masterKubelet", "nonMasqueradeCIDR"), "masterKubelet nonMasqueradeCIDR did not match cluster nonMasqueradeCIDR"))
+					if fi.ValueOf(c.Spec.ControlPlaneKubelet.NonMasqueradeCIDR) != nonMasqueradeCIDRString {
+						if strict || c.Spec.ControlPlaneKubelet.NonMasqueradeCIDR != nil {
+							allErrs = append(allErrs, field.Forbidden(fieldSpec.Child("controlPlaneKubelet", "nonMasqueradeCIDR"), "controlPlaneKubelet nonMasqueradeCIDR did not match cluster nonMasqueradeCIDR"))
```
rifelpet, line 220: we haven't defaulted to (or "advertised") v1alpha3 yet, correct? Will there be confusion when users provide v1alpha2 manifests as recommended by our documentation, but the field names in validation errors are for the v1alpha3 schema?

johngmyers: That's how I've been making changes so far. We don't want to lose track of the validation changes, ending up with a new API and old field names in the validation errors.

rifelpet: It seems like that could cause confusion until we default to v1alpha3. Do we have any docs around the API changes for v1alpha3? Perhaps that could at least serve as a reference for how to manually translate the validation errors into the necessary fixes users need to make in their v1alpha2 manifests.

johngmyers: I suppose we could add mapping tables to `cluster_spec.md` and `instance_groups.md`. It's a pity GitHub Markdown doesn't support sortable tables. I think we'd want to sort by the v1alpha2 name.

rifelpet: Addressed in https://github.com/kubernetes/kops/pull/14688

rifelpet (approved):

PR #14671, opened 2022-11-26
Update Calico documentation and validation

johngmyers:
/assign @olemarkus

### DIFF
--- a/docs/networking/calico.md
+++ b/docs/networking/calico.md
@@ -30,11 +30,13 @@ kops create cluster \
 
 ## Configuring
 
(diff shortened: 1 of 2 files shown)


johngmyers:
Does IPv6 + Typha make sense?

IPv6 + MTU seems like it would be possible, but I don't know why anyone would want to configure that way.

hakman:
Typha is a cache from my understanding, so it should makes sense:
https://projectcalico.docs.tigera.io/reference/typha/overview

olemarkus (approved):

PR #14672, opened 2022-11-26
Don't disable AWS src/dst checks in Calico IPv6

johngmyers:
/assign @olemarkus

### DIFF
--- a/pkg/apis/kops/v1alpha2/networking.go
+++ b/pkg/apis/kops/v1alpha2/networking.go
@@ -108,8 +108,8 @@ type CalicoNetworkingSpec struct {
 	// AllowIPForwarding enable ip_forwarding setting within the container namespace.
 	// (default: false)
 	AllowIPForwarding bool `json:"allowIPForwarding,omitempty"`
-	// AWSSrcDstCheck enables/disables ENI source/destination checks (AWS only)
-	// Options: Disable (default), Enable, or DoNothing
+	// AWSSrcDstCheck enables/disables ENI source/destination checks (AWS IPv4 only)
+	// Options: Disable (default for IPv4), Enable, or DoNothing
 	AWSSrcDstCheck string `json:"awsSrcDstCheck,omitempty"`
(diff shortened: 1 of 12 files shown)


johngmyers:
Looks like I made it impossible to disable src/dst checks on IPv6 in the template. Should I support that or add validation against it?

hakman:
> Looks like I made it impossible to disable src/dst checks on IPv6 in the template. Should I support that or add validation against it?

Let's not support it at all for IPv6. Should not be needed.

pkg/apis/kops/v1alpha2/networking.go
```diff
@@ -108,8 +108,8 @@ type CalicoNetworkingSpec struct {
 	// AllowIPForwarding enable ip_forwarding setting within the container namespace.
 	// (default: false)
 	AllowIPForwarding bool `json:"allowIPForwarding,omitempty"`
-	// AWSSrcDstCheck enables/disables ENI source/destination checks (AWS only)
-	// Options: Disable (default), Enable, or DoNothing
+	// AWSSrcDstCheck enables/disables ENI source/destination checks (AWS IPv4 only)
+	// Options: Disable (default for IPv4), Enable, or DoNothing
```
hakman, line 112: Isn't the above `(AWS IPv4 only)` enough?
```suggestion
	// Options: Disable (default), Enable, or DoNothing
```

johngmyers: That could imply one has to override the default when on IPv6

hakman: To me the initial versions implies that there is some default for IPv4 and undefined for IPv6 and has to be set by the operator.
Would be nice to also add validation to not be able to set it for IPv6 at all.

johngmyers: This PR includes validation preventing it from being set to anything other than `DoNothing` on IPv6.

hakman: K, I'm good/

PR #14676, opened 2022-11-26
Update documentation and validation for Cilium

johngmyers:
/assign @olemarkus


olemarkus:
Should we add something to the release notes as well?

johngmyers:
What would we add to the release notes?

olemarkus:
That cilium eni is now considered stable.

johngmyers:
That change was done by #14622, not this PR.

olemarkus:
Yeah, most of this should probably have been a part of that PR. But it's in the same vein as https://github.com/kubernetes/kops/pull/14676/files#diff-664156d0f9d53d9d867cf3470ef83bfc1a7965e7ced8f0e61df746b8f9021786R113

But happy to lgtm if you feel this should be a separate PR.

/lgtm

PR #14680, opened 2022-11-27
Refactor terraformWriter.Literal

johngmyers:
This is probably best reviewed commit-by-commit.

Moves logic for Literal into the constructors.

/assign @rifelpet

### DIFF
--- a/util/pkg/vfs/tests/gsfs_test.go
+++ b/util/pkg/vfs/tests/gsfs_test.go
@@ -56,15 +56,15 @@ func TestGSRenderTerraform(t *testing.T) {
 			{
 				"bucket": "foo",
 				"name": "bar",
-				"provider": "${google.files}",
+				"provider": "google.files",
 				"source": "\"${path.module}/data/google_storage_bucket_object_bar_content\""
 			}
 			`,
 			expectedACLJSON: `
 			{
 				"bucket": "foo",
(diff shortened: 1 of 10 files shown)


util/pkg/vfs/tests/gsfs_test.go
```diff
@@ -56,15 +56,15 @@ func TestGSRenderTerraform(t *testing.T) {
...
+				"provider": "google.files",
 				"source": "\"${path.module}/data/google_storage_bucket_object_bar_content\""
 			}
 			`,
 			expectedACLJSON: `
 			{
 				"bucket": "foo",
-				"object": "${google_storage_bucket_object.bar.output_name}",
-				"provider": "${google.files}",
+				"object": "google_storage_bucket_object.bar.output_name",
```
rifelpet, line 66: I think this needs to be:
```suggestion
				"object": "${google_storage_bucket_object.bar.output_name}",
```

but I'm not sure if that's just a side effect of how we're using json for unit tests

johngmyers: That's just a side effect of how JSON is being used for unit tests.  JSON is only being used for unit tests at this point, so the "expected" value is arbitrary.

rifelpet (approved):

PR #14682, opened 2022-11-27
Make `--cloud` and `--zones` flags required

hakman:
kOps has some logic that is inferring the cloud provider. This is done by matching the specified zones with cloud provider specific list of zones/regions. These lists are static and unlikely to be maintained over time. All this logic could be gone if the `--cloud` flag would be mandatory. As this flag is only passed during cluster creation, it seems to me a better approach than to rely on flaky guessing. Would also open the possibility to start initialising the cloud components earlier and do better validation in the future (like for AWS instance types).

/cc @johngmyers @olemarkus @rifelpet

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -231,11 +231,13 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 		sort.Strings(validClouds)
 	}
 	cmd.Flags().StringVar(&options.CloudProvider, "cloud", options.CloudProvider, fmt.Sprintf("Cloud provider to use - %s", strings.Join(validClouds, ", ")))
+	cmd.MarkFlagRequired("cloud")
 	cmd.RegisterFlagCompletionFunc("cloud", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return validClouds, cobra.ShellCompDirectiveNoFileComp
 	})
 
 	cmd.Flags().StringSliceVar(&options.Zones, "zones", options.Zones, "Zones in which to run the cluster")
+	cmd.MarkFlagRequired("zones")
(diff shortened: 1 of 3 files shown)


johngmyers:
I think that's an extreme reaction, leading to a significant drop in usability.

If keeping the list up to date is important, we could create an e2e job on each relevant cloud that enumerates available zones and ensures they're in the list.

hakman:
> I think that's an extreme reaction, leading to a significant drop in usability.

How would adding `--cloud=aws/gce/hetzner/...` be a significant drop in usability?

johngmyers:
It is an additional point of friction when creating clusters from the command-line.

johngmyers:
The PR doesn't remove the need to maintain the list of zones, as it is still used for command completion.

hakman:
Not for long :)

rifelpet:
I don't think making one flag required is a significant drop in usability. Every situation in which this command is being executed, the appropriate value for this flag is known so its just a matter of ensuring the flag and value are included in the command. How many required flags does the command have currently? I think something as primitive as a cluster's cloud provider is a reasonable value that must be specified when creating clusters. We'd want to audit our docs to ensure it is present in every example command.

johngmyers:
So the plan is to also remove completion for the zone names?

These little UI conveniences are worth a lot: they're why TOPS-20 was such a joy to work with. The state of the Linux CLI is tragic.

Marking `--zones` as required in Cobra is worthwhile. The cloud guessing code could be hoisted up out of `NewCluster`. The construction of `fi.Cloud`, however, is too complicated to hoist up completely.

hakman:
We could auto-complete zones from cloud. For AWS at least it's pretty easy to get at least the regions without cloud calls.
For GCP and others, it requires cloud calls. Those should not be so bad and would be simpler to catch missing credentials early.

hakman:
https://github.com/kubernetes/kops/pull/14768 and https://github.com/kubernetes/kops/pull/14683 make this idea obsolete.

PR #14683, opened 2022-11-27
Get AWS zones list from AWS SDK

johngmyers:


pkg/zones/wellknown.go
```diff
@@ -389,7 +392,18 @@ func GuessCloudForZone(zone string) (kops.CloudProviderID, bool) {
 func WellKnownZonesForCloud(matchCloud kops.CloudProviderID) []string {
 	switch matchCloud {
 	case kops.CloudProviderAWS:
-		return awsZones
+		awsCloud, err := awsup.NewAWSCloud("us-east-1", map[string]string{})
+		zones, err := awsCloud.EC2().DescribeAvailabilityZones(&ec2.DescribeAvailabilityZonesInput{
+			AllAvailabilityZones: aws.Bool(true),
```
johngmyers, line 397: We possibly shouldn't set this for the command-completion case.

hakman: Do we want to get the exact zones, or regions would be good enough and allow operators to type the final character? For AWS we could use the partitions list:
https://github.com/aws/aws-sdk-go/blob/1356aada4806053cf6a64f66288912c9625fa016/aws/endpoints/defaults.go#L103

johngmyers: For the command-completion case, we're supposed to list all the possibilities so that the shell can list them out.

In the infer-the-cloud case we could be lax.

johngmyers:
This is slow when there is no prefix, probably because `NewAWSCloud` initializes clients for services we don't end up using. We might want to create an "EC2 only" case.

pkg/zones/wellknown.go
hakman: This approach can only be used for zones. Cloud inference is not possible as you don't have credentials for all types of clouds.

johngmyers: Returning the region name as the completion gives a better command-completion experience, so I changed it to that.

pkg/zones/wellknown.go
```diff
@@ -386,10 +391,52 @@ func GuessCloudForZone(zone string) (kops.CloudProviderID, bool) {
 	return matches[0], true
 }
 
-func WellKnownZonesForCloud(matchCloud kops.CloudProviderID) []string {
+func WellKnownZonesForCloud(matchCloud kops.CloudProviderID, prefix string) []string {
 	switch matchCloud {
 	case kops.CloudProviderAWS:
-		return awsZones
+		prefix = strings.ToLower(prefix)
+		awsCloud, err := awsup.NewAWSCloud("us-east-1", map[string]string{})
```
rifelpet, line 398: We'll need to figure out a way to have this work for other AWS partitions

johngmyers: I don't have the ability to test other partitions.

pkg/zones/wellknown.go
johngmyers: You can infer clouds that you have credentials for. Creating clusters in clouds you don't have credentials for isn't going to work anyway.

pkg/zones/wellknown.go
rifelpet: I can test govcloud. We just need a way to use a partition-specific region. This would likely require parsing the resulting ARN from sts.GetCallerIdentity and using a map of partitions to region. It may be wise to respect AWS_REGION or any region in the AWS SDK's config lookup sequence as well.

[endpoints.DefaultPartitions](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/) and their Regions() method might be able to help too.

johngmyers: We could probably just query one region in each partition.

I don't think we want to limit to AWS_REGION. I see that as a default, not a restriction.

johngmyers: Rewrote to use `endpoints.DefaultPartitions` to get the list of region names.

pkg/zones/wellknown.go
hakman: I appreciate this change. It is a very good approach for the region command-completion for AWS.
Though, looping through N clouds just to infer the cloud provider is not an optimal approach. It would most likely become slow and unpredictable the more cloud providers we add to the mix.

johngmyers: For one, is there any reason this PR shouldn't go in as-is?

This AWS code is able to quickly determine, without making remote calls, that it doesn't have any matches when it is given regions from another cloud provider.

pkg/zones/wellknown.go
```diff
@@ -386,10 +258,49 @@ func GuessCloudForZone(zone string) (kops.CloudProviderID, bool) {
...
+		for _, partition := range endpoints.DefaultResolver().(endpoints.EnumPartitions).Partitions() {
+			for _, region := range partition.Regions() {
+				regionName := strings.ToLower(region.ID())
+				if !strings.HasPrefix(regionName, prefix) {
+					// For the guess-cloud case, if the zone name is in the form of zones belonging
+					// to a region, that's good enough.
+					if len(prefix) == len(regionName)+1 && strings.HasPrefix(prefix, regionName) {
+						found = append(found, prefix)
+					}
+					if !strings.HasPrefix(prefix, regionName+"-") {
```
hakman, line 276: What does `regionName+"-"` represent?

johngmyers: That's the prefix for Local Zones and Wavelength Zones. Basically, any zone suffix that isn't a single letter.

johngmyers: Added comment as requested in Office Hours.

johngmyers: @hakman is this good to go now?

johngmyers:
@hakman is there anything holding up approval?

hakman:
@johngmyers What do you think about this reshuffle of the code?
```go
				regionName := strings.ToLower(region.ID())
				if regionName == prefix {
					// If the prefix is a region name, return all its zones as the completion options.
					awsCloud, err := awsup.NewAWSCloud(regionName, map[string]string{})
					if err != nil {
						continue
					}
					var zones *ec2.DescribeAvailabilityZonesOutput
					zones, err = awsCloud.EC2().DescribeAvailabilityZones(&ec2.DescribeAvailabilityZonesInput{
						AllAvailabilityZones: aws.Bool(true),
					})
					if err != nil {
						continue
					}
					for _, zone := range zones.AvailabilityZones {
						found = append(found, *zone.ZoneName)
					}
				} else if strings.HasPrefix(prefix, regionName) && len(prefix) > len(regionName) {
					// If the prefix is in the form of zones belonging to a region, that's good enough.
					found = append(found, prefix)
				} else if strings.HasPrefix(regionName, prefix) {
					// Return the region name as the completion option. After the user completes
					// that much, the code will then look up the specific zone options.
					found = append(found, regionName)
				}
```

johngmyers:
@hakman If `prefix` is "us-east-1-b" then it should return `[]string{"us-east-1-bue-1a", "us-east-1-bos-1a"}`, not `[]string{"us-east-1-b"}`.

hakman:
I updated the code a little. Could you check one more time?

hakman:
> Does that work with a prefix of `""`?

Now it does.
I meant this as a simplification. Do you think it's more clear or not than the initial code?

hakman (approved):

PR #14685, opened 2022-11-27
Revert "Calico: Work around host port/conntrack problem"

johngmyers:
This reverts commit 9105368a411d47b4a519d25ca7c369a7db2976e8.


seh (commented):
Though it's not purely part of reverting the earlier patch, we now have a sibling file named _k8s-1.25.yaml.template_ that includes the same code to be removed [between lines 4677 and 4682, inclusive](https://github.com/kubernetes/kops/blob/5d1c3a83619f6c7868050bbc6086656d71756012/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.25.yaml.template#L4677-L4682).

Should we include that deletion here, or in a separate PR? I prefer to include it here.

seh (approved):
Thank you for covering both places in which this environment variable appears.

hakman (approved):

PR #14687, opened 2022-11-28
[WIP] Use viper to read `KOPS_STATE_STORE` and `KOPS_CLUSTER_NAME` env vars

hakman:
Maybe I am missing something, but `viper.BindPFlag()` doesn't seem to work (for me).

### DIFF
--- a/cmd/kops/root.go
+++ b/cmd/kops/root.go
@@ -147,13 +147,14 @@ func NewCmdRoot(f *util.Factory, out io.Writer) *cobra.Command {
 		return []string{"yaml", "json"}, cobra.ShellCompDirectiveFilterFileExt
 	})
 
-	cmd.PersistentFlags().StringVar(&rootCommand.RegistryPath, "state", "", "Location of state storage (kops 'config' file). Overrides KOPS_STATE_STORE environment variable")
-	viper.BindPFlag("KOPS_STATE_STORE", cmd.PersistentFlags().Lookup("state"))
 	viper.BindEnv("KOPS_STATE_STORE")
+	viperStateStore := strings.TrimSuffix(viper.GetString("KOPS_STATE_STORE"), "/")
+	cmd.PersistentFlags().StringVar(&rootCommand.RegistryPath, "state", viperStateStore, "Location of state storage (kops 'config' file). Overrides KOPS_STATE_STORE environment variable")
 	// TODO implement completion against VFS
 
-	defaultClusterName := os.Getenv("KOPS_CLUSTER_NAME")
-	cmd.PersistentFlags().StringVarP(&rootCommand.clusterName, "name", "", defaultClusterName, "Name of cluster. Overrides KOPS_CLUSTER_NAME environment variable")
+	viper.BindEnv("KOPS_CLUSTER_NAME")
+	viperClusterName := viper.GetString("KOPS_CLUSTER_NAME")
+	cmd.PersistentFlags().StringVar(&rootCommand.clusterName, "name", viperClusterName, "Name of cluster. Overrides KOPS_CLUSTER_NAME environment variable")
 	cmd.RegisterFlagCompletionFunc("name", commandutils.CompleteClusterName(rootCommand.factory, false, false))
 
 	// create subcommands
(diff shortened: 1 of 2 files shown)


cmd/kops/root.go
```diff
@@ -147,13 +147,12 @@ func NewCmdRoot(f *util.Factory, out io.Writer) *cobra.Command {
...
-	cmd.PersistentFlags().StringVar(&rootCommand.RegistryPath, "state", "", "Location of state storage (kops 'config' file). Overrides KOPS_STATE_STORE environment variable")
-	viper.BindPFlag("KOPS_STATE_STORE", cmd.PersistentFlags().Lookup("state"))
 	viper.BindEnv("KOPS_STATE_STORE")
+	cmd.PersistentFlags().StringVar(&rootCommand.RegistryPath, "state", viper.GetString("KOPS_STATE_STORE"), "Location of state storage (kops 'config' file). Overrides KOPS_STATE_STORE environment variable")
 	// TODO implement completion against VFS
 
-	defaultClusterName := os.Getenv("KOPS_CLUSTER_NAME")
-	cmd.PersistentFlags().StringVarP(&rootCommand.clusterName, "name", "", defaultClusterName, "Name of cluster. Overrides KOPS_CLUSTER_NAME environment variable")
+	viper.BindEnv("KOPS_CLUSTER_NAME")
+	cmd.PersistentFlags().StringVar(&rootCommand.clusterName, "name", viper.GetString("KOPS_CLUSTER_NAME"), "Name of cluster. Overrides KOPS_CLUSTER_NAME environment variable")
```
johngmyers, line 155: `make gen-cli-docs` should clear the `KOPS_CLUSTER_NAME` environment variable, lest it put its value in the generated documentation.

johngmyers: Maybe the `unexport` at the top does that? But gen-cli-docs explicitly clears KOPS_STATE_STORE?

In any case, this wouldn't be a new problem.

johngmyers:
The BindPFlag is working for me for KOPS_STATE_STORE.

I notice that we ignore the return value of BindPFlag. Perhaps the lookup was failing?

hakman:
> I notice that we ignore the return value of BindPFlag. Perhaps the lookup was failing?

I tried doing the same for `KOPS_CLUSTER_NAME` and fails. `KOPS_STATE_STORE` seems to work, but still annoying to have 2 approaches. Also, I don't see the value of BindPFlag over the GetString as default.

johngmyers:
GetString as default causes the viper-sourced value to leak into the command-line help.

hakman:
OK, that's not so bad.
Any thoughts on why `KOPS_CLUSTER_NAME` when switched to `BindPFlag`?

johngmyers:
I believe KOPS_STATE_STORE works because of https://github.com/kubernetes/kops/blob/master/cmd/kops/root.go#L211

johngmyers:
So I think we should be consistent in how we use viper. 

There's also the viper use for `lifecycle-overrides` in update_cluster.go. That is using a lowercase, hyphenated viper key.

johngmyers:
Perhaps there should be a key `state-store` with an alias `KOPS_STATE_STORE` for backwards compatibility?

hakman:
Quite confusing...

johngmyers:
Or maybe `state` to match the flag name.

I see `lifecycle-overrides` doesn't have a `viper.GetString()`. I wonder if it works if the viper key matches the flag name.

johngmyers:
https://github.com/carolynvs/stingoftheviper

remyleone:
I've tried to do a project wide change using viper https://github.com/kubernetes/kops/pull/14644 and a smaller one here: https://github.com/kubernetes/kops/pull/14642

hakman:
Thanks @remyleone. Your changes inspired my experiment.
Regarding the project wide change in https://github.com/kubernetes/kops/pull/14644, it is a little too wide. I see viper as useful for command line, feature flags and maybe some random env vars used for dev purposes. I don't see a value in using vipers for nodeup, protokube and various containers.
https://github.com/kubernetes/kops/pull/14642 is quite nice and I think it's a good start to use viper more.

PR #14691, opened 2022-11-30
Stop trying to populate arbitrary cluster fields from the channel

johngmyers:
When I refactored `Spec.Networking` to be a non-pointer, this code started forcing new clusters to have Kubenet.

This seems to be an extremely dubious feature.

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -202,16 +202,13 @@ func NewCluster(opt *NewClusterOptions, clientset simple.Clientset) (*NewCluster
 		},
 	}
 
-	if channel.Spec.Cluster != nil {
-		cluster.Spec = *channel.Spec.Cluster
-
+	cluster.Spec.Channel = opt.Channel
+	if opt.KubernetesVersion == "" {
 		kubernetesVersion := api.RecommendedKubernetesVersion(channel, kops.Version)
 		if kubernetesVersion != nil {
 			cluster.Spec.KubernetesVersion = kubernetesVersion.String()
 		}
-	}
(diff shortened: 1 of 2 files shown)


hakman:
Would it be safe to also start ignoring the the cluster field in `ChannelSpec` in general (maybe even removing it)?
/lgtm

johngmyers:
This does ignore the cluster field in `ChannelSpec`. We can't remove it from the published channels, as the previous code requires it be present in order to look at the `kubernetesVersion` field of the appropriate kOps version.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -202,16 +202,13 @@ func NewCluster(opt *NewClusterOptions, clientset simple.Clientset) (*NewCluster
 		},
 	}
 
-	if channel.Spec.Cluster != nil {
-		cluster.Spec = *channel.Spec.Cluster
```
hakman, line 206: I think the intention here was to be able to populate ClusterSpec from channels.

hakman:
> This does ignore the cluster field in `ChannelSpec`. We can't remove it from the published channels, as the previous code requires it be present in order to look at the `kubernetesVersion` field of the appropriate kOps version.

I meant that, we could do something in API like `Cluster *ClusterSpec  `json:"-"``, to make it clear that it's ignored and deprecated.

upup/pkg/fi/cloudup/new_cluster.go
johngmyers: Except it then overwrites large sections of the config.

The current code works only because it overwrites the kubenet setting in the channel with a blank networking struct.

hakman:
Let's go with it, we can optimise late if needed.
/approve

Issue #14690, opened 2022-11-29
Concerns about the MPL-2.0 of Hashicorp deps


hakman:
/kind office-hours

CNCF did a license scan as part of https://github.com/kubernetes/community/issues/6992 and kOps is mentioned with the as using deps under the MPL-2.0 license which is weak copyleft (**Be sure that it is used only as dynamic libraries, to be safe if it's not required remove it from your repo.**):
```
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/errwrap/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-cleanhttp/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-immutable-radix/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-multierror/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-plugin/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-retryablehttp/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-rootcerts/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-secure-stdlib/mlock/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-secure-stdlib/parseutil/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-secure-stdlib/strutil/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-sockaddr/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-uuid/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/go-version/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/golang-lru/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/hcl/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/hcl/v2/LICENSE
kubernetes-2022-10-13.zip/kops/vendor/github.com/hashicorp/memberlist/LICENSE
[...]

PR #14692, opened 2022-11-30
Remove support for using Vault as state store

hakman:
Refs #14690 

/cc @olemarkus @johngmyers


johngmyers:
If the CNCF's timeline allows, I'd suggest we deprecate in 1.26 for removal in 1.27.

hakman:
It is an experimental feature, under feature flag. There are not guarantees of deprecation and removal.

johngmyers:
Oh right, feature flagged.

johngmyers (approved):
/hold for consideration of wording change

PR #14700, opened 2022-12-01
gce: Update cluster-autoscaler config

hakman:
/cc @olemarkus @justinsb


johngmyers:
Is this covered by an integration test? It doesn't appear to be covered by bootstrapchannelbuilder_test.go

hakman:
There integration tests like the `many-addons*` ones covering cluster-autoscaler for AWS.
I can one add for GCE too in a separate PR.

hakman:
https://github.com/kubernetes/kops/pull/14712 should add some more integration tests that can be used to see how the output changes.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -824,6 +825,37 @@ func (tf *TemplateFunctions) GetNodeInstanceGroups() map[string]kops.InstanceGro
 	return nodegroups
 }
 
+type ClusterAutoscalerNodeGroup struct {
+	AutoScale *bool
+	MinSize   int32
+	MaxSize   int32
+	Other     string
```
olemarkus, line 832: Can we call this `Name`?

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -824,6 +825,37 @@ func (tf *TemplateFunctions) GetNodeInstanceGroups() map[string]kops.InstanceGro
...
+	for _, ig := range tf.KopsModelContext.InstanceGroups {
+		if ig.Spec.Role == kops.InstanceGroupRoleNode && (ig.Spec.Autoscale == nil || fi.ValueOf(ig.Spec.Autoscale)) {
+			group := ClusterAutoscalerNodeGroup{
+				AutoScale: ig.Spec.Autoscale,
+				MinSize:   fi.ValueOf(ig.Spec.MinSize),
+				MaxSize:   fi.ValueOf(ig.Spec.MaxSize),
+			}
+			if cluster.Spec.GetCloudProvider() == kops.CloudProviderGCE {
+				cloud := tf.cloud.(gce.GCECloud)
+				format := "https://www.googleapis.com/compute/v1/projects/%s/zones/%s/instanceGroups/%s"
```
olemarkus, line 848: I've seen this output used in `kops get instances` as well, so wondering if there may be some generic function for this somewhere.

hakman: Not to my knowledge...

upup/pkg/fi/cloudup/template_functions.go
hakman: It comes from cluster-autoscaler args:
https://github.com/kubernetes/autoscaler/blob/f3242d8485ab3d53a6ac3fbe430918408295a8cf/cluster-autoscaler/main.go#L144-L146

olemarkus: Ah okay. I guess it can be key/val and whatnot with some providers

upup/pkg/fi/cloudup/template_functions.go
olemarkus: Looks like that same sprintf is used everywhere. But there is also a function here: upup/pkg/fi/cloudup/gcetasks/instance.go

But cleanup for another time, I suppose

olemarkus (approved):

PR #14702, opened 2022-12-02
Fix the addons/authentication.aws startup failure

nnmin-aws:
The aws-iam-authenticator pod always fails from [kops update cluster ](https://kops.sigs.k8s.io/getting_started/aws/#build-the-cluster) complaining the following error. This PR fix the error. 
The added configMap is only placeholder to avoid the mounting error as below. It still prevents unauthorized user/role from authenticating. 

```
Events:
  Type     Reason       Age                   From     Message
  ----     ------       ----                  ----     -------
  Warning  FailedMount  44m                   kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[output kube-api-access-rkvtd config state]: timed out waiting for the condition
  Warning  FailedMount  31m (x5 over 60m)     kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[kube-api-access-rkvtd config state output]: timed out waiting for the condition
  Warning  FailedMount  26m (x2 over 56m)     kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[state output kube-api-access-rkvtd config]: timed out waiting for the condition
  Warning  FailedMount  6m19s (x16 over 65m)  kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[config state output kube-api-access-rkvtd]: timed out waiting for the condition
  Warning  FailedMount  2m7s (x40 over 67m)   kubelet  MountVolume.SetUp failed for volume "config" : configmap "aws-iam-authenticator" not found
```
/cc @olemarkus
/assign @olemarkus

### DIFF
--- a/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
@@ -110,6 +110,21 @@ subjects:
   namespace: kube-system
 
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: kube-system
+  name: aws-iam-authenticator
+  labels:
+    k8s-app: aws-iam-authenticator
+data:
+  config.yaml: |
+    # a unique-per-cluster identifier to prevent replay attacks
+    # (good choices are a random token or a domain name that will be unique to your cluster)
+
+    clusterID: {{ or .Authentication.AWS.ClusterID ClusterName }}
+
+---
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
@@ -158,7 +173,7 @@ spec:
       # - output (output kubeconfig to plug into your apiserver configuration, mounted from the host)
       containers:
       - name: aws-iam-authenticator
-        image: {{ or .Authentication.AWS.Image "602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-iam-authenticator:v0.5.9" }}
+        image: {{ or .Authentication.AWS.Image "602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-iam-authenticator:v0.5.12" }}
         args:
         - server
         {{- if or (not .Authentication.AWS.BackendMode) (contains "MountedFile" .Authentication.AWS.BackendMode) }}
--- a/tests/integration/update_cluster/complex/data/aws_s3_object_complex.example.com-addons-authentication.aws-k8s-1.12_content
+++ b/tests/integration/update_cluster/complex/data/aws_s3_object_complex.example.com-addons-authentication.aws-k8s-1.12_content
@@ -134,6 +134,26 @@ subjects:
 
 ---
 
+apiVersion: v1
+data:
+  config.yaml: |
+    # a unique-per-cluster identifier to prevent replay attacks
+    # (good choices are a random token or a domain name that will be unique to your cluster)
+
+    clusterID: complex.example.com
+kind: ConfigMap
+metadata:
+  creationTimestamp: null
+  labels:
+    addon.kops.k8s.io/name: authentication.aws
+    app.kubernetes.io/managed-by: kops
+    k8s-app: aws-iam-authenticator
+    role.kubernetes.io/authentication: "1"
+  name: aws-iam-authenticator
+  namespace: kube-system
+
+---
+
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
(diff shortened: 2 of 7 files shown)


johngmyers:
This might allow the pods to start up, but it would have the effect of preventing anyone from authenticating. The ConfigMap specifying authorized identities would be periodically overwritten with the blank version being added here.

nnmin-aws:
> This might allow the pods to start up, but it would have the effect of preventing anyone from authenticating. The ConfigMap specifying authorized identities would be periodically overwritten with the blank version being added here.

thanks @johngmyers for your comments!
This PR also prevents anyone from authenticating because all three maps are empty. You can check https://github.com/kubernetes-sigs/aws-iam-authenticator to find how three maps are used. I am also a contributor for aws-iam-authenticator.

Also I attached test result. you can also checkout this PR and it will clear your concern.

`% kubectl logs aws-iam-authenticator-xgsmr -n kube-system --kubeconfig "/Users/nnmin/kops/e2e-test-artifacts/test-cluster-222.k8s.local.kubeconfig" 
error: You must be logged in to the server (Unauthorized)`

upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
```diff
@@ -109,6 +109,25 @@ subjects:
...
+  namespace: kube-system
+  name: aws-iam-authenticator
+  labels:
+    k8s-app: aws-iam-authenticator
+data:
+  config.yaml: |
+    # a unique-per-cluster identifier to prevent replay attacks
+    # (good choices are a random token or a domain name that will be unique to your cluster)
+
+    clusterID: {{ ClusterName }}
```
johngmyers, line 125: Why set `clusterID` here when it's already being passed as a flag? And this appears to ignore the `ClusterID` setting in `AWSAuthenticationSpec`.

johngmyers:
I am failing to understand how an aws-iam-authenticator that prevents anyone from authenticating is useful.

nnmin-aws:
> I am failing to understand how an aws-iam-authenticator that prevents anyone from authenticating is useful.

thanks @johngmyers for your comments!

In this example, the three 


> I am failing to understand how an aws-iam-authenticator that prevents anyone from authenticating is useful.

thanks @johngmyers for the comments!

First, it is better than having a failed pod. agree? 😆 
Second, you can update either mapUsers/mapRoles/mapAccounts to have valid users or roles and start the cluster, then aws-iam-authenticator will allow these assigned users or roles to log in. In this way, it will save your time/manual work to re-apply the aws-iam-authenticator daemonset from the pod failure.

johngmyers:
Having the resource in the manifest means kOps takes control of that resource. If the admin updates the content of that ConfigMap, kOps will later revert those changes.

This is worse than having the pods not start up.

nnmin-aws:
> Having the resource in the manifest means kOps takes control of that resource. If the admin updates the content of that ConfigMap, kOps will later revert those changes.
> 
> This is worse than having the pods not start up.

there are already other resources such as CRD/ClusterRole/ServiceAccount/ClusterRoleBinding included in the same template. does it seem concern to add a placeholder configmap to make it run instead of failing?

johngmyers:
kOps manages those resources entirely. There is no expectation for an admin to modify them behind kOps's back in order to get the subsystem to function.

It would be a much better experience if we could make CRD the default `backendMode`, but it's inexplicably still marked as alpha. ConfigMaps are a terrible mechanism for configuring things like roles, users, and accounts.

And the concept of a "placeholder ConfigMap" is faulty. If aws-iam-authenticator doesn't need the information that comes from the ConfigMap, it shouldn't fail to start when the ConfigMap isn't present.

upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
nnmin-aws: Could you please give details to reproduce the ignorance you observed?

I have observed the failing aws-iam-authenticator from [kops update cluster](https://kops.sigs.k8s.io/getting_started/aws/#build-the-cluster), this PR can fix it.

nnmin-aws:
> kOps manages those resources entirely. There is no expectation for an admin to modify them behind kOps's back in order to get the subsystem to function.
> 
> It would be a much better experience if we could make CRD the default `backendMode`, but it's inexplicably still marked as alpha. ConfigMaps are a terrible mechanism for configuring things like roles, users, and accounts.
> 
> And the concept of a "placeholder ConfigMap" is faulty. If aws-iam-authenticator doesn't need the information that comes from the ConfigMap, it shouldn't fail to start when the ConfigMap isn't present.

No, aws-iam-authenticator binary doesn't need the information at all. But the container specify to mount this configmap and fails when it doesn't exist 🤣 . 

aws-iam-authenticator already support a new mode `dynamicfile`. it monitor a file(same content with configmap) and dynamic reload to memory whenever there is change on the file.

johngmyers:
> No, aws-iam-authenticator binary doesn't need the information at all. But the container specify to mount this configmap and fails when it doesn't exist

So perhaps a fix for the startup failure would be better directed to aws-iam-authenticator instead of kOps?

johngmyers:
Might need to mark the volume `optional` in the pod template.

PR #14706, opened 2022-12-02
v1alpha3: move networking fields under networking

johngmyers:


rifelpet:
Could we merge networkCIDR and additionalNetworkCIDRs into a single networkCIDRs field?

PR #14707, opened 2022-12-02
aws: Skip Route53 cleanup for clusters without DNS

hakman:
Small cleanup improvement to get tests passing again in https://github.com/kubernetes/cloud-provider-aws/pull/532:
```
Ginkgo ran 1 suite in 2m57.709627616s
Test Suite Passed
/home/prow/go/src/k8s.io/cloud-provider-aws
I1202 08:29:47.704238   27602 delete_cluster.go:128] Looking for cloud resources to delete
Error: error querying for route53 zones: AccessDenied:
	User: arn:aws:iam::607362164682:user/awstester is not authorized to perform: route53:ListHostedZones
	because no identity-based policy allows the route53:ListHostedZones action
		status code: 403, request id: 911b0c35-ba05-4950-880f-3d78acabfddb
make: *** [Makefile:133: test-e2e] Error 1
```

### DIFF
--- a/pkg/resources/aws/aws.go
+++ b/pkg/resources/aws/aws.go
@@ -32,7 +32,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/route53"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
-	"k8s.io/kops/pkg/dns"
+	kopsapi "k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/pkg/resources"
 	"k8s.io/kops/pkg/resources/spotinst"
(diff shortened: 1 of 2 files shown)


olemarkus (approved):

johngmyers:
I don't think `route53:ListHostedZones` is too much to ask for.

I don't like passing in the `Cluster`. It might not exist anymore: one of our testing AWS accounts had all of its S3 buckets deleted.

I was thinking of using `ListResourcesAWS()` for a utility to clean up EKS clusters.

hakman:
> I don't think route53:ListHostedZones is too much to ask for.

It is not in general, but for getting started or restricted accounts it may be a problem. Gossip and None DNS don't require it, so most optimal is not to get there.

> I don't like passing in the Cluster.

Would passing only the value of `cluster.UsesNoneDNS()` make it better?

johngmyers:
> Would passing only the value of cluster.UsesNoneDNS() make it better?

Perhaps an exclusion list of (a subset of) AWS services?

PR #14713, opened 2022-12-03
Refactor writing of Terraform data sources

johngmyers:


upup/pkg/fi/cloudup/terraform/element.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	writeIndent(buffer, indent)
+	buffer.WriteString("}")
+}
+
+func ToElement(item interface{}) (element, error) {
+	if s, ok := item.(string); ok {
+		return terraformWriter.LiteralFromStringValue(s), nil
+	}
+	if s, ok := item.(*string); ok {
```
justinsb, line 84: Nit: Could probably swap the `string` and `*string` checks and just fall-through (as we do with reflect.Pointer below)

upup/pkg/fi/cloudup/terraform/element.go
```diff
@@ -0,0 +1,130 @@
...
+		v = v.Elem()
+	}
+	if v.Kind() == reflect.Struct {
+		o := &object{
+			field: map[string]element{},
+		}
+		for _, field := range reflect.VisibleFields(v.Type()) {
+			key := field.Tag.Get("cty")
+			if key == "" {
+				var b strings.Builder
```
justinsb, line 104: Maybe refactor this into its own method, `defaultCtyName` or similar, just because it would make this method much easier to read?

justinsb:
LGTM.  I see the tests are passing.  Do we have test coverage on the formatting, or are we testing only after passing through a `tf fmt`?

upup/pkg/fi/cloudup/terraform/element.go
johngmyers: I was trying to avoid calling `reflect` on strings. But that's an unnecessary micro-optimization.

upup/pkg/fi/cloudup/terraform/element.go
johngmyers: Refactored

johngmyers:
The integration tests give us coverage. This particular change is covered by `TestSharedVPCIPv6()`.

justinsb:
Thanks @johngmyers - looks great

/approve
/lgtm

PR #14714, opened 2022-12-03
WIP: Use kops-controller for discovery (instead of gossip)

justinsb:
This should allow more reuse.


johngmyers:
I prefer to keep nodeup-specific code under `nodeup/pkg`. I don't see the need to authenticate the client for discovery requests and would prefer if the discovery clients didn't have credentials that could be used to generate certs.

What is the value of adding the complexity of GRPC?

justinsb:
> I prefer to keep nodeup-specific code under nodeup/pkg. I don't see the need to authenticate the client for discovery requests and would prefer if the discovery clients didn't have credentials that could be used to generate certs.

As kops-controller is going to end up relatively exposed in some configurations (not on AWS, don't worry!), I'm wary of defaulting to no authentication.  I agree that we don't want to be doing the full node auth all the time though.   I'm thinking that node bootstrap can include a client certificate for use with kops-controller - WDYT?

> What is the value of adding the complexity of GRPC?

It's specifically the need for immediate updates that made me think of GRPC, as I figure this is a good match a streaming request, and polling would be quite expensive.

johngmyers:
Once the node has bootstrapped, anything running on it should be able to use a Kubernetes service account token.

PR #14715, opened 2022-12-04
Upgrade Node Termination Handler to 1.18.1 and fix some issues

johngmyers:


upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
```diff
@@ -291,103 +298,118 @@ spec:
...
         kubernetes.io/os: linux
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: eks.amazonaws.com/compute-type
+                operator: NotIn
+                values:
+                - fargate
```
hakman, line 411: Do we really need/want this?

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
```diff
@@ -291,103 +298,118 @@ spec:
     matchLabels:
       app.kubernetes.io/name: aws-node-termination-handler
       app.kubernetes.io/instance: aws-node-termination-handler
+      app.kubernetes.io/component: daemonset
```
hakman, line 301: I don't think this can be changed:
https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#pod-selector

hakman (changes requested):

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
olemarkus: I wonder if we want a presubmit test that upgrade addons from master to PR to catch things like this.

hakman: Would be nice, but just validation and maybe some conformance. Regular tests are already taking **VERY** long time.

olemarkus: The upgrade scenario already allows disabling e2e by setting `KOPS_SKIP_E2E`.

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
```diff
@@ -130,6 +130,7 @@ spec:
             - matchExpressions:
               - key: node-role.kubernetes.io/master
                 operator: Exists
+      hostNetwork: true
```
olemarkus, line 133: This one assumes IMDSv2 hop limit = 1, right? I don't think addons in general does this. Although I personally am in the "hostNetwork is safer than max hop limit > 1" camp, I know this is contentious.

johngmyers: Non-host-network pods in an IPv6 cluster can't talk to the IPv4 metadata service.

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
johngmyers: It was in the upstream manifest and didn't seem harmful, but whatever.

hakman: Ok, let's keep it to make it easier to diff with future updates.

johngmyers:
Put the fargate thing back in and changed the log format to 2.

hakman (approved):

PR #14721, opened 2022-12-04
Change default for NTH Queue Processor mode to enabled

johngmyers:
Queue Processor mode uses fewer resources.


olemarkus:
This is a good change!

There is a resource tracking e2e test that tests the conversion from imds mode to queue mode. That one would probably have to be updated as well: tests/e2e/scenarios/addon-resource-tracking/run-test.sh

olemarkus (approved):

PR #14723, opened 2022-12-05
Refactor writing Terraform resources

johngmyers:
This is the big one; a lot of type combinations needed to be covered.

The risk is that there's some combination of types that exists in our code but isn't exercised by our integration tests. I tested a version where all nil pointers were replaced by zero values and all empty slices were replaced by a slice containing a zero value. That didn't produce any panics. So the risk would be a resource type that isn't exercised at all in the integration tests.


upup/pkg/fi/cloudup/terraform/element.go
```diff
@@ -102,8 +160,50 @@ func ToElement(item interface{}) (element, error) {
...
+			}
+		}
+		switch elemType.Kind() {
+		case reflect.String:
+			elements := make([]*terraformWriter.Literal, v.Len())
+			for i := range elements {
+				elem := v.Index(i)
+				if elem.Kind() == reflect.Pointer {
+					// TODO can these ever be nil?
+					elem = elem.Elem()
```
justinsb, line 186: Agree that panic-ing here probably makes sense.  It would be surprising to have a nil in a list of strings...

upup/pkg/fi/cloudup/terraform/element.go
```diff
@@ -71,24 +72,81 @@ func (o *object) Write(buffer *bytes.Buffer, indent int, key string) {
...
+	members []element
+}
+
+func (s *sliceObject) IsSingleValue() bool {
+	return false
+}
+
+func (s *sliceObject) Write(buffer *bytes.Buffer, indent int, key string) {
+	for _, member := range s.members {
+		member.Write(buffer, indent, key)
```
justinsb, line 90: I was surprised we didn't need `[` or `,` here.  But now I see that it's because these are slices of objects, which are just written without delimeters.  Makes sense, but a comment might be helpful here.

justinsb:
This LGTM, and we do have test coverage.

/lgtm
/approve

/hold to give others a chance to comment, but I think you can self-remove the hold after a few hours @johngmyers

johngmyers:
Argh, pushed commit to wrong branch. Removed.

hakman:
> Argh, pushed commit to wrong branch. Removed.

You sure?

johngmyers:
I'm having a bad git day, I suppose.

PR #14727, opened 2022-12-05
Move kops-controller client into its own package

justinsb:
This should allow more reuse.

### DIFF
--- a/nodeup/pkg/model/bootstrap_client.go
+++ b/nodeup/pkg/model/bootstrap_client.go
@@ -24,6 +24,7 @@ import (
 
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/bootstrap"
+	"k8s.io/kops/pkg/kopscontrollerclient"
 	"k8s.io/kops/pkg/wellknownports"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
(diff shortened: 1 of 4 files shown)


johngmyers:
I can see moving `Client` and `dial`, but I'd prefer we leave the bootstrap-specific portions of `QueryBootstrap` behind. Perhaps pass in a `req, resp interface{}`?

Possibly pass in the authenticator as well, since non-bootstrap requests will probably have a different authenticator?

johngmyers:
Actually, `req` and `resp` should use type parameters.

justinsb:
I don't think we can use type parameters today with struct methods (unless you know a trick).

I'm guessing you want to not move `nodeup.BootstrapRequest` because of versioning skew issues?  But we do have those skew issues, I think.  Although maybe if we always update the control plane first, and don't bootstrap control-plane nodes, it's not a problem in practice?

This refactor is to support updating gossip to just work through kops-controller.  For that I was thinking of using GRPC, because I think we want streaming support rather than polling.  I could just leave the existing QueryBootstrap client/logic where it is and make this a pure GRPC client - that would work too.  A bit of a code duplication for connecting, but not _too_ bad as the GRPC interfaces are a little different anyway.  Would that be better for you?

johngmyers:
I sent you a PR to remove knowledge of the request and response types. I think it would be good to make the authenticator optional and possibly passed-in per request.

I see there are two bootstrap requests to kops-controller. It would be good to consolidate them, preferably into the one that is a `Task`. But that is probably a different PR.

johngmyers:
I don't want to move BootstrapRequest because I don't want that knowledge outside of nodeup. It should be obvious that only nodeup can bootstrap or authenticate as a bootstrapping instance.

If you need GRPC features, then I suppose you need GRPC. I'd prefer to keep the bootstraprequest simple if possible. You could probably extract the connection logic into something shared by the REST and GRPC query methods.

justinsb:
Thanks @johngmyers - pulled in your change :-)

johngmyers (approved):

PR #14728, opened 2022-12-05
/etc/hosts: fix spurious messages about guard block

justinsb:


hakman (approved):

hakman:
Seems something broke.
/lgtm cancel

hakman:
@justinsb Considering we are trying to deprecate Gossip, I don't think this is worth pursuing.

justinsb:
Agreed

/close

Issue #14726, opened 2022-12-05
Unable to resolve Kubernetes cluster API URL dns during kops validate cluster


fbozic:
/kind bug

I'm trying to deploy a new GCE cluster with Cloud DNS. Validation after creation fails due to error: `unexpected error during validation: unable to resolve Kubernetes cluster API URL dns: lookup api.cluster.kops.mydomain.io` 

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Client version: 1.25.3 (git-v1.25.3)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.4", GitCommit:"872a965c6c6526caa949f0c6ac028ef7aff3fb78", GitTreeState:"clean", BuildDate:"2022-11-09T13:36:36Z", GoVersion:"go1.19.3", Compiler:"gc", Platform:"linux/amd64"}

**3. What cloud provider are you using?**
GCE

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
```
kops create -f kops/kops-dns.yaml
kops update cluster --name cluster.kops.mydomain.io --yes
kops export kubecfg --name cluster.kops.mydomain.io
kops validate cluster --name cluster.kops.mydomain.io
```

**5. What happened after the commands executed?**
Cluster validation never completes due to missing DNS record, or invalid kubeconfig file. 

**6. What did you expect to happen?**
Exported kubeconfig is valid. 

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2022-12-05T09:52:16Z"
  name: cluster.kops.mydomain.io
spec:
  api:
    loadBalancer:
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudConfig: {}
  cloudProvider: gce
  configBase: gs://kops-state-store/cluster.kops.mydomain.io
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - instanceGroup: master-europe-west3-a
      name: a
    - instanceGroup: master-europe-west3-b
      name: b
    - instanceGroup: master-europe-west3-c
      name: c
[...]

PR #14729, opened 2022-12-05
Always use load balancer address in kubeconfig

hakman:
kOps only pre-creates DNS records when there is no API load balancer. They are created later, by `dns-controller` or `external-dns`. In the mean time, the DNS name is cached as failed and operators will not be able to access the API.

Using the load balancer address instead of a DNS record should be a better approach from availability point of view, as DNS changes may be available late and take a long time to be visible for operators.

https://github.com/kubernetes/kops/blob/a9c6c4e5fa26d64d04fbf9f58aee9154090e7943/upup/pkg/fi/cloudup/dns.go#L252-L266

/cc @justinsb @johngmyers @rifelpet @olemarkus 
/hold for feedback

Closes https://github.com/kubernetes/kops/issues/14726

### DIFF
--- a/pkg/kubeconfig/create_kubecfg.go
+++ b/pkg/kubeconfig/create_kubecfg.go
@@ -35,42 +35,20 @@ const DefaultKubecfgAdminLifetime = 18 * time.Hour
 func BuildKubecfg(cluster *kops.Cluster, keyStore fi.Keystore, secretStore fi.SecretStore, cloud fi.Cloud, admin time.Duration, configUser string, internal bool, kopsStateStore string, useKopsAuthenticationPlugin bool) (*KubeconfigBuilder, error) {
 	clusterName := cluster.ObjectMeta.Name
 
-	var master string
+	var server string
 	if internal {
-		master = cluster.APIInternalName()
+		server = "https://" + cluster.APIInternalName()
 	} else {
-		master = cluster.Spec.API.PublicName
-		if master == "" {
-			master = "api." + clusterName
+		if cluster.Spec.API.PublicName != "" {
+			server = "https://" + cluster.Spec.API.PublicName
+		} else {
+			server = "https://api." + clusterName
 		}
 	}
 
-	server := "https://" + master
-
-	// We use the LoadBalancer where we know the master DNS name is otherwise unreachable
-	useELBName := false
-
-	// If the master DNS is a gossip DNS name; there's no way that name can resolve outside the cluster
-	if cluster.IsGossip() {
-		useELBName = true
-	}
-
-	// If the cluster has DNS disabled, must use the load balancer name
-	if cluster.UsesNoneDNS() {
-		useELBName = true
-	}
-
-	// If the DNS is set up as a private HostedZone, but here we have to be
-	// careful that we aren't accessing the API over DirectConnect (or a VPN).
-	// We differentiate using the heuristic that if we have an internal ELB
-	// we are likely connected directly to the VPC.
-	privateDNS := cluster.Spec.Networking.Topology != nil && cluster.Spec.Networking.Topology.DNS == kops.DNSTypePrivate
-	internalELB := cluster.Spec.API.LoadBalancer != nil && cluster.Spec.API.LoadBalancer.Type == kops.LoadBalancerTypeInternal
-	if privateDNS && !internalELB {
-		useELBName = true
-	}
-
-	if useELBName {
+	// If a load balancer exists we use it, except for when an SSL certificate is set.
+	// This should avoid a lot of pain with DNS pre-creation.
+	if cluster.Spec.API.LoadBalancer != nil && (cluster.Spec.API.LoadBalancer.SSLCertificate == "" || admin != 0) {
 		ingresses, err := cloud.GetApiIngressStatus(cluster)
 		if err != nil {
 			return nil, fmt.Errorf("error getting ingress status: %v", err)
--- a/pkg/kubeconfig/create_kubecfg_test.go
+++ b/pkg/kubeconfig/create_kubecfg_test.go
@@ -121,12 +121,14 @@ func buildMinimalCluster(clusterName string, masterPublicName string, lbCert boo
 	cluster := testutils.BuildMinimalCluster(clusterName)
 	cluster.Spec.API.PublicName = masterPublicName
 	cluster.Spec.KubernetesVersion = "1.24.0"
-	cluster.Spec.API.LoadBalancer = &kops.LoadBalancerAccessSpec{}
-	if lbCert {
-		cluster.Spec.API.LoadBalancer.SSLCertificate = "cert-arn"
-	}
-	if nlb {
-		cluster.Spec.API.LoadBalancer.Class = kops.LoadBalancerClassNetwork
+	if lbCert || nlb {
+		cluster.Spec.API.LoadBalancer = &kops.LoadBalancerAccessSpec{}
+		if lbCert {
+			cluster.Spec.API.LoadBalancer.SSLCertificate = "cert-arn"
+		}
+		if nlb {
+			cluster.Spec.API.LoadBalancer.Class = kops.LoadBalancerClassNetwork
+		}
 	}
 	return cluster
 }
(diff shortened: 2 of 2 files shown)


hakman:
> Don't we allow people to use API public name + ACM? I suppose this is equally secure, but it wouldn't make the ACM feature very useful.

I think for the ACM option we may need to revert to DNS name.

pkg/kubeconfig/create_kubecfg_test.go
```diff
@@ -330,20 +334,12 @@ func TestBuildKubecfg(t *testing.T) {
 			name: "Test Kube Config Data For Gossip cluster with admin and secondary NLB port",
```
hakman, line 334: @olemarkus Does this test make sense to you? Gossip implies cluster without DNS. You would not have a public name with Gossip and adding a cert would be pretty hard.

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -35,42 +35,20 @@ const DefaultKubecfgAdminLifetime = 18 * time.Hour
...
-	// We differentiate using the heuristic that if we have an internal ELB
-	// we are likely connected directly to the VPC.
-	privateDNS := cluster.Spec.Networking.Topology != nil && cluster.Spec.Networking.Topology.DNS == kops.DNSTypePrivate
-	internalELB := cluster.Spec.API.LoadBalancer != nil && cluster.Spec.API.LoadBalancer.Type == kops.LoadBalancerTypeInternal
-	if privateDNS && !internalELB {
-		useELBName = true
-	}
-
-	if useELBName {
+	// If a load balancer exists we use it, except for when an SSL certificate is set.
```
johngmyers, line 49: Perhaps also use the LB when a certificate is set? The later code will cause the kubeconfig to use the secondary port when an admin cert is being written.

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -35,42 +35,20 @@ const DefaultKubecfgAdminLifetime = 18 * time.Hour
...
-	if cluster.UsesNoneDNS() {
-		useELBName = true
-	}
-
-	// If the DNS is set up as a private HostedZone, but here we have to be
-	// careful that we aren't accessing the API over DirectConnect (or a VPN).
-	// We differentiate using the heuristic that if we have an internal ELB
-	// we are likely connected directly to the VPC.
-	privateDNS := cluster.Spec.Networking.Topology != nil && cluster.Spec.Networking.Topology.DNS == kops.DNSTypePrivate
-	internalELB := cluster.Spec.API.LoadBalancer != nil && cluster.Spec.API.LoadBalancer.Type == kops.LoadBalancerTypeInternal
```
johngmyers, line 68: Are we sure we want to always use internal LBs? Perhaps we should avoid using internal LBs when we have public DNS?

pkg/kubeconfig/create_kubecfg.go
hakman: Cert means that the operator set `API.PublicName` is used for it so not sure how using the LB address in kubeconfig would work.

pkg/kubeconfig/create_kubecfg.go
hakman: How would a cluster with internal LB and public DNS work. Is it possible to configure such a thing?

pkg/kubeconfig/create_kubecfg.go
johngmyers: `API.PublicName` gets included in the SAN of the kube-apiserver cert.

pkg/kubeconfig/create_kubecfg.go
johngmyers: Haven't looked at the code, but I would think access from inside would use LB, whereas access from outside would use DNS.

hakman: Internal is always DNS, unless you set `API.LoadBalancer.UseForInternalAPI`.
This code is for generating the kubeconfig for the user and for that would expect to use the cert + public DNS name.

pkg/kubeconfig/create_kubecfg.go
hakman: When the cert is set, there will be the `API.PublicName` pointing to the LB with cert and address, so LB itself will be used, but not via the LB address.

pkg/kubeconfig/create_kubecfg.go
johngmyers: API validation prevents the combination of NLB + LoadBalancerTypeInternal + UseForInternalAPI

I don't see any cross-validation between DNS topology and load balancer.

hakman: kOps offers a huge number of config possibilities, with not so much validation.
API LB internal means that the cluster has private topology?
API LB means access to API has to be done via LB. DNS would be a CNAME to the LB?

johngmyers: By my reading, if there is a loadbalancer, internal or no, then DNSModelBuilder will point external DNS to the loadbalancer. So for AWS the point is largely moot.

pkg/kubeconfig/create_kubecfg.go
johngmyers: `API.PublicName` will point to the LB, but the kubeconfig will be written to use port 8443. That port won't have a certificate; it will pass through to use the cert on kube-apiserver.

hakman: Only if `admin != 0`. In any case, this is a more advanced case.

johngmyers: I had thought this PR changed the path through that port remapping logic, but on a closer look it doesn't. It just makes it not use the load balancer address in some cases where it previously would. But the apiserver cert would have `API.PublicName` in a SAN so still work.

johngmyers (approved):

pkg/kubeconfig/create_kubecfg.go
johngmyers: ...but this might exacerbate #14726 since that's likely an `admin != 0` case.

johngmyers:
On second thought, I think this PR should at least not use DNS name in `admin != 0`-with-certificate case where it previously used load balancer name.
/approve cancel
/lgtm cancel

pkg/kubeconfig/create_kubecfg.go
```diff
@@ -35,42 +35,20 @@ const DefaultKubecfgAdminLifetime = 18 * time.Hour
...
-	if privateDNS && !internalELB {
-		useELBName = true
-	}
-
-	if useELBName {
+	// If a load balancer exists we use it, except for when an SSL certificate is set.
+	// This should avoid a lot of pain with DNS pre-creation.
+	if cluster.Spec.API.LoadBalancer != nil && cluster.Spec.API.LoadBalancer.SSLCertificate == "" {
```
johngmyers, line 51:
```suggestion
	if cluster.Spec.API.LoadBalancer != nil && (cluster.Spec.API.LoadBalancer.SSLCertificate == "" || admin != 0) {
```

pkg/kubeconfig/create_kubecfg.go
hakman: Why exacerbate?
The `SSLCertificate` is an AWS only feature, so unrelated to #14726.
We always add the LB address(es) as SANs, so there's no issue when using the kOps generated cert.

johngmyers: The issue could be during validation after cluster creation, when the public name is negatively cached but the load balancer name would work.

hakman: That's why I don't want to use public names wherever possible. LB names or IP addresses are always available.
For SSLCertificate users, it's a problem that we can't solve.

pkg/kubeconfig/create_kubecfg.go
hakman: Sorry, but don't get why `admin != 0` should be a condition...

johngmyers (approved):

johngmyers:
Looks like this broke all the IPv6 e2e tests.

johngmyers:
Hopefully fixed by #14800

PR #14731, opened 2022-12-06
scaleway: list and delete resources

Mia-Cross:
Now that the `kops create cluster` command actually creates a Scaleway cluster, this new PR should add everything that is needed for the `kops delete cluster` command to be able to destroy it, which will be useful to implement automated testing.

### DIFF
--- a/pkg/resources/scaleway/resources.go
+++ b/pkg/resources/scaleway/resources.go
@@ -1,0 +1,153 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scaleway
+
+import (
+	"k8s.io/kops/pkg/resources"
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/cloudup/scaleway"
+
+	iam "github.com/scaleway/scaleway-sdk-go/api/iam/v1alpha1"
+	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
+)
+
+const (
+	resourceTypeServer = "server"
+	resourceTypeSSHKey = "ssh-key"
+	resourceTypeVolume = "volume"
+)
+
+type listFn func(fi.Cloud, string) ([]*resources.Resource, error)
+
+func ListResources(cloud scaleway.ScwCloud, clusterInfo resources.ClusterInfo) (map[string]*resources.Resource, error) {
+	resourceTrackers := make(map[string]*resources.Resource)
+	clusterName := clusterInfo.Name
+
+	listFunctions := []listFn{
+		listServers,
+		listSSHKeys,
+		listVolumes,
+	}
+
+	for _, fn := range listFunctions {
+		rt, err := fn(cloud, clusterName)
+		if err != nil {
+			return nil, err
+		}
+		for _, t := range rt {
+			resourceTrackers[t.Type+":"+t.ID] = t
+		}
+	}
+
+	return resourceTrackers, nil
+}
+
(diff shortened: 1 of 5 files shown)


pkg/resources/scaleway/resources.go
```diff
@@ -0,0 +1,106 @@
...
+
+type listFn func(fi.Cloud, string) ([]*resources.Resource, error)
+
+func ListResources(cloud scaleway.ScwCloud, clusterInfo resources.ClusterInfo) (map[string]*resources.Resource, error) {
+	resourceTrackers := make(map[string]*resources.Resource)
+	clusterName := clusterInfo.Name
+
+	listFunctions := []listFn{
+		listServers,
+		listVolumes,
```
hakman, line 28: SSH keys?

pkg/resources/scaleway/resources.go
```diff
@@ -0,0 +1,106 @@
...
+	"k8s.io/kops/upup/pkg/fi"
+	"k8s.io/kops/upup/pkg/fi/cloudup/scaleway"
+
+	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
+)
+
+const (
+	resourceTypeDNSRecord    = "dns-record"
+	resourceTypeGateway      = "gateway"
+	resourceTypeLoadBalancer = "load-balancer"
```
hakman, line 14: Please keep only implemented ones.

cmd/kops/create_cluster.go
```diff
@@ -57,6 +54,8 @@ import (
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup"
 	"k8s.io/kops/upup/pkg/fi/utils"
+	"k8s.io/kubectl/pkg/util/i18n"
+	"k8s.io/kubectl/pkg/util/templates"
```
hakman, line 58: Unrelated to this PR, please remove.

hakman (changes requested):
Please also split the Identify part to a separate PR. From now on you should try to separate changes into different PRs when unrelated.

pkg/resources/scaleway/resources.go
Mia-Cross: I trimmed the PR to keep it shorter, I was going to just add the function to list and delete ssh keys but by doing that I noticed that our account API that I was using will soon be deprecated and is being replaced by our IAM API. I should probably make an other PR to do the migration right away before adding the ListSSHKeys function to this one.

hakman: OK

hakman:
Except for the boilerplate failure...
/lgtm

Mia-Cross:
This PR needs #14776 in order to work properly, otherwise checking if dependencies are satisfied always returns false.

hakman:
I think `Resource.Blocked` and `Resource.Blocks` are not just the types, but the exact resource that is blocked:
https://github.com/kubernetes/kops/blob/97869057129b30ea284c3ed1bdf1db36e752701d/pkg/resources/aws/routetable.go#L79-L89

Mia-Cross:
Thanks @hakman  for clearing that up, I closed the other PR and this one is ready for another review :+1:

pkg/resources/scaleway/resources.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resourceTrackers := []*resources.Resource(nil)
+	for _, server := range servers {
+		blocks := []string(nil)
+		for _, volume := range server.Volumes {
+			blocks = append(blocks, resourceTypeVolume+":"+volume.ID)
+		}
```
hakman, line 71: Seems strange that bocks/blocked is set in both volumes and servers. Was this necessary?

Mia-Cross: Not particularly but it seemed strange to me not to set it both way, I thought the code was more explicit/exhaustive this way but I can drop one of them if it's weird.

hakman: Generally you would set one that it's the easiest to set, or it makes more sense to you.
In many cases, these were not set at all, assuming that kOps will retry to delete them anyway on error. This is why you don't see that many references to blocks/blocked.

hakman:
@Mia-Cross Could you also change the commit message to something like "scaleway list and delete resources"? Would make it easier to understand what it's later.

pkg/resources/scaleway/resources.go
Mia-Cross: I only kept the `blocked by` field in volume as it was the easiest to set.

hakman (approved):

PR #14735, opened 2022-12-07
scaleway: documentation

Mia-Cross:
This PR adds documentation to the Scaleway provider.

### DIFF
--- a/docs/getting_started/scaleway.md
+++ b/docs/getting_started/scaleway.md
@@ -10,7 +10,6 @@
 
 ### Coming soon
 
-* Load-balancers
 * Scaleway DNS (to create clusters with a custom domain name)
 * Private network
 
--- a/docs/getting_started/scaleway.md
+++ b/docs/getting_started/scaleway.md
@@ -1,0 +1,76 @@
+# Getting Started with kops on Scaleway
+
+**WARNING**: Scaleway support on kOps is currently in **alpha**, which means that it is in the early stages of development and subject to change, please use with caution.
+
+## Features
+
+* Create, update and delete clusters
+* Create, edit and delete instance groups 
+* Migrating from single to multi-master
+
+### Coming soon
+
+* Load-balancers
+* Scaleway DNS (to create clusters with a custom domain name)
+* Private network
+
+### Next features to implement
+
+* `kops rolling-update`
+* BareMetal servers
+* [Terraform](https://github.com/scaleway/terraform-provider-scaleway) support
+* [Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/scaleway) support
+
+## Requirements
+
+* [kops version >= 1.26 installed](../install.md)
+* [kubectl installed](../install.md)
+* [Scaleway credentials](https://www.scaleway.com/en/docs/generate-api-keys/) : you will need at least an access key, a secret key and a project ID.
+* [S3 bucket and its credentials](https://www.scaleway.com/en/docs/storage/object/quickstart/) : the bucket's credentials may differ from the one used for provisioning the resources needed by the cluster. If you use a Scaleway bucket, you will need to prefix the bucket's name with `scw://` in the `KOPS_STATE_STORE` environment variable. For more information about buckets, see [here](../state.md)
+
+### Optional
+
+* [SSH key](https://www.scaleway.com/en/docs/configure-new-ssh-key/) : creating a cluster can be done without an SSH key, but it is required to update it. `id_rsa` and `id_ed25519` keys are supported
+
+
+## Environment Variables
+
+It is important to set the following [environment variables](https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/README.md):
+```bash
+export SCW_ACCESS_KEY="my-access-key"
+export SCW_SECRET_KEY="my-secret-key"
+export SCW_DEFAULT_PROJECT_ID="my-project-id"
+export SCW_DEFAULT_REGION="fr-par"
+export SCW_DEFAULT_ZONE="fr-par-1"
+# Configure the bucket name to store kops state
+export KOPS_STATE_STORE=scw://<bucket-name> # where <bucket-name> is the name of the bucket you set earlier
+# Scaleway Object Storage is S3 compatible so we just override some S3 configurations to talk to our bucket
+export S3_REGION=fr-par                     # or another scaleway region providing Object Storage
+export S3_ENDPOINT=s3.$S3_REGION.scw.cloud  # define provider endpoint
+export S3_ACCESS_KEY_ID="my-access-key"     # where <my-access-key> is the S3 API Access Key for your bucket
+export S3_SECRET_ACCESS_KEY="my-secret-key" # where <my-secret-key> is the S3 API Secret Key for your bucket
+# this is required since Scaleway support is currently in alpha so it is feature gated
+export KOPS_FEATURE_FLAGS="Scaleway"
+```
+
+## Creating a Single Master Cluster
+
+Note that for now you can only create a kops cluster in a single availability zone (fr-par-1, fr-par-2, fr-par-3, nl-ams-1, nl-ams-2, pl-waw-1, pl-waw-2).
+
+```bash
+# The default cluster uses ubuntu images on DEV1-M machines with cilium as Container Network Interface
+  # This creates a cluster with the gossip DNS in zone fr-par-1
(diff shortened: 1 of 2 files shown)


docs/getting_started/scaleway.md
```diff
@@ -0,0 +1,59 @@
...
+
+In the following examples, `example.com` should be replaced with the Scaleway domain you created when going through the [Requirements](#requirements).
+Note that for now you can only create a kops cluster in a single availability zone (fr-par-1, fr-par-2, fr-par-3, nl-ams-1, nl-ams-2, pl-waw-1).
+
+```bash
+# The default cluster uses ubuntu images on DEV1-M machines with cilium as Container Network Interface
+  # This creates a cluster with the gossip DNS in zone fr-par-1
+kops create cluster --cloud=scaleway --name=mycluster.k8s.local --zones=fr-par-1 --yes
+  # This creates a cluster with a custom domain name example.com in zone fr-par-1
+kops create cluster --cloud=scaleway --name=mycluster.example.com --zones=fr-par-1 --yes
```
hakman, line 46: I don't think there is DNS support at the moment.

docs/getting_started/scaleway.md
```diff
@@ -0,0 +1,59 @@
+# Getting Started with kops on Scaleway
+
+**WARNING**: Scaleway support on kOps is currently in **alpha**, which means that it is in the early stages of development and subject to change, please use with caution.
```
johngmyers, line 3: Should also add a kops_feature_table block.

docs/getting_started/scaleway.md
```diff
@@ -0,0 +1,59 @@
...
+### Optional
+* If you wish to use your own domain name for your cluster, you'll need a [Scaleway managed domain](https://www.scaleway.com/en/docs/network/domains-and-dns/quickstart). If not, you can always use the [gossip DNS](https://kops.sigs.k8s.io/gossip/) by naming your cluster with the suffix `.k8s.local`.
+
+## Environment Variables
+
+It is important to set the following [environment variables](https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/README.md):
+```bash
+export SCW_ACCESS_KEY="my-access-key"
+export SCW_SECRET_KEY="my-secret-key"
+export SCW_DEFAULT_PROJECT_ID="my-project-id"
```
johngmyers, line 22: Can this not go in the cluster spec and `kops create cluster` flag, like it does for GCP?

docs/getting_started/scaleway.md
```diff
@@ -0,0 +1,59 @@
...
+export SCW_DEFAULT_REGION="fr-par"
+export SCW_DEFAULT_ZONE="fr-par-1"
```
johngmyers, line 24: Why do these have to be defined in the environment?

docs/getting_started/scaleway.md
```diff
@@ -0,0 +1,59 @@
...
+# Configure the bucket name to store kops state
+export KOPS_STATE_STORE=scw://<bucket-name> # where <bucket-name> is the name of the bucket you set earlier
+# Scaleway Object Storage is S3 compatible so we just override some S3 configurations to talk to our bucket
+export S3_REGION=fr-par                     # or another scaleway region providing Object Storage
+export S3_ENDPOINT=s3.$S3_REGION.scw.cloud  # define provider endpoint
+export S3_ACCESS_KEY_ID="my-access-key"     # where <access-key-id> is the Spaces API Access Key for your bucket
+export S3_SECRET_ACCESS_KEY="my-secret-key" # where <secret-key> is the Spaces API Secret Key for your bucket
```
johngmyers, line 31: Can't the `scw:` code use `SCW_ACCESS_KEY`, etc?

johngmyers (changes requested):

docs/getting_started/scaleway.md
```diff
@@ -0,0 +1,59 @@
+# Getting Started with kops on Scaleway
+
+**WARNING**: Scaleway support on kOps is currently in **alpha**, which means that it is in the early stages of development and subject to change, please use with caution.
+
+## Requirements
+
+* [kops version >= 1.18 installed](../install.md)
+* [kubectl installed](../install.md)
+* [Setup your SSH key](https://www.scaleway.com/en/docs/configure-new-ssh-key/) : id_rsa and id_ed25519 keys are supported
```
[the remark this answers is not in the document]
Mia-Cross: I don't think an SSH key is necessary to create a cluster but to update I had to run `kops create secret --name <cluster name> sshpublickey admin -i ~/.ssh/id_ed25519.pub`

docs/getting_started/scaleway.md
Mia-Cross: S3 credentials and Scaleway credentials can be different so I think we should keep the option to give S3 credentials, but I could make them optional (if no S3 credentials are given, use Scaleway credentials instead). Would it be better ?

docs/getting_started/scaleway.md
Mia-Cross: Scaleway APIs use these environment variables to create the client. We can also use profiles from a config file but we have to keep in mind that not all our clients have one.

docs/getting_started/scaleway.md
```diff
@@ -0,0 +1,72 @@
...
+### Next features to implement
+
+* `kops rolling-update`
+* BareMetal servers
+* [Terraform](https://github.com/scaleway/terraform-provider-scaleway) support
+* [Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/scaleway) support
+
+## Requirements
+
+* [kops version >= 1.18 installed](../install.md)
```
hakman, line 26:
```suggestion
* [kops version >= 1.26 installed](../install.md)
```

docs/getting_started/scaleway.md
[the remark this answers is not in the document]
johngmyers: If an SSH key is not a requirement, it should not be listed under the "Requirements" section.

docs/getting_started/scaleway.md
johngmyers: Could the kOps Scaleway integration code deal with that?

docs/getting_started/scaleway.md
johngmyers: My naïve expectation would be for it to use Scaleway environment variables if configured, falling back to S3 environment variables otherwise. Then I would expect the documentation to prefer the Scaleway config file / environment variables. But there could be reasons to do it this way.

Putting credentials in environment variables is a security weakness, as that exposes them to other processes on the host.

docs/getting_started/scaleway.md
remyleone: In case you are not specifying the zone from the command line, the SDK will use those environment variables to configure the locality of the nodes.

docs/getting_started/scaleway.md
remyleone: I see two possibles ways:
Option 1: Kops uses the scaleway configuration file already used for scaleway-cli and terraform. The advantage of this method is that it offers more comfort for scaleway users that already use the scaleway dev tools ecosystem. The disadvantage is that it multiplies potential files where to look for this.

Option 2: We only use the kops.yaml configuration file. Advantage: people that are using Kops are only maintaining one configuration file. Would the kops team be open to fetching secrets from different locations? We can use the `viper` library to bring both environment variables and configuration file-defined variables.

docs/getting_started/scaleway.md
remyleone: https://github.com/kubernetes/kops/pull/14642

Mia-Cross: I opened #14787 that adds the option to try to get and use scaleway credentials if S3 credentials are not given. Reading your comment once again I noticed that it was not what you suggested, but I'm confused because it makes no sense to go further if it's the scaleway credentials that are missing since nothing else will work.

docs/getting_started/scaleway.md
Mia-Cross: @remyleone I should indeed try to get rid of these variables because `zones` is required in the command line when creating a cluster. It was simpler to pass an environment variable because sometimes I have to call the SDK from functions that do not have access to the cluster spec (where the zone is stored) but I can try again to find a way.

remyleone: I think they could be defined in either a configuration file, in an environment variable or a command line argument. viper will handle those cases :)

Mia-Cross: fixed in #14972

PR #14736, opened 2022-12-07
scaleway : nodeidentity

Mia-Cross:
This PR allows the kops-controller to identify Scaleway nodes.

### DIFF
--- a/pkg/nodeidentity/scaleway/identify.go
+++ b/pkg/nodeidentity/scaleway/identify.go
@@ -19,7 +19,6 @@ package scaleway
 import (
 	"context"
 	"fmt"
-	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/nodeidentity/scaleway/identify.go
```diff
@@ -0,0 +1,155 @@
...
+// stringKeyFunc is a string as cache key function
+func stringKeyFunc(obj interface{}) (string, error) {
+	key := obj.(*nodeidentity.Info).InstanceID
+	return key, nil
+}
+
+// getServer queries Scaleway for the server with the specified ID, returning an error if not found
+func (i *nodeIdentifier) getServer(ctx context.Context, id string) (*instance.Server, error) {
+	api := instance.NewAPI(i.client)
+	zone := os.Getenv("SCW_DEFAULT_ZONE")
```
johngmyers, line 141: Would it be possible to get the zone through configuration instead of the environment? Not a blocking issue.

Mia-Cross: I will eventually try to get rid of the environment variables as you suggested in your review of  #14735, for now I just removed it from here.

johngmyers (approved):

PR #14740, opened 2022-12-07
Make curl fail on 404 in scenario tests

olemarkus:
This should hopefully make scenario tests fail earlier than the rather obscure `exec format error` when trying to execute the xml error output.


hakman:
do we want to add some retries in the mix?

olemarkus:
Don't think so. Most (all I have seen) cases this happens, a kops version is gone from staging.

hakman (approved):

Issue #14546, opened 2022-11-14
Changes to new_cluster.go break OpenStack creation (and others?)


kciredor:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
v1.25.2

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
v1.25.x

**3. What cloud provider are you using?**
OpenStack

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
`kops create cluster` according to https://kops.sigs.k8s.io/getting_started/openstack/

**5. What happened after the commands executed?**
Panic, stack trace:
```
I1114 10:58:21.505521      36 create_cluster.go:831] Using SSH public key: /path/ssh_id_rsa.pub
W1114 10:58:22.587804      36 new_cluster.go:887] Running with masters in the same AZs; redundancy will be reduced
I1114 10:58:22.587876      36 new_cluster.go:1279]  Cloud Provider ID = openstack
panic: interface conversion: interface is nil, not openstack.OpenstackCloud

goroutine 1 [running]:
k8s.io/kops/upup/pkg/fi/cloudup.defaultMachineType({0x0?, 0x0}, 0xc000b36f00, 0xc000600800)
        k8s.io/kops/upup/pkg/fi/cloudup/populate_instancegroup_spec.go:330 +0x4b6
k8s.io/kops/upup/pkg/fi/cloudup.NewCluster(0xc0005f9200, {0x5407080, 0xc00096a180})
        k8s.io/kops/upup/pkg/fi/cloudup/new_cluster.go:423 +0x1ca5
main.RunCreateCluster({0x53f5c40, 0xc00019c008}, 0x14?, {0x53cd760, 0xc000200008}, 0xc0005f9200)
        k8s.io/kops/cmd/kops/create_cluster.go:517 +0x33f
main.NewCmdCreateCluster.func1(0xc000b34500?, {0xc0005f5440?, 0x24?, 0x24?})
        k8s.io/kops/cmd/kops/create_cluster.go:203 +0x177
github.com/spf13/cobra.(*Command).execute(0xc000b34500, {0xc0005f5200, 0x24, 0x24})
        github.com/spf13/cobra@v1.5.0/command.go:872 +0x694
github.com/spf13/cobra.(*Command).ExecuteC(0x76904a0)
[...]

PR #14742, opened 2022-12-08
OpenStack: Refactor Load Balancer builder

olemarkus:
This should make the code more easier to reason about. It should also help uncover what breaks in #14546.

An additional error case is tested after this refactor. If octavia is configured, but no router is set, no LB client would be set on the cloud. Now we return error instead. This should be caught earlier by validation, but these additional tests doesn't hurt.

/cc @zetaab


zetaab (approved):
/lgtm
/approve

looks good to me! thanks for fixing it

PR #14744, opened 2022-12-08
OpenStack: set load balancer spec before initialising client

olemarkus:


zetaab:
similar PR was merged already https://github.com/kubernetes/kops/pull/14887 

sorry did not remember that you did this already

/close

PR #14746, opened 2022-12-08
Overflow wrap the content column so content doesn't go under right sidebar and not be readable

NorseGaud:
On https://kops.sigs.k8s.io/iam_roles/ the json links go under the right sidebar. I also can't copy the whole URL because as soon as I get to the part that's under the right sidebar it freaks out and cancels my click and drag highlighting. 

This means the content is not wrapped when it's wider than the content column. Easy CSS fix for this.

Screenshot 2022-12-08 at 2 41 21 PM

### DIFF
--- a/docs/extra.css
+++ b/docs/extra.css
@@ -7,3 +7,8 @@
 .hidden {
 	display: none;
 }
+
+.md-content {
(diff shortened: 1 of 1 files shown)


hakman:
Pretty cool. Thanks!
/ok-to-test

hakman (approved):

PR #14749, opened 2022-12-09
Drop support for Canal as of Kubernetes 1.26

johngmyers:


olemarkus:
Do you happen to have a link to the deprecation notice? Quick search and I couldn't find anything.

johngmyers:
As this is as of k8s 1.26, we don't need an advance notice.

johngmyers:
Or were you talking about the upstream deprecation? That's the blue box in https://projectcalico.docs.tigera.io/getting-started/kubernetes/flannel/flannel

olemarkus:
I was thinking of the upstream notice. That blue box is more a discouragement than deprecation perhaps.

I'm all for deprecating canal since it's so infrequently used. But I know some active users do use it, so we just have to make our reasoning clear to avoid unnecessary unhappiness.

johngmyers:
That box seems to clearly express disapproval to me, so "deprecate" seems appropriate. But I could go with "disrecommended" or somesuch.

olemarkus:
/hold for feedback
Pasted a heads-up in the user channel. Not sure we want to discuss this in offer hours either? Not a must since we can alwas the drop is rather soft code-wise and we can always remove the blocking condition if we change our minds.
So thinking if there are no big reasons not to do this popping up, we can merge over the weekend.

hakman:
@lwr20 Any thoughts on this topic? It's not very clear for us what the status of Canal actually is these days. Some feedback would be appreciated. :)

lwr20:
I'd definitely want @caseydavenport's view

caseydavenport:
canal itself isn't officially deprecated, although we are recommending that users prefer pure Calico as it's generally simpler to run one component rather than two, and for most users Calico provides everything that they get from flannel plus more.

While there are still some small set of users who need canal for one reason or another, and that's enough for us to keep supporting it for now, we aren't seeing substantial adoption or any development work on the project at this point in time.

hakman:
Thanks for the explanation @caseydavenport & @lwr20, and for all your in general.

johngmyers:
To fill in more context, kOps has had Canal in experimental status since kOps 1.12. kOps has no e2e coverage of it.

johngmyers:
Deferred to 1.27 at the earliest.

PR #14753, opened 2022-12-09
Always use the latest Go patch version for builds

hakman:


johngmyers:
I'm of two minds for this. On one hand, we don't have to spend effort updating Go. On the other, it makes it more difficult to find out which version of Go was used for any given release. And the PRs to update Go let me know to update it in my IDE.

hakman:
> I'm of two minds for this. On one hand, we don't have to spend effort updating Go. On the other, it makes it more difficult to find out which version of Go was used for any given release. And the PRs to update Go let me know to update it in my IDE.

@johngmyers I mostly agree with you. The effort for updating Go on the `master` branch is minimal, but taking care of 2 release branches also is quite annoying.
Go version used for build should be pretty easy to find: `go version .build/dist/darwin/arm64/kops`.
For keeping up with releases, I prefer https://groups.google.com/g/golang-announce.

johngmyers:
If someone else wants to approve, I won't object.

hakman:
> If someone else wants to approve, I won't object.

I don't feel that strongly about it. Let's see if there's someone that like the idea on Friday, otherwise will close the PR.

hakman:
Decided to postpone this for now :).

PR #14761, opened 2022-12-10
Enable NTH capacity rebalance handling by default

johngmyers:
Since (I believe) capacity rebalance recommendations need to be explicitly enabled, this is a more sensible default. Why would someone enable rebalance recommendations if they don't want some sort of action taken on them?

### DIFF
--- a/docs/releases/1.26-NOTES.md
+++ b/docs/releases/1.26-NOTES.md
@@ -29,7 +29,7 @@ with "control-plane-". The names of groups for existing clusters are unchanged.
 
 * New clusters can more easily be configured to use Cilium in ENI mode by setting `--networking=cilium-eni`.
 
-* Node Termination Handler now defaults to Queue-Processor mode. It also now enables Scheduled Event Draining by default.
+* Node Termination Handler now defaults to Queue-Processor mode. It also now enables Scheduled Event Draining and Rebalance Draining by default.
 
 ## GCP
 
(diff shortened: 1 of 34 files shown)


olemarkus:
Rebalance recommendations in itself is not a feature one toggles on or off. But one can configure autoscaling to respond to them: https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-capacity-rebalancing.html

We have discussed the use of rebalance recommendations together with NTH in slack extensivelty, and several spot+NTH users have strongly discouraged using this feature as rebalance recommendations happens far more frequently than actual spot interruptions. So when cordoning is enabled, the cluster will be left with dozens of cordoned nodes. With draining enabled, it will lead to dozens of unnecessary interruptions although karpenter/CAS can reap the nodes in this case (given that they are fully drained).

johngmyers:
And yet in Queue Processor mode we drain on rebalance recommendations with no option to turn it off? I'm failing to see how that makes sense.

olemarkus:
In queue processor mode, we hardcoded the values for rebalance recommendations to false, I think. Enabling it in the cluster spec only took effect on the daemonset.

johngmyers:
Queue processor mode has it hardcoded to drain on rebalance recommendations. I've submitted #14771, which will cause existing queue processor users to have this turned off unless they configure otherwise.
/close

PR #14764, opened 2022-12-10
Announce deprecation of the Kube-router CNI

johngmyers:
While upstream has made releases, we have not upgraded Kube-router in over a year and a half.

It has a presubmit and is in the CNI e2e suite, but is not in the grid.

At the very least, we need a maintainer to keep it up to date.


johngmyers:
Deferred to 1.27 at the earliest.

PR #14767, opened 2022-12-11
Bump cilium to 1.13

olemarkus:
/hold until cilium 1.13 is GA

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -974,8 +974,8 @@ func validateNetworkingCilium(cluster *kops.Cluster, v *kops.CiliumNetworkingSpe
 			allErrs = append(allErrs, field.Invalid(versionFld, v.Version, "Could not parse as semantic version"))
 		}
 
-		if version.Minor != 11 && version.Patch < 5 {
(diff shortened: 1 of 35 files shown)


johngmyers:
Does Cilium support upgrading directly from 11 to 13? They haven't supported two-minor-version upgrades before.

This newer approach of only supporting one minor version of Cilium at a time means that users won't be able to skip kops minor versions. I don't think we've made that clear in our documentation.

olemarkus:
If we want multiple cilium versions support, I want those other versions in the grid though. Sometimes RBAC and other things change in minor versions, or there can be other things we do outside of cilium (e.g distro change) that needs testing.

johngmyers:
The [Cilium upgrade documentation](https://docs.cilium.io/en/stable/operations/upgrade/#version-specific-notes) only makes assertions about single-minor-version upgrades between 1.9 and 1.12. There's no information about 1.13 yet.

anthonyhaussman:
Cilium `1.13` is out now :tada:

zadjadr:
@olemarkus can we reopen this ? I believe 1.14.0 is not out, while Kops is still on 1.12.0

hakman:
Should be addressed by https://github.com/kubernetes/kops/pull/15730.

PR #14769, opened 2022-12-11
gce: Add support for clusters without DNS

hakman:
Quick version of `--dns=none` for GCE.


pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -61,28 +61,32 @@ func createPublicLB(b *APILoadBalancerBuilder, c *fi.ModelBuilderContext) error
...
 		Name:       s(b.NameForForwardingRule("api")),
 		Lifecycle:  b.Lifecycle,
-		PortRange:  s("443-443"),
+		PortRange:  s(strconv.Itoa(wellknownports.KubeAPIServer) + "-" + strconv.Itoa(wellknownports.KubeAPIServer)),
 		TargetPool: targetPool,
 		IPAddress:  ipAddress,
 		IPProtocol: "TCP",
+	})
+	if b.Cluster.UsesNoneDNS() {
+		c.AddTask(&gcetasks.ForwardingRule{
```
justinsb, line 79: We may be able to lock this down e.g. only to the nodes... We/I can try that separately though!

pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -93,8 +97,17 @@ func createPublicLB(b *APILoadBalancerBuilder, c *fi.ModelBuilderContext) error
 			Network:      network,
 			SourceRanges: b.Cluster.Spec.API.Access,
 			TargetTags:   []string{b.GCETagForRole(kops.InstanceGroupRoleControlPlane)},
-			Allowed:      []string{"tcp:443"},
+			Allowed:      []string{"tcp:" + strconv.Itoa(wellknownports.KubeAPIServer)},
 		})
+		if b.Cluster.UsesNoneDNS() {
```
justinsb, line 102: Do we need this?  This would be for e.g. humans running `kubectl`?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1410,20 +1410,32 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+						break
+					}
+				}
+				if bootConfig.APIServerIP != "" {
 					break
 				}
 			}
-			if bootConfig.APIServerIP != "" {
+		case kops.CloudProviderGCE:
+			// Use any IP address that is found (including public ones)
```
justinsb, line 1432: Nit (for a future PR): We probably should just have nodeup accept a list of IPs, probably easier than choosing!

justinsb (approved):

justinsb:
This looks good.  One quibble about whether we need to open the ports on the Load Balancers or can restrict it a bit more, but we can likely merge this and then try locking it down.

/approve
/lgtm

/hold in case you want to address why we need to open the kops-controller port to the kubernetesAPIAccessRange - I don't understand that, I don't think it necessarily includes the nodes!  But if not, no worries, we can look at this as part of follow-on

upup/pkg/fi/cloudup/apply_cluster.go
hakman: I thought about this a little. Would be nice to do the choosing on nodeup side. For sure it would be better equipped to find the best bootstrap address for itself.

pkg/model/gcemodel/api_loadbalancer.go
hakman: Not quite pretty, but with Public LB, the nodes need access to it and this was the best idea at the time except for allowing all traffic.

pkg/model/gcemodel/api_loadbalancer.go
hakman: My idea is that, we could add another internal LB for this purpose, making it secure. Though, that would diverge from the original concept and a refactor would be needed to return the IP(s) for kops-controller separately from API ones.
My goal in this PR was to get it supported as simple as possible. To the very least, this gives us a chance to add some tests and see how the output changes later.

hakman:
Thanks for the review @justinsb. I also agree about the firewall rule for kops-controller. Hopefully we can find a solution there and implement it in a separate PR.
/hold cancel

PR #14771, opened 2022-12-12
Don't drain on rebalance recommendations in SQS mode unless configured

johngmyers:


olemarkus:
Just to ensure I understand the motivation for this PR ...
This only uses the NTH configuration to add/remove specific rules, right? You don't change the templates or NTH behaviour so you don't disable draining as such, only that NTH doesn't have to consume rebalance recommendations it isn't using. The title of the PR suggests to me that it configures NTH behaviour.

johngmyers:
This PR causes draining to be disabled by not sending NTH the events. If NTH gets a rebalance event, it drains the node.

olemarkus (approved):

Issue #165, opened 2016-07-18
kops export kubecfg should be careful where it writes


justinsb:
We should document that we write to `KUBECONFIG`, and/or create an `--output` option

kenden:
`kops export kubecfg --help`
returns
`--config string                    config file (default is $HOME/.kops.yaml)`

I would expect `--config mykubeconfig.yaml` to create a kubeconfig file with that name.
`--output` sounds even clearer though.

PR #14772, opened 2022-12-12
The aws-iam-authenticator pod always fails from kops update cluster c…

nnmin-aws:
The aws-iam-authenticator pod always fails from [kops update cluster ](https://kops.sigs.k8s.io/getting_started/aws/#build-the-cluster)complaining the following error. 
This PR fix the error by adding a config.

```
Events:
  Type     Reason       Age                   From     Message
  ----     ------       ----                  ----     -------
  Warning  FailedMount  44m                   kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[output kube-api-access-rkvtd config state]: timed out waiting for the condition
  Warning  FailedMount  31m (x5 over 60m)     kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[kube-api-access-rkvtd config state output]: timed out waiting for the condition
  Warning  FailedMount  26m (x2 over 56m)     kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[state output kube-api-access-rkvtd config]: timed out waiting for the condition
  Warning  FailedMount  6m19s (x16 over 65m)  kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[config state output kube-api-access-rkvtd]: timed out waiting for the condition
  Warning  FailedMount  2m7s (x40 over 67m)   kubelet  MountVolume.SetUp failed for volume "config" : configmap "aws-iam-authenticator" not found
```

### DIFF
--- a/tests/integration/update_cluster/complex/data/aws_s3_object_complex.example.com-addons-authentication.aws-k8s-1.12_content
+++ b/tests/integration/update_cluster/complex/data/aws_s3_object_complex.example.com-addons-authentication.aws-k8s-1.12_content
@@ -134,6 +134,26 @@ subjects:
 
 ---
 
+apiVersion: v1
+data:
+  config.yaml: |
+    # a unique-per-cluster identifier to prevent replay attacks
+    # (good choices are a random token or a domain name that will be unique to your cluster)
+
+    clusterID: complex.example.com
+kind: ConfigMap
+metadata:
+  creationTimestamp: null
+  labels:
+    addon.kops.k8s.io/name: authentication.aws
+    app.kubernetes.io/managed-by: kops
+    k8s-app: aws-iam-authenticator
+    role.kubernetes.io/authentication: "1"
+  name: aws-iam-authenticator
+  namespace: kube-system
+
+---
+
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
--- a/tests/integration/update_cluster/complex/data/aws_s3_object_complex.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/complex/data/aws_s3_object_complex.example.com-addons-bootstrap_content
@@ -55,7 +55,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.12
     manifest: authentication.aws/k8s-1.12.yaml
-    manifestHash: e5087fe4df1676c9d497b29a8aee52c45ddb6e4352665c32417cec755c6ba012
+    manifestHash: c390bd8dcefed07b9a6aabaf9948c04dd4fe228c6c300f95db7e1f5226fe9578
     name: authentication.aws
     selector:
       role.kubernetes.io/authentication: "1"
--- a/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/authentication.aws/k8s-1.12.yaml.template
@@ -110,6 +110,21 @@ subjects:
   namespace: kube-system
 
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: kube-system
+  name: aws-iam-authenticator
+  labels:
+    k8s-app: aws-iam-authenticator
+data:
+  config.yaml: |
+    # a unique-per-cluster identifier to prevent replay attacks
+    # (good choices are a random token or a domain name that will be unique to your cluster)
+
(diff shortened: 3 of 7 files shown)


johngmyers:
As mentioned in #14702, this is the wrong way to address the problem. If kOps provides an empty config, then nobody will be able to authenticate to the cluster using IAM.

Instead, the mount of the configmap needs to be made `optional`.

olemarkus:
When using the CRD mode, kOps can also add the CRs. Can't the same mapping be used for the configmap?
Usually we do things like this by default, but also add a flag that allows people to manage configuration outside of the cluster spec.

Similar to this: https://github.com/kubernetes/kops/pull/14519

nnmin-aws:
> As mentioned in #14702, this is the wrong way to address the problem. If kOps provides an empty config, then nobody will be able to authenticate to the cluster using IAM.
> 
> Instead, the volume of the configmap needs to be made `optional`.

_If kOps provides an empty config, then nobody will be able to authenticate to the cluster using IAM._. ---this is not right. if CRD mode is on, then any entries in CRD still can authenticate.

johngmyers:
CRD mode is alpha, so we should not rely on it.

This change will break existing deployments using ConfigMap.

nnmin-aws:
> When using the CRD mode, kOps can also add the CRs. Can't the same mapping be used for the configmap? Usually we do things like this by default, but also add a flag that allows people to manage configuration outside of the cluster spec.
> 
> Similar to this: #14519

thanks @olemarkus for comments. Currently authenticator.aws template doesn't have a config file, is it expected?

However the template still mount the config file and it caused the error. So I added a config file here. then it will not complain the below error.


`Events:
  Type     Reason       Age                   From     Message
  ----     ------       ----                  ----     -------
  Warning  FailedMount  44m                   kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[output kube-api-access-rkvtd config state]: timed out waiting for the condition
  Warning  FailedMount  31m (x5 over 60m)     kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[kube-api-access-rkvtd config state output]: timed out waiting for the condition
  Warning  FailedMount  26m (x2 over 56m)     kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[state output kube-api-access-rkvtd config]: timed out waiting for the condition
  Warning  FailedMount  6m19s (x16 over 65m)  kubelet  Unable to attach or mount volumes: unmounted volumes=[config], unattached volumes=[config state output kube-api-access-rkvtd]: timed out waiting for the condition
  Warning  FailedMount  2m7s (x40 over 67m)   kubelet  MountVolume.SetUp failed for volume "config" : configmap "aws-iam-authenticator" not found`

nnmin-aws:
> I suggest you try to resolve this as John suggests using `optional`.

Tried and it didn't work because the following snippet in authentication.aws. It always generate the config to mount config in container. 

```
{{- if or (not .Authentication.AWS.BackendMode) (contains "MountedFile" .Authentication.AWS.BackendMode) }}
        - name: config
          mountPath: /etc/aws-iam-authenticator/
{{- end }}
```

the logic in authentication.aws template is kind of conflicting. It doesn't provide a basic config, but mounts the config everywhere. Hence I propose to give a basic config.

olemarkus:
Just to make sure we are talking about the same thing. What we are suggesting is adding `optional: true` around line 209. Also suggested you can create the configmap using the same struct as CRDs (but allow people to continue to supply their own).

rifelpet:
Here's the relevant API field: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#configmapvolumesource-v1-core

nnmin-aws:
> Here's the relevant API field: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#configmapvolumesource-v1-core

Thanks @olemarkus and @rifelpet!

I have tried adding optional: true around line 209. Then authenticator.aws will fail at line #165 because the config.yaml doesn't exist.

#line 164-166 is below.
```
{{- if or (not .Authentication.AWS.BackendMode) (contains "MountedFile" .Authentication.AWS.BackendMode) }}
{{- end }}
```

johngmyers:
Then we're back to the observation in #14702 that we would need an aws-iam-authenticator fix to handle not having a config file.

PR #14774, opened 2022-12-12
Support Google Cloud Storage for serviceAccountIssuerDiscovery

justinsb:
The main challenge here is that on S3 it's relatively easy to set
per-object ACLs, but GCS generally steers users away from that to
whole-bucket IAM permissions.

We expect GCS to be used with a separate public bucket for discovery,
and we steer users in that direction.

- OIDC / serviceAccountIssuerDiscovery: support Google Cloud Storage
- Add irsa-gce test
- Allow custom vfs.Context to exist in context.Context
- Update test output for PublicACL rename


pkg/model/issuerdiscovery.go
```diff
@@ -71,13 +74,45 @@ func (b *IssuerDiscoveryModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
 		return err
 	}
+
+	publicFileACL := fi.PtrTo(true)
+
+	discoveryStorePath := b.Cluster.Spec.ServiceAccountIssuerDiscovery.DiscoveryStore
+	vfsContext := vfs.FromContext(ctx)
+	discoveryStore, err := vfsContext.BuildVfsPath(discoveryStorePath)
+	if err != nil {
+		return fmt.Errorf("error building VFS path for %q: %w", discoveryStorePath, err)
```
johngmyers, line 84:
```suggestion
		return fmt.Errorf("building VFS path for %q: %w", discoveryStorePath, err)
```

pkg/model/issuerdiscovery.go
```diff
@@ -71,13 +74,45 @@ func (b *IssuerDiscoveryModelBuilder) Build(c *fi.ModelBuilderContext) error {
...
+	}
+
+	switch discoveryStore := discoveryStore.(type) {
+	case *vfs.GSPath:
+		isPublic, err := discoveryStore.IsBucketPublic(ctx)
+		if err != nil {
+			return fmt.Errorf("error checking if bucket was public: %w", err)
```
johngmyers, line 91:
```suggestion
			return fmt.Errorf("checking if bucket was public: %w", err)
```

upup/pkg/fi/fitasks/managedfile.go
```diff
@@ -43,8 +43,9 @@ type ManagedFile struct {
 
 	Contents fi.Resource
 
-	// Public controls whether the object is world-readable
-	Public *bool
+	// PublicACL controls whether the _object_ has an ACL which grants world-readable status.
+	// Note that the _bucket_ may itself have a grant for world-readable, that is separate.
```
johngmyers, line 47:
```suggestion
	// Note that the _bucket_ may itself have a grant for world-readable; that is separate.
```

util/pkg/vfs/gsfs.go
```diff
@@ -386,6 +386,37 @@ func (p *GSPath) Hash(a hashing.HashAlgorithm) (*hashing.Hash, error) {
 	return &hashing.Hash{Algorithm: hashing.HashAlgorithmMD5, HashValue: md5Bytes}, nil
 }
 
+func (p *GSPath) IsBucketPublic(ctx context.Context) (bool, error) {
+	policy, err := p.client.Buckets.GetIamPolicy(p.bucket).Context(ctx).Do()
+	if err != nil {
+		return false, fmt.Errorf("error getting IAM configuration for Cloud Storage bucket %q: %w", p.bucket, err)
```
johngmyers, line 392:
```suggestion
		return false, fmt.Errorf("getting IAM configuration for Cloud Storage bucket %q: %w", p.bucket, err)
```

johngmyers (changes requested):

pkg/model/components/discovery.go
```diff
@@ -61,6 +61,11 @@ func (b *DiscoveryOptionsBuilder) BuildOptions(o interface{}) error {
 				if err != nil {
 					return err
 				}
+			case *vfs.GSPath:
+				serviceAccountIssuer, err = base.GetHTTPsUrl(clusterSpec.IsIPv6Only())
```
olemarkus, line 65: Not a blocker, but we should use an interface and cast instead of using this switch.

pkg/model/issuerdiscovery.go
```diff
@@ -51,6 +53,8 @@ type oidcDiscovery struct {
 }
 
 func (b *IssuerDiscoveryModelBuilder) Build(c *fi.ModelBuilderContext) error {
+	ctx := context.TODO()
```
olemarkus: Wonder if we should start propagating context through `fi.Context` or something here.

pkg/model/issuerdiscovery.go
```diff
@@ -53,7 +52,7 @@ type oidcDiscovery struct {
 }
 
 func (b *IssuerDiscoveryModelBuilder) Build(c *fi.ModelBuilderContext) error {
-	ctx := context.TODO()
+	ctx := c.Context()
```
olemarkus, line 55: Ah :)

olemarkus (changes requested):
I am thinking the change to embed `context.Context` into builder contexts and add VFS context into `context.Context` is a separate PR. It's a very good change, but somewhat huge to just add a as a side-effect of adding GCS support for issuer discovery :)

upup/pkg/fi/fitasks/managedfile.go
hakman: Moved this to https://github.com/kubernetes/kops/pull/14775 to reduce the number of touched by this PR.

pkg/model/components/discovery.go
justinsb: I agree, I just wasn't sure yet what it should look like - and the PR was already getting big...

justinsb:
> I am thinking the change to embed context.Context into builder contexts and add VFS context into context.Context is a separate PR. It's a very good change, but somewhat huge to just add a as a side-effect of adding GCS support for issuer discovery :)

I tried doing it from the point of view of "can we add it consistently to the methods that call vfs.Context", and it is _very_ viral, so the change ended up being huge: https://github.com/kubernetes/kops/pull/14777  It also uncovered that we likely need to pass the context in to the `Render<Target>` methods, which likely overlaps with the work @johngmyers did on refactoring that interface e.g. for generics.

I do agree though, this doesn't really belong in this PR.  I will try to split out just the bits I needed here, which isn't very logically consistent, but the alternative is ... not great!

pkg/model/issuerdiscovery.go
justinsb: I did split out "just this PR" in https://github.com/kubernetes/kops/pull/14778, and I tried adding the context to every callsite of vfs.Context in https://github.com/kubernetes/kops/pull/14777.  Adding to every callsite is pretty overwhelming.

I like the idea of basically extending context.Context very much.  That feels very much within the spirit of the rules.  I might give that a go later on :-)

johngmyers:
It would be good to extend the whole-bucket IAM permissions support to AWS as well. AWS is pushing in that direction. That can be a separate PR.

justinsb:
> It would be good to extend the whole-bucket IAM permissions support to AWS as well. AWS is pushing in that direction. That can be a separate PR.

I can do that!

johngmyers:
The Terraform code assumes that serviceAccountIssuerDiscovery.discoveryStore is on the same provider as configBase. The irsa_gce puts the former on google and the latter on aws. Thus verify-terraform flags up the fact that the google files provider isn't defined.

Do we want to support the case where the two are on different clouds and using terraform or should we switch configBase to GSFS?

justinsb:
> The Terraform code assumes that serviceAccountIssuerDiscovery.discoveryStore is on the same provider as configBase. The irsa_gce puts the former on google and the latter on aws. Thus verify-terraform flags up the fact that the google files provider isn't defined.

> Do we want to support the case where the two are on different clouds and using terraform or should we switch configBase to GSFS?

I think eventually both, but auth is going to be tricky for cross-cloud in real life, so I'm starting with configBase on Google Cloud Storage.  I thought I had uploaded a PR but don't see it... it's probably at the end of the context PRs, as that was the problem (but I think we can achieve that another way now)

johngmyers (approved):

PR #14776, opened 2022-12-12
split key to resolve dependencies when deleting resources

Mia-Cross:
When I was implementing the listing and deletion of Scaleway resources (#14731), I noticed that every cloud provider builds the key of the resourceTrackers map by concatenating the type and the ID of the resource `resourceTrackers[t.Type+":"+t.ID] = t`, then in `pkg/resources/ops/delete.go`, the way dependencies completion is checked is by comparing this key to just the type, which always fails and so dependencies are never complete.
Am I the only only experiencing this problem ? If not, here's a proposition on how to fix it.

### DIFF
--- a/pkg/resources/ops/delete.go
+++ b/pkg/resources/ops/delete.go
@@ -18,6 +18,7 @@ package ops
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Mia-Cross:
It turns out that the reason that it didn't work is that I should have given the resource type AND the id of the resource blocking or blocked by the resource to delete in the resourceTracker (instead of just the resource type).

PR #14785, opened 2022-12-13
Load etcd binaries dynamically from container images

hakman:
Based on the idea discussed during office hours about using init containers to populate the required `etcd` binaries, instead of having them bundled with the `etcd-manager` images. 

This could also easily be extended to use any version specified by the user.

/cc @justinsb @johngmyers

### DIFF
--- a/pkg/model/components/etcdmanager/model.go
+++ b/pkg/model/components/etcdmanager/model.go
@@ -180,7 +180,7 @@ metadata:
 spec:
   containers:
   - name: etcd-manager
-    image: registry.k8s.io/etcdadm/etcd-manager:v3.0.20221209
+    image: registry.k8s.io/etcdadm/etcd-manager:v3.0.20230119-slim
     resources:
       requests:
         cpu: 100m
--- a/pkg/model/components/etcdmanager/model.go
+++ b/pkg/model/components/etcdmanager/model.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/blang/semver/v4"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
@@ -195,6 +196,8 @@ spec:
       name: run
     - mountPath: /etc/kubernetes/pki/etcd-manager
       name: pki
+    - mountPath: /opt
+      name: bin
   hostNetwork: true
   hostPID: true # helps with mounting volumes from inside a container
   volumes:
@@ -210,6 +213,8 @@ spec:
       path: /etc/kubernetes/pki/etcd-manager
       type: DirectoryOrCreate
     name: pki
+  - name: bin
+    emptyDir: {}
 `
 
 // buildPod creates the pod spec, based on the EtcdClusterSpec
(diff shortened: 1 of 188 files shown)


hakman:
@johngmyers Retesting won't help pass the tests. The path to the etcd binaries is hardcoded to `linux-amd64` and the test is on `arm64`. Something like https://github.com/kubernetes-sigs/etcdadm/pull/352 would be required to address this.

johngmyers:
If I read this correctly, we load all the versions as initcontainers?

I don't suppose apply_cluster can figure out which etcd version is currently in use? It would be good to avoid loading old versions that aren't needed, especially in environments which need to ensure that vulnerabilities are remediated.

hakman:
> If I read this correctly, we load all the versions as initcontainers?

Yes

> I don't suppose apply_cluster can figure out which etcd version is currently in use? It would be good to avoid loading old versions that aren't needed, especially in environments which need to ensure that vulnerabilities are remediated.

The main problem is that we don't know the version of the etcd state store on each disk and what would it take to upgrade it to the desired version.

> Admittedly we're shipping crusty, old etcd binaries now, but it would be good to improve the situation.

This can be improved for sure. For example, there is very little reason to load more than etcd 3.5.6 and etcd 3.4.13 with latest kOps versions.

johngmyers:
We'd have to somehow get the versions of etcd from the disks to apply_cluster, either by way of the state store or by having kops read them.

Probably a directory in the state store that etcd-manager can write to.

hakman:
> We'd have to somehow get the versions of etcd from the disks to apply_cluster, either by way of the state store or by having kops read them.
> Probably a directory in the state store that etcd-manager can write to.

We could write last good version in state store when etcd starts. Though, going this way will create more problems than it solves when trying to update the cluster.

I would just go with some static minor versions. One would need at most 2 etcd versions extracted and could be based on k8s version mapping:
https://github.com/kubernetes/kubernetes/blob/d2ed6d355e2e483675331715f76cf666243042ae/cmd/kubeadm/app/constants/constants.go#L468-L484

pkg/model/components/etcdmanager/model.go
```diff
@@ -212,6 +215,8 @@ spec:
       path: /etc/kubernetes/pki/etcd-manager
       type: DirectoryOrCreate
     name: pki
+  - name: bin
+    emptyDir: {}
```
justinsb, line 219: We should double check what `medium` is used here - as we don't normally run with swap, I think we would prefer this not be in RAM

pkg/model/components/etcdmanager/model.go
```diff
@@ -238,7 +243,43 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
...
+					{
+						MountPath: "/opt",
+						Name:      "bin",
+					},
+				},
+			}
+			// Add the command for copying the etcd binaries
+			var command string
+			if semver.MustParse(etcdVersion).GE(semver.MustParse("3.4.13")) {
+				command = "cp /usr/local/bin/etcd /opt/etcd-v" + etcdVersion + "/etcd && cp /usr/local/bin/etcdctl /opt/etcd-v" + etcdVersion + "/etcdctl"
```
justinsb, line 267: Nit: I'm guessing we somehow don't, but we don't need the mkdir here?  I'm also confused because this looks like it is doing the same thing as the "other" command

pkg/model/components/etcdmanager/options.go
```diff
@@ -53,23 +54,43 @@ func (b *EtcdManagerOptionsBuilder) BuildOptions(o interface{}) error {
...
+	"3.5.4":  "registry.k8s.io/etcd:3.5.4-0",
+	"3.5.6":  "registry.k8s.io/etcd:3.5.6-0",
+}
+
+func etcdSupportedVersions() []string {
+	var versions []string
+	for etcdVersion := range etcdSupportedImages {
+		versions = append(versions, etcdVersion)
+	}
+	sort.Strings(versions)
```
justinsb, line 86: I like this strategy - start with the equivalent of what we have now, cut the list down in a subsequent PR.  And yes, you're right, we need to cut this list down!

justinsb:
LGTM but maybe you can explain the copying to me in office hours :-)

pkg/model/components/etcdmanager/model.go
hakman: Newer images bundle a custom go based `cp` utility that recursively creates the necessary dirs.
https://github.com/kubernetes/kubernetes/pull/91171

pkg/model/components/etcdmanager/model.go
hakman: By default, emptyDir volumes are mounted using local ephemeral storage.
https://kubernetes.io/docs/concepts/storage/volumes/#emptydir
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#local-ephemeral-storage

pkg/model/components/etcdmanager/model.go
johngmyers: Longer term we should take the shell out of the container and use a custom Go binary for this.

johngmyers (approved):

PR #14788, opened 2022-12-15
WIP: Support public buckets for serviceAccountIssuers on S3

justinsb:
- OIDC / serviceAccountIssuerDiscovery: support Google Cloud Storage
- Add irsa-gce test
- Support mocking of GCS via custom vfs.Context
- Support public buckets for serviceAccountIssuers on S3


justinsb:
@johngmyers see the last commit; this is trying to do what you suggested on AWS where we don't need to set an S3 object ACL if the bucket is readable.

One thing I wasn't sure on was whether we should be using bucket ACLs or bucket policies.  The docs seem to point to bucket policies, but ACLs were much easier to use when I tried it...  It looks like GetBucketPolicyStatus does cover both ACLs and Policies though.

pkg/model/issuerdiscovery.go
```diff
@@ -97,7 +98,15 @@ func (b *IssuerDiscoveryModelBuilder) Build(c *fi.ModelBuilderContext) error {
 			publicFileACL = nil
 		}
 	case *vfs.S3Path:
-		// ok
+		isPublic, err := discoveryStore.IsBucketPublic(ctx)
+		if err != nil {
+			return fmt.Errorf("checking if bucket was public: %w", err)
+		}
+		if !isPublic {
+			klog.Infof("serviceAccountIssuers bucket %q is not public, will use object ACL", discoveryStore.Bucket())
```
johngmyers, line 106: Can simplify by putting `publicFileACL = fi.PtrTo(true)` here, then define it above with `var publicFileACL *bool`

johngmyers:
The Terraform code assumes that `serviceAccountIssuerDiscovery.discoveryStore` is on the same provider as `configBase`. The irsa_gce puts the latter on google and the former on aws. Thus verify-terraform flags up the fact that the google files provider isn't defined.

Do we want to support the case where the two are on different clouds and using terraform or should we switch `configBase` to GSFS?

johngmyers:
Addressed by #14799 
/close

PR #14797, opened 2022-12-17
Context threading: more wiring

justinsb:
We're aiming to use this for testing immediately and better
logging/tracing in future, but to make the changes manageable breaking
them into a smaller series that don't directly achieve much.

### DIFF
--- a/cmd/kops/create.go
+++ b/cmd/kops/create.go
@@ -198,7 +198,7 @@ func RunCreate(ctx context.Context, f *util.Factory, out io.Writer, c *CreateOpt
 				}
 
 				sshKeyArr := []byte(v.Spec.PublicKey)
-				err = sshCredentialStore.AddSSHPublicKey(sshKeyArr)
+				err = sshCredentialStore.AddSSHPublicKey(ctx, sshKeyArr)
 				if err != nil {
 					return err
 				}
--- a/cmd/kops/create.go
+++ b/cmd/kops/create.go
@@ -76,7 +76,7 @@ func NewCmdCreate(f *util.Factory, out io.Writer) *cobra.Command {
 		Example: createExample,
 		Args:    cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return RunCreate(context.TODO(), f, out, options)
+			return RunCreate(cmd.Context(), f, out, options)
 		},
 	}
 
--- a/cmd/kops/create_keypair.go
+++ b/cmd/kops/create_keypair.go
@@ -128,10 +128,10 @@ func NewCmdCreateKeypair(f *util.Factory, out io.Writer) *cobra.Command {
 			return nil
 		},
 		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-			return completeCreateKeypair(f, options, args, toComplete)
+			return completeCreateKeypair(cmd.Context(), f, options, args, toComplete)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return RunCreateKeypair(context.TODO(), f, out, options)
+			return RunCreateKeypair(cmd.Context(), f, out, options)
 		},
 	}
 
@@ -304,16 +304,15 @@ func completeKeyset(ctx context.Context, cluster *kopsapi.Cluster, clientSet sim
 	return keyset, keyStore, nil, cobra.ShellCompDirectiveNoFileComp
 }
 
-func completeCreateKeypair(f commandutils.Factory, options *CreateKeypairOptions, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+func completeCreateKeypair(ctx context.Context, f commandutils.Factory, options *CreateKeypairOptions, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 	commandutils.ConfigureKlogForCompletion()
-	ctx := context.TODO()
 
 	cluster, clientSet, completions, directive := GetClusterForCompletion(ctx, f, nil)
 	if cluster == nil {
 		return completions, directive
 	}
 
-	keyset, _, completions, directive := completeKeyset(cluster, clientSet, args, rotatableKeysetFilter)
+	keyset, _, completions, directive := completeKeyset(ctx, cluster, clientSet, args, rotatableKeysetFilter)
 	if keyset == nil {
 		return completions, directive
 	}
--- a/cmd/kops/create_keypair.go
+++ b/cmd/kops/create_keypair.go
@@ -164,7 +164,7 @@ func RunCreateKeypair(ctx context.Context, f *util.Factory, out io.Writer, optio
 	}
 
 	if options.Keyset != "all" {
-		return createKeypair(out, options, options.Keyset, keyStore)
+		return createKeypair(ctx, out, options, options.Keyset, keyStore)
 	}
 
 	keysets, err := keyStore.ListKeysets()
@@ -174,7 +174,7 @@ func RunCreateKeypair(ctx context.Context, f *util.Factory, out io.Writer, optio
 
 	for name := range keysets {
 		if rotatableKeysetFilter(name, nil) {
-			if err := createKeypair(out, options, name, keyStore); err != nil {
+			if err := createKeypair(ctx, out, options, name, keyStore); err != nil {
 				return fmt.Errorf("creating keypair for %s: %v", name, err)
 			}
 		}
@@ -183,7 +183,7 @@ func RunCreateKeypair(ctx context.Context, f *util.Factory, out io.Writer, optio
 	return nil
 }
 
-func createKeypair(out io.Writer, options *CreateKeypairOptions, name string, keyStore fi.CAStore) error {
+func createKeypair(ctx context.Context, out io.Writer, options *CreateKeypairOptions, name string, keyStore fi.CAStore) error {
 	var err error
 	var privateKey *pki.PrivateKey
 	if options.PrivateKeyPath != "" {
@@ -252,7 +252,7 @@ func createKeypair(out io.Writer, options *CreateKeypairOptions, name string, ke
 		return err
 	}
 
-	err = keyStore.StoreKeyset(name, keyset)
+	err = keyStore.StoreKeyset(ctx, name, keyset)
 	if err != nil {
 		return fmt.Errorf("error storing user provided keys %q %q: %v", options.CertPath, options.PrivateKeyPath, err)
 	}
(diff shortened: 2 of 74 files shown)


cmd/kops/create_keypair.go
```diff
@@ -268,7 +268,9 @@ func createKeypair(out io.Writer, options *CreateKeypairOptions, name string, ke
 }
 
 func completeKeyset(cluster *kopsapi.Cluster, clientSet simple.Clientset, args []string, filter func(name string, keyset *fi.Keyset) bool) (keyset *fi.Keyset, keyStore fi.CAStore, completions []string, directive cobra.ShellCompDirective) {
-	keyStore, err := clientSet.KeyStore(cluster)
+	ctx := context.TODO()
```
johngmyers, line 271: I don't think this qualifies for TODO. This is a new interactive context.

upup/pkg/fi/vfs_castore.go
```diff
@@ -59,7 +60,7 @@ func NewVFSCAStore(cluster *kops.Cluster, basedir vfs.Path) *VFSCAStore {
 }
 
 // NewVFSSSHCredentialStore creates a SSHCredentialStore backed by VFS
-func NewVFSSSHCredentialStore(cluster *kops.Cluster, basedir vfs.Path) SSHCredentialStore {
+func NewVFSSSHCredentialStore(ctx context.Context, cluster *kops.Cluster, basedir vfs.Path) SSHCredentialStore {
```
johngmyers, line 63: Why is this changing to take a Context? We're adding a Context to all methods, so why one when creating the store?

justinsb: You're right - we don't need it - removed!

cmd/kops/create_keypair.go
justinsb: So I've been using TODO at the "frontier" to limit how far we have to require context.Context in the callchains.

However, in most of these we do have the cobra Command available, which includes the context.Context, so I added another commit to add those to this PR.

pkg/client/simple/clientset.go
```diff
@@ -53,10 +53,10 @@ type Clientset interface {
 	SecretStore(cluster *kops.Cluster) (fi.SecretStore, error)
 
 	// KeyStore builds the key store for the specified cluster
-	KeyStore(cluster *kops.Cluster) (fi.CAStore, error)
+	KeyStore(ctx context.Context, cluster *kops.Cluster) (fi.CAStore, error)
```
johngmyers, line 56: This shouldn't take a Context

pkg/client/simple/clientset.go
```diff
@@ -53,10 +53,10 @@ type Clientset interface {
...
 
 	// SSHCredentialStore builds the SSHCredential store for the specified cluster
-	SSHCredentialStore(cluster *kops.Cluster) (fi.SSHCredentialStore, error)
+	SSHCredentialStore(ctx context.Context, cluster *kops.Cluster) (fi.SSHCredentialStore, error)
```
johngmyers, line 59: Shouldn't take a context

johngmyers (changes requested):

pkg/client/simple/clientset.go
justinsb: Not sure I follow.  I think all our methods are going to end up taking a context.  And this one directly needs a context for VFSClientset::KeyStore -> pkiPath -> VFSContext::BuildVfsPath

johngmyers: The methods on a CAStore might take a context, but it doesn't make much sense for the CAStore itself to hold a context.

I'm thinking a lot of the problem has to do with `vfs.Path` implementing `io.WriterTo`. Perhaps one should call a method taking a context in order to get a `io.WriterTo`.

justinsb: I think that's a valid concern, and I've been wondering if it's possible to change where we _need_ the context.  The reason it's baked into the VFSContext right now is that we are caching a GCS client, which includes the http.Client, which is what we're going to intercept/override in our tests.  Arguably the GCS Client and the http.Client should be built on every method call, but I _think_ the GCS client is not just using the standard http.Transport.  So maybe we should be caching it ... etc etc

But: I don't think we should assume that a method that takes a context.Context is going to embed the context.Context.  If the context means anything, it means that the method is non-trivial.  Anything that logs, anything that emits an otel span, or anything that calls any method that does that is going to take a context parameter.

So, while I think we can absolutely look at where we store the GCS client and friends, overall we're going to have a context on basically every exported function, and probably most internal functions also.

I suspect we'll do this for a few years until we decide that maybe goroutine-local-storage is a good idea, and then we'll hopefully just have a bulk refactoring tool to undo all the context args ;-)

johngmyers: The design pattern I'm applying is that contexts don't belong in clientsets. One should mock out the client through the context; one should mock out the client through the clientset.

The VFSContext (which is a clientset) should not be a global variable; it should be passed in explicitly. And `vfs.Path` is not just a path but something which currently inherently needs to embed a context.

justinsb: OK, that makes sense and I will look at doing that.  I think we need to pass it into every callsite to BuildVfsPath (which can then pass a context to buildGCSPath).  I think the GCS vfs.Path will end up embedding a GCS client, which I think is equivalent.

But, in order to get there, we need to wire up the context into a lot of methods.  I can either do that piecemeal in "waves" - one idea or set of functions at a time - with the hope that these smaller PRs can be more easily reviewed, or I can do a "big bang" where we change every callsite.  I feel that changing every callsite at once will be almost impossible to review though...

This particular function calls VFSClientset::KeyStore -> pkiPath -> VFSContext::BuildVfsPath anyway, so we'd need the context argument even if I did the "BuildVfsPath" wave immediately.  (And that "wave" looks like it'll be _very_ big)

justinsb: (The PR where I tried doing _some_ of this is https://github.com/kubernetes/kops/pull/14777 and it rapidly becomes all-encompassing...)

johngmyers: I don't think we should hang a `VFSContext` off of a `context.Context` and I don't think we should be adding `context.Context` parameters to enable hanging a `VFSContext` off of a `context.Context`.

It would be reasonable for a `VFSClientset` to have a `VFSContext`.

justinsb: OK, I changed this PR to _only_ wire up context.Context into more functions.

Once we have context.Context available in enough callsites, we can try it the more logically coherent way.

cmd/kops/create.go
```diff
@@ -192,13 +192,13 @@ func RunCreate(ctx context.Context, f *util.Factory, out io.Writer, c *CreateOpt
...
 
-				sshCredentialStore, err := clientset.SSHCredentialStore(cluster)
+				sshCredentialStore, err := clientset.SSHCredentialStore(ctx, cluster)
 				if err != nil {
 					return err
 				}
 
 				sshKeyArr := []byte(v.Spec.PublicKey)
-				err = sshCredentialStore.AddSSHPublicKey(sshKeyArr)
+				err = sshCredentialStore.AddSSHPublicKey(ctx, sshKeyArr)
```
johngmyers, line 201: So you have to pass a context to create a `SSHCredentialStore` and then you have to pass it again to call a receiver on that `SSHCredentialStore`. This just doesn't make sense to me.

I think you'd be better off passing a `VFSContext` to create the `simple.Clientset`. Then you'd be able to have tests create a mock `VFSContext`.

justinsb: I think the core of it is that all methods are going to take a context.

Once all methods take a context, then we can make good choices regarding where to consume values from the context.  It's hard right now to start using the context, because it's tempting to use the subset of methods that have a context.  (As you've seen, where I was trying to add context partially and use it in the wrong places).

johngmyers: One doesn't pass a context to `k8sClient.CoreV1()` or `k8sClient.CoreV1().Nodes()`. Similarly, one shouldn't pass a context to `clientset.SSHCredentialStore()`.

justinsb: I'm proposing we just add them, then we can prune if we feel they aren't needed.  Otherwise this is going to take a very long time....

cmd/kops/create.go
```diff
@@ -76,7 +76,7 @@ func NewCmdCreate(f *util.Factory, out io.Writer) *cobra.Command {
 		Example: createExample,
 		Args:    cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return RunCreate(context.TODO(), f, out, options)
+			return RunCreate(cmd.Context(), f, out, options)
```
olemarkus, line 79: Doesn't this return `nil` as we don't run with `cobraCommand.ExecuteContext`?

justinsb: Ugh I hope it returns context.TODO() or context.Background() but I'll check - thanks for pointing it out!

justinsb: I think it actually defaults to context.Background, but I switched us to call ExecuteContext so it's clearer (which might also allow us to reply to control-C more promptly in future :-) )

justinsb:
On the more contentious issue, @johngmyers has defined a nice distinction between builder methods (broadly: no side effects, no network calls, not expected to fail for good input) vs request methods (make network calls, do things, could fail unexpectedly if e.g. the network flakes).   The argument is that builder methods should not take context.Context.

It's potentially requires some deeper refactoring, but I think we/I can do it.   However, I want to be sure that we don't end up revisiting these methods, so I think we need some clear design principles - ideally that we could enforce with a linter.

I'm thinking something along these lines for the rules:
* "builder" methods would not take a context, and would be named `NewFoo`.  (NewVFSPath, NewSSHCredentialStore etc).  The prefix could also be "Build" or something else.
* "request" methods would always take a context.

An alternative rule is that "every method takes a context", but builders should be builders (no network requests etc).  We can't enforce that with a linter, but we don't enforce it today.

The problem with the stricter rules is that logging and tracing are also tied into the context (or will be in the next year or two).  I think tracing is OK, but if a builder wants to log that is awkward...  We could suggest passing in the logger there (effectively, a "limited context")

olemarkus:
I definitely think we should use klog from context. But I don't think logging from deterministic functions makes a whole lot of sense.

It's not uncommon to create loggers that are passed into builders though, so that may be an option. Not a huge fan of this pattern myself.

johngmyers:
To quibble with the rules, in `k8sClient.CoreV1().Nodes()` both `CoreV1()` and `.Nodes()` are builder methods. I don't think a prefix of "New" or "Build" is desirable in that case of a receiver on a client that configures a different client.

justinsb:
Well this is the place to iterate on the "rules" (and we don't even need to have rules, and they could be guidelines).

Another candidate I thought of:  Builder's don't get a prefix.  So Builders should start with Nouns, Requests should start with Verbs.

I don't think we can lint that reliably in general, but we might be able to catch the obvious ones.  It also gives us an answer to whether methods should include the Get prefix as they often do in Java; they do if they are a Request (GetInstance), they don't if they are a Builder (i.e. don't do GetSSHCredentialStore, do SSHCredentialStore)

justinsb:
So I tried the "no prefix" option and it seems pretty reasonable (and has the advantage of being the smallest change ... )

johngmyers (approved):

PR #14798, opened 2022-12-17
WIP: pass context into all run methods

justinsb:
- Context threading: more wiring
- More context threading
- Pass context into Run methods

### DIFF
--- a/pkg/client/simple/vfsclientset/clientset.go
+++ b/pkg/client/simple/vfsclientset/clientset.go
@@ -81,8 +81,9 @@ func (c *VFSClientset) AddonsFor(cluster *kops.Cluster) simple.AddonsClient {
 }
 
 func (c *VFSClientset) SecretStore(cluster *kops.Cluster) (fi.SecretStore, error) {
+	ctx := context.TODO()
 	if cluster.Spec.SecretStore == "" {
-		configBase, err := registry.ConfigBase(cluster)
+		configBase, err := registry.ConfigBase(ctx, cluster)
 		if err != nil {
 			return nil, err
 		}
@@ -94,8 +95,8 @@ func (c *VFSClientset) SecretStore(cluster *kops.Cluster) (fi.SecretStore, error
 	}
 }
 
-func (c *VFSClientset) KeyStore(cluster *kops.Cluster) (fi.CAStore, error) {
-	basedir, err := pkiPath(cluster)
+func (c *VFSClientset) KeyStore(ctx context.Context, cluster *kops.Cluster) (fi.CAStore, error) {
+	basedir, err := pkiPath(ctx, cluster)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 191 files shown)


justinsb:
I'd like to sequence this after the generics merge, but I think we will automatically pass the context in, so I want to get this under testing...

johngmyers:
Perhaps the context should go into the Target? I'm not seeing much of a problem keeping it in the fi.Context, since I can't think of a use case for the scopes being different.

pkg/client/simple/vfsclientset/clientset.go
```diff
@@ -64,9 +64,9 @@ func (c *VFSClientset) ListClusters(ctx context.Context, options metav1.ListOpti
 }
 
 // ConfigBaseFor implements the ConfigBaseFor method of simple.Clientset for a VFS-backed state store
-func (c *VFSClientset) ConfigBaseFor(cluster *kops.Cluster) (vfs.Path, error) {
+func (c *VFSClientset) ConfigBaseFor(ctx context.Context, cluster *kops.Cluster) (vfs.Path, error) {
 	if cluster.Spec.ConfigBase != "" {
-		return vfs.Context.BuildVfsPath(cluster.Spec.ConfigBase)
+		return vfs.FromContext(ctx).BuildVfsPath(cluster.Spec.ConfigBase)
```
johngmyers, line 69: I'm struggling to see why the `vfs.Path` should depend on a passed-in `context.Context`. And only when the `ConfigBase` is set.

Shouldn't this either take a VFSContext or hang the VFSContext off of the VFSClientset?

johngmyers: It looks like a VFSContext is closer to a Clientset than it is to a context.Context.

justinsb: > It looks like a VFSContext is closer to a Clientset than it is to a context.Context.

Yes, I think that's fair.  For example, if we had two different identities (which we don't, but if we did)... then I would expect we would have implement that with different VFSContext instances.

We could rename VFSContext - perhaps VFSScope ?

Our current use-case is for tests, so I didn't want to do _too_ much refactoring, but it prompts introducing contexts so perhaps that ship has sailed.  I do think that introducing contexts is good, even if we're not using the tracing / logging etc that it enables yet.

johngmyers: It's sort of a `VFSClient`. It's a pity we have `VFSClientSet`; perhaps that should be `VFSSimpleClientset`?

johngmyers: I don't have an objection to introducing `context.Context`s; it's just that this doesn't seem an appropriate use of them.

justinsb: That's reasonable - there's certainly an element of expedience in my thinking here.  If we add context.Context to a bunch more functions, then I think we can do it the more logical way you're suggesting.  I was trying to show some value before we added context.Context everywhere, but it sounds like we're already bought in to / resigned to the need for context.Context!

PR #14800, opened 2022-12-18
Always include load balancer domain in APIServer certificate

johngmyers:


hakman:
Thanks @johngmyers!
/lgtm
/approve
/retest

PR #14807, opened 2022-12-18
Context threading: finish up ClientSet methods

justinsb:
Builds on #14797

More context: all ClientSet methods now accept a Context
    
The next change in the series!


justinsb:
As agreed, we're going to avoid contexts on "builder" methods

/close

PR #14808, opened 2022-12-18
Context injection: Add to Render methods

justinsb:
~Builds on #14807.~ (no longer builds on this)

By passing the fi.Context into the Render methods, we can get a context.Context, and we can also work towards a standard Task signature.

Defined an interface for AWSTasks that defines the expected signature, and asserted that all types implement it (the seccond commit)

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
+++ b/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
@@ -100,6 +100,8 @@ type AutoscalingGroup struct {
 	CapacityRebalance *bool
 }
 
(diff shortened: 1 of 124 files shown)


johngmyers:
The `Target` is stored in the `fi.Context`, so I'm wondering if we can reduce the redundancy.

Do we really want the `Render` methods to have access to the `SubContext`? Have we considered putting a copy of the `context.Context` into the `Target`?

johngmyers:
This really muddies the distinction between what is within the purview of the `Task` and what is within the purview of the `Target`. I am not in favor of giving Targets access to the entire `fi.Context`.

PR #14809, opened 2022-12-18
Trim some fields from fi.Context

johngmyers:


justinsb:
This is a nice cleanup - thank you!

My only nit is that I don't love that we have to access the Cloud through .T.Cloud, because .T. isn't very self-documenting, so if we could hide this from all the tasks that would be good IMO.

A few ideas:
* Create c.Cloud() on the CloudupContext (but I think this is hard at the moment, may be easier if NodeupContext and CloudupContext actually become separate types, but that's probably a few PRs away)
* Should we aim for fi.AWSCloudupContext (I think the last commit in https://github.com/kubernetes/kops/pull/14808 is interesting because it suggests maybe we don't need reflection...).  Again, a few PRs away..
* A helper type for AWSTasks, implements the AWSTask interface, and has a `Cloud(c *fi.CloudupContext)` helper.  That might be easier implementation wise.  But arguably it pulls us away from the better(?) options above.

Anyway I think this PR is getting us closer to being able to do that, so I'm going to approve.

/approve
/lgtm

/hold for whenever you're ready @johngmyers (e.g. if you wanted to tackle the nit, but I tried and it's probably a topic for another PR)

johngmyers:
Yes, I was annoyed I couldn't embed T.

johngmyers:
I could call the field something like "Sub", but eventually most fields are going to be under it. I don't think having a helper function per field will make things better.

PR #14813, opened 2022-12-19
v1alpha3: Move some GCE-specific fields to CloudProvider.GCE

johngmyers:


pkg/model/gcemodel/context.go
```diff
@@ -142,12 +142,12 @@ func (c *GCEModelContext) NetworkingIsGCERoutes() bool {
 
 // LinkToServiceAccount returns a link to the GCE ServiceAccount object for VMs in the given role
 func (c *GCEModelContext) LinkToServiceAccount(ig *kops.InstanceGroup) *gcetasks.ServiceAccount {
-	if c.Cluster.Spec.CloudConfig.GCEServiceAccount != "" {
+	if c.Cluster.Spec.CloudProvider.GCE.ServiceAccount != "" {
 		// This is a legacy setting because the nodes & control-plane run under the same serviceaccount
-		klog.Warningf("using legacy spec.cloudConfig.gceServiceAccount=%q setting", c.Cluster.Spec.CloudConfig.GCEServiceAccount)
+		klog.Warningf("using legacy spec.cloudProvider.gce.serviceAccount=%q setting", c.Cluster.Spec.CloudProvider.GCE.ServiceAccount)
```
justinsb, line 147: Nit: until we make v1alpha3 the default, we probably should keep the v1alpha2 names?  I recognize this is a pain though...

tests/integration/update_cluster/ha_gce/data/google_compute_instance_template_master-us-test1-a-ha-gce-example-com_metadata_startup-script
```diff
@@ -123,8 +123,6 @@ ensure-install-dir
 
 cat > conf/cluster_spec.yaml << '__EOF_CLUSTER_SPEC'
 cloudConfig:
-  gcpPDCSIDriver:
```
justinsb, line 126: We might need to embed the GCE config so we know when to restart instances?

justinsb:
One question about whether we need to output the GCE config to ensure we know when to rolling-update instances...

hakman (approved):

tests/integration/update_cluster/ha_gce/data/google_compute_instance_template_master-us-test1-a-ha-gce-example-com_metadata_startup-script
johngmyers: The only thing in there is an "enabled" flag. If there were config, it would go into the addon manifest, which I don't think needs an instance restart to be applied. If we add something that requires nodeup to change its build-out behavior, I'd hope we add that setting to nodeup.Config.

johngmyers: The GCE service account goes into the ASGs and the bucket ACLs. I would think the former would cause instances to be marked for rolling update.

pkg/model/gcemodel/context.go
johngmyers: We're maintaining a mapping table in cluster_spec.md. If we don't adjust as we go, we're bound to miss a few when we change the default.

tests/integration/update_cluster/ha_gce/data/google_compute_instance_template_master-us-test1-a-ha-gce-example-com_metadata_startup-script
justinsb: OK - sgtm!  Merge and iterate :-)

pkg/model/gcemodel/context.go
justinsb: We could create a helper function that can translate between API versions (in future), even if we don't do the mapping at first...  But maybe that's good for another PR

PR #14814, opened 2022-12-19
Separate out a fi.InstallContext

johngmyers:
This clarifies that the phase that installs the kops-configuration unit needs a lot less state. It sets us up to move more state into `NodeupSubContext` without having to create fake state in the install path.


nodeup/pkg/bootstrap/install.go
```diff
@@ -44,35 +44,16 @@ func (i *Installation) Run() error {
...
-	tasks := make(map[string]fi.NodeupTask)
+	tasks := make(map[string]fi.InstallTask)
 
-	buildContext := &fi.NodeupModelBuilderContext{
+	buildContext := &fi.InstallModelBuilderContext{
 		Tasks: tasks,
 	}
 	i.Build(buildContext)
 
-	// If there is a package task, we need an update packages task
```
justinsb, line 54: I guess the idea is that we know there isn't a Package task?  That makes sense if so!

nodeup/pkg/bootstrap/install.go
```diff
@@ -165,16 +146,16 @@ func (i *Installation) buildEnvFile() *nodetasks.File {
 		sysconfig += key + "=" + value + "\n"
 	}
 
-	task := &nodetasks.File{
+	task := &nodetasks.InstallFile{File: nodetasks.File{
```
justinsb, line 149: A little awkward codewise, but I see the value!

(Would be great if these ended up as interfaces where the File task could implement both, but we have that conflict on Run... we could get rid of Run maybe, but ... not in this PR!)

justinsb:
/approve
/lgtm

Thanks - would be nice if we didn't have to wrap File and Service, but it seems like we're moving forwards here!

nodeup/pkg/bootstrap/install.go
johngmyers: There's only a File and a Service.

johngmyers:
I was actually wrapping in the other direction until I figured I was touching too much code.

PR #14821, opened 2022-12-20
kubetest2: fix bucket non-determinism on state store

justinsb:


tests/e2e/kubetest2-kops/gce/gcs.go
```diff
@@ -17,26 +17,25 @@ limitations under the License.
...
 
+// GCSBucketName constructs a bucket name to use for the state store.
+// Note that this must be deterministic across invocations of kubetest2,
+// otherwise we won't use the same bucket between runs.
 func GCSBucketName(projectID string) string {
-	var s string
 	if jobID := os.Getenv("PROW_JOB_ID"); len(jobID) >= 2 {
-		s = jobID[:2]
+		s := jobID[:2]
+		bucket := strings.Join([]string{projectID, "state", s}, "-")
```
hakman, line 33: Should we add the `kops-state` prefix here also?

tests/e2e/kubetest2-kops/gce/gcs.go
```diff
@@ -17,26 +17,25 @@ limitations under the License.
...
+		return bucket
 	} else {
-		b := make([]byte, 2)
-		rand.Read(b)
-		s = hex.EncodeToString(b)
+		bucket := "kops-state-" + projectID
```
hakman, line 36: Is this case for manual runs?

olemarkus:
I have questions about this function ...

It wants to be deterministic. It wants to have some form of uniqueness based on job name, yet the entropy (two chars from the job ID) isn't very high, so certainly conflicts.

It is only used by a hidden CLI argument, so I assume we are not using this in our setups.

PR #14824, opened 2022-12-20
Update cert-manager to v1.10.1

hakman:


upup/models/cloudup/resources/addons/certmanager.io/k8s-1.16.yaml.template
```diff
@@ -5187,20 +5202,20 @@ apiVersion: v1
...
-    app.kubernetes.io/version: "v1.9.1"
+    app.kubernetes.io/version: "v1.10.1"
 spec:
   type: ClusterIP
   ports:
   - name: https
     port: 443
     protocol: TCP
-    targetPort: 10250
+    targetPort: "https"
```
hakman, line 5218: Not quite sure why this was left to `10250` since this change has been around for some time.

upup/models/cloudup/resources/addons/certmanager.io/k8s-1.16.yaml.template
```diff
@@ -5349,26 +5371,27 @@ spec:
...
+            capabilities:
+              drop:
+              - ALL
           env:
           - name: POD_NAMESPACE
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-          resources:
-            {}
```
hakman, line 5358: I think this is safe now that we are using a new approach to applying manifests.

johngmyers (approved):

PR #14825, opened 2022-12-20
Move Cluster into CloudupSubContext

johngmyers:


upup/pkg/fi/nodeup/command.go
```diff
@@ -181,7 +181,7 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
 		return fmt.Errorf("nodeup config hash mismatch (was %q, expected %q)", got, want)
 	}
 
-	cloudProvider := api.CloudProviderID(bootConfig.CloudProvider)
+	cloudProvider := bootConfig.CloudProvider
```
justinsb, line 184: This makes me happy :-)

PR #14826, opened 2022-12-20
Update AWS test scenarios

hakman:
/cc @johngmyers @olemarkus

PR #14827, opened 2022-12-20
Fix LBC e2e test after cluster API change

olemarkus:


hakman:
Duplicate of https://github.com/kubernetes/kops/pull/14826 ?

olemarkus:
Yeah

justinsb:
Should we be versioning these field paths?  I hadn't realized we were exposing v1alpha3 to users...

olemarkus:
It's actually working on the internal unversioned API.

justinsb:
> It's actually working on the internal unversioned API.

Well that is a mirror of v1alpha3, so I hadn't realized that we were breaking the `--set` command here.

I guess we should look at targeting an explicit version for the set command, and maybe even allowing that version to be explicitly specified (e.g. --set-version=v1alpha3)

olemarkus:
I am thinking we need a strategy for multi-version support in general. Like e.g getting cluster spec in various versions.

johngmyers (approved):

johngmyers:
To follow kubectl, it'd have to be something like `kubectl create cluster.v1alpha2 ...`. But writing a `--set` that uses versioned APIs would be a pain.

johngmyers:
I suppose it would convert unversioned to versioned, do the reflection, then convert back

johngmyers:
This change was applied by #14826
/close

PR #14830, opened 2022-12-20
Move more state out of the shared fi.Context

johngmyers:


justinsb:
Looks good!

/approve
/lgtm

PR #14834, opened 2022-12-20
WIP: Allow built-in manifests to be replaced by external addons

justinsb:
For discussion, in conjunction with https://github.com/kubernetes/cloud-provider-gcp/pull/434

Allows built in manifests to be replaced by manifests supplied via the "additional objects" mechanism.  That's behind a feature flag so we can make changes.

This allows use of kOps to test components (hopefully meaning that we can run components that are tested with kOps).  It also allows sophisticated users to plug in replacement manifests for the built-ins and allows us to make progress towards a more modular world.

OTOH the approach feels a _little_ hacky - I was wondering whether we should use labels or similar - but OTOH creating the same object with a built-in and an additional object is going to result in "bad things (tm)", so it's probably a good signaling mechanism.

### DIFF
--- a/pkg/kubemanifest/containerargs.go
+++ b/pkg/kubemanifest/containerargs.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubemanifest
+
+import (
+	"fmt"
+	"strings"
+)
+
+type ContainerVisitorFunction func(container map[string]interface{}) error
+
+func (m *Object) VisitContainers(visitorFn ContainerVisitorFunction) error {
+	visitorObj := &containerVisitor{
+		visitor: visitorFn,
+	}
+	err := m.accept(visitorObj)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+type containerVisitor struct {
+	visitorBase
+	visitor ContainerVisitorFunction
+}
+
+func (m *containerVisitor) VisitMap(path []string, v map[string]interface{}) error {
+	n := len(path)
+	if n < 2 || path[n-2] != "containers" || !strings.HasPrefix(path[n-1], "[") {
+		return nil
+	}
+
+	if err := m.visitor(v); err != nil {
+		return fmt.Errorf("error visiting container %v: %w", v, err)
+	}
+
+	return nil
+}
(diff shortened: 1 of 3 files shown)


olemarkus:
We have a pattern where users can make kops behave as the addon is installed, but allow users to manually manage the manifests. Both EBS CSI driver and cert-manager use this (`managed: false`).

johngmyers:
Is the desire here to have tighter integration than `managed: false`? A big question would be how to stabilize the interface between kops and the manifest templates.

If we want to consume manifest templates from upstream, perhaps we should adopt Helm's .Values? BootstrapChannelBuilder could then map from kops config to upstream's values schema.

The way it is figuring out whether the external addon is a known kops addon is overly complicated. Instead of looking for a key resource it should just look for an annotation on the external addon.

It should be able to clean up the resources from the kops addon when that addon starts being overridden by an external one.

justinsb:
Great feedback both - thanks @olemarkus and @johngmyers !

I'd say the hope is that we would have _less_ coupling between kOps and the addons, and ideally one day kOps wouldn't have to define a configuration block in the cluster object at all, rather we'd have a set of additional "cluster addons" (they might or might not use operators, the more important thing is that kOps is more of a modular distro and less of a monolith IMO).  Testing is intrinsically monolithic - we have to test these things together - and that's why I think it would be great to work with cloud-provider-gcp (and others) to ensure they're helping us with the testing burden (and from their perspective, putting their testing energy into production scenarios).

And how we propagate values "across the fence" is indeed difficult!   I do agree it'll end up with something like helm's values.yaml, but I also suspect we'd prefer some sort of typed ClusterInfo CRD (or maybe a well-known configmap, which has the advantage that it can be mounted into env vars).  But we'd have to agree that, and there's a chicken/egg aspect here.  I do think we can break the deadlock by doing "both sides" at the same time here.

I don't love the triggering mechanism, looking for a well-known object.  If we could agree some labels maybe that would work too!   Maybe a variant of our kops `addon.kops.k8s.io/name` label that is not tied to kops.  Suggestions welcome!

Good point on cleanup/pruning also :-)

johngmyers:
The coupling is a large part of the value of kOps.

I can't imagine what it would have been like implementing IPv6 if we had loose coupling between kOps and Calico/Cilium.

johngmyers:
If we're listing problem spaces, there's the aspect of how the external addons communicate requirements to kOps. A  key problem is how to construct the IAM policies for each cloud provider.

This is a big problem with AWS addons: upstreams document an over-broad policy. The fact that it's documentation means that the policies almost never restrict to resources tagged to the cluster, as that's too complicated for humans to deal with. AWS has a project that exposes Terraform code for creating policies for the addons they manage, but despite that being able to solve the human-comprehension problem, those policies are also over-broad.

Another problem is that CNIs frequently need specific ports and protocols opened up in firewalls. This is a reaon that kopeio CNI, for example, cannot simply plug-in using the "cni" networking provider.

justinsb:
Right, I figure this is how we're going to tackle those issues :-)  We know that these addons would like better ability to test with production tooling like kOps, it should help us with our testing burden if they're testing with kOps, if we can figure out some way to communicate the contract and get the addons to define more production suitable manifests it should help us with our overall burden as well.

I don't think we'd be any worse off in kOps, we don't have to consume the manifests that the addons produce.  I think it's a win all round if we can, but if not I think that's OK too.

And yes, kopeio-networking is [on my list](https://github.com/kopeio/networking/pull/84) of early integrations!

justinsb:
Pushed #14955 which does the same thing as this but using labels to identify the overrides, as we discussed in office hours this morning.

They're pretty similar, but I think #14955 unblocks us working with the GCE CCM team (including fixing the breakage!)

justinsb:
We did this the other way in #14955 :-)

/close

PR #14837, opened 2022-12-21
v1alpha3: Move GCE-specific CloudConfig settings to GCESpec

johngmyers:
Getting close to emptying CloudConfig out.

A good question is where `ManageStorageClasses` belongs. Perhaps a `Storage` field under `ClusterSpec`? We could put `SnapshotController` there. Not sure what else, unless we want to put the CSI drivers there.


hakman:
`ManageStorageClasses` is more of a per CSI driver setting. Not sure if all of kOps implementations respect it. Would be nice to be added to each CSI driver separately to avoid misconfiguration.

olemarkus:
While I probably was the one encouraging a common config, I agree that this config makes more sense to have per CSI implementation.

johngmyers:
I still see it as common. The feature is to create a default StorageClass using whatever is appropriate for the cloud provider. (and maintain StorageClasses that used to be the default in older clusters.)

hakman:
In theory it is common. In practice, there is no requirement to be implemented and it is dependent on the cloud provider implementation in kOps.

hakman (approved):

PR #14839, opened 2022-12-21
Remove the SpecOverrideFlag feature flag

hakman:
`SpecOverrideFlag` has been around for a long time. It was enabled by default for some time also and it's not going away.

Any missing feature should be addressed in separate issues.

### DIFF
--- a/tests/e2e/scenarios/lib/common.sh
+++ b/tests/e2e/scenarios/lib/common.sh
@@ -44,7 +44,6 @@ export KOPS_BASE_URL
 export KOPS
 export CHANNELS
 
-export KOPS_FEATURE_FLAGS="SpecOverrideFlag"
 export KOPS_RUN_TOO_NEW_VERSION=1
 
 if [[ -z "${DISCOVERY_STORE-}" ]]; then 
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -389,16 +389,14 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 		return []string{"json", "yaml"}, cobra.ShellCompDirectiveNoFileComp
 	})
 
-	if featureflag.SpecOverrideFlag.Enabled() {
-		cmd.Flags().StringSliceVar(&options.Sets, "set", options.Sets, "Directly set values in the spec")
-		cmd.RegisterFlagCompletionFunc("set", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-			return nil, cobra.ShellCompDirectiveNoFileComp
-		})
-		cmd.Flags().StringSliceVar(&options.Unsets, "unset", options.Unsets, "Directly unset values in the spec")
-		cmd.RegisterFlagCompletionFunc("unset", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-			return nil, cobra.ShellCompDirectiveNoFileComp
-		})
-	}
+	cmd.Flags().StringSliceVar(&options.Sets, "set", options.Sets, "Directly set values in the spec")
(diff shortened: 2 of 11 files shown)


johngmyers:
See #11639. There was a request to make sure the syntax works well for lists and maps.

I wouldn't want to de-feature-flag it while we're changing the internal API.

olemarkus:
I think which api version it operates on needs debating. Currently, it operates on the unversioned api, which is not very stable and should not be exposed to users.

/cc @justinsb

johngmyers:
Well, I missed that it was enabled in 1.25. Had I noticed, I would have objected; the syntax needs work with respect to lists and maps.

hakman:
Let's step away from lists and maps. None of us has time to do that work. It works well for us AS IS.
We can define the desired behaviour for lists and maps and put it into issues with "help needed".
Regarding API version, that's something that we need to add for sure, but again, someone has to do it. As is, it's functional and `--set` and `--unset` behaviour will not change much in the future.

justinsb:
Honestly the biggest problem with lists and maps is finding a syntax that we like.  In one of my [toy projects[(https://github.com/justinsb/kweb/blob/main/templates/mustache/fieldpath/expression.go) I implemented up simple indexing with `[]`, but I'm not sure it's that useful for lists (it's good for maps I guess, but we don't actually have many maps here).

The tricky case is an "associative list" (a list that is really more like a map, keyed by a field like name, sometimes keyed by two fields - notably in the case of ports):

* k/k used jsonpath in kubectl `{.users[?(@.name=="e2e")].user.password}`, which is a slightly awkward syntax.
* k/k used CEL in CRD validation, but that is also a little awkward here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md#why-not-represent-associative-lists-as-maps-in-cel  CEL seems to be much more oriented towards (read-only) expressions, so the current syntax would be something like `associativeList.all(e, e.key1 == 'a' && e.key2 == 'b' && e.val == 100)`

My personal opinion is that the ideal syntax would be to treat associative lists as maps (though see the caveats above), and simply accept the caveats listed in https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md#why-not-represent-associative-lists-as-maps-in-cel (which seem acceptable).  That syntax would be something like `pod.spec.containers["nginx"].image` .  This does require some type awareness (to know that the index key is "name" here), but we know our own types.

However, I do agree with @hakman - in the absence of a strong use case for setting these values (we said we were going to wait until we needed it), and knowing that these setters are widely used and very helpful (I personally use them a lot), I think we should remove the feature flag.

Even if we don't remove the feature flag, I think we should be careful about breaking users here.  A few ideas - we could use a particular version, we could try both versions (I don't think we're likely to have any ambiguous paths), we could manually construct path parsers for the edge cases (like json.Unmarshal), or we could maintain a list of "known useful" paths that we always support (which is how this feature was first implemented).

tests/e2e/scenarios/lib/common.sh
```diff
@@ -44,7 +44,6 @@ export KOPS_BASE_URL
 export KOPS
 export CHANNELS
 
-export KOPS_FEATURE_FLAGS="SpecOverrideFlag"
```
johngmyers, line 47: I think the upgrade-ab scenario might need this.

hakman: Should we set this specifically in those tests?

johngmyers: That would be best.

johngmyers: @hakman This environment variable needs to be added to the upgrade-ab scenario.

hakman: Done. :)

johngmyers (approved):

PR #14840, opened 2022-12-21
Update resource tracking test scenario

hakman:
Not sure if this test is still useful, but updating it to not fail.

/cc @olemarkus @johngmyers


olemarkus:
We still need this one. It's how we detect that channels correctly prune stuff.

hakman:
> We still need this one. It's how we detect that channels correctly prune stuff.

It's stuck in k8s & kOps 1.21. Not very useful.

olemarkus:
It starts with those versions, yes. But it upgrades to latest kops. I think  it's fine to move the k8s target at the lowest supported k8s version.

hakman:
Missed the `kops-acquire-latest` call...

olemarkus (approved):

PR #14846, opened 2022-12-22
Allow migration from Gossip DNS to no DNS

hakman:


johngmyers:
I think we might need to change https://github.com/kubernetes/kops/blob/85026145a1be32628815509cdeadcf7d8634dc1e/upup/pkg/fi/cloudup/openstack/cloud.go#L401

hakman:
> I think we might need to change
> 
> https://github.com/kubernetes/kops/blob/85026145a1be32628815509cdeadcf7d8634dc1e/upup/pkg/fi/cloudup/openstack/cloud.go#L401

That part needs changing for sure, but `--dns=none` is not supported for OpenStack yet.

johngmyers:
I'd prefer we don't leave behind a trap for later.

hakman:
> I'd prefer we don't leave behind a trap for later.

Done in #14847

hakman:
At the moment, the GCE implementation requires access from anywhere in the internet to work.
There is no way to allow only the server addresses as they are random public IPs.

upup/pkg/fi/cloudup/template_functions.go
```diff
@@ -127,7 +127,7 @@ func (tf *TemplateFunctions) AddTo(dest template.FuncMap, secretStore fi.SecretS
 	dest["GossipDomains"] = func() []string {
 		var names []string
 
-		if cluster.IsGossip() {
+		if dns.IsGossipClusterName(cluster.Name) {
```
johngmyers, line 130: I don't understand this. If running NoneDNS in a cluster with a gossip name, `IsGossip()` is going to be false, yet all of the addons are going to enable the Gossip machinery?

hakman: This is not quite needed in general, only for the `[sig-auth] ServiceAccounts ServiceAccountIssuerDiscovery should support OIDC discovery of service account issuer [Conformance]` test to pass.

johngmyers:
Could you explain what's going on here? It looks like a cluster with a Gossip name but enabling NoneDNS will, with this change, enable Gossip in both kops-controller and coredns. So it woudn't exactly be migrating from Gossip since it will still be doing some Gossip.

hakman:
> Could you explain what's going on here? It looks like a cluster with a Gossip name but enabling NoneDNS will, with this change, enable Gossip in both kops-controller and coredns. So it woudn't exactly be migrating from Gossip since it will still be doing some Gossip.

Nothing, just some permissions to list endpoints and modify coredns configmap, which could be removed.
Most Gossip magic is done by Protokube, which is gone.

johngmyers (approved):

PR #14854, opened 2022-12-22
AWS: Use pagination for DescribeImages call

johngmyers:
New AWS feature!

### DIFF
--- a/cloudmock/aws/mockec2/images.go
+++ b/cloudmock/aws/mockec2/images.go
@@ -46,18 +46,18 @@ func (m *MockEC2) DescribeImagesWithContext(aws.Context, *ec2.DescribeImagesInpu
 	panic("Not implemented")
 }
 
-func (m *MockEC2) DescribeImages(request *ec2.DescribeImagesInput) (*ec2.DescribeImagesOutput, error) {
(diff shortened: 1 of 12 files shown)


cloudmock/aws/mockec2/images.go
```diff
@@ -46,18 +46,18 @@ func (m *MockEC2) DescribeImagesWithContext(aws.Context, *ec2.DescribeImagesInpu
 	panic("Not implemented")
 }
 
-func (m *MockEC2) DescribeImages(request *ec2.DescribeImagesInput) (*ec2.DescribeImagesOutput, error) {
+func (m *MockEC2) DescribeImagesPages(request *ec2.DescribeImagesInput, callback func(output *ec2.DescribeImagesOutput, b bool) bool) error {
```
justinsb, line 49: Can we use DescribeImagesPagesWithContext (rather than having to introduce it later)

justinsb:
Looks good.  One nit to start using the context form right away rather than introducing it later (because we have to mock it as well).  Are we actually fetching lots of images though?  Any way we could avoid doing so (e.g. can we sort by timestamp?)

johngmyers:
I don't see any way to request sorting. I don't think it's common to fetch a large number of images, but it seems possible someone might have a huge number of versions of an image.

justinsb (approved):

justinsb:
Thanks - lgtm!

/approve
/lgtm

PR #14860, opened 2022-12-23
Update release notes for kOps v1.26

hakman:


docs/releases/1.26-NOTES.md
```diff
@@ -15,6 +15,8 @@ with "control-plane-". The names of groups for existing clusters are unchanged.
 
 ## AWS
 
+* Clusters can be created without DNS or Gossip, but using the `--dns=none` flag.
```
johngmyers, line 18: Using the `--dns=none` flag, clusters can be created without DNS or Gossip.

Clusters without DNS or Gossip can be created using the `--dns=none` flag.

Instead of listing this three times, perhaps list it in the general section with "on AWS, GCP, and Hetzner".

hakman: I would prefer to keep separate. Updated the wording.

johngmyers (approved):

PR #14864, opened 2022-12-23
WIP: Don't embed VFSContext into GSPath

justinsb:


justinsb:
WIP because I wanted to confirm this is the direction we are aiming for.  Once we get a context into all the VFS Path methods, we will source the VFSContext from that context.Context (falling back to the "global" VFSContext, which we might no longer need to export ... TBD).

cc @johngmyers

johngmyers:
I think this is the wrong direction. VFSContext is misnamed: it is a clientset, not a context.

johngmyers:
Possibly `vfs.Path` should have a `VFSContext()` accessor. Then `VFSClientset` could use `basePath.VFSContext` to construct new paths.

PR #14865, opened 2022-12-23
apimachinery: Enable round-trip tests now we have two versions in play

justinsb:
Will help us catch problems where values don't convert correctly.

### DIFF
--- a/vendor/k8s.io/apimachinery/pkg/api/apitesting/codec.go
+++ b/vendor/k8s.io/apimachinery/pkg/api/apitesting/codec.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apitesting
+
+import (
+	"fmt"
+	"mime"
+	"os"
+
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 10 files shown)


justinsb:
/assign @johngmyers 

Looks like it might be catching some things :-)

johngmyers:
```
         - 				ExternalPolicies:   &nil,
        + 				ExternalPolicies:   nil,
 ```

Not really worried about `&nil`. Perhaps it should be `map` instead of `*map`?

         - 						Provider:              "Ǚ頻蓊ż$č!Ťp\u00ad屳`lʒň琐*",
        + 						Provider:              "",

         - 						LeaderElectionTimeout: s"&Duration{Duration:-1869857h12m53.878519543s,}",
        + 						LeaderElectionTimeout: nil,

Unused fields deliberately omitted from new API.

         - 					Hosts:        []string{},
        + 					Hosts:        nil,

Not worried about the distinction. Don't know why the test even cares.

         - 					NodeLabels:                  map[string]string{"z": `H騦TǟBȓŘ;硲#+ĳ楀\K~鿣`},
        + 					NodeLabels:                  nil,

Not admin configurable, thus ignored, thus deliberately omitted from new API.

         - 					LyftVPC:   &kops.LyftVPCNetworkingSpec{},
        + 					LyftVPC:   nil,

         - 					Legacy:                 true,
        + 					Legacy:                 false,

If set causes API validation error, thus deliberately omitted from new API.

justinsb:
Right, I think the printing might be over-printing here, so not sure that all the differences that are printed "count" as failures.

I think this is what we likely need to figure out: "Unused fields deliberately omitted from new API."  I know what we've done elsewhere is to map an unused field `spec.foo` -> `spec.deprecatedFoo`.  But at some point those are dropped, and I think if we are happy with dropping values of `foo` (which I think we are), then we don't need to introduce `deprecatedFoo`.  What I haven't found yet is how to tell the fuzzer that some fields are "special".  I think having a table of these fields in our code would end up being super-useful - not sure if we would generate docs from them, but I suspect once we have it we'll find more uses (e.g. automatically logging warnings in v1alpha2 if the field is used?)

johngmyers:
There are some fields which, if used, cause an API validation error. There is a much larger set of others which have been ignored for a long time.

PR #14879, opened 2022-12-26
aws: Set IMDS defaults for existing clusters

hakman:
IMDSv2 has been enabled for newly created clusters since k8s 1.22, which will be the minimum supported version in kOps 1.27. I think it's time to enable IMDSv2 by default, unless configured otherwise.

/cc @olemarkus @johngmyers

### DIFF
--- a/pkg/model/awsmodel/autoscalinggroup.go
+++ b/pkg/model/awsmodel/autoscalinggroup.go
@@ -181,7 +181,7 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.CloudupMode
 		Lifecycle:                    b.Lifecycle,
 		CPUCredits:                   fi.PtrTo(fi.ValueOf(ig.Spec.CPUCredits)),
 		HTTPPutResponseHopLimit:      fi.PtrTo(int64(1)),
-		HTTPTokens:                   fi.PtrTo(ec2.LaunchTemplateHttpTokensStateOptional),
+		HTTPTokens:                   fi.PtrTo(ec2.LaunchTemplateHttpTokensStateRequired),
 		HTTPProtocolIPv6:             fi.PtrTo(ec2.LaunchTemplateInstanceMetadataProtocolIpv6Disabled),
 		IAMInstanceProfile:           link,
 		ImageID:                      fi.PtrTo(ig.Spec.Image),
@@ -281,10 +281,14 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.CloudupMode
 
 	if ig.Spec.InstanceMetadata != nil && ig.Spec.InstanceMetadata.HTTPPutResponseHopLimit != nil {
 		lt.HTTPPutResponseHopLimit = ig.Spec.InstanceMetadata.HTTPPutResponseHopLimit
+	} else if ig.IsControlPlane() && (b.Cluster.IsKubernetesLT("1.26") || !b.UseServiceAccountExternalPermissions()) {
+		lt.HTTPPutResponseHopLimit = fi.PtrTo[int64](3)
 	}
 
 	if ig.Spec.InstanceMetadata != nil && ig.Spec.InstanceMetadata.HTTPTokens != nil {
 		lt.HTTPTokens = ig.Spec.InstanceMetadata.HTTPTokens
+	} else if b.IsKubernetesLT("1.27") {
+		lt.HTTPTokens = fi.PtrTo(ec2.LaunchTemplateHttpTokensStateOptional)
 	}
 
 	if rootVolumeType == ec2.VolumeTypeIo1 || rootVolumeType == ec2.VolumeTypeIo2 {
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -877,14 +877,16 @@ func setupControlPlane(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubne
 				g.Spec.Zones = []string{zone}
 			}
 
-			if cloudProvider == api.CloudProviderAWS {
-				g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
-					HTTPPutResponseHopLimit: fi.PtrTo(int64(3)),
-					HTTPTokens:              fi.PtrTo("required"),
+			if cluster.IsKubernetesLT("1.27") {
+				if cloudProvider == api.CloudProviderAWS {
+					g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+						HTTPPutResponseHopLimit: fi.PtrTo(int64(3)),
+						HTTPTokens:              fi.PtrTo("required"),
+					}
 				}
-			}
-			if cluster.IsKubernetesGTE("1.26") && fi.ValueOf(cluster.Spec.IAM.UseServiceAccountExternalPermissions) {
-				g.Spec.InstanceMetadata.HTTPPutResponseHopLimit = fi.PtrTo(int64(1))
+				if cluster.IsKubernetesGTE("1.26") && fi.ValueOf(cluster.Spec.IAM.UseServiceAccountExternalPermissions) {
+					g.Spec.InstanceMetadata.HTTPPutResponseHopLimit = fi.PtrTo(int64(1))
+				}
 			}
 
 			g.Spec.MachineType = opt.ControlPlaneSize
@@ -1006,10 +1008,12 @@ func setupNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap ma
 			g.Spec.Zones = []string{zone}
 		}
 
-		if cloudProvider == api.CloudProviderAWS {
-			g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
-				HTTPPutResponseHopLimit: fi.PtrTo(int64(1)),
-				HTTPTokens:              fi.PtrTo("required"),
+		if cluster.IsKubernetesLT("1.27") {
+			if cloudProvider == api.CloudProviderAWS {
+				g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+					HTTPPutResponseHopLimit: fi.PtrTo(int64(1)),
+					HTTPTokens:              fi.PtrTo("required"),
+				}
 			}
 		}
 
@@ -1028,9 +1032,11 @@ func setupKarpenterNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSub
 	g.Spec.Manager = api.InstanceManagerKarpenter
 	g.ObjectMeta.Name = "nodes"
 
-	g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
-		HTTPPutResponseHopLimit: fi.PtrTo(int64(1)),
-		HTTPTokens:              fi.PtrTo("required"),
+	if cluster.IsKubernetesLT("1.27") {
+		g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
+			HTTPPutResponseHopLimit: fi.PtrTo(int64(1)),
+			HTTPTokens:              fi.PtrTo("required"),
+		}
 	}
 
 	return []*api.InstanceGroup{g}, nil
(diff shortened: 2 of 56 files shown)


johngmyers:
I think I'm leaning towards soft-changing the default. Keep the default the same for k8s <= 1.26, make the default IMDS hop 1 for k8s >= 1.27.

hakman:
> I think I'm leaning towards soft-changing the default. Keep the default the same for k8s <= 1.26, make the default IMDS hop 1 for k8s >= 1.27.

So, make it default to optional until k8s 1.26?

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -281,10 +281,14 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.CloudupMode
 
 	if ig.Spec.InstanceMetadata != nil && ig.Spec.InstanceMetadata.HTTPPutResponseHopLimit != nil {
 		lt.HTTPPutResponseHopLimit = ig.Spec.InstanceMetadata.HTTPPutResponseHopLimit
+	} else if ig.IsControlPlane() && (b.Cluster.IsKubernetesLT("1.27") || !(b.Cluster.Spec.IAM != nil && fi.ValueOf(b.Cluster.Spec.IAM.UseServiceAccountExternalPermissions))) {
```
olemarkus, line 284: Can we move this to a function? My head hurts a bit from trying to read this expression :)

hakman: Is there a good reason to not use IRSA for k8s < 1.27 (see https://github.com/kubernetes/kops/pull/14572)?

hakman: I think latest change makes it a little more readable. Even if it were a function it would still be hard to even name it. :))

pkg/model/awsmodel/autoscalinggroup.go
```diff
@@ -281,10 +281,14 @@ func (b *AutoscalingGroupModelBuilder) buildLaunchTemplateTask(c *fi.CloudupMode
 
 	if ig.Spec.InstanceMetadata != nil && ig.Spec.InstanceMetadata.HTTPPutResponseHopLimit != nil {
 		lt.HTTPPutResponseHopLimit = ig.Spec.InstanceMetadata.HTTPPutResponseHopLimit
+	} else if ig.IsControlPlane() && (b.Cluster.IsKubernetesLT("1.27") || !b.UseServiceAccountExternalPermissions()) {
+		lt.HTTPPutResponseHopLimit = fi.PtrTo[int64](3)
```
johngmyers, line 285: Is there any reason we can't set this to 1 for k8s >= 1.27? Possibly in a follow-up PR to be done in 1.27.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -877,16 +877,6 @@ func setupControlPlane(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubne
...
 			}
 
-			if cloudProvider == api.CloudProviderAWS {
-				g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
-					HTTPPutResponseHopLimit: fi.PtrTo(int64(3)),
-					HTTPTokens:              fi.PtrTo("required"),
-				}
-			}
-			if cluster.IsKubernetesGTE("1.26") && fi.ValueOf(cluster.Spec.IAM.UseServiceAccountExternalPermissions) {
-				g.Spec.InstanceMetadata.HTTPPutResponseHopLimit = fi.PtrTo(int64(1))
```
johngmyers, line 887: I think we need to leave this in for k8s == 1.26.x && fi.ValueOf(cluster.Spec.IAM.UseServiceAccountExternalPermissions)

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -877,16 +877,6 @@ func setupControlPlane(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubne
 				g.Spec.Zones = []string{zone}
 			}
 
-			if cloudProvider == api.CloudProviderAWS {
```
johngmyers, line 880: I think we need to leave this in for k8s < 1.27

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1006,13 +996,6 @@ func setupNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetMap ma
 			g.Spec.Zones = []string{zone}
 		}
 
-		if cloudProvider == api.CloudProviderAWS {
```
johngmyers, line 1009: I think we need to leave this in for k8s < 1.27

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1028,11 +1011,6 @@ func setupKarpenterNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSub
 	g.Spec.Manager = api.InstanceManagerKarpenter
 	g.ObjectMeta.Name = "nodes"
 
-	g.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
```
johngmyers, line 1031: Leave for k8s < 1.27

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1073,13 +1051,6 @@ func setupAPIServers(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetM
 			g.Spec.Zones = []string{zone}
 		}
 
-		if cloudProvider == api.CloudProviderAWS {
```
johngmyers, line 1076: leave for k8s < 1.27

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1275,11 +1246,6 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
 				bastionGroup.Spec.Zones = allZones.List()
 			}
 
-			bastionGroup.Spec.InstanceMetadata = &api.InstanceMetadataOptions{
```
johngmyers, line 1278: Leave for k8s < 1.27

johngmyers (changes requested):

pkg/model/awsmodel/autoscalinggroup.go
hakman: It is the condition from https://github.com/kubernetes/kops/pull/14572.

upup/pkg/fi/cloudup/new_cluster.go
hakman: See the comment ^^^.

upup/pkg/fi/cloudup/new_cluster.go
hakman: It is the default and has been moved to model.

pkg/model/awsmodel/autoscalinggroup.go
johngmyers: I'm asking why we can't set it to 1 for control planes not using IRSA.

upup/pkg/fi/cloudup/new_cluster.go
johngmyers: Which comment were you referring to?

upup/pkg/fi/cloudup/new_cluster.go
johngmyers: The only thing in model I see is the new default for k8s >= 1.27. When creating new clusters with k8s < 1.27 we should still have the behavior of explicitly setting these fields.

upup/pkg/fi/cloudup/new_cluster.go
hakman: This logic is replaced by the one where you asked:
> Is there any reason we can't set this to 1 for k8s >= 1.27? Possibly in a follow-up PR to be done in 1.27.

upup/pkg/fi/cloudup/new_cluster.go
hakman: Model has defaults and those take care of keeping the old behaviour:
```go
	lt := &awstasks.LaunchTemplate{
		HTTPPutResponseHopLimit:      fi.PtrTo(int64(1)),
		HTTPTokens:                   fi.PtrTo(ec2.LaunchTemplateHttpTokensStateRequired),
	}
```

pkg/model/awsmodel/autoscalinggroup.go
hakman: https://github.com/kubernetes/kops/pull/10702 was the motivation for setting hop limit to 3 on control plane.

johngmyers: I think it would be better to make such pods be host-network. I don't think this issue blocks this PR, but we should make a decision before we release 1.27.

upup/pkg/fi/cloudup/new_cluster.go
johngmyers: I asked to change the default for no-supplied-value. I didn't ask to change the default for new clusters.

johngmyers: The "this to 1" applies to control-plane nodes in non-IRSA clusters.

upup/pkg/fi/cloudup/new_cluster.go
johngmyers: New clusters with k8s < 1.27 created by kOps >= 1.27 should have IMDSv2 hop limit 1, just like new clusters with k8s < 1.27 created by kOps 1.26.

hakman: They do, just diferent approach.

upup/pkg/fi/cloudup/new_cluster.go
hakman: It is an inverted condition. As the default is "1" for ALL, the condition to switch to 3 for control plane without IRSA had to be updated.

pkg/model/awsmodel/autoscalinggroup.go
olemarkus: Ideally, we'd have some admission controller that ensures only kops-stuff can run on the control plane. But this is a very tricky thing to do. With PSS, it is at least easier to ensure only "privileged" namespaces can run with hostNetworking on, which in turn is a form of safety mechanism. So I am inclinded to agree with hostNetworking being better than max hops > 1.

hakman: I agree too, just unrelated to this PR. The only purpose was to not pollute the IG config with these settings and hide them in `awsmodel` as defaults, not duplicating code.
Another topic is that it's already complicated to keep track of defaults or changes between versions. Would be nice to not have https://github.com/kubernetes/kops/pull/10702 in 1.26, this in 1.27 and so on...

johngmyers, line 285: Previously, clusters created long ago and successively upgraded to k8s 1.26 would have the hop limit defaulting to 1, but with HTTP tokens optional. With this change, non-IRSA control-plane nodes in such clusters will have the hop limit set to 3 (but HTTP tokens still optional).

This is technically a change in behavior for k8s < 1.27, though it might not matter.

upup/pkg/fi/cloudup/new_cluster.go
johngmyers: I can now see how we were previously defaulting the hop limit to 1. But it looks like there's an issue with requiring tokens in new clusters with k8s < 1.27.

New clusters with k8s < 1.27 created by kOps >= 1.27 should have tokens required, just like new clusters with k8s < 1.27 created by kOps 1.26.

hakman: I would like ALL clusters to have tokens required. It is a change in default behaviour, but it's something that we pushed for all new clusters since kOps 1.22. It's been a long time to accept this behaviour. Operators should move to required tokens sooner rather than later.

johngmyers:
I think we need a higher-bandwidth discussion.
/kind office-hours
/hold for office-hours

hakman:
@johngmyers I think this is ready for another pass.

johngmyers (approved):
If we backport #14913 to 1.26, this logic could be simpler.
/hold for consideration of #14913

hakman:
@johngmyers Could you remove the hold? We can improve this when we merge https://github.com/kubernetes/kops/pull/14913.

PR #14880, opened 2022-12-26
Use short service name with discovery labels

hakman:
Closes https://github.com/kubernetes/kops/issues/14759
Ref https://github.com/kubernetes/kops/issues/14877

/cc @olemarkus @johngmyers

### DIFF
--- a/cmd/kops-controller/controllers/hosts_controller.go
+++ b/cmd/kops-controller/controllers/hosts_controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
+	"k8s.io/kops/cmd/kops-controller/pkg/config"
 	"k8s.io/kops/pkg/apis/kops"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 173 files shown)


johngmyers:
Could this perhaps break upgrade of a cilium-etcd cluster?

hakman:
> Could this perhaps break upgrade of a cilium-etcd cluster?

You mean with dedicated API servers? That is still experimental, so no expectations there. Also, the current DNS names for the etcd clusters are not quite right without the `.internal.` part.

johngmyers:
I was wondering about cilium-etcd without apiserver nodes. We don't appear to have an upgrade test for that. I haven't researched the details; I'd like @olemarkus to give his opinion on the upgrade implications of changing those DNS names.

hakman:
The DNS names for the etcd clusters are only used when the apiserver nodes feature is enabled. Other than that, DNS names remain the same, so any upgrades should work just fine.
https://github.com/kubernetes/kops/blob/4cf5418eb83f5aa29e22caae1bc39fb1e7b9da7b/pkg/model/components/kopscontroller/template_functions.go#L87-L104

johngmyers (approved):
Okay, it's fine for apiserver-node clusters to have a disruptive upgrade

PR #14884, opened 2022-12-26
AWS: set IMDS hop limit 1 on all new clusters

johngmyers:
Since this is for new clusters, we don't have to be particular about k8s versions.

The default hop limit if not specified in the cluster spec has been 1 ever since the field was added.

### DIFF
--- a/docs/instance_groups.md
+++ b/docs/instance_groups.md
@@ -47,14 +47,14 @@ spec:
 
 ## instanceMetadata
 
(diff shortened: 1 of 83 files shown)


hakman (approved):

hakman:
Docs and release notes need a small update:
* https://kops.sigs.k8s.io/instance_groups/#instancemetadata
* https://kops.sigs.k8s.io/releases/1.27-notes/#aws

/lgtm cancel
/hold cancel

PR #14893, opened 2022-12-27
Don't expose v1alpha3 API

johngmyers:
It's not ready for 1.26.

### DIFF
--- a/pkg/apis/kops/install/install.go
+++ b/pkg/apis/kops/install/install.go
@@ -23,12 +23,10 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/apis/kops/v1alpha2"
-	"k8s.io/kops/pkg/apis/kops/v1alpha3"
 )
 
 func Install(scheme *runtime.Scheme) {
 	utilruntime.Must(kops.AddToScheme(scheme))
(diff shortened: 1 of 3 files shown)


hakman:
I think in the previous releases we left it around, just not made it the default yet.
/cc @olemarkus @justinsb

olemarkus:
I believe we did not remove it in 1.25, yes.

johngmyers:
It escaped into 1.25. The 1.26 one is incompatible. Technically we should rename the 1.26 one to v1alpha4 and master branch to v1alpha5. Or we can remove from 1.26 and save some hassle.

justinsb:
If you're OK to remove it from 1.26 (because it constrains your ability to iterate), then I'm in favor of removing it.  We could also feature flag it if that's easier (?).

I don't think we're going to break anyone by removing it, because I don't think we've suggested to anyone to use it and it's not the default / not documented, and we do risk breaking people by leaving it in when we aren't maintaining compatability.

My vote: remove, easy enough to add back if someone is using it, but if they are we should discuss with them that they probably should be sticking with alpha2 for the moment.

johngmyers:
Anyone who used v1alpha3 in 1.25 is going to be broken. We didn't maintain compatibility.

I would prefer we not leave a similar attractive nuisance in 1.26. APIMachinery rules allow removal of alpha apiversions with no deprecation period.

hakman (approved):

PR #14894, opened 2022-12-28
v1alpha3: Move most OIDC settings to authentication.oidc

johngmyers:


pkg/apis/kops/componentconfig.go
```diff
@@ -366,28 +366,28 @@ type KubeAPIServerConfig struct {
 	// OIDCUsernameClaim is the OpenID claim to use as the user name.
 	// Note that claims other than the default ('sub') is not guaranteed to be
 	// unique and immutable.
-	OIDCUsernameClaim *string `json:"oidcUsernameClaim,omitempty" flag:"oidc-username-claim"`
+	OIDCUsernameClaim *string `json:"-" flag:"oidc-username-claim"`
```
justinsb, line 369: Is this just because we don't want to be serializing from the unversioned form, or is it something deeper?

nodeup/pkg/model/kube_apiserver.go
```diff
@@ -216,12 +218,39 @@ func (b *KubeAPIServerBuilder) Build(c *fi.NodeupModelBuilderContext) error {
 	return nil
 }
 
+func (b *KubeAPIServerBuilder) configureOIDC(kubeAPIServer *kops.KubeAPIServerConfig) {
+	if b.NodeupConfig.APIServerConfig.Authentication == nil || b.NodeupConfig.APIServerConfig.Authentication.OIDC == nil {
```
justinsb, line 222: Could we do this on the kops-cli side?  The reason being that I expect the componentconfig objects eventually to become real component-config objects, and that we can create config files on the kops-cli side (also that they will come out of the cluster).  And in general, if we _can_ avoid logic in nodeup, I think we shoud.

But maybe there's a reason?

justinsb:
Generally I like the idea of exposing the OIDC options as "first class".  The most likely change to them is that we will support more than one in future (if/when kube-apiserver does), but I don't think that's on the cards right now and it's an easy-ish change to make (well, it's actually surprisingly difficult if you want to support existing readers that aren't aware of the plural form - e.g. PodIP -> PodIPs - but I think we could sacrifice that goal).

However, do we need to stop it being set in componentconfig?  Generally I think of componentconfig as the direct mapping of the flags, and the higher-level options as being our managed form.  We don't offer as many compatability guarantees around componentconfig options (because we can't, because flags get deprecated), but we do offer guarantees around the higher-level mappings.  But this does mean that we do end up with the ability to specify the same thing twice - once at high level, and once at low level.

kube-apiserver will probably be the last to go, but we do want to be able to use their versioned componentconfig when they do support it (I think!).

pkg/apis/kops/componentconfig.go
johngmyers: We only want to take values from the field in `authentication.oidc`. The field exists here for the `flag` annotation: so we can pass it to kube-apiserver. Longer term, we should probably have different struct types for the cluster API and the thing that passes data to kube-apiserver.

nodeup/pkg/model/kube_apiserver.go
johngmyers: It's the mindset. The admin should be thinking of configuring authentication to the cluster, not configuring the kube-apiserver binary. That OIDC is configured directly through kube-apiserver, whereas IAM and Kopeio are webhooks is an implementation detail that kOps should abstract away. As it is, it is difficult to discover how to configure OIDC because it is so buried.

johngmyers:
I'm having a hard time figuring out the benefit of allowing things to be specified in two places, much less figuring out how to reconcile situations when there are values in both places. There were already situations where the values specified in the API were overridden by the code; I have removed those fields from the v1alpha3 API.

johngmyers:
This PR doesn't pull  up `oidcCAFile` because specifying a filename isn't helpful without something provisioning the content of the file. I was thinking a future PR could add a `authentication.oidc.caCerts` or somesuch file, allowing specification of the CA public certs which nodeup would then provision onto the nodes.

hakman (approved):

PR #14898, opened 2022-12-28
Add documentation for switching to memberlist gossip protocol

efekete:


johngmyers (changes requested):
I don't see why this new document is in contributing; it appears to be operational.

I think kOps might be migrating off of Weave Mesh a bit less optionally.

efekete:
Yes, this is operational but to us it was quite important to it right without downtime. I am not sure what would be the right place for this.  I would gladly move it if you point me to it.
Either way, I think it is important to note that using weaveworks mesh has its (quite low) limitations in terms of cluster size. It also has high resource usage.

hakman:
@efekete What do you think about https://github.com/kubernetes/kops/blob/master/docs/gossip.md ?

efekete:
Do you think this whole guide and notes about limitations should go there? It makes sense, I'll move it there.

Should I open new PR then and close this one, or make that as new commits here?

hakman:
Please keep this PR. You should probably squash the commits.

efekete:
@hakman, @johngmyers I moved the docs, let me know if there is something else

justinsb:
And to the comment above, I think we may be able to replace gossip entirely, because in general we have to seed gossip by discovering some VMs with a cloud-api call _anyway_, and we can probably use that mechanism to just find the control-plane VMs.

However, that's more directional and not going to happen immediately, and these docs are super helpful  - thank you!

docs/gossip.md
```diff
@@ -7,9 +7,10 @@ Gossip does not suffer potential disruptions due to out of date records in DNS c
 
 Gossip is also the only option if you want to deploy a cluster in any of the AWS regions without Route 53, such as the China and GovCloud ones.
 
+There is two supported gossip protocols: weaveworks mesh and memberlist. Memberlist is suggested on larger clusters due to known high resource usage of mesh protocol on larger clusters.
```
johngmyers, line 10:
```suggestion
There are two supported gossip protocols: weaveworks mesh and memberlist. Memberlist is suggested on larger clusters due to known high resource usage of the mesh protocol on larger clusters.
```

docs/gossip.md
```diff
@@ -23,4 +24,82 @@ If you are using [bastion hosts](bastion.md), it is a bit tricky to find the bas
...
+      listen: 0.0.0.0:4000
+      protocol: memberlist
+
+  dnsControllerGossipConfig:
+    protocol: mesh
+    listen: 0.0.0.0:3998
+    seed: 127.0.0.1:3999
+    secondary:
+      protocol: memberlist
+      listen: 0.0.0.0:3993
```
johngmyers, line 55: Could these two be put in consistent order?

docs/gossip.md
```diff
@@ -23,4 +24,82 @@ If you are using [bastion hosts](bastion.md), it is a bit tricky to find the bas
...
+```
+
+### 2 Switch memberlist to primary and remove secondary protocol
+
+```yaml
+  gossipConfig:
+    protocol: memberlist
+    listen: 0.0.0.0:4000
+    secondary:
+      protocol: ""
```
johngmyers, line 89: Are they supposed to remove it or set it to `""`?

docs/gossip.md
```diff
@@ -23,4 +24,82 @@ If you are using [bastion hosts](bastion.md), it is a bit tricky to find the bas
...
+    protocol: memberlist
+    listen: 0.0.0.0:3993
+    seed: 127.0.0.1:4000
+    secondary:
+      protocol: ""
+```
+
+Restart worker nodes and then restart control-plane nodes
+
+Note that this will not flag nodes as *Needs update*; you will have to do *rolling-update* with `--force` flag.
```
johngmyers, line 101: Perhaps move this up a paragraph? It's not the rolling update that is expected to cause the nodes to be needs-update.

It would be better to modify kOps to cause the nodes to be marked as needs-update. One would do this by including the relevant config in the `nodeup.Config` or userdata.

docs/gossip.md
```diff
@@ -23,4 +24,82 @@ If you are using [bastion hosts](bastion.md), it is a bit tricky to find the bas
...
+
+  dnsControllerGossipConfig:
+    protocol: memberlist
+    listen: 0.0.0.0:3993
+    seed: 127.0.0.1:4000
+    secondary:
+      protocol: ""
+```
+
+Restart worker nodes and then restart control-plane nodes
```
johngmyers, line 99: Perhaps say "Perform a rolling update on" instead of "Restart"? Presumably a reboot wouldn't be effective.

docs/gossip.md
```diff
@@ -23,4 +24,82 @@ If you are using [bastion hosts](bastion.md), it is a bit tricky to find the bas
...
+
+Note that this will not flag nodes as *Needs update*; you will have to do *rolling-update* with `--force` flag.
+
+**!Restart control-plane nodes last, othwerwise all nodes that still have mesh as primary will get into NotReady state!**
```
johngmyers, line 103: This could be designed as three steps: add secondary, flip primary and secondary, then remove secondary. This would remove the failure if control plane updates before worker nodes.

docs/gossip.md
```diff
@@ -23,4 +24,82 @@ If you are using [bastion hosts](bastion.md), it is a bit tricky to find the bas
...
+
+**Do not remove secondary protocol completly, otherwise new nodes won't join** [duplicate metrics issue, Protokube won't ever start](https://github.com/kubernetes/kops/issues/9006)
```
johngmyers, line 105: What's going on here? The secondary protocol is defaulting to something? Can kOps not be fixed to avoid this problem?

johngmyers:
Ordinarily I'd suggest fixing kOps to make the procedure more straightforward, but in this case there's a separate initiative to phase out both Weaveworks mesh and memberlist.

docs/gossip.md
efekete: Original idea was:
1st add second protocol -> rolling update control planes
2nd switch protocols -> rolling update all nodes
3rd remove second protocol

but on the second step we had the the problem after rolling-update control planes as all the worker nodes went into not ready state. If we are going to do it that way, I guess we would need to do rolling-update of all nodes 3 times (on every step). This way it is just once (actually twice control planes, and once worker nodes)

I am fine with putting either in docs. We were doing switch on one of our production clusters with the two step process as doing rolling-update of all nodes three times would take us just too much time.

docs/gossip.md
efekete: Agree, I'll change

docs/gossip.md
efekete: As I understand, the problem is with the Prometheus metrics for some reason. I haven't investigate to be honest.

As already suggested here in comments, the idea is of removing both of these protocols so I am not sure if this is worth fixing. 

Main reason I think this i worth documenting is to note that there could be issues with current gossip protocol and to suggest temporary fix for it. 
It is difficult to find this problem when some nodes randomly start OOMing and you have to spend quite some time debugging to find that the protokube gossip protocol is the the problem. There are Github issues about that, but it's hard to find them when you don't know what you are searching for.

Therefore I think this is worth documenting, but not worth spending time to simplify this through some kOps optimizations.

docs/gossip.md
efekete: Of course

docs/gossip.md
efekete: Set it to ""

I'll change the instruction to clarify this

efekete:
I agree, but I think this is still worth documenting before that is ready.

I wouldn't make a big deal of make this switch easier as this is just temporary for people that actually run into issues with gossip mesh protocol.

PR #14905, opened 2022-12-30
Use private topology for apiserver e2e test

johngmyers:
Because anyone needing dedicated apiserver nodes is unlikely to be using public topology.


hakman (approved):

johngmyers:
Needs the fix for #14907, but I don't know why that wasn't exposed with public topology.
/lgtm cancel

hakman:
@johngmyers Why didn't you just add that commit here?

johngmyers:
@hakman That commit should be backported, this PR doesn't need to be.

johngmyers:
/
/retest

PR #14910, opened 2022-12-30
WIP: consistent paths for config and secrets scheduler

justinsb:
- kube-scheduler: Use /etc/kubernetes/kube-scheduler for config/secrets
- Update expected output


johngmyers:
Does `/etc/kubernetes` work on ContainerOS and Flatcar?

This will be a breaking change for some people with hooks.

justinsb:
> Does /etc/kubernetes work on ContainerOS and Flatcar?

Yes - we always write /etc/kubernetes/manifests here.  In general I think it's a better idea to use bind-mounts to create the directories we want with the permissions we want, rather than moving things around.

> This will be a breaking change for some people with hooks.

Good point.  I would like to get everything moved to consistent locations, so even if we think this is OK for the scheduler, it might be more problematic elsewhere.  I'll ponder...

/hold for pondering

PR #14911, opened 2022-12-30
kops replace: can now replace additional objects

justinsb:
Behind the ClusterAddons feature flag, but can create or replace
matching objects also.

Because some componentconfig objects don't have metadata, support the
cluster name flag (--name), but don't default it to kubeconfig
current-context.


cmd/kops/replace.go
```diff
@@ -84,138 +99,314 @@ func NewCmdReplace(f *util.Factory, out io.Writer) *cobra.Command {
...
+
+		clusterName := ""
+
+		switch v := obj.(type) {
+		case *kopsapi.Cluster:
+			clusterName = v.Name
+		case *kopsapi.InstanceGroup:
+			clusterName = v.ObjectMeta.Labels[kopsapi.LabelClusterName]
+		case *kopsapi.SSHCredential:
+			clusterName = v.ObjectMeta.Labels[kopsapi.LabelClusterName]
```
johngmyers, line 145: Existing behavior, but this seems to me to be an abuse of a label. We're treating it sort-of like `namespace`, but we're shoehorning it into a magic label because `ObjectMeta` doesn't have a `Cluster` field.

I think we need a better way of handling cluster-namespaced objects. Perhaps by putting the cluster name in the `Namespace` field?

cmd/kops/replace.go
```diff
@@ -84,138 +99,314 @@ func NewCmdReplace(f *util.Factory, out io.Writer) *cobra.Command {
...
+		case *unstructured.Unstructured:
+			if !featureflag.ClusterAddons.Enabled() {
+				klog.V(2).Infof("Type of object was %T", v)
+				return fmt.Errorf("unhandled kind %v", gvk)
+			}
+			clusterName = v.GetLabels()[kopsapi.LabelClusterName]
```
johngmyers, line 151: And we're extending the abuse of the label to new objects...

justinsb:
Agreed on the label; I don't really like it looking at it again a few years later.

I was thinking that by introducing the cluster name flag we can basically force "one invocation per cluster", and then we can work towards making the cluster-name argument mandatory / strongly encouraged.  If you pass the cluster-name, the label can be omitted; if you specify the label it has to match.  I think that'll nudge people towards not using the label.

I think if these were real cluster objects we would have a spec.clusterRef on the InstanceGroup etc.  That's a bit awkward for KubeSchedulerConfig (where we don't control the schema, and where it doesn't even have a metadata yet, though I think consensus is that was likely a mistake).  But we would probably wrap "additional objects" anyway, because we don't want them to actually do anything on a "management cluster".

That wrapper would probably look something like the "ClusterPackage" type that is currently only used for internal book-keeping inside kOps. but maybe with some objects embedded.  If we did that, we could add spec.clusterRef  to our objects.

cluster-api did spec.clusterName, which would also work as we do know the type: https://github.com/kubernetes-sigs/cluster-api/blob/main/api/v1beta1/machinedeployment_types.go#L68

I think though that for back-compatibility we can't drop the multi-cluster functionality enabled here.  But how about if I don't recognize the label on additional objects, then we effectively force anyone using additional objects to explicitly pass the cluster name?

johngmyers:
I split the clusterName discussion to #14928

PR #14912, opened 2022-12-30
gce: try to avoid concurrent IAM project operations

justinsb:
We set up a process-wide table of mutexes, to avoid concurrent IAM
operations on GCE projects.  Best-effort is reasonable here, we will
retry, but avoiding concurrent operations just avoids logspam and a
needless retry from self-conflicts.


justinsb:
I think we can probably use this in some other places as we encounter them, but starting with GCE as it gets logged on-screen.

johngmyers (approved):

PR #14913, opened 2022-12-30
Run pods needing control-plane instance credentials on hostNetwork

johngmyers:
Allows setting IMDS hop limit to 1 on control-plane nodes.

### DIFF
--- a/pkg/wellknownports/wellknownports.go
+++ b/pkg/wellknownports/wellknownports.go
@@ -73,6 +73,9 @@ const (
 	// VxlanUDP is the port used by VXLAN tunneling over UDP
 	VxlanUDP = 8472
 
+	// AWSLBCMetricsPort is reserved for the AWS Load Balancer Controller's metrics.
+	AWSLBCMetricsPort = 9442
+
 	// KubeletAPI is the port where kubelet listens
 	KubeletAPI = 10250
 )
--- a/tests/integration/update_cluster/additionalobjects/data/aws_s3_object_additionalobjects.example.com-addons-aws-ebs-csi-driver.addons.k8s.io-k8s-1.17_content
+++ b/tests/integration/update_cluster/additionalobjects/data/aws_s3_object_additionalobjects.example.com-addons-aws-ebs-csi-driver.addons.k8s.io-k8s-1.17_content
@@ -746,6 +746,7 @@ spec:
         volumeMounts:
         - mountPath: /csi
           name: socket-dir
+      hostNetwork: true
       nodeSelector: null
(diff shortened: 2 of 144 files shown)


johngmyers:
many-addons broken due to cluster-autoscaler

rifelpet:
I believe this has a side effect of preventing CiliumNetworkPolicies from restricting the network access for these pods

johngmyers:
@olemarkus What is with the upgrade-ab test? It's failing because the PDB is not satisfied 5 seconds after the new manifest was applied. It takes time for a Deployment to roll out.

We probably need a loop around the `${CHANNELS} apply channel` command, waiting for it to reconcile.

johngmyers:
I don't see a problem with CiliumNetworkPolicies not applying to these pods. They are privileged, after all. If they care that much about restricting their access, they can enable IRSA.

pkg/wellknownports/wellknownports.go
```diff
@@ -73,6 +73,21 @@ const (
...
+	AWSLBCHealthPort = 9441
+
+	//AWSLBCMetricsPort is reserved for the AWS Load Balancer Controller's metrics.
+	AWSLBCMetricsPort = 9442
+
+	// AWSLBCWebhookPort is reserved for the AWS Load Balancer Controller's webhook server.
+	AWSLBCWebhookPort = 9443
+
+	// CloudCSIDriverPort is reserved for the cloud provider's primary CSI driver.
+	CloudCSIDriverPort = 9808
```
hakman, line 89: This is only needed for AWS, no need to complicate other providers with it.
```suggestion
	// AWSCSIDriverPort is reserved for the cloud provider's primary CSI driver.
	AWSCSIDriverPort = 9808
```

johngmyers: The OpenStack cinder daemonset runs on host network and uses a port.

The GCP PD controller also runs on host network, though it uses three ports.

@justinsb should we not try to share port assignments across the primary CSI drivers for cloud providers?

johngmyers: We should be more proactive in avoiding port conflicts. We don't have the coverage to find them all through testing.

olemarkus: We probably should have a map of the ports in `wellknownports.go` and use that map in manifests that are using host networking. That would make it easier to discover any conflicts.

johngmyers:
Making this PR minimal and leaving the avoidance of other port conflicts to future work.

hakman (approved):

PR #14915, opened 2022-12-30
Reduce use of cluster topology field

johngmyers:
I'm questioning why we have the `topology.nodes` and `topology.controlPlane` fields in the cluster spec. They're redundant with the cluster subnet specs.

This changes AWS subnet tagging to be based on the existence or not of private subnets. It changes GCE assignment of public IPs to nodes to behave like AWS: based on the subnet the instance is in, not whether the topology spec thinks the control plane is public.


hakman:
The DNS topology is not that useful either. Private DNS has some value for AWS, but that's pretty much it.

hakman (approved):

PR #14916, opened 2022-12-30
openstack: use subnet type instead of topology

johngmyers:
The desire here is to remove references to the `topology.nodes` and `topology.controlPlane` fields in the cluster spec. They're redundant with the cluster `subnet` specs.

This is my attempt at moving the Openstack code so that it uses the subnet specifications instead of what is stored in `topology`. I don't quite know enough about Openstack to know if I'm approaching things the right way, so would appreciate some help.

The `minimal_openstack` and `openstack_floatingip` cluster specs supply a `Public` subnet, but say topology is `private`, which doesn't make sense. Suggestions?

/area provider/openstack
/cc @olemarkus @zetaab


pkg/model/openstackmodel/servergroup.go
```diff
@@ -191,39 +195,21 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.CloudupModelBuilderContex
...
 		// Associate a floating IP to the instances if we have external network in router
-		// and respective topology is "public"
+		// and respective subnet is "Public" or "Utility".
 		if b.Cluster.Spec.CloudProvider.Openstack.Router != nil {
 			if ig.Spec.AssociatePublicIP != nil && !fi.ValueOf(ig.Spec.AssociatePublicIP) {
 				continue
 			}
-			switch ig.Spec.Role {
-			case kops.InstanceGroupRoleBastion:
+			if havePublicSubnet {
```
zetaab, line 203: the only problem with this setup is that now if people had "Private" topology subnets (and bastion in those) it means that after this PR kops will create bastion to private subnet without floating ip and there is no way to connect to it. Openstack provider does not create loadbalancer just for bastion like AWS (as I see it its waste of resources). I would still create floatingip always for bastion to have backwards compatibility.

zetaab: the real problem here is that openstack is not similar for each company, sometimes its possible that people can connect already to fixed ips, sometimes not (like in our case). The network setup varies a lot between companies.

If we check old code: it created floatingip always for bastion, but only for master/nodes if those had Public topology.

zetaab (changes requested):

pkg/model/openstackmodel/servergroup.go
zetaab: "Utility" subnets are not either used in openstack, at least we have not used those in anywhere

johngmyers: Thank you for the feedback and context. I have changed the PR to always create FloatingIP for bastions.

Would it be appropriate to change the subnet types in the failing tests? Do you think there might be users with private topology but subnets incorrectly marked as type public?

zetaab: its little bit difficult how people are thinking what is the different between private and public topologies in OpenStack. The problem is that OpenStack can be installed in so many different ways.

For me and how I am thinking it should be:

private topology:
- instances does have ip address from internal network

public topology:
- instances does have ip address from internal network
- instances do have also floating ip address (from external network)

However, in case of private topology in some OpenStack installations it is already possible to connect to instances. It depends how the network is installed.

zetaab (changes requested):

pkg/model/openstackmodel/servergroup.go
```diff
@@ -130,11 +131,14 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.CloudupModelBuilderContex
 				az = fi.PtrTo(subnet)
 			}
 
-			subnetName, err := b.findSubnetClusterSpec(subnet)
+			subnetName, subnetType, err := b.findSubnetClusterSpec(subnet)
 			if err != nil {
 				return err
 			}
 			subnets = append(subnets, b.LinkToSubnet(s(subnetName)))
+			if subnetType == kops.SubnetTypePublic || subnetType == kops.SubnetTypeUtility {
```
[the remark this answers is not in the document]
johngmyers: Including `SubnetTypeUtility` makes it consistent with other cloud providers (even though nodes shouldn't be in utility subnets).

pkg/model/openstackmodel/servergroup_test.go
```diff
@@ -67,6 +67,7 @@ func getServerGroupModelBuilderTestInput() []serverGroupModelBuilderTestInput {
 						Subnets: []kops.ClusterSubnetSpec{
 							{
 								Name:   "subnet",
+								Type:   kops.SubnetTypePublic,
```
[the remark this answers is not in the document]
johngmyers: I think that would be a change for a future PR. I'm not sure I have enough context to do that.

johngmyers:
I fixed the tests to have the subnet types be consistent with the topology. After that, the substantive change does not affect test output.

zetaab (approved):

PR #14926, opened 2023-01-01
Support multiple terraform file providers

justinsb:
By adding them on-demand, it is arguably a little simpler as well.

### DIFF
--- a/upup/pkg/fi/cloudup/terraform/target_hcl2.go
+++ b/upup/pkg/fi/cloudup/terraform/target_hcl2.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sort"
 
+	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/featureflag"
 	"k8s.io/kops/upup/pkg/fi/cloudup/terraformWriter"
@@ -132,11 +133,13 @@ func (t *TerraformTarget) writeProviders(buf *bytes.Buffer) {
 		Write(buf, 0, fmt.Sprintf("provider %q", providerName))
 	buf.WriteString("\n")
 
-	// Add the second provider definition for managed files
-	if t.filesProvider != nil {
+	// Add any additional provider definition for managed files
+	keys := sortedKeysForMap(t.TerraformWriter.Providers)
+	for _, key := range keys {
+		provider := t.TerraformWriter.Providers[key]
 		providerBody := map[string]string{}
 		providerBody["alias"] = "files"
-		for k, v := range t.filesProvider.Arguments {
+		for k, v := range provider.Arguments {
 			providerBody[k] = v
 		}
 		for k, v := range tfGetFilesProviderExtraConfig(t.clusterSpecTarget) {
@@ -144,9 +147,18 @@ func (t *TerraformTarget) writeProviders(buf *bytes.Buffer) {
 		}
 		mapToElement(providerBody).
 			ToObject().
-			Write(buf, 0, fmt.Sprintf("provider %q", t.filesProvider.Name))
+			Write(buf, 0, fmt.Sprintf("provider %q", provider.Name))
 		buf.WriteString("\n")
 	}
+}
+
+func sortedKeysForMap[K ~string, V any](m map[K]V) []K {
+	var keys []K
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
+	return keys
 }
 
 func (t *TerraformTarget) writeResources(buf *bytes.Buffer, resourcesByType map[string]map[string]interface{}) {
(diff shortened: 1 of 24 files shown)


upup/pkg/fi/cloudup/terraform/target_hcl2.go
```diff
@@ -132,23 +133,34 @@ func (t *TerraformTarget) writeProviders(buf *bytes.Buffer) {
 		Write(buf, 0, fmt.Sprintf("provider %q", providerName))
 	buf.WriteString("\n")
 
-	// Add the second provider definition for managed files
-	if t.filesProvider != nil {
+	// Add any additional provider definition for managed files
+	keys := keysForMap(t.TerraformWriter.Providers)
+	sort.Strings(keys)
+	for _, key := range keys {
+		filesProvider := t.TerraformWriter.Providers[key]
```
johngmyers, line 140: It's not necessarily the files provider at this point. Can this be named `provider`?

upup/pkg/fi/cloudup/terraform/target_hcl2.go
```diff
@@ -132,23 +133,34 @@ func (t *TerraformTarget) writeProviders(buf *bytes.Buffer) {
...
 		providerBody := map[string]string{}
 		providerBody["alias"] = "files"
```
johngmyers, line 142: I don't think you can have multiple providers with the same alias.

upup/pkg/fi/cloudup/terraform/target_hcl2.go
```diff
@@ -132,23 +133,34 @@ func (t *TerraformTarget) writeProviders(buf *bytes.Buffer) {
...
 		buf.WriteString("\n")
 	}
 }
 
+func keysForMap[K comparable, V any](m map[K]V) []K {
+	var keys []K
+	for k := range m {
+		keys = append(keys, k)
+	}
+	return keys
```
johngmyers, line 161: Sort here instead of in the caller? `K` is always `string`.

upup/pkg/fi/cloudup/terraformWriter/writer.go
```diff
@@ -88,6 +92,28 @@ func (t *TerraformWriter) AddFileBytes(resourceType string, resourceName string,
...
+
+	key := tfProvider.Name
+
+	existing := t.Providers[key]
+	if existing != nil {
+		if reflect.DeepEqual(tfProvider, existing) {
+			// already exists and matches
+			return nil
+		}
+		return fmt.Errorf("attempt to add different tfProvider with key %q", key)
```
johngmyers, line 111: Consider panic() instead?

Actually, consider having this take name and arguments as parameters, then return an interned `*TerraformProvider`.

johngmyers (changes requested):

upup/pkg/fi/cloudup/terraform/target_hcl2.go
rifelpet: They can have the same alias as long as they are not the same provider type, which the `Providers` map should ensure

```tf
provider "null" {
  alias = "foo"
}

provider "random" {
  alias = "foo"
}

resource "null_resource" "bar" {
  provider = null.foo
}

resource "random_string" "bar" {
  provider = random.foo
  length = 1
}
```

upup/pkg/fi/cloudup/terraform/target_hcl2.go
justinsb: Good call!

upup/pkg/fi/cloudup/terraform/target_hcl2.go
justinsb: Sure... the challenge with generics seems to be when to stop, but this is a clear win and it's comparable anyway...

upup/pkg/fi/cloudup/terraformWriter/writer.go
justinsb: Changed to panic.

I was thinking along similar lines, it would be nice if the vfs package did not depend on the terraform package, and refactoring EnsureProvider to take the parameters would be good here (after this PR, we don't need to expose the TerraformProvider function I don't think).  But that is a much bigger refactor, and I wanted to keep this one smallish.  I can give it a go though, and we can see what it looks like in a separate PR (it'll build on this one anyway!)

upup/pkg/fi/cloudup/terraform/target_hcl2.go
```diff
@@ -196,31 +209,69 @@ func (t *TerraformTarget) writeTerraform(buf *bytes.Buffer) {
...
 		}
 	}
 
+	for _, tfProvider := range t.TerraformWriter.Providers {
+		providers[tfProvider.Name] = true
+		providerAliases[tfProvider.Name] = append(providerAliases[tfProvider.Name], "files")
+	}
+
+	providerKeys := sortedKeysForMap(providers)
+	sort.Strings(providerKeys)
```
johngmyers, line 231:
```suggestion
```

johngmyers (approved):
Just some nits here. Holding in case you want to fix, otherwise can be fixed up in subsequent commit.
/hold

upup/pkg/fi/cloudup/terraform/target_hcl2.go
```diff
@@ -132,23 +133,35 @@ func (t *TerraformTarget) writeProviders(buf *bytes.Buffer) {
 		Write(buf, 0, fmt.Sprintf("provider %q", providerName))
 	buf.WriteString("\n")
 
-	// Add the second provider definition for managed files
-	if t.filesProvider != nil {
+	// Add any additional provider definition for managed files
+	keys := sortedKeysForMap(t.TerraformWriter.Providers)
+	sort.Strings(keys)
```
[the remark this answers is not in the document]
justinsb: Ooops - thanks :-)

upup/pkg/fi/cloudup/terraform/target_hcl2.go
```diff
@@ -196,31 +209,69 @@ func (t *TerraformTarget) writeTerraform(buf *bytes.Buffer) {
...
+	for _, tfProvider := range t.TerraformWriter.Providers {
+		providers[tfProvider.Name] = true
+		providerAliases[tfProvider.Name] = append(providerAliases[tfProvider.Name], "files")
+	}
+
+	providerKeys := sortedKeysForMap(providers)
+	sort.Strings(providerKeys)
+	for _, provider := range providerKeys {
+		var tf map[string]*terraformWriter.Literal
+		switch provider {
```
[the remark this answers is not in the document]
justinsb: Will be the last commit in https://github.com/kubernetes/kops/pull/14948

johngmyers (approved):

PR #14927, opened 2023-01-01
EnsureTask should panic

justinsb:
This means that we automatically check the error code.  A linter could
detect errors here (maybe), but in practice we can't recover from
errors here anyway.


johngmyers:
I'll buy that. It looks like there's remnants of an `ErrorsTo` refactor.

johngmyers (approved):

Issue #14859, opened 2022-12-23
Add support for clusters without DNS or Gossip


hakman:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

kOps requires DNS alias(es) for various services, like `kube-apiserver`, `kops-controller`, ... .
DNS (in general) is unreliable on cluster creation and not always easy to create the prerequisites.
Gossip, as an alternative is nice, but it requires permissions for the cloud provider to list peers. Limiting those permissions to something reasonable is not always possible, like discussed in #14315.

Most supported cloud providers allow use of a load balancer for Kubernetes API, which could be used also for clusters that don't use any form of DNS.

**2. Feel free to provide a design supporting your feature request.**

Load balancers are created at the same time as the cluster and are not recreated unless deleted. The (usually) provide a stable address, which can be used in `/etc/hosts` aliases for the `kube-apiserver`, `kops-controller`, ... services.

To enable the feature, one could use the `--dns=none` flag on cluster creation.
There would be no restriction to the cluster name as there are for Gossip.

Unfortunately each cloud provider has its own load balancer implementation and features, so how clusters without DNS work will be dependent on cloud provider features.

- [x] Support for AWS
- [x] Support for Hetzner
- [x] Support for GCE/GCP
- [x] Support for Azure
- [x] Support for Openstack
- [x] Support for Scaleway
- [x] Allow migration from Gossip
- [ ] Documentation
- [ ] Release notes

johngmyers:
@hakman close this out and open new issue for remaining work or retarget this to the 1.27 milestone?

hakman:
I will close it before the release. Some of the tasks still need addressing before the 1.26 release.

protosam:
Getting stable seeds pretty much requires knowing something in advance. Using a load balancer will not reliably route to the desired server sometimes.

I saw etcd-manager using this public discovery service maintained by Core OS (maybe Red Hat now?): https://etcd.io/docs/v3.5/dev-internal/discovery_protocol/#public-discovery-service

First request returns a random UUID, which is used as an endpoint to PUT and GET seed info. 
```
DISCO_SVC=$(curl -Ls https://discovery.etcd.io/new)
curl -Ls ${DISCO_SVC}
curl -Ls -X PUT ${DISCO_SVC}/key-name-here -d value=value-goes-here
curl -Ls ${DISCO_SVC}/key-name-here
```

[...]

PR #14930, opened 2023-01-02
OpenStack: Add support for clusters without DNS

zetaab:
Refs https://github.com/kubernetes/kops/issues/14859

```
% kubectl get nodes
NAME                          STATUS   ROLES                AGE     VERSION
control-plane-espsti-omfcjm   Ready    control-plane,node   3m20s   v1.25.5
control-plane-esptnl-hxjzpy   Ready    control-plane,node   3m6s    v1.25.5
control-plane-helpa-kszzmr    Ready    control-plane,node   2m22s   v1.25.5
nodes-espsti-lo2kne           Ready    node                 3m6s    v1.25.5
nodes-esptnl-bq14jb           Ready    node                 3m5s    v1.25.5
```

node hosts file

```
...
# Begin host entries managed by kops - do not edit
10.222.156.209  api.internal.jessetesti.foobar.com
# End host entries managed by kops
```

so it seems working.

### DIFF
--- a/pkg/model/openstackmodel/servergroup.go
+++ b/pkg/model/openstackmodel/servergroup.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/model"
 	"k8s.io/kops/pkg/truncate"
+	"k8s.io/kops/pkg/wellknownports"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstack"
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstacktasks"
@@ -165,6 +166,10 @@ func (b *ServerGroupModelBuilder) buildInstances(c *fi.CloudupModelBuilderContex
 			Lifecycle:                b.Lifecycle,
 		}
 		c.AddTask(portTask)
+
+		if b.Cluster.UsesNoneDNS() && ig.Spec.Role == kops.InstanceGroupRoleControlPlane {
+			portTask.ForAPIServer = true
+		}
 
 		metaWithName := make(map[string]string)
 		for k, v := range igMeta {
@@ -315,8 +320,10 @@ func (b *ServerGroupModelBuilder) Build(c *fi.CloudupModelBuilderContext) error 
 		}
 		c.AddTask(poolTask)
 
+		nameForResource := fi.ValueOf(lbTask.Name)
 		listenerTask := &openstacktasks.LBListener{
-			Name:      lbTask.Name,
+			Name:      fi.PtrTo(nameForResource),
+			Port:      fi.PtrTo(wellknownports.KubeAPIServer),
 			Lifecycle: b.Lifecycle,
 			Pool:      poolTask,
 		}
@@ -334,7 +341,7 @@ func (b *ServerGroupModelBuilder) Build(c *fi.CloudupModelBuilderContext) error 
 		c.AddTask(listenerTask)
 
 		monitorTask := &openstacktasks.PoolMonitor{
-			Name:      lbTask.Name,
+			Name:      fi.PtrTo(nameForResource),
 			Pool:      poolTask,
 			Lifecycle: b.Lifecycle,
 		}
@@ -350,7 +357,7 @@ func (b *ServerGroupModelBuilder) Build(c *fi.CloudupModelBuilderContext) error 
 				Pool:          poolTask,
 				ServerGroup:   mastersg,
 				InterfaceName: fi.PtrTo(ifName),
-				ProtocolPort:  fi.PtrTo(443),
+				ProtocolPort:  fi.PtrTo(wellknownports.KubeAPIServer),
 				Lifecycle:     b.Lifecycle,
 				Weight:        fi.PtrTo(1),
 			}
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -1398,28 +1398,24 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 	// Set API server address to an IP from the cluster network CIDR
 	if cluster.UsesNoneDNS() {
 		switch cluster.Spec.GetCloudProvider() {
-		case kops.CloudProviderAWS, kops.CloudProviderHetzner:
+		case kops.CloudProviderAWS, kops.CloudProviderHetzner, kops.CloudProviderOpenstack:
 			// Use a private IP address that belongs to the cluster network CIDR (some additional addresses may be FQDNs or public IPs)
-			for _, networkCIDR := range append(cluster.Spec.Networking.AdditionalNetworkCIDRs, cluster.Spec.Networking.NetworkCIDR) {
-				_, cidr, err := net.ParseCIDR(networkCIDR)
-				if err != nil {
-					return nil, nil, fmt.Errorf("failed to parse network CIDR %q: %w", networkCIDR, err)
-				}
-				for _, additionalIP := range apiserverAdditionalIPs {
+			for _, additionalIP := range apiserverAdditionalIPs {
+				for _, networkCIDR := range append(cluster.Spec.Networking.AdditionalNetworkCIDRs, cluster.Spec.Networking.NetworkCIDR) {
+					_, cidr, err := net.ParseCIDR(networkCIDR)
+					if err != nil {
+						return nil, nil, fmt.Errorf("failed to parse network CIDR %q: %w", networkCIDR, err)
+					}
 					if cidr.Contains(net.ParseIP(additionalIP)) {
-						bootConfig.APIServerIP = additionalIP
-						break
+						bootConfig.APIServerIPs = append(bootConfig.APIServerIPs, additionalIP)
 					}
 				}
-				if bootConfig.APIServerIP != "" {
-					break
-				}
 			}
+
 		case kops.CloudProviderGCE:
 			// Use any IP address that is found (including public ones)
 			for _, additionalIP := range apiserverAdditionalIPs {
-				bootConfig.APIServerIP = additionalIP
-				break
+				bootConfig.APIServerIPs = append(bootConfig.APIServerIPs, additionalIP)
 			}
 		default:
 			return nil, nil, fmt.Errorf("'none' DNS topology is not supported for cloud %q", cluster.Spec.GetCloudProvider())
@@ -1429,8 +1425,8 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 	useConfigServer := apiModel.UseKopsControllerForNodeBootstrap(cluster) && !ig.HasAPIServer()
 	if useConfigServer {
 		host := "kops-controller.internal." + cluster.ObjectMeta.Name
-		if cluster.UsesNoneDNS() {
-			host = bootConfig.APIServerIP
+		if cluster.UsesNoneDNS() && len(bootConfig.APIServerIPs) > 0 {
+			host = bootConfig.APIServerIPs[0] // TODO: how we could support array?
 		}
 		baseURL := url.URL{
 			Scheme: "https",
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -333,7 +333,7 @@ func NewCluster(opt *NewClusterOptions, clientset simple.Clientset) (*NewCluster
 				MaxRetries: fi.PtrTo(3),
 			},
 		}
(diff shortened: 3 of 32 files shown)


pkg/model/openstackmodel/servergroup.go
```diff
@@ -308,57 +313,75 @@ func (b *ServerGroupModelBuilder) Build(c *fi.CloudupModelBuilderContext) error
...
+			if useVIPACL {
+				var AllowedCIDRs []string
+				// currently kOps openstack supports only ipv4 addresses
+				for _, CIDR := range b.Cluster.Spec.API.Access {
+					if net.IsIPv4CIDRString(CIDR) {
+						AllowedCIDRs = append(AllowedCIDRs, CIDR)
+					}
+				}
+				sort.Strings(AllowedCIDRs)
+				if isKopsController(name) {
```
zetaab, line 350: I think other cloudproviders is allowing kops controller port to same targets than API itself. But I do not see any reason why that should be allowed from outside cluster?

hakman: I did that to not allow `0.0.0.0/0` for public topologies.
In particular, GCP is quite inflexible when using a public LB, as it cannot allow communication with the private net, only public IPs, which are unknown. So, you are right, it's a hack that will be fixed at some point.
If you can, allow access to that port only from the cluster network CIDR.

hakman: Though, the kops-controller access is required when using it for bootstrap.
See: https://github.com/kubernetes/kops/pull/14460.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1408,6 +1415,17 @@ func initializeOpenstackAPI(opt *NewClusterOptions, cluster *api.Cluster) {
...
+	// this is needed in new clusters, otherwise openstack clients will automatically try to use openstack designate
+	if strings.ToLower(opt.DNSType) == "none" {
+		if cluster.Spec.Networking.Topology == nil {
+			cluster.Spec.Networking.Topology = &api.TopologySpec{
+				DNS: api.DNSTypeNone,
+			}
+		} else {
+			cluster.Spec.Networking.Topology.DNS = api.DNSTypeNone
+		}
+	}
```
hakman, line 1428: I thought that we already set this in a generic way for all cloud providers:
https://github.com/kubernetes/kops/blob/e85051b709e57a07fea062495c32fde6c4bc5c5a/upup/pkg/fi/cloudup/new_cluster.go#L1304-L1311

pkg/model/openstackmodel/servergroup.go
zetaab: Openstack does not use kops controller for bootstrap

hakman: I guess nodes don't need access to kops-controller. It would be nice to use kops-controller for bootstrap though.

upup/pkg/fi/cloudup/new_cluster.go
zetaab: @hakman yes its done for new clusters. But like the comment says the code is executed too late for openstack. 

openstack needs that information already in row https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster.go#L337

but its initialised in https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster.go#L378

if that information is not place in row 337, it will automatically try to initialise openstack designate because cluster.UsesNoneDNS() returns still false. Openstack designate will not work if that is not installed to openstack setup (like in our case).

So we could have generic model IF we could move that before row 337, but its difficult because in that case we should move other parts as well

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1408,6 +1415,17 @@ func initializeOpenstackAPI(opt *NewClusterOptions, cluster *api.Cluster) {
...
+
+	// this is needed in new clusters, otherwise openstack clients will automatically try to use openstack designate
+	if strings.ToLower(opt.DNSType) == "none" {
+		if cluster.Spec.Networking.Topology == nil {
+			cluster.Spec.Networking.Topology = &api.TopologySpec{
+				DNS: api.DNSTypeNone,
+			}
+		} else {
+			cluster.Spec.Networking.Topology.DNS = api.DNSTypeNone
+		}
```
hakman, line 1427:
```suggestion
		if cluster.Spec.Networking.Topology == nil {
			cluster.Spec.Networking.Topology = &api.TopologySpec{}
		}
		cluster.Spec.Networking.Topology.DNS = api.DNSTypeNone
```

pkg/model/openstackmodel/servergroup.go
zetaab: what is the advantage of using kops-controller for bootstrapping? We do not need that much information in metadata in future in normal nodes?

hakman: You do need config and cert(s) passed to nodes. This requires tokens to read the config and cert(s) files from s3 or similar.

zetaab:
@hakman is there something missing from this? One thing that I was thinking that is it correct that I use basically external loadbalancer ip (same than clients uses externally with kubectl)? In openstack the ports are usually going to same as same in internal network, which means that also the internal ips for each master is quite static. Could we instead use those (like currently with gossip?). 

example:
```
% openstack port list|grep master
+--------------------------------------+------------------------------------------------------+-------------------+-------------------------------------------------------------------------------+--------+
| ID                                   | Name                                                 | MAC Address       | Fixed IP Addresses                                                            | Status |
+--------------------------------------+------------------------------------------------------+-------------------+-------------------------------------------------------------------------------+--------+
| 0de4bfb9-758a-463a-a31a-0f9f72fddcaf | port-master-helpa-1-rofa3-k8s-local                  | fa:16:3e:51:06:26 | ip_address='100.76.3.162', subnet_id='680dbcb6-1a6e-4f2a-a8b0-a86eb8733ff1'   | ACTIVE |
| 3406cb48-3812-4b50-88cb-c1c65addc299 | port-master-espsti-1-rofa3-k8s-local                 | fa:16:3e:1c:43:da | ip_address='100.68.1.59', subnet_id='5dcd7e63-cfc2-46a5-8660-10e1932c27ce'    | ACTIVE |
| b707c06d-3b74-497d-bf46-c5995389d65e | port-master-esptnl-1-rofa3-k8s-local                 | fa:16:3e:ff:f5:9a | ip_address='100.72.0.36', subnet_id='70da818a-34f2-400b-a851-32cd9adb5294'    | ACTIVE |
+--------------------------------------+------------------------------------------------------+-------------------+-------------------------------------------------------------------------------+--------+
```

so one possibility is that instead of LB ip we could use these internal 100.x.x.x ips like with gossip. However, there is small probability that these are going to change if someone will delete the ports (same applies if someone will delete the loadbalancer). These are not deleted during lifetime of cluster, updating master (control-plane) means that instance is rolled over but port is going to same as same.

hakman:
> Is there something missing from this?

This includes changes to expose `kops-controller` or something similar in `servergroup.go`. As I explained, that is used for bootstrap and config, but OpenStack doesn't have that feature yet (btw, you should sooner or later). Otherwise, there should be no changes related to `kops-controller`.

> Could we instead use those (like currently with gossip?)

In theory yes, see what I did for AWS NLB internal network interfaces.
There is an extra change that would be needed, to send multiple IPs to the nodes. Most linux software would not understand multiple `/etc/hosts` entries, but Go binaries build without CGO should.

zetaab:
current status

```
% kubectl get nodes
NAME                          STATUS   ROLES                AGE   VERSION
control-plane-espsti-dhowof   Ready    control-plane,node   83s   v1.25.5
control-plane-esptnl-exyhow   Ready    control-plane,node   67s   v1.25.5
control-plane-helpa-qe3fj2    Ready    control-plane,node   81s   v1.25.5
nodes-espsti-yegjri           Ready    node                 77s   v1.25.5
nodes-esptnl-xdala8           Ready    node                 77s   v1.25.5
nodes-helpa-l4vjnj            Ready    node                 76s   v1.25.5


root@nodes-esptnl-xdala8:/home/ubuntu# cat /etc/hosts
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1 ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts

# Begin host entries managed by kops - do not edit
10.2.32.182     api.internal.jessetesti.k8s.local
# End host entries managed by kops

% openstack port list|grep control
| 89476442-2c3a-457b-9208-fefc61df4c49 | port-control-plane-espsti-1-jessetesti-k8s-local     | fa:16:3e:6d:62:d6 | ip_address='10.2.32.182', subnet_id='56d1af54-558d-4efd-b34c-b030eef8be9a'    | ACTIVE |
| a5454a19-a010-4466-b880-7bb4abfc1ae2 | port-control-plane-helpa-1-jessetesti-k8s-local      | fa:16:3e:6d:59:ab | ip_address='10.2.98.6', subnet_id='e9318872-d580-4cd1-9152-6fe54774cbb6'      | ACTIVE |
| d9af17cb-efc5-42a0-91ac-00b6a8dfa274 | port-control-plane-esptnl-1-jessetesti-k8s-local     | fa:16:3e:9f:59:02 | ip_address='10.2.64.70', subnet_id='c4a49b69-6689-4975-bc46-85524a6e74ca'     | ACTIVE |
```

I removed kops-controller part because its not implemented yet, better to add that in same PR. Now the only missing thing is multiple IPs to /etc/hosts?

zetaab:
```
% kubectl get nodes
NAME                          STATUS   ROLES                AGE   VERSION
control-plane-espsti-6ru9ix   Ready    control-plane,node   62s   v1.25.5
control-plane-esptnl-qnfuku   Ready    control-plane,node   69s   v1.25.5
control-plane-helpa-n0lzpz    Ready    control-plane,node   69s   v1.25.5
nodes-espsti-sg5tsl           Ready    node                 63s   v1.25.5
nodes-esptnl-znotb6           Ready    node                 62s   v1.25.5
nodes-helpa-anaxhf            Ready    node                 62s   v1.25.5


root@nodes-esptnl-znotb6:/home/ubuntu# cat /etc/hosts
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1 ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts

# Begin host entries managed by kops - do not edit
10.2.35.36      api.internal.jessetesti.k8s.local
10.2.64.25      api.internal.jessetesti.k8s.local
10.2.96.133     api.internal.jessetesti.k8s.local
# End host entries managed by kops

% openstack port list|grep control                        
| 021e10c9-fddf-40fc-bed2-30cddfbdc53d | port-control-plane-esptnl-1-jessetesti-k8s-local     | fa:16:3e:ef:3f:a5 | ip_address='10.2.64.25', subnet_id='c4a4a470-e321-418a-bb8d-779ed7a80a41'     | ACTIVE |
| 28de7678-9440-4992-98f5-92dbfc108290 | port-control-plane-espsti-1-jessetesti-k8s-local     | fa:16:3e:9e:e1:c1 | ip_address='10.2.35.36', subnet_id='e5df8b59-a2c0-423b-908d-7ac84e736c5a'     | ACTIVE |
| f43a60ed-82aa-4404-8004-e221eb53ca5a | port-control-plane-helpa-1-jessetesti-k8s-local      | fa:16:3e:18:07:9e | ip_address='10.2.96.133', subnet_id='0d5521d5-bf9c-46e7-b8fa-a9b7c75b9aa6'    | ACTIVE |
```

Openstack project credentials does exist only in control plane when using None dns.
However, if people uses Openstack swift those are still needed in nodes (because node needs VFS access).
Kops-controller provisioning could remove this if I understood correctly.

It seems now working like should?

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1429,8 +1425,8 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 	useConfigServer := apiModel.UseKopsControllerForNodeBootstrap(cluster) && !ig.HasAPIServer()
 	if useConfigServer {
 		host := "kops-controller.internal." + cluster.ObjectMeta.Name
-		if cluster.UsesNoneDNS() {
-			host = bootConfig.APIServerIP
+		if cluster.UsesNoneDNS() && len(bootConfig.APIServerIPs) > 0 {
+			host = bootConfig.APIServerIPs[0] // TODO: how we could support array?
```
zetaab, line 1429: how to deal with this? or is one ip enough?

hakman: Let's discuss this separately.

PR #14945, opened 2023-01-03
vfs: don't cache bucketDetails in S3Path

justinsb:
We cache it in the S3Context anyway, and this potentially avoids some
race conditions (albeit likely harmless race conditions).

### DIFF
--- a/util/pkg/vfs/s3fs.go
+++ b/util/pkg/vfs/s3fs.go
@@ -37,11 +37,10 @@ import (
 )
 
 type S3Path struct {
-	s3Context     *S3Context
-	bucket        string
-	bucketDetails *S3BucketDetails
-	key           string
-	etag          *string
+	s3Context *S3Context
+	bucket    string
+	key       string
+	etag      *string
 
 	// scheme is configurable in case an S3 compatible custom
 	// endpoint is specified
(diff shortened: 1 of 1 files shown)


util/pkg/vfs/s3fs.go
```diff
@@ -442,26 +442,22 @@ func (p *S3Path) ReadTree() ([]Path, error) {
...
 
-func (p *S3Path) ensureBucketDetails(ctx context.Context) error {
-	if p.bucketDetails == nil || p.bucketDetails.region == "" {
-		bucketDetails, err := p.s3Context.getDetailsForBucket(ctx, p.bucket)
-
-		p.bucketDetails = bucketDetails
-		if err != nil {
-			return err
-		}
+func (p *S3Path) ensureBucketDetails(ctx context.Context) (*S3BucketDetails, error) {
```
johngmyers, line 445:
```suggestion
func (p *S3Path) getBucketDetails(ctx context.Context) (*S3BucketDetails, error) {
```

johngmyers (changes requested):

util/pkg/vfs/s3fs.go
justinsb: Good call - done!

PR #14947, opened 2023-01-04
Miscellaneous NodeupConfig refactoring

johngmyers:


nodeup/pkg/model/context.go
```diff
@@ -355,6 +358,17 @@ func (c *NodeupModelContext) BuildBootstrapKubeconfig(name string, ctx *fi.Nodeu
 	}
 }
 
+// RemapImage applies any needed remapping to an image reference.
+func (c *NodeupModelContext) RemapImage(image string) string {
+	if c.remapRegistry {
```
hakman, line 363: Why not inline the condition here?
```suggestion
	if components.IsBaseURL(c.NodeupConfig.KubernetesVersion) && c.IsKubernetesLT("1.25") {
```

johngmyers: Mostly because of the order in which I refactored things.

hakman (approved):

PR #14948, opened 2023-01-04
Refactor terraform provider version logic

justinsb:
As suggested in review of #14926, this approach might be more concise and clearer.

Builds on #14926


justinsb:
/retest

Looks like an unrelated flake

johngmyers (approved):

PR #14954, opened 2023-01-05
gce terraform: Use google_storage_object_acl

justinsb:
We are using the schema of google_storage_object_acl.

### DIFF
--- a/util/pkg/vfs/gsfs.go
+++ b/util/pkg/vfs/gsfs.go
@@ -456,7 +456,7 @@ func (p *GSPath) RenderTerraform(w *terraformWriter.TerraformWriter, name string
 			// https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_object_acl#role_entity
 			tfACL.RoleEntity = append(tfACL.RoleEntity, fmt.Sprintf("%v:%v", re.Role, re.Entity))
 		}
-		return w.RenderResource("google_storage_object_access_control", name, tfACL)
+		return w.RenderResource("google_storage_object_acl", name, tfACL)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


justinsb:
/assign @rifelpet 

I think you introduced this, I wasn't sure which terraform object we should really be using here (do we want to manage all the ACLs?) but I figure we have to use the right schema and this is the smaller change.

rifelpet:
I'm not sure which we should use either. The rest of the fields in that struct suggest I was intending to use [google_storage_object_acl](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_object_acl) which is authoritative whereas [storage_object_access_control](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_object_access_control) is not authoritative.

Do we want this to be authoritative? users wouldn't have a way to grant any other permissions on the objects: we dont expose a way to add Role Entities to this resource's list. In practice I'm not sure if thats something uses would need to do. For reference, in S3 we use ACLs which are discouraged in favor of bucket/object policies, so we aren't authoritative with our S3 object permissions.

johngmyers:
Not competent to review
/uncc

justinsb:
So this is tricky.  I think we're actually in luck because GCS is moving to uniform-bucket-level-access instead of ACLs, so this isn't the recommended path anyway (and elsewhere we're going to support UBLA better).

I think generally the non-authoritative form feels like the better one to use in terraform - it's more extensible, and if the terraform author wants to add more permissions in another object then they should be able to do so.  However, storage_object_access_control only supports a single ACL per terraform object, so we'd have to start figuring out how to name them in the case of multiple ACLs.

Also, in general, I'd say the kOps approach has been to either try to manage objects fully or not manage them at all (external), so from that point of view we should probably just manage the object entirely.

Given all that, I think this is the better approach (and I did try using `google_storage_object_access_control` in another branch and it got tricky with multiple ACLs)

justinsb:
/override pull-kops-e2e-k8s-gce-ipalias

We'll get that test to pass separately :-)

rifelpet (approved):

PR #14955, opened 2023-01-05
Allow built-in manifests to be replaced by external addons

justinsb:
We identify the external manifests by checking for our labels.
Currently that label is kOps specific, and we'll likely have to evolve
that to something ecosystem-netural.

We only support the GCE CCM addon and the kopeio-networking addon at
first.

For the GCE CCM addon, we need to replace the arguments, in particular
we likely need the Pod CIDR.  Here we need to work with the GCE CCM to
find a mechanism that can allow some of these flags to be communicated
via a more extensible mechanism (env vars or config maps, likely,
though possibly CRDs).

This is all behind the ClusterAddons feature flag at the moment, so we
can figure this out with other projects safely.

### DIFF
--- a/upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
+++ b/upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
@@ -870,7 +870,47 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.CloudupModelBuilderContext) 
 		if b.Cluster.Spec.GetCloudProvider() == kops.CloudProviderGCE {
 			if b.Cluster.Spec.ExternalCloudControllerManager != nil {
 				key := "gcp-cloud-controller.addons.k8s.io"
-				{
+				useBuiltin := !b.hasExternalAddon(key)
+
+				if !useBuiltin {
+					klog.Infof("Found cloud-controller-manager in addons; won't use builtin")
+
+					// Until we make the manifest extensible, we still need to inject our arguments.
+					// TODO(justinsb): we don't really want to do this, it limits the ability for users to override things.
+					// However, this is behind a feature flag at the moment, and this way we can work towards something better.
+					gkDaemonset := schema.GroupKind{Group: "apps", Kind: "DaemonSet"}
+					for _, addon := range b.ClusterAddons {
+						if addon.GroupVersionKind().GroupKind() == gkDaemonset &&
+							addon.GetName() == "cloud-controller-manager" &&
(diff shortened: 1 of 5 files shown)


upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
```diff
@@ -870,7 +870,47 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.CloudupModelBuilderContext)
...
+
+							klog.Infof("replacing arguments in externally provided cloud-controller-manager")
+
+							fnAny, ok := b.templates.TemplateFunctions["CloudControllerConfigArgv"]
+							if !ok {
+								return nil, nil, fmt.Errorf("unable to find TemplateFunction CloudControllerConfigArgv")
+							}
+							fn, ok := fnAny.(func() ([]string, error))
+							if !ok {
+								return nil, nil, fmt.Errorf("unexpected typed for TemplateFunction CloudControllerConfigArgv: %T", fnAny)
```
johngmyers, line 895:
```suggestion
								return nil, nil, fmt.Errorf("unexpected type for TemplateFunction CloudControllerConfigArgv: %T", fnAny)
```

upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
```diff
@@ -870,7 +870,47 @@ func (b *BootstrapChannelBuilder) buildAddons(c *fi.CloudupModelBuilderContext)
...
+							}
+							args, err := fn()
+							if err != nil {
+								return nil, nil, fmt.Errorf("error from TemplateFunction CloudControllerConfigArgv: %w", err)
```
johngmyers, line 899:
```suggestion
								return nil, nil, fmt.Errorf("in TemplateFunction CloudControllerConfigArgv: %w", err)
```

johngmyers (commented):
A few nits, but otherwise looks good.

upup/pkg/fi/cloudup/bootstrapchannelbuilder/bootstrapchannelbuilder.go
justinsb: Thanks - merged this and the other one, squashed and made you a co-author

justinsb:
/retest

Test failure likely unrelated

hakman:
I think something (maybe the janitor) deletes the DNS record for the API server. In any case, the test is not "required" to pass.

johngmyers (approved):

PR #14959, opened 2023-01-06
move openstack cloud config to k8s secrets

zetaab:
This will move openstack cloud.config usage to k8s secrets instead.

after this etcd-manager is only component that really needs cloud.config from control-plane local volume. in-tree configs can be removed afaik (remove the --cloud-config flag from kubelet, apiserver, control-manager?). https://github.com/kubernetes-sigs/etcdadm/blob/master/etcd-manager/pkg/volumes/openstack/volumes.go#L268

### DIFF
--- a/nodeup/pkg/model/cloudconfig.go
+++ b/nodeup/pkg/model/cloudconfig.go
@@ -19,11 +19,11 @@ package model
 import (
 	"encoding/json"
 	"fmt"
-	"os"
 	"strings"
 
 	"k8s.io/kops/pkg/apis/kops"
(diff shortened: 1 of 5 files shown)


johngmyers (approved):
It's a little disconcerting that there are no integration tests, but I suppose no Terraform (or other Target suitable for testing)

How did the environment variables get passed to nodeup? Can some of those be no longer passed to nodeup?

For future PRs, I'm not a fan of using environment variables for the non-credential things such as `OS_REGION_NAME`. They would be better put in the cluster spec since they are specific to the cluster, not the user.

PR #14960, opened 2023-01-07
Add VFSContext to various clientsets

johngmyers:
`VFSContext` is not a context, it is a clientset. This starts to refactor some references to `VFSContext` so that the pull it from a higher-level clientset.

There will be more work to push either a `VFSContext` or a higher-level clientset into the things that currently reference the global variable. The next thing to hold a reference to `VFSContext` is probably `ModelBuilderContext`.

/cc @justinsb

### DIFF
--- a/pkg/instancegroups/rollingupdate_os_test.go
+++ b/pkg/instancegroups/rollingupdate_os_test.go
@@ -60,7 +60,7 @@ func getTestSetupOS(t *testing.T, ctx context.Context) (*RollingUpdateCluster, *
 
 	assetBuilder := assets.NewAssetBuilder(inCluster.Spec.Assets, inCluster.Spec.KubernetesVersion, false)
 	basePath, _ := vfs.Context.BuildVfsPath(inCluster.Spec.ConfigBase)
-	clientset := vfsclientset.NewVFSClientset(basePath)
+	clientset := vfsclientset.NewVFSClientset(vfs.Context, basePath)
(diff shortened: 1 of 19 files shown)


pkg/instancegroups/rollingupdate_os_test.go
```diff
@@ -60,7 +60,7 @@ func getTestSetupOS(t *testing.T, ctx context.Context) (*RollingUpdateCluster, *
 
 	assetBuilder := assets.NewAssetBuilder(inCluster.Spec.Assets, inCluster.Spec.KubernetesVersion, false)
 	basePath, _ := vfs.Context.BuildVfsPath(inCluster.Spec.ConfigBase)
-	clientset := vfsclientset.NewVFSClientset(basePath)
+	clientset := vfsclientset.NewVFSClientset(vfs.Context, basePath)
```
justinsb, line 63: We may need to think of a better name for this vfsclientset.  Maybe `StateStoreClient` (the alternative is when we back with kubernetes CRDs, which is still experimental).  Not a blocker though!

justinsb:
Thanks @johngmyers 

/approve
/lgtm

pkg/instancegroups/rollingupdate_os_test.go
johngmyers: We could call it `VFSSimpleClientset` or `VFSKopsClientset`

PR #14963, opened 2023-01-07
nodeup: default nonMasqueradeCIDR to the podCIDR

justinsb:
Rather than failing, we can at least try the podCIDR here, as it is
likely the value we want.

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -407,12 +407,17 @@ iptables -w -t nat -A IP-MASQ -d {{.NonMasqueradeCIDR}} -m comment --comment "ip
 iptables -w -t nat -A IP-MASQ -m comment --comment "ip-masq: outbound traffic is subject to MASQUERADE (must be last in chain)" -j MASQUERADE
 `
 
-	if b.Cluster.Spec.Networking.NonMasqueradeCIDR == "" {
-		// We could fall back to the pod CIDR, that is likely more correct anyway
-		return fmt.Errorf("NonMasqueradeCIDR is not set")
+	nonMasqueradeCIDR := b.Cluster.Spec.Networking.NonMasqueradeCIDR
+	if nonMasqueradeCIDR == "" {
+		if podCIDR := b.Cluster.Spec.Networking.PodCIDR; podCIDR != "" {
+			klog.Warningf("defaulting nonMasqueradeCIDR to podCIDR %q", podCIDR)
+			nonMasqueradeCIDR = podCIDR
+		} else {
+			return fmt.Errorf("nonMasqueradeCIDR is not set (and podCIDR is also empty)")
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


nodeup/pkg/model/containerd.go
```diff
@@ -407,12 +407,17 @@ iptables -w -t nat -A IP-MASQ -d {{.NonMasqueradeCIDR}} -m comment --comment "ip
...
 `
 
-	if b.Cluster.Spec.Networking.NonMasqueradeCIDR == "" {
-		// We could fall back to the pod CIDR, that is likely more correct anyway
-		return fmt.Errorf("NonMasqueradeCIDR is not set")
+	nonMasqueradeCIDR := b.Cluster.Spec.Networking.NonMasqueradeCIDR
+	if nonMasqueradeCIDR == "" {
+		if podCIDR := b.Cluster.Spec.Networking.PodCIDR; podCIDR != "" {
+			klog.Warningf("defaulting nonMasqueradeCIDR to podCIDR %q", podCIDR)
+			nonMasqueradeCIDR = podCIDR
```
johngmyers, line 414: It's going to masquerade the `ServiceClusterIPRange`? That doesn't seem correct.

nodeup/pkg/model/containerd.go
```diff
@@ -407,12 +407,17 @@ iptables -w -t nat -A IP-MASQ -d {{.NonMasqueradeCIDR}} -m comment --comment "ip
 iptables -w -t nat -A IP-MASQ -m comment --comment "ip-masq: outbound traffic is subject to MASQUERADE (must be last in chain)" -j MASQUERADE
 `
 
-	if b.Cluster.Spec.Networking.NonMasqueradeCIDR == "" {
-		// We could fall back to the pod CIDR, that is likely more correct anyway
-		return fmt.Errorf("NonMasqueradeCIDR is not set")
+	nonMasqueradeCIDR := b.Cluster.Spec.Networking.NonMasqueradeCIDR
+	if nonMasqueradeCIDR == "" {
+		if podCIDR := b.Cluster.Spec.Networking.PodCIDR; podCIDR != "" {
+			klog.Warningf("defaulting nonMasqueradeCIDR to podCIDR %q", podCIDR)
```
johngmyers, line 413: Why not populate `NonMasqueradeCIDR` in the completed cluster spec, perhaps in `performNetworkAssignmentsIPAliases()`?

If iptables is disabling masquerading for the CIDR, then it truly is the `NonMasqueradeCIDR`. Why should GCP in particular bind the value so much later than other cloud providers?

olemarkus: I would also do this in cloudup. Doing it here would be a bit surprising and I would probably not think to look here.

nodeup/pkg/model/containerd.go
justinsb: The service ips are dealt with "earlier" ... if we put a service IP onto the wire (i.e. dev eth0), then - yes - we would NAT it.

PR #14988, opened 2023-01-12
etcd-manager: Add option to set backup retention

hakman:


pkg/apis/kops/cluster.go
```diff
@@ -642,6 +642,8 @@ type EtcdManagerSpec struct {
 	Env []EnvVar `json:"env,omitempty"`
 	// BackupInterval which is used for backups. The default is 15 minutes.
 	BackupInterval *metav1.Duration `json:"backupInterval,omitempty"`
+	// BackupRetention which is used for backups. The default is 90 days.
```
johngmyers, line 645: This doesn't give much information on what the syntax of the string is. Neither do the validation errors.

pkg/apis/kops/validation/validation.go
```diff
@@ -1286,6 +1288,37 @@ func validateEtcdClusterSpec(spec kops.EtcdClusterSpec, c *kops.Cluster, fieldPa
...
+}
+
+// validateEtcdBackupRetention checks that backupRetention can be parsed by ParseHumanDuration() from etcd-manager.
+func validateEtcdBackupRetention(backupRetention string, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if strings.HasSuffix(backupRetention, "d") || strings.HasSuffix(backupRetention, "y") {
+		s := backupRetention
+		s = strings.TrimSuffix(s, "d")
+		s = strings.TrimSuffix(s, "y")
```
johngmyers, line 1308: No `w` or `m`?

pkg/apis/kops/cluster.go
hakman: Any suggestion on how to explain it?

pkg/apis/kops/validation/validation.go
hakman: https://github.com/kubernetes-sigs/etcdadm/blob/ccd19ba88d5cb47e09482e5f8b7211318f5e2504/etcd-manager/pkg/backupcontroller/cleanup.go#L41-L62

pkg/apis/kops/cluster.go
hakman: One idea would be to rename the field to `BackupRetentionDays`. It should 99.99% of the use cases. WDYT?

pkg/apis/kops/v1alpha2/cluster.go
```diff
@@ -665,8 +665,10 @@ type EtcdManagerSpec struct {
 	// This allows etcd setting to be configured/overwriten. No config validation is done.
 	// A list of etcd config ENV vars can be found at https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/configuration.md
 	Env []EnvVar `json:"env,omitempty"`
-	// BackupInterval which is used for backups. The default is 15 minutes.
+	// BackupInterval which is used for backups.
```
johngmyers, line 668: Why is the mention of a default removed from v1alpha2 but not the internal or v1alpha3 APIs?

pkg/apis/kops/v1alpha2/cluster.go
```diff
@@ -665,8 +665,10 @@ type EtcdManagerSpec struct {
...
 	BackupInterval *metav1.Duration `json:"backupInterval,omitempty"`
+	// BackupRetentionDays which is used for backups. The default is 90 days.
```
johngmyers, line 670: Why is the default only mentioned in v1alpha2?

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1433,8 +1433,12 @@ func initializeOpenstack(opt *NewClusterOptions, cluster *api.Cluster) {
 }
 
 func createEtcdCluster(etcdCluster string, controlPlanes []*api.InstanceGroup, encryptEtcdStorage bool, etcdStorageType string) api.EtcdClusterSpec {
-	etcd := api.EtcdClusterSpec{}
-	etcd.Name = etcdCluster
+	etcd := api.EtcdClusterSpec{
+		Name: etcdCluster,
+		Manager: &api.EtcdManagerSpec{
+			BackupRetentionDays: fi.PtrTo[uint32](90),
```
johngmyers, line 1439: Why explicitly lock this down in new clusters?

pkg/model/components/etcdmanager/model.go
```diff
@@ -342,6 +343,13 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
 		config.BackupInterval = fi.PtrTo(etcdCluster.Manager.BackupInterval.Duration.String())
 	}
 
+	if etcdCluster.Manager != nil && etcdCluster.Manager.BackupRetentionDays != nil {
```
johngmyers, line 346: This doesn't appear to be defaulting to 90 days; it appears to default to not setting a retention. Or does etcd-manager do the 90-day default?

pkg/apis/kops/v1alpha2/cluster.go
hakman: Typo. Fixed.

pkg/apis/kops/v1alpha2/cluster.go
hakman: Not sure how to put it. There is a new default for newly created clusters, otherwise the etcd-manager default, which varies.

upup/pkg/fi/cloudup/new_cluster.go
hakman: 1. don't want to change behaviour for those that expect it to be longer
2. want it explicit to make it easy to change by operators

pkg/model/components/etcdmanager/model.go
hakman: I don't remember agreeing to change the behaviour for existing clusters. If I'm mistaking, I don't mind updating the PR.

hakman:
@johngmyers Could you take another look at this? Thanks!

pkg/model/components/etcdmanager/model.go
justinsb: Let's discuss today, I think we should change the default if unspecified to match the 90 days we documented, but I think changing the default wouldn't go specifically here anyway (it should go where we populate the full cluster config, so it is more self-documenting).

upup/pkg/fi/cloudup/new_cluster.go
justinsb: Right - the idea (AIUI) is to populate the value explicitly (only) for new clusters created via `kops create cluster`, so users see the value and can easily change it.

justinsb:
I think this LGTM but will hold so we can discuss in office hours today seeing as we're so close to it!

/approve
/lgtm
/hold

johngmyers:
The API documentation says the default is 90 days. When the API documents a default, that means the value one gets when the option is not set. That is not the case here, so the documentation is incorrect.

johngmyers:
We should have a release note about the new default.

If we have a new default in the API then I don't see the reason to explicitly specify it in new clusters.

docs/operations/etcd_backup_restore_encryption.md
```diff
@@ -22,8 +22,8 @@ Backups are done periodically and before cluster modifications using [etcd-manag
 are stored in object storage (like S3) together with the cluster configuration.
 
 By default, backups are taken every 15 min. Hourly backups are kept for 1 week and
-daily backups are kept for 1 year, before being automatically removed. The retention
-duration for backups [can be adjusted](../cluster_spec.md#etcd-backups-retention)
+daily backups are kept for 90 days (or 2 years before kOps 1.27), before beingautomatically removed.
```
johngmyers, line 25:
```suggestion
daily backups are kept for 90 days (or 2 years before kOps 1.27), before being automatically removed.
```

johngmyers (changes requested):

johngmyers (approved):

PR #14991, opened 2023-01-13
Allow custom service account issuer without public bucket

hakman:
/cc @olemarkus @johngmyers

### DIFF
--- a/pkg/model/issuerdiscovery.go
+++ b/pkg/model/issuerdiscovery.go
@@ -87,13 +87,22 @@ func (b *IssuerDiscoveryModelBuilder) Build(c *fi.CloudupModelBuilderContext) er
 
 	switch discoveryStore := discoveryStore.(type) {
 	case *vfs.S3Path:
-		isPublic, err := discoveryStore.IsBucketPublic(ctx)
+		discoveryStoreURL, err := discoveryStore.GetHTTPsUrl(b.Cluster.Spec.IsIPv6Only())
 		if err != nil {
-			return fmt.Errorf("checking if bucket was public: %w", err)
+			return err
 		}
-		if !isPublic {
-			klog.Infof("serviceAccountIssuers bucket %q is not public; will use object ACL", discoveryStore.Bucket())
-			publicFileACL = fi.PtrTo(true)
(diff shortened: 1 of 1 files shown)


johngmyers:
As stated in Office Hours, I would like there to be more documentation around this use case.

johngmyers:
I read over the existing documentation. It states:

> The `enableAWSOIDCProvider` configures AWS to trust the service account issuer to
authenticate service accounts for IAM Roles for Service Accounts (IRSA). In order for this to work,
the service account issuer discovery URL must be publicly readable.

This is true, though it should probably more specifically state that `spec.kubeAPIServer.serviceAccountIssuer` is what must be publicly readable.

The must-be-publicly-readable requirement could be made conditional on `enableAWSOIDCProvider: true` since it is AWS IAM that makes that requirement. Technically, we should be applying the requirement of being publicly readable to `serviceAccountIssuer` instead of `discoveryStore`, but that's a bit difficult now.

We should document the conditions under which kOps sets object-level ACLs. I'm wondering if we should deprecate the object-level ACLs.

olemarkus:
> We should document the conditions under which kOps sets object-level ACLs. I'm wondering if we should deprecate the object-level ACLs.

There is something off with the logic today, I believe. But the object-level ACL is somewhat critical in the cases where public buckets are used as state store (like e.g our e2e tests). But the ACL should only tighten the restrictions. Right now, it is also trying to loosen the restrictions (public objects in private buckets), which fail.

johngmyers:
I was referring to the `PublicACL` field of `ManagedFile`.

There's other code that sets public-read for the `fileRepository` on S3. That should probably also be deprecated. The `fileRepository` doesn't have to be public in the first place.

johngmyers:
I think what we need is to update the documentation in `cluster_spec.md` stating the conditions in which it will set object-level public-read ACLs, namely that it is in S3, there is no whole-bucket public ACL, and the `kubeAPIServer.serviceAccountIssuer` hasn't been overridden to point somewhere else.

justinsb:
We discussed in office hours, this unblocks some users hitting this problem, so we're going to merge this and hopefully document it more fully over time.

Thanks @hakman 

/approve
/lgtm

PR #14996, opened 2023-01-13
Determine default API access method by IG subnet type

johngmyers:
I believe this is the last substantive reference to the control-plane and node topology field in the cluster spec.

### DIFF
--- a/upup/pkg/fi/cloudup/populate_cluster_spec.go
+++ b/upup/pkg/fi/cloudup/populate_cluster_spec.go
@@ -44,6 +44,9 @@ type populateClusterSpec struct {
 	// We build it up into a complete config, but we write the values as input
 	InputCluster *kopsapi.Cluster
 
+	// InputInstanceGroups are the instance groups in the cluster
+	InputInstanceGroups []*kopsapi.InstanceGroup
+
 	// fullCluster holds the built completed cluster spec
 	fullCluster *kopsapi.Cluster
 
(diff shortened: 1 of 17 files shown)


upup/pkg/fi/cloudup/populate_cluster_spec.go
```diff
@@ -217,6 +221,47 @@ func (c *populateClusterSpec) run(ctx context.Context, clientset simple.Clientse
...
+				if ig.Spec.Role == kopsapi.InstanceGroupRoleAPIServer {
+					haveAPIServerNodes = true
+				}
+			}
+			for _, ig := range c.InputInstanceGroups {
+				if ig.Spec.Role == kopsapi.InstanceGroupRoleAPIServer || (!haveAPIServerNodes && ig.Spec.Role == kopsapi.InstanceGroupRoleControlPlane) {
+					for _, subnet := range ig.Spec.Subnets {
+						switch subnetTypesByName[subnet] {
+						case kopsapi.SubnetTypePrivate:
+							if cluster.Spec.API.LoadBalancer == nil {
```
hakman, line 243:
```suggestion
							if cluster.Spec.API.LoadBalancer == nil && cluster.Spec.API.DNS == nil {
```

upup/pkg/fi/cloudup/populate_cluster_spec.go
```diff
@@ -217,6 +221,47 @@ func (c *populateClusterSpec) run(ctx context.Context, clientset simple.Clientse
...
+								cluster.Spec.API.LoadBalancer = &kopsapi.LoadBalancerAccessSpec{}
+							}
+						case kopsapi.SubnetTypePublic:
+							if cluster.Spec.API.DNS == nil {
```
hakman, line 247:
```suggestion
							if cluster.Spec.API.LoadBalancer == nil && cluster.Spec.API.DNS == nil {
```

hakman (changes requested):

upup/pkg/fi/cloudup/populate_cluster_spec.go
johngmyers: Actually, either `if` condition will always be true due to the enclosing `if` statement.

upup/pkg/fi/cloudup/populate_cluster_spec.go
johngmyers: Actually, either `if` condition will always be true due to the enclosing `if` statement.

upup/pkg/fi/cloudup/populate_cluster_spec.go
johngmyers: Oh, nevermind, I had it correct before. The for loop means that if there's a mix of subnet types both could get set. So we pick the one that is listed first.

johngmyers: Hrm, that's not what my previous code did; it would assign both. I think if there's both it should go with loadbalancer.

hakman: Should one have both public and private IGs for API?

hakman: Also, the DNS part is probably valid just for AWS.

johngmyers: Having both public and private IGs for API would be  unusual. The DNS part was in the previous code, for the case of TopologyPublic for the control plane.

hakman (approved):

PR #14997, opened 2023-01-14
Validate nonMasqueradeCIDR doesn't overlap additionalNetworkCIDRs

johngmyers:


pkg/apis/kops/validation/validation.go
```diff
@@ -290,17 +291,21 @@ func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *fie
...
 type cloudProviderConstraints struct {
-	requiresSubnets      bool
-	requiresNetworkCIDR  bool
-	prohibitsNetworkCIDR bool
-	requiresSubnetCIDR   bool
+	requiresSubnets                                 bool
+	requiresNetworkCIDR                             bool
+	prohibitsNetworkCIDR                            bool
+	requiresNonMasqueradeCIDR                       bool
+	requiresServiceClusterSubnetOfNonMasqueradeCIDR bool
```
justinsb, line 298: I thought this was the case everywhere (per discussion in other PR), but maybe it isn't?

(Also, naming nit - maybe we should call it ServiceCIDR instead of ServiceClusterSubnet, because I don't think anyone actually allocates a cloud "subnetwork" to it.  But maybe this is where I'm getting confused!)

pkg/apis/kops/validation/validation.go
```diff
@@ -907,6 +924,47 @@ func validateNetworking(cluster *kops.Cluster, v *kops.NetworkingSpec, fldPath *
...
+	var nonMasqueradeCIDRs []*net.IPNet
+	{
+		if v.NonMasqueradeCIDR == "" {
+			if providerConstraints.requiresNonMasqueradeCIDR {
+				allErrs = append(allErrs, field.Required(fldPath.Child("nonMasqueradeCIDR"), "Cluster does not have nonMasqueradeCIDR set"))
+			}
+		} else {
+			allErrs = append(allErrs, validateCIDR(v.NonMasqueradeCIDR, fldPath.Child("nonMasqueradeCIDR"), &nonMasqueradeCIDRs)...)
+
+			if strings.Contains(v.NonMasqueradeCIDR, ":") && v.NonMasqueradeCIDR != "::/0" {
```
justinsb, line 936: Nit: this always sets off my spidey-sense, because I think we've all been burned by this comparison many times!  we should create an IsIPv6 helper (and ideally we would use net.CIDR instead of string comparison).  Not a blocker though

justinsb:
Generally seems good to me, I assume this was hit accidentally by somebody?

I also thought that it's OK for a service cidr to not overlap the nonMasqueradeCIDR, because those IPs should be rewritten before we get to the masquerade rule, but maybe I'm missing something.

My nits are pre-existing, I believe, so not blockers.

/approve
/lgtm

/hold in case we want to discuss the service cidr, but that was pre-existing

pkg/apis/kops/validation/validation.go
johngmyers: The current semantics are that GCE doesn't require this. If we bring GCE in line with other cloud providers then we can remove this exception.

The reading was intended to be "the cloud provider _requires_ the _SericeCluster_ be a _subnet_ of the _nonMasquradeCIDR_".

johngmyers:
I think there might be a dual meaning of NonMasqueradeCIDR: addresses that shouldn't be masqueraded and addresses that you can allocate cluster-scoped networks out of.

Certainly cluster-scoped networks don't need to be masqueraded. You're pointing out that, technically, some cluster-scoped networks are a "don't care" case for masquerading.

I think there's more value in having a single CIDR for all the cluster-scoped networks than allowing the cluster-scoped networks to be disjoint. We don't currently have a separate check that the ServiceCIDR doesn't overlap a Subnet, so that won't get caught on GCE.

This work is in preparation for combining `NetworkCIDR` and `AdditionalNetworkCIDRs` into a single field. Right now the `AdditionalNetworkCIDRs` aren't considered for things they should be. I don't think you can currently allocate a Subnet out of them, which kinda defeats their point.

johngmyers:
I think the handling of the service CIDR can be adjusted in followup work.
/hold cancel

johngmyers:
I suppose in non-overlay CNIs the Pod CIDRs aren't local-scoped, so the combination of them and the service CIDR doesn't make sense.

I suppose we do need to add that check that subnet CIDRs don't overlap the service CIDR.

PR #14999, opened 2023-01-14
Use kops-controller to boostrap nodes in OpenStack

zetaab:
I have tested this PR using gossip and None dns clusters in OpenStack. Both will bootstrap fine and everything is working like should.


pkg/model/bootstrapscript.go
```diff
@@ -167,9 +167,9 @@ func (b *BootstrapScript) buildEnvironmentVariables(cluster *kops.Cluster) (map[
 			)
 		}
 
-		// credentials needed always when using swift but when using None dns only in control plane
+		// credentials needed always in control-plane and when using gossip also in nodes
```
zetaab, line 170: I cannot test the behaviour with openstack designate, hopefully it will work

pkg/model/bootstrapscript.go
```diff
@@ -167,10 +167,10 @@ func (b *BootstrapScript) buildEnvironmentVariables(cluster *kops.Cluster) (map[
 			)
 		}
 
-		// credentials needed always when using swift but when using None dns only in control plane
-		passEnvs := true
-		if !strings.HasPrefix(cluster.Spec.ConfigBase, "swift://") && cluster.UsesNoneDNS() && !b.ig.IsControlPlane() {
-			passEnvs = false
+		// credentials needed always in control-plane and when using gossip also in nodes
```
zetaab, line 170: I cannot test the behaviour with openstack designate, hopefully it will work. It should not do any dns changes from worker nodes? If kops uses route53, the dns changes are done by dns-controller and that is running always in control-plane?

hakman:
@zetaab Please move the commit multiple ConfigServers to a separate PR.

zetaab:
@hakman done

hakman (approved):

upup/pkg/fi/cloudup/openstack/verifier.go
```diff
@@ -0,0 +1,117 @@
...
+	err = mapstructure.Decode(instance.Addresses, &addresses)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode addresses: %w", err)
+	}
+
+	for _, addrList := range addresses {
+		for _, props := range addrList {
+			addrs = append(addrs, props.Addr)
+		}
+	}
```
johngmyers, line 106: I don't see this verifying the body at all. How do we know the body came from the instance, as opposed to something that received a copy of the metadata ServerID from an unrelated request from the instance?

upup/pkg/fi/cloudup/openstack/authenticator.go
```diff
@@ -0,0 +1,42 @@
...
+func NewOpenstackAuthenticator() (bootstrap.Authenticator, error) {
+	return &openstackAuthenticator{}, nil
+}
+
+func (o openstackAuthenticator) CreateToken(body []byte) (string, error) {
+	metadata, err := GetLocalMetadata()
+	if err != nil {
+		return "", fmt.Errorf("unable to fetch metadata: %w", err)
+	}
+	return OpenstackAuthenticationTokenPrefix + metadata.ServerID, nil
```
johngmyers, line 41: The ServerID looks like it's a UUID. Is it actually a secret? Couldn't something that is not the instance get the UUID and then impersonate the instance?

It's sending the UUID as a plaintext password, which is incredibly weak.

johngmyers (commented):
I am concerned that the authentication mechanism here, a plaintext password sent in the clear, is far too weak. At the very least it should send an HMAC over the request body and a shared secret.

I'm also concerned that the secret, the UUID of the instance, might be easily accessible to adversaries.

upup/pkg/fi/cloudup/openstack/authenticator.go
zetaab: @johngmyers yep its weak. I was thinking like, should I use combination of some variables? this is implemented in same way than hetzner was https://github.com/kubernetes/kops/pull/14460#discussion_r1011368171

PR #15000, opened 2023-01-14
support multiple ConfigServers

zetaab:
Instead of supporting one boostrap ConfigServer, I modified it to understand multiple ConfigServers. The logic is that it will try to find the working one from the all available configservers. It seems that Kops-controller will answer only from one master


upup/pkg/fi/nodeup/command.go
```diff
@@ -162,7 +162,9 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
 			return fmt.Errorf("error parsing BootConfig config response: %v", err)
 		}
 		nodeupConfigHash = sha256.Sum256([]byte(nodeConfig.NodeupConfig))
-		nodeupConfig.CAs[fi.CertificateIDCA] = bootConfig.ConfigServer.CACertificates
+		if len(bootConfig.ConfigServers) > 0 {
+			nodeupConfig.CAs[fi.CertificateIDCA] = bootConfig.ConfigServers[0].CACertificates
```
zetaab, line 166: this is little bit dirty, but all configservers do have same CA Certificate

pkg/apis/nodeup/config.go
```diff
@@ -115,8 +115,8 @@ type BootConfig struct {
 	CloudProvider kops.CloudProviderID
 	// ConfigBase is the base VFS path for config objects.
 	ConfigBase *string `json:",omitempty"`
-	// ConfigServer holds the configuration for the configuration server.
-	ConfigServer *ConfigServerOptions `json:",omitempty"`
+	// ConfigServers holds the configuration for the configuration servers.
+	ConfigServers []*ConfigServerOptions `json:",omitempty"`
```
hakman, line 119: Not quite sure if `ConfigServers` or adding `Servers` to `ConfigServerOptions`. `Server = Servers[0]` could be kept around for backward compatibility.

upup/pkg/fi/nodeup/command.go
```diff
@@ -760,20 +762,34 @@ func getNodeConfigFromServer(ctx context.Context, bootConfig *nodeup.BootConfig,
 		Resolver:      resolver,
 	}
 
-	u, err := url.Parse(bootConfig.ConfigServer.Server)
-	if err != nil {
-		return nil, fmt.Errorf("unable to parse configuration server url %q: %w", bootConfig.ConfigServer.Server, err)
-	}
-	client.BaseURL = *u
-	client.CAs = []byte(bootConfig.ConfigServer.CACertificates)
+	return loopThroughConfigservers(ctx, client, bootConfig.ConfigServers)
```
hakman, line 765: I don't think an extra function is needed for now.

upup/pkg/fi/nodeup/command.go
```diff
@@ -760,20 +762,34 @@ func getNodeConfigFromServer(ctx context.Context, bootConfig *nodeup.BootConfig,
...
-		APIVersion:        nodeup.BootstrapAPIVersion,
-		IncludeNodeConfig: true,
+// loopThroughConfigservers returns BootstrapResponse we need to loop through all configservers because only one of them is active. Normally loadbalancer healthcheck will take care of detecting the active one.
+// However, like in case of OpenStack there are 3 config servers and no loadbalancer in front of which means that we need loop through all of them to see which one is the active.
+func loopThroughConfigservers(ctx context.Context, client *kopscontrollerclient.Client, configServers []*nodeup.ConfigServerOptions) (*nodeup.BootstrapResponse, error) {
+	var err error
+	for _, configServer := range configServers {
+		u, loopErr := url.Parse(configServer.Server)
+		if loopErr != nil {
+			return nil, fmt.Errorf("unable to parse configuration server url %q: %w", configServer.Server, loopErr)
```
hakman, line 775: Would be simpler to just use `"go.uber.org/multierr"`:
```suggestion
	var merr error
	for _, configServer := range configServers {
		u, loopErr := url.Parse(configServer.Server)
		if loopErr != nil {
			merr = multierr.Append(merr, fmt.Errrorf("unable to parse configuration server url %q: %w", configServer.Server, loopErr))
```

upup/pkg/fi/nodeup/command.go
```diff
@@ -760,20 +762,34 @@ func getNodeConfigFromServer(ctx context.Context, bootConfig *nodeup.BootConfig,
...
+		request := nodeup.BootstrapRequest{
+			APIVersion:        nodeup.BootstrapAPIVersion,
+			IncludeNodeConfig: true,
+		}
+		var resp nodeup.BootstrapResponse
+		err = client.Query(ctx, &request, &resp)
+		if err == nil {
+			return &resp, nil
+		} else {
+			klog.Infof("loopThroughConfigservers error %w", err)
```
hakman, line 789:
```suggestion
			merr = multierr.Append(merr, err)
```

hakman (changes requested):
I am generally ok with the direction. Leaving the final review to @justinsb or @johngmyers.
/assign @johngmyers 
/assign @justinsb

johngmyers:
I'll have to research why it only answers from one instance.

johngmyers:
Okay, this was introduced in #12997 

The server should not be put under control of the manager, since that subjects it to leader election.

johngmyers:
@zetaab I'd appreciate if you could try #15002.

hakman:
I think without putting all 3 servers (when available) there is less redundancy. We already send all 3 IPs for the bootstrap check.
Also, seems this is the direction in which @justinsb was going:
https://github.com/kubernetes/kops/pull/14501#discussion_r1020973316
https://github.com/kubernetes/kops/pull/14769#discussion_r1049717186

pkg/apis/nodeup/config.go
johngmyers: It would be better to have `Servers` in `ConfigServerOptions`. The servers aren't going to have different CAs.

upup/pkg/fi/nodeup/command.go
```diff
@@ -721,8 +723,8 @@ func seedRNG(ctx context.Context, bootConfig *nodeup.BootConfig, region string)
 	return nil
 }
 
-// getNodeConfigFromServer queries kops-controller for our node's configuration.
-func getNodeConfigFromServer(ctx context.Context, bootConfig *nodeup.BootConfig, region string) (*nodeup.BootstrapResponse, error) {
+// getNodeConfigFromServer queries kops-controllers for our node's configuration.
```
johngmyers, line 726:
```suggestion
// getNodeConfigFromServers queries kops-controllers for our node's configuration.
```

upup/pkg/fi/nodeup/command.go
johngmyers: Concur

upup/pkg/fi/nodeup/command.go
johngmyers, line 775:
```suggestion
			return nil, fmt.Errorf("parsing configuration server url %q: %w", configServer.Server, loopErr)
```

johngmyers (changes requested):
Next time, when both the change and the changes to golden outputs are nontrivial please put the golden output changes in a separate commit.

upup/pkg/fi/nodeup/command.go
```diff
@@ -163,7 +164,9 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
 			return fmt.Errorf("error parsing BootConfig config response: %v", err)
 		}
 		nodeupConfigHash = sha256.Sum256([]byte(nodeConfig.NodeupConfig))
-		nodeupConfig.CAs[fi.CertificateIDCA] = bootConfig.ConfigServer.CACertificates
+		if len(bootConfig.ConfigServer.Servers) > 0 {
```
johngmyers, line 167: `nodeConfig` can only be non-nil if this condition is true, so there's no need to test it.

upup/pkg/fi/nodeup/command.go
```diff
@@ -767,20 +770,28 @@ func getNodeConfigFromServer(ctx context.Context, bootConfig *nodeup.BootConfig,
 		Resolver:      resolver,
 	}
 
-	u, err := url.Parse(bootConfig.ConfigServer.Server)
-	if err != nil {
-		return nil, fmt.Errorf("unable to parse configuration server url %q: %w", bootConfig.ConfigServer.Server, err)
-	}
-	client.BaseURL = *u
-	client.CAs = []byte(bootConfig.ConfigServer.CACertificates)
```
johngmyers, line 775: No need to move this inside the loop.

johngmyers (approved):
/hold to give opportunity to fix nits

upup/pkg/fi/nodeup/command.go
```diff
@@ -112,8 +113,8 @@ func (c *NodeUpCommand) Run(out io.Writer) error {
 	// If we're using a config server instead of vfs, nodeConfig will hold our configuration
 	var nodeConfig *nodeup.NodeConfig
 
-	if bootConfig.ConfigServer != nil {
-		response, err := getNodeConfigFromServer(ctx, &bootConfig, region)
+	if len(bootConfig.ConfigServer.Servers) > 0 {
```
johngmyers, line 116: You need to keep the previous nil check or it will NPE if no ConfigServer

upup/pkg/fi/nodeup/command.go
```diff
@@ -765,22 +766,30 @@ func getNodeConfigFromServer(ctx context.Context, bootConfig *nodeup.BootConfig,
...
-	if err != nil {
-		return nil, fmt.Errorf("unable to parse configuration server url %q: %w", bootConfig.ConfigServer.Server, err)
-	}
-	client.BaseURL = *u
-	client.CAs = []byte(bootConfig.ConfigServer.CACertificates)
+	var merr error
+	for _, server := range bootConfig.ConfigServer.Servers {
+		u, loopErr := url.Parse(server)
+		if loopErr != nil {
+			merr = multierr.Append(merr, fmt.Errorf("unable to parse configuration server url %q: %w", server, loopErr))
```
hakman, line 776:
```suggestion
		u, err := url.Parse(server)
		if err != nil {
			merr = multierr.Append(merr, fmt.Errorf("unable to parse configuration server url %q: %w", server, loopErr))
			continue			
```

upup/pkg/fi/nodeup/command.go
```diff
@@ -765,22 +766,30 @@ func getNodeConfigFromServer(ctx context.Context, bootConfig *nodeup.BootConfig,
...
+			APIVersion:        nodeup.BootstrapAPIVersion,
+			IncludeNodeConfig: true,
+		}
+		var resp nodeup.BootstrapResponse
+		err := client.Query(ctx, &request, &resp)
+		if err == nil {
+			return &resp, nil
+		} else {
+			merr = multierr.Append(merr, err)
+		}
```
hakman, line 790:
```suggestion
		err = client.Query(ctx, &request, &resp)
		if err != nil {
			merr = multierr.Append(merr, err)
			continue
		}
		return &resp, nil
```

hakman (changes requested):

hakman:
Please restage/squash the commits before merging. Will be easier to cherrypick.
/lgtm

hakman:
I meant to keep all code in one PR and the golden outputs, as @johngmyers said.
When we cherry-pick, most likely the code will not have any changes, but the expected will be regenerated. It's much easier later to track if a commit was ported to a branch or not.

hakman (approved):

PR #15002, opened 2023-01-14
Run kops-controller server on non-leaders as well

johngmyers:
Fixes regression from #12997 which caused only one of the kops-controller pods to respond to client requests.


hakman:
If I remember correctly, one of the ideas was the server response should always come from the latest `kops-controller`, in case the API changes.

johngmyers:
It needs to come from the latest version, yes, but I don't see what that has to do with leader election.

Rolling updates always update the control planes before worker nodes. If a new worker node comes up shortly after an apply_cluster, it will have to wait until it queries a kops-controller on the newer version.

PR #15003, opened 2023-01-14
Document security properties of Hetzner and Openstack node bootstrap

johngmyers:


hakman:
Don't get why you did this. It just puts a lot of pressure to find a better solution.
/hold

johngmyers:
This allows admins to know what it is they have to protect. It also forestalls what would otherwise be the inevitable security bounty claim.

docs/getting_started/openstack.md
```diff
@@ -218,6 +218,16 @@ spec:
       insecureSkipVerify: true
 ```
 
-## Next steps
+# Security of node bootstrap
+
+As of kOps 1.27, the node bootstrap process uses an instance's UUID as a plaintext password.
```
zetaab, line 223: well I am thinking to make it at least little bit better: if I can somehow to verify that the request ip address is the one which the instance should have. And also if there is way to make better password algorithm

johngmyers: Does Openstack prevent interception/spoofing of IP within the network like Amazon VPC does?

Would it be possible to pass a per-instance credential through userdata or vendordata? The `Instance` task could mint a specific, time-limited credential if it had a way to pass it to the instance reasonably securely.

zetaab: @johngmyers yep there is thing called "portsecurity", you cannot fake your ip address.

Is the credential needed only one time? I am just thinking is it one possible way to check is instance < 1 hour old in verifier? After that it will not accept requests

docs/getting_started/openstack.md
```diff
@@ -218,6 +218,16 @@ spec:
...
+Anything able to obtain this UUID and query port 3988 on the control plane will be
+able to impersonate the instance and join the cluster as that node.
+
+Furthermore, any instance in the account which has a `KopsInstanceGroup` tag
```
zetaab, line 227: this is not true. The server needs to be in same OpenStack network, at the same time it means that user needs to have admin access to OpenStack project in order to provision these servers/modify metadata in existing servers. If someone gets somehow admin access to OpenStack project, it is much more bigger issue than kops-controller auth. Basically this can be compared that someone will get access in AWS to provision ec2 instances / modify existing ones. Why we are not documenting that issue?

zetaab: also instance needs to be in nodes security group, the 3988 port is open only for that.

PR #15004, opened 2023-01-15
Move some networking settings to nodeup.Config

johngmyers:


hakman (approved):

zetaab:
@johngmyers I think this is breaking master currently?


I am trying to create new cluster like previously:

```
Jan 16 06:32:42 control-plane-esptnl-xpuosc nodeup[1094]: W0116 06:32:42.189482    1094 main.go:133] got error running nodeup (will retry in 30s): error building loader: building *model.KubeAPIServerBuilder: error parsing ServiceClusterIPRange "": invalid CIDR address:
^C
```

johngmyers:
That failure makes no sense. `ServiceClusterIPRange` is unconditionally set in `populate_cluster_spec` and unconditionally copied into `nodeup.Config`. Are you sure you don't have a newer nodeup than kops?

zetaab:
@johngmyers I am compiling always new version of nodeup + kops cli using `make upload`. So yes: I am 100% sure that I am using same version from both.

zetaab:
It might be that I was wrong: I compiled everything again and it is working. So it might be that I had older kops cli than nodeup

PR #15007, opened 2023-01-15
OpenStack: add request address check for node bootstrap

zetaab:
still this is not perfect. In order to use something like hmac, we need somehow share the secret with worker nodes. The current situation is that there are no secrets in user_data or no credentials to state store. I would not like to add any secrets to user_data when its now cleaned.


hakman:
Would the NodeupConfigHash work as secret?

PR #15008, opened 2023-01-15
Use jwt RS256 between kops-controller and worker node

zetaab:


zetaab:
@hakman @johngmyers could you check this. 

Disclaimer: it does not work. In order to work it needs certificates mounted to kops-controller and kops-controller needs to verify JWT token against certificates.

But what is your idea, could this be something that I could continue with or do you have better ideas how to implement this feature to cloudproviders which does not provide any method to sign requests from instances.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1422,6 +1424,36 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+			key, err := jwt.ParseRSAPrivateKeyFromPEM(b)
+			if err != nil {
+				return nil, nil, fmt.Errorf("parsing pem key failed: %w", err)
+			}
+			now := time.Now().UTC()
+			claims := jwt.MapClaims{
+				"ig":  ig.Name,
+				"iat": now.Unix(),
+				"nbf": now.Unix(),
+				"exp": now.Add(time.Hour).Unix(),
```
zetaab, line 1447: the idea is that JWT token is short lived only one hour long

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1422,6 +1424,36 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+				return nil, nil, fmt.Errorf("unable to find privatekey: %w", err)
+			}
+
+			key, err := jwt.ParseRSAPrivateKeyFromPEM(b)
+			if err != nil {
+				return nil, nil, fmt.Errorf("parsing pem key failed: %w", err)
+			}
+			now := time.Now().UTC()
+			claims := jwt.MapClaims{
+				"ig":  ig.Name,
```
zetaab, line 1444: gives access to specific instancegroup, not all

upup/pkg/fi/cloudup/openstack/verifier.go
```diff
@@ -84,7 +84,14 @@ func (o openstackVerifier) VerifyToken(ctx context.Context, token string, body [
...
 	}
-	serverID := strings.TrimPrefix(token, OpenstackAuthenticationTokenPrefix)
+	token = strings.TrimPrefix(token, OpenstackAuthenticationTokenPrefix)
+	tmp := strings.Split(token, ":")
+	if len(tmp) != 2 {
+		return nil, fmt.Errorf("invalid token")
+	}
+	serverID := tmp[0]
+	jwtToken := tmp[1]
+	klog.Infof("'%s' '%s'", serverID, jwtToken)
```
zetaab, line 94: output example

```
'e7ef16b6-ab41-4444-b749-ccfd24233e76' 'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2NzM3OTc2NTcsImlhdCI6MTY3Mzc5NDA1NywiaWciOiJub2Rlcy1lc3BzdGkiLCJuYmYiOjE2NzM3OTQwNTd9.jQSUBs5jFrqkR7L-kqaRxiG5QIHLWUEga1v_DxnhZhesc9h7OGrU3Wd8Vlmko15bFpmofunDUoUbUSLalAQ5UPTofawEuwDsplUclcXpnshF61jaHIromIJ20yPQvh1MBU-0IJD0tUadX_N1M6AiACUhDrhV9i_QsK0_jmpweVCKiK_DrbiWm0_QqJ1Rv_tmmbBJyPOSfiqSJ3FpVaPVN17ALDh-kAmbOYX4BboZwujTFenckfZSJykcCI54YSRA9N1vTMDcQ8zm7v1Qzd4lDR0nVnRBmAKlxxQwzoM053AFeskZs6TvWpW0GRbK0re6HL_svHbawRHm9wkn4JiPmQ'
```

zetaab:
current status: this PR works already, but the solution is not pretty.

pkg/model/bootstrapscript.go
```diff
@@ -247,7 +247,7 @@ func (b *BootstrapScript) buildEnvironmentVariables(cluster *kops.Cluster) (map[
 // ResourceNodeUp generates and returns a nodeup (bootstrap) script from a
 // template file, substituting in specific env vars & cluster spec configuration
 func (b *BootstrapScriptBuilder) ResourceNodeUp(c *fi.CloudupModelBuilderContext, ig *kops.InstanceGroup) (fi.Resource, error) {
-	keypairs := []string{"kubernetes-ca", "etcd-clients-ca"}
+	keypairs := []string{"kubernetes-ca", "etcd-clients-ca", "kops-controller-bootstrap"}
```
johngmyers, line 250: I would call it just `bootstrap`, but whatever.

Not all nodes need this. It should only be added to keypairs if cloudprovider is openstack and instanceGroup is control plane.

pkg/model/pki.go
```diff
@@ -45,6 +45,15 @@ func (b *PKIModelBuilder) Build(c *fi.CloudupModelBuilderContext) error {
...
 	c.AddTask(defaultCA)
 
+	kopsControllerBootstrapJWT := &fitasks.Keypair{
+		Name:      fi.PtrTo("kops-controller-bootstrap"),
+		Lifecycle: b.Lifecycle,
+		Subject:   "cn=kops-controller",
+		Type:      "client",
+		Signer:    defaultCA,
+	}
+	c.AddTask(kopsControllerBootstrapJWT)
```
johngmyers, line 55: Only add if cloud provider is Openstack

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1422,6 +1424,41 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+					return nil, nil, fmt.Errorf("sign token failed: %w", err)
+				}
+				bootConfig.BootstrapToken = token
+			}
+		} else {
+			kopsControllerPublicKeys, err := keyset.ToPublicKeys()
+			if err != nil {
+				return nil, nil, fmt.Errorf("encoding kops-controller-bootstrap keys: %w", err)
+			}
+			bootConfig.BootstrapPublicKeys = kopsControllerPublicKeys
```
johngmyers, line 1458: This code is not necessary.

pkg/apis/nodeup/config.go
```diff
@@ -126,6 +126,10 @@ type BootConfig struct {
 	InstanceGroupRole kops.InstanceGroupRole
 	// NodeupConfigHash holds a secure hash of the nodeup.Config.
 	NodeupConfigHash string
+	// BootstrapToken contains token for downloading bootstrap setup from kops-controller
+	BootstrapToken string `json:",omitempty"`
+	// BootstrapPublicKeys are the bootstrap public keys to trust.
+	BootstrapPublicKeys string `json:",omitempty"`
```
johngmyers, line 132: This field is not necessary. Nodeup should get the keys out of the `CAs` map.

johngmyers (changes requested):
This is the right direction. These are the initial comments; I'll give it a more thorough review later.

pkg/bootstrap/authenticate.go
```diff
@@ -22,7 +22,7 @@ import (
 
 // Authenticator generates authentication credentials for requests.
 type Authenticator interface {
-	CreateToken(body []byte) (string, error)
+	CreateToken(body []byte, jwtToken string) (string, error)
```
johngmyers, line 25: nit: the "T" in "JWT" stands for "Token", so "jwtToken" is redundant.

```suggestion
	CreateToken(body []byte, jwt string) (string, error)
```

pkg/model/pki.go
```diff
@@ -45,6 +45,15 @@ func (b *PKIModelBuilder) Build(c *fi.CloudupModelBuilderContext) error {
 	}
 	c.AddTask(defaultCA)
 
+	kopsControllerBootstrapJWT := &fitasks.Keypair{
+		Name:      fi.PtrTo("kops-controller-bootstrap"),
+		Lifecycle: b.Lifecycle,
+		Subject:   "cn=kops-controller",
```
johngmyers, line 51:
```suggestion
		Subject:   "cn=bootstrap",
```

pkg/model/pki.go
```diff
@@ -45,6 +45,15 @@ func (b *PKIModelBuilder) Build(c *fi.CloudupModelBuilderContext) error {
...
+		Type:      "client",
+		Signer:    defaultCA,
```
johngmyers, line 53:
```suggestion
		Type:      "ca",
```

This would probably be why it didn't work.

pkg/model/bootstrapscript.go
johngmyers: Correction: only add this to keypairs if Openstack.

And the name should end in `-ca` in order for the keypair to be rotatable. So `bootstrap-ca`. Otherwise we need to modify `rotatableKeysetFilter()`.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1422,6 +1424,41 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 		}
 	}
 
+	if cluster.Spec.GetCloudProvider() == kops.CloudProviderOpenstack {
+		keyset := keysets["kops-controller-bootstrap"]
+		if !ig.HasAPIServer() {
+			if keyset != nil && keyset.Primary != nil && keyset.Primary.PrivateKey != nil {
```
johngmyers, line 1430: If the keyset isn't there or doesn't have a primary private key, it should error out.

upup/pkg/fi/cloudup/apply_cluster.go
johngmyers: Actually, here it needs to call `loadCertificates()` with `includeKeypairID` of `false`.

johngmyers (changes requested):

hakman:
This looks very cool. Looking forward to try it with Hetzner.
/lgtm

nodeup/pkg/model/kops_controller.go
```diff
@@ -87,6 +88,19 @@ func (b *KopsControllerBuilder) Build(c *fi.NodeupModelBuilderContext) error {
 	if model.UseCiliumEtcd(b.Cluster) {
 		caList = append(caList, "etcd-clients-ca-cilium")
 	}
+	if model.UseJWTForBootstrap(b.Cluster) {
```
johngmyers, line 91: Please don't add new references to the deprecated `b.Cluster`. This should be using something that tests `b.BootConfig.CloudProvider`.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1422,6 +1424,43 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
 	}
 
+	if apiModel.UseJWTForBootstrap(cluster) {
+		if !ig.HasAPIServer() {
+			keyset := keysets["bootstrap-ca"]
+			if keyset == nil {
+				return nil, nil, fmt.Errorf("key bootstrap-ca not found")
+			}
+
+			if keyset.Primary != nil && keyset.Primary.PrivateKey != nil {
```
johngmyers, line 1434: If either of these are nil, it should error out, not silently omit the token.

upup/pkg/fi/cloudup/openstack/verifier.go
```diff
@@ -111,7 +147,13 @@ func (o openstackVerifier) VerifyToken(ctx context.Context, token string, body [
...
 	value, ok := instance.Metadata[TagKopsInstanceGroup]
 	if ok {
-		result.InstanceGroupName = value
+		igValue, found := claims["ig"]
+		if found {
+			if igValue != value {
+				return nil, fmt.Errorf("claim mismatch %s != %s", igValue, value)
+			}
+			result.InstanceGroupName = value
+		}
```
johngmyers, line 156: In the case of `!found` shouldn't it return error?

upup/pkg/fi/cloudup/openstack/verifier.go
```diff
@@ -80,11 +81,46 @@ func NewOpenstackVerifier(opt *OpenStackVerifierOptions) (bootstrap.Verifier, er
...
 	if !strings.HasPrefix(token, OpenstackAuthenticationTokenPrefix) {
 		return nil, fmt.Errorf("incorrect authorization type")
 	}
-	serverID := strings.TrimPrefix(token, OpenstackAuthenticationTokenPrefix)
+	token = strings.TrimPrefix(token, OpenstackAuthenticationTokenPrefix)
+	tmp := strings.Split(token, ":")
+	if len(tmp) != 2 {
+		return nil, fmt.Errorf("invalid token")
+	}
+	serverID := tmp[0]
```
johngmyers, line 119: Would there be value in putting the serverID into the claims? Or is it not available at the time we have to construct the JWT?

johngmyers (changes requested):
This is really close.

upup/pkg/fi/cloudup/apply_cluster.go
zetaab: @johngmyers if I return error and execute only `kops create cluster foobar.k8s.local` its returning error in that case. It is not executing the command successfully because its returning error from row 1435 (because private key does not exists it cannot convert it as a bytes)

upup/pkg/fi/cloudup/openstack/verifier.go
zetaab: its not available when constructing the JWT, the only thing that is available is instancegroup name.

upup/pkg/fi/cloudup/apply_cluster.go
zetaab: I do understand that it should return error if I have `--yes` flag, but how to avoid returning error without yes?

johngmyers: I think you might need to check if the target is `DryRunTarget`, returning error only if it isn't. I'll do some research tomorrow.

upup/pkg/fi/cloudup/openstack/verifier.go
johngmyers: Okay, thanks for the info. I'm doubting the additional check is worth the effort.

upup/pkg/fi/cloudup/apply_cluster.go
zetaab: looks like that is not passed to this function either

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1422,6 +1424,43 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
...
+				key, err := jwt.ParseRSAPrivateKeyFromPEM(b)
+				if err != nil {
+					return nil, nil, fmt.Errorf("parsing pem key failed: %w", err)
+				}
+				now := time.Now().UTC()
+				claims := jwt.MapClaims{
+					"ig":  ig.Name,
+					"iat": now.Unix(),
+					"nbf": now.Unix(),
+					"exp": now.Add(time.Hour).Unix(),
```
zetaab, line 1449: the problem with this is that if someone/some reason will restart the instance. It cannot authenticate back:

```
I0118 07:52:14.295244       1 server.go:145] bootstrap 10.2.97.28:36072 verify err: validate: Token is expired
```

Why we are fetching the certificates again from kops-controller if node will restart? It do have certificates already in disk?

upup/pkg/fi/cloudup/openstack/verifier.go
zetaab: Well actually all other cloudproviders than openstack do use servergroups which makes it impossible to generate user_data per server. This means that we do not get big benefit of doing that for OpenStack only, because I am pretty sure that for instance hetzner (+scaleway?) is going to use this in instancegroup level. Also we have chicken - egg problem here. We cannot have serverID before the server is provisioned, we cannot either provision server before we have user_data. If we use server name instead, that is something people can modify (at least in openstack). So basically server name is not way to authorise.

PR #15016, opened 2023-01-16
Move more networking settings into nodeup.Config

johngmyers:


pkg/apis/nodeup/config.go
```diff
@@ -63,6 +64,10 @@ type Config struct {
 	KubeProxy *kops.KubeProxyConfig
 	// Networking configures networking.
 	Networking kops.NetworkingSpec
+	// UseCiliumEtcd is true when a Cilium etcd cluster is present.
+	UseCiliumEtcd bool `json:",omitempty"`
+	// UsesKubenet specifies that the CNI is derived from Kubenet.
+	UsesKubenet bool `json:",omitempty"`
```
hakman, line 70: Should we be consistent with `Use` vs `Uses`?

johngmyers: I was being consistent with the existing `components.UsesKubenet()`.

hakman (approved):

PR #15020, opened 2023-01-18
Upgrade Node Termination Handler to 1.18.3

johngmyers:


upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
```diff
@@ -107,15 +107,14 @@ spec:
     matchLabels:
       app.kubernetes.io/name: aws-node-termination-handler
       app.kubernetes.io/instance: aws-node-termination-handler
-      kubernetes.io/os: linux
+      app.kubernetes.io/component: deployment
```
hakman, line 110:
```
error from apply on apps/v1, Kind=Deployment kube-system/aws-node-termination-handler:
error from apply: error patching object: Deployment.apps "aws-node-termination-handler" is invalid: spec.selector: Invalid value:
v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"deployment", "app.kubernetes.io/instance":"aws-node-termination-handler", "app.kubernetes.io/name":"aws-node-termination-handler"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}:
field is immutable
```

upup/models/cloudup/resources/addons/node-termination-handler.aws/k8s-1.11.yaml.template
```diff
@@ -299,7 +299,6 @@ spec:
         app.kubernetes.io/component: daemonset
         kubernetes.io/os: linux
         k8s-app: aws-node-termination-handler
-        kubernetes.io/os: linux
```
hakman, line 302: This is used in `matchLabels`. I don't think we can remove it.

johngmyers: It's duplicate with two lines above.

hakman (approved):

PR #15024, opened 2023-01-19
make openstack kops-controller boostrap auth better

zetaab:


nodeup/pkg/bootstrap/install.go
```diff
@@ -161,6 +161,7 @@ func (i *Installation) buildSystemdJob() *nodetasks.InstallService {
 	manifest := &systemd.Manifest{}
 	manifest.Set("Unit", "Description", "Run kOps bootstrap (nodeup)")
 	manifest.Set("Unit", "Documentation", "https://github.com/kubernetes/kops")
+	manifest.Set("Unit", "ConditionPathExists", "!/srv/kubernetes/kubelet-server.crt")
```
zetaab, line 164: if I understood correctly we cannot use this in AWS? Is there easy way how we could pass cloud information to this package @olemarkus

zetaab:
@johngmyers @hakman could you check this? Currently in master (and in 1.26) we do not have bullet proof version

upup/pkg/fi/cloudup/openstack/verifier.go
```diff
@@ -75,12 +82,42 @@ func NewOpenstackVerifier(opt *OpenStackVerifierOptions) (bootstrap.Verifier, er
...
+	}
+	return kubernetes.NewForConfig(config)
+}
+
+// readKubeConfig ...
+func readKubeConfig() (*restclient.Config, error) {
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		clientcmd.NewDefaultClientConfigLoadingRules(),
+		&clientcmd.ConfigOverrides{}).ClientConfig()
+}
```
hakman, line 118: This should be inlined in `newClientSet()`.

hakman (changes requested):
/lgtm

PR #15032, opened 2023-01-21
Clarify comment on nonMasqueradeCIDR

justinsb:
Attempt to produce a more descriptive definition of nonMasqueradeCIDR.


justinsb:
I thought I would have a go at defining better what it is that this CIDR does.

One thing I had forgotten / wasn't clear to me is that this matches again the _destination_ IP.  So it's really saying "what destination IPs can accept a Pod IP and route back to us".  At least I think it is :-)

justinsb:
Also, I'm not 100% sure on the definition here - I _think_ it's right - and I'm certainly open to other suggestions on the phrasing :-)

johngmyers:
My take:

NonMasqueradeCIDR is the network range that can route directly to pod IPs. Traffic from pod IPs to within this range will not be masqueraded using NAT with the pod's node's address.

johngmyers:
Additionally: "A value of `::/0` specifies an IPv6-only pod network."

justinsb:
> NonMasqueradeCIDR is the network range that can route directly to pod IPs. Traffic from pod IPs to within this range will not be masqueraded using NAT with the pod's node's address.

I like that ... one challenge is that (at least on GCE) it's actually several CIDRs.  So I guess the question is whether we want this field to map only to the masquerade rules that nodeup sets up.   And further (assuming we shouldn't set those up with ip-masq-agent) how best to signal that.

Another thing I'm trying to figure out is what value we should be passing e.g. to AWS CCM `--cluster-cidr` - I'm wondering what it uses it for!

johngmyers:
This is specific to masquerading in the node. If the cloud provider is doing NAT from NetworkCIDR to the internet, that's out of scope of this setting.

johngmyers:
NonMasqueradeCIDR is the network range that can, assuming it can reach the nodes' network, route directly to pod IPs. Traffic from pod IPs to within this range will not be masqueraded using NAT with the pod's node's address. Any NAT external to the node is not considered for this definition. A value of `::/0` specifies an IPv6-only pod network.

PR #15033, opened 2023-01-21
nodeup: don't set up masquerade if nonMasqueradeCIDR not set

justinsb:
If the non-masquerade CIDR is not set, take that as an indication that
we don't want masquerade, rather than failing nodeup.

Not setting a non-masquerade CIDR means that we likely won't preserve
pod IPs for pod-to-pod traffic, but likely just means that more
NATting is done than might be needed.

Omitting the value can also be useful if we're using something like
the ip-masq-agent to manage masquerade rules for us.

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -390,6 +390,12 @@ runtime-endpoint: unix:///run/containerd/containerd.sock
 func (b *ContainerdBuilder) buildIPMasqueradeRules(c *fi.NodeupModelBuilderContext) error {
 	// TODO: Should we just rely on running nodeup on every boot, instead of setting up a systemd unit?
 
+	if b.NodeupConfig.Networking.NonMasqueradeCIDR == "" {
+		// We could fall back to the pod CIDR, that is likely a good universal
+		klog.Infof("not setting up masquerade, as NonMasqueradeCIDR is not set")
+		return nil
+	}
+
 	// This is based on rules from gce/cos/configure-helper.sh and the old logic in kubenet_linux.go
 
 	// We stick closer to the logic in kubenet_linux, both for compatibility, and because the GCE logic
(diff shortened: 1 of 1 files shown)


justinsb:
https://github.com/kubernetes-sigs/ip-masq-agent#configuring-the-agent has the suggestion not to have the kubelet also manage non-masquerade-cidr, which I take to also mean nodeup shouldn't.

I do want to see if we can pass tests without any rules here though, or whether it only works with the agent also.

justinsb:
Proposed https://github.com/kubernetes/test-infra/pull/28478 to allow us to e2e tests these on-demand.

johngmyers:
We currently use a `NonMasqueradeCIDR` of `::/0` to indicate an IPv6-only cluster. So either we need a separate field to indicate the pod network's address family or we should have this use case inicated by a `NonMasqueradeCIDR` of `0.0.0.0/0`.

johngmyers:
I would prefer we not have the API default to IPv4 when it did not previously.
/hold

johngmyers:
Hrm, it currently defaults to `100.64.0.0/10` for cloud providers other than GCE. GCE effectively defaults it to some other value, probably `0.0.0.0/0`. So it's arguable that the field already defaults to IPv4.

johngmyers:
I still think this field should be populated in the completed cluster spec. That will reduce the probability of bugs.

justinsb:
What I'm trying to figure out is what this field actually really _is_ now :-)  In various places it seems to be:

- The CIDR that contains pods + services
- The CIDR that pod IPs are drawn from
- The CIDR that is used to determine when we use source masquerade (here)

That's also why I opened https://github.com/kubernetes/kops/pull/15032 so we could try to tease this out.  I've triggered the new ipalias test (thanks for lgtm on that hakman!) so hopefully we can start to understand consequences here.

One thing we might need is some signal to nodeup to not install these rules, if indeed ip-masq-agent is the preferred option.  But AFAICT just using a fixed list of the private CIDRs is just as good and simpler (e.g. 172.16, 10, 192.168, 100.64 etc)

PR #15035, opened 2023-01-21
Set the nonMasqueradeCIDR for GCE networking

justinsb:
We do need a non-masquerade CIDR, and we can use the range we draw the pod CIDRs from (10.0.0.0/8).

### DIFF
--- a/upup/pkg/fi/cloudup/defaults.go
+++ b/upup/pkg/fi/cloudup/defaults.go
@@ -107,11 +107,7 @@ func PerformAssignments(c *kops.Cluster, cloud fi.Cloud) error {
 	}
 
 	if c.Spec.Networking.NonMasqueradeCIDR == "" {
-		if c.Spec.Networking.GCE != nil {
-			// Don't set NonMasqueradeCIDR
-		} else {
-			c.Spec.Networking.NonMasqueradeCIDR = "100.64.0.0/10"
-		}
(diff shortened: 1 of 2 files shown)


justinsb:
Trying both approaches in parallel here :-)

johngmyers:
You might want `0.0.0.0/0`

justinsb:
So in https://github.com/kubernetes/kops/pull/15186 I combined this and #15040 and the fix in #15186 itself also; all three are needed I believe.

* #15186: don't run the CCM routes controller here - there's no need for any route control; these are "real" IPs
* #15040: Allow access to apiserver from pods IPs, the network tag method which we use for the nodes does not work on these IPs (so maybe they're not quite so "real" after all!)
* #15035 (this PR): For nodeports, I _think_ we are blocking the return traffic because it isn't an established connection and the response is coming from a different IP.  I think it may be possible to do better here, for example by recognizing the port range, but I think starting with this non masquerade CIDR is more consistent with what we do elsewhere.  I think what changes here is that traffic to destinations outside the cluster is NAT-ted to appear to come from the node ip, which then means that it matches the established TCP connection.

justinsb:
/hold cancel

@johngmyers are you OK with this PR as-is?  I'd like to unblock the testing with cloud-provider-gcp

johngmyers (approved):

PR #15036, opened 2023-01-22
Improve support for AdditionalNetworkCIDRs

johngmyers:
Ensure that everything on AWS that tests NetworkCIDR also considers AdditionalNetworkCIDRs appropriately.

### DIFF
--- a/pkg/apis/kops/validation/aws.go
+++ b/pkg/apis/kops/validation/aws.go
@@ -58,18 +58,6 @@ func awsValidateCluster(c *kops.Cluster) field.ErrorList {
 		allErrs = append(allErrs, awsValidateIAMAuthenticator(field.NewPath("spec", "authentication", "aws"), c.Spec.Authentication.AWS)...)
 	}
 
-	for i, subnet := range c.Spec.Networking.Subnets {
-		f := field.NewPath("spec", "networking", "subnets").Index(i)
-		if subnet.AdditionalRoutes != nil {
-			if len(subnet.ID) > 0 {
-				allErrs = append(allErrs, field.Invalid(f, subnet, "additional routes cannot be added if the subnet is shared"))
-			} else if subnet.Type != kops.SubnetTypePrivate {
-				allErrs = append(allErrs, field.Invalid(f, subnet, "additional routes can only be added on private subnets"))
-			}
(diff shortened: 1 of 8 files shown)


johngmyers:
There's still one more issue: If one uses an API loadbalancer, one cannot put a control plane node in an AdditionalNetworkCIDR. That is because the load balancer's security group only opens the ports to the primary NetworkCIDR.

johngmyers:
I think it might be the other way around: only nodes in the primary network CIDR can talk to apiserver or kops-controller.

pkg/apis/kops/validation/aws_test.go
```diff
@@ -811,37 +839,45 @@ func TestAWSAdditionalRoutes(t *testing.T) {
...
-		cluster := kops.Cluster{
-			Spec: kops.ClusterSpec{
-				Networking: kops.NetworkingSpec{
-					NetworkCIDR: test.clusterCidr,
-					Subnets: []kops.ClusterSubnetSpec{
-						{
-							ID:               test.providerId,
-							Type:             test.subnetType,
-							AdditionalRoutes: test.route,
+		t.Run(test.name, func(t *testing.T) {
```
justinsb, line 859: Nice :+1:

pkg/apis/kops/validation/aws.go
```diff
@@ -416,39 +404,39 @@ func hasAWSEBSCSIDriver(c kops.ClusterSpec) bool {
...
+			allErrs = append(allErrs, field.Invalid(f.Child("target"), r, "unknown target type for route"))
+		}
 
-			ipRoute, _, e := net.ParseCIDR(r.CIDR)
-			if e != nil {
-				allErrs = append(allErrs, field.Invalid(f.Child("cidr"), r, "invalid cidr"))
-			} else if clusterNet.Contains(ipRoute) && strings.HasPrefix(r.Target, "pcx-") {
-				allErrs = append(allErrs, field.Forbidden(f.Child("target"), "target is more specific than cluster CIDR block. This route can target only an interface or an instance."))
+		var routeCIDRs []*net.IPNet
+		allErrs = append(allErrs, validateCIDR(r.CIDR, f.Child("cidr"), &routeCIDRs)...)
```
justinsb, line 424: Pre-existing, but I'm not the biggest fan of the validateCIDR method signature, it's pretty hard to grok what's happening there.... might send a follow on PR to clean that up.

justinsb (approved):

justinsb:
Thanks @johngmyers 

/approve
/lgtm

PR #15037, opened 2023-01-22
Don't set up masquerade if NonMasqueradeCIDR is /0

johngmyers:
Extend #15033 to cover cases where masquerade is excluded for the entire IP family's range.


johngmyers:
I think the step after this is to get GCP to default `NonMasqueradeCIDR` to `0.0.0.0/0`, probably in populate_cluster_spec. But we can discuss that at the time.

justinsb:
Thanks @johngmyers - this looks good; if it's just for GCE I'm going to try to page this all back in to my memory and get this (and whatever else we need) merged together this weekend

hakman (approved):

PR #15039, opened 2023-01-22
tests: Capture iptables chains

justinsb:
This is helpful debugging information when we see network issues.


hakman (approved):

justinsb:
/override pull-kops-e2e-k8s-gce-ipalias

We'll get that test to pass separately :-)

PR #15040, opened 2023-01-22
gce: When using network native pod IPs, open firewall to apiserver

justinsb:
If we're not masquerading the pod IPs, we need an explicit firewall
rule for the pods to reach the kube-apiserver.  Normally this is
permitted anyway, but if the apiserver has a locked-down CIDR range
(as the e2e tests do) then we need our own rule.

### DIFF
--- a/pkg/model/gcemodel/api_loadbalancer.go
+++ b/pkg/model/gcemodel/api_loadbalancer.go
@@ -99,6 +99,18 @@ func createPublicLB(b *APILoadBalancerBuilder, c *fi.CloudupModelBuilderContext)
 			TargetTags:   []string{b.GCETagForRole(kops.InstanceGroupRoleControlPlane)},
 			Allowed:      []string{"tcp:" + strconv.Itoa(wellknownports.KubeAPIServer)},
 		})
+
+		if b.NetworkingIsIPAlias() {
+			c.AddTask(&gcetasks.FirewallRule{
+				Name:         s(b.NameForFirewallRule("pod-cidrs-to-https-api")),
+				Lifecycle:    b.Lifecycle,
+				Network:      network,
+				SourceRanges: []string{b.Cluster.Spec.Networking.PodCIDR},
+				TargetTags:   []string{b.GCETagForRole(kops.InstanceGroupRoleControlPlane)},
+				Allowed:      []string{"tcp:" + strconv.Itoa(wellknownports.KubeAPIServer)},
+			})
+		}
+
 		if b.Cluster.UsesNoneDNS() {
 			b.AddFirewallRulesTasks(c, "kops-controller", &gcetasks.FirewallRule{
 				Lifecycle:    b.Lifecycle,
--- a/pkg/model/gcemodel/external_access.go
+++ b/pkg/model/gcemodel/external_access.go
@@ -17,8 +17,11 @@ limitations under the License.
 package gcemodel
 
 import (
+	"strconv"
+
 	"k8s.io/klog/v2"
 	"k8s.io/kops/pkg/apis/kops"
+	"k8s.io/kops/pkg/wellknownports"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/gcetasks"
 )
@@ -112,6 +115,17 @@ func (b *ExternalAccessModelBuilder) Build(c *fi.CloudupModelBuilderContext) err
 			SourceRanges: b.Cluster.Spec.API.Access,
 			Network:      network,
 		})
+
+		if b.NetworkingIsIPAlias() {
+			c.AddTask(&gcetasks.FirewallRule{
+				Name:         s(b.NameForFirewallRule("pod-cidrs-to-https-api")),
+				Lifecycle:    b.Lifecycle,
+				Network:      network,
+				SourceRanges: []string{b.Cluster.Spec.Networking.PodCIDR},
+				TargetTags:   []string{b.GCETagForRole(kops.InstanceGroupRoleControlPlane)},
+				Allowed:      []string{"tcp:" + strconv.Itoa(wellknownports.KubeAPIServer)},
+			})
+		}
 	}
 
 	return nil


justinsb:
/retest

I want to be see if this is a flake or not

justinsb:
It looks like this one might be real, the others might be flakes:

```
Kubernetes e2e suite: [It] [sig-network] Services should be able to update service type to NodePort listening on same port number but different protocols expand_less
```

That sort of makes sense, the networking for NodePort is tricky.

johngmyers:
What if the cluster is using an internal loadbalancer in front of the Kubernetes API?

justinsb:
> What if the cluster is using an internal loadbalancer in front of the Kubernetes API?

Tweaked the code to set this up alongside other places we use API.Access CIDRs.  It looks like we don't currently enforce CIDR restrictions on internal load balancers on GCE, that's probably fine but I'll think about whether we should add that separately.

My current theory is that we need this and some sort of masquerading for nodeport at least, so in practice #15035 if we don't want ip-masq-agent, but I'm trying to figure out exactly why.

pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -99,6 +99,18 @@ func createPublicLB(b *APILoadBalancerBuilder, c *fi.CloudupModelBuilderContext)
...
+				Lifecycle:    b.Lifecycle,
+				Network:      network,
+				SourceRanges: []string{b.Cluster.Spec.Networking.PodCIDR},
+				TargetTags:   []string{b.GCETagForRole(kops.InstanceGroupRoleControlPlane)},
+				Allowed:      []string{"tcp:" + strconv.Itoa(wellknownports.KubeAPIServer)},
+			})
+		}
+
 		if b.Cluster.UsesNoneDNS() {
 			b.AddFirewallRulesTasks(c, "kops-controller", &gcetasks.FirewallRule{
```
johngmyers, line 115: Unrelated to this PR, but why does all of `b.Cluster.Spec.API.Access` need to talk to kops-controller? Shouldn't this be restricted to `NetworkCIDR`?

pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -99,6 +99,18 @@ func createPublicLB(b *APILoadBalancerBuilder, c *fi.CloudupModelBuilderContext)
 			TargetTags:   []string{b.GCETagForRole(kops.InstanceGroupRoleControlPlane)},
 			Allowed:      []string{"tcp:" + strconv.Itoa(wellknownports.KubeAPIServer)},
 		})
+
+		if b.NetworkingIsIPAlias() {
+			c.AddTask(&gcetasks.FirewallRule{
```
johngmyers, line 104: Presumably this is to allow traffic from the LB to the control plane nodes.

Do we need a rule to allow the PodCIDRs to get traffic to the internal load balancer?

johngmyers (commented):
It looks like GCE requires nodes to be covered by a CIDR in `Cluster.Spec.API.Access`. AWS adds explicit rules allowing each worker instance group to talk to control plane nodes.

pkg/model/gcemodel/api_loadbalancer.go
justinsb: I believe that's right, I can send a PR, unless you want to :-)

It might be that on some platforms it's not possible to have a separate CIDR, but it is possible here :-)

pkg/model/gcemodel/api_loadbalancer.go
justinsb: We don't have a rule for the nodes, so I'm assuming not.  I think the internal load balancer is open to anything on the subnets.  I can try sending another PR to lock it down to the source tags (for node IPs) + the pod CIDR (source tags don't work for secondary IPs), but we don't have a test for it either at the moment, I believe, so likely another PR...

justinsb:
So this - alongside https://github.com/kubernetes/kops/pull/15186 - should be the two missing pieces to get networking=gce to pass e2e reliably.

This PR is still setting `--configure-cloud-routes=true` on the nodes, which I think was causing the nodes to get into unexpected states.  The fix for that is #15186.  We can see the errors in the [CCM logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/15040/pull-kops-e2e-k8s-gce-ipalias/1617220117144276992/artifacts/cluster-info/kube-system/cloud-controller-manager-dt2zb/logs.txt):

```
I0122 18:14:37.727626       1 event.go:294] "Event occurred" object="nodes-us-east1-b-pvxr" kind="Node" apiVersion="" type="Warning" reason="FailedToCreateRoute" message="Could not create route 61156e03-375e-4622-ba9a-fd3b54ae2c36 10.4.2.0/24 for node nodes-us-east1-b-pvxr after 5.29389529s: googleapi: Error 400: UNSUPPORTED_OPERATION - 10.4.2.0/24 hides the address space of the network (10.4.0.0/14). Cannot change the routing of packets destined for the network."
```

We need to open up the pod IP range to the apiserver because network tags do not apply to secondary IPs.  And of course it "works on my machine" because e2e locks down admin-access, but I don't typically lock it down when doing test/dev.

We do likely need test coverage (and possibly some improvements) for internal load balancers, but I don't _think_ anything there blocks this PR?  So I'd ask that we try to get this in first, so we can get the GCE CCM + kOps tests going, and then we can fix e.g. internal load balancer CIDR restrictions.

I think I'm going to cherry-pick this to the other PR also, just so we can see them pass together (hopefully!)

justinsb:
@johngmyers are you OK with this one also?  You have good observations above (internal LB & can we narrow the kops-controller CIDR are the two I noted) but is it OK if I do them in follow-on PRs?  (Trying to unblock cloud-provider-gcp testing)

pkg/model/gcemodel/api_loadbalancer.go
hakman: Actually, in most cases, this will have to be `0.0.0.0/0`. With most clouds, there is no way to track the internal IP through an external load balancer.

johngmyers (approved):

PR #15045, opened 2023-01-23
Handle disappearing LBs during tear-down

johngmyers:
Found in https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-k126-kolatest-to-klatest-kolatest/1615188933384081408/build-log.txt

Handle the case where a load balancer in the account is deleted between the time we list it and the time we query it for tags.

### DIFF
--- a/pkg/resources/aws/aws.go
+++ b/pkg/resources/aws/aws.go
@@ -1497,8 +1497,20 @@ func DescribeELBs(cloud fi.Cloud) ([]*elb.LoadBalancerDescription, map[string][]
 
 		tagResponse, err := c.ELB().DescribeTags(tagRequest)
 		if err != nil {
-			innerError = fmt.Errorf("error listing elb Tags: %v", err)
-			return false
+			tagResponse = &elb.DescribeTagsOutput{}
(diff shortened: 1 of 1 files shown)


justinsb:
I assume this happens if there is a concurrent deletion operation.  Can we add a log message so people know why it's switching behaviours (e.g. if they notice the slowdown)?

An alternative strategy could be simply to retry any list operation if the error is one of a set of candidate errors?

johngmyers:
This could be for any LB in the account, possibly one for a different cluster. I think any logging would be noise; it's just switching tactics for the 20 LBs in the page of results.

It is a good question whether we should try to address these piecemeal or whether we should have a retry loop over the whole "list resources and delete" process.

justinsb:
In office hours, I think we said we were going to look at retrying all list operations and see if that was in fact easier.

Not sure if you've had time @johngmyers - I don't think there's anything wrong with this PR as-is, so we can merge as-is, or I can just have a go at the list handling if that's helpful.

But no rush, I don't think this happens very often (currently) ... just going through the recent PRs!

johngmyers:
Will address this at a higher level

PR #15048, opened 2023-01-23
Automated cherry pick of #15024: make openstack kops-controller boostrap auth better

zetaab:
Cherry pick of #15024 on release-1.26.

#15024: make openstack kops-controller boostrap auth better

```release-note

```

### DIFF
--- a/cmd/kops-controller/pkg/server/server.go
+++ b/cmd/kops-controller/pkg/server/server.go
@@ -140,7 +140,7 @@ func (s *Server) bootstrap(w http.ResponseWriter, r *http.Request) {
 
 	ctx := r.Context()
 
-	id, err := s.verifier.VerifyToken(ctx, r.Header.Get("Authorization"), body, s.opt.Server.UseInstanceIDForNodeName)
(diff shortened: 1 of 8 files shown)


cmd/kops-controller/pkg/server/server.go
```diff
@@ -140,7 +140,7 @@ func (s *Server) bootstrap(w http.ResponseWriter, r *http.Request) {
 
 	ctx := r.Context()
 
-	id, err := s.verifier.VerifyToken(ctx, r.Header.Get("Authorization"), body, s.opt.Server.UseInstanceIDForNodeName)
+	id, err := s.verifier.VerifyToken(ctx, r, r.Header.Get("Authorization"), body, s.opt.Server.UseInstanceIDForNodeName)
```
olemarkus, line 143: Do we need to pass in both `r` the token here? Can't we fetch the header from `r` within the function?

zetaab: we can, but I tried to keep it simple for most of the providers. Instead parsing headers in each provider, I still like the way that authorization header is pre-parsed here. If they want to more advanced stuff the `r` is available (like openstack needs)

hakman (approved):
This is a cherry-pick. Let's go with it as is.

Issue #15065, opened 2023-01-27
Enable cluster autoscaler metrics scraping via service monitor


Schmitze333:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
The Kubernetes Cluster Autoscaler offers the possibility to scrape its metrics based on a service monitor. (see [here](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/templates/servicemonitor.yaml))
However, when setting up cluster autoscaler with kOps, one has to provide an additional scrape config in order to get the metrics.
It would be handy to utilize a service monitor also via kOps.

**2. Feel free to provide a design supporting your feature request.**
Extend the `clusterAutoscaler` sections with an optional `serviceMonitor` section:
```yaml
spec:
  clusterAutoscaler:
    enabled: true   
    serviceMonitor:
      enabled: true
      namespace: monitoring
      scrapeInterval: 60s
```
`namespace` and `scrapeInterval` are reasonable options that one might possibly specify.

PR #15066, opened 2023-01-27
Add optional servicemonitor for cluster autoscaler

Schmitze333:
This PR proposes options within the `clusterAutoscaler` section of a cluster spec that will add a service monitor to the cluster autoscaler, which would allow scraping metrics from the autoscaler without an additional scrape config.

Relates to: #15065


pkg/apis/kops/componentconfig.go
```diff
@@ -1035,6 +1035,21 @@ type ClusterAutoscalerConfig struct {
 	// CustomPriorityExpanderConfig overides the priority-expander ConfigMap with the provided configuration. Any InstanceGroup configuration will be ignored if this is set.
 	// This could be useful in order to use regex on priorities configuration
 	CustomPriorityExpanderConfig map[string][]string `json:"customPriorityExpanderConfig,omitempty"`
+	// ClusterAutoscalerServiceMonitorConfig specifies the configuration for an optional service monitor
```
justinsb, line 1038: Nit: the comment should be `serviceMonitor specifies...` because that's what makes the most sense in the CRD (and thus in any docs etc).

(There's a debate about whether it should be `ServiceMonitor` or `serviceMonitor`, one is more correct for the go comment and one is more correct for the CRD comment, I guess I'd favor users over developers, so I prefer `serviceMonitor` but I suspect I've made both choices personally in the past, probably depending on what linters I was using at the time!)

pkg/apis/kops/componentconfig.go
```diff
@@ -1035,6 +1035,21 @@ type ClusterAutoscalerConfig struct {
...
+type ClusterAutoscalerServiceMonitorConfig struct {
+	// Enables the service monitor for the cluster autoscaler
+	// Default: false
+	Enabled *bool `json:"enabled,omitempty"`
+	// Namespace of the service monitor
+	// Default: kube-system
+	Namespace *string `json:"namespace,omitempty"`
+	// Scrape interval for the cluster autoscaler
+	// Default: 30s
+	ScrapeInterval *string `json:"scrapeInterval,omitempty"`
```
justinsb, line 1052: I thought you could use a time.Duration here, but I don't see any other examples so ... I guess not!

pkg/model/components/clusterautoscaler_test.go
```diff
@@ -0,0 +1,106 @@
...
+			clusterSpec := buildSpec(test.inSM)
+
+			err := ob.BuildOptions(clusterSpec)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			got := *clusterSpec.ClusterAutoscaler.ServiceMonitor
+
+			if !reflect.DeepEqual(got, test.want) {
```
justinsb, line 99: Tip: I personally like to use cmp.Diff (https://pkg.go.dev/github.com/google/go-cmp/cmp) even though I have to look up the recommended syntax every time:

```
	if diff := cmp.Diff(want, got); diff != "" {
		t.Errorf("MakeGatewayInfo() mismatch (-want +got):\n%s", diff)
	}
```

Not a blocker though!

upup/models/cloudup/resources/addons/cluster-autoscaler.addons.k8s.io/k8s-1.15.yaml.template
```diff
@@ -257,8 +257,34 @@ spec:
   selector:
     app.kubernetes.io/name: "cluster-autoscaler"
   type: "ClusterIP"
+
+{{- if WithDefaultBool .ServiceMonitor.Enabled false }}
 ---
+# Source: cluster-autoscaler/templates/service-monitor.yaml
+apiVersion: monitoring.coreos.com/v1
```
justinsb, line 264: Oh - `monitoring.coreos.com` a surprising api group to see.  Obviously coreos/redhat/ibm were/are very much part of the kubernetes project, but this does mean that users will have to install the prometheus operator I guess (if they enable this option), and that operator isn't under CNCF I believe - though it is Apache 2.  Honestly I'm just not sure how we should handle that - I don't think we have a simple "yes/no" policy ....

justinsb:
Look good from a code perspective - thanks @Schmitze333 

The thing that gives me pause here is that we now have a dependency on a CRD that is defined in a different project (the prometheus-operator).  I would sort of have expected/hoped the CRD would be defined in a "neutral" project, and that the prometheus-operator would support it.  (More like the Ingress model, with different providers).

From that perspective, I guess it also raises the question of whether we should be configuring this for each addon, or whether we want to have some cluster-wide setting like `spec.exposeServiceMonitor` or `spec.enableMonitoring: prometheus-operator` (those are not great API design probably, but I hope the difference makes sense!)

Schmitze333:
> The thing that gives me pause here is that we now have a dependency on a CRD that is defined in a different project (the prometheus-operator). I would sort of have expected/hoped the CRD would be defined in a "neutral" project, and that the prometheus-operator would support it. (More like the Ingress model, with different providers).

Hi @justinsb, that's true and it makes totally sense. Honestly, I didn't think about that since the PR was driven by a personal use case where the presence of the prometheus-operator is guaranteed.

> From that perspective, I guess it also raises the question of whether we should be configuring this for each addon, or whether we want to have some cluster-wide setting like `spec.exposeServiceMonitor` or `spec.enableMonitoring: prometheus-operator` (those are not great API design probably, but I hope the difference makes sense!)

Good question. I'm not sure if users need the option to choose whether they want to scrape single resources or not.

Both, the CDR topic and the API design from a global perspective are bigger topics. I'm afraid, as a first-time contributor I won't be much of a great help here.
I wonder if and how there's a way forward. Anyway, I'm happy to help.

pkg/model/components/clusterautoscaler_test.go
Schmitze333: Ah cool! Definitively a better test output.

olemarkus:
Thanks for the contribution, but unfortunately relying on CRDs kOps do not own is far from straightforward. The PromOperator CRDs are versioned too, may need upgrade strategies, and so on. 

I think we have to close this one.

PR #15067, opened 2023-01-28
GCE: Create per-machine certificate

justinsb:
For GCE (at first), we get a machine certificate as part of bootstrapping (or self-sign for control-plane machines)

Breaking up the gossip-via-kops-controller PR, also looking at authentication.


justinsb:
I think the issue was an AWS upstream issue, so trying again without the test hacks :-)

justinsb:
/test pull-kops-e2e-k8s-gce-cilium

The failure was one I hadn't seen before - Pod and PodStatus

justinsb:
Removing WIP - this was WIP while I got the tests to pass, and we had a little detour thanks to IPv6 on AWS  - but I think we're ready for review :-)

PR #15069, opened 2023-01-28
openstack verifier: support IPv6

justinsb:
Add IPv6 support to the openstack verifier and polish up a few error messages.

### DIFF
--- a/cmd/kops-controller/pkg/server/server.go
+++ b/cmd/kops-controller/pkg/server/server.go
@@ -144,7 +144,8 @@ func (s *Server) bootstrap(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		klog.Infof("bootstrap %s verify err: %v", r.RemoteAddr, err)
 		w.WriteHeader(http.StatusForbidden)
-		_, _ = w.Write([]byte(fmt.Sprintf("failed to verify token: %v", err)))
+		// don't return the error; this allows us to have richer errors without security implications
+		_, _ = w.Write([]byte("failed to verify token"))
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


justinsb:
cc @zetaab and @johngmyers 

I think the openstack verifier is doing the right things here, checking the IP address against the cloud.

This is an important endpoint to protect, but I think what we're doing is pretty reasonable here.  We are slightly vulnerable to IP spoofing inside the network, but I think we should have a warning that generally we recommend ensuring that you have protection against IP spoofing generally.   I don't think this warning is even specific to OpenStack - if I can spoof IP addresses, depending on my CNI I think I can intercept pod or service traffic also in kubernetes, so this is probably good advice regardless (given that we can't guarantee that all pod to pod traffic is encrypted and authenticated).

We could be _slightly_ better off if we connected back from the verifier to the machine, but I _think_ even that is spoofable if you can put ARP packets on the wire, so I think the right protection is to protect against IP address takeover.

zetaab (approved):
@justinsb you are correct that ip spoofing is possible problem. However, in OpenStack there is port securities in VPC level. It means that you cannot fake the ip address basically in OpenStack layer. However, if we have Kubernetes worker and pods running in top of that, all of those pods do have same ip address that the worker itself has against verifier. That is why I did protection that boostrap can be requested only once and if node is in K8s cluster it does not return anymore anything (so then pods are not problem).

Anyway this is not easy problem and its even more difficult issue in case of Hetzner (ping @hakman). The issue is that Hetzner cannot see the real source ip from where the request is coming from. Hetzner is calling verifier though loadbalancer and loadbalancer will lose the source ip

Of course it would be nice if all clouds could provider some mechanism to verify the instance identity, but that is not the reality. Do you have any better ideas how we could solve this issue?

PR #15077, opened 2023-01-30
scaleway: CCM and CSI driver

Mia-Cross:
This PR adds the manifests for the cloud-controller-manager and the container-storage-interface driver. This is the last step before we can have the cluster verified by `kops validate cluster`. After that we can start implementing e2e tests.

### DIFF
--- a/upup/models/cloudup/resources/addons/scaleway-cloud-controller.addons.k8s.io/k8s-1.24.yaml.template
+++ b/upup/models/cloudup/resources/addons/scaleway-cloud-controller.addons.k8s.io/k8s-1.24.yaml.template
@@ -1,0 +1,185 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: scaleway-secret
+  namespace: kube-system
+type: Opaque
+stringData:
+  # insert your Scaleway credentials here
+  SCW_ACCESS_KEY: {{ SCW_ACCESS_KEY }}
+  SCW_SECRET_KEY: {{ SCW_SECRET_KEY }}
+  # Project ID could also be an Organization ID
+  SCW_DEFAULT_PROJECT_ID: {{ SCW_DEFAULT_PROJECT_ID }}
+  # Region is where your loadbalancer will be created, ex: nl-ams, nl-ams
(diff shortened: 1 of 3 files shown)


upup/models/cloudup/resources/addons/scaleway-cloud-controller.addons.k8s.io/k8s-1.24.yaml.template
```diff
@@ -0,0 +1,185 @@
...
+stringData:
+  # insert your Scaleway credentials here
+  SCW_ACCESS_KEY: {{ SCW_ACCESS_KEY }}
+  SCW_SECRET_KEY: {{ SCW_SECRET_KEY }}
+  # Project ID could also be an Organization ID
+  SCW_DEFAULT_PROJECT_ID: {{ SCW_DEFAULT_PROJECT_ID }}
+  # Region is where your loadbalancer will be created, ex: nl-ams, nl-ams
+  SCW_DEFAULT_REGION: {{ SCW_DEFAULT_REGION }}
+  # Zone is where your servers and volumes will be created, ex: nl-ams-1, nl-ams-2
+  SCW_DEFAULT_ZONE: {{ SCW_DEFAULT_ZONE }}
```
hakman, line 17: I don't think this works. See:
https://github.com/kubernetes/kops/blob/f4780157b598210f91ed586d512c376908344e45/upup/pkg/fi/cloudup/template_functions.go#L179-L187

hakman (changes requested):

upup/models/cloudup/resources/addons/scaleway-cloud-controller.addons.k8s.io/k8s-1.24.yaml.template
hakman: My bad, seems that these were already added previously, even though unused.
https://github.com/kubernetes/kops/blob/f4780157b598210f91ed586d512c376908344e45/upup/pkg/fi/cloudup/template_functions.go#L194-L209

hakman (approved):

PR #15078, opened 2023-01-30
disable kops-configuration.service after successful execution

### LINKED ISSUE #15057: nodeup in crashloop after node restart in OpenStack
/kind bug

 this information.**

master

 a `kops` flag.**

1.26.1

openstack

I have kops openstack cluster installed using None dns.

After I will restart normal worker node, the node will come back online and start serving workloads from kubelet. However, nodeup (kops-configuration.service) will be in crashloop.

```
Jan 25 14:17:05 nodes-espsti-pn7tti nodeup[662]: W0125 14:17:05.188771     662 main.go:133] got error running nodeup (will retry in 30s): failed to get node config from server: kops-controller returned status code 403: failed to verify token: server is already joined to kubernetes cluster; kops-controller returned status code 403: failed to verify token: server is already joined to kubernetes cluster; kops-controller returned status code 403: failed to verify token: server is already joined to kubernetes cluster
```

The problem here is that OpenStack allows nodes fetch bootstrap configuration only once, because otherwise it could mean that someone else will fetch the certs also from the workloads. This leads to issue that in case of openstack we should not call bootstrap again if the needed configs and certificates do exist already in node.

nodeup should not end to that state, either it should succeed or we should disable it (if its not anymore needed)

some possibilities:
- after nodeup has done all of its tasks (when this is?) we could disable nodeup service?
- I tried to cache nodeconfig to disk and then check does certificates exist https://github.com/zetaab/kops/commit/09b327b023713bc7155441a3efedb4584c97117e This logic will work, but nodeup still requires that certificates needs to be loaded internally that it can generate the files again. So it is kind of missing Find - Render structure that we have in normal models..

zetaab:
fixes #15057

cc @johngmyers 

I do not know is this mistake that we enable the `kops-configuration.service` in cloud-init after the nodeup is successfully executed? If we do not do that (this PR), we can reboot machines easily in OpenStack and kops controller bootstrap will work fine (because its not executed anymore after initial execution)

### DIFF
--- a/nodeup/pkg/bootstrap/install.go
+++ b/nodeup/pkg/bootstrap/install.go
@@ -175,6 +175,7 @@ func (i *Installation) buildSystemdJob() *nodetasks.InstallService {
 	service := &nodetasks.InstallService{Service: nodetasks.Service{
 		Name:       serviceName,
 		Definition: fi.PtrTo(manifestString),
+		Enabled:    fi.PtrTo(false),
 	}}
 
 	service.InitDefaults()
--- a/nodeup/pkg/bootstrap/tests/simple/tasks.yaml
+++ b/nodeup/pkg/bootstrap/tests/simple/tasks.yaml
@@ -17,7 +17,7 @@ definition: |
 
   [Install]
   WantedBy=multi-user.target
-enabled: true
+enabled: false
 manageState: true
 running: true
 smartRestart: true


sjtoik (approved):
I would consider refactoring how the configuration is handled.
Having env variables coming from three sources seems excessive.

nodeup/pkg/bootstrap/install.go
```diff
@@ -178,6 +178,6 @@ func (i *Installation) buildSystemdJob() *nodetasks.InstallService {
 	}}
```
hakman, line 178: Probably better placed here:
```suggestion
		Enabled:    fi.PtrTo(false),
	}}
```

hakman (changes requested):

nodeup/pkg/bootstrap/install.go
```diff
@@ -175,9 +175,9 @@ func (i *Installation) buildSystemdJob() *nodetasks.InstallService {
 	service := &nodetasks.InstallService{Service: nodetasks.Service{
 		Name:       serviceName,
 		Definition: fi.PtrTo(manifestString),
+		Enabled:    fi.PtrTo(false),
 	}}
 
 	service.InitDefaults()
-
```
hakman, line 181: @zetaab Do you mind reverting this line also?

hakman (approved):

PR #15083, opened 2023-01-31
Allow setting 'ignore-volume-microversion' for OCCP

ederst:
This will allow setting the option `ignore-volume-microversion` for the cinder-csi-plugin.

Setting this is necessary for older OpenStack APIs so that OCCP can create PVs.

Note: This will work with cinder-csi-plugin >= 1.25.

For reference:
* https://github.com/kubernetes/cloud-provider-openstack/pull/1986/

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -438,6 +438,8 @@ spec:
                             type: boolean
                           ignore-volume-az:
                             type: boolean
+                          ignore-volume-microversion:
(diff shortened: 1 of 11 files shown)


zetaab (approved):
seems legit

hakman:
@zetaab Please cherry-pick to 1.26 if you think it's useful there also.

ederst:
@zetaab, @hakman  added some tests, and can do the cherry-pick afterwards too if wanted

zetaab (approved):

PR #15095, opened 2023-02-02
Pass actual cluster name to cinder-csi-plugin

ederst:
This passes the acutal cluster name to the cinder-csi-plugin, so that the plugin will add the name as metadata to the backing volume in OpenStack.

Effectively, the change will help to better identify which volume in OpenStack belongs to which cluster, which is especially helpful when running multiple clusters in one OpenStack tenant/project.

Setting the cluster name in both - the controller and the nodeplugin - pods will ensure that dynamic and ephemeral volumes will receive the correct metadata.

### DIFF
--- a/upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml.template
+++ b/upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml.template
@@ -333,7 +333,7 @@ spec:
             - name: CLOUD_CONFIG
               value: /etc/kubernetes/cloud.config
             - name: CLUSTER_NAME
-              value: kubernetes
+              value: "{{- if .CloudProvider.Openstack.BlockStorage.ClusterName -}} {{ .CloudProvider.Openstack.BlockStorage.ClusterName }} {{- else -}} kubernetes {{- end -}}"
           imagePullPolicy: "IfNotPresent"
           ports:
             - containerPort: 9808
@@ -465,11 +465,18 @@ spec:
             - /bin/cinder-csi-plugin
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--cloud-config=$(CLOUD_CONFIG)"
+{{- if .CloudProvider.Openstack.BlockStorage.ClusterName }}
+            - "--cluster=$(CLUSTER_NAME)"
+{{- end }}
           env:
             - name: CSI_ENDPOINT
               value: unix://csi/csi.sock
             - name: CLOUD_CONFIG
               value: /etc/kubernetes/cloud.config
+{{- if .CloudProvider.Openstack.BlockStorage.ClusterName }}
+            - name: CLUSTER_NAME
+              value: {{ .CloudProvider.Openstack.BlockStorage.ClusterName }}
+{{- end }}
           imagePullPolicy: "IfNotPresent"
           ports:
             - containerPort: 9808
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -428,6 +428,10 @@ spec:
                         properties:
                           bs-version:
                             type: string
+                          clusterName:
+                            description: ClusterName sets the --cluster flag for the
+                              cinder-csi-plugin to the provided name
+                            type: string
                           createStorageClass:
                             description: CreateStorageClass provisions a default class
                               for the Cinder plugin
--- a/pkg/apis/kops/componentconfig.go
+++ b/pkg/apis/kops/componentconfig.go
@@ -790,6 +790,8 @@ type OpenstackBlockStorageConfig struct {
 	CreateStorageClass *bool  `json:"createStorageClass,omitempty"`
 	CSIPluginImage     string `json:"csiPluginImage,omitempty"`
 	CSITopologySupport *bool  `json:"csiTopologySupport,omitempty"`
+	// ClusterName sets the --cluster flag for the cinder-csi-plugin to the provided name
+	ClusterName string `json:"clusterName,omitempty"`
 }
 
 // OpenstackMonitor defines the config for a health monitor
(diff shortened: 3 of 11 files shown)


upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml.template
```diff
@@ -333,7 +333,7 @@ spec:
             - name: CLOUD_CONFIG
               value: /etc/kubernetes/cloud.config
             - name: CLUSTER_NAME
-              value: kubernetes
+              value: {{ ClusterName }}
```
zetaab, line 336: hmm, what will happen to existing installations? It might work correctly to newer one, but what about old ones?

zetaab:
also I think we need to add this also to occm?

zetaab:
if I read code correctly, modifying existing CSI cinder cluster name should not matter(?). The cluster name is used only when creating snapshots or new volumes, but it will not affect anything else (old volumes should work?). Anyway this might be risky move, if later on the behaviour will change in csi cinder driver (the labels should be modified at least).

ederst:
> also I think we need to add this also to occm? Seems that loadbalancer names have the cluster name, so if you create two similar named LBs to two different clusters it is facing an issue. Anyway this change is not backwards compatible, so I would like to see some boolean flag for that feature. It should be turned on for new clusters, but not to old ones.

Don't know about that, I know there was an issue resolved in OCCM (https://github.com/kubernetes/cloud-provider-openstack/pull/1809), but AFAIK the `--cluster-name` is already configurable for the CCMs in kOps:

https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/componentconfig.go#L675-L676

So in case someone experiences this issue, they can do that already.

> if I read code correctly, modifying existing CSI cinder cluster name should not matter(?). The cluster name is used only when creating snapshots or new volumes, but it will not affect anything else (old volumes should work?). Anyway this might be risky move, if later on the behaviour will change in csi cinder driver (the labels should be modified at least).
> IMO the best what we can do here, is to add that new boolean value which will define the behaviour for both: occm and csi. After that it can be used for new clusters but not the old ones

Yes, right now the label seems to be just informative, that's what I read from code and doc as well.  And I tried it out on a few of our test clusters, does not seem to impact the old volumes, and new volumes just receive the cluster name instead of "kubernetes" in the property.

But you got a point with when the behavior changes later on. So this might be better off as a config flag. But maybe not a boolean but also a clusterName flag here:

https://github.com/kubernetes/kops/blob/ffdffb3d6bbcc8e097b42c0a3f9340d178381545/pkg/apis/kops/componentconfig.go#L784-L793

For example like this:

```golang
type OpenstackBlockStorageConfig struct {
	Version                  *string `json:"bs-version,omitempty"`
	IgnoreAZ                 *bool   `json:"ignore-volume-az,omitempty"`
	OverrideAZ               *string `json:"override-volume-az,omitempty"`
	IgnoreVolumeMicroVersion *bool   `json:"ignore-volume-microversion,omitempty"`
	// CreateStorageClass provisions a default class for the Cinder plugin
	CreateStorageClass *bool  `json:"createStorageClass,omitempty"`
	CSIPluginImage     string `json:"csiPluginImage,omitempty"`
	CSITopologySupport *bool  `json:"csiTopologySupport,omitempty"`
        ClusterName  string `json:"clusterName,omitempty"`  // default value: "kubernetes"
}
```

upup/models/cloudup/resources/addons/storage-openstack.addons.k8s.io/k8s-1.16.yaml.template
ederst: tried to answer this in one place, here: https://github.com/kubernetes/kops/pull/15095#issuecomment-1415685298

zetaab:
/retest

@ederst IMO this looks almost correct in case of CSI cinder. The only missing item is that we should populate this clusterName in case of new clusters. If we do that, we make it better for users hopefully in future. Perhaps good place for that is in https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster.go#L326

However, I checked our kops occm installations and those does not have cluster-name flag at all for some reason(?).. anyways that can be handled as separate issue/pr if needed.

ederst:
> /retest
> 
> @ederst IMO this looks almost correct in case of CSI cinder. The only missing item is that we should populate this clusterName in case of new clusters. If we do that, we make it better for users hopefully in future. Perhaps good place for that is in https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster.go#L326

Ah, OK now I think I know what you mean. Sonds good, will check out the issue (on Monday probably).

> However, I checked our kops occm installations and those does not have cluster-name flag at all for some reason(?).. anyways that can be handled as separate issue/pr if needed.
> 

Strange, but maybe it is only available in OCCM >=1.26 b/c it was not cherry-picked to older versions? Had this issue with some features.

zetaab (approved):

hakman:
Please squash the commits first.
/hold

zetaab:
/approve cancel

also we need that default value to https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster.go#L326

sorry I was too fast, did not remember that missing piece :)

ederst:
@hakman @zetaab good to go now, or should the `--cluster` flag be set to the value of the cloud controller manager?

https://github.com/kubernetes/kops/pull/15139/files#diff-cf0e850db1e4461eab5d4dac860d20f2176b940f236e50825ae538ee23688a1e

hakman:
Thanks @ederst. Leaving the final review to @zetaab.
/hold cancel

zetaab:
@ederst if you could add the clustername by default here https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster.go#L326

something like 

```
			BlockStorage: &api.OpenstackBlockStorageConfig{
				Version:     fi.PtrTo("v3"),
				IgnoreAZ:    fi.PtrTo(opt.OpenstackStorageIgnoreAZ),
				ClusterName: opt.ClusterName
			},
```

ederst:
@zetaab done like suggested, updated commits again so that code changes are separate from the generated stuff

zetaab (approved):
thank you

Issue #15094, opened 2023-02-01
kops GPU support not working due to broken nvidia apt repo


darintay:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
Client version: 1.25.2 (git-v1.25.2)

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:58:47Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:52:18Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
Created a cluster with default image (ubuntu 20.04) using the kops gpu support (https://github.com/kubernetes/kops/blob/master/docs/gpu.md) and tried to bring up a node

**5. What happened after the commands executed?**
Node never joined cluster
It is stuck in nodeup trying to install nvidia-container-runtime and failing, due to https://github.com/NVIDIA/libnvidia-container/issues/202

**6. What did you expect to happen?**
Node joins cluster

**9. Anything else do we need to know?**

I can fix the problem manually by ssh'ing into the node and editing /etc/apt/sources.list.d/nvidia-container-runtime.list to point at the 18.04 link, which still works.  I'm not sure there's a way I can do that automatically with cloud-init or hooks - I tried a bit but it seems like the nodeup process is getting stuck on the package install before anything I can script in my cluster config.

Reverting to an 18.04 base image also works, as expected.

I'm not sure there's anything to be done on the kops side without nvidia fixing their repo, but thought I'd file for visibility and in case there are workarounds I haven't thought of.

elezar:
@darintay do you have a link to the `Dockerfile` used to build the image being used?

The non-`ubuntu18.04` repo should not be used directly if possible and I would like to validate how the repo lists are set up in the image.

[...]

PR #15096, opened 2023-02-02
Fix nvidia container toolkit repos

elezar:
This PR updates the repositories and the installation of the NVIDIA Container Toolkit.

This ensures that #15094 is less likely to occur in future.

### DIFF
--- a/nodeup/pkg/model/nvidia.go
+++ b/nodeup/pkg/model/nvidia.go
@@ -32,13 +32,13 @@ var _ fi.NodeupModelBuilder = &NvidiaBuilder{}
 func (b *NvidiaBuilder) Build(c *fi.NodeupModelBuilderContext) error {
 	if b.InstallNvidiaRuntime() && b.Distribution.IsUbuntu() {
 		c.AddTask(&nodetasks.AptSource{
-			Name:    "nvidia-container-runtime",
+			Name:    "nvidia-container-toolkit",
 			Keyring: "https://nvidia.github.io/libnvidia-container/gpgkey",
 			Sources: []string{
 				"deb https://nvidia.github.io/libnvidia-container/stable/ubuntu18.04/$(ARCH) /",
 			},
 		})
-		c.AddTask(&nodetasks.Package{Name: "nvidia-container-runtime"})
+		c.AddTask(&nodetasks.Package{Name: "nvidia-container-toolkit"})
 		c.AddTask(&nodetasks.Package{Name: b.NodeupConfig.NvidiaGPU.DriverPackage})
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


hakman:
Thanks @elezar. Much appreciated!
/lgtm
/assign @olemarkus

nodeup/pkg/model/nvidia.go
```diff
@@ -33,24 +31,14 @@ var _ fi.NodeupModelBuilder = &NvidiaBuilder{}
...
 		c.AddTask(&nodetasks.AptSource{
-			Name:    "nvidia-container-runtime",
-			Keyring: "https://nvidia.github.io/nvidia-container-runtime/gpgkey",
+			Name:    "nvidia-container-toolkit",
+			Keyring: "https://nvidia.github.io/libnvidia-container/gpgkey",
 			Sources: []string{
-				fmt.Sprintf("deb https://nvidia.github.io/nvidia-container-runtime/stable/ubuntu%s/$(ARCH) /", version),
-				fmt.Sprintf("deb https://nvidia.github.io/libnvidia-container/stable/ubuntu%s/$(ARCH) /", version),
-				fmt.Sprintf("deb https://nvidia.github.io/nvidia-docker/ubuntu%s/$(ARCH) /", version),
+				"deb https://nvidia.github.io/libnvidia-container/stable/ubuntu18.04/$(ARCH) /",
```
zetaab, line 38: well we cannot hardcode it like this

zetaab (changes requested):

nodeup/pkg/model/nvidia.go
elezar: Sure, I can update this. As a matter of interest why can't it be hardcoded? If the only value is ubuntu18.04 the string manipulation is superfluous.

olemarkus: I assume the ubuntu18 means 18+ in this case? So it will work with linking against libc and other libraries used in other distros. The important bit is that it works on all our supported versions.

olemarkus: Based on the issue this resolves, that seems indeed the case.

olemarkus (approved):
Thanks for taking the time to fix this.

nodeup/pkg/model/nvidia.go
elezar: Yes, the `ubuntu18.04` packages are used for all newer Ubuntu versions. This is also the URL that is included in our repo lists that are downloaded as part of our official documentation.

elezar: @zetaab does this explanation resolve your questions?

zetaab:
@elezar do you want to do cherry pick of this PR to release-1.26 as well? Its broken there as well

hakman:
@zetaab Probably easier if you do it. I can approve those PRs.

zetaab (approved):

elezar:
Thanks for the quick turnaround on this!

PR #15102, opened 2023-02-03
Retry creating failed OpenStack servers

ederst:
This aims to improve the experience when creating openstack servers and they run into issues during scheduling, which is not covered by the create API request.

So after creating the instance kops waits for it to become ACTIVE and if not, tries to reprovision the instance by deleting the failed instance and creating a new one.

If the last attempt was still not successful, erroneous instances will persist to allow further investigation, and the task will fail, which will ultimately fail the update call.

### DIFF
--- a/upup/pkg/fi/cloudup/openstack/instance.go
+++ b/upup/pkg/fi/cloudup/openstack/instance.go
@@ -44,6 +44,10 @@ const (
 	BOOT_FROM_VOLUME          = "osVolumeBoot"
 	BOOT_VOLUME_SIZE          = "osVolumeSize"
 	SERVER_GROUP_AFFINITY     = "serverGroupAffinity"
+
+	defaultActiveTimeout = time.Second * 120
+	activeStatus         = "ACTIVE"
+	errorStatus          = "ERROR"
 )
 
 // floatingBackoff is the backoff strategy for listing openstack floatingips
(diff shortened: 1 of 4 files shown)


upup/pkg/fi/cloudup/openstack/instance.go
```diff
@@ -91,6 +100,20 @@ func createInstance(c OpenstackCloud, opt servers.CreateOptsBuilder, portID stri
 			return false, fmt.Errorf("error creating server %v: %v", opt, err)
 		}
 		server = v
+
+		// TODO: 30 is such a magic number, configurable? or just const?
+		err = servers.WaitForStatus(c.ComputeClient(), server.ID, "ACTIVE", 30)
```
zetaab, line 105: when we create the server it should go from BUILD -> ACTIVE, if it goes anything else than ACTIVE we should fail this without waiting?

https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/util.go#L9 this will poll until the server is in ACTIVE and will ignore the ERROR state (when we could already exit the function).

zetaab: but ok, 30 secs is quite fast anyway.. in our case 30 secs is not maybe enough of waiting

ederst: > when we create the server it should go from BUILD -> ACTIVE, if it goes anything else than ACTIVE we should fail this without waiting?
> 
> https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/util.go#L9 this will poll until the server is in ACTIVE and will ignore the ERROR state (when we could already exit the function).

yep, that would probably make more sense: waiting until the new server transitions away from `BUILD` and then see what state it's in:
* if `ACTIVE` everything's fine
* if other state (probably only `ERROR` possible in this case?), then fail

in some cases transitioning from `BUILD` to `ACTIVE` or `ERROR` will take some time though - we have some overloaded OpenStack clouds where our CI system provisions lots of servers and provisioning sometimes takes longer than 30secs - so "failing immediately" is relative. :)

I'll try to cook something up in that regard, and maybe also make the waiting for the state optional (feature flag?), so that the current way of provisioning will not be changed (someone may rely on that behavior?).

upup/pkg/fi/cloudup/openstack/instance.go
```diff
@@ -65,10 +69,45 @@ func IsPortInUse(err error) bool {
...
+			return false, fmt.Errorf("unable to create server: %v", server.Fault)
+		}
+
+		if server.Status == activeStatus {
+			return true, nil
+		}
+
+		return false, nil
+	})
+}
```
ederst, line 93: @zetaab this `WaitFor()` will now stop immediately when a server transitions into status `ERROR` and i have increased the timeout to 120 seconds (like in the former `WaitForStatusActive()` func in [openstacktasks/instance.go](https://github.com/kubernetes/kops/pull/15102/files#diff-948b4589cbdb93053f2dbe3ee8a1e710d3d171030bcfe2b4efdbba88ad8c80eeL99-L102), which AFAIK is not used anywhere.

ederst: i will also try to create a test for this, but i struggle a bit with the mockcloud implementation right now

ederst:
@zetaab now why I struggled with the OpenStack MockCloud:

Because there is no implementations for getting a server:

https://github.com/kubernetes/kops/blob/9edc7ec67886890e83d6cadedf32b00f1835441c/cloudmock/openstack/mockcompute/servers.go#L133-L147

This is also why the lifecycle integration tests currently fail (e.g. `TestLifecycleMinimalOpenstack`).

Well, will adapt... :)

zetaab (approved):
/approve
/lgtm

PR #15110, opened 2023-02-05
gce: Always apply the metadata-proxy-ready node label

justinsb:
This restores the behaviour before #14127, which wasn't documented /
intended.


zetaab (approved):

zetaab:
@justinsb can you cherry pick this to 1.26 and 1.25

PR #15111, opened 2023-02-06
Switch contained config file path to `/etc/containerd/config.toml`

hakman:
Using `/etc/containerd/config-kops.toml` was useful when kOps installed deb/rpm packages from https://download.docker.com and was not safe to overwrite the package config file.
Since switching to `.tgz` packages, there is no longer a need for a special config file name, which is mostly confusing for operators.

### DIFF
--- a/docs/releases/1.26-NOTES.md
+++ b/docs/releases/1.26-NOTES.md
@@ -55,7 +55,9 @@ error is encountered while updating an InstanceGroup.
 
 # Other changes of note
 
-* Add unreachable route for pod IP on deletion option in Cilium.
+* containerd config is now written to `/etc/containerd/config.toml`.
+
+* Cilium can be configured to add unreachable route for pod IP on deletion.
 
 # Breaking changes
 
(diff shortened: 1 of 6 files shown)


docs/releases/1.26-NOTES.md
```diff
@@ -63,6 +63,8 @@ error is encountered while updating an InstanceGroup.
 
 * Support for Kubernetes version 1.20 has been removed.
 
+* containerd config is now written to `/etc/containerd/config.toml`.
```
olemarkus, line 66: I don't really consider this breaking change. It's not like we support operators interfering directly with this file. I'd move this to "other changes of note".

olemarkus (changes requested):

docs/releases/1.26-NOTES.md
hakman: True, they should not be messing with it.

olemarkus (approved):

PR #15112, opened 2023-02-06
Remove obsolete etcd versions

hakman:


justinsb:
Thanks @hakman - this should make startup time noticeably faster!

/approve
/lgtm

PR #15109, opened 2023-02-04
Bump AWS CNI to 1.12.2

moshevayner:
https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.12.2

PR #15114, opened 2023-02-06
Disable kops-configuration after package updates

hakman:
When the `systemd` package is updated, it restarts and `kops-configuration` will not continue because of #15078.

See test failure in https://github.com/kubernetes/kops/pull/15109.

### DIFF
--- a/nodeup/pkg/bootstrap/install.go
+++ b/nodeup/pkg/bootstrap/install.go
@@ -175,7 +175,7 @@ func (i *Installation) buildSystemdJob() *nodetasks.InstallService {
 	service := &nodetasks.InstallService{Service: nodetasks.Service{
 		Name:       serviceName,
 		Definition: fi.PtrTo(manifestString),
-		Enabled:    fi.PtrTo(false),
+		Enabled:    fi.PtrTo(true),
 	}}
 
 	service.InitDefaults()
--- a/nodeup/pkg/bootstrap/tests/simple/tasks.yaml
+++ b/nodeup/pkg/bootstrap/tests/simple/tasks.yaml
@@ -17,7 +17,7 @@ definition: |
 
   [Install]
   WantedBy=multi-user.target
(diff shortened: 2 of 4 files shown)


nodeup/pkg/bootstrap/install.go
```diff
@@ -175,7 +175,7 @@ func (i *Installation) buildSystemdJob() *nodetasks.InstallService {
 	service := &nodetasks.InstallService{Service: nodetasks.Service{
 		Name:       serviceName,
 		Definition: fi.PtrTo(manifestString),
-		Enabled:    fi.PtrTo(false),
+		Enabled:    fi.PtrTo(true),
```
zetaab, line 178: like said in slack: if we just apply this change the service will be always enabled in systemd level. That another model kops-configuration.go does not matter anything.

zetaab: I am thinking could we enable this service BEFORE we start running normal kops-configuration process? In that case if systemd will restart the process it will start again? However, finally we could disable the service? Does it make sense? No idea does it help really

hakman: You are right, this is not the right way to go about this. @justinsb suggested to return some specific message when node already registered and exit gracefully.

zetaab: that is actually good idea, then we could keep service enabled like before

hakman: Exactly! :)

hakman: Let's start with the reverts and we can do ^^^ later next week.
https://github.com/kubernetes/kops/pull/15129
https://github.com/kubernetes/kops/pull/15130

zetaab:
this pr perhaps not needed anymore see https://github.com/kubernetes/kops/pull/15138

PR #15117, opened 2023-02-09
Update golangci-lint to v1.51.1

hakman:
This should address the recent failures in `pull-kops-verify-golangci-lint`:
https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kops-verify-golangci-lint


justinsb:
We discussed in office hours,cancelling hold :-)

/hold cancel

PR #15121, opened 2023-02-10
Use cloud-discovery on GCE in gossip mode

justinsb:
It's a little simpler and should speed up our boot.

### DIFF
--- a/nodeup/pkg/model/bootstrap_client.go
+++ b/nodeup/pkg/model/bootstrap_client.go
@@ -25,9 +25,11 @@ import (
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/bootstrap"
 	"k8s.io/kops/pkg/kopscontrollerclient"
+	"k8s.io/kops/pkg/resolver"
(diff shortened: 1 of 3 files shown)


nodeup/pkg/model/bootstrap_client.go
```diff
@@ -45,14 +47,20 @@ func (b BootstrapClientBuilder) Build(c *fi.NodeupModelBuilderContext) error {
...
 	case kops.CloudProviderGCE:
 		authenticator, err = gcetpmsigner.NewTPMAuthenticator()
-		// We don't use the custom resolver here in gossip mode (though we could);
-		// instead we use this as a check that protokube has now started.
+
+		discovery, err := gcediscovery.New()
+		if err != nil {
+			return err
+		}
+		resolver = discovery
```
hakman, line 63: `gcetpmsigner.NewTPMAuthenticator` also returns error, so this would overwrite that. Move this below the error check for `authenticator`?

justinsb: Ouch - good call, thank you!

hakman (approved):

PR #15122, opened 2023-02-10
scaleway: get credentials from Scaleway profile

Mia-Cross:
The Scaleway Developer Tools ecosystem allows users to define profiles in a yaml file that looks like this :
```yaml
profiles:
  default-profile:
    access_key: SCWXXXXXXXXXXXXXXXX
    secret_key: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
    default_project_id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
    default_zone: fr-par-1
    default_region: fr-par
```

This PR adds the possibility for kOps to get the credentials from this file rather than from the environment by defining `SCW_PROFILE=default-profile` in the environment.

They can still choose to provide their credentials by defining SCW_ACCESS_KEY, SCW_SECRET_KEY and SCW_DEFAULT_PROJECT_ID but I will update the documentation in  #14735 in order to encourage them to use the profiles instead since it's more secure.


hakman (approved):
Seems a good approach. Thanks @Mia-Cross.
I think we are getting close now :).

PR #15125, opened 2023-02-11
Perform challenge callbacks into a node

justinsb:
In order to verify that the caller is running on the specified node, we source the expected IP address from the cloud, and require that the node set up a simple challenge/response server to answer requests.

Because the challenge server runs on a port outside of the nodePort range, this also makes it harder for pods to impersonate their host nodes - though we do combine this with TPM and similar functionality where it is available.


zetaab:
one notable thing: people can run pods in hostNetwork and then its possible to impersonate. But as I see this: why we should accept calls anymore from instances that are already members of kubernetes cluster? If we do not accept that, then its pretty easy to protect against normal pods. 

One thing that is coming to my mind: if node lifetime is enough long it might be that kubelet certificates are going to expire and re-register is needed (kops-controller should accept in that case?)?

justinsb:
> one notable thing: people can run pods in hostNetwork and then its possible to impersonate. But as I see this: why we should accept calls anymore from instances that are already members of kubernetes cluster? If we do not accept that, then its pretty easy to protect against normal pods.

So I think we need to do this PR _and_ prevent the node from re-registering.  We need this so that we can have higher-confidence when the node joins (particularly on clouds where we don't get strong node-attestation, for example DigitalOcean).  And we want to prevent the node from re-registering to avoid attacks where a pod tries to impersonate a node.  We do want to allow some re-registration (as you've pointed out, where the node reboots, or where the cert expires) ... I am thinking we want a machine-key or similar, but I think we've agreed that we can iterate on that!

justinsb:
/retest

Going to look into each of the failures, but they all appear to be unrelated

justinsb:
Test failures matched https://github.com/kubernetes/kubernetes/issues/117363 , i.e. a data race in k8s

justinsb:
I made the changes here to only run this on hetzner (and I'll rebase the digitalocean branch after this merges).

justinsb:
Thanks for reviewing @hakman 

/hold cancel

PR #15126, opened 2023-02-11
Use cloud-based discovery for AWS in gossip mode

justinsb:
Use cloud-based discovery for AWS in gossip mode

A little faster and more consistent.


hakman:
@justinsb Anything you want to keep here?

PR #15130, opened 2023-02-12
Revert "Automated cherry pick of #15078: disable kops-configuration.service after successful execution"

hakman:
Reverts kubernetes/kops#15085

/cc @zetaab @justinsb


zetaab (approved):
We need to implement that graceful exit instead

PR #15138, opened 2023-02-12
exit nodeup gracefully if server already exists in k8s

zetaab:
Original issue for this is https://github.com/kubernetes/kops/issues/15057 but it was reverted due to other issues in https://github.com/kubernetes/kops/pull/15129

some discussion available in https://github.com/kubernetes/kops/pull/15114#pullrequestreview-1287780870

I have verified that this solution works in OpenStack. When existing instance tries to join again to cluster (it does that when rebooting server OR restarting kops-configuration.service):

kops-controller log:
```
kops-controller-t4b8j kops-controller I0212 14:11:02.063861       1 server.go:148] 10.2.96.248:59480: instance already exists in kubernetes cluster
```

node log
```
Feb 12 14:11:03 nodes-helpa-fctnib nodeup[669]: I0212 14:11:03.187949     669 client.go:156] bootstrap request responded with code 204
Feb 12 14:11:03 nodes-helpa-fctnib systemd[1]: kops-configuration.service: Succeeded.
Feb 12 14:11:03 nodes-helpa-fctnib systemd[1]: Finished Run kOps bootstrap (nodeup).
```

So it works like planned. The only possible problem in this solution is that if someone do have older than 455 days https://github.com/kubernetes/kops/blob/master/cmd/kops-controller/pkg/server/server.go#L188 kubelet, the cert will expiry and it cannot renew the cert because it will be still part of the Kubernetes cluster (node is maybe in `NotReady` state). Currently the workaround for that is delete the node manually `kubectl delete node...` and restart it in cloudprovider (or kops-configuration.service in node itself). After that the node can request the new certificate. I am just hoping that people could update their clusters in time and they should NOT have 455 day old nodes in their cluster.

cc @justinsb @hakman

### DIFF
--- a/cmd/kops-controller/pkg/server/server.go
+++ b/cmd/kops-controller/pkg/server/server.go
@@ -142,6 +142,12 @@ func (s *Server) bootstrap(w http.ResponseWriter, r *http.Request) {
 
 	id, err := s.verifier.VerifyToken(ctx, r, r.Header.Get("Authorization"), body, s.opt.Server.UseInstanceIDForNodeName)
 	if err != nil {
+		// means that we should exit nodeup gracefully
+		if err == bootstrap.ErrAlreadyExists {
+			w.WriteHeader(http.StatusNoContent)
+			klog.Infof("%s: %v", r.RemoteAddr, err)
+			return
+		}
 		klog.Infof("bootstrap %s verify err: %v", r.RemoteAddr, err)
 		w.WriteHeader(http.StatusForbidden)
 		// don't return the error; this allows us to have richer errors without security implications
--- a/cmd/kops-controller/pkg/server/server.go
+++ b/cmd/kops-controller/pkg/server/server.go
@@ -144,7 +144,7 @@ func (s *Server) bootstrap(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		// means that we should exit nodeup gracefully
 		if err == bootstrap.ErrAlreadyExists {
-			w.WriteHeader(http.StatusNoContent)
+			w.WriteHeader(http.StatusConflict)
 			klog.Infof("%s: %v", r.RemoteAddr, err)
 			return
 		}
--- a/pkg/bootstrap/authenticate.go
+++ b/pkg/bootstrap/authenticate.go
@@ -18,8 +18,11 @@ package bootstrap
 
 import (
 	"context"
+	"errors"
 	"net/http"
 )
+
+var ErrAlreadyExists = errors.New("node already exists")
 
 // Authenticator generates authentication credentials for requests.
 type Authenticator interface {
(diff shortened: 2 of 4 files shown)


pkg/kopscontrollerclient/client.go
```diff
@@ -150,6 +151,12 @@ func (b *Client) Query(ctx context.Context, req any, resp any) error {
 		defer response.Body.Close()
 	}
 
+	// if we receive StatusNoContent it means that we should exit gracefully
+	if response.StatusCode == http.StatusNoContent {
+		klog.Infof("bootstrap request responded with code %d", response.StatusCode)
```
hakman, line 156:
```suggestion
		klog.Infof("kops-controller returned status code %d", response.StatusCode)
```

pkg/bootstrap/authenticate.go
```diff
@@ -18,9 +18,12 @@ package bootstrap
 
 import (
 	"context"
+	"errors"
 	"net/http"
 )
 
+var ErrAlreadyExists = errors.New("instance already exists in kubernetes cluster")
```
hakman, line 25:
```suggestion
var ErrAlreadyExists = errors.New("node already exists")
```

hakman (changes requested):

cmd/kops-controller/pkg/server/server.go
```diff
@@ -142,6 +142,12 @@ func (s *Server) bootstrap(w http.ResponseWriter, r *http.Request) {
 
 	id, err := s.verifier.VerifyToken(ctx, r, r.Header.Get("Authorization"), body, s.opt.Server.UseInstanceIDForNodeName)
 	if err != nil {
+		// means that we should exit nodeup gracefully
+		if err == bootstrap.ErrAlreadyExists {
+			w.WriteHeader(http.StatusNoContent)
```
justinsb, line 147: Nit-picking, but we could also do http.StatusConflict - that's what I've seen from GCP APIs at least.

justinsb:
One nit pick on the http status code, but I like this approach - a specific response, and exit with 0 :+1:

justinsb:
On the expiry side, I think we should figure out what behaviour we want here when the node already exists.  IIUC this is mostly a security defense-in-depth, so as we beef up auth here (e.g. the [node callback](https://github.com/kubernetes/kops/pull/15125) ) then we might not need it at all.  Whatever we decide, I'd like for us to try to converge the various clouds so they have similar behvaior here.

Anyway, I don't think my nit on the status code is blocking ....

/lgtm

hakman:
> On the expiry side, I think we should figure out what behaviour we want here when the node already exists. IIUC this is mostly a security defense-in-depth, so as we beef up auth here (e.g. the [node callback](https://github.com/kubernetes/kops/pull/15125) ) then we might not need it at all. Whatever we decide, I'd like for us to try to converge the various clouds so they have similar behvaior here.

It is also nice to know that the node config is sent only once and we don't update it dynamically on reboot, if something changes in the IG config.

hakman (approved):

PR #15139, opened 2023-02-12
Add cluster name to OCCM opts

### LINKED ISSUE #15108: Openstack CCM should have --cluster-name defined
/kind feature

Like discussed in https://github.com/kubernetes/kops/pull/15095#issuecomment-1414352848 we should have `--cluster-name` defined in our clusters if we want to run multiple Kubernetes clusters in same project.

Currently when its not enabled it uses default value `kubernetes` and we cannot modify the flag for existing clusters. I am not 100% sure but we could try populating the https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/componentconfig.go#L676 value in https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/new_cluster.go#L326 and see will it be added to ccm configuration.

zetaab:
fixes #15108

We cannot add it to existing clusters (this value is used things like loadbalancer and routes), we can add it only to new clusters.

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -1438,6 +1438,11 @@ func initializeOpenstack(opt *NewClusterOptions, cluster *api.Cluster) {
 			cluster.Spec.Networking.Topology.DNS = api.DNSTypeNone
 		}
 	}
+
+	if cluster.Spec.ExternalCloudControllerManager == nil {
+		cluster.Spec.ExternalCloudControllerManager = &api.CloudControllerManagerConfig{}
+	}
+	cluster.Spec.ExternalCloudControllerManager.ClusterName = opt.ClusterName
 }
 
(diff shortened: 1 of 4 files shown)


upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1438,6 +1438,14 @@ func initializeOpenstack(opt *NewClusterOptions, cluster *api.Cluster) {
...
 		}
 	}
+
+	if cluster.Spec.ExternalCloudControllerManager == nil {
+		cluster.Spec.ExternalCloudControllerManager = &api.CloudControllerManagerConfig{
+			ClusterName: opt.ClusterName,
+		}
+	} else {
+		cluster.Spec.ExternalCloudControllerManager.ClusterName = opt.ClusterName
+	}
```
hakman, line 1448:
```suggestion
	if cluster.Spec.ExternalCloudControllerManager == nil {
		cluster.Spec.ExternalCloudControllerManager = &api.CloudControllerManagerConfig{}
	}
	cluster.Spec.ExternalCloudControllerManager.ClusterName = opt.ClusterName
```

hakman (changes requested):

hakman (approved):

PR #15141, opened 2023-02-12
hetzner: add dependency logic to deletion

justinsb:
This avoids warnings about deleting e.g. the volume while it is still
attached to a server.

### DIFF
--- a/pkg/resources/hetzner/resources.go
+++ b/pkg/resources/hetzner/resources.go
@@ -182,6 +182,10 @@ func listServers(cloud fi.Cloud, clusterName string) ([]*resources.Resource, err
 			Obj:     server,
 		}
 
+		for _, firewall := range server.PublicNet.Firewalls {
+			resourceTracker.Blocks = append(resourceTracker.Blocks, fmt.Sprintf("%s:%d", resourceTypeFirewall, firewall.Firewall.ID))
+		}
+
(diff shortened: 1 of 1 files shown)


pkg/resources/hetzner/resources.go
```diff
@@ -269,16 +277,22 @@ func deleteLoadBalancer(cloud fi.Cloud, r *resources.Resource) error {
...
 	server := r.Obj.(*hcloud.Server)
-	_, err := client.Delete(context.TODO(), server)
+	result, _, err := serverClient.DeleteWithResult(ctx, server)
 	if err != nil {
 		return fmt.Errorf("failed to delete server %s(%s): %w", r.Name, r.ID, err)
 	}
 
+	if _, err := hetzner.WaitForAction(ctx, c, result.Action); err != nil {
+		return fmt.Errorf("error waiting for server deletion: %w", err)
+	}
```
hakman, line 294: Wouldn't using `actionClient.WatchProgress()` be simpler?
```suggestion
	_, errCh := actionClient.WatchProgress(ctx, result.Action)
	if err := <-errCh; err != nil {
		return err
	}
```

justinsb: That first return value is suspicious, because it seems like most of the messages to it will be dropped.  But as we're not using it, yes - good call...

hakman (approved):

PR #15144, opened 2023-02-12
Upgrade Karpenter to v0.27.5

anthonyhaussman:
## [feat(karpenter): Upgrade to version 0.27.0](https://github.com/kubernetes/kops/pull/15144/commits/1f1c2e9e6d933696a9a0a2e651020c019bfdb4c6)

Upgrade Karpenter to the current last stable version `0.27.0`. Templates have been updated to use the same templates as the Helm chart.

Some noticeable changes:
 - Rename `karpenter-config-logging` ConfigMap to `config-logging`
 - Some environment variables configurations are now in the `karpenter-global-settings` ConfigMap
 - Karpenter use now a common pod for the controller and webhook.

## [feat(karpenter): Use AWSNodeTemplate for launchTemplate](https://github.com/kubernetes/kops/pull/15144/commits/5bd655e09eaf6a6e2ae297edc9d6a912504eca16) 

To set Launch Templates is deprecated into the provisioner, it is recommended to use the `AWSNodeTemplate` to set it.
Ref:
 - https://karpenter.sh/v0.27.0/concepts/node-templates/

## [feat(karpenter): Drop the karpenter feature flag](https://github.com/kubernetes/kops/pull/15144/commits/1d457c438e77dd356de15b8db70ca46154bf2e97)

Drop the `Karpenter` feature flag.  it's fairly well-supported now.

### DIFF
--- a/docs/operations/karpenter.md
+++ b/docs/operations/karpenter.md
@@ -2,15 +2,7 @@
 
 [Karpenter](https://karpenter.sh) is a Kubernetes-native capacity manager that directly provisions Nodes and underlying instances based on Pod requirements. On AWS, kOps supports managing an InstanceGroup with either Karpenter or an AWS Auto Scaling Group (ASG).
 
-Karpenter is a fairly new project, and it is still not determined how Karpenter should work with kOps. Because of this, Karpenter is behind the `Karpenter` feature flag.
-
 ## Installing
-
-Enable the Karpenter feature flag:
-
-```sh
-export KOPS_FEATURE_FLAGS="Karpenter"
-```
 
 Karpenter requires that external permissions for ServiceAccounts be enabled for the cluster. See [AWS IAM roles for ServiceAccounts documentation](/cluster_spec#service-account-issuer-discovery-and-aws-iam-roles-for-service-accounts-irsa) for how to enable this. 
 
--- a/docs/operations/karpenter.md
+++ b/docs/operations/karpenter.md
@@ -4,6 +4,12 @@
 
 ## Installing
 
+If using kOps 1.26 or older, enable the Karpenter feature flag :
+
+```sh
+export KOPS_FEATURE_FLAGS="Karpenter"
+```
+
 Karpenter requires that external permissions for ServiceAccounts be enabled for the cluster. See [AWS IAM roles for ServiceAccounts documentation](/cluster_spec#service-account-issuer-discovery-and-aws-iam-roles-for-service-accounts-irsa) for how to enable this. 
 
 ### Existing clusters
--- a/docs/releases/1.27-NOTES.md
+++ b/docs/releases/1.27-NOTES.md
@@ -11,6 +11,8 @@ This behaviour can be overridden by setting `spec.etcdClusters[*].manager.backup
 
 * external-dns is now supported in IPv6 clusters.
 
+* `Karpenter` has been upgraded to version `0.27.x` and the feature flag is no longer necessary.
+
 ## AWS
 
 * As of Kubernetes version 1.27, all nodes will default to running with instance-metadata-service tokens required, with a max hop limit of 1.
--- a/tests/e2e/scenarios/upgrade-ab/run-test.sh
+++ b/tests/e2e/scenarios/upgrade-ab/run-test.sh
@@ -14,8 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-export KOPS_FEATURE_FLAGS="SpecOverrideFlag,"
-
 REPO_ROOT=$(git rev-parse --show-toplevel);
 source "${REPO_ROOT}"/tests/e2e/scenarios/lib/common.sh
 
--- a/tests/e2e/scenarios/upgrade-ab/run-test.sh
+++ b/tests/e2e/scenarios/upgrade-ab/run-test.sh
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+export KOPS_FEATURE_FLAGS="SpecOverrideFlag,"
+
 REPO_ROOT=$(git rev-parse --show-toplevel);
 source "${REPO_ROOT}"/tests/e2e/scenarios/lib/common.sh
 
@@ -78,7 +80,10 @@ else
   KOPS="${KOPS_A}"
 fi
 
-
+create_args=""
+if [[ ${KOPS_IRSA-} = true ]]; then
+  create_args="${create_args} --discovery-store=${DISCOVERY_STORE}/${CLUSTER_NAME}/discovery"
+fi
 
 ${KUBETEST2} \
     --up \
@@ -86,7 +91,7 @@ ${KUBETEST2} \
     --kubernetes-version="${K8S_VERSION_A}" \
     --control-plane-size="${KOPS_CONTROL_PLANE_SIZE:-1}" \
     --template-path="${KOPS_TEMPLATE:-}" \
-    --create-args="--networking calico"
+    --create-args="--networking calico ${KOPS_EXTRA_FLAGS:-} ${create_args}"
 
 # Export kubeconfig-a
 KUBECONFIG_A=$(mktemp -t kops.XXXXXXXXX)
--- a/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
+++ b/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
@@ -961,7 +961,7 @@ spec:
       dnsPolicy: Default
       containers:
         - name: controller
-          image: public.ecr.aws/karpenter/controller:v0.27.3
+          image: public.ecr.aws/karpenter/controller:v0.27.5
           imagePullPolicy: IfNotPresent
           env:
             - name: KUBERNETES_MIN_VERSION
--- a/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
+++ b/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.2
+    controller-gen.kubebuilder.io/version: v0.11.3
   creationTimestamp: null
   name: provisioners.karpenter.sh
 spec:
@@ -373,12 +373,13 @@ spec:
     storage: true
     subresources:
       status: {}
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.2
+    controller-gen.kubebuilder.io/version: v0.11.3
   creationTimestamp: null
   name: awsnodetemplates.karpenter.k8s.aws
 spec:
@@ -697,7 +698,7 @@ data:
     "aws.enableENILimitedPodDensity": "false"
     {{ end }}
     "aws.enablePodENI": "false"
-    "aws.interruptionQueue": "false"
+    "aws.interruptionQueueName": ""
     "aws.isolatedVPC": "false"
     {{ if UsesInstanceIDForNodeName }}
     "aws.nodeNameConvention": "resource-name"
@@ -1093,32 +1094,6 @@ webhooks:
           - awsnodetemplates
           - awsnodetemplates/status
         scope: '*'
-      - apiGroups:
-          - karpenter.sh
-        apiVersions:
-          - v1alpha5
-        resources:
-          - provisioners
-          - provisioners/status
-        operations:
-          - CREATE
-          - UPDATE
----
-# Source: karpenter/templates/webhooks-core.yaml
-apiVersion: admissionregistration.k8s.io/v1
-kind: MutatingWebhookConfiguration
-metadata:
-  name: defaulting.webhook.karpenter.sh
-webhooks:
-  - name: defaulting.webhook.karpenter.sh
-    admissionReviewVersions: ["v1"]
-    clientConfig:
-      service:
-        name: karpenter
-        namespace: kube-system
-    failurePolicy: Fail
-    sideEffects: None
-    rules:
       - apiGroups:
           - karpenter.sh
         apiVersions:
--- a/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
+++ b/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.11.2
   creationTimestamp: null
   name: provisioners.karpenter.sh
 spec:
@@ -42,6 +42,11 @@ spec:
               Node properties are determined from a combination of provisioner and
               pod scheduling constraints.
             properties:
+              annotations:
+                additionalProperties:
+                  type: string
+                description: Annotations are applied to every node.
+                type: object
               consolidation:
                 description: Consolidation are the consolidation parameters
                 properties:
@@ -63,6 +68,10 @@ spec:
                     description: ContainerRuntime is the container runtime to be used
                       with your worker nodes.
                     type: string
+                  cpuCFSQuota:
+                    description: CPUCFSQuota enables CPU CFS quota enforcement for
+                      containers that specify CPU limits.
+                    type: boolean
                   evictionHard:
                     additionalProperties:
                       type: string
@@ -87,6 +96,27 @@ spec:
                     description: EvictionSoftGracePeriod is the map of signal names
                       to quantities that define grace periods for each eviction signal
                     type: object
+                  imageGCHighThresholdPercent:
+                    description: ImageGCHighThresholdPercent is the percent of disk
+                      usage after which image garbage collection is always run. The
+                      percent is calculated by dividing this field value by 100, so
+                      this field must be between 0 and 100, inclusive. When specified,
+                      the value must be greater than ImageGCLowThresholdPercent.
+                    format: int32
+                    maximum: 100
+                    minimum: 0
+                    type: integer
+                  imageGCLowThresholdPercent:
+                    description: ImageGCLowThresholdPercent is the percent of disk
+                      usage before which image garbage collection is never run. Lowest
+                      disk usage to garbage collect to. The percent is calculated
+                      by dividing this field value by 100, so the field value must
+                      be between 0 and 100, inclusive. When specified, the value must
+                      be less than imageGCHighThresholdPercent
+                    format: int32
+                    maximum: 100
+                    minimum: 0
+                    type: integer
                   kubeReserved:
                     additionalProperties:
                       anyOf:
@@ -159,6 +189,8 @@ spec:
                   name:
                     description: 'Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names'
                     type: string
+                required:
+                - name
                 type: object
               requirements:
                 description: Requirements are layered with Labels and applied to every
@@ -341,18 +373,12 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.11.2
   creationTimestamp: null
   name: awsnodetemplates.karpenter.k8s.aws
 spec:
@@ -475,6 +501,10 @@ spec:
               context:
                 description: Context is a Reserved field in EC2 APIs https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html
                 type: string
+              detailedMonitoring:
+                description: DetailedMonitoring controls if detailed monitoring is
+                  enabled for instances that are launched
+                type: boolean
               instanceProfile:
                 description: InstanceProfile is the AWS identity that instances use.
                 type: string
@@ -562,17 +592,43 @@ spec:
                   being provisioned with the correct configuration.
                 type: string
             type: object
+          status:
+            description: AWSNodeTemplateStatus contains the resolved state of the
+              AWSNodeTemplate
+            properties:
+              securityGroups:
+                description: SecurityGroups contains the current Security Groups values
+                  that are available to the cluster under the SecurityGroups selectors.
+                items:
+                  description: SecurityGroupStatus contains resolved SecurityGroup
+                    selector values utilized for node launch
+                  properties:
+                    id:
+                      description: Id of the security group
+                      type: string
+                  type: object
+                type: array
+              subnets:
+                description: Subnets contains the current Subnet values that are available
+                  to the cluster under the subnet selectors.
+                items:
+                  description: SubnetStatus contains resolved Subnet selector values
+                    utilized for node launch
+                  properties:
+                    id:
+                      description: Id of the subnet
+                      type: string
+                    zone:
+                      description: The associated availability zone
+                      type: string
+                  type: object
+                type: array
+            type: object
         type: object
     served: true
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
 ---
 # Source: karpenter/templates/serviceaccount.yaml
 apiVersion: v1
@@ -581,19 +637,19 @@ metadata:
   name: karpenter
   namespace: kube-system
 ---
-# Source: karpenter/templates/webhook/deployment.yaml
+# Source: karpenter/templates/secret-webhook-cert.yaml
 apiVersion: v1
 kind: Secret
 metadata:
-  name: karpenter-webhook-cert
+  name: karpenter-cert
   namespace: kube-system
-data: {} # Injected by karpenter-webhook
+data: {} # Injected by karpenter
 ---
-# Source: karpenter/templates/100-config-logging.yaml
+# Source: karpenter/templates/configmap-logging.yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: karpenter-config-logging
+  name: config-logging
   namespace: kube-system
   labels:
     app.kubernetes.io/part-of: karpenter
@@ -602,7 +658,7 @@ data:
   zap-logger-config: |
     {
       "level": "debug",
-      "development": true,
+      "development": false,
       "disableStacktrace": true,
       "disableCaller": true,
       "sampling": {
@@ -611,7 +667,7 @@ data:
       },
       "outputPaths": ["stdout"],
       "errorOutputPaths": ["stderr"],
-      "encoding": "console",
+      "encoding": "json",
       "encoderConfig": {
         "timeKey": "time",
         "levelKey": "level",
@@ -623,9 +679,7 @@ data:
         "timeEncoder": "iso8601"
       }
     }
-  # Log level overrides
-  loglevel.controller: debug
-  loglevel.webhook: debug
+  loglevel.webhook: "debug"
 ---
 # Source: karpenter/templates/configmap.yaml
 apiVersion: v1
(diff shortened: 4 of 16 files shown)


upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -581,19 +637,19 @@ metadata:
...
   namespace: kube-system
 data: {} # Injected by karpenter-webhook
 ---
-# Source: karpenter/templates/100-config-logging.yaml
+# Source: karpenter/templates/configmap-logging.yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: karpenter-config-logging
+  name: config-logging
```
olemarkus, line 652: This name is the default name for any knative logging config. It is way too generic for `kube-system`.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -623,9 +679,8 @@ data:
         "timeEncoder": "iso8601"
       }
     }
-  # Log level overrides
-  loglevel.controller: debug
-  loglevel.webhook: debug
+  loglevel.webhook: "error"
```
olemarkus, line 682: Since we frequently get asked about debug logs, this probably should be configurable if we move to using error.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -639,286 +694,295 @@ data:
...
+    "aws.clusterEndpoint": "https://{{ APIInternalName }}"
+    "aws.clusterName": "{{ ClusterName }}"
+    "aws.defaultInstanceProfile": ""
+    {{ if not .Networking.AmazonVPC }}
+    "aws.enableENILimitedPodDensity": "true"
+    {{ else }}
+    "aws.enableENILimitedPodDensity": "false"
+    {{ end }}
+    "aws.enablePodENI": "false"
+    "aws.interruptionQueueName": "{{ ClusterName }}"
```
olemarkus, line 713: Do we need to set this? I would like to keep the NTH functionality out of Karpenter on kOps and point people to using the NTH addon instead.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -639,286 +694,295 @@ data:
...
+    {{ end }}
+    "aws.vmMemoryOverheadPercent": "0.075"
+    "batchIdleDuration": "1s"
+    "batchMaxDuration": "10s"
+---
+# Source: karpenter/templates/clusterrole.yaml
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: karpenter-controller
```
olemarkus, line 646: Since this PR changes many of the resource names, pruning must be enabled to clean up old resources: https://github.com/kubernetes/kops/pull/14601/files#diff-13f16294f2d4be71ad5d7207a320b919ec0eb217c5be43a7ee0c91e3b17e1f99R1205

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -639,286 +694,295 @@ data:
...
-      priorityClassName: system-cluster-critical
       serviceAccountName: karpenter
+      securityContext:
+        fsGroup: 1000
+      priorityClassName: "system-cluster-critical"
       dnsPolicy: Default
       containers:
         - name: controller
-          image: public.ecr.aws/karpenter/controller:v0.16.3
+          image: public.ecr.aws/karpenter/controller:v0.24.0@sha256:83e7a21200efd4e17314c931aa715c9c73b27fcbf08673d263c007029ad754da
```
olemarkus, line 973: Please don't include these sha. kOps will automatically add the SHAs for all manifests during cloudup. This allows a bit more flexibility for those who want to rehost/rebuild images.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -639,286 +694,295 @@ data:
...
+      app.kubernetes.io/name: karpenter
+      app.kubernetes.io/instance: karpenter
   template:
     metadata:
       labels:
-        karpenter: controller
         app.kubernetes.io/name: karpenter
+        app.kubernetes.io/instance: karpenter
     spec:
-      affinity:
```
olemarkus, line 862: Why is this being removed?

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -929,243 +993,203 @@ spec:
...
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-            - matchExpressions:
-              - key: kubernetes.io/os
-                operator: In
-                values:
-                - linux
-              - key: karpenter.sh/provisioner-name
-                operator: DoesNotExist
-      tolerations:
```
olemarkus, line 943: Why is this being removed?

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -639,286 +694,295 @@ data:
...
 spec:
   replicas: {{ ControlPlaneControllerReplicas false }}
+  revisionHistoryLimit: 10
   strategy:
-    type: Recreate
+    rollingUpdate:
+      maxUnavailable: 1
   selector:
     matchLabels:
-      karpenter: controller
```
olemarkus, line 855: This will break upgrades I think. Selectors are immutable.

olemarkus (changes requested):
With these massive changes, I think we need an upgrade e2e test with Karpenter. There are quite a few potential breakages here.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: I prefer to re-set it to debug mode and propose an other future PR to be configurable to not overcharge this PR
 - https://github.com/kubernetes/kops/pull/15144/commits/c08c7b5ef17e8b7f4095b90ddf2b462914370a2a

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: Let disable it then :+1: 
 - https://github.com/kubernetes/kops/pull/15144/commits/ba5384b6a1320d7e93027be73106554ad9b935c9

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: Done :pray: 
 - https://github.com/kubernetes/kops/pull/15144/commits/213d11cfb8aa083e6930094e452229b62d1b02b4

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: Good catch, I have changed it to the old value
 - https://github.com/kubernetes/kops/pull/15144/commits/28da422b8429990d6117d9b76bdc1007e21c164b

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: Pruning enabled :ok_hand: 
 - https://github.com/kubernetes/kops/pull/15144/commits/096eb62c33fa2574fc70ed86fef96ba406480e40

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: It's present here
https://github.com/kubernetes/kops/blob/5656e2f37983a99565d264e5542f00ea552665f9/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template#L1021-L1027

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: https://github.com/kubernetes/kops/blob/5656e2f37983a99565d264e5542f00ea552665f9/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template#L1021-L1027

anthonyhaussman: Fixed
 - https://github.com/kubernetes/kops/pull/15144/commits/dee14fb84330644298db6daa3c18840954b963f7

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: Fixed
 - Fixed
 - https://github.com/kubernetes/kops/pull/15144/commits/dee14fb84330644298db6daa3c18840954b963f7

anthonyhaussman:
@olemarkus I have fixed most of the templates et done some tests on a private environment.
Karpenter starts correctly and new nodes appear but do not join the cluster on my side (I think it's a different problem as I have another "classic" InstanceGroup that does not appear also).

On the `e2e` test, I see the Karpenter pod staying in `Pending` status and I'm not really sure why.
I don't have any experience with `e2e` unfortunately :confused: 

I would like to have your and/or the community's helps on this :pray:

anthonyhaussman:
I have locally backported modifications on `kops-1.26.0-beta.2` to try a bootstrap.
Karpenter works like a charm :ok_hand:

olemarkus:
You can see a list of all the pod resources here: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/15144/pull-kops-e2e-aws-karpenter/1625454977755385856/artifacts/cluster-info/kube-system/pods.yaml

Includes their status and reason.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -929,108 +989,35 @@ spec:
...
-              value: karpenter-webhook
-            - name: SYSTEM_NAMESPACE
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.namespace
+              path: /readyz
+              port: http
+          resources:
+            limits:
+              cpu: 1
```
olemarkus, line 1015: This seems a bit much. The pod won't fit on the control plane nodes.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
```diff
@@ -1059,113 +1046,161 @@ spec:
       - key: node-role.kubernetes.io/control-plane
         operator: Exists
       topologySpreadConstraints:
-      - maxSkew: 1
-        topologyKey: topology.kubernetes.io/zone
-        whenUnsatisfiable: ScheduleAnyway
-        labelSelector:
-          matchLabels:
-            karpenter: webhook
-      - maxSkew: 1
```
olemarkus, line 1068: Why is this one being removed?

olemarkus (changes requested):

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: The last version of Karpenter merges the controller and the webhook into the same container.
The new `topologySpreadConstraints` is now (based ib the default output of the helm chart):
https://github.com/kubernetes/kops/blob/dee14fb84330644298db6daa3c18840954b963f7/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template#L1048-L1056

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: Resources updated :ok_hand: 
 - https://github.com/kubernetes/kops/pull/15144/commits/322d94d03043c423b72b0a72e821acf73d68a3c4

hakman:
```
==== START logs for container controller of pod kube-system/karpenter-775f4bbbf5-dsfwd ====
{"level":"info","ts":1676409234.094117,"logger":"fallback","caller":"injection/injection.go:63","msg":"Starting informers..."}
panic: failed to get region from metadata server: EC2MetadataRequestError: failed to get EC2 instance identity document
caused by: EC2MetadataError: failed to make EC2Metadata request

	status code: 401, request id: 

goroutine 1 [running]:
github.com/samber/lo.must({0x1f12b40, 0x4000c67200}, {0x4000abf930, 0x1, 0x1})
	github.com/samber/lo@v1.37.0/errors.go:51 +0x1b8
github.com/samber/lo.Must[...]({0x0?, 0x0}, {0x1f12b40, 0x4000c67200?}, {0x4000abf930?, 0x1?, 0x0?})
	github.com/samber/lo@v1.37.0/errors.go:65 +0x44
github.com/aws/karpenter/pkg/context.NewOrDie({{0x2732758, 0x4000c1da70}, 0x40000f2000, {0x2752518, 0x40000f6300}, {0x273dda8, 0x400056ab40}, {0x270b160, 0x4000c541e0}, {0x2737c70, ...}, ...})
	github.com/aws/karpenter/pkg/context/context.go:64 +0x3dc
main.main()
	github.com/aws/karpenter/cmd/controller/main.go:35 +0x128
==== END logs for container controller of pod kube-system/karpenter-775f4bbbf5-dsfwd ====
```

hakman:
I dont' think the following env vars should have been removed, for both controller and webhook. Not sure about the other ones.
```
- name: AWS_REGION
  value: {{ Region }}
- name: CLUSTER_NAME
  value: {{ ClusterName }}
- name: CLUSTER_ENDPOINT
  value: https://{{ APIInternalName }}
```

anthonyhaussman:
> I dont' think the following env vars should have been removed, for both controller and webhook. Not sure about the other ones.
> 
> ```
> - name: AWS_REGION
>   value: {{ Region }}
> - name: CLUSTER_NAME
>   value: {{ ClusterName }}
> - name: CLUSTER_ENDPOINT
>   value: https://{{ APIInternalName }}
> ```

@hakman, those configuration has been moved to the `karpenter-global-settings` ConfigMap:
https://github.com/kubernetes/kops/blob/b2213eb7fb45d5d2ebf51dbc95f82cb7b2e69458/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template#L697-L722

But I see there's a duplicate. Let me fix it. :eyes:

hakman:
As you can see, AWS SDK complains about the region. I would say it doesn't care much about what `karpenter-global-settings`. Let's see what the tests tell us.
Also, could you tell me how you generated the updated manifest (helm command?).

anthonyhaussman:
> As you can see, AWS SDK complains about the region. I would say it doesn't care much about `karpenter-global-settings`. Let's see what the tests tell us. Also, could you tell me how you generated the updated manifest (helm command?).

Indeed, I have re-set the region as it seems to be still an environment variable.
 - https://github.com/kubernetes/kops/pull/15144/commits/1295c0bbcab6323216c8103516eaede18d621fa7

The helm template command:
```
> helm template karpenter oci://public.ecr.aws/karpenter/karpenter  --version v0.24.0
```

hakman:
You may need to set the region in the webhook also. Let's see the test logs.

anthonyhaussman:
> You may need to set the region in the webhook also. Let's see the test logs.

Karpenter uses now a common pod and container for the controller and webhook.

anthonyhaussman:
> ```
> ==== START logs for container controller of pod kube-system/karpenter-775f4bbbf5-dsfwd ====
> {"level":"info","ts":1676409234.094117,"logger":"fallback","caller":"injection/injection.go:63","msg":"Starting informers..."}
> panic: failed to get region from metadata server: EC2MetadataRequestError: failed to get EC2 instance identity document
> caused by: EC2MetadataError: failed to make EC2Metadata request
> 
> 	status code: 401, request id: 
> 
> goroutine 1 [running]:
> github.com/samber/lo.must({0x1f12b40, 0x4000c67200}, {0x4000abf930, 0x1, 0x1})
> 	github.com/samber/lo@v1.37.0/errors.go:51 +0x1b8
> github.com/samber/lo.Must[...]({0x0?, 0x0}, {0x1f12b40, 0x4000c67200?}, {0x4000abf930?, 0x1?, 0x0?})
> 	github.com/samber/lo@v1.37.0/errors.go:65 +0x44
> github.com/aws/karpenter/pkg/context.NewOrDie({{0x2732758, 0x4000c1da70}, 0x40000f2000, {0x2752518, 0x40000f6300}, {0x273dda8, 0x400056ab40}, {0x270b160, 0x4000c541e0}, {0x2737c70, ...}, ...})
> 	github.com/aws/karpenter/pkg/context/context.go:64 +0x3dc
> main.main()
> 	github.com/aws/karpenter/cmd/controller/main.go:35 +0x128
> ==== END logs for container controller of pod kube-system/karpenter-775f4bbbf5-dsfwd ====
> ```

@hakman could you tell me how you found those logs :pray: ?

hakman:
When the test fails you get a details link. Click on it, then on `Artifacts` (top right), then go to `artifacts/cluster-info/<NAMESPACE>/`. You should see logs for all running pods in named dirs, as well as their status in `pods.yaml`.

anthonyhaussman:
And it seems we are good now :tada:

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
olemarkus: The block you removed is one we use for all addons so it shouldn't be removed just because upstream doesn't have it.

olemarkus (changes requested):

hakman:
Cool, will do some more passes through the template changes, but passing the tests is great.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: Well, we should decide which selector we want to use as Karpenter is a single pod now:
```
karpenter: webhook
```
or
```
karpenter: controller
```
?

The `update-expected` script seems to accept only 1 label "karpenter".

olemarkus: Not sure what you mean what update-expected accepts., It only renders the manifests. These labels need to select the appropriate pods. Usually the same selector one uses for the Deployment selector (and things like PDB).
I am guessing the `karpenter` label isn't required as part of the selector at all.

olemarkus:
I'd like to have an upgrade e2e before this goes in. But I won't be able to create one until later today.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
anthonyhaussman: I have added the webhook label in this commit https://github.com/kubernetes/kops/pull/15144/commits/8baad20aca0d9db2dc7bdbb7aa637d31be969a5e but the update-expected script only keeps the `karpenter: webhook` label in tests files.

olemarkus: Just to be clear, it's the entry using hostname as topology key that is missing. Not selecting specifically for the webhook label.

anthonyhaussman: Haaaaaaaaa ^^'
Sorry I didn't understand that. Let me fix that. :ok_hand:

anthonyhaussman: Fixed in https://github.com/kubernetes/kops/pull/15144/commits/6a76b51cdaef8c2cbdb79eb43ddc1f35f4d7f2e7 :+1:

olemarkus:
Karpenter upgrade e2e did not actually use karpenter. The scenario probably lacks some features for passing on flags.

anthonyhaussman:
> Karpenter upgrade e2e did not actually use karpenter. The scenario probably lacks some features for passing on flags.

Can I do something to help?

olemarkus:
It might be you could add a commit to your PR (or I can push one to your branch) when I've done what's needed on the test-infra side (should be done soon).

anthonyhaussman:
> It might be you could add a commit to your PR (or I can push one to your branch) when I've done what's needed on the test-infra side (should be done soon).

I stay available for this. I'll push a commit when you need it.

hakman:
@anthonyhaussman I was looking at the diff between your changes and what I get from the `helm template` output. Seems there are quite a few differences. Could you check?

anthonyhaussman:
> @anthonyhaussman I was looking at the diff between your changes and what I get from the `helm template` output. Seems there are quite a few differences. Could you check?

There are some differences concerning some labels and selectors to have an easy upgrade process (as selectors are immutable).
Beware of this, I don't catch big problems here @hakman :thinking: 

Maybe I miss something.

anthonyhaussman:
> This is what I have: https://gist.github.com/hakman/968fbc62401393ac49b16fff700fcda1.

I have crossed check. Looks good to me. :slightly_smiling_face:

olemarkus:
Multiple feature flags still not working properly in that upgrade scenario. But wonder if we can drop the feature flag with this change. Should be fairly well supported now.

anthonyhaussman:
> Multiple feature flags still not working properly in that upgrade scenario. But wonder if we can drop the feature flag with this change. Should be fairly well supported now.

It could be dropped from my point of view. I use the Karpenter feature since several month now without problems.

anthonyhaussman:
@olemarkus Shall we change the default Karpenter flag to `true` as a first step?
https://github.com/kubernetes/kops/blob/c56a983355cd675435a29e575b4a4ee810eaf268/pkg/featureflag/featureflag.go#L85

anthonyhaussman:
I have pushed https://github.com/kubernetes/kops/pull/15144/commits/67fe9a0664eddf613884855cf95170b611808b6b to drop the `karpenter` feature flag :ok_hand:

anthonyhaussman:
@olemarkus  I have pushed https://github.com/kubernetes/kops/pull/15144/commits/67fe9a0664eddf613884855cf95170b611808b6b to drop the `karpenter` feature flag :ok_hand:

tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -88,7 +88,7 @@ ${KUBETEST2} \
     --kubernetes-version="${K8S_VERSION_A}" \
     --control-plane-size="${KOPS_CONTROL_PLANE_SIZE:-1}" \
     --template-path="${KOPS_TEMPLATE:-}" \
-    --create-args="--networking calico ${KOPS_EXTRA_FLAGS:-}"
```
olemarkus: This wasn't intentional, I assume. We at least need to add this back so the scenario actually runs Karpenter.

olemarkus (changes requested):

tests/e2e/scenarios/upgrade-ab/run-test.sh
anthonyhaussman: I added it back in https://github.com/kubernetes/kops/pull/15144/commits/684809d452a8741c74d30c0a821b3fbb2d56f823.
Is it really necessary as we drop the necessity of the Karpenter flag in https://github.com/kubernetes/kops/pull/15144/commits/1d457c438e77dd356de15b8db70ca46154bf2e97?

olemarkus: Well. That is the flag that tells kOps to use karpenter-managed IGs instead of ASG-managed ones. When that one is dropped, kOps won't provision karpenter.

olemarkus:
Myeah, we can't just drop the feature flag like that. Because the upgrade test will start with an older kOps version where it's still gated. I am guessing the scenario test doesn't properly pass the feature flag variable, which is why it fails.

anthonyhaussman:
> Myeah, we can't just drop the feature flag like that. Because the upgrade test will start with an older kOps version where it's still gated. I am guessing the scenario test doesn't properly pass the feature flag variable, which is why it fails.

I have re-added the `KOPS_FEATURE_FLAGS` export too @olemarkus  : https://github.com/kubernetes/kops/pull/15144/commits/7f801f3133d31bc994aacabc90f17925375b7788

olemarkus:
Tests seems to be working now. The failure is a legit error.

anthonyhaussman:
> Tests seems to be working now. The failure is a legit error.

@olemarkus Does this PR okay to be approved?

olemarkus:
You still need to look at the tests and make them pass (in this case, it is a namespaced resource that doesn't have a namespace set).

You also need to squash your PR in the end.

anthonyhaussman:
@olemarkus I was able to recreate the problem and it seems that the old `mutatingwebhookconfiguration` and `validatingwebhookconfiguration.admissionregistration.k8s.io` Kind resources are not pruned before the application of the new Karpenter Manifest.

Especially if the deletion of `validation.webhook.provisioners.karpenter.sh` and `defaulting.webhook.provisioners.karpenter.sh`  are not done, it concludes to some conflict.

I was unable to retrieve prune logs info to see if those resources are deleted successfully.

Do you have an idea why those resources are not deleted? :confused:

olemarkus:
Ah. All resources must be applied before old ones are pruned. I suggest keeping the old name so I that it's patches instead.

olemarkus:
The service name used in the certificate doesn't need to change.

upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template
noony: Maybe we can switch to a new deployment named "karpenter" without controller or webhook in the name because now one pods handle both.

olemarkus:
The name of the webhook resources shouldn't affect anything. They are arbitrary.
The name of the service referred to in the webhooks need to match the name of the service though, and the service name must also be configured correctly on the controller (the KARPENTER_SERVICE env variable). As long as these match up, there shouldn't be any cert issues.

anthonyhaussman:
> The name of the webhook resources shouldn't affect anything. They are arbitrary. The name of the service referred to in the webhooks need to match the name of the service though, and the service name must also be configured correctly on the controller (the KARPENTER_SERVICE env variable). As long as these match up, there shouldn't be any cert issues.

It's what I have configured on my last commits:

https://github.com/kubernetes/kops/blob/b993c99f3abca4cadee5527625a0a2271eba96c9/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template#L897-L900

https://github.com/kubernetes/kops/blob/b993c99f3abca4cadee5527625a0a2271eba96c9/upup/models/cloudup/resources/addons/karpenter.sh/k8s-1.19.yaml.template#L949-L950

Still, I have still certificate issues:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/15144/pull-kops-e2e-aws-upgrade-k123-ko125-to-k124-kolatest-karpenter/1629922422935261184#1:build-log.txt%3A6013

anthonyhaussman:
bbb

> From what I can tell, the webhook has a hardcoded name from upstream, and when you change the name in the templates, you end up with a double set of webhook specs. The old one does not get any CA certificate injected and thus the validation fails.
> 
> I don't think there is any way of fixing this except for making the webhook name configurable in Karpenter. Probably want to file an upstream bug about this.

@olemarkus I have raised an issue to have configurable webhook names:
 - https://github.com/aws/karpenter-core/issues/226

olemarkus:
kOps should now try to apply, then prune, then try to apply again if prune is successful. This should take care of the renaming issues.

anthonyhaussman:
> kOps should now try to apply, then prune, then try to apply again if prune is successful. This should take care of the renaming issues.

Ho nice. I'm gonna rebase my PR when I'll have some time.

anthonyhaussman:
@olemarkus,

I have rebased/squashed and updated the template to be more coherent as the prune should run on all resources.
(Sorry for the label mess due to one of my bad rebase).

The `pull-kops-e2e-aws-upgrade-k123-ko125-to-k124-kolatest-karpenter` seems to pass now but I am a little worried about not retrieving the message `error applying update` at the first application of the template based on your [PR](https://github.com/kubernetes/kops/pull/15192).

Do you have an idea? :thinking:

olemarkus:
I wouldn't expect to see those failures in the log as the control plane should have had time to apply them. But looking at the logs, it seems like karpenter isn't running at all. So you may have done something wrong when rebasing.

anthonyhaussman:
The test  `pull-kops-e2e-aws-karpenter` launch Karpenter without problem :thinking: 
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/15144/pull-kops-e2e-aws-karpenter/1633373183472046080#

olemarkus:
You seem to have removed the commit that adds the `--instance-manager` flag.

anthonyhaussman:
@olemarkus I think I miss a commit bit I can't retrieve it  :disappointed:

olemarkus:
You have to dig through `git reflog` then.

anthonyhaussman:
My five last commit was cherry-picked from `git reflog` searching for commits concerning flags.
I didn't find more commits on this.
 - https://github.com/kubernetes/kops/pull/15144/commits/9948f4118e03a1ef0dbe016ddaa534bf6ce6e066
 - https://github.com/kubernetes/kops/pull/15144/commits/b765ee9bad90116f204803d2ed6de31f1f744fe2
 - https://github.com/kubernetes/kops/pull/15144/commits/c9586f6c9cae0f5e52ad7969e456eb984d5dfad9
 - https://github.com/kubernetes/kops/pull/15144/commits/6067902e6b5ddcaa53a215a013ae2353832a69fc
 - https://github.com/kubernetes/kops/pull/15144/commits/ca8bc94df0f3366fb84fe0cc42f2745f2c9f85bd

Weird behavior, last tests always failed on the same error:

```
The dns-controller Kubernetes deployment has not updated the Kubernetes cluster's API DNS entry to the correct IP address.  The API DNS IP address is the placeholder address that kops creates: 203.0.113.123.  Please wait about 5-10 minutes for a control plane node to start, dns-controller to launch, and DNS to propagate.  The protokube container and dns-controller deployment logs may contain more diagnostic information.  Etcd and the API DNS entries must be updated for a kops Kubernetes cluster to start.
```

anthonyhaussman:
@olemarkus,
Seems we have still a problem with the health check on the `provisioner`:
```
I0309 11:14:32.446734   36017 health.go:61] status conditions not found for Provisioner.karpenter.sh:nodes
```

I have done a manual test and it works on my side.
Two questions:
 - At the second apply after pruning, does the process retry the apply? It seems the `provisioner` need to be passed a second time, the first time return 
```
for: "./k8s-1.19.yaml": error when patching "./k8s-1.19.yaml": admission webhook "validation.webhook.karpenter.sh" denied the request: validation failed: expected exactly one, got both: spec.provider, spec.providerRef
```

 - Does the process check the status condition returned by the `describe` command? If it's the case, `provisioner` and `awsnodetemplates` resources do not return any.

olemarkus:
The health error log is just informational. You can ignore it.

The process now runs apply, prune, apply. In the e2e we sleep for 120 seconds, so the process should already have tried apply/prune 1-2 times before the e2e test itself runs its apply/prune. Ideally when e2e runs channes, there should be no changes to apply at all.

anthonyhaussman:
> The health error log is just informational. You can ignore it.
> 
> The process now runs apply, prune, apply. In the e2e we sleep for 120 seconds, so the process should already have tried apply/prune 1-2 times before the e2e test itself runs its apply/prune. Ideally when e2e runs channes, there should be no changes to apply at all.

Not sure why it does not work here :thinking: 
Shall we add the `provisioners.karpenter.sh` resource to be pruned into
https://github.com/kubernetes/kops/blob/8023192818cbbdd41f13ffd6fe3bd314047983a1/upup/pkg/fi/cloudup/bootstrapchannelbuilder/pruning.go#L63-L77
?

olemarkus:
You aren't renaming provisioners are you? Removing provisioner also terminates all instances associated with it, so it's super dangerous. For now, I think we should leave provisioners alone.

anthonyhaussman:
> You aren't renaming provisioners are you? Removing provisioner also terminates all instances associated with it, so it's super dangerous. For now, I think we should leave provisioners alone.

Nop I don't rename it.
Just change the way to call `provider` using `providerRef` via the `AWSNodeTemplate`.
Not sure what not working here.

anthonyhaussman:
@olemarkus,

I was navigating into the [protokube.log](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/15144/pull-kops-e2e-aws-upgrade-k123-ko125-to-k124-kolatest-karpenter/1635523560434831360/artifacts/ip-172-20-121-220.eu-central-1.compute.internal/protokube.log) and see:
```
updating "karpenter.sh": error pruning manifest from "s3://k8s-kops-prow/e2e-e0f2048229-326c9.test-cncf-aws.k8s.io/addons/karpenter.sh/k8s-1.19.yaml": failed to prune objects of kind RoleBinding.rbac.authorization.k8s.io: failed to delete kube-system/karpenter-controller: rolebindings.rbac.authorization.k8s.io "karpenter-controller" not found
W0314 06:32:47.270879    2978 kube_boot.go:89] error applying channel "s3://k8s-kops-prow/e2e-e0f2048229-326c9.test-cncf-aws.k8s.io/addons/bootstrap-channel.yaml": error running channels: exit status 1
```

And indeed, the `karpenter-controller`  rolebindings does not exist anymore after a first pass.
I think it's related to why the test does not pass.

olemarkus:
Thanks for keeping this one going. I hope to have time to look at this and other stuff this weekend.

anthonyhaussman:
> Thanks for keeping this one going. I hope to have time to look at this and other stuff this weekend.

Thanks :pray: 
I have some struggles finding out why it's not working.

For the `pull-kops-e2e-aws-karpenter` test, it seems not related to the Karpenter installation. I have tested locally the Karpenter template and I didn't have any issues.
For the `pull-kops-e2e-aws-upgrade-k123-ko125-to-k124-kolatest-karpenter`, it seems the issue still coming from the pruning not happening before the apply of the template.

If you need I am available to work on this.

hakman:
@anthonyhaussman Could you rebase this PR? Thanks!

hakman:
The only failing tests is now `[sig-network] Services should serve endpoints on same port and different protocols` which will be skipped once https://github.com/kubernetes/test-infra/pull/29616 merges.

hakman:
@olemarkus Do you have some time to take one more look?

olemarkus:
Looks like `--instance-manager` flag has been rebased out of the PR again.

hakman:
> Looks like `--instance-manager` flag has been rebased out of the PR again.

At least `pull-kops-e2e-aws-karpenter` is passing 😄.

tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+export KOPS_FEATURE_FLAGS="SpecOverrideFlag,"
```
hakman, line 17: I don't think I want this flag to get back in.

hakman (changes requested):

docs/operations/karpenter.md
```diff
@@ -2,16 +2,8 @@
...
-Karpenter is a fairly new project, and it is still not determined how Karpenter should work with kOps. Because of this, Karpenter is behind the `Karpenter` feature flag.
-
 ## Installing
 
-Enable the Karpenter feature flag:
-
-```sh
-export KOPS_FEATURE_FLAGS="Karpenter"
-```
-
```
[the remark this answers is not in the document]
anthonyhaussman: Done :slightly_smiling_face: 
 - https://github.com/kubernetes/kops/pull/15144/commits/3371d3c77964d6a5768abf6591c1230255e91d0a

docs/releases/1.27-NOTES.md
```diff
@@ -9,6 +9,8 @@ This is a document to gather the release notes prior to the release.
 * The default retention duration for the etcd backups is now set to 90 days.
 This behaviour can be overridden by setting `spec.etcdClusters[*].manager.backupRetentionDays` in the cluster spec.
 
+* `Karpenter` feature flag is no more necessary and upgraded to version `0.27.0`.
```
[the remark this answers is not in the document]
anthonyhaussman: Done during rebase :ok_hand:

tests/e2e/scenarios/upgrade-ab/run-test.sh
anthonyhaussman: I have removed it in https://github.com/kubernetes/kops/pull/15144/commits/5b95d3073aacbb80d14c3be7a9958cce2043cd1b

anthonyhaussman:
> Looks like `--instance-manager` flag has been rebased out of the PR again.

@olemarkus I had re-pushed all commits concerning extra flags after my past bad rebase into https://github.com/kubernetes/kops/pull/15144/commits/294f3438a92a776e1234ac78c45667b3e80c078c

I have rechecked my `reflog` and didn't find a commit about the `--intance-manager` flag. :confused: 
If I can have your help on this. :pray:

hakman:
@anthonyhaussman Could you also take care of the indentation diffs inside the template? I think you are using a different indentation than before.

hakman:
> @anthonyhaussman Could you also take care of the indentation diffs inside the template? I think you are using a different indentation than before. It should reduce the diff a little.

ltellesfl:
Any updates on this? Karpenter is already at version 0.28.1

olemarkus:
I think it's time to get this one in. @anthonyhaussman can you squash/clean up your commits and update the version in the PR title? Then we're good to go.

hakman:
@anthonyhaussman & @olemarkus Also https://github.com/kubernetes/kops/pull/15144#issuecomment-1584102984 please.

hakman (approved):

anthonyhaussman:
@olemarkus / @hakman Sorry for the late response. :pray: 
I was on vacation and was not able to have a look at the PR during that time.

Really sorry about that. Thanks for the merge. :pray:

hakman:
@anthonyhaussman No worries, just wanted to get it in 1.27. Great work!

BTW, if you get a chance, maybe you can take a look at https://github.com/kubernetes/kops/pull/15585.

PR #15145, opened 2023-02-12
remove AWS S3 access from nodes if using none dns

zetaab:
Normal nodes does not need S3 access after using none dns

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -465,9 +465,11 @@ func (r *NodeRoleNode) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 
 	b.addNodeupPermissions(p, r.enableLifecycleHookPermissions)
 
-	var err error
-	if p, err = b.AddS3Permissions(p); err != nil {
-		return nil, fmt.Errorf("failed to generate AWS IAM S3 access statements: %v", err)
+	if !b.Cluster.UsesNoneDNS() {
+		var err error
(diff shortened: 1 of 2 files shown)


hakman (approved):

zetaab:
/hold

I am running tests for this to ensure that its not needed

rifelpet:
what happens if a node created from an older kops version is rebooted after this change is applied? will the old nodeup still try (and fail) to fetch files from the state store?

zetaab:
@rifelpet afaik dns none is quite new thing and it has not been any stable release yet(?). Also it has not required s3 access since its lifetime, so it should be quite safe move?

zetaab:
yeah I can confirm that at least new cluster with dns none starts after this. 

/hold cancel

rifelpet:
Do we (or will we) support migrating existing clusters from `gossip` to `none` ? If so my concern still stands

hakman:
> Do we (or will we) support migrating existing clusters from `gossip` to `none` ? If so my concern still stands

We will support migrating from `gossip` to `none`, but it will not be without downtime and full cluster roll.
Also, I doubt that anyone will restart nodes during this process. Probably there should be a note about it.
This seems more like something that was missed during https://github.com/kubernetes/kops/pull/14501.

avdhoot:
@zetaab This change caused an [issue](https://github.com/kubernetes/kops/issues/15287) . The image pull is no longer working.  @rifelpet your [concern](https://github.com/kubernetes/kops/issues/15287) is valid.  cc @hakman

PR #15165, opened 2023-02-16
WIP Add Ability to Configure Custom PriorityClasses to Be Used in Cluster Validation

### LINKED ISSUE #13150: Add Ability to Configure Custom PriorityClasses to Be Used in Cluster Validation
/kind feature

Currently `kops validate cluster` checks pods with the priority classes `system-cluster-critical` and `system-node-critical`. It would be great for users to be able to specify additional priority classes for kops to validate associated pods.

ederst:
Currently, kOps  only validates pods with the priority classes `system-cluster-critical` and `system-node-critical`.

This will enable the users to specify their own priority classes kOps should consider when validating a cluster.

TODO:

- [ ] Revise naming, doc
- [ ] Add tests

TL;DRcode:

It is possible to specify `--additional-filters <key>=<value>`, multiple times with it. Where key can be `priority-class-name` and `namespace`. The latter is not implemented yet.

fixes #13150

### DIFF
--- a/cmd/kops/delete_instance.go
+++ b/cmd/kops/delete_instance.go
@@ -71,9 +71,12 @@ type DeleteInstanceOptions struct {
 
 	Surge bool
 
-	// AdditionalPriorities specifies that the validation will consider pods with these PriorityClassNames as well,
-	// in addition to the default system critical ones (todo: improve wording...).
-	AdditionalPriorities []string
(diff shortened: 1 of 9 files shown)


ederst:
as discussed in the office hours/bonus meeting: i'll close this due to not [having a use case which cannot be handled by PDBs](https://github.com/kubernetes/kops/issues/13150#issuecomment-1468068252)

PR #15169, opened 2023-02-20
update openstack csi & ccm versions

zetaab:


hakman:
I don't think there's any 1.24 release planned anymore.

hakman (approved):

PR #15172, opened 2023-02-20
cherry pick of #15169: update openstack csi & ccm versions

zetaab:


zetaab:
perhaps we can close this

PR #15173, opened 2023-02-21
[Docs] Fix typo

jare:


justinsb:
Thanks for the contribution - and fixing typos is a great way to get started!

Are you able to sign the CLA?  I'm afraid we can't take PRs without it, but the good news is that the CLA then applies all the kubernetes projects (maybe some CNCF ones also!)

hakman:
Thanks @jare, but without the CLA we cannot accept this contribution.

PR #7837, opened 2019-10-28
Add External Policies (AWS managed policy attachments)

mattouille:
This feature allows Managed Policies to be attached to Instance Group Roles. It will manage the full lifecycle of _all_ Managed Policies attached to `master`, `node`, and `bastion` by first adding new managed policies followed by removing unspecified policies (policies not present in the config) in order to avoid service disruption.

PR #15177, opened 2023-02-22
[docs] Fix External IAM Policy Docs

philnielsen:
This feature was renamed during its [development](https://github.com/kubernetes/kops/pull/7837) and a remnant of that original name was in the docs. Fixing this because it confused me enough to have to look up what the correct config was in the code.


rifelpet (approved):
/ok-to-test

PR #15181, opened 2023-02-27
[DigitalOcean] support terraform provider for DO

srikiz:
Initial work to get the terraform support for DO.

- Tested by creating a cluster with `--out terraform` like below.

```
kops create cluster --cloud=digitalocean --name=dev6.****.co.in --node-count=2 --networking=calico --zones=sfo3 --ssh-public-key=~/.ssh/id_ed25519.pub --target=terraform --out=./terraform-test -v=6
```

- kubernetes.tf file gets created with digitial ocean specific resources. snippet below.
```

output "cluster_name" {
  value = "dev6.test.co.in"
}

output "region" {
  value = "sfo3"
}

provider "digitalocean" {
}

resource "digitalocean_droplet" "control-plane-sfo3-1-masters-dev6-test-co-in" {
  image     = "ubuntu-20-04-x64"
  name      = "control-plane-sfo3-1.masters.dev6.test.co.in"
  region    = "sfo3"
  size      = "s-2vcpu-4gb"
  ssh_keys  = ["c5:e2:db:15:b1:36:5f:f4:21:13:1a:59:cf:a4:39:05"]
  tags      = ["KubernetesCluster:dev6-test-co-in", "k8s-index:etcd-1", "KubernetesCluster-Master:dev6-test-co-in", "kops-instancegroup:control-plane-sfo3-1"]
  user_data = file("${path.module}/data/digitalocean_droplet_control-plane-sfo3-1.masters.dev6.test.co.in_user_data")
}

resource "digitalocean_droplet" "nodes-sfo3-dev6-test-co-in" {
  image     = "ubuntu-20-04-x64"
  name      = "nodes-sfo3.dev6.test.co.in"
  region    = "sfo3"
  size      = "s-2vcpu-4gb"
  ssh_keys  = ["c5:e2:db:15:b1:36:5f:f4:21:13:1a:59:cf:a4:39:05"]
  tags      = ["KubernetesCluster:dev6-test-co-in", "kops-instancegroup:nodes-sfo3"]
  user_data = file("${path.module}/data/digitalocean_droplet_nodes-sfo3.dev6.test.co.in_user_data")
}

resource "digitalocean_spaces_bucket_object" "cluster-completed-spec" {
  bucket  = "sri-kops-space"
  content = file("${path.module}/data/digitalocean_spaces_bucket_object_cluster-completed.spec_content")
  key     = "dev6.test.co.in/cluster-completed.spec"
  region  = "sfo2"
}

```

- terraform init/plan/apply works as expected.
- able to see the cluster created as expected.

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -168,6 +168,9 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
 		if !found {
 			return fmt.Errorf("cloud provider %v does not support the terraform target", c.Cloud.ProviderID())
 		}
+		if c.Cloud.ProviderID() == kops.CloudProviderDO && !featureflag.DOTerraform.Enabled() {
+			return fmt.Errorf("DO Terraform requires the DOTerraform feature flag to be enabled")
+		}
 	}
 	if c.InstanceGroups == nil {
 		list, err := c.Clientset.InstanceGroupsFor(c.Cluster).List(ctx, metav1.ListOptions{})
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -92,6 +92,7 @@ var TerraformCloudProviders = []kops.CloudProviderID{
 	kops.CloudProviderGCE,
 	kops.CloudProviderHetzner,
 	kops.CloudProviderScaleway,
+	kops.CloudProviderDO,
 }
 
 type ApplyClusterCmd struct {
--- a/upup/pkg/fi/cloudup/dotasks/droplet.go
+++ b/upup/pkg/fi/cloudup/dotasks/droplet.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/kops/upup/pkg/fi/cloudup/do"
 	"k8s.io/kops/upup/pkg/fi/cloudup/terraform"
 	_ "k8s.io/kops/upup/pkg/fi/cloudup/terraform"
+	"k8s.io/kops/upup/pkg/fi/cloudup/terraformWriter"
 )
 
 // Droplet represents a group of droplets. In the future it
@@ -222,14 +223,14 @@ func (_ *Droplet) CheckChanges(a, e, changes *Droplet) error {
 }
 
 type terraformDropletOptions struct {
-	Image    *string  `cty:"image"`
-	Size     *string  `cty:"size"`
-	Region   *string  `cty:"region"`
-	Name     *string  `cty:"name"`
-	Tags     []string `cty:"tags"`
-	SSHKey   []string `cty:"ssh_keys"`
-	UserData *string  `cty:"user_data"`
-	VPCUUID  *string  `cty:"vpc_uuid"`
+	Image    *string                  `cty:"image"`
+	Size     *string                  `cty:"size"`
+	Region   *string                  `cty:"region"`
+	Name     *string                  `cty:"name"`
+	Tags     []string                 `cty:"tags"`
+	SSHKey   []string                 `cty:"ssh_keys"`
+	UserData *terraformWriter.Literal `cty:"user_data"`
+	VPCUUID  *string                  `cty:"vpc_uuid"`
 }
 
 func (_ *Droplet) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *Droplet) error {
@@ -242,11 +243,18 @@ func (_ *Droplet) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *D
 		SSHKey: []string{fi.ValueOf(e.SSHKey)},
 	}
 
-	userData, err := fi.ResourceAsString(e.UserData)
-	if err != nil {
-		return err
+	if e.UserData != nil {
+		d, err := fi.ResourceAsBytes(e.UserData)
+		if err != nil {
+			return err
+		}
+		if d != nil {
+			tf.UserData, err = t.AddFileBytes("digitalocean_droplet", fi.ValueOf(e.Name), "user_data", d, false)
+			if err != nil {
+				return err
+			}
+		}
 	}
-	tf.UserData = &userData
 
 	return t.RenderResource("digitalocean_droplet", *e.Name, tf)
 }
--- a/upup/pkg/fi/cloudup/dotasks/droplet.go
+++ b/upup/pkg/fi/cloudup/dotasks/droplet.go
@@ -25,6 +25,7 @@ import (
 
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/do"
+	"k8s.io/kops/upup/pkg/fi/cloudup/terraform"
 	_ "k8s.io/kops/upup/pkg/fi/cloudup/terraform"
 )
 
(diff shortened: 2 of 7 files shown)


upup/pkg/fi/cloudup/terraform/target_hcl2.go
```diff
@@ -122,7 +122,7 @@ func (t *TerraformTarget) writeProviders(buf *bytes.Buffer) {
 	if t.Cloud.ProviderID() == kops.CloudProviderGCE {
 		providerBody["project"] = t.Project
 	}
-	if t.Cloud.ProviderID() != kops.CloudProviderHetzner {
+	if t.Cloud.ProviderID() != kops.CloudProviderHetzner && t.Cloud.ProviderID() != kops.CloudProviderDO {
```
timoreimann, line 125: I understand that we'd set the region here unless the cloud provider is Hetzner or DO. To confirm, is that the desired behavior? DO does support regions, right?

upup/pkg/fi/cloudup/dotasks/droplet.go
```diff
@@ -210,3 +212,40 @@ func (_ *Droplet) CheckChanges(a, e, changes *Droplet) error {
...
+		Region: e.Region,
+		Name:   e.Name,
+		Tags:   e.Tags,
+		SSHKey: []string{fi.ValueOf(e.SSHKey)},
+	}
+
+	if e.UserData != nil {
+		d, err := fi.ResourceAsBytes(e.UserData)
+		if err != nil {
+			return err
```
timoreimann, line 240: Should we wrap this error and the one from line 245 in a `fmt.Errorf` in order to add context distinguishing the two?

util/pkg/vfs/s3fs.go
```diff
@@ -587,40 +594,66 @@ func (p *S3Path) RenderTerraform(w *terraformWriter.TerraformWriter, name string
...
+		}
 
-	content, err := w.AddFileBytes("aws_s3_object", name, "content", bytes, false)
-	if err != nil {
-		return fmt.Errorf("rendering S3 file: %v", err)
-	}
+		// retrieve space region from endpoint
+		endpoint := os.Getenv("S3_ENDPOINT")
+		if endpoint == "" {
+			return fmt.Errorf("S3 Endpoint is empty")
```
timoreimann, line 608: `errors.New` is sufficient here.

util/pkg/vfs/s3fs.go
```diff
@@ -587,40 +594,66 @@ func (p *S3Path) RenderTerraform(w *terraformWriter.TerraformWriter, name string
...
+		}
+		region := strings.Split(endpoint, ".")[0]
```
timoreimann, line 610: Do we always expect a dot separator? If so, should we return an error if it's missing?

timoreimann (commented):
Did mostly a mechanic review since I'm not too familiar with Terraform and kops' usage. Someone with more expertise on that front should likely review as well.

upup/pkg/fi/cloudup/terraform/target_hcl2.go
srikiz: Yes that's correct. Per DO Terraform provider [reference here](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs#argument-reference), it doesn't have the region attribute specified.
Though, we fetch the spaces region from the `S3_ENDPOINT` environment variable. And the droplet and volume region is associated to the terraform resource based on the `region` attribute as [referenced here.](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/droplet#region)

upup/pkg/fi/cloudup/dotasks/droplet.go
srikiz: sure, updated.

util/pkg/vfs/s3fs.go
srikiz: Sure, modified it.

util/pkg/vfs/s3fs.go
srikiz: Per DO documentation, it is always in the format `{region}.digitaloceanspaces.com` and if user attempts to specify an incorrect endpoint, the region will contain the entire string from `S3_ENDPOINT`. I think that shouldn't be a problem since eventually the terraform resource will not be created. So I am keeping it as-is unless you think we need that check.

upup/pkg/fi/cloudup/dotasks/droplet.go
```diff
@@ -210,3 +212,40 @@ func (_ *Droplet) CheckChanges(a, e, changes *Droplet) error {
...
+}
+
+func (_ *Droplet) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *Droplet) error {
+	tf := &terraformDropletOptions{
+		Image:  e.Image,
+		Size:   e.Size,
+		Region: e.Region,
+		Name:   e.Name,
+		Tags:   e.Tags,
+		SSHKey: []string{fi.ValueOf(e.SSHKey)},
```
rifelpet, line 234: `VPCUUID` should be set here

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -91,6 +91,7 @@ var TerraformCloudProviders = []kops.CloudProviderID{
 	kops.CloudProviderAWS,
 	kops.CloudProviderGCE,
 	kops.CloudProviderHetzner,
+	kops.CloudProviderDO,
```
rifelpet, line 94: I'm wondering if we should guard the terraform support behind a feature flag until it is fully implemented. I think LoadBalancer and VPC still need support?

util/pkg/vfs/s3fs.go
```diff
@@ -579,6 +580,13 @@ type terraformS3File struct {
 	Provider *terraformWriter.Literal `json:"provider,omitempty" cty:"provider"`
 }
 
+type terraformDOFile struct {
```
rifelpet, line 583: Do you happen to know why DO reuses S3Path rather than implement its own DOPath?

This is where it is reused: https://github.com/kubernetes/kops/blob/e63cb44bd06e4b5d079f53f164d526035d29f84e/util/pkg/vfs/context.go#L348-L364

srikiz: I think it's because we could access DO Spaces using S3 complaint API and that was already in-place and we started using that. We didn't go back and change that to use DO API's instead. May be take this separately?

rifelpet: yea a separate PR would be fine 👍🏻

srikiz:
@rifelpet @timoreimann - please re-review whenever you get a chance. Thanks !

upup/pkg/fi/cloudup/dotasks/droplet.go
```diff
@@ -219,3 +221,40 @@ func (_ *Droplet) CheckChanges(a, e, changes *Droplet) error {
...
+
+type terraformDropletOptions struct {
+	Image    *string                  `cty:"image"`
+	Size     *string                  `cty:"size"`
+	Region   *string                  `cty:"region"`
+	Name     *string                  `cty:"name"`
+	Tags     []string                 `cty:"tags"`
+	SSHKey   []string                 `cty:"ssh_keys"`
+	UserData *terraformWriter.Literal `cty:"user_data"`
+	VPCUUID  *string                  `cty:"vpc_uuid"`
```
rifelpet, line 233: This field is never being set in RenderTerraform

upup/pkg/fi/cloudup/dotasks/droplet.go
```diff
@@ -219,3 +221,40 @@ func (_ *Droplet) CheckChanges(a, e, changes *Droplet) error {
...
+}
+
+func (_ *Droplet) RenderTerraform(t *terraform.TerraformTarget, a, e, changes *Droplet) error {
+	tf := &terraformDropletOptions{
+		Image:  e.Image,
+		Size:   e.Size,
+		Region: e.Region,
+		Name:   e.Name,
+		Tags:   e.Tags,
+		SSHKey: []string{fi.ValueOf(e.SSHKey.KeyFingerprint)},
```
rifelpet, line 243: Based on the model code below, `e.SSHKey` can be nil so we should conditionally set this field to avoid panics

https://github.com/kubernetes/kops/blob/140591f72336bb29b6933ed30f9249121e36de9c/pkg/model/domodel/droplets.go#L40-L41

upup/pkg/fi/cloudup/dotasks/droplet.go
srikiz: sure, updated. Most of my testing, that was set to empty and was not being used. I have a note to test the vpc and Load balancer feature separately and more PR's to follow.

rifelpet:
Once we have sufficient resource coverage we should add terraform instructions to the docs, mentioning the feature flag:

https://kops.sigs.k8s.io/getting_started/digitalocean/

PR #15183, opened 2023-02-28
feat(NodeLocalDNS): Add possibility to set an ExternalCoreFile

anthonyhaussman:
Allow users to provide entirely custom CoreFile for NodeLocalDNS to provide improved flexibility.


hakman:
Quick observation, `Corefile` instead of `CoreFile`.
/ok-to-test

anthonyhaussman:
> Quick observation, `Corefile` instead of `CoreFile`. /ok-to-test

@hakman I wanted to keep the same "vocabulary" as the CoreDNS app which is `CoreFile`

hakman (approved):
Thanks @anthonyhaussman.
LGTM

PR #15186, opened 2023-03-01
gce: Don't enable all controllers with the CCM.

justinsb:
If running with GCE "native" networking, we do not need the route
controller (and it causes problems).

In general we do not need the gkenetworkparams controller (and it
complains about missing CRDs).


justinsb:
See discussion here: https://github.com/kubernetes/cloud-provider-gcp/pull/482#issuecomment-1450203175

Obviously we need a better way to differentiate the parameters that are cluster-specific (the controllers) vs component-specific (these options), and it's further complicated because these options actually vary by networking mode...  But for now we can unblock testing with this.

justinsb:
This is the test we really need to pass

Some of the flakes look like a conflict between ko and distroless.dev:
```
 GET https://distroless.dev/v2/static/manifests/latest: MANIFEST_UNKNOWN: OCI index found, but Accept header does not support OCI indexes
 ```

**Comment by justinsb** [comment]
```
"gkenetworkparams" is not in the list of known controllers
Error: "gkenetworkparams" is not in the list of known controllers
```

Ouch - so we can't exclude an unknown controller ... we should probably fix that!

File: pkg/model/components/gcpcloudcontrollermanager.go [hunk 1]
```diff
@@ -57,6 +57,21 @@ func (b *GCPCloudControllerManagerOptionsBuilder) BuildOptions(options interface
 	if ccmConfig.ClusterCIDR == "" {
 		ccmConfig.ClusterCIDR = clusterSpec.Networking.PodCIDR
 	}
+
+	if ccmConfig.Controllers == nil {
+		if clusterSpec.Networking.GCE != nil {
+			ccmConfig.Controllers = []string{
+				"*",
+				// Don't run route controller, not needed here with "native" pod IPs
+				"-route",
```
aojea, line 66: but if the users want to run in route mode (is that possible)? you need to enable it

jprzychodzen: > ```
> "gkenetworkparams" is not in the list of known controllers
> Error: "gkenetworkparams" is not in the list of known controllers
> ```
> 
> Ouch - so we can't exclude an unknown controller ... we should probably fix that!
> 
> Edit: Opened [kubernetes/kubernetes#116186](https://github.com/kubernetes/kubernetes/issues/116186) (and sent a fix)

It seems I've made a typo: it should be `gkenetworkparamset`

aojea: you should only skip in native, but I think that is safer to depend on the flag `--configure-cloud-routes`

https://github.com/kubernetes/cloud-provider-gcp/blob/427c014c890a013cec192d783e00d1cb576e5662/vendor/k8s.io/cloud-provider/app/core.go#L102-L106

justinsb: Ah - thanks for letting me know re gkenetworkparamset.  I guess that supports the idea that it should be an error, because there are two possibilities for an incorrect controller name: (1) it isn't compiled into this version, or (2) typo.  If we make "not-compiled-in" more user friendly, we make "typo" worse.

I summarized what I interpret the flags should be [here](https://github.com/kubernetes/cloud-provider-gcp/pull/482#issuecomment-1450478516), and we should be testing that right now :-)

justinsb:
lol ...

```
"gkenetworkparamset" is not in the list of known controllers
Error: "gkenetworkparamset" is not in the list of known controllers
```

So both a typo and a version issue.  The image we're using here is `image: k8scloudprovidergcp/cloud-controller-manager:latest` (not sure why we're not using a tagged image...)

justinsb:
Cherry-picked in #15040 into this PR also, as I think we need both to reliably pass e2e

justinsb:
> Services should be able to update service type to NodePort listening on same port number but different protocols 

I'm guessing we also need a NonMasqueradeCIDR....

justinsb:
Looks like we have the magic combo:

* Don't run configure-routes in CCM
* Open the apiserver to the pod CIDR (because we lock it down pretty heavily in e2e)
* Use a NonMasqueradeCIDR that covers the local network (at least) needed for NodePorts

Let's double check:

/test pull-kops-e2e-k8s-gce-cilium

upup/pkg/fi/cloudup/gce/network.go
```diff
@@ -183,6 +183,16 @@ func performNetworkAssignmentsIPAliases(ctx context.Context, c *kops.Cluster, cl
...
 
+	// NonMasqueradeCIDR should include all the pods and any hosts which can route to the pod IP.
+	// Here, that is any IP on the network.
+	// Networks on GCE don't have a well-defined CIDR (instead, subnets do).
+	// We use networkCIDR instead; these IPs are routable on the GCE network.
+	// Technically this means that the service CIDR would be subject to masquerade,
+	// but that traffic is already remapped before it reaches the masquerade rule.
+	// Ideally we would support all the ip ranges in use (all the private IP ranges),
+	// but as long as we cover the pod CIDR we should be OK.
+	c.Spec.Networking.NonMasqueradeCIDR = networkCIDR
```
aojea, line 194: you mean podCIDR here, no?

aojea: I see, networkCIDR  contains all

justinsb:
Other PRs have merged, rebase and tidied up the commit message here.

justinsb:
Hooray!  This PR is now passing, the other PRs that were needed have also merged so this PR is "clean", and the logs for CCM look good.  Thanks for helping us figure out the options @aojea and @jprzychodzen

hakman (approved):

aojea:
thanks for supporting it :)

PR #15196, opened 2023-03-07
fix create cluster doc, using 'master' instead of 'control-plane'

plucury:


hakman:
The recommended terminology is `control-plane`. Please read more from:
https://github.com/kubernetes/community/blob/5a324d2198e6530b12261917387e5dbdff86d12d/sig-architecture/naming/recommendations/001-master-control-plane.md

plucury:
> https://github.com/kubernetes/community/blob/5a324d2198e6530b12261917387e5dbdff86d12d/sig-architecture/naming/recommendations/001-master-control-plane.md

OK. So we need rename parameters in CLI?

hakman:
You can still use the old parameter names for now, but it would be recommended to rename.

PR #15208, opened 2023-03-09
Bump channels

hakman:
/cc @olemarkus @rifelpet @zetaab


moshevayner (approved):
/lgtm

zetaab (approved):
We are already using 1.26.2 and same ubuntu image and can confirm everything works like should

/lgtm
/approve

PR #15221, opened 2023-03-10
OpenStack: Allow setting allowed address pairs for ports

ederst:
This enables setting allowed address pairs for ports via IG annotations, which can be beneficial when hosting workloads on the nodes which need to work around the spoofing protection of OpenStack (keepalived, Calico without IPIP or other encapsulation, etc.). 

It works by adding multiple annotations to an IG config, like so:

```yaml
metadata:
  annotations:
    openstack.kops.io/allowedAddressPair/N: <IP>(,<MAC>)
```

Where `N` can be an arbitrary identifier, and the MAC address is optional.

### DIFF
--- a/upup/pkg/fi/cloudup/openstacktasks/port.go
+++ b/upup/pkg/fi/cloudup/openstacktasks/port.go
@@ -40,6 +40,7 @@ type Port struct {
 	Lifecycle                fi.Lifecycle
 	Tags                     []string
 	ForAPIServer             bool
+	AllowedAddressPairs      []ports.AddressPair
 }
 
 // GetDependencies returns the dependencies of the Port task
@@ -83,6 +84,29 @@ func (s *Port) FindAddresses(context *fi.CloudupContext) ([]string, error) {
 
 func (s *Port) IsForAPIServer() bool {
 	return s.ForAPIServer
+}
+
+// getActualAllowedAddressPairs returns the actual allowed address pairs which kOps currently manages.
+func getActualAllowedAddressPairs(port *ports.Port, find *Port) []ports.AddressPair {
+	if find == nil {
+		return port.AllowedAddressPairs
+	}
+
+	var allowedAddressPairs []ports.AddressPair
+	for _, portAddressPair := range port.AllowedAddressPairs {
+		// TODO: what if user set the macaddress in the config to the same one as the port?
+		if portAddressPair.MACAddress == port.MACAddress {
+			portAddressPair.MACAddress = ""
+		}
+
+		allowedAddressPairs = append(allowedAddressPairs, portAddressPair)
+	}
+
+	sort.Slice(allowedAddressPairs, func(i, j int) bool {
+		return allowedAddressPairs[i].IPAddress < allowedAddressPairs[j].IPAddress
+	})
+
+	return allowedAddressPairs
 }
 
 func newPortTaskFromCloud(cloud openstack.OpenstackCloud, lifecycle fi.Lifecycle, port *ports.Port, find *Port) (*Port, error) {
@@ -150,14 +174,15 @@ func newPortTaskFromCloud(cloud openstack.OpenstackCloud, lifecycle fi.Lifecycle
 	}
 
 	actual := &Port{
-		ID:                fi.PtrTo(port.ID),
-		InstanceGroupName: cloudInstanceGroupName,
-		Name:              fi.PtrTo(port.Name),
-		Network:           &Network{ID: fi.PtrTo(port.NetworkID)},
-		SecurityGroups:    sgs,
-		Subnets:           subnets,
-		Lifecycle:         lifecycle,
-		Tags:              tags,
+		ID:                  fi.PtrTo(port.ID),
+		InstanceGroupName:   cloudInstanceGroupName,
+		Name:                fi.PtrTo(port.Name),
+		Network:             &Network{ID: fi.PtrTo(port.NetworkID)},
+		SecurityGroups:      sgs,
+		Subnets:             subnets,
+		Lifecycle:           lifecycle,
+		Tags:                tags,
+		AllowedAddressPairs: getActualAllowedAddressPairs(port, find),
 	}
 	if find != nil {
 		find.ID = actual.ID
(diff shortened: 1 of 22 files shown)


zetaab:
@ederst are you planning to implement own loadbalancer instead of octavia? Just curious because we just tried that and it was working fine and was faster from performance perspective than default octavia

ederst:
> @ederst are you planning to implement own loadbalancer instead of octavia? Just curious because we just tried that and it was working fine and was faster from performance perspective than default octavia

We kind of implemented our own loadbalancing since day one of using kOps, because we had some really bad experiences with LBaasv2 and then with octavia too in our OpenStack; not so much performance related but the lack of stability put us off.

Currently, we use a relatively simple setup with DNS RR and keepalived which runs on the control plane and ingress nodes. We do not have the FIPs directly attached to the nodes, but to ports which act as VIPs, and when every node is up and running the VIPs are spread out equally (every node has one VIP due to having highest prio on one of the VIPs). The basics are kind of described [here](https://docs.openshift.com/container-platform/4.8/networking/configuring-ipfailover.html) and [there](https://blog.codecentric.de/highly-available-vips-openstack-vms-vrrp).

Not 100% perfect like real loadbalancing but does the trick ATM.

zetaab:
@ederst funny, we had also openshift 3.11 before kOps and we had quite similar setup. Ingress machines with vrrp VIP (+keepalived). 

I actually created similar "demo setup" with keepalived (running in k8s) and haproxy (running in k8s) to kOps OpenStack. Current kOps is supporting that quite well, only thing that was needed to modify manually was these allowed address pairs in ports. So I am supporting this change because it means that people can basically implement own (better) loadbalancing solutions after this is in place.

ederst:
> @ederst funny, we had also openshift 3.11 before kOps and we had quite similar setup. Ingress machines (and controlplane) with vrrp VIP (+keepalived).
> 
> I actually created similar "demo setup" with keepalived (running in k8s) and haproxy (running in k8s) to kOps OpenStack.

Funny indeed, because at first we wanted to do the keepalived + haproxy/some other suitable LB setup in K8s too. But we then just stayed with the setup as it is right now, because it works well enough and we figured the added complexity, work, etc. for the fully-fledged, self-managed LB solution would not be worth it.

>Current kOps is supporting that quite well, only thing that was needed to modify manually was these allowed address pairs in ports. So I am supporting this change because it means that people can basically implement own (better) loadbalancing solutions after this is in place.

Yep, deactivating the managed LB in kOps was a breeze and Pulumi, where we wrapped the kOps binary in a dynamic provider and provision some of the infrastructure and modify the ports to include the allowed address pairs, did the rest for us.

> The correct solution would be to use OVN network (then the lb solution will scale) and octavia with that, but it basically means reinstall for whole network layer because at least we have OVS currently. (if someone is reading this who have OVN networking and using kOps, feel free to confirm does kOps work with OVN. I have no idea, since I do not have working environment for that)

Well, for OpenStack we have a separate supplier so we are just users there. But we definitely have some older SUSE OpenStack clouds and the newer ones are RedHat AFAIK. If I am not completely wrong the newer ones do use OVN.

We actually never tested the performance due to esperiencing stability issues as the octavia LBs randomly went into "ERROR" state when creating or updating them and there was no visible error message attached to it why it happened. Then we always had to ask the support what went wrong, which was tedious.

Thus we kept the "ipfailover" setup in place and opted to only roll out some test environments with octavia (and on the older clouds with lbaasv2), just to see if the general rollout still works. But I think we roll out those LBs with Pulumi as well and not with kOps.

Anyhow, glad you support the change. Will see that I can do a final version of it ASAP.

ederst:
@zetaab for your consideration

ederst:
Back to WIP because:
* when allowed address pairs are already set, change is not detected accordingly:

is:
```
Will modify resources:
  Port/port-worker-1-my-cluster-k8s-local
  	AllowedAddressPairs 	 <nil> -> [{"ip_address":"10.135.0.0/16"}, {"ip_address":"192.168.0.0/16"}]
```

but should be:
```
Will modify resources:
  Port/port-worker-1-my-cluster-k8s-local
  	AllowedAddressPairs 	 [{"ip_address":"10.135.0.0/8"}, {"ip_address":"192.168.0.0/8"}] -> [{"ip_address":"10.135.0.0/16"}, {"ip_address":"192.168.0.0/16"}]
```

* when removing the annotations, address pairs will not get deleted

ederst:
@zetaab the current version of the code now should take care of managing the allowed address pairs for the ports, for example:

```log
Will modify resources:
  Port/port-control-plane-0-1-my-cluster-k8s-local
  	AllowedAddressPairs 	 [{"ip_address":"10.135.0.0/16"}, {"ip_address":"192.168.0.0/16"}, {"ip_address":"192.168.0.150"}, {"ip_address":"192.168.1.34"}, {"ip_address":"192.168.1.51"}] -> <nil>

   Port/port-worker-1-my-cluster-k8s-local
  	AllowedAddressPairs 	 [{"ip_address":"10.135.0.0/8"}, {"ip_address":"192.168.0.0/8"}] -> [{"ip_address":"10.135.0.0/16"}]
```

This means that also existing allowed address pairs would be deleted if not properly configured in IG manifest, which might not be what we want?

One possibility to mitigate this would be to explicitly configure that kOps should manage the allowed address pairs as well, [like so](https://github.com/kubernetes/kops/pull/15221#discussion_r1146041249), to not override any existing configuration, if not wanted. Don't know right now how to handle it [here](https://github.com/kubernetes/kops/pull/15221/files#diff-1b24f08f9ecc517253dfb510cc5b190ae17eae79f61548a49eef5a7a5a002722R185) as there is no IG config available afaik to check (maybe differentiatie between nil and empty slice?, sound a bit hacky).

Anyhow, I think that a mixture of letting kOps set allowed address pairs and letting the user also do that is hard to do. At least I found no correct way of doing it (e.g. when deleting an allowed address pair from the IG config, how should kOps know that this was an address pair it added).

upup/pkg/fi/cloudup/openstacktasks/port.go
```diff
@@ -85,6 +86,29 @@ func (s *Port) IsForAPIServer() bool {
 	return s.ForAPIServer
 }
 
+// getActualAllowedAddressPairs blubbs
```
ederst, line 89: whoops, need to improve the doc string here

zetaab (approved):
/approve
/lgtm

IMO configuring allowed address pairs by both user and kops is quite difficult to maintain. Instead if kops has mechanism to configure these, users should use kops to do it.

PR #15225, opened 2023-03-13
Spotinst: Setting the VNG Size Limits in Launch Spec

IdanShohamNetApp:
This PR adds setting support for VNG size limits in launch spec.


hakman:
Thanks for contributing this change @IdanShohamNetApp.
Could you squash the commits and rebase? https://github.com/kubernetes/kops/pull/15225#issuecomment-1466242436

PR #15230, opened 2023-03-14
scaleway: fix client for protokube

Mia-Cross:
In #15122 I removed the default region from the client options when creating a Scaleway client for protokube (as I thought it was no use since protokube only makes requests to zoned APIs). 
But it turns out I have to rollback this change because now I get this error, which I haven't encountered earlier :
`Error initializing Scaleway: "error creating client: scaleway-sdk-go: default region cannot be empty"`


hakman (approved):
/ok-to-test

PR #15231, opened 2023-03-14
Initial support for Debian 12 Bookworm

ReillyBrogan:
Initial support for the upcoming Debian 12

This currently requires a custom image to be built. That image will need to have the debian package base-files 12.4 installed which is not yet available in testing (but is available in unstable and is expected to be migrated soon).

### DIFF
--- a/docs/operations/images.md
+++ b/docs/operations/images.md
@@ -40,6 +40,7 @@ The following table provides the support status for various distros with regards
 | Debian 9                            |          1.8 |   1.10 |       1.21 |    1.23 |
 | [Debian 10](#debian-10-buster)      |         1.13 |   1.17 |          - |       - |
 | [Debian 11](#debian-11-bullseye)    |       1.21.1 |      - |          - |       - |
+| [Debian 12](#debian-12-bookworm)    |       1.26.3 |      - |          - |       - |
 | [Flatcar](#flatcar)                 |       1.15.1 |   1.17 |          - |       - |
 | Kope.io                             |            - |      - |       1.18 |    1.23 |
 | RHEL 7                              |            - |    1.5 |       1.21 |    1.23 |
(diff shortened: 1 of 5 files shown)


docs/operations/images.md
```diff
@@ -123,6 +123,12 @@ az vm image list --all --output table \
   --publisher Debian --offer debian-11 --sku 11-gen2
 ```
 
+### Debian 12 (Bookworm)
+
+Debian 12 is based on Kernel version **6.1** which has no known major Kernel bugs and fully supports all Cilium features.
+
+
```
hakman, line 130: Please add a link to this in the Matrix.
```suggestion
At the moment there is no official image published.
```

hakman (changes requested):

docs/operations/images.md
ReillyBrogan: What version should I put as the starting version? `1.27`? `1.26.3`?

hakman: I would go with 1.26.3.

ReillyBrogan:
@hakman I was actually going to mention you anyway. I see that there's a check for Debian 11 in `sysctls.go` regarding a Router Advertisement sysctl. I read through all of the PRs/issue I found on it and wasn't able to figure out the exact reason for this distro-specific workaround. 

My best guess is that this is because Debian 11 does not use systemd-networkd while having a new-ish kernel, is this workaround likely to be required for Debian 12 as well if it does not use systemd-networkd? I haven't yet created a "fresh" Debian 12 install to see what it uses by default...

hakman:
This setting is needed only for IPv6, where addresses are not allocated without it (due to forwarding being enabled).
For `systemd-networkd`, there's no need to do such things:
https://www.freedesktop.org/software/systemd/man/systemd.network.html#IPv6AcceptRA=

hakman (approved):

ReillyBrogan:
I guess we'll just wait and see if the official Bookworm images ship with systemd-networkd enabled or not.

PR #15234, opened 2023-03-15
Make it possible to use assume role

zetaab:
What is our current problem? We are hosting kOps management portal in AWS (which has AWS role). We are using that to provision clusters to another AWS accounts. Currently we are using management portal role and calling assumerole to another account role. However, because its called "role chaining" we can get maximum 1 hour long credentials from STS. In all cases the 1 hour long credentials are not enough (for instance kops rolling-update). My goal in this PR is that kOps could automatically refresh assumed role. So in future I could just specify `AWS_ASSUME_ROLE_ARN` to kOps cli binary, and it will automatically assume and refresh itself if needed.

### DIFF
--- a/upup/pkg/fi/cloudup/awsup/aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/aws_cloud.go
@@ -276,7 +276,7 @@ func NewAWSCloud(region string, tags map[string]string) (AWSCloud, error) {
 		}
 
 		// assumes the role before executing commands
-		roleARN := os.Getenv("AWS_ASSUME_ROLE_ARN")
+		roleARN := os.Getenv("KOPS_AWS_ROLE_ARN")
 		if roleARN != "" {
 			creds := stscreds.NewCredentials(sess, roleARN)
 			config = &aws.Config{Credentials: creds}
--- a/upup/pkg/fi/cloudup/awsup/aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/aws_cloud.go
@@ -19,6 +19,7 @@ package awsup
 import (
 	"context"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -33,6 +34,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
@@ -240,6 +242,13 @@ func ResetAWSCloudInstances() {
 	awsCloudInstances = make(map[string]AWSCloud)
 }
 
+func setConfig(config *aws.Config) *aws.Config {
+	// This avoids a confusing error message when we fail to get credentials
+	// e.g. https://github.com/kubernetes/kops/issues/605
+	config = config.WithCredentialsChainVerboseErrors(true)
+	return request.WithRetryer(config, newLoggingRetryer(ClientMaxRetries))
+}
+
 func NewAWSCloud(region string, tags map[string]string) (AWSCloud, error) {
 	raw := awsCloudInstances[region]
 	if raw == nil {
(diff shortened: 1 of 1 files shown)


zetaab:
cc @hakman @olemarkus I need some help here, if you read the description and check the solution. Does it make sense at all? Is there easier way to solve this? 

func stscreds.NewCredentials():

```
// NewCredentials returns a pointer to a new Credentials value wrapping the
// AssumeRoleProvider. The credentials will expire every 15 minutes and the
// role will be named after a nanosecond timestamp of this operation. The
// Credentials value will attempt to refresh the credentials using the provider
// when Credentials.Get is called, if the cached credentials are expiring.
```

so if I understand this correctly it will handle the credential refresh automatically out of the box?

rifelpet:
Can you define the role ARN in a profile in the AWS config file, and specify the profile via `AWS_PROFILE`?

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html

zetaab:
@rifelpet we are not using aws cli to call anything, its just go code. I will make sample code about this and test it inside kubernetes

zetaab:
I will test that, does it help if I generate ~/.aws/config and export AWS_PROFILE=foobar

zetaab:
hmm actually

```
	tmpl := fmt.Sprintf(`[profile assumed]
	role_arn = %s`, os.Getenv("KOPS_ROLE_ARN"))

	err := os.WriteFile("/code/.aws/config", []byte(tmpl), 0644)
	if err != nil {
		log.Fatalf("aliveCheckNew2 writefile %s", err)
	}
	os.Setenv("HOME", "/code")
	os.Setenv("AWS_PROFILE", "assumed")
```

solves the same issue (before executing kops cli)

zetaab:
it seems that its not possible, similar issue https://github.com/aws/aws-cli/issues/3875

the problem is that the portal is getting the original aws role as env variables:

```
% kubectl exec -it credtest-d7c469dc4-lcrrf -- env|grep AWS
AWS_REGION=eu-central-1
AWS_DEFAULT_REGION=eu-central-1
AWS_STS_REGIONAL_ENDPOINTS=regional
AWS_ROLE_ARN=arn:aws:iam::xxx:role/kapi-role
AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

after that it cannot assume role again (see issue)

zetaab:
Screenshot 2023-03-16 at 10 31 11

setup is something like this. We have pod running in account x, using role r1. We are executing `kops` inside that pod, but before we can really execute anything we need to assume the role to r2 (account y role which is allowed to do stuff in account y).

zetaab:
this does not break any current behaviour, but makes it work for this use-case. I do not see huge risk of merging this. However, if someone have better idea how to make it work - that is maybe better to discuss first.

test code available here https://github.com/zetaab/credtest/blob/main/main.go

explanations:
func aliveCheck - current behaviour, we will create new credentials with assumerole before starting kops cli. The credentials will expire after one hour (which is problem when rolling updating cluster)
func aliveCheckNew - the same behaviour than in this PR, works as planned. It will automatically refresh the credentials when needed
func aliveCheckNew2 - tried to use aws configfile to override the behaviour if AWS clients could assume role before when executing commands. However, it does not work - it does not assume role again if its already using role.

zetaab:
@hakman any news to this?

hakman:
@zetaab I will try to discuss it during today's office hours with @justinsb.

justinsb:
So I was wondering about whether this should be AWS_ASSUME_ROLE_ARN or AWS_ROLE_ARN, .  AWS_ROLE_ARN is used by the aws cli and most SDKs, but it is only used for web identity.  It also sounds like we want to avoid potential conflicts here, and anyone using IRSA (as you are) will have AWS_ROLE_ARN set, so there would be a high chance of conflicts.

Code LGTM, just trying to understand the nuances here.

zetaab:
@justinsb yep `AWS_ROLE_ARN` is used by AWS CDKs in case of web identity. We have the use-case that pod is started using web identity role, which means that it will inject role to pod which is used for commands. However, after that we would like to assume it again using another role (otherwise we need to add own oidc providers for ALL aws accounts). So when we start pod using IRSA, the contoller will inject env variable `AWS_ROLE_ARN` to pod. After that using `AWS_ASSUME_ROLE_ARN` works fine and it will not conflict with anything.

zetaab:
@hakman / @justinsb how we could proceed with this one? IMO it should not break any existing setup, but it makes it possible to use roles with roles :)

hakman:
@zetaab what do you think about `KOPS_AWS_ROLE_ARN` for now?

hakman (approved):

zetaab:
@hakman I rebased, needs lgtm again

Issue #15232, opened 2023-03-15
kops 1.26.2 no longer applying additionalPolicies.master policy


jim-barber-he:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

```
$ kops version
Client version: 1.26.2 (git-v1.26.2)
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
 
 ```
 $ kubectl version    
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.2", GitCommit:"fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b", GitTreeState:"clean", BuildDate:"2023-02-22T13:39:03Z", GoVersion:"go1.19.6", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.6", GitCommit:"ff2c119726cc1f8926fb0585c74b25921e866a28", GitTreeState:"clean", BuildDate:"2023-01-18T19:15:26Z", GoVersion:"go1.19.5", Compiler:"gc", Platform:"linux/amd64"}
 ```

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

I have a 1.25.6 cluster created with kops 1.25.3  with the following in the cluster spec
```yaml
  additionalPolicies:
    master: |
      [
        {
          "Action": [
            "sts:AssumeRole"
          ],
          "Effect": "Allow",
          "Resource": [
            "arn:aws:iam::REDACTED:role/kiam.*"
          ]
        }
      ]
```
I am using kops 1.26.2 to update the cluster like so:
`kops update cluster`

**5. What happened after the commands executed?**

The changes shown by kops contained:
```
  IAMRolePolicy/additional.masters.$CLUSTER_NAME
        PolicyDocument
                                - {
                                -   "Statement": [
                                -     {
                                -       "Action": [
                                -         "sts:AssumeRole"
                                -       ],
                                -       "Effect": "Allow",
                                -       "Resource": [
                                -         "arn:aws:iam::REDACTED:role/kiam.*"
                                -       ]
                                -     }
                                -   ],
[...]

PR #15238, opened 2023-03-16
aws: Use `control-plane` for additional policies instead of `master`

hakman:
Fixes https://github.com/kubernetes/kops/issues/15232

Same as for external policies:
https://github.com/kubernetes/kops/blob/071f9a86ad5ee2c0fcf7388206a92beeed743714/pkg/model/awsmodel/iam.go#L326-L332


rifelpet (approved):
For reference, the conversion from v1alpha2 to the unversioned API happens here:

https://github.com/kubernetes/kops/blob/036bb8c62ba0574a0a666e073d6ed55ef47265b9/pkg/apis/kops/v1alpha2/conversion.go#L81-L99

PR #15251, opened 2023-03-18
Outdated etcdctl certs path in official docs

dishavirk:


hakman:
Hi @dishavirk. I may be missing something here. For me, the cert dir, with latest kOps version is `/rootfs/srv/kubernetes/kube-apiserver/`. What version are you using?

PR #15256, opened 2023-03-19
feat(spot): Setting the VNG Size Limits in Launch Spec

IdanShohamNetApp:
This PR adds setting support for VNG size limits in launch spec.


hakman:
@IdanShohamNetApp How does this change affect work with the initial implementation in https://github.com/kubernetes/kops/pull/13234?

IdanShohamNetApp:
> @IdanShohamNetApp How does this change affect work with the initial implementation in #13234?

hello @hakman 
this change adds support for limit size inside the definition of a VNG, in the initial implementation that wasn't supported.

hakman (approved):

Issue #15245, opened 2023-03-17
Generated Terraform code related to vpc_ipv6_cidr_block fails


felixfriedrich:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

`v1.26.2`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

`v1.25.2`

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

We generate Terraform code based on a `cluster.yaml`:
```
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
...
```

* `kops replace -f cluster.yml`
* `kops update cluster --target=terraform --name $(CLUSTER_NAME) --out=.`

The two commands above generate Terraform code which contains the following bits, which are for the sake of this bug report as minimal as possible:

```
data "aws_vpc" "my_vpc" {
  id = "vpc-REDACTED"
}

locals {
  vpc_ipv6_cidr_block  = data.aws_vpc.my_vpc.ipv6_cidr_block
  vpc_ipv6_cidr_length = local.vpc_ipv6_cidr_block == null ? null : tonumber(regex(".*/(\\d+)", local.vpc_ipv6_cidr_block)[0])
}
```

**5. What happened after the commands executed?**

This code fails whoever with:
```
│ Error: Error in function call
│
│   on kops-bug.tf line 7, in locals:
│    7:   vpc_ipv6_cidr_length = local.vpc_ipv6_cidr_block == null ? null : tonumber(regex(".*/(\\d+)", local.vpc_ipv6_cidr_block)[0])
│     ├────────────────
│     │ while calling regex(pattern, string)
│     │ local.vpc_ipv6_cidr_block is ""
│
│ Call to function "regex" failed: pattern did not match any part of the given string.
```

**6. What did you expect to happen?**

Correcting the code like below works:
```
local.vpc_ipv6_cidr_block == null ? null : tonumber(regex(".*/(\\d+)", local.vpc_ipv6_cidr_block)[0])
```

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: null
  name: cluster-1.example.com
[...]

PR #15262, opened 2023-03-21
aws: Set local.vpc_ipv6_cidr_block only for IPv6 managed clusters

hakman:
Fixes https://github.com/kubernetes/kops/issues/15245

/cc @rifelpet @johngmyers

### DIFF
--- a/tests/integration/update_cluster/shared_vpc_ipv6/kubernetes.tf
+++ b/tests/integration/update_cluster/shared_vpc_ipv6/kubernetes.tf
@@ -22,7 +22,6 @@ locals {
   vpc_cidr_block                    = data.aws_vpc.minimal-ipv6-example-com.cidr_block
   vpc_id                            = "vpc-12345678"
   vpc_ipv6_cidr_block               = data.aws_vpc.minimal-ipv6-example-com.ipv6_cidr_block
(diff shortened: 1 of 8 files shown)


tests/integration/update_cluster/shared_vpc_ipv6/kubernetes.tf
```diff
@@ -22,7 +22,6 @@ locals {
   vpc_cidr_block                    = data.aws_vpc.minimal-ipv6-example-com.cidr_block
   vpc_id                            = "vpc-12345678"
   vpc_ipv6_cidr_block               = data.aws_vpc.minimal-ipv6-example-com.ipv6_cidr_block
```
rifelpet, line 24: This integration test case uses ipv6 so we probably shouldn't be removing it here. In general we may not be able to know if a "shared VPC" has ipv6 enabled or not

johngmyers:
The code suppresses `local.vpc_ipv6_cidr_length`, not `local.vpc_ipv6_cidr_block`. It doesn't appear to be correct.

I think the correct fix would be to compare against `""` instead of `null`.

hakman:
> The code suppresses `local.vpc_ipv6_cidr_length`, not `local.vpc_ipv6_cidr_block`. It doesn't appear to be correct.
> 
> I think the correct fix would be to compare against `""` instead of `null`.

Agreed.

PR #15266, opened 2023-03-24
OpenStack: Add network-cidr config for etcd-manager

ederst:
This add the possibility of setting `--network-cidr` for etcd-manager when using OpenStack, which was introduced with https://github.com/kubernetes-sigs/etcdadm/pull/317

### DIFF
--- a/pkg/model/components/etcdmanager/model.go
+++ b/pkg/model/components/etcdmanager/model.go
@@ -478,6 +478,7 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
 				fmt.Sprintf("%s=%s", openstack.TagClusterName, b.Cluster.Name),
 			}
 			config.VolumeNameTag = openstack.TagNameEtcdClusterPrefix + etcdCluster.Name
+			config.NetworkCIDR = fi.PtrTo(b.Cluster.Spec.Networking.NetworkCIDR)
 
 		case kops.CloudProviderScaleway:
 			config.VolumeProvider = "scaleway"
@@ -589,6 +590,7 @@ type config struct {
 	VolumeTag             []string `flag:"volume-tag,repeat"`
 	VolumeNameTag         string   `flag:"volume-name-tag"`
 	DNSSuffix             string   `flag:"dns-suffix"`
+	NetworkCIDR           *string  `flag:"network-cidr"`
 }
 
 // SelectorForCluster returns the selector that should be used to select our pods (from services)


pkg/model/components/etcdmanager/model.go
```diff
@@ -478,6 +478,7 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
 				fmt.Sprintf("%s=%s", openstack.TagClusterName, b.Cluster.Name),
 			}
 			config.VolumeNameTag = openstack.TagNameEtcdClusterPrefix + etcdCluster.Name
+			config.NetworkCIDR = etcdCluster.Manager.NetworkCIDR
```
zetaab, line 481: can etcdCluster.Manager be nil?

pkg/apis/kops/validation/validation.go
```diff
@@ -1382,6 +1382,14 @@ func validateEtcdClusterSpec(spec kops.EtcdClusterSpec, c *kops.Cluster, fieldPa
 		allErrs = append(allErrs, validateEtcdMemberSpec(m, fieldPath.Child("etcdMembers").Index(i))...)
 	}
 
+	if spec.Manager != nil && spec.Manager.NetworkCIDR != nil {
+		networkCIDRPath := fieldPath.Child("manager").Child("networkCIDR")
+		if c.Spec.CloudProvider.Openstack == nil {
```
zetaab, line 1387: if c.Spec.GetCloudProvider() != api.CloudProviderOpenstack

perhaps better way to check the cloud, that is what is used in most of the places in kops

zetaab (changes requested):

zetaab:
@ederst basically kops does not support currently adding more than one interface to server which is created by kops? So you are adding that port manually after kops created the server? Anyway I am not against this PR, just thinking what kind of use-cases people are doing

pkg/model/components/etcdmanager/model.go
ederst: Yes I think so (added a check for it at another place but not here), and there should be a check if the network cidr should even be set.

ederst:
> @ederst basically kops does not support currently adding more than one interface to server which is created by kops? So you are adding that port manually after kops created the server? Anyway I am not against this PR, just thinking what kind of use-cases people are doing

Yes, we add a port with an IPv6 address with a little "controller" I wrote, to work around the issue that kOps does not support IPv6/adding multiple ports/ips per port for OpenStack.

This revealed that the IPs returned by the OpenStack API are not consistently sorted, even when on the same port, sometimes the IPv6 was first then the IPv4, which led to phenomens like:
* Not being able to attach to pods due to the Ipv6 being used as ip for the node (set by OCCM as received by the API but mitigated with a PR of mine as well)
* ETCD manager picking the Ipv6 instead of Ipv4, and the next time the Ipv4 again

So, this will help to conistently choose an IP when adding multiple ports like we do, and when mb kOps will support adding multiple ports and/or Ipv6 for OpenStack (which will probably require dual stack hosts like in AWS atm).

hakman:
My understanding was that we will pass to etcd `cluster.Spec.Networking.NetworkCIDR`. Am I missing something?
/hold

ederst:
> My understanding was that we will pass to etcd `cluster.Spec.Networking.NetworkCIDR`. Am I missing something? /hold

I think this will be better because no config is needed and the "primary" CIDR is just passed on to etcd-manager which at least in my case is what I want. Don't know if it makes sense to make it overridable in case someone wants to configure a different CIDR - for example to force etcd-manager to use IPv6 or a "manage/utility" subnet. But this is not my use case, so I tend to go that easier route.

hakman:
> > My understanding was that we will pass to etcd `cluster.Spec.Networking.NetworkCIDR`. Am I missing something? /hold
> 
> I think this will be better because no config is needed and the "primary" CIDR is just passed on to etcd-manager which at least in my case is what I want. Don't know if it makes sense to make it overridable in case someone wants to configure a different CIDR - for example to force etcd-manager to use IPv6 or a "manage/utility" subnet. But this is not my use case, so I tend to go that easier route.

I don't think it makes sense to allow overrides.
If there is a valid use case, we can look at it at that point in the future.

ederst:
@hakman @zetaab adapted accordingly: only sets the `Networking.NetworkCIDR` now

zetaab:
the interesting question is: does etcd-manager crash if we insert flags to it that is not supported? If yes, then we cannot insert --network-cidr to all images. In case it does not crash, then I would say we do not care

ederst:
> the interesting question is: does etcd-manager crash if we insert flags to it that is not supported? If yes, then we cannot insert --network-cidr to all images. In case it does not crash, then I would say we do not care

Yes, as I suspected it will fail:

```shell
❯ docker run -it --rm -v /tmp/blubb:/tmp/blubb registry.k8s.io/etcdadm/etcd-manager:v3.0.20230201 --network-cidr 123.123.123.123/12
flag provided but not defined: -network-cidr
Usage of /etcd-manager:
  -add_dir_header
    	If true, adds the file directory to the header of the log messages
  -address string
```

So we need to care. Maybe a simple on/off switch, because I think for example doing this based on the version of the image might be a bit dodgy.

hakman:
We don't need to care that much. I don't think it's expected to run kOps with older version or etcd-manager.
In the case of kOps 1.27, there are already breaking changes in that area and there will be more.

ederst:
> We don't need to care that much. I don't think it's expected to run kOps with older version or etcd-manager. In the case of kOps 1.27, there are already breaking changes in that area and there will be more.

Ok, then I guess a simple hint in "Breaking Changes" will do.

ederst:
> > > We don't need to care that much. I don't think it's expected to run kOps with older version or etcd-manager. In the case of kOps 1.27, there are already breaking changes in that area and there will be more.
> > 
> > 
> > Ok, then I guess a simple hint in "Breaking Changes" will do.
> 
> Not sure we have to go so far. etcd-manager should be kept in sync with kOps.

Well, OK then, I guess if @zetaab is daccord with the changes then let's do the minimal approach :)

zetaab (approved):
/approve
/lgtm

but we cannot cherry-pick this to older releases, because otherwise we need to cherry-pick etcd-managers as well

PR #15267, opened 2023-03-24
OpenStack: Add OCCM address sort order config

ederst:
This will add the OCCM config to specify an address sort order:
* https://github.com/kubernetes/cloud-provider-openstack/pull/1946


zetaab (approved):
/approve
/lgtm

PR #15270, opened 2023-03-26
Update kube-router to v1.5.3

chenliu1993:
Update kube-router to v1.5.3

Tried in aws provider using kops ans other components built based on master and it worked as it was in v1.2.3


chenliu1993:
hi @hakman , I am trying to understand the process from promoting the kube-router version to stable channel and do a version-up. May I know what should be done to make sure the kube-router cni is ok to put in release? Should I simply add a test-infra kops-grid option for kube-router?

hakman:
Thanks @chenliu1993.
It looks good now that `pull-kops-e2e-cni-kuberouter` is passing. Generally, there are other changes to the manifest, but there were no changes to any of the files in https://github.com/cloudnativelabs/kube-router/tree/master/daemonset for the last 3 years. Maybe once 2.0.0 will be out, there will be something more to change.
/lgtm
/approve

PR #15273, opened 2023-03-28
test scenarios: create simple scalability test

justinsb:
We want to try some scalability testing on AWS.


hakman (approved):

hakman:
`--dns=none` would be nice to use for this scenario. It would make it easier to run locally, less prone to failure in different envs and could benefit from the cluster name generation from kubetest.

PR #15275, opened 2023-03-28
Fix IPv6 Terraform code generation

felixfriedrich:
This PR fixes https://github.com/kubernetes/kops/issues/15245

`LiteralEqualConditionExpression` is a bit more generic than necessary. I couldn't find a good name for a function, which would do an equal comparison to empty String, but would return null in that case.
Creating a function, which would generically do a equal comparison felt more natural. Also, the return values are now explicit at the place of calling.

Happy to change this in case there are any concerns.

### DIFF
--- a/upup/pkg/fi/cloudup/terraformWriter/literal.go
+++ b/upup/pkg/fi/cloudup/terraformWriter/literal.go
@@ -140,13 +140,13 @@ func LiteralListExpression(args ...*Literal) *Literal {
 	}
 }
 
-// LiteralNullConditionalExpression constructs a Literal which returns `null`
-// if the supplied "nullable" expression is null, otherwise returns "value".
+// LiteralEmptyStrConditionalExpression constructs a Literal which returns `null`
+// if the supplied "empty" expression is an empty string, otherwise returns "value".
 // It is the caller's responsibility to ensure the supplied parameters do not use operators
 // with lower precedence than the conditional operator.
-func LiteralNullConditionalExpression(nullable, value *Literal) *Literal {
+func LiteralEmptyStrConditionalExpression(empty, value *Literal) *Literal {
 	return &Literal{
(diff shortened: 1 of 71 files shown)


felixfriedrich:
@hakman All GitHub Actions that ran succeeded. What else needs to be done in terms of testing? What's the next step to get this merged?

felixfriedrich:
@johngmyers I am replacing a function you write. Would you have a look at me change? It fixes the issue linked to in the description.

upup/pkg/fi/cloudup/terraformWriter/literal.go
```diff
@@ -140,13 +140,14 @@ func LiteralListExpression(args ...*Literal) *Literal {
...
 
-// LiteralNullConditionalExpression constructs a Literal which returns `null`
-// if the supplied "nullable" expression is null, otherwise returns "value".
+// LiteralEqualConditionExpression constructs a Literal which compares "equal" to "to".
+// If "equal" is equal to "to" "either" is being returned, otherwise "or".
+// The result looks like this: "equal == to ? either : or"
 // It is the caller's responsibility to ensure the supplied parameters do not use operators
 // with lower precedence than the conditional operator.
-func LiteralNullConditionalExpression(nullable, value *Literal) *Literal {
+func LiteralEqualConditionExpression(equal, to, either, or *Literal) *Literal {
```
hakman, line 148: Could you keep this simple, just change from `null` to empty str?
```suggestion
func LiteralEmptyStrConditionalExpression(equal, to, either, or *Literal) *Literal {
```

felixfriedrich: This is indeed way simpler. Thanks for the input.

hakman (approved):
Thanks @felixfriedrich! :)

PR #15278, opened 2023-03-29
kubetest2: only set storage.testdriver if manifest is present

justinsb:


hakman (approved):

rifelpet:
> I'm not entirely sure why we're doing this, so uploading for discussion / in case anyone has context.

https://github.com/kubernetes/kops/pull/14098#issuecomment-1207173222 explains why this e2e flag is set. This is causing problems because it assumes the working directory is the kops repo root, correct? I wonder if we could detect the repo root in a different manner or even use go:embed to store the driver manifests in the kubetest2-tester-kops binary directly, have it write the manifest to a temp file, and pass its path to the e2e flag.

PR #15294, opened 2023-04-05
Update bastion.md

recollir:
This feature is not available in 1.22. It started to become available with version 1.23


hakman (approved):
Thanks for correcting this @recollir.
/ok-to-test

PR #15299, opened 2023-04-06
OpenStack: Use interface attach when reconciling server ports 

ederst:
This makes use of the interface attach method when reconciling server ports.

The difference between just setting the `DeviceID` on the Port and this is that with the attachment process the server is also validated. Which means for example that a Port cannot be attached to a server in `ERROR`
state.


ederst:
@zetaab can you have a look at this too please?

zetaab (approved):
/ok-to-test
/approve
/lgtm

zetaab:
@ederst run `make gomod`

hakman:
> @ederst run `make gomod`

Maybe also rebase.

ederst:
> > @ederst run `make gomod`
> 
> Maybe also rebase.

@hakman @zetaab whynotboth.gif :)

Rebased and ran `make gomod` with up-to-date go, should do it.

PR #15300, opened 2023-04-06
OpenStack: Build InstanceGroup despite missing IP

ederst:
This will allow kOps to build an OpenStack InstanceGroup with a missing fixed IP.

In essence this will get rid of `interface name X not found` errors, when there are servers present which do not have an interface attached or are in a state (e.g. `ERROR`) which does not allow this. And will mark such a server with `NeedsUpdate`.

Currently this will happen:

```log
❯ openstack server list --status ERROR
2b2a77cc-8edb-4b21-b23c-dca515264aea worker-avjtq4    ERROR
6cf7570c-78d2-49e9-af5a-7376246d7cce worker-ijaqik    ERROR

❯ kops get instances
Error: error getting cloud instance group "worker": error creating cloud instance group member: server `6cf7570c-78d2-49e9-af5a-7376246d7cce` interface name `my-cluster-network` not found

❯ kops validate cluster
Validating cluster my-cluster.k8s.local

Error: Validation failed: unexpected error during validation: error getting cloud instance group "worker": error creating cloud instance group member: server `6cf7570c-78d2-49e9-af5a-7376246d7cce` interface name `my-cluster-network` not found
```

By ignoring the errors of `GetFixedIP()` when building the IGs the above commands will show now something more sensible (node without an IP and in e.g. `ERROR` state, validation will show node as "has not yet joined cluster").

### DIFF
--- a/upup/pkg/fi/cloudup/openstack/server_group.go
+++ b/upup/pkg/fi/cloudup/openstack/server_group.go
@@ -135,7 +135,7 @@ func osBuildCloudInstanceGroup(c OpenstackCloud, cluster *kops.Cluster, ig *kops
 		generationName := fmt.Sprintf("%d-%d", cluster.GetGeneration(), ig.Generation)
 
 		status := cloudinstances.CloudInstanceStatusUpToDate
-		if generationName != observedName {
+		if generationName != observedName || server.Status == errorStatus {
 			status = cloudinstances.CloudInstanceStatusNeedsUpdate
 		}
 		cm, err := cg.NewCloudInstance(instanceId, status, nodeMap[instanceId])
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/openstack/server_group.go
```diff
@@ -147,10 +147,7 @@ func osBuildCloudInstanceGroup(c OpenstackCloud, cluster *kops.Cluster, ig *kops
 			cm.MachineType = server.Flavor["original_name"].(string)
 		}
 
-		ip, err := GetServerFixedIP(server, server.Metadata[TagKopsNetwork])
-		if err != nil {
-			return nil, fmt.Errorf("error creating cloud instance group member: %v", err)
-		}
+		ip, _ := GetServerFixedIP(server, server.Metadata[TagKopsNetwork])
```
zetaab, line 150: we could still handle the error and print like klog.Warningf just in case?

zetaab (changes requested):

upup/pkg/fi/cloudup/openstack/server_group.go
ederst: @zetaab added a warn log

zetaab (approved):
thanks

/approve
/lgtm

PR #15301, opened 2023-04-06
OpenStack: Use task engine to retry failed servers

ederst:
This will make use of the kOps taks engine to retry failed servers.

The [former approach](https://github.com/kubernetes/kops/pull/15102) had the side effect of not making kOps fail when the last retry failed:

Because there is now a server present - although in an erroneous state - the "instance task" which the task engine retried reconciled the server (port, floating ip) instead of recreating it again.

With the approach of letting the task engine retry the failed servers this will be handled correctly.


zetaab (approved):
/approve
/lgtm

Thank you for fixing it

PR #15316, opened 2023-04-13
Create func to reset awsCloudInstances

ltellesfl:
Since we use kops as lib, we need to reset this variable so that we can create clients with another AWS account.

Here is an example that also follows the same idea and was recently changed:

https://github.com/kubernetes/kops/blob/v1.26.2/util/pkg/vfs/context.go#L65

### DIFF
--- a/upup/pkg/fi/cloudup/awsup/aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/aws_cloud.go
@@ -234,10 +234,14 @@ func (c *awsCloudImplementation) Region() string {
 	return c.region
 }
 
-var awsCloudInstances map[string]AWSCloud = make(map[string]AWSCloud)
+var AWSCloudInstances = NewAWSCloudInstances()
+
+func NewAWSCloudInstances() map[string]AWSCloud {
+	return make(map[string]AWSCloud)
+}
 
 func NewAWSCloud(region string, tags map[string]string) (AWSCloud, error) {
-	raw := awsCloudInstances[region]
+	raw := AWSCloudInstances[region]
 	if raw == nil {
 		c := &awsCloudImplementation{
 			region: region,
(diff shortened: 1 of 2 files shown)


rafatio:
Hey folks! Any updates about this?

upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -234,10 +234,14 @@ func (c *awsCloudImplementation) Region() string {
 	return c.region
 }
 
-var awsCloudInstances map[string]AWSCloud = make(map[string]AWSCloud)
+var AWSCloudInstances = NewAWSCloudInstances()
```
olemarkus, line 237: Making this one public doesn't seem like a good idea. Perhaps create a function for setting a specific `AWSCloud` to a specific region.

olemarkus (changes requested):

upup/pkg/fi/cloudup/awsup/aws_cloud.go
ltellesfl: The problem is not the region, it's the AWS account. The region is the same but not the account, the second time it runs it keeps the credentials for the account that was set on the first run. When NewAWSCloud() func runs the second time it will not create again the clients because of this [if](https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/awsup/aws_cloud.go#L241). Since we want to use multi-account we need to reset those clients to use with a different AWS account, [here is an example](https://github.com/topfreegames/kubernetes-kops-operator/blob/main/utils/kops_utils.go#L110) of how we did using a fork with the proposed changes.

olemarkus: Yes, I figured that. So what I am proposing is that you can create a new AWSCloud instance for the same region rather than just resetting the map and rerun the entire instantiation logic.

Alternatively, make a public function that resets the private map. Either is better than opening the whole map up for manipulation anywhere in the code.

ltellesfl:
Hello, any updates?

olemarkus (approved):

PR #15318, opened 2023-04-14
hetzer: update CSI driver to v2.3.2

metatarz:
This update is needed since the old hcloud-csi-driver v.2.0.0 image lacks ARM64 support and hence, it fails to run on new hetzner ARM64 servers.


metatarz:
@hakman

hakman:
Thanks for the effort @metatarz.
A more complete version of this was done in https://github.com/kubernetes/kops/pull/15440 and should be available in 1.27.0.

PR #15319, opened 2023-04-15
Update containerd to v1.6.20

anthonyhaussman:
- https://github.com/containerd/containerd/releases/tag/v1.6.19
 - https://github.com/containerd/containerd/releases/tag/v1.6.20


hakman:
@anthonyhaussman This looks good. Could you split the Go part into a separate commit, like in https://github.com/kubernetes/kops/pull/15159/commits? It would make it easier to cherry-pick.

anthonyhaussman:
> @anthonyhaussman This looks good. Could you split the Go part into a separate commit, like in https://github.com/kubernetes/kops/pull/15159/commits? It would make it easier to cherry-pick.

It's done @hakman :slightly_smiling_face:

hakman:
Thanks @anthonyhaussman!
/retest

hakman (approved):

olemarkus:
It's been released, but kOps needs to start supporting downloading and using it. The credentials provider is also not really tested upstream, so there may be more work needed there as well.

Currently working on what's needed kOps-side and will then add e2e in the aws cloud provider.

rifelpet:
How about reverting the deps update? That way the head of master is not broken

olemarkus:
The problem is that the functionality for pulling ECR-containers are removed from kubelet. So we'd have to run these tests with K8s 1.26 instead. Which I am not necessarily against. If the ECR credentials provider is also broken, we won't get the PR to start using it into the repo either.

rifelpet:
If we know k8s 1.27 doesn't work with kops yet then temporarily pinning our presubmit tests to k8s 1.26 seems reasonable

PR #15324, opened 2023-04-16
docs: Add note to 1.22 about the CA CN rename

agilgur5:
## Summary

Add a note in 1.22 changelog about https://github.com/kubernetes/kops/pull/11921 et al's CA changes

## Details

- this broke some of my teams' automation code that relied on the CN, so thought it would be good to call out in case anyone else stumbles upon this in order to not spend a few hours debugging 😅 
  - this change may particularly impact Prod environments, where the cluster has not been rebuilt in some time (years), and so they will have the old CN while new clusters in lower environments will have the new CN
    - so code that relies on the CN may unexpectedly break Production while working fine in lower environments 😬 😬 😬 
      - fortunately we caught this in our QA env, but it passed Dev fine
      
## Testing Evidence

From an internal fix my team made to another team's automation:
k8s ca change

(We've recommend they pull the CA from `/etc/kubernetes/pki/ca.crt` or `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` to be a bit more resilient to changes like this)

## Review Notes

Feel free to change the language / wording as you see fit or put in a different part of the release notes (I put this under "Other changes of note" right now).

I thought specifically warning about older clusters was important, as otherwise this is particularly easy to overlook

### DIFF
--- a/docs/releases/1.22-NOTES.md
+++ b/docs/releases/1.22-NOTES.md
@@ -163,6 +163,9 @@ For file assets, it means adding an explicit path as shown below:
   The previous behavior of using self-signed certs may be restored by setting `kubeControllerManager.tlsCertFile` and/or
   `kubeScheduler.tlsCertFile` to `""` in the cluster spec.
 
+* kOps has changed the CN of the Kubernetes general CA from `kubernetes` to `kubernetes-ca`.
+  This change is backward-compatible, but if you were relying on the CN for some code, you will need to ensure it is compatible with both the old name and the new name as you may have older clusters with the old CN and newer clusters with the new CN.
+
 * Cilium now supports the wireguard protocol for transparent encryption.
 
 
--- a/docs/releases/1.22-NOTES.md
+++ b/docs/releases/1.22-NOTES.md
@@ -163,8 +163,7 @@ For file assets, it means adding an explicit path as shown below:
   The previous behavior of using self-signed certs may be restored by setting `kubeControllerManager.tlsCertFile` and/or
   `kubeScheduler.tlsCertFile` to `""` in the cluster spec.
 
-* kOps has changed the CN of the Kubernetes general CA from `kubernetes` to `kubernetes-ca`.
-  This change is backward-compatible, but if you were relying on the CN for some code, you will need to ensure it is compatible with both the old name and the new name as you may have older clusters with the old CN and newer clusters with the new CN.
+* kOps has changed the CN of the Kubernetes general CA from `kubernetes` to `kubernetes-ca`. This change is backward-compatible, but any tools using the CN will need to be updated.
 
 * Cilium now supports the wireguard protocol for transparent encryption.
 


agilgur5:
@johngmyers any chance you could take a look at this PR?

johngmyers:
I must admit to being at a loss as to why anything would depend on the CN having a particular value.

agilgur5:
The automation I mentioned here was for the Vault k8s integration, which requires [the k8s CA](https://developer.hashicorp.com/vault/docs/auth/kubernetes#configuration).

Can see in the screenshot above that this code does TLS inspection to pull the CA, but ends up getting two certs, the CA and the `kubernetes-master` cert. It uses the CN to distinguish between the two.

I don't think that's the best way to get the CA, as I mentioned above, but regardless, a team I work with _already_ had code that _specifically_ depended on the CN. I.e. the CN was part of the "public surface" that kOps exposes, and changing the CN broke existing code. Or, as is oft stated in Linux kernel development, anything that is exposed will be used (aka the ["Workflow" xkcd](https://xkcd.com/1172/)).

So at the very least, I think this should be documented in the changelog.
Per above, I think it is doubly important since this has a higher likelihood to break code that depends on the CN in Prod, as those clusters are often older / rebuilt less often and so may still have the old CN.

agilgur5:
@johngmyers can you take a look again?

johngmyers:
I don't think this rises to the level of a release note. This is an implementation detail; the CN of a root CA is not something that is supposed to matter to anything.

agilgur5:
It's a breaking change that broke actual code in production exactly because it wasn't mentioned in the release notes...

johngmyers:
kOps 1.22 has been out of support for about a year. One person making an incorrect assumption about internals does not merit a release note.

agilgur5:
> kOps 1.22 has been out of support for about a year.

I'll repeat once more that this breaking change affects clusters that upgraded to 1.22 _and beyond_ and were not rebuilt in that time, which is, again, much more likely for Prod clusters that have zero downtime.
This affects all clusters that existed pre-1.22 and were upgraded to 1.22+ through the standard kOps upgrade process.
A cluster that was upgraded from 1.21 to 1.26, which is current, would also have this issue.

> One person making an incorrect assumption about internals

I'm not the one who made that assumption but I have no idea how anyone should know not to rely on that. That is, similarly, undocumented behavior.

agilgur5:
> the CN of a root CA is not something that is supposed to matter to anything.

One could say that this too, is an assumption. And there is an existing counter-example that shows it did, in fact, matter to something 🤷  So one could further say that _that_ is an incorrect assumption.

I'm not here to play word games and state opinions, I am stating facts and events that _literally already happened_. I am here to warn other users about that, because if we did not catch that in QA, that would have resulted in a major issue to a large enterprise company. I would not wish production outages upon anybody, and the absolute least I can do is add a tiny warning.

docs/releases/1.22-NOTES.md
```diff
@@ -163,6 +163,9 @@ For file assets, it means adding an explicit path as shown below:
   The previous behavior of using self-signed certs may be restored by setting `kubeControllerManager.tlsCertFile` and/or
   `kubeScheduler.tlsCertFile` to `""` in the cluster spec.
 
+* kOps has changed the CN of the Kubernetes general CA from `kubernetes` to `kubernetes-ca`.
+  This change is backward-compatible, but if you were relying on the CN for some code, you will need to ensure it is compatible with both the old name and the new name as you may have older clusters with the old CN and newer clusters with the new CN.
```
hakman:
```suggestion
* kOps has changed the CN of the Kubernetes general CA from `kubernetes` to `kubernetes-ca`. This change is backward-compatible, but any tools using the CN will need to be updated.
```

agilgur5: I appreciate the succinctness here, although this is less explicit about how to handle the change since if you only have clusters built prior to 1.22, you _wouldn't_ need to update any tooling until you rebuilt a cluster vs upgraded (which may not happen frequently in prod in particular). That's what makes it particularly confusing and hard to track down, as two clusters on the same version may have different CNs.

Although now that this is merged, it is much easier to find and search for, and folks can `blame` and read the commit or PR for more details at least

PR #15331, opened 2023-04-18
gce: set ip address family on all FirewallRule tasks

justinsb:
We had missed a few code paths previously.


rifelpet:
I think there are a few tasks defined in firewall.go that need updating too

hakman (approved):

justinsb:
@rifelpet it's a good point; I should haven't said "all" FirewallRule tasks, because I _think_ we only need to set this where we use SourceRanges.  I think I've got them all...

PR #15332, opened 2023-04-18
gce: Update logic for internal LB

hakman:
/cc @justinsb

### DIFF
--- a/pkg/model/gcemodel/api_loadbalancer.go
+++ b/pkg/model/gcemodel/api_loadbalancer.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/gce"
 	"k8s.io/kops/upup/pkg/fi/cloudup/gcetasks"
+	"k8s.io/utils/strings/slices"
 )
 
 // APILoadBalancerBuilder builds a LoadBalancer for accessing the API
@@ -132,7 +133,6 @@ func createPublicLB(b *APILoadBalancerBuilder, c *fi.CloudupModelBuilderContext)
 // GCP this entails creating a health check, backend service, and one forwarding rule
 // per specified subnet pointing to that backend service.
 func createInternalLB(b *APILoadBalancerBuilder, c *fi.CloudupModelBuilderContext) error {
-	lbSpec := b.Cluster.Spec.API.LoadBalancer
 	hc := &gcetasks.HealthCheck{
 		Name:      s(b.NameForHealthCheck("api")),
 		Port:      wellknownports.KubeAPIServer,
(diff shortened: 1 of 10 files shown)


upup/pkg/fi/cloudup/gcetasks/address.go
```diff
@@ -130,9 +140,15 @@ func (_ *Address) CheckChanges(a, e, changes *Address) error {
 func (_ *Address) RenderGCE(t *gce.GCEAPITarget, a, e, changes *Address) error {
 	cloud := t.Cloud
 	addr := &compute.Address{
-		Name:    *e.Name,
-		Address: fi.ValueOf(e.IPAddress),
-		Region:  cloud.Region(),
+		Name:        *e.Name,
+		Address:     fi.ValueOf(e.IPAddress),
+		AddressType: fi.ValueOf(e.IPAddressType),
+		Purpose:     "SHARED_LOADBALANCER_VIP",
```
justinsb, line 146: Do we need to specify this?

pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -161,48 +161,49 @@ func createInternalLB(b *APILoadBalancerBuilder, c *fi.CloudupModelBuilderContex
...
-		if subnetNotSpecified(sn, b.Cluster.Spec.Networking.Subnets) {
-			c.AddTask(subnet)
-		}
+
+	network, err := b.LinkToNetwork()
+	if err != nil {
+		return err
+	}
+
+	if len(b.Cluster.Spec.Networking.Subnets) != 1 {
```
justinsb, line 170: We can't support multiple subnets?  Is it that we would need multiple IPs?

hakman: Not quite sure we support multiple subnets. I thought we don't, but I may be wrong.
If that's the case, then yes, we need multiple IPs.

upup/pkg/fi/cloudup/gcetasks/address.go
hakman: We need it, because we have it shared between `kops-controller` and `api-server`.

pkg/model/gcemodel/api_loadbalancer.go
hakman: @justinsb Seems we don't support multiple subnets for GCE/GCP:
https://github.com/kubernetes/kops/blob/60a2e153a9590797e2e0a44efd3ea00c1589f0ad/upup/pkg/fi/cloudup/new_cluster.go#L643-L646

hakman: I did something along the lines of supporting multiple subnets, but it's not pretty.

hakman:
retest

justinsb:
Thanks @hakman 

/approve
/lgtm

justinsb:
BTW I think we also need a firewall rule for access from outside the cluster (but inside the VPC), but we can do that separately

PR #15336, opened 2023-04-20
fix(cilium): install CNI plugin binary in an InitContainer

anthonyhaussman:
Starting cilium version `1.12.8` and to reduces the potential security surface of the agent, Cilium removes the bind-mount of `/opt/cni/bin` into the template. Instead, write the binaries once in an initContainer.

Ref:
 - https://github.com/cilium/cilium/pull/24075

### DIFF
--- a/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.12.yaml.template
@@ -782,8 +782,10 @@ spec:
           {{- end }}
         - mountPath: /var/run/cilium
           name: cilium-run
+          {{- if not (semverCompare "~1.11.15 || ~1.12.8 || >=1.13.1" $semver) }}
         - mountPath: /host/opt/cni/bin
           name: cni-path
+          {{- end }}
         - mountPath: /host/etc/cni/net.d
           name: etc-cni-netd
 {{ if .EtcdManaged }}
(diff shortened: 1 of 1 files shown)


upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.12.yaml.template
```diff
@@ -817,6 +819,26 @@ spec:
 {{ end }}
       hostNetwork: true
       initContainers:
+      {{- if semverCompare ">=1.12.8" $semver }}
```
ReillyBrogan, line 822: `{{- if semverCompare "~1.11.15 || ~1.12.8 || >= 1.13.1" $semver }}`

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.12.yaml.template
```diff
@@ -782,8 +782,10 @@ spec:
           {{- end }}
         - mountPath: /var/run/cilium
           name: cilium-run
+          {{- if semverCompare "<1.12.8" $semver }}
```
ReillyBrogan, line 785: `{{- if not semverCompare "~1.11.15 || ~1.12.8 || >= 1.13.1" $semver }}`

ReillyBrogan (changes requested):
The commit in question was in 1.11.15 and 1.13.1 as well. I updated the semverCompare string accordingly (these are untested though, I'd recommend making sure they work).

upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.12.yaml.template
anthonyhaussman: Thanks for the suggestion, I have made the modification. :slightly_smiling_face:

s3than:
@olemarkus Is there anything needed for this PR? currently with the changes to Cilium it's blocking updating

hakman (approved):

s3than:
@hakman 🙇 Thank you

PR #15339, opened 2023-04-20
fix(firewall): open node to master ICMP security group

### LINKED ISSUE #15329: Cilium health check fails towards control plane node
/kind bug

**1. What `kops` version are you running?**

`Client version: 1.26.2 (git-v1.26.2)`

 a `kops` flag.**

`v1.25.9`

AWS

Trying to follow [kOps Getting Started](https://kops.sigs.k8s.io/getting_started/aws/) with cilium enabled

```
kops create cluster --name=${NAME} --cloud=aws --zones=us-west-1a --discovery-store=s3://xxx-xxx-xxx/${NAME}/discovery
```

cilium-agent running on worker nodes reports that the health check towards control-plane nodes fails.

```
$ cilium status --verbose
  Name                              IP                Node          Endpoints
  i-018a7b5c13a2332ae               10.4.4.191        unreachable   reachable
  i-08674cf4137ca05a5               10.4.5.111        unreachable   reachable
  i-0a2ccc6a8bc4dc881               10.4.6.163        unreachable   reachable
  
  
$ cilium-health status
  [output omitted]
  i-08674cf4137ca05a5:
    Host connectivity to 10.4.5.111:
      ICMP to stack:   Connection timed out
      HTTP to agent:   OK, RTT=298.557us
    Endpoint connectivity to 100.96.47.26:
      ICMP to stack:   OK, RTT=1.46466ms
      HTTP to agent:   OK, RTT=344.044us
```

However on cilium-agent running on control plane nodes, cilium health checks are working not matter which nodes the target is. **Basically, the AWS security group on control plane nodes does not allow inbound ICMP request from worker nodes.**

image

Kops should have configured the control plane Security group to allow inbound ICMP request from worker nodes.

 N/A

  N/A

N/A

maximumG:
Fixes #15329.

Cilium health check mechanism requires ICMP and TCP/4240 port to be opened between every nodes in the cluster, including control plane nodes. If one of these 2 checks fails the node is considered as unreachable.

We only open ICMP echo request (type 8) from node to master.

### DIFF
--- a/pkg/model/awsmodel/firewall.go
+++ b/pkg/model/awsmodel/firewall.go
@@ -196,13 +196,12 @@ func (b *FirewallModelBuilder) applyNodeToMasterBlockSpecificPorts(c *fi.Cloudup
 			}
 			for _, protocol := range protocols {
 				awsName := strconv.Itoa(int(protocol))
-				name := awsName
 
 				switch protocol {
 				case ProtocolIPIP:
-					name = "ipip"
+					awsName = "ipip"
 					t := &awstasks.SecurityGroupRule{
-						Name:          fi.PtrTo(fmt.Sprintf("node-to-master-protocol-%s%s", name, suffix)),
+						Name:          fi.PtrTo(fmt.Sprintf("node-to-master-protocol-%s%s", awsName, suffix)),
 						Lifecycle:     b.Lifecycle,
 						SecurityGroup: masterGroup.Task,
 						SourceGroup:   nodeGroup.Task,
@@ -210,9 +209,9 @@ func (b *FirewallModelBuilder) applyNodeToMasterBlockSpecificPorts(c *fi.Cloudup
 					}
 					AddDirectionalGroupRule(c, t)
 				case ProtocolICMP:
-					name = "icmp"
+					awsName = "icmp"
 					t := &awstasks.SecurityGroupRule{
-						Name:          fi.PtrTo(fmt.Sprintf("node-to-master-protocol-%s%s", name, suffix)),
+						Name:          fi.PtrTo(fmt.Sprintf("node-to-master-protocol-%s%s", awsName, suffix)),
 						Lifecycle:     b.Lifecycle,
 						SecurityGroup: masterGroup.Task,
 						SourceGroup:   nodeGroup.Task,
--- a/pkg/model/awsmodel/firewall.go
+++ b/pkg/model/awsmodel/firewall.go
@@ -31,6 +31,7 @@ type Protocol int
 
 const (
 	ProtocolIPIP Protocol = 4
+	ProtocolICMP Protocol = 1
 )
 
 // FirewallModelBuilder configures firewall network objects
@@ -135,9 +136,11 @@ func (b *FirewallModelBuilder) applyNodeToMasterBlockSpecificPorts(c *fi.Cloudup
 	udpRanges := []portRange{{From: 1, To: 65535}}
 	protocols := []Protocol{}
 
-	if b.Cluster.Spec.Networking.Cilium != nil && b.Cluster.Spec.Networking.Cilium.EtcdManaged {
-		// Block the etcd peer port
-		tcpBlocked[2382] = true
+	if b.Cluster.Spec.Networking.Cilium != nil {
+		protocols = append(protocols, ProtocolICMP)
+		if b.Cluster.Spec.Networking.Cilium.EtcdManaged {
+			tcpBlocked[2382] = true
+		}
 	}
 
 	if b.Cluster.Spec.Networking.Calico != nil {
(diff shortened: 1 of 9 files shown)


olemarkus:
Thanks. I think I am somewhat convinced it is practical to allow this traffic.

/ok-to-test

pkg/model/awsmodel/firewall.go
```diff
@@ -193,21 +196,33 @@ func (b *FirewallModelBuilder) applyNodeToMasterBlockSpecificPorts(c *fi.Cloudup
...
+					}
+					AddDirectionalGroupRule(c, t)
+				case ProtocolICMP:
+					name = "icmp"
+					t := &awstasks.SecurityGroupRule{
+						Name:          fi.PtrTo(fmt.Sprintf("node-to-master-protocol-%s%s", name, suffix)),
+						Lifecycle:     b.Lifecycle,
+						SecurityGroup: masterGroup.Task,
+						SourceGroup:   nodeGroup.Task,
+						FromPort:      fi.PtrTo(int64(8)),
```
olemarkus, line 218: nit: I think `net` may have some enums that can be used here instead of magic number 8

olemarkus (changes requested):

pkg/model/awsmodel/firewall.go
maximumG: You're right makes more sense to use the embedded golang ICMPType constants rather than the raw number.

olemarkus:
You probably need to run `./hack/update-expected.sh` to fix the tests.

tests/integration/update_cluster/minimal-ipv6-calico/kubernetes.tf
maximumG, line 1: weird to see a change in here, especially regading the protocol parameter, switching from a valid value 4 to -1 🤔 . Any idea why this may happened @olemarkus

maximumG (commented):
some calico related test are failing because the integration tests were modified and the ipip protocol switch from value 4 to -1.

maximumG:
@olemarkus is there any way I can help to finalize this PR ?

tests/integration/update_cluster/minimal-ipv6-cilium/kubernetes.tf
```diff
@@ -1029,6 +1029,15 @@ resource "aws_security_group_rule" "from-nodes-minimal-ipv6-example-com-ingress-
   type                     = "ingress"
 }
 
+resource "aws_security_group_rule" "from-nodes-minimal-ipv6-example-com-ingress-icmp-8to8-masters-minimal-ipv6-example-com" {
+  from_port                = 8
```
johngmyers, line 1033: ICMPv6 echo is not type 8.

johngmyers (changes requested):

tests/integration/update_cluster/privatecalico/kubernetes.tf
```diff
@@ -1134,21 +1134,21 @@ resource "aws_security_group_rule" "from-nodes-privatecalico-example-com-egress-
   type              = "egress"
 }
 
-resource "aws_security_group_rule" "from-nodes-privatecalico-example-com-ingress-4-0to0-masters-privatecalico-example-com" {
+resource "aws_security_group_rule" "from-nodes-privatecalico-example-com-ingress-all-0to0-nodes-privatecalico-example-com" {
   from_port                = 0
-  protocol                 = "4"
-  security_group_id        = aws_security_group.masters-privatecalico-example-com.id
+  protocol                 = "-1"
```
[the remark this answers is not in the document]
johngmyers: It's an artifact of the diff. Changing "4" to "ipip" caused that rule to sort after the "all-0to0-nodes" rule.

pkg/model/awsmodel/firewall.go
```diff
@@ -192,22 +196,33 @@ func (b *FirewallModelBuilder) applyNodeToMasterBlockSpecificPorts(c *fi.Cloudup
 			}
 			for _, protocol := range protocols {
 				awsName := strconv.Itoa(int(protocol))
-				name := awsName
+
 				switch protocol {
 				case ProtocolIPIP:
-					name = "ipip"
+					awsName = "ipip"
```
johngmyers, line 202: `ipip` is not a valid value for the `IpProtocol` field of an `IpPermission`.

> 	// The IP protocol name (tcp, udp, icmp, icmpv6) or number (see Protocol Numbers
	// (http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml)).

johngmyers:
Please test your change works on an IPv6 cluster.

bkmzio:
@maximumG, may I inquire if there's a chance to consolidate or merge the changes? Your input would be greatly appreciated!

PR #15340, opened 2023-04-20
update deps

zetaab:


go.mod
```diff
@@ -3,8 +3,9 @@ module k8s.io/kops
 go 1.20
 
 replace (
-	github.com/docker/cli => github.com/docker/cli v20.10.21+incompatible
-	github.com/docker/docker => github.com/docker/docker v20.10.21+incompatible
+	github.com/docker/cli => github.com/docker/cli v23.0.1+incompatible
+	github.com/docker/docker => github.com/docker/docker v23.0.1+incompatible
+	k8s.io/apiserver => k8s.io/apiserver v0.26.4
```
zetaab, line 8: the problem with newer library is that https://github.com/kubernetes/kops/blob/master/pkg/applylib/applyset/managedfields.go#L187 `DecodeManagedFields` function seems to be removed without deprecation? No idea is that moved to some another package or what is the case here.

zetaab:
cc @olemarkus @hakman

zetaab:
looks like sigs.k8s.io/controller-runtime needs older client-go (0.26.x) but other components needs newer. There are on-going PR https://github.com/kubernetes-sigs/controller-runtime/pull/2189 

I think we cannot proceed before that is done?

zetaab:
@olemarkus copied those functions from apimachinery/internal, but still https://github.com/kubernetes-sigs/controller-runtime/pull/2189 is the problem

zetaab:
cc @hakman @olemarkus it seems that builds are passing now. sigs.k8s.io/controller-runtime is not using release (because there are not yet new release for that) instead its using latest master commit. From my perspective we could merge this and make 1.27 beta?

hakman:
@zetaab I think it's a good idea. Will put it on the list for tomorrow.

hakman:
@zetaab I think https://github.com/kubernetes/kops/pull/15373 is a little cleaner, without holding back apiserver.

PR #15342, opened 2023-04-20
Switch to using external ECR credential provider for k8s 1.27

olemarkus:


nodeup/pkg/model/kubelet.go
```diff
@@ -413,6 +434,56 @@ func (b *KubeletBuilder) usesContainerizedMounter() bool {
 	}
 }
 
+// addECRCP installs the ECR credential provider
+func (b *KubeletBuilder) addECCRCP(c *fi.NodeupModelBuilderContext) error {
```
hakman, line 438:
```suggestion
func (b *KubeletBuilder) addECRCP(c *fi.NodeupModelBuilderContext) error {
```

pkg/apis/kops/cluster.go
```diff
@@ -220,6 +220,9 @@ type AWSSpec struct {
 	// Spotinst cloud-config specs
 	SpotinstProduct     *string `json:"spotinstProduct,omitempty"`
 	SpotinstOrientation *string `json:"spotinstOrientation,omitempty"`
+
+	// BinaryLocation is the location of the AWS cloud provider binaries.
+	BinaryLocation *string `json:"binaryLocation,omitempty"`
```
hakman, line 225: `BinaryLocation` seems a little vague to me.

olemarkus: Yeah I know. I hate it. But I couldn't find a good name for this.

FTR: the motivation behind this one is to use the PR/staging bucket in the AWS cloud provider e2e tests.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1056,6 +1056,23 @@ func (c *ApplyClusterCmd) addFileAssets(assetBuilder *assets.AssetBuilder) error
 			c.Assets[arch] = append(c.Assets[arch], mirrors.BuildMirroredAsset(u, hash))
 		}
 
+		if c.Cluster.UsesExternalECRCredentialsProvider() {
+			binaryLocation := c.Cluster.Spec.CloudProvider.AWS.BinaryLocation
+			if binaryLocation == nil {
+				binaryLocation = fi.PtrTo("https://artifacts.k8s.io/binaries/cloud-provider-aws/v1.27.1")
```
hakman, line 1062: Should `v1.27.1` be something like AWS CCM version? Not sure if we have access to that here.

olemarkus: Its a good catch. I was thinking of a helper method here where we feed in the K8s version and returns the cloud provider version. And move the case statement from CCM into that function. But I thought there wasn't a rush to do this right now.

hakman: No rush.

hakman (approved):

PR #15350, opened 2023-04-27
scaleway: doc: temporary env var to pull protokube

Mia-Cross:
Until the integration of #15230 in the next release of protokube, we specify in the doc that we have to use the environment variable KOPS_BASE_URL to pull the version from master.


Mia-Cross:
I'm doing a (small-scale) talk tonight where I will invite people to try out kops so I might have some users sooner than the next release, and I don't want to leave them without a solution to the problem. I'll write in my to-do list to remove this part as soon as the new release is available.

hakman (approved):

PR #15351, opened 2023-04-27
Block bootstrap when the node already exists

justinsb:
We now do this across all clouds, as it has been demonstrated on
OpenStack.

### DIFF
--- a/cmd/kops-controller/main.go
+++ b/cmd/kops-controller/main.go
@@ -45,6 +45,7 @@ import (
 	"k8s.io/kops/upup/pkg/fi/cloudup/hetzner"
 	"k8s.io/kops/upup/pkg/fi/cloudup/openstack"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/yaml"
 	// +kubebuilder:scaffold:imports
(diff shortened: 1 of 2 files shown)


hakman:
@justinsb `pull-kops-e2e-k8s-gce-ipalias` always running intentionally?

justinsb:

Seems very unrelated: `creating apiservice v1alpha1.wardle.example.com: apiservices.apiregistration.k8s.io "v1alpha1.wardle.example.com" already exists`

This is the brand-new job though, so could be real (though unlikely to be related to the PR itself)

hakman:
@justinsb Should this be removed from the OpenStack side?
/lgtm

justinsb:
> @justinsb Should this be removed from the OpenStack side?

I think it should, but I thought it was harmless to double-check it first.  I can send a separate PR to @zetaab to approve, maybe that's the best way?

zetaab:
@justinsb there is one possible problem: if you restart the node itself and after it will come back up, it will try to re-register itself if I remember correctly. We should think how it should behave in that case. Nodes are not restarting often but anyway it should work somehow.

justinsb:
> if you restart the node itself and after it will come back up, it will try to re-register itself if I remember correctly.

Thanks for reminding me!  I think what we should do is to try to save keys to disk and reuse them.  (I think we're saving them to disk anyway).  I don't think we're reusing them today, but I don't think it's too hard, and with any luck, that should mean we don't need to issue the bootstrap request again.

But I will double-check this scenario.  If it doesn't work in general, we could create a "machine key" that we issue as normal, but then we can use this key if the node does need to do reauthentication.  We were talking about whether we should introduce a machine key anyway, but it's more complexity because we probably have to store that machine key and it isn't obvious where to store it.  It would probably have to be a new CRD, because the node likely doesn't exist until after this call completes and the kubelet eventually registers.  Anyway, @hakman and I were hoping to get away without it, but good call on testing node restart!

hakman:
OK, let's iterate in separate PRs.
/approve

PR #15355, opened 2023-04-30
Spotinst: update docs about setting the VNG Size Limits in Launch Spec

IdanShohamNetApp:
This PR update spot-ocean doc about setting support for VNG size limits in launch spec.


IdanShohamNetApp:
a new PR regarding this change will be open

PR #15358, opened 2023-05-02
update runc to 1.1.7

heybronson:
Bumping runc to 1.1.7 to resolve GPU issues described in https://github.com/NVIDIA/nvidia-docker/issues/1671#issuecomment-1526913755

### DIFF
--- a/pkg/model/components/containerd.go
+++ b/pkg/model/components/containerd.go
@@ -49,7 +49,7 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 			if b.IsKubernetesGTE("1.23") {
 				containerd.Version = fi.PtrTo("1.6.20")
 				containerd.Runc = &kops.Runc{
-					Version: fi.PtrTo("1.1.4"),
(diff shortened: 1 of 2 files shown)


pkg/model/components/containerd.go
```diff
@@ -49,7 +49,7 @@ func (b *ContainerdOptionsBuilder) BuildOptions(o interface{}) error {
 			if b.IsKubernetesGTE("1.23") {
 				containerd.Version = fi.PtrTo("1.6.20")
 				containerd.Runc = &kops.Runc{
-					Version: fi.PtrTo("1.1.4"),
+					Version: fi.PtrTo("1.1.7"),
```
hakman, line 52: @heybronson Could you try setting the default to 1.1.5?
I think there's something broken with the latest releases of runc. We should probably create an issue with https://github.com/opencontainers/runc.

```suggestion
					Version: fi.PtrTo("1.1.5"),
```

heybronson: Will do, thanks for the response!

hakman (approved):

PR #15367, opened 2023-05-04
DigitalOcean support for node bootstrap

justinsb:
Update DigitalOcean to support bootstrapping nodes through kops-controller.

This means we no longer need to put the DO credentials onto the nodes.

### DIFF
--- a/cmd/kops-controller/main.go
+++ b/cmd/kops-controller/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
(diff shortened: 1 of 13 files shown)


justinsb:
/retest

The AWS failure (e.g. `cannot determine challenge endpoint for instance id: i-003af7b182b5c840f`) looks like it might be real, but we should be handling ipv6...

justinsb:
Hmmm... I guess it was a flake after all, which makes sense - we shouldn't be affecting the AWS or GCE paths!

hakman (approved):

PR #15373, opened 2023-05-06
Update dependencies to K8s v1.27

hakman:


pkg/applylib/applyset/managedfields.go
```diff
@@ -184,15 +184,10 @@ func mergeManagedFields(managedFields []metav1.ManagedFieldsEntry) (*fieldpath.S
 
 // toFieldPathSet converts an encoded ManagedFieldsEntry to a set of managed fields (a fieldpath.Set)
 func toFieldPathSet(fields *metav1.ManagedFieldsEntry) (*fieldpath.Set, error) {
-	decoded, err := fieldmanager.DecodeManagedFields([]metav1.ManagedFieldsEntry{*fields})
-	if err != nil {
+	// https://github.com/kubernetes/kubernetes/blob/v1.27.1/staging/src/k8s.io/client-go/util/csaupgrade/upgrade.go#L310-L315
+	var fieldSet fieldpath.Set
+	if err := fieldSet.FromJSON(bytes.NewReader(fields.FieldsV1.Raw)); err != nil {
```
hakman, line 189: @justinsb Does this look right to you?

justinsb: That looks right to me.  Previously we used fieldmanager.DecodeManagedFields which did some version checks, but I don't know if they're really necessary today.

@apelisse what we're actually doing here is merging two field managers, to transition from kubectl to server-side-apply management of fields.  I recall you saying that functionality may be available somewhere as a function, rather than us having to do the low-level stuff we're doing here?

apelisse: I've left this function if you want to make the various checks:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/managedfields/fieldmanager.go#L54-L57

Please take a look at that PR: https://github.com/kubernetes/kubernetes/pull/112905/, code available here: staging/src/k8s.io/client-go/util/csaupgrade/upgrade.go. Let me know if that doesn't solve your problem!

PR #15393, opened 2023-05-10
scaleway: dns support

Mia-Cross:
This PR adds support for DNS in Scaleway clusters.

[OUTDATED]
As it is, the integration tests can only be run by me because of the 2 constant variables : 
- `validDNSZone` is a domain name on which I have the rights.
- `validScalewayProfileName` is used to load my credentials from a local file in order to interact with it.
I'm not really sure what I should put instead for anybody to be able to run the tests, so I'll take any suggestion.

### DIFF
--- a/dnsprovider/pkg/dnsprovider/providers/scaleway/dns.go
+++ b/dnsprovider/pkg/dnsprovider/providers/scaleway/dns.go
@@ -108,7 +108,7 @@ type zones struct {
 func (z *zones) List() ([]dnsprovider.Zone, error) {
 	dnsZones, err := z.domainAPI.ListDNSZones(&domain.ListDNSZonesRequest{}, scw.WithAllPages())
 	if err != nil {
-		return nil, fmt.Errorf("failed to list DNS zones: %v", err)
+		return nil, fmt.Errorf("failed to list DNS zones: %w", err)
 	}
 
 	zonesList := []dnsprovider.Zone(nil)
@@ -210,8 +210,9 @@ func (r *resourceRecordSets) List() ([]dnsprovider.ResourceRecordSet, error) {
 		// The scaleway API returns the record without the zone
 		// but the consumers of this interface expect the zone to be included
 		recordName := dns.EnsureDotSuffix(record.Name) + r.Zone().Name()
-		if rrset, ok := rrsetsWithoutDups[recordName]; !ok {
-			rrsetsWithoutDups[recordName] = &resourceRecordSet{
+		recordKey := recordName + "_" + record.Type.String()
+		if rrset, ok := rrsetsWithoutDups[recordKey]; !ok {
+			rrsetsWithoutDups[recordKey] = &resourceRecordSet{
 				name:       recordName,
 				data:       []string{record.Data},
 				ttl:        int(record.TTL),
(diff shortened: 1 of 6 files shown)


hakman:
@Mia-Cross Unrelated to this PR, is there a plan to add support for `--dns=none`?

Mia-Cross:
@hakman Yes, I plan to do it soon !

hakman:
kOps integration tests are based on mock clients. Not all clouds implement those though.
See: https://github.com/kubernetes/kops/tree/master/cloudmock

Not quite sure how to proceed here without the mock client, except by removing the integration tests and adding them back in a later PR, when a mock client is available.

dnsprovider/pkg/dnsprovider/providers/scaleway/dns.go
```diff
@@ -0,0 +1,456 @@
...
+			})
+		}
+	}
+
+	if len(r.upserts) > 0 {
+		for _, rrset := range r.upserts {
+			for _, rrdata := range rrset.Rrdatas() {
+				for _, record := range records {
+					recordNameWithZone := fmt.Sprintf("%s.%s.", record.Name, r.zone.Name())
+					if recordNameWithZone == dns.EnsureDotSuffix(rrset.Name()) {
```
justinsb, line 382: Do we want to check the type here also?

dnsprovider/pkg/dnsprovider/providers/scaleway/dns.go
```diff
@@ -0,0 +1,456 @@
...
+				}
+			}
+		}
+	}
+
+	if len(r.removals) > 0 {
+		for _, rrset := range r.removals {
+			for _, record := range records {
+				recordNameWithZone := fmt.Sprintf("%s.%s.", record.Name, r.zone.Name())
+				if recordNameWithZone == dns.EnsureDotSuffix(rrset.Name()) && record.Data == rrset.Rrdatas()[0] {
```
justinsb, line 407: And do we want to check the type here also?

dnsprovider/pkg/dnsprovider/providers/scaleway/dns_test.go
```diff
@@ -0,0 +1,301 @@
...
+		t.Errorf("unexpected zone name: %s", outZone.Name())
+	}
+}
+
+func TestAddShouldFail(t *testing.T) {
+	domainAPI := setUpFakeZonesNil()
+	zs := getDNSProviderZones(domainAPI)
+
+	inZone := &zone{
+		name:      "dns.test.fr",
```
justinsb, line 120: Nit: We normally use example.com - it's reserved for tests/examples: https://en.wikipedia.org/wiki/Example.com

(Maybe test.fr is also though?)

Not a blocker!

justinsb:
This looks great - thank you @Mia-Cross 

One question about whether the key of DNS records should be the record name + record type, that is how the DNS works on most other clouds.  Other than that, just a few non-blocking nits.

Mia-Cross:
@justinsb I've applied the recommended changes. About the record type, I had only tested with type A records so I didn't think of this but it seems like a good idea indeed to include the type of record in the key. Is there anywhere else where I should make the change ?

justinsb:
Thanks @Mia-Cross  - this looks great!  I'm not aware of any other places that the record type is not included (at least any place where it is significant, as we do mostly use A and AAAA records!)

/approve
/lgtm

vendor/github.com/scaleway/scaleway-sdk-go/api/domain/v2beta1/domain_sdk.go
```diff
@@ -0,0 +1,3916 @@
...
+	_ = strings.Join
+
+	_ scw.ScalewayRequest
+	_ marshaler.Duration
+	_ scw.File
+	_ = parameter.AddToQuery
+	_ = namegenerator.GetRandomName
+)
+
+// API: domains and DNS API documetation.
```
hakman, line 42:
```suggestion
// API: domains and DNS API.
```

vendor/github.com/scaleway/scaleway-sdk-go/api/domain/v2beta1/domain_sdk.go
```diff
@@ -0,0 +1,3916 @@
...
+}
+
+// NewAPI returns a API object from a Scaleway client.
+func NewAPI(client *scw.Client) *API {
+	return &API{
+		client: client,
+	}
+}
+
+// RegistrarAPI: domains registrar API.
```
hakman, line 55:
```suggestion
// RegistrarAPI: domains and DNS - Registrar API.
```

hakman (changes requested):
I think `make gomod` is needed.

PR #15395, opened 2023-05-10
add removeAll to vfs

Codelax:
Adds `RemoveAll` to vfs interface.
The goal of this method is to avoid deleting files one by one if the implementation allows it.
When deleting a cluster that ran for a while, it can take a very long time to delete all backups.

I personally use s3 as state-store, I tried to implement it for other providers when I could find useful methods in their APIs. For other providers I resolved to use a loop, it does not change from the current implementation and it could be improved later on.

For the swift implementation, I used the default 10k bulk items but I'm not sure about this one as this seems to be configurable per swift instance. 

I also found partially the same code with `RemoveAllVersion` although this only target object versions/snapshots. But this one does not seems to be used elsewhere in the code.

### DIFF
--- a/pkg/client/simple/vfsclientset/clientset.go
+++ b/pkg/client/simple/vfsclientset/clientset.go
@@ -175,27 +175,9 @@ func DeleteAllClusterState(basePath vfs.Path) error {
 		return fmt.Errorf("refusing to delete: unknown file found: %s", path)
 	}
 
-	for _, path := range paths {
-		err = path.Remove()
-		if err != nil {
-			return fmt.Errorf("error deleting cluster file %s: %v", path, err)
-		}
-	}
-
-	return nil
-}
-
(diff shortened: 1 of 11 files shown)


hakman:
@Codelax Could you try using `RemoveAllVersions()`? It is used by etcd-manager.

Codelax:
@hakman Yes I have thought of that but it seems like a drastic change to go from soft delete to hard delete.
If one use a versioned bucket to store state and backups it would be a breaking change.

justinsb:
Thanks @Codelax - this looks like a great change.

One tip for the future, I think if you're not personally using OpenStack/Swift I would have just left it as the no-change implementation.    Then we can let whoever uses OpenStack/Swift implement and verify it in their scenario.

That said, I think the changes look good, so I'm going to approve!

Thanks for the contribution :-)

/approve
/lgtm

PR #15399, opened 2023-05-11
do not mount same dir twice

zetaab:
I am trying the current master and the error is

```
syslog:May 11 07:52:10 master-helpa-6hqpcj kubelet[3235]: E0511 07:52:10.874853    3235 file.go:187] "Could not process manifest file" err="invalid pod: [spec.containers[0].volumeMounts[14].mountPath: Invalid value: \"/srv/kubernetes/kube-apiserver\": must be unique]" path="/etc/kubernetes/manifests/kube-apiserver.manifest"
```

full kube-apiserver spec https://gist.github.com/zetaab/bb55d42b4a998a9ee7695b4e98d40274

like can be seen `/srv/kubernetes/kube-apiserver` is two times there. We have been using following kops spec for years:

```
    auditPolicyFile: /srv/kubernetes/kube-apiserver/policy.yaml
    auditWebhookConfigFile: /srv/kubernetes/kube-apiserver/audit.yaml
```

but now it looks like it will break the apiserver because it will try to mount the same folder twice.

```
    - mountPath: /srv/kubernetes/kube-apiserver
      name: srvkapi
      readOnly: true
    - mountPath: /srv/kubernetes/kube-apiserver
      name: auditconfigdir
      readOnly: true
...
  - hostPath:
      path: /srv/kubernetes/kube-apiserver
    name: srvkapi
  - hostPath:
      path: /srv/kubernetes/kube-apiserver
    name: auditconfigdir

### DIFF
--- a/nodeup/pkg/model/kube_apiserver.go
+++ b/nodeup/pkg/model/kube_apiserver.go
@@ -723,7 +723,9 @@ func (b *KubeAPIServerBuilder) buildPod(ctx context.Context, kubeAPIServer *kops
 	if kubeAPIServer.AuditPolicyFile != "" {
 		// The audit config dir will be used for both the audit policy and the audit webhook config
 		auditConfigDir := filepath.Dir(kubeAPIServer.AuditPolicyFile)
-		kubemanifest.AddHostPathMapping(pod, container, "auditconfigdir", auditConfigDir)
+		if pathSrvKAPI != auditConfigDir {
+			kubemanifest.AddHostPathMapping(pod, container, "auditconfigdir", auditConfigDir)
+		}
 	}
 
 	if b.NodeupConfig.APIServerConfig.Authentication != nil {


zetaab:
cc @hakman can you review

hakman:
@zetaab Could we just return early from `AddHostPathMapping()` if the path is already mapped?

zetaab:
Screenshot 2023-05-11 at 14 34 39


well I do not know does this look awful.

zetaab:
/hold 

lets see could we get better solution after https://github.com/kubernetes/kops/pull/15406

hakman (approved):

math3vz:
@hakman @zetaab can we backport this to 1.26? It looks like it only made it to 1.27 and above

hakman:
@math3vz I don't think there's any planned release for 1.26 anymore. Why is it not possible to use 1.27+?

PR #15405, opened 2023-05-11
scaleway: tag resources with cloud labels

Mia-Cross:
In this PR, we make sure that every resource (that supports tags) gets tagged with the cloud labels at creation. 
For now only load-balancers had this feature.


hakman:
@Mia-Cross Please run `hack/update-expected`

hakman (approved):

PR #15406, opened 2023-05-11
nodeup: Use functional options pattern for HostPathMapping

justinsb:
This means that the object is not mutated after construction, making
it easier to do validity checks (such as whether we have mounted the
same path twice).


zetaab:
@justinsb can you check PR https://github.com/kubernetes/kops/pull/15399 it basically modifies almost same code and its existing bug.

hakman:
> Does this PR solve that issue with duplicate items mounting same path?

This PR should make it easier to chec for duplicate mounts in a generic way.

hakman (approved):

Issue #15402, opened 2023-05-11
replace references of https://storage.googleapis.com/kubernetes-release with https://dl.k8s.io


rjsadow:
## Problem:

There are still references to https://storage.googleapis.com/kubernetes-release instead of https://dl.k8s.io

dl.k8s.io is the correct advertised download host and will eventually move to be fastly shielding a fully community-owned bucket

ref: https://github.com/kubernetes/k8s.io/issues/2396

## Proposed Solution:

Replace https://storage.googleapis.com/kubernetes-release with https://dl.k8s.io

Also replace any gs://kubernetes-release references with using https://dl.k8s.io/ if applicable.

/kind bug
/assign

PR #15408, opened 2023-05-11
replace references of https://storage.googleapis.com/kubernetes-release with https://dl.k8s.io

KlwntSingh:
fixes: https://github.com/kubernetes/kops/issues/15402


upup/pkg/fi/cloudup/defaults.go
```diff
@@ -167,10 +167,10 @@ func ensureKubernetesVersion(c *kops.Cluster) error {
 }
 
 // FindLatestKubernetesVersion returns the latest kubernetes version,
-// as stored at https://storage.googleapis.com/kubernetes-release/release/stable.txt
+// as stored at https://dl.k8s.io/release/stable.txt
 // This shouldn't be used any more; we prefer reading the stable channel
 func FindLatestKubernetesVersion() (string, error) {
-	stableURL := "https://storage.googleapis.com/kubernetes-release/release/stable.txt"
+	stableURL := "https://dl.k8s.io/release/stable.txt"
```
hakman, line 173: Not quite sure what's wrong here:
```
% curl https://storage.googleapis.com/kubernetes-release/release/stable.txt
v1.27.1

% curl https://dl.k8s.io/release/stable.txt                                
<html>
<head><title>302 Found</title></head>
<body>
<center><h1>302 Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

rjsadow: Try adding a `-L` to your curl

rjsadow: We should ensure redirects are used anywhere `stableURL` is referenced.

rifelpet:
It looks like you need to run ./hack/update-expected.sh to pick up more integration test changes.

We'll also want to update our e2e tests to reference the new location too: https://github.com/kubernetes/test-infra/blob/2326b828d25f04c676703ceda0782a3eed596a12/config/jobs/kubernetes/kops/helpers.py#L78

upup/pkg/fi/cloudup/defaults.go
hakman: Thanks @rjsadow. 😁

hakman (approved):

PR #15412, opened 2023-05-12
Use registry.k8s.io for legacy addons

Nunuscloud:
In AWS, we use a Network load balancer (NLB) to expose the Ingress-Nginx Controller behind a Service of Type=LoadBalancer.


hakman:
These addons have been deprecated a long time ago and are only kept for compatibility reasons with very old kOps versions:

PR #15413, opened 2023-05-12
Use registry.k8s.io for legacy addons

Nunuscloud:
The provided templates illustrate the setup for legacy in-tree service load balancer for AWS NLB. 
The v1.6.0 version that was previously uploaded is legacy and does not run. We recommend that you use that version.


hakman:
These addons have been deprecated a long time ago and are only kept for compatibility reasons with very old kOps versions:

PR #15414, opened 2023-05-12
OpenStack Octavia FlavorID

sn4psh0t:
Support for OpenStack Octavia FlavorID parameter

### DIFF
--- a/upup/pkg/fi/cloudup/openstack/cloud.go
+++ b/upup/pkg/fi/cloudup/openstack/cloud.go
@@ -315,7 +315,6 @@ type openstackCloud struct {
 	zones           []string
 	floatingEnabled bool
 	useVIPACL       *bool
-	flavorID        *string
 }
 
 var _ fi.Cloud = &openstackCloud{}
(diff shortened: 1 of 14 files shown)


zetaab:
@sn4psh0t could you run `./hack/update-expected.sh` and commit changes to own commit

upup/pkg/fi/cloudup/openstack/cloud.go
```diff
@@ -496,6 +497,9 @@ func buildLoadBalancerClient(c *openstackCloud, spec *kops.OpenstackSpec, provid
 			return fmt.Errorf("error building lb client: %w", err)
 		}
 		lbClient = client
+		if spec.Loadbalancer.FlavorID != nil {
```
zetaab, line 500: hmm is these 3 rows really needed? I do not see where c.flavorID is used?

zetaab (changes requested):

upup/pkg/fi/cloudup/openstack/cloud.go
sn4psh0t: @zetaab you're right: openstackCloud struct should not be influenced here, sorry for the oversight. Can I push another commit reverting the changes to cloud.go back?

zetaab: Just make new commit to remove that

sn4psh0t: done @zetaab

zetaab (approved):
/approve
/lgtm

Thank you!

PR #15415, opened 2023-05-12
Add a mutex lock to 'awsCloudInstances' map

fcgravalos:
We're using terraform kops provider to manage our AWS kops clusters. Time to time we hit a race condition with the stack trace points to `awsup.NewAWSCloud` function, when writing to a concurrent map, that maintains a map between regions and `AWSCloud` objects.

This PR changes this to variable so it belongs to a new type, that wraps the map into its own struct where access is controlled by a mutex lock.

Let me know if that makes sense to you all.

Thanks for building this awesome project!

### DIFF
--- a/upup/pkg/fi/cloudup/awsup/aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/aws_cloud.go
@@ -236,10 +236,23 @@ func (c *awsCloudImplementation) Region() string {
 	return c.region
 }
 
-var awsCloudInstances map[string]AWSCloud = make(map[string]AWSCloud)
+type awsCloudInstancesRegionMap struct {
+	mutex     sync.Mutex
+	regionMap map[string]AWSCloud
+}
+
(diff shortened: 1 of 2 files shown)


fcgravalos:
This is a piece of the stack trace of our terraform run:

image

fcgravalos:
@rifelpet would you mind re-triggering tests again? Thanks!

fcgravalos:
@hakman made another change to make sure concurrent access to the `ResetAWSCloudInstances` is safe too, I believe this now requires again the /ok-to-test command? sorry about that.

fcgravalos:
it seems like all the ci checks passed! :rocket:

fcgravalos:
Hey @hakman @zetaab ! wdyt? Does it look like a reasonable PR?

hakman:
I think mutexes need to be used both in write and read places, not just write.

fcgravalos:
@hakman added locking on reads too

justinsb:
Thanks @fcgravalos 

/label tide/merge-method-squash

/approve
/lgtm

PR #15419, opened 2023-05-16
use registry.k8s.io for cinder csi

zetaab:


zetaab:
@hakman this was missing from the another PR where I updated the CCM

hakman (approved):

PR #15422, opened 2023-05-16
scaleway: none DNS option available

Mia-Cross:
This PR adds the possibility for Scaleway clusters to use no DNS.


justinsb:
I do think it's OK to merge this as-is, but yes if we can have the callback-check on scaleway also that would be a great bonus!

hakman:
@Mia-Cross I don't mind either way, so your choice if you want to add now or later.

PR #15423, opened 2023-05-16
OpenStack: Set FloatingIP support if router spec exists

### LINKED ISSUE #15401: OpenStack: kOps cannot create router when LB support is disabled
/kind bug

 this information.**

```
Client version: 1.26.3 (git-v1.26.3)
```

 a `kops` flag.**

```
1.26.x
```

`openstack`

```
kops update cluster --yes
```

Configure a cluster where kOps has to provision the network by itself but without letting kOps provision an LB; excerpt:

```
kind: Cluster
metadata:
  name: some-test-cluster.k8s.local
spec:
  api: {}
  cloudConfig:
    openstack
      router:
        externalNetwork: <name or id of extnet>
  cloudProvider: openstack
  # do not set to let kOps provision network
  # networkID: <TBD>
```

```
goroutine 325 [running]:
k8s.io/kops/upup/pkg/fi/cloudup/openstack.(*openstackCloud).GetExternalNetwork(0x140014afa40?)
        k8s.io/kops/upup/pkg/fi/cloudup/openstack/network.go:154 +0x20
k8s.io/kops/upup/pkg/fi/cloudup/openstacktasks.(*Router).RenderOpenstack(0x5?, 0x14000d38170, 0x14001054d80?, 0x14001716540, 0x2?)
        k8s.io/kops/upup/pkg/fi/cloudup/openstacktasks/router.go:103 +0x320
```

I expected kOps to provision the cluster and its infrastructure - network, router, nodes, etc.

see above

-

The problem seems to be that when kOps does not manage the LBs it will not set the `openstackCloud.extNetworkName`:

https://github.com/kubernetes/kops/blob/3b5c99d3c5025b604d426a7f0c79d066ce2ad755/upup/pkg/fi/cloudup/openstack/cloud.go#L450-L466

This leads to the the nil pointer here:

https://github.com/kubernetes/kops/blob/3b5c99d3c5025b604d426a7f0c79d066ce2ad755/upup/pkg/fi/cloudup/openstack/network.go#L153-L155

This was probably introduced with https://github.com/kubernetes/kops/pull/14742 which mitigated some issues of https://github.com/kubernetes/kops/issues/14546.

The solution to the bug would be to set the `extNetworkName` even if LB support is not wanted, so that the router can be created.

ederst:
With this change the external network info will be set, even if there is no loadbalancer support enabled. Otherwise this leads to an error when kOps creates a network with router.

fixes #15401

### DIFF
--- a/upup/pkg/fi/cloudup/openstack/cloud.go
+++ b/upup/pkg/fi/cloudup/openstack/cloud.go
@@ -435,6 +435,7 @@ func buildClients(provider *gophercloud.ProviderClient, tags map[string]string, 
 		useOctavia:    false,
 	}
 
+	setFloatingIPSupport(c, spec)
 	err = buildLoadBalancerClient(c, spec, provider, region)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build load balancer client: %w", err)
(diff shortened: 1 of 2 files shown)


zetaab (approved):
/approve
/lgtm

PR #15436, opened 2023-05-21
Update etcd to v3.5.9

hakman:
https://github.com/kubernetes/kubernetes/pull/115733 drops `cp` from `etcd` container images, which kOps was using to copy the binaries from init containers. To fix that, `kops-copy` was born.

https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.5.md


olemarkus (approved):

hakman:
Thanks @olemarkus. Let's merge and iterate.
/hold cancel

pkg/model/components/etcdmanager/model.go
```diff
@@ -249,9 +260,11 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
 			initContainer := v1.Container{
 				Name:    "init-etcd-" + strings.ReplaceAll(etcdVersion, ".", "-"),
 				Image:   etcdSupportedImages[etcdVersion],
-				Command: []string{"/bin/sh"},
+				Command: []string{"/opt/bin/kops-copy"},
 				Args: []string{
-					"-c",
+					"/usr/local/bin/etcd",
```
justinsb, line 265: Nice - this is so much simpler!

pkg/model/components/etcdmanager/options.go
```diff
@@ -69,7 +69,7 @@ var etcdSupportedImages = map[string]string{
 	"3.2.24": "registry.k8s.io/etcd:3.2.24-1",
 	"3.3.17": "registry.k8s.io/etcd:3.3.17-0",
 	"3.4.13": "registry.k8s.io/etcd:3.4.13-0",
-	"3.5.7":  "registry.k8s.io/etcd:3.5.7-0",
+	"3.5.9":  "registry.k8s.io/etcd:3.5.9-0",
```
justinsb, line 72: Ah - maybe we never shipped 3.5.7?

cmd/kops-copy/main.go
```diff
@@ -0,0 +1,81 @@
...
+		return fmt.Errorf("unable to open source file %q: %w", source, err)
+	}
+	defer sf.Close()
+
+	fi, err := sf.Stat()
+	if err != nil {
+		return fmt.Errorf("unable to stat source file %q: %w", source, err)
+	}
+
+	fn := path.Join(target, path.Base(source))
```
justinsb, line 43: Nit: technically we should probably be using filepath when they are file paths, so it's one less thing to fix for windows...

pkg/model/components/etcdmanager/model.go
```diff
@@ -201,6 +200,18 @@ spec:
...
+    - /ko-app/kops-copy
+    - /opt/bin
+    command:
+    - /ko-app/kops-copy
+    image: registry.k8s.io/kops/kops-copy:1.27.0-alpha.2
+    name: kops-copy
+    resources: {}
+    volumeMounts:
+    - mountPath: /opt
+      name: bin
```
justinsb, line 214: Maybe this should be `kops-copy` to avoid conflicts (in case an image has an /opt/bin) & to "leave room" for copying other images?

justinsb:
Looks really great.  A few nits/suggestions, but nothing blocking, so will approve & hold.  Feel free to remove hold unless you want to address any of the nits (in this PR!)

/approve
/lgtm
/hold

pkg/model/components/etcdmanager/model.go
hakman: We use this volume for all binaries, including etcd ones. Probably naming it `opt` would be better.

PR #15441, opened 2023-05-23
Default to dns=none for k8s.local clusters, on a few clouds

justinsb:
For DigitalOcean, Hetzner and GCE (initially), we treat k8s.local as
dns=none, instead of using the heavier gossip mechanism.

### DIFF
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -902,6 +902,15 @@ func (c *Cluster) UsesNoneDNS() bool {
 	if c.Spec.Networking.Topology != nil && c.Spec.Networking.Topology.DNS == DNSTypeNone {
 		return true
 	}
+
+	if dns.IsGossipClusterName(c.Name) {
+		// Default implementation of k8s.local clusters is now "dns=none" for a few clouds
+		switch c.Spec.GetCloudProvider() {
+		case CloudProviderDO, CloudProviderHetzner, CloudProviderGCE:
+			return true
+		}
+	}
+
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


justinsb:
cc @hakman - I think we should roll out your support for dns=none as the default implementation for gossip on a few clouds.  I think we should probably do it for all clouds, but I figure this is a good place to start.

One issue - you can see in the second commit that `kops create cluster` now switches from dns=private to dns=none for a hetzner cluster with name minimal.k8s.local.  I think that is correct though.

I do wonder if this line in setupDNSTopology should be `else if cluster.UsesNoneDNS()`, but I'll send a separate PR for that I think: https://github.com/kubernetes/kops/blob/7668eb41954e6fa7763adf815b96be3cd9379eba/upup/pkg/fi/cloudup/new_cluster.go#L1328

hakman:
@justinsb The problem with switching to `--dns=none` for existing clusters is that they will break. @zetaab managed to migrate from gossip to none DNS, but I don't remember being seamless.

zetaab:
yeah I have migrated something like 100+ OpenStack kOps clusters from gossip to none dns. 

Steps:

1. kops edit cluster 

from
```
  topology:
    dns:
      type: Private (or Public)
```

to
```
  topology:
    dns:
      type: None
```

2. `kubectl delete deployment dns-controller -n kube-system`
3. `kops update cluster --yes`
4. `kops rolling-update cluster --yes`

However, quite often the rolling updates failed to issue with kops-controller. Kops-controller in some master went to crashloopbackoff. This usually happened after first or first two masters were updated to none dns configuration. It was quite easy to solve: we deleted always the master instance which was containing the kops-controller which is not behaving well. After that we executed  `kops update cluster --yes` which created the master again with latest none dns configuration and it started to work.

The migration is not that smooth, but if you do it once your life will be easier after that (and normal nodes does not have any secrets)

zetaab:
if we want get rid off gossip, we could perhaps ignore kops-controller crashloop in rolling-update with some flag during migration? Then at least in openstack it will work without issues.

hakman:
/close

PR #15453, opened 2023-05-27
kops-controller: Return `http.StatusConflict` when node already exists

hakman:
As @justinsb commented in https://github.com/kubernetes/kops/pull/15138#discussion_r1107168105, `kops-controller` should return `http.StatusConflict` when node already exits.

Also, for reference, https://github.com/kubernetes/kops/pull/15351#issuecomment-1527644755.

Due to this mismatch, some tests are failing for Karpenter:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/15144/pull-kops-e2e-aws-karpenter/1662005360124235776

/cc @zetaab @justinsb


zetaab (approved):

hakman:
Thanks @zetaab. https://github.com/kubernetes/kops/pull/15456 includes this too, so closing...

PR #15456, opened 2023-05-27
kops-controller: Return `http.StatusConflict` only when node is ready

hakman:
/cc @justinsb @zetaab 
/assign @justinsb 
/hold


zetaab (approved):
remove hold when needed, looks lgtm from my perspective

PR #15458, opened 2023-05-31
Support intstr.IntOrString type  in cmd

peaaceChoi:
When using the **edit instancegroup with --set** option , a parsing error occurs if it is not a primitive type. The **instance group** struct has a custom type, and supporting it makes using cmd more comfortable.

e.g)
When trying to **updatespec.rollingUpdate.maxUnavailable** with edit instancegroup, a type error occurs because maxUnavailable is of type **intstr.IntOrString**.

```
./kops edit instancegroup --set 'spec.rollingUpdate.maxUnavailable=100%'
=> Unhandled type...
```

This PR supports processing the **intstr.IntOrString** type.

### DIFF
--- a/util/pkg/reflectutils/access.go
+++ b/util/pkg/reflectutils/access.go
@@ -21,6 +21,8 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 func SetString(target interface{}, targetPath string, newValue string) error {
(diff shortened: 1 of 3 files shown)


util/pkg/reflectutils/access.go
```diff
@@ -159,6 +161,9 @@ func setPrimitive(v reflect.Value, newValue string) error {
 		default:
 			panic("missing case in switch")
 		}
+
```
hakman, line 164: Additional space not liked by gofmt test.
```suggestion

```

hakman:
Interesting approach. Thanks for the PR @peaaceChoi.
/lgtm
/cc @justinsb

PR #15460, opened 2023-06-01
Default to dns=none for hetzner and digitalocean

justinsb:
For new clusters only, default to dns=none on hetzner and
digitalocean.

Hetzner previously used dns=private (with the same behaviour), but
dns=none is more specific.

DigitalOcean previously defaulted to peer-to-peer gossip, but dns=none
is simpler and requires fewer permissions.


hakman (approved):

hakman:
@justinsb we should probably default to none for new clusters for all could providers.

PR #15466, opened 2023-06-03
Add additional config for node local dns

zetaab:


zetaab:
no idea why that upgrade test fails

hakman:
> no idea why that upgrade test fails

Unrelated, etcd upgrade fails.

hakman:
@justinsb The only question here is which is preferred `ExtraConfig` or `AdditionalConfig`?

hakman:
@zetaab Based on the chat I had with @justinsb yesterday, I think it would be preferable to use `AdditionalConfig` instead of  `ExtraConfig`. Does it sound reasonable to you?

zetaab:
yeah I can change that

zetaab:
blah, I need to fix this with time. rebase did not went quite well

zetaab:
fixed

hakman (approved):

PR #15470, opened 2023-06-05
fix(cilium): cilium-operator must be able to patch K8S nodes resources

### LINKED ISSUE #15464: cilium-operator is missing RBAC permission to remove `node.cilium.io/agent-not-ready` taint
/kind bug

 this information.**

1.26

 a `kops` flag.**

1.25.9

AWS

1. Add the `node.cilium.io/agent-not-ready` taint to a kops instance group
2. Enable cilium `debug` log switch in the kops config
3. Check logs from the cilium-operator which is stating that it cannot patch the nodes resources

```
level=debug msg="Removing Node Taint" nodeName=i-0de0ebddfb1f1a317 subsys=watchers taint=node.cilium.io/agent-not-ready
level=debug msg="Controller func execution time: 1.045493ms" name=mark-k8s-node-i-0de0ebddfb1f1a317-as-available subsys=controller uuid=0a29a389-734c-4430-9ed2-5311fc45cd4a
level=debug msg="Controller run failed" consecutiveErrors=8 error="nodes \"i-0de0ebddfb1f1a317\" is forbidden: User \"system:serviceaccount:kube-system:cilium-operator\" cannot patch resource \"nodes\" in API group \"\" at the cluster scope" name=mark-k8s-node-i-0de0ebddfb1f1a317-as-available subsys=controller uuid=0a29a389-734c-4430-9ed2-5311fc45cd4a
```

No pods can be scheduled anymore on nodes as the cilium related taint cannot be removed.

Cilium-operator pods should be allowed to patch nodes in order to remove the cilium related taint.

We added this taint as per the [cilium recommended installation guide](https://docs.cilium.io/en/v1.12/gettingstarted/taints/#taint-effects) and to avoid pods being scheduled before the CNI is actually working as expected.

It seems that the ClusterRole definition for the cilium-operator SA is not following the actual values from the official cilium helm chart. https://github.com/kubernetes/kops/blob/1bef61969cfde84538a8a6787b67639f7d102d75/upup/models/cloudup/resources/addons/networking.cilium.io/k8s-1.16-v1.12.yaml.template#L419-L579

maximumG:
Fixes #15464.

As per the cilium [recommended installation](https://docs.cilium.io/en/latest/installation/taints/), the cilium-operator pod is in charge of removing the `node.cilium.io/agent-not-ready` that can be put on nodes upon startup. As such the operator needs the `PATCH` permission on the node resource.

### DIFF
--- a/tests/integration/update_cluster/minimal-ipv6-cilium/data/aws_s3_object_minimal-ipv6.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/minimal-ipv6-cilium/data/aws_s3_object_minimal-ipv6.example.com-addons-bootstrap_content
@@ -55,7 +55,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.16
     manifest: networking.cilium.io/k8s-1.16-v1.12.yaml
-    manifestHash: dea487bf6b1b7fe738189959345233264860eb0476be3aa9bf2adea26c8d62e2
+    manifestHash: 4469a0401fedf0e4aaf2fd42b569af37824d4cd2db6cbb2d8738a057f968efdc
     name: networking.cilium.io
     needsRollingUpdate: all
     selector:
--- a/tests/integration/update_cluster/minimal-ipv6-cilium/data/aws_s3_object_minimal-ipv6.example.com-addons-networking.cilium.io-k8s-1.16_content
+++ b/tests/integration/update_cluster/minimal-ipv6-cilium/data/aws_s3_object_minimal-ipv6.example.com-addons-networking.cilium.io-k8s-1.16_content
@@ -218,6 +218,13 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - nodes
+  - nodes/status
+  verbs:
+  - patch
+- apiGroups:
+  - ""
+  resources:
   - services
   - endpoints
   - namespaces
--- a/tests/integration/update_cluster/minimal-warmpool/data/aws_s3_object_minimal-warmpool.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/minimal-warmpool/data/aws_s3_object_minimal-warmpool.example.com-addons-bootstrap_content
@@ -48,7 +48,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.16
     manifest: networking.cilium.io/k8s-1.16-v1.12.yaml
-    manifestHash: e47a9b297b7164c269de1f5218bbf5112ce68771648075156819f04c151d0814
+    manifestHash: 70e84f7fd93599d773d4ebf9be53b9f3e79d4b94edaf87a8ca22bf656baa4f39
     name: networking.cilium.io
     needsRollingUpdate: all
     selector:
--- a/tests/integration/update_cluster/minimal-warmpool/data/aws_s3_object_minimal-warmpool.example.com-addons-networking.cilium.io-k8s-1.16_content
+++ b/tests/integration/update_cluster/minimal-warmpool/data/aws_s3_object_minimal-warmpool.example.com-addons-networking.cilium.io-k8s-1.16_content
@@ -218,6 +218,13 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - nodes
+  - nodes/status
+  verbs:
+  - patch
+- apiGroups:
+  - ""
+  resources:
   - services
   - endpoints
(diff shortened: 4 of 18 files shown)


hakman:
@maximumG To fix the failing tests you need to run `hack/update-exptected.sh` and commit the changes.

olemarkus:
But that taint is not used in our setup. Does it then need this permission?

maximumG:
> But that taint is not used in our setup. Does it then need this permission?

I must agree that by default the cilium taint is not configured. However I might argue that users are able to [customize the taint per kops instance groups](https://kops.sigs.k8s.io/tutorial/working-with-instancegroups/#adding-taints-or-labels-to-an-instance-group), leaving the possibility to user to add this specific taint.

Actually this is exactly what we tried out, :smiley:  we put this taint to ensure no workload got scheduled before the cilium-agent was ready. And we were surprised that it was not removed after all.

maximumG:
@olemarkus Do you need additional info to validate this PR on your side ?

h3poteto:
I also have the same issue. And I can understand what @olemarkus says.

`node.cilium.io/agent-not-ready` taint is only used when we use ENI as IPAM in Cilium. And now we need to add the taint manually. Nevertheless, is it correct to prepare only permissions with kOps?
I'm thinking of 3 ideas.

1. Just write a document about the taint and permissions. 
    - This is an optional usage.
    - So users need to add the taint and create ClusterRole for the cilium-operator manually.
2. The permission and taint are automatically added by kOps when users specify ENI as IPAM.
    - The taint is automatically added by kOps only when users specify `networking.cilium.ipam = eni`.
    - At that time, permissions are also added by kOps.
4. Same as this PR.
    - Users add the taint manually.
    - But the permission required at that time is prepared by kOps.

What do you think?

hakman:
@h3poteto @maximumG Added to the topics list for kOps Office Hours next week.

hakman:
@johngmyers Any thoughts?

johngmyers:
Why is `nodes/status` needed?

The referenced documentation describes a situation where there is a second CNI that Cilium has to wrest control from, but that shouldn't apply to kOps.

If an extra taint was useful in some common configuration, I would prefer that kOps automatically apply it, in addition to giving cilium-operator the permssion to remove it.

zadjadr:
I believe this is needed long term anyway, because in v1.14 this change has been introduced:

> The Cilium operator now taints nodes where Cilium is scheduled to run but is not running.
> This prevents pods from being scheduled on nodes without Cilium.
The CNI configuration file is no longer removed on agent shutdown.
This means that pod deletion will always succeed; previously it would fail if Cilium was down for an upgrade.
> This should help prevent nodes accidentally entering an unmanageable state.
> It also means that nodes are not removed from cloud LoadBalancer backends during Cilium upgrades. 

https://github.com/cilium/cilium/pull/23486

johngmyers:
When Cilium requires it we will need to add the permission.

maximumG:
> Why is nodes/status needed?

On our production K8S clusters, we had some weird scheduling on new joining nodes. Basically, business pods were sometimes scheduled before Cilium daemonset, exhausting the node ressources (CPU/memory). As a result, the cilium-agent was not able to get scheduled on the node and trigger another node scale up.

With this taint added and removed by the cilium-operator, we would avoid such situation.

olemarkus:
If your business pods gets scheduled before the cilium DS, you must have specific tolerations for that. If you e.g tolerate any taint, adding more taints won't work.

johngmyers:
> With this taint added and removed by the cilium-operator, we would avoid such situation.

Taints are not under `nodes/status`. So I ask again: why is `nodes/status` needed?

zadjadr:
@johngmyers node/status is used to [annotate the node](https://github.com/cilium/cilium/blob/v1.13.5/install/kubernetes/cilium/templates/cilium-agent/clusterrole.yaml#L40-L48).

maximumG:
> node/status is used to [annotate the node](https://github.com/cilium/cilium/blob/v1.13.5/install/kubernetes/cilium/templates/cilium-agent/clusterrole.yaml#L40-L48). But I don't know if this is needed 100%.

Checking the latest cilium helm documentation for v1.12 [here](https://docs.cilium.io/en/v1.12/helm-reference/), I would say that this permission is indeed useless. The `annotateK8sNode` is set to `false` by default and AFAIK there is no way to modify this using cilium kops integration.

PR #15474, opened 2023-06-06
scaleway: use kops controller for bootstrap

Mia-Cross:
This PR adds the possibility to use the kops-controller to bootstrap the nodes so there is no need to pass the credentials to the nodes.


pkg/model/bootstrapscript.go
```diff
@@ -139,6 +139,9 @@ func (b *BootstrapScript) buildEnvironmentVariables() (map[string]string, error)
 		if b.ig.IsControlPlane() || !b.builder.UseKopsControllerForNodeBootstrap() {
 			passEnvs = true
 		}
+		if cluster.Spec.GetCloudProvider() == kops.CloudProviderScaleway && cluster.UsesLegacyGossip() {
+			passEnvs = true
+		}
```
hakman, line 144: Why is this still needed?

Mia-Cross: This is the only way I managed to have gossip clusters still working. Should we deprecate them right away and only support NoneDNS clusters ?

hakman: That would be a good idea.
Also, please take a look at https://github.com/kubernetes/kops/pull/15493. It may be interesting for you.

Mia-Cross: Yes this is what I meant. Thanks for the link to the PR, I made a few changes according to it and gossip clusters work fine without this fix now :+1:

hakman (approved):

PR #15475, opened 2023-06-07
Use Ubuntu 22.04 (Jammy) as the default distro for K8s 1.27+

hakman:


hakman:
/cc @justinsb @srikiz 
@Mia-Cross

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1526,30 +1526,36 @@ func addCiliumNetwork(cluster *api.Cluster) {
...
 	}
 
-	switch cluster.Spec.GetCloudProvider() {
-	case api.CloudProviderDO:
-		return defaultDOImage, nil
-	case api.CloudProviderHetzner:
-		return defaultHetznerImage, nil
-	case api.CloudProviderScaleway:
-		return defaultScalewayImage, nil
+	if kubernetesVersion.LT(semver.MustParse("1.27.0")) {
```
justinsb, line 1541: Something for discussion / another PR, but maybe we should just embed a version of the channel.   As it looks like what we're doing is having a fallback where we encode the channel in code, but now we have to keep them in sync.

justinsb:
This lgtm!

/approve
/lgtm
/hold cancel

PR #15479, opened 2023-06-07
New OpenStack security group rules for metrics

fchiacchiaretta:
These new security group rules to allow scraping of metrics for kubeControllerManager on port 10257/tcp and kubeScheduler on port 10259/tcp


zetaab (approved):
/approve
/lgtm

PR #15481, opened 2023-06-08
Add RHEL9 support

jsafrane:
Add support for RHEL 9.

RHEL 9 does not ship python2 and libcgroup. I am able to create a RHEL9 cluster with this PR, I did not check any older distro.

### DIFF
--- a/nodeup/pkg/model/miscutils.go
+++ b/nodeup/pkg/model/miscutils.go
@@ -46,7 +46,7 @@ func (b *MiscUtilsBuilder) Build(c *fi.NodeupModelBuilderContext) error {
 	if b.Distribution.IsRHELFamily() {
 		// TODO: These packages have been auto-installed for a long time, and likely we don't need all of them any longer
 		packages = append(packages, "wget")
-		if b.Distribution != distributions.DistributionAmazonLinux2023 {
+		if b.Distribution != distributions.DistributionAmazonLinux2023 && b.Distribution != distributions.DistributionRhel9 {
 			packages = append(packages, "curl")
 			packages = append(packages, "python2")
 		}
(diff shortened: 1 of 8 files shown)


nodeup/pkg/model/packages.go
```diff
@@ -61,7 +61,6 @@ func (b *PackagesBuilder) Build(c *fi.NodeupModelBuilderContext) error {
 		c.AddTask(&nodetasks.Package{Name: "ebtables"})
 		c.AddTask(&nodetasks.Package{Name: "ethtool"})
 		c.AddTask(&nodetasks.Package{Name: "iptables"})
-		c.AddTask(&nodetasks.Package{Name: "libcgroup"})
```
hakman, line 64: Let's add a case for RHEL 9 and move the `libcgroup` to `default` case to keep behaviour for existing distros.

jsafrane: I removed `switch`, there would be too much duplication (i.e. install libcgroup in `DistributionAmazonLinux2023:` in in `default:`)

nodeup/pkg/model/miscutils.go
```diff
@@ -48,7 +48,6 @@ func (b *MiscUtilsBuilder) Build(c *fi.NodeupModelBuilderContext) error {
 		packages = append(packages, "wget")
 		if b.Distribution != distributions.DistributionAmazonLinux2023 {
```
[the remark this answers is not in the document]
jsafrane: done

jsafrane:
/test pull-kops-aws-distro-rhel8
/test pull-kops-aws-distro-rocky8
just to be sure...

PR #15486, opened 2023-06-09
scaleway: use latest version of the ccm

Mia-Cross:
We had to rollback the version of the CCM in #15333 while waiting for the next release. It is now available so we can use the latest version again.


hakman:
This is generally not a recommended solution and, when going to beta, would need to switch to fixed version.
The recommended approach would be to make the image value configurable.
/ok-to-test

hakman (approved):

PR #15493, opened 2023-06-11
Use kops-controller on hetzner, even with gossip

justinsb:
This is a more secure configuration.

### DIFF
--- a/pkg/model/hetznermodel/loadbalancer.go
+++ b/pkg/model/hetznermodel/loadbalancer.go
@@ -60,7 +60,7 @@ func (b *LoadBalancerModelBuilder) Build(c *fi.CloudupModelBuilderContext) error
 		},
 	}
 
-	if b.Cluster.UsesNoneDNS() {
+	if b.Cluster.UsesNoneDNS() || b.UseKopsControllerForNodeBootstrap() {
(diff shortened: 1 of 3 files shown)


pkg/model/hetznermodel/loadbalancer.go
```diff
@@ -60,7 +60,7 @@ func (b *LoadBalancerModelBuilder) Build(c *fi.CloudupModelBuilderContext) error
 		},
 	}
 
-	if b.Cluster.UsesNoneDNS() {
+	if b.Cluster.UsesNoneDNS() || b.UseKopsControllerForNodeBootstrap() {
```
justinsb, line 63: We probably could just do `b.UseKopsControllerForNodeBootstrap()` at this point, but this seems like the more obvious change.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1419,44 +1419,64 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 	}
 
 	// Set API server address to an IP from the cluster network CIDR
+	var controlPlaneIPs []string
```
justinsb, line 1422: This looks like a bigger change, but is really saying "also use the fixed IP on hetzner"

hakman (approved):

PR #15496, opened 2023-06-13
Pass feature flags to nodeup

jsafrane:
When kops is running with feature flags, it should pass the same flags to `nodeup` on all nodes. `nodeup` can then render all manifests with the feature flags in mind.

Add a copy of `KOPS_FEATURE_FLAG` both to to `/etc/sysconfig/kops-configuration` and the bootstrap script.


jsafrane:
Fixed unit test - sort `/etc/sysconfig/kops-configuration` to have stable output.

hakman:
This seems like an interesting idea. Will add it to the Office Hours topic. Thanks @jsafrane.

jsafrane:
With the latest https://github.com/kubernetes/kops/pull/15487, I don't need this PR any longer. Do you still want to merge it?

hakman:
I checked with the team and the intent of feature flags is to be used only in kOps CLI, not pass further. For that, there are other mechanisms.
Thanks for the effort.

PR #15512, opened 2023-06-18
etcd-manager: support symlinking versions

justinsb:
This is an easy way for us to signal that certain versions are
compatible with each to etcd-manager, which is otherwise
overly-cautious when it comes to unknown versions.

We extend kops-utils to support ln commands, and try to match the
syntax more closely align with traditional unix syntax.

### DIFF
--- a/cmd/kops-utils-cp/main.go
+++ b/cmd/kops-utils-cp/main.go
@@ -17,17 +17,17 @@ limitations under the License.
 package main
 
 import (
+	"flag"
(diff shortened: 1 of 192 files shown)


justinsb:
This is what I came up, the reason is that etcd-manager really wants to use the older version during an etcd upgrade, before the upgrade takes place.  If we symlink though, we can say "please use 3.5.9 when 3.5.7 is requested"

PR #15514, opened 2023-06-18
Remap all init container images of etcd-manager

alasdairtran:
I was testing out kops v1.27.0-beta.1 and noticed that the new `kops-utils-cp` init container image for etcd-manager hasn't been remapped via the asset builder and thus not copied to our custom container proxy when we run `kops get assets --copy`. 

This PR adds the missing code to remap the `kops-utils-cp` image.

### DIFF
--- a/pkg/model/components/etcdmanager/model.go
+++ b/pkg/model/components/etcdmanager/model.go
@@ -273,13 +273,16 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
 					},
 				},
 			}
-			// Remap image via AssetBuilder
-			remapped, err := b.AssetBuilder.RemapImage(initContainer.Image)
-			if err != nil {
-				return nil, fmt.Errorf("unable to remap container image %q: %w", initContainer.Image, err)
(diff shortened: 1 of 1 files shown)


pkg/model/components/etcdmanager/model.go
```diff
@@ -256,6 +256,17 @@ func (b *EtcdManagerBuilder) buildPod(etcdCluster kops.EtcdClusterSpec, instance
 	}
 
 	{
+		// Remap all init container images via AssetBuilder
+		if len(pod.Spec.InitContainers) > 0 {
```
hakman, line 260: There is no need to check the length of a slice before a loop.
This should be moved after the `etcdSupportedVersions` loop and do all init containers in a single loop.

hakman (changes requested):

pkg/model/components/etcdmanager/model.go
alasdairtran: Thanks for the suggestions. I've made those changes.

hakman (approved):
Thanks @alasdairtran. Nice find and fix!

PR #15516, opened 2023-06-18
etcd-manager: set environment variables once

justinsb:
Previously we were setting the env variable up to 3 times in an HA
control-plane, because we were adding to the etcd-manager
configuration once for each replica.


justinsb:
See (for example) the tests/integration/update_cluster/ha/data/aws_launch_template_master-us-test-1a.masters.ha.example.com_user_data output, where the env var is repeated 3 times.

One thing I'm not sure of, is whether we want this to appear once in the data files we write.  I think it's superfluous because we have the `backupRetentionDays: 90` as well.  But if we wanted to it wouldn't be too hard to just scan for duplicate env vars and simply remove them.  This is only a cleanup - I don't think it breaks anything to specify the env var 3 times - so maybe that is the more defensive-coding approach?

cc @hakman

PR #15542, opened 2023-06-22
kubetest2-kops: add support for boskos-resource-type flag

justinsb:
This will cause AWS tests to acquire an aws-account from boskos (when
set to aws-account), and run the tests with those credentials.

### DIFF
--- a/tests/e2e/go.mod
+++ b/tests/e2e/go.mod
@@ -8,6 +8,7 @@ replace k8s.io/kops => ../../.
 replace k8s.io/client-go => k8s.io/client-go v0.31.0
 
 require (
+	github.com/aws/aws-sdk-go v1.44.283
 	github.com/blang/semver/v4 v4.0.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/octago/sflags v0.2.0
--- a/tests/e2e/kubetest2-kops/aws/s3.go
+++ b/tests/e2e/kubetest2-kops/aws/s3.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


justinsb:
I can verify this once we have https://github.com/kubernetes/test-infra/pull/29907 merged

justinsb:
@ameukam so I hit some problems with the existing `aws-account` boskos pool.  I got accounts like `capa-user-XX` (e.g. capa-user-02).  I don't think they have access to the shared S3 bucket we were using previously.  So I tried having kubetest2-kops create a bucket in each AWS account (GCE does something similar already).  However, I think maybe the capa-user-XX accounts don't have S3 access (or I've messed up credentials somewhere, which is definitely plausible :-) )

Maybe we try creating some new accounts in a new boskos pool (e.g. resource-type == `kops-aws-account` ?), and make sure they have permissions to do S3?  Doing separate pools will also help us with tracking back spend (e.g. kops vs capa).

ameukam:
> @ameukam so I hit some problems with the existing `aws-account` boskos pool. I got accounts like `capa-user-XX` (e.g. capa-user-02). I don't think they have access to the shared S3 bucket we were using previously. So I tried having kubetest2-kops create a bucket in each AWS account (GCE does something similar already). However, I think maybe the capa-user-XX accounts don't have S3 access (or I've messed up credentials somewhere, which is definitely plausible :-) )
> 
> Maybe we try creating some new accounts in a new boskos pool (e.g. resource-type == `kops-aws-account` ?), and make sure they have permissions to do S3? Doing separate pools will also help us with tracking back spend (e.g. kops vs capa).

@justinsb we can use those accounts https://github.com/kubernetes/k8s.io/blob/main/infra/aws/terraform/management-account/organization-accounts-boskos.tf#L17-L42 for kOps. They are currently unused. WDYT ?

justinsb:
@ameukam that sounds good to me - it's great that they under terraform control because if we do need any other permissions we have a good path forwards.  Are they loaded into a boskos pool?  Any idea how to select them if so - I think I can select a different `resource-type`, not sure if there's any other way to select them...

ameukam:
@justinsb They are currently not onboarded. I don't have access to the `default` cluster. Would be interesting to have a dedicated EKS cluster with a dedicated boskos where can test those changes. That means also move some (or all the) presubmits to this new cluster.

hakman:
@justinsb @ameukam Anything new here?

ameukam:
> @justinsb @ameukam Anything new here?

Nothing new for the moment. A few things needs to be done on my side.

justinsb:
/reopen

Let's discuss some more!

PR #15547, opened 2023-06-23
Fix long auth helper cache file name

norseto:
The `kops helpers kubectl-auth` command tries to create a cache file that is long in proportion to the length of the `cluster` and `state` parameters, and outputs the following message. This PR corrects this behavior so that it always generates cache file names of the same length.

> I0623 12:21:51.479433    8889 kubectl_auth.go:114] cached credential "/home/vscode/.kube/cache/kops-authentication/my_kops_cluster_test_my_domain_net_73333a2f2f6d792d6b6f70732d73746174652d6d792d6177732d6163636f756e742d69642d61702d6e6f727468656173742d31006d792d6b6f70732d636c75737465722e746573742e6d792d646f6d61696e2e6e657400e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" was not valid: open /home/vscode/.kube/cache/kops-authentication/my_kops_cluster_test_my_domain_net_73333a2f2f6d792d6b6f70732d73746174652d6d792d6177732d6163636f756e742d69642d61702d6e6f727468656173742d31006d792d6b6f70732d636c75737465722e746573742e6d792d646f6d61696e2e6e657400e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855: input/output error

### DIFF
--- a/pkg/commands/helpers/kubectl_auth.go
+++ b/pkg/commands/helpers/kubectl_auth.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math/big"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -173,7 +174,11 @@ func cacheFilePath(kopsStateStore string, clusterName string) string {
 	b.WriteString(clusterName)
 	b.WriteByte(0)
 
-	hash := fmt.Sprintf("%x", sha256.Sum256(b.Bytes()))
+	var i big.Int
+	hb := sha256.Sum224(b.Bytes())
+	i.SetBytes(hb[:])
+
+	hash := i.Text(62)
 	return filepath.Join(homedir.HomeDir(), ".kube", "cache", "kops-authentication", hash)
 }
 
--- a/pkg/commands/helpers/kubectl_auth.go
+++ b/pkg/commands/helpers/kubectl_auth.go
@@ -27,7 +27,6 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -174,20 +173,8 @@ func cacheFilePath(kopsStateStore string, clusterName string) string {
 	b.WriteString(clusterName)
 	b.WriteByte(0)
 
-	hash := fmt.Sprintf("%x", sha256.New().Sum(b.Bytes()))
-	sanitizedName := strings.Map(func(r rune) rune {
-		switch {
-		case r >= 'a' && r <= 'z':
-			return r
-		case r >= 'A' && r <= 'Z':
-			return r
-		case r >= '0' && r <= '9':
-			return r
-		default:
-			return '_'
-		}
-	}, clusterName)
-	return filepath.Join(homedir.HomeDir(), ".kube", "cache", "kops-authentication", sanitizedName+"_"+hash)
+	hash := fmt.Sprintf("%x", sha256.Sum256(b.Bytes()))
+	return filepath.Join(homedir.HomeDir(), ".kube", "cache", "kops-authentication", hash)
 }
 
 func loadCachedExecCredential(cacheFilePath string) (*ExecCredential, error) {
--- a/pkg/commands/helpers/kubectl_auth.go
+++ b/pkg/commands/helpers/kubectl_auth.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


justinsb:
This does seem reasonable.  We could also truncate the sha256 without any real risk of collisions.

The advantage of keeping the cluster name is that it is more obvious which file is which.  I think that's the only reason.  But the length is potentially unbounded.  Though in practice long cluster names tend to hit up against problems in kube, so I don't think it's a real risk either.

That said, this is clearly a real problem - thanks for the fix @norseto.  We'll discuss in office hours, but I think your proposal is reasonable.  The one thing I want to check is whether this name is still too long for some filesystems (I think it's 512 characters), if that is the case then I think another approach could be to just take the first e.g. 64 characters of the hash or something like that (and _maybe_ the first e.g. 32 characters of the cluster name)

pkg/commands/helpers/kubectl_auth.go
```diff
@@ -174,20 +173,8 @@ func cacheFilePath(kopsStateStore string, clusterName string) string {
...
-		case r >= 'A' && r <= 'Z':
-			return r
-		case r >= '0' && r <= '9':
-			return r
-		default:
-			return '_'
-		}
-	}, clusterName)
-	return filepath.Join(homedir.HomeDir(), ".kube", "cache", "kops-authentication", sanitizedName+"_"+hash)
+	hash := fmt.Sprintf("%x", sha256.Sum256(b.Bytes()))
```
hakman, line 176: This should reduce the size of the hash a little. Should allow to keep at least a partial sanitized name.
WDYT @justinsb?
```suggestion
	h := md5.New()
	h.Write([]byte(kopsStateStore))
	h.Write([]byte(clusterName))
	hash := hex.EncodeToString(h.Sum(nil))
```

pkg/commands/helpers/kubectl_auth_test.go
```diff
@@ -0,0 +1,44 @@
...
+				"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijk." +
+				"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijk." +
+				"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijk.com",
+		},
+	}
+
+	output1 := cacheFilePath(inputs[0].kopsStateStore, inputs[0].clusterName)
+	_, file := path.Split(output1)
+
+	if len(file) > 64 {
```
hakman, line 41: @norseto Why 64 chars is too long?
```suggestion
	if len(file) > 64 {
```

norseto: @hakman 
This tests that the length of the sha256 string does not exceed 64 characters with this implementation.

hakman: OK, to put it differently, what is the minimum size for which it fails to read the file for you?

norseto: It fails at 256 characters file name. It will not fail with 255 characters.

pkg/commands/helpers/kubectl_auth.go
justinsb: Yes, the worry is that now the scanners will start false-positive triggering on "uses md5".  I have yet to see a tool smart enough to detect a truncation of a hash though :-)

johngmyers:
I would suggest using the base64 URLEncoding instead of hex.

norseto:
Since different file systems have different case sensitivity, wouldn't base36 be safer?

johngmyers:
We're already truncating, so we're losing bits one way or another. But base32 would also work.

norseto:
Updated to use SHA224 and Base62. File names can be up to 38 characters long.
It would be possible to give the cluster name first as in the current implementation, but the maximum length to be given would need to be determined. For example, the first 24 characters maximum, etc. In that case the length of the filename would be 63 characters...

justinsb:
Thanks for tracking this down @norseto and for providing a fix!  We talked about this in office hours; the root cause here was a bug I'm pretty sure I introduced, specifically `sha256.New().Sum(b.Bytes())` just isn't right - it is not computing the hash at all!  I should probably have used `sha256.Sum256` or the more normal `hasher := sha256.New(); hasher.Write(b); hash := hasher.Sum(nil)`.  So that's why the filename was _so_ long, because it wasn't even a sha256.

I think for most modern filesystems we will get away with a length up to 255.  So I do think including the cluster name would be good.

I think you've fixed a big bug here, so if you prefer I will just LGTM this (and then I might reintroduce the cluster name).

But if you're game to do one last iteration, the cluster name is helpful for people that want to know what these files are, so I think the preference that we discussed in office hours is something like this:


```
hash := fmt.Sprintf("%x", sha256.Sum224(b.Bytes()))
if len(hash) > 32 {
hash = hash[:32]
}
	sanitizedName := strings.Map(func(r rune) rune {
		switch {
		case r >= 'a' && r <= 'z':
			return r
		case r >= 'A' && r <= 'Z':
			return r
		case r >= '0' && r <= '9':
			return r
		default:
			return '_'
		}
	}, clusterName)
if len(sanitizedName) > 32 {
sanitizedName = sanitizedName[:32]
}
	return filepath.Join(homedir.HomeDir(), ".kube", "cache", "kops-authentication", sanitizedName+"_"+hash)
	```

Obviously we can tweak the max lengths to hit any desired length, it's a compromise on length vs keeping the cluster identifiable (for sanitizedName)  and avoiding collisions (for hash)

Thanks again for the contribution, and sorry that we are going round a bit on this one.  It took a while for us to see the actual underlying bug, because the hash API is not the most intuitive :-)

norseto:
@justinsb 
I think it is reasonable to include sanitizedName for identifing the target cluster.
However, since it will be combined with sanitizedName, I think the risk of conflict is low to truncate the sha224 result to 32 characters, 128bits. But I do not have proof of how low.
I think it is best if it can be used without truncation.
I have just pushed a fix. Please take a look.

PR #15551, opened 2023-06-27
Allow overriding uint values

hakman:


util/pkg/reflectutils/access.go
```diff
@@ -161,6 +161,24 @@ func setType(v reflect.Value, newValue string) error {
...
+		case "uint":
+			newV = reflect.ValueOf(v)
+		case "uint32":
+			v32 := uint32(v)
+			newV = reflect.ValueOf(v32)
+		case "uint64":
+			v64 := uint64(v)
+			newV = reflect.ValueOf(v64)
+		default:
+			panic("missing case in switch")
```
prateekgogia, line 180: nit: mention this panic happened in uint case because while debugging it can become harder to tell where this occurred in `int` case or `uint` case.
```suggestion
			panic("missing case for uint switch")
```

hakman: Thanks for the feedback. Done!

hakman:
> (What's the uint value, out of interest?)

Don't remember, I did not use it in the end, but already had this patch. 🤣

PR #15552, opened 2023-06-27
Add create cluster flag for specifying the list of etcd clusters

hakman:


hakman:
This should be useful for scaling tests.
/cc @dims @justinsb

PR #15557, opened 2023-06-28
make cni plugin configurable in scaling test scenario

prateekgogia:


tests/e2e/scenarios/scalability/run-test.sh
```diff
@@ -86,7 +86,7 @@ echo "ADMIN_ACCESS=${ADMIN_ACCESS}"
 # cilium does not yet pass conformance tests (shared hostport test)
 #create_args="--networking cilium"
 create_args=()
-create_args+=("--networking=calico")
+create_args+=("--networking=${CNI_PLUGIN:-calico}")
```
hakman, line 89: Use Ubuntu 20.04 as base, or make this also configurable:
```suggestion
create_args+=("--image=ssm:/aws/service/canonical/ubuntu/server/20.04/stable/current/arm64/hvm/ebs-gp2/ami-id")
create_args+=("--networking=${CNI_PLUGIN:-calico}")
```

prateekgogia:
will cut a follow up PR in test-infra to make use of these variables.

tests/e2e/scenarios/scalability/run-test.sh
```diff
@@ -96,7 +98,8 @@ create_args+=("--node-count=${KUBE_NODE_COUNT:-101}")
 # However, it currently fails two tests (HostPort & OIDC) so need to track that down
 #create_args="--dns none"
 create_args+=("--node-size=c6g.medium")
-create_args+=("--master-size=c6g.2xlarge")
+create_args+=("--control-plane-count=2")
```
hakman, line 101: It should be either 1 or 3. Also, not sure what the reason is to increase the control-plane number. It will probably decrease performance.
```suggestion
create_args+=("--control-plane-count=3")
```

prateekgogia: Thanks I was just trying to figure this out, should kops have this validation to fail the cluster creation if the size is 2?

prateekgogia: Also, the reason we want to add more CP nodes is to be able to perform a 5k kube nodes scale test. I have made this count configurable.

hakman (approved):

tests/e2e/scenarios/scalability/run-test.sh
hakman: 2 CP nodes should also be doable, never tried it though. Let's do that separately.
Each CP has etcd running. The more etcd replicas, the slower it gets. (my reasoning)

PR #15562, opened 2023-06-30
Update dependencies

hakman:
Ref: https://github.com/google/go-tpm/releases/tag/v0.9.0


upup/pkg/fi/cloudup/gce/tpm/gcetpmsigner/tpm_other.go
```diff
@@ -23,7 +23,7 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/legacy/tpm2"
```
hakman, line 26: @justinsb WDYT?

justinsb: https://github.com/kubernetes/cloud-provider-gcp/blob/e64cf3b0fcb1958cee1fe55d7e30f4573c34bf4e/pkg/tpmattest/tpmattest.go#L11 hasn't yet updated, so I don't think we want to cherry-pick without a long soak time though!

PR #15564, opened 2023-06-30
Revert "Remove obsolete etcd versions"

hakman:
/cc @justinsb 
/assign @justinsb


justinsb:
As discussed in office hours, we're going to add the versions and then use symlinks in the future to make that more efficient (possibly in a future kOps version, that is TBD)

/approve
/lgtm

PR #15565, opened 2023-07-01
etcd-manager: support symlinking versions

justinsb:
This is an easy way for us to signal that certain versions are
compatible with each to etcd-manager, which is otherwise
overly-cautious when it comes to unknown versions.
    
We extend kops-utils-cp to support the `--target-dir` to write to a
directory; and the `--symlink` flag to use symlinks.  The syntax
isn't identical to cp, but should be semi-familiar and allows us to
minimize the number of initContainers we use.

### DIFF
--- a/cmd/kops-utils-cp/main.go
+++ b/cmd/kops-utils-cp/main.go
@@ -17,17 +17,18 @@ limitations under the License.
 package main
 
 import (
+	"flag"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"k8s.io/klog/v2"
 )
 
-func copyFile(source, target string) error {
-	klog.Infof("Copying source file %q to target directory %q", source, target)
+func copyFile(source, targetDir string, force bool) error {
+	klog.Infof("Copying source file %q to target directory %q", source, targetDir)
 
 	sf, err := os.Open(source)
 	if err != nil {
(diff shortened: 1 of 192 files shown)


justinsb:
This is similar to #15512 but we're not trying to be compatible with cp and ln, and we can minimize the number of initContainers.

hakman:
@justinsb Should we merge https://github.com/kubernetes/kops/pull/15564 first?

pkg/model/components/etcdmanager/model.go
```diff
@@ -229,6 +218,8 @@ spec:
     emptyDir: {}
 `
 
+const kopsUtilsImage = "registry.k8s.io/kops/kops-utils-cp:1.28.0-alpha.1"
```
hakman, line 221: Should we move the etcd-manager image here too?

pkg/model/components/etcdmanager/options.go
```diff
@@ -65,34 +69,41 @@ func (b *EtcdManagerOptionsBuilder) BuildOptions(o interface{}) error {
...
-	"3.5.3":  "registry.k8s.io/etcd:3.5.3-0",
-	"3.5.4":  "registry.k8s.io/etcd:3.5.4-0",
-	"3.5.6":  "registry.k8s.io/etcd:3.5.6-0",
-	"3.5.7":  "registry.k8s.io/etcd:3.5.7-0",
-	"3.5.9":  "registry.k8s.io/etcd:3.5.9-0",
+// etcdVersion describes how we want to support each etcd version.
+type etcdVersion struct {
+	Version          string
+	Image            string
+	SymlinkToVersion string
```
hakman, line 76: Should we start using `SymlinkToVersion`?

cmd/kops-utils-cp/main.go
```diff
@@ -62,20 +64,67 @@ func copyFile(source, target string) error {
...
+	var targetDirs stringSliceFlags
+	flag.Var(&targetDirs, "t", "copy to directory")
+
+	flag.Parse()
+
+	sources := flag.Args()
+
+	if len(targetDirs) == 0 {
+		if len(sources) < 2 {
+			log.Fatal("Usage: kops-utils-cp SOURCE ... TARGETDIR")
```
hakman, line 101: Not sure this is still accurate.

hakman (changes requested):

pkg/model/components/etcdmanager/model.go
justinsb: Probably ... but maybe we do that separately/later?  (Unless you can think of some reason to do it at the same time...)

pkg/model/components/etcdmanager/options.go
justinsb: Good point.  I added some commits that now do.  I thought it would be useful to first show comparable behaviour, then add the symlinks.

hakman:
Hmm... not quite sure what's wrong here. Clusters are not coming up.

justinsb:
Figured it out I think; two issues:

1) Flag parsing (at least with flag.Parse) requires that positional arguments come strictly last, so had to make sure this was the case.
2) Added a flag `-f` to force, so that if there are any failures we can be idempotent.  We weren't seeing (1) clearly because the container was being retried, and files had previous been copied (AFAICT)

hakman:
> 1. Flag parsing (at least with flag.Parse) requires that positional arguments come strictly last, so had to make sure this was the case.

@justinsb Should we just drop any positional args? I don't think there's much value for them.

justinsb:
Changed to make `force` / overwrite always true (removed the flag), because `cp` overwrites automatically by default.

By "positional arguments", I mean the list of source files (as opposed to the flag values, e.g. the list of target directories).  I might be using the wrong name :-)

hakman:
> By "positional arguments", I mean the list of source files (as opposed to the flag values, e.g. the list of target directories). I might be using the wrong name :-)

Yes, I meant that we may use `-f` to specify the source file(s), `-d` to specify target dir(s) and `-s` to create symlinks instead of copy.

justinsb:
> Yes, I meant that we may use -f to specify the source file(s), -d to specify target dir(s) and -s to create symlinks instead of copy.

If we do that, we should probably use long-form flags to avoid confusion, like `--src`, --target-dir` and `--symlink`.  That could work though...

hakman:
> This is similar to #15512 but we're not trying to be compatible with cp and ln, and we can minimize the number of initContainers.

Can https://github.com/kubernetes/kops/pull/15512 be closed now?

pkg/model/components/etcdmanager/options.go
```diff
@@ -65,34 +69,53 @@ func (b *EtcdManagerOptionsBuilder) BuildOptions(o interface{}) error {
...
+	{Version: "3.4.3", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.4", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.5", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.6", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.7", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.8", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.9", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.10", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.11", SymlinkToVersion: "3.4.13"},
+	{Version: "3.4.12", SymlinkToVersion: "3.4.13"},
```
hakman, line 92: We only supported 3.4.3 and 3.4.13, we should not add more here.
Similar to the 3.5.x, we should skip 2, 5 and 9.

hakman (approved):

PR #15570, opened 2023-07-01
azure: Add support for network security groups

hakman:
/cc @rifelpet @justinsb


pkg/model/azuremodel/network.go
```diff
@@ -41,14 +45,134 @@ func (b *NetworkModelBuilder) Build(c *fi.CloudupModelBuilderContext) error {
...
 
+	nsgTask := &azuretasks.NetworkSecurityGroup{
+		Name:          fi.PtrTo(b.NameForVirtualNetwork()),
+		Lifecycle:     b.Lifecycle,
+		ResourceGroup: b.LinkToResourceGroup(),
+		Tags:          map[string]*string{},
+	}
+	var sshAccessIPv4, sshAccessIPv6 []string
+	for _, cidr := range b.Cluster.Spec.SSHAccess {
+		switch net.IPFamilyOfCIDRString(cidr) {
```
justinsb, line 56: Oooh - I didn't know about this function, I think we have a lot of places that check if an IP is an IPv4 or IPv6 that could probably benefit from this!

justinsb:
LGTM, I think we can go ahead and merge given where we are in the development cycle (I'm assuming you don't want to cherry-pick this to 1.27)

/approve
/lgtm

pkg/model/azuremodel/network.go
hakman: Very true. Also we have some of our own util for this that we could replace.

PR #15579, opened 2023-07-02
Don't download container runtime assets when skipping the installation

### LINKED ISSUE #15558: containerd/runc assets are still downloaded if skipInstall is true
/kind bug

 this information.**
1.26.4

 a `kops` flag.**
1.26.6

AWS/Spotinst

Have a cluster with `.spec.containerd.skipInstall = true`.

The node logs indicate that containerd and runc assets are downloaded and extracted. They are never used.

For this to not happen

hakman:
For now, this only works when `SkipInstall=true` is set at cluster level.

Closes #15558

### DIFF
--- a/upup/pkg/fi/cloudup/apply_cluster.go
+++ b/upup/pkg/fi/cloudup/apply_cluster.go
@@ -72,7 +72,6 @@ import (
 	"k8s.io/kops/upup/pkg/fi/cloudup/terraform"
 	"k8s.io/kops/upup/pkg/fi/cloudup/terraformWriter"
 	"k8s.io/kops/util/pkg/architectures"
-	"k8s.io/kops/util/pkg/hashing"
 	"k8s.io/kops/util/pkg/mirrors"
 	"k8s.io/kops/util/pkg/vfs"
 )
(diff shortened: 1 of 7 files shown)


upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1072,37 +1071,49 @@ func (c *ApplyClusterCmd) addFileAssets(assetBuilder *assets.AssetBuilder) error
...
+				break
+			}
+
+			dockerAssetUrl, dockerAssetHash, err := findDockerAsset(c.Cluster, assetBuilder, arch)
+			if err != nil {
+				return err
+			}
+			if dockerAssetUrl != nil && dockerAssetHash != nil {
+				c.Assets[arch] = append(c.Assets[arch], mirrors.BuildMirroredAsset(dockerAssetUrl, dockerAssetHash))
+			}
```
johngmyers, line 1094: In the future, it might be worth extracting this out to a method, possibly a reciever on `assetBuilder`. The method would take a `func(Cluster, assetBuillder, arch) (url, hash, error)`

johngmyers (approved):

upup/pkg/fi/cloudup/apply_cluster.go
hakman: Yep, should be pretty easy.

Issue #6318, opened 2019-01-09
Upgrading to CoreDNS does not remove kube-dns deployments


syscll:
**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
```
Version 1.11.0 (git-2c2042465)
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
1.11.6
```

**3. What cloud provider are you using?**
```
AWS
```

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
```
kops edit cluster
```

Add CoreDNS provider:
```
kubeDNS:
  provider: CoreDNS
```

Update cluster:
```
kops update cluster --yes
```

**5. What happened after the commands executed?**

Everything ran successfully. However, I still see `kube-dns` deployments:
```
$ kubectl get deployments -n=kube-system
NAME                   DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
coredns                2         2         2            2           1d
dns-controller         1         1         1            1           1y
kube-dns               2         2         2            2           1y
kube-dns-autoscaler    1         1         1            1           1y
...
```

**6. What did you expect to happen?**

I expected `kube-dns` to be replaced by CoreDNS.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```
apiVersion: kops/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: 2017-12-19T15:09:49Z
  name: [REDACTED]
spec:
  additionalPolicies:
    node: |
      [{"Effect": "Allow","Action": ["route53:ChangeResourceRecordSets"],"Resource": ["arn:aws:route53:::hostedzone/*"]},{"Effect": "Allow","Action": ["route53:ListHostedZones","route53:ListResourceRecordSets"],"Resource": ["*"]}]
  api:
    loadBalancer:
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudProvider: aws
  configBase: [REDACTED]
  docker:
    version: 18.06.1
  etcdClusters:
  - etcdMembers:
    - instanceGroup: master-eu-central-1b
      name: b
    name: main
  - etcdMembers:
    - instanceGroup: master-eu-central-1b
      name: b
    name: events
  iam:
    allowContainerRegistry: true
    legacy: false
[...]

PR #15584, opened 2023-07-02
docs: remove `kube-dns-autoscaler` when upgrading to CoreDNS

agilgur5:
## Summary

Per https://github.com/kubernetes/kops/issues/6318#issuecomment-830490064 and related comments, the `kube-dns-autoscaler` is no longer needed when upgrading to CoreDNS.

## Details

- there is a separate `coredns-autoscaler` that is spun up in more recent kOps versions
  - so the `kube-dns-autoscaler` is no longer needed after an upgrade to CoreDNS - and possibly could conflict if they both targeted the same deployment

- updating the `kube-dns-autoscaler` was the most confusing part, so this simplifies things a good bit!

- also fix some minor inconsistent spacing in this doc

## Misc

Was teaching folks about k8s DNS so came back to the issue to see what the progress was w/ CoreDNS as I had to do this upgrade a couple years ago

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -244,7 +244,7 @@ etcdClusters:
 ### etcd backups retention
 {{ kops_feature_table(kops_added_default='1.18') }}
 
-As of kOps 1.27, the default etcd backup retention duration is 90 days.  
+As of kOps 1.27, the default etcd backup retention duration is 90 days.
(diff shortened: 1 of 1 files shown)


hakman (approved):
Thanks @agilgur5. Should be very useful for those that upgrade clusters.

PR #15588, opened 2023-07-03
doc: Added documentation about loadbalancer and security group configuration

valentin-ricard:
OpenStack needs an additional option to get access to NodePorts for loadbalancers.

This PR is aiming to document this behavior.


hakman (approved):
Thanks @valentin-ricard!
/ok-to-test

PR #15594, opened 2023-07-05
ipv6: containerd routes support for IPv6

justinsb:
If using IPv6 and a kubenet-style CNI (which is more common with
IPv6), we need to support an IPv6 route on the pod, or else Pods will
be unable to reach other Pods.

### DIFF
--- a/nodeup/pkg/model/containerd.go
+++ b/nodeup/pkg/model/containerd.go
@@ -17,6 +17,7 @@ limitations under the License.
 package model
 
 import (
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"regexp"
@@ -72,7 +73,9 @@ func (b *ContainerdBuilder) Build(c *fi.NodeupModelBuilderContext) error {
 		// This is a temporary backwards-compatible solution for kubenet users and will be deprecated when Kubenet is deprecated:
 		// https://github.com/containerd/containerd/blob/master/docs/cri/config.md#cni-config-template
 		if b.NodeupConfig.UsesKubenet {
-			b.buildCNIConfigTemplateFile(c)
+			if err := b.buildCNIConfigTemplateFile(c); err != nil {
+				return err
+			}
 			if err := b.buildIPMasqueradeRules(c); err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


nodeup/pkg/model/containerd.go
```diff
@@ -461,6 +461,13 @@ func (b *ContainerdBuilder) buildCNIConfigTemplateFile(c *fi.NodeupModelBuilderC
     ]
 }
 `
+
+	routes := `[{ "dst": "0.0.0.0/0" }]"`
```
hakman, line 465:
```suggestion
	routes := `[{ "dst": "0.0.0.0/0" }]`
```

johngmyers, line 465: Not blocking, but I dislike putting JSON strings in code. It's more maintainable if you call `json.Marshal()`.

justinsb: Yes, I agree and I did debate structifying this, but I don't think it has a well-defined schema, much less a go type we can import - so it would be a big change.   Or I could start with routes and we can gradually expand, but then it's weird that we are mixing approaches here.

johngmyers: You can use ad-hoc types like `[]map[string]string{{"dst":"0.0.0.0/0"}}` or something like that.

justinsb: Done

johngmyers:
Looks good, just need to update-expected.sh

johngmyers:
We should probably have a containerdbuilder test that exercises the new functionality.

justinsb:
@johngmyers  I tried adding a test and sadly we can't get more test coverage on the IPv6 side until we tweak which CNI providers work with IPv6 - right now kubenet is blocked.  I'm not sure whether we want to enable IPv6 with kubenet itself or work on a different CNI.  My current approach - which I'm whittling down to something more manageable - has kops-controller setting the node networking status condition ready with IPv6, but that relies on the cloudprovider setting the providerID, which relies on the CCM, which _on some clouds_ relies on the node being ready.  That might be an issue with the CCM providers (Hetzner has this dependency, GCE does not).

We do have coverage of the IPv4 case - you can see that the spacing changed because of json.Marshal

johngmyers:
We could probably unblock `External`, or we'd have to wait until `GCE` is unblocked.

/lgtm
/approve

johngmyers:
Ah, `External` not supported for 1.26+

PR #15598, opened 2023-07-06
Release 1.27.0

hakman:


justinsb:
/hold

I say we hold this until we decide on #15608, as I think that will mess up node labelling on control-plane nodes on GCE (and DO?)

hakman:
@justinsb @johngmyers Any objection to releasing 1.27.0 now that everything is merged?

johngmyers:
I think we need to release 1.26.5, then add a release note advising GCE and DO to upgrade to 1.26.5 before going to 1.27+

hakman:
> I think we need to release 1.26.5, then add a release note advising GCE and DO to upgrade to 1.26.5 before going to 1.27+

Sure, but we can trigger the release and just publish them one after the other.

johngmyers:
We should test to verify the 1.26.5 to 1.27.0 upgrade fixes the GCE problem.

johngmyers:
I'm skeptical backporting that patch fixed the node-labeling problem on GCE 1.26 to 1.27 upgrades. We might need to rolling-update the control plane twice. Could someone run a test?

PR #15585, opened 2023-07-03
Update Karpenter to v0.28.1

hakman:
/cc @olemarkus @justinsb

PR #15601, opened 2023-07-07
feat(karpenter): Variabilize image, logFormat and logLevel

anthonyhaussman:
After discussion on https://github.com/kubernetes/kops/pull/15585#pullrequestreview-1512237273, add variables to give the possibility to set the log format for Karpenter and set the version.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -1508,6 +1508,12 @@ spec:
                 properties:
                   enabled:
                     type: boolean
+                  image:
+                    type: string
+                  logEncoding:
+                    type: string
+                  logLevel:
+                    type: string
                 type: object
               keyStore:
                 description: KeyStore is the VFS path to where SSL keys and certificates
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -249,7 +249,10 @@ type ScalewaySpec struct {
 }
 
 type KarpenterConfig struct {
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled     bool   `json:"enabled,omitempty"`
+	LogEncoding string `json:"logFormat,omitempty"`
+	LogLevel    string `json:"logLevel,omitempty"`
+	Image       string `json:"image,omitempty"`
 }
 
 // ServiceAccountIssuerDiscoveryConfig configures an OIDC Issuer.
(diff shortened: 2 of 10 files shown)


pkg/apis/kops/cluster.go
```diff
@@ -249,7 +249,9 @@ type ScalewaySpec struct {
 }
 
 type KarpenterConfig struct {
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled   bool   `json:"enabled,omitempty"`
+	Version   string `json:"version,omitempty"`
+	LogFormat string `json:"logFormat,omitempty"`
```
hakman, line 254: Let's keep this in alphabetical order.
```suggestion
	LogFormat string `json:"logFormat,omitempty"`
	Version   string `json:"version,omitempty"`
```

anthonyhaussman: Fixed :slightly_smiling_face:

k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -1508,6 +1508,10 @@ spec:
                 properties:
                   enabled:
                     type: boolean
+                  logFormat:
+                    type: string
+                  version:
+                    type: string
```
hakman, line 1514: @olemarkus Should we go with `version` or `image` in this case? Any thoughts?

hakman: The consensus is that we would like `image` instead of `version`. Could you update the PR @anthonyhaussman?

k8s/crds/kops.k8s.io_clusters.yaml
```diff
@@ -1508,6 +1508,10 @@ spec:
                 properties:
                   enabled:
                     type: boolean
+                  logFormat:
```
hakman, line 1511: To keep naming consistent with official manifest.
```suggestion
                  logEncoding:
```

k8s/crds/kops.k8s.io_clusters.yaml
hakman: @anthonyhaussman Please also add `logLevel`.

k8s/crds/kops.k8s.io_clusters.yaml
anthonyhaussman: Done :+1:

k8s/crds/kops.k8s.io_clusters.yaml
anthonyhaussman: I have added `logLevel` :slightly_smiling_face:

PR #15604, opened 2023-07-07
scaleway: documentation improvement

Mia-Cross:
This PR improves the documentation for Scaleway. It adds :

- information about the profiles feature
- information about a few default values and how to override them
- links to other pages

It also removes the temporary fix added in #15350 which is no longer needed.

### DIFF
--- a/docs/getting_started/scaleway.md
+++ b/docs/getting_started/scaleway.md
@@ -33,49 +33,79 @@
 
 ## Environment Variables
 
-It is important to set the following [environment variables](https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/README.md):
+### Enable Scaleway
+
+Since Scaleway support is currently in alpha, it is feature gated and you will need to set this variable:
+```bash
+export KOPS_FEATURE_FLAGS="Scaleway"
+```
+
+### Scaleway Credentials
+
+To be able to use Scaleway APIs, it is required to set up your credentials in the [environment](https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/README.md).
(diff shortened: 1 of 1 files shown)


hakman:
@Mia-Cross I would suggest to start using example with `--dns=none`. This is recommended instead of the less secure Gossip, which will probably be deprecated in the future.
We should release kOps v1.27.0 soon and will have all the new nice features that you added lately.

docs/getting_started/scaleway.md
```diff
@@ -33,49 +33,79 @@
...
+
+Since Scaleway support is currently in alpha, it is feature gated and you will need to set this variable:
+```bash
+export KOPS_FEATURE_FLAGS="Scaleway"
+```
+
+### Scaleway Credentials
+
+To be able to use Scaleway APIs, it is required to set up your credentials in the [environment](https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/README.md).
+You have 2 ways to pass your credentials:
```
bene2k1, line 46:
```suggestion
You have two ways to pass your credentials:
```

docs/getting_started/scaleway.md
```diff
@@ -33,49 +33,79 @@
...
+
+1. If you're already familiar with Scaleway's DevTools, then you probably have a config file (its default location is `$HOME/.config/scw/config.yaml`).
```
bene2k1, line 48:
```suggestion
1. If you are already familiar with Scaleway's DevTools, then you probably have a config file (its default location is `$HOME/.config/scw/config.yaml`).
```

hakman (approved):

PR #15607, opened 2023-07-09
gce: Use `user-data` instead of `startup-script` metadata key

### LINKED ISSUE #15589: MIME Headers in Startup Script causing "Command not found" error
/kind bug

I'm using Kops on GCE. I'm able to successfully set up a cluster but there appears to be a bug when I use additionalUserData to add custom code to the metadata startup script.

The MIME headers are getting included in the script which then throws a "command not found" error since it attempts to execute those headers as shell commands.

 this information.**

1.26.4

 a `kops` flag.**

1.26.6

GCE

kops create -f desktop-kops.yml --state ${KOPS_STATE_STORE}
kops update cluster --name sample.k8s.local --yes --state ${KOPS_STATE_STORE} --admin

The Cluster gets created successfully but the startup script has the MIME headers in it which causes problems.

I expected the command I added via the additionalUserData key to run at startup. I see it added to the startup script but the fact the MIME headers are there causes the script to fail.

```yaml

kind: Cluster
metadata:
  name: sample.k8s.local
spec:
  api:
    loadBalancer:
      type: Public
  authorization:
    alwaysAllow: {}
  channel: stable
  kubernetesApiAccess:
  - 0.0.0.0/0
  sshAccess:
  - 0.0.0.0/0
  project: ubuntu-desktop-migration
  subnets:
  - cidr: 10.10.0.0/21
    name: node-subnet
    type: Public
    region: us-central1
    stackType: IPV4_ONLY
    natGateway: true
    PublicIp: 34.xxx.xx.xx
  cloudProvider: gce
  etcdClusters:
  - etcdMembers:
    - instanceGroup: controller-instance
      name: a
    name: main
  - etcdMembers:
    - instanceGroup: controller-instance
      name: a
    name: events
  networking:
    calico: {}
  metricsServer:
    enabled: true
    insecure: true
  kubeDNS:
    provider: coredns
  kubelet:
    anonymousAuth: false
  topology:
    dns:
      type: Public
    masters: public
    nodes: public
---
kind: InstanceGroup
metadata:
  labels:
    kops.k8s.io/cluster: sample.k8s.local
  name: controller-instance
spec:
  additionalUserData:
  - name: arunstartupctrl.sh
    type: text/x-shellscript
    content: |
      #!/bin/bash
       sudo chmod 1777 /tmp
  role: ControlPlane
  image: ubuntu-os-cloud/ubuntu-2204-jammy-v20230616
  rootVolume: 20
  associatePublicIP: true
  machineType: e2-standard-4
  minSize: 1
  maxSize: 1
  gcpProvisioningModel: STANDARD
  zones: 
  - us-central1-a
  subnets:
  - node-subnet
  
---
kind: InstanceGroup
metadata:
  labels:
    kops.k8s.io/cluster: sample.k8s.local
  name: node-instance
spec:
  role: Node
  image: ubuntu-os-cloud/ubuntu-2204-jammy-v20230616
  rootVolume: 30
  associatePublicIP: false
  machineType: e2-standard-8
  minSize: 1
  maxSize: 10
  gcpProvisioningModel: SPOT
  zones:
  - us-central1-a
  subnets:
  - node-subnet
  additionalUserData:
  - name: arunstartupnodes.sh
    type: text/x-shellscript
    content: |
      #!/bin/bash
      sudo chmod 1777 /tmp

```

When I remove the additionalUserData part of the yml file the startup script doesn't have any MIME headers and runs successfully. 

startup script **with** additionalUserData:
```
startup-script	
Content-Type: multipart/mixed; boundary="MIMEBOUNDARY"
MIME-Version: 1.0

--MIMEBOUNDARY
Content-Disposition: attachment; filename="nodeup.sh"
Content-Transfer-Encoding: 7bit
Content-Type: text/x-shellscript
Mime-Version: 1.0

#!/bin/bash
set -o errexit
set -o nounset
set -o pipefail

```
startup script **without** additionalUserData:
```
startup-script	
#!/bin/bash
set -o errexit
set -o nounset
set -o pipefail
```

hakman:
See: https://github.com/hashicorp/terraform-provider-google/issues/1816

Also: https://cloudinit.readthedocs.io/en/23.2.1/reference/datasources/gce.html
> The GCE datasource gets its data from the internal compute metadata server. Metadata can be queried at the URL http://metadata.google.internal/computeMetadata/v1/ from within an instance. For more information see the [GCE metadata docs](https://cloud.google.com/compute/docs/storing-retrieving-metadata).
> user-data and user-data-encoding can be provided to cloud-init by setting those custom metadata keys for an instance.

Closes #15589


justinsb:
Thanks @hakman  - verified that is now generally supported, even on COS which I think was the problem a few years ago.

/approve
/lgtm

hakman:
> Thanks @hakman - verified that is now generally supported, even on COS which I think was the problem a few years ago.

Thanks for checking! This will make some people happy. 😄

PR #15608, opened 2023-07-09
kops-controller: load objects with version conversion

justinsb:
If we deserialize the yaml, we don't go through the version-conversion
logic.  That logic maps from Master -> ControlPlane, so without that
logic we see unexpected values in the "string enums".

### DIFF
--- a/cmd/kops-controller/controllers/legacy_node_controller.go
+++ b/cmd/kops-controller/controllers/legacy_node_controller.go
@@ -26,12 +26,11 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/v2"
-	"k8s.io/kops/pkg/apis/kops"
+	api "k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/apis/kops/registry"
 	"k8s.io/kops/pkg/kopscodecs"
 	"k8s.io/kops/pkg/nodeidentity"
 	"k8s.io/kops/pkg/nodelabels"
-	"k8s.io/kops/upup/pkg/fi/utils"
 	"k8s.io/kops/util/pkg/vfs"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -152,9 +151,9 @@ func (r *LegacyNodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-// getClusterForNode returns the kops.Cluster object for the node
+// getClusterForNode returns the api.Cluster object for the node
 // The cluster is actually loaded when we first start
-func (r *LegacyNodeReconciler) getClusterForNode(node *corev1.Node) (*kops.Cluster, error) {
+func (r *LegacyNodeReconciler) getClusterForNode(node *corev1.Node) (*api.Cluster, error) {
 	clusterPath := r.configBase.Join(registry.PathClusterCompleted)
 	cluster, err := r.loadCluster(clusterPath)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


justinsb:
This suggests that enum constants should be in the versioned API: I would have preferred to load a particular version like v1alpha2, but then I would have had to use the constants from a different version - the internal version - which would have been confusing.  If we put the enum constants into the API version, we can more easily map different values between versions if we have to (like with Master and ControlPlane).  cc @johngmyers

johngmyers:
@justinsb we should really get the remaining clouds off of the legacy node reconciler. It's just GCE and DigitalOcean left. Kops-controller shouldn't be reading the cluster or instance group specs in the first place.

justinsb:
@johngmyers yes, agreed & that is next on my plate :-)

But I think this might also need to be backported to prior versions, I'm not sure but I think I saw it there also.

upup/pkg/fi/utils/yaml.go
```diff
@@ -28,6 +28,8 @@ func YAMLToJSON(yamlBytes []byte) ([]byte, error) {
 }
 
 // YamlUnmarshal unmarshals the yaml content to an interface
+// Deprecated: normally you want something like kopscodecs.Decode,
+// so that we can convert apiVersions.
```
johngmyers, line 32: One doesn't need to and can't convert apiVersions for three uses of this: BootConfig, NodupConfig, and the `--values` flag of `kops toolbox template`. So I don't think it's appropriate to deprecate.

justinsb: Yes, that's fair.  Changing to a note.

cmd/kops-controller/controllers/legacy_node_controller.go
```diff
@@ -199,27 +198,28 @@ func (r *LegacyNodeReconciler) getInstanceGroupForNode(ctx context.Context, node
...
 	ttl := time.Hour
 
 	b, err := r.cache.Read(p, ttl)
 	if err != nil {
 		return nil, fmt.Errorf("error loading Cluster %q: %v", p, err)
 	}
 
-	o, _, err := kopscodecs.Decode(b, nil)
+	defaultGVK := api.SchemeGroupVersion.WithKind("Cluster")
+	o, _, err := kopscodecs.Decode(b, &defaultGVK)
```
johngmyers, line 211: Why do we need to pass a `defaultGVK`?

johngmyers:
I would be inclined to backpatch the new reconciler, but whatever.

Does kops-controller label control-plane nodes? If so, I could see annoying upgrade problems.

justinsb:
> I would be inclined to backpatch the new reconciler

I worry that it's too large a change to qualify for a backport.  We also want to make sure that e.g. instance group AdditionalLabels are supported, and that requires setting the cluster-autoscaler labels, which would be another backport etc

cmd/kops-controller/controllers/legacy_node_controller.go
justinsb: The intent is that this triggers the version-conversion logic, so we get either v1alpha2 or v1alpha3, but I'll verify this is doing what I think it's doing.

justinsb:
/retest

"Kubernetes e2e suite: [It] [sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: tmpfs] [Testpattern: Pre-provisioned PV (default fs)] volumes should store data" should be unrelated

johngmyers:
Landing this fix to 1.27 won't address the problem of 1.26 and previous kops-controllers mislabling 1.27 control plane nodes during an upgrade. We'd have to backport to previous versions and tell users to upgrade to the new previous version before upgrading across the 1.27 boundary. Or we'd have to tell users to do a disruptive upgrade of the control plane.

Longer term, we're going to want control-plane nodes to set their own labels and have kops-controller ignore control-plane nodes. This could also become an issue for the new node controller.

cmd/kops-controller/controllers/legacy_node_controller.go
johngmyers: By my reading, `kopscodecs.Decode()` always decodes to the unversioned protocol because it uses `Codecs.UniversalDecoder(kops.SchemeGroupVersion)`.

The version in `b` always takes precedence over `defaultGVK`. The completed-cluster is always written versioned, so I'm not seeing how `defaultGVK` is ever used.

justinsb:
Marking as blocks-next because if this bug is real, it means that control-plane nodes may be marked as node, which we don't want.

justinsb:
/retest

Failure seems unrelated

justinsb:
Good point on the complexities of a backport @johngmyers .  I propose we fix it going forwards first and then I'll go back and see how we can backport where we need to.

johngmyers:
@justinsb I agree with fixing this and think we should backport to 1.26 at least. I would like to remove the `defaultGVK`if possible as I believe that to be a red herring.

cmd/kops-controller/controllers/legacy_node_controller.go
justinsb: I'll give it a go, to be more homogenous with the other code.  We're type checking it anyway, so if we start seeing the type check fail with the wrong package, we'll know we need a deeper look!

johngmyers (approved):

PR #15611, opened 2023-07-09
gce: Set firewall rules for Internal LBs also

justinsb:
It seems we can use the exact same rules.

### DIFF
--- a/pkg/model/gcemodel/api_loadbalancer.go
+++ b/pkg/model/gcemodel/api_loadbalancer.go
@@ -38,7 +38,7 @@ var _ fi.CloudupModelBuilder = &APILoadBalancerBuilder{}
 
 // createPublicLB validates the existence of a target pool with the given name,
 // and creates an IP address and forwarding rule pointing to that target pool.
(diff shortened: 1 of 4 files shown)


justinsb:
cc @hakman I had to do this to get _any_ access (from the bastion), but I haven't found a nice way to actually test this properly.  Have you found any tricks?  (Maybe running kOps on a GCE VM in the same project/network?)

pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -215,8 +219,7 @@ func createInternalLB(b *APILoadBalancerBuilder, c *fi.CloudupModelBuilderContex
 			})
 		}
 	}
-
-	return nil
+	return b.addFirewallRules(c)
```
hakman, line 222: This may be better outside the `create*LB` methods.

hakman (commented):
Small nit, but otherwise nice find.
/lgtm
/approve
/hod in case you want to move the call to `b.addFirewallRules()` outside the `create*LB` functions.

PR #15617, opened 2023-07-11
azure: Add mode dependency logic to deletion

hakman:
/cc @justinsb @rifelpet

### DIFF
--- a/pkg/resources/azure/azure.go
+++ b/pkg/resources/azure/azure.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/kops/pkg/resources"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/azure"
-	"k8s.io/kops/upup/pkg/fi/cloudup/azuretasks"
 )
 
 const (
@@ -173,7 +172,7 @@ func (g *resourceGetter) toVirtualNetworkResource(vnet *network.VirtualNetwork) 
 	if vnet.Subnets != nil {
 		for _, sn := range *vnet.Subnets {
 			if sn.NetworkSecurityGroup != nil {
-				nsgID, err := azuretasks.ParseNetworkSecurityGroupID(*sn.NetworkSecurityGroup.ID)
+				nsgID, err := azure.ParseNetworkSecurityGroupID(*sn.NetworkSecurityGroup.ID)
 				if err != nil {
 					return nil, fmt.Errorf("parsing network security group ID: %s", err)
 				}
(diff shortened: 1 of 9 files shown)


pkg/resources/azure/azure.go
```diff
@@ -161,16 +164,35 @@ func (g *resourceGetter) listVirtualNetworksAndSubnets(ctx context.Context) ([]*
 	return rs, nil
 }
 
-func (g *resourceGetter) toVirtualNetworkResource(vnet *network.VirtualNetwork) *resources.Resource {
+func (g *resourceGetter) toVirtualNetworkResource(vnet *network.VirtualNetwork) (*resources.Resource, error) {
+	var blocks []string
+	blocks = append(blocks, toKey(typeResourceGroup, g.resourceGroupName()))
+
+	nsgs := map[string]struct{}{}
```
johngmyers, line 171: Why not use a `k8s.io/utils/set/set.Set[string]`?

hakman: I think it's just an awkward way to not use memory. Should probably change to bool, to keep things simple?

hakman: Interesting approach. Thanks.

pkg/resources/azure/azure.go
```diff
@@ -319,16 +342,26 @@ func (g *resourceGetter) toVMScaleSetResource(vmss *compute.VirtualMachineScaleS
...
 			}
-			vnets[subnetID.VirtualNetworkName] = struct{}{}
-			subnets[subnetID.SubnetName] = struct{}{}
+			vnets.Insert(subnetID.VirtualNetworkName)
+			subnets.Insert(subnetID.SubnetName)
+			if ip.LoadBalancerBackendAddressPools != nil {
+				for _, lb := range *ip.LoadBalancerBackendAddressPools {
+					lbID, err := azure.ParseLoadBalancerID(*lb.ID)
+					if err != nil {
+						return nil, fmt.Errorf("parsing load balancer ID: %s", err)
```
justinsb, line 360: Nit: should probably be `...: %w` (?)

upup/pkg/fi/cloudup/azure/azure_utils.go
```diff
@@ -40,3 +40,119 @@ func ZoneToAvailabilityZoneNumber(zone string) (string, error) {
...
+		s.ResourceGroupName,
+		s.VirtualNetworkName,
+		s.SubnetName)
+}
+
+// ParseSubnetID parses a given subnet ID string and returns a SubnetID.
+func ParseSubnetID(s string) (*SubnetID, error) {
+	l := strings.Split(s, "/")
+	if len(l) != 11 {
+		return nil, fmt.Errorf("malformed format of subnet ID: %s, %d", s, len(l))
```
justinsb, line 65: Nit: I normally prefer to print with %q to highlight the "wrong" value, and I'm not sure that printing the length really adds any information (though I know it is handy during development!)

justinsb:
Thanks @hakman 

/approve
/lgtm

upup/pkg/fi/cloudup/azure/azure_utils.go
hakman: It's just copy/paste from the original code, wanted to be easy diff.

pkg/resources/azure/azure.go
hakman: It's just copy/paste from the original code, wanted to be easy diff.

PR #15623, opened 2023-07-12
Improve validation of PodCIDR and ServiceClusterIPRange

### LINKED ISSUE #15034: Adjust ServiceClusterIPRange validation restrictions
Per conversations in #14963, #14997 and #15033.

Determine and adjust the validation restrictions on `ServiceClusterIPRange`, making them consistent for GCE and the other cloud providers.

Currently, cloud providers other than GCE require `ServiceClusterIPRange` be within `NonMasqueradeCIDR`. This should be required for either all cloud providers or none of them.

johngmyers:
Fixes #15034 

Removes the requirement (on non-GCE) that the ServiceClusterIPRange be within NonMasqueradeCIDR, as kube-proxy reroutes the ServiceClusterIPRange before masquerading.

Prohibits overlap between the PodCIDR and the ServiceClusterIPRange

Prohibits overlap between either the PodCIDR and the ServiceClusterIPRange and any subnet CIDR or IPv6CIDR. (except IPv6 doesn't have a PodCIDR, so can't overlap).

Adds validation of any podCIDR and requires it be within any NonMasqueradeCIDR.

### DIFF
--- a/pkg/apis/kops/validation/validation_test.go
+++ b/pkg/apis/kops/validation/validation_test.go
@@ -197,7 +197,7 @@ func TestValidateSubnets(t *testing.T) {
 			},
 		}
 		_, ipNet, _ := net.ParseCIDR(cluster.Networking.NetworkCIDR)
-		errs := validateSubnets(cluster, cluster.Networking.Subnets, field.NewPath("subnets"), true, &cloudProviderConstraints{}, []*net.IPNet{ipNet})
+		errs := validateSubnets(cluster, cluster.Networking.Subnets, field.NewPath("subnets"), true, &cloudProviderConstraints{}, []*net.IPNet{ipNet}, nil, nil)
 
 		testErrors(t, g.Input, errs, g.ExpectedErrors)
 	}
@@ -394,6 +394,7 @@ func Test_Validate_Networking_Flannel(t *testing.T) {
 				Networking: kops.NetworkingSpec{
 					NetworkCIDR:           "10.0.0.0/8",
 					NonMasqueradeCIDR:     "100.64.0.0/10",
+					PodCIDR:               "100.96.0.0/11",
 					ServiceClusterIPRange: "100.64.0.0/13",
 					Subnets: []kops.ClusterSubnetSpec{
 						{
@@ -460,6 +461,7 @@ func Test_Validate_AdditionalPolicies(t *testing.T) {
 			Networking: kops.NetworkingSpec{
 				NetworkCIDR:           "10.10.0.0/16",
 				NonMasqueradeCIDR:     "100.64.0.0/10",
+				PodCIDR:               "100.96.0.0/11",
 				ServiceClusterIPRange: "100.64.0.0/13",
 				Subnets: []kops.ClusterSubnetSpec{
 					{
--- a/pkg/apis/kops/validation/legacy.go
+++ b/pkg/apis/kops/validation/legacy.go
@@ -244,17 +244,6 @@ func validateServiceAccountIssuerDiscovery(c *kops.Cluster, said *kops.ServiceAc
 	return allErrs
 }
 
-// validateSubnetCIDR is responsible for validating subnets are part of the CIDRs assigned to the cluster.
-func validateSubnetCIDR(networkCIDRs []*net.IPNet, subnetCIDR *net.IPNet) bool {
-	for _, additionalNetworkCIDR := range networkCIDRs {
-		if subnet.BelongsTo(additionalNetworkCIDR, subnetCIDR) {
-			return true
-		}
-	}
-
-	return false
-}
-
 // DeepValidate is responsible for validating the instancegroups within the cluster spec
 func DeepValidate(c *kops.Cluster, groups []*kops.InstanceGroup, strict bool, cloud fi.Cloud) error {
 	if errs := ValidateCluster(c, strict); len(errs) != 0 {
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -293,23 +293,21 @@ func validateClusterSpec(spec *kops.ClusterSpec, c *kops.Cluster, fieldPath *fie
 }
 
 type cloudProviderConstraints struct {
-	requiresSubnets                                 bool
-	requiresNetworkCIDR                             bool
-	prohibitsNetworkCIDR                            bool
-	prohibitsMultipleNetworkCIDRs                   bool
-	requiresNonMasqueradeCIDR                       bool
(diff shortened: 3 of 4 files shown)


pkg/apis/kops/validation/validation_test.go
```diff
@@ -460,6 +461,7 @@ func Test_Validate_AdditionalPolicies(t *testing.T) {
 			Networking: kops.NetworkingSpec{
 				NetworkCIDR:           "10.10.0.0/16",
 				NonMasqueradeCIDR:     "100.64.0.0/10",
+				PodCIDR:               "100.96.0.0/11",
```
hakman, line 464: Shouldn't `PodCIDR` and `ServiceClusterIPRange` be both part of `NonMasqueradeCIDR`?

johngmyers: ServiceClusterIPRange doesn't have to be part of NonMasqueradeCIDR. That is because only things within the cluster network can talk to ServiceClusterIPRange and kube-proxy reroutes ServiceClusterIPRange before masquerading. GCE currently doesn't require ServiceClusterIPRange to be in NonMasqueradeCIDR; this PR drops that validation for all of the other cloud providers as well.

hakman: In this case, does is the NonMasqueradeCIDR still used for anything? It was dropped by kubelet when Docker shim was removed.

johngmyers: It's used by ContainerdBuilder and AWSCloudControllerManagerOptionsBuilder. It is also used as an exclusion for the egress proxy.

johngmyers: The use by AWSCloudControllerManagerOptionsBuilder looks to me like it's a bug. It should be using the PodCIDR instead.

hakman: Maybe we should use PodCIDR instead:
```
eccm.ClusterCIDR = clusterSpec.Networking.NonMasqueradeCIDR
```

johngmyers: We can handle AWSCloudControllerManagerOptionsBuilder as a separate PR. That one we might want to backport.

hakman (approved):

minkimipt:
we are running a number of production clusters that were setup with kops 1.18 or even earlier on AWS. We’ve been upgrading kops and k8s over the last years and we are currently on kops 1.27 and k8s 1.25. We are expanding our infra into other regions and in aim to keep our older clusters upgraded too. We’ve recently kicked off creation of new clusters, that we do directly with kops 1.28. We are following the same subnetting scheme, that we used to apply for all of our clusters, but we’ve hit some problems in kops 1.28, which leaves us in doubt how to continue further. Here’s the particular problem that we are facing when trying to create a new cluster:

```
Error: completed cluster failed validation: spec.networking.serviceClusterIPRange: Forbidden: serviceClusterIPRange "10.22.128.0/17" must not overlap podCIDR "10.22.128.0/17"
```

Those subnets indeed overlap, but that hasn’t been an issue until kops 1.28 and we’ve been able to track that change down to this [PR](https://github.com/kubernetes/kops/pull/15623), which introduced the serviceClusterIPRange check, i.e.:
```
			if subnet.Overlap(podCIDR, serviceClusterIPRange) {
				allErrs = append(allErrs, field.Forbidden(fldPath.Child("serviceClusterIPRange"), fmt.Sprintf("serviceClusterIPRange %q must not overlap podCIDR %q", serviceClusterIPRange, podCIDR)))
			}
```
Since our clusters used to work well while those subnets were overlapping, could we make this check optional? Is there any risk adding a command line parameters that disables this check? Will we hit any issues in kops 1.28 and later and k8s 1.28 if this check is not done?

@johngmyers may I ask you to elaborate about the reason of this change?

justinsb:
@minkimipt thanks for reporting and sorry about the issue.  IIRC when we merged this we believed this was not supported / caused other problems.  However, as it's a regression and existings clusters at least appear to be working fine, I think we should reduce this to a warning / disable it entirely while we track down exactly what goes wrong when these CIDRs overlap, and if it is indeed a problem, figure out how to move them.  I propose we track as https://github.com/kubernetes/kops/issues/16340

ebdekock:
> @minkimipt thanks for reporting and sorry about the issue. IIRC when we merged this we believed this was not supported / caused other problems. However, as it's a regression and existings clusters at least appear to be working fine, I think we should reduce this to a warning / disable it entirely while we track down exactly what goes wrong when these CIDRs overlap, and if it is indeed a problem, figure out how to move the CIDRs to be non-overlapping. I propose we track as #16340

@justinsb @johngmyers we are running into the same issue, except for the subnet range. We are trying to update a cluster where the subnet and service IP range overlap. We are hitting this validation, preventing the update:

https://github.com/kubernetes/kops/pull/15623/files#diff-ae412ac68b83570fd50e9d5d63873f060eb8f503f5e44be78d710076294c3285R623-R625

I have the same questions here:

> Since our clusters used to work well while those subnets were overlapping, could we make this check optional? Is there any risk adding a command line parameters that disables this check? Will we hit any issues in kops 1.28 and later and k8s 1.28 if this check is not done?

I see we've reverted the pod overlap validation: https://github.com/kubernetes/kops/pull/16345/commits/a1bba9dbc73e90d9d525a015a565389a03a61160
Are the subnets different or would the same be possible?

minkimipt:
@ebdekock we were able to upgrade kops to 1.28.5 where that validation was disabled in a cluster where we had those networks overlapping. Just to confirm that it's not a problem anymore.

ebdekock:
> @ebdekock we were able to upgrade kops to 1.28.5 where that validation was disabled in a cluster where we had those networks overlapping. Just to confirm that it's not a problem anymore.

I was waiting for the 1.28.5 release, as I thought it would also fix my issue, but our overlap is with the subnet CIDR and not the PodCIDR

ebdekock:
> I would say the "appear to be working fine" is merely that problems are rarely detected, despite being there. I would not be in favor of weakening the validations and would suggest the clusters' networking configurations be fixed.

Yup, that makes total sense and we agree that fixing the underlying issue is the best way forward. We would like to get the cluster updated and need to prioritize getting the networking fixed. Would it be possible to get this validation behind a temporary flag (that's off by default) to give us time to correct the issue?

ebdekock:
> I would say the "appear to be working fine" is merely that problems are rarely detected, despite being there. I would not be in favor of weakening the validations and would suggest the clusters' networking configurations be fixed.

We have fixed the networking like you suggested and no longer need this, thanks!

PR #15626, opened 2023-07-12
scaleway: refactoring: utils functions to get info from tags

Mia-Cross:
This PR adds 3 new utils functions to get:
- the cluster's name
- the instance's role
- the instance's instance group name

from the given tags.

This avoids a few `for _, tag := range tags` type loops.


johngmyers (approved):
This is a good refactor. You could refine it further with something like:

```go
func TagValue(server *instance.Server, prefix string) string {
```

PR #15627, opened 2023-07-13
azure: Add support for dns=none

hakman:
/cc @justinsb 
/assign @justinsb
/kind office-hours


upup/pkg/fi/cloudup/dns_test.go
```diff
@@ -30,84 +30,6 @@ func TestPrecreateDNSNames(t *testing.T) {
 		cluster  *kops.Cluster
 		expected []recordKey
 	}{
-		{
```
hakman, line 33: Azure doesn't do any DNS operations, so these test cases are useless.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1321,7 +1321,7 @@ func setupDNSTopology(opt *NewClusterOptions, cluster *api.Cluster) error {
 	switch strings.ToLower(opt.DNSType) {
 	case "":
 		switch cluster.Spec.GetCloudProvider() {
-		case api.CloudProviderHetzner, api.CloudProviderDO:
+		case api.CloudProviderHetzner, api.CloudProviderDO, api.CloudProviderAzure:
```
hakman, line 1324: We should move most, if not all cloud providers to `dns=none` soon.
Gossip is a much worse choice and Public DNS is only available on a few cloud providers.

upup/pkg/fi/cloudup/azuretasks/loadbalancer.go
```diff
@@ -151,6 +179,7 @@ func (*LoadBalancer) RenderAzure(t *azure.AzureAPITarget, a, e, changes *LoadBal
 			ID: to.StringPtr(fmt.Sprintf("/%s/virtualNetworks/%s/subnets/%s", idPrefix, *e.Subnet.VirtualNetwork.Name, *e.Subnet.Name)),
 		}
 	}
+	// TODO: Move hardcoded values to the model
```
hakman, line 182: Almost everything about LB is hardcoded. We should move this to model soon. For now it is good enough.

upup/pkg/fi/cloudup/dns_test.go
```diff
@@ -30,84 +30,6 @@ func TestPrecreateDNSNames(t *testing.T) {
...
-				{"api.cluster1.example.com", rrstype.A},
-				{"api.cluster1.example.com", rrstype.AAAA},
-				{"api.internal.cluster1.example.com", rrstype.AAAA},
-			},
-		},
-		{
-			cluster: &kops.Cluster{
-				Spec: kops.ClusterSpec{
-					API: kops.APISpec{
-						LoadBalancer: &kops.LoadBalancerAccessSpec{},
```
johngmyers, line 67: We need an AWS version of this test.

upup/pkg/fi/cloudup/dns_test.go
```diff
@@ -30,84 +30,6 @@ func TestPrecreateDNSNames(t *testing.T) {
...
-			},
-			expected: []recordKey{
-				{"api.internal.cluster1.example.com", rrstype.A},
-			},
-		},
-		{
-			cluster: &kops.Cluster{
-				Spec: kops.ClusterSpec{
-					API: kops.APISpec{
-						LoadBalancer: &kops.LoadBalancerAccessSpec{},
```
johngmyers, line 82: We need an AWS version of this test

upup/pkg/fi/cloudup/dns_test.go
```diff
@@ -30,84 +30,6 @@ func TestPrecreateDNSNames(t *testing.T) {
...
-			expected: []recordKey{
-				{"api.internal.cluster1.example.com", rrstype.AAAA},
-			},
-		},
-		{
-			cluster: &kops.Cluster{
-				Spec: kops.ClusterSpec{
-					API: kops.APISpec{
-						LoadBalancer: &kops.LoadBalancerAccessSpec{
-							UseForInternalAPI: true,
```
johngmyers, line 101: We need an AWS version of this test.

johngmyers (changes requested):

pkg/apis/kops/model/features.go
```diff
@@ -24,7 +24,7 @@ import (
 
 // UseKopsControllerForNodeBootstrap is true if nodeup should use kops-controller for bootstrapping.
 func UseKopsControllerForNodeBootstrap(cloudProvider kops.CloudProviderID) bool {
-	return cloudProvider != kops.CloudProviderAzure
+	return true
```
johngmyers, line 27: 👍 👍 👍

upup/pkg/fi/cloudup/azure/verifier.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &azureVerifier{
+		client: azureClient,
+	}, nil
+}
+
+func (a azureVerifier) VerifyToken(ctx context.Context, rawRequest *http.Request, token string, body []byte, useInstanceIDForNodeName bool) (*bootstrap.VerifyResult, error) {
```
rifelpet, line 55: This doesn't validate that the VM should actually be allowed to join the cluster, it seems to accept a token from any VM in the resource group. For comparison, the AWS verifier checks that the assumed role in the GetCallerIdentity response matches an IAM role used by any of the cluster's instance groups.

Can we do something similar here?

hakman: True. It checks at least to be in same resource group. Will check to see if I can lock down things more here.

upup/pkg/fi/cloudup/dns_test.go
hakman: Done

upup/pkg/fi/cloudup/dns_test.go
hakman: Done

upup/pkg/fi/cloudup/dns_test.go
hakman: Done

pkg/model/azuremodel/network.go
```diff
@@ -122,7 +124,32 @@ func (b *NetworkModelBuilder) Build(c *fi.CloudupModelBuilderContext) error {
 			SourceAddressPrefixes:    &k8sAccessIPv6,
 			SourcePortRange:          fi.PtrTo("*"),
 			DestinationAddressPrefix: fi.PtrTo("*"),
-			DestinationPortRange:     fi.PtrTo("443"),
+			DestinationPortRange:     fi.PtrTo(strconv.Itoa(wellknownports.KubeAPIServer)),
+		})
+	}
+	if b.Cluster.UsesNoneDNS() {
+		// TODO: Limit access to necessary source address prefixes instead of "0.0.0.0/0" and "::/0"
```
johngmyers, line 131: It would be better to limit this out of the gate. At the bare minimum, it should be limited to `networkCIDR`.

hakman: It can't be limited, because you have internal IPs (with private address) and external LB (with public address).
Whitelisting the networkCIDR does nothing.

upup/pkg/fi/cloudup/azure/verifier.go
hakman: There may be a more secure solution by using the attested data, though will be something for the future.
For now I added the VM ID, which should be pretty hard to guess without access to the resource group.
https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#attested-data

rifelpet: Can we verify the VM's VMSS is of a known kOps InstanceGroup?

hakman: That should be checked further along in the process. This is why we set `result.InstanceGroupName`.

hakman: It may be safer to take the IG name from the VMSS name, instead of the tag. WDYT?

justinsb:
This lgtm, and I think is a good net improvement for security already.

I do think we should check that the Azure VM is part of our cluster, but maybe we do that as a separate PR?  WDYT @rifelpet  / @hakman ?

/approve
/lgtm
/hold to agree whether we check the cluster as part of this PR or another

pkg/apis/kops/model/features.go
johngmyers: I'll take on removing the dead code this leaves behind.

rifelpet:
I'm fine with the security improvements happening in a follow up

hakman:
Thanks @justinsb & @rifelpet I will create a separate PR with a small idea and we can continue the discussion there.
In short, my impression was that after we set the IG name in the response, it will be used to get config for that IG name. If it doesn't exist, it will fail.
/hold cancel

PR #15352, opened 2023-04-27
Add support for MemorySwap

tomesm:
This PR adds support for `MemorySwapConfiguration` based on the following logic:

- If `memorySwapConfiguration.swapBehavior` is not set, there will be no change to the kubelet configuration.
- If `memorySwapConfiguration.swapBehavior` is set to an empty string, kubelet config will contain `memorySwap: ""`.
- If `memorySwapConfiguration.swapBehavior` is set to `"UnlimitedSwap"`, kubelet config will contain `memorySwap: UnlimitedSwap`.

Fixes: #14950

PR #15632, opened 2023-07-14
Add support for using swap memory

### LINKED ISSUE #14950: Swap Support
/kind feature

Kubernetes 1.22 introduced support for enabling swap on nodes. There are many usecases that benefit from this, so it would be great if Kops supported enabling this. 

Unfortunately while it appears that the `NodeSwap` feature gate and `failSwapOn: false` work as expected it appears that there is no way to modify the `MemorySwapConfiguration` in the kubelet config to change the default setting from `LimitedSwap` to `UnlimitedSwap`. This means that it is still not possible to actually use swap for Kubernetes workloads, only that it can be used for non-kubernetes processes running on the host.

Perhaps `kubelet` and `masterKubelet` could be extended with the following:
```
kubelet:
  memorySwapConfiguration:
    swapBehavior: "", "LimitedSwap", or "UnlimitedSwap"
```

Reference: 
- https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-MemorySwapConfiguration
- https://kubernetes.io/blog/2021/08/09/run-nodes-with-swap-alpha/

hakman:
Reduced version of #15352 (thanks the inspiration @tomesm)

Closes #14950

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -3808,6 +3808,10 @@ spec:
                       Kubelet.
                     format: int32
                     type: integer
+                  memorySwapBehavior:
+                    description: 'MemorySwapBehavior defines how swap is used by container
+                      workloads. Supported values: LimitedSwap, "UnlimitedSwap.'
+                    type: string
                   networkPluginMTU:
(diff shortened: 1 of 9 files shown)


justinsb:
Thanks @hakman 

/approve
/lgtm

PR #15643, opened 2023-07-16
WIP: prune dns-controller

justinsb:
When switching from gossip to dns=none, the rolling-update fails because the dns-controller addon is still present (because it isn't removed, as we never merged #13962).  However, the configuration for the dns-controller addon still assumes gossip is running on protokube (which it isn't).

WIP because there are a few ways to tackle this: we could implement addon removal (but I would rather we did it very carefully, perhaps with explicit removal).  We could simply keep dns-controller but not install it.  Or we could do what we do here - install a different (empty) manifest.

### DIFF
--- a/upup/models/cloudup/resources/addons/dns-controller.addons.k8s.io/not-enabled.yaml.template
+++ b/upup/models/cloudup/resources/addons/dns-controller.addons.k8s.io/not-enabled.yaml.template
@@ -1,0 +1,143 @@
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: dns-controller
+  namespace: kube-system
+  labels:
+    k8s-addon: dns-controller.addons.k8s.io
+    k8s-app: dns-controller
+    version: v{{ KopsVersion }}
+spec:
+  replicas: 1
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      k8s-app: dns-controller
+  template:
+    metadata:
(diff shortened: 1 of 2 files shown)


johngmyers:
Couldn't one start with gossip plus external-dns?

johngmyers:
I can't find validation preventing gossip + external-dns, but external-dns doesn't have gossip configuration support.

johngmyers:
Still, switching from external-dns to none should probably uninstall external-dns.

johngmyers:
Okay, this is confusing. There's `ExternalDNSProviderNone`, which is different than `DNSTypeNone`. Changing the `ExternalDNS.Provider` should probably prune any previous provider, though this is arguably a separate PR.

hakman:
DNSTypeNone means no DNS work at all. Anyone that wants to use external DNS is free to deploy it on its own.

zetaab:
@justinsb fyi:


in case of AWS there is some downtime:
Screenshot 2023-07-16 at 20 48 51

In case of AWS the nodes will lose its connection to control plane.

OpenStack migration is almost working:

Screenshot 2023-07-16 at 20 49 54

I have executed gossip -> dns none to about 100+ openstack clusters. The only problem that I saw was that kops-controller goes to crashloopbackoff and that will block the rolling-update. The solution for that was that I just terminated the instance where that was located and when new controlplane instance started the problem was gone.

PR #15645, opened 2023-07-16
Remove references to more ClusterSpec fields from nodeup

johngmyers:


upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -1411,6 +1411,7 @@ func (n *nodeUpConfigBuilder) BuildConfig(ig *kops.InstanceGroup, apiserverAddit
 		}
 
 		if isMaster || usesLegacyGossip {
+			config.Channels = n.channels
```
hakman, line 1414: I think this is needed only on master nodes.

johngmyers: I didn't see logic to keep it from being passed on gossip worker nodes

hakman (approved):

PR #15647, opened 2023-07-16
Remove more references to ClusterSpec fields from nodeup

johngmyers:


hakman (approved):

doryer:
@johngmyers I think the commit should be called `Remove more references to nodeup from ClusterSpec fields` instead of `Remove more references to ClusterSpec fields from nodeup`

PR #15650, opened 2023-07-17
Support removal of managed node labels

justinsb:
kops-controller manages a few node-role node-labels.  We
now remove any extra managed labels that land on the node.

This means we will now actively remove the extra node label if we
previously erroneously applied to a control-plane node; previous code
changes stopped applying it.

### DIFF
--- a/cmd/kops-controller/controllers/legacy_node_controller.go
+++ b/cmd/kops-controller/controllers/legacy_node_controller.go
@@ -132,12 +132,23 @@ func (r *LegacyNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		}
 	}
 
-	if len(updateLabels) == 0 {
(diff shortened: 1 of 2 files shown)


cmd/kops-controller/controllers/legacy_node_controller.go
```diff
@@ -132,12 +132,23 @@ func (r *LegacyNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	}
 
-	if len(updateLabels) == 0 {
+	deleteLabels := make(map[string]struct{})
```
johngmyers, line 135: Why not use a `k8s.io/utils/set/set.Set[string]`?

johngmyers (approved):
Pity there are no tests for any of this stuff.

PR #15659, opened 2023-07-18
azure: Verify node identity using VMSS name instead of tags

hakman:
Continue discussion from https://github.com/kubernetes/kops/pull/15627#discussion_r1264531162.

/cc @rifelpet @justinsb @johngmyers

### DIFF
--- a/upup/pkg/fi/cloudup/azure/verifier.go
+++ b/upup/pkg/fi/cloudup/azure/verifier.go
@@ -28,15 +28,16 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2022-05-01/network"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"k8s.io/kops/pkg/bootstrap"
-	"k8s.io/kops/pkg/nodeidentity/azure"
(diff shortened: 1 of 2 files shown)


pkg/apis/kops/model/features.go
```diff
@@ -31,6 +31,8 @@ func UseChallengeCallback(cloudProvider kops.CloudProviderID) bool {
 		return true
 	case kops.CloudProviderScaleway:
 		return true
+	case kops.CloudProviderAzure:
+		return true
```
rifelpet, line 35: Does this require additional NetworkSecurityGroup rules to allow access from the control plane VMs to the nodes?

hakman: Was just looking at that. Updated with a new rule to allow it.

justinsb:
This LGTM, but we should give @rifelpet time to look again today.

/approve
/lgtm
/hold in case @rifelpet sees anything

rifelpet:
/hold cancel

Looks good

hakman:
Thanks @rifelpet @justinsb 🎉

PR #15660, opened 2023-07-18
Update channels

hakman:


johngmyers:
Shouldn't we wait a bit before updating stable?

hakman:
> Shouldn't we wait a bit before updating stable?

* images are tested in this PR
* only bumped kOps to 1.26.5 in stable

What would you prefer to hold back?

moshevayner (approved):

PR #15667, opened 2023-07-19
azure: Populate node labels from tags

hakman:
/cc @justinsb @rifelpet

### DIFF
--- a/pkg/nodeidentity/azure/identify.go
+++ b/pkg/nodeidentity/azure/identify.go
@@ -26,7 +26,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	expirationcache "k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
+	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/nodeidentity"
+	"k8s.io/kops/pkg/nodelabels"
+	"k8s.io/kops/upup/pkg/fi/cloudup/azure"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/nodeidentity/azure/identify.go
```diff
@@ -100,24 +103,41 @@ func (i *nodeIdentifier) IdentifyNode(ctx context.Context, node *corev1.Node) (*
...
+		if k == azure.TagClusterName && v != nil {
+			labels[kops.LabelClusterName] = *v
+		}
+		if k == InstanceGroupNameTag && v != nil {
+			labels[kops.NodeLabelInstanceGroup] = *v
+		}
+		if strings.HasPrefix(k, azure.TagNameRolePrefix) {
+			role := strings.TrimPrefix(k, azure.TagNameRolePrefix)
+			switch {
+			case strings.HasSuffix(nodelabels.RoleLabelControlPlane20, role):
```
justinsb, line 116: Should this just be checking against the instancegroup roles (and I think you've already stripped the prefix)?

https://github.com/kubernetes/kops/blob/d5c24585187ddb50647fa939cb4e7fa8d997b077/pkg/model/azuremodel/context.go#L117C1-L121

justinsb:
Thanks @hakman - and thanks for tweaking the code I didn't grok!

/approve
/lgtm

pkg/nodeidentity/azure/identify.go
hakman: Azure tag is in the format `"k8s.io_role_control-plane": "1"`, so we have to trim the `k8s.io_role_` prefix.
I agree that it looks better now with the `InstanceGroupRoleX.ToLowerString()`.

PR #15670, opened 2023-07-19
Fix AWS CCM defaults for IPAM to match KCM

johngmyers:
When reviewing #15623 we noticed that AWS CCM was incorrectly defaulting ClusterCIDR to NonMasqueradeCIDR. Updating the CCM code to match that for KCM.

GCP networking doesn't work on AWS, so removing that code.

I'm not sure whether we should default UseServiceAccountCredentials to true like KCM does.

/cc @hakman


johngmyers:
KCM also has code for defaulting NodeCIDRMaskSize when ClusterCIDR is IPv6, but we never set ClusterCIDR for IPv6. I don't think we need to copy that.

PR #15672, opened 2023-07-19
Bump k8s and kops versions used in scenario scripts

rifelpet:
Some of these were failing because the versions used were too old:

https://testgrid.k8s.io/kops-misc#kops-aws-addon-resource-tracking

```
 Error: kubernetes upgrade is required 
```

This bumps the scenario scripts to use newer versions.


johngmyers:
I'm not finding test-infra jobs referencing either of the cilium-connectivity-test or digital-ocean scenarios. Should these be removed instead?

PR #15678, opened 2023-07-20
v1alpha3: remove redundant ConfigStore

johngmyers:


hakman:
I would say that ConfigBase is the redundant one. Everything else has the Store suffix...

johngmyers:
Many fewer places use ConfigStore; it only affects ACLs.  As it derives from ConfigBase it is unlikely anyone would set it.

My plan is to change to a `configStore` sub-struct: `configStore.base`, `configStore.keypairs`, and `configStore.secrets`.

hakman (approved):

Issue #15684, opened 2023-07-21
S3 config lost in openstack for worker nodes


pcmid:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
```
kops version
Client version: 1.27.0
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"archive", BuildDate:"2023-06-15T08:14:06Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.6", GitCommit:"11902a838028edef305dfe2f96be929bc4d114d8", GitTreeState:"clean", BuildDate:"2023-06-14T09:49:08Z", GoVersion:"go1.19.10", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
`openstack`

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
```
S3_ENDPOINT=http://ceph_rgw_url S3_ACCESS_KEY_ID=XXX S3_SECRET_ACCESS_KEY=XXX kops --name name.k8s.local --state do://kops rolling-update cluster --yes
```
> The `--state` parameter starts with `do://` due to https://github.com/kubernetes/kops/issues/9926

**5. What happened after the commands executed?**
Timeout to wait woker node join cluster.

**6. What did you expect to happen?**
The worker node successfully joins the cluster

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2023-06-11T11:51:45Z"
  generation: 4
  name: name.k8s.local
spec:
  api:
    loadBalancer:
      type: Public
  authorization:
    rbac: {}
  certManager:
    defaultIssuer: letsencrypt-prod
    enabled: true
  channel: stable
  cloudConfig:
    openstack:
      blockStorage:
        bs-version: v3
        clusterName: name.k8s.local
[...]

PR #15691, opened 2023-07-22
Pass `S3_*` env vars to nodes when using Gossip

hakman:
Closes https://github.com/kubernetes/kops/issues/15684


johngmyers (commented):
Can we suppress the credentials entirely when the cloud provider has an instance role or equivalent?

We should be moving the non-credential stuff to the cluster spec.

hakman:
> Can we suppress the credentials entirely when the cloud provider has an instance role or equivalent?
> We should be moving the non-credential stuff to the cluster spec.

Not sure what you mean, but closing this for now. It should not be needed.

PR #15694, opened 2023-07-23
update logs with "cannot render instance groups" instead of "cannot render nodes" in validate_cluster.go

haojue:
update logs with "cannot render instance groups" instead of "cannot render nodes"

### DIFF
--- a/cmd/kops/validate_cluster.go
+++ b/cmd/kops/validate_cluster.go
@@ -264,7 +264,7 @@ func validateClusterOutputTable(result *validation.ValidationCluster, cluster *k
 	fmt.Fprintln(out, "INSTANCE GROUPS")
 	err := t.Render(instanceGroups, out, "NAME", "ROLE", "MACHINETYPE", "MIN", "MAX", "SUBNETS")
 	if err != nil {
-		return fmt.Errorf("cannot render nodes for %q: %v", cluster.Name, err)
+		return fmt.Errorf("cannot render instance groups for %q: %w", cluster.Name, err)
 	}
 
 	{


cmd/kops/validate_cluster.go
```diff
@@ -222,7 +222,7 @@ func RunValidateCluster(ctx context.Context, f *util.Factory, out io.Writer, opt
 					time.Sleep(pollInterval)
 					continue
 				} else {
-					return nil, fmt.Errorf("cluster passed validation %d consecutive times", consecutive)
+					return nil, fmt.Errorf("cluster passed validation %d consecutive times, failed because of negative options.wait %d", consecutive, options.wait)
```
johngmyers, line 225: This message is inaccurate. Options.wait is not necessarily negative. 

The lack of a wait option is not a cause of the cluster's failure to pass validation.

johngmyers (changes requested):

cmd/kops/validate_cluster.go
haojue: Hi @johngmyers , you mean when it fails, Options.wait is not necessarily negative.?  I just noticed ` if options.wait > 0 {` , so what's your recommended error message for it? thanks

cmd/kops/validate_cluster.go
```diff
@@ -264,7 +264,7 @@ func validateClusterOutputTable(result *validation.ValidationCluster, cluster *k
 	fmt.Fprintln(out, "INSTANCE GROUPS")
 	err := t.Render(instanceGroups, out, "NAME", "ROLE", "MACHINETYPE", "MIN", "MAX", "SUBNETS")
 	if err != nil {
-		return fmt.Errorf("cannot render nodes for %q: %v", cluster.Name, err)
+		return fmt.Errorf("cannot render instancegroups for %q: %v", cluster.Name, err)
```
haojue, line 267: Hi @johngmyers , this is ok to you?

johngmyers:
```suggestion
		return fmt.Errorf("cannot render instance groups for %q: %w", cluster.Name, err)
```

cmd/kops/validate_cluster.go
hakman: > so what's your recommended error message for it? thanks

I is recommended to leave it as it is, without changes.

haojue:
Hi @johngmyers , thank you for your review comments, I updated per your comments.

johngmyers:
@haojue please change the commit and PR descriptions to match the remaining change.

haojue:
Hi @johngmyers, I updated the description and the commit message

johngmyers (approved):
Thanks!

PR #15709, opened 2023-07-28
gce: Set labels on ForwardingRules

justinsb:
We add the cluster-name label, now that labels are supported on
ForwardingRules.

### DIFF
--- a/cloudmock/gce/mockcompute/forwarding_rule.go
+++ b/cloudmock/gce/mockcompute/forwarding_rule.go
@@ -71,6 +71,26 @@ func (c *forwardingRuleClient) Insert(project, region string, fr *compute.Forwar
 	return doneOperation(), nil
 }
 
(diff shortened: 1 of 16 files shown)


hakman (approved):

justinsb:
Both look like dns lookup errors; unrelated (?)

/retest

justinsb:
Looks like we can't set labels on creation, which is ... annoying.  Added code to set them explicitly immediately after creation.

hakman:
> Looks like we can't set labels on creation, which is ... annoying. Added code to set them explicitly immediately after creation.

Pretty annoying also that you have to re-read the resource. Would be nice to at least return the fingerprint as part of the result of the create.

justinsb:
> Pretty annoying also that you have to re-read the resource. 

Agreed, and technically we can get it from the operation, it's just a pain to do so.

hakman (approved):

Issue #15679, opened 2023-07-20
GCE second cluster is given first cluster's API endpoint IP


codingkriggs:
/kind bug

Note, I have used substitutes `<bucket>` and `<project>` to hide private information in this report.

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

```
$ kops version
Client version: 1.27.0 (git-v1.27.0)
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

```
$ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"clean", BuildDate:"2023-06-14T09:53:42Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"clean", BuildDate:"2023-06-14T09:47:40Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**

GCE

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

`$ kops create cluster cluster1.k8s.local --zones us-west1-c --state gs://<bucket>/ --project=<project> --cloud gce --yes ; echo ---------- ; kops create cluster cluster2.k8s.local --zones us-west1-c --state gs://<bucket>/ --project=<project> --cloud gce --yes`

**5. What happened after the commands executed?**

A warning line in the output reads (I modified the IPs slightly):

`W0720 05:18:57.220910    5431 create_kubecfg.go:72] Found multiple API endpoints ([34.82.159.115 35.227.35.198]), choosing arbitrarily`

when I view the contents of ~/.kube/config, I have (some bits truncated with "[...]"):

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: [...]
    server: https://34.82.159.115
    tls-server-name: api.internal.cluster1.k8s.local
  name: cluster1.k8s.local
- cluster:
[...]

PR #15711, opened 2023-07-28
GCE: Fix issue around multiple clusters in project

justinsb:
We now match the forwardingRules explicitly by name.  In another PR, we add labels to them.

Issue #15679


justinsb:
As we can match by label, I believe this PR is no longer needed.

Refactoring the names into its own package might be handy later, but if we need it, it's always there!

/close

PR #15714, opened 2023-07-29
gce: Use dns=none for newly created clusters

hakman:
/cc @justinsb @rifelpet 
/assign @justinsb

### DIFF
--- a/pkg/model/gcemodel/api_loadbalancer.go
+++ b/pkg/model/gcemodel/api_loadbalancer.go
@@ -112,7 +112,7 @@ func (b *APILoadBalancerBuilder) addFirewallRules(c *fi.CloudupModelBuilderConte
 		b.AddFirewallRulesTasks(c, "https-api", &gcetasks.FirewallRule{
 			Lifecycle:    b.Lifecycle,
 			Network:      network,
-			SourceRanges: b.Cluster.Spec.API.Access,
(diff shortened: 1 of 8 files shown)


pkg/model/gcemodel/api_loadbalancer.go
```diff
@@ -112,7 +112,7 @@ func (b *APILoadBalancerBuilder) addFirewallRules(c *fi.CloudupModelBuilderConte
 		b.AddFirewallRulesTasks(c, "https-api", &gcetasks.FirewallRule{
 			Lifecycle:    b.Lifecycle,
 			Network:      network,
-			SourceRanges: b.Cluster.Spec.API.Access,
+			SourceRanges: []string{"0.0.0.0/0"},
```
hakman, line 115: @justinsb The GCE resolver looks for the IP address of the control plane nodes and this is how it manages to bootstrap the node. This is possible to GCE due to IAM roles, but not possible in smaller cloud providers.

In general, I would prefer to not resolve at all and allow access to the LB (external or internal address).

Also, after bootstrap, the node has to connect to the API server, to join the cluster, which needs  access to the LB (external or internal address). Is there a better way to do this than adding a NAT gateway with a static IP address?

hakman:
/close

PR #15153, opened 2023-02-14
Add terraform target support for configuring Warm Pool

rsafonseca:
Addresses https://github.com/kubernetes/kops/issues/11633

PR #15722, opened 2023-07-30
Fix WarmPool with --target direct

rifelpet:
Fixes https://testgrid.k8s.io/kops-misc#kops-warm-pool

I believe kops hasn't created the nodes ASG because it interpreted the WarmPool field as a dependency (introduced in https://github.com/kubernetes/kops/pull/15153), so it tries to create the WarmPool before the ASG.

The toolbox dump doesn't mention any instances from the nodes ASG: https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-warm-pool/1685665134501957632/artifacts/toolbox-dump.yaml

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
+++ b/upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
@@ -110,6 +110,31 @@ func (e *AutoscalingGroup) CompareWithID() *string {
 	return e.Name
 }
 
+// Track dependencies here to explicitly ignore WarmPool
+// because the WarmPool should be created after the ASG, not the other way around.
+// The WarmPool struct field is only used for RenderTerraform.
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/cloudup/awstasks/autoscalinggroup.go
```diff
@@ -110,6 +110,37 @@ func (e *AutoscalingGroup) CompareWithID() *string {
 	return e.Name
 }
 
+// Track dependencies here to explicitly ignore WarmPool
+// because the WarmPool should be created after the ASG, not the other way around.
+// The WarmPool struct field is only used for RenderTerraform.
+func (e *AutoscalingGroup) GetDependencies(tasks map[string]fi.CloudupTask) []fi.CloudupTask {
+	var deps []fi.CloudupTask
+
+	if e.LoadBalancers != nil {
```
johngmyers, line 119: The `if` statements aren't necessary; one can `range` over a `nil` slice.

johngmyers (approved):

PR #15726, opened 2023-08-01
Don't set object-level public ACL in S3 FileRepository

johngmyers:
As mentioned in a comment on #14991 

S3 is moving away from object-level ACLs, so we should stop trying to set them. This removes code that set public-read ACLs on objects written to the FileRepository.


hakman (approved):
Should we cherry-pick this to 1.27?

johngmyers:
It's not a particularly high-value change. I wouldn't object to backporting, but think it's under the threshold.

PR #15705, opened 2023-07-28
Bump to Cilium 1.14

zadjadr:
I would love some help with the PR.

Stuff still missing: 

- [x] Manifest hashes (how to get them?)
- [x] First upgrade Cilium to v1.13
- [x] Deploy working cluster using roling update
- [x] Deploy working cluster from scratch
- [x] Allow wireguard + L7 proxy
- [x] Test hubble-relay

Can't test myself:

- [ ] AWS ENI stuff (new lifecycle.postStart for cilium-agent)

Better done in separate PRs: 
- [ ] Test wireguard + L7 proxy
- [ ] Add new functionality (e.g L2 Announcement, bgp, etc.)

Not 100% related changes:

- Switched from mounting /var/run/cilium in the hubble-relay deployment to using a service and exposing the hubble-peer port in the cilium agent (better for security, better practice etc.)

PR #15730, opened 2023-08-01
Bump cilium to v1.13.5

zadjadr:
Unfortunately, Cilium 1.14 is not ready yet (https://github.com/kubernetes/kops/pull/15705)

So here is a bump to 1.13.5

Tested this on a K8s Deployment in Openstack with Cilium ETCD.

image

image

### DIFF
--- a/tests/integration/update_cluster/minimal-ipv6-cilium/data/aws_s3_object_minimal-ipv6.example.com-addons-networking.cilium.io-k8s-1.16_content
+++ b/tests/integration/update_cluster/minimal-ipv6-cilium/data/aws_s3_object_minimal-ipv6.example.com-addons-networking.cilium.io-k8s-1.16_content
@@ -26,6 +26,7 @@ metadata:
 
 apiVersion: v1
 data:
+  agent-health-port: "9879"
   auto-direct-node-routes: "false"
   bpf-ct-global-any-max: "262144"
   bpf-ct-global-tcp-max: "524288"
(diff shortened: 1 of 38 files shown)


upup/pkg/fi/cloudup/bootstrapchannelbuilder/cilium.go
```diff
@@ -35,7 +35,7 @@ func addCiliumAddon(b *BootstrapChannelBuilder, addons *AddonList) error {
 		klog.Infof("found cilium (%q) in addons; won't use builtin", key)
 	} else {
 		id := "k8s-1.16"
-		location := key + "/" + id + "-v1.12.yaml"
+		location := key + "/" + id + "-v1.13.5.yaml"
```
hakman, line 38:
```suggestion
		location := key + "/" + id + "-v1.13.yaml"
```

zadjadr: Fixed!

tests/integration/update_cluster/minimal-ipv6-cilium/data/aws_s3_object_minimal-ipv6.example.com-addons-networking.cilium.io-k8s-1.16_content
```diff
@@ -26,6 +26,7 @@ metadata:
 
 apiVersion: v1
 data:
+  agent-health-port: "9879"
```
hakman, line 29: Out of curiosity, isn't this the default? Why do we need to set it explicitly?

johngmyers:
Why does the second commit revert the template change for `routing-mode: "native"` ?

zadjadr:
@johngmyers I believe the `routing-mode` is something that was added in 1.14.0 and this PR is based on my changes in the Bump to 1.14.0 PR, so I removed it for now

johngmyers:
It is misleading to have semantic changes in a commit labeled `hack/update-expected.sh`. Such commits should only have changes produced by said script.

zadjadr:
@johngmyers Hope it is fine now

PR #15735, opened 2023-08-02
OpenStack: add server group name override annotation

### LINKED ISSUE #15734: ControlPlane anti affinity on OpenStack cluster with one AZ
/kind feature

As discussed with @zetaab in the kops-users Slack channel, there is currently no way to have instance anti affinity between control plane nodes, if the target OpenStack environment only has 1 AZ.

kOps currently requires control plane IGs to have a min and max size of 1 -> Resulting in 3 controlplane IGs
These IGs create their own separate server groups with 1 instance in them.
If the target OpenStack environment only has 1 AZ, these server groups are useless.

kOps should support anti affinity between controlplane IGs, even if there is only one AZ.

Maybe allow the specification of a custom server group name via an annotation like: "openstack.kops.io/servergroupname"
This would allow the operator to specify the same "openstack.kops.io/servergroupname" for every controlplane IG.

zetaab:
fixes #15734

makes it possible to override server group name used in instance group.

Example:

```yaml
kind: InstanceGroup
metadata:
  annotations:
    openstack.kops.io/serverGroupName: control-plane
```

### DIFF
--- a/docs/getting_started/openstack.md
+++ b/docs/getting_started/openstack.md
@@ -123,29 +123,6 @@ kops update cluster my-cluster.k8s.local --state ${KOPS_STATE_STORE} --yes
 
 kOps should create instances to all three zones, but provision volumes from the same zone.
 
-## Using external cloud controller manager
-
-If you want to use [External CCM](https://github.com/kubernetes/cloud-provider-openstack) in your installation, this section contains instructions what you should do to get it up and running.
-
-Create cluster without `--yes` flag (or modify existing cluster):
-
-```bash
-kops edit cluster <cluster>
-```
-
-Add the following to clusterspec:
-
-```yaml
-spec:
-  cloudControllerManager: {}
-```
-
-Finally, update the cluster:
-
-```bash
-kops update cluster --name <cluster> --yes
-```
-
 ## Using CCM created Loadbalancers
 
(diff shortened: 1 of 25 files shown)


docs/getting_started/openstack.md
```diff
@@ -123,29 +123,6 @@ kops update cluster my-cluster.k8s.local --state ${KOPS_STATE_STORE} --yes
 
 kOps should create instances to all three zones, but provision volumes from the same zone.
 
-## Using external cloud controller manager
```
zetaab, line 126: default behaviour is ccm in openstack, so this is not needed

pkg/model/openstackmodel/servergroup.go
```diff
@@ -268,24 +268,36 @@ func (b *ServerGroupModelBuilder) Build(c *fi.CloudupModelBuilderContext) error
...
+		}
+
+		sgTask, ok := sgs[sgName]
+		if !ok {
+			igMap := make(map[string]*int32)
+			igMap[ig.Name] = ig.Spec.MaxSize
+			sgTask = &openstacktasks.ServerGroup{
+				Name:        s(sgName),
+				ClusterName: s(clusterName),
+				IGMap:       igMap,
```
zetaab, line 284: the problem with this was that earlier one servergroup was mapped to one instancegroup. However, now when its possible to customise the servergroup name, it can be that multiple instancegroups are mapped to single servergroup. It makes the logic little bit advanced.

hakman (approved):

PR #15739, opened 2023-08-04
Stop skipping some cilium tests on newer k8s versions

rifelpet:
It would be most accurate to skip based on kops version but we dont have access to that, so keying off of the k8s version is better than always skipping these tests


rifelpet:
These are the only two cilium jobs that use CI builds of k8s (1.28)

/test pull-kops-e2e-aws-cloud-controller-manager
/test pull-kops-e2e-k8s-gce-ci

rifelpet:
superseded by https://github.com/kubernetes/kops/pull/15753

PR #15743, opened 2023-08-05
Mark flags as deprecated instead of normalizing

hakman:
/cc @johngmyers @justinsb


justinsb:
We discussed in office hours, lgtm - thanks @hakman !

/approve
/lgtm

PR #15745, opened 2023-08-05
Use localhost for k8s api address in cert-manager pods

rifelpet:
The aws-load-balancer-controller [tests are failing](https://testgrid.k8s.io/kops-misc#kops-aws-aws-load-balancer-controller) because the cert-manager pods are [timing out trying to reach the k8s api](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-aws-load-balancer-controller/1687699439508000768/artifacts/cluster-info/kube-system/cert-manager-cainjector-7fbb85f9c6-b6jkn/logs.txt) via the service IP.

Since all cert-manager pods have hard affinity towards control plane nodes it makes sense to use localhost, similar to [what we did for AWS CCM](https://github.com/kubernetes/kops/pull/11939/commits/af0aefd2e7b1ec07730a843e43e93f4527b6094c)

### DIFF
--- a/tests/integration/update_cluster/aws-lb-controller/data/aws_s3_object_minimal.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/aws-lb-controller/data/aws_s3_object_minimal.example.com-addons-bootstrap_content
@@ -41,7 +41,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.16
     manifest: certmanager.io/k8s-1.16.yaml
-    manifestHash: 1d9ddc6d025438aa22182e434f9534e983ddefa857913ef0a456d9cf9cea360b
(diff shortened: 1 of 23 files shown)


rifelpet:
Ah right, we'd need to run these as hostNetwork. I'm debating if thats a good idea vs digging in to why they can't reach the API over the service IP.

hakman:
> Ah right, we'd need to run these as hostNetwork. I'm debating if thats a good idea vs digging in to why they can't reach the API over the service IP.

I think the reason this is failing is that we are using AWS VPC CNI in this scenario and that is incompatible with Ubuntu 22.04.
https://github.com/kubernetes/kops/blob/4fcd5943e62b377d52685e4afc3a53f75f088184/tests/e2e/scenarios/aws-lb-controller/run-test.sh#L24

PR #15753, opened 2023-08-07
Trim e2e skip regexes for Cilium

### LINKED ISSUE #15448: Reduce the list of skipped e2e test for Cilium
https://github.com/kubernetes/kops/blob/f2fe04438f322b0b248ed8ab49b5105cf20ae106/tests/e2e/pkg/tester/skip_regex.go#L60-L85

Some of them got fixed in 
https://github.com/cilium/cilium/pull/24202
https://github.com/cilium/cilium/pull/24174

you can compare against the regex in cilium repo 
https://github.com/cilium/cilium/pull/25258/files#diff-01b72d17856efb38bd2914aec50791eef6b9bc8238b6975d12f87b39cb24b008R214

Feel free to tag me to review the PR

johngmyers:
Closes #15448

### DIFF
--- a/tests/e2e/pkg/tester/skip_regex.go
+++ b/tests/e2e/pkg/tester/skip_regex.go
@@ -34,6 +34,12 @@ func (t *Tester) setSkipRegexFlag() error {
 		return nil
 	}
 
+	kopsVersion, err := t.getKopsVersion()
+	if err != nil {
+		return err
+	}
+	isPre28 := kopsVersion < "1.28"
+
 	cluster, err := t.getKopsCluster()
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


rifelpet:
The kops master branch is used to build kubetest2-kops for all e2e jobs including upgrade tests that upgrade to older kops versions that still use cilium 1.12. This change will stop skipping those tests for cilium 1.12 clusters which will cause failures. 

https://github.com/kubernetes/test-infra/blob/17464456700f8d6158e5957e81b2472d6eb0e02a/config/jobs/kubernetes/kops/kops-periodics-upgrades.yaml#L17-L22

Maybe we could update build_jobs.py to clone the release branch for upgrade tests that upgrade to non-latest kops versions?

hakman:
K8s 1.28 will be released a week from now. I think what @rifelpet was doing with checking K8s version instead of kOps version was pretty good. We already do checks against `k8sVersion.Minor`.

tests/e2e/pkg/tester/skip_regex.go
```diff
@@ -34,6 +34,12 @@ func (t *Tester) setSkipRegexFlag() error {
 		return nil
 	}
 
+	kopsVersion, err := t.getKopsVersion()
+	if err != nil {
+		return err
+	}
+	isPre28 := kopsVersion < "1.28"
```
rifelpet, line 41: nit: This will evaluate to false for any future 1.28 pre-release builds

rifelpet: And we can't inspect the Cluster manifest because `spec.networking.cilium.version` isn't guaranteed to be set. I suppose one alternative would be to inspect the daemonset itself with the kubectl binary that the tester downloads: https://github.com/kubernetes/kops/blob/7e06e4c4162cd7d5229c227f89815977934f8fc4/tests/e2e/pkg/tester/kubectl.go#L132-L134

johngmyers: Evaluating false for future 1.28 pre release builds is desired behavior

johngmyers:
Checking k8s version has the undesirable behavior of skipping those tests on kops 1.28+ branch kops and <1.28 k8s

hakman:
> Checking k8s version has the undesirable behavior of skipping those tests on kops 1.28+ branch kops and <1.28 k8s

In my opinion, that is irrelevant. We aim for most maintainable code here, not for exact skip list.

johngmyers:
Being able to remove the extra skips earlier leads to more maintainable code.

rifelpet (approved):

PR #15755, opened 2023-08-09
Remove use of ClusterSpec in nodeup

johngmyers:
Finishes the refactoring of nodeup to no longer directly reference the ClusterSpec.


hakman:
May be needed to force push or something to fix the netlify failures. Not sure if you can just re-trigger those.
/lgtm

hakman (approved):

PR #15761, opened 2023-08-10
aws: Default ENABLE_PREFIX_DELEGATION=true with AWS VPC CNI

hakman:
Ref https://github.com/kubernetes/kops/issues/15759


hakman:
`pull-kops-e2e-cni-amazonvpc-ipv6` is supposed to fail, so nothing to worry about.

hakman:
This was useful for the scalability test. If anyone thinks we should still change the default, please reopen.

PR #15785, opened 2023-08-16
aws: implement paginator for DescribeLaunchTemplate on buildKarpenterGroup

ltellesfl:
That's a fix for buildKarpenterGroup func.

That's an edge case that happens only when an AWS account has more than 200 launch templates (the max per page) and the one with Karpenter as a manager is on the second page. 

The aws-sdk ec2 func DescribeLaunchTemplates does not return directly a single launch template given the filters if the launch template is on the second page. Instead of that the return is empty with a nextToken, and the paginator fixes that behavior.

Without the paginator the code panics if the function returns a nextToken


hakman (approved):
Thanks for your contribution @ltellesfl!

PR #15787, opened 2023-08-16
Add a new field for using a custom registry for Cilium

### LINKED ISSUE #15786: Allow users to override the default Cilium Registry in the networking addon
/kind feature

This is a general request for the ability to change the registry for Cilium images in the networking addon. Currently, it is hard coded to `quay.io` in the addon template, but there are a variety of use cases where it would be nice to override the registry to something different:
* `quay.io` is not allowed on internal intranets for whatever reason
* `quay.io` is down, and another registry needs to be used instead
* `quay.io` is accessed through a dedicated pull through registry cache

This is already supported in other networking addons such as [Calico](https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.25.yaml.template#L4510), so the design is really to just duplicate the method to Cilium's networking add on: Add a `Registry` field to the `CiliumNetworkingSpec` and then use that within the template (defaulting to `quay.io`): `{{ or .Registry "quay.io" }}`

jandersen-plaid:
Fixes #15786.

There is a bit more context in the issue, but this adds a `Registry` field to the `CiliumNetworkingSpec` so that users can override the image registry used to pull Cilium images.

This copies the method used for similar Networking add ons when specifying a `Registry` argument in their NetworkingSpec; defaulting to `quay.io` when the `Registry` field is empty.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -5117,6 +5117,10 @@ spec:
                       reconfigureKubelet:
                         description: ReconfigureKubelet is unused.
                         type: boolean
+                      registry:
+                        description: Registry overrides the default Cilium container
+                          registry (quay.io)
+                        type: string
                       removeCbrBridge:
                         description: RemoveCbrBridge is unused.
                         type: boolean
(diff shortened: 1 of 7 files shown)


hakman:
@jandersen-plaid Please rebase to pass the tests.

hakman (approved):
Thanks for adding this @jandersen-plaid! 😄

hakman:
Sorry, I think you will need to rebase one more time it tests don't pass...

PR #15791, opened 2023-08-17
Create clusters with bigger default subnets

hakman:
When creating a cluster, the network CIRD is split into 8 subnets.
For AWS, the network CIDR is at max a /16, which results in /19 subnets (8192 IPs). This is quite small for AWS VPC CNI.

When doing scalability tests, even 500 nodes are too many for such small subnets, considering 16 IPs / node.
Even using multi-AZ, we still get a max of 48 IPs / node, again, not so great.

This PR proposes to split the network CIDR into bigger subnets, when the number of subnets allows it.

This only affects AWS and OpenStack:
https://github.com/kubernetes/kops/blob/73159cd6f50f72e004d972ce2ade21dcf40b7860/upup/pkg/fi/cloudup/defaults.go#L122-L128

/cc @johngmyers @justinsb @zetaab

### DIFF
--- a/pkg/apis/kops/validation/validation.go
+++ b/pkg/apis/kops/validation/validation.go
@@ -611,7 +611,7 @@ func validateSubnet(subnetSpec *kops.ClusterSubnetSpec, c *kops.ClusterSpec, fie
 				allErrs = append(allErrs, field.Forbidden(fieldPath.Child("cidr"), fmt.Sprintf("subnet %q cidr %q is not a subnet of the networkCIDR %q%s", subnetSpec.Name, subnetSpec.CIDR, c.Networking.NetworkCIDR, extraMsg)))
 			}
 		}
-		if subnet.Overlap(subnetCIDR, podCIDR) {
+		if subnet.Overlap(subnetCIDR, podCIDR) && c.Networking.AmazonVPC == nil {
 			allErrs = append(allErrs, field.Forbidden(fieldPath.Child("cidr"), fmt.Sprintf("subnet %q cidr %q must not overlap podCIDR %q", subnetSpec.Name, subnetSpec.CIDR, podCIDR)))
 		}
 		if subnet.Overlap(subnetCIDR, serviceClusterIPRange) {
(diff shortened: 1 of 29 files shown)


zetaab:
one thing that in my opinion we should keep in mind: there should be free space in VPC if for instance someone wants to migrate from availability zone a to availability zone b. It means that new subnets needs to be created before the old ones are removed. I have seen need for this in two different clusters during my kops usage.

In case of openstack I do not see huge issue with this, people can anyway change the behaviour with config if needed. I am more worried about default configuration for AWS.

hakman:
> one thing that in my opinion we should keep in mind: there should be free space in VPC if for instance someone wants to migrate from availability zone a to availability zone b. It means that new subnets needs to be created before the old ones are removed. I have seen need for this in two different clusters during my kops usage.

There is still room left with this approach. The first subnet CIRD is not allocated for cluster subnets, only for utility ones. It should be good enough for such migrations.

dims:
cc @jayanthvn @jdn5126

pkg/apis/kops/validation/validation.go
```diff
@@ -611,7 +611,7 @@ func validateSubnet(subnetSpec *kops.ClusterSubnetSpec, c *kops.ClusterSpec, fie
 				allErrs = append(allErrs, field.Forbidden(fieldPath.Child("cidr"), fmt.Sprintf("subnet %q cidr %q is not a subnet of the networkCIDR %q%s", subnetSpec.Name, subnetSpec.CIDR, c.Networking.NetworkCIDR, extraMsg)))
 			}
 		}
-		if subnet.Overlap(subnetCIDR, podCIDR) {
+		if subnet.Overlap(subnetCIDR, podCIDR) && c.Networking.AmazonVPC == nil {
```
johngmyers, line 614: I believe Cilium in ENI mode is in the same situation.

I wonder if the correct behavior for ENI-based CNIs is to leave `podCIDR` unset, like we do for IPv6. CCM is not assigning the pod IPs, the CNI is, deriving it from the ENI.

johngmyers (changes requested):
Consider doing the same thing when splitting the little subnets.

johngmyers (approved):

PR #15801, opened 2023-08-19
WIP: Find kops-controller using a SAN IP in the apiserver cert

justinsb:
We connect to kube-apiserver, issue a dummy GET request, but inspect
the TLS certificate to discovery the internal IP.

This does expose a small bit of information, but we are exposing it on
other clouds anyway.


justinsb:
cc @hakman .  I'm not sure if this is a good idea; we expose a bit of information publicly (that we're already exposing on other clouds), and then we are able to make a direct connection from the node to kops-controller.  Because the kops-controller connection doesn't go through the load balancer, this means we can lock down kops-controller to internal networks/subnets on clouds where that is possible.

justinsb:
Probably a bad idea I'm now thinking, so...

/close

PR #15805, opened 2023-08-20
aws: Create subnets for additional network CIDRs

hakman:
This will be needed for scale tests, to be able to use clusters with more than 500 nodes.

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -68,7 +68,6 @@ type CreateClusterOptions struct {
 	ContainerRuntime           string
 	OutDir                     string
 	DisableSubnetTags          bool
-	NetworkCIDR                string
 	DNSZone                    string
 	NodeSecurityGroups         []string
 	ControlPlaneSecurityGroups []string
@@ -300,7 +299,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.RegisterFlagCompletionFunc("subnets", completeSubnetID(options))
 	cmd.Flags().StringSliceVar(&options.UtilitySubnetIDs, "utility-subnets", options.UtilitySubnetIDs, "Shared utility subnets to use")
 	cmd.RegisterFlagCompletionFunc("utility-subnets", completeSubnetID(options))
-	cmd.Flags().StringVar(&options.NetworkCIDR, "network-cidr", options.NetworkCIDR, "Network CIDR to use")
+	cmd.Flags().StringSliceVar(&options.NetworkCIDRs, "network-cidr", options.NetworkCIDRs, "Network CIDR(s) to use")
 	cmd.RegisterFlagCompletionFunc("network-cidr", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	})
@@ -614,8 +613,12 @@ func RunCreateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Cr
 		cluster.Spec.ContainerRuntime = c.ContainerRuntime
 	}
 
-	if c.NetworkCIDR != "" {
-		cluster.Spec.Networking.NetworkCIDR = c.NetworkCIDR
+	for i, cidr := range c.NetworkCIDRs {
+		if i == 0 {
+			cluster.Spec.Networking.NetworkCIDR = cidr
+		} else {
+			cluster.Spec.Networking.AdditionalNetworkCIDRs = append(cluster.Spec.Networking.AdditionalNetworkCIDRs, cidr)
+		}
 	}
 
 	if c.DisableSubnetTags {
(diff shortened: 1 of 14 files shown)


cmd/kops/create_cluster.go
```diff
@@ -300,7 +299,7 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.RegisterFlagCompletionFunc("subnets", completeSubnetID(options))
 	cmd.Flags().StringSliceVar(&options.UtilitySubnetIDs, "utility-subnets", options.UtilitySubnetIDs, "Shared utility subnets to use")
 	cmd.RegisterFlagCompletionFunc("utility-subnets", completeSubnetID(options))
-	cmd.Flags().StringVar(&options.NetworkCIDR, "network-cidr", options.NetworkCIDR, "Network CIDR to use")
+	cmd.Flags().StringSliceVar(&options.NetworkCIDRs, "network-cidr", options.NetworkCIDRs, "Network CIDR(s) to use")
```
prateekgogia, line 302: Do we need to make this change here as well - https://github.com/kubernetes/kops/blob/master/tests/e2e/kubetest2-kops/deployer/up.go#L112?

prateekgogia, line 302: My suggestion would be to add a new flag called `additional-network-cidrs` to be backward compatible.

hakman: It is already backwards compatible. Nothing changes from user point of view.

hakman: I don't see any mention of `network-cidr` there.

prateekgogia: Sorry, might be missing something here but won't going from `StringVar` to `StringSliceVar` require users to make a change to pass a slice of strings now?

prateekgogia: IIUC, kubetest2 uses that interface which I linked earlier and we would need to provide this flag there as well.

hakman: you can pass a slice as multiple `--network-cidr` flags or comma separated.
For example, with this change, one can just do:
```
--network-cidr=10.0.0.0/16 --network-cidr=10.1.0.0/16 --network-cidr=10.2.0.0/16 --network-cidr=10.3.0.0/16 --network-cidr=10.4.0.0/16 --network-cidr=10.5.0.0/16
```

justinsb: I think we can pass them via the "passthrough" CreateArgs: https://github.com/kubernetes/kops/blob/9e0db3db74fbe5e435bbf7d1af805ab07f34bf6e/tests/e2e/kubetest2-kops/deployer/up.go#L126-L135

justinsb: Right, it's a nice feature of the flags library!

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -745,10 +747,32 @@ func setupZones(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.Stri
...
 		}
-		zoneToSubnetMap[zoneName] = subnet
+		zoneToSubnetsMap[zoneName] = append(zoneToSubnetsMap[zoneName], subnet)
 	}
 
-	return zoneToSubnetMap, nil
+	if len(opt.NetworkCIDRs) > 1 {
+		for i, cidr := range opt.NetworkCIDRs[1:] {
+			zoneName := opt.Zones[i%len(opt.Zones)]
+			subnetName := fmt.Sprintf("%s-%d", zoneName, i+1)
```
justinsb, line 756: I think this is pretty specific to AWS.  Also the way we round-robin around the zones is convenient, but might be surprising (along with the numbering).

Should we feature-flag this feature (and/or restrict to AWS only at first)?

tests/integration/create_cluster/complex/expected-v1alpha2.yaml
```diff
@@ -40,17 +44,29 @@ spec:
...
+    type: Public
+    zone: us-test-1a
+  - cidr: 10.2.0.0/16
+    name: us-test-1a-2
+    type: Public
+    zone: us-test-1a
+  - cidr: 10.3.0.0/16
+    name: us-test-1a-3
+    type: Public
+    zone: us-test-1a
```
johngmyers, line 69: I'm thinking this test should have a second zone. And we should have a private_complex test.

upup/pkg/fi/cloudup/new_cluster.go
johngmyers: One possibility would be to put the `-%d` suffix on the first subnet as well when `len(opt.NetworkCIDRs) > 1` but on AWS the primary CIDR of a VPC does have slightly different semantics.

johngmyers:
/lgtm

Holding off on approval to give opportunity to add those additional test cases. I would be okay with this going in as-is.

upup/pkg/fi/cloudup/new_cluster.go
hakman: I agree about the slightly different semantics. This is why I preferred to have different names for additional network CIDRs.

tests/integration/create_cluster/complex/expected-v1alpha2.yaml
hakman: Agreed, done.

justinsb:
Thanks @hakman 

We discussed in office hours, this is good to go I think, my concern was whether we should have a feature flag. @johngmyers agreed with you that we didn't, so...

/approve
/lgtm

PR #15816, opened 2023-08-24
add mirror

justinsb:
- Add cdn.dl.k8s.io as a mirror for dl.k8s.io
- Update expected output for additional mirrors


hakman (approved):
Should we cherry-pick to 1.27?

rifelpet:
This is worthy of a release note

justinsb:
Yes, I will add a release note, and then I think we should cherry-pick (unless anyone thinks otherwise).

The idea is that this should help with issues where dl.k8s.io is unreachable.  Normally that just redirects to cdn.dl.k8s.io, but we've had some reports of dl.k8s.io not always being reachable.

One thing I am thinking about though is whether we should download the shas from mirrors / whether we should be checking the sig files now that there are signature files (and whether there's any reason to check them on the primary mirrors).

PR #15824, opened 2023-08-26
Add support for --experimental-allocatable-ignore-eviction kubelet flag

### LINKED ISSUE #15821: Support --experimental-allocatable-ignore-eviction to allow usage of swap
/kind feature

Thanks to #15632, #15352, #14950 we can enable swap in kops kubernetes clusters. We enabled it and it works as expected. However, when we use more memory than RAM on the node pods get eviced by Kubernetes even though the node has a lot of memory available (due to be able to use swap).

This is caused by kubelet which calculcates `memory.available := node.status.capacity[memory] - node.stats.memory.workingSet` (see https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#eviction-signals). Here `node.status.capacity[memory]` is just RAM (without swap) and `node.stats.memory.workingSet` is the used memory by our containers (even if they use swap). The resulting value `memory.available` is used to determine if the node has memory pressure and if it should start to evict pods. Via kops we can set the hard eviction threshold for `memory.available` (which has to be a positive number). If we do not set the threshold it will fallback to a default of `100Mi`. As a result we cannot use more memory than RAM available even though we got swap available.

To use swap beyond the available memory we need to set `--experimental-allocatable-ignore-eviction` in the kubelet commandline which will disable fallback defaults for hard eviction thresholds. This would allow us to remove the threshold for `memory.available` and soly rely on the Linux OOM killer instead.

We would like to see a new option `ignoreEvictionDefaults` (or similar):

```
  kubelet:
    ignoreEvictionDefaults: "true".    # new option
    evictionHard: nodefs.available<10%,nodefs.inodesFree<5%,imagefs.available<10%,imagefs.inodesFree<5%.   # no setting for memory here
    failSwapOn: false
    featureGates:
      NodeSwap: "true"
```

When `ignoreEvictionDefaults` is set to `true` kops should add `--experimental-allocatable-ignore-eviction` to the commandline of kubelet on all nodes. This should allow usage of swap in a kops cluster.

hakman:
Closes #15821

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -3741,6 +3741,10 @@ spec:
                     description: Comma-delimited list of grace periods for each soft
                       eviction signal.  For example, 'memory.available=30s'.
                     type: string
+                  experimentalAllocatableIgnoreEviction:
+                    description: ExperimentalAllocatableIgnoreEviction enables ignoring
+                      Hard Eviction Thresholds while calculating Node Allocatable
+                    type: boolean
                   experimentalAllowedUnsafeSysctls:
                     description: ExperimentalAllowedUnsafeSysctls are passed to the
                       kubelet config to whitelist allowable sysctls Was promoted to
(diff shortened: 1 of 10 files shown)


justinsb:
Thanks @hakman 

/approve
/lgtm

PR #15828, opened 2023-08-26
Don't rely on kubectl being installed

justinsb:
- Avoid depending on kubectl from kops update-cluster
- Remove now-unused code to get config from kubectl


justinsb:
Non-urgent so assigning to 1.29 milestone

hakman:
> Non-urgent so assigning to 1.29 milestone
@justinsb Looks pretty good even for 1.28. In case of trouble, it can be easily reverted.

johngmyers (approved):

PR #15829, opened 2023-08-26
Refactor: Replace ForAPIServer with WellKnownServices

justinsb:
We instead return a list of the services we are supporting.

We can in future split out internal and external apiserver services.

### DIFF
--- a/pkg/model/bootstrapscript.go
+++ b/pkg/model/bootstrapscript.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/kops/pkg/apis/kops"
 	"k8s.io/kops/pkg/apis/nodeup"
 	"k8s.io/kops/pkg/model/resources"
+	"k8s.io/kops/pkg/wellknownservices"
 	"k8s.io/kops/upup/pkg/fi"
 	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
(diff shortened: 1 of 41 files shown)


pkg/model/bootstrapscript.go
```diff
@@ -40,7 +41,16 @@ import (
 )
 
 type NodeUpConfigBuilder interface {
-	BuildConfig(ig *kops.InstanceGroup, apiserverAdditionalIPs []string, keysets map[string]*fi.Keyset) (*nodeup.Config, *nodeup.BootConfig, error)
+	BuildConfig(ig *kops.InstanceGroup, wellKnownAddresses WellKnownAddresses, keysets map[string]*fi.Keyset) (*nodeup.Config, *nodeup.BootConfig, error)
+}
+
+// WellKnownAddresses holds known addresses for well-known services
+type WellKnownAddresses struct {
```
johngmyers, line 48: Perhaps something more like a `map[wellknownservices.WellKnownService][]string`?

justinsb: Good idea - converted to a type alias on the map.

justinsb:
I remembered why this one is important ... doing this will let us set up two apiservers, one for internal traffic and one for external traffic, which is needed for GCE firewalling, which is needed for GCE dns=none

PR #15835, opened 2023-08-29
scaleway: rolling-update feature

Mia-Cross:
This PR adds the possibility to edit the image and the size of the instances on a live cluster by performing a rolling-update.

This feature will also be useful for the Terraform support that's coming up next.

### DIFF
--- a/upup/pkg/fi/cloudup/scalewaytasks/instance.go
+++ b/upup/pkg/fi/cloudup/scalewaytasks/instance.go
@@ -18,11 +18,14 @@ package scalewaytasks
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
 	"github.com/scaleway/scaleway-sdk-go/api/lb/v1"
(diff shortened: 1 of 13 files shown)


upup/pkg/fi/cloudup/scalewaytasks/instance.go
```diff
@@ -359,3 +426,54 @@ func (_ *Instance) RenderTerraform(t *terraform.TerraformTarget, actual, expecte
...
+}
+
+func checkUserDataDifferences(c *fi.CloudupContext, cloud scaleway.ScwCloud, actualServer *instance.Server, expectedUserData *fi.Resource) (bool, error) {
+	actualUserData, err := cloud.InstanceService().GetServerUserData(&instance.GetServerUserDataRequest{
+		Zone:     actualServer.Zone,
+		ServerID: actualServer.ID,
+		Key:      "cloud-init",
+	}, scw.WithContext(c.Context()))
+	if err != nil {
+		return false, fmt.Errorf("error getting actual user-data: %w", err)
```
johngmyers, line 453: Error messages shouldn't start with the word "error", as they're printed with an "error: " prefix.

johngmyers (changes requested):

hakman (approved):
Looking forward to Terraform output. 👍

PR #15839, opened 2023-08-30
Spotinst: Feature/add resource tag specification double ami2 try2

yehielnetapp:
Feature add
1)It will be possible to set 2 AMI per Virtual Node Group, one for each architecture (x86, arm)
2)add tag to track storage costs on their end on the VNG level.

### DIFF
--- a/docs/getting_started/spot-ocean.md
+++ b/docs/getting_started/spot-ocean.md
@@ -197,15 +197,17 @@ metadata:
   name: "example"
   labels:
     spotinst.io/strategy-cluster-spread-nodes-by: "count"
-    spotinst.io/strategy-cluster-orientation-availability-vs-cost: "balanced"    
+    spotinst.io/strategy-cluster-orientation-availability-vs-cost: "balanced"
+    spotinst.io/resource-tag-specification-volumes: "true"
   ...
 ```
 
 
-| Label | Description                                                                        | Default |
-|---|---|---|
-| `spotinst.io/strategy-cluster-spread-nodes-by` | Specify how Ocean will spread the nodes across markets by this value [vcpu,count]. | `count` |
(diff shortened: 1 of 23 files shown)


hakman:
> /cc @hakman is there any thing to do here ?

Generally, we were not planning to support mixed X86 & ARM instance groups.
I don't think that's a problem if only for Spotinst, but will have to check with the other maintainers.
/cc @justinsb 
/assign @justinsb

justinsb:
Hi @yehielnetapp - sorry about the delay here.  `spotinst.io/resource-tag-specification-volumes` looks good, but we have discussed `spotinst.io/other-architecture-image` a few times now, and we haven't really been able to reach a great conclusion.  If you wanted to split out  `spotinst.io/resource-tag-specification-volumes` into a separate PR that one is non-controversial.

The biggest challenge with `other-architecture-image` is that it assumes there are only two architectures.  Maybe if the architecture was specified it would be ok, something like `spotinst.io/image-for-arm64` (or whatever name makes sense).

In general, for kOps, we would normally expect that people that want to mix architectures would use different instanceGroups, but technically I guess this isn't required.  We did discuss whether we should do something generic to allow instanceGroups to have multiple images, but we didn't know of other people with the use case.

We looked at how karpenter does it: https://karpenter.sh/docs/concepts/node-templates/#specamiselector  That solves the problem, but we also wanted the image spec to be "locked" (i.e. if a new image is published, we would expect that the instanceGroup object should change, at least with how we manage instanceGroups today).

Looking at the code here though, it does look like spotinst is smart enough to automatically figure out which image is which architecture, and use the right one.  So maybe the answer could be to rename the annotation to `other-architecture-images`, allow a comma-separated list (at least in theory, even if today there will only be one).  I think that overcomes the concerns we had (though @hakman or @johngmyers might recall some additional challenges?) - I would be OK to merge this with that tweak.

yehielnetapp:
Hi  @justinsb i have resolved your  comments to new PR : 
https://github.com/kubernetes/kops/pull/16099

Issue #15751, opened 2023-08-07
Error running AutoscalingLifecycleHook for NTH task


hakman:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

CI

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

v1.27.4

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

```
kops create cluster --name e2e-4342699135-9b4dd.test-cncf-aws.k8s.io --cloud aws --kubernetes-version v1.27.4 --ssh-public-key /tmp/kops/e2e-4342699135-9b4dd.test-cncf-aws.k8s.io/id_ed25519.pub --set cluster.spec.nodePortAccess=0.0.0.0/0 --networking amazonvpc --set=cluster.spec.cloudProvider.aws.loadBalancerController.enabled=true --set=cluster.spec.certManager.enabled=true --master-size=t4g.medium --node-size=t4g.medium --image=ssm:/aws/service/canonical/ubuntu/server/20.04/stable/current/arm64/hvm/ebs-gp2/ami-id --zones=eu-west-1a,eu-west-1b,eu-west-1c --discovery-store=s3://k8s-kops-prow/e2e-4342699135-9b4dd.test-cncf-aws.k8s.io/discovery --admin-access 34.123.236.93/32 --master-count 1 --master-volume-size 48 --node-count 4 --node-volume-size 48
```

**5. What happened after the commands executed?**

```
W0807 05:40:43.008620    6028 executor.go:139] error running task "AutoscalingLifecycleHook/nodes-eu-west-1a-NTHLifecycleHook" (9m58s remaining to succeed): ValidationError: AutoScalingGroup name not found - No such Auto Scaling group: nodes-eu-west-1a.e2e-4342699135-9b4dd.test-cncf-aws.k8s.io
	status code: 400, request id: 538d0c0d-de15-41c0-a1d9-eec0c53d847d
W0807 05:40:43.008698    6028 executor.go:139] error running task "AutoscalingLifecycleHook/control-plane-eu-west-1a-NTHLifecycleHook" (9m58s remaining to succeed): ValidationError: AutoScalingGroup name not found - No such Auto Scaling group: control-plane-eu-west-1a.masters.e2e-4342699135-9b4dd.test-cncf-aws.k8s.io
[...]

PR #15846, opened 2023-08-31
Dependency analysis: include the direct task

justinsb:
We weren't previously adding the direct dependency (though we were
adding indirect dependencies) when a task implemented
fi.HasDependencies.

Issue #15751


hakman:
@johngmyers Should we cherry-pick this too?

PR #15847, opened 2023-08-31
Improve status printing when tasks are in progress

justinsb:
Previously we were implying that the tasks were somehow failing,but
this isn't quite right when they emit the try-again-later error.

Also pretty-print the pluralization while we're here.

### DIFF
--- a/upup/pkg/fi/executor.go
+++ b/upup/pkg/fi/executor.go
@@ -164,14 +164,10 @@ func (e *executor[T]) RunTasks(taskMap map[string]Task[T]) error {
 				}
 			}
 			formatTaskCount := func(n int) string {
-				if n == 1 {
-					return "1 task"
-				} else {
-					return fmt.Sprintf("%d tasks", n)
-				}
+				return fmt.Sprintf("%d task(s)", n)
 			}
 			if tryAgainLaterCount == n {
-				klog.Infof("continuing to run %s", formatTaskCount(tryAgainLaterCount))
+				klog.Infof("Continuing to run %s", formatTaskCount(tryAgainLaterCount))
 			} else {
 				klog.Infof("No progress made, sleeping before retrying %s", formatTaskCount(n))
 			}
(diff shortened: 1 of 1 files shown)


upup/pkg/fi/executor.go
```diff
@@ -148,11 +149,32 @@ func (e *executor[T]) RunTasks(taskMap map[string]Task[T]) error {
...
+			}
+			formatTaskCount := func(n int) string {
+				if n == 1 {
+					return "1 task"
+				} else {
+					return fmt.Sprintf("%d tasks", n)
+				}
+			}
+			if tryAgainLaterCount == n {
+				klog.Infof("%s are continuing to execute", formatTaskCount(tryAgainLaterCount))
```
justinsb, line 174: Ugh ... presumably this should be "1 task is continuing to execute" :-(

justinsb: Changed to "continuing to run 1 task"

upup/pkg/fi/executor.go
```diff
@@ -148,11 +149,32 @@ func (e *executor[T]) RunTasks(taskMap map[string]Task[T]) error {
...
+				if !errors.Is(err, &tryAgainLaterError) {
+					tryAgainLaterCount++
+				}
+			}
+			formatTaskCount := func(n int) string {
+				if n == 1 {
+					return "1 task"
+				} else {
+					return fmt.Sprintf("%d tasks", n)
+				}
```
hakman, line 171: Would this work?
```suggestion
				return fmt.Sprintf("%d task(s)", n)
```

hakman:
@justinsb I guess something broke when pushing dns-conroller.tgz to bucket. Can you force push to try and fix it?

upup/pkg/fi/executor.go
```diff
@@ -148,11 +149,32 @@ func (e *executor[T]) RunTasks(taskMap map[string]Task[T]) error {
...
+			}
+			if tryAgainLaterCount == n {
+				klog.Infof("continuing to run %s", formatTaskCount(tryAgainLaterCount))
```
hakman, line 174:
```suggestion
				klog.Infof("Continuing to run %s", formatTaskCount(tryAgainLaterCount))
```

hakman (approved):

PR #15848, opened 2023-08-31
Fix warmpool to expose dependencies for dependency analysis

justinsb:
We should populate the AutoscalingGroup field, so that it can be used
by dependency analysis.


justinsb:
Not particularly urgent, as we seem to be getting away with this one, and worst case it should just be an extra retry.

johngmyers (approved):

PR #15854, opened 2023-09-02
aws: Add option to provide multiple instance types on cluster creation

hakman:
Alternative to #15849, for use in scalability tests.

/cc @justinsb @johngmyers @dims

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -162,8 +162,8 @@ type NewClusterOptions struct {
 	NodeImage         string
 	ControlPlaneImage string
 	BastionImage      string
-	ControlPlaneSize  string
-	NodeSize          string
+	ControlPlaneSizes []string
+	NodeSizes         []string
 }
 
 func (o *NewClusterOptions) InitDefaults() {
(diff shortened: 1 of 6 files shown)


upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1055,7 +1066,21 @@ func setupNodes(opt *NewClusterOptions, cluster *api.Cluster, zoneToSubnetsMap m
 			g.Spec.NodeLabels["cloud.google.com/metadata-proxy-ready"] = "true"
 		}
 
-		g.Spec.MachineType = opt.NodeSize
+		for i, size := range opt.NodeSizes {
+			if i == 0 {
```
dims, line 1070: same if condition twice?

hakman: Thanks for catching. Was in a hurry earlier.

justinsb:
As this is for scalability tests, I'm guessing it's not something we would need in 1.27 or 1.28

/milestone v1.29

johngmyers:
@hakman it looks like this broke the kops-aws-cni-cilium-eni periodic. We need to de-duplicate the instance types.

hakman:
@johngmyers https://github.com/kubernetes/test-infra/pull/30664

Issue #15852, opened 2023-09-01
Can not create cluster (following the getting_started)


wo9999999999:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
1.27.0

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
Client Version: v1.28.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3

**3. What cloud provider are you using?**
aws

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
kops create cluster

**5. What happened after the commands executed?**
cluster validation keep fail 
```INSTANCE GROUPS
NAME				ROLE		MACHINETYPE	MIN	MAX	SUBNETS
control-plane-us-east-1a	ControlPlane	t3.medium	1	1	us-east-1a
nodes-us-east-1a		Node		t3.medium	1	1	us-east-1a

NODE STATUS
NAME			ROLE		READY
i-0be13550669d0b732	control-plane	True
i-0c78304642cc52ef8	node		True

VALIDATION ERRORS
KIND	NAME						MESSAGE
Pod	kube-system/ebs-csi-controller-75fc64d98f-4dbzk	system-cluster-critical pod "ebs-csi-controller-75fc64d98f-4dbzk" is pending

Validation Failed
W0901 16:32:30.328628   45600 validate_cluster.go:232] (will retry): cluster not yet healthy

```

**6. What did you expect to happen?**
start cluster successfully

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2023-09-01T08:22:42Z"
  name: woleung.k8s.local
spec:
  api:
    loadBalancer:
      class: Network
      type: Public
  authorization:
    rbac: {}
  channel: stable
  cloudProvider: aws
  configBase: s3://testwokops5-example-com-state-store/woleung.k8s.local
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - encryptedVolume: true
      instanceGroup: control-plane-us-east-1a
      name: a
    manager:
      backupRetentionDays: 90
    memoryRequest: 100Mi
    name: main
  - cpuRequest: 100m
    etcdMembers:
[...]

PR #15868, opened 2023-09-04
Only run one replica of controller pods on non-HA clusters

justinsb:
Even when running on workers (using IRSA), if we try to run multiple
controllers we may have problems with node-spreading, and we don't
necessarily gain any availability, as we need an apiserver lease.

Issue #15852


justinsb:
Assuming we agree with this approach, there is a risk that we break e.g. a PodDisruptionBudget during a rolling update.  I propose we merge to 1.28 first, then cherry pick to 1.27 (but not necessarily 1.27.1, as I don't think this is a regression, it only occurs with IRSA + a cluster with a single worker node)

johngmyers (approved):

PR #15872, opened 2023-09-04
Remove dead code for in-tree CCM

johngmyers:


pkg/apis/kops/componentconfig.go
```diff
@@ -883,9 +883,9 @@ type CloudConfiguration struct {
 
 // EBSCSIDriverSpec is the config for the AWS EBS CSI driver
 type EBSCSIDriverSpec struct {
-	// Enabled enables the AWS EBS CSI driver
-	// Default: false
-	Enabled *bool `json:"enabled,omitempty"`
+	// Enabled enables the AWS EBS CSI driver. Can only be set to true.
+	// Default: true
+	Enabled *bool `json:"-"`
```
rifelpet, line 888: Why is running with EBS CSI Driver disabled an invalid configuration?

I imagine operators that dont want to allow stateful workloads would disable it to avoid the extra controller and daemonset pods as well as IAM role.

johngmyers: Running external CCM without EBS CSI Driver has been prohibited since at least #11795

johngmyers: This requirement goes back to #10467

johngmyers: If we want to start supporting running without a `PersistentVolume` driver, that could be done in a follow-up PR.

rifelpet: Agreed

hakman (approved):

PR #15884, opened 2023-09-07
ENI Cleanup Improvements

rifelpet:
Attempting to fix the scalability tests: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kops-aws-scale-amazonvpc-using-cl2/1699576183240265728

* Paginate DescribeNetworkInterfaces which should help with large VPCs containing many ENIs. A similar change was made to the AWS VPC CNI: https://github.com/aws/amazon-vpc-cni-k8s/pull/1333
* Filter for unattached ENIs in the API rather than after the fact, reducing the number of response pages.

### DIFF
--- a/pkg/resources/aws/eni.go
+++ b/pkg/resources/aws/eni.go
@@ -66,18 +66,21 @@ func DumpENI(op *resources.DumpOperation, r *resources.Resource) error {
 func DescribeENIs(cloud fi.Cloud, clusterName string) (map[string]*ec2.NetworkInterface, error) {
 	c := cloud.(awsup.AWSCloud)
 
+	statusFilter := &ec2.Filter{
+		Name: aws.String("status"),
+		Values: []*string{
+			aws.String(ec2.NetworkInterfaceStatusDetaching),
+			aws.String(ec2.NetworkInterfaceStatusAvailable),
+		},
+	}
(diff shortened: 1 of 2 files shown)


hakman:
/cc @dims
@hakuna-matatah

dims:
/approve
/lgtm

thanks @rifelpet

pkg/resources/aws/eni.go
```diff
@@ -66,24 +66,28 @@ func DumpENI(op *resources.DumpOperation, r *resources.Resource) error {
 func DescribeENIs(cloud fi.Cloud, clusterName string) (map[string]*ec2.NetworkInterface, error) {
 	c := cloud.(awsup.AWSCloud)
 
+	statusFilter := &ec2.Filter{
+		Name: aws.String("status"),
+		Values: []*string{
+			aws.String(ec2.NetworkInterfaceStatusDetaching),
```
hakman, line 72: Do we care about the detaching interfaces in our case?
I think the List function is called again and again, so those detaching ENIs will become available at some point.

johngmyers: No, ENIs that aren't returned won't get a resource tracker assigned. I suppose the assumption is that ENIs that are attached will be implicitly deleted once the thing they're attached to is deleted.

PR #15892, opened 2023-09-08
scaleway: terraform support

Mia-Cross:
This PR adds the support for the generation of Terraform files when creating and updating clusters.


upup/pkg/fi/cloudup/scalewaytasks/loadbalancer.go
```diff
@@ -220,10 +219,11 @@ func (_ *LoadBalancer) RenderTerraform(t *terraform.TerraformTarget, actual, exp
 	}
 
 	tfLB := terraformLoadBalancer{
-		Type: "LB-S",
-		Name: expected.Name,
-		Tags: expected.Tags,
-		IPID: terraformWriter.LiteralProperty("scaleway_lb_ip", tfName, "id"),
+		Type:        "LB-S",
```
remyleone, line 222: Could it be a default const that could be changed easily in case of a rename or change of product?

tests/integration/update_cluster/minimal_scaleway/kubernetes.tf
```diff
@@ -169,31 +175,40 @@ resource "scaleway_iam_ssh_key" "kubernetes-scw-minimal-k8s-local-be_9e_c3_eb_cb
...
-  tags  = ["noprefix=kops.k8s.io/cluster=scw-minimal.k8s.local", "noprefix=kops.k8s.io/instance-group=nodes-fr-par-1"]
-  type  = "DEV1-M"
+resource "scaleway_instance_server" "nodes-fr-par-1-0" {
+  enable_dynamic_ip      = true
+  image                  = "ubuntu_focal"
+  ip_id                  = scaleway_instance_ip.nodes-fr-par-1-0.id
+  name                   = "nodes-fr-par-1-0"
+  replace_on_type_change = false
+  tags                   = ["noprefix=kops.k8s.io/cluster=scw-minimal.k8s.local", "noprefix=kops.k8s.io/instance-group=nodes-fr-par-1"]
+  type                   = "DEV1-M"
```
remyleone, line 209: Could this type be put in a variable to allow quick changes?

tests/integration/update_cluster/minimal_scaleway/kubernetes.tf
```diff
@@ -169,31 +175,40 @@ resource "scaleway_iam_ssh_key" "kubernetes-scw-minimal-k8s-local-be_9e_c3_eb_cb
...
 
-resource "scaleway_instance_server" "nodes-fr-par-1" {
-  image = "ubuntu_focal"
-  ip_id = scaleway_instance_ip.nodes-fr-par-1.id
-  name  = "nodes-fr-par-1"
-  tags  = ["noprefix=kops.k8s.io/cluster=scw-minimal.k8s.local", "noprefix=kops.k8s.io/instance-group=nodes-fr-par-1"]
-  type  = "DEV1-M"
+resource "scaleway_instance_server" "nodes-fr-par-1-0" {
+  enable_dynamic_ip      = true
+  image                  = "ubuntu_focal"
```
remyleone, line 204: Could this be put in a variable to be quickly change for the next LTS when it's out?

tests/integration/update_cluster/minimal_scaleway/kubernetes.tf
```diff
@@ -212,24 +227,29 @@ resource "scaleway_instance_volume" "etcd-1-etcd-main-scw-minimal-k8s-local" {
...
 resource "scaleway_lb" "api-scw-minimal-k8s-local" {
-  ip_id = scaleway_lb_ip.api-scw-minimal-k8s-local.id
-  name  = "api.scw-minimal.k8s.local"
-  tags  = ["noprefix=kops.k8s.io/cluster=scw-minimal.k8s.local", "noprefix=kops.k8s.io/role=ControlPlane"]
-  type  = "LB-S"
+  description = "Load-balancer for kops cluster scw-minimal.k8s.local"
+  ip_id       = scaleway_lb_ip.api-scw-minimal-k8s-local.id
+  name        = "api.scw-minimal.k8s.local"
+  tags        = ["noprefix=kops.k8s.io/cluster=scw-minimal.k8s.local", "noprefix=kops.k8s.io/role=ControlPlane"]
+  type        = "LB-S"
```
remyleone, line 234: Could this be put in a variable to be changed quickly in case the default type of LB would change name?

pkg/model/scalewaymodel/instances.go
```diff
@@ -75,6 +78,15 @@ func (b *InstanceModelBuilder) Build(c *fi.CloudupModelBuilderContext) error {
 			instance.Role = fi.PtrTo(scaleway.TagRoleWorker)
 		}
 
+		// If the instance's commercial type is one that has no local storage, we have to specify for the
+		// block storage volume a big enough size (default size is 10GB)
+		for _, commercialType := range commercialTypesWithBlockStorageOnly {
+			if strings.HasPrefix(ig.Spec.MachineType, commercialType) {
+				instance.VolumeSize = fi.PtrTo(50)
```
remyleone, line 85: Could you put the default size in a const? Also are you sure that this is the correct way to specify a size ? I thought it was: 		volumeSize    = 20 * scw.GB for instance

docs/getting_started/scaleway.md
```diff
@@ -111,3 +107,69 @@ kops update cluster mycluster.k8s.local --yes
...
+
+# Terraform support
+
+kOps offers the possibility to generate a Terraform configuration corresponding to the cluster that would have been created directly otherwise.
+
+You can find more information on the dedicated page on [kOps Terraform support](../terraform.md) or [Scaleway's Terraform provider's documentation](https://github.com/scaleway/terraform-provider-scaleway).
+
+## For clusters without load-balancers
+
+This concerns clusters using Scaleway DNS. For this type of clusters, things are pretty simple.
```
remyleone, line 119: Please rephrase:
https://justsimply.dev/

docs/getting_started/scaleway.md
```diff
@@ -111,3 +107,69 @@ kops update cluster mycluster.k8s.local --yes
...
+
+```bash
+kops create cluster --cloud=scaleway --name=mycluster.mydomain.com --zones=fr-par-1 --target=terraform --out=$OUTPUT_DIR
+cd $OUTPUT_DIR
+terraform init
+terraform apply
+```
+kOps will generate a `kubernetes.tf` file in the output directory of your choice, you just have to initialize Terraform and apply the configuration.
```
remyleone, line 127: Can this name be changed using an argument?

dnsprovider/pkg/dnsprovider/providers/scaleway/dns.go
```diff
@@ -353,75 +353,35 @@ func (r *resourceRecordChangeset) Apply(ctx context.Context) error {
...
-			klog.V(8).Infof("adding new DNS record %q to zone %q", recordName, r.zone.name)
-			updateRecordsRequest = append(updateRecordsRequest, &domain.RecordChange{
-				Add: &domain.RecordChangeAdd{
-					Records: recordsToAdd,
-				},
-			})
-		}
-	}
-
+	// Scaleway's Domain API doesn't allow edits to the same record if one request, so we have to check for duplicates
```
remyleone, line 356:
```suggestion
	// Scaleway's Domain API doesn't allow edits to the same record in one request, so we have to check for duplicates
```
I'm not sure I'm understanding this sentence correctly.

tests/integration/update_cluster/minimal_scaleway/kubernetes.tf
Mia-Cross: This file is generated with the default values. The value for the instance type is defined [here](https://github.com/kubernetes/kops/blob/9816fd941624f73e5b2a819bce58f443d703042e/upup/pkg/fi/cloudup/populate_instancegroup_spec.go#L43)

tests/integration/update_cluster/minimal_scaleway/kubernetes.tf
Mia-Cross: Same here, the default value for the image is defined [there](https://github.com/kubernetes/kops/blob/9816fd941624f73e5b2a819bce58f443d703042e/upup/pkg/fi/cloudup/populate_instancegroup_spec.go#L60) and updated when possible.

docs/getting_started/scaleway.md
Mia-Cross: I don't think so, the name of the file is defined [here](https://github.com/kubernetes/kops/blob/9816fd941624f73e5b2a819bce58f443d703042e/upup/pkg/fi/cloudup/terraform/target_hcl2.go#L57)

pkg/model/scalewaymodel/instances.go
Mia-Cross: I put the value in a constant, but for the second remark, in the model part I have to store the value as an int and not a Scaleway-specific type. That's why I only put the size in GB so that it doesn't overflow. 
When I actually make the request to create the instance ([here](https://github.com/kubernetes/kops/blob/9816fd941624f73e5b2a819bce58f443d703042e/upup/pkg/fi/cloudup/scalewaytasks/instance.go#L244)) , I retrieve the value and multiply it by size.GB.
NB: I noticed that `50` was also hardcoded in that file, I'll use the constant instead.

tests/integration/update_cluster/minimal_scaleway/kubernetes.tf
Mia-Cross: Same here

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -762,6 +770,9 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
 
 		// Avoid making changes on a dry-run
 		shouldPrecreateDNS = false
+		if cluster.Spec.GetCloudProvider() == kops.CloudProviderScaleway {
+			shouldPrecreateDNS = true
+		}
```
hakman, line 775: During dry-run should not pre-create DNS records.

upup/pkg/fi/cloudup/apply_cluster.go
```diff
@@ -738,6 +738,14 @@ func (c *ApplyClusterCmd) Run(ctx context.Context) error {
 			return err
 		}
 
+		if cluster.Spec.GetCloudProvider() == kops.CloudProviderScaleway {
+			scwCloud := cloud.(scaleway.ScwCloud)
+			err := tf.AddOutputVariable("zone", terraformWriter.LiteralFromStringValue(scwCloud.Zone()))
+			if err != nil {
+				return err
+			}
+		}
```
hakman, line 747: See how `project` is set for GCE and use the same pattern.
```suggestion
		if scwZone != "" {
			if err := tf.AddOutputVariable("zone", terraformWriter.LiteralFromStringValue(scwZone)); err != nil {
				return err
			}
		}
```

hakman (changes requested):
@Mia-Cross mostly LGTM, with a few small nits. Could you take a look?

upup/pkg/fi/cloudup/apply_cluster.go
Mia-Cross: Sorry I meant to write this line in the Terraform part and not the DryRun part.
This may be temporary as I am considering creating a model and a task for the DNS record resource. But I have not had any conflict if I pre-create DNS records even for Terraform clusters.

upup/pkg/fi/cloudup/apply_cluster.go
Mia-Cross: Done :+1:

upup/pkg/fi/cloudup/apply_cluster.go
hakman: Could you remove it from here? I don't want to forget about it and have different behaviour from other cloud providers.

Mia-Cross: The model and task for DNS record are ready. I removed the exception for the pre-creation of DNS records with Terraform from `apply_cluster.go`.
Instead the records get pre-created in the kubernetes.tf file with the placeholder IP and with `ignore_changes` on the `data` field because its value will be changed outside of Terraform (similar to the way the etcd volumes are handled)

hakman (approved):
Thanks @Mia-Cross, LGTM!

PR #15897, opened 2023-09-10
aws: Use dns=none for newly created clusters

hakman:
/cc @justinsb @johngmyers @zetaab 
/assign @justinsb  
/hold for feedback

### DIFF
--- a/upup/pkg/fi/cloudup/new_cluster.go
+++ b/upup/pkg/fi/cloudup/new_cluster.go
@@ -145,6 +145,8 @@ type NewClusterOptions struct {
 	Topology string
 	// DNSType is the DNS type to use; "public" or "private". Defaults to "public".
 	DNSType string
+	// DNSZone is the DNS zone to use.
+	DNSZone string
 
 	// APILoadBalancerClass determines whether to use classic or network load balancers for the API
 	APILoadBalancerClass string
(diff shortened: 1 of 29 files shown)


johngmyers:
I don't think this is a good change. In AWS and other cloud providers with a public-DNS service, if the cluster domain is not under `cluster.local` then it should default to Public DNS. If the cluster domain is under `cluster.local` then it should default to dns=none.

zetaab:
@johngmyers just fix your comment, you perhaps meant `k8s.local` which is the suffix in old "gossip" way.

upup/pkg/fi/cloudup/new_cluster.go
```diff
@@ -1369,15 +1371,20 @@ func setupTopology(opt *NewClusterOptions, cluster *api.Cluster, allZones sets.S
...
-			if cluster.UsesLegacyGossip() {
-				cluster.Spec.Networking.Topology.DNS = api.DNSTypePrivate
-			} else {
-				cluster.Spec.Networking.Topology.DNS = api.DNSTypePublic
+		if opt.DNSZone != "" {
+			// Use dns=public if zone is specified
+			cluster.Spec.Networking.Topology.DNS = api.DNSTypePublic
+		} else {
+			switch cluster.Spec.GetCloudProvider() {
+			case api.CloudProviderAWS, api.CloudProviderHetzner, api.CloudProviderDO, api.CloudProviderAzure:
```
zetaab, line 1379: openstack can be added to this list

zetaab:
What is the real reason to hardcode clusters to real dns names(dns=public)? The certificates are anyways self-signed. What will happen when the real dns name domain will expire and some hacker will buy that? Right. (yeah I have heard at least one case)

What are the benefits of using real dns instead of dns=none by default? Because as I see it: the default configuration should be easy to get started with. Sometimes its not that easy to get dns (sub)zones delegated to AWS R53.

johngmyers:
Thinking about this a bit, I believe the main benefit of dns=public is that you don't have to update all clients' kubeconfigs in case the front-end LB gets reprovisioned.

We use dns=public under a split-visiblity zone, publishing to a zone that is only delegated to internally.

The certificates are not self-signed, they are out of the cluster's Kubernetes general CA. Reasonable clients would only talk to the cluster if they verify the server's cert against a CA in the kubeconfig.

hakman:
/close

Issue #15853, opened 2023-09-01
Add NodeEncryption (beta) to Cilium 


zadjadr:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

I want to add the ability to [encrypt node-to-node traffic](https://github.com/cilium/cilium/blob/90a9402d2342a12774b8c3eebd67de5bad572472/Documentation/security/network/encryption-wireguard.rst#node-to-node-encryption-beta)

In 1.13 and below, the node encryption only works with IPsec and from 1.14 and above it only works with wireguard.

**2. Feel free to provide a design supporting your feature request.**

I'll try to provide a PR as soon as possible :)

PR #15901, opened 2023-09-11
Cilium: Implement node encryption

zadjadr:
Fixes: https://github.com/kubernetes/kops/issues/15853

In 1.13 and below, the node encryption only works with IPsec and from 1.14 and above it only works with wireguard.

There is no extra validation added, because the setting will be ignored if you have `wiredguard` + `nodeEncryption` + cilum 1.13 or `ipsec` + `nodeEncryption` + cilium 1.14.


zadjadr:
@hakman can this be merged?

hakman (approved):

PR #15902, opened 2023-09-11
Handle 404 properly from Service Accounts API

upodroid:
- Capitalized a few resources to match their names
- If you cancel kubetest2 before the tests start running, you end up in a weird state where it tries to delete service accounts after they had been deleted. This change fixes that

### DIFF
--- a/pkg/resources/gce/gce.go
+++ b/pkg/resources/gce/gce.go
@@ -377,10 +377,10 @@ func deleteGCEDisk(cloud fi.Cloud, r *resources.Resource) error {
 	op, err := c.Compute().Disks().Delete(u.Project, u.Zone, u.Name)
 	if err != nil {
 		if gce.IsNotFound(err) {
-			klog.Infof("disk not found, assuming deleted: %q", t.SelfLink)
+			klog.Infof("Disk not found, assuming deleted: %q", t.SelfLink)
 			return nil
 		}
-		return fmt.Errorf("error deleting disk %s: %v", t.SelfLink, err)
+		return fmt.Errorf("error deleting Disk %s: %v", t.SelfLink, err)
 	}
 
 	return c.WaitForOp(op)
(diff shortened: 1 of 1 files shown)


pkg/resources/gce/gce.go
```diff
@@ -1044,7 +1044,14 @@ func deleteServiceAccount(cloud fi.Cloud, r *resources.Resource) error {
...
 	_, err := c.IAM().ServiceAccounts().Delete(o.Name)
-	return err
+
+	if err != nil {
+		if gce.IsNotFound(err) {
+			klog.Infof("ServiceAccount not found, assuming deleted: %q", o.Name)
+			return nil
+		}
+	}
+	return fmt.Errorf("error deleting ServiceAccount %s: %v", o.Name, err)
```
hakman, line 1054:
```suggestion
		return fmt.Errorf("error deleting ServiceAccount %s: %v", o.Name, err)
	}
	return nil
```

hakman (changes requested):

pkg/resources/gce/gce.go
upodroid: Fixed

pkg/resources/gce/gce.go
```diff
@@ -377,10 +377,10 @@ func deleteGCEDisk(cloud fi.Cloud, r *resources.Resource) error {
 	op, err := c.Compute().Disks().Delete(u.Project, u.Zone, u.Name)
 	if err != nil {
 		if gce.IsNotFound(err) {
-			klog.Infof("disk not found, assuming deleted: %q", t.SelfLink)
+			klog.Infof("Disk not found, assuming deleted: %q", t.SelfLink)
 			return nil
 		}
-		return fmt.Errorf("error deleting disk %s: %v", t.SelfLink, err)
+		return fmt.Errorf("error deleting Disk %s: %v", t.SelfLink, err)
```
johngmyers, line 383: Error messages should not start with the word "error"

```suggestion
		return fmt.Errorf("deleting Disk %s: %v", t.SelfLink, err)
```

hakman (approved):

pkg/resources/gce/gce.go
hakman: sorry... missed this

upodroid: I'll do it in a followup PR

PR #15903, opened 2023-09-12
aws: Filter ENI list by VPC

hakman:
This should add filtering by VPC for ENIs, which seems needed for scale tests, but also should be a nice addition.
https://kubernetes.slack.com/archives/CCK68P2Q2/p1694457549877619?thread_ts=1694437331.729519&cid=CCK68P2Q2

/cc @rifelpet @justinsb @dims


pkg/resources/aws/eni.go
```diff
@@ -79,6 +79,9 @@ func DescribeENIs(cloud fi.Cloud, clusterName string) (map[string]*ec2.NetworkIn
 		request := &ec2.DescribeNetworkInterfacesInput{
 			Filters: append(filters, statusFilter),
 		}
+		if vpcID != "" {
```
hakuna-matatah, line 82: nice.

hakman: actually, here we may as well return. there's no ENI without VPC and it's probably valid for most other things.

pkg/resources/aws/routetable.go
```diff
@@ -51,7 +51,7 @@ func DescribeRouteTables(cloud fi.Cloud, clusterName string) (map[string]*ec2.Ro
 	return routeTables, nil
 }
 
-func ListRouteTables(cloud fi.Cloud, clusterName string) ([]*resources.Resource, error) {
+func ListRouteTables(cloud fi.Cloud, vpcID, clusterName string) ([]*resources.Resource, error) {
```
hakuna-matatah, line 54: I see that many of the functions are now taking vpc-id which is nice, not just  in the `eni.go` file but also in  other files like sqs, routetable etc. What I'm trying to understand is, are we not using these functions anywhere else in the project ? Because I don't see any consumers of these functions are being changed to accommodate this change i.e; the callers ?

hakman: These are used only on cleanup, which is quite demanding.
`vpc-id` can be used as a filter later in other places. For now, let's see how the review goes.

hakuna-matatah:
Thanks for adding these changes very quickly, as this will make life easier for us internally on large scale vpc's. Appreciate the efforts.

johngmyers (approved):

PR #15904, opened 2023-09-12
Increase vpc cidrs to allocate more ip's for large scale k8s clusters

hakuna-matatah:
- Pods and nodes are not coming up due to insufficient IPs, so increasing CIDRs to accommodate this.


hakman (approved):

hakuna-matatah:
> Perhaps larger subnets instead of more subnets?

This is max size of subnets we could have with our requirements i.e; /16.

PR #15900, opened 2023-09-11
Fix e2e testing logic

upodroid:
I cleaned up some logic in the kubetest2-kops deployer. Some flags are automatically enabled when testing on GCE.

PR #15907, opened 2023-09-13
Fix COS image mounter URL

upodroid:
There is a bug caused by the CDN switch recently that is causing COS e2e clusters to not provision properly.

https://github.com/kubernetes/k8s.io/issues/5659

### DIFF
--- a/nodeup/pkg/model/kubelet.go
+++ b/nodeup/pkg/model/kubelet.go
@@ -495,7 +495,7 @@ func (b *KubeletBuilder) addContainerizedMounter(c *fi.NodeupModelBuilderContext
 	// TODO: leverage assets for this tar file (but we want to avoid expansion of the archive)
 	c.AddTask(&nodetasks.Archive{
 		Name:      "containerized_mounter",
-		Source:    "https://dl.k8s.io/gci-mounter/mounter.tar",
+		Source:    "https://storage.googleapis.com/kubernetes-release/gci-mounter/mounter.tar",
 		Hash:      "6a9f5f52e0b066183e6b90a3820b8c2c660d30f6ac7aeafb5064355bf0a5b6dd",
 		TargetDir: path.Join(containerizedMounterHome, "rootfs"),
 	})


nodeup/pkg/model/kubelet.go
```diff
@@ -495,7 +495,7 @@ func (b *KubeletBuilder) addContainerizedMounter(c *fi.NodeupModelBuilderContext
 	// TODO: leverage assets for this tar file (but we want to avoid expansion of the archive)
 	c.AddTask(&nodetasks.Archive{
 		Name:      "containerized_mounter",
-		Source:    "https://dl.k8s.io/gci-mounter/mounter.tar",
+		Source:    "https://storage.googleapis.com/kubernetes-release/gci-mounter/mounter.tar",
```
hakman, line 498: Ugh...

tests/e2e/pkg/tester/skip_regex.go
```diff
@@ -104,10 +104,6 @@ func (t *Tester) setSkipRegexFlag() error {
 	}
 
 	if cluster.Spec.LegacyCloudProvider == "gce" {
-		// Firewall tests expect a specific format for cluster and control plane host names
-		// which kOps does not match
-		// ref: https://github.com/kubernetes/kubernetes/blob/1bd00776b5d78828a065b5c21e7003accc308a06/test/e2e/framework/providers/gce/firewall.go#L92-L100
-		skipRegex += "|Firewall"
```
rifelpet, line 110: The master branch of kubetest2-kops is used to test against older kops and k8s versions, so we likely need to keep this skipped for any kubernetes versions that will not have https://github.com/kubernetes/kubernetes/pull/120628 merged or cherrypicked.

rifelpet: We test against unsupported kubernetes versions. For example, the latest kops supports k8s 1.24: https://github.com/kubernetes/kops/blob/e31bf381abe96d7cb7e47b9afe6873e549abc1c9/upup/pkg/fi/cloudup/apply_cluster.go#L83

And we use the kops master branch's kubetest2-kops to test older kops versions, like [this upgrade test](https://github.com/kubernetes/test-infra/blob/d4d54e2cb23c92bbe4a1e7adeedcdbc71b7c6f8a/config/jobs/kubernetes/kops/build_jobs.py#L984) which runs the k/k e2e suite for k8s 1.23. We should probably reconsider which kops branch we use for running kubetest2-kops on older kops versions, but until then I think we'd need to keep skipping Firewall tests on unsupported k/k branches.

upodroid: I reverted the change to this file. For testing upstream k/k, skip and focus regexes will be set so I can't use this file.

/hold cancel

rifelpet:
I confirmed the --tooling flag is available in all kubernetes versions that kubetest2-kops tests against, so we dont need to guard setting the flag behind a k8s version check 👍🏻 
/lgtm

rifelpet:
I confirmed the --tooling flag is available in all kubernetes versions that kubetest2-kops tests against, so we dont need to guard setting the flag behind a k8s version check 👍🏻 
/lgtm

upodroid:
Can we merge this PR now? I moved the flag change to the #15900 PR

upodroid:
/retest

flake

hakman (approved):

PR #15911, opened 2023-09-14
Make `--cloudonly` sound less innocuous

danports:
How's this wording @hakman?

### DIFF
--- a/cmd/kops/rolling-update_cluster.go
+++ b/cmd/kops/rolling-update_cluster.go
@@ -58,6 +58,9 @@ var (
 	updated with the --force flag.  Rolling update drains and validates the cluster by default.  A cluster is
 	deemed validated when all required nodes are running and all pods with a critical priority are operational.
 
+	If the cluster is in a broken state and cannot be validated, rolling-update will get stuck and eventually 
+	fail; you can force the update to proceed with the --cloudonly flag, which will skip validation.
+
 	Note: terraform users will need to run all of the following commands from the same directory
 	` + pretty.Bash("kops update cluster --target=terraform") + ` then ` + pretty.Bash("terraform plan") + ` then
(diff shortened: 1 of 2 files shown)


cmd/kops/rolling-update_cluster.go
```diff
@@ -189,7 +189,7 @@ func NewCmdRollingUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 
 	cmd.Flags().BoolVarP(&options.Yes, "yes", "y", options.Yes, "Perform rolling update immediately; without --yes rolling-update executes a dry-run")
 	cmd.Flags().BoolVar(&options.Force, "force", options.Force, "Force rolling update, even if no changes")
-	cmd.Flags().BoolVar(&options.CloudOnly, "cloudonly", options.CloudOnly, "Perform rolling update without confirming progress with Kubernetes")
+	cmd.Flags().BoolVar(&options.CloudOnly, "cloudonly", options.CloudOnly, "Perform rolling update without validating cluster status, which may cause downtime - only recommended for troubleshooting broken clusters")
```
hakman, line 192: What do you think?
```suggestion
	cmd.Flags().BoolVar(&options.CloudOnly, "cloudonly", options.CloudOnly, "Perform rolling update without validating cluster status (will cause downtime)")
```

danports: Yeah, that's shorter so that's probably better for CLI output. I added a bit of explanatory text to the command description.

IMO it would be better to rename `--cloudonly` to `--skip-validation` since that is a more accurate description of what it does (and should also raise the right alarm bells in users' heads), but I'm not sure it's worth changing that at this point.

hakman (approved):

PR #15919, opened 2023-09-15
Add `SeccompDefault` kubelet config

colinhoglund:
Updating the kubelet config spec with the `SeccompDefault` feature gate and `--seccomp-default` boolean flag.

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -816,6 +816,20 @@ spec:
 
 Note that Kubelet will fail to install the shutdown inhibtor on systems where logind is configured with an `InhibitDelayMaxSeconds` lower than `shutdownGracePeriod`. On Ubuntu, this setting is 30 seconds.
 
+### SeccompDefault
+
+[SeccompDefault](https://kubernetes.io/blog/2021/08/25/seccomp-default/) enables the use of `RuntimeDefault` as the default seccomp profile for all workloads. (Default: false)
+
+Note that a feature gate is required to enable the feature, and the feature is turned on using kubelet config.
+
+```yaml
+spec:
+  kubelet:
+    featureGates:
+      SeccompDefault: "true"
+    seccompDefault: true
+```
+
 ## kubeScheduler
 
 This block contains configurations for `kube-scheduler`.  See https://kubernetes.io/docs/admin/kube-scheduler/
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -3966,6 +3966,10 @@ spec:
                     description: RuntimeRequestTimeout is timeout for runtime requests
                       on - pull, logs, exec and attach
                     type: string
+                  seccompDefault:
+                    description: SeccompDefault enables the use of `RuntimeDefault`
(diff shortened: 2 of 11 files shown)


pkg/apis/kops/validation/validation.go
```diff
@@ -922,6 +922,10 @@ func validateKubelet(k *kops.KubeletConfigSpec, c *kops.Cluster, kubeletPath *fi
 		if k.MemorySwapBehavior != "" {
 			allErrs = append(allErrs, IsValidValue(kubeletPath.Child("memorySwapBehavior"), &k.MemorySwapBehavior, []string{"LimitedSwap", "UnlimitedSwap"})...)
 		}
+
+		if k.SeccompDefault != nil && c.IsKubernetesLT("1.22") {
+			allErrs = append(allErrs, field.Forbidden(kubeletPath.Child("seccompDefault"), "seccompDefault is only available in Kubernetes >=1.22"))
+		}
```
hakman, line 928: Please remove the validation. K8s 1.22 is no longer supported for some time now.

hakman (changes requested):

docs/cluster_spec.md
```diff
@@ -816,6 +816,20 @@ spec:
...
+[SeccompDefault](https://kubernetes.io/blog/2021/08/25/seccomp-default/) enables the use of `RuntimeDefault` as the default seccomp profile for all workloads. (Default: false)
+
+Note that a feature gate is required to enable the feature, and the feature is turned on using kubelet config.
+
+```yaml
+spec:
+  kubelet:
+    featureGates:
+      SeccompDefault: "true"
+    seccompDefault: true
```
colinhoglund, line 830: This feels a bit unnecessary. Would it make sense to just automatically enable the feature flag when `seccompDefault: true`?

colinhoglund: I went ahead and pushed a commit for this change. Happy to revert if there is a preferred way to handle this type of configuration.

hakman: Thanks for the thought, but it is not the direction we want to go. We expect operators to set these flags and maintain them.

nodeup/pkg/model/kubelet_test.go
```diff
@@ -411,3 +412,24 @@ func Test_BuildComponentConfigFile(t *testing.T) {
 		t.Errorf("Failed to build component config file: %v", err)
 	}
 }
+
+func TestSeccompDefault(t *testing.T) {
+	b := &KubeletBuilder{
+		&NodeupModelContext{
+			NodeupConfig: &nodeup.Config{
+				KubeletConfig: kops.KubeletConfigSpec{
+					SeccompDefault: ptr.To(true),
```
hakman, line 421:
```suggestion
					SeccompDefault: fi.PtrTo(true),
```

hakman (changes requested):

nodeup/pkg/model/kubelet_test.go
hakman: Once you do these changes, please also squash the commits.

hakman:
Thanks @colinhoglund! 🚀 
/lgtm
/approve

PR #15930, opened 2023-09-19
gce: when using only ipv6, we don't need a CNI to set up the network.

justinsb:
IPv6 addresses are "real" addresses, and don't need any further
configuration on GCE.

### DIFF
--- a/cmd/kops-controller/controllers/gceipam.go
+++ b/cmd/kops-controller/controllers/gceipam.go
@@ -18,7 +18,9 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"net"
 	"net/url"
 	"strings"
 
@@ -26,6 +28,8 @@ import (
 	"google.golang.org/api/compute/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


cmd/kops-controller/controllers/gceipam.go
```diff
@@ -139,6 +147,46 @@ func (r *GCEIPAMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			}
+		}
+
+		if allIPv6 {
+			// IPv6 does not require a route to be set up, so mark the node as NetworkReady
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == "NetworkUnavailable" {
+					if condition.Status == corev1.ConditionTrue {
+						newCondition := metav1.Condition{
+							Message: "Node has IPv6",
```
johngmyers, line 176:
```suggestion
							Message: "Node is IPv6-only",
```

johngmyers:
Unrelated to this PR, but consider updating the GCP node identifier to use the non-legacy interface.

cmd/kops-controller/controllers/gceipam.go
```diff
@@ -139,6 +147,46 @@ func (r *GCEIPAMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 	}
 
+	if len(node.Spec.PodCIDRs) != 0 {
```
aojea, line 150: you can make this simpler , this field is already validated in the apiserver https://github.com/kubernetes/kubernetes/blob/54d42e6a658f84ea66593d3107f41db6931a5fbf/pkg/apis/core/validation/validation.go#L4247-L4281

```suggestion
       netutils "k8s.io/utils/net" 


	if len(node.Spec.PodCIDRs) != 1 {
	    return ctrl.Result{}, nil
	 }
	 if netutils.IsIPv4CIDR(node.Spec.PodCIDRs[0]) {
	      return ctrl.Result{}, nil
	 }
	 ...
	 patch condition here to unset NetworkUnavailable
	 ...
	    
```

aojea:
> /hold as @aojea suggested this might be done for us in the latest CCM

/hold cancel

the cloud provider gcp will take some time because the ipam code changes are not straightforward, this ipam here is pretty neat and simple and will allow us to iterate faster

PR #15931, opened 2023-09-19
protokube: support writing AAAA records to /etc/hosts

justinsb:
Though we likely want to stop using this code altogether, it is still
used on GCE (for now).  dns-controller will "gossip" AAAA records, and
they are valid (and work) in /etc/hosts.

### DIFF
--- a/protokube/pkg/gossip/dns/hosts.go
+++ b/protokube/pkg/gossip/dns/hosts.go
@@ -38,18 +38,24 @@ func (h *HostsFile) Update(snapshot *DNSViewSnapshot) error {
 			klog.Warningf("ignoring unexpected lines in /etc/hosts: %v", badLines)
 		}
 
+		hostsToAddr := make(map[string][]string)
 		zones := snapshot.ListZones()
 		for _, zone := range zones {
 			records := snapshot.RecordsForZone(zone)
 
 			for _, record := range records {
-				if record.RrsType != "A" {
-					klog.Warningf("skipping record of unhandled type: %v", record)
-					continue
-				}
+				switch record.RrsType {
+				case "A", "AAAA":
+					hostsToAddr[record.Name] = append(hostsToAddr[record.Name], record.Rrdatas...)
 
-				hostMap.ReplaceRecords(record.Name, record.Rrdatas)
+				default:
+					klog.Warningf("skipping record of unhandled type: %v", record)
+				}
 			}
+		}
+
+		for host, addresses := range hostsToAddr {
+			hostMap.ReplaceRecords(host, addresses)
 		}
 
 		return hostMap, nil


justinsb:
This exercises the node -> apiserver ipv6 code path, rather than relying on ipv4.  Suggested by @aojea , because then we can get one step closer to a pure-IPv6 world.

protokube/pkg/gossip/dns/hosts.go
```diff
@@ -38,20 +38,29 @@ func (h *HostsFile) Update(snapshot *DNSViewSnapshot) error {
 			klog.Warningf("ignoring unexpected lines in /etc/hosts: %v", badLines)
 		}
 
+		hostsToAddr := make(map[string][]string)
```
aojea, line 41: if you make a map then one hostname can only have one IP family, if that is your preference I suggest you to default always to the same one, usually ipv4. Otherwise you can create one mapping per ip familiy, this will give you a cleaner separation in the code between families

justinsb: So this is a map from hosts to ip addresses, the addresses can be a mix of ipv4 and ipv6 (i.e. we build this map by appending to the existing map entries).

I actually introduced this map because _previously_ the code did replace the addresses, so ipv4 would overwrite the ipv6 address.

justinsb: (It's confusing because this map is "backwards" compared to /etc/hosts; I find the mapping in /etc/hosts unintuitive considering how we are using it, so we try to keep the address -> names mapping isolated in `hosts.HostMap`)

aojea: osh, so `record.Name` is an IP address?

aojea: are not `Rrdatas`the ip addresses?

protokube/pkg/gossip/dns/hosts.go
```diff
@@ -38,20 +38,29 @@ func (h *HostsFile) Update(snapshot *DNSViewSnapshot) error {
...
 			records := snapshot.RecordsForZone(zone)
 
 			for _, record := range records {
-				if record.RrsType != "A" {
+				switch record.RrsType {
+				case "A":
+					hostsToAddr[record.Name] = append(hostsToAddr[record.Name], record.Rrdatas...)
+
+				case "AAAA":
+					hostsToAddr[record.Name] = append(hostsToAddr[record.Name], record.Rrdatas...)
```
johngmyers, line 52:
```suggestion
				case "A", "AAAA":
					hostsToAddr[record.Name] = append(hostsToAddr[record.Name], record.Rrdatas...)
```

johngmyers (changes requested):

protokube/pkg/gossip/dns/hosts.go
aojea: oh, nevermind, I got it now

aojea:
LGTM

johngmyers (approved):

PR #15934, opened 2023-09-20
scale-test: Optimise node dumping for large clusters

hakman:
/cc @dims @justinsb @rifelpet

### DIFF
--- a/pkg/dump/dumper.go
+++ b/pkg/dump/dumper.go
@@ -34,6 +34,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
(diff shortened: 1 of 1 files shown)


pkg/dump/dumper.go
```diff
@@ -34,6 +34,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	// MaxDumped is the maximum number of nodes to dump
+	MaxDumped = 500
```
justinsb, line 39: Nit: MaxNodesToDump might be clearer

justinsb:
One naming nit, generally lgtm!

PR #15935, opened 2023-09-20
aws: Fix minor typos for Karpenter setup

junm-cloudnatix:
`on-demand` is the right string to indicate OnDemand in Karpenter. See: https://github.com/aws/karpenter-core/blob/main/pkg/apis/v1alpha5/labels.go#L30 As the result it does not fall back to ondemand instances.

Also add `ec2:DescribeImages` to karpenter IAM policies -- it's noted in https://karpenter.sh/docs/getting-started/migrating-from-cas/#create-iam-roles (the list also has DeleteLaunchTemplates but I don't think this is necessary for kOps).

### DIFF
--- a/pkg/model/components/addonmanifests/karpenter/iam.go
+++ b/pkg/model/components/addonmanifests/karpenter/iam.go
@@ -51,9 +51,11 @@ func addKarpenterPermissions(p *iam.Policy) {
 		// Not included because we require Karpenter
 		// use existing kOps instance group launch templates
 		// "ec2:CreateLaunchTemplate",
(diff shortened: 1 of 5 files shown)


hakman:
@junm-cloudnatix We are having some trouble with our Karpenter tests. I will try to find the root cause early next week. Sorry for the trouble validating the PR.

ndallavalentina:
https://github.com/kubernetes/kops/issues/15948

hakman (approved):

hakman:
Thanks @junm-cloudnatix!

PR #15939, opened 2023-09-21
Update audit configuration in cluster_spec.md

michalbolek:
Since [1.22](https://github.com/kubernetes/kops/blob/ff5b06586d99e86c9dcf1680022e7a0e626305f1/docs/releases/1.22-NOTES.md#control-plane-pods-no-longer-mount-srvkubernetes) pods only mount specific hostPaths that are reachable for kube-apiserver which means the audit policy config has to be there as well if we want it to be reachable by the pod. This has not been reflected in the documentation

### DIFF
--- a/docs/cluster_spec.md
+++ b/docs/cluster_spec.md
@@ -431,10 +431,10 @@ spec:
     auditLogMaxBackups: 1
     auditLogMaxSize: 100
     auditLogPath: /var/log/kube-apiserver-audit.log
-    auditPolicyFile: /etc/kubernetes/audit/policy-config.yaml
+    auditPolicyFile: /srv/kubernetes/kube-apiserver/audit/policy-config.yaml
   fileAssets:
   - name: audit-policy-config
-    path: /etc/kubernetes/audit/policy-config.yaml
+    path: /srv/kubernetes/kube-apiserver/audit/policy-config.yaml
     roles:
     - Master
     content: |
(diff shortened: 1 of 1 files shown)


hakman:
Please explain how you tested, what version of kOps you used.
/hold

michalbolek:
I'm using kops-provider 1.24.0 for terraform which is based on kOps 1.24.0

I have envs with auditing enabled with the policy file created as `/srv/kubernetes/kube-apiserver/audit/policy.yaml` and the resulting api-server params:
```
--audit-log-maxage=7
--audit-log-maxbackup=1
--audit-log-maxsize=1024
--audit-log-path=/var/log/kube-apiserver-audit.log
--audit-policy-file=/srv/kubernetes/kube-apiserver/audit/policy.yaml
```

And the `/srv/kubernetes/kube-apiserver` is a default mount for kube-apiserver from what I've seen.

hakman:
The changes you are doing are valid only for kOps v1.23 and v1.24. This was later fixed in 1.25.

hakman:
I don't mind updating the docs to say that 1.22-1.25 would need to use different dirs instead. What do you think?

michalbolek:
I'm not sure, think it might cause more confusion and make the docs less readable with added. On the other hand it could be just a simple "note" to mind the difference and maybe link to the 1.22 change?

hakman:
> On the other hand it could be just a simple "note" to mind the difference and maybe link to the 1.22 change?

I think a note would be nice, if you have the chance.

PR #15979, opened 2023-09-29
Promote alpha to stable

moshevayner:
Promote changes from #15660 and #15967 onto stable channel

/cc @hakman
Re #15660 I see that it's been over two month since we bumped the recommended version in alpha to 1.27.0, so I took the liberty of updating stable as well. 
Let me know if you think it's still too soon and I can revert that change.

Thanks!


hakman (approved):
Wonderful, thanks @moshevayner! :)

PR #15984, opened 2023-10-02
fix(service): Add missing port 9999 to pod-identity-webhook service

### LINKED ISSUE #15946: pod-identity-webhook Prometheus port missing on service?
/kind bug

 this information.**

1.26.5

 a `kops` flag.**

1.26.8

AWS

`spec.podIdentityWebhook.enabled = true`

The deployment `pod-identity-webhook` was installed. The service `pod-identity-webhook` is created with Prometheus annotations, but it seems that the port 9999 for Prometheus is missing from the `spec.ports` section:

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    prometheus.io/port: "9999"
    prometheus.io/scheme: http
    prometheus.io/scrape: "true"
  labels:
    addon.kops.k8s.io/name: eks-pod-identity-webhook.addons.k8s.io
    app.kubernetes.io/managed-by: kops
    k8s-addon: eks-pod-identity-webhook.addons.k8s.io
  name: pod-identity-webhook
  namespace: kube-system
spec:
  clusterIP: x.x.x.x
  clusterIPs:
  - x.x.x.x
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - port: 443
    protocol: TCP
    targetPort: 443
  selector:
    app: pod-identity-webhook
  sessionAffinity: None
  type: ClusterIP
```

The port 9999 ist defined in the service.

see above

-

The ConfigMap `pod-identity-webhook` has these Prometheus annotations as well. Is this on purpose?

Subham-sarkar:
### What does this PR introduce?

This PR addresses the issue where port 9999 was missing in the `pod-identity-webhook` service definition, causing problems with Prometheus scraping.

### Why is this change needed?

The missing port 9999 was required for Prometheus to scrape metrics from the `pod-identity-webhook` service. Without this port, monitoring and metrics collection were incomplete.

### How does it address the issue?

This PR modifies the service definition to include port 9999 in the `spec.ports` section, ensuring that Prometheus can scrape metrics correctly.

### Related Issue

Fixes #15946

### DIFF
--- a/tests/integration/update_cluster/irsa/data/aws_s3_object_minimal.example.com-addons-eks-pod-identity-webhook.addons.k8s.io-k8s-1.16_content
+++ b/tests/integration/update_cluster/irsa/data/aws_s3_object_minimal.example.com-addons-eks-pod-identity-webhook.addons.k8s.io-k8s-1.16_content
@@ -257,6 +257,9 @@ spec:
   - port: 443
     protocol: TCP
     targetPort: 443
+  - port: 9999
+    protocol: TCP
+    targetPort: 9999
   selector:
     app: pod-identity-webhook
 


hakman:
Hi @Subham-sarkar. Thank you for taking the time to try and fix this issue.
The change should be done in https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/k8s-1.16.yaml.template and run `hack/update-expected.sh` to update the tests output.

PR #15986, opened 2023-10-03
WIP Put control-plane and apiserver nodes in IPv6-only subnets

johngmyers:
AWS now supports IPv6 instance targets, which means we can stop using `DualStack` subnets for the control plane, APIServer nodes, and bastion hosts.


johngmyers:
IPv6 instance targets need to have a primary IPv6 address. Unfortunately the Terraform AWS provider doesn't support setting the necessary field in the launch template. Filed hashicorp/terraform-provider-aws#33733.

johngmyers:
Status: instances are registering as targets, but are failing health checks.

johngmyers:
The issue is that the protocol conversion from IPv4 to IPv6 is causing the traffic between LB and backend to not match any security group rule.

PR #15987, opened 2023-10-03
aws: Don't add dependency on additional CIDR for shared VPC

hakman:
`VPCCIDRBlock` tasks are not created when the VPC is shared:
https://github.com/kubernetes/kops/blob/ecda9e8652e660e29dc8f25a56e7a8b834eeb574/pkg/model/awsmodel/network.go#L97-L116


zetaab (approved):
/approve
/lgtm

great!

PR #15989, opened 2023-10-03
Allow setting map[string]string from the command line

hakman:
```shell
> kops create cluster \
    --set spec.kubeAPIServer.runtimeConfig=extensions/v1beta1=true \
    --set spec.kubeAPIServer.runtimeConfig=scheduling.k8s.io/v1alpha1=true
...

> kops get cluster -o yaml
...
  kubeAPIServer:
    runtimeConfig:
      extensions/v1beta1: "true"
      scheduling.k8s.io/v1alpha1: "true"
...
```
/cc @upodroid

### DIFF
--- a/util/pkg/reflectutils/access.go
+++ b/util/pkg/reflectutils/access.go
@@ -134,6 +134,23 @@ func setType(v reflect.Value, newValue string) error {
 
 	t := v.Type().String()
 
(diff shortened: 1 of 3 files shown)


upodroid (approved):
Thanks

I'll rebase the alpha PR with this change

PR #15997, opened 2023-10-05
TECH-12913 Removing Alert for KubeDaemonSetRolloutStuck

rex-invoca:
TECH-12913 Removing Alert for KubeDaemonSetRolloutStuck

Ticket:  https://invoca.atlassian.net/browse/TECH-12913

### DIFF
--- a/addons/prometheus-operator/v0.26.0.yaml
+++ b/addons/prometheus-operator/v0.26.0.yaml
@@ -15373,18 +15373,6 @@ spec:
       for: 15m
       labels:
         severity: critical
-    - alert: KubeDaemonSetRolloutStuck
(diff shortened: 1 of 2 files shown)


rex-invoca:
This is my first tim working with these, so I converted this to a draft to ensure it wouldn't be merged.

This and:

https://github.com/Invoca/invocaops_docker/pull/3728

Are for discussion purposes, please ignore for the time being.

hakman:
Hi @rex-invoca, these addons are deprecated and unmaintained. They are kept only to not break very old setups.

Could you explain your use case and why you need the change?

rex-invoca:
no-op

PR #15999, opened 2023-10-05
Introduce KubernetesE2ETesting feature flag

upodroid:
We want kops to do things in E2E clusters that it shouldn't be doing by default in prod clusters.

One of these examples is installing nfs clients. kops used to do this but it was changed last year in #13577

### DIFF
--- a/nodeup/pkg/model/packages.go
+++ b/nodeup/pkg/model/packages.go
@@ -17,6 +17,7 @@ limitations under the License.
 package model
 
 import (
+	"k8s.io/kops/pkg/featureflag"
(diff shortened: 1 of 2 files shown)


nodeup/pkg/model/packages.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package model
 
 import (
+	"k8s.io/kops/pkg/featureflag"
```
hakman, line 20: Feature flags are not passed to NodeUp.

upodroid:
closing for now

PR #16004, opened 2023-10-06
WIP Use droplet ID for node names in DO

hakman:
/cc @rifelpet

### DIFF
--- a/docs/releases/1.29-NOTES.md
+++ b/docs/releases/1.29-NOTES.md
@@ -12,6 +12,10 @@ This is a document to gather the release notes prior to the release.
 
 ## Openstack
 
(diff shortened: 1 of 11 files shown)


rifelpet:
Whats the advantage of using the droplet ID over the IP address for the node name? I suppose it matches AWS' instance ID node names, any other benefits?

hakman:
> Whats the advantage of using the droplet ID over the IP address for the node name? I suppose it matches AWS' instance ID node names, any other benefits?

See https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/docs/getting-started.md#--provider-iddigitaloceandroplet-id

PR #16008, opened 2023-10-08
Configure ginkgo to use the `KUBE_BASTION_SSH` env variable

upodroid:
I need to fix a bug in kubetest2 to make this work. https://github.com/kubernetes-sigs/kubetest2/pull/247


upup/pkg/fi/cloudup/gce/instancegroups.go
```diff
@@ -257,3 +263,30 @@ func matchInstanceGroup(mig *compute.InstanceGroupManager, c *kops.Cluster, inst
...
+			cm.Roles = append(cm.Roles, role)
+			cm.PrivateIP = fi.ValueOf(&instance.NetworkInterfaces[0].NetworkIP)
+			cm.ExternalIP = fi.ValueOf(&instance.NetworkInterfaces[0].AccessConfigs[0].NatIP)
+		}
+	}
+	if isControlPlane {
+		cm.Roles = append(cm.Roles, "control-plane")
+		cm.PrivateIP = fi.ValueOf(&instance.NetworkInterfaces[0].NetworkIP)
+		cm.ExternalIP = fi.ValueOf(&instance.NetworkInterfaces[0].AccessConfigs[0].NatIP)
+	}
```
johngmyers, line 291: Why does this avoid setting the PrivateIP and ExternalIP when there is no role label? Why not unconditionally set those regardless of the role label?

tests/e2e/pkg/tester/tester.go
```diff
@@ -141,6 +129,29 @@ func (t *Tester) getKopsInstanceGroups() ([]*api.InstanceGroup, error) {
...
+	}
+	if cluster.Spec.LegacyCloudProvider == "gce" {
+		instances, err := kops.GetInstances("kops", cluster.Name, nil)
+		if err != nil {
+			return err
+		}
+		for _, instance := range instances {
+			for _, role := range instance.Roles {
+				if role == "control-plane" && instance.ExternalIP != "" {
+					t.Env = append(t.Env, fmt.Sprintf("KUBE_SSH_BASTION=%s", instance.ExternalIP))
```
johngmyers, line 147: For a three-node control plane we would set KUBE_SSH_BASTION three times?

hakman:
This is more or less a hack (as you also said it in Slack). It is probably the easy way to do it, but not quite sure it's the right thing to do.
Why can't something be done in k/k to list the control-plane IPs and choose one of them? I think I've seen in the past some tests that do that.

@aojea Any thoughts?

tests/e2e/pkg/tester/tester.go
upodroid: Forgot about that, we need just one IP.

upup/pkg/fi/cloudup/gce/instancegroups.go
upodroid: It doesn't, look at 282 to 284.

PR #16011, opened 2023-10-10
feat: add cpu option setting for aws

Yitaek:
Adding CPU options setting for AWS EC2 instances


Yitaek:
@johngmyers or @zetaab could i get a review please? i mostly followed what @hakman's example of adding instance metadata. aws sdk already had the required fields for cpu options so mostly exposing it to kops api. 

admittedly there could be more validation done for cpu options, but could not find an easy way to pull the specific settings per instance without creating the mapping myself (any suggestions there welcome)

PR #16016, opened 2023-10-11
scaleway: IPAM integration

Mia-Cross:
Since the release of IPAM at Scaleway, some services (like load-balancers) no longer work with the IP provided by the instance API.
This PR makes all kOps components use the IPs from IPAM instead.

A PR has also been made on the [etcd-manager](https://github.com/kubernetes-sigs/etcdadm/pull/385) and works together with this one. 

This is the first step for the integration of private networks, which is coming soon.

### DIFF
--- a/protokube/pkg/gossip/scaleway/seeds.go
+++ b/protokube/pkg/gossip/scaleway/seeds.go
@@ -19,7 +19,6 @@ package scaleway
 import (
 	"fmt"
 
-	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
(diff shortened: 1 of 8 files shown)


Mia-Cross:
As it is in this PR, the code works for clusters with DNS and with no DNS, but not for those with gossip DNS because the initialization of the secondary gossip fails. This is due to the fact that the `memberlistmesh` package expects only private IPs, which are not provided by IPAM, whereas the legacy instance private IPs don't work anymore.

This problem will be fixed in the next PR implementing private networks since it will allow instances to have a private IP provided by IPAM.

hakman (approved):

PR #16020, opened 2023-10-14
Experimental support for tracing

justinsb:
Implement simple otel tracing, starting with the kOps CLI.

* We implement a custom collector which writes to a otel proto to a file.  Eventually we can gather these artifacts in tests.
* Implement a simple server that serves the jaeger UI.  It's very inefficient currently (and loads everything into memory), but may be good enough for small traces.

Prompted by discussion in kOps office hours about how it would be understanding where we are spending time e.g. on node startup.


justinsb:
I'm working on cleaning this up to be a reasonable PR (reasonable size, better factored).  I started by splitting out #16021 which is mostly orthogonal (but supporting) changes to add context.Context to more methods in the vfs layer.

justinsb:
cc @rifelpet

justinsb:
Added some basic docs in the last commit (including describing it as very early experimental support), and so removing WIP marker.

cmd/kops/main.go
```diff
@@ -16,9 +16,44 @@ limitations under the License.
...
+	}
+
+	otelShutdown, err := setupOTelSDK(ctx, serviceName, serviceVersion)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		return err
+	}
+	// Handle shutdown properly so nothing leaks.
+	defer func() {
+		if err := otelShutdown(context.Background()); err != nil {
```
rifelpet, line 49: Should this use the existing context from above? if not it might be worth a comment explaining why it has its own

util/pkg/vfs/s3fs.go
```diff
@@ -286,6 +288,9 @@ func (p *S3Path) getRequestACL(aclObj ACL) (*string, error) {
 }
 
 func (p *S3Path) WriteFile(ctx context.Context, data io.ReadSeeker, aclObj ACL) error {
+	ctx, span := tracer.Start(ctx, "S3Path::WriteFile")
```
rifelpet, line 291:
```suggestion
	ctx, span := tracer.Start(ctx, "S3Path::WriteFile", trace.WithAttributes(attribute.String("path", p.String())))
```

docs/opentelemetry.md
```diff
@@ -0,0 +1,23 @@
+# OpenTelemetry support
+
+kOps is experimenting with initial support for OpenTelemetry, starting with tracing.
+
+The support should be considered experimental; the trace file format and schema will likely change, and these initial experiments might be removed entirely.
+
+kOps supports a "serverless" mode of operation, where log the OpenTracing output to a file.  We do this because our e2e test runner (prow) doesn't yet have a destination for OpenTelemetry data.
```
rifelpet, line 7:
```suggestion
kOps supports a "serverless" mode of operation, where it logs the OpenTracing output to a file.  We do this because our e2e test runner (prow) doesn't yet have a destination for OpenTelemetry data.
```

rifelpet (commented):
Nice work! Do you think we could either set that env var in the presubmit prow jobs, or push a temporary commit to this PR that sets it by default to a filepath within the prow artifacts path? Just to see an example file

util/pkg/vfs/s3fs.go
justinsb: Thanks!   I'm still trying to figure out when best to use attributes etc!

cmd/kops/main.go
justinsb: Added a comment "// We use a background context because the main context has probably been shut down."  I think that's why we're doing this.  I got it from some otel example boilerplate.

justinsb:
> Do you think we could either set that env var in the presubmit prow jobs, or push a temporary commit to this PR that sets it by default to a filepath within the prow artifacts path? Just to see an example file

Good idea.  In the last two commits I tweaked it to (1) write files to a directory and set up kubetest to default that directory if artifacts is set and (2) tweaked the reader to use the VFS layer.

So now from this branch you can do:

```
cd tools/otel/traceserver
go run . --src gs://kubernetes-jenkins/pr-logs/pull/kops/16020/pull-kops-e2e-k8s-gce-cilium/1722995655665782784/artifacts/otlp/ --run jaeger
```

http://127.0.0.1:16686/trace/5152af7cbbfc499316aba7df80ce1a76 is then the most interesting trace (I think).

rifelpet (commented):
This might be worthy of a release note. Even though it is only relevant to kops development, it might head off any concerns about introducing telemetry to the project, clarifying that it doesn't "phone home" or connect to any remote servers.

otherwise LGTM!

justinsb:
Good idea @rifelpet (and thanks for uploading the picture) - added a release note to highlight the fact that this is no telemetry in the "phone home" sense!

rifelpet (approved):

PR #16022, opened 2023-10-14
openstack: fix instance group validation if using serverGroupName annotation

zetaab:


zetaab:
/hold

Still running rolling-update tests for clusters which contains serverGroupName annotation, and cluster which does not have that.

upup/pkg/fi/cloudup/openstack/server_group.go
```diff
@@ -84,80 +84,52 @@ func listServerGroups(c OpenstackCloud, opts servergroups.ListOptsBuilder) ([]se
...
+func osBuildCloudInstanceGroup(c OpenstackCloud, cluster *kops.Cluster, ig *kops.InstanceGroup, nodeMap map[string]*v1.Node) (*cloudinstances.CloudInstanceGroup, error) {
 	cg := &cloudinstances.CloudInstanceGroup{
-		HumanName:     newLaunchConfigName,
+		HumanName:     ig.Name,
 		InstanceGroup: ig,
 		MinSize:       int(fi.ValueOf(ig.Spec.MinSize)),
-		TargetSize:    int(fi.ValueOf(ig.Spec.MinSize)), // TODO: Retrieve the target size from OpenStack?
+		TargetSize:    int(fi.ValueOf(ig.Spec.MinSize)),
 		MaxSize:       int(fi.ValueOf(ig.Spec.MaxSize)),
-		Raw:           &g,
```
zetaab, line 120: what is purpose of this Raw field? Just thinking is it in use in some another place

zetaab:
/hold cancel

everything seems working fine

hakman (approved):

Wieneo:
@zetaab 
Sorry for reopening this old issue but we just tried the new kOps version and it seems to break with multiple kOps clusters in one OpenStack project.
kOps update fails because instances from other clusters get marked as "has not yet joined cluster".

Wieneo:
I have opened #16106, explaining the issue further 🙂

PR #15866, opened 2023-09-04
Default to 100.64.0.0/13 as IPv4 service cluster IP range

hakman:
Closes https://github.com/kubernetes/kops/issues/15862

PR #16028, opened 2023-10-18
Document breaking changes for serviceClusterIPRange and OpenStack with S3 state store

ederst:
When upgrading from kOps 1.25/26 to 1.27/1.28 I experienced some breaking changes:

I relied on the calculation of the `serviceClusterIPRange` from `nonMasqueradeCIDR` which was removed in https://github.com/kubernetes/kops/pull/15866. This resulted in an error with the certificate as the certificate for the API was still issued for the previous service IP. Setting the `serviceClusterIPRange` to the previously calculated one was doing the trick.

Also, when having clusters running on OpenStack with gossip and the state stored in S3, newly created nodes will not join the cluster, as the `S3_*` variables are not passed on to the nodes anymore, [due to the change here](https://github.com/kubernetes/kops/pull/15646/files#diff-c1c852aea92883d3117fae99c1610c5cdcbc217f5aa3b340f23b5dd02e971d99L137-L142). The [attempt to fix this was closed](https://github.com/kubernetes/kops/pull/15691). Could be that this affects other clouds as well, but I have not checked this.

So, if those changes are legit (and not bugs), then I'd say let's document them at least as breaking changes.

### DIFF
--- a/docs/releases/1.27-NOTES.md
+++ b/docs/releases/1.27-NOTES.md
@@ -33,6 +33,10 @@ they would do so when the respective `topology` was set to `public`.
 
 # Breaking changes
 
+## OpenStack
+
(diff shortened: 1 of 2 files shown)


zetaab (approved):
/ok-to-test

crypticophelia:
Having recently gone through the pain of upgrading to 1.28.1 caused by this issue, it would have been superb if perhaps this **_BREAKING CHANGE_** note had been cherry picked back into the release notes. It took us 3 broken clusters, and two rebuilds to finally figure out what shenanigans were occurring here, and then only by pure chance of this issue sifting to the top of the search results.

Never once did it occur to any of us to check the 1.29 alpha release notes for 1.28 breaking change information, and I cannot fathom a world where it would be normal to operate in that manner.

Unimpressed.

ederst:
@crypticophelia The documentation is only generated from the main branch, so at least the information is there (https://kops.sigs.k8s.io/releases/1.28-notes/). But yeah it can be misleading if information like this is not backported to the appropriate release branch and then only looking at the MD file in there.

@hakman @zetaab  So I am not sure if this is an oversight or just a "flaw" in how documentation is treated (no cherry picks necessary due to gen from main branch).

PR #16037, opened 2023-10-21
Thread context through some slower AWS tasks

justinsb:
There are still too many context.TODOs here for this to join all the
way up, but we should be able to better understand the slowest tasks.


justinsb:

Looks like an unrelated storage flake

rifelpet (approved):

PR #16045, opened 2023-10-23
Fix tests for multi-arch clusters

hakman:
kOps assumes that both ARM64 and AMD64 binaries are built when creating a new cluster.
When building Kubernetes for testing, we want a specific build arch, but only in that case.

/cc @upodroid @aojea @justinsb


upodroid (approved):
LGTM

PR #16061, opened 2023-10-25
Limit sanitized URLs to 200 characters

hakman:


justinsb:
Thanks @hakman.  I was a little concerned that this is a utility function and truncation might not be appropriate, but in fact it's only used either with a hash or in a unique temp directory, so this works!

/approve
/lgtm

justinsb:
(We might want to see if we can come up with a better name though, like "BuildSafeFileName" or something like that )

hakman:
> (We might want to see if we can come up with a better name though, like "BuildSafeFileName" or something like that )

@justinsb It is actually used only for URLs, so could just change to `SanitizeURL`?

PR #16069, opened 2023-10-28
dump: include the various gateways on AWS

justinsb:
I think we were leaking egress-only-internet-gateways, it would be
helpful to see their status in the dump.


justinsb:
I think this is related to issue https://github.com/kubernetes/kops/issues/16049

hakman (approved):

PR #16072, opened 2023-10-30
WIP: Wait for AWS NAT Gateways to be ready

justinsb:
At least in theory, NAT-Gateway creation is asynchronous on AWS.  We
now wait to be sure it is successfully created.

### DIFF
--- a/cloudmock/aws/mockec2/natgateway.go
+++ b/cloudmock/aws/mockec2/natgateway.go
@@ -38,6 +38,9 @@ func (m *MockEC2) CreateNatGatewayWithId(request *ec2.CreateNatGatewayInput, id 
 		Tags:         tags,
 	}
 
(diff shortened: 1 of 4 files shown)


justinsb:
Looking at the toolbox dump, I don't think this is the problem with the IPv6 tests, but I think I can probably try a few things here and see if any of them make a difference!

justinsb:
OK, so unexpectedly that passed with a hack to always send a multi-part mime cloud-init config, and to configure debugging.

"Only works with debug enabled" is going to be fun to figure out :joy: 

Running the test again to see if it was just a random fluke (there was also an oddity in that some other VM was dumped by IP address, still trying to figure out what that one was):

justinsb:
Rebased this one, but I don't think we need it.  If we do, we can revisit.

/close

PR #16073, opened 2023-11-01
toolbox dump: include AWS route table objects

justinsb:


justinsb:
This "can't" be the fix, but #16072 passed with this included.  I think the cloud-init change is more likely to be the cause (though that doesn't make a lot of sense), but this can serve as our "control"

justinsb:
/wip so that it doesn't merge, I don't think it's particularly time-critical

justinsb:
This turned out not to be related, but I think we should be dumping all these objects.  Debating whether we should just default the Dump function, but as we still have to set object I though a bit of repetition here wasn't the end of the world.

hakman (approved):

Issue #15720, opened 2023-07-30
amazonvpc is not working with Ubuntu 22.04(Jammy)


h3poteto:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
```
$ kops version
Client version: 1.27.0 (git-v1.27.0)
```

I tried the same thing with the master branch (a8fa8952ba9a489ba335e29484075df5639fb02f).

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
```
$ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"archive", BuildDate:"2023-06-15T08:14:06Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.4", GitCommit:"fa3d7990104d7c1f16943a67f11b154b71f6a132", GitTreeState:"clean", BuildDate:"2023-07-19T12:14:49Z", GoVersion:"go1.20.6", Compiler:"gc", Platform:"linux/amd64"}
```

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
```
$ kops create -f cluster.yaml
$ kops update cluster --name $CLUSTER_NAME --admin --yes
```

**5. What happened after the commands executed?**
A Kubernetes cluster is created, but some pods are not working. So `kops validate` command fails.
For example, cert-manager and ebs-csi-node say errors.

```
$ k get pods -n kube-system 
NAME                                     READY   STATUS              RESTARTS        AGE
aws-cloud-controller-manager-hlxjj       1/1     Running             0               14m
aws-cloud-controller-manager-p2ggb       1/1     Running             0               14m
aws-cloud-controller-manager-p5nk6       1/1     Running             0               13m
aws-iam-authenticator-9zz6q              0/1     ContainerCreating   0               13m
aws-iam-authenticator-mnmgq              0/1     ContainerCreating   0               13m
aws-iam-authenticator-pnncn              0/1     ContainerCreating   0               13m
[...]

PR #16083, opened 2023-11-08
Document incompatability with Amazon VPC CNI and 22.04

rifelpet:
Relates to https://github.com/kubernetes/kops/issues/15720


hakman (approved):
Thanks @rifelpet! 😄

PR #16094, opened 2023-11-11
Release 1.27.2

hakman:


hakman:
> [It] [sig-cli] Kubectl client Simple pod should return command exit codes should handle in-cluster config
> [It] [sig-cli] Kubectl client Simple pod should support exec using resource/name

/retest

rifelpet (approved):

PR #16103, opened 2023-11-14
gce: Support additional StorageClasses and change default

sl1pm4t:
When the the GCE CSI driver is enabled, kOps installs a single StorageClass called `standard-csi`. This SC is backed by the Google Cloud `pd-standard` disk type, which is the slowest spinning disk variety.
https://cloud.google.com/compute/docs/disks#disk-types

On top of being slow, the `pd-standard` disks are not supported on some of the newer GCP Machine Types, such as the C3 family:
https://cloud.google.com/compute/docs/general-purpose-machines#c3_disks

For these reasons it would be great if kOps installed additional StorageClasses, for better performance and broader compatibility.

This change adds two additional StorageClasses named `balanced-csi` and `ssd-csi` to support the `pd-balanced` and `pd-ssd` GCP Persistent Disk types respectively. (I was trying to follow the pattern of the existing `standard-csi` when naming these, but I see over in the AWS side, kOps uses a SC names like `kops-ssd-1-17`, so let me know if that is required here also). 

~~This PR also updates the default StorageClass to be `balanced-csi`. Which is the default set on GKE clusters.
Looking at the AWS SC history, I see kOps has switched the default in the past, but let me know if this is something that needs better backwards compatibility.~~

### DIFF
--- a/tests/integration/update_cluster/ha_gce/data/aws_s3_object_ha-gce.example.com-addons-bootstrap_content
+++ b/tests/integration/update_cluster/ha_gce/data/aws_s3_object_ha-gce.example.com-addons-bootstrap_content
@@ -48,7 +48,7 @@ spec:
     version: 9.99.0
   - id: k8s-1.23
     manifest: gcp-pd-csi-driver.addons.k8s.io/k8s-1.23.yaml
-    manifestHash: 841d9adcf46c3c58b460d6a8eff1ef315288d5700f75007bb3944261c8405e17
+    manifestHash: b8815d8ac98d4fd8003d2429cfb23bb11160f3c1295b8266fda1900910061aab
     name: gcp-pd-csi-driver.addons.k8s.io
     selector:
       k8s-addon: gcp-pd-csi-driver.addons.k8s.io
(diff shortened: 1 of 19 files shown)


sl1pm4t:
I removed the change to the default storage class as discussed in the office hours.

hakman (approved):
Thanks @sl1pm4t! 👍

justinsb:
Thanks - I had a look at what GKE does as well, and it looks like it actually has two different classes:

```
> k get storageclass -A | grep pd
premium-rwo                 pd.csi.storage.gke.io          Delete          WaitForFirstConsumer   true                   4d20h
regional-fast               kubernetes.io/gce-pd           Delete          WaitForFirstConsumer   false                  4d20h
standard                    kubernetes.io/gce-pd           Delete          Immediate              true                   4d20h
standard-rwo (default)      pd.csi.storage.gke.io          Delete          WaitForFirstConsumer   true                   4d20h
```

Thank matches with https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/gce-pd-csi-driver#create_a_storageclass

Focusing on the CSI drivers:

* standard-rwo is pd-balanced
* premium-rwo is pd-ssd

I don't love the "rwo" - I think it means "read-write one" vs rwx which maps to filestore and somehow means "shared".

But I guess "standard" and "premium" are more acceptable ways of saying "cheap" and "fast", which I think is the big distinction people want to make (as we talked about in office hours).

Not sure how best to proceed (i.e. should we care about matching names, when they don't match on AWS vs GCP anyway?), I guess a good way to start is to make sure that the GCE-PD CSI driver has a manifest, and then we can have the debate there!

PR #16110, opened 2023-11-19
Update kube-router to 1.6

aauren:
This PR, updates kube-router to the latest stable 1.X release (1.6.0). It is primarily a dependency update release and has been released for several months. If this goes well, the next step would be to upgrade to kube-router 2.X which is the most recent stable release, but comes with a few breaking changes.

I have also removed the deprecation notices for kube-router since I'm now committing to maintaining it. I have split out a 3rd commit that removes the deprecation notices for kube-router from the kops release notes. However, I left it as a separate commit, because I'm not sure if the project considers release notes like that immutable. Let me know the best way to proceed, and I can modify the PR accordingly.

After upgrading it, I have tested it against my own kops cluster and I have also run the kubetest2 e2e tests. kube-router's test results against the `sig-network` conformance tests are the same both before and after the upgrade. It currently fails 9 of the `sig-network` conformance tests. This is something that I was already interested in fixing upstream, so I'll work on that on the 2.X release line over the course of the next several months.

FYI @hakman

### DIFF
--- a/docs/networking/kube-router.md
+++ b/docs/networking/kube-router.md
@@ -1,7 +1,5 @@
 # Kube-router
 
-&#9888; The Kube-router CNI is not supported for Kubernetes 1.28 or later.
-
 [Kube-router](https://github.com/cloudnativelabs/kube-router) is project that provides one cohesive solution that provides CNI networking for pods, an IPVS based network service proxy and iptables based network policy enforcement.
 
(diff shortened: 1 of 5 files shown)


hakman:
Thanks @aauren, much appreciated! :)
/lgtm
/approve

Issue #16027, opened 2023-10-17
IRSA - Can't create IAM role when namespace has wildcard char


federicopires:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Client version: 1.28.0

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

Client Version: v1.28.2
Server Version: v1.28.2

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
kops update cluster

**5. What happened after the commands executed?**

**6. What did you expect to happen?**
Tried to update cluster with:

```
iam:
   serviceAccountExternalPermissions:
      - name: some-service
        namespace: "*"
        aws:
          policyARNs:
            - arn:aws:iam::xxxxxxx:policy/my_policy
```
(Same applies for `namespace: "foo-*"`)

`W1017 14:57:05.689738   69794 executor.go:139] error running task "IAMRole/some-service.wildcard.sa.k8s.xxx.com" (9m9s remaining to succeed): error creating IAMRole: ValidationError: 1 validation error detected: Value '*' at 'tags.7.member.value' failed to satisfy constraint: Member must satisfy regular expression pattern: [\p{L}\p{Z}\p{N}_.:/=+\-@]*`

Probably need to replace `*` with `wildcard` whenever a tag in AWS is created for a resource. In this case maybe it's in https://github.com/kubernetes/kops/blob/master/pkg/model/context.go#L204 and maybe also in https://github.com/kubernetes/kops/blob/master/pkg/model/context.go#L251 similar as done in https://github.com/kubernetes/kops/blob/master/pkg/model/awsmodel/iam.go#L264

Maybe something like:

`return strings.ReplaceAll(tags, "*", "wildcard")`

could do the trick for both `CloudTagsForServiceAccount` and `CloudTags` functions?

PR #16113, opened 2023-11-19
Replace * in IAM tags when IRSA namespace has wildcard

h3poteto:
fix: https://github.com/kubernetes/kops/issues/16027

I think `*` char is allowed in namespace of IRSA, so it should be possible to create.


hakman:
@h3poteto To update golden output automatically, run hack/update-expected.sh:
```diff
+     "service-account.kops.k8s.io/namespace"     = "test-wildcard"
-     "service-account.kops.k8s.io/namespace"     = "test-*"
```

hakman:
Thanks @h3poteto.
/lgtm
/approve

PR #16125, opened 2023-11-23
kops: Don't try to split --set flags on commas

justinsb:
Passing quotes in very difficult with the --set flag, because the
pflag library tries to parse it using strict CSV semantics.  We switch
to using the StringArrayVar, which does not attempt splitting.

This is a breaking change; compound flags must now be written as two flags, but I think that is clearer.

Example: `--set=a,b` must now be `--set=a --set=b`

### DIFF
--- a/cmd/kops/create_cluster.go
+++ b/cmd/kops/create_cluster.go
@@ -408,9 +408,9 @@ func NewCmdCreateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 		return []string{"json", "yaml"}, cobra.ShellCompDirectiveNoFileComp
 	})
 
-	cmd.Flags().StringSliceVar(&options.Sets, "override", options.Sets, "Directly set values in the spec")
+	cmd.Flags().StringArrayVar(&options.Sets, "override", options.Sets, "Directly set values in the spec")
 	cmd.Flags().MarkDeprecated("override", "use --set instead")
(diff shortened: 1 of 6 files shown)


hakman:
This would break some scenarios, like scalability tests.
/hold

justinsb:
> This would break some scenarios, like scalability tests.

Ah, bummer.  I do think `--set=foo=bar --set=foo2=bar2` is more readable than `--set=foo=bar,foo2=bar2`, particularly when the paths get long, but it is a breaking change.

We could also just do a more naive splitting on commas.  We can also use the CSV splitter with LazyQuotes=true, this was proposed in https://github.com/spf13/pflag/pull/371

hakman:
I think just fixing the scalability scenario would be enough. It's not such a bad change.
https://github.com/kubernetes/kops/blob/b4d787f390abcf1a090fc07a8cbf921fba2ccbb8/tests/e2e/scenarios/scalability/run-test.sh#L89

justinsb:
I do think you raise a good point though, and it's possible that people are setting compound flags "in the wild" that we don't know about, so I don't particularly want to break them if we can avoid it.

The fix is actually pretty easy, in #16128 - we just need to set LazyQuotes: true.  It's a bit harder than it should be, but it would be very easy if pflag merged the support.

My 2c: #16128 unblocks us without breaking anyone...

hakman:
https://github.com/kubernetes/kops/pull/16128 was merged.
/close

PR #16128, opened 2023-11-24
kops: Support quotes in --set flags

justinsb:
The pflag library tries to split comma separated values using strict
CSV semantics, which requires quoting of the full value if the value
contains a quote.  We switch to LazyQuotes which is much more
tolerant.


justinsb:
This is an alternative to #16125, which should be non-breaking (at the cost of more code)

hakman (approved):

PR #16130, opened 2023-11-24
aws: support custom naming with external ccm

JBOClara:
Allow to customize ResourceBasedNaming subnet configuration

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -5801,6 +5801,10 @@ spec:
                       description: Region is the region the subnet is in, set for
                         subnets that are regionally scoped
                       type: string
+                    resourceBasedNaming:
+                      description: ResourceBasedNaming is used to enable/disable resource
+                        based naming for this subnet
+                      type: boolean
                     type:
                       description: SubnetType string describes subnet types (public,
                         private, utility)
--- a/pkg/apis/kops/cluster.go
+++ b/pkg/apis/kops/cluster.go
@@ -777,6 +777,8 @@ type ClusterSubnetSpec struct {
 	PublicIP string `json:"publicIP,omitempty"`
 	// AdditionalRoutes to attach to the subnet's route table
 	AdditionalRoutes []RouteSpec `json:"additionalRoutes,omitempty"`
+	// ResourceBasedNaming is used to enable/disable resource based naming for this subnet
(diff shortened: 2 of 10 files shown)


hakman:
Hi @JBOClara. Did you test this change?
/ok-to-test

JBOClara:
I have tested without applying for now. I need to build a test cluster to test with '--yes' today.

azman0101:
I'll check the failed tests and make some changes

hakman:
> I have tested without applying for now. I need to build a test cluster to test with '--yes' today.
> Do I need to add a unit test for this change ?

I think the main focus should be to have a functional approach and we can discuss later what needs to be added or changed to merge the PR. As is at the moment, I don't see this having any effect in tests, as nodes still use resource based names.

To help, here is the commit hat removed support for non resource based names:
https://github.com/kubernetes/kops/pull/15872/commits/1ea0fd30045a6adcb5e213a7688e70593b0263ef

JBOClara:
I'm not entirely clear on the reasoning behind the node naming change when using CCM. We're just starting our migration to 1.24.

According to https://cloud-provider-aws.sigs.k8s.io/prerequisites/#node-name-conventions, AWS supports two naming conventions: IP-based or resource-based naming.

We want to be able to use the 'IP name' subnet configuration, but since we manage our VPC and subnet with kops, there's no way to switch this parameter with kops. I tried to change this parameter on a cluster's subnets, but kops rolled back all my changes.

In response to https://github.com/kubernetes/kops/pull/12864#issuecomment-986405851, it was suggested that if admins don't want to use resource-based domain names, they can create their own shared subnets. However, we want to keep, as far as possible, using our subnets managed by kops.
Could you please clarify why we need to use shared subnets?

hakman:
> I'm not entirely clear on the reasoning behind the node naming change when using CCM. We're just starting our migration to 1.24.

I will try to be as clear as I can. This was done 2 years ago and the reasons are irrelevant. We asked people that want to keep the old behaviour to contribute it and we will be happy to review the change.
I am not aware of any way to make this work the way you want with latest kOps. I don't think the suggestion in that comment will work anymore. You can try to set the ownership tag on the subnet to `"kubernetes.io/cluster/<cluster-name>" = "shared"` and kOps will not change it anymore, but that is at best a temporary solution.

azman0101:
Ok, I'm closing this PR so

JBOClara:
Ok 👍, thanks for your help, I'm closing this PR

Issue #16084, opened 2023-11-09
Have periodic E2E jobs use the same branch for kops and kubetest2-kops


rifelpet:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**

Currently all E2E jobs use the HEAD of master for building kubetest2-kops, regardless of the kops branch being tested.

This was convenient during initial kubetest2-kops development, to make sweeping changes that impact all jobs.

The downside of this is when kubetest2-kops changes require backwards-incompatible changes to kops and its release process. We need to cherrypick changes to kops and/or kubetest2-kops to ensure all tested versions remain in sync.

For example, [this job](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-calico-deb10-k26-ko26/1721326387475779584) is failing because something changed with the kops staging bucket location in master, but it hasn't been cherrypicked back to all tested kops versions.  

Additionally, newer kubetest2 versions only support testing k8s versions newer than kops' current grid.

For example, [this job](https://testgrid.k8s.io/kops-grid#kops-grid-calico-amzn2-k24-ko27) uses kops 1.27 to test k8s 1.24 and is currently failing because k8s 1.24's e2e suite uses ginkgo v1, but the master branch of kubetest2-kops depends on a version of kubetest2 that [requires ginkgo v2](https://github.com/kubernetes-sigs/kubetest2/pull/236/files#diff-51419d4eb061f005f23df1255e63a2c27bce80ed5995415f3179da05947a67a6L81-L84)

```
 Copying gs://kubernetes-release/release/v1.24.17/kubernetes-test-linux-amd64.tar.gz...
/ [0 files][    0.0 B/114.2 MiB]                                                
-
- [1 files][114.2 MiB/114.2 MiB]                                                
\
Operation completed over 1 objects/114.2 MiB.                                    
Copying gs://kubernetes-release/release/v1.24.17/bin/linux/amd64/kubectl...
/ [0 files][    0.0 B/ 44.5 MiB]                                                
/ [1 files][ 44.5 MiB/ 44.5 MiB]                                                
Operation completed over 1 objects/44.5 MiB.                                     
F1106 02:16:51.700377   13960 tester.go:505] failed to run ginkgo tester: unsupported ginkgo version: 1 
```

If we instead used the kops 1.27 branch for kubetest2-kops, we could keep it on an older kubetest2 dependency version that still supports ginkgo v1.

[...]

PR #16056, opened 2023-10-24
Enable containerd test-handler runtimeclass for k/k e2e tests

upodroid:
Some k/k e2e tests require the test-handler RuntimeClass to be configured.

https://testgrid.k8s.io/sig-cluster-lifecycle-kubeup-to-kops#ci-kubernetes-e2e-cos-gce-canary&include-filter-by-regex=RuntimeClass

https://testgrid.k8s.io/sig-cluster-lifecycle-kubeup-to-kops#ci-kubernetes-e2e-cos-gce-serial-canary&include-filter-by-regex=RuntimeClass

Tests got skipped because it is missing.

PR #16133, opened 2023-11-30
Only set containerd plugins on kops versions that include quote fixes

rifelpet:
https://github.com/kubernetes/kops/pull/16056 broke jobs that use any kops build that doesn't include https://github.com/kubernetes/kops/pull/16128 ([example](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-addon-resource-tracking/1729964344453632000)). Until we can implement https://github.com/kubernetes/kops/issues/16084 we can guard the new --set assignment behind a kops version check.

This will eval to true for all 1.29 builds. We dont have any jobs that test 1.29 prereleases (which dont include #16128) we only have jobs that test CI builds of the kops master branch, so any kops 1.29 job _will_ include the quote fix.

### DIFF
--- a/tests/e2e/kubetest2-kops/deployer/up.go
+++ b/tests/e2e/kubetest2-kops/deployer/up.go
@@ -130,8 +130,17 @@ func (d *deployer) createCluster(zones []string, adminAccess string, yes bool) e
 		"--kubernetes-version", d.KubernetesVersion,
 		"--ssh-public-key", d.SSHPublicKeyPath,
 		"--set", "cluster.spec.nodePortAccess=0.0.0.0/0",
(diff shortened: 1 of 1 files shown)


upodroid:
On that note, is it worth running the kops e2e tests from release branches to avoid these problems?

hakman:
> On that note, is it worth running the kops e2e tests from release branches to avoid these problems?

Release branches are unreliable, because artifacts get removed after 1-2 months.
We test upgrades with those releases, even if they are not supported anymore.

rifelpet:
🤞🏻 
/retest

PR #16139, opened 2023-11-30
WIP: Simple baremetal nodes

justinsb:
This is a cleaned-up and revamped version of experimental bare-metal support, whittled down to a handful of commits.

I plan on submitting these PRs individually, but wanted to give the "big picture" also.


docs/metal.md
```diff
@@ -0,0 +1,164 @@
+# Bare Metal Support
+
+***Bare metal support is experimental, and may be removed at any time***
+
+## Introduction
+
+kOps has some experimental bare-metal support, specifically for nodes.  The idea
```
zetaab, line 7: worth of mentioning that control-plane (and other nodes) should be with public topology? If the topology is private, its going to be difficult to have working cluster internal network

PR #16140, opened 2023-11-30
feat: Support multiple token verifiers in kops-controller

justinsb:
This will allow us to support nodes running in multiple clouds.  If we
don't configure multiple verifiers, this should be a no-op.

### DIFF
--- a/pkg/bootstrap/chain.go
+++ b/pkg/bootstrap/chain.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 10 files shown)


justinsb:
Breaking #16139 into more manageable chunks

pkg/bootstrap/chain.go
```diff
@@ -0,0 +1,50 @@
...
+		result, err := verifier.VerifyToken(ctx, rawRequest, token, body)
+		if err == nil {
+			return result, nil
+		}
+		if err == ErrNotThisVerifier {
+			continue
+		}
+		klog.Infof("failed to verify token: %v", err)
+	}
+	return nil, fmt.Errorf("unable to verify token")
```
rifelpet, line 49: I think this should wrap one of the non-nil errors that was received from the verifiers

PR #16141, opened 2023-11-30
Support bootstrapping nodes using Public Key authentication

justinsb:
- feat: Create CRD for Host
- feat: Support PKI bootstrap

We modify the GCE TPM support to support signing with a fixed private key.  The public key must be registered in a Host CRD instance (this will be done automatically by the `enroll` command).


k8s/crds/kops.k8s.io_hosts.yaml
```diff
@@ -0,0 +1,44 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1111
```
hakman, line 6:
```suggestion
    controller-gen.kubebuilder.io/version: v0.13.0
```

hakman (approved):

PR #16151, opened 2023-12-03
Don't set LimitNoFile for containerd systemd unit file

upodroid:
@dims 

Slack Thread for context: https://kubernetes.slack.com/archives/C09QZ4DQB/p1701357762061459

This is an upstream change in containerd that should fix NFS test issues on al2023

https://github.com/containerd/containerd/pull/8924


hakman (approved):

zetaab:
@hakman @dims @upodroid could you read https://github.com/kubernetes/kops/issues/16298

we are now hitting this in production. Why this change was made and how the production workloads should work with ulimit 1024?

PR #16156, opened 2023-12-07
Order instances in terms of cost

hakuna-matatah:
Order instances in terms of lowest to highest cost basis for cost efficiency - https://docs.aws.amazon.com/autoscaling/ec2/userguide/allocation-strategies.html#on-demand-allocation-strategy


hakuna-matatah:
Related issue - https://github.com/kubernetes/k8s.io/issues/6165

PR #16157, opened 2023-12-07
Reduce disk volume size to reduce cost

hakuna-matatah:
Related issue - https://github.com/kubernetes/k8s.io/issues/6165


hakuna-matatah:
@dims @hakman PTAL

hakman (approved):

dims:
/retitle Reduce disk volume size to reduce cost

PR #16162, opened 2023-12-07
Allow EKS Pod identity env variables

ameukam:
Allow env variables required by EKS Pod identity for authentification are passed to the kops binary.

See https://docs.aws.amazon.com/eks/latest/userguide/pod-id-how-it-works.html


ameukam:
cc @justinsb @rifelpet @hakman

hakman (approved):

PR #16176, opened 2023-12-15
test: Set num-nodes flag

upodroid:
There is a bug in the test framework where the -num-nodes flag must be explicitly set to avoid tests being skipped. It should be autocalculated but it is not working as intended.

image

Also, node-os-distro should be set to skip various tests that are OS specific in the kubernetes e2e test suite.

@dims

### DIFF
--- a/tests/e2e/pkg/tester/skip_regex.go
+++ b/tests/e2e/pkg/tester/skip_regex.go
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	skipRegexBase = "\\[Slow\\]|\\[Serial\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]|nfs|NFS|Gluster"
+	skipRegexBase = "\\[Slow\\]|\\[Serial\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]|nfs|NFS|Gluster|NodeProblemDetector"
 )
 
 func (t *Tester) setSkipRegexFlag() error {
@@ -81,6 +81,8 @@ func (t *Tester) setSkipRegexFlag() error {
 		skipRegex += "|same.hostPort.but.different.hostIP.and.protocol"
 		// https://github.com/cilium/cilium/issues/9207
 		skipRegex += "|serve.endpoints.on.same.port.and.different.protocols"
+		// https://github.com/kubernetes/kubernetes/blob/418ae605ec1b788d43bff7ac44af66d8b669b833/test/e2e/network/networking.go#L135
+		skipRegex += "|should.check.kube-proxy.urls"
 
 		if isPre28 {
 			// These may be fixed in Cilium 1.13 but skipping for now
--- a/tests/e2e/pkg/tester/tester.go
+++ b/tests/e2e/pkg/tester/tester.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/octago/sflags/gen/gpflag"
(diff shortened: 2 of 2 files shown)


upodroid:
/retest 

this is ready to be merged

tests/e2e/pkg/tester/tester.go
```diff
@@ -203,6 +199,22 @@ func (t *Tester) addNodeIG() error {
 			ig = v
 		}
 	}
+	numNodes := int(*ig.Spec.MaxSize) // we assume that MinSize = Maxsize, this is true for e2e testing
```
rifelpet, line 202: This will be the size of the last nodes InstanceGroup. Should we instead sum the max size of all nodes InstanceGroups?

upodroid: No, the ginkgo wants the number of all nodes except the master IG. When we are using kops for e2e testing we assume there are exactly 2 IGs(one master and one worker). If we add the master ig to the count, you'll have an extra node that you can't schedule workloads onto without tweaking manifests to tolerate control plane taints

rifelpet: Ah I was thinking in the case of there being multiple nodes IGs. This tester code can be used in cluster configurations that have multiple node IGs so I don't know if your assumption is always true.

tests/e2e/kubetest2-kops/deployer/up.go
```diff
@@ -130,6 +130,7 @@ func (d *deployer) createCluster(zones []string, adminAccess string, yes bool) e
 		"--kubernetes-version", d.KubernetesVersion,
 		"--ssh-public-key", d.SSHPublicKeyPath,
 		"--set", "cluster.spec.nodePortAccess=0.0.0.0/0",
+		"--set", "spec.nodeProblemDetector.enabled=true",
```
hakman, line 133: @upodroid Could we add a `Feature` tag on this, seems like a niche thing to test all the time.

hakman (changes requested):

tests/e2e/pkg/tester/skip_regex.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	skipRegexBase = "\\[Slow\\]|\\[Serial\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]|nfs|NFS|Gluster"
+	skipRegexBase = "\\[Slow\\]|\\[Serial\\]|\\[Disruptive\\]|\\[Flaky\\]|\\[Feature:.+\\]|nfs|NFS|Gluster|NodeProblemDetector"
```
upodroid, line 29: I tried to adjust the prow job generator and it caused a big mess. We run ubuntu by default so it would need to be running most of the time. It is easier to skip, will open a PR in k/k now to park this behind a Feature tag

upodroid:
/retest

We merged a PR in k/k to stop running NPD tests, the skip is needed for release branches

tests/e2e/pkg/tester/tester.go
```diff
@@ -203,6 +199,22 @@ func (t *Tester) addNodeIG() error {
...
+	klog.Infof("Setting -num-nodes=%v", numNodes)
+	t.TestArgs += " -num-nodes=" + strconv.Itoa(numNodes)
+
+	// Set the node-os-distro flag for ginkgo
+	distro := CalculateDistroFromIG(ig)
+	if distro != "" {
+		klog.Infof("Setting --node-os-distro=%s", distro)
+		t.TestArgs += " --node-os-distro=" + distro
```
hakman, line 210: Don't remember, did we say we will move this to the generator for now?

upodroid: Yes

upodroid: This is ready to merged

hakman (approved):

PR #16184, opened 2023-12-21
Enable pagination on GCE Instance Groups

rifelpet:
This is required to support MIGs with >1000 instances.

See https://cloud.google.com/compute/docs/instance-groups/add-remove-vms-in-mig#increase_the_groups_size_limit

Kops already uses pagination to get instances from instance groups so no other changes are necessary https://github.com/kubernetes/kops/blob/62fcb00f8c6ff3f86adbbaddf9cfe0dbf79c4498/upup/pkg/fi/cloudup/gce/compute.go#L712-L721


hakman (approved):

upodroid:
thank you

PR #16186, opened 2023-12-21
add support for devcontainer

remyleone:
Related to https://github.com/kubernetes/kubernetes/pull/121561


justinsb:
/retest

This can't be related, and I was just working on multiple branches and thought this would be a good thing to add!

justinsb:
There has been some skew in the spec it looks like, and we need to update to a newer version of go, but I think it would be good to get the ball rolling here.

/approve
/lgtm

PR #16188, opened 2023-12-21
skip_regex.go: kube-router add back in service afinity test

aauren:
@hakman 

Requires https://github.com/kubernetes/kubernetes/pull/122440 to be merged first.

Once those two are merged, kube-router should be passing both of the service affinity tests in the e2e suite, so we can allow those to run and no longer worry about failures.

### DIFF
--- a/tests/e2e/pkg/tester/skip_regex.go
+++ b/tests/e2e/pkg/tester/skip_regex.go
@@ -97,7 +97,7 @@ func (t *Tester) setSkipRegexFlag() error {
 			skipRegex += "|should.create.a.Pod.with.SCTP.HostPort"
 		}
 	} else if networking.KubeRouter != nil {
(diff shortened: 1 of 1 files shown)


aauren:
Closing this given the feedback on https://github.com/kubernetes/kubernetes/pull/122440

aauren:
Accidentally closed this one too quickly. This one can be merged now that https://github.com/kubernetes/kubernetes/pull/122440 has been merged.

rifelpet:
We should update this logic to only skip for k8s versions < 1.30. We wont see the impact of this skip removal until 1.30 is stable though because both the periodic and presubmit tests for kuberouter use the latest stable builds of k/k, not the CI builds.


We could also [switch them to use CI](https://github.com/kubernetes/test-infra/blob/b1eaeb8e5d77319242296ee9cbabdedd69f879b8/config/jobs/kubernetes/kops/build_jobs.py#L1287-L1288) until then.

hakman:
@rifelpet https://github.com/kubernetes/test-infra/pull/31511

PR #16193, opened 2023-12-28
Add cert-manager resource types to kubetest2-kops artifacts

rifelpet:
The LBC prow job is flaking. I believe its due to a race condition that still exists. I'm adding the cert-manager resources that the LBC addon includes to help with troubleshooting

https://testgrid.k8s.io/kops-misc#kops-aws-aws-load-balancer-controller


hakman:
@rifelpet We should investigate the dumper at some point. All the new namespaces and resource types are taking a very long time to be dumped.
/lgtm
/approve

PR #16196, opened 2023-12-31
Parallelize k8s resource dumps with kops toolbox dump

rifelpet:
This adds support for dumping k8s resources with `kops toolbox dump` rather than kubetest2-kops. A few changes compared to the kubetest2-kops logic:

* Resources are dumped in parallel with a worker pool. This should drastically speed up the resource dumps in scalability tests.
* The list of resource kinds to dump is sourced dynamically from the discovery client rather than relying on `kubectl cluster-info dump` + our own static list

One regression is that when the control plane isn't available (due to a kops bug, for example), kubetest2-kops will SSH into control plane nodes and run `kubectl cluster-info dump` on the control plane node and copy the artifacts back locally. `kops toolbox dump` doesn't support this yet. We haven't relied on that for troubleshooting recent prow job failures but I'll work on adding that support back next.

### DIFF
--- a/cmd/kops/toolbox_dump.go
+++ b/cmd/kops/toolbox_dump.go
@@ -56,6 +56,8 @@ var (
 	`))
 
 	toolboxDumpShort = i18n.T(`Dump cluster information`)
+
+	k8sResources = os.Getenv("KOPS_TOOLBOX_DUMP_K8S_RESOURCES")
 )
 
 type ToolboxDumpOptions struct {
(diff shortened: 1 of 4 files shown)


hakman (approved):
Very nice!
/cc @upodroid @hakuna-matatah

hakman:
@rifelpet Could you do something similar for dumping nodes? Should make dumping 500 nodes much faster.

PR #16198, opened 2024-01-01
Include pod logs in toolbox dump

rifelpet:
Followup to #16196 I forgot to include pod logs which were being handled by `kubectl cluster-info` ([example artifacts with missing logs](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kops/16181/presubmit-kops-gce-scale-calico-using-cl2/1741588176922742784/artifacts/cluster-info/kube-system/)). This adds back pod logs with a few changes:

* Separate log files per container rather than concatenating them all together in one `logs.txt` file
* Include logs from previous containers when they exist. This should help with troubleshooting cluster validation issues that cause controllers to restart.

### DIFF
--- a/pkg/dump/podlogs.go
+++ b/pkg/dump/podlogs.go
@@ -1,0 +1,162 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


pkg/dump/podlogs.go
```diff
@@ -0,0 +1,162 @@
...
+			errors.Join(dumpErr, result.err)
+		}
+	}
+	close(results)
+	return dumpErr
+}
+
+func (d *podLogDumper) getPodLogs(ctx context.Context, pods chan v1.Pod, results chan podLogDumpResult) {
+	for pod := range pods {
+		for _, container := range pod.Spec.Containers {
```
hakman, line 94: I think we should dump the init container logs also:
```suggestion
		for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
```

hakman (approved):

PR #16212, opened 2024-01-04
aws: Skip WarmPool checks when it's not enabled

hakman:
/cc @rifelpet @hakuna-matatah 

Ref: https://github.com/kubernetes/kops/pull/16172#issuecomment-1876191111

PR #16213, opened 2024-01-04
aws: Use instance metadata to get warm pool state

rifelpet:
This should eliminate some of the API throttling during scale tests and help with time for nodes to join the cluster

See the example userdata [here](https://docs.aws.amazon.com/autoscaling/ec2/userguide/tutorial-lifecycle-hook-instance-metadata.html#instance-metadata-create-hello-world-function) for usage of this instance metadata endpoint. I confirmed the 404 behavior works on non-ASG instances.

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -1116,6 +1116,9 @@ func addASLifecyclePolicies(p *Policy, enableHookSupport bool) {
 			"autoscaling:DescribeLifecycleHooks",
 		)
 	}
+	// TODO: remove this after k8s 1.29 support is removed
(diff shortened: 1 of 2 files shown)


rifelpet:
Apparently we dont have a presubmit job to test warmpool behavior, just [this](https://testgrid.k8s.io/kops-misc#kops-warm-pool) periodic. but that job should make it clear if this cause any regressions.

hakman:
@rifelpet Could you add the changes from https://github.com/kubernetes/kops/pull/16212 into this?

rifelpet:
> @rifelpet Could you add the changes from #16212 into this?

done

hakman (approved):

PR #16217, opened 2024-01-04
Allow override of the DNS domain used by the tests.

ameukam:
In order to move the kOps prowjobs to the new community infrastructure, we need to be able to use a custom domain for the required AWS public zone.


ameukam:
cc @rifelpet @hakman

Issue #16218, opened 2024-01-04
Problems creating first LoadBalancer in brand-new AWS account


justinsb:
/kind bug

@ameukam observed a problem creating services of type LoadBalancer in a brand new AWS account.

In https://storage.googleapis.com/kubernetes-jenkins/logs/ci-aws-kops-eks-pod-identity-sandbox/1742899966826778624/artifacts/cluster-info/kube-system/aws-cloud-controller-manager-sshfn/aws-cloud-controller-manager.log we found this error:

```
I0104 13:38:18.926043 1 aws_loadbalancer.go:1017] Creating load balancer for deployment-8855/test-rolling-update-with-lb with name: adaa9d6372b254401baef4dd8faeb0d4 E0104 13:38:19.807666 1 controller.go:298] error processing service deployment-8855/test-rolling-update-with-lb (retrying with exponential backoff): failed to ensure load balancer: AccessDenied: User: arn:aws:sts::808842816990:assumed-role/aws-cloud-controller-manager.kube-system.sa.e2e-kops-eks--9pmkdr/1704375497398309729 is not authorized to perform: iam:CreateServiceLinkedRole on resource: arn:aws:iam::808842816990:role/aws-service-role/elasticloadbalancing.amazonaws.com/AWSServiceRoleForElasticLoadBalancing because no identity-based policy allows the iam:CreateServiceLinkedRole action status code: 403, request id: 7edfd319-c55f-4381-9f12-32242008b8f7 

I0104 13:38:19.807812 1 event.go:307] "Event occurred" object="deployment-8855/test-rolling-update-with-lb" fieldPath="" kind="Service" apiVersion="v1" type="Warning" reason="SyncLoadBalancerFaile
```

(Specifically: "no identity-based policy allows the iam:CreateServiceLinkedRole action status code")

@rifelpet pointed out that _someone_ needs to call iam:CreateServiceLinkedRole before we can use ELB (and possibly other services).  That could be in our CCM permissions, we could do it from the kOps CLI, we could require users to do it manually.

@rifelpet also pointed out that CCM and LBC both document the permission as needed:

https://github.com/kubernetes/cloud-provider-aws/blob/65e4f1ac4dbed33744b1e935892417d5c71ae43c/docs/prerequisites.md?plain=1#L77

https://github.com/search?q=repo%3Akubernetes-sigs%2Faws-load-balancer-controller%20CreateServiceLinkedRole&type=code

ameukam:
[...]

PR #16219, opened 2024-01-04
Add permission needed for service-linked role creation

ameukam:
Attempting to fix:
  - https://github.com/kubernetes/kops/issues/16218

by adding the permission needed for the AWS CCM to create a service-linked role for the elastic lb service.


ameukam:
cc @rifelpet @hakman @justinsb

rifelpet (approved):
The upstream docs have this wildcard action so it should be fine to include here as a wildcard too.

https://github.com/kubernetes/cloud-provider-aws/blob/master/docs/prerequisites.md#iam-policies

PR #16222, opened 2024-01-05
Dont list and terminate ASG-managed EC2 instances

rifelpet:
/hold for feedback

Looking at [scale tests](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/16204/presubmit-kops-aws-scale-amazonvpc-using-cl2/1742962895290896384/build-log.txt), kops delete cluster spends 6 minutes trying to terminate all of the ec2 instances. In reality this shouldn't be necessary because kops also deletes the ASGs (which appear to happen much faster) and the ASGs will terminate the instances.

I dont know how the timing compares but it should at least eliminate the throttled API calls.

### DIFF
--- a/pkg/resources/aws/aws.go
+++ b/pkg/resources/aws/aws.go
@@ -63,7 +63,6 @@ func ListResourcesAWS(cloud awsup.AWSCloud, clusterInfo resources.ClusterInfo) (
 	listFunctions := []listFn{
 		// EC2
 		ListAutoScalingGroups,
-		ListInstances,
 		ListKeypairs,
 		ListSecurityGroups,
 		ListVolumes,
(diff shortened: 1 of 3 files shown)


hakman:
Both me and @upodroid thought about this and I think it's a good idea.
The problem is that other resources depend on instances, like volumes and would leave those resources hanging (maybe).

rifelpet:
> Both me and @upodroid thought about this and I think it's a good idea.
> 
> The problem is that other resources depend on instances, like volumes and would leave those resources hanging (maybe).

Could we have volumes depend on the ASGs instead? What I'm not sure about is if all instances are gone at the time that the ASG deletion returns, or whether AWS is still in the process of terminating them.

rifelpet:
Replaced by https://github.com/kubernetes/kops/pull/16229

PR #16229, opened 2024-01-06
aws: Skip deleting ASG instances without volumes

hakman:
/cc @rifelpet 

This seems to be slower though


rifelpet:
slower than trying to delete all ASG instances?

hakman:
> slower than trying to delete all ASG instances?

No idea on larger scale, but we can find out.

rifelpet (approved):

PR #16233, opened 2024-01-08
gce: Switch from using targetpools to backend services

upodroid:
Google recommends creating NLBs using Backend Services instead of TargetPools to take advantage of newer features

https://cloud.google.com/load-balancing/docs/network/networklb-target-pools

Also, this change "almost" supports global LBs(missing TCP target proxy resource)
Google splits some services in to regional vs global with identical object types. I fixed the methods to detect if a region is being supplied.

Why:
1. I want to try and see if global TCP Proxy LBs don't trigger ddos protection when running scale tests.
2. If Ddos protection kicks in, we can write a cloud-armor policy that allows 1k rps(10k reqs per 10s) from any IP.

### DIFF
--- a/cloudmock/gce/mockcompute/api.go
+++ b/cloudmock/gce/mockcompute/api.go
@@ -137,11 +137,11 @@ func (c *MockClient) HTTPHealthChecks() gce.HttpHealthChecksClient {
 	return c.httpHealthChecksClient
 }
 
-func (c *MockClient) RegionHealthChecks() gce.RegionHealthChecksClient {
+func (c *MockClient) HealthChecks() gce.HealthChecksClient {
 	return c.healthCheckClient
 }
 
-func (c *MockClient) RegionBackendServices() gce.RegionBackendServiceClient {
+func (c *MockClient) BackendServices() gce.BackendServiceClient {
 	return c.backendServiceClient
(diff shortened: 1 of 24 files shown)


justinsb:
Thanks for this @upodroid ... it looks like we do use backend services with internal load balancers.  I am proposing that we have internal load balancers for both api & kops-controller in all circumstances, so we should be using backend services (my motivation was the firewall rule bug).

I think the issue you hit was on the node/pod -> apiserver traffic maybe getting rate limited, so it might be good to validate that if/when we make that switch, that the rate limiting goes away.

That said, I don't oppose the idea of using backend services on the "user-facing" traffic also - the IPv6 support seems compelling in particular!  I think we should sequence this after the better internal LB support though, do you agree?

upodroid:
> That said, I don't oppose the idea of using backend services on the "user-facing" traffic also - the IPv6 support seems compelling in particular! I think we should sequence this after the better internal LB support though, do you agree?

Yes

I need to split this PR into smaller pieces
1. The gce client changes to support regional/global resources with a unified function
2. The LB changes

upodroid:
Hmm, I'll get the merge conflicts fixed and ship my open PRs in early June

PR #16237, opened 2024-01-09
aws: Use `domain` instead of `vpc` when rendering `aws_eip`

hakman:
```
/home/prow/go/src/k8s.io/kops/tests/integration/update_cluster/bastionadditional_user-data
╷
│ Warning: Argument is deprecated
│ 
│   with aws_eip.us-test-1a-bastionuserdata-example-com,
│   on kubernetes.tf line 414, in resource "aws_eip" "us-test-1a-bastionuserdata-example-com":
│  414:   vpc = true
│ 
│ use domain attribute instead
╵
Success! The configuration is valid, but there were some
validation warnings as shown above.
```
Ref: https://github.com/hashicorp/terraform-provider-aws/issues/26714

### DIFF
--- a/tests/integration/update_cluster/additionalobjects/kubernetes.tf
+++ b/tests/integration/update_cluster/additionalobjects/kubernetes.tf
@@ -980,7 +980,7 @@ terraform {
     aws = {
       "configuration_aliases" = [aws.files]
       "source"                = "hashicorp/aws"
-      "version"               = ">= 4.0.0"
+      "version"               = ">= 5.0.0"
     }
   }
 }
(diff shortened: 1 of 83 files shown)


rifelpet:
That PR was merged in the terraform-provider-aws version v5.0.0. I think we'll want to bump our minimum version too https://github.com/hakman/kops/blob/70bba2edfc45002e5c3d78f284f77f34b8f9ac53/upup/pkg/fi/cloudup/terraform/target_hcl2.go#L244

rifelpet (approved):

PR #16249, opened 2024-01-12
WIP: Integration branch for gce with dual load balancers

justinsb:
An integration branch to show how the pieces fit together / so we can test them together.

We add a second load balancer on GCE, and use that (fixed) address for dns=none, replacing the previous gcediscovery mechanism.

- Refactor ForAPIServer
- Prefer external endpoints when building kubeconfig
- Split internal and external kube-apiserver addresses
- gce: Always create an internal load balancer


upodroid:
FYI: https://github.com/kubernetes/kops/pull/16233 can you consider this change as well when reworking the GCE loadbalancers? Mine is ready(just missing the upgrade logic to replace the loadbalancer components on existing clusters)

justinsb:
This should be done in #16379 

@upodroid that PR should make node -> control-plane traffic internal, which I think is the goal of #16233 also?  I don't object to #16233 (particularly now we have pruning), but let's see if the scale tests are happier now that #16379 has landed.

/close

Issue #16064, opened 2023-10-26
Cilium , No IP's left to assign on aws instances


victorgitmain:
/kind bug
Running on AWS ec2 instances with cilium for network.
Sometimes newly created pods will get stuck indefinitely in "ContainerCreating" state.
When using: 
```
kubectl describe pod $POD_NAME
```
We get: 
```
  Warning  FailedCreatePodSandBox  2s    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "0f06846efa22ac25671ad6657f3540a4f4f9879c345b4ebeb7331769fffdd24f": plugin type="cilium-cni" name="cilium" failed (add): unable to allocate IP via local cilium agent: [POST /ipam][502] postIpamFailure 
```
after check we indeed see that the maximum number of pods has been scheduled
but because the instance still has resources left, pods still get scheduled on those instances 
and get stuck in "ContainerCreating"

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

1.
```
Client version: 1.28.0 (git-v1.28.0)
Last applied server version: 1.28.0
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
2.
```
1.27.6
```

**3. What cloud provider are you using?**
3.
```
AWS
```
**4. What commands did you run?  What is the simplest way to reproduce this issue?**
   Use AWS for cloud provider with networking set to Cilium 
   Use small instance type with low amount of "Maximum network interfaces"
   happened on "r5.large,m5.large" can happen on any node as long there is enough space to 
   schedule the pods.
   Create large amount of small pods and see that they get scheduled to Instances with no IPs 
   left.
   Can use following Yaml:
```
kubectl apply -f nginx.yaml
```
   ```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 200
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: public.ecr.aws/nginx/nginx:1.21
        ports:
        - containerPort: 80
   ```
**5. What happened after the commands executed?**
Pods got stuck in "ContainerCreating" state with
```
[...]

PR #16253, opened 2024-01-15
aws: fix maxPods when cilium ipam=eni is used

argusua:
**What does this PR introduce?**
This PR addresses the issue where kubelet is configured with maxPods 100 in the AWS environment and when cilium with ipam=eni is used.

**Why is this change needed?**
The number of IP addresses is connected to the AWS instance type and is much less than 100 which is the default maxPods value. So, for small instance types we in most cases end up with the pod not being able to schedule because of 'No IP's left to assign on aws instances'

**How does it address the issue?**
We use the same code to determine maxPods per AWS instance type as for amazonvpc networking.

**Related Issue**
Fixes https://github.com/kubernetes/kops/issues/16064

### DIFF
--- a/nodeup/pkg/model/kubelet.go
+++ b/nodeup/pkg/model/kubelet.go
@@ -634,7 +634,7 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 	c.ClientCAFile = filepath.Join(b.PathSrvKubernetes(), "ca.crt")
 
 	// Respect any MaxPods value the user sets explicitly.
-	if b.NodeupConfig.Networking.AmazonVPC != nil && c.MaxPods == nil {
+	if (b.NodeupConfig.Networking.AmazonVPC != nil || (b.NodeupConfig.Networking.Cilium != nil && b.NodeupConfig.Networking.Cilium.IPAM == kops.CiliumIpamEni)) && c.MaxPods == nil {
 		sess := session.Must(session.NewSession())
 		metadata := ec2metadata.New(sess)
 


hakman:
Hi @argusua. Thanks for taking the time to submit a PR.
Please sign the CLA to be able to accept it.

argusua:
> Hi @argusua. Thanks for taking the time to submit a PR. Please sign the CLA to be able to accept it.

Done.

argusua:
Can someone help with failed tests? Because for me it's not clear why they are failing. I've tried building and running nodeup on my test cluster with amazonvpc and cilium cni and it works correctly.

nodeup/pkg/model/kubelet.go
```diff
@@ -634,7 +634,7 @@ func (b *KubeletBuilder) buildKubeletConfigSpec() (*kops.KubeletConfigSpec, erro
 	c.ClientCAFile = filepath.Join(b.PathSrvKubernetes(), "ca.crt")
 
 	// Respect any MaxPods value the user sets explicitly.
-	if b.NodeupConfig.Networking.AmazonVPC != nil && c.MaxPods == nil {
+	if (b.NodeupConfig.Networking.AmazonVPC != nil || b.NodeupConfig.Networking.Cilium.IPAM == kops.CiliumIpamEni) && c.MaxPods == nil {
```
hakman, line 637: Probably this is why tests are failing.
```suggestion
	if (b.NodeupConfig.Networking.AmazonVPC != nil || (b.NodeupConfig.Networking.Cilium != nil && b.NodeupConfig.Networking.Cilium.IPAM == kops.CiliumIpamEni)) && c.MaxPods == nil {
```

hakman (changes requested):

argusua:
Thanks, @hakman It was also my thought but not clear why I can't recreate this on my local cluster. Anyway, I've commited your suggestion. Let's if this helps.

argusua:
Hi, @hakman Was able to recreate the issue locally too, and your suggestion fixed it. Also, looks like tests here is green now. Thanks.

argusua:
Hi, @hakman 
Done

hakman (approved):

PR #16181, opened 2023-12-20
Add GCE scale testing on kops

upodroid:
Part of https://github.com/kubernetes/enhancements/issues/4224
Fixes: https://github.com/kubernetes/kops/issues/16241

kops will be replacing the current kubeup 5k scale test, which will be developed here

PR #16266, opened 2024-01-20
Fix dumping logs for GCE scale tests

upodroid:
Splitting https://github.com/kubernetes/kops/pull/16181 to separate PRs

1. When a bastion is being used, kops tries to use it by its hostname which doesn't work with `--dns=none`. Using the IP address is much easier
2. we also dump the bastion IP in the logs
3. `--max-nodes` flag is now respected by the DumpAllNodes function.
4. kubetest2 now sets the `--ssh-private-key` automatically when running kops gce clusters, this flag must be set for the `kops toolbox dump` cmd to work
5. We calculate the correct internal IP addresses of GCE instances to use with bastions.

### DIFF
--- a/cmd/kops/toolbox_dump.go
+++ b/cmd/kops/toolbox_dump.go
@@ -210,7 +210,14 @@ func RunToolboxDump(ctx context.Context, f commandutils.Factory, out io.Writer, 
 			return fmt.Errorf("adding key to SSH agent: %w", err)
 		}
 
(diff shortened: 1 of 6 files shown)


hakman:
@upodroid Please explain in the description what issue you are trying to fix here and how.

upodroid:
> Please explain in the description what issue you are trying to fix here and how.

Done

hakman (approved):

PR #16269, opened 2024-01-22
gce: Change default storage class to balanced-csi

sl1pm4t:
This change updates the default `StorageClass` to be `balanced-csi` on GCP clusters. 

Balanced persistent disk type is the default set on GKE clusters.

I've updated the release notes with the breaking change, as discussed in the office hours call last year. LMK if there is anywhere else that needs updating.


justinsb:
No time like the present!

/approve
/lgtm

PR #16272, opened 2024-01-23
gce: Set node IP Alias range to match NodeCIDRMaskSize

sl1pm4t:
Prior to this change, GCE Node IP Alias CIDR range was hardcoded to be a `/24` which can be a limiting factor for scaling out a cluster when using smaller instance types, and could cause IP pool exhaustion despite many unallocated IPs.

This also means it's possible to bring up a cluster where the actual node CIDR size disagrees with what is configured in `cluster.spec.kubeControllerManager.NodeCIDRMaskSize` - although I do not know how detrimental that would be.

This change updates the IP Alias range to match the value configured in `cluster.spec.kubeControllerManager.NodeCIDRMaskSize`.

### DIFF
--- a/pkg/model/gcemodel/autoscalinggroup.go
+++ b/pkg/model/gcemodel/autoscalinggroup.go
@@ -190,8 +190,12 @@ func (b *AutoscalingGroupModelBuilder) buildInstanceTemplate(c *fi.CloudupModelB
 			if gce.UsesIPAliases(b.Cluster) {
 				t.CanIPForward = fi.PtrTo(false)
 
+				nodeCIDRMaskSize := int32(24)
+				if b.Cluster.Spec.KubeControllerManager.NodeCIDRMaskSize != nil {
+					nodeCIDRMaskSize = *b.Cluster.Spec.KubeControllerManager.NodeCIDRMaskSize
+				}
 				t.AliasIPRanges = map[string]string{
-					b.NameForIPAliasRange("pods"): "/24",
+					b.NameForIPAliasRange("pods"): fmt.Sprintf("/%d", nodeCIDRMaskSize),
 				}
 			} else {
 				t.CanIPForward = fi.PtrTo(true)


s3than:
Thanks for this @sl1pm4t

justinsb:
> This also means it's possible to bring up a cluster where the actual node CIDR size disagrees with what is configured in cluster.spec.kubeControllerManager.NodeCIDRMaskSize - although I do not know how detrimental that would be.

It shouldn't matter, I think, because when we're using GCE networking (aka IP Alias), we bring up instances and GCE allocates the CIDR which becomes the node's pod cidr.

I think `cluster.spec.kubeControllerManager.NodeCIDRMaskSize` field is unused in GCE networking mode.  We also normally suggest that the fields under kubeControllerManager are more for power-users than for things that we expect users to set directly, the reason being that these fields map to flags, and kubernetes doesn't have the same compatibility guarantees around flags - flags are often deprecated.

What we would normally suggest is creating a top level field.    In this case it would be something like "NodeCIDRMaskSize" under [GCPNetworkingSpec](https://github.com/kubernetes/kops/blob/8f36682a3550f7cb21e906b9f740724d08b7ade4/pkg/apis/kops/v1alpha2/networking.go#L669).  Would that make more sense here?  (Or any other field name that makes sense to you!)

If not, I'm aware this PR has been sitting for a long time (sorry!), so I think we could just merge as-is and then have `spec.gce.nodeCIDRMaskSize` take precedence if set when/if we add that field.

justinsb:
We're coming up to the cutoff point for 1.29.  I think we could merge this as-is, though if you're able to add another field that would probably be better in the long term.  But the reason I think we could merge this as-is because the current field (in this PR) is currently unused anyway, so I think we could just slot in the new field with higher priority if/when we add it

justinsb:
We discussed a bit in office hours, and just merging this as-is is probably the best approach; it doesn't block adding the "more correct" field and the current version should be harmless.

upodroid:
+1

PR #16277, opened 2024-01-24
Add support to configure HPA Controller concurrent syncs flag in HPA/KCM Controller

hakuna-matatah:
Add support to configure HPA Controller concurrent syncs flag in HPA/KCM Controller
This value requires to be tweaked to get more performance out of HPA controller on large scale clusters using HPAs.
Plan to leverage this in scale tests in future.

### fixes 
https://github.com/kubernetes/kops/issues/16278

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -2147,6 +2147,11 @@ spec:
                       sync concurrently.
                     format: int32
                     type: integer
+                  concurrentHorizontalPodAustoscalerSyncs:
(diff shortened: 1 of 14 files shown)


hakuna-matatah:
/cc @hakman

PTAL

mengqiy:
@hakuna-matatah Can you please also add job-syncs here?

hakuna-matatah:
> @hakuna-matatah Can you please also add job-syncs here?

Will do that in the next PR ? Do we also need `cron-job-syncs` or just `job-syncs` is enough ?

hakman (approved):

PR #16282, opened 2024-01-26
Split internal and external kube-apiserver addresses

justinsb:
This allows us to select the right IP address when we have multiple
load balancers (one for internal, one for external) as we are planning
to have on GCE.

### DIFF
--- a/pkg/model/hetznermodel/loadbalancer.go
+++ b/pkg/model/hetznermodel/loadbalancer.go
@@ -65,7 +65,11 @@ func (b *LoadBalancerModelBuilder) Build(c *fi.CloudupModelBuilderContext) error
 			hetzner.TagKubernetesClusterName: b.ClusterName(),
 		},
 
-		WellKnownServices: []wellknownservices.WellKnownService{wellknownservices.KubeAPIServer, wellknownservices.KopsController},
+		WellKnownServices: []wellknownservices.WellKnownService{
+			wellknownservices.KopsControllerInternal,
+			wellknownservices.KubeAPIServerExternal,
+			wellknownservices.KubeAPIServerInternal,
+		},
 	}
 
 	c.AddTask(&loadbalancer)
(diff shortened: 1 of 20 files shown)


pkg/model/hetznermodel/loadbalancer.go
```diff
@@ -65,7 +65,11 @@ func (b *LoadBalancerModelBuilder) Build(c *fi.CloudupModelBuilderContext) error
 			hetzner.TagKubernetesClusterName: b.ClusterName(),
 		},
 
-		WellKnownServices: []wellknownservices.WellKnownService{wellknownservices.KubeAPIServer, wellknownservices.KopsController},
+		WellKnownServices: []wellknownservices.WellKnownService{
+			wellknownservices.KopsController,
```
hakman, line 69: Could we rename this to make it easy to understand that it will be always internal?
```suggestion
			wellknownservices.KopsControllerInternal,
```

hakman (changes requested):

pkg/model/hetznermodel/loadbalancer.go
justinsb: Sure!  I do want to check if it (will be) publicly exposed on any clouds though, as that might be more confusing!

hakman: It probably will be exposed on some clouds so we may need to add that too and use it when required.

justinsb:
Renamed to KopsControllerInternal, and also expanded the comments for the enum values (in pkg/wellknownservices/wellknownservices.go)

I'm using "Internal" for in-cluster traffic, and "External" for other traffic.  InCluster might be clearer than Internal, but I don't know what a better name would be for External...

justinsb:
WDYT @hakman ... I can probably make #16249 work in a different way, but right now that PR depends on this one (and I think that one is comparatively more important!)

justinsb:
The approach we took in #16379 replaces the need for this (at least that's the belief!)

/close

PR #16291, opened 2024-01-29
WIP: Introduce cleanup phase

justinsb:
A large "integration branch" PR to show the big picture; I will whittle this down with smaller PRs but want to show & agree the direction.

Because of the challenges of adding security groups to an existing NLB - see issue #16276 - the best way around this seems to be to create a second NLB during `kops update`, do a `kops rolling-update` (the old nodes use the old NLB, the new nodes use the new NLB), then do a cleanup phase where we delete the original (now unused) NLB and associated resources.

Even with this, as @zetaab pointed out the kubeconfig DNS name will change if we're using the ELB, though this "just" means that users will need to distribute a new kubeconfig.  Users using a real domain will not need to distribute a new kubeconfig.

Generally, the trick here is to introduce the idea of "revisions" of a Task or Cloud Resource.  Two NLBs / TargetGroups / etc might have the same Name tag on AWS, but would have a different revision tag (`kops.k8s.io/revision`).

Another trick we use is to stash more state in private fields in the Task; we already have a "linking" phase such that if you refer to a Task with its "ID" (as defined by CompareWithID) that gets replaced with the canonical Task object.

### DIFF
--- a/cmd/kops/update_cluster.go
+++ b/cmd/kops/update_cluster.go
@@ -80,6 +80,8 @@ type UpdateClusterOptions struct {
 	// LifecycleOverrides is a slice of taskName=lifecycle name values.  This slice is used
 	// to populate the LifecycleOverrides struct member in ApplyClusterCmd struct.
 	LifecycleOverrides []string
+
+	DoDeferredDeletions bool
 }
 
 func (o *UpdateClusterOptions) InitDefaults() {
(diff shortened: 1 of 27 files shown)


justinsb:
This is getting relatively minimal I think, we do need #16293 to go in to remove a lot of the noise, but it's not obvious to me how to split what's left into any more standalone parts.

(I do need to clean up the commits though, they are a mess on this branch!)

justinsb:
This was done in #16356 (and other earlier PRs), closing

/close

PR #16294, opened 2024-01-29
refactor: wait for load balancer readiness using a private field

justinsb:
This approach is more explicit than looking at the names of the target
groups, and using a private field is simpler.

### DIFF
--- a/pkg/model/awsmodel/api_loadbalancer.go
+++ b/pkg/model/awsmodel/api_loadbalancer.go
@@ -193,6 +193,12 @@ func (b *APILoadBalancerBuilder) Build(c *fi.CloudupModelBuilderContext) error {
 			Type:              fi.PtrTo("network"),
 		}
 
+		// Wait for all load balancer components to be created (including network interfaces needed for NoneDNS).
+		// Limiting this to clusters using NoneDNS because load balancer creation is quite slow.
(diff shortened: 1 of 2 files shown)


justinsb:
Looks like a known flake https://github.com/kubernetes/kubernetes/issues/122333

I was worried tht if this was the first test to run, it might be an error with the "wait for ready" logic.  But it wasn't, it was about 8 minutes into the test run.

/retest

hakman:
@justinsb Wouldn't it be simpler to just always wait for NLB?
It was an optimisation, but we can live without it.

justinsb:
> Wouldn't it be simpler to just always wait for NLB?
> It was an optimisation, but we can live without it.

I'm fine with that if that's what we want to do.  It does take a while, but I figure people are going to be waiting anyway...

The important end-goal here is to get rid of the TargetGroups field (notice they aren't really used once we split out listeners!), so we can version them based on the LoadBalancer version.  I also wanted to experiment with the idea of private fields for internal options, they don't get diffed etc so they're a little easier to work with (no need to sync actual & expected, won't show up in dry-run output).  Because they're private we do have to set them via a method but ... that seems like a good trade-off.

In this particular case though, I don't mind either way - as long as we stop using TargetGroups!

hakman (approved):

PR #16334, opened 2024-02-10
Set KUBECONFIG for LBC's ginkgo tests

rifelpet:
Hoping to fix these tests: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-aws-load-balancer-controller/1755828286962798592

The test failures seem to be using the wrong kubeconfig settings, potentially using the "in cluster" config of the prow job pod itself.

This explicitly sets KUBECONFIG so that the ginkgo tests will use the kops cluster's admin credentials 

ref: https://kubernetes.slack.com/archives/C8MKE2G5P/p1707502477184889?thread_ts=1707242089.155659&cid=C8MKE2G5P

### DIFF
--- a/tests/e2e/scenarios/aws-lb-controller/run-test.sh
+++ b/tests/e2e/scenarios/aws-lb-controller/run-test.sh
@@ -33,6 +33,10 @@ ZONES="eu-west-1a,eu-west-1b,eu-west-1c"
 
 kops-up
 
+KUBECONFIG=$(mktemp -t kops.XXXXXXXXX)
+export KUBECONFIG
+"${KOPS}" export kubecfg --name "${CLUSTER_NAME}" --admin --kubeconfig "${KUBECONFIG}"
+
 VPC=$(${KOPS} toolbox dump -o json | jq -r .vpc.id)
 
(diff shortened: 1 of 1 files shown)


tests/e2e/scenarios/aws-lb-controller/run-test.sh
```diff
@@ -33,6 +33,9 @@ ZONES="eu-west-1a,eu-west-1b,eu-west-1c"
 
 kops-up
 
+export KUBECONFIG=$(mktemp -t kops.XXXXXXXXX)
```
hakman, line 36: https://www.shellcheck.net/wiki/SC2155
```suggestion
KUBECONFIG=$(mktemp -t kops.XXXXXXXXX)
export KUBECONFIG
```

hakman: Do we need to export?

rifelpet: I believe so because the ginkgo process is what is loading the client-go config and reading from the KUBECONFIG env var, not just a shell variable.

hakman (approved):

PR #16339, opened 2024-02-10
target group: refactor discovery into awsup

justinsb:
This sets us up to support multiple generations of target groups,
needed if we want to support adding SecurityGroups to the NLB.

### DIFF
--- a/upup/pkg/fi/cloudup/awsup/elbv2_targetgroups.go
+++ b/upup/pkg/fi/cloudup/awsup/elbv2_targetgroups.go
@@ -26,25 +26,25 @@ import (
 	"k8s.io/klog/v2"
 )
 
-type LoadBalancerInfo struct {
-	LoadBalancer *elbv2.LoadBalancer
-	Tags         []*elbv2.Tag
-	arn          string
+type TargetGroupInfo struct {
+	TargetGroup *elbv2.TargetGroup
+	Tags        []*elbv2.Tag
+	arn         string
 }
 
 // ARN returns the ARN of the load balancer.
-func (i *LoadBalancerInfo) ARN() string {
+func (i *TargetGroupInfo) ARN() string {
 	return i.arn
 }
 
 // NameTag returns the value of the tag with the key "Name".
-func (i *LoadBalancerInfo) NameTag() string {
+func (i *TargetGroupInfo) NameTag() string {
 	s, _ := i.GetTag("Name")
 	return s
 }
 
 // GetTag returns the value of the tag with the given key.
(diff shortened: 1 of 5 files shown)


hakman (approved):

upup/pkg/fi/cloudup/awsup/elbv2_targetgroups.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+type TargetGroupInfo struct {
+	TargetGroup *elbv2.TargetGroup
+	Tags        []*elbv2.Tag
+}
+
+// ARN returns the ARN of the load balancer.
+func (i *TargetGroupInfo) ARN() string {
+	return aws.StringValue(i.TargetGroup.TargetGroupArn)
```
rifelpet, line 36: perhaps return an empty string if `i.TargetGroup` is nil rather than panic

justinsb: aws.StringValue does this :-)  It's the equivalent of our fi.ValueOf (and previously fi.StringValue).  I try to use the aws functions when dealing with aws inputs, and the fi functions when using with our own, though the function definitions are identical (modulo generics)

justinsb:

Looks unrelated to this PR, but an odd failure, looks like the load balancer was not ready.  I will double-check whether I removed some important "wait for ready" logic in the reshuffling

upup/pkg/fi/cloudup/awsup/elbv2_targetgroups.go
rifelpet: Accessing TargetGroupArn if TargetGroup is nil will panic before even entering the StringValue function though, right?

justinsb: Doh, you're right, sorry!

justinsb:
So I was not imagining that we would build these `*Info` objects without setting the core object they wrap.  I don't know whether the right way to express that is just use a struct or whether to make the wrapped-object private.

I side-stepped the issue for now by storing an `arn string` field - we have the arn anyway during construction.

I have noticed that these are looking a lot like our fi.Resource types, that might be a productive refactoring direction: deferred-deletion is a limited form of cluster deletion - both are about cluster resource-discovery and pruning.

justinsb:
And thanks for spotting it @rifelpet!  It may be safe today, but we don't want to lay traps for our future selves :-)

rifelpet (approved):

Issue #16336, opened 2024-02-10
Private dns=none clusters incorrectly creating bastion DNS name tasks


rifelpet:
/kind bug
/kind failing-test

https://testgrid.k8s.io/kops-misc#kops-aws-private

kops is trying to create a DNS record for the bastion hosts:

```
 I0208 03:29:19.850427   14272 loader.go:90] Known tasks:
...
I0208 03:29:19.850581   14272 loader.go:92]   DNSName/bastion.e2e-e2e-kops-aws-private.test-cncf-aws.k8s.io
I0208 03:29:19.850589   14272 loader.go:92]   DNSName/bastion.e2e-e2e-kops-aws-private.test-cncf-aws.k8s.io-AAAA 
```

which then fails because the DNSZone task hasn't been defined, because the cluster uses dns=none:

`Error: error building tasks: unexpected error resolving task "DNSName/bastion.e2e-e2e-kops-aws-private.test-cncf-aws.k8s.io": unable to find task "DNSZone/", referenced from DNSName/bastion.e2e-e2e-kops-aws-private.test-cncf-aws.k8s.io:Zone`

The [cluster.yaml](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-private/1755432422633115648/artifacts/cluster.yaml) shows a bastion's public name field is set:

```yaml
  topology:
    bastion:
      bastionPublicName: bastion.e2e-e2e-kops-aws-private.test-cncf-aws.k8s.io
    dns:
      type: None
```

which is what creates the bastion DNSName tasks:

https://github.com/kubernetes/kops/blob/b73a2dd7aa0b0b5e69f7bda265bdd8b004d33510/pkg/model/awsmodel/bastion.go#L399-L407

what I dont understand is how that field is being set in the cluster spec, because new_cluster.go only sets it if we're publishing DNS records:

https://github.com/kubernetes/kops/blob/b73a2dd7aa0b0b5e69f7bda265bdd8b004d33510/upup/pkg/fi/cloudup/new_cluster.go#L1364-L1368

https://github.com/kubernetes/kops/blob/446f3acaf177d9257bd8a1c8710d20085fdb4977/pkg/apis/kops/cluster.go#L893-L898

If we can have kops stop populating that cluster spec field for dns=none clusters, the DNSName tasks should stop being added.

PR #16342, opened 2024-02-10
Move DNS topology setup earlier in cluster creation

rifelpet:
Fixes https://github.com/kubernetes/kops/issues/16336

This is needed because setting the bastion public name field depends on the DNS topology.
We were incorrectly setting bastion.publicName for dns=none clusters because the dns=none field wasn't yet set on the cluster.


justinsb:
This LGTM - thanks for fixing!

/approve
/lgtm

Issue #16340, opened 2024-02-10
Treatment of overlapping ServiceCIDR and PodCIDRs


justinsb:
We may have introduced a regression, where previously overlapping ServiceCIDR and PodCIDRs were allowed, and now they are blocked by validation.

It does seem that the node-podCIDR allocation logic in kube-controller-manager automatically excludes the ServiceCIDR, so that an overlap should not give a conflict in practice (though we probably still want to discourage it).  For example:
 https://github.com/kubernetes/kubernetes/blob/98b589a76d57a137d91afc487304d0f699c23288/pkg/controller/nodeipam/ipam/range_allocator.go#L103

/kind bug

(Edited to change "should yield a conflict" to "should not give a conflict" - doh!)

minkimipt:
Thanks @justinsb for creating the issue.

PR #16344, opened 2024-02-10
validation: Allow overlap of pod/node CIDR and service CIDR

justinsb:
We allowed this previously, so this is a regression for existing clusters.

These clusters are not obviously broken, and the
kube-controller-manager (for example) will exclude the service range
when issuing node CIDRs.  As such, remove validation until we can
determine if anything is actually broken by an overlap (and a path
forwards if so).

Issue #16340


rifelpet (approved):
/retest

PR #16346, opened 2024-02-11
Update Go to v1.22.0

hakman:


rifelpet:
k/k's `release-1.29` is still on go 1.21: https://github.com/kubernetes/kubernetes/blob/release-1.29/.go-version
should we wait until we create kops' release-1.29 branch so that this lands in kops 1.30?

hakman:
> k/k's `release-1.29` is still on go 1.21: https://github.com/kubernetes/kubernetes/blob/release-1.29/.go-version should we wait until we create kops' release-1.29 branch so that this lands in kops 1.30?

k/k will backport go updates to all supported branches, so nothing to worry about.

PR #16335, opened 2024-02-10
Have kops-controller create headless k8s services for dns=none clusters 

rifelpet:
Fixes #16332

Like gossip clusters, dns=none clusters wont be able to resolve these DNS names which causes e2e test failures for OIDC tests. This ensures the OIDC test pods can resolve the DNS names.

Example failure: https://testgrid.k8s.io/kops-grid#kops-grid-calico-amzn2-k26

`I0204 15:43:18.460415       1 log.go:198] Get "https://api.internal.e2e-e2e-kops-grid-calico-amzn2-k26.test-cncf-aws.k8s.io/.well-known/openid-configuration": dial tcp: lookup api.internal.e2e-e2e-kops-grid-calico-amzn2-k26.test-cncf-aws.k8s.io on 100.64.0.10:53: no such host`

PR #16347, opened 2024-02-11
Include /etc/hosts coredns mounts for dns=none clusters

rifelpet:
Ref: https://github.com/kubernetes/kops/pull/16335#issuecomment-1937413388

### DIFF
--- a/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
+++ b/upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
@@ -76,8 +76,8 @@ data:
           fallthrough in-addr.arpa ip6.arpa
           ttl 30
         }
-    {{- if GossipName }}
-        hosts /rootfs/etc/hosts k8s.local {
+    {{- if not PublishesDNSRecords }}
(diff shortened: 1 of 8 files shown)


rifelpet:
cc @hakman 

You're right, this is more relevant to the failing tests. I'm not sure what the headless services in https://github.com/kubernetes/kops/pull/16335 are actually used for. Do you happen to know?

upup/models/cloudup/resources/addons/coredns.addons.k8s.io/k8s-1.12.yaml.template
```diff
@@ -76,7 +76,7 @@ data:
           fallthrough in-addr.arpa ip6.arpa
           ttl 30
         }
-    {{- if GossipName }}
+    {{- if not PublishesDNSRecords }}
         hosts /rootfs/etc/hosts k8s.local {
```
rifelpet, line 80: /hold Actually this needs to be updated to the cluster's domain

rifelpet: /hold cancel

hakman (approved):

rifelpet:
Confirmed this fixed the e2e failures: https://testgrid.k8s.io/kops-grid#kops-grid-calico-flatcar-k28

PR #16354, opened 2024-02-15
Allow scaling events to be logged during cluster validation

rifelpet:
The [e2e upgrade jobs](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-upgrade-k127-ko127-to-k128-ko128-many-addons/1755930711879061504) that have been migrated to the new prow cluster are failing to validate mid rolling-update

`I0209 13:36:37.740937   14123 instancegroups.go:560] Cluster did not pass validation within deadline: InstanceGroup "nodes-us-west-2a" did not have enough nodes 1 vs 4.`

We can get scaling activities on the ASG which should mention if the AWS autoscaling service is failing to launch nodes for some reason (resource quota, capacity, etc.)

This allows those events to be logged at `--v=4` level, and sets that level on the upgrade scripts.

I included autoscaling activity for both AWS and GCE. other providers can add their implementations separately.

### DIFF
--- a/upup/pkg/fi/cloudup/awsup/aws_cloud.go
+++ b/upup/pkg/fi/cloudup/awsup/aws_cloud.go
@@ -1196,6 +1196,21 @@ func awsBuildCloudInstanceGroup(ctx context.Context, c AWSCloud, cluster *kops.C
 		return nil, fmt.Errorf("failed to fetch instances: %v", err)
 	}
 
+	scalingReq := &autoscaling.DescribeScalingActivitiesInput{
+		AutoScalingGroupName: g.AutoScalingGroupName,
+	}
+	scalingEvents := make([]cloudinstances.ScalingEvent, 0)
+	c.Autoscaling().DescribeScalingActivitiesPagesWithContext(ctx, scalingReq, func(p *autoscaling.DescribeScalingActivitiesOutput, lastPage bool) bool {
+		for _, activity := range p.Activities {
+			event := cloudinstances.ScalingEvent{
+				Timestamp:   aws.TimeValue(activity.StartTime),
+				Description: aws.StringValue(activity.Description),
+			}
+			scalingEvents = append(scalingEvents, event)
+		}
+		return true
+	})
+
 	cg := &cloudinstances.CloudInstanceGroup{
 		HumanName:     aws.StringValue(g.AutoScalingGroupName),
 		InstanceGroup: ig,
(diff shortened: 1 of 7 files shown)


rifelpet:
trying this out:

/test pull-kops-e2e-aws-upgrade-k127-ko127-to-klatest-kolatest-many-addons

though this job sounds invalid, since a k8s 1.27 cluster can't be upgraded to k8s latest (1.29)

justinsb:
I like the idea!

> though this job sounds invalid, since a k8s 1.27 cluster can't be upgraded to k8s latest (1.29)

(Do we know that it can't?  It's not considered a supported path, but in practice most skip-upgrades do work)

upup/pkg/fi/cloudup/gce/compute.go
```diff
@@ -709,6 +710,17 @@ func (c *instanceGroupManagerClientImpl) List(ctx context.Context, project, zone
 	return ms, nil
 }
 
+func (c *instanceGroupManagerClientImpl) ListErrors(ctx context.Context, project, zone, name string) ([]*compute.InstanceManagedByIgmError, error) {
```
justinsb, line 713: I sort of regret our having this layer in GCE, I'm not sure it adds much!

upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -1196,13 +1196,29 @@ func awsBuildCloudInstanceGroup(ctx context.Context, c AWSCloud, cluster *kops.C
 		return nil, fmt.Errorf("failed to fetch instances: %v", err)
 	}
 
+	scalingReq := &autoscaling.DescribeScalingActivitiesInput{
```
justinsb, line 1199: I don't know whether this gets expensive, but I would if we should plumb through a flag here as to whether we want this information (or add a method or func to CloudInstanceGroup that could query it on-demand)

upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -1196,13 +1196,29 @@ func awsBuildCloudInstanceGroup(ctx context.Context, c AWSCloud, cluster *kops.C
...
+		AutoScalingGroupName: g.AutoScalingGroupName,
+	}
+	scalingEvents := make([]cloudinstances.ScalingEvent, 0)
+	c.Autoscaling().DescribeScalingActivitiesPagesWithContext(ctx, scalingReq, func(p *autoscaling.DescribeScalingActivitiesOutput, lastPage bool) bool {
+		for _, activity := range p.Activities {
+			event := cloudinstances.ScalingEvent{
+				Timestamp:   aws.TimeValue(activity.StartTime),
+				Description: aws.StringValue(activity.Description),
```
justinsb, line 1207: Looks like there's some extra fields; I'm not sure if Description has all/most of the information, but you might consider including the `Raw any` field (which will then get printed!)

justinsb (commented):
Cool idea!

PR #16050, opened 2023-10-23
Add support for publishing Service Account Issuer documents to GCS.

sl1pm4t:
This change adds for support for GCS as a target for `cluster.spec.serviceAccountIssuerDiscovery.discoveryStore`.

The generated the service account issuer URL is in the form:
`https://storage.googleapis.com/<BUCKET>` 
meaning the JWKS URI becomes:
`https://storage.googleapis.com/<BUCKET>/openid/v1/jwks`

This is the first step towards supporting Workload Identity on GCP kOps clusters.

PR #16355, opened 2024-02-15
gce: Update GCE storage service scope to DevstorageFullControlScope to resolve permission error.

sl1pm4t:
This is a fix for the functionality added in https://github.com/kubernetes/kops/pull/16050.

In some cases, kops was unable to query the GCP API to retrieve the bucket IAM policy. 
The API would return a 403 Forbidden. 
e.g.
```
Error: error building tasks: checking if bucket was public: googleapi: Error 403: Access denied., forbidden
```

This PR increases the auth scopes that the GCS client is created with to resolve the permission problem

Strangely this 403 error did not happen until kops was authenticating to GCS using a Service Account created in a different GCP project than where the OIDC Issuer bucket was created.

### DIFF
--- a/util/pkg/vfs/context.go
+++ b/util/pkg/vfs/context.go
@@ -418,7 +418,7 @@ func (c *VFSContext) getGCSClient(ctx context.Context) (*storage.Service, error)
 	}
 
 	// TODO: Should we fall back to read-only?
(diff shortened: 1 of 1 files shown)


justinsb:
Thank you for the fix; @sl1pm4t 

/approve
/lgtm

PR #16356, opened 2024-02-15
Generate revisions of NLB objects, and introduce cleanup phase

justinsb:
This lets us safely make changes to otherwise immutable fields, in
particular for adding security groups to NLBs created without them.

We detect the older versions, and create deletion tasks to remove
them.  These tasks can be deferred, and we expect them to be
deferred to a "prune" phase that runs after cluster apply.

### DIFF
--- a/cmd/kops/update_cluster.go
+++ b/cmd/kops/update_cluster.go
@@ -80,6 +80,12 @@ type UpdateClusterOptions struct {
 	// LifecycleOverrides is a slice of taskName=lifecycle name values.  This slice is used
 	// to populate the LifecycleOverrides struct member in ApplyClusterCmd struct.
 	LifecycleOverrides []string
+
+	// Prune is true if we should clean up any old revisions of objects.
+	// Typically this is done in after we have rolling-updated the cluster.
+	// The goal is that the cluster can keep running even during more disruptive
+	// infrastructure changes.
+	Prune bool
 }
 
 func (o *UpdateClusterOptions) InitDefaults() {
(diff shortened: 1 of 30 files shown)


justinsb:
This is the remaining work from #16291; I think it makes sense to consider as a single PR.  (I sent a new PR to keep the PR history understandable, even though they should now be the same!)

cc @hakman @rifelpet

justinsb:
One possible thing I've realized I could do ... I could split out the deletion tasks/printing/processing into a separate PR, and then only really use them in this PR.  That would make this one smaller... let me know!

cmd/kops/update_cluster.go
```diff
@@ -133,6 +137,8 @@ func NewCmdUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	viper.BindEnv("lifecycle-overrides", "KOPS_LIFECYCLE_OVERRIDES")
 	cmd.RegisterFlagCompletionFunc("lifecycle-overrides", completeLifecycleOverrides)
 
+	cmd.Flags().BoolVar(&options.DoDeferredDeletions, "cleanup-after-upgrade", options.DoDeferredDeletions, "Delete old revisions of cloud resources that were needed during an upgrade")
```
hakman, line 140: Do we want to use `prune` instead?
```suggestion
	cmd.Flags().BoolVar(&options.DoDeferredDeletions, "prune", options.DoDeferredDeletions, "Delete old revisions of cloud resources that were needed during an upgrade")
```

hakman (changes requested):

cmd/kops/update_cluster.go
justinsb: I like prune, let me make the change!

justinsb:
Changed to --prune, added a few more comments, and squashed the commits.

hakman (approved):

PR #16358, opened 2024-02-17
azure: Avoid spurious changes in VirtualNetwork

hakman:


justinsb:
Agreed - having the subnet field also makes the planner think there's a dependency

/approve
/lgtm

PR #16363, opened 2024-02-18
Update Calico to v3.27.3

hakman:


justinsb:
From one the mount-bpffs logs: `calico-node: error while loading shared libraries: libpcap.so.0.8: cannot open shared object file: No such file or directory`

It's on all the nodes, but e.g. https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kops/16363/pull-kops-e2e-cni-calico-ipv6/1759139852261003264/artifacts/cluster-info/kube-system/calico-node-8xp9w/

Maybe we need to install libpcap in nodeup when using calico?

hakman:
Related to https://github.com/projectcalico/calico/issues/8541. We will wait for v3.27.3.

hakman:
@rifelpet Could you check this error with IPv6? Looks related to https://github.com/kubernetes/kops/pull/16448.
```
Error: error determining default DNS zone: error querying zones:
error listing hosted zones: operation error Route 53:
ListHostedZones, failed to resolve service endpoint, endpoint rule error,
Invalid Configuration: Missing Region
```

rifelpet:
Yes I'll take a look. The old code wasn't specifying a region either so there must be a behavioral change in v2 that isnt documented in the migration guide.

https://github.com/kubernetes/kops/pull/16448/files#diff-8b59c898fa44da5fbbb0cf9868a918db03cdefe41c6855ef5840762ba9837455

it only happens in the ipv6 job too.

hakman:
> Yes I'll take a look. The old code wasn't specifying a region either so there must be a behavioral change in v2 that isnt documented in the migration guide.
> 
> https://github.com/kubernetes/kops/pull/16448/files#diff-8b59c898fa44da5fbbb0cf9868a918db03cdefe41c6855ef5840762ba9837455
> 
> it only happens in the ipv6 job too.

Exactly. Could we merge this and the cherry-pick in the mean time?

rifelpet (approved):

PR #16372, opened 2024-02-22
chore: update dependencies in submodules

justinsb:
Our submodules should also get auto-updated and auto-tidied by the
Makefile targets; this should then be done automatically by our github
action scheduled task.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -324,12 +324,16 @@ ko-kops-utils-cp-push:
 gomod:
 	go mod tidy
 	go mod vendor
-	cd tests/e2e; go mod tidy
 	cd hack; go mod tidy
+	cd tests/e2e; go mod tidy
+	cd tools/otel/traceserver; go mod tidy
 
 .PHONY: goget
 goget:
 	go get $(shell go list -f '{{if not (or .Main .Indirect)}}{{.Path}}{{end}}' -mod=mod -m all | grep -v spotinst-sdk-go)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -324,12 +324,16 @@ ko-kops-utils-cp-push:
...
-	cd tests/e2e; go mod tidy
 	cd hack; go mod tidy
+	cd tests/e2e; go mod tidy
+	cd tools/otel/traceserver; go mod tidy
 
 .PHONY: goget
 goget:
 	go get $(shell go list -f '{{if not (or .Main .Indirect)}}{{.Path}}{{end}}' -mod=mod -m all | grep -v spotinst-sdk-go)
+	cd hack; go get $(shell go list -f '{{if not (or .Main .Indirect)}}{{.Path}}{{end}}' -mod=mod -m all | grep -v spotinst-sdk-go)
+	cd tests/e2e; go get $(shell go list -f '{{if not (or .Main .Indirect)}}{{.Path}}{{end}}' -mod=mod -m all | grep -v spotinst-sdk-go)
```
hakman, line 335: I don't think I want ginkgo and kubetest updated, at least not as part of this flow.
Also, we need to remove  `grep -v spotinst-sdk-go` everywhere.

hakman (changes requested):

Makefile
justinsb: Done; we don't import ginkgo directly anywhere, and we only import kubetest2 in one submodule; added the exclusion there.

hakman (approved):

PR #16376, opened 2024-02-23
Update aws-sdk-go to support new testing (prow) environment

justinsb:
The newer SDK version lets us use built-in authentication in our new
prow testing environment.


hakman (approved):

justinsb:
Finally got everything passing CI:
*  version bumps for libraries with vulnerabilities - golang.org/x/crypto and	golang.org/x/net  (to make the dependency checker happy)
* updated controller-runtime so it would not crash with new go (and then had to update CRDs) 
* cherry-picked a simple logging fix to keep go vet happy also

PR #16377, opened 2024-02-23
Update aws-sdk-go to support new testing (prow) environment

justinsb:
The newer SDK version lets us use built-in authentication in our new
prow testing environment.


justinsb:
Actually, looks like this was already done on 1.26 - ignore me!

/close

PR #16379, opened 2024-02-24
GCE: Use internal load balancer for node to control-plane traffic

justinsb:
- gce: Always create an internal load balancer
- gce: Prune old forwarding rules
- gce: use internal IP address for node -> control-plane communication

### DIFF
--- a/pkg/model/gcemodel/api_loadbalancer.go
+++ b/pkg/model/gcemodel/api_loadbalancer.go
@@ -87,23 +87,6 @@ func (b *APILoadBalancerBuilder) createPublicLB(c *fi.CloudupModelBuilderContext
 			"name":           "api",
 		},
 	})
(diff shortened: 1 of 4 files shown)


justinsb:
/assign @hakman 

As we discussed in office hours, it looks like we can use the CIDR matching (though on the subnets on GCE) and side-step the need to match internal vs external addresses.

We can probably also match to the IP address in the subnet in future, minimizing cross-zone traffic, though I don't think this is actually used for more than a few KB of transfer so it shouldn't move the needle.

justinsb:
/retest

Calico failure on AWS should be unrelated

hakman (approved):

rifelpet:
Confirmed the GCE e2e jobs are passing now: https://testgrid.k8s.io/kops-gce#kops-gce-stable

Issue #15969, opened 2023-09-27
Install crictl/nerdctl on kops nodes


upodroid:
/kind feature

**1. Describe IN DETAIL the feature/behavior/change you would like to see.**
crictl/nerdctl allow you to debug containerd easily

**2. Feel free to provide a design supporting your feature request.**

hakman:
Seems like a good idea. Not very hard to do.

hakman:
Most of the work was done by @h3poteto. Thank you!
/remove-lifecycle rotten

PR #16383, opened 2024-02-25
Install nerdctl and crictl on nodes

h3poteto:
refs #15969 

Add CLI tools nerdctl and crictl on nodes.
nerdctl is for only containerd, so it will be installed only when the container runtime is containerd. crictl will be installed always.


upup/pkg/fi/cloudup/crictl.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"k8s.io/kops/pkg/apis/kops"
+	"k8s.io/kops/pkg/assets"
+	"k8s.io/kops/util/pkg/architectures"
+	"k8s.io/kops/util/pkg/hashing"
+)
+
+const (
+	crictlAssetUrlAmd64 = "https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.29.0/crictl-v1.29.0-linux-amd64.tar.gz"
+	crictlAssetUrlArm64 = "https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.29.0/crictl-v1.29.0-linux-arm64.tar.gz"
```
h3poteto, line 15: I'm wondering how to update this version when a new version is released. Do you have any ideas?

nodeup/pkg/model/crictl.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+var _ fi.NodeupModelBuilder = &CrictlBuilder{}
+
+func (b *CrictlBuilder) Build(c *fi.NodeupModelBuilderContext) error {
+	assetName := "crictl"
+	assetPath := ""
+	asset, err := b.Assets.Find(assetName, assetPath)
+	if err != nil {
+		return fmt.Errorf("unable to locate asset %q", err)
```
justinsb, line 22: Nit: `"unable to locate asset %q: %w", assetName, err)` I think you mean - although I'm not sure what happens if you %q an err

nodeup/pkg/model/crictl.go
```diff
@@ -0,0 +1,48 @@
...
+		path = "/opt/kops/bin"
+	}
+	if b.Distribution == distributions.DistributionContainerOS {
+		path = "/home/kubernetes/bin"
+	}
+	return path
+}
+
+func (b *CrictlBuilder) crictlPath() string {
+	return b.binaryPath() + "/crictl"
```
justinsb, line 47: Nit: You should be using filepath.Join or path.Join to avoid hard-coding the path separator.  (And in most cases it's filepath.Join for files, path.Join for files).  In this case though it really doesn't matter because we don't support windows and presumably it would be crictl.exe anyway!

nodeup/pkg/model/nerdctl.go
```diff
@@ -0,0 +1,65 @@
...
+
+type NerdctlBuilder struct {
+	*NodeupModelContext
+}
+
+var _ fi.NodeupModelBuilder = &NerdctlBuilder{}
+
+func (b *NerdctlBuilder) Build(c *fi.NodeupModelBuilderContext) error {
+	if b.skipInstall() {
+		klog.Info("SkipInstall is set to true; won't install nerdctl")
```
justinsb, line 20: Nit: Maybe `containerd.skipInstall is set to true; won't install nerdctl` or maybe `not installing containerd; won't install nerdctl`

tests/integration/update_cluster/minimal_scaleway/data/scaleway_instance_server_control-plane-fr-par-1-0_user_data
```diff
@@ -130,7 +130,7 @@ ClusterName: scw-minimal.k8s.local
 ConfigBase: memfs://tests/scw-minimal.k8s.local
 InstanceGroupName: control-plane-fr-par-1
 InstanceGroupRole: ControlPlane
-NodeupConfigHash: Eju0QG9LIx4osURNz5K7RCCbZ5hWBWrmyJ83jQfXFeE=
+NodeupConfigHash: qkMcgEhTy3oHXOw6o8xCcTcSY7T7PLlQIesc5/8Ntvg=
```
justinsb, line 133: (Aside): I'm starting to wonder if we should just regexp replace these values in the golden files :-)

upup/pkg/fi/cloudup/crictl.go
justinsb: Great question; we have done it by hand so far.  The other problem is that it would be nice if we "hard-coded" the sha somehow, both for security and to avoid having to download the sha file "every time".  One thing we could do is put it into a yaml file like the channels file(s), but find a format that dependabot could use (or just create our own dependabot like tool for our yaml file, if we can't fit it into dependabot).  We could either download that channels file on demand, or we could use `go:embed` (or both!)

As this is not something this PR is introducing, I don't consider it a blocker for this PR.  But cc @hakman to see if he has any thoughts here.

upup/pkg/fi/cloudup/nerdctl.go
```diff
@@ -0,0 +1,38 @@
...
+const (
+	nerdctlAssetUrlAmd64  = "https://github.com/containerd/nerdctl/releases/download/v1.7.4/nerdctl-1.7.4-linux-amd64.tar.gz"
+	nerdctlAssetUrlArm64  = "https://github.com/containerd/nerdctl/releases/download/v1.7.4/nerdctl-1.7.4-linux-arm64.tar.gz"
+	nerdctlAssetHashAmd64 = "71aee9d987b7fad0ff2ade50b038ad7e2356324edc02c54045960a3521b3e6a7"
+	nerdctlAssetHashArm64 = "d8df47708ca57b9cd7f498055126ba7dcfc811d9ba43aae1830c93a09e70e22d"
+)
+
+func findNerdctlAsset(c *kops.Cluster, assetBuilder *assets.AssetBuilder, arch architectures.Architecture) (*url.URL, *hashing.Hash, error) {
+	if c.Spec.Containerd == nil {
+		return nil, nil, fmt.Errorf("unable to find containerd config")
```
justinsb, line 22: Nit: We don't actually depend on containerd config in this function; is the expectation that we would e.g. match versions in future?  I think the intent is to help people not add the nerdctl asset if nodeup is not going to install it.  A comment would be helpful here - though I think the only people likely to hit this error would be developers changing the flow of the code.

justinsb:
A few nits but nothing blocking.

PR looks good to me, I will approve in a few hours after tests pass unless you want to address any nits (unless someone else weighs in).

Thanks for doing this - I'm hoping to get to your other PRs as well today!

upup/pkg/fi/cloudup/nerdctl.go
h3poteto: Oh, that's right. There is no problem in calling this function even if containerd is not set up.

justinsb:
/ok-to-test

(Whoops)

justinsb:
Thanks for addressing the nits @h3poteto - they were nits so not needed, but thank you all the same!

/approve
/lgtm

PR #16387, opened 2024-02-26
Update NVIDIA Container Toolkit URL

elezar:
The NVIDIA Container Toolkit packaging has been simplified to produce a single deb (or rpm) package. This means that the URL is no longer distribution dependent and the stable/deb repository path is used instead.

See https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html#installing-with-apt


hakman (approved):
Thanks @elezar!

PR #16390, opened 2024-02-28
chore: update boilerplate.py to recognize new build tags

justinsb:
Sync up with upstream https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.py

We want to pick up the new-style go:build tags.


justinsb:
Aside: this really should be containerized!

hakman (approved):

PR #16394, opened 2024-03-03
Fix shellcheck warnings for bootstrap script

hakman:
```
In x.sh line 37:
local -r urls=( $(split-commas "$3") )
^------------------^ SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In x.sh line 50:
"curl -f --compressed -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10"
^-----^ SC2027 (warning): The surrounding quotes actually unquote this. Remove or escape them.
^-----^ SC2206 (warning): Quote to prevent word splitting/globbing, or split robustly with mapfile or read -a.


In x.sh line 51:
"wget --compression=auto -O "${file}" --connect-timeout=20 --tries=6 --wait=10"
^-----^ SC2027 (warning): The surrounding quotes actually unquote this. Remove or escape them.
^-----^ SC2206 (warning): Quote to prevent word splitting/globbing, or split robustly with mapfile or read -a.


In x.sh line 52:
"curl -f -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10"
^-----^ SC2027 (warning): The surrounding quotes actually unquote this. Remove or escape them.
^-----^ SC2206 (warning): Quote to prevent word splitting/globbing, or split robustly with mapfile or read -a.


In x.sh line 53:
"wget -O "${file}" --connect-timeout=20 --tries=6 --wait=10"
^-----^ SC2027 (warning): The surrounding quotes actually unquote this. Remove or escape them.
^-----^ SC2206 (warning): Quote to prevent word splitting/globbing, or split robustly with mapfile or read -a.


In x.sh line 81:
actual=$(sha256sum ${file} | awk '{ print $1 }') || true
^-----^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
actual=$(sha256sum "${file}" | awk '{ print $1 }') || true


In x.sh line 89:
echo $1 | tr "," "\n"
^-- SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
echo "$1" | tr "," "\n"

For more information:
https://www.shellcheck.net/wiki/SC2027 -- The surrounding quotes actually u...
https://www.shellcheck.net/wiki/SC2206 -- Quote to prevent word splitting/g...
https://www.shellcheck.net/wiki/SC2207 -- Prefer mapfile or read -a to spli...
```

### DIFF
--- a/pkg/model/bootstrapscript.go
+++ b/pkg/model/bootstrapscript.go
@@ -398,10 +398,12 @@ func (b *BootstrapScript) createProxyEnv(ps *kops.EgressProxySpec) (string, erro
 		}
 
 		// Set env variables for base environment
-		buffer.WriteString(`echo "http_proxy=` + httpProxyURL + `" >> /etc/environment` + "\n")
-		buffer.WriteString(`echo "https_proxy=` + httpProxyURL + `" >> /etc/environment` + "\n")
-		buffer.WriteString(`echo "no_proxy=` + ps.ProxyExcludes + `" >> /etc/environment` + "\n")
-		buffer.WriteString(`echo "NO_PROXY=` + ps.ProxyExcludes + `" >> /etc/environment` + "\n")
+		buffer.WriteString(`{` + "\n")
+		buffer.WriteString(`  echo "http_proxy=` + httpProxyURL + `"` + "\n")
+		buffer.WriteString(`  echo "https_proxy=` + httpProxyURL + `"` + "\n")
+		buffer.WriteString(`  echo "no_proxy=` + ps.ProxyExcludes + `"` + "\n")
+		buffer.WriteString(`  echo "NO_PROXY=` + ps.ProxyExcludes + `"` + "\n")
+		buffer.WriteString(`} >> /etc/environment` + "\n")
 
 		// Load the proxy environment variables
 		buffer.WriteString("while read -r in; do export \"${in?}\"; done < /etc/environment\n")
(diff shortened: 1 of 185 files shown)


rifelpet:
This makes the userdata script longer which may cause IGs with additionalUserData to exceed the cloud provider's max size. We have CompressUserData as an API field but i wonder if we should be compressing it automatically if the total size exceeds known limits.

hakman:
> This makes the userdata script longer which may cause IGs with additionalUserData to exceed the cloud provider's max size.

The increase is from 4388 to 4506 chars in `bootstrapscript_0.txt`. I don't think 120 chars this is that relevant.
We could probably remove all the "==" prefixes and suffixes, to reduce it.

> We have CompressUserData as an API field but i wonder if we should be compressing it automatically if the total size exceeds known limits.

We probably should auto-compress, or exit with error.

pkg/model/resources/nodeup.go
```diff
@@ -79,28 +81,28 @@ download-or-bust() {
   while true; do
     for url in "${urls[@]}"; do
       commands=(
-        "curl -f --compressed -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10"
-        "wget --compression=auto -O "${file}" --connect-timeout=20 --tries=6 --wait=10"
-        "curl -f -Lo "${file}" --connect-timeout 20 --retry 6 --retry-delay 10"
-        "wget -O "${file}" --connect-timeout=20 --tries=6 --wait=10"
+        "curl -f --compressed -Lo ${file} --connect-timeout 20 --retry 6 --retry-delay 10"
```
justinsb, line 84: I'm sort of surprised we don't need quotes here to keep shellcheck happy, but agree that we shouldn't need quotes in practice.

pkg/model/resources/nodeup.go
```diff
@@ -64,9 +64,11 @@ function ensure-install-dir() {
 
 # Retry a download until we get it. args: name, sha, urls
 download-or-bust() {
+  echo "== Downloading $1 with hash $2 from $3 =="
   local -r file="$1"
   local -r hash="$2"
-  local -r urls=( $(split-commas "$3") )
+  local -a urls
+  mapfile -t urls < <(split-commas "$3")
```
justinsb, line 71: It looks like mapfile is a bash-ism, but I think we're already relying on bash-specific features.

justinsb:
Some comments, but no objection if we're OK with the size.  I'm sort of surprised that we change the size at all here!  I thought we already had auto-compression on AWS (where the limit is so strict), but yeah, maybe trimming off the `==` suffixes is a good way to get a few bytes back?  (The prefix is arguably useful for regexes and humans scanning, the suffixes I would guess less so)

pkg/model/resources/nodeup.go
hakman: Actually, with quotes is not working quite right...

pkg/model/resources/nodeup.go
hakman: See: https://www.shellcheck.net/wiki/SC2207

hakman:
If @rifelpet is ok with the size for now, I can iterate and get it smaller. I already have some additional commits on the topic.

rifelpet:
We dont have autocompression on any cloud providers:

https://github.com/kubernetes/kops/blob/045bd4932b7bd9888b99755b00b1106b535753cd/pkg/model/bootstrapscript.go#L353

https://github.com/kubernetes/kops/blob/045bd4932b7bd9888b99755b00b1106b535753cd/pkg/model/resources/nodeup.go#L163-L165

I think its worthy of a release note given that upgrading the kops version to include this PR could cause `kops update cluster --yes` to break on an existing cluster, but if we could automatically compress based on each cloud provider's max user-data size then we could remove the release note. I think handling either option in a follow up is fine.

hakman:
@rifelpet @justinsb I reduced the output to 4387 bytes with the new update, so 1 byte below what we had before 😅 .
Could you take another look?

pkg/model/bootstrapscript.go
```diff
@@ -398,28 +398,28 @@ func (b *BootstrapScript) createProxyEnv(ps *kops.EgressProxySpec) (string, erro
 		}
 
 		// Set env variables for base environment
-		buffer.WriteString(`echo "http_proxy=` + httpProxyURL + `" >> /etc/environment` + "\n")
-		buffer.WriteString(`echo "https_proxy=` + httpProxyURL + `" >> /etc/environment` + "\n")
-		buffer.WriteString(`echo "no_proxy=` + ps.ProxyExcludes + `" >> /etc/environment` + "\n")
-		buffer.WriteString(`echo "NO_PROXY=` + ps.ProxyExcludes + `" >> /etc/environment` + "\n")
+		buffer.WriteString(`{` + "\n")
```
justinsb, line 401: Sneaky!  I bet a heredoc would be even shorter, but let's keep that one in the bank for the future!

pkg/model/resources/nodeup.go
justinsb: Right - I would guess we would need to get fancy with `'` and `"` and escaping

justinsb:
LGTM.  I'm going to approve but hold to give @rifelpet a chance to see it

/approve
/lgtm

/hold for a few hours for @rifelpet

PR #16395, opened 2024-03-03
e2e tests: When upgrading, wait for the new configuration

justinsb:
In our upgrade-ab tests, we might be making non-trivial cloud
infrastructure changes, such as waiting for a new load balancer.

Add a `kops validate cluster` step to wait for those changes to be
applied.


justinsb:
I hope this will fix https://testgrid.k8s.io/kops-upgrades#kops-aws-upgrade-bug16276-calico, which is currently failing at the `channels apply` line (the next line), I think because it is using kubeconfig-B.  We just did a `kubectl get nodes` with kubeconfig-A, which is why that line isn't telling us anything about kubeconfig-B.  We expect kubeconfig-A and kubeconfig-B to _both_ be working (if the upgrade is not disruptive).

rifelpet (approved):

Issue #16024, opened 2023-10-16
NLB for apiserver port 8443 is unreachable


jim-barber-he:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
 
It's actually a build from the `master` branch since I was testing a fix for another issue that I had previously encountered (that is now fixed).
It was built like so:
```
$ go version  
go version go1.21.3 linux/amd64
$ export S3_BUCKET=hetest-kops
$ export VERSION=1.28.0-dev.1 
$ make kops-install VERSION=$VERSION
$ make upload S3_BUCKET=s3://$S3_BUCKET VERSION=$VERSION
```
And results in 
```
$ kops version
Client version: 1.28.0-dev.1 (git-v1.29.0-alpha.1-139-gab5b8a873a)
```

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

```
$ kubectl version                                            
Client Version: v1.28.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.2
```

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

I have created a cluster from a manifest as per how usually create clusters.
This time using this custom kops to deploy a 1.28 k8s cluster.

**5. What happened after the commands executed?**

I couldn't talk to the API server externally.

**6. What did you expect to happen?**

Be able to use the cluster.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

I'll just provide the part that probably matters.
We use a custom SSL certificate for the API server with a name of `api.$CLUSTER_NAME`.
It is provided in the cluster spec like so:

```yaml
spec:
  api:
    loadBalancer:
      class: Network
      sslCertificate: arn:aws:acm:$AWS_REGION:$AWS_ACCOUNT:certificate/REDACTED
      sslPolicy: ELBSecurityPolicy-TLS13-1-3-2021-06
      type: Internal
```

When using a custom SSL certificate, rules are created in the NLB for port `8443` in addition to `443` and the `kops export kubecfg --admin --name $CLUSTER` created entries in `.kube/config` referring to port `8443`.

**8. Please run the commands with most verbose logging by adding the `-v 10` flag.
[...]

PR #16403, opened 2024-03-13
aws: Expose port 8443 when using NLB with a custom certificate

justinsb:
We dropped the security group rules for this when moving to NLB
security groups; add them back.

Issue #16024


hakman:
Thanks @justinsb. LGTM
I will approve once the expected output is updated.

hakman (approved):

Issue #15680, opened 2023-07-20
Further tighten IAM permissions for cert-manager DNS01 challenges


andsens:
/kind feature

Big thumbs up for already tightening the rather loose policy that cert-manager suggests in their guide by restricting the hosted zone. I have a suggestion for tighten it even further. Split up the `ChangeResourceRecordSets` and `ListResourceRecordSets` and restrict any changes to TXT records on `_acme-challenge.*` subdomains.

```
{
  "Effect": "Allow",
  "Action": "route53:ChangeResourceRecordSets",
  "Resource": "arn:aws:route53:::hostedzone/XXXXXXXXXXXXXX",
  "Condition": {
    "ForAllValues:StringLike":{
      "route53:ChangeResourceRecordSetsNormalizedRecordNames": ["_acme-challenge.*"]
    },
    "ForAllValues:StringEquals":{
      "route53:ChangeResourceRecordSetsRecordTypes": ["TXT"]
    }
  }
}
```

I modified the policy kops created with the suggested change and it works like a charm :-)

andsens:
/remove-lifecycle stale

Might send a PR if I get the time.

PR #16412, opened 2024-03-18
aws/cert-manager: Tighten IAM permissions for cert-manager

andsens:
This change restricts which record types and domain prefixes cert-manager is allowed to change for DNS01 acme challenges.

Only `_acme-challenge.*` TXT records may be created/updated/removed.

Implements kubernetes/kops#15680

I ran `make all examples test` and it compiles, but beyond that I have not created any tests for this (knowledge of go is ankle deep) so I hope someone is kind enough to help me out and add that :-)


hakman:
Thank you @andsens!
/lgtm
/approve

PR #16433, opened 2024-03-29
chore: update testdata to stop using alpha versions

justinsb:
- chore: update testdata to stop using alpha versions
- chore: update generated testdata


justinsb:
I'm setting us up here to store well-known artifact hashes by embedding them in the kops binary.  Not strictly necessary, because we will fall back on unknown versions I think, but it does make the tests cleaner / faster.

Issue #16424, opened 2024-03-25
Migrate to aws-sdk-go-v2


rifelpet:
/kind cleanup

https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

We'll need to migrate to github.com/aws/aws-sdk-go-v2 before v1 enters maintenance mode

Grouped into packages that likely need to be updated together
- [x] EC2 cloudmock, pkg/resources/aws, awsup, awstasks, pkg/apis/kops/validation, pkg/model, cloudup/new_cluster.go, template_functions.go, testutils
- [x] Autoscaling cloudmock, pkg/resources/aws, awsup, awstasks, pkg/instancegroups, pkg/model
- [x] ELB cloudmock, pkg/resources/aws, awsup, awstasks 
- [x] ELBv2 cloudmock, pkg/resources/aws, awsup, awstasks, testutils
- [x] Eventbridge cloudmock, pkg/resources/aws, awsup, awstasks, pkg/model
- [x] IAM cloudmock, pkg/resources/aws, awsup, awstasks, pkg/model, testutils
- [x] SQS cloudmock, pkg/resources/aws, awsup, awstasks 
- [x] Route53 cloudmock, pkg/resources/aws, awsup, awstasks, dnscontroller, testutils
- [x] SSM awsup
- [x] EC2 pkg/nodeidentity
- [x] EC2 protokube
- [x] EC2 nodeup
- [x] EC2 cmd/kops
- [ ] STS nodeup, kops-controller
- [x] pkg/zones

rifelpet:
Two design questions I'm soliciting input on:

1. Most integer types in the SDK have changed from *int64 to *int32. Kops API has *int64 fields that are passed directly through the model and tasks into the SDK types:
   https://github.com/kubernetes/kops/blob/393aac40dfb668073516c11e055594324a33a480/pkg/apis/kops/v1alpha3/cluster.go#L848-L855
   As-is we have to convert from *int64 to *int32:
   https://github.com/kubernetes/kops/blob/d0d6f8b2f5e68e6d4c6926e34d2554219a654f0b/pkg/model/awsmodel/autoscalinggroup.go#L104
   Should we update the Kops API to use *int32 ? I realize some of them are used by other cloud providers, but at least some of them are AWS-only.

[...]

PR #16435, opened 2024-03-30
Migrate IAM to aws-sdk-go-v2

rifelpet:
ref: #16424


pkg/model/awsmodel/iam.go
```diff
@@ -21,8 +21,8 @@ import (
 	"sort"
 	"strings"
 
+	awsIam "github.com/aws/aws-sdk-go-v2/service/iam"
```
justinsb, line 24: Nit: we should probably switch `awsIam` -> `awsiam` at some stage, it's more go idiomatic I believe

pkg/model/awsmodel/iam.go
```diff
@@ -473,20 +473,25 @@ func (b *IAMModelBuilder) buildAWSIAMRolePolicy(role iam.Subject) (fi.Resource,
...
+		}
+		for _, role := range page.Roles {
 			if !strings.HasSuffix(fi.ValueOf(role.RoleName), "."+b.Cluster.ObjectMeta.Name) {
 				continue
 			}
 			getRequest := &awsIam.GetRoleInput{RoleName: role.RoleName}
-			roleOutput, err := iamapi.GetRole(getRequest)
+			roleOutput, err := iamapi.GetRole(ctx, getRequest)
 			if err != nil {
-				getRoleErr = fmt.Errorf("calling IAM GetRole on %s: %w", fi.ValueOf(role.RoleName), err)
```
justinsb, line 488: Ah, this is going to be a much more ergonomic API :-)

pkg/model/awsmodel/iam.go
```diff
@@ -500,13 +505,9 @@ func (b *IAMModelBuilder) FindDeletions(context *fi.CloudupModelBuilderContext,
 				}
 			}
 		}
-		return true
-	})
+	}
 	if getRoleErr != nil {
```
justinsb, line 509: Nit: this isn't used any more, I believe?

pkg/resources/aws/aws.go
```diff
@@ -2078,11 +2078,17 @@ func ListIAMOIDCProviders(cloud fi.Cloud, vpcID, clusterName string) ([]*resourc
...
-				continue
-			} else if err != nil {
-				return nil, fmt.Errorf("error getting IAM OIDC Provider: %v", err)
+			resp, err := c.IAM().GetOpenIDConnectProvider(ctx, descReq)
+			if err != nil {
+				var nse *iamtypes.NoSuchEntityException
+				if errors.As(err, &nse) {
+					klog.Warningf("could not find IAM OIDC Provider %q. Resource may already have been deleted: %v", aws.StringValue(arn), nse)
+					continue
+				} else if awserror, ok := err.(smithy.APIError); ok && awserror.ErrorCode() == "403" {
```
justinsb, line 2087: This feels like a more fragile API than we had before.  Maybe we should hide `if awserror, ok := err.(smithy.APIError); ok && awserror.ErrorCode() == "403" ` behind a function like `IsHTTP403Error(err error) bool` ?

upup/pkg/fi/cloudup/awstasks/iamrolepolicy.go
```diff
@@ -95,27 +100,26 @@ func (e *IAMRolePolicy) Find(c *fi.CloudupContext) (*IAMRolePolicy, error) {
 		PolicyName: e.Name,
 	}
 
-	response, err := cloud.IAM().GetRolePolicy(request)
-	if awsErr, ok := err.(awserr.Error); ok {
-		if awsErr.Code() == iam.ErrCodeNoSuchEntityException {
+	response, err := cloud.IAM().GetRolePolicy(ctx, request)
+	if err != nil {
+		var nse *iamtypes.NoSuchEntityException
```
justinsb, line 105: I guess we could also have a helper for this check, but this one is simpler.

upup/pkg/fi/cloudup/awsup/aws_cloud.go
```diff
@@ -333,16 +332,7 @@ func NewAWSCloud(region string, tags map[string]string) (AWSCloud, error) {
...
-			Config:            *config,
-			SharedConfigState: session.SharedConfigEnable,
-		})
-		if err != nil {
-			return c, err
-		}
-		c.iam = iam.New(sess, config)
-		c.iam.Handlers.Send.PushFront(requestLogger)
-		c.addHandlers(region, &c.iam.Handlers)
+		c.iam = iam.NewFromConfig(cfgV2)
```
justinsb, line 335: So we are losing 3 things:
* SharedConfigState 
* requestlogger
* addHandlers - which configures the cross-request delay

Do we think any of those are still important?

justinsb:
Thanks for doing this @rifelpet 

Generally looks good, but my one "blocker" is whether we've lost things by "just" doing `iam.NewFromConfig(cfgV2)`.  I'm fine if the answer is "we may be losing things, and if we are we can add them back, but we don't think we need them" - but I just want to make sure we're doing it consciously.

rifelpet:
> Thanks for doing this @rifelpet
> 
> Generally looks good, but my one "blocker" is whether we've lost things by "just" doing `iam.NewFromConfig(cfgV2)`. I'm fine if the answer is "we may be losing things, and if we are we can add them back, but we don't think we need them" - but I just want to make sure we're doing it consciously.

I mention the config changes [here](https://github.com/kubernetes/kops/pull/16432#issue-2215225398). I'm hoping that the SDK's new built-in retryer and retry logger will be sufficient for our needs but we'll find out once we run scale tests (and likely wont find out until the largest AWS service is migrated - EC2)

I'll address your nits in a followup

upup/pkg/fi/cloudup/awsup/aws_cloud.go
rifelpet: SharedConfigState is now default: https://aws.github.io/aws-sdk-go-v2/docs/migrating/#client-construction

there is no way to disable or enable SharedConfigState anymore, only specify which shared config file to use (default `~/.aws/config`)

PR #16439, opened 2024-03-30
Use embedded hashes for our well-known assets

justinsb:
- Create assetdata
- Use embedded hashes for our well-known assets


hakman (approved):
I think some of the logic from the initial PR got lost somewhere. If I recall correctly we were removing all the unused binaries and architectures. Though this is a nit, we can address it later.
Thanks! 😄

PR #16440, opened 2024-03-31
Update kube-router to v2.1.0

aauren:
@hakman for review

Changes:

* Bump kube-router version to v2.1.0 from v1.6.0
* Enable hairpin mode for kube-router by default - From the e2e tests this seems to be expected functionality of a CNI
* Add the runtime endpoint into the container so that user's are able to utilize DSR functionality
* Mount `/etc/iproute2/rt_tables` into the container as well which is used by kube-router to configure source routing for DSR
* Add `hostPID: true` functionality for DSR
* Add environment var `POD_NAME` so that user's can easily enable load-balancing mode for kube-router
* Add additional RBAC for the following:
  * leases / services/status - For loadbalancer controller which can be optionally enabled
  * endpointslices - This is needed in kube-router v2.X and on so that kube-router is able to handle dual-stack endpoints

I found that in the current main branch of kops that kube-router's old method of adding instance modify permissions to the nodes no longer works correctly. I believe that this permission should be coming in when kube-router is chosen as the networking provider:

* https://github.com/kubernetes/kops/blob/master/pkg/model/components/addonmanifests/kuberouter/iam.go#L35
* https://github.com/kubernetes/kops/blob/master/pkg/model/iam/iam_builder.go#L1086-L1090

However, this no longer seems to be applying to nodes correctly and when this is missing kube-router is not able to forward pod traffic to other nodes.

After poking around a bit, I also noticed that other CNIs like Calico have special logic in the IAM builder like the following to add similar functionality in the IAM node builder code: https://github.com/kubernetes/kops/blob/master/pkg/model/iam/iam_builder.go#L467-L469

So I added this functionality for kube-router as well and found that it works correctly and enables kube-router to disable the src-dest check in AWS.

Additionally removed all of the skip_regex.go exceptions that were kept for kube-router as with this version update kube-router now passes all of the upstream e2e conformance tests.

### DIFF
--- a/pkg/model/iam/iam_builder.go
+++ b/pkg/model/iam/iam_builder.go
@@ -468,6 +468,10 @@ func (r *NodeRoleNode) BuildAWSPolicy(b *PolicyBuilder) (*Policy, error) {
 		addCalicoSrcDstCheckPermissions(p)
 	}
 
+	if b.Cluster.Spec.Networking.KubeRouter != nil {
+		addKubeRouterSrcDstCheckPermissions(p)
+	}
+
 	return p, nil
 }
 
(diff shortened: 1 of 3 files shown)


aauren:
I went ahead and re-added the skip regex for the 2 failing kube-router tests TCP close/wait which appears to be flaky as this one isn't failing consistently in my similar test-bed and the kube-proxy URLs one as something appears to have changed in the test package that causes this to fail.

I'll look more into these later, but this should hopefully get it through.

hakman:
Nice bump. Thanks @aauren!
/lgtm
/approve

PR #16444, opened 2024-03-31
Migrate aws-sdk-go/aws to aws-sdk-go-v2/aws

rifelpet:
Ref: #16424

In an effort to reduce the diffs when migrating the last few largest services, this migrates most of the usage of `aws` pointer conversion methods to use the v2 SDK. 

I left files that still use v1's `aws.Config`


hakman:
Nice approach. Thanks!

PR #16460, opened 2024-04-10
Migrate EC2 to aws-sdk-go-v2

rifelpet:
Ref: #16424

This largely finishes the migration to aws-sdk-go-v2. I'll update the issue with remaining steps and blockers.

This is a large PR and I did my best to break up commits into logical chunks, though each commit isnt necessarily able to compile. To get each commit to compile would have involved adding an extra `AWSCloud.EC2V2()` interface method and gradually migrating consumers to use it and then clean it up afterwards, which would have had a much wider impact.

/cc @hakman


hakman:
@rifelpet I think I am seeing a new error that prevents deleting the AWS clusters:
> security-group:sg-062e8760d65edf1da	error deleting resources, will retry: cannot revoke ingress for ID "sg-062e8760d65edf1da": operation error EC2: RevokeSecurityGroupIngress, https response error StatusCode: 400, RequestID: ea681b1f-28fa-4c1c-8b58-7ccd9b795e8d, api error InvalidRequest: The request received was invalid.

hakman:
This is indeed huge, but looks great. Very nice work @rifelpet!
Still needs the fix for deleting the clusters, but LGTM.

rifelpet:
> @rifelpet I think I am seeing a new error that prevents deleting the AWS clusters:
> 
> > security-group:sg-062e8760d65edf1da	error deleting resources, will retry: cannot revoke ingress for ID "sg-062e8760d65edf1da": operation error EC2: RevokeSecurityGroupIngress, https response error StatusCode: 400, RequestID: ea681b1f-28fa-4c1c-8b58-7ccd9b795e8d, api error InvalidRequest: The request received was invalid.

I was able to repro this with a very simple script. It seems like a bug in the v2 SDK given that it worked fine with the v1 SDK, but I changed the cleanup to use SecurityGroupRuleIDs instead and now it is passing 🤞🏻

hakman (approved):

PR #16467, opened 2024-04-18
kube-router: fix version and cluster ip range

aauren:
FYI @hakman 

* Updates the kube-router init container to also be v2.1.0 which is latest at this point
* Adds the service Cluster IP range from the kops configuration so that cluster IPs can be ingressed to the node and then later be processed by NetworkPolicy


hakman (approved):
💯

PR #16473, opened 2024-04-18
nodeup: error message instead of segfault

multi-io:
in case nodeupConfig is local and ConfigStore is not set.


hakman (approved):
Thanks @multi-io! 😄

PR #16479, opened 2024-04-20
fix: Detect resolv.conf being a symlink, rather than relying on distro detection

justinsb:
Distro detection is less reliable than understanding the actual
configuration in use, and with Debian 12 multiple configurations are
(more) probable.

### DIFF
--- a/util/pkg/distributions/distributions.go
+++ b/util/pkg/distributions/distributions.go
@@ -18,6 +18,10 @@ package distributions
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+
+	"k8s.io/klog/v2"
 )
 
 // Distribution represents a particular version of an operating system.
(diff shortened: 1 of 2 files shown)


justinsb:
cc @hakman 

This is the code I had written before that may be related to the issue brought up in office hours with debian12 on hetzner.  I will test debian12 on a few places.

It was "stuck" behind a more complicated PR, AFAICT ... sorry!

justinsb:
I validated the problem on debian12 on hetzner (we aren't running systemd-resolved there), and I updated the PR to fix it.  Essentially we want to strictly detect systemd-resolved, which is the problem.  kubelet can read /etc/resolv.conf even if it's a symlink, the issue only arises if /etc/resolv.conf includes 127.0.0.53 or something like that, which doesn't work in a container.

Another approach could be to look at /etc/resolv.conf for a `nameserver 127.0.0.x` line (or a similar approach), but that feels tricky  - we still have to find the "real" resolv.conf file (the one without proxies that only work in the host network-namespace).

hakman:
This is nice. We also have the `resolv-conf` flag override if needed in some distro:
https://github.com/kubernetes/kops/blob/47b12ba8410e0d35e77d2b2fee1ee9675c033866/pkg/apis/kops/componentconfig.go#L114-L115

util/pkg/distributions/distributions.go
```diff
@@ -96,17 +100,55 @@ func (d *Distribution) DefaultUsers() ([]string, error) {
...
+//     which contains 127.0.0.53 (that won't work in a container), and upstream
+//     nameservers are in /run/systemd/resolve/resolv.conf.
+//     We must use /run/systemd/resolve/resolv.conf
+//
+// * systemd-resolved with libnss_resolve:
+//     /etc/resolv.conf has upstream nameservers. /etc/nsswitch.conf includes
+//     "resolve", which is a direct-path to systemd-resolved.
+//     We can use /etc/resolv.conf
+
+func (d *Distribution) HasLoopbackEtcResolvConf() (string, bool) {
```
hakman, line 122: Do we still need the bool, could just return a string ptr or `nil`.
```suggestion
func (d *Distribution) HasLoopbackEtcResolvConf() *string {
```

hakman (changes requested):

util/pkg/distributions/distributions.go
hakman: Seems like a minor nuisance.

rifelpet:
/lgtm cancel

for test failure

hakman:
I think https://github.com/kubernetes/kops/pull/16495 will fix the problem for now, as discussed during Office Hours.
Let's revisit this if the issue reappears.
/close

PR #16483, opened 2024-04-21
Migrate AWS Verifier to aws-sdk-go-v2

rifelpet:
Closes https://github.com/kubernetes/kops/issues/16424

From [this comment](https://github.com/kubernetes/kops/issues/16424#issuecomment-2046489720):

[While presigned requests are still supported in V2](https://aws.github.io/aws-sdk-go-v2/docs/migrating/#presigned-requests), the [presign methods and types](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/sts#PresignClient) no longer provide access to the request body, only their url and headers. See https://github.com/aws/aws-sdk-go-v2/issues/1137. Kops-controller currently reads the request body to perform some validation:
https://github.com/kubernetes/kops/blob/1c24423d9be20396e62755f6952b415567a56ced/upup/pkg/fi/cloudup/awsup/aws_verifier.go#L153-L157

In V1 the presigned request is a POST however in V2 it is converted to a GET request and the normal  `Action=GetCallerIdentity&Version=2011-06-15` body is moved to URL query parameters:
https://github.com/aws/aws-sdk-go-v2/blob/bc2a669d3241023e20194cdfe042b8c275887e51/service/sts/api_client.go#L641-L645

I've removed the validation that checks the Content-Length signed header matches the size of the request body, given the request body is now empty.

[This thread](https://github.com/kubernetes/kops/pull/9653#discussion_r463585666) on the original kops-controller PR discusses potential upgrade challenges. In this case I believe we'll have the normal race with this type of change:

1. Old node is launched
2. New `kops update cluster --yes` and `kops rolling-update --yes` is ran
3. New control plane node is launched, new kops-controller is running
4. Old node attempts to join and fails because the GetCallerIdentity requests dont match.

In reality I dont anticipate this being a problem because the old node would have needed to not join the cluster in the time it took the control plane to bootstrap and launch the kops-controller daemonset pod (a significantly longer process than normal k8s node bootstrap). Eventually the failed node will be cleaned up by cluster-autoscaler or karpenter.

My only concern would be disruption to `kops rolling-update` if the node never joins which causes cluster validation to fail until it is cleaned up.

### DIFF
--- a/cmd/kops-controller/main.go
+++ b/cmd/kops-controller/main.go
@@ -130,7 +130,7 @@ func main() {
 		var verifiers []bootstrap.Verifier
 		var err error
 		if opt.Server.Provider.AWS != nil {
-			verifier, err := awsup.NewAWSVerifier(opt.Server.Provider.AWS)
+			verifier, err := awsup.NewAWSVerifier(ctx, opt.Server.Provider.AWS)
 			if err != nil {
 				setupLog.Error(err, "unable to create verifier")
 				os.Exit(1)
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/amazon-ec2-instance-selector/v2 v2.4.2-0.20231216170552-14d4dfcbaadf
-	github.com/aws/aws-sdk-go v1.52.1
 	github.com/aws/aws-sdk-go-v2 v1.26.1
 	github.com/aws/aws-sdk-go-v2/config v1.27.11
 	github.com/aws/aws-sdk-go-v2/credentials v1.17.11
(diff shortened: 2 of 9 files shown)


justinsb:
We don't have great unit test coverage here, so I figured it might be good to add some: #16487

TBH I'm struggling to get the test to pass with v2, trying to figure it out...

I think we were already very sensitive to the aws-sdk version, so I don't think this is a huge regression version-wise.

rifelpet:
> I think I've figured it out ... the V2 API puts a lot more in the query URL. So we likely need to pass along the url/method/headers. The good thing is that eliminates the dependency on the aws-sdk-go version (going forwards!)
> 
> Here's my WIP / PoC branch ... let me know if you want me e.g. to send a PR to your branch or something like that. I'm happy to fix up the tests, the code change itself here is pretty simple (we encode an `awsV2Token`, including the url & method)
> 
> [`master...justinsb:kops:aws-sdk-go-v2-verifier`?expand=1 (compare)](https://github.com/kubernetes/kops/compare/master...justinsb:kops:aws-sdk-go-v2-verifier?expand=1)

Yes, feel free to PR to my branch or just push directly to it. I just rebased to resolve conflicts. Adding the other data in to the token makes sense to me 👍🏻

justinsb:
Thanks @rifelpet 

/approve
/lgtm

PR #16497, opened 2024-04-27
Support multiple service-account-issuers in apiserver

zetaab:
currently kOps supports only one --service-account-issuer flag in apiserver. However, it is problem if people want to change the issuer without downtime. https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection says

>  You can specify the --service-account-issuer argument multiple times, this can be useful to enable a non-disruptive change of the issuer. When this flag is specified multiple times, the first is used to generate tokens and all are used to determine which issuers are accepted. You must be running Kubernetes v1.22 or later to be able to specify --service-account-issuer multiple times.

This PR makes it possible to define issuer multiple times

### DIFF
--- a/nodeup/pkg/model/kube_apiserver.go
+++ b/nodeup/pkg/model/kube_apiserver.go
@@ -678,6 +678,7 @@ func (b *KubeAPIServerBuilder) buildPod(ctx context.Context, kubeAPIServer *kops
 			"/usr/local/bin/kube-apiserver",
 		}
 		container.Args = append(container.Args, sortedStrings(flags)...)
+		container.Args = sortServiceAccountIssuers(container.Args, fi.ValueOf(kubeAPIServer.ServiceAccountIssuer), kubeAPIServer.AdditionalServiceAccountIssuers)
 	}
 
 	for _, path := range b.SSLHostPaths() {
@@ -754,4 +755,27 @@ func (b *KubeAPIServerBuilder) buildAnnotations() map[string]string {
 	}
 
 	return annotations
+}
+
+func sortServiceAccountIssuers(in []string, currentIssuer string, oldIssuers []string) []string {
+	if len(oldIssuers) == 0 {
+		return in
+	}
+
+	positionMap := make(map[int]string)
+	positionMap[0] = fmt.Sprintf("--service-account-issuer=%s", currentIssuer)
+	for k, v := range oldIssuers {
+		positionMap[k+1] = fmt.Sprintf("--service-account-issuer=%s", v)
+	}
+
+	out := make([]string, 0, len(in))
+	issuerPosition := 0
+	for _, flag := range in {
+		if strings.HasPrefix(flag, "--service-account-issuer=") {
+			flag = positionMap[issuerPosition]
+			issuerPosition++
+		}
+		out = append(out, flag)
+	}
+	return out
 }
--- a/nodeup/pkg/model/kube_apiserver.go
+++ b/nodeup/pkg/model/kube_apiserver.go
@@ -678,7 +678,9 @@ func (b *KubeAPIServerBuilder) buildPod(ctx context.Context, kubeAPIServer *kops
 			"/usr/local/bin/kube-apiserver",
 		}
 		container.Args = append(container.Args, sortedStrings(flags)...)
-		container.Args = sortServiceAccountIssuers(container.Args, fi.ValueOf(kubeAPIServer.ServiceAccountIssuer), kubeAPIServer.AdditionalServiceAccountIssuers)
+		for _, issuer := range kubeAPIServer.AdditionalServiceAccountIssuers {
+			container.Args = append(container.Args, "--service-account-issuer="+issuer)
+		}
 	}
 
 	for _, path := range b.SSLHostPaths() {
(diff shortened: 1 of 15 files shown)


nodeup/pkg/model/kube_apiserver.go
```diff
@@ -678,6 +678,7 @@ func (b *KubeAPIServerBuilder) buildPod(ctx context.Context, kubeAPIServer *kops
 			"/usr/local/bin/kube-apiserver",
 		}
 		container.Args = append(container.Args, sortedStrings(flags)...)
+		container.Args = sortServiceAccountIssuers(container.Args, kubeAPIServer.ServiceAccountIssuers)
```
zetaab, line 681: needed because otherwise the issuers will be sorted just by sortedStrings. We want the order to be correct, like it was specified in the configuration

pkg/apis/kops/validation/validation.go
```diff
@@ -829,6 +829,13 @@ func validateKubeAPIServer(v *kops.KubeAPIServerConfig, c *kops.Cluster, fldPath
 		}
 	}
 
+	if v.ServiceAccountIssuer != nil && len(v.ServiceAccountIssuers) > 0 {
```
zetaab, line 832: not sure is this correct way of handling this? I would not like to deprecate the api field either.

tests/integration/update_cluster/additionalobjects/data/aws_s3_object_cluster-completed.spec_content
```diff
@@ -101,7 +101,6 @@ spec:
     requestheaderUsernameHeaders:
     - X-Remote-User
     securePort: 443
-    serviceAccountIssuer: https://api.internal.additionalobjects.example.com
```
zetaab, line 104: hmm? why its missing.. need to figure out

tests/integration/update_cluster/apiservernodes/data/aws_s3_object_cluster-completed.spec_content
```diff
@@ -94,7 +94,6 @@ spec:
     requestheaderUsernameHeaders:
     - X-Remote-User
     securePort: 443
-    serviceAccountIssuer: https://api.internal.minimal.example.com
```
zetaab, line 97: same here

tests/integration/update_cluster/aws-lb-controller/data/aws_s3_object_cluster-completed.spec_content
```diff
@@ -100,7 +100,6 @@ spec:
     requestheaderUsernameHeaders:
     - X-Remote-User
     securePort: 443
-    serviceAccountIssuer: https://discovery.example.com/minimal.example.com
```
zetaab, line 103: and here

nodeup/pkg/model/kube_apiserver.go
hakman: We also have lots of places where we use `Additonal...` to add secondary fields. This seems like one of those places, because the order matters.

nodeup/pkg/model/tests/golden/multiple-sa-issuers/cluster.yaml
```diff
@@ -0,0 +1,70 @@
...
+  kubernetesVersion: v1.28.0
+  masterPublicName: api.multisaissuer.example.com
+  networkCIDR: 172.20.0.0/16
+  kubeAPIServer:
+    serviceAccountIssuer: https://huh.com/multisaissuer.example.com
+    additionalServiceAccountIssuers:
+    - https://foobar.com
+    - https://aaa.com
+    - https://hello.com
+    - https://dorld.com
```
hakman, line 42: Could this be put in the complex test or one of the IRSA ones? I don't think we gain much by adding another test.

pkg/apis/kops/validation/validation.go
```diff
@@ -829,6 +829,9 @@ func validateKubeAPIServer(v *kops.KubeAPIServerConfig, c *kops.Cluster, fldPath
 		}
 	}
 
+	if c.IsKubernetesLT("1.22") && len(v.AdditionalServiceAccountIssuers) > 0 {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("additionalServiceAccountIssuers"), "additionalServiceAccountIssuers requires Kubernetes 1.22+"))
+	}
```
hakman, line 834: No need for this:
https://github.com/kubernetes/kops/blob/c6bb40ed328c99ca1490113334f61622600c5155/upup/pkg/fi/cloudup/apply_cluster.go#L73

pkg/apis/kops/componentconfig.go
```diff
@@ -500,6 +500,9 @@ type KubeAPIServerConfig struct {
 	// in "iss" claim of issued tokens. This value is a string or URI.
 	ServiceAccountIssuer *string `json:"serviceAccountIssuer,omitempty" flag:"service-account-issuer"`
 
+	// AdditionalServiceAccountIssuers can contain additional service account token issuers.
+	AdditionalServiceAccountIssuers []string `json:"additionalServiceAccountIssuers,omitempty" flag:"service-account-issuer,repeat"`
```
hakman, line 504: Maybe remove the `flag` and just append these after the flags are built, like here:
https://github.com/kubernetes/kops/blob/c6bb40ed328c99ca1490113334f61622600c5155/nodeup/pkg/model/kube_apiserver.go#L576
```suggestion
	AdditionalServiceAccountIssuers []string `json:"additionalServiceAccountIssuers,omitempty"`
```

justinsb:
We discussed in office hours, "additional" makes sense because the first one is special.

/approve

PR #16504, opened 2024-04-29
chore: optimize the method for obtaining gopath

weidongkl:
```bash
$ time(go env | grep GOPATH | cut -f 2 -d '"' | sed 's/ /\\ /g')
/root/go

real    0m0.004s
user    0m0.003s
sys     0m0.004s
$ time(go env GOPATH)
/root/go

real    0m0.003s
user    0m0.003s
sys     0m0.000s

```


hakman (approved):
Nice simplification. Thanks @weidongkl!

PR #16532, opened 2024-05-06
Increase QPS - reduce total test duration

hakuna-matatah:
- Based on recent experiments, 500 QPS/Bursts seems to be achievable out of KCM.

- This should also help reduce overall test duration time.


rifelpet (approved):

dims:
> Roughly we gained ~30mins less compute time on the overall test duration, this should reduce the cost of the test runs further.

Very cool! thanks @hakuna-matatah

PR #16578, opened 2024-05-18
tests: add mock for STS to better test node handshake

justinsb:
- aws node handshake: support both v1 and v2 signatures, default to v1
- tests: add mock for STS to better test node handshake


justinsb:
This includes #16577 (as the first commit)

rifelpet (approved):

PR #16583, opened 2024-05-21
Make ASG Warmpool depend on ASG Lifecycle hook

jim-barber-he:
Fixes: https://github.com/kubernetes/kops/issues/16582

When creating a new ASG with a Warmpool using Lifecycle hooks, the instances that first join the Warmpool when the ASG is created could come up before the Lifecycle hook is in effect.
This can lead to problems such as those instances not calling the Lifecycle hook notification when pressed into service, causing the ASG to terminate them approximately 10 mins after they've been performing work in the cluster.

Setting a dependency on for the Warmpool to take into account any Lifecycle hooks means that the Warmpool won't be created until the hooks are ready.

### DIFF
--- a/upup/pkg/fi/cloudup/awstasks/warmpool.go
+++ b/upup/pkg/fi/cloudup/awstasks/warmpool.go
@@ -44,6 +44,28 @@ type WarmPool struct {
 	AutoscalingGroup *AutoscalingGroup
 }
 
+var _ fi.CloudupHasDependencies = &WarmPool{}
+
+// Warmpool depends on any Lifecycle hooks being in place first.
+func (e *WarmPool) GetDependencies(tasks map[string]fi.CloudupTask) []fi.CloudupTask {
+	var deps []fi.CloudupTask
+
+	// Depend on the ASG.
+	if e.AutoscalingGroup != nil {
+		deps = append(deps, e.AutoscalingGroup)
+	}
+
+	// Depend on any Lifecycle hooks assigned to the ASG.
+	for _, task := range tasks {
+		if l, ok := task.(*AutoscalingLifecycleHook); ok {
+			if l.AutoscalingGroup == e.AutoscalingGroup {
+				deps = append(deps, task)
+			}
+		}
+	}
+	return deps
+}
+
 // Find is used to discover the ASG in the cloud provider.
 func (e *WarmPool) Find(c *fi.CloudupContext) (*WarmPool, error) {
 	ctx := c.Context()


justinsb:
Question: is this race specific to the instances in the WarmPool, or can it apply to the (main) ASG instances also?  I'm wondering if we should be setting LifecycleHookSpecificationList in CreateAutoScalingGroup...

jim-barber-he:
> Question: is this race specific to the instances in the WarmPool, or can it apply to the (main) ASG instances also?

It happens to all the instances that are created at the time the ASG is created and before its lifecycle hook is in effect.
But the instances that skipped the Warmpool and went straight into service won't be reaped by the ASG because the Lifecycle hook wasn't active when they joined (even though they tried to send the lifecycle notification but couldn't).

So all hosts are affected, but only the Warmpool ones are reaped when they later go into service, because by that time the Lifecycle hook is in place on the ASG but they don't seem to attempt to call it.

>  I'm wondering if we should be setting LifecycleHookSpecificationList in CreateAutoScalingGroup...

I attempted to fix it for all hosts using the `LifecycleHookSpecificationList` and had success in that now no hosts started on the ASG until the hook was in place.
But my change had issues if you tried to edit already existing instances groups to add or remove warmpools.
I'm still learning Go and struggled with getting this work properly, so I gave up.
The Slack thread talking about this with Hakman is here: https://kubernetes.slack.com/archives/C3QUFP0QM/p1715733980241579

Since the first instances that skip the Warmpool  when the ASG is created are safe from being reaped by the ASG and it was only the Warmpool instances at risk I figured I'd raise this PR as it gets us out of hot water.
But feel free to close this if you want to develop the `LifecycleHookSpecificationList` approach, but at the moment that's beyond my abilities.

justinsb:
Thanks @jim-barber-he , this makes a lot of sense and is a good fix for warmpools, even if we end up (somehow) _also_ precreating the hooks with LifecycleHookSpecificationList (for non-warmpools).

Thanks also for the detailed comments on the issue (#16582), I'm going to look that over and see if I have any ideas!

/approve
/lgtm

jim-barber-he:
@justinsb 

Are you okay if I raise a PR to cherry pick this back to kOps 1.29 as well?

justinsb:
> Are you okay if I raise a PR to cherry pick this back to kOps 1.29 as well?

That would be wonderful, thank you!  If you haven't seen them, there is a script you can use described here: https://github.com/kubernetes/kops/blob/master/docs/contributing/proposing-a-cherry-pick.md

jim-barber-he:
@justinsb The cherry-pick PR is here: https://github.com/kubernetes/kops/pull/16603

PR #16593, opened 2024-05-30
aws: Add support for using ECR as pull-through image cache

rsafonseca:
This PR introduces a simple way to enable using [ECR as a Pull-through image cache](https://docs.aws.amazon.com/AmazonECR/latest/userguide/pull-through-cache.html), without having to mutate images on the cluster using tools like Kyverno.

Containerd already has support for specifying registry mirrors in kops, but since ECR uses short lived tokens, it's not trivial (or even impossible without adding a few extra hacks on top of it) to use it as a pull-through cache.

This PR also bumps the ecr-credential-provider binary, which before version 1.29.0 specifically tried to parse an ECR repo URL from the image passed, leading to not being possible to enable this feature. This is now resolved in the latest versions.

This PR uses a flag to enable the feature when needed, and adds any server addresses configured in the mirrors to be allowed on the CredentialProviderConfig object that kops configures on the kubelet.

To configure this:

- Create your [pull through cache rule in Amazon ECR](https://docs.aws.amazon.com/AmazonECR/latest/userguide/pull-through-cache-creating-rule.html)
- Set spec.containerd.useECRCredentialsForMirrors to true
- Configure your mirrors in spec.containerd.registryMirrors

example:
```
spec:
  containerd:
    useECRCredentialsForMirrors: true
    registryMirrors:
      docker.io:
        - https://<MY-AWS-ACCOUNT>.dkr.ecr.us-east-1.amazonaws.com/v2/<MY cache rule namespace, e.g docker-hub>
```


rsafonseca:
Any idea why these e2e tests might be failing @hakman?
It seems odd that they never come up and the log dump references not finding an unrelated role that should belong to another job
e.g.
pull-kops-e2e-cni-amazonvpc:
```
0530 19:12:47.346328   14816 dumplogs.go:51] /home/prow/go/src/k8s.io/kops/.build/dist/linux/amd64/kops toolbox dump --name e2e-pr16593.pull-kops-e2e-cni-amazonvpc.test-cncf-aws.k8s.io --dir /logs/artifacts --private-key /tmp/kops/e2e-pr16593.pull-kops-e2e-cni-amazonvpc.test-cncf-aws.k8s.io/id_ed25519 --ssh-user ubuntu
W0530 19:12:59.466784   50124 aws.go:2055] could not find role "masters.e2e-pr16593.pull-kops-e2e-cni-cilium-ipv6.test-cn-40qra0". Resource may already have been deleted: operation error IAM: GetInstanceProfile, https response error StatusCode: 404, RequestID: 013035b7-7eba-43a9-98dd-abf842145433, NoSuchEntity: Instance Profile masters.e2e-pr16593.pull-kops-e2e-cni-cilium-ipv6.test-cn-40qra0 cannot be found.
W0530 19:13:02.485316   50124 aws.go:2055] could not find role "nodes.e2e-pr16593.pull-kops-e2e-cni-cilium-ipv6.test-cncf-47o1f8". Resource may already have been deleted: operation error IAM: GetInstanceProfile, https response error StatusCode: 404, RequestID: 4bc7f2c4-9c38-4723-85bd-b493cf0794c8, NoSuchEntity: Instance Profile nodes.e2e-pr16593.pull-kops-e2e-cni-cilium-ipv6.test-cncf-47o1f8 cannot be found.
```
could something be leaking somewhere? or do you think there's any change here that could be somehow related to this weird behaviour?

hakman:
Please ignore the IPv6 tests

rsafonseca:
All good then, though it looks like pull-kops-e2e-cni-cilium-eni is flaky, passed on retest.
Any thoughts on this PR @hakman ? Does it make sense to you to add the config option under containerd on the CRD? :)

sczelo:
Any update about this PR? Do you need any help or test, to be merged faster?
I built the images locally from this repo and tested it on AWS. It's worked for me.

rsafonseca:
Did you get a chance to review this yet @hakman @johngmyers ? :)

SzilardS:
@rifelpet @hakman @johngmyers is there any way we can help you guys to get it merged before/into the 1.30 release?

elliotdobson:
> This PR also bumps the ecr-credential-provider binary, which before version 1.29.0 specifically tried to parse an ECR repo URL from the image passed, leading to not being possible to enable this feature. This is now resolved in the latest versions.

@rsafonseca can you provide some more details about this? An issue or PR from the source repo?

rsafonseca:
There is no specific issue regarding this, simply the failure logic was changed some commits ago to not error out when it fails to parse an ECR specific url. This is easy to test with the old vs newer binaries, the version that kops currently has in the assets will drop out and not pass the ecr auth token when the referenced image repo is not ECR, so when containerd tries to use ECR as a mirror it has no creds.

rsafonseca:
Hey folks, can anyone take a look at this? :)

nodeup/pkg/model/kubelet.go
```diff
@@ -459,12 +459,21 @@ func (b *KubeletBuilder) addECRCredentialProvider(c *fi.NodeupModelBuilderContex
...
 
 	{
+
 		configContent := `apiVersion: kubelet.config.k8s.io/v1
 kind: CredentialProviderConfig
 providers:
   - name: ecr-credential-provider
     matchImages:
-      - "*.dkr.ecr.*.amazonaws.com"
+`
```
justinsb, line 468: In office hours, we talked about whether UseECRCredentialsForMirrors should be per mirror or whether we only apply this to urls that "look like" ECR.  The (compelling) argument that @rsafonseca made is that people aren't likely to have multiple mirrors so it's not necessary.

👍

rsafonseca:
@hakman I've removed the for loop from the inline string literal as you had suggested in the office hours, and in fact i removed the whole string literal. Overall the number of lines increased, but it should be more readable and easier to manipulate in the future. https://github.com/kubernetes/kops/pull/16593/commits/2852bccdbee7291a8a5a2bc4e32f2a83a88a525b

rsafonseca:
rebased again, the number of files that needed their hashes updated inside tests/integration increased dramatically since the last rebase so the bot just moved this from L to XXL 🫠

pkg/nodemodel/fileassets.go
```diff
@@ -107,7 +107,7 @@ func BuildKubernetesFileAssets(ig model.InstanceGroup, assetBuilder *assets.Asse
 			case kops.CloudProviderAWS:
 				binaryLocation := ig.RawClusterSpec().CloudProvider.AWS.BinariesLocation
 				if binaryLocation == nil {
-					binaryLocation = fi.PtrTo("https://artifacts.k8s.io/binaries/cloud-provider-aws/v1.27.1")
+					binaryLocation = fi.PtrTo("https://artifacts.k8s.io/binaries/cloud-provider-aws/v1.29.0")
```
hakman: Please remove this from the PR.  Will be done as part of https://github.com/kubernetes/kops/pull/17486.

rsafonseca: cool, i'll remove and rebase as soon as you merge #17486 as this PR won't work with v1.27.1 (it was only tested with 1.29, but hopefully no breaking changes in 1.31)

rsafonseca: done

pkg/model/iam/iam_builder.go
```diff
@@ -773,6 +773,10 @@ func addECRPermissions(p *Policy) {
 		"ecr:DescribeRepositories",
 		"ecr:ListImages",
 		"ecr:BatchGetImage",
+		"ecr:ReplicateImage",
+		"ecr:BatchImportUpstreamImage",
+		"ecr:CreateRepository",
+		"ecr:TagResource",
```
hakman, line 798: @rsafonseca Please move these to a separate `addECRPullThroughPermissions(p *Policy)` function and call it only when the flag is set.

hakman (changes requested):

pkg/model/iam/iam_builder.go
hakman: Also, add the new option to this test, to see how the generated output looks like. Other than these, LGTM.
https://github.com/kubernetes/kops/blob/93d0bed9ae9fa26b7cc3f6115a5d4c017515b0ec/tests/integration/update_cluster/containerd-custom/in-v1alpha2.yaml#L13-L22

hakman: Kind ping on this @rsafonseca. Let me know if you have any feedback here.

rsafonseca: oops, sorry i missed your comment earlier, sorting this out now as requested :)

rsafonseca: all done @hakman 👍

hakman: Thanks, looks good to merge. One final squash, please.

hakman:
The tests are broken, so please ignore for now. Will fix them and merge this.

pkg/model/iam/iam_builder.go
rsafonseca: done :)

nodeup/pkg/model/kubelet.go
jim-barber-he: Is this still the case?

Unfortunately ECR pull-through only supports specific registries.
I have 4 other registries that I'd like to handle with harbor, while using ECR for the ones it supports.

PR #16600, opened 2024-06-04
[DOCS] Add docs for Debian 12 on AWS

dannyeuu:
Now there is an official Debian 12 images on AWS


rifelpet (approved):
Confirmed this matches what is published to https://wiki.debian.org/Cloud/AmazonEC2Image/Bookworm

Thanks for the update!

PR #16601, opened 2024-06-05
Make sure we got DNS during disruptions

jabdoa2:
Currently, you might have no DNS anymore if you loose the wrong zone.


hakman:
@jabdoa2 please rebase the PR. My guess is that `hack/update-expected.sh` must also be run to update test outputs.

hakman:
@jabdoa2 Thanks for this, it should be covered in #17472.
/close

PR #16624, opened 2024-06-20
tests: Add test for dl.k8s.io kubernetes version parsing

justinsb:
We saw an error with older kOps versions here, so add a test for this
case.


justinsb:
cc @ameukam

hakman (approved):

PR #16625, opened 2024-06-20
Raise errors earlier if we cannot parse the kube version

justinsb:
If we check this later, in some code-paths we can only panic (because
some code paths assume this check has already been done)


justinsb:
Example error (as pointed out by @ameukam ): https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-gcr-mirror-canary/1803777786335203328

hakman (approved):

Issue #16637, opened 2024-06-26
Finish Migrating To Community CI By August 1st


BenTheElder:
/kind bug

https://github.com/kubernetes/test-infra/blob/master/docs/job-migration-todo.md

kOps is one of the few projects with many remaining jobs, please migrate by August 1st or risk losing CI coverage.

ameukam:
cc @hakman @justinsb @rifelpet

rifelpet:
The remaining jobs that havent been migrated can be categorized into:

* Periodic jobs that publish version markers to a GCS bucket, currently `gs://kops-ci`. Example usage ([1](https://github.com/kubernetes/test-infra/blob/f7aaf47e91f1c97571d423d643f67aa81fe877a6/config/jobs/kubernetes/kops/build_jobs.py#L1779), [2](https://github.com/kubernetes/test-infra/blob/f7aaf47e91f1c97571d423d643f67aa81fe877a6/config/jobs/kubernetes/kops/build_jobs.py#L1804), [3](https://github.com/kubernetes/test-infra/blob/f7aaf47e91f1c97571d423d643f67aa81fe877a6/config/jobs/kubernetes/kops/build_jobs.py#L757))

  I see a `k8s-infra-kops-ci-results` bucket was created years ago: https://github.com/kubernetes/k8s.io/pull/2678 but is not mentioned in any prow job nor in kubetest2-kops. Should we use this bucket or is there a more appropriate project we should create a new bucket in? Currently these jobs use a static AWS credential secret and also have write access to the `kops-ci` GCS bucket. Which prow cluster should we migrate to and how will we acquire both AWS and GCP credentials in one job? We currently [acquire random GCP projects](https://github.com/kubernetes/kops/blob/0d7b2a301b5819e4b5c18e65ca5e20c9633d022c/tests/e2e/kubetest2-kops/deployer/common.go#L70-L86) from boskos but don't use boskos with AWS.

* The presubmit E2E jobs publish artifacts to a `k8s-staging-kops` GCS bucket, used by the E2E cluster provisioned in the test. Can you confirm `gs://k8s-staging-kops` is in a community-owned project? Similar to above we need a way to get credentials for its project while also using AWS credentials for cluster provisioning. If it isn't community-owned then we'll need a new bucket.

[...]

PR #16659, opened 2024-07-06
Support publishing CI artifacts to S3

rifelpet:
Ref: https://github.com/kubernetes/kops/issues/16637 https://github.com/kubernetes/k8s.io/issues/5127 https://github.com/kubernetes/k8s.io/issues/2625

More context in https://github.com/kubernetes/k8s.io/pull/6950#issue-2393322023

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,10 @@ DOCKER_REGISTRY?=gcr.io/must-override
 S3_BUCKET?=s3://must-override/
 UPLOAD_DEST?=$(S3_BUCKET)
 GCS_LOCATION?=gs://must-override
+S3_REGION?=us-east-1
 GCS_URL=$(GCS_LOCATION:gs://%=https://storage.googleapis.com/%)
+S3_HOSTNAME=$(subst %,$(S3_REGION),https://s3.%.amazonaws.com)
+S3_URL=$(S3_BUCKET:s3://%=${S3_HOSTNAME}/%)
 LATEST_FILE?=latest-ci.txt
 GOPATH_1ST:=$(shell go env GOPATH)
 UNIQUE:=$(shell date +%s)
@@ -260,7 +263,7 @@ gcs-upload-and-tag: gsutil gcs-upload
 	echo "${GCS_URL}${VERSION}" > ${UPLOAD}/latest.txt
 	gsutil -h "Cache-Control:private, max-age=0, no-transform" cp ${UPLOAD}/latest.txt ${GCS_LOCATION}${LATEST_FILE}
 
-# gcs-publish-ci is the entry point for CI testing
+# gcs-publish-ci and s3-publish-ci are the entry points for CI testing
 # In CI testing, always upload the CI version.
 .PHONY: gcs-publish-ci
 gcs-publish-ci: VERSION := ${KOPS_CI_VERSION}+${GITSHA}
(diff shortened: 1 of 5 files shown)


tests/e2e/kubetest2-kops/deployer/build.go
```diff
@@ -34,6 +36,7 @@ import (
 const (
 	defaultJobName = "pull-kops-e2e-kubernetes-aws"
 	defaultGCSPath = "gs://k8s-staging-kops/pulls/%v/pull-%v"
+	defaultS3Path  = "s3://k8s-infra-kops-ci-results/pulls/%v/pull-%v"
```
rifelpet, line 39: Created by https://github.com/kubernetes/k8s.io/pull/6950

upodroid:
We want kubetest2-kops to dynamically create a temporary s3 bucket when the presubmit is running, similar to this code
https://github.com/kubernetes/kops/blob/master/tests/e2e/kubetest2-kops/gce/gcs.go

I made the IAM changes, but it allows presubmits to write to our CI bucket which isn't ideal(existing behaviour we should change).

rifelpet:
Yes I see the migrated job on your PR passed, using k8s-staging-kops:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/16648/pull-kops-e2e-k8s-aws-calico-k8s-infra/1809653969442574336

I believe this means we could migrate all presubmit jobs to use this, but you're right giving presubmit jobs write permissions to the same bucket used for post-merge artifacts like version markers is not great. 

We have the `k8s-infra-kops-ci-results` bucket created in https://github.com/kubernetes/k8s.io/pull/2678, is that in the correct project and can we use it for either presubmit artifacts or version markers, and then revoke permissions for presubmits to whichever bucket we choose for version markers?

upodroid:
For now, let's retain the existing behaviour as AWS presubmits have been writing to the k8s-staging-kops bucket and let's try to switch to dynamic buckets on AWS.

As for gs://k8s-infra-kops-ci-results, I don't want to use this bucket and we should delete it.

Makefile
```diff
@@ -271,6 +274,15 @@ gcs-publish-ci: gsutil version-dist-ci
 	echo "${GCS_URL}/${VERSION}" > ${UPLOAD}/${LATEST_FILE}
 	gsutil -h "Cache-Control:private, max-age=0, no-transform" cp ${UPLOAD}/${LATEST_FILE} ${GCS_LOCATION}
 
+.PHONY: s3-publish-ci
+s3-publish-ci: VERSION := ${KOPS_CI_VERSION}+${GITSHA}
+s3-publish-ci: version-dist-ci
```
justinsb, line 279: We do also have `hack/upload`, which _might_ make this easier?

tests/e2e/kubetest2-kops/builder/build.go
```diff
@@ -83,21 +85,37 @@ func (b *BuildOptions) Build() (*BuildResults, error) {
...
+
 	cmd.SetDir(b.KopsRoot)
 	exec.InheritOutput(cmd)
 	if err := cmd.Run(); err != nil {
 		return nil, err
 	}
 
 	// Get the full path (including subdirectory) that we uploaded to
-	// It is written by gcs-publish-ci to .build/upload/latest-ci.txt
+	// It is written by the *-publish-ci make tasks to .build/upload/latest-ci.txt
```
justinsb, line 118: It might be nice to spell out the two tasks here (assuming we can't combine them), to make it more explicit and to help with searching...

justinsb (commented):
This actually LGTM.  We're probably holding this PR for now (given that the test-infra one is closed), but I think a good reason to upload artifacts to S3 would be to cut down on bandwidth etc costs for testing.  But I guess they aren't that large....

upodroid:
We do need to ship this to fix the AWS presubmits publishing to the k8s-staging-kops bucket.

PR #16662, opened 2024-07-08
Set Ubuntu 24.04 as default in k8s 1.31

rifelpet:
kops 1.30 is already far enough into the release cycle that I think it makes sense to wait for 1.31 for this. The 24.04 e2e testing has been fine so far.


justinsb:
SGTM ... cc @hakman for consensus

hakman:
> SGTM ... cc @hakman for consensus

Same, wanted to hear from you first.

PR #16693, opened 2024-07-19
feat(cluster-autoscaler) Implement emitPerNodegroupMetrics parameter

alexandresavicki:
## Context:

/kind feature

The current use case that these metrics are needed for is to create alerts when the number of nodes in the cluster reach a certain percentage of the max allowed nodes for example.

## References:
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca

### DIFF
--- a/docs/addons.md
+++ b/docs/addons.md
@@ -71,6 +71,7 @@ spec:
     enabled: true
     expander: least-waste
     balanceSimilarNodeGroups: false
+    emitPerNodegroupMetrics: false
     awsUseStaticInstanceList: false
     scaleDownUtilizationThreshold: 0.5
     skipNodesWithCustomControllerPods: true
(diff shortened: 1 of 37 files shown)


hakman:
@alexandresavicki Please run `run hack/update-expected.sh` to update some generated files.

fernandrone (approved):

alexandresavicki:
@olemarkus @johngmyers can you please review this PR?

olemarkus:
Looks good to me.

Could you add a small note to the release notes?

alexandresavicki:
@olemarkus done! Can you please take a look?

alexandresavicki:
@olemarkus Thanks for merging.
Can we cherry pick to `1.29.0` ?

PR #16701, opened 2024-07-27
test: Use latest K8s build for scale tests

hakuna-matatah:
Want to test 1.31.beta to see if it works on AWS.

### DIFF
--- a/tests/e2e/scenarios/scalability/run-test.sh
+++ b/tests/e2e/scenarios/scalability/run-test.sh
@@ -27,7 +27,7 @@ fi
 echo "CLUSTER_NAME=${CLUSTER_NAME}"
 
 if [[ -z "${K8S_VERSION:-}" ]]; then
-  K8S_VERSION="$(curl -s -L https://dl.k8s.io/release/stable.txt)"
+  K8S_VERSION="$(curl -s -L https://dl.k8s.io/release/latest.txt)"
 fi
 
 # A temp patch for kubetest2 https://github.com/kubernetes-sigs/kubetest2/pull/256


hakman:
presubmit-kops-aws-small-scale-amazonvpc-using-cl2

hakman (approved):

hakuna-matatah:
cc: @dims @shyamjvs @BenTheElder

hakuna-matatah:
cc: @dims @shyamjvs @BenTheElder

dims:
@hakuna-matatah next time we could add `K8S_VERSION` in the test-infra ci job definition too right?

hakuna-matatah:
Looks like test run succeeded but it looks like Prow job is stuck in running state - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/16701/presubmit-kops-aws-scale-amazonvpc-using-cl2/1818245175466528768

BenTheElder:
> 
Looks like test run succeeded but it looks like Prow job is stuck in running state - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/16701/presubmit-kops-aws-scale-amazonvpc-using-cl2/1818245175466528768


that's probably https://github.com/kubernetes-sigs/prow/issues/210

BenTheElder:
> I think we could also facilitate that, but wondering why would we want to do that ? Do we want to run for many different versions ? Or just the latest ?

it's also nice to be explicit in the job config, so we have more confidence that e.g. release-master-informing jobs _actually_ run code from master, though we should always confirm by investigating the uploaded artifacts

Issue #16697, opened 2024-07-24
Installation fails on GCE + RHEL or Rocky Linux


jsafrane:
/kind bug

Default RHEL 9 and Rocky Linux 9 images on GCE do not have cloud-init installed, while kops expects it there. A cluster fails to install on those VMs. CentOS and RHEL 8 / Rocky 8 most probably have the same issue.

It seems that kOps got broken by https://github.com/kubernetes/kops/pull/15607 - RHEL / Rocky VMs support `startup-script`, but they don't have cloud-init installed and don't support scripts in `user-data` out of the box. I can fix the installation by ssh to the VMs and `yum install -y cloud-init; reboot`.

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Recent master, `1.30.0-beta.1 (git-v1.30.0-beta.1-46-gae8fa06e07`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

v1.30.2 / amd64

**3. What cloud provider are you using?**

GCE

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

Just a default installation on GCE using either Rocky Linux or RHEL (and probably CentOS too):

```
kops create cluster jsafrane.mydomain.com --zones us-central1-a --state ${KOPS_STATE_STORE}/ --project=${PROJECT}  --cloud=gce --yes  --image=rocky-linux-cloud/rocky-linux-9-optimized-gcp-v20240717
```

**5. What happened after the commands executed?**

Both control-plane and worker start, however, they just boot and don't do anything useful. There is no cloud-init installed on GCE RHEL or Rocky images.

Machine logs:
```
Jul 24 11:09:02 control-plane-us-central1-a-tgqq google_metadata_script_runner[2239]: Starting startup scripts (version dev).
Jul 24 11:09:02 control-plane-us-central1-a-tgqq google_metadata_script_runner[2239]: No startup scripts to run.
```
and just does nothing useful.

**6. What did you expect to happen?**

Cluster installs.

**7. Please provide your cluster manifest. Execute
  `kops get --name my.example.com -o yaml` to display your cluster manifest.
  You may want to remove your cluster name and other sensitive information.**

```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: Cluster
metadata:
  creationTimestamp: "2024-07-24T11:06:45Z"
  name: jsafrane.gcp.devcluster.openshift.com
[...]

PR #16705, opened 2024-07-29
gce: Add option to use startup script instead of user-data

hakman:
Partial revert of https://github.com/kubernetes/kops/pull/15607

Fixes https://github.com/kubernetes/kops/issues/16697

/hold for feedback
/cc @justinsb @rifelpet @jsafrane


justinsb:
Awesome, thanks for fixing this!

/approve
/lgtm

PR #16717, opened 2024-08-01
spotinst: Bump k8s cluster controller to v2

yehielnetapp:
update controller v2, inform user on conflict with AWS nodeTerminationHandler and Ocean

### DIFF
--- a/upup/models/cloudup/resources/addons/spotinst-kubernetes-cluster-controller.addons.k8s.io/v1.14.0.yaml.template
+++ b/upup/models/cloudup/resources/addons/spotinst-kubernetes-cluster-controller.addons.k8s.io/v1.14.0.yaml.template
@@ -1,67 +1,265 @@
+#helm upgrade --install --wait ocean-controller spot/ocean-kubernetes-controller \
+#--namespace "kube-system" --create-namespace \
+#--set spotinst.account="${SPOTINST_ACCOUNT}" \
+#--set spotinst.clusterIdentifier="prefeat.yehiel.ek8s.com" \
+#--set spotinst.token="${SPOTINST_TOKEN}" --set spotinst.disableAutoUpdate="true"  --dry-run > helmdrupdry.yam
+# Source: ocean-kubernetes-controller/templates/configmap.yaml
+---
 apiVersion: v1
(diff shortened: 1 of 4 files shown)


upup/models/cloudup/resources/addons/spotinst-kubernetes-cluster-controller.addons.k8s.io/v1.14.0.yaml.template
```diff
@@ -1,67 +1,265 @@
...
+#--set spotinst.account="${SPOTINST_ACCOUNT}" \
+#--set spotinst.clusterIdentifier="prefeat.yehiel.ek8s.com" \
+#--set spotinst.token="${SPOTINST_TOKEN}" --set spotinst.disableAutoUpdate="true"  --dry-run > helmdrupdry.yam
+# Source: ocean-kubernetes-controller/templates/configmap.yaml
+---
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: spotinst-kubernetes-cluster-controller-config
+  name: ocean-controller-ocean-kubernetes-controller
```
olemarkus, line 11: As it is now, the renaming here will create new resources while leaving the ones with the old names behind.
It may be best to avoid renaming if possible, otherwise enable pruning old resources by setting `BuildPrune` on the addon in `bootstrapchannelbuilder.go`.

You probably want to thoroughly test migrating a cluster through this change.
This change also warrants a rather visible notice in the release notes.

olemarkus (changes requested):

upup/models/cloudup/resources/addons/spotinst-kubernetes-cluster-controller.addons.k8s.io/v1.14.0.yaml.template
yehielnetapp: adding BuildPrune remove old resources, we notify client about clean up he need to do after upgrade.

olemarkus:
/lgtm
/assign @hakman 

I'd like more eyes on this one.

hakman (approved):
Thanks @yehielnetapp & @olemarkus!

PR #16737, opened 2024-08-06
Fixes already existing nodes behaviour in kops controller

### LINKED ISSUE #16736: kOps leave connections open in OpenStack when rebooting instance
/kind bug

 this information.**

master

 a `kops` flag.**

1.30.x

openstack

if we restart any instance, it will start making new connections in loop.

example:

```
root@master-helpa-gvapl8:/home/ubuntu# netstat -n|grep 3988|grep ESTABLISHED|wc -l
28213
```

We are seeing this in many kOps openstack clusters. Port 3988 is the kops-controller port that is used in control plane. All nodes are connecting to it and retrying forever (and making new connections)

jauru:
fixes #16736


zetaab (approved):
/approve
/lgtm

thanks for fixing

Issue #16159, opened 2023-12-07
ig nodeLabels not passed to kubernetes nodes in Hetzner


lukasredev:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
`v1.28.1`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
`v1.27.8`

**3. What cloud provider are you using?**
Hetzner

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
Create the cluster
```bash
kops create cluster --name=my-cluster.lukasre.k8s.local \
  --ssh-public-key=path-to-pub --cloud=hetzner --zones=fsn1 \
  --image=ubuntu-20.04 --networking=calico --network-cidr=10.10.0.0/16
```
Add a new instance group with different node labels
```bash
kops create ig nodes-immich-fsn1 --subnet fsn1
``` 

Edit the instance group with the following config:
```yaml
apiVersion: kops.k8s.io/v1alpha2
kind: InstanceGroup
metadata:
  creationTimestamp: "2023-12-06T20:46:51Z"
  generation: 3
  labels:
    kops.k8s.io/cluster: my-cluster.lukasre.k8s.local
  name: nodes-immich-fsn1
spec:
  image: ubuntu-22.04
  kubelet:
    anonymousAuth: false
    nodeLabels:
      lukasre.ch/instancetype: immich
      node-role.kubernetes.io/node: ""
  machineType: cx21
  manager: CloudGroup
  maxSize: 1
  minSize: 1
  nodeLabels:
    kops.k8s.io/instancegroup: nodes-immich-fsn1
    lukasre.ch/instancetype: immich
  role: Node
  subnets:
  - fsn1
```

Update the cluster (including forcing a rolling update) with
```bash
kops update cluster --yes
kops rolling-update cluster --yes --force
```

**5. What happened after the commands executed?**
Commands are successful, but node labels are not added.

The yaml representation of the newly created node is the following (only metadata)
``` yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    alpha.kubernetes.io/provided-node-ip: 10.10.0.7
    csi.volume.kubernetes.io/nodeid: '{"csi.hetzner.cloud":"40260390"}'
    node.alpha.kubernetes.io/ttl: "0"
    projectcalico.org/IPv4Address: 10.10.0.7/32
[...]

PR #16739, opened 2024-08-07
Propagate IG NodeLabels to k8s nodes in Hetzner

rifelpet:
Fixes https://github.com/kubernetes/kops/issues/16159

This adds Hetzner server labels for the instance group nodeLabels. kops-controller then recognizes these server labels and sets them on the associated k8s node.

I'm marking this WIP as we may need to change this approach due to label length limits in the Hetzner API:

https://docs.hetzner.com/cloud/servers/getting-started/creating-a-server/

> Both key and value must be 63 characters or less, beginning and ending with an alphanumeric character and alphanumerics can be used inbetween

The Hetzner CCM mentions support for the `node.kubernetes.io/exclude-from-external-load-balancers` label included in the integration test output changes: https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/182#issuecomment-830577642 which is 55 characters on its own, so with the prefix I've added it exceeds the length limit. I considered storing these elsewhere like in userdata and having kops-controller parse them out of the `hcloud.Server`'s userdata, except hcloud doesn't expose userdata in read operations, it is a write-only field.

Therefore I think our only option is to use a shorter prefix

/wip

### DIFF
--- a/upup/pkg/fi/cloudup/hetzner/cloud.go
+++ b/upup/pkg/fi/cloudup/hetzner/cloud.go
@@ -41,7 +41,7 @@ const (
 	TagKubernetesInstanceUserData    = "kops.k8s.io/instance-userdata"
 	TagKubernetesInstanceNeedsUpdate = "kops.k8s.io/needs-update"
 	TagKubernetesVolumeRole          = "kops.k8s.io/volume-role"
-	TagKubernetesNodeLabelPrefix     = "kops.k8s.io/node-label/"
+	TagKubernetesNodeLabelPrefix     = "node-label.kops.k8s.io."
 )
 
 // HetznerCloud exposes all the interfaces required to operate on Hetzner Cloud resources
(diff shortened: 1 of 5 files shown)


upup/pkg/fi/cloudup/hetzner/cloud.go
```diff
@@ -41,6 +41,7 @@ const (
 	TagKubernetesInstanceUserData    = "kops.k8s.io/instance-userdata"
 	TagKubernetesInstanceNeedsUpdate = "kops.k8s.io/needs-update"
 	TagKubernetesVolumeRole          = "kops.k8s.io/volume-role"
+	TagKubernetesNodeLabelPrefix     = "kops.k8s.io/node-label/"
```
hakman, line 44: Hetzner does not allow multiple slashes for labels:
> Valid label keys have two segments: an optional prefix and name, separated by a slash ( / ).

https://docs.hetzner.cloud/#labels

```suggestion
	TagKubernetesNodeLabelPrefix     = "kops.k8s.io/node-label."
```

hakman (changes requested):

upup/pkg/fi/cloudup/hetzner/cloud.go
rifelpet: this wont work because the node labels themselves may have slashes:

`kops.k8s.io/node-label.node-role.kubernetes.io/node`

I just pushed a potential fix. Its not great but given the label requirements I'm not sure we can do any better.

hakman (approved):

PR #16745, opened 2024-08-09
Fix table not format of docs

lou-lan:
This table is not formatted, and the rendering is wrong, which affects reading. Just format the table.

截屏2024-08-09 18 18 01


hakman:
Thanks @lou-lan, should help with readability.
/lgtm
/approve

PR #16749, opened 2024-08-15
WIP: Try not installing cgroupfs-mount

justinsb:
It seems to cause a restart of kops-configuration, right in the middle of an apt-get install


justinsb:
I don't think the issue is necessarily this particular package, there's an async effect here when we restart packages.

We _probably_ don't need this package, but it isn't the cause of the problems here.

PR #16751, opened 2024-08-15
WIP: Try bumping regex skiplist version to include 1.31

justinsb:

PR #16755, opened 2024-08-15
nodeup: if apt-get tells us to run dpkg configure, run it

justinsb:
This is a fallback against some of nastier edge cases we can fall into
otherwise.

### DIFF
--- a/tests/e2e/pkg/tester/skip_regex.go
+++ b/tests/e2e/pkg/tester/skip_regex.go
@@ -105,7 +105,7 @@ func (t *Tester) setSkipRegexFlag() error {
 
 	if cluster.Spec.LegacyCloudProvider == "digitalocean" {
 		skipRegex += "|Services.should.respect.internalTrafficPolicy=Local.Pod.and.Node,.to.Pod"
-		if k8sVersion.Minor < 31 {
+		if k8sVersion.Minor < 32 {
 			// https://github.com/kubernetes/kubernetes/issues/121018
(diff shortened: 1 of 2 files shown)


hakman:
This looks good. Probably https://github.com/kubernetes/kops/pull/16751 should be combined with this to pass all the tests.

justinsb:
Added the skiplist here, but I _think_ there's something happening on GCE also which might still fail....

hakman:
> Added the skiplist here, but I _think_ there's something happening on GCE also which might still fail....

One step closer.

rifelpet:
@justinsb I believe this is causing some regressions in GCE:

https://testgrid.k8s.io/kops-network-plugins#kops-gce-cni-cilium-k8s-ci

A node is failing to join the cluster, with:

```
W0828 12:39:10.230865    4618 executor.go:141] error running task "Package/cgroupfs-mount" (6m19s remaining to succeed): error installing package "cgroupfs-mount": exit status 100: Reading package lists...
Building dependency tree...
Reading state information...
cgroupfs-mount is already the newest version (1.4).
0 upgraded, 0 newly installed, 0 to remove and 103 not upgraded.
1 not fully installed or removed.
Need to get 0 B/316 kB of archives.
After this operation, 0 B of additional disk space will be used.
dpkg: error processing package chrony (--configure):
 package is in a very bad inconsistent state; you should
 reinstall it before attempting configuration
Errors were encountered while processing:
 chrony
needrestart is being skipped since dpkg has failed
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

PR #16758, opened 2024-08-18
WIP Upgrade Karpenter to 1.0.0

rifelpet:
TODO list:
- [ ] Set the EC2NodeClass userdata - this is a bit tricky because userdata is normally generated via `fi.CloudupTask` dependencies but none of the bootstrapchannelbuilder code uses `fi.Resource`
- [ ] Update awsmodel to not add awstasks.LaunchTemplates for Karpenter instance groups
- [ ] Review `rolling-update`, `delete cluster`, and `toolbox dump` code to confirm it works without kops-managed launch templates
- [ ] Add release note and update documentation
- [ ] Figure out why the --kubeconfig flag is being removed from the CLI options
- [ ] Reuse the root volume defaulting logic from pkg/awsmodel

Note that this doesn't follow the required karpenter upgrade procedure - therefore existing kops clusters with karpenter enabled may not have an upgrade path to this version of karpenter. We'll need to figure out a low-effort workaround to include in the upgrade instructions.


tests/integration/update_cluster/karpenter/data/aws_s3_object_minimal.example.com-addons-karpenter.sh-k8s-1.19_content
```diff
@@ -2050,15 +4332,47 @@ metadata:
...
-  launchTemplate: karpenter-nodes-default.minimal.example.com
-  subnetSelector:
-    kops.k8s.io/instance-group/karpenter-nodes-default: '*'
-    kubernetes.io/cluster/minimal.example.com: '*'
+  amiFamily: Custom
+  amiSelectorTerms:
+  - id: ami-12345678
+  blockDeviceMappings:
+  - deviceName: /dev/xvda
+    ebs: {}
```
rifelpet, line 4340: looks like the awsmodel's default root volume fields aren't being set on the InstanceGroup that we're referencing here so we'll need to reuse that defaulting logic.

justinsb:
I had a look at the userdata problem.  As part of the bare metal WIP, I've refactored the bootstrap script building so it isn't as tricky as it used to be: https://github.com/kubernetes/kops/blob/master/pkg/commands/toolbox_enroll.go#L401

I had a go at piping that through, it isn't too bad except for the wellKnownAddresses.  But getting those would indeed be tricky...

But - I was wondering if we should build this in the template anyway, vs uploading the kops InstanceGroups and having kops-controller (for example) expand them.  This is what I've been playing with for cluster-api, and it seems to work I think (though I should probably revive and clean that up)

Another advantage of doing it that way is that users could then dynamically create InstanceGroups.  The gotcha is that those would not be stored in the state store so would have to be managed differently, but we don't _have_ to support (or encourage) dynamic InstanceGroups.

PR #16768, opened 2024-08-24
refactor: OptionsBuilder works on kops.Cluster

justinsb:
Previously it worked on ClusterSpec, but that means we can't read the
labels or annotations.  We want to use those labels/annotations to
experiment with new cloud providers (e.g. bare metal) before adding
this to the API.

Introduce generics so we also get type-safety.

### DIFF
--- a/pkg/model/components/apiserver.go
+++ b/pkg/model/components/apiserver.go
@@ -34,11 +34,11 @@ type KubeAPIServerOptionsBuilder struct {
 	*OptionsContext
 }
 
-var _ loader.OptionsBuilder = &KubeAPIServerOptionsBuilder{}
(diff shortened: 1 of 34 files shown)


justinsb:
I _think_ `operation error ec2imds: GetRegion, canceled, context deadline exceeded` is unrelated (???)

/retest

hakman (approved):

justinsb:
/retest

I am confused as to how this could be failing when initializing DNS, but other jobs are not failing.... retesting one more time to validate my assumptions!

rifelpet:
I think the imds error is legitimate and a result of https://github.com/kubernetes/kops/pull/16647 but I'm not sure why it is happening inconsistently.

PR #16769, opened 2024-08-24
experiment: get test results for cilium without kube-proxy replacement

justinsb:


hakman:
Not needed for now.
/close

PR #16774, opened 2024-08-25
Update deps to K8s v0.31.0

hakman:
The "Update Dependencies" workflow has been failing for some time:
https://github.com/kubernetes/kops/actions/workflows/update-deps.yml

/cc @rifelpet @justinsb


tests/e2e/scenarios/upgrade-ha-leader-migration/run-test.sh
rifelpet, line 1: why is this being removed?

hakman: leader migration is no longer needed and was generating some dependencies mess.

PR #16778, opened 2024-08-25
dns: use resolved region rather than re-resolving every time

justinsb:
This should allow the fallback behaviour to work as expected.

### DIFF
--- a/dnsprovider/pkg/dnsprovider/providers/aws/route53/route53.go
+++ b/dnsprovider/pkg/dnsprovider/providers/aws/route53/route53.go
@@ -67,6 +67,8 @@ func newRoute53() (*Interface, error) {
 	imdsRegionResp, err := imdsClient.GetRegion(ctx, &imds.GetRegionInput{})
 	if err != nil {
 		klog.V(4).Infof("Unable to discover region by IMDS, using SDK defaults: %s", err)
+		// Don't use imdsClient if it's erroring (we're probably not running on EC2 here, e.g. kops update)
+		imdsClient = nil
 	} else {
 		region = imdsRegionResp.Region
 	}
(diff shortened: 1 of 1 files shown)


justinsb:
/hold

As I'm not very sure about this

rifelpet:
I want to make sure we dont reintroduce https://github.com/kubernetes/kops/issues/16645

My initial idea for this fix was to only include the `awsconfig.WithEC2IMDSRegion()` call if IMDS is available - that is if `imdsClient.GetRegion()` returns a nil error.

justinsb:
> My initial idea for this fix was to only include the awsconfig.WithEC2IMDSRegion() call if IMDS is available - that is if imdsClient.GetRegion() returns a nil error.

Ah - that does sound like a better approach than mine here!  Do you want to send a PR, or if you want I can adapt this one?

rifelpet:
Feel free to test my theory in this PR :)

rifelpet:
Looks like this didn't fix the problem:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-addon-resource-tracking/1827929845212909568

```
Error: error reading cluster configuration "e2e-c0d41e2af2-13250.tests-kops-aws.k8s.io": error reading s3://k8s-kops-ci-prow-state-store/e2e-c0d41e2af2-13250.tests-kops-aws.k8s.io/config: Unable to list AWS regions: NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, .
CredentialsEndpointError: invalid endpoint host, "169.254.170.23", only loopback hosts are allowed.
Error: exit status 1 
```

PR #16782, opened 2024-08-26
refactor: Move GetCloudProvider to cluster

justinsb:
This lets us use labels (or annotations), meaning we can experiment
with different clouds without changing the API.

We also add initial (experimental/undocumented) support for exposing a "Metal" provider.

### DIFF
--- a/pkg/apis/kops/v1alpha2/conversion.go
+++ b/pkg/apis/kops/v1alpha2/conversion.go
@@ -419,7 +419,27 @@ func Convert_kops_ClusterSpec_To_v1alpha2_ClusterSpec(in *kops.ClusterSpec, out 
 	if out.API.IsEmpty() {
 		out.LegacyAPI = nil
 	}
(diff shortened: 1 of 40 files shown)


pkg/apis/kops/v1alpha2/conversion.go
```diff
@@ -419,7 +419,27 @@ func Convert_kops_ClusterSpec_To_v1alpha2_ClusterSpec(in *kops.ClusterSpec, out
 	if out.API.IsEmpty() {
 		out.LegacyAPI = nil
 	}
-	out.LegacyCloudProvider = string(in.GetCloudProvider())
+	if in.CloudProvider.AWS != nil {
```
justinsb, line 422: We don't support experimental cloud providers here (we don't have the cluster, and thus don't have the labels), but I think that's OK.

hakman (approved):

PR #16789, opened 2024-08-30
metal: more functions to enable `kops update cluster`

justinsb:
We still have to add VMs, but this is another step closer.


justinsb:
Turns out to be mostly s3 mock, but it looks like this might be the functionality we need to bring up a cluster (at least until we do `kops delete cluster`)

hakman (approved):

PR #16794, opened 2024-08-30
vfs: fix file replacement for SFTP

justinsb:
This makes it consistent with other VFS systems, where we can replace existing files.


justinsb:
This should not be relevant to cilium

/retest

PR #16816, opened 2024-09-07
Discover a bastion load balancer and use it for dumping artifacts

rifelpet:
Currently the e2e prow jobs for ipv6 clusters can't fetch artifacts:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-cni-cilium-ipv6/1832137312717246464#1:build-log.txt%3A1954

In my local testing, I can't reach the bastion host directly but can SSH through the bastion NLB that targets the instance. This adds support in `kops toolbox dump` to use the bastion NLB if it exists.

### DIFF
--- a/cmd/kops/toolbox_dump.go
+++ b/cmd/kops/toolbox_dump.go
@@ -211,10 +211,18 @@ func RunToolboxDump(ctx context.Context, f commandutils.Factory, out io.Writer, 
 		}
 
 		// look for a bastion instance and use it if exists
+		// Prefer a bastion load balancer if exists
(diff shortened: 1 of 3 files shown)


rifelpet:
Artifacts of ipv6-only nodes are now being dumpted: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kops/16816/pull-kops-e2e-cni-cilium-ipv6/1832401672555466752/artifacts/

hakman (approved):

PR #16820, opened 2024-09-10
tests: build and push locally for metal tests

justinsb:
Because we now push much larger files, replace our in-memory storage with disk-backed storage.


justinsb:
/retest

(cilium failures should not be related)

hakman (approved):

PR #16855, opened 2024-09-24
correct hubble tls file names as mapped from secret hubble-server-certs

mwiede:
I realized, that after upgrading to Kops master branch, the cilium-relay deployment was not ok.

Further analyzing brought me to https://github.com/cilium/cilium/issues/31516 where people say, there is a problem with dns, but not for me.

`hubble-relay` connects though `hubble-peer`-service to the hubble processes on port 4244.

But the problem was, that Hubble server inside cilium pods did not start, because the files mounted from `hubble-server-certs` have gotten a new name after upgrade to cilium 1.16.

`cilium status` inside cilium pod reported:
```
Hubble:                  Warning Server not initialized
```

while `hubble status` reported
```
Healthcheck (via unix:///var/run/cilium/hubble.sock): Ok
```


rifelpet:
Thanks for the fix @mwiede, can you also run `./hack/update-expected.sh` and commit the resulting changes?

rifelpet:
These tests have been flakey recently, so the failures are unrelated.

/retest

PR #16868, opened 2024-09-28
aws: Update VPC CNI to v1.18.5

hakman:
/cc @rifelpet @justinsb


nodeup/pkg/model/networking/amazon-vpc-routed-eni.go
```diff
@@ -68,7 +68,7 @@ ManageForeignRoutingPolicyRules=no
 	// Running Amazon VPC CNI on Ubuntu 22.04 and later or any version of al2023 requires
```
rifelpet, line 68:
```suggestion
	// Running Amazon VPC CNI on Ubuntu 22.04 or any version of al2023 requires
```

rifelpet (approved):

PR #16892, opened 2024-10-11
fix: print the correct field path when spec.api.publicName is missing

justinsb:
We introduce a field-mapping layer to help us map v1alpha2 -> v1alpha3 and vice-versa


hakman (approved):

pkg/apis/kops/fieldmap.go
```diff
@@ -0,0 +1,71 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kops
+
+// Because we have multiple versions of the API, fields hvae different paths in different API versions.
```
hakman, line 19:
```suggestion
// Because we have multiple versions of the API, fields have different paths in different API versions.
```

PR #16905, opened 2024-10-14
Release 1.31.0-alpha.1

rifelpet:


hakman (approved):

rifelpet:
@hakman can you tell what is failing here? I'm not up to speed on the metal support.

justinsb:
The bare metal scenario test is pretty early, and I don't think I intended for it to block unrelated PRs.  It brings up a few VMs and tries to get to the point where the apiserver is responding and some nodes are registered.  It doesn't have many _tests_ per-se, in this case the issue is that nodeup is timing out for one reason or another.

I'll give it a retry, but we can probably bypass the test and merge, as everything else is working.

hakman:
@justinsb @rifelpet Probably we need to add the `tests-e2e-scenarios-bare-metal` test to this list:
https://github.com/kubernetes/test-infra/blob/6884c41c969f516bf5ff7aa3ad2582638fabc169/config/prow/config.yaml#L352

justinsb:
No objection to adding it to the list, but doesn't the fact that it's not on the list mean that it's not blocking?

(i.e. maybe it's not merging because it's missing lgtm)

justinsb:
Changes LGTM by the way :-)

hakman:
> No objection to adding it to the list, but doesn't the fact that it's not on the list mean that it's not blocking?
> 
> (i.e. maybe it's not merging because it's missing lgtm)

We had LGTM until earlier today, but did not help.

justinsb:
I force-merged it 🤞

PR #16911, opened 2024-10-22
openstack: Upgrade to Gophercloud v2

pierreprinetti:
Gophercloud v1 is frozen and lacks a few improvements that can be found in the v2 branch. This patch bumps Gophercloud to the current v2 branch and makes the required modifications to the kops code.

Essentially:

* Add a context argument to Gophercloud functions
* Refactor scheduler hints for server creation

Incidentally, Gophercloud also got stricter in the accepted response codes. I have changed the mocked `servers.Create` call to return `202` instead of `201`: [docs](https://docs.openstack.org/api-ref/compute/?expanded=restore-soft-deleted-instance-restore-action-detail#create-server)


pierreprinetti:
Despite the huge LOC change count, this patch should be relatively straightforward, as it’s almost exclusively the same regex applied all over. Let me know if I can help with the review @zetaab @olemarkus.

hakman:
> Despite the huge LOC change count, this patch should be relatively straightforward, as it’s almost exclusively the same regex applied all over. Let me know if I can help with the review @zetaab @olemarkus.

Thank you @pierreprinetti, the review seems pretty straightforward.
I will wait for @zetaab's final approval, as we don't have any e2e tests for OpenStack.
/lgtm

zetaab (approved):
/approve

PR #16920, opened 2024-10-25
Add support for configuring environment variables on kube-apiserver

rsafonseca:
This PR allows configuring environment variables for the kube-apiserver.
This presents high value in order to be able to control a number of runtime settings, such as the ones in the code examples.

main motivation:
The apiserver is a memory hungry beast, and currently we have no way of controlling garbage collection settings, which can easily lead to OOM, or the need to run on much larger instances to prevent that, making the kops control plane very costly.

future proofing:
This provides the flexibility to add any other unforeseen variables that could be useful for the kops cluster admin


hakman:
Cool change 😄 
/ok-to-test
/lgtm
/cc @rifelpet @justinsb

rifelpet (approved):

PR #16922, opened 2024-10-26
WIP Allow updating the cluster one instance group at a time

hakman:
/cc @justinsb @rifelpet

### DIFF
--- a/cmd/kops/update_cluster.go
+++ b/cmd/kops/update_cluster.go
@@ -27,6 +27,8 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/cmd/kops/util"
@@ -75,6 +77,14 @@ type UpdateClusterOptions struct {
 	user          string
 	internal      bool
 
+	// InstanceGroups is the list of instance groups to update;
+	// if not specified, all instance groups will be updated
+	InstanceGroups []string
+
+	// InstanceGroupRoles is the list of roles we should update
+	// if not specified, all instance groups will be updated
+	InstanceGroupRoles []string
+
 	Phase string
 
 	// LifecycleOverrides is a slice of taskName=lifecycle name values.  This slice is used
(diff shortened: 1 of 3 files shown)


cmd/kops/update_cluster.go
```diff
@@ -285,24 +306,75 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Up
 		lifecycleOverrideMap[taskName] = lifecycleOverride
 	}
 
+	list, err := clientset.InstanceGroupsFor(cluster).List(ctx, metav1.ListOptions{})
```
justinsb, line 309: Nit: might be nice to split this into a function?

justinsb:
Remind me - I think you said there's a gotcha with this approach @hakman ?

hakman:
> Remind me - I think you said there's a gotcha with this approach @hakman ?

@justinsb If you try to apply the config to a single IG, the KubeletConfig is removed.

justinsb:
Looked at this for a while, and I think I know what is happening!

So the issue is:

```
// upgradeSpecs ensures that fields are fully populated / defaulted
func (c *ApplyClusterCmd) upgradeSpecs(ctx context.Context, assetBuilder *assets.AssetBuilder) error {
```

We store the instanceGroup as the user specified it.  But there are many computed fields - the so called "full" instanceGroup or cluster spec.  upgradeSpecs converts the instanceGroups to populate these fields.

In this PR, we aren't calling upgradeSpecs or the population logic.  So these instance groups don't have many fields populated, including (I would guess) the kubelet config.

I suggest that the easy thing to do is to convert `FilteredInstanceGroups` into a predicate (i.e. `func() bool`), and then basically filter these out at the last minute.  Then we don't need to duplicate the defaulting logic.

justinsb:
Applying a predicate does seem to work ... https://github.com/justinsb/kops/tree/update-cluster-per-ig

The problem is that I don't think we can just do ASGs.  Even on AWS, we need to deal with NodeTerminationHandler object.  But I think this problem also exists on other clouds, so we need to solve it there also IIUC

justinsb:
I opened another PR #16936 with the filter.  I also had to do something similar to what you did - _all_ tasks have access to InstanceGroups (the filtered instance groups) and AllInstanceGroups (unfiltered).  The "trick" is that most tasks work with "InstanceGroups" (the filtered), there are just a handful that need to be repointed to work with AllInstanceGroups (e.g. things that need a list of roles)

Likely we'll need a few tweaks for other clouds, but they shouldn't be too big!

hakman:
> I opened another PR #16936 with the filter. I also had to do something similar to what you did - _all_ tasks have access to InstanceGroups (the filtered instance groups) and AllInstanceGroups (unfiltered). The "trick" is that most tasks work with "InstanceGroups" (the filtered), there are just a handful that need to be repointed to work with AllInstanceGroups (e.g. things that need a list of roles)
> 
> Likely we'll need a few tweaks for other clouds, but they shouldn't be too big!

Awesome! 🎉

PR #16932, opened 2024-11-04
Fix version upgrade kubelet support

rsafonseca:
Automatically preserve kubelet supported version skew on worker nodes while control plane is being updated

Fixes: https://github.com/kubernetes/kops/issues/16907

### DIFF
--- a/cmd/kops/update_cluster.go
+++ b/cmd/kops/update_cluster.go
@@ -27,11 +27,14 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 	"k8s.io/kops/cmd/kops/util"
 	"k8s.io/kops/pkg/apis/kops"
+	apisutil "k8s.io/kops/pkg/apis/kops/util"
 	"k8s.io/kops/pkg/assets"
 	"k8s.io/kops/pkg/commands/commandutils"
 	"k8s.io/kops/pkg/kubeconfig"
@@ -67,6 +70,10 @@ type UpdateClusterOptions struct {
 	SSHPublicKey       string
 	RunTasksOptions    fi.RunTasksOptions
 	AllowKopsDowngrade bool
+	// Bypasses kubelet vs control plane version skew checks,
+	// which by default prevent non-control plane instancegroups
+	// from being updated to a version greater than the control plane
+	IgnoreKubeletVersionSkew bool
 	// GetAssets is whether this is invoked from the CmdGetAssets.
 	GetAssets bool
 
@@ -103,6 +110,8 @@ func (o *UpdateClusterOptions) InitDefaults() {
 	o.Target = "direct"
 	o.SSHPublicKey = ""
 	o.OutDir = ""
+	// By default we enforce the version skew between control plane and worker nodes
+	o.IgnoreKubeletVersionSkew = false
 
 	// By default we export a kubecfg, but it doesn't have a static/eternal credential in it any more.
 	o.CreateKubecfg = true
@@ -163,6 +172,7 @@ func NewCmdUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.RegisterFlagCompletionFunc("lifecycle-overrides", completeLifecycleOverrides)
 
 	cmd.Flags().BoolVar(&options.Prune, "prune", options.Prune, "Delete old revisions of cloud resources that were needed during an upgrade")
+	cmd.Flags().BoolVar(&options.IgnoreKubeletVersionSkew, "ignore-kubelet-version-skew", options.IgnoreKubeletVersionSkew, "Setting this to true will force updating the kubernetes version on all instance groups, regardles of which control plane version is running")
 
 	return cmd
 }
@@ -318,20 +328,30 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Up
 		return nil, err
 	}
 
+	minControlPlaneRunningVersion := cluster.Spec.KubernetesVersion
+	if !c.IgnoreKubeletVersionSkew {
+		minControlPlaneRunningVersion, err = checkControlPlaneRunningVersion(ctx, cluster.ObjectMeta.Name, minControlPlaneRunningVersion)
+		if err != nil {
+			klog.Warningf("error checking control plane running version, assuming no k8s upgrade in progress: %v", err)
+		} else {
+			klog.V(2).Infof("successfully checked control plane running version: %v", minControlPlaneRunningVersion)
+		}
+	}
 	applyCmd := &cloudup.ApplyClusterCmd{
-		Cloud:               cloud,
-		Clientset:           clientset,
-		Cluster:             cluster,
-		DryRun:              isDryrun,
-		AllowKopsDowngrade:  c.AllowKopsDowngrade,
-		RunTasksOptions:     &c.RunTasksOptions,
-		OutDir:              c.OutDir,
-		InstanceGroupFilter: predicates.AllOf(instanceGroupFilters...),
-		Phase:               phase,
-		TargetName:          targetName,
-		LifecycleOverrides:  lifecycleOverrideMap,
-		GetAssets:           c.GetAssets,
-		DeletionProcessing:  deletionProcessing,
+		Cloud:                      cloud,
+		Clientset:                  clientset,
+		Cluster:                    cluster,
+		DryRun:                     isDryrun,
+		AllowKopsDowngrade:         c.AllowKopsDowngrade,
+		RunTasksOptions:            &c.RunTasksOptions,
+		OutDir:                     c.OutDir,
+		InstanceGroupFilter:        predicates.AllOf(instanceGroupFilters...),
+		Phase:                      phase,
+		TargetName:                 targetName,
+		LifecycleOverrides:         lifecycleOverrideMap,
+		GetAssets:                  c.GetAssets,
+		DeletionProcessing:         deletionProcessing,
+		ControlPlaneRunningVersion: minControlPlaneRunningVersion,
 	}
 
 	applyResults, err := applyCmd.Run(ctx)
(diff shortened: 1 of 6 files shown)


rsafonseca:
cc @danports @hakman @justinsb as discussed in the office hours
I've only had time to do some basic testing, but haven't gone through an actual upgrade, though the outputs look as expected :)

danports:
Thanks @rsafonseca! On the one hand, I can't say I'm the biggest fan of the ergonomics of this approach, because it basically means we have to run the identical `kops update cluster` and `kops rolling-update cluster` commands twice in our CD script - and not having a command line switch like `--instance-group-role` makes the script less self-documenting. But on the other hand, I am very happy to have any solution that works and this also has the advantage of being safe by default in terms of automatically enforcing the version skew policy. An "uber command" as @rifelpet called it would still be great, and while I haven't looked at all at the code, I wondered whether one could implement it by making a rolling update (of CP or worker nodes) just another type of task that needs to be run in the update dependency tree.

cmd/kops/update_cluster.go
```diff
@@ -290,19 +299,29 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Up
 		return nil, err
 	}
 
+	minControlPlaneRunningVersion := cluster.Spec.KubernetesVersion
+	if !c.IgnoreVersionSkew {
+		minControlPlaneRunningVersion, err = checkControlPlaneRunningVersion(ctx, cluster.ObjectMeta.Name, minControlPlaneRunningVersion)
+		if err != nil {
+			klog.Warningf("error checking control plane running verion: %v", err)
+		} else {
+			klog.Warningf("successfully checked control plane running version: %v", minControlPlaneRunningVersion)
```
justinsb, line 308: Nit: any reason to make this a warning (as opposed to Infof)?

cmd/kops/update_cluster.go
```diff
@@ -290,19 +299,29 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Up
 		return nil, err
 	}
 
+	minControlPlaneRunningVersion := cluster.Spec.KubernetesVersion
+	if !c.IgnoreVersionSkew {
+		minControlPlaneRunningVersion, err = checkControlPlaneRunningVersion(ctx, cluster.ObjectMeta.Name, minControlPlaneRunningVersion)
+		if err != nil {
+			klog.Warningf("error checking control plane running verion: %v", err)
```
justinsb, line 306: I think we should just return the error, unless there's some reason to expect this error to happen which we need to tolerate

cmd/kops/update_cluster.go
```diff
@@ -518,3 +537,38 @@ func completeLifecycleOverrides(cmd *cobra.Command, args []string, toComplete st
 	}
 	return completions, cobra.ShellCompDirectiveNoFileComp
 }
+
+func checkControlPlaneRunningVersion(ctx context.Context, clusterName string, version string) (string, error) {
+
+	configLoadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
```
justinsb, line 543: Nit:  I think we should try to share this "get kubeconfig" code with the rolling-update logic (I think that also connects to the cluster).  Not a blocker though.

cmd/kops/update_cluster.go
```diff
@@ -518,3 +537,38 @@ func completeLifecycleOverrides(cmd *cobra.Command, args []string, toComplete st
...
+
+	parsedVersion, err := apisutil.ParseKubernetesVersion(version)
+	if err != nil {
+		return version, fmt.Errorf("cannot parse kubernetes version %q: %v", clusterName, err)
+	}
+	nodeList, err := k8sClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{
+		LabelSelector: "node-role.kubernetes.io/control-plane",
+	})
+	if err != nil {
+		return version, fmt.Errorf("cannot list nodes in cluster %q: %v", clusterName, err)
```
justinsb, line 564: Ah - this is tricky.  If the control plane is offline, we presumably don't want to update the nodes (?)

pkg/nodemodel/fileassets.go
```diff
@@ -42,22 +42,40 @@ type FileAssets struct {
...
 
-// AddFileAssets adds the file assets within the assetBuilder
 func (c *FileAssets) AddFileAssets(assetBuilder *assets.AssetBuilder) error {
+	var err error
+	c.Assets, err = c.addFileAssets(assetBuilder, c.Cluster.Spec.KubernetesVersion)
+	if err != nil {
+		return err
+	}
+	if len(assetBuilder.KubeletSupportedVersion) > 0 && assetBuilder.KubeletSupportedVersion != c.Cluster.Spec.KubernetesVersion {
+		c.AssetsSupportedKubelet, err = c.addFileAssets(assetBuilder, assetBuilder.KubeletSupportedVersion)
```
justinsb, line 58: Ah - great point - we need different assets.  Although I'm starting to worry about complexity now...

justinsb:
Thanks for this!  A few random comments on the code but I'm going to mull on the overall approach.

cmd/kops/update_cluster.go
rsafonseca: oh no... that was just for my testing i fully intended to make it info or debug and then forgot :)

cmd/kops/update_cluster.go
rsafonseca: If the control plane is offline, this is the least of our worries lol. But it might be an expected situation (e.g. creating a new cluster) so we just log the error and continue as normal

pkg/nodemodel/fileassets.go
rsafonseca: Yes, the complexity here worried me a bit as well as I dug a bit into how to achieve this. The code isn't perfect but i tried to do it without having to refactor many things, if you have any better ideas we can either improve this down the line or i can make any adjustments :)

justinsb:
So we have added the ability to update only some of the instance groups, and in https://github.com/kubernetes/kops/pull/16939 I am proposing to make it easy to drive those operations in a predefined sequence.

One thing I don't like about that approach is that we do have to predefine the sequence, and one of the things I like in this PR is that we're being more reasoned about the sequence: rather than hard-coding the sequence we effectively infer the sequence at run-time. 

It might be that the predefined approach is good enough to unblock 1.31, but I would like to pursue this direction more.  I am wondering if we should make it more explicit, adding a field for the Kube version on each instance groups and managing that.

danports:
@justinsb Could we combine the two approaches by having the anneal/reconcile cluster command call update & rolling-update repeatedly, until the update command reports that there are no tasks left that it had to skip due to version skew or other dependencies? Then the anneal/reconcile command wouldn't have to hard-code any sequences, and that structure might also be able to incorporate running the `--prune` logic as well.

justinsb:
@danports yes, I think we might need a combination.  Something brought up in office hours was that we need to support this for terraform also, where we can't emit partial configuration (at least I don't think we can!). I've gone round in circles a lot on other ways to make this work, but I am feeling like the approach in this PR is probably going to be the only one that we can really do, short of introducing a kubernetesVersion field in the instance group which is its own management problem.

I am doing some work to try to clean up our assets logic so that this PR doesn't have to plumb through multiple asset versions - that should make this PR smaller I hope!

justinsb:
OK, I think I've got something working, and it is based on this PR.  The approach here is very clever, and because terraform will delete resources if we stop emitting them, I think this is the right way forward.  The downside is that we will have to run update / rolling-update / update / rolling-update, but we can address that with messages to the user and potentially a command like #16939 (though that will have to be reworked)

I do think we should merge #16954 first though, as with that PR the code in this PR becomes much easier.  The whole building of fileAssets looks to be unnecessarily complicated, which is part of the reason why this PR is somewhat complicated.  I have a rebase of it after #16954 so once that merges I think I can probably just update this PR branch (as a maintainer).

cc @hakman

rsafonseca:
Cool, let me know if you need a hand with anything @justinsb :)

justinsb:
/ok-to-test

I rebased, it's simpler now that we can resolve the file assets "at the last minute"

rsafonseca:
Awesome, much cleaner change now :D

justinsb:
> Awesome, much cleaner change now :D

Thanks @rsafonseca  - I want to emphasize that you did the hard work here of figuring out the approach.  I just cleaned up some of the existing over-complicated code in kOps that was resolving file assets in an awkward way.

cc @hakman if we get this merged I have a few small follow-ons (e.g. there's a problem with featuregates being removed upstream), but I think this is our best approach, given terraform.

cmd/kops/update_cluster.go
```diff
@@ -163,6 +171,7 @@ func NewCmdUpdateCluster(f *util.Factory, out io.Writer) *cobra.Command {
 	cmd.RegisterFlagCompletionFunc("lifecycle-overrides", completeLifecycleOverrides)
 
 	cmd.Flags().BoolVar(&options.Prune, "prune", options.Prune, "Delete old revisions of cloud resources that were needed during an upgrade")
+	cmd.Flags().BoolVar(&options.IgnoreVersionSkew, "ignore-version-skew", options.IgnoreVersionSkew, "Setting this to true will force updating the kubernetes version on all instance groups, regardles of which control plane version is running")
```
rifelpet, line 174: given how many policies are defined here: https://kubernetes.io/releases/version-skew-policy/

and that this only violates one of them, perhaps we should make the flag more specific? `--ignore-kubelet-version-skew`

rsafonseca: I see your point @rifelpet, but i think instead what we could do is keep this naming and add an additional check of whether the user is trying to update more than 3 minor versions at a time, and throw an error if that's the case unless the --ignore-version-skew flag is set. That would take care of all the version skews listed there and the flag name would then make sense. WDYT?

justinsb: So we _can_ do this, using something like:

```
	// By default we enforce the version skew between control plane and worker nodes
	options.IgnoreVersionSkew = "DoNotIgnore"
	cmd.Flags().StringVar(&options.IgnoreVersionSkew, "ignore-version-skew", options.IgnoreVersionSkew, "Setting this to true will force updating the kubernetes version on all instance groups, regardless of which control plane version is running")
	cmd.Flags().Lookup("ignore-version-skew").NoOptDefVal = "IgnoreAllSkew"
```

The thing I don't like is that it won't be very discoverable, particularly once we have more options.  The help with the above looks like this:

```
      --ignore-version-skew string[="IgnoreAllSkew"]   Setting this to true will force updating the kubernetes version on all instance groups, regardless of which control plane version is running (default "DoNotIgnore")
```

So we're going to be struggling to get a clear description here.  Maybe that just doesn't matter - this is blocking the 1.31 beta release, so maybe this is good enough?

The alternative is that we have separate flags, `--ignore-kubelet-version-skew`, `--ignore-upgrade-skip-versions` and maybe that's all (but if not we just add more).  We'd have more flags, but they would be simple.

rifelpet: Having separate flags makes sense to me because users may want to toggle them separately which is easier to understand as independent flags rather than trying to choose the most appropriate flag value.

justinsb: Yeah, I think it does seem simpler as I don't think we're going to end up with _too many_ of them (and if we do, we can always hide the flags and move to an aggregated flag)

Any objections to `--ignore-kubelet-version-skew` ?  I think I can push a change if we all agree cc @rsafonseca

rsafonseca: No concerns from me, feel free to push the change @justinsb  :)

cmd/kops/update_cluster.go
```diff
@@ -318,20 +328,30 @@ func RunUpdateCluster(ctx context.Context, f *util.Factory, out io.Writer, c *Up
 		return nil, err
 	}
 
+	minControlPlaneRunningVersion := cluster.Spec.KubernetesVersion
+	if !c.IgnoreKubeletVersionSkew {
+		minControlPlaneRunningVersion, err = checkControlPlaneRunningVersion(ctx, cluster.ObjectMeta.Name, minControlPlaneRunningVersion)
+		if err != nil {
+			klog.Warningf("error checking control plane running verion: %v", err)
+		} else {
+			klog.Warningf("successfully checked control plane running version: %v", minControlPlaneRunningVersion)
```
rifelpet, line 337:
```suggestion
			klog.V(2).Infof("successfully checked control plane running version: %v", minControlPlaneRunningVersion)
```

cmd/kops/update_cluster.go
rifelpet: I agree we shouldn't log this warning when we know the check wont succeed otherwise it will cause confusion. For example, the e2e tests show this log line:

`W1202 14:53:23.985493   47436 update_cluster.go:335] error checking control plane running verion: cannot load kubecfg settings for "e2e-pr16932.pull-kops-e2e-k8s-aws-calico.test-cncf-aws.k8s.io": context "e2e-pr16932.pull-kops-e2e-k8s-aws-calico.test-cncf-aws.k8s.io" does not exist`

Another failure case would be if the kubeconfig context _does_ exist but the k8s API isn't up - perhaps a user is recreating a previously-deleted cluster. In that case it may be more acceptable to log the warning.

Perhaps we could also avoid confusion by clarifying the impact:

```suggestion
			klog.Warningf("error checking control plane running version, assuming no k8s upgrade in progress: %v", err)
```

justinsb:
Thanks @rifelpet ; I incorporated your log message tweaks.  We do need some follow on PR (some node feature flags have been removed in 1.31), so we can also make more tweaks in follow on as we observe things (many of which will be more of the "oh, we should consider the instance group version, not the cluster version" variety)

rifelpet (approved):

PR #16937, opened 2024-11-08
Fix: Normalize the SQS policies before comparing them

justinsb:
This avoids spurious comparison errors.


pkg/jsonutils/transform.go
```diff
@@ -0,0 +1,166 @@
...
+
+	for k, v := range m {
+		childPath := path + "." + k
+
+		v2, err := o.visitAny(v, childPath)
+		if err != nil {
+			return err
+		}
+		m[k] = v2
+		v = v2
```
hakman, line 95: Linter says this needs to be removed.
```suggestion
```

hakman (approved):

PR #16939, opened 2024-11-09
reconcile command to combine update and rolling-update

justinsb:
This is adding the "combined command" now that the control-plane must be upgraded before the nodes.

I'm not clear as to what the best name is (or even whether it should be a separate command, or a flag).  Thoughts / suggestions very welcome!


danports:
Some name suggestions: reconcile cluster, converge cluster?

danports:
Thanks for getting #16932 over the finish line @justinsb! Do you think this PR will land in 1.31 too or does it need some more time to cook?

PR #16944, opened 2024-11-11
WIP: tests: add test for bare-metal with ipv6

justinsb:
IPv6 brings some new complexities, particularly around IPAM.


justinsb:
I am trying to upload this and then I can rebasing as I/we fix each problem.

Current problem is from nodeup:

```
vm0 nodeup[703]: W1111 17:07:00.322041     703 main.go:133] got error running nodeup (will retry in 30s): error building loader: building *model.PrefixBuilder: kOps IPAM controller not supported on cloud "metal"
```

So we need to decide how the podCIDR is assigned!

justinsb:
cc @hakman 

I think this is now uncontroversial (I hope).  We assign podCIDRs to nodes if they are configured on the Host object.  If users don't want to do that, they just don't set podCIDRs on the Host object.

hakman:
> cc @hakman
> 
> I think this is now uncontroversial (I hope). We assign podCIDRs to nodes if they are configured on the Host object. If users don't want to do that, they just don't set podCIDRs on the Host object.

Cool, I will take a look soon. 🚀

pkg/apis/kops/v1alpha2/host.go
```diff
@@ -36,6 +36,9 @@ type Host struct {
 type HostSpec struct {
 	PublicKey     string `json:"publicKey,omitempty"`
 	InstanceGroup string `json:"instanceGroup,omitempty"`
+
+	// PodCIDRs configures the IP ranges to be used for pods on this node/host.
+	PodCIDRs []string `json:"podCIDRs,omitempty"`
```
hakman, line 41: Should this be added to the `v1apha3` API too? What about the main API?
https://github.com/kubernetes/kops/blob/5f564fe1abc5526e4554845a5518f14443cae5e7/pkg/apis/kops/v1alpha3/host.go#L36-L39

justinsb: Great call.  But I propose doing it as a separate PR, because I think we are also missing a round-trip test that would have caught this!

justinsb:
/hold cancel

I propose adding a round-trip test alongside fixing the missing field in v1alpha3

PR #16945, opened 2024-11-11
2024 Scale dev

upodroid:
We have plenty of GCP credits to burn between now and the end of the year. 

Restarting https://github.com/kubernetes/kops/pull/16181
Part of https://github.com/kubernetes/enhancements/issues/4224


dims:
are we trying to land this or should it be on `/hold`?

upodroid:
/hold

I need to get these jobs to launch and then write their periodic and run it the rest of Nov everyday and see if they can replace the existing kubeup scale jobs.

I think the 500 will pass easily. 5K will be a stretch.

PR #16954, opened 2024-11-24
refactor: build assets for kubernetes and nodeup explicitly

justinsb:
Also remove the (potentially) confusing FileAssets type.


justinsb:
I think if we can get this PR in, then this will greatly simplify #16932.  I then have a few more small cleanup PRs and then we are able to update the cluster safely according to the version skew policy :-)

cc @hakman

hakman (approved):

PR #16955, opened 2024-11-25
refactor: create abstraction over cluster/instancegroup for building assets

justinsb:
This abstraction should let us change the version on an instance group level.


justinsb:
I think we need this one also, it introduces the idea that the kubernetes version is a property of the instance group, without needing to fully wire up the field.

cc @hakman

hakman (approved):

PR #16961, opened 2024-11-26
fix(openstack): determine ports to delete based on tags

ederst:
This will determine the ports to delete when deleting an InstanceGroup in OpenStack based on the tags, rather than comparing the prefix of the name of the port to a pattern.

The current approach for deleting ports associated with an InstanceGroup in OpenStack can be problematic.

For example, when creating an InstanceGroup called `worker` and another one called `worker-2`. Some of the ports kOps creates for those groups will look like this:

* `port-worker-2-1-<clusterName>`
* `port-worker-2-2-<clusterName>`
* `port-worker-1-<clusterName>`
* `port-worker-2-<clusterName>`
* `port-worker-20-<clusterName>`

Deleting the `worker-2` IG will ultimately delete all the ports prefixed with `port-worker-2`, which means that kOps will also delete ports of the `worker` IG.

TODO:

- [x] Write tests

### DIFF
--- a/cloudmock/openstack/mocknetworking/ports.go
+++ b/cloudmock/openstack/mocknetworking/ports.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"slices"
 	"strings"
 
 	"k8s.io/kops/upup/pkg/fi"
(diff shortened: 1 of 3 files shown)


ederst:
@hakman @zetaab friendly bump (took me some time to finish)

hakman:
Thanks for the effort @ederst, and sorry for the delay.
/lgtm
/approve

PR #16975, opened 2024-12-04
Fixup kubelet and controlPlaneKubelet config building

justinsb:
We need to differentiate between the version of the control plane
and the version of the nodes.


justinsb:
/retest

Both on cilium, so _could_ be real, but worth a retest

hakman:
> /retest
> 
> Both on cilium, so _could_ be real, but worth a retest

eni always fails on those 2 tests

hakman (approved):

PR #16983, opened 2024-12-06
feat: add custom topologySpreadConstraints support to coredns

thiagoluiznunes:
### Context on the change

Recently at WildlifeStudios, we had a short temporary outage in CoreDNS in one of our clusters, since a single node crashed, and all CoreDNS pods were running on it.

So, to avoid that, we would like to set our own topologySpreadContraints parameters according to our user case.

### What does this PR do?
This PR adds support for customizing the field topologySpreadConstraints in the template of the CoreDNS add-on.

### DIFF
--- a/k8s/crds/kops.k8s.io_clusters.yaml
+++ b/k8s/crds/kops.k8s.io_clusters.yaml
@@ -3742,6 +3742,183 @@ spec:
                           type: string
                       type: object
                     type: array
+                  topologySpreadConstraints:
+                    description: TopologySpreadConstraints describes how a group of
+                      pods ought to spread across topology domains.
+                    items:
+                      description: TopologySpreadConstraint specifies how to spread
+                        matching pods among the given topology.
+                      properties:
+                        labelSelector:
+                          description: |-
+                            LabelSelector is used to find matching pods.
+                            Pods that match this label selector are counted to determine the number of pods
+                            in their corresponding topology domain.
+                          properties:
+                            matchExpressions:
+                              description: matchExpressions is a list of label selector
+                                requirements. The requirements are ANDed.
+                              items:
+                                description: |-
+                                  A label selector requirement is a selector that contains values, a key, and an operator that
+                                  relates the key and values.
+                                properties:
+                                  key:
+                                    description: key is the label key that the selector
+                                      applies to.
+                                    type: string
+                                  operator:
+                                    description: |-
+                                      operator represents a key's relationship to a set of values.
+                                      Valid operators are In, NotIn, Exists and DoesNotExist.
+                                    type: string
+                                  values:
+                                    description: |-
+                                      values is an array of string values. If the operator is In or NotIn,
+                                      the values array must be non-empty. If the operator is Exists or DoesNotExist,
+                                      the values array must be empty. This array is replaced during a strategic
+                                      merge patch.
+                                    items:
+                                      type: string
+                                    type: array
+                                    x-kubernetes-list-type: atomic
+                                required:
+                                - key
+                                - operator
+                                type: object
+                              type: array
+                              x-kubernetes-list-type: atomic
+                            matchLabels:
+                              additionalProperties:
+                                type: string
+                              description: |-
+                                matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels
+                                map is equivalent to an element of matchExpressions, whose key field is "key", the
+                                operator is "In", and the values array contains only "value". The requirements are ANDed.
+                              type: object
+                          type: object
+                          x-kubernetes-map-type: atomic
+                        matchLabelKeys:
+                          description: |-
+                            MatchLabelKeys is a set of pod label keys to select the pods over which
+                            spreading will be calculated. The keys are used to lookup values from the
+                            incoming pod labels, those key-value labels are ANDed with labelSelector
+                            to select the group of existing pods over which spreading will be calculated
+                            for the incoming pod. The same key is forbidden to exist in both MatchLabelKeys and LabelSelector.
+                            MatchLabelKeys cannot be set when LabelSelector isn't set.
+                            Keys that don't exist in the incoming pod labels will
+                            be ignored. A null or empty list means only match against labelSelector.
+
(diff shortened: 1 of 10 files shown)


thiagoluiznunes:
Hey folks, could you review the PR?
cc @hakman @justinsb

hakman:
@justinsb Could you please take a look?
/lgtm
/assign @justinsb

hakman:
> /retest

@thiagoluiznunes Don't worry, it's a known issue.
/override pull-kops-e2e-cni-flannel

Could you also share what value you would like to use in your case and why?

thiagoluiznunes:
> > /retest
> 
> @thiagoluiznunes Don't worry, it's a known issue. /override pull-kops-e2e-cni-flannel
> 
> Could you also share what value you would like to use in your case and why?

@hakman when you ask about my case, is it about the /retest or the feature?
Enabling the customization of the topologySpreadConstraints is necessary because we have some small clusters at Wildlife with 2 or 3 nodes of workload. Eventually, the coredns pods are scheduled in the same node due to the small quantity, and this scenario is dangerous if a failure occurs in that node. Does it make sense?

thiagoluiznunes:
> > > /retest
> > 
> > 
> > @thiagoluiznunes Don't worry, it's a known issue. /override pull-kops-e2e-cni-flannel
> > Could you also share what value you would like to use in your case and why?
> 
> @hakman when you ask about my case, is it about the /retest or the feature? Enabling the customization of the topologySpreadConstraints is necessary because we have some small clusters at Wildlife with 2 or 3 nodes of workload. Eventually, the coredns pods are scheduled in the same node due to the small quantity, and this scenario is dangerous if a failure occurs in that node. Does it make sense?

Hi @hakman , I wanted to follow up and see if you have any updates on this discussion.

thiagoluiznunes:
Hi @hakman, We would like to set the following values as the default in the topologySpreadConstraints:

```
topologySpreadConstraints:
- labelSelector:
    matchLabels:
      k8s-app: kube-dns
  maxSkew: 1
  topologyKey: topology.kubernetes.io/zone
  whenUnsatisfiable: DoNotSchedule
- labelSelector:
    matchLabels:
      k8s-app: kube-dns
  maxSkew: 1
  topologyKey: kubernetes.io/hostname
  whenUnsatisfiable: DoNotSchedule
```

hakman:
What you are trying makes sense, but I don't think it will work due to `maxSkew`.

Have you tried using `maxSurge` and `maxUnavailable`?
https://kops.sigs.k8s.io/operations/rolling-update/

thiagoluiznunes:
> What you are trying makes sense, but I don't think it will work due to `maxSkew`.
> 
> Have you tried using `maxSurge` and `maxUnavailable`? https://kops.sigs.k8s.io/operations/rolling-update/

@hakman So, these fields have no effect, since there is no kops rollout, but Karpenter prunes the nodes that fail, and this ends up affecting CoreDNS. Therefore, using maxSurge or maxUnavailable would not make a difference.

thiagoluiznunes:
Hi @hakman . Do you have updates about the discussion?

hakman:
@thiagoluiznunes We had lots of discussions on this topic (thanks @justinsb). The conclusion is that we would prefer to change the default to `whenUnsatisfiable: DoNotSchedule`. Would that be ok with you?

thiagoluiznunes:
@hakman  I would like to understand why the custom option for topologySpread is not enabled. It could potentially be beneficial for us (host: DoNotSchedule, zone: ScheduleAnyway), but enforcing this configuration might introduce issues in unanticipated corner cases. What drawbacks would Kops face by incorporating this custom configuration? I believe that adopting this custom option would primarily provide us with advantages. Does it make sense?

hakman:
> @hakman I would like to understand why the custom option for topologySpread is not enabled. It could potentially be beneficial for us (host: DoNotSchedule, zone: DotNotSchedule), but enforcing this configuration might introduce issues in unanticipated corner cases. What drawbacks would Kops face by incorporating this custom configuration? I believe that adopting this custom option would primarily provide us with advantages. Does it make sense?

We prefer to not add another config option here, or another dependency on the `topologySpreadConstraints` type.

thiagoluiznunes:
> > @hakman I would like to understand why the custom option for topologySpread is not enabled. It could potentially be beneficial for us (host: DoNotSchedule, zone: DotNotSchedule), but enforcing this configuration might introduce issues in unanticipated corner cases. What drawbacks would Kops face by incorporating this custom configuration? I believe that adopting this custom option would primarily provide us with advantages. Does it make sense?
> 
> We prefer to not add another config option here, or another dependency on the `topologySpreadConstraints` type.

Hi @hakman, 
Could we change the default value to DoNotSchedule? This adjustment would resolve our issue. What are your thoughts? I can update the merge request accordingly.

thiagoluiznunes:
Hi @hakman,
After discussing with my team regarding the recent office hours, we raised a point about the inability to customize the configuration. Essentially, we identified a new corner case related to our infrastructure. We have two scenarios: small clusters and large clusters. If we set the parameters `DoNotSchedule` for both host and zone, it will solve the problem for small clusters, but it will increase costs for medium and large clusters. For instance, we have a production cluster with 11 CoreDNS pods running smoothly with the current topology set to `ScheduleAnyway`. However, if we change the parameter to DoNotSchedule, new nodes will be created unnecessarily just to fit the topology. It’s possible that other users could encounter the same scenario. What are your thoughts?

thiagoluiznunes:
> Hi @hakman, After discussing with my team regarding the recent office hours, we raised a point about the inability to customize the configuration. Essentially, we identified a new corner case related to our infrastructure. We have two scenarios: small clusters and large clusters. If we set the parameters `DoNotSchedule` for both host and zone, it will solve the problem for small clusters, but it will increase costs for medium and large clusters. For instance, we have a production cluster with 11 CoreDNS pods running smoothly with the current topology set to `ScheduleAnyway`. However, if we change the parameter to DoNotSchedule, new nodes will be created unnecessarily just to fit the topology. It’s possible that other users could encounter the same scenario. What are your thoughts?

@hakman ping :D

hakman:
@thiagoluiznunes https://github.com/kubernetes/kops/pull/17472 change should add the desired behaviour to CoreDNS. Sorry for the delay.
/close

PR #16985, opened 2024-12-06
Dial down warning level when asset is not found

justinsb:
We have to download the hash file for assets that are not well-known,
but this is not an error, and this causes a lot of noise in our logs.


justinsb:
cc @hakman  I think you said this was confusing to users?  We could also skip the warning only if the URL matches KOPS_BASE_URL, meaning we would still be alerted when we have not updated the hashes in a while!

hakman (commented):
Awesome! 😀

hakman (approved):

PR #17138, opened 2024-12-14
Release 1.31.0-beta.1

justinsb:
Release 1.31.0-beta.1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,16 +66,16 @@ GITSHA := $(shell cd ${KOPS_ROOT}; git describe --always)
 # We lock the versions of our controllers also
 # We need to keep in sync with:
 #   pkg/model/components/etcdmanager/model.go
-KOPS_UTILS_CP_TAG=1.31.0-alpha.1
+KOPS_UTILS_CP_TAG=1.31.0-beta.1
 KOPS_UTILS_CP_PUSH_TAG=$(shell tools/get_workspace_status.sh | grep STABLE_KOPS_UTILS_CP_TAG | awk '{print $$2}')
 #   upup/models/cloudup/resources/addons/dns-controller/
-DNS_CONTROLLER_TAG=1.31.0-alpha.1
+DNS_CONTROLLER_TAG=1.31.0-beta.1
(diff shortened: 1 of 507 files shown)


hakman:
> [It] [sig-network] Services should implement NodePort and HealthCheckNodePort correctly when ExternalTrafficPolicy changes
This test always fails for `pull-kops-e2e-k8s-gce-cilium`.
/override pull-kops-e2e-k8s-gce-cilium

hakman (approved):

justinsb:
Thanks for the cilium pointer - and I am looking at the bare metal test!

hakman:
> Thanks for the cilium pointer - and I am looking at the bare metal test!

@justinsb I think we had the same issue at previous release, right @rifelpet?
We just manually merged the PR at that time.

justinsb:
> > Thanks for the cilium pointer - and I am looking at the bare metal test!
> 
> @justinsb I think we had the same issue at previous release, right @rifelpet? We just manually merged the PR at that time.

Ah, I think that's the right thing to do.

First of all, this test is experimental functionality anyway, so shouldn't block a release.

In this case, the issue is that kops-controller is not getting KOPS_BASE_URL, I think, because this is the error from those logs:

```
I1214 19:42:14.016404       1 builder.go:317] asset "https://artifacts.k8s.io/binaries/kops/1.31.0-beta.1/linux/amd64/nodeup" is not well-known, downloading hash
I1214 19:42:14.055707       1 context.go:317] hit maximum retries 1 with error file does not exist
I1214 19:42:14.055743       1 builder.go:337] Unable to read hash file "https://artifacts.k8s.io/binaries/kops/1.31.0-beta.1/linux/amd64/nodeup.sha256": file does not exist
I1214 19:42:14.199812       1 context.go:317] hit maximum retries 1 with error file does not exist
I1214 19:42:14.199846       1 builder.go:337] Unable to read hash file "https://github.com/kubernetes/kops/releases/download/v1.31.0-beta.1/nodeup-linux-amd64.sha256": file does not exist
I1214 19:42:14.199855       1 builder.go:352] Unable to read new sha256 hash file (is this an older/unsupported kubernetes release?)
I1214 19:42:14.236349       1 context.go:317] hit maximum retries 1 with error file does not exist
```

So yes, agree with manual merge!

PR #17152, opened 2024-12-27
tests: use reconcile command for kOps 1.31+

justinsb:
We expect people to start using the reconcile command
particularly for upgrades.

### DIFF
--- a/tests/e2e/scenarios/upgrade-ab/run-test.sh
+++ b/tests/e2e/scenarios/upgrade-ab/run-test.sh
@@ -121,24 +121,37 @@ else
   "${KOPS_B}" edit cluster "${CLUSTER_NAME}" "--set=cluster.spec.kubernetesVersion=${K8S_VERSION_B}"
 fi
 
-"${KOPS_B}" update cluster
-"${KOPS_B}" update cluster --admin --yes
-# Verify no additional changes
-"${KOPS_B}" update cluster
+if [[ "${KOPS_VERSION_B}" =~ 1.(20|21|22|23|24|25|26|27|28|29|30). ]]; then
+  # kOps introduced the reconcile command in 1.31
+  # TODO: remove this block once we stop testing upgrades to kops <1.31
+  "${KOPS_B}" update cluster
+  "${KOPS_B}" update cluster --admin --yes
+  # Verify no additional changes
+  "${KOPS_B}" update cluster
 
-# Verify kubeconfig-a still works
-kubectl get nodes -owide --kubeconfig "${KUBECONFIG_A}"
+  # Verify kubeconfig-a still works
+  kubectl get nodes -owide --kubeconfig "${KUBECONFIG_A}"
 
-# Sleep to ensure channels has done its thing
-sleep 120s
+  # Sleep to ensure channels has done its thing
+  sleep 120s
 
-# Make sure configuration B has been applied (e.g. new load balancer is ready)
-"${KOPS_B}" validate cluster --wait=10m
+  # Make sure configuration B has been applied (e.g. new load balancer is ready)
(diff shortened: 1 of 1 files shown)


tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -121,24 +121,35 @@ else
   "${KOPS_B}" edit cluster "${CLUSTER_NAME}" "--set=cluster.spec.kubernetesVersion=${K8S_VERSION_B}"
 fi
 
-"${KOPS_B}" update cluster
-"${KOPS_B}" update cluster --admin --yes
-# Verify no additional changes
-"${KOPS_B}" update cluster
+if [[ "${KOPS_VERSION_B}" =~ 1.3[123456789] ]]; then
```
rifelpet, line 124: maybe we should invert this, using the old method when the version matches `1.2` or `1.30`. That way we avoid surprises when 1.40 is tested.

stl-victor-sudakov:
Are there any plans to document the reconcile command better and in more detail? As a kOps user, I do not quite understand the relation of this command especially to the "upgrade cluster" command.

As far as I understand, it does something like "if a rolling-update is required after an update, perform a rolling update now", is this right? But when do you run the single "kops reconcile cluster" command instead of "kops update cluster --reconcile" ?

tests/e2e/scenarios/upgrade-ab/run-test.sh
```diff
@@ -121,24 +121,35 @@ else
...
+  # Preview changes
+  "${KOPS_B}" update cluster
```
hakman, line 126: Would be nice if `kops reconcile cluster` would show the same output as `kops update cluster`.

tests/e2e/scenarios/upgrade-ab/run-test.sh
justinsb: Good idea 👍

tests/e2e/scenarios/upgrade-ab/run-test.sh
justinsb: We've done this now, so switched to use `reconcile cluster` here

rifelpet (approved):

justinsb:
@stl-victor-sudakov 

> Are there any plans to document the reconcile command better and in more detail? As a kOps user, I do not quite understand the relation of this command especially to the "upgrade cluster" command.

Definitely.  kube 1.31 changed the rules so that kubelet 1.31 cannot join a 1.30 apiserver cluster, which actually interacted poorly with autoscaling and how we did `kops update` previously.  We worked around that by updating one instance group at a time, but then the UX of that was sort of tedious.  So `kops reconcile` today effectively does `kops update cluster --instance-group-roles control-plane,apiserver`, `kops rolling-update cluster --instance-group-roles control-plane,apiserver`, `kops update cluster`, `kops rolling-update cluster`.  There are a few details, but that's what we are trying to streamline with `kops reconcile cluster`

In future we might further improve the mechanism, for example being more precise about how we update addons.

But yes, we were sort of pushed into the reconcile command by upstream changes which we didn't catch till quite late in the process, and now that we've got a reasonable UX we are going to get the docs in place :-)

> As far as I understand, it does something like "if a rolling-update is required after an update, perform a rolling update now", is this right? But when do you run the single "kops reconcile cluster" command instead of "kops update cluster --reconcile" ?

Right, `kops reconcile cluster --yes` is a one-stop command that can be used instead of `kops update cluster` and `kops rolling-update cluster`.  You can still use those commands, but `kops reconcile cluster` makes it easy to obey the more complicated kubernetes upgrade rules that were introduced with k8s 1.31

PR #17157, opened 2024-12-28
Use ephemeral S3 buckets for E2E tests

ameukam:
Use S3 buckets created during the lifecycle of a test instead of a
static one.

### DIFF
--- a/tests/e2e/kubetest2-kops/aws/s3.go
+++ b/tests/e2e/kubetest2-kops/aws/s3.go
@@ -1,0 +1,172 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aws
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"k8s.io/klog/v2"
+)
+
+// We need to pick some region to query the AWS APIs through, even if we are not running on AWS.
+const defaultRegion = "us-east-2"
+
+// It contains S3Client, an Amazon S3 service client that is used to perform bucket
+// and object actions.
+type awsClient struct {
+	S3Client *s3.Client
+}
+
+func NewAWSClient(ctx context.Context) (*awsClient, error) {
+	cfg, err := awsconfig.LoadDefaultConfig(ctx,
+		awsconfig.WithRegion(defaultRegion))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+	return &awsClient{
+		S3Client: s3.NewFromConfig(cfg),
+	}, nil
+}
+
+// AWSBucketName constructs an unique bucket name using the AWS account ID on region us-east-2
+func AWSBucketName(ctx context.Context) (string, error) {
+	config, err := awsconfig.LoadDefaultConfig(ctx, awsconfig.WithRegion(string(types.BucketLocationConstraintUsEast2)))
+	if err != nil {
+		return "", fmt.Errorf("failed to load AWS config: %w", err)
+	}
+
+	stsSvc := sts.NewFromConfig(config)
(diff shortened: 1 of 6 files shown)


tests/e2e/kubetest2-kops/aws/s3.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to load AWS config: %w", err)
+	}
+
+	stsSvc := sts.NewFromConfig(config)
+	callerIdentity, err := stsSvc.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+	if err != nil {
+		return "", fmt.Errorf("building AWS STS presigned request: %w", err)
+	}
+	bucket := "k8s-infra-kops-" + *aws.String(*callerIdentity.Account)
```
rifelpet, line 50: We might want to add a random or timestamp suffix. S3 has a delay before a deleted bucket's name can be reused, and if a boskos account is quickly reused between kops jobs then the bucket creation could fail.

rifelpet (commented):
We'll need a second bucket for the cluster's oidc discovery store, we can reuse this logic but it will need a policy that allows anonymous read access.

tests/e2e/kubetest2-kops/aws/s3.go
```diff
@@ -0,0 +1,158 @@
...
+			klog.Errorf("Failed to set public read policy on bucket %s: %v", bucketName, err)
+			return err
+		}
+		klog.Infof("Public read policy set on bucket %s", bucketName)
+	}
+
+	return err
+}
+
+func (client awsClient) DeleteS3Bucket(ctx context.Context, bucketName string) error {
```
BenTheElder, line 108: TODO: boskos?

ameukam: Yeah. Boskos is the next thing we can get rid of those static buckets.

tests/e2e/kubetest2-kops/aws/s3.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+	return &awsClient{
+		S3Client: s3.NewFromConfig(cfg),
+	}, nil
+}
+
+// AWSBucketName constructs an unique bucket name using the AWS account ID on region us-east-2
+func AWSBucketName(ctx context.Context) (string, error) {
```
xmudrii, line 56: Nit: why is this not a receiver function of `awsClient`?

tests/e2e/kubetest2-kops/aws/s3.go
```diff
@@ -0,0 +1,172 @@
...
+	stsSvc := sts.NewFromConfig(config)
+	callerIdentity, err := stsSvc.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+	if err != nil {
+		return "", fmt.Errorf("building AWS STS presigned request: %w", err)
+	}
+
+	// Add timestamp suffix
+	timestamp := time.Now().Format("01022006")
+	bucket := fmt.Sprintf("k8s-infra-kops-%s", *callerIdentity.Account)
+	bucket = fmt.Sprintf("%s-%s", bucket, timestamp)
```
xmudrii, line 71: Why don't we use a random name here? Isn't there a risk that if cleanup fails, the account gets reused, we are either stuck with an existing bucket that's not cleaned up or not being able to create another bucket for that day?

tests/e2e/kubetest2-kops/aws/s3.go
```diff
@@ -0,0 +1,172 @@
...
+		CreateBucketConfiguration: &types.CreateBucketConfiguration{
+			LocationConstraint: types.BucketLocationConstraintUsEast2,
+		},
+	})
+
+	var exists *types.BucketAlreadyExists
+	if err != nil {
+		if errors.As(err, &exists) {
+			klog.Infof("Bucket %s already exists.\n", bucketName)
+			err = exists
```
xmudrii, line 95: Is "bucket exists" considered to be an error or do we want to reuse the bucket?

tests/e2e/kubetest2-kops/aws/s3.go
xmudrii: @ameukam This function is not used, right?

tests/e2e/kubetest2-kops/aws/s3.go
xmudrii: Also, why don't we handle other errors here?

tests/e2e/kubetest2-kops/aws/s3.go
ameukam: We could even make it more unique by having `timestamp := time.Now().Format("20060102150405")`  but I'm not opposed to use `math/rand`

tests/e2e/kubetest2-kops/aws/s3.go
ameukam: Not yet. will be used in https://github.com/kubernetes/kops/blob/master/tests/e2e/kubetest2-kops/deployer/down.go

tests/e2e/kubetest2-kops/aws/s3.go
ameukam: the function is meant to construct the bucket's name. We only rely on the STS service to do it. I didn't see the need to access to `S3Client`.

tests/e2e/kubetest2-kops/aws/s3.go
ameukam: Done.

ameukam:
@rifelpet @hakman PTAL.
We may need to override some contexts.

hakman (approved):

Issue #16849, opened 2024-09-20
AWS sts:AssumeRole stopped working with role/OrganizationAccountAccessRole in 1.30.x


vitaliyf:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**

Testing upgrade from `Client version: 1.29.2 (git-v1.29.2)` to `Client version: 1.30.1 (git-v1.30.1)`

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**

`v1.29.9`

**3. What cloud provider are you using?**

AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**

`kops_v1.30.1 update cluster` - no other changes to manifest or environment, only executing newer kops binary.

**5. What happened after the commands executed?**

`$ export AWS_PROFILE=company-name-dev3`
`$ kops_v1.30.1 update cluster`

`SDK 2024/09/20 14:31:06 DEBUG request failed with unretryable error https response error StatusCode: 403, RequestID: 623bd87e-11e1-4b06-9f16-10f60ba2f030, api error AccessDenied: User: arn:aws:sts::[redacted]006:assumed-role/OrganizationAccountAccessRole/aws-go-sdk-1726842666098977639 is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::[redacted]006:role/OrganizationAccountAccessRole
Error: error determining default DNS zone: error querying zones: error listing hosted zones: operation error Route 53: ListHostedZones, get identity: get credentials: failed to refresh cached credentials, operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 623bd87e-11e1-4b06-9f16-10f60ba2f030, api error AccessDenied: User: arn:aws:sts::[redacted]006:assumed-role/OrganizationAccountAccessRole/aws-go-sdk-1726842666098977639 is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::[redacted]006:role/OrganizationAccountAccessRole`

**6. What did you expect to happen?**

With kops-1.29.2 the output shows proposed changes that need to be applied with `--yes`

AWS CLI is able to successfully get Route53 zones from the same shell:

```
$ aws route53 list-hosted-zones
{
    "HostedZones": [
        {
            "Id": "/hostedzone/Z0[redacted]",
            "Name": "k8s.dev3.us-west-2.example.com.",
[...]

PR #17161, opened 2025-01-02
Only configure STS region for Route 53 when we obtain it using IDMS

jValdron:
This PR fixes https://github.com/kubernetes/kops/issues/16849


rifelpet (approved):

aramhakobyan:
Still same issue is present, see https://github.com/kubernetes/kops/issues/17247

jValdron:
@aramhakobyan this wasn't backported to v1.30, so using kOps 1.31 should fix the issue

PR #17162, opened 2025-01-02
Install CNI network plugins only for specific CNIs

hakman:
/cc @ameukam @rifelpet @justinsb


pkg/nodemodel/wellknownassets/cni.go
```diff
@@ -34,14 +34,24 @@ import (
...
 const (
-	// CNI tarball for k8s >= 1.22
 	defaultCNIAssetAmd64K8s_22 = "https://storage.googleapis.com/k8s-artifacts-cni/release/v0.9.1/cni-plugins-linux-amd64-v0.9.1.tgz"
 	defaultCNIAssetArm64K8s_22 = "https://storage.googleapis.com/k8s-artifacts-cni/release/v0.9.1/cni-plugins-linux-arm64-v0.9.1.tgz"
 
-	// CNI tarball for k8s >= 1.27
 	defaultCNIAssetAmd64K8s_27 = "https://storage.googleapis.com/k8s-artifacts-cni/release/v1.2.0/cni-plugins-linux-amd64-v1.2.0.tgz"
 	defaultCNIAssetArm64K8s_27 = "https://storage.googleapis.com/k8s-artifacts-cni/release/v1.2.0/cni-plugins-linux-arm64-v1.2.0.tgz"
 
+	defaultCNIAssetAmd64K8s_29 = "https://storage.googleapis.com/k8s-artifacts-cni/release/v1.3.0/cni-plugins-linux-amd64-v1.3.0.tgz"
```
ameukam, line 43: We keep this is in WIP until we get some consensus about mirroring CNI to `artifacts.k8s.io`?

hakman: Doesn't matter much, as this is for kOps 1.32, which is early alpha. For sure we will add mirrors for this.

hakman: Also, please not that the top 3 use CNIs don't need those binaries, so even without mirrors, it should not be very demanding.

ameukam:
We don't want to generate assets for CNI plugins ? I don't see any change in `pkg/assets/assetdata/cni.yaml` or am I wrong ?

hakman:
> We don't want to generate assets for CNI plugins ? I don't see any change in `pkg/assets/assetdata/cni.yaml` or am I wrong ?

We want, but did not have time for that yet. 😁

pkg/nodemodel/wellknownassets/cni.go
hakman: Mirrors added 😄

hakman:
@rifelpet If you get a chance, could you take a look at this? Thanks!

nodeup/pkg/model/networking/common.go
```diff
@@ -30,8 +30,12 @@ var _ fi.NodeupModelBuilder = &CommonBuilder{}
 
 // Build is responsible for copying the common CNI binaries
 func (b *CommonBuilder) Build(c *fi.NodeupModelBuilderContext) error {
-	if err := b.AddCNIBinAssets(c); err != nil {
-		return err
+	if b.NodeupConfig.Networking.AmazonVPC == nil &&
+		b.NodeupConfig.Networking.Calico == nil &&
+		b.NodeupConfig.Networking.Cilium == nil {
```
rifelpet, line 35: can we reuse the new logic in pkg/apis/kops/model/instance_group.go ? that way we dont have to remember to update the same conditional in two places.

hakman: sure, done

rifelpet (approved):

pkg/assets/mirrors.go
```diff
@@ -40,6 +40,12 @@ var wellKnownMirrors = []mirrorConfig{
 			"https://github.com/kubernetes/kops/releases/download/v{kopsVersion}/",
 		},
 	},
+	{
+		Base: "https://storage.googleapis.com/k8s-artifacts-cni/release/",
```
BenTheElder, line 44: shouldn't we invert this?

long term we'd like to eliminate this host, we have been exiting making kubernetes responsible for hosting third party assets, especially on non-rate-limited hosts that we pay for (And it seems the main use case to not just fetch from github is bypassing their rate limits)

BenTheElder: ref: https://github.com/kubernetes/k8s.io/issues/7584

hakman: Agreed, the main use case is bypassing the rate limits.
We started discussing inverting this today, but no conclusion yet. The initial intention was to keep this consistent and decide later which ones to change to the mirror as primary.
@justinsb any preference on how to approach this?

BenTheElder: > Agreed, the main use case is bypassing the rate limits.

Yeah, the problem is this then makes us responsible for no rate limit, I think if users are pulling so much that github is blocking them, then they probably need to use their own mirror or something?

We have been rate limiting other hosts like registry.k8s.io, but that's not possible here without moving to a different host (at which point ... just use github / official downloads?)

PR #17180, opened 2025-01-07
fix the random order of block_device_mappings render

AldoFusterTurpin:
### What this PR does 
Include a new function to get the keys of the map used for block_device_mappings to access elements in deterministic order and avoid false changes in terraform diff.

### Why we need it ?
When we render the terraform in kops, sometimes we get a diff like this
image

stating that there are changes in the block_device_mappings, but the only thing that changes is the order of those devices, and this happens randomly because we use a map to store those [here](https://github.com/kubernetes/kops/blob/5ccbcd05607e641b8e9a988a970f5aced5a54d0e/upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go#L266).

### The solution
Simply sort the keys of the map and always access the maps using a deterministic order to avoid getting false changes shown in the terraform diff.

### Additional notes
I have used generics for simplicity and included a unit test for the new function. 

The only thing that changes is that now we do not iterate the maps directly using a "for" construct but instead iterate over the keys of the map in order and access the values of the maps using the keys. This change does not affect the behavior, just ensures that we always iterate over the devices in the same order.


upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
```diff
@@ -262,55 +265,42 @@ func (t *LaunchTemplate) RenderTerraform(target *terraform.TerraformTarget, a, e
...
-					VolumeType:          fi.PtrTo(string(x.EbsVolumeType)),
-				},
-			},
-		})
+
+	devicesKeys := getSortedKeysFromMap(devices)
+	for _, key := range devicesKeys {
+		v := devices[key]
+		terraformLaunchTemplateBlockDevice := createTerraformLaunchTemplateBlockDevice(key, v)
+		tf.BlockDeviceMappings = append(tf.BlockDeviceMappings, terraformLaunchTemplateBlockDevice)
```
hakman, line 278: This should not be needed starting Go 1.22:
https://go.dev/blog/loopvar-preview
```suggestion
		tf.BlockDeviceMappings = append(tf.BlockDeviceMappings, createTerraformLaunchTemplateBlockDevice(key, devices[key]))
```

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
```diff
@@ -262,55 +265,42 @@ func (t *LaunchTemplate) RenderTerraform(target *terraform.TerraformTarget, a, e
...
-					VolumeType:          fi.PtrTo(string(x.EbsVolumeType)),
-				},
-			},
-		})
+
+	additionalsKeys := getSortedKeysFromMap(additionals)
+	for _, key := range additionalsKeys {
+		v := additionals[key]
+		terraformLaunchTemplateBlockDevice := createTerraformLaunchTemplateBlockDevice(key, v)
+		tf.BlockDeviceMappings = append(tf.BlockDeviceMappings, terraformLaunchTemplateBlockDevice)
```
hakman, line 290: This should not be needed starting Go 1.22:
https://go.dev/blog/loopvar-preview
```suggestion
		tf.BlockDeviceMappings = append(tf.BlockDeviceMappings, createTerraformLaunchTemplateBlockDevice(key, additionals[key]))
```

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
```diff
@@ -262,55 +265,42 @@ func (t *LaunchTemplate) RenderTerraform(target *terraform.TerraformTarget, a, e
...
 
 	devices, err = buildEphemeralDevices(cloud, fi.ValueOf(e.InstanceType))
 	if err != nil {
 		return err
 	}
-	for n, x := range devices {
+
+	devicesKeys = getSortedKeysFromMap(devices)
+	for _, key := range devicesKeys {
+		v := devices[key]
```
hakman, line 300: This should not be needed starting Go 1.22:
https://go.dev/blog/loopvar-preview

hakman (changes requested):
Thanks for the PR @AldoFusterTurpin. There are a few nits that should be addressed before merging the PR.

AldoFusterTurpin:
> Thanks for the PR @AldoFusterTurpin. There are a few nits that should be addressed before merging the PR.

Thanks for the review! Checking now. 👌

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
AldoFusterTurpin: @hakman Yes, the temp vars are not needed, but I think the code is easier to read that way because 
```
tf.BlockDeviceMappings = append(tf.BlockDeviceMappings, createTerraformLaunchTemplateBlockDevice(key, devices[key]))
```
is a line too long IMHO
WDYT ?

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
```diff
@@ -262,55 +265,42 @@ func (t *LaunchTemplate) RenderTerraform(target *terraform.TerraformTarget, a, e
...
-					KmsKeyID:            x.EbsKmsKey,
-					IOPS:                x.EbsVolumeIops,
-					Throughput:          x.EbsVolumeThroughput,
-					VolumeSize:          x.EbsVolumeSize,
-					VolumeType:          fi.PtrTo(string(x.EbsVolumeType)),
-				},
-			},
-		})
+
+	devicesKeys := getSortedKeysFromMap(devices)
```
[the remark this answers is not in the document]
AldoFusterTurpin: True! Cool, thanks! Updating it now!

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
hakman: Sure, but at least inline the `v := devices[key]` part.

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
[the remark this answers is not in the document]
AldoFusterTurpin: Update:
I think the .Keys function is not available in go 1.22.7, the current Go version in go.mod.

[Docs for the current version](https://pkg.go.dev/maps@go1.22.7). Are you okay with keeping the current approach ?

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
AldoFusterTurpin: That sounds good 🙂 , doing so!

AldoFusterTurpin: Done!

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
AldoFusterTurpin: Done! (inlined v)

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
AldoFusterTurpin: Done! (Inlined v)

upup/pkg/fi/cloudup/awstasks/launchtemplate_target_terraform.go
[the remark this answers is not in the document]
AldoFusterTurpin: Maybe I misunderstood what you meant (using the STL maps package), but I think it is not available in our Go version.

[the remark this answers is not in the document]
hakman: I think we have something like this already in kOps code:
https://github.com/kubernetes/kops/blob/77256f84c3f6c8d4e774bd2bc4822e11e9961edc/util/pkg/maps/maps.go#L32-L37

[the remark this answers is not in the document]
AldoFusterTurpin: I thought you were talking about the STL. I missed that package. Great, thanks! 
Done.

hakman:
Thanks @AldoFusterTurpin!
/lgtm

rifelpet (approved):

PR #17181, opened 2025-01-07
Bump alpha channel with Dec 24 K8s releases

hakman:
/cc @rifelpet @moshevayner


hakman:
> /lgtm This is new! Do we now need to run `hack/generate-asset-hashes.sh` every time `alpha` or `stable` channels are updated? Thanks @hakman !

@moshevayner It's good as we don't need to pull hashes for those artifacts during running. It doesn't break anything.

PR #17182, opened 2025-01-07
make sure use correct endpoint for China cluster

neda1985:
My boyfriend was working with Kops and ran into an issue where the URI generator wasn’t behaving as expected in the China cluster. Turns out, the problem was with the endpoint (used for setting up IRSA) in China—it had an extra .cn at the end of the string. After digging into it, I found the fix! 😎
For reference: [AWS China Endpoints Documentation](https://docs.amazonaws.cn/en_us/aws/latest/userguide/endpoints-Beijing.html)

PR #17183, opened 2025-01-08
Use SDK's built-in resolver for S3Path.GetHTTPsUrl

rifelpet:
Alternative to https://github.com/kubernetes/kops/pull/17182

The aws-sdk-go-v2 includes an endpoint resolver that uses the region and partition information embedded in the SDK without making any API call to AWS.

This makes the VFS code more flexible, supporting all regions and partitions that the SDK recognizes.


nuved:
Hi, 
Do we have the changes for 1.28 or 1.29 as well?

PR #17187, opened 2025-01-08
Refactor validation logic around checking for multiple options

justinsb:
We also have to move some of the EnvVar handling around to avoid a circular reference.


justinsb:
WIP because I want to validate that this detects that the json names aren't yet right in v1alpha3 (also I need to rebase for kindnet)

hakman (approved):

PR #17190, opened 2025-01-08
Add IPv6 support for kindnet

hakman:
/cc @justinsb @rifelpet @aojea


aojea:
test failing are related to statefulset that can not be scheduled

```
I0108 21:02:10.663885 49753 dump.go:53] At 2025-01-08 20:52:09 +0000 UTC - event for datadir-ss-0: {ebs.csi.aws.com_ebs-csi-controller-65dbd65d98-v8sgz_184282f2-0eed-4f1d-b96a-af2e043c46a0 } ProvisioningFailed: failed to provision volume with StorageClass "kops-csi-1-21": rpc error: code = Internal desc = Could not create volume "pvc-1e7a5728-e634-490b-8187-78a1627d64e8": could not create volume in EC2: operation error EC2: CreateVolume, get identity: get credentials: failed to refresh cached credentials, failed to retrieve credentials, operation error STS: AssumeRoleWithWebIdentity, failed to get rate limit token, retry quota exceeded, 0 available, 5 requested

```

hakman:
Pretty odd, let's retry first 🙂 
/test pull-kops-e2e-cni-kindnet-ipv6

aojea:
still the same problem

> I0109 07:29:50.210015       1 controller.go:843] CreateVolume failed, supports topology = true, node selected true => may reschedule = true => state = Finished: rpc error: code = Internal desc = Could not create volume "pvc-3e9b3371-9520-4b73-bec0-3f626907832f": could not create volume in EC2: operation error EC2: CreateVolume, get identity: get credentials: failed to refresh cached credentials, failed to retrieve credentials, operation error STS: AssumeRoleWithWebIdentity, exceeded maximum number of attempts, 3, https response error StatusCode: 0, RequestID: , request send failed, Post "https://sts.us-west-2.amazonaws.com/": EOF

yeah, it may be to the masquerading

aojea:
the kindnet in the same node logs these messages

> I0109 07:26:04.498470       1 proxy.go:310] Failed to connect to original destination [[64:ff9b::3477:a3dd]:443]: dial tcp4 52.119.163.221:443: connect: network is unreachable

https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kops/17190/pull-kops-e2e-cni-kindnet-ipv6/1877249288690470912/artifacts/cluster-info/kube-system/kindnet-k89jm/kindnet-cni.log

that is an amazon ip https://urlscan.io/ip/52.94.181.70

in the docs https://kops.sigs.k8s.io/topology/#private-subnet

> If the subnet is capable of IPv6, egress to the internet is typically routed through a connection-tracking firewall, such as an AWS Egress-only Internet Gateway. Egress to the NAT64 range 64:ff9b::/96 is typically routed to a NAT64 device, such as an AWS NAT Gateway.

so maybe to access that Service you need to use the provided nat64 by aws and is not available from the instances?

pkg/model/components/kindnet.go
```diff
@@ -40,19 +40,27 @@ func (b *KindnetOptionsBuilder) BuildOptions(o *kops.Cluster) error {
 		c.Version = "v1.8.0"
 	}
 
-	// Kindnet should masquerade well known ranges if kops is not doing it
-	if c.Masquerade == nil {
-		c.Masquerade = &kops.KindnetMasqueradeSpec{
-			Enabled: fi.PtrTo(true),
+	if clusterSpec.IsIPv6Only() {
```
aojea, line 43: We need to disable NAT64 in Kindnet If is AWS and if is IPv6 only

aojea:
ok, talked offline with @hakman 

Kops has a NAT64 gateway setup in aws
The instance does not have IPv4 connectivity
Disabling Kindnet NAT64 on AWS with IPv6 only is one option, so the NAT64 gateway is used.

However, my feedback from users is that they want to avoid NAT gateways at all cost, they are expensive and causes problems, kindnet offers an alternative to this setup if the instance has also IPv4 connectivity by enabling NAT64

pkg/model/components/kindnet.go
aojea: ... or disable NAT64 gateway and allow IPv4 connectivity from the instance , I just throw the options, no strong opinion

aojea:
fantastic

> pull-kops-e2e-cni-kindnet-ipv6 — Job succeeded.

hakman:
Thanks for all the help getting IPv6 to work, @aojea! 🙂

rifelpet (approved):

Issue #17146, opened 2024-12-17
`kops reconcile cluster` fails if `kops export kubecfg` hasn't been run first


danports:
/kind bug

**1. What `kops` version are you running? The command `kops version`, will display
 this information.**
1.31.0-beta.1

**2. What Kubernetes version are you running? `kubectl version` will print the
 version if a cluster is running or provide the Kubernetes version specified as
 a `kops` flag.**
1.30.8

**3. What cloud provider are you using?**
AWS

**4. What commands did you run?  What is the simplest way to reproduce this issue?**
`kops reconcile cluster --yes`

**5. What happened after the commands executed?**
```
Updating control plane configuration
W1217 16:09:34.669456    2974 update_cluster.go:362] error checking control plane running version, assuming no k8s upgrade in progress: cannot load kubecfg settings for "my.cluster.com": context "my.cluster.com" does not exist
W1217 16:09:55.576232    2974 pruning.go:115] manifest includes an object of GroupKind Secret, which will not be pruned
I1217 16:09:55.669254    2974 issuerdiscovery.go:101] serviceAccountIssuers bucket "my-discovery-store" is not public; will use object ACL
I1217 16:09:59.567931    2974 executor.go:113] Tasks: 0 done / 135 total; 66 can run
I1217 16:10:00.038363    2974 executor.go:113] Tasks: 66 done / 135 total; 32 can run
I1217 16:10:00.413219    2974 default_methods.go:122] not deleting SecurityGroupRule/sg-xxx: port=-1 protocol=4 group=sg-xxx ip= ipv6= because it is marked for deferred-deletion
I1217 16:10:01.366491    2974 executor.go:113] Tasks: 98 done / 135 total; 25 can run
I1217 16:10:02.104563    2974 executor.go:113] Tasks: 123 done / 135 total; 3 can run
I1217 16:10:02.299848    2974 executor.go:113] Tasks: 126 done / 135 total; 6 can run
I1217 16:10:02.414822    2974 executor.go:113] Tasks: 132 done / 135 total; 3 can run
I1217 16:10:02.527061    2974 executor.go:113] Tasks: 135 done / 135 total; 0 can run
I1217 16:10:07.750745    2974 dns.go:235] Pre-creating DNS records
Cluster changes have been applied to the cloud.
Changes may require instances to restart: kops rolling-update cluster
Doing rolling-update for control plane
Error: cannot load kubecfg settings for "my.cluster.com": context "my.cluster.com" does not exist
```

[...]

PR #17191, opened 2025-01-09
Add docs for the new kops reconcile cluster command

rifelpet:
/hold for feedback

A few open questions:
1. Do we update all the `rolling-update cluster` docs references to `reconcile cluster`?
2. Do we return an error when a user tries to upgrade from k8s 1.30 to 1.31 using `update cluster --yes`? (with no `--instance-group*` filtering)
3. Do we add a new [permalink](https://github.com/kubernetes/kops/tree/master/permalinks) that the error message links to?
4. Do we mention the new `update cluster --reconcile` flag? When would a user use it instead of `kops reconcile cluster` ?

### DIFF
--- a/docs/tutorial/upgrading-kubernetes.md
+++ b/docs/tutorial/upgrading-kubernetes.md
@@ -4,7 +4,7 @@
 
 Kops' upgrade procedure has hostorically risked violating the [Kubelet version skew policy](https://kubernetes.io/releases/version-skew-policy/#kubelet). Between `kops update cluster --yes` and every kube-apiserver being rotated with `kops rolling-update cluster --yes`, newly launched nodes running new kubelet versions could be connecting to older `kube-apiserver` nodes.
 
-**Violating this policy when upgrading to Kubernetes 1.31 can cause newer kubelets to crash.**
+**Violating this policy when upgrading to Kubernetes 1.31 can cause newer kubelets to crash.** [This kubernetes issue](https://github.com/kubernetes/kubernetes/issues/127316) provides details though it was not addressed because the change does not actually violate the version skew policy, it merely breaks tooling that was already violating the policy.
 
 To upgrade a cluster to Kubernetes 1.31 or newer, use the new `kops reconcile cluster` command introduced in Kops 1.31. This replaces both `kops update cluster --yes` and `kops rolling-update cluster --yes`.
 
--- a/docs/tutorial/upgrading-kubernetes.md
+++ b/docs/tutorial/upgrading-kubernetes.md
@@ -1,15 +1,59 @@
 # Upgrading kubernetes
 
-Upgrading kubernetes is very easy with kOps, as long as you are using a compatible version of kOps.
-The kOps `1.18.x` series (for example) supports the kubernetes 1.16, 1.17 and 1.18 series,
-as per the kubernetes deprecation policy. Older versions of kubernetes will likely still work, but these
-are on a best-effort basis and will have little if any testing. kOps `1.18` will not support the kubernetes
-`1.19` series, and for full support of kubernetes `1.19` it is best to wait for the kOps `1.19` series release.
-We aim to release the next major version of kOps within a few weeks of the equivalent major release of kubernetes,
-so kOps `1.19.0` will be released within a few weeks of kubernetes `1.19.0`. We try to ensure that a 1.19 pre-release
-(alpha or beta) is available at the kubernetes release, for early adopters.
+## **NOTE for Kubernetes >1.31**
 
-Upgrading kubernetes is similar to changing the image on an InstanceGroup, except that the kubernetes version is
+Kops' upgrade procedure has hostorically risked violating the [Kubelet version skew policy](https://kubernetes.io/releases/version-skew-policy/#kubelet). Between `kops update cluster --yes` and every kube-apiserver being rotated with `kops rolling-update cluster --yes`, newly launched nodes running new kubelet versions could be connecting to older `kube-apiserver` nodes.
+
+**Violating this policy when upgrading to Kubernetes 1.31 can cause newer kubelets to crash.**
+
+To upgrade a cluster to Kubernetes 1.31 or newer, use the new `kops reconcile cluster` command introduced in Kops 1.31. This replaces both `kops update cluster --yes` and `kops rolling-update cluster --yes`.
+
+`kops reconcile cluster` will interleave the cloud provider updates of `kops update cluster --yes` with the node rotations of `kops rolling-update cluster --yes`.
+
+It is comparable to the following sequence:
+1. `kops update cluster --instance-group-roles=control-plane,apiserver --yes`
+2. `kops rolling-update cluster --instance-group-roles=control-plane,apiserver --yes`
+3. `kops update cluster --yes`
+4. `kops rolling-update cluster --yes`
+
+**Terraform** users will need to use a targeted terraform apply with the normal `kops rolling-update cluster --yes`:
+
+```sh
+$ kops update cluster --target terraform ...
+
+# Get the terraform resource IDs of the instance groups with a spec.role of `ControlPlane`, `Master`, or `APIServer`
+# The exact output may vary.
+$ terraform state list | grep -E 'aws_autoscaling_group|google_compute_instance_group_manager|hcloud_server|digitalocean_droplet|scaleway_instance_server'
+aws_autoscaling_group.controlplane-us-east-1a-example-com
+aws_autoscaling_group.controlplane-us-east-1b-example-com
+aws_autoscaling_group.controlplane-us-east-1c-example-com
+aws_autoscaling_group.nodes-example-com
+aws_autoscaling_group.bastion-example-com
+
+# Apply the changes to all control plane instance groups
+$ terraform apply -target 'aws_autoscaling_group.controlplane-us-east-1a-example-com' -target 'aws_autoscaling_group.controlplane-us-east-1b-example-com' -target 'aws_autoscaling_group.controlplane-us-east-1c-example-com'
+
+# Roll the apiserver nodes
+$ kops rolling-update cluster --yes --instance-group-roles control-plane,apiserver
+
+# Apply everything else
+$ terraform apply
+
+# Roll the remaining nodes
+$ kops rolling-update cluster --yes
+```
+
+## Upgrades before Kops 1.31
+
+Upgrading kubernetes is very easy with kOps, as long as you are using a compatible version of kOps.
+The kOps `1.30.x` series (for example) supports the kubernetes 1.28, 1.29, and 1.30 series,
+as per the kubernetes deprecation policy. Older versions of kubernetes will likely still work, but these
+are on a best-effort basis and will have little if any testing. kOps `1.30` will not support the kubernetes
+`1.31` series, and for full support of kubernetes `1.31` it is best to wait for the kOps `1.31` series release.
+We aim to release the next major version of kOps within a few weeks of the equivalent major release of kubernetes.
+We try to ensure that a pre-release (alpha or beta) is available at the kubernetes release date, for early adopters.
+
+Upgrading kubernetes is similar to changing the image on an InstanceGroup, the kubernetes version is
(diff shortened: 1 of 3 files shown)


stl-victor-sudakov:
> Do we return an error when a user tries to upgrade from k8s 1.30 to 1.31 using `update cluster --yes`? 

Why would this be an error?

rifelpet:
> > Do we return an error when a user tries to upgrade from k8s 1.30 to 1.31 using `update cluster --yes`?
> 
> Why would this be an error?

Because updating both the cluster's control plane launch templates (or other cloud provider equivalents) _and_ node launch templates at the same time will cause new nodes to fail to join the cluster until all control plane instances have been upgraded. So if Cluster Autoscaler or Karpenter scale up nodes before or during the control plane rolling-update, they will fail to join and workloads will be stuck in Pending. This is almost certainly not what the user wants and is why we're introducing the new command.

rifelpet:
We could allow the user to bypass the error if they know what they're doing. for example, on clusters that dont use Cluster Autoscaler or Karpenter.

stl-victor-sudakov:
> > > Do we return an error when a user tries to upgrade from k8s 1.30 to 1.31 using `update cluster --yes`?
> > 
> > 
> > Why would this be an error?
> 
> Because updating both the cluster's control plane launch templates (or other cloud provider equivalents) _and_ node launch templates at the same time will cause new nodes to fail to join the cluster until all control plane instances have been upgraded. So if Cluster Autoscaler or Karpenter scale up nodes before or during the control plane rolling-update, they will fail to join and workloads will be stuck in Pending. This is almost certainly not what the user wants and is why we're introducing the new command.

Hold on, hasn't this sequence:
```
kops upgrade cluster $NAME --yes
kops update cluster $NAME --yes
kops rolling-update cluster $NAME --yes
```
always been the standard upgrade sequence? These steps are even documented in https://kops.sigs.k8s.io/operations/updates_and_upgrades/#automated-update And now it is an error?

stl-victor-sudakov:
> Now it may cause node failures during the k8s 1.31 upgrade, yes. Hence the bold release note being added in this PR and my proposal to prevent users from making this mistake by returning a (skippable) error.
> 
> I'll update that docs page to note this change as well.

Sorry for my persistence, what has changed in k8s 1.31 that the regular kOps upgrade procedure has become dangerous?

rifelpet:
> Sorry for my persistence, what has changed in k8s 1.31 that the regular kOps upgrade procedure has become dangerous?

I updated this PR to link to the k/k issue that goes into more detail: https://github.com/kubernetes/kubernetes/issues/127316

stl-victor-sudakov:
> > Sorry for my persistence, what has changed in k8s 1.31 that the regular kOps upgrade procedure has become dangerous?
> 
> I updated this PR to link to the k/k issue that goes into more detail: [kubernetes/kubernetes#127316](https://github.com/kubernetes/kubernetes/issues/127316)

Oh, what a longread! Maybe https://github.com/kubernetes/kops/issues/16907 would be shorter and more to the point, it is also mentioned within the longer post. 

However I think I understand the innovation. The new `reconcile` command does "update --yes && rolling-update --yes" on CP nodes first, and then does the same on worker nodes, thus enforcing that the CP is fully updated first. Is this correct?

danports:
> The new reconcile command does "update --yes && rolling-update --yes" on CP nodes first, and then does the same on worker nodes, thus enforcing that the CP is fully updated first. Is this correct?

Yes, that's correct.

docs/tutorial/upgrading-kubernetes.md
```diff
@@ -1,15 +1,59 @@
 # Upgrading kubernetes
 
+## **NOTE for Kubernetes >1.31**
+
+Kops' upgrade procedure has hostorically risked violating the [Kubelet version skew policy](https://kubernetes.io/releases/version-skew-policy/#kubelet). Between `kops update cluster --yes` and every kube-apiserver being rotated with `kops rolling-update cluster --yes`, newly launched nodes running new kubelet versions could be connecting to older `kube-apiserver` nodes.
```
danports, line 5:
```suggestion
Kops' upgrade procedure has historically risked violating the [Kubelet version skew policy](https://kubernetes.io/releases/version-skew-policy/#kubelet). After `kops update cluster --yes` completes and before every kube-apiserver is replaced with `kops rolling-update cluster --yes`, newly launched nodes running newer kubelet versions could be connecting to older `kube-apiserver` nodes.
```

docs/tutorial/upgrading-kubernetes.md
```diff
@@ -1,15 +1,59 @@
...
+To upgrade a cluster to Kubernetes 1.31 or newer, use the new `kops reconcile cluster` command introduced in Kops 1.31. This replaces both `kops update cluster --yes` and `kops rolling-update cluster --yes`.
+
+`kops reconcile cluster` will interleave the cloud provider updates of `kops update cluster --yes` with the node rotations of `kops rolling-update cluster --yes`.
+
+It is comparable to the following sequence:
+1. `kops update cluster --instance-group-roles=control-plane,apiserver --yes`
+2. `kops rolling-update cluster --instance-group-roles=control-plane,apiserver --yes`
+3. `kops update cluster --yes`
+4. `kops rolling-update cluster --yes`
+
```
danports, line 18:
```suggestion
5. `kops update cluster --prune --yes`
```

docs/tutorial/upgrading-kubernetes.md
```diff
@@ -1,15 +1,59 @@
...
-We aim to release the next major version of kOps within a few weeks of the equivalent major release of kubernetes,
-so kOps `1.19.0` will be released within a few weeks of kubernetes `1.19.0`. We try to ensure that a 1.19 pre-release
-(alpha or beta) is available at the kubernetes release, for early adopters.
+are on a best-effort basis and will have little if any testing. kOps `1.30` will not support the kubernetes
+`1.31` series, and for full support of kubernetes `1.31` it is best to wait for the kOps `1.31` series release.
+We aim to release the next major version of kOps within a few weeks of the equivalent major release of kubernetes.
+We try to ensure that a pre-release (alpha or beta) is available at the kubernetes release date, for early adopters.
 
-Upgrading kubernetes is similar to changing the image on an InstanceGroup, except that the kubernetes version is
+Upgrading kubernetes is similar to changing the image on an InstanceGroup, the kubernetes version is
```
danports, line 56: I think the language here was clearer with "except that".

danports (changes requested):

danports:
> 1. Do we update all the `rolling-update cluster` docs references to `reconcile cluster`?

Yes, eventually, but I don't think it needs to happen right away, since `reconcile cluster` probably needs more time to mature and should support additional options that `update cluster`/`rolling-update cluster` already support. See #17146 for instance.

> 2. Do we return an error when a user tries to upgrade from k8s 1.30 to 1.31 using `update cluster --yes`? (with no `--instance-group*` filtering)

That would be a smart idea, though perhaps it should only be a warning if the cluster doesn't have CAS/Karpenter enabled.

> 3. Do we add a new [permalink](https://github.com/kubernetes/kops/tree/master/permalinks) that the error message links to?

👍 More context for error messages is always good.

> 4. Do we mention the new `update cluster --reconcile` flag? When would a user use it instead of `kops reconcile cluster` ?

I am confused about this myself. Based on the commit history I think maybe @justinsb just added the `--reconcile` flag as an initial step before adding the full `reconcile cluster` command, so maybe we don't need `--reconcile` anymore?

justinsb:
> I am confused about this myself. Based on the commit history I think maybe @justinsb just added the --reconcile flag as an initial step before adding the full reconcile cluster command, so maybe we don't need --reconcile anymore?

Good point, we can probably just get rid of this now as I don't think we've shipped it in any releases.  I don't _think_ there's any reason to keep it?  All `kops update --reconcile` does is "redirect" to the full `reconcile` command.

justinsb:
Summarizing what I think we've agreed in office hours discussions:

> Do we update all the rolling-update cluster docs references to reconcile cluster?

We think it's too early to "hide" rolling-update cluster, but we should maybe add pointers at key places suggesting that users try "reconcile cluster"

> Do we return an error when a user tries to upgrade from k8s 1.30 to 1.31 using update cluster --yes? (with no --instance-group* filtering)

We _think_ no, but we might add a warning.  The argument for "no" is we break users (and only users with rapidly autoscaling clusters are likely to be affected by the underlying problem).  The argument for "yes" is it might be better to break users and get them to consider whether they should switch to "reconcile".

I think we decided to push users to "reconcile" more in kOps 1.32, we're going to treat "reconcile" as a preview for now (kOps 1.31).

> Do we add a new [permalink](https://github.com/kubernetes/kops/tree/master/permalinks) that the error message links to?

I don't think we discussed in office hours, but sounds like a good idea in 1.32

> Do we mention the new update cluster --reconcile flag? When would a user use it instead of kops reconcile cluster ?

I don't think we discussed in office hours, but sounds like we should remove it.

stl-victor-sudakov:
If you remove the new `update cluster --reconcile` flag, then will the procedure to update a cluster always consist of two steps? First you run `update --yes`,  and then `reconcile` ? Would it not be convenient to have just one command to update and reconcile if necessary? On the other hand, the requirement to always run two steps may be more consistent.

stl-victor-sudakov:
> Right, kops reconcile cluster --yes is a one-stop command that can be used instead of kops update cluster and kops rolling-update cluster. You can still use those commands, but kops reconcile cluster makes it easy to obey the more complicated kubernetes upgrade

Hold on, if I change something by `kops edit cluster`, a security group or some other cloud property which does not require a rolling-update of nodes, should I use the old `update` command or the new `reconcile` command?

justinsb:
> Or, on the other hand, if reconcile already implies an update --yes step, then the the new update cluster --reconcile flag is really unnecessary. 

That's right, reconcile == update + rolling-update.  It's done by instance group to comply with the "update control plane first" rules.   (Technically it's done by instance group role). And yes, I think the `--reconcile` flag on update is unnecessary and I'll send a PR :-)

We always complied with the rules about "control plane first", just there is an edge case because `kops update` bumps the ASG versions (on AWS), but normally that new configuration is only rolled out by `kops rolling-update cluster`.  The edge case though is that if the node ASG auto-scales, it will come up with the new version, even if the control-plane has not yet updated.  That was tolerated previously, but kube 1.31 now makes it an error which causes problems in rapidly autoscaling clusters.

> (all the above is being said from the point of view of a (slightly) confused kOps user).

No worries, we're all trying to figure out what makes the most sense and how best to explain it :-)

justinsb:
I think this PR looks pretty good.  I agree that "except that" was probably clearer, but I propose we merge this PR and iterate on the docs.

docs/tutorial/upgrading-kubernetes.md
```diff
@@ -1,15 +1,59 @@
...
+To upgrade a cluster to Kubernetes 1.31 or newer, use the new `kops reconcile cluster` command introduced in Kops 1.31. This replaces both `kops update cluster --yes` and `kops rolling-update cluster --yes`.
+
+`kops reconcile cluster` will interleave the cloud provider updates of `kops update cluster --yes` with the node rotations of `kops rolling-update cluster --yes`.
+
+It is comparable to the following sequence:
+1. `kops update cluster --instance-group-roles=control-plane,apiserver --yes`
+2. `kops rolling-update cluster --instance-group-roles=control-plane,apiserver --yes`
+3. `kops update cluster --yes`
+4. `kops rolling-update cluster --yes`
+5. `kops update cluster --prune --yes`
```
danports, line 18:
```suggestion
5. `kops update cluster --prune --yes`

```
My bad, we need an extra line break in here to fix the list formatting.

justinsb:
We agreed in office hours to merge and polish in follow on PRs (because the GitHub UI isn't the best for iterating on docs PRs!)

/approve
/lgtm

stl-victor-sudakov:
I have run "kops reconcile cluster" several times, works well, but during the reconcile process kops outputs many information messages about the necessity to run "kops update" or "kops rolling-update". I think they are irrelevant and even misleading now? Should I open a new issue, or we just ignore them?

PR #17220, opened 2025-01-20
Remove reconcile flag from `kops update`

justinsb:
We have `kops reconcile`, and it's confusing having both.

We didn't ship the --reconcile flag in any released version.


rifelpet (approved):
/override pull-kops-e2e-k8s-aws-amazonvpc-u2404

I mistakenly made this new presubmit required instead of optional, fixed in https://github.com/kubernetes/test-infra/pull/34184

